PR #282, opened 2017-07-24
Zipkin api/v1/spans endpoint

### LINKED ISSUE #248: Define zipkin compatibility
I would like to more formally define Zipkin compatibility. What is currently supported and what will be supported in the future.

Currently we accept zipkin spans in thrift format over HTTP. Default endpoint is `http://localhost:14268/api/traces?format=zipkin.thrift`.

Default Zipkin endpoint is `http://localhost:9411/api/v1/spans` with `Content-Type` `thrift` or `json` and optional `gzip` encoding.

The difference makes it hard to send data from zipkin instrumentations to jaeger server. In some cases it requires changes in zipkin tracers because of the hardcoded path (`/api/v1/spans`) and often it's not possible to configure sender with a custom query param (`format=zipkin.thrift`).

We should consider supporting the same endpoint on the jaeger side to make migrations easier.

pavolloffay:
Fixes #248

### DIFF
--- a/cmd/collector/app/http_handler.go
+++ b/cmd/collector/app/http_handler.go
@@ -35,7 +35,8 @@ import (
 )
 
 const (
-	formatParam               = "format"
+	formatParam = "format"
+	// UnableToReadBodyErrFormat is an error message for invalid requests
 	UnableToReadBodyErrFormat = "Unable to process request body: %v"
 )
 
@@ -89,6 +90,5 @@ func (aH *APIHandler) saveSpan(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	w.WriteHeader(http.StatusOK)
+	w.WriteHeader(http.StatusAccepted)
 }
-
--- a/cmd/collector/app/http_handler.go
+++ b/cmd/collector/app/http_handler.go
@@ -21,6 +21,7 @@
 package app
 
 import (
+	"compress/gzip"
 	"fmt"
 	"io/ioutil"
 	"net/http"
@@ -62,6 +63,34 @@ func (aH *APIHandler) RegisterRoutes(router *mux.Router) {
 	router.HandleFunc("/api/traces", aH.saveSpan).Methods(http.MethodPost)
 }
 
+func (aH *APIHandler) RegisterZipkinRoutes(router *mux.Router) {
+	router.HandleFunc("/api/v1/spans", aH.zipkinHandler).Methods(http.MethodPost)
+}
+
+func (aH *APIHandler) zipkinHandler(w http.ResponseWriter, r *http.Request) {
+	bRead := r.Body
+	if strings.Contains(r.Header.Get("Content-Encoding"), "gzip") {
+		result, err := gzip.NewReader(r.Body)
+		if err != nil {
+			http.Error(w, "Wrong encoding", http.StatusBadRequest)
+		}
+		bRead = result
+	}
+
+	bodyBytes, err := ioutil.ReadAll(bRead)
+	if err != nil {
+		http.Error(w, "Could not read request body", http.StatusBadRequest)
+	}
+
+	if r.Header.Get("Content-Type") == "application/x-thrift" {
+		handleZipkinThrift(aH.zipkinSpansHandler, bodyBytes, w)
+	} else {
+		http.Error(w, "Only Content-Type:application/x-thrift is supported at the moment", http.StatusBadRequest)
+	}
+
+	w.WriteHeader(http.StatusAccepted)
+}
+
 func (aH *APIHandler) saveSpan(w http.ResponseWriter, r *http.Request) {
 	bodyBytes, err := ioutil.ReadAll(r.Body)
 	r.Body.Close()
@@ -89,24 +118,27 @@ func (aH *APIHandler) saveSpan(w http.ResponseWriter, r *http.Request) {
 		}
 
 	case "zipkin.thrift":
-		spans, err := deserializeZipkin(bodyBytes)
-		if err != nil {
-			http.Error(w, fmt.Sprintf(unableToReadBodyErrFormat, err), http.StatusBadRequest)
-			return
-		}
-
-		ctx, _ := tchanThrift.NewContext(time.Minute)
-		if _, err = aH.zipkinSpansHandler.SubmitZipkinBatch(ctx, spans); err != nil {
-			http.Error(w, fmt.Sprintf("Cannot submit Zipkin batch: %v", err), http.StatusInternalServerError)
-			return
-		}
-
+		handleZipkinThrift(aH.zipkinSpansHandler, bodyBytes, w)
 	default:
 		http.Error(w, fmt.Sprintf("Unsupported format type: %v", format), http.StatusBadRequest)
 		return
 	}
 
 	w.WriteHeader(http.StatusOK)
+}
+
+func handleZipkinThrift(zHandler ZipkinSpansHandler, bodyBytes []byte, w http.ResponseWriter) {
+	spans, err := deserializeZipkin(bodyBytes)
+	if err != nil {
+		http.Error(w, fmt.Sprintf(unableToReadBodyErrFormat, err), http.StatusBadRequest)
+		return
+	}
+
+	ctx, _ := tchanThrift.NewContext(time.Minute)
+	if _, err = zHandler.SubmitZipkinBatch(ctx, spans); err != nil {
+		http.Error(w, fmt.Sprintf("Cannot submit Zipkin batch: %v", err), http.StatusInternalServerError)
+		return
+	}
 }
 
 func deserializeZipkin(b []byte) ([]*zipkincore.Span, error) {
--- a/cmd/collector/app/http_handler.go
+++ b/cmd/collector/app/http_handler.go
@@ -21,7 +21,6 @@
 package app
 
 import (
-	"compress/gzip"
 	"fmt"
 	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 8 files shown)


yurishkuro:
I think a bit of refactoring is called for. Can we move all of zipkin-related logic to `app/zipkin` package? `zipkin.APIHandler` will refer just to Zipkin functionality.

pavolloffay:
I can give it a try. Do you agree with other changes in this PR? e.g. new flag for zipkin port. Or do we want a flag to enable zipkin port?

yurishkuro:
> Do you agree with other changes in this PR? e.g. new flag for zipkin port. Or do we want a flag to enable zipkin port?

new flag for a port is fine. The main question is: do we want this on by default? If not,  we may want to use 0 as the default for it and treat it as "don't enable zipkin port". Or have another flag `collector.zipkin.enabled`.

pavolloffay:
> I think a bit of refactoring is called for. Can we move all of zipkin-related logic to app/zipkin package? zipkin.APIHandler will refer just to Zipkin functionality. 

see https://github.com/uber/jaeger/pull/282/commits/4752a3c2563c4a5a1389b7ad0e97fe0743dc948f

We could also move zipkin stuff from `collector/app/span_handler.go` to `app/zipkin/span_handler.go`

cmd/collector/app/zipkin/http_handler.go
```diff
@@ -0,0 +1,111 @@
...
+		zipkinSpansHandler: zipkinSpansHandler,
+	}
+}
+
+func (aH *APIHandler) RegisterRoutes(router *mux.Router) {
+	router.HandleFunc("/api/v1/spans", aH.zipkinHandler).Methods(http.MethodPost)
+}
+
+func (aH *APIHandler) zipkinHandler(w http.ResponseWriter, r *http.Request) {
+	bRead := r.Body
```
black-adder: defer close the request body to avoid memory leaks

cmd/collector/app/zipkin/http_handler.go
```diff
@@ -0,0 +1,111 @@
...
+	"strconv"
+	"time"
+
+	"go.uber.org/zap"
+	"github.com/apache/thrift/lib/go/thrift"
+	"github.com/gorilla/mux"
+
+	"github.com/uber/jaeger/cmd/collector/app"
+	zipkincore "github.com/uber/jaeger/thrift-gen/zipkincore"
+	tchanThrift "github.com/uber/tchannel-go/thrift"
```
black-adder: to the second import group

cmd/collector/app/http_handler.go
```diff
@@ -109,26 +92,3 @@ func (aH *APIHandler) saveSpan(w http.ResponseWriter, r *http.Request) {
 	w.WriteHeader(http.StatusOK)
```
black-adder: I actually prefer returning 202 like you've done in the zipkin http handler, mind making the change here too?

cmd/collector/app/zipkin/http_handler.go
```diff
@@ -0,0 +1,111 @@
...
+		if err = zs.Read(transport); err != nil {
+			return nil, err
+		}
+		spans = append(spans, zs)
+	}
+
+	return spans, nil
+}
+
+func StartHttpAPI(logger *zap.Logger, zipkinSpansHandler app.ZipkinSpansHandler, recoveryHandler func (http.Handler) http.Handler) {
```
black-adder: given that this is going to have to be covered, it might be better to just bite the bullet and inline it in the mains

cmd/collector/main.go
```diff
@@ -83,10 +84,15 @@ func main() {
...
 	r := mux.NewRouter()
-	apiHandler := app.NewAPIHandler(jaegerBatchesHandler, zipkinSpansHandler)
-	apiHandler.RegisterRoutes(r)
+	app.NewAPIHandler(jaegerBatchesHandler).RegisterRoutes(r)
 	httpPortStr := ":" + strconv.Itoa(*builder.CollectorHTTPPort)
 	recoveryHandler := recoveryhandler.NewRecoveryHandler(logger, true)
+
+	go zipkin.StartHttpAPI(logger, zipkinSpansHandler, recoveryHandler)
+
+	zipkinHandler := zipkin.NewAPIHandler(zipkinSpansHandler)
```
black-adder: hmmm, I'm not seeing much point of this. Given that we've swapped zipkin under a new endpoint, doesn't make sense for this to run on the old port.

cmd/collector/app/http_handler.go
pavolloffay: No objections. We should use the same for all endpoints.

cmd/collector/main.go
pavolloffay: agree https://github.com/uber/jaeger/issues/239#issuecomment-317865183

pavolloffay:
Rebased and improved tests. Could somebody re-review?

cmd/collector/app/zipkin/http_handler_test.go
```diff
@@ -0,0 +1,283 @@
...
+}
+
+func initializeTestServer(err error) (*httptest.Server, *APIHandler) {
+	r := mux.NewRouter()
+	handler := NewAPIHandler(&mockZipkinHandler{err: err})
+	handler.RegisterRoutes(r)
+	return httptest.NewServer(r), handler
+}
+
+func TestFormatsViaClient(t *testing.T) {
```
black-adder: this test name doesn't have relevance anymore

cmd/collector/app/zipkin/http_handler_test.go
```diff
@@ -0,0 +1,283 @@
...
+	header.Add("Content-Encoding", "gzip")
+
+	statusCode, resBodyStr, err := postBytes(server.URL+`/api/v1/spans`, bodyBytes, header)
+	assert.NoError(t, err)
+	assert.EqualValues(t, http.StatusBadRequest, statusCode)
+	assert.EqualValues(t, "Unable to process request body: unexpected EOF\n", resBodyStr)
+}
+
+func TestUnsupportedContentType(t *testing.T) {
+	span := &zipkincore.Span{}
```
black-adder: these spans not used in this test

cmd/collector/app/zipkin/http_handler_test.go
```diff
@@ -0,0 +1,283 @@
...
+	handler.zipkinSpansHandler.(*mockZipkinHandler).err = fmt.Errorf("Bad times ahead")
+	statusCode, resBodyStr, err = postBytes(server.URL+`/api/v1/spans`, bodyBytes, createHeader("application/x-thrift"))
+	assert.NoError(t, err)
+	assert.EqualValues(t, http.StatusInternalServerError, statusCode)
+	assert.EqualValues(t, "Cannot submit Zipkin batch: Bad times ahead\n", resBodyStr)
+}
+
+func TestGzipEncoding(t *testing.T) {
+	span := &zipkincore.Span{}
+	spans := []*zipkincore.Span{}
```
black-adder: ditto as above

cmd/collector/app/zipkin/http_handler_test.go
```diff
@@ -0,0 +1,283 @@
...
+		}
+		break
+	}
+
+	assert.Equal(t, 1, len(handler.zipkinSpansHandler.(*mockZipkinHandler).getSpans()))
+}
+
+func TestThriftFormat(t *testing.T) {
+	span := &zipkincore.Span{}
+	spans := []*zipkincore.Span{}
```
black-adder: given that in this test you don't actually use span, you can inline it like so:

`spans := []*zipkincore.Span{{}}`

cmd/collector/app/zipkin/http_handler_test.go
```diff
@@ -0,0 +1,283 @@
...
+	bodyBytes := []byte("not good")
+	statusCode, resBodyStr, err := postBytes(server.URL+`/api/v1/spans`, bodyBytes, createHeader("application/x-thrift"))
+	assert.NoError(t, err)
+	assert.EqualValues(t, http.StatusBadRequest, statusCode)
+	assert.EqualValues(t, "Unable to process request body: *zipkincore.Span field 0 read error: EOF\n", resBodyStr)
+}
+
+func TestDeserializeZipkinWithBadListStart(t *testing.T) {
+	span := &zipkincore.Span{TraceID: 12, Name: "test"}
+	spans := []*zipkincore.Span{}
```
black-adder: can inline intiailization like above

cmd/collector/app/zipkin/http_handler_test.go
```diff
@@ -0,0 +1,283 @@
...
+	z.Close()
+	return buffer.Bytes()
+}
+
+func postBytes(urlStr string, bytesBody []byte, header *http.Header) (int, string, error) {
+	req, err := http.NewRequest(http.MethodPost, urlStr, bytes.NewBuffer(bytesBody))
+	if err != nil {
+		return 0, "", err
+	}
+	if strings.Contains(http.DetectContentType(bytesBody), "gzip") {
```
black-adder: why not explicitly pass this header in when you're testing gzip? Feels weird(?) that the test util is adding the header for you after detecting it's gzip.

cmd/collector/app/zipkin/http_handler_test.go
pavolloffay: changed to `TestViaClient`

cmd/collector/app/zipkin/http_handler_test.go
pavolloffay: you are right, it's better to don't do any fanciness here.

pavolloffay:
@black-adder thanks, all comments addressed.

pavolloffay:
ES test failed, It should not be related to this PR

PR #284, opened 2017-07-25
Add ES Curator script

mh-park:
Cleans up old indices

### DIFF
--- a/scripts/esCleaner.py
+++ b/scripts/esCleaner.py
@@ -1,0 +1,31 @@
+import elasticsearch
+import curator
+import sys
+
+
+def main():
+    if len(sys.argv) == 1:
+        print('USAGE: %s HOSTNAME[:PORT] ...' % sys.argv[0])
+        sys.exit(1)
+
+    client = elasticsearch.Elasticsearch(sys.argv[1:])
+
+    ilo = curator.IndexList(client)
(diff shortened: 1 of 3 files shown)


scripts/esCleaner.py
```diff
@@ -0,0 +1,31 @@
...
+    if len(sys.argv) == 1:
+        print('USAGE: %s HOSTNAME[:PORT] ...' % sys.argv[0])
+        sys.exit(1)
+
+    client = elasticsearch.Elasticsearch(sys.argv[1:])
+
+    ilo = curator.IndexList(client)
+    empty_list(ilo, 'ElasticSearch has no indices')
+
+    ilo.filter_by_age(source='name', direction='older', timestring='%Y-%m-%d', unit='days', unit_count=4)
```
black-adder: unit_count should be configurable. What if the company using this script has other indexes in elastic search? we don't want to delete those.

scripts/esCleaner.py
```diff
@@ -0,0 +1,31 @@
+import elasticsearch
+import curator
+import sys
+
+
+def main():
```
black-adder: There should be a bash script that tells users how to use this, does pip install, etc and runs this script. The cron job can then be configured to run the script. Check out https://github.com/uber/jaeger/blob/master/plugin/storage/cassandra/schema/create.sh

PR #288, opened 2017-07-26
Add README for ElasticSearch plugin

mh-park:


plugin/storage/es/README.md
```diff
@@ -0,0 +1,49 @@
...
+This provides a plugin to use Jaeger with [ElasticSearch](https://www.elastic.co). This currently supports ElasticSearch 5.x.
+
+## Indices
+Indices will be created depending on the spans timestamp. i.e., a span with
+a timestamp on 2017/04/21 will be stored in an index named `jaeger-2017-04-21`.
+ElasticSearch also has no support for TTL, so there exists a script `./es_indices_clean.sh`
+that deletes older indices automatically. The [Elastic Curator](https://www.elastic.co/guide/en/elasticsearch/client/curator/current/about.html)
+can also be used instead to do a similar job.
+
+#### Using `./es_indices_clean.sh`
```
yurishkuro: did you mean three `###`?

plugin/storage/es/README.md
```diff
@@ -0,0 +1,49 @@
...
+Parameters: 
+ * a number that will delete any indices older than that number in days
+ * ElasticSearch hostnames
```
yurishkuro: show example

plugin/storage/es/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+### Timestamps
+Because ElasticSearch's `Date` datatype has only millisecond granularity and Jaeger
+requires microsecond granularity, Jaeger spans' `StartTime` is saved as a long type.
+The conversion is done automatically.
+
+### Separation of Spans and Service:Operation Pairs
+The current commit has `span` and `service:operation` documents under the same index for a given date.
+This is to be separated into two indices in the near future in preparation for ElasticSearch v6.0.
```
yurishkuro: Please open an issue and link it here. In the issue describe what problems the existing implementation causes, and which feature of 6.x helps to address them.

plugin/storage/es/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+### Nested fields (tags)
```
yurishkuro: seems like this should go under Limitations, with the title describing the limit, rather than implementation.

plugin/storage/es/README.md
```diff
@@ -0,0 +1,49 @@
...
+`Tags` are [nested](https://www.elastic.co/guide/en/elasticsearch/reference/current/nested.html) fields in the 
+ElasticSearch schema used for Jaeger. This allows for better search capabilities and data retention. However, because
+ElasticSearch creates a new document for every nested field, there is currently a limit of 50 nested fields per document.
+
+## Limitations
+
+### Tag query over multiple spans
+Because each span is its own document in ElasticSearch, if a query has multiple tags that exist in different spans in the
```
yurishkuro: maybe rephrase without the negative at the end?

plugin/storage/es/README.md
```diff
@@ -0,0 +1,49 @@
...
+### Tag query over multiple spans
+Because each span is its own document in ElasticSearch, if a query has multiple tags that exist in different spans in the 
+same trace, the trace will not be retrieved.
+
+### Case-sensitivity
+Queries are case-sensitive. For example, if a document with service name `ABC` is searched using a query `abc`, 
+the document will not be retrieved.
+
+## Testing
+To locally test the ElasticSearch storage plugin, run `make es-integration-test` in the top folder.
```
yurishkuro: This looks a bit odd, how can the same test be used with Cassandra? I thought we'd need something like this instead: `ES_HOST_PORT=.... make integration-test` or `CASSANDRA_HOST_PORT=.... make integration-test`

yurishkuro:
Another thing - have a look at the top-level README and `/docs`, they should mention that ES is also available now.

plugin/storage/es/README.md
mh-park: did you mean this for the separation of span and service:operation documents?

plugin/storage/es/README.md
mh-park: There are some plugin-specific code that needs to be written for each, i.e. the es.Client for ElasticSearch and cassandra.Session for C*. So there's another file that exists: `/plugin/storage/integration/es_integration_test.go`. We'll have to create another file if we want to do an integration test on C*.

black-adder:
let's hold off on the top-level README until we've added viper flags for this and we've fully tested it out internally.

plugin/storage/es/README.md
yurishkuro: I understand, but the make target currently doesn't invoke any file specifically, it just runs all tests under integration/...

The approach with env variables works by having the specific backend tests call t.Skip() if the env variable they need is not set. That way `make integration-test` could be anything from no-op to testing all backends at once, if the databases happened to be running somewhere.

PR #290, opened 2017-07-27
Fix bug where test fails due to date error

black-adder:
There's a mock that expects exactly 4 indices however this test breaks during UTC [0:1) hours because during that time window we generate 5 indices. Mocking time.Now to always use a specific date for tests.

### DIFF
--- a/plugin/storage/es/spanstore/reader.go
+++ b/plugin/storage/es/spanstore/reader.go
@@ -91,14 +91,16 @@ type SpanReader struct {
 	// this will be rounded down to UTC 00:00 of that day.
 	maxLookback             time.Duration
 	serviceOperationStorage *ServiceOperationStorage
+
+	timeNow func() time.Time
 }
 
 // NewSpanReader returns a new SpanReader with a metrics.
 func NewSpanReader(client es.Client, logger *zap.Logger, maxLookback time.Duration, metricsFactory metrics.Factory) spanstore.Reader {
-	return storageMetrics.NewReadMetricsDecorator(newSpanReader(client, logger, maxLookback), metricsFactory)
+	return storageMetrics.NewReadMetricsDecorator(newSpanReader(client, logger, maxLookback, time.Now), metricsFactory)
 }
 
-func newSpanReader(client es.Client, logger *zap.Logger, maxLookback time.Duration) *SpanReader {
+func newSpanReader(client es.Client, logger *zap.Logger, maxLookback time.Duration, timeNow func() time.Time) *SpanReader {
 	ctx := context.Background()
 	return &SpanReader{
 		ctx:                     ctx,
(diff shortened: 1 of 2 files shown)


plugin/storage/es/spanstore/reader.go
```diff
@@ -124,6 +127,8 @@ func (s *SpanReader) GetTrace(traceID model.TraceID) (*model.Trace, error) {
 func (s *SpanReader) readTrace(traceID string, traceQuery *spanstore.TraceQueryParameters) (*model.Trace, error) {
```
yurishkuro: not related to this diff, but needs to be fixed. Why does this function take TraceQueryParameters? It has nothing to do with what that type actually represents. Should simply pass upper and lower time boundary

plugin/storage/es/spanstore/reader.go
```diff
@@ -124,6 +127,8 @@ func (s *SpanReader) GetTrace(traceID model.TraceID) (*model.Trace, error) {
...
 	query := elastic.NewTermQuery(traceIDField, traceID)
 
+	// NB. Add an hour in both directions so that traces that straddle two indexes are retrieved.
+	// ie starts in one and ends in another.
```
yurishkuro: why "hour"? And why did we hardcode the granularity of index rotation to 24hrs in L196? It could've been configurable.

plugin/storage/es/spanstore/reader.go
```diff
@@ -91,27 +91,30 @@ type SpanReader struct {
...
 	maxLookback             time.Duration
 	serviceOperationStorage *ServiceOperationStorage
+
+	timeNow func() time.Time
 }
 
 // NewSpanReader returns a new SpanReader with a metrics.
 func NewSpanReader(client es.Client, logger *zap.Logger, maxLookback time.Duration, metricsFactory metrics.Factory) spanstore.Reader {
-	return storageMetrics.NewReadMetricsDecorator(newSpanReader(client, logger, maxLookback), metricsFactory)
+	return storageMetrics.NewReadMetricsDecorator(newSpanReader(client, logger, maxLookback, time.Now), metricsFactory)
```
yurishkuro: please break to multiple lines, at least for the top-level function (newSpanReader can stay on one line)

plugin/storage/es/spanstore/reader.go
```diff
@@ -91,27 +91,30 @@ type SpanReader struct {
...
 }
 
-func newSpanReader(client es.Client, logger *zap.Logger, maxLookback time.Duration) *SpanReader {
+func newSpanReader(client es.Client, logger *zap.Logger, maxLookback time.Duration, timeNow func() time.Time) *SpanReader {
```
yurishkuro: break signature to multiple lines, it's too long (124 chars by the looks of it)

plugin/storage/es/spanstore/reader.go
mh-park: if the granularity of index rotation isn't 24 hrs, the naming convention for indices will have to change

mh-park:
Resolved in #301

PR #293, opened 2017-07-27
Add exists check to ES writer

mh-park:
Without the exists check, if a new collector is brought up against the same ElasticSearch host, the writer would fail because the index would already exist and index creation would fail.


plugin/storage/es/spanstore/writer.go
```diff
@@ -107,16 +107,19 @@ func (s *SpanWriter) WriteSpan(span *model.Span) error {
...
 }
 
 func (s *SpanWriter) createIndex(indexName string, jsonSpan *jModel.Span) error {
 	if !keyInCache(indexName, s.indexCache) {
 		start := time.Now()
-		_, err := s.client.CreateIndex(indexName).Body(spanMapping).Do(s.ctx)
-		s.writerMetrics.indexCreate.Emit(err, time.Since(start))
-		if err != nil {
-			return s.logError(jsonSpan, err, "Failed to create index", s.logger)
+		exists, _ := s.client.IndexExists(indexName).Do(s.ctx) // don't need to check the error because the exists variable will be false anyway if there is an error
```
black-adder: add a comment around here detailing that we might drop a couple of spans every time a new index is created.

PR #298, opened 2017-07-31
Adjust ES indices size to be configurable

mh-park:


pkg/es/config/config.go
```diff
@@ -36,6 +36,7 @@ type Configuration struct {
 	Password   string
 	Sniffer    bool          // https://github.com/olivere/elastic/wiki/Sniffing
 	MaxSpanAge time.Duration // configures the maximum lookback on span reads
+	Timeframe  time.Duration // configures how large in duration each index is
```
yurishkuro: please choose a descriptive & narrow name

plugin/storage/es/spanstore/reader.go
```diff
@@ -90,21 +91,40 @@ type SpanReader struct {
...
 
-func newSpanReader(client es.Client, logger *zap.Logger, maxLookback time.Duration) *SpanReader {
+func newSpanReader(
+	client es.Client,
+	logger *zap.Logger,
+	maxLookback time.Duration,
+	timeframe time.Duration,
+) *SpanReader {
 	ctx := context.Background()
+	if timeframe == time.Duration(0) {
```
yurishkuro: Constants are auto-typed, don't need to cast it

plugin/storage/es/spanstore/reader.go
```diff
@@ -183,34 +203,34 @@ func (s *SpanReader) unmarshalJSONSpan(esSpanRaw *elastic.SearchHit) (*jModel.Sp
 }
 
 // Returns the array of indices that we need to query, based on query params
-func findIndices(startTime time.Time, endTime time.Time) []string {
+func findIndices(startTime time.Time, endTime time.Time, timeframe time.Duration) []string {
 	var indices []string
-	firstIndex := IndexWithDate(startTime)
-	currentIndex := IndexWithDate(endTime)
+	firstIndex := indexWithDate(startTime, timeframe)
+	currentIndex := indexWithDate(endTime, timeframe)
```
yurishkuro: why not make these indices member functions of the struct? Then you won't need to pass extra parameter which doesn't change anyway

plugin/storage/es/spanstore/reader_test.go
```diff
@@ -261,37 +261,37 @@ func TestSpanReaderFindIndices(t *testing.T) {
...
 	}
 	for _, testCase := range testCases {
-		actual := findIndices(testCase.startTime, testCase.endTime)
+		actual := findIndices(testCase.startTime, testCase.endTime, time.Hour*24)
 		assert.EqualValues(t, testCase.expected, actual)
 	}
 }
 
 func TestSpanReader_indexWithDate(t *testing.T) {
 	withSpanReader(func(r *spanReaderTest) {
```
yurishkuro: what's the point of wrapping this in withSpanReader?

PR #299, opened 2017-08-01
Modify ES Reader to use MultiSearch for bulk reading

mh-park:


plugin/storage/es/spanstore/reader.go
```diff
@@ -238,19 +238,40 @@ func (s *SpanReader) FindTraces(traceQuery *spanstore.TraceQueryParameters) ([]*
...
+	return s.multiRead(uniqueTraceIDs, traceQuery)
+}
+
+func (s *SpanReader) multiRead(traceIDs []string, traceQuery *spanstore.TraceQueryParameters) ([]*model.Trace, error) {
+	if len(traceIDs) == 0 {
+		return []*model.Trace{}, nil
+	}
+	searchRequests := make([]*elastic.SearchRequest, len(traceIDs))
+
+	for i := range searchRequests {
```
black-adder: I prefer: `for i, traceID := range traceIDs` so you don't have to do index on L251

plugin/storage/es/spanstore/reader.go
```diff
@@ -238,19 +238,40 @@ func (s *SpanReader) FindTraces(traceQuery *spanstore.TraceQueryParameters) ([]*
...
+		query := elastic.NewTermQuery("traceID", traceIDs[i])
+		searchRequests[i] = elastic.NewSearchRequest().IgnoreUnavailable(true).Type("span").
+			Source(elastic.NewSearchSource().Query(query).Size(10000))
```
black-adder: defaultDocCount

plugin/storage/es/spanstore/reader.go
```diff
@@ -238,19 +238,40 @@ func (s *SpanReader) FindTraces(traceQuery *spanstore.TraceQueryParameters) ([]*
...
+		Add(searchRequests...).
+		Index(indices...).
+		Do(s.ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	var traces []*model.Trace
+	for _, result := range results.Responses {
+		spans, err := s.collectSpans(result.Hits.Hits)
```
black-adder: could result.Hits ever be nil? we could get a NPE

plugin/storage/es/spanstore/reader.go
```diff
@@ -238,19 +238,40 @@ func (s *SpanReader) FindTraces(traceQuery *spanstore.TraceQueryParameters) ([]*
...
-	var retMe []*model.Trace
-	for _, traceID := range uniqueTraceIDs {
-		if len(retMe) >= traceQuery.NumTraces {
-			break
-		}
-		trace, err := s.readTrace(traceID, traceQuery)
+	return s.multiRead(uniqueTraceIDs, traceQuery)
+}
+
+func (s *SpanReader) multiRead(traceIDs []string, traceQuery *spanstore.TraceQueryParameters) ([]*model.Trace, error) {
```
black-adder: looks like we can remove the readTrace function entirely given it's a subset of this function

mh-park: as in use this instead of readTrace? we still use readTrace for GetTrace

plugin/storage/es/spanstore/reader.go
```diff
@@ -238,19 +214,43 @@ func (s *SpanReader) FindTraces(traceQuery *spanstore.TraceQueryParameters) ([]*
...
+	}
+	searchRequests := make([]*elastic.SearchRequest, len(traceIDs))
+
+	for i, traceID := range traceIDs {
+		query := elastic.NewTermQuery("traceID", traceID)
+		searchRequests[i] = elastic.NewSearchRequest().IgnoreUnavailable(true).Type("span").
+			Source(elastic.NewSearchSource().Query(query).Size(defaultDocCount))
+	}
+
+	indices := findIndices(traceQuery.StartTimeMin.Add(-time.Hour), traceQuery.StartTimeMax.Add(time.Hour))
```
black-adder: Add a little note describing why we do this

PR #303, opened 2017-08-02
Make ES number of shards and replicas configurable

### LINKED ISSUE #300: Allow users to pass in flag for ElasticSearch # of shards and RF for the index

mh-park:
Resolves #300

### DIFF
--- a/pkg/es/config/config.go
+++ b/pkg/es/config/config.go
@@ -36,8 +36,8 @@ type Configuration struct {
 	Password    string
 	Sniffer     bool          // https://github.com/olivere/elastic/wiki/Sniffing
 	MaxSpanAge  time.Duration `yaml:"max_span_age"` // configures the maximum lookback on span reads
-	NumShards   int           `yaml:"shards"`
-	NumReplicas int           `yaml:"replicas"`
+	NumShards   int64         `yaml:"shards"`
+	NumReplicas int64         `yaml:"replicas"`
 }
 
 // NewClient creates a new ElasticSearch client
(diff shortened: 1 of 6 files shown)


pkg/es/config/config.go
```diff
@@ -31,11 +31,13 @@ import (
 
 // Configuration describes the configuration properties needed to connect to a ElasticSearch cluster
```
black-adder: an

plugin/storage/es/spanstore/writer.go
```diff
@@ -65,8 +69,20 @@ type Service struct {
...
+func NewSpanWriter(
+	client es.Client,
+	logger *zap.Logger,
+	metricsFactory metrics.Factory,
+	numShards int64,
+	numReplicas int64,
+) *SpanWriter {
 	ctx := context.Background()
+	if numShards == 0 {
+		numShards = 5
```
black-adder: consts

plugin/storage/es/spanstore/writer_test.go
```diff
@@ -178,10 +179,16 @@ func TestSpanWriter_WriteSpan(t *testing.T) {
 				spanExistsService.On("Do", mock.AnythingOfType("*context.emptyCtx")).Return(testCase.spanIndexExists, nil)
 
 				serviceCreateService := &mocks.IndicesCreateService{}
+				serviceMapping = strings.Replace(serviceMapping, "${__NUMBER_OF_SHARDS__}", strconv.FormatInt(w.writer.numShards, 10), 1)
```
black-adder: any better way to test this? right now you're using the same replace logic in the code and in the test to ensure the mapping was replaced.

PR #308, opened 2017-08-03
Refactor Viperize function

black-adder:


black-adder:
tests already exist for this function (it's more of a test utility function than anything else)

yurishkuro:
I don't see any tests in this package

pavolloffay:
Does the coverage check work for all packages e.g. cmd, pkg...?

black-adder:
go only checks coverage for directories that have at least one _test.go file

PR #309, opened 2017-08-03
Add ElasticSearch flags to query/collector

mh-park:


cmd/query/main.go
```diff
@@ -43,6 +44,8 @@ import (
 func main() {
 	logger, _ := zap.NewProduction()
 	casOptions := casFlags.NewOptions("cassandra", "cassandra.archive")
+	// currently no archive for es
+	esOptions := esFlags.NewOptions("es")
```
yurishkuro: what's preventing it?

yurishkuro: although even for Cassandra I don't think we're actually wiring in the archive storage

PR #316, opened 2017-08-04
Remove default cassandra ProtoVersion

black-adder:
We currently hard code the version to 4; this doesn't seem necessary given default protoVersion is set on the primary namespace and gocql also discovers the best protoVersion for the current session if protoVersion is not set.

As a side note we run cassandra 2 internally for "reasons" and this prevents it.


pavolloffay:
I was wondering why it is there. I have found another occurrence https://github.com/uber/jaeger/blob/master/cmd/query/app/builder/cassandra.go#L56

PR #317, opened 2017-08-04
Raise timeout for ES cleanup script

mh-park:


plugin/storage/es/esCleaner.py
```diff
@@ -18,14 +19,15 @@ def main():
 
     for index in ilo.working_list():
         print "Removing", index
-    delete_indices = curator.DeleteIndices(ilo)
```
black-adder: can we make this configurable and have it default to 120?

plugin/storage/es/esCleaner.py
```diff
@@ -1,11 +1,13 @@
...
 import curator
+import logging
 import sys
+import os
 
 
 def main():
     if len(sys.argv) == 1:
-        print('USAGE: %s NUM_OF_DAYS HOSTNAME[:PORT] ...' % sys.argv[0])
+        print('USAGE: [TIMEOUT=(default 120)] %s NUM_OF_DAYS HOSTNAME[:PORT] ...' % sys.argv[0])
```
yurishkuro: print a brief explanation of the script and explain all parameters

mh-park: this happens in `./es_indices_clean.sh`
technically i don't need this piece of code in the python script, but should i add an explanation to both anyways?

mh-park: https://github.com/uber/jaeger/blob/b31ca51b47d7ca7c9e78fbe6fd64638fedf30175/plugin/storage/es/es_indices_clean.sh#L4

plugin/storage/es/esCleaner.py
```diff
@@ -18,14 +21,16 @@ def main():
...
+    timeout = int(os.getenv("TIMEOUT", 120))
+    delete_indices = curator.DeleteIndices(ilo, master_timeout=timeout)
     delete_indices.do_action()
 
 
 def empty_list(ilo, error_msg):
     try:
         ilo.empty_list_check()
     except curator.NoIndices:
+        logging.info(error_msg)
```
black-adder: why is this needed if print is happening below?

PR #322, opened 2017-08-07
Refactor storage configuration

pavolloffay:
Fixes #296

PR #323, opened 2017-08-07
Add basic authenticator to configs for Cassandra

### LINKED ISSUE #255: Unable to start collector with PasswordAuthenticator
Is there an option to pass username/password for cassandra while starting collector?

I dont not see any option to do that

  -cassandra.connections-per-host int
        The number of Cassandra connections from a single backend instance (default 2)
  -cassandra.keyspace string
        The Cassandra keyspace for Jaeger data (default "jaeger_v1_local")
  -cassandra.max-retry-attempts int
        The number of attempts when reading from Cassandra (default 3)
  -cassandra.port int
        The port for cassandra
  -cassandra.proto-version int
        The Cassandra protocol version (default 4)
  -cassandra.servers string
        The comma-separated list of Cassandra servers (default "127.0.0.1")
  -cassandra.socket-keep-alive duration
        Cassandra's keepalive period to use, enabled if > 0
  -cassandra.timeout duration
        Timeout used for queries

mh-park:
Resolves #255

### DIFF
--- a/cmd/flags/cassandra/options.go
+++ b/cmd/flags/cassandra/options.go
@@ -38,6 +38,8 @@ const (
 	suffixKeyspace         = ".keyspace"
 	suffixProtoVer         = ".proto-version"
 	suffixSocketKeepAlive  = ".socket-keep-alive"
+	suffixUsername         = ".username"
+	suffixPassword         = ".password"
 )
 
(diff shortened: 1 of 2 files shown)


yurishkuro:
will this conflict with #322? If so, please don't merge before that one lands

mh-park:
@yurishkuro none of the files conflict, unless there are more changes on the way for #322

PR #324, opened 2017-08-08
Change default ElasticSearch server flag value

mh-park:


cmd/flags/es/options.go
```diff
@@ -109,7 +109,7 @@ func addFlags(flagSet *flag.FlagSet, nsConfig *namespaceConfig) {
 	flagSet.String(
 		nsConfig.namespace+suffixServers,
```
black-adder: maybe change suffixServers to suffixServerURLs? but the internal variable can be kept as servers.

PR #325, opened 2017-08-08
Change ElasticSearch replicas default value

mh-park:
the default value of replicas is 1, not 2; passing in 1 as the number of replicas means 1 primary + 1 replica, so 2 in total.
https://www.elastic.co/guide/en/elasticsearch/guide/current/replica-shards.html


mh-park:
@yurishkuro this may conflict with #322, so this should wait until that pr is merged

PR #327, opened 2017-08-09
Zipkin error tag sanitizer

### LINKED ISSUE #184: zipkin.thrift error tag value set as a string
When using zipkin.thrift error tag is set as a string value (maybe every value is set as a string). This may affect future data analysis pipeline.

Fix for UI was done here https://github.com/uber/jaeger-ui/pull/32

pavolloffay:
Fixes #184

### DIFF
--- a/cmd/collector/app/sanitizer/zipkin/span_sanitizer.go
+++ b/cmd/collector/app/sanitizer/zipkin/span_sanitizer.go
@@ -22,6 +22,7 @@ package zipkin
 
 import (
 	"strconv"
+	"strings"
 
 	"go.uber.org/zap"
 
@@ -119,5 +120,39 @@ func (s *parentIDSanitizer) Sanitize(span *zc.Span) *zc.Span {
 	}
 	span.BinaryAnnotations = append(span.BinaryAnnotations, &annotation)
 	span.ParentID = nil
+	return span
+}
+
+// NewErrorTagSanitizer returns a sanitizer that changes error binary annotations to boolean type
+// and sets appropriate value, in case value was a string message it adds a 'error.message' binary annotation with
+// this message.
+func NewErrorTagSanitizer() Sanitizer {
+	return &errorTagSanitizer{}
+}
+
+type errorTagSanitizer struct {
+}
+
+func (s *errorTagSanitizer) Sanitize(span *zc.Span) *zc.Span {
+	for _, binAnno := range span.BinaryAnnotations {
+		if binAnno.AnnotationType != zc.AnnotationType_BOOL && strings.EqualFold("error", binAnno.Key) {
+			binAnno.AnnotationType = zc.AnnotationType_BOOL
+
+			if strings.EqualFold("true", string(binAnno.Value)) || len(binAnno.Value) == 0 {
+				binAnno.Value = []byte{1}
+			} else if strings.EqualFold("false", string(binAnno.Value)) {
+				binAnno.Value = []byte{0}
+			} else {
+				// value is different to true/false, create another bin annotation with error message
+				annoErrorMsg := &zc.BinaryAnnotation{
+					Key:   "error.message",
+					Value: binAnno.Value,
+				}
+				span.BinaryAnnotations = append(span.BinaryAnnotations, annoErrorMsg)
+				binAnno.Value = []byte{1}
+			}
+		}
+	}
+
 	return span
 }
--- a/cmd/collector/app/builder/span_handler_builder.go
+++ b/cmd/collector/app/builder/span_handler_builder.go
@@ -134,6 +134,7 @@ func (spanHb *SpanHandlerBuilder) BuildHandlers() (app.ZipkinSpansHandler, app.J
 	zSanitizer := zs.NewChainedSanitizer(
 		zs.NewSpanDurationSanitizer(spanHb.logger),
 		zs.NewParentIDSanitizer(spanHb.logger),
+		zs.NewErrorTagSanitizer(),
 	)
 
 	spanWriter, err := spanHb.storageBuilder()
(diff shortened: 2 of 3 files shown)


cmd/collector/app/sanitizer/zipkin/span_sanitizer.go
```diff
@@ -121,3 +122,25 @@ func (s *parentIDSanitizer) Sanitize(span *zc.Span) *zc.Span {
 	span.ParentID = nil
 	return span
 }
+
+func NewErrorTagSanitizer() Sanitizer {
+	return &errorTagSanitizer{}
+}
+
+type errorTagSanitizer struct {
+	log spanLogger
```
black-adder: not used

cmd/collector/app/sanitizer/zipkin/span_sanitizer.go
```diff
@@ -121,3 +122,25 @@ func (s *parentIDSanitizer) Sanitize(span *zc.Span) *zc.Span {
...
+}
+
+func (s *errorTagSanitizer) Sanitize(span *zc.Span) *zc.Span {
+	for i := 0; i < len(span.BinaryAnnotations); i++ {
+		anno := span.BinaryAnnotations[i]
+		if strings.EqualFold("error", anno.Key) {
+			anno.AnnotationType = zc.AnnotationType_BOOL
+			anno.Value = []byte("error")
```
black-adder: I thought the issue was that we are submitting both boolean and string values from our clients according to this UI fix: https://github.com/uber/jaeger-ui/pull/32/files

pavolloffay: The issue was that error binary annotation was passed as a string and not as a boolean.

yurishkuro: I think you need `anno.Value = []byte{0x01}`

Also, in theory someone can set a tag `error=false`, so the correct thing to do would be to convert the value from string to bool correctly.

cmd/collector/app/sanitizer/zipkin/span_sanitizer.go
```diff
@@ -124,7 +124,7 @@ func (s *parentIDSanitizer) Sanitize(span *zc.Span) *zc.Span {
 }
 
 // NewErrorTagSanitizer returns a sanitizer that changes error binary annotations to boolean type
-// and sets value to lower case 'error'
+// and sets value to false
```
yurishkuro: you mean sets the value to "true"?

cmd/collector/app/sanitizer/zipkin/span_sanitizer.go
pavolloffay: https://github.com/openzipkin/zipkin/blob/2755987aa31699bf02d9e6e0366453092ba4a944/zipkin/src/main/java/zipkin/Constants.java#L147

>  the value is a human readable message associated with an error.

I think we should not override the value. There will be an annotation with error key plus bin annotation with an error key and value error message. What we should probably do is:
1. create an error tag for annotation 
2. create a log entry for binary annotation

I am not sure how this would affect older jaeger-ot clients using zipkin model.

black-adder: I hadn't considered that. Older clients emitting zipkin directly might still be emitting Annotation{key: error, value: true} so we might have to add an extra check which only adds a log entry if the value is not "true" or True

yurishkuro:
I think we're considering several different use cases here. One is instrumentation doing `span.SetTag("error", "true")`, which is sort-of OT-compliant except for the value type being string. Another is instrumentation doing it right with a Boolean but the tracer converting the value to string, for whatever reason. And the third one is a native Zipkin instrumentation sending spans to Jaeger that is equivalent to `span.SetTag("error", "<error message.")` (which is not an OT-compliant way).

For the first two use cases we should simple parse the value as boolean and convert the binary annotation to BOOL type with the converted value as 0 or 1.

For the latter use case the only sensible thing I see doing is to convert that single binary annotation to two: `("error", BOOL, 0x1)` and `("error.message", STRING, <error message>)`. We cannot do "logs" here because we don't have a timestamp (the OT-compatible way would be to record the error message in the log, separate from the `error=true` tag).

pavolloffay:
>  We cannot do "logs" here because we don't have a timestamp (the OT-compatible way would be to record the error message in the log, separate from the error=true tag). 

We could do it. Native zipkin instrumentation sets annotation (with timestamp) to error and creates binary annotation with error and error string. I will try to include this use case if it gets too messy we could skip it.

adriancole:
>
> We could do it. Native zipkin instrumentation sets annotation (with
> timestamp) to error and creates binary annotation with error and error
> string.
>
this is not quite correct. the error timestamp annotation is not always
present and when present it is often a substring not an exact string. I
wouldn't rely on finding an "error" timestamped annotation. The closest
would be the end of the span (timestamp+duration or last annotation)

pavolloffay:
@adriancole thanks for the clarification. Let's better not add a log entry. 

PR updated, It does mostly what @yurishkuro described.

michael-barker:
@pavolloffay I was trying out Sleuth with both Zipkin and Jaeger.  I noticed that my error tag which is a string is reported with the string value in Zipkin under the erro tag.  In Jaeger I see that the error tag is set to true but that error.message is set to the value false.  Based on this pull request should I expect the value of error.message to be the same string value that was originally present in the error tag?  Is this an issue or am I misunderstanding how it should function?

pavolloffay:
It seems that default type of the annotation was set to bool.

It's been fixed in https://github.com/jaegertracing/jaeger/pull/891

Issue #207, opened 2017-06-14
SURVEY: Who is using Jaeger


badiib:
Hi, you are in a group of individuals who have create or commented on issues in the Jaeger repository and we are doing a simple informal survey about Jaeger usage. If you could answer the following questions, it would be very valuable to gauge interest in the project:

- If applicable, what company/organization do you represent?
- How are you using Jaeger? E.g. full production deployment, considering, experimenting, or I am not using Jaeger etc.
- If you are not using Jaeger, why not?
- How many services (or microservices) exist in your system layout?

@jkandasa @sunfaces @jbdalido @princeop @pavolloffay @mabn @jpkrohling @nlamirault @JodeZer @prestonprice57 @jrbury @objectiser @sloev @hwinkel @Madhu1512 @yuekui2 @valichek @dianvaltodorov @ZhouZiHe @LoungeFlyZ @jeluard @diegofernandes @d-ulyanov @jyothepro @yqf3139 @tomersimis @ruinanchen @szdavid92 @anuptalwalkar @hekike @sul4bh @Strandedpirate @julianste @awhiteside @nklmish @sweatybridge @kevinearls @felixbarny @hzariv @nlamirault @longXboy @drzero42 @xdralex @philipgian @bharat-p

JodeZer:
* A financial system service providers company From Shanghai
* experimenting
* To be honest, we are considering zipkin more.My team members are more familiar with ES and MySql than  Cassandra, and our java coders like zipkin. Though l as a pure gocoder perfer jaeger.
* We are during a microservice transformation.And our system is half java and half golang.
* BTW, we select a kafka x ES solution with zipkin which jaeger does not provide.

yqf3139:
**If applicable, what company/organization do you represent?**

I am a contributor to [fission](https://github.com/fission/fission), which is a FaaS solution on top of Kubernetes.

**How are you using Jaeger? E.g. full production deployment, considering, experimenting, or I am not using Jaeger etc.**

We need to integrate a distributed tracer for two usage: 
- Help to troubleshoot performance problems with fission itself. 
- Provide tracer handler to user so that they can instrument their code easily and trace the functions as part of a bigger solution.

Currently I am doing some experiments on the integration.

**If you are not using Jaeger, why not?**

Will find myself some time to try Jaeger. It seems Jaeger has better client library support.

**How many services (or microservices) exist in your system layout?**

Around 10 microservices. Excluding user functions, which are also services evolving over time.

[...]

PR #332, opened 2017-08-13
Add adopters list

caniszczyk:
Add an ADOPTERS.md to track production usage of Jaeger. It's also easier for potential adopters to add themselves to the list. Also update the README with this information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,12 @@ Reach project contributors via these channels:
  * [Gitter chat room](https://gitter.im/jaegertracing/Lobby)
  * [Github issues](https://github.com/uber/jaeger/issues)
 
+## Adopters
+
+Jaeger is [used in production](./ADOPTERS.md) by organizations such as Atlassian, Orange, Red Hat and Uber.
+
+If you use Jaeger in production, please consider adding yourself to our [ADOPTERS.md](./ADOPTERS.md)
+
 ## License
 
 [The MIT License](./LICENSE).
--- a/ADOPTERS.md
+++ b/ADOPTERS.md
@@ -1,0 +1,15 @@
+* Atlassian
+* Base CRM
+* Booking.com
+* Circonus
+* Massachusetts Open Cloud
+* Nets
+* Orange
+* Squarespace
+* Stagemonitor
+* Red Hat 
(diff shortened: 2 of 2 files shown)


caniszczyk:
Hey folks, can you please verify this is the list of actual organizations that are using Jaeger in production. Having a list of adopters is one of the requirements in becoming part of CNCF.

README.md
```diff
@@ -53,6 +53,12 @@ Reach project contributors via these channels:
  * [Gitter chat room](https://gitter.im/jaegertracing/Lobby)
  * [Github issues](https://github.com/uber/jaeger/issues)
 
+## Adopters
+
+Jaeger is [used in production](./ADOPTERS.md) by organizations such as Atlassian, Orange, Red Hat and Uber.
```
adriancole: @caniszczyk doesn't this statement conflate "adopting" with production usage? For example, vendors typically are not end users, though sometimes they are. For example, I'd be surprised if @postwait and others at circonus replaced their core infrastructure with jaeger, even if they have a zipkin-jaeger plug-in for one of their projects.

Similarly Red Hat do run production infrastructure I think, does this mean the public openshift infrastructure is monitored with jaeger? or does it mean they support and build plugins and tooling for others?

TL;DR; is I think this is "leading the witness" in so far as it is conflating a list of folks participating in the ecosystem with being production users. Users have skin in the game, when something breaks, their alerts go off. Production usually means things like this. Maybe it is better to just call it an adopters list and leave bullets in the list for people to offer information about whether they are in production or not?

I CNCF projects would have higher credibility if folks get accurate info about their projects.

yurishkuro: I think it's a fair point. There are different types of users on this list
  * some have full prod deployment
  * some are using Jaeger end to end, but in dev/qa (which could also a valid use case)
  * some are using parts of Jaeger like the client libs
  * some are using Jaeger as part of other products (e.g. vendors who may not even want to reveal their actual customers)
  * some are just experimenting

Rather than listing them as "Adopters", maybe we can name the readme section "Testimonials" and point to the survey (#207)?

postwait: We o the use the whole stack, but we use large parts and we use it in a production capacity.

yurishkuro:
done in #340

PR #333, opened 2017-08-15
Zipkin span start time and duration sanitizer

pavolloffay:
Similar to: https://github.com/openzipkin/zipkin/blob/master/zipkin/src/main/java/zipkin/internal/ApplyTimestampAndDuration.java

This is required for follow-up PR with zipkin JSON encoding.

### DIFF
--- a/cmd/collector/app/sanitizer/zipkin/span_sanitizer.go
+++ b/cmd/collector/app/sanitizer/zipkin/span_sanitizer.go
@@ -83,8 +83,29 @@ type spanDurationSanitizer struct {
 func (s *spanDurationSanitizer) Sanitize(span *zc.Span) *zc.Span {
 	if span.Duration == nil {
 		span.Duration = &defaultDuration
+		if len(span.Annotations) < 2 {
+			return span
+		}
+		// Prefer RPC one-way (cs -> sr) vs arbitrary annotations.
+		first := span.Annotations[0].Timestamp
+		last := span.Annotations[len(span.Annotations) - 1].Timestamp
+		for _, anno := range span.Annotations {
+			if anno.Value == zc.CLIENT_SEND {
+				first = anno.Timestamp
+			} else if anno.Value == zc.CLIENT_RECV {
+				last = anno.Timestamp
+			}
+		}
+		if first != last {
+			duration := last - first
+			span.Duration = &duration
+			if span.Timestamp == nil {
+				span.Timestamp = &first
+			}
+		}
 		return span
 	}
+
 	duration := *span.Duration
 	if duration >= 0 {
 		return span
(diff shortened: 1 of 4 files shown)


cmd/collector/app/sanitizer/zipkin/span_sanitizer.go
```diff
@@ -99,6 +120,32 @@ func (s *spanDurationSanitizer) Sanitize(span *zc.Span) *zc.Span {
...
+func (s *spanStartTimeSanitizer) Sanitize(span *zc.Span) *zc.Span {
+	if span.Timestamp != nil || len(span.Annotations) == 0 {
+		return span
+	}
+
+	for _, anno := range span.Annotations {
+		if anno.Value == zc.CLIENT_SEND {
+			span.Timestamp = &anno.Timestamp
+			return span
+		} else if anno.Value == zc.SERVER_RECV && span.ParentID == nil {
```
yurishkuro: `else` is redundant since previous if ends with `return`

cmd/collector/app/sanitizer/zipkin/span_sanitizer.go
```diff
@@ -99,6 +120,32 @@ func (s *spanDurationSanitizer) Sanitize(span *zc.Span) *zc.Span {
...
+			span.Timestamp = &anno.Timestamp
```
yurishkuro: comment explaining why not breaking out of loop in this case?

cmd/collector/app/sanitizer/zipkin/span_sanitizer.go
```diff
@@ -83,8 +83,29 @@ type spanDurationSanitizer struct {
...
+		for _, anno := range span.Annotations {
+			if anno.Value == zc.CLIENT_SEND {
+				first = anno.Timestamp
+			} else if anno.Value == zc.CLIENT_RECV {
+				last = anno.Timestamp
+			}
+		}
+		if first != last {
+			duration := last - first
+			span.Duration = &duration
```
yurishkuro: double allocation with L85 - better to declare local var `duration` and do `span.Duration = &duration` at the end of `if`

cmd/collector/app/sanitizer/zipkin/span_sanitizer.go
```diff
@@ -83,8 +83,29 @@ type spanDurationSanitizer struct {
...
+		first := span.Annotations[0].Timestamp
+		last := span.Annotations[len(span.Annotations) - 1].Timestamp
+		for _, anno := range span.Annotations {
+			if anno.Value == zc.CLIENT_SEND {
+				first = anno.Timestamp
+			} else if anno.Value == zc.CLIENT_RECV {
+				last = anno.Timestamp
+			}
+		}
+		if first != last {
```
yurishkuro: maybe `first < last`? Otherwise you can assign negative TS

pavolloffay:
@yurishkuro thanks for the quick look. PR updated

pavolloffay:
close/open to trigger travis build.

PR #334, opened 2017-08-15
Zipkin json encoding via HTTP

pavolloffay:
Fixes https://github.com/uber/jaeger/issues/225

Zipkin API definition: https://github.com/openzipkin/zipkin-api/blob/master/zipkin-api.yaml

Docker Image to try it: https://hub.docker.com/r/pavolloffay/tmp-jaeger-zipkin-json/

### DIFF
--- a/cmd/collector/app/zipkin/http_handler.go
+++ b/cmd/collector/app/zipkin/http_handler.go
@@ -33,7 +33,7 @@ import (
 	tchanThrift "github.com/uber/tchannel-go/thrift"
 
 	"github.com/uber/jaeger/cmd/collector/app"
-	zipkincore "github.com/uber/jaeger/thrift-gen/zipkincore"
+	"github.com/uber/jaeger/thrift-gen/zipkincore"
 )
 
 // APIHandler handles all HTTP calls to the collector
@@ -75,30 +75,36 @@ func (aH *APIHandler) saveSpans(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if r.Header.Get("Content-Type") == "application/x-thrift" {
-		handleZipkinThrift(aH.zipkinSpansHandler, bodyBytes, w)
+	contentType := r.Header.Get("Content-Type")
+	var tSpans []*zipkincore.Span
+	if contentType == "application/x-thrift" {
+		tSpans, err = deserializeThrift(bodyBytes)
+		if err != nil {
+			http.Error(w, fmt.Sprintf(app.UnableToReadBodyErrFormat, err), http.StatusBadRequest)
+			return
+		}
+	} else if contentType == "application/json" {
+		tSpans, err = deserializeJSON(bodyBytes)
+		if err != nil {
+			http.Error(w, fmt.Sprintf(app.UnableToReadBodyErrFormat, err), http.StatusBadRequest)
+			return
+		}
 	} else {
-		http.Error(w, "Only Content-Type:application/x-thrift is supported at the moment", http.StatusBadRequest)
+		http.Error(w, "Not supported Content-Type", http.StatusBadRequest)
+	}
+
+	if len(tSpans) > 0 {
+		ctx, _ := tchanThrift.NewContext(time.Minute)
+		if _, err = aH.zipkinSpansHandler.SubmitZipkinBatch(ctx, tSpans); err != nil {
+			http.Error(w, fmt.Sprintf("Cannot submit Zipkin batch: %v", err), http.StatusInternalServerError)
+			return
+		}
 	}
 
 	w.WriteHeader(http.StatusAccepted)
 }
 
-func handleZipkinThrift(zHandler app.ZipkinSpansHandler, bodyBytes []byte, w http.ResponseWriter) {
-	spans, err := deserializeZipkin(bodyBytes)
-	if err != nil {
-		http.Error(w, fmt.Sprintf(app.UnableToReadBodyErrFormat, err), http.StatusBadRequest)
-		return
-	}
-
-	ctx, _ := tchanThrift.NewContext(time.Minute)
-	if _, err = zHandler.SubmitZipkinBatch(ctx, spans); err != nil {
-		http.Error(w, fmt.Sprintf("Cannot submit Zipkin batch: %v", err), http.StatusInternalServerError)
-		return
-	}
-}
-
-func deserializeZipkin(b []byte) ([]*zipkincore.Span, error) {
+func deserializeThrift(b []byte) ([]*zipkincore.Span, error) {
 	buffer := thrift.NewTMemoryBuffer()
 	buffer.Write(b)
 
--- a/cmd/collector/app/zipkin/http_handler.go
+++ b/cmd/collector/app/zipkin/http_handler.go
@@ -90,7 +90,7 @@ func (aH *APIHandler) saveSpans(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 	} else {
-		http.Error(w, "Not supported Content-Type", http.StatusBadRequest)
+		http.Error(w, "Unsupported Content-Type", http.StatusBadRequest)
 	}
 
 	if len(tSpans) > 0 {
--- a/cmd/collector/app/zipkin/http_handler_test.go
+++ b/cmd/collector/app/zipkin/http_handler_test.go
@@ -131,34 +131,35 @@ func TestJsonFormat(t *testing.T) {
 	binAnnoJSON := createBinAnno("http.status_code", "200", endpJSON)
 	spanJSON := createSpan("bar", "1234567891234565", "1234567891234567", "1234567891234568", 156, 15145, false,
 		annoJSON, binAnnoJSON)
-
 	statusCode, resBodyStr, err := postBytes(server.URL+`/api/v1/spans`, []byte(spanJSON), createHeader("application/json"))
 	assert.NoError(t, err)
 	assert.EqualValues(t, http.StatusAccepted, statusCode)
 	assert.EqualValues(t, "", resBodyStr)
 
+	endpErrJSON := createEndpoint("", "127.0.0.A", "", 80)
+
 	// error zipkinSpanHandler
 	handler.zipkinSpansHandler.(*mockZipkinHandler).err = fmt.Errorf("Bad times ahead")
-	statusCode, resBodyStr, err = postBytes(server.URL+`/api/v1/spans`, []byte(spanJSON), createHeader("application/json"))
-	assert.NoError(t, err)
-	assert.EqualValues(t, http.StatusInternalServerError, statusCode)
-	assert.EqualValues(t, "Cannot submit Zipkin batch: Bad times ahead\n", resBodyStr)
+	tests := []struct {
+		payload    string
+		expected   string
+		statusCode int
+	}{
+		{spanJSON, "Cannot submit Zipkin batch: Bad times ahead\n", http.StatusInternalServerError},
+		{createSpan("bar", "", "1", "1", 156, 15145, false, annoJSON, binAnnoJSON),
+			"Unable to process request body: id is not an unsigned long\n", http.StatusBadRequest},
+		{createSpan("bar", "ZTA", "1", "1", 156, 15145, false, "", ""),
+			"Unable to process request body: id is not an unsigned long\n", http.StatusBadRequest},
+		{createSpan("bar", "1", "", "1", 156, 15145, false, "", createAnno("cs", 1, endpErrJSON)),
+			"Unable to process request body: wrong ipv4\n", http.StatusBadRequest},
+	}
 
-	// error json no id
-	spanJSON = createSpan("bar", "", "1234567891234567", "1234567891234568", 156, 15145, false,
-		annoJSON, binAnnoJSON)
-	statusCode, resBodyStr, err = postBytes(server.URL+`/api/v1/spans`, []byte(spanJSON), createHeader("application/json"))
-	require.NoError(t, err)
-	assert.EqualValues(t, http.StatusBadRequest, statusCode)
-	assert.EqualValues(t, "Unable to process request body: id is not an unsigned long\n", resBodyStr)
-
-	// error toThrift no id no integer
-	spanJSON = createSpan("bar", "Z23456789123456A", "1234567891234567", "1234567891234568", 156, 15145, false,
-		annoJSON, binAnnoJSON)
-	statusCode, resBodyStr, err = postBytes(server.URL+`/api/v1/spans`, []byte(spanJSON), createHeader("application/json"))
-	require.NoError(t, err)
-	assert.EqualValues(t, http.StatusBadRequest, statusCode)
-	assert.EqualValues(t, "Unable to process request body: id is not an unsigned long\n", resBodyStr)
+	for _, test := range tests {
+		statusCode, resBodyStr, err = postBytes(server.URL+`/api/v1/spans`, []byte(test.payload), createHeader("application/json"))
+		require.NoError(t, err)
+		assert.EqualValues(t, test.statusCode, statusCode)
+		assert.EqualValues(t, test.expected, resBodyStr)
+	}
 }
 
 func TestGzipEncoding(t *testing.T) {
--- a/cmd/collector/app/zipkin/http_handler_test.go
+++ b/cmd/collector/app/zipkin/http_handler_test.go
@@ -145,13 +145,13 @@ func TestJsonFormat(t *testing.T) {
 		expected   string
 		statusCode int
 	}{
-		{spanJSON, "Cannot submit Zipkin batch: Bad times ahead\n", http.StatusInternalServerError},
-		{createSpan("bar", "", "1", "1", 156, 15145, false, annoJSON, binAnnoJSON),
-			"Unable to process request body: id is not an unsigned long\n", http.StatusBadRequest},
-		{createSpan("bar", "ZTA", "1", "1", 156, 15145, false, "", ""),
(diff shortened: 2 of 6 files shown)


cmd/collector/app/zipkin/http_handler.go
```diff
@@ -75,30 +75,36 @@ func (aH *APIHandler) saveSpans(w http.ResponseWriter, r *http.Request) {
...
+			return
+		}
+	} else if contentType == "application/json" {
+		tSpans, err = deserializeJSON(bodyBytes)
+		if err != nil {
+			http.Error(w, fmt.Sprintf(app.UnableToReadBodyErrFormat, err), http.StatusBadRequest)
+			return
+		}
 	} else {
 		http.Error(w, "Only Content-Type:application/x-thrift is supported at the moment", http.StatusBadRequest)
```
yurishkuro: extend err msg

cmd/collector/app/zipkin/http_handler_test.go
```diff
@@ -148,7 +187,7 @@ func TestGzipBadBody(t *testing.T) {
 func TestUnsupportedContentType(t *testing.T) {
 	server, _ := initializeTestServer(nil)
 	defer server.Close()
-	statusCode, _, err := postBytes(server.URL+`/api/v1/spans`, []byte{}, createHeader("application/json"))
+	statusCode, _, err := postBytes(server.URL+`/api/v1/spans`, []byte{}, createHeader("text/html"))
```
yurishkuro: xml would be funnier

cmd/collector/app/zipkin/zipkin_json.go
```diff
@@ -0,0 +1,215 @@
...
+}
+type annotation struct {
+	Endpoint  endpoint `json:"endpoint"`
+	Value     string   `json:"value"`
+	Timestamp int64    `json:"timestamp"`
+}
+type binaryAnnotation struct {
+	Endpoint endpoint `json:"endpoint"`
+	Key      string   `json:"key"`
+	Value    string   `json:"value"`
```
yurishkuro: why did [Adrian refer](https://github.com/openzipkin/zipkin/issues/1499#issuecomment-322213158) to `oneOf`? Is this the right JSON format?

cmd/collector/app/zipkin/zipkin_json.go
```diff
@@ -0,0 +1,215 @@
...
+func decode(body []byte) ([]zipkinSpan, error) {
+	type zipkinSpans []zipkinSpan
+	var spans zipkinSpans
+
+	if err := json.Unmarshal(body, &spans); err != nil {
+		return nil, err
+	}
+
+	for _, span := range spans {
+		if len(span.TraceID) != 16 {
```
yurishkuro: how are we dealing with 128bit IDs? Does Zipkin JSON encode them as one string, or hi/lo?

cmd/collector/app/zipkin/zipkin_json.go
```diff
@@ -0,0 +1,215 @@
...
+			return nil, errTraceID
+		}
+		if len(span.ID) != 16 && len(span.ID) != 32 {
+			return nil, errSpanID
+		}
+		if len(span.ParentID) > 0 && len(span.ParentID) != 16 {
```
yurishkuro: I never understood this restriction in Zipkin to have the exact length - what if some leading zeroes were dropped for whatever reason, we can still parse the hex string. Especially for parsing out-of-band span data,  I am not sure why we need to be this strict.

pavolloffay:
How are we going to write integration tests for this? `zipkin-go-opentracing` supports only thrift. We could use `brave` but it's in java.  I think we have two options:
1. modify xdock-java to use brave
2. crate new xdock-zipkin which would verify zipkin related stuff.

cmd/collector/app/zipkin/zipkin_json.go
```diff
@@ -0,0 +1,215 @@
...
+
+	return &zipkincore.BinaryAnnotation{
+		Key:            binAnno.Key,
+		Value:          []byte(binAnno.Value),
+		Host:           endpoint,
+		AnnotationType: zipkincore.AnnotationType_STRING,
+	}, nil
+}
+
+func parseIpv4(str string) (int32, error) {
```
yurishkuro: so we're only parsing IPv4 from Zipkin, not IPv6?

Note that our zipkin.thrift (in jaeger-idl) is probably behind the official Zipkin IDL, e.g. we don't have IPv6 or TraceIdHigh fields.

pavolloffay: yes, the reason is that our zipkin.thrift does not support ipv6. I will leave TODO comment there.

yurishkuro:
> How are we going to write integration tests for this?

I think we might be able to use the same xdock driver, but a different "client" that's implemented with Zipkin's JSON and Thrift official senders (two clients, really)

@black-adder - can we come up with a unified end-to-end testing framework? Ultimately it boils down to two parts
  1. running some configuration of the backend (could be all-in-one, or docker-compose, or k8s)
  1. having the driver his _some sort of client_ that can report spans to the backend (either via agent or via direct POST)

cmd/collector/app/zipkin/zipkin_json.go
yurishkuro: Yeah, we should sync it up with official IDL (should be backwards compatible) - #335

cmd/collector/app/zipkin/http_handler_test.go
pavolloffay: `video/mp4` :P

cmd/collector/app/zipkin/zipkin_json.go
pavolloffay: I think `oneOf` refers to `bool`, `string` `num` so on. In the current https://github.com/openzipkin/zipkin-api/blob/master/zipkin-api.yaml it's defined as string

cmd/collector/app/zipkin/http_handler.go
pavolloffay: The comment is odd, but why custom `error`? I don't see a value for it

yurishkuro: I meant err msg says only x-thrift is supported, not json, so we can extend it to mention both

cmd/collector/app/zipkin/zipkin_json.go
pavolloffay: Honestly, I don't know, it's defined in api definition.

@adriancole hi, would you give us some insights here?

cmd/collector/app/zipkin/zipkin_json.go
pavolloffay: a good point, 128bit IDs is long 32 chars. It's using this https://github.com/openzipkin/zipkin/blob/cf4df43952f2b82496c2e4c730765e7b70eba22b/zipkin/src/main/java/zipkin/internal/Util.java#L114

cmd/collector/app/zipkin/zipkin_json.go
pavolloffay: @yurishkuro updated to align with your comment. It seems that zipkin sever does the same.

cmd/collector/app/zipkin/zipkin_json.go
pavolloffay: Currently parsing only `traceId[16:]` (https://github.com/openzipkin/zipkin/blob/cf4df43952f2b82496c2e4c730765e7b70eba22b/zipkin/src/main/java/zipkin/internal/Util.java#L105). Once new thrift is updated `traceIdHigh` will be set with `[:16]`

cmd/collector/app/zipkin/zipkin_json.go
```diff
@@ -0,0 +1,234 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
```
black-adder: file can be json.go since it's already namespaced under zipkin.

cmd/collector/app/zipkin/zipkin_json.go
```diff
@@ -0,0 +1,234 @@
...
+	spans, err := decode(body)
+	if err != nil {
+		return nil, err
+	}
+
+	return spansToThrift(spans)
+}
+
+func decode(body []byte) ([]zipkinSpan, error) {
+	type zipkinSpans []zipkinSpan
```
black-adder: is this intermediate type definition necessary?

cmd/collector/app/zipkin/zipkin_json.go
```diff
@@ -0,0 +1,234 @@
...
+		if err != nil {
+			return nil, err
+		}
+		tSpan.BinaryAnnotations = append(tSpan.BinaryAnnotations, binAnno)
+	}
+
+	return tSpan, nil
+}
+
+func endpointToThrift(endp endpoint) (*zipkincore.Endpoint, error) {
```
black-adder: go prefers incomprehensible variable names for short functions: https://talks.golang.org/2014/names.slide#7

so here s/endp/e

cmd/collector/app/zipkin/zipkin_json_test.go
```diff
@@ -0,0 +1,390 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package zipkin
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/stretchr/testify/assert"
```
black-adder: import ordering

cmd/collector/app/zipkin/zipkin_json_test.go
```diff
@@ -0,0 +1,390 @@
...
+}
+
+func TestUnmarshalEndpoint(t *testing.T) {
+	endp := &endpoint{}
+	err := json.Unmarshal([]byte(createEndpoint("foo", "127.0.0.1", "2001:db8::c001", 66)), endp)
+	require.NoError(t, err)
+	assert.Equal(t, "foo", endp.ServiceName)
+	assert.Equal(t, "127.0.0.1", endp.IPv4)
+	assert.Equal(t, "2001:db8::c001", endp.IPv6)
+	// only service name
```
black-adder: this test and a couple of other tests might be slightly more readable with fixtures

cmd/collector/app/zipkin/zipkin_json.go
pavolloffay: it can be defined in global scope, is that better?

black-adder:
> 1. running some configuration of the backend (could be all-in-one, or docker-compose, or k8s)
> 2. having the driver hit some sort of client that can report spans to the backend (either via agent or via direct POST)

This can be easily done with in the all-in-one integration test, however doing it via crossdock tests does seem more semantically correct. The all-in-one integration test is supposed to be a simple test that ensures the pipeline works, I don't want to conflate the test with zipkin testing logic.

Creating a test zipkin client which sends http and thrift spans to the collectors and making this run with the xdock tests seems the best approach. But I'm not entirely sure how to get this to work without creating another docker image just for the zipkin client, I'll have to experiment a bit.

cmd/collector/app/zipkin/zipkin_json.go
black-adder: sgtm

yurishkuro:
btw, their new format with Go: https://github.com/openzipkin/zipkin-api-example/tree/master/go

cmd/collector/app/zipkin/zipkin_json.go
yurishkuro: whaaa?

pavolloffay: _incomprehensible_ :smiley:  then I will rename all local vars to `i`

pavolloffay:
This is updated and should be ready for re-review.

cmd/collector/app/zipkin/json.go
```diff
@@ -0,0 +1,236 @@
...
+
+	return &zipkincore.BinaryAnnotation{
+		Key:            ba.Key,
+		Value:          []byte(ba.Value),
+		Host:           endpoint,
+		AnnotationType: zipkincore.AnnotationType_STRING,
+	}, nil
+}
+
+func parseIpv4(str string) (int32, error) {
```
pavolloffay: it did not work https://paste.fedoraproject.org/paste/mMkVJUks1K6pexIej3XDRA/ https://paste.fedoraproject.org/paste/DwfZ8v4DtYrLKjeb2bNbjw

cmd/collector/app/zipkin/json.go
```diff
@@ -0,0 +1,236 @@
...
+		return 0, errIsNotUnsignedLog
+	}
+
+	start := 0
+	if len > 16 {
+		start = len - 16
+	}
+
+	var result uint64
+	for i := start; i < len; i++ {
```
pavolloffay: it produces different numbers

cmd/collector/app/zipkin/http_handler.go
```diff
@@ -75,30 +75,36 @@ func (aH *APIHandler) saveSpans(w http.ResponseWriter, r *http.Request) {
...
+		}
+	} else if contentType == "application/json" {
+		tSpans, err = deserializeJSON(bodyBytes)
+		if err != nil {
+			http.Error(w, fmt.Sprintf(app.UnableToReadBodyErrFormat, err), http.StatusBadRequest)
+			return
+		}
 	} else {
-		http.Error(w, "Only Content-Type:application/x-thrift is supported at the moment", http.StatusBadRequest)
+		http.Error(w, "Unsupported Content-Type", http.StatusBadRequest)
```
yurishkuro: missing return?

cmd/collector/app/zipkin/http_handler.go
```diff
@@ -75,30 +75,36 @@ func (aH *APIHandler) saveSpans(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if r.Header.Get("Content-Type") == "application/x-thrift" {
-		handleZipkinThrift(aH.zipkinSpansHandler, bodyBytes, w)
+	contentType := r.Header.Get("Content-Type")
+	var tSpans []*zipkincore.Span
+	if contentType == "application/x-thrift" {
+		tSpans, err = deserializeThrift(bodyBytes)
+		if err != nil {
```
yurishkuro: can do this outside of the `if contentType==`

cmd/collector/app/zipkin/http_handler_test.go
```diff
@@ -122,6 +122,46 @@ func TestThriftFormat(t *testing.T) {
...
+	endpErrJSON := createEndpoint("", "127.0.0.A", "", 80)
+
+	// error zipkinSpanHandler
+	handler.zipkinSpansHandler.(*mockZipkinHandler).err = fmt.Errorf("Bad times ahead")
+	tests := []struct {
+		payload    string
+		expected   string
+		statusCode int
+	}{
+		{payload: spanJSON, expected: "Cannot submit Zipkin batch: Bad times ahead\n", statusCode: http.StatusInternalServerError},
```
yurishkuro: nit: this is more readable
```
{
    payload: spanJSON, 
    expected: "Cannot submit Zipkin batch: Bad times ahead\n", 
    statusCode: http.StatusInternalServerError,
},
```

cmd/collector/app/zipkin/json.go
```diff
@@ -0,0 +1,236 @@
...
+	ParentID          string             `json:"parentId,omitempty"`
+	TraceID           string             `json:"traceId"`
+	Name              string             `json:"name"`
+	Timestamp         *int64             `json:"timestamp"`
+	Duration          *int64             `json:"duration"`
+	Debug             bool               `json:"debug"`
+	Annotations       []annotation       `json:"annotations"`
+	BinaryAnnotations []binaryAnnotation `json:"binaryAnnotations"`
+}
+type zipkinSpans []zipkinSpan
```
yurishkuro: only used once, better to inline: `var spans []zipkinSpan`

cmd/collector/app/zipkin/json.go
yurishkuro: ParseIP returns IP, which is an alias to []byte. If nil return 0 else pack bytes into long

cmd/collector/app/zipkin/json.go
yurishkuro: why not use model functions like `model/TraceIDFromString`? 

https://github.com/uber/jaeger/blob/master/model/span.go#L146

It automatically takes care of high/low.

cmd/collector/app/zipkin/json.go
pavolloffay: It was like that before. @black-adder wanted to move it here.

black-adder: I think it's fine to define it here as long as it's used in multiple places (which looks like it can). Every reference to `[]zipkinSpan` can be replaced with `zipkinSpans`. I count at least 3 instances where this happens.

yurishkuro: I prefer to use `[]zipkinSpan` directly, less obfuscation. The type alias is useful when you add extra behavior to it, like we do with `[]model.Tag` (sorting, etc.)

pavolloffay:
@yurishkuro I am on PTO until 28th. I don't want to block other people.. If there are no serious blockers I can submit fix PR to address latest comments:
* IP parsing
* reuse functions from model to handle 128 bit trace ids
* minor improvements in http handler

Once this is in #342 could be merged (if it's ok).

PR #341, opened 2017-08-16
Document zipkin http endpoint

pavolloffay:


docs/deployment.md
```diff
@@ -56,7 +56,8 @@ At default settings the collector exposes the following ports:
 Port  | Protocol | Function
 ----- | -------  | ---
 14267 | TChannel | used by **jaeger-agent** to send spans in jaeger.thrift format
-14268 | HTTP     | can accept spans directly from clients in Jaeger or Zipkin Thrift 
+14268 | HTTP     | can accept spans directly from clients in jaeger.thrift format
+9411  | HTTP     | can accept Zipkin spans in JSON or Thrift (disabled by default)
```
black-adder: do we want to be more precise? ie hostport:/api/v1/spans?

pavolloffay: I think this is fine it uses the same port so users assume that it is the same API and it's mostly for users migrating from zipkin (I guess)

pavolloffay: There is also this http://jaeger.readthedocs.io/en/latest/getting_started/#migrating-from-zipkin. We should probably say there that we have zipkin compatible REST API.

black-adder: +1

PR #342, opened 2017-08-16
Update Zipkin toDomain to handle 128bit traceId and ipv6

pavolloffay:


model/converter/thrift/zipkin/fixtures/jaeger_01.json
```diff
@@ -61,7 +61,7 @@
       }
     },
     {
-      "traceID": "1",
+      "traceID": "20000000000000001",
```
pavolloffay: How can be this split into `lo` and `hi`? 

https://github.com/uber/jaeger/blob/5e5a361babba385bd0b9868bace1e10c75fc0829/model/span.go#L55
https://github.com/uber/jaeger/blob/5e5a361babba385bd0b9868bace1e10c75fc0829/model/span.go#L42-L44

pavolloffay: `traceID: {"hi": 1, "lo": 2}` did not worked

black-adder: checkout https://github.com/uber/jaeger/blob/5e5a361babba385bd0b9868bace1e10c75fc0829/model/span.go#L173 it overwrites the unmarshalling logic so that it can take a 128bit hex string and convert it into high and low

yurishkuro: ot TraceIDFromString

yurishkuro:
done in #349

PR #345, opened 2017-08-18
Implement forfeit lock.

isaach1000:
Implement forfeit method for `distributedlock` interface and plugin.

### DIFF
--- a/plugin/pkg/distributedlock/cassandra/lock.go
+++ b/plugin/pkg/distributedlock/cassandra/lock.go
@@ -40,6 +40,7 @@ const (
 	leasesTable   = `leases`
 	cqlInsertLock = `INSERT INTO ` + leasesTable + ` (name, owner) VALUES (?,?) IF NOT EXISTS USING TTL ?;`
 	cqlUpdateLock = `UPDATE ` + leasesTable + ` USING TTL ? SET owner = ? WHERE name = ? IF owner = ?;`
+	cqlDeleteLock = `DELETE FROM ` + leasesTable + ` WHERE name = ? IF owner = ?;`
 )
 
 var (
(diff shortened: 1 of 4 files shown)


plugin/pkg/distributedlock/cassandra/lock.go
```diff
@@ -78,6 +79,20 @@ func (l *Lock) Acquire(resource string, ttl time.Duration) (bool, error) {
...
+	var name, owner string
+	applied, err := l.session.Query(cqlDeleteLock, resource, l.tenantID).ScanCAS(&name, &owner)
+	if err != nil {
+		return false, errors.Wrap(err, "Failed to forfeit resource lock due to cassandra error")
+	}
+	if applied {
+		// The lock was successfully deleted
+		return true, nil
+	}
+	return false, errors.Wrap(errLockOwnership, "Failed to forfeit resource lock")
```
black-adder: We need a new error, this will return: "This host does not own the resource lock: Failed to foreit resource lock"

isaach1000: Ya I'll fix that.

isaach1000: Actually, I think the format is the reverse: `"Failed to forfeit resource lock: This host does not own the resource lock"`. I followed the pattern I saw for acquire/extending the lock ownership, where a similar error is returned (`"Failed to extend lease on resource lock: This host does not own the resource lock"`). See `expectedErrMsg` in `lock_test.go` for details. If this is wrong, please let me know what would be a better message.

black-adder: You're right, retracting statement

black-adder:
@isaach1000 thanks for the contribution!

isaach1000:
No problem. Thank you for accepting!

Issue #261, opened 2017-07-11
Agent unable to connect to local collector


mnaga:
I have a collector running locally. And am starting a agent with command line options for my collector server/port as below. Couple of questions

1) Am not sure why the service discovery is enabled when I am explicitly passing the collector info
2) How do I met more info on the tchannel error?

$ go run ./cmd/agent/main.go -collector.host-port 127.0.0.1:14267
{"level":"info","ts":1499793930.997451,"caller":"tchannel/bulider.go:95","msg":"Enabling service discovery","service":"jaeger-collector"}
{"level":"info","ts":1499793930.9975188,"caller":"peerlistmgr/peer_list_mgr.go:117","msg":"Registering active peer","peer":"127.0.0.1:14267"}
{"level":"info","ts":1499793930.9978952,"caller":"agent/main.go:52","msg":"Starting agent"}
{"level":"info","ts":1499793931.998254,"caller":"peerlistmgr/peer_list_mgr.go:165","msg":"Not enough connected peers","connected":0,"required":1}
{"level":"info","ts":1499793931.998381,"caller":"peerlistmgr/peer_list_mgr.go:172","msg":"Trying to connect to peer","host:port":"127.0.0.1:14267"}

yurishkuro:
Can you try running all-in-one from source? The components still communicate over the same ports.

```
$ go run ./cmd/standalone/main.go --span-storage.type=memory
{"level":"info","ts":1499795331.6866407,"caller":"tchannel/bulider.go:95","msg":"Enabling service discovery","service":"jaeger-collector"}
{"level":"info","ts":1499795331.6867087,"caller":"peerlistmgr/peer_list_mgr.go:117","msg":"Registering active peer","peer":"127.0.0.1:14267"}
{"level":"info","ts":1499795331.687073,"caller":"standalone/main.go:82","msg":"Starting agent"}
{"level":"info","ts":1499795331.68803,"caller":"standalone/main.go:117","msg":"Starting jaeger-collector TChannel server","port":14267}
{"level":"info","ts":1499795331.6881118,"caller":"standalone/main.go:124","msg":"Starting jaeger-collector HTTP server","http-port":14268}
{"level":"info","ts":1499795331.69117,"caller":"standalone/main.go:174","msg":"Starting jaeger-query HTTP server","port":16686}
[...]

PR #356, opened 2017-08-27
log error message when unable to connect to jaeger-collector

jnewmano:
Additional log messages will help in diagnosing connection issues between the agent and the collector, as seen in #261.

### DIFF
--- a/pkg/discovery/peerlistmgr/peer_list_mgr.go
+++ b/pkg/discovery/peerlistmgr/peer_list_mgr.go
@@ -178,16 +178,12 @@ func (m *PeerListManager) ensureConnections() {
 			continue
 		}
 
-		if conn.IsActive() == false {
-			m.logger.Info("Connection is not active")
-			continue
+		if conn.IsActive() {
+			m.logger.Info("Connected to peer", zap.String("host:port", conn.RemotePeerInfo().HostPort))
+			numConnected++
+			if numConnected >= minPeers {
+				return
+			}
 		}
-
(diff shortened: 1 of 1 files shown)


pkg/discovery/peerlistmgr/peer_list_mgr.go
```diff
@@ -173,12 +173,21 @@ func (m *PeerListManager) ensureConnections() {
...
-			if numConnected >= minPeers {
-				return
-			}
+		if err != nil {
+			m.logger.Error("Unable to connect", zap.String("host:port", peer.HostPort()), zap.Duration("connCheckTimeout", m.connCheckTimeout), zap.Error(err))
+			continue
+		}
+
+		if conn.IsActive() == false {
+			m.logger.Info("Connection is not active")
```
yurishkuro: Is there a way to write a unit test for this condition? Currently it fails code coverage:

https://coveralls.io/builds/13009121/source?filename=pkg%2Fdiscovery%2Fpeerlistmgr%2Fpeer_list_mgr.go

jnewmano: conn.IsActive() == false isn't currently exercised in current tests, which caused the decrease in code coverage.

I will push a new commit that structures the code how it was before, which avoids the decrease in code coverage.

This will leave the case where the connection is inactive without any logging (as it currently is).

yurishkuro:
Thanks, @jnewmano !

PR #363, opened 2017-09-01
Replaced base Docker images for agent/collector/query

jpkrohling:
Tested this with the OpenShift templates and they work as expected. No changes are necessary to the existing OpenShift/Kubernetes templates, so, I don't expect it to break any other client.


jpkrohling:
@kevinearls  , would you test this with a sample application, just as a sanity check?

PR #371, opened 2017-09-02
Make sure when --es.num-replicas=0, ES indices have number_of_replica=0 not 1

### LINKED ISSUE #369: ES replicas defaults to 1 even when specifying --es.num-replicas=0 when starting collector and query
With a local single node ES cluster if I start Jaeger collector and query to use Elasticsearch as backend with following options:
```
-span-storage.type=elasticsearch --es.num-shards=1 --es.num-replicas=0 --es.server-urls=http://jaeger-es:9200 --dependency-storage.type=elasticsearch
```

Upon checking output of localhost:9200/_cluster/health?pretty
```
{
  "cluster_name" : "jaeger-cluster",
  "status" : "yellow",
  "timed_out" : false,
  "number_of_nodes" : 1,
  "number_of_data_nodes" : 1,
  "active_primary_shards" : 4,
  "active_shards" : 4,
  "relocating_shards" : 0,
  "initializing_shards" : 0,
  "unassigned_shards" : 4,
  "delayed_unassigned_shards" : 0,
  "number_of_pending_tasks" : 0,
  "number_of_in_flight_fetch" : 0,
  "task_max_waiting_in_queue_millis" : 0,
  "active_shards_percent_as_number" : 50.0
}
```
ES cluster health check shows status 'yellow' and you will notice active_primary_shards=4 and unassigned_shards=4, which means some indexes are configured with value replicas=1, this can be confirmed by checking output of :

http://localhost:9200/jaeger-span-2017-09-02/_settings?pretty
```
{
  "jaeger-span-2017-09-02" : {
    "settings" : {
      "index" : {
        "mapping" : {
          "nested_fields" : {
            "limit" : "50"
          }
        },
        "number_of_shards" : "1",
        "provided_name" : "jaeger-span-2017-09-02",
        "mapper" : {
          "dynamic" : "false"
        },
        "creation_date" : "1504368267533",
        "requests" : {
          "cache" : {
            "enable" : "true"
          }
        },
        "number_of_replicas" : "1",
        "uuid" : "AkWNKfQSRlCa9pIoxilj5g",
        "version" : {
          "created" : "5050299"
        }
      }
    }
  }
}
```
This confirms that when creating indexes `number_of_repliacas` is getting set to 1, when it should have been set to 0. 
This is a problem when running a single node ES cluster where we want replicas to be set to 0 intentionally.

bharat-p:
Resolves #369 

ES number of replicas was getting set to 1 even when specifying --es.num-replicas=0 when starting collector and query.

**Before fix:**

output of localhost:9200/_cluster/health?pretty
```
{
  "cluster_name" : "jaeger-cluster",
  "status" : "yellow",
  "timed_out" : false,
  "number_of_nodes" : 1,
  "number_of_data_nodes" : 1,
  "active_primary_shards" : 4,
  "active_shards" : 4,
  "relocating_shards" : 0,
  "initializing_shards" : 0,
  "unassigned_shards" : 4,
  "delayed_unassigned_shards" : 0,
  "number_of_pending_tasks" : 0,
  "number_of_in_flight_fetch" : 0,
  "task_max_waiting_in_queue_millis" : 0,
  "active_shards_percent_as_number" : 50.0
}
```
ES cluster health check shows status 'yellow' and you will notice active_primary_shards=4 and unassigned_shards=4, which means some indexes are configured with value replicas=1, this can be confirmed by checking output of :

http://localhost:9200/jaeger-span-2017-09-02/_settings?pretty
```
{
  "jaeger-span-2017-09-02" : {
    "settings" : {
      "index" : {
        "mapping" : {
          "nested_fields" : {
            "limit" : "50"
          }
        },
        "number_of_shards" : "1",
        "provided_name" : "jaeger-span-2017-09-02",
        "mapper" : {
          "dynamic" : "false"
        },
        "creation_date" : "1504368267533",
        "requests" : {
          "cache" : {
            "enable" : "true"
          }
        },
        "number_of_replicas" : "1",
        "uuid" : "AkWNKfQSRlCa9pIoxilj5g",
        "version" : {
          "created" : "5050299"
        }
      }
    }
  }
}
```

**Verified after fix:**
output of http://localhost:9200/_cluster/health?pretty
```
{
  "cluster_name" : "jaeger-cluster",
  "status" : "green",
  "timed_out" : false,
  "number_of_nodes" : 1,
  "number_of_data_nodes" : 1,
  "active_primary_shards" : 10,
  "active_shards" : 10,
  "relocating_shards" : 0,
  "initializing_shards" : 0,
  "unassigned_shards" : 0,
  "delayed_unassigned_shards" : 0,
  "number_of_pending_tasks" : 0,
  "number_of_in_flight_fetch" : 0,
  "task_max_waiting_in_queue_millis" : 0,
  "active_shards_percent_as_number" : 100.0
}
```
After fix output of http://localhost:9200/jaeger-span-2017-09-02/_settings?pretty
```
{
  "jaeger-span-2017-09-02" : {
    "settings" : {
      "index" : {
        "mapping" : {
          "nested_fields" : {
            "limit" : "50"
          }
        },
        "number_of_shards" : "5",
        "provided_name" : "jaeger-span-2017-09-02",
        "mapper" : {
          "dynamic" : "false"
        },
        "creation_date" : "1504369115915",
        "requests" : {
          "cache" : {
            "enable" : "true"
          }
        },
        "number_of_replicas" : "0",
        "uuid" : "Ou2TIOavSCyIDDqlj_4Gxg",
        "version" : {
          "created" : "5050299"
        }
      }
    }
  }
}
```


yurishkuro:
Personally I would rather have a parameter called "replication factor" that defines the total number of copies of the data, similar to Cassandra. Then the setting `es.numReplicas = replicationFactor-1` could be done internally, which would allow RF=0 to be treated as empty value and overwritten by a default, since RF=0 is never valid (while NumReplicas=0 is valid).

I know that this would go against the nomenclature of ES configuration, thus a bit controversial.

bharat-p:
@yurishkuro that should be ok, but as you already pointed out it goes against ES' nomenclature and might be confusing to ES users, also will cause issue with folks who are already relying on this parameter.

yurishkuro:
maybe we can keep the command line switch name & meaning the same but default it to something like `-1` if not specified, and handle the rest internally (like converting internal option to "rf" instead of "numReplicas"). My main concern with this change is that it breaks the convention in Jaeger codebase (and in Go language in general) that empty value (0) can/will be replaced by a default. After this change the value of 0 starts having a meaning rather than being an empty value.

bharat-p:
Another option is we change default value for `es.num-replicas` to `0` instead of current `1`.
cc @mh-park as he is the original author of ES storage plugin for his opinion.

yurishkuro:
There's a new PR #754, but the discussion above was left unresolved.

@pavolloffay Since you were doing work with ES and its configuration, do you have an opinion?

cc @manannayak

PR #376, opened 2017-09-04
Docker images do not push SHA and travis build number tags

pavolloffay:
related to #283 

It also fixes all-in-one image, latest is published per commit and added release tags.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ PASS=$(shell printf "\033[32mPASS\033[0m")
 FAIL=$(shell printf "\033[31mFAIL\033[0m")
 COLORIZE=sed ''/PASS/s//$(PASS)/'' | sed ''/FAIL/s//$(FAIL)/''
 DOCKER_NAMESPACE?=$(USER)
-DOCKER_TAG?=local
+DOCKER_TAG?=latest
 
 .DEFAULT_GOAL := test-and-lint
 
--- a/crossdock/docker-compose.yml
+++ b/crossdock/docker-compose.yml
@@ -48,6 +48,7 @@ services:
 
     test_driver:
         build: .
+        image: jaegertracing/test_driver:latest
         links:
             - cassandra
         depends_on:
--- a/travis/upload-to-docker.sh
+++ b/travis/upload-to-docker.sh
@@ -3,7 +3,7 @@
 set -e
 
 BRANCH=${BRANCH:?'missing BRANCH env var'}
-IMAGE="${REPO:?'missing REPO env var'}:${COMMIT:?'missing COMMIT env var'}"
+IMAGE="${REPO:?'missing REPO env var'}:latest"
 
 unset major minor patch
 if [[ "$BRANCH" == "master" ]]; then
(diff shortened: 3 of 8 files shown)


yurishkuro:
The title is unclear - do you mean to turn off pushing commit & travis build tags? The reason we tag with the commit sha is because sometimes in-progress changes in one project depend on in-progress changes in another one, so integration tests can run by pulling images via SHA rather than `latest`

pavolloffay:
See your last comment in the linked issue. If you now think something different I can bring back commit SHA tags.

yurishkuro:
Right, so as you notice it explicitly mentioned "remove other tags like travis build number and commit hash -- *we might still want those for crossdock images*"

pavolloffay:
I have added back sha images for xdock

crossdock/docker-compose.yml
```diff
@@ -48,6 +48,7 @@ services:
 
     test_driver:
         build: .
+        image: jaegertracing/test_driver:latest
```
yurishkuro: what does this directive do?  We want to make sure we still build from local source

Makefile
```diff
@@ -29,7 +29,7 @@ PASS=$(shell printf "\033[32mPASS\033[0m")
 FAIL=$(shell printf "\033[31mFAIL\033[0m")
 COLORIZE=sed ''/PASS/s//$(PASS)/'' | sed ''/FAIL/s//$(FAIL)/''
 DOCKER_NAMESPACE?=$(USER)
-DOCKER_TAG?=local
+DOCKER_TAG?=latest
```
yurishkuro: this seems unsafe

travis/upload-to-docker.sh
```diff
@@ -3,7 +3,7 @@
 set -e
 
 BRANCH=${BRANCH:?'missing BRANCH env var'}
-IMAGE="${REPO:?'missing REPO env var'}:${COMMIT:?'missing COMMIT env var'}"
+IMAGE="${REPO:?'missing REPO env var'}:latest"
```
yurishkuro: this feels unsafe. Using COMMIT pinpoints the image precisely, while `latest` is kind of vague. What's the harm in keeping COMMIT? Is it because it clutters dockerhub?

travis/upload-to-docker.sh
```diff
@@ -20,21 +20,19 @@ else
...
+    fi
   fi
 fi
 
-# TOOO why are we pushing as $REPO instead of $IMAGE?
-$DOCKER push $REPO
+# Do not enable echo before the `docker login` command to avoid revealing the password.
+set -x
+docker login -u $DOCKER_USER -p $DOCKER_PASS
+# push all tags, therefore push to repo
```
yurishkuro: should we start the whole thing by running some sort of purge on the local repo, to avoid the possibility of some other versions handing around?  Especially if this runs in development.

crossdock/docker-compose.yml
pavolloffay: It uses this name when it builds the image. If you do `docker-compose pull` it will pull that image, not build

travis/upload-to-docker.sh
pavolloffay: >  Is it because it clutters dockerhub? 

basically yes, But I don't have anything against it. Although I haven't seen other projects doing it.

Makefile
pavolloffay: I would rather build latest than local, it's more convenient. 

Note that if you are building images using this Makefile the docker namespace is `$USER` so not really docker hub user. I also assume that anybody who is pushing images is cautious.

travis/upload-to-docker.sh
pavolloffay: we can clear docker registry when starting the CI, but it seems that travis clears registry anyway.

Or we can explicitly push everything:
```
docker push $REPO:latest
docker push $REPO:sha
```
, patch, minor, major would be in the if statements.

travis/upload-to-docker.sh
pavolloffay: Would somebody depend on the `SHA` tag? latest features will be present in the `latest` tag and now we have proper release tags.

Makefile
yurishkuro: The assumption that "anybody who is pushing images is cautious" doesn't comfort me :-), but since it goes under `$USER` namespace it should be fine.

PR #377, opened 2017-09-04
Makefile use  GNU sed and force thrift to use the same user

### LINKED ISSUE #339: Change sed commands to GNU compatible
Currently, `sed` commands in `Makefile` are for `bsd` set we should consider changing them to be `GNU` compatible.

pavolloffay:
fixes #339

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -181,8 +181,8 @@ test_ci: build_examples
 thrift: idl/thrift/jaeger.thrift thrift-image
 	[ -d $(THRIFT_GEN_DIR) ] || mkdir $(THRIFT_GEN_DIR)
 	$(THRIFT) -o /data --gen go:$(THRIFT_GO_ARGS) --out /data/$(THRIFT_GEN_DIR) /data/idl/thrift/agent.thrift
-	sed -i'' -e 's|"zipkincore"|"$(PROJECT_ROOT)/thrift-gen/zipkincore"|g' $(THRIFT_GEN_DIR)/agent/*.go
-	sed -i'' -e 's|"jaeger"|"$(PROJECT_ROOT)/thrift-gen/jaeger"|g' $(THRIFT_GEN_DIR)/agent/*.go
+	sed -i.bak 's|"zipkincore"|"$(PROJECT_ROOT)/thrift-gen/zipkincore"|g' $(THRIFT_GEN_DIR)/agent/*.go
+	sed -i.bak 's|"jaeger"|"$(PROJECT_ROOT)/thrift-gen/jaeger"|g' $(THRIFT_GEN_DIR)/agent/*.go
 	$(THRIFT) -o /data --gen go:$(THRIFT_GO_ARGS) --out /data/$(THRIFT_GEN_DIR) /data/idl/thrift/jaeger.thrift
 	$(THRIFT) -o /data --gen go:$(THRIFT_GO_ARGS) --out /data/$(THRIFT_GEN_DIR) /data/idl/thrift/sampling.thrift
 	$(THRIFT) -o /data --gen go:$(THRIFT_GO_ARGS) --out /data/$(THRIFT_GEN_DIR) /data/idl/thrift/baggage.thrift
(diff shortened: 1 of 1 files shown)


yurishkuro:
I'm getting this:
```
sed -i 's|"zipkincore"|"github.com/uber/jaeger/thrift-gen/zipkincore"|g' thrift-gen/agent/*.go
sed: 1: "thrift-gen/agent/agent.go": undefined label 'hrift-gen/agent/agent.go'
```

pavolloffay:
hmm It worked well with my sed, also with @jpkrohling and @objectiser. 

```
sed (GNU sed) 4.2.2
Copyright (C) 2012 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by Jay Fenlason, Tom Lord, Ken Pizzini,
and Paolo Bonzini.
GNU sed home page: <http://www.gnu.org/software/sed/>.
General help using GNU software: <http://www.gnu.org/gethelp/>.
E-mail bug reports to: <bug-sed@gnu.org>.
Be sure to include the word ``sed'' somewhere in the ``Subject:'' field.
```

yurishkuro:
I'm using the default MacOS sed, which has no version, but its man page says

```
     -i extension
             Edit files in-place, saving backups with the specified extension.  If a zero-length extension is given, no
             backup will be saved.  It is not recommended to give a zero-length extension when in-place editing files, as
             you risk corruption or partial content in situations where disk space is exhausted, etc.

...

STANDARDS
     The sed utility is expected to be a superset of the IEEE Std 1003.2 (``POSIX.2'') specification.

     The -E, -a and -i options are non-standard FreeBSD extensions and may not be available on other operating systems.

HISTORY
     A sed command, written by L. E. McMahon, appeared in Version 7 AT&T UNIX.

AUTHORS
     Diomidis D. Spinellis <dds@FreeBSD.org>

BSD                              May 10, 2005                              BSD
```

basically the `-i` switch requires an argument

pavolloffay:
> The -E, -a and -i options are non-standard FreeBSD extensions and may not be available on other operating systems.

I have added a simple [hack](https://stackoverflow.com/a/14813278/4158442). It would be better to use just the standard GNU version https://stackoverflow.com/a/34815955/4158442.

When I run it, it produces some changes like: 
```diff 
diff --git a/thrift-gen/jaeger/tchan-jaeger.go b/thrift-gen/jaeger/tchan-jaeger.go
index 393b8f1..89e1e9c 100644
--- a/thrift-gen/jaeger/tchan-jaeger.go
+++ b/thrift-gen/jaeger/tchan-jaeger.go
@@ -43,6 +43,10 @@ func (c *tchanCollectorClient) SubmitBatches(ctx thrift.Context, batches []*Batc
        }
        success, err := c.client.Call(ctx, c.thriftService, "submitBatches", &args, &resp)
        if err == nil && !success {
+               switch {
+               default:
+                       err = fmt.Errorf("received no result or unknown exception for submitBatches")
+               }
        }
 
        return resp.GetSuccess(), err
```

Is it fine?

yurishkuro:
don't think the hack quite worked.

when I ran `make thrift`, it did not produce any differences in the existing files, but created three new files

```
	thrift-gen/agent/agent.go-e
	thrift-gen/agent/constants.go-e
	thrift-gen/agent/ttypes.go-e
```

I think the `-e` was interpreted as a suffix.

pavolloffay:
This is bad I will revert to pure GNU sed. Can we just use the standard GNU sed which is available on all systems? It's definitely a good thing for the community. Gnu sed can be easily installed on MacOS.

pavolloffay:
I have tried the last workaroud using `-i.bak` and then remove those files. But this does not prevent that new BSD commands will be added and it will fail for other contributors.

vprithvi:
>This is bad I will revert to pure GNU sed. Can we just use the standard GNU sed which is available on all systems? It's definitely a good thing for the community

I have mixed feelings about it, installing GNU sed is simple using brew, but one needs to finagle the path to make it the default sed that is used for everything. (Also, it would probably break other software that assumes that BSD sed is installed)

Is it too bad to have a conditional on `uname -s`, with the appropriate `sed` commands for a given environment?

pavolloffay:
@vprithvi now it should be compatible with BSD and GNU could you please test it?

vprithvi:
Works for me, could you add a comment on why we are doing these shenanigans with `sed` in the makefile?

yurishkuro:
works for me. I restarted the builds.

Issue #337, opened 2017-08-15
Cannot find service name in Zipkin span


pavolloffay:
When testing #334 I discovered that `brave` (or maybe any zipkin instrumentation)  can send span with just one binary annotation 
```
 {  
      "traceId":"549d6e48d3a5bfcbe4a4fbc1284ff98b",
      "id":"b5e1ead09289b182",
      "name":"",
      "parentId":"e4a4fbc1284ff98b",
      "binaryAnnotations":[  
         {  
            "key":"http.status_code",
            "value":"200",
            "endpoint":{  
               "serviceName":"spring-boot",
               "ipv4":"10.43.17.27"
            }
         }
      ]
   }
```
trace: https://paste.fedoraproject.org/paste/4iUxY0cOkfGpz5jQj-K-lA

It results in `Cannot find service name in Zipkin span [traceID=e4a4fbc1284ff98b, spanID=b5e1ead09289b182]` https://github.com/uber/jaeger/blob/79fba1278743840dff6b84aa059ec58f7ec21a0a/model/converter/thrift/zipkin/to_domain.go#L161-L182

![screenshot of jaeger ui](https://user-images.githubusercontent.com/5618424/29317293-847b81c0-81cb-11e7-8580-2d27f9f248ba.jpg)

Reproducer:
https://github.com/pavolloffay/opentracing-java-examples/tree/jaeger-unknown-service-error-and-ui-issues
run spring-boot and `curl localhost:8080/chaining`, Note that it's not deterministic.

yurishkuro:
yeah, the code checks for service name AND ipv4 being defined, I think the IP should be optional. Having said that, the above JSON seems to have ipv4, so not sure why it's not working.

pavolloffay:
fixed in https://github.com/uber/jaeger/pull/378

PR #378, opened 2017-09-04
Zipkin json and thrift xdock tests

### LINKED ISSUE #181: Create integration tests for zipkin.thrift over http
Add integration tests for zipkin.thrift over http using popular zipkin instrumentation libraries.

pavolloffay:
Fixes #181 #337. 

It uses https://hub.docker.com/r/pavolloffay/xdock-zipkin-brave/ with https://github.com/pavolloffay/xdock-zipkin-brave/pull/1/files. Once this is approved I will move it to jaegertracing.

### DIFF
--- a/crossdock/docker-compose.yml
+++ b/crossdock/docker-compose.yml
@@ -13,7 +13,7 @@ services:
             - zipkin-brave-thrift
 
         environment:
-            - WAIT_FOR=test_driver,go,node,java,python
+            - WAIT_FOR=test_driver,go,node,java,python,zipkin-brave-thrift,zipkin-brave-json
             - WAIT_FOR_TIMEOUT=60s
 
             - CALL_TIMEOUT=60s
--- a/crossdock/docker-compose.yml
+++ b/crossdock/docker-compose.yml
@@ -46,14 +46,14 @@ services:
             - "8080-8082"
 
     zipkin-brave-json:
-        image: pavolloffay/xdock-zipkin-brave
+        image: jaegertracing/xdock-zipkin-brave
         ports:
             - "8080-8081"
         environment:
             - ENCODING=JSON
 
     zipkin-brave-thrift:
-        image: pavolloffay/xdock-zipkin-brave
+        image: jaegertracing/xdock-zipkin-brave
         ports:
             - "8080-8081"
         environment:
--- a/crossdock/docker-compose.yml
+++ b/crossdock/docker-compose.yml
@@ -9,6 +9,8 @@ services:
             - node
             - java
             - python
+            - zipkin-brave-json
+            - zipkin-brave-thrift
 
         environment:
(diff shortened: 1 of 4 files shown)


model/converter/thrift/zipkin/to_domain.go
```diff
@@ -178,6 +178,11 @@ func (td toDomain) findServiceNameAndIP(zSpan *zipkincore.Span) (string, int32,
 			return a.Host.ServiceName, a.Host.Ipv4, nil
 		}
 	}
+	for _, a := range zSpan.BinaryAnnotations {
+		if a.Host != nil && a.Host.ServiceName != "" {
+			return a.Host.ServiceName, a.Host.Ipv4, nil
+		}
+	}
```
yurishkuro: if you're adding this you obviously ran into an issue, but I am curios that kind of binary annotation (aside from LOCAL_COMPONENT already checked for above) could be considered a reliable source of service name/ip? 

NB a comment above this loop would be good explaining.

yurishkuro:
test seems to fail with `dial tcp: lookup zipkin-brave-json on 127.0.0.11:53: no such host` - why such a strange port number, 53?

yurishkuro:
looks great if we can get it to work!

pavolloffay:
> test seems to fail with dial tcp: lookup zipkin-brave-json on 127.0.0.11:53: no such host - why such a strange port number, 53?

It was late I forgot to push the latest image. Now it should work.

model/converter/thrift/zipkin/to_domain.go
pavolloffay: I have added a comment, basically tracer can send a span with just one binary annotation, see also liked issue it happened with correct OT instrumentation and brave-opentracing.

crossdock/docker-compose.yml
```diff
@@ -9,6 +9,8 @@ services:
             - node
             - java
             - python
+            - zipkin-brave-json
+            - zipkin-brave-thrift
 
         environment:
             - WAIT_FOR=test_driver,go,node,java,python
```
black-adder: we should wait for zipkin-brave-json/zipkin just in case (assuming they have a health check endpoint)

model/converter/thrift/zipkin/to_domain.go
```diff
@@ -178,6 +178,12 @@ func (td toDomain) findServiceNameAndIP(zSpan *zipkincore.Span) (string, int32,
 			return a.Host.ServiceName, a.Host.Ipv4, nil
 		}
 	}
+	// Tracer can also report a span witch just binary annotation
```
black-adder: s/witch/with

pavolloffay:
The repo has been moved to https://github.com/jaegertracing/xdock-zipkin-brave

pavolloffay:
this should be ready for re-review/merge.

PR #382, opened 2017-09-06
Remove thrift image when it is done

pavolloffay:


yurishkuro:
what's the objective here?

pavolloffay:
Make thrift leaves a dead container after the run. It should be removed it's not needed.

yurishkuro:
but are you removing the container or the image?  Title says image

pavolloffay:
Wrong title/commit message, it's container

yurishkuro:
cool, thanks

PR #385, opened 2017-09-06
cassandra/spanstore: speed up query by service name

Zariel:
The service_name_index table is orded on disk by start_time DESC so
reads by full partition key will return ordered by start_time. Removing
the order by in the query to list spans significantly improves span
listing performance, observed going from >20s per query to <500ms.

### DIFF
--- a/plugin/storage/cassandra/spanstore/reader.go
+++ b/plugin/storage/cassandra/spanstore/reader.go
@@ -50,7 +50,6 @@ const (
 		SELECT trace_id
 		FROM service_name_index
 		WHERE bucket IN ` + bucketRange + ` AND service_name = ? AND start_time > ? AND start_time < ?
-		ORDER BY start_time DESC
 		LIMIT ?`
 	queryByServiceAndOperationName = `
 		SELECT trace_id


yurishkuro:
```
CREATE TABLE IF NOT EXISTS ${keyspace}.service_name_index (
    service_name      text,
    bucket            int,
    start_time        bigint,
    trace_id          blob,
    PRIMARY KEY ((service_name, bucket), start_time)
) WITH CLUSTERING ORDER BY (start_time DESC)
```

yeah, makes sense.

plugin/storage/cassandra/spanstore/reader.go
```diff
@@ -50,7 +50,6 @@ const (
 		SELECT trace_id
 		FROM service_name_index
 		WHERE bucket IN ` + bucketRange + ` AND service_name = ? AND start_time > ? AND start_time < ?
-		ORDER BY start_time DESC
```
yurishkuro: what about the other queries that do the same?

yurishkuro: I am actually starting to think it's not right to remove the order by. Yes, the records are already ordered on disk, but only within a partition, which includes the bucket number. Bucket numbers are used to avoid hot spots.

Zariel: As far as I understand the query is using the full partition key, I've tested this using the same query here and ensured that all the results are sorted in the correct order

yurishkuro: what I mean is, the partition key is a pair of service_name + bucket:

```
PRIMARY KEY ((service_name, bucket), start_time)
```

The query always runs against all buckets, so for any given service name the query checks 10 different partitions. The results are only sorted within each partition, not across the whole result set, unless we add the order-by clause.

Zariel:
I did some experimentation to simulate whats going on and found that my assumption was incorrect and in fact the following is happening.

Where N is the query LIMIT and a partition is (key, bucket) and clustering key is time

### When not using ORDER BY
Reads N results single partition, data is sorted by time as expected, result is a single partition

### When using ORDER BY
Reads N results for EVERY partition, data is then sorted by time and top N results are returned, result is from every partition.

This surprises me somewhat, I think that when not using ORDER BY Cassandra is returning the first N results which are already sorted within a single partition but not across partitions. If the read goes across multiple partitions then the result is not sorted.

I think the reason why the read performance is so poor for this query is that the coordinator is having to read N * buckets results in memory to return N results which causes heap and gc pressure. Here it may be a better idea to do this sort and top N from the query node and we should see better read performance and better Cassandra heap usage at the expense of more complicated query code.

Tool to check read orders if anyone is interested, https://github.com/Zariel/cassandra-sorted-reads

PR #397, opened 2017-09-14
Speed up queryByServiceName by removing orderBy from query By Service

oibe:


oibe:
@yurishkuro any reason not to do this?

yurishkuro:
this is a duplicate of already closed #385

PR #408, opened 2017-09-19
Define import path in preparation of repository move to jaegertracing

yurishkuro:
per https://github.com/uber/jaeger/issues/409

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 sudo: required
 
 language: go
+go_import_path: github.com/uber/jaeger
 
 dist: trusty
 
--- a/crossdock/services/tracehandler.go
+++ b/crossdock/services/tracehandler.go
@@ -129,10 +129,7 @@ func (h *TraceHandler) runTest(service string, request *traceRequest, tFunc test
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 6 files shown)


crossdock/services/tracehandler.go
```diff
@@ -129,10 +129,7 @@ func (h *TraceHandler) runTest(service string, request *traceRequest, tFunc test
 	if err != nil {
 		return err
 	}
-	if err := vFunc(request, traces); err != nil {
-		return err
-	}
-	return nil
+	return vFunc(request, traces)
```
vprithvi: These seem to be changes unrelated to this PR.

yurishkuro: Because go vet complained

.travis.yml
```diff
@@ -1,6 +1,7 @@
 sudo: required
 
 language: go
+go_import_path: github.com/uber/jaeger
```
black-adder: just so I'm on the same boat, why is this required?

yurishkuro: after the repo move the github path is going to be different, but all the imports in the code refer to uber/jaeger, so it won't  build. This instruction tells Travis to relocate the repo to this path under GOPATH before building.

PR #410, opened 2017-09-20
Upgrade to Go 1.9

yurishkuro:


vprithvi:
`gotestcover` is deprecated and recommends using the following

```
echo 'mode: atomic' > coverage.txt && go list ./... | xargs -n1 -I{} sh -c 'go test -covermode=atomic -coverprofile=coverage.tmp {} && tail -n +2 coverage.tmp >> coverage.txt' && rm coverage.tmp
```

See: https://github.com/pierrre/gotestcover#deprecated

vprithvi:
The branch name should be updated to go-1.9

PR #428, opened 2017-09-25
Fixes a couple of documentation links

craigglennie:
Adds missing URLs for Uber Open Source and Google's Dapper paper.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -7,8 +7,8 @@ Welcome to Jaeger's documentation portal! Below, you'll find information for beg
 If you can't find what you are looking for, or have an issue not covered here, we'd love to hear from you either on [Github](https://github.com/uber/jaeger/issues), [Gitter chat](https://gitter.im/jaegertracing/Lobby), or on our [mailing list](https://groups.google.com/forum/#!forum/jaeger-tracing).
 
 ## About
-Jaeger, inspired by [Dapper][dapper] and [OpenZipkin](http://zipkin.io),
-is a distributed tracing system released as open source by [Uber Technologies][ubeross].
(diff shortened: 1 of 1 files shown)


yurishkuro:
there's nothing wrong with the links, as can be seen here https://github.com/jaegertracing/jaeger

craigglennie:
Huh, that's weird. They're definitely broken sometimes - when I visit https://github.com/jaegertracing/jaeger/blob/master/docs/index.md they render like this

screen shot 2017-09-25 at 12 55 40 pm

yurishkuro:
the docs are hosted here: http://jaeger.readthedocs.org/en/latest/

craigglennie:
They're broken there, too - see the links for Dapper and Uber Technologies under "About" 

screen shot 2017-09-25 at 1 20 37 pm 2

yurishkuro:
you're right. Seems like gitbook does not respect the same syntax as Github's Markdown.

There are other pages that use these references too.

yurishkuro:
cc @vprithvi

yurishkuro:
according to https://toolchain.gitbook.com/syntax/markdown.html it should be supported
```
This is [an example][id] reference-style link.

[id]: http://example.com/  "Optional Title Here"
```

yurishkuro:
oh, I see the problem, the syntax is wrong

yurishkuro:
ok I have a fix

yurishkuro:
#431

yurishkuro:
ok I think the links are fixed

craigglennie:
👍 great! I'll close this.

yurishkuro:
@craigglennie thanks for raising it

PR #442, opened 2017-09-29
Add cassandra tag filter

black-adder:
tag_filter allows the ability to decide which tags we want to index in cassandra. Some tags aren't worth indexing and this will allow you to customize which ones to filter out.

### DIFF
--- a/cmd/collector/app/builder/span_handler_builder.go
+++ b/cmd/collector/app/builder/span_handler_builder.go
@@ -29,7 +29,6 @@ import (
 	cascfg "github.com/uber/jaeger/pkg/cassandra/config"
 	escfg "github.com/uber/jaeger/pkg/es/config"
 	casSpanstore "github.com/uber/jaeger/plugin/storage/cassandra/spanstore"
-	casSpanstoreModel "github.com/uber/jaeger/plugin/storage/cassandra/spanstore/dbmodel"
 	esSpanstore "github.com/uber/jaeger/plugin/storage/es/spanstore"
 	"github.com/uber/jaeger/storage/spanstore"
 )
@@ -96,7 +95,6 @@ func (spanHb *SpanHandlerBuilder) initCassStore(builder cascfg.SessionBuilder) (
 		spanHb.collectorOpts.WriteCacheTTL,
 		spanHb.metricsFactory,
 		spanHb.logger,
-		casSpanstoreModel.DefaultTagFilter(),
 	), nil
 }
 
--- a/cmd/collector/app/builder/span_handler_builder.go
+++ b/cmd/collector/app/builder/span_handler_builder.go
@@ -29,6 +29,7 @@ import (
 	cascfg "github.com/uber/jaeger/pkg/cassandra/config"
 	escfg "github.com/uber/jaeger/pkg/es/config"
 	casSpanstore "github.com/uber/jaeger/plugin/storage/cassandra/spanstore"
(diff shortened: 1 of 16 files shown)


black-adder:
I guess I need to signoff on every commit?

plugin/storage/cassandra/spanstore/dbmodel/tag_filter.go
```diff
@@ -0,0 +1,41 @@
...
+// limitations under the License.
+
+package dbmodel
+
+import (
+	"github.com/uber/jaeger/model"
+)
+
+// FilterTags returns span tags that should be inserted into cassandra.
+type FilterTags func(span *model.Span) []TagInsertion
```
yurishkuro: I am not too psyched with the data dependencies here. Filtering is unrelated to building `TagInsertion`
Why not have the filter only depend on the domain model (e.g. Span and Tag for tags, and Span and Log for logs) and return true/false?

yurishkuro:
@caniszczyk [FOSSA complains about GPL license](https://app.fossa.io/projects/git%2Bgithub.com%2Fjaegertracing%2Fjaeger/refs/branch/master/48d6781a4b3d02fa5f6ba3a83b1b72e086204839/issues/policy_flag/active) in some of the dependencies, is that valid?  Looks like 4 of the come from `spf13/cobra`, which is a library used by Docker, Hugo, etc.

cmd/collector/app/builder/span_handler_builder.go
```diff
@@ -95,6 +96,7 @@ func (spanHb *SpanHandlerBuilder) initCassStore(builder cascfg.SessionBuilder) (
 		spanHb.collectorOpts.WriteCacheTTL,
 		spanHb.metricsFactory,
 		spanHb.logger,
+		casSpanstoreModel.DefaultTagFilter(),
```
yurishkuro: this would be better as a functional option, as the arg list is starting to get too long (and this one is clearly optional).

plugin/storage/cassandra/spanstore/dbmodel/unique_tags.go
```diff
@@ -16,27 +16,26 @@ package dbmodel
...
 import "github.com/uber/jaeger/model"
 
-// GetAllUniqueTags creates a list of all unique tags found in a span and process
-func GetAllUniqueTags(span *model.Span) []TagInsertion {
-	process := span.Process
-	allTags := span.Tags
-	allTags = append(allTags, process.Tags...)
+// GetAllUniqueTags creates a list of all unique tags from a set of filtered tags.
+func GetAllUniqueTags(span *model.Span, tagFilter TagFilter) []TagInsertion {
+	tags := tagFilter.FilterProcessTags(span.Process.Tags)
```
yurishkuro: - allTags was a better name imo (and would cause fewer spurious changes in this diff)
- are you sure you don't need to init to nil? if FilterProcessTags simply returns its arg, and the underlying array had extra capacity, then append might just add to that, creating potential side effects.

black-adder: Just wrote some primitive code: https://play.golang.org/p/5MbPA7Siya looks like the existing process tags doesn't get touched even if it has extra capacity

yurishkuro: https://play.golang.org/p/jI1r1-T9jY

PR #452, opened 2017-10-04
Expose missing port on docker run inside hotrod example

rgamba:
I ran into this error message when following the README instructions on the hotrod example. Exposing the 5775 port fixes the issue.
```
ERROR	log/logger.go:42	write udp 127.0.0.1:63246->127.0.0.1:5775: write: connection refused	{"service": "customer"}
```

### DIFF
--- a/examples/hotrod/README.md
+++ b/examples/hotrod/README.md
@@ -24,7 +24,7 @@ to view the traces. A tutorial / walkthough is available:
 An all-in-one Jaeger backend is packaged as a Docker container with in-memory storage.
 
 ```
-docker run -d -p6831:6831/udp -p16686:16686 jaegertracing/all-in-one:latest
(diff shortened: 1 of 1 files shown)


black-adder:
interesting, port 5775 is used for zipkin span emission which the hotrod example shouldn't be using. Are you by any chance starting up a separate service (python?) and using the all-in-one?

rgamba:
@black-adder nope, no other service using the all in one. First time testing the service so it was pretty much starting from scratch.

black-adder:
gotcha, this might be a bug inside the hot rod demo itself then since we don't actually need to expose that port for the demo to work. I'll take a look at the demo to see what we're missing.

yurishkuro:
I wonder if glide is installing an older version of jaeger client

yurishkuro:
This should be fixed by #470.

PR #453, opened 2017-10-04
Split zipkin span which contains server and client annotations

### LINKED ISSUE #451: Span kind for Zipkin shared spans
Zipkin span can contain both server and client annotations. For this span we should consider adding server and client span kind. 

This fix could help to determine dependency link between services. The current implementation has to use logs in addition to tags.

The current implementation adds only one span kind tag. The decision is based on the order of annotations https://github.com/jaegertracing/jaeger/blob/master/model/converter/thrift/zipkin/to_domain.go#L313 (client annotations should be first)

pavolloffay:

Fixes #451 

Zipkin logic is here: 
Collector: https://github.com/openzipkin/zipkin/blob/5c8a17620674ada895acf05038bfdb209924a47d/zipkin/src/main/java/zipkin/collector/Collector.java#L143
v1 to v2 model: https://github.com/openzipkin/zipkin/blob/eeeeb3c02b0e0c578eb36c08d80cf2bf400067fc/zipkin/src/main/java/zipkin/internal/V2SpanConverter.java#L46

### DIFF
--- a/cmd/collector/app/span_handler.go
+++ b/cmd/collector/app/span_handler.go
@@ -110,10 +110,10 @@ func NewZipkinSpanHandler(logger *zap.Logger, modelHandler SpanProcessor, saniti
 
 // SubmitZipkinBatch records a batch of spans already in Zipkin Thrift format.
 func (h *zipkinSpanHandler) SubmitZipkinBatch(ctx thrift.Context, spans []*zipkincore.Span) ([]*zipkincore.Response, error) {
-	mSpans := make([]*model.Span, len(spans))
-	for i, span := range spans {
+	mSpans := make([]*model.Span, 0, len(spans))
+	for _, span := range spans {
 		sanitized := h.sanitizer.Sanitize(span)
-		mSpans[i] = ConvertZipkinToModel(sanitized, h.logger)
+		mSpans = append(mSpans, convertZipkinToModel(sanitized, h.logger)...)
 	}
 	bools, err := h.modelProcessor.ProcessSpans(mSpans, ZipkinFormatType)
 	if err != nil {
@@ -129,7 +129,7 @@ func (h *zipkinSpanHandler) SubmitZipkinBatch(ctx thrift.Context, spans []*zipki
 }
 
 // ConvertZipkinToModel is a helper function that logs warnings during conversion
-func ConvertZipkinToModel(zSpan *zipkincore.Span, logger *zap.Logger) *model.Span {
+func convertZipkinToModel(zSpan *zipkincore.Span, logger *zap.Logger) []*model.Span {
 	mSpan, err := zipkin.ToDomainSpan(zSpan)
 	if err != nil {
 		logger.Warn("Warning while converting zipkin to domain span", zap.Error(err))
--- a/cmd/collector/app/span_handler.go
+++ b/cmd/collector/app/span_handler.go
@@ -130,9 +130,9 @@ func (h *zipkinSpanHandler) SubmitZipkinBatch(ctx thrift.Context, spans []*zipki
 
 // ConvertZipkinToModel is a helper function that logs warnings during conversion
 func convertZipkinToModel(zSpan *zipkincore.Span, logger *zap.Logger) []*model.Span {
-	mSpan, err := zipkin.ToDomainSpan(zSpan)
+	mSpans, err := zipkin.ToDomainSpan(zSpan)
 	if err != nil {
 		logger.Warn("Warning while converting zipkin to domain span", zap.Error(err))
 	}
-	return mSpan
+	return mSpans
 }
--- a/model/converter/thrift/zipkin/to_domain.go
+++ b/model/converter/thrift/zipkin/to_domain.go
@@ -35,10 +35,10 @@ const (
 
 var (
 	coreAnnotations = map[string]string{
-		zipkincore.CLIENT_RECV: string(ext.SpanKindRPCClientEnum),
-		zipkincore.CLIENT_SEND: string(ext.SpanKindRPCClientEnum),
 		zipkincore.SERVER_RECV: string(ext.SpanKindRPCServerEnum),
 		zipkincore.SERVER_SEND: string(ext.SpanKindRPCServerEnum),
+		zipkincore.CLIENT_RECV: string(ext.SpanKindRPCClientEnum),
+		zipkincore.CLIENT_SEND: string(ext.SpanKindRPCClientEnum),
 	}
 
 	// Some tags on Zipkin spans really describe the process emitting them rather than an individual span.
--- a/model/converter/thrift/zipkin/to_domain.go
+++ b/model/converter/thrift/zipkin/to_domain.go
@@ -158,13 +158,13 @@ func (td toDomain) transformSpan(zSpan *zipkincore.Span) []*model.Span {
 		}
 		// if the first span is a client span we create server span and vice-versa.
 		if result[0].IsRPCClient() {
-			s.Tags = []model.KeyValue{model.String(ext.SpanKindRPCServer.Key, string(ext.SpanKindRPCServer.Value.(ext.SpanKindEnum)))}
+			s.Tags = []model.KeyValue{model.String(string(ext.SpanKind), string(ext.SpanKindRPCServerEnum))}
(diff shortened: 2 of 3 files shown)


model/converter/thrift/zipkin/to_domain.go
```diff
@@ -128,7 +162,20 @@ func (td toDomain) transformSpan(zSpan *zipkincore.Span) *model.Span {
 		Duration:      model.MicrosecondsAsDuration(uint64(zSpan.GetDuration())),
 		Tags:          tags,
 		Logs:          td.getLogs(zSpan.Annotations),
+	}}
+
+	if td.isClientAndServerSpan(zSpan) {
```
pavolloffay: Zipkin impl checks for `cs`,`sr` only.

model/converter/thrift/zipkin/to_domain.go
```diff
@@ -309,7 +356,10 @@ func (td toDomain) getLogFields(annotation *zipkincore.Annotation) []model.KeyVa
 
 func (td toDomain) getSpanKindTag(annotations []*zipkincore.Annotation) (model.KeyValue, bool) {
 	for _, a := range annotations {
-		if spanKind, ok := coreAnnotations[a.Value]; ok {
+		if spanKind, ok := coreClientAnnotations[a.Value]; ok {
```
pavolloffay: I have split core annotations to be sure that client kind is always created first. We could keep annotations in one map and then in `transformSpan` check which one has been transformed and create missing one based on that

yurishkuro: > to be sure that client kind is always created first

why does it matter?

pavolloffay: Then in `transformSpan` I know that only server span kind has to be added.

pavolloffay:
@yurishkuro it should be ready for review

cmd/collector/app/span_handler.go
```diff
@@ -129,7 +129,7 @@ func (h *zipkinSpanHandler) SubmitZipkinBatch(ctx thrift.Context, spans []*zipki
 }
 
 // ConvertZipkinToModel is a helper function that logs warnings during conversion
-func ConvertZipkinToModel(zSpan *zipkincore.Span, logger *zap.Logger) *model.Span {
+func convertZipkinToModel(zSpan *zipkincore.Span, logger *zap.Logger) []*model.Span {
 	mSpan, err := zipkin.ToDomainSpan(zSpan)
```
yurishkuro: s/mSpan/mSpans/

model/converter/thrift/zipkin/to_domain.go
```diff
@@ -76,39 +76,62 @@ func ToDomain(zSpans []*zipkincore.Span) (*model.Trace, error) {
...
+	}
+	return jSpans, err
+}
+
+// isClientAndServerSpan determines whether span represents both client and server processing
+func (td toDomain) isClientAndServerSpan(zSpan *zipkincore.Span) bool {
+	var csOk bool
+	var srOk bool
+	for _, ann := range zSpan.Annotations {
+		if ann.Value == "cs" {
```
yurishkuro: please use constants from zipkincore namespace instead of strings

model/converter/thrift/zipkin/to_domain.go
```diff
@@ -76,39 +76,62 @@ func ToDomain(zSpans []*zipkincore.Span) (*model.Trace, error) {
...
+			jSpan.Process = processes.add(jSpan.Process)
+			trace.Spans = append(trace.Spans, jSpan)
+		}
 	}
 	return trace, multierror.Wrap(errors)
 }
 
-func (td toDomain) ToDomainSpan(zSpan *zipkincore.Span) (*model.Span, error) {
-	jSpan := td.transformSpan(zSpan)
+func (td toDomain) ToDomainSpan(zSpan *zipkincore.Span) ([]*model.Span, error) {
```
yurishkuro: please make the naming consistent across the board: ToDomainSpans (plural)

model/converter/thrift/zipkin/to_domain.go
```diff
@@ -128,7 +152,25 @@ func (td toDomain) transformSpan(zSpan *zipkincore.Span) *model.Span {
...
 		Logs:          td.getLogs(zSpan.Annotations),
+	}}
+
+	if td.isClientAndServerSpan(zSpan) {
+		// if the span is client and server we split it into two separate spans
+		s := &model.Span{
+			TraceID:       model.TraceID{High: uint64(traceIDHigh), Low: uint64(zSpan.TraceID)},
+			SpanID:        model.SpanID(zSpan.ID),
+			OperationName: zSpan.Name,
+			ParentSpanID:  model.SpanID(parentID),
```
yurishkuro: missing Flags, StartTime, Duration?

model/converter/thrift/zipkin/to_domain.go
```diff
@@ -128,7 +152,25 @@ func (td toDomain) transformSpan(zSpan *zipkincore.Span) *model.Span {
...
+		}
+		// if the first span is a client span we create server span and vice-versa.
+		if spanKindTag.VStr == string(ext.SpanKindRPCClient.Value.(ext.SpanKindEnum)) {
+			s.Tags = []model.KeyValue{model.String(ext.SpanKindRPCServer.Key, string(ext.SpanKindRPCServer.Value.(ext.SpanKindEnum)))}
```
yurishkuro: why cast the value `ext.SpanKindRPCServer.Value.(ext.SpanKindEnum)`?

yurishkuro: also, why is the first argument `SpanKindRPCServer` instead of just `SpanKind`?

model/converter/thrift/zipkin/to_domain.go
```diff
@@ -76,39 +76,62 @@ func ToDomain(zSpans []*zipkincore.Span) (*model.Trace, error) {
...
+	}
+	return false
 }
 
 // transformSpan transforms a zipkin span into a Jaeger span
-func (td toDomain) transformSpan(zSpan *zipkincore.Span) *model.Span {
+func (td toDomain) transformSpan(zSpan *zipkincore.Span) []*model.Span {
 	tags := td.getTags(zSpan.BinaryAnnotations, td.isSpanTag)
-	if spanKindTag, ok := td.getSpanKindTag(zSpan.Annotations); ok {
+	var spanKindTag model.KeyValue
```
yurishkuro: I don't think you need to capture this tag, the model span has helper methods like `span.IsRPCClient()`

model/converter/thrift/zipkin/to_domain.go
pavolloffay: >why cast the value ext.SpanKindRPCServer.Value.(ext.SpanKindEnum)? 

because it fails on ` cannot convert ext.SpanKindRPCClient.Value (type interface {}) to type string`

model/converter/thrift/zipkin/to_domain.go
pavolloffay: Added with similar impl to zipkin.

model/converter/thrift/zipkin/to_domain.go
```diff
@@ -118,17 +131,48 @@ func (td toDomain) transformSpan(zSpan *zipkincore.Span) *model.Span {
...
+		s := &model.Span{
+			TraceID:       model.TraceID{High: uint64(traceIDHigh), Low: uint64(zSpan.TraceID)},
+			SpanID:        model.SpanID(zSpan.ID),
+			OperationName: zSpan.Name,
+			ParentSpanID:  model.SpanID(parentID),
+			Flags:         flags,
+		}
+		// if the first span is a client span we create server span and vice-versa.
+		if result[0].IsRPCClient() {
+			s.Tags = []model.KeyValue{model.String(ext.SpanKindRPCServer.Key, string(ext.SpanKindRPCServer.Value.(ext.SpanKindEnum)))}
```
yurishkuro: I think this is a lot simpler: `model.String(string(ext.SpanKind), string(ext.SpanKindRPCServerEnum))`

pavolloffay: +1 changed

PR #454, opened 2017-10-05
Trace the dedicated query server

rbtcollins:
The standalone query server already has tracing enabled. This is a
trivial parity patch.

### DIFF
--- a/cmd/query/main.go
+++ b/cmd/query/main.go
@@ -24,6 +24,7 @@ import (
 	"github.com/gorilla/mux"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
+	jaegerClientConfig "github.com/uber/jaeger-client-go/config"
 	"github.com/uber/jaeger-lib/metrics/go-kit"
 	"github.com/uber/jaeger-lib/metrics/go-kit/expvar"
 	"go.uber.org/zap"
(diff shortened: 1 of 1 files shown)


cmd/query/main.go
```diff
@@ -65,6 +66,18 @@ func main() {
 
 			metricsFactory := xkit.Wrap("jaeger-query", expvar.NewFactory(10))
 
+			tracer, closer, err := jaegerClientConfig.Configuration{
+				Sampler: &jaegerClientConfig.SamplerConfig{
+					Type:  "probabilistic",
```
yurishkuro: use `const` instead of probabilistic, it's cheaper.

yurishkuro: actually, I think this needs to be overridable vie cmd line switches

rbtcollins: This will be via https://github.com/uber/jaeger-client-go/issues/206 no? I think we should have one set of variables for all the Jaeger bindings, so that its really easy to configure jaeger enabled components in one consistent fashion. That hasn't been done yet though.

I can switch this to const easily - I think its worth noting though that this is the code thats in the all-in-one entry point. Should I change that at the same time?

Given that this is a very simple and relatively low-use UI: scales O(developers) not O(users), I think a const of 1 is fine to start with.

Issue #448, opened 2017-10-02
clockskew.go adjuster should skip FollowsFrom spans


black-adder:
Right now, followsFrom spans are incorrectly clock skew adjusted to fit inside the parent

bogatuadrian:
I opened a PR at https://github.com/jaegertracing/jaeger/pull/460 that tries to fix this issue. I'm not sure if it's the right approach, but I'm up to make any needed changes requested in the review.

bogatuadrian:
At least is it possible to add an option to exclude the clock skew adjuster from the list of adjusters at startup? Or, maybe even better, to add that to the API request?

yurishkuro:
There is a feature in the query API that we're missing to return the trace without any kind of adjustments, which is especially useful for looking at raw JSON.

yurishkuro:
> to return the trace without any kind of adjustments

#617

jpkrohling:
Is this still relevant?

yurishkuro:
Yes

PR #460, opened 2017-10-06
Fix FollowsFrom reference clock skew adjuster

bogatuadrian:
As stated by @black-adder in https://github.com/jaegertracing/jaeger/issues/448, the clockskew adjuster does not handle spans that reference the parent span using FollowsFrom reference. This PR tries to fix that by considering spans that have that reference as root spans. This way, the span will not be adjusted (specifically will be adjusted with a skew with `delta = 0`).

Note: if you find the use of labels ugly, please comment and I will change it to the alternative approach.

### DIFF
--- a/model/adjuster/clockskew.go
+++ b/model/adjuster/clockskew.go
@@ -113,12 +113,21 @@ func (a *clockSkewAdjuster) buildNodesMap() {
 // or points to an ID for which there is no span.
 func (a *clockSkewAdjuster) buildSubGraphs() {
 	a.roots = make(map[model.SpanID]*node)
+IterateSpans:
 	for _, n := range a.spans {
-		// TODO handle FOLLOWS_FROM references
 		if n.span.ParentSpanID == 0 {
 			a.roots[n.span.SpanID] = n
 			continue
 		}
+
+		// Treat spans that have a FOLLOWS_FROM reference to their parent as root spans.
+		for _, ref := range n.span.References {
+			if ref.RefType == model.FollowsFrom && ref.SpanID == n.span.ParentSpanID {
+				a.roots[n.span.SpanID] = n
+				continue IterateSpans
+			}
+		}
+
 		if p, ok := a.spans[n.span.ParentSpanID]; ok {
 			p.children = append(p.children, n)
 		} else {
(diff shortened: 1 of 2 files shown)


model/adjuster/clockskew.go
```diff
@@ -113,12 +113,21 @@ func (a *clockSkewAdjuster) buildNodesMap() {
...
 	for _, n := range a.spans {
-		// TODO handle FOLLOWS_FROM references
 		if n.span.ParentSpanID == 0 {
 			a.roots[n.span.SpanID] = n
 			continue
 		}
+
+		// Treat spans that have a FOLLOWS_FROM reference to their parent as root spans.
+		for _, ref := range n.span.References {
+			if ref.RefType == model.FollowsFrom && ref.SpanID == n.span.ParentSpanID {
```
yurishkuro: what if the References list has both ChildOf and FollowsFrom?

bogatuadrian: Both referencing the parent? Well, first of all this should not happen, but if it does and there is at least one FollowsFrom to the parent I think the time should not be adjusted.

yurishkuro: > first of all this should not happen

I don't know about that, there can be all kinds of DAGs people might want to model. My point is that the original implementation of the clock skew was ignoring the FollowsFrom altogether. If we want to support it properly, then the graph may need to be normalized better. I.e. it should be looking at all references, not just jump to a conclusion when we see one specific type. And it should not be relying on the parentId anywhere, only on references.

Parent ID is actually an optimization detail for the cases where there's a single ChildOf reference to that parent - the pattern occurs in the majority of spans, and representing it with a single ID is more efficient than with a ChildOf reference. But the processing should really be based on the references (which might require an enrichment step where add a ChildOf reference if we see `parentId != 0 and len(references) == 0` - this is what we're doing before returning JSON to the UI, we could probably refactor the enrichment chain to do that before the clockskew).

nziebart: does this change actually work? just because we add it to the set of roots doesn't mean we will not adjust it. it's still a part of the sub-DAG for some other root.

nziebart: what if we consider the span as a root `iff` all of its references are `FOLLOWS_FROM`? Otherwise, we only add it as a child for parents where there is a `CHILD_OF` relationship?

PR #461, opened 2017-10-08
gocql: use NewNativeType instead of unsafe hacks

Zariel:
Use gocql.NewNativeType instead of unsafe hacks to produce a NativeType

Signed-off-by: Chris Bannister <c.bannister@gmail.com>

### DIFF
--- a/pkg/cassandra/gocql/testutils/udt.go
+++ b/pkg/cassandra/gocql/testutils/udt.go
@@ -16,7 +16,6 @@ package testutils
 
 import (
 	"testing"
-	"unsafe"
 
 	"github.com/gocql/gocql"
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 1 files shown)


yurishkuro:
The Travis build says `pkg/cassandra/gocql/testutils/udt.go:45: undefined: gocql.NewNativeType`

We're pinning to an older version of gocql driver. Internally we have one installation with an old version of Cassandra, but should be decommissioning it in a couple of months. Let's keep this PR open for now.

mlaccetti:
@yurishkuro Any word on when the upgrade to a newer `gocql` will happen, with health checks?

yurishkuro:
sorry, we have not prioritized the upgrade of gocql yet. We tried it once internally and it resulted in much worse performance, so we put it on hold.

isaachier:
Is this already done?

black-adder:
Yes, gocql driver was updated. This can be closed

PR #462, opened 2017-10-09
Support configuration files

pavolloffay:
Support configuration files. This allows us to configure Jaeger with kubernetes config maps. 

By default configuration files are disabled. Lists are currently not handled because we parse them manually. For instance `cassandra.servers` requires a string with comma separated values. 

Follows some examples. Note that flags with dots can be written as nested objects.
```json
{
  "query": {
    "prefix": "prefix",
    "static-files": "static"
  },
  "cassandra": {
    "keyspace": "from_file",
    "password": "pas_from_file",
    "servers": "localhost,localhost2",
    "archive": {
      "keyspace": "archive_keyspace"
    }
  },
  "es.archive.max-span-age": "75h0m0s"
}
```
```yml
query:
  prefix: prefix
  static-files: static
cassandra:
  keyspace: from_file
  password: pas_from_file
  servers: localhost,localhost2
  archive:
    keyspace: archive_keyspace
es.archive.max-span-age: 75h0m0s
```

### DIFF
--- a/cmd/collector/main.go
+++ b/cmd/collector/main.go
@@ -132,7 +132,7 @@ func main() {
 	config.AddFlags(
 		v,
 		command,
-		flags.AddConfFileFlag,
+		flags.AddConfigFileFlag,
 		flags.AddFlags,
 		builder.AddFlags,
 		casOptions.AddFlags,
--- a/cmd/collector/main.go
+++ b/cmd/collector/main.go
@@ -15,6 +15,7 @@
 package main
 
 import (
+	"fmt"
 	"net"
 	"net/http"
 	"os"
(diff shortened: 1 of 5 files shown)


cmd/collector/main.go
```diff
@@ -130,6 +139,7 @@ func main() {
 	config.AddFlags(
 		v,
 		command,
+		flags.AddConfFileFlag,
 		flags.AddFlags,
```
yurishkuro: any reason not to have config flag added as part of AddFlags? It would probably also avoid duplicating the code in L65-71 in all binaries, only do it in `sFlags := new(flags.SharedFlags).InitFromViper(v)`

pavolloffay: I did a split for two reasons:

1. agent does use `flags.AddFlags`
2. `flags.AddFlags` can be specified in config file, so there would be twice `sFlags := new(flags.SharedFlags).InitFromViper(v)`

yurishkuro: I see - makes sense. In that case can we move L65-71 into that package too, so that we can have just a one liner in the mains?  E.g. something like `flags.TryProcessExternalConfig(v, logger)`

yurishkuro: also, by doing that we can make `flags.ExternalConfFlags` type private. I suggest renaming it to `ConfigFile(Flag)` ("external" is non-informative)

cmd/flags/flags.go
```diff
@@ -36,8 +37,25 @@ const (
...
+	flagSet.String(configFile, "", "Configuration file in JSON, TOML, YAML, HCL, or Java properties formats (default none). See spf13/viper for precedence.")
+}
+
+// TryLoadConfigFile initializes viper with config file specified as flag
+func TryLoadConfigFile(v *viper.Viper, logger *zap.Logger) {
+	if file := v.GetString(configFile); file != "" {
+		v.SetConfigFile(file)
+		err := v.ReadInConfig()
+		if err != nil {
+			logger.Fatal(fmt.Sprintf("Fatal error config file: %s \n", file), zap.Error(err))
```
yurishkuro: Nit: don't use string formating. File name can be passed as a string field. Message could be "error loading config".

cmd/query/main.go
```diff
@@ -123,6 +125,7 @@ func main() {
 	config.AddFlags(
 		v,
 		command,
+		flags.AddConfFileFlag,
```
yurishkuro: Nit: s/conf/config for making consistency

PR #482, opened 2017-10-18
replace "github.com/olivere/elastic" with "gopkg.in/olivere/elastic.v5"  

NeoCN:
change import `github.com/olivere/elastic` to `gopkg.in/olivere/elastic.v5`

the code in `elastic` package import sub package like `gopkg.in/olivere/elastic.v5/uritemplates`、`gopkg.in/olivere/elastic.v5/config`, so, unify the import to avoid get the same dependency in two different path and vendor twice


yurishkuro:
afaik the use of `gopkg.in` is not recommended. I checked the repo github.com/olivere/elastic, it uses semver for releases, so we get no benefits from using gopkg.in

If we are using the import inconsistently, then I'd rather switch everything to use github.com/olivere/elastic, not the gopkg.in

NeoCN:
@yurishkuro the code under `github.com/olivere/elastic` import `gopkg.in/olivere/elastic.v5` make the dependency graph looks as following graph
jaeger-deps
Indeed, A and B is the same one under different path.

> If we are using the import inconsistently, then I'd rather switch everything to use github.com/olivere/elastic, not the gopkg.in

this must change the code under `github.com/olivere/elastic` which not maintained by jaeger project.

there is no problem that keep the current code unchanged

Issue #445, opened 2017-10-02
create.sh fails when the DC name contains a hyphen.


rbtcollins:
The generated CQL is invalid - I'm not sure whether its missing quoting, or just unusable.

Example output:
```
system_traces  system_schema  system_auth  system  system_distributed

Generating the schema for the keyspace jaeger_v1_DC1-K8Demo and datacenter DC1-K8Demo
Using template file /cassandra-schema/v001.cql.tmpl with parameters:
    mode = prod
    datacenter = DC1-K8Demo
    keyspace = jaeger_v1_DC1-K8Demo
    replication = {'class': 'NetworkTopologyStrategy', 'DC1-K8Demo': '2' }
    trace_ttl = 172800
    dependencies_ttl = 0
<stdin>:3:SyntaxException: line 1:43 mismatched input '-' expecting K_WITH (... KEYSPACE IF NOT EXISTS jaeger_v1_DC1[-]K8Demo...)
<stdin>:13:SyntaxException: line 1:39 mismatched input '-' expecting '(' (... TYPE IF NOT EXISTS jaeger_v1_DC1[-]K8Demo...)
<stdin>:18:SyntaxException: line 1:39 mismatched input '-' expecting '(' (... TYPE IF NOT EXISTS jaeger_v1_DC1[-]K8Demo...)
<stdin>:24:SyntaxException: line 1:39 mismatched input '-' expecting '(' (... TYPE IF NOT EXISTS jaeger_v1_DC1[-]K8Demo...)
<stdin>:29:SyntaxException: line 1:39 mismatched input '-' expecting '(' (... TYPE IF NOT EXISTS jaeger_v1_DC1[-]K8Demo...)
<stdin>:54:SyntaxException: line 1:40 no viable alternative at input '-' (CREATE TABLE IF NOT EXISTS [jaeger_v1_DC1]-...)
<stdin>:68:SyntaxException: line 1:40 no viable alternative at input '-' (CREATE TABLE IF NOT EXISTS [jaeger_v1_DC1]-...)
<stdin>:83:SyntaxException: line 1:40 no viable alternative at input '-' (CREATE TABLE IF NOT EXISTS [jaeger_v1_DC1]-...)
<stdin>:100:SyntaxException: line 1:40 no viable alternative at input '-' (CREATE TABLE IF NOT EXISTS [jaeger_v1_DC1]-...)
[...]

PR #487, opened 2017-10-22
 passes error message if KEYSPACE does not have valid character

dexter816:
if cassandra fails because of using invalid character the user will now
know why, thus can make changes to KEYSPACE parameter

fixes: #445
Signed-off-by: Deepika Upadhyay <deepikaupadhyay01@gmail.com>


yurishkuro:
Thanks!

dexter816:
\o/ Yay ! Thanks for helping ! :)

PR #488, opened 2017-10-23
Handle ports > 32k in Zipkin JSON

### LINKED ISSUE #467: Zipkin v1 endpoint uses int16 for port, cannot represent ports > 32k
My docker network granted me ports in the range of 56000, which used as ports in endpoint cannot be unmarshalled by the Zipkin Endpoint, because it defines int16 for port field.

yurishkuro:
Resolves #467

### DIFF
--- a/cmd/collector/app/zipkin/http_handler_test.go
+++ b/cmd/collector/app/zipkin/http_handler_test.go
@@ -93,7 +93,7 @@ func waitForSpans(t *testing.T, handler *mockZipkinHandler, expecting int) {
 			t.Error("never received a span")
 			return
 		}
-		if want, have := expecting, len(handler.getSpans()); want != have {
+		if have := len(handler.getSpans()); expecting != have {
 			time.Sleep(time.Millisecond)
 			continue
 		}
--- a/cmd/collector/app/zipkin/http_handler_test.go
+++ b/cmd/collector/app/zipkin/http_handler_test.go
@@ -83,20 +83,23 @@ func TestViaClient(t *testing.T) {
 
 	tracer.StartSpan("root").Finish()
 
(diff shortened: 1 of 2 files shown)


cmd/collector/app/zipkin/http_handler_test.go
```diff
@@ -83,20 +83,23 @@ func TestViaClient(t *testing.T) {
...
+
+func waitForSpans(t *testing.T, handler *mockZipkinHandler, expecting int) {
 	deadline := time.Now().Add(2 * time.Second)
 	for {
 		if time.Now().After(deadline) {
 			t.Error("never received a span")
 			return
 		}
-		if want, have := 1, len(handler.zipkinSpansHandler.(*mockZipkinHandler).getSpans()); want != have {
+		if want, have := expecting, len(handler.getSpans()); want != have {
```
black-adder: is "want" necessary? just do 

`if have := ....; expected != have {`

cmd/collector/app/zipkin/json.go
```diff
@@ -147,10 +147,15 @@ func endpointToThrift(e endpoint) (*zipkincore.Endpoint, error) {
 	if err != nil {
 		return nil, err
 	}
+	port := e.Port
+	if port >= (1 << 15) {
+		// Zipkin.thrift defines port as i16, so values between (2^16 and 2^16-1) must be encoded as negative
```
black-adder: 2^15 and 2^16-1?

yurishkuro: what's the question?

yurishkuro: oh, yeah

Issue #489, opened 2017-10-23
Query service: /api/config returning index.html


tiffon:
**Description**

The query-service is returning the body of `index.html` when the UI requests `/api/config`.

This is causing errors in the browser, but fortunately they are not interrupting the UI experience.

The ideal scenario is the config is dumped into the body of `index.html` when it is served. This will allow us to remove the HTTP call to `/api/config` entirely. 

Per a previous discussion, the `index.html` can contain the following, which I believe can be replaced with a JSON serialization of the config:

```
JAEGER_CONFIG = DEFAULT_CONFIG;
```

Where `DEFAULT_CONFIG` is replaced with the UI config contents.

The UI config contents are `{}` by default, and can be overridden via a envar or cmd line parameter (maybe `query.ui-config`?) that conveys a filepath with a YAML or JSON file.

**Steps to reproduce the issue:**
1. Follow the steps [here](http://jaeger.readthedocs.io/en/latest/getting_started/#sample-application)
2. Create a trace via the HotROD app
3. View a trace

**Describe the results you received:**

`/api/config` returns the `index.html` content.

**Describe the results you expected:**

Either `/api/config` returns the `index.html` content or we switch to dumping the config into the body of the HTML.

yurishkuro:
Let's make a change in the UI project first that ensures that the generated index.html contains
```
DEFAULT_CONFIG = {};
JAEGER_CONFIG = DEFAULT_CONFIG;
```
Then we can add code to the query service to assign a different value to JAEGER_CONFIG when serving index.html

tiffon:
Sounds good, will post back when it's merged.

yurishkuro:
UI change: https://github.com/jaegertracing/jaeger-ui/pull/107

yurishkuro:
All done & tested, keeping this open until we add docs on how to use it.

pavolloffay:
https://github.com/jaegertracing/jaeger-ui could list all configurable options.

yurishkuro:
@pavolloffay elaborate? Like a help page?

yurishkuro:
or you mean the README?  +1 then

pavolloffay:
just the readme for now

yurishkuro:
I think we have one outstanding issue (https://github.com/jaegertracing/jaeger-ui/issues/123) before we can document (although the current solution already works, so we _could_ just document it as is - https://github.com/jaegertracing/jaeger-ui/issues/124).

[...]

PR #490, opened 2017-10-24
Allow embedding custom UI config in index.html

yurishkuro:
For #489

```
$ cat yy.yaml
x: abcd
z:
  - a
  - b

$ go run ./cmd/standalone/main.go --span-storage.type=memory --query.ui-config yy.yaml
```

replaces

```
<!-- JAEGER_CONFIG = DEFAULT_CONFIG; -->
```

in `index.html` with

```
<!-- JAEGER_CONFIG = {"x":"abcd","z":["a","b"]}; -->
```

TODO:
  * [x] tests
  * [x] UI version that includes the config var in `index.html`


cmd/query/app/fixture/index.html
```diff
@@ -2,4 +2,5 @@
 <html lang="en">
     <meta charset="UTF-8">
     <title>Test Page</title>
+    <!-- JAEGER_CONFIG=DEFAULT_CONFIG; -->
```
black-adder: change this to: "JAEGER_CONFIG = DEFAULT_CONFIG;" to be a more accurate fixture?

cmd/query/app/static_handler.go
```diff
@@ -27,22 +34,66 @@ const (
...
+		return nil, nil
+	}
+	ext := filepath.Ext(uiConfig)
+	bytes, err := ioutil.ReadFile(uiConfig)
+	if err != nil {
+		return nil, errors.Wrapf(err, "Cannot read UI config file %v", uiConfig)
+	}
+
+	var c map[string]interface{}
+	var unmarshall func([]byte, interface{}) error
```
black-adder: s/unmarshall/unmarshal

cmd/query/app/fixture/index.html
yurishkuro: how is it more accurate?  The regexp ignores the whitespace.

cmd/query/app/static_handler.go
```diff
@@ -27,22 +33,67 @@ const (
...
 	if staticAssetsRoot == "" {
 		staticAssetsRoot = defaultStaticAssetsRoot
 	}
 	if !strings.HasSuffix(staticAssetsRoot, "/") {
 		staticAssetsRoot = staticAssetsRoot + "/"
 	}
-	return &StaticAssetsHandler{staticAssetsRoot: staticAssetsRoot}
+	indexBytes, err := ioutil.ReadFile(staticAssetsRoot + "index.html")
+	if err != nil {
+		return nil, errors.Wrap(err, "Cannot read UI static assets")
```
pavolloffay: Now it fails without UI. Before it was working. Workaround is to create empty `index.html` or do not include static handler if flag is empty 

https://github.com/jaegertracing/jaeger/pull/493 uses query as a docker container, when xdock is executed it builds all images, Now we have to build UI which takes significant time.

yurishkuro: let's discuss on #497

Issue #469, opened 2017-10-12
Docker compose for deployment of individual components


jpkrohling:
See #466 for context and https://github.com/jaegertracing/jaeger-kubernetes/blob/master/production/jaeger-production-template.yml for reference

This should be done in a separate repository, like there is for Kubernetes and OpenShift.

jpkrohling:
See #306 for a starting point.

ishanjain28:
Hi, I'd like to take this up if it is still available.

allieoop:
Note: `jaeger-kubernetes/production/jaeger-production-template.yml` was renamed to `jaeger-kubernetes/production/cassandra.yml` (see https://github.com/jaegertracing/jaeger-kubernetes/commit/f930d0ab75bd042bdcc539acefd81f7fc9992016#diff-0b3583916142302afe2c114d88c1f7c1) - Issue description with this reference needs to be updated.

allieoop:
@ishanjain28 are you interested in working together on this?

I was just wondering what name we should give to the new repo. Maybe something like `jaeger-compose`?

ishanjain28:
@allieoop Yeah, Sure. `jaeger-compose` sounds good to me.

pavolloffay:
@ishanjain28 see #493

allieoop:
If I'm understanding this correctly, it sounds like the goal of this issue is to move the docker compose file into a separate repo so that it can be used to support multiple storages. I see the following question asked in Issue 493 (https://github.com/jaegertracing/jaeger/pull/493#issuecomment-339730092): do we want to move the docker compose file into a separate repo? But I don't see any further discussion. If anyone has any thoughts on why this should be done, it would be helpful to discuss it here in the issue to move the compose file out of this repo and into a new one, like what storages do we want to support that we aren't currently supporting or anything else?

pavolloffay:
>  it sounds like the goal of this issue is to move the docker compose file into a separate repo so that it can be used to support multiple storages.

Compose file can be here on in a different repo and still support different storages.

I think it's easier to have it directly here because:
*  it's used in crossdock - it's automatically tested
* can more quickly adopt new changes

cc @jpkrohling @yurishkuro

pavolloffay:
We can reuse this compose file in different jaeger repositories to run xdock tests:

```bash
docker-compose -f https://github.com/jaegertracing/jaeger/master/compose.yml -f $(XDOCK_YAML) run crossdock 
```

[...]

PR #493, opened 2017-10-25
docker-compose and reuse it in crossdock

pavolloffay:
* Docker-compose yaml with cassandra deployment. 
* The compose file is reused in crossdock. 
* `test-driver` is not starting jaeger services

Tests are not passing because xdock containers send tracing data to `test-driver` and not `jaeger-*`.

If we can use docker compose 3.3 (it requires docker engine 17.06.0+) then we can make this even more generic -- split this compose file and support multiple storages:
* one compose file to define only jaeger services with enabled `--config-file=`
* one compose file per storage. compose file would define config map for all jaeger services.

This is related to #469 #405


Makefile
```diff
@@ -31,7 +31,7 @@ THRIFT_GEN_DIR=thrift-gen
 PASS=$(shell printf "\033[32mPASS\033[0m")
 FAIL=$(shell printf "\033[31mFAIL\033[0m")
 COLORIZE=$(SED) ''/PASS/s//$(PASS)/'' | $(SED) ''/FAIL/s//$(FAIL)/''
-DOCKER_NAMESPACE?=$(USER)
+DOCKER_NAMESPACE?=jaegertracing
```
pavolloffay: Setting the repository to the correct one. I added user input to docker push to avoid unwanted pushes.

yurishkuro:
> test-driver is not starting jaeger services

awesome!

Makefile
jpkrohling: This was on purpose, so that we don't push accidentally to the *official* repository. Just pass `DOCKER_NAMESPACE` before the `make` command to override it:

```
DOCKER_NAMESPACE=jaegertracing make
```

Or, like this:
https://github.com/jaegertracing/jaeger/blob/f0193c63aefb9be980b639ea59d612ad47b78ba7/travis/build-docker-images.sh#L23

Makefile
```diff
@@ -117,44 +117,45 @@ build-collector-linux:
...
 .PHONY: docker
-docker: build_ui build-agent-linux build-collector-linux build-query-linux docker-images-only
+docker: build_ui build-agent-linux build-collector-linux build-query-linux build-crossdock-linux docker-images-only
 
 .PHONY: docker-images-only
 docker-images-only:
-	cp -r jaeger-ui-build/build/ cmd/query/jaeger-ui-build
 	docker build -t $(DOCKER_NAMESPACE)/jaeger-cassandra-schema:${DOCKER_TAG} plugin/storage/cassandra/
 	@echo "Finished building jaeger-cassandra-schema =============="
+	cp -r jaeger-ui-build/build/ cmd/query/jaeger-ui-build
```
jpkrohling: It's easier for this command to fail than the `jaeger-cassandra-schema`, so, keeping it as the first will make it fail "faster". Or were you having problems with this order of actions?

Makefile
```diff
@@ -117,44 +117,45 @@ build-collector-linux:
...
 		docker build -t $(DOCKER_NAMESPACE)/jaeger-$$component:${DOCKER_TAG} cmd/$$component ; \
 		echo "Finished building $$component ==============" ; \
 	done
 	rm -rf cmd/query/jaeger-ui-build
+	docker build -t $(DOCKER_NAMESPACE)/test-driver:${DOCKER_TAG} crossdock/
+	@echo "Finished building test-driver ==============" ; \
 
 .PHONY: docker-push
 docker-push:
+	@while [ -z "$$CONTINUE" ]; do \
```
jpkrohling: Will this work in non-interactive mode? Like, on Travis?

Makefile
pavolloffay: It makes sense, I just moved it closer to the place where it is used.

Makefile
pavolloffay: I have to find out :). 

This target at the moment is not used on travis. However I would like to simplify travis config a bit and use travis stages and have deploy stage which would consistently deploy all images (potentially using this command)

Makefile
pavolloffay: >This was on purpose, so that we don't push accidentally to the official repository. 

I know, because of this I have added user input/verification before docker push.

This change is because now xdock uses only docker images so it has to build. We could use different image names/repository in docker-compose but that seems more odd.

pavolloffay: This should work `CONTINUE=y make docker-push`

pavolloffay:
The question also is if we want to have compose file in this repository or move it to a separate.

crossdock/docker-compose.yml
```diff
@@ -27,48 +30,49 @@ services:
         image: jaegertracing/xdock-go
         ports:
             - "8080-8082"
+        links:
+          - "jaeger-agent:test_driver"
```
pavolloffay: Temporary fix, xdock images expect jaeger services on this host.

pavolloffay:
It should be ready for the review.

cmd/standalone/main.go
```diff
@@ -221,12 +221,16 @@ func startQuery(
...
 		queryApp.HandlerOptions.Tracer(tracer))
+	r := mux.NewRouter()
 	staticHandler, err := queryApp.NewStaticAssetsHandler(qOpts.StaticAssets, qOpts.UIConfig)
 	if err != nil {
 		logger.Fatal("Could not create static assets handler", zap.Error(err))
 	}
-	r := mux.NewRouter()
-	apiHandler.RegisterRoutes(r)
+	if staticHandler != nil {
+		apiHandler.RegisterRoutes(r)
```
yurishkuro: this is confusing. Better to group related code (and use scopes to hide vars that's not needed later)
```
r := mux.NewRouter()

{ 
    h := New...
    h.RegisterRoutes(r)
}

if h, err := queryApp.NewStaticAssetsHandler(qOpts.StaticAssets, qOpts.UIConfig); err != nil {
    logger.Fatal("Could not create static assets handler", zap.Error(err))
} else if h != nil {
    h.RegisterRoutes(r)
} else {
    logger.Info("Static handler is not registered")
}
```

pavolloffay: I have added a function which handles static handler registration

yurishkuro:
@pavolloffay nice! there are 3 different tickets mentioned here, can you close those that are resolved by this pr?

yurishkuro:
@pavolloffay could you take a look at jaeger-client-go builds? Crossdock tests are now failing there, I suspect it might be related to this change

https://travis-ci.org/jaegertracing/jaeger-client-go/jobs/295424981

pavolloffay:
Yes I will, all xdocs has to be updated. Sorry I could have been more proactive here.

PR #496, opened 2017-10-27
Zipkin json binAnnotation value number

pavolloffay:
Fix deserialization of zipkin binary annotations values of number type. cc @golonzovsky 

The following json fails to deserialize with this error `Unable to process request body: json: cannot unmarshal number into Go value of type string`

```json
[
{
    "traceId": "7d3277b087a7c05a",
    "name": "/",
    "id": "3c1813d224db60fe",
    "parentId": "a50ac2ce7f3bf056",
    "timestamp": 1508924339988597,
    "duration": 1781,
    "annotations": [],
    "binaryAnnotations": [
      {
        "key": "http.status_code",
        "value": 200,
        "type": "I64"
      }
    ]
  }
]
```

### DIFF
--- a/cmd/collector/app/zipkin/json.go
+++ b/cmd/collector/app/zipkin/json.go
@@ -20,6 +20,7 @@ import (
 	"encoding/binary"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"math"
 	"strconv"
 	"strings"
@@ -186,10 +187,6 @@ func binAnnoToThrift(ba binaryAnnotation) (*zipkincore.BinaryAnnotation, error) 
 		return nil, err
 	}
 
-	if ba.Type == "" {
-		ba.Type = "STRING"
-	}
-
 	var val []byte
 	var valType zipkincore.AnnotationType
 	switch ba.Type {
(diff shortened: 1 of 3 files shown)


cmd/collector/app/zipkin/json.go
```diff
@@ -37,7 +37,7 @@ type annotation struct {
 type binaryAnnotation struct {
 	Endpoint endpoint `json:"endpoint"`
 	Key      string   `json:"key"`
-	Value    string   `json:"value"`
+	Value    json.Number   `json:"value"`
```
pavolloffay: This feels hacky, I am using this https://stackoverflow.com/a/24480906/4158442. 

At the moment we also do not parse type.

pavolloffay: @yurishkuro what about using `json.RawMessage` instead? it's basically `[]byte`

yurishkuro: it's harder to parse, up to you. Does zipkin allow other unquoted values like `true`?

pavolloffay: `json.RawMessage` keeps `"`  there `"value": "str"` is `"str"`

pavolloffay: we should also base64 decode if the type is bytes. I will add type deserialization and some logic for it. Do not merge yet

adriancole:
>
> it's harder to parse, up to you. Does zipkin allow other unquoted values like true?
>
> While zipkin accepts v1 format with nuanced types like I32, they aren't
supported moving forward (v2). In many cases they will end up converted to
strings.

pavolloffay:
it should be ready for re-review

yurishkuro:
@pavolloffay was there an issue for this?

pavolloffay:
No, the issue was reported only on gitter.

PR #500, opened 2017-10-30
Zipkin use net.parseIP and model functions for IDs

pavolloffay:

### DIFF
--- a/cmd/collector/app/zipkin/json.go
+++ b/cmd/collector/app/zipkin/json.go
@@ -92,8 +92,7 @@ func spansToThrift(spans []zipkinSpan) ([]*zipkincore.Span, error) {
 }
 
 func spanToThrift(s zipkinSpan) (*zipkincore.Span, error) {
-	// id can be padded with zeros
-	id, err := model.SpanIDFromString(s.ID[:16])
+	id, err := model.SpanIDFromString(cutLongID(s.ID))
 	if err != nil {
 		return nil, err
 	}
@@ -116,7 +115,7 @@ func spanToThrift(s zipkinSpan) (*zipkincore.Span, error) {
 	}
 
 	if len(s.ParentID) > 0 {
-		parentID, err := model.SpanIDFromString(s.ParentID[:16])
+		parentID, err := model.SpanIDFromString(cutLongID(s.ParentID))
 		if err != nil {
 			return nil, err
 		}
@@ -140,6 +139,15 @@ func spanToThrift(s zipkinSpan) (*zipkincore.Span, error) {
 	}
 
 	return tSpan, nil
+}
+
+// id can be padded with zeros. We let it fail later in case it's longer than 32
+func cutLongID(id string) string {
+	l := len(id)
+	if l > 16 && l <= 32 {
+		return id[:16]
+	}
+	return id
 }
 
 func endpointToThrift(e endpoint) (*zipkincore.Endpoint, error) {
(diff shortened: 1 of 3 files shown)


cmd/collector/app/zipkin/json.go
```diff
@@ -149,6 +141,15 @@ func spanToThrift(s zipkinSpan) (*zipkincore.Span, error) {
 	return tSpan, nil
 }
 
+// id can be padded with zeros. We let it fail later in case it's longer than 32
+func cutLongID(id string) string {
+	l := len(id)
+	if l > 16 && l <= 32 {
+		return id[:16]
```
yurishkuro: afaik zipkin does not support span IDs longer than 64bit, so what's the use case for truncating?

cmd/collector/app/zipkin/json.go
```diff
@@ -248,49 +249,20 @@ func float64bytes(float float64) []byte {
...
 func parseIpv4(str string) (int32, error) {
-	// TODO use net.ParseIP
-	segments := strings.Split(str, ".")
-	if len(segments) == 1 {
+	if str == "" {
 		return 0, nil
 	}
 
+	ip := net.ParseIP(str).To4()
+	if ip == nil {
```
yurishkuro: note that zipkin model currently supports ipv6 as well

cmd/collector/app/zipkin/json.go
pavolloffay: This is not used for trace ids. 

There can be span id `1` padded with 16 zeros.

cmd/collector/app/zipkin/json.go
pavolloffay: At the moment we construct that as `Ipv6: []byte(e.IPv6)` Which I guess is not ideal.

cmd/collector/app/zipkin/json.go
yurishkuro: > There can be span id 1 padded with 16 zeros.

do you mean for a total of 17 digits? Doesn't sound like it would be valid for zipkin.

pavolloffay: Try this:
```json
[
  {
    "traceId": "7d3277b087a7c05a",
    "name": "/",
    "id": "000000000000000001",
    "timestamp": 1508924339988597,
    "duration": 1781,
    "annotations": [],
    "binaryAnnotations": [
      {
        "key": "lc",
        "value": "nginx",
        "endpoint": {
          "ipv4": "",
          "port": 0,
          "serviceName": "nginx"
        }
      }
    ]
  }
]
```

zipkin consumes it without problems. The ID parsing function what we had here before was exact impl from zipkin and there were tests for these types of ids.

yurishkuro: I still think it's invalid ID - what if it was padded with 1s instead of 0s? By truncating it you are essentially reading a different number.

If you only want to truncate the 0s, you may want to (a) optionally check that the chopped off characters are all 0s, and (b) parse the right-most 16 digits `id[l-16:]`.

pavolloffay: It certainly works in zipkin. Here is the zipkin code https://github.com/openzipkin/zipkin/blob/eeeeb3c02b0e0c578eb36c08d80cf2bf400067fc/zipkin/src/main/java/zipkin/internal/Util.java#L101

yurishkuro: you still need (b), zipkin does it too: `int beginIndex = length > 16 ? length - 16 : 0;`

pavolloffay: I did this
```
func cutLongID(id string) string {
 +	l := len(id)
 +	if l > 16 && l <= 32 {
 +		start := l - 16
 +		return id[start:]
 +	}
 +	return id
 +}
```

PR #504, opened 2017-10-31
use the now exported expvarHandler https://github.com/golang/go/issue…

dstroot:
This is simply some code cleanup and simplification.  Prior to Go 1.8 the expvarHandler was not exposed.  So you had to copy the code into your project.  You no longer have to do that.

This makes the example hotrod code simpler and easier to follow. Downside it requires Go 1.8 or above, which should be OK now that we are at 1.9.2 right?

See: https://github.com/golang/go/issues/15030


yurishkuro:
love it! please sign your commit, e.g. `git commit -s --amend; git push --force`

examples/hotrod/services/route/server.go
```diff
@@ -59,7 +59,7 @@ func (s *Server) Run() error {
 func (s *Server) createServeMux() http.Handler {
 	mux := tracing.NewServeMux(s.tracer)
 	mux.Handle("/route", http.HandlerFunc(s.route))
-	mux.Handle("/debug/vars", http.HandlerFunc(httpexpvar.Handler))
+	mux.Handle("/debug/vars", expvar.Handler()) // use exported handler
```
yurishkuro: note - this is not the only place the handler was used
```
CGO_ENABLED=0 GOOS=linux installsuffix=cgo go build -o ./cmd/agent/agent-linux ./cmd/agent/main.go
pkg/metrics/builder.go:28:2: cannot find package "github.com/uber/jaeger/examples/hotrod/pkg/httpexpvar" in any of:
	/home/travis/gopath/src/github.com/uber/jaeger/vendor/github.com/uber/jaeger/examples/hotrod/pkg/httpexpvar (vendor tree)
	/home/travis/.gimme/versions/go1.9.linux.amd64/src/github.com/uber/jaeger/examples/hotrod/pkg/httpexpvar (from $GOROOT)
	/home/travis/gopath/src/github.com/uber/jaeger/examples/hotrod/pkg/httpexpvar (from $GOPATH)
```

dstroot: OK - my bad - I am on it.  👍

dstroot: OK - as far as I can tell that *is* the only place in the code.  However I forked `jaeger-tracing/jaeger' and this is still looking back at `uber/jaeger`.  I think I need to fork the repo from Uber and change it there and submit a PR.  Does that sound right to you Yuri?  If so you can just cancel this one.

Or, this can be a bigger PR where I change the imports over to `jaeger-tracing/jaeger' for hotrod...

yurishkuro: jaegertracing/jaeger is the new github location of uber/jaeger, the old location is redirected here. Your current PR is fine as is, you just need to fix another place:
```
./pkg/metrics/builder.go:28:	"github.com/uber/jaeger/examples/hotrod/pkg/httpexpvar"
```

dstroot: Hmm... this is likely an issue with the automated build tooling.  My change eliminates the need for that package (and thus the package is missing) in my PR.  That was kind of the point of my PR - that package was only needed prior to Go 1.8.

yurishkuro: yes, but your change doesn't touch the file `./pkg/metrics/builder.go` which still imports the old package. It needs to change similar to how this file is changed

dstroot: Ah!  Sorry I misunderstood.  I thought that package was only part of Hotrod!  Changes made and signed.  Tests passed.

yurishkuro:
Please make sure all commits are signed (or squash them into one)

examples/hotrod/services/route/server.go
dstroot: OK - commits squashed and all changes made.  Sorry - thought this would be a quick easy change.

dstroot:
No - it was all my fault.  I was not 100% familiar with DCO, and I made an assumption the Hotrod example packages were only used in the Hotrod example. That was totally on me.  I LOVE the example by the way.  I am trying to use open tracing and Jaeger now in my projects.  Cheers!

PR #505, opened 2017-11-01
Xdock compose, test driver links jaeger services

pavolloffay:
A minor improvement. `test-driver` should depend on jaeger services and not xdock.

### DIFF
--- a/crossdock/docker-compose.yml
+++ b/crossdock/docker-compose.yml
@@ -4,9 +4,6 @@ services:
     crossdock:
         image: crossdock/crossdock
         links:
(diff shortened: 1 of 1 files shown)


crossdock/docker-compose.yml
```diff
@@ -76,6 +73,10 @@ services:
         image: jaegertracing/test-driver
         ports:
             - "8080"
+        links:
```
pavolloffay: U think it's even better to use `depends_on` container should be started in a separate network and reachable by name.

PR #514, opened 2017-11-06
Introduce logging level configuration

pavolloffay:
Logging struct was defined in flag package but not used. I have also removed unused code from there.

### DIFF
--- a/cmd/collector/main.go
+++ b/cmd/collector/main.go
@@ -15,6 +15,7 @@
 package main
 
 import (
+	"fmt"
 	"net"
 	"net/http"
 	"os"
@@ -50,7 +51,6 @@ func main() {
 	var signalsChannel = make(chan os.Signal, 0)
 	signal.Notify(signalsChannel, os.Interrupt, syscall.SIGTERM)
 
-	logger, _ := zap.NewProduction()
 	serviceName := "jaeger-collector"
 	casOptions := casFlags.NewOptions("cassandra")
 	esOptions := esFlags.NewOptions("es")
@@ -61,10 +61,18 @@ func main() {
 		Short: "Jaeger collector receives and processes traces from Jaeger agents and clients",
 		Long: `Jaeger collector receives traces from Jaeger agents and agent and runs them through
 				a processing pipeline.`,
-		Run: func(cmd *cobra.Command, args []string) {
-			flags.TryLoadConfigFile(v, logger)
+		RunE: func(cmd *cobra.Command, args []string) error {
+			err := flags.TryLoadConfigFile(v)
+			if err != nil {
+				return err
+			}
 
 			sFlags := new(flags.SharedFlags).InitFromViper(v)
+			logger, err := sFlags.NewLogger(zap.NewProductionConfig())
+			if err != nil {
+				return err
+			}
+
 			casOptions.InitFromViper(v)
 			esOptions.InitFromViper(v)
 
@@ -127,6 +135,7 @@ func main() {
 			case <-signalsChannel:
 				logger.Info("Jaeger Collector is finishing")
 			}
+			select {}
 		},
 	}
 
@@ -143,7 +152,8 @@ func main() {
 	)
 
 	if error := command.Execute(); error != nil {
-		logger.Fatal(error.Error())
+		fmt.Println(error.Error())
+		os.Exit(1)
 	}
 }
 
--- a/cmd/collector/main.go
+++ b/cmd/collector/main.go
@@ -135,7 +135,7 @@ func main() {
 			case <-signalsChannel:
 				logger.Info("Jaeger Collector is finishing")
 			}
-			select {}
+			return nil
 		},
 	}
 
--- a/cmd/flags/flags.go
+++ b/cmd/flags/flags.go
@@ -15,14 +15,14 @@
 package flags
 
 import (
-	"errors"
 	"flag"
 	"fmt"
-	"strings"
 	"time"
 
+	"github.com/pkg/errors"
 	"github.com/spf13/viper"
 	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
 )
 
 const (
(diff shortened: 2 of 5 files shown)


cmd/collector/main.go
```diff
@@ -126,6 +134,7 @@ func main() {
 			case <-signalsChannel:
 				logger.Info("Jaeger Collector is finishing")
 			}
+			select {}
```
yurishkuro: why is this needed?

cmd/flags/flags.go
```diff
@@ -33,7 +33,7 @@ const (
 	// ESStorageType is the storage type flag denoting an ElasticSearch backing store
 	ESStorageType                  = "elasticsearch"
 	spanStorageType                = "span-storage.type"
-	logLevel                       = "log-level"
+	logLevel                       = "log-Level"
```
yurishkuro: shouldn't it be lower case?

cmd/query/main.go
```diff
@@ -118,6 +126,7 @@ func main() {
 			case <-serverChannel:
 				logger.Info("Jaeger Query is finishing")
 			}
+			select {}
```
yurishkuro: ?

cmd/flags/flags.go
```diff
@@ -60,19 +72,22 @@ type SharedFlags struct {
...
 	flags.DependencyStorage.DataFrequency = v.GetDuration(dependencyStorageDataFrequency)
+	flags.Logging.Level = v.GetString(logLevel)
 	return flags
 }
 
 // AddFlags adds flags for SharedFlags
 func AddFlags(flagSet *flag.FlagSet) {
 	flagSet.String(spanStorageType, CassandraStorageType, fmt.Sprintf("The type of span storage backend to use, options are currently [%v,%v,%v]", CassandraStorageType, ESStorageType, MemoryStorageType))
-	flagSet.String(logLevel, "info", "Minimal allowed log level")
+	flagSet.String(logLevel, "info", "Minimal allowed log Level. For more levels see zap logger documentation")
```
yurishkuro: instead of `zap logger documentation` I'd print the url

cmd/flags/flags.go
```diff
@@ -44,14 +44,26 @@ func AddConfigFileFlag(flagSet *flag.FlagSet) {
...
 		if err != nil {
-			logger.Fatal("Error loading config file", zap.Error(err), zap.String(configFile, file))
+			errors.Wrap(err, fmt.Sprintf("Error loading config file %s", file))
 		}
 	}
+	return nil
+}
+
+// NewLogger pareses log Level
+func (flags *SharedFlags) NewLogger(conf zap.Config, options ...zap.Option) (*zap.Logger, error) {
```
yurishkuro: it looks like in all cases you're passing zap.Production() as the config. Is there a use case you're thinking of when it's not going to be production?  Also, do we need a way to affect that config from cmd line / our config?

cmd/flags/flags.go
```diff
@@ -44,14 +44,26 @@ func AddConfigFileFlag(flagSet *flag.FlagSet) {
...
 		err := v.ReadInConfig()
 		if err != nil {
-			logger.Fatal("Error loading config file", zap.Error(err), zap.String(configFile, file))
+			errors.Wrap(err, fmt.Sprintf("Error loading config file %s", file))
 		}
 	}
+	return nil
+}
+
+// NewLogger pareses log Level
```
yurishkuro: typo "pareses"

cmd/flags/flags.go
```diff
@@ -88,19 +103,5 @@ type spanStorage struct {
...
-	MaxRetryAttempts   int
-	QueryTimeout       time.Duration
-	Servers            string
-	Port               int
-	Keyspace           string
-}
-
-func (co cassandraOptions) ServerList() []string {
-	return strings.Split(co.Servers, ",")
-}
```
yurishkuro: nice catch

cmd/query/main.go
```diff
@@ -132,7 +141,8 @@ func main() {
 	)
 
 	if error := command.Execute(); error != nil {
-		logger.Fatal(error.Error())
+		fmt.Println(error.Error())
```
yurishkuro: this is unfortunate, because a logger may print the error quite differently, e.g. with a stack trace, write to file, report to Sentry, etc. If we guaranteed that the command only returned errors when it can't init the logger, then it wouldn't be too bad, but I don't think this is generally the case.

cmd/flags/flags.go
pavolloffay: added but it looks a bit weird. 

Message for conf file: 
```
Configuration file in JSON, TOML, YAML, HCL, or Java properties formats (default none). See spf13/viper for precedence.
```

cmd/flags/flags.go
pavolloffay: > Is there a use case you're thinking of when it's not going to be production? Also, do we need a way to affect that config from cmd line / our config?

Not right now. Maybe a debug flag or something like that in the future. I made this as flexible as possible. 

So do you want to change it to the following?
```go
func (flags *SharedFlags) NewLogger() (*zap.Logger, error) 
```

cmd/query/main.go
pavolloffay: This statement is invoked when: 
* an error is propagated from `runE` -> in all cases except logger creation and config file parsing `zap.Fatal` is called which directly calls `os.exit`.
* cobra fails to parse flags e.g. a wrong/undefined flag

Other solution might be to use a default logger here.

cmd/flags/flags.go
```diff
@@ -44,14 +44,26 @@ func AddConfigFileFlag(flagSet *flag.FlagSet) {
...
 
 // TryLoadConfigFile initializes viper with config file specified as flag
-func TryLoadConfigFile(v *viper.Viper, logger *zap.Logger) {
+func TryLoadConfigFile(v *viper.Viper) error {
 	if file := v.GetString(configFile); file != "" {
 		v.SetConfigFile(file)
 		err := v.ReadInConfig()
 		if err != nil {
-			logger.Fatal("Error loading config file", zap.Error(err), zap.String(configFile, file))
+			errors.Wrap(err, fmt.Sprintf("Error loading config file %s", file))
```
yurishkuro: missing `return`?  And you can use `Wrapf` for extra formatting.

pavolloffay: :+1: ouch, the change from log Fatal :)

pavolloffay:
@yurishkuro could you please re-review?

yurishkuro:
@pavolloffay could you please rebase from master to make sure all is clean? Then ok to merge.

pavolloffay:
@yurishkuro thanks, rebased, I will merge on green

PR #515, opened 2017-11-06
Add info on UI configuration to docs

tiffon:
Add information on UI configuration to the deployment docs.


black-adder:
I'll land since the broken tests seem to be crossdock-java related

Issue #273, opened 2017-07-16
Enable some form of metrics from Jaeger backend processes


yurishkuro:
At the moment even though all Jaeger processes are instrumented for metrics, the binaries do not enable them by default. Options are:
  * enable expvar metrics by default (at /debug/vars endpoints)
  * enable Prometheus metrics by default
  * make the type of metrics configurable (expvar, prometheus, whatever other types are supported by go-kit)

At the moment we're missing a factory for Prometheus metrics, either need to implement it or wait for go-kit's new "provider" interface. Implementing in manually should be very straightforward.

yurishkuro:
Prometheus metrics are now default in all Jaeger components.

TODO:
  * [ ] review & normalize metric names similar to https://github.com/jaegertracing/jaeger-client-go/pull/222
  * [ ] update documentation on where the metrics are exposed in each component

yurishkuro:
Closing since both TODOs have their own issues

PR #516, opened 2017-11-07
Support Prometheus metrics as default mode

### LINKED ISSUE #360: Use prometheus as metrics backend in jaeger agent
I am runing agent separately, I try to use prometheus as metrics backend. Like this:
**go run cmd/agent/main.go --metrics-backend prometheus**
But i got an error
**panic:  duplicate metrics collector registration attempted.**

Is there any extra configuration for prometheus? Or prometheus is not supported for now?

yurishkuro:
Depends in https://github.com/jaegertracing/jaeger-lib/pull/29

Resolves #360, partially #273 .

### DIFF
--- a/cmd/collector/app/builder/span_handler_builder.go
+++ b/cmd/collector/app/builder/span_handler_builder.go
@@ -116,7 +116,7 @@ func (spanHb *SpanHandlerBuilder) initElasticStore(esBuilder escfg.ClientBuilder
 // BuildHandlers builds span handlers (Zipkin, Jaeger)
 func (spanHb *SpanHandlerBuilder) BuildHandlers() (app.ZipkinSpansHandler, app.JaegerBatchesHandler) {
 	hostname, _ := os.Hostname()
-	hostMetrics := spanHb.metricsFactory.Namespace(hostname, nil)
+	hostMetrics := spanHb.metricsFactory.Namespace("", map[string]string{"host": hostname})
 
 	zSanitizer := zs.NewChainedSanitizer(
 		zs.NewSpanDurationSanitizer(),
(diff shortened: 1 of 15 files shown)


cmd/collector/app/builder/span_handler_builder.go
```diff
@@ -116,7 +116,7 @@ func (spanHb *SpanHandlerBuilder) initElasticStore(esBuilder escfg.ClientBuilder
 // BuildHandlers builds span handlers (Zipkin, Jaeger)
 func (spanHb *SpanHandlerBuilder) BuildHandlers() (app.ZipkinSpansHandler, app.JaegerBatchesHandler) {
 	hostname, _ := os.Hostname()
-	hostMetrics := spanHb.metricsFactory.Namespace(hostname, nil)
+	hostMetrics := spanHb.metricsFactory.Namespace("", map[string]string{"host": hostname})
```
yurishkuro: NOTE: this is a breaking change, but without it the crossdock test was failing with `"49d962b96482:save_latency" is not a valid metric name` because the host name was used as a prefix.

pkg/metrics/builder.go
```diff
@@ -18,23 +18,22 @@ import (
...
 	"net/http"
 
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/spf13/viper"
 	xkit "github.com/uber/jaeger-lib/metrics/go-kit"
 	kitexpvar "github.com/uber/jaeger-lib/metrics/go-kit/expvar"
-	kitprom "github.com/uber/jaeger-lib/metrics/go-kit/prometheus"
+	jprom "github.com/uber/jaeger-lib/metrics/prometheus"
 
 	"github.com/uber/jaeger-lib/metrics"
```
black-adder: imports for this is wonky

PR #519, opened 2017-11-08
[DONT REVIEW] Add priority channels to bounded queue

black-adder:
The bounded queue should allow items to be prioritized so that less important items (spans) are dropped.

- [ ] tests

### DIFF
--- a/pkg/queue/bounded_queue.go
+++ b/pkg/queue/bounded_queue.go
@@ -75,15 +75,11 @@ func (q *BoundedQueue) StartConsumers(num int, consumer func(item interface{})) 
 					for _, items := range q.items {
 						select {
 						case item := <-items:
+							atomic.AddInt32(&q.size, -1)
 							consumer(item)
 							break
 						default:
 						}
-						// if item, ok := <-items; ok {
-						// atomic.AddInt32(&q.size, -1)
-						// consumer(item)
-						// break
-						// }
 					}
 				}
 			}
--- a/pkg/queue/bounded_queue.go
+++ b/pkg/queue/bounded_queue.go
@@ -22,21 +22,26 @@ import (
 	"github.com/uber/jaeger-lib/metrics"
 )
 
+const (
+	totalPriorities = 2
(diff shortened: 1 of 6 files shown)


pkg/queue/bounded_queue.go
```diff
@@ -60,11 +69,16 @@ func (q *BoundedQueue) StartConsumers(num int, consumer func(item interface{}))
...
 			for {
 				select {
-				case item := <-q.items:
-					atomic.AddInt32(&q.size, -1)
-					consumer(item)
 				case <-q.stopCh:
 					return
+				default:
+					for _, items := range q.items {
+						if item, ok := <-items; ok {
```
black-adder: this doesn't work as intended, it blocks forever. Back to the drawing board

pkg/queue/bounded_queue.go
```diff
@@ -60,11 +69,18 @@ func (q *BoundedQueue) StartConsumers(num int, consumer func(item interface{}))
...
 				select {
-				case item := <-q.items:
-					atomic.AddInt32(&q.size, -1)
-					consumer(item)
 				case <-q.stopCh:
 					return
+				default:
+					for _, items := range q.items {
+						select {
+						case item := <-items:
```
black-adder: sooooo, this iterates over the priority queue and starting at the highest priority channel, checks if there's an item. If there is, it consumes it. This repeats for all channels.

it feels wrong

yurishkuro: It is wrong - hot cycle, no blocking.

black-adder: i'm gonna try using a priority queue but I feel like the all the extra syncing required will kill performance vs hot cycling.

I'll benchmark both and give you the numbers

yurishkuro: > all the extra syncing required

this is a common misconception - do you think channels are free from syncing? There are many cases where a simple mutex is a lot more efficient than achieving the same functionality with channels.

PR #523, opened 2017-11-09
Provide a more honest view into the size of the collector queue

kwojcik:
Update the queue size Gauge when it gets full, and do the periodic update less often so that it's less likely to overwrite an update from a full queue.

Signed-off-by: Kevin Wojcik kwojcik@uber.com

### DIFF
--- a/cmd/collector/app/span_processor.go
+++ b/cmd/collector/app/span_processor.go
@@ -57,7 +57,7 @@ func NewSpanProcessor(
 		sp.processItemFromQueue(value)
 	})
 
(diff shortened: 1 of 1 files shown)


kwojcik:
Signed-off-by: Kevin Wojcik <kwojcik@uber.com>

yurishkuro:
Hm, I am not a big fan of reporting _less often_.

I was actually thinking of changing jaeger-lib/metrics to support Histogram instead of Timer. It would allow using Histogram for queue depth.

PR #525, opened 2017-11-10
Update README and docs with new C++ client lib.

hypnoce:
Hi, 

with the new [C++ client](https://github.com/jaegertracing/cpp-client) merged, maybe it would be nice to expose it into the document materials of Jaeger.
Let me know if you find this useful or maybe premature.
I also wanted to update [this image](https://github.com/jaegertracing/jaeger/blob/master/docs/images/architecture.png) by adding C++ but I could not find the source.

Thanks !


yurishkuro:
Thanks!

PR #526, opened 2017-11-10
Added HTTP reporter with support for sending auth credentials

jpkrohling:


jpkrohling:
This PR adds auth support for the Agent when communicating with the Collector. It does *not* add auth support for communication between the Client (Tracer) and the Agent.

At the current state, I need a few clarifications, specially regarding the sampling strategies and baggage restriction features. It does not look like they are fully implemented and I didn't find them exposed via regular HTTP endpoint, so, I did not implement this part.

It also requires some documentation, but basically, this is how one should start the Agent to send an OAuth token with every request:

```
go run cmd/agent/main.go --collector.host-port 192.168.178.20:8180 --collector.auth-token eyJhbGciOiJ...
```

When a batch can't be submitted, it prints the following log:
```
{"level":"error","ts":1510314302.2487879,"caller":"http/reporter.go:214","msg":"Could not submit jaeger batch","error":"failed to submit batch: 401 Unauthorized","stacktrace":"github.com/uber/jaeger/cmd/agent/app/reporter/http.(*Reporter).submitAndReport\n\t/mnt/storage/jpkroehling/Projects/src/github.com/uber/jaeger/cmd/agent/app/reporter/http/reporter.go:214\ngithub.com/uber/jaeger/cmd/agent/app/reporter/http.(*Reporter).EmitBatch\n\t/mnt/storage/jpkroehling/Projects/src/github.com/uber/jaeger/cmd/agent/app/reporter/http/reporter.go:200\ngithub.com/uber/jaeger/thrift-gen/jaeger.(*agentProcessorEmitBatch).Process\n\t/mnt/storage/jpkroehling/Projects/src/github.com/uber/jaeger/thrift-gen/jaeger/agent.go:137\ngithub.com/uber/jaeger/thrift-gen/jaeger.(*AgentProcessor).Process\n\t/mnt/storage/jpkroehling/Projects/src/github.com/uber/jaeger/thrift-gen/jaeger/agent.go:111\ngithub.com/uber/jaeger/cmd/agent/app/processors.(*ThriftProcessor).processBuffer\n\t/mnt/storage/jpkroehling/Projects/src/github.com/uber/jaeger/cmd/agent/app/processors/thrift_processor.go:110"}                               
```

As far as I know, this deviates from the TChannel reporter, which fails silently.

jpkrohling:
By the way: besides the tests included in this PR, I also did some manual testing. With a properly configured Keycloak and an Auth Proxy running on port 8180 (see blog post on the Jaeger blog, soon to be published), the `all-in-one` can be started like this (without exposing the agent ports):

```
docker run \
    -p 16686:16686 \
    -p 14268:14268 \
    --name=jaeger \
    jaegertracing/all-in-one:latest
```

An auth token can be obtained via:
```
curl \
    -X POST \
    -u instrumented-application:THE_SECRET \
    http://YOUR_IP:8080/auth/realms/jaeger/protocol/openid-connect/token \
    -d 'grant_type=client_credentials'
```

Then, the agent can be started as:
```
go run cmd/agent/main.go --collector.host-port 192.168.178.20:8180 --collector.auth-token eyJhbGciOiJ...
```

Finally, a target application can just be started as usual and the client will send spans via UDP as it would normally do and I confirmed that spans generated by this application were visible on the Query UI.

cmd/agent/app/reporter/http/reporter.go
```diff
@@ -0,0 +1,229 @@
...
+	// Number of successful span submissions to collector
+	SpansSubmitted metrics.Counter `metric:"spans.submitted"`
+
+	// Number of failed span submissions to collector
+	SpansFailures metrics.Counter `metric:"spans.failures"`
+}
+
+var httpClient = &http.Client{Timeout: 2 * time.Second}
+
+// Reporter forwards received spans to central collector tier over TChannel.
```
pavolloffay: over TChannel?

cmd/agent/app/reporter/http/reporter_test.go
```diff
@@ -0,0 +1,214 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"sync"
+	"testing"
+
+	"go.uber.org/zap"
```
pavolloffay: move to the second group

cmd/agent/app/reporter/http/reporter_test.go
```diff
@@ -0,0 +1,214 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/uber/jaeger-lib/metrics"
+	mTestutils "github.com/uber/jaeger-lib/metrics/testutils"
+
+	cApp "github.com/uber/jaeger/cmd/collector/app"
+	"github.com/uber/jaeger/cmd/collector/app/zipkin"
+	"github.com/uber/jaeger/thrift-gen/jaeger"
+	"github.com/uber/jaeger/thrift-gen/zipkincore"
+	tchanThrift "github.com/uber/tchannel-go/thrift"
```
pavolloffay: second group

cmd/agent/app/flags.go
```diff
@@ -56,6 +60,22 @@ func AddFlags(flags *flag.FlagSet) {
 		"",
 		"comma-separated string representing host:ports of a static list of collectors to connect to directly (e.g. when not using service discovery)")
 	flags.String(
+		scheme,
+		"http",
+		"protocol scheme to use when talking to the collector")
```
pavolloffay: What is the scheme for the current protocol (agent-collector)? Can be made that default?

jpkrohling:
> As far as I know, this deviates from the TChannel reporter, which fails silently.

Scratch that. It does complain when it can't submit a batch:

```
{"level":"error","ts":1510323644.892156,"caller":"tchannel/reporter.go:133","msg":"Could not submit jaeger batch","error":"no peers available","stacktrace":"..."}
```

cmd/agent/app/flags.go
jpkrohling: Good question. If TChannel is over HTTP (I don't think it is), then it's certainly `http` :) In any case, this is either `http` or `https`, as it's used only for the new reporter and will be ignored when the TChannel reporter is used.

pavolloffay: I am asking because this flag adds more confusion

jpkrohling: Added a clarification at the end:

> used only when auth-related properties are specified

Hopefully this clears up the confusion.

cmd/agent/app/builder.go
```diff
@@ -190,13 +231,29 @@ func (b *Builder) GetProcessors(rep reporter.Reporter, mFactory metrics.Factory)
 	return retMe, nil
 }
 
+func (b *Builder) useTChannelReporter() bool {
+	// if we don't have credentials, we use the tchannel reporter
+	// if we have an auth token or a pair of username+password, we should use the http reporter
+	return b.AuthToken == "" && (b.Username == "" || b.Password == "")
```
yurishkuro: I prefer not to introduce these indirect conditions. We should have a cmd line argument telling which reporter to use (default to tchannel). Then HTTP reporter can in addition check the other params.

cmd/agent/app/builder.go
```diff
@@ -190,13 +231,29 @@ func (b *Builder) GetProcessors(rep reporter.Reporter, mFactory metrics.Factory)
...
+}
+
 // GetHTTPServer creates an HTTP server that provides sampling strategies and baggage restrictions to client libraries.
-func (c HTTPServerConfiguration) GetHTTPServer(svc string, channel *tchannel.Channel, mFactory metrics.Factory) *http.Server {
-	mgr := httpserver.NewCollectorProxy(svc, channel, mFactory)
-	if c.HostPort == "" {
-		c.HostPort = defaultHTTPServerHostPort
+func (b *Builder) GetHTTPServer(r reporter.Reporter, mFactory metrics.Factory) *http.Server {
+	// TODO: this manager is used for the sampling and baggage restrictions, not sure we need for this here:
+	// is there a non-tchannel sampling/baggage restriction endpoint on the collector side?
```
yurishkuro: I don't think there is non-thrift endpoint for those.

cmd/agent/app/builder.go
```diff
@@ -190,13 +231,29 @@ func (b *Builder) GetProcessors(rep reporter.Reporter, mFactory metrics.Factory)
...
+	// for now, we let it be nil for non-TChannel reporters
```
yurishkuro: not sure how it would work then, when clients request their configs, the nil mgr will cause a panic

cmd/agent/app/reporter/http/reporter.go
```diff
@@ -0,0 +1,236 @@
...
+	"github.com/uber/jaeger/thrift-gen/jaeger"
+	"github.com/uber/jaeger/thrift-gen/zipkincore"
+)
+
+const (
+	jaegerBatches = "jaeger"
+	zipkinBatches = "zipkin"
+)
+
+type batchMetrics struct {
```
yurishkuro: is this different from tchannel reporter?  seems like it could be reused, e.g. via `reporter/metrics` package

cmd/agent/app/reporter/http/reporter.go
```diff
@@ -0,0 +1,236 @@
...
+var httpClient = &http.Client{Timeout: 2 * time.Second}
+
+// Reporter forwards received spans to central collector tier over plain HTTP.
+type Reporter struct {
+	scheme             string   // http or https
+	collectorHostPorts []string // only the first value is used
+
+	authToken string
+	username  string
+	password  string
```
yurishkuro: is this really a possibility that someone will be running the agent with basic auth?

cmd/agent/app/reporter/http/reporter.go
```diff
@@ -0,0 +1,236 @@
...
+		submissionFunc,
+		"Could not submit jaeger batch",
+		int64(len(batch.Spans)),
+		r.batchesMetrics[jaegerBatches],
+	)
+}
+
+// Endpoint returns the endpoint used when communicating with the remote collector
+func (r *Reporter) Endpoint() string {
+	// TODO: do we want to do client-side load balancing? Or use a retry logic?
```
yurishkuro: yes we do, if we care about performance. TChannel always sends to less-busy connection internally.

We could use LB from go-kit.

cmd/agent/app/reporter/tchannel/reporter.go
```diff
@@ -57,6 +57,7 @@ type Reporter struct {
 	peerListMgr    *peerlistmgr.PeerListManager
 	batchesMetrics map[string]batchMetrics
 	logger         *zap.Logger
+	mFactory       metrics.Factory
```
yurishkuro: why do we need to store it?

cmd/agent/app/builder.go
jpkrohling: A counter-argument is that we keep the actual implementation hidden behind the requirements: if we see auth data, we know we need to perform auth and we shall use whatever transport provides this feature. Once we change to gRPC, this distinction should disappear.

cmd/agent/app/builder.go
jpkrohling: What should be done then? Should I just implement a dummy endpoint, returning some default data?

cmd/agent/app/reporter/http/reporter.go
jpkrohling: I might be missing something. I understand that this reporter is to be used for the communication between the Agent and Collector. This is not for protecting the agent.

cmd/agent/app/reporter/tchannel/reporter.go
jpkrohling: Oops, this is a leftover from a previous version.

cmd/agent/app/reporter/http/reporter.go
jpkrohling: Would it be OK to have it in a next iteration, or is it required for this one? I would then create an issue and assign to myself.

jpkrohling:
I'm not up to date on what's expected from Travis. Looks like `master` is also failing, so, I don't think it's related to this PR.

cmd/agent/app/builder.go
yurishkuro: yes, that's precisely the hidden magic that goes against Go's "no magic" principle - adding an argument changes which transport is used, which port needs to be configured, etc. I would much rather be explicit. When we implement gRPC we can revisit.

cmd/agent/app/builder.go
yurishkuro: That seems like the weak spot of this PR - we're trying to hack just one flow between the agent and collector where in practice we need to replace the whole API with TLS-capable transport. Maybe we shouldn't do it piecemeal and just Do The Right Thing by implementing gRPC.

jpkrohling: I might have misunderstood this back in November. Based on the last meeting, looks like we want gRPC *in addition* to this, but then, the problem reported on this comment would still exist, wouldn't it?

jpkrohling:
@yurishkuro : is the direction of this PR still the path we want to follow, or should I close this?

yurishkuro:
> is the direction of this PR still the path we want to follow, or should I close this?

My main concern is that this PR is affecting a single flow between the agent and collector, rather than holistically all flows. I am not a big fan of doing that, especially when it comes to security features. There are other ways people can secure connections between agent and collector, e.g. with ssh tunnels. My preference is we wait for gRPC switch.

pavolloffay:
I have created https://github.com/jaegertracing/jaeger/issues/673

PR #533, opened 2017-11-13
Bump idl with zipkin swagger and messaging annos

pavolloffay:
* updated idl to master/head
* run `make thrift`

### DIFF
--- a/thrift-gen/baggage/tchan-baggage.go
+++ b/thrift-gen/baggage/tchan-baggage.go
@@ -43,6 +43,10 @@ func (c *tchanBaggageRestrictionManagerClient) GetBaggageRestrictions(ctx thrift
 	}
 	success, err := c.client.Call(ctx, c.thriftService, "getBaggageRestrictions", &args, &resp)
 	if err == nil && !success {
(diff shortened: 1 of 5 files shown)


thrift-gen/baggage/tchan-baggage.go
```diff
@@ -43,6 +43,10 @@ func (c *tchanBaggageRestrictionManagerClient) GetBaggageRestrictions(ctx thrift
 	}
 	success, err := c.client.Call(ctx, c.thriftService, "getBaggageRestrictions", &args, &resp)
 	if err == nil && !success {
+		switch {
```
black-adder: this looks fine but it's weird this got generated given that these thrift files haven't changed and we're always used thrift 0.9.3 to generate them. Maybe they patched something into 0.9.3 recently since it does seem like the docker image was updated: https://hub.docker.com/r/library/thrift/tags/

pavolloffay: It's weird I don't know why it's changed. I have used a standard make target.

PR #541, opened 2017-11-16
Docs: update migration from zipkin

pavolloffay:

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -95,9 +95,9 @@ go run ./cmd/agent/main.go
 
 ## Migrating from Zipkin
 
-Collector service exposes Zipkin compatible REST API `/api/v1/spans` and `/api/v2/spans`.
+Collector service exposes Zipkin compatible REST API `/api/v1/spans` and `/api/v2/spans` for both
+JSON and thrift encoding.
 By default it's disabled. It can be enabled with `--collector.zipkin.http-port=9411`. 
-Agent uses `TBinaryProtocol` and it is available on `UDP` port `5775`.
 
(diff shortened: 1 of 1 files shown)


docs/getting_started.md
```diff
@@ -93,16 +93,13 @@ make install
...
 
-In the near future we will provide individual Docker images for each component,
-as well as the Kubernetes templates.
-
 ## Migrating from Zipkin
 
-Collector service exposes Zipkin compatible REST API `/api/v1/spans` and can be enabled by
-`--collector.zipkin.http-port=9411`. It supports Thrift and JSON format.
-Agent uses `TBinaryProtocol` and is available on `UDP` port `5775`.
+Collector service exposes Zipkin compatible REST API `/api/v1/spans` and `/api/v2/spans`.
```
yurishkuro: We should clarify : both thrift and JSON? For both versions?

docs/getting_started.md
```diff
@@ -93,16 +93,13 @@ make install
...
+By default it's disabled. It can be enabled with `--collector.zipkin.http-port=9411`. 
+Agent uses `TBinaryProtocol` and it is available on `UDP` port `5775`.
```
yurishkuro: This is a bit terse, I would clarify that the agent can receive zipkin thrift over UDP. Having said that, not sure how useful it is as a migration mechanism, since zipkin clients typically don't use UDP. I think it would be great to achieve party between the agent and the collector for accepting spans over HTTP.

pavolloffay: I am not aware of any zipkin tracer/collector using UPD. Perhaps it's better to remove it from here because only jaeger clients are/were using it.

docs/getting_started.md
pavolloffay: done

PR #543, opened 2017-11-16
Prepare release 0.10.0

black-adder:


CHANGELOG.md
```diff
@@ -1,6 +1,26 @@
 Changes by Version
 ==================
 
+0.9.1 (unreleased)
```
yurishkuro: I think given the amount of changes, it should be 0.10

PR #548, opened 2017-11-17
Use healthcheck to mark cassandra-schema container as ready

### LINKED ISSUE #547: Crossdock test unstable
https://travis-ci.org/jaegertracing/jaeger/jobs/303621441

It seems that Cassandra takes too long to start, but the dependencies in docker-compose are not strict. E.g. jaeger-query depends on cassandra-schema, but cassandra-schema does start pretty quickly, even though it doesn't actually succeed until much later. Perhaps we can use [Docker's health checks](https://docs.docker.com/compose/compose-file/#healthcheck) to ensure that cassandra-schema container is not considered healthy until the schema has been actually created, so that the other components do not start prematurely. For example, after the last statement in plugin/storage/cassandra/schema/docker.sh we can write a status to a file and use that file in the health check

yurishkuro:
Resolves #547


crossdock/docker-compose.yml
```diff
@@ -1,28 +1,29 @@
...
             - WAIT_FOR_TIMEOUT=60s
 
             - CALL_TIMEOUT=60s
 
-            - AXIS_CLIENT=test_driver
+            - AXIS_CLIENT=test-driver
             - AXIS_SERVICES=go,node,java,python,zipkin-brave-json,zipkin-brave-json-v2,zipkin-brave-thrift
 
             - BEHAVIOR_ENDTOEND=client,services
+        depends_on:
```
pavolloffay: this is not necessary because there is already a link.

docker-compose/jaeger-docker-compose.yml
```diff
@@ -1,4 +1,4 @@
-version: '2'
+version: '3'
```
pavolloffay: this can break xdocks in other clients. But am +1 on version 3.

yurishkuro: yeah, I am aware that it will break other clients. We don't need to bump it if we don't have a use for the healthcheck, which was added in 2.1

pavolloffay: with version `3` we could perhaps also use configs (something like configmap in k8s) to split templates for ES and C*.

PR #549, opened 2017-11-19
Docker compose specify restart limit

pavolloffay:
Related to https://github.com/jaegertracing/jaeger/issues/547

### DIFF
--- a/docker-compose/jaeger-docker-compose.yml
+++ b/docker-compose/jaeger-docker-compose.yml
@@ -9,7 +9,7 @@ services:
         - "14268:14268"
         - "14267"
         - "9411:9411"
-      restart: on-failure
+      restart: on-failure:10
       depends_on:
         - cassandra-schema
 
@@ -19,7 +19,7 @@ services:
       ports:
         - "16686:16686"
         - "16687"
-      restart: on-failure
+      restart: on-failure:10
       depends_on:
         - cassandra-schema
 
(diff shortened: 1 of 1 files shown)


pavolloffay:
The first run passed https://travis-ci.org/jaegertracing/jaeger/builds/304432958. 

We should restart it a couple of times, to be sure whether it's stable

yurishkuro:
what is the default restart count if this value is not given?

10 actually looks rather low to me, since Cassandra can take a while to start.

yurishkuro:
restarting the build

pavolloffay:
the latest restart failed, so this does not fix it.

yurishkuro:
Driver cannot find query service:

`Get http://jaeger-query:16687: dial tcp: lookup jaeger-query on 127.0.0.11:53: no such host`

however it doesn't look like it was restarted 20 times, only 10:

```
$ grep jaeger-query_1 ~/Downloads/log.txt | grep fatal | wc -l
      10
```

It would be nice if we could specify the wait period before the restarts. Cassandra took about 45 seconds to start.

pavolloffay:
The problem is that compose increases delay between restarts so the last delay was higher that wait in `test-driver`.

pavolloffay:
This does not fix the issue, closing in favour of https://github.com/jaegertracing/jaeger/pull/565

PR #555, opened 2017-11-24
Enable TLS client connections to Cassandra

rbtcollins:
I'm not entirely sure how to test this in the context of Jaeger's test suite; it does work for me...

### DIFF
--- a/cmd/flags/cassandra/options.go
+++ b/cmd/flags/cassandra/options.go
@@ -34,6 +34,12 @@ const (
 	suffixSocketKeepAlive  = ".socket-keep-alive"
 	suffixUsername         = ".username"
 	suffixPassword         = ".password"
+	suffixTLS              = ".tls"
+	suffixCert             = ".tls.cert"
+	suffixKey              = ".tls.key"
+	suffixCA               = ".tls.ca"
+	suffixServerName       = ".tls.server-name"
+	suffixVerifyHost       = ".tls.verify-host"
 )
 
 // TODO this should be moved next to config.Configuration struct (maybe ./flags package)
@@ -62,6 +68,10 @@ func NewOptions(primaryNamespace string, otherNamespaces ...string) *Options {
 	options := &Options{
 		primary: &namespaceConfig{
 			Configuration: config.Configuration{
+				TLS: config.TLS{
+					Enabled:                false,
+					EnableHostVerification: true,
+				},
 				MaxRetryAttempts:   3,
 				Keyspace:           "jaeger_v1_local",
 				ProtoVersion:       4,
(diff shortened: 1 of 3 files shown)


yurishkuro:
This is one of those things that are difficult to integration-test. We're not testing basic auth either.

Could you please add some instructions to the docs section https://github.com/jaegertracing/jaeger/blob/master/docs/deployment.md#cassandra, e.g. links to Cassandra documentation or blogs about how these settings are meant to be used.

rbtcollins:
@yurishkuro I've put some words in.

rbtcollins:
@yurishkuro I think the failure is a flaky test?

cmd/flags/cassandra/options.go
```diff
@@ -34,6 +34,12 @@ const (
 	suffixSocketKeepAlive  = ".socket-keep-alive"
 	suffixUsername         = ".username"
 	suffixPassword         = ".password"
+	suffixTLS              = ".tls"
+	suffixCert             = ".cert"
```
yurishkuro: I think it would better to prefix all params with `tls-`, e.g. `.tls-cert`, `.tls-key`, etc.

cmd/flags/cassandra/options.go
```diff
@@ -34,6 +34,12 @@ const (
...
+	suffixKey              = ".key"
+	suffixCA               = ".ca"
+	suffixServerName       = ".servername"
```
yurishkuro: should this be `server-name`?

docs/deployment.md
```diff
@@ -106,6 +112,18 @@ where `{datacenter}` is the name used in the Cassandra configuration / network t
...
 Run the script without arguments to see the full list of recognized parameters.
 
+#### TLS support
+
+Jaeger supports TLS client to node connections as long as you've configured
+your Cassandra cluster correctly. After verifying with e.g. `cqlsh`, you can
+configure the collector and query like so:
+
+```
+collector-linux  --cassandra.servers=cassandra --cassandra.tls \
```
yurishkuro: I would prefer not to direct people to command line arguments, but instead use env variables. So this could be the standard Docker command for the collector with env vars defined. Also, one param per line to make it more readable.
```
docker run \
  -e SPAN_STORAGE_TYPE=cassandra \
  -e CASSANDRA_SERVERS=<...> \
  -e CASSANDRA_TLS=true \
  . . . etc . . .
  jaegertracing/jaeger-collector
```

cmd/flags/cassandra/options.go
```diff
@@ -129,6 +139,30 @@ func addFlags(flagSet *flag.FlagSet, nsConfig *namespaceConfig) {
...
+		nsConfig.TLS.KeyPath,
+		"Path to TLS key file")
+	flagSet.String(
+		nsConfig.namespace+suffixCA,
+		nsConfig.TLS.CaPath,
+		"Path to TLS CA file")
+	flagSet.String(
+		nsConfig.namespace+suffixServerName,
+		nsConfig.TLS.ServerName,
+		"Override the TLS server name")
```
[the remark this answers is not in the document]
rbtcollins: No. The main reason is to allow connecting to a host where the server name in the cert, or one of the SANs, differs.

e.g. you have a cert for foo.svc.cluster.local but you're connecting to just 'foo', you could set ServerName=foo.svc.cluster.local. 

There is something odd in the dialer setup for jaeger at the moment which is leading to the TLS layer missing the hostname, and ServerName can be used to work around that. I'm going to poke around and see if I can make it more JustWorks in future - but there are still cases where ServerName is needed.

docs/deployment.md
rbtcollins: ok, theres some automatic conversion to env variables from the command line args right?

cmd/flags/cassandra/options.go
rbtcollins: ¯\_(ツ)_/¯

docs/deployment.md
yurishkuro: yes, `s/[.-]/_/g + toUpperCase`

yurishkuro:
Thanks!

PR #568, opened 2017-11-29
Docs add dependencies tab option

pavolloffay:


docs/deployment.md
```diff
@@ -182,6 +182,7 @@ Two aspects of the UI can be configured:
```
black-adder: above this line, let's change from "Two aspects" to "Multiple aspects"

yurishkuro:
Since this is not code submission, I am going to merge without the DCO.

PR #558, opened 2017-11-25
use utc time zone in es storage getIndices test

seregayoga:
I changed test `TestGetIndices` to use UTC timezone instead of local
Fixes #557

PR #569, opened 2017-11-29
Use UTC time zone in es storage findIndices test

vprithvi:
This test fails when the local timezone is IST.


black-adder:
A simple fix for this would be to physically move somewhere where the test passes.

Do you mind checking the entire ES codebase for Time.Local references? We had a similar PR  https://github.com/jaegertracing/jaeger/pull/558/files so let's just fix them all.

vprithvi:
>Do you mind checking the entire ES codebase for Time.Local references?

Already did, there are no more!

>A simple fix for this would be to physically move somewhere where the test passes.

"simple"

PR #570, opened 2017-11-29
Fix jaeger agent embedded reporter builder bug

vprithvi:
- When overriding the collector service name using the agent app builder's embedded
  `WithCollectorServiceName`, the service name is stored in the embedded struct and
  not in the outer struct.

- When constructing an agent using `createAgent`, `GetHttpServer` reads the outer
  struct leading to different behavior when reading configuration from a file, as
  opposed to using the builder.

- This change uses go-yaml's `inline` keyword to deserialize embedded structs.

### DIFF
--- a/cmd/agent/app/builder.go
+++ b/cmd/agent/app/builder.go
@@ -43,9 +43,6 @@ const (
 
 	defaultHTTPServerHostPort = ":5778"
 
-	agentServiceName            = "jaeger-agent"
-	defaultCollectorServiceName = "jaeger-collector"
-
 	jaegerModel model = "jaeger"
 	zipkinModel       = "zipkin"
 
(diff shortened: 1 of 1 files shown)


cmd/agent/app/builder.go
```diff
@@ -71,12 +68,7 @@ type Builder struct {
 	HTTPServer HTTPServerConfiguration  `yaml:"httpServer"`
 	Metrics    jmetrics.Builder         `yaml:"metrics"`
 
-	// These 3 fields are copied from tchreporter.Builder because yaml does not parse embedded structs
```
vprithvi: image

cmd/agent/app/builder.go
```diff
@@ -71,12 +68,7 @@ type Builder struct {
...
-	CollectorHostPorts   []string `yaml:"collectorHostPorts"`
-	DiscoveryMinPeers    int      `yaml:"minPeers"`
-	CollectorServiceName string   `yaml:"collectorServiceName"`
-
-	tchreporter.Builder
+	tchreporter.Builder `yaml:",inline"`
```
yurishkuro: this is awesome. Could you search the rest of the code for this though? I think there may be other examples of this copy-pasta.

btw, why comma before `inline`?

vprithvi: >btw, why comma before inline?

It declares an empty yaml tag for `tchreporter.Builder`. Simply using `inline` doesn't work.

 >Could you search the rest of the code for this though? I think there may be other examples of this copy-pasta.

I wasn't able to come up with a good way of searching for anonymous structs; Nothing else seems to have similar comments.

PR #579, opened 2017-12-04
Fix issue where Domain to UI model converter double reports the same …

black-adder:
…Span Reference

Problem was that the converter removes the parentSpanID and replaces it with a parent child-of Reference. But if the child-of reference is already present in the model span, it gets double reported.

It seems like the whole `if span.ParentSpanID != 0 && !preserveParentID {` code path could be removed. Can any part of the jaeger system send in a span where ParentSpanID is filled out but the child-of reference is not present? Also, the logic here makes a couple of faulty assumptions.

### DIFF
--- a/model/converter/json/from_domain.go
+++ b/model/converter/json/from_domain.go
@@ -105,7 +105,6 @@ func (fd fromDomain) convertReferences(span *model.Span, preserveParentID bool) 
 			SpanID:  json.SpanID(ref.SpanID.String()),
 		})
 		if ref.TraceID == span.TraceID && ref.SpanID == span.ParentSpanID {
-			// Check if the parent reference already exists
 			parentRefAdded = true
 		}
 	}
--- a/model/converter/json/from_domain.go
+++ b/model/converter/json/from_domain.go
@@ -97,27 +97,27 @@ func (fd fromDomain) convertReferences(span *model.Span, preserveParentID bool) 
 		length++
 	}
 	out := make([]json.Reference, 0, length)
-	var parentRef json.Reference
-	if span.ParentSpanID != 0 && !preserveParentID {
-		// TODO this (wrongly) assumes that the reference type is always ChildOf
-		parentRef = json.Reference{
-			RefType: json.ChildOf,
-			TraceID: json.TraceID(span.TraceID.String()),
-			SpanID:  json.SpanID(span.ParentSpanID.String()),
-		}
-		out = append(out, parentRef)
-	}
+	var parentRefAdded bool
 	for _, ref := range span.References {
-		// TODO if the parentRef was wrongly added as a ChildOf ref above, this could
-		// potentially add a FollowsFrom ref and still keep the ChildOf ref
 		newRef := json.Reference{
 			RefType: fd.convertRefType(ref.RefType),
 			TraceID: json.TraceID(ref.TraceID.String()),
 			SpanID:  json.SpanID(ref.SpanID.String()),
 		}
-		if newRef != parentRef {
-			out = append(out, newRef)
+		out = append(out, newRef)
+		if newRef.TraceID == json.TraceID(span.TraceID.String()) &&
+			newRef.SpanID == json.SpanID(span.ParentSpanID.String()) {
+			// Check if the parent reference already exists
+			parentRefAdded = true
 		}
+	}
+	if span.ParentSpanID != 0 && !preserveParentID && !parentRefAdded {
+		// TODO this (wrongly) assumes that the reference type is always ChildOf
+		out = append(out, json.Reference{
+			RefType: json.ChildOf,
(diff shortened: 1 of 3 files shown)


model/converter/json/from_domain.go
```diff
@@ -97,19 +97,27 @@ func (fd fromDomain) convertReferences(span *model.Span, preserveParentID bool)
...
 			RefType: json.ChildOf,
 			TraceID: json.TraceID(span.TraceID.String()),
 			SpanID:  json.SpanID(span.ParentSpanID.String()),
-		})
+		}
+		out = append(out, parentRef)
 	}
 	for _, ref := range span.References {
-		out = append(out, json.Reference{
+		// TODO if the parentRef was wrongly added as a ChildOf ref above, this could
```
black-adder: For this TODO, it might be easier to keep track of parentRef but not add it until we're done with this for loop and if the reference isn't added, add it post. This way, if the reference should be a FollowsFrom, we don't double report.

yurishkuro: do you want to fix it for good then?

yurishkuro:
> Can any part of the jaeger system send in a span where ParentSpanID is filled out but the child-of reference is not present? 

iirc the only reason ParentSpanID even exists in the model is an optimization. In our deployment we hardly have any apps that create FollowsFrom references, so the overwhelming majority of spans have just a single parent of the child-of type. Instead of representing that link as a full blown Reference object we simply record the ParentSpanID (a single uint64).

model/converter/json/from_domain.go
```diff
@@ -97,18 +97,26 @@ func (fd fromDomain) convertReferences(span *model.Span, preserveParentID bool)
...
+	var parentRefAdded bool
 	for _, ref := range span.References {
-		out = append(out, json.Reference{
+		newRef := json.Reference{
 			RefType: fd.convertRefType(ref.RefType),
 			TraceID: json.TraceID(ref.TraceID.String()),
 			SpanID:  json.SpanID(ref.SpanID.String()),
+		}
+		out = append(out, newRef)
+		if newRef.TraceID == json.TraceID(span.TraceID.String()) &&
```
yurishkuro: comparison via strings is inefficient - you have the strongly typed `ref` in scope, why not use that?

model/converter/json/from_domain.go
```diff
@@ -97,18 +97,26 @@ func (fd fromDomain) convertReferences(span *model.Span, preserveParentID bool)
...
+			newRef.SpanID == json.SpanID(span.ParentSpanID.String()) {
+			// Check if the parent reference already exists
+			parentRefAdded = true
+		}
+	}
+	if span.ParentSpanID != 0 && !preserveParentID && !parentRefAdded {
+		// TODO this (wrongly) assumes that the reference type is always ChildOf
```
yurishkuro: it assumes that correctly, because if ParentSpanID != 0 but there are no other references then that ParentSpanID does refer to child-of type (you may want to leave that as a comment)

model/converter/json/from_domain.go
```diff
@@ -97,19 +97,26 @@ func (fd fromDomain) convertReferences(span *model.Span, preserveParentID bool)
...
-	}
+	var parentRefAdded bool
 	for _, ref := range span.References {
 		out = append(out, json.Reference{
 			RefType: fd.convertRefType(ref.RefType),
 			TraceID: json.TraceID(ref.TraceID.String()),
 			SpanID:  json.SpanID(ref.SpanID.String()),
 		})
+		if ref.TraceID == span.TraceID && ref.SpanID == span.ParentSpanID {
+			// Check if the parent reference already exists
```
yurishkuro: not sure this comment explains anything, maybe remove it?

model/converter/json/from_domain.go
```diff
@@ -97,19 +97,26 @@ func (fd fromDomain) convertReferences(span *model.Span, preserveParentID bool)
...
+			parentRefAdded = true
+		}
+	}
+	if span.ParentSpanID != 0 && !preserveParentID && !parentRefAdded {
```
yurishkuro: not directly related to this diff, but I see that the only time `preserveParentID == true` happens is when this conversion is called from [ES storage WriteSpan function](https://github.com/jaegertracing/jaeger/blob/412baf60db217d6f41d7fdca04358069362b29ca/plugin/storage/es/spanstore/writer.go#L118). What was the reason why we needed to preserve ParentSpanID in ES? Same performance/storage optimization?

yurishkuro:
Incidentally, I wonder if it would've been better to write an Adjuster for the domain model instead, and not have any of this business logic in the toJSON conversion.

yurishkuro:
btw was there a ticket describing what the problem this was causing?

black-adder:
only internally, I thought it was an internal build issue. I'll put one up on github

PR #583, opened 2017-12-05
Report card fixes

### LINKED ISSUE #582: golint
I am trying to get jaeger on [awesome-go](https://github.com/avelino/awesome-go). The pull request is https://github.com/avelino/awesome-go/pull/1711. It seems @cassiobotaro wants jaeger to fix up the remaining warnings from golint, etc. The warnings are listed on https://goreportcard.com/report/github.com/jaegertracing/jaeger.

isaachier:
Fixes #582.


examples/hotrod/services/driver/redis.go
```diff
@@ -53,7 +53,7 @@ func (r *Redis) FindDriverIDs(ctx context.Context, location string) []string {
 		span.SetTag("param.location", location)
 		ext.SpanKindRPCClient.Set(span)
 		defer span.Finish()
-		ctx = opentracing.ContextWithSpan(ctx, span)
+		_ = opentracing.ContextWithSpan(ctx, span)
```
yurishkuro: This is wrong. The context is used later.

storage/samplingstore/mocks/Store.go
```diff
@@ -19,10 +19,12 @@ import "github.com/stretchr/testify/mock"
 
 import "time"
 
+// Store writes and retrieves sampling data to and from storage.
```
yurishkuro: This is generated code. It should not be edited.

pkg/distributedlock/mocks/Lock.go
```diff
@@ -20,10 +20,13 @@ import (
 	"github.com/stretchr/testify/mock"
 )
 
+// Lock mocks distributed lock for control of a resource.
```
yurishkuro: Generated code

model/span.go
```diff
@@ -96,6 +96,16 @@ func (s *Span) NormalizeTimestamps() {
 	}
 }
 
+// FlattenTags combines span tags, process tags, and log fields into one KeyValues collection
```
yurishkuro: This is used just once and the semantic value of mixing tags with log fields is highly dubious. I would avoid having this in the model.

model/keyvalue.go
```diff
@@ -236,6 +236,11 @@ func (kv *KeyValue) IsLess(two *KeyValue) bool {
 	}
 }
 
+// Matches checks if this KeyValue matches the query key/value pair
+func (kv *KeyValue) Matches(key, value string) bool {
```
yurishkuro: Of dubious value as public method.

model/keyvalue.go
```diff
@@ -272,6 +277,18 @@ func (kvs KeyValues) Equal(other KeyValues) bool {
 	return true
 }
 
+// FindMatch checks for a matching KeyValue in the KeyValues for the given
+// query key/value pair. Returns the match and true when the match is found.
+// Returns empty KeyValue and false when no match is found.
+func (kvs KeyValues) FindMatch(key, value string) (KeyValue, bool) {
```
yurishkuro: I can't see on the phone but I thought we already had a find-like method for a list of tags. Also, the method seems suspicious given that it takes bake as string.

isaachier: There was a note in the existing code explaining the `FindKey` method only handles one value rather than multiple.

model/span.go
isaachier: The only reason I extracted this as a public method was a `TODO` by the existing function suggesting this fix. Can easily remove.

examples/hotrod/services/driver/redis.go
isaachier: Actually it looks like this is correct. It is not used again in the existing method and the `ctx` argument is copied by value (not a pointer). Unless context is an interface or something like that, cannot see how this works.

isaachier: I guess it is an interface: https://golang.org/pkg/context/#Context.

yurishkuro: Hard for me to verify on the phone, but isn't the context later used for logging?

storage/spanstore/memory/memory.go
```diff
@@ -201,22 +210,14 @@ func (m *Store) validSpan(span *model.Span, query *spanstore.TraceQueryParameter
...
-		keyValueFoundAndMatches := false
-		// (NB): we cannot find the KeyValue.Find function because there can be multiple tags with the same key
-		for _, keyValue := range spanKVs {
-			if keyValue.Key == queryK && keyValue.AsString() == queryV {
-				keyValueFoundAndMatches = true
-				break
-			}
-		}
-		if !keyValueFoundAndMatches {
+		// (NB): we cannot find the KeyValues.FindKey function because there can be multiple tags with the same key
```
yurishkuro: we cannot use...

isaachier: Yes added back note from master. Had to write local helper function instead.

isaachier:
Anyone want to merge this?

storage/spanstore/memory/memory.go
yurishkuro: I meant changing the comment "we cannot **use**", not "find"

isaachier: oh lol didn't read carefully

PR #584, opened 2017-12-05
Add log to avoid warning for ineffectual assignment

isaachier:

### DIFF
--- a/examples/hotrod/services/driver/redis.go
+++ b/examples/hotrod/services/driver/redis.go
@@ -19,7 +19,6 @@ import (
 	"errors"
 	"fmt"
 	"math/rand"
-	"strings"
 	"sync"
 
 	"github.com/opentracing/opentracing-go"
--- a/examples/hotrod/services/driver/redis.go
+++ b/examples/hotrod/services/driver/redis.go
@@ -54,6 +54,7 @@ func (r *Redis) FindDriverIDs(ctx context.Context, location string) []string {
 		ext.SpanKindRPCClient.Set(span)
 		defer span.Finish()
 		ctx = opentracing.ContextWithSpan(ctx, span)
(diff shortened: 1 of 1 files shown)


examples/hotrod/services/driver/redis.go
```diff
@@ -54,6 +54,7 @@ func (r *Redis) FindDriverIDs(ctx context.Context, location string) []string {
 		ext.SpanKindRPCClient.Set(span)
 		defer span.Finish()
 		ctx = opentracing.ContextWithSpan(ctx, span)
+		r.logger.For(ctx).Info("FindDriverIDs at " + location)
```
yurishkuro: that's not useful. I would rather have this just before the `return` statement and log the `drivers` array.

isaachier: Got it OK

examples/hotrod/services/driver/redis.go
```diff
@@ -62,6 +63,8 @@ func (r *Redis) FindDriverIDs(ctx context.Context, location string) []string {
 	for i := range drivers {
 		drivers[i] = fmt.Sprintf("T7%05dC", rand.Int()%100000)
 	}
+	msg := fmt.Sprintf("FindDriverIDs(%s) = [%s]", location, strings.Join(drivers, ", "))
+	r.logger.For(ctx).Info(msg)
```
yurishkuro: Zap has a field for arrays: https://github.com/uber-go/zap/blob/master/array.go#L98

```
r.logger.For(ctx).Info("Found drivers", zap.Strings("drivers", drivers))
```

isaachier: Much cleaner thanks. Don't know zap too well.

PR #587, opened 2017-12-06
make cassandra service_name_index inserts idempotent

kwojcik:
Use the span hash as the source of variation for bucket choice instead of an atomic counter so that the inserts are idempotent.

Signed-off-by: Kevin Wojcik <kwojcik@uber.com>

### DIFF
--- a/plugin/storage/cassandra/spanstore/writer.go
+++ b/plugin/storage/cassandra/spanstore/writer.go
@@ -17,7 +17,6 @@ package spanstore
 import (
 	"encoding/json"
 	"strings"
-	"sync/atomic"
 	"time"
 	"unicode/utf8"
 
@@ -85,7 +84,6 @@ type SpanWriter struct {
 	writerMetrics        spanWriterMetrics
 	logger               *zap.Logger
 	tagIndexSkipped      metrics.Counter
-	bucketCounter        uint32
 	tagFilter            dbmodel.TagFilter
 }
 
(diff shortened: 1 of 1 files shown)


plugin/storage/cassandra/spanstore/writer.go
```diff
@@ -200,7 +198,7 @@ func (s *SpanWriter) indexByDuration(span *dbmodel.Span, startTime time.Time) er
 }
 
 func (s *SpanWriter) indexBySerice(traceID model.TraceID, span *dbmodel.Span) error {
```
black-adder: while you're here, do you mind doing s/indexBySerice/indexByService :D

plugin/storage/cassandra/spanstore/writer.go
```diff
@@ -199,8 +197,8 @@ func (s *SpanWriter) indexByDuration(span *dbmodel.Span, startTime time.Time) er
 	return err
 }
 
-func (s *SpanWriter) indexBySerice(traceID model.TraceID, span *dbmodel.Span) error {
-	bucketNo := atomic.AddUint32(&s.bucketCounter, 1) % defaultNumBuckets
+func (s *SpanWriter) indexByService(traceID model.TraceID, span *dbmodel.Span) error {
+	bucketNo := span.SpanHash % defaultNumBuckets
```
yurishkuro: good call

PR #595, opened 2017-12-15
Use unsigned for span hash

pavolloffay:
span hash is used for cassandra buckets, with signed it produced
negative numbers

Related to https://github.com/jaegertracing/jaeger/pull/587 and fixes https://github.com/jaegertracing/spark-dependencies/issues/23

### DIFF
--- a/plugin/storage/cassandra/spanstore/dbmodel/converter.go
+++ b/plugin/storage/cassandra/spanstore/dbmodel/converter.go
@@ -57,7 +57,7 @@ func (c converter) fromDomain(span *model.Span) *Span {
 		Refs:          refs,
 		Process:       udtProcess,
 		ServiceName:   span.Process.ServiceName,
-		SpanHash:      int64(spanHash),
+		SpanHash:      spanHash,
 	}
 }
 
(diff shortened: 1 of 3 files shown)


yurishkuro:
can we instead convert to uint when calculating the bucket? (a) it would be a smaller change, and (b) I am not sure how Cassandra will take full unsigned i64 as a numeric field

pavolloffay:
> can we instead convert to uint when calculating the bucket? (a) it would be a smaller change, and (b) I am not sure how Cassandra will take full unsigned i64 as a numeric field

Isn't it better to keep it as unsigned? Our hash impl originally returns unsigned.
https://github.com/pavolloffay/jaeger/blob/d16de316a8bfc0ba7a6d5d1b7e43d3c48156d8ae/model/hash.go#L30

b) Is a good point and we are fine because max value is 9 
 see https://github.com/gocql/gocql/blob/master/marshal.go#L449 
Cassandra internally converts it to signed https://github.com/gocql/gocql/blob/master/marshal.go#L426

yurishkuro:
> b) Is a good point and we are fine because max value is 9 see https://github.com/gocql/gocql/blob/master/marshal.go#L449

When I look at our version of the gocql driver, I see this
```
        case uint64:
                if v > math.MaxUint16 {
                        return nil, marshalErrorf("marshal smallint: value %d out of range", v)
                }
                return encShort(int16(v)), nil
```

Makes no sense to me whatsoever, how does it even work for us? `MaxUint16` is like 65k, wouldn't we be getting errors almost all the time?

In the new driver it's still bound by `MaxUint32`.

yurishkuro:
putting aside the truncation that happens in the driver, I suggest a simpler change that addresses the actual problem without affecting the rest of the code - #599

PR #597, opened 2017-12-16
Fix #556

### LINKED ISSUE #556: when trace id is invalid, jaeger-ui send this request forever
jaeger verison: 3136216e0636e67e39c6815b909f0cf9e0888884
jaeger-ui version : 27c3d58e224f0ff78a4ce0ba4622d47ec4bffa7c

```sh
$ query 2>&1 > query.log --query.static-files=jaeger-ui/build/ --span-storage.type=elasticsearch --es.server-urls="${ESServerUrls}"
```

NeoCN:
When query result got zero hit, should not collectSpans any more

### DIFF
--- a/plugin/storage/es/spanstore/reader.go
+++ b/plugin/storage/es/spanstore/reader.go
@@ -72,8 +72,6 @@ var (
 	// ErrUnableToFindTraceIDAggregation occurs when an aggregation query for TraceIDs fail.
 	ErrUnableToFindTraceIDAggregation = errors.New("Could not find aggregation of traceIDs")
 
-	errNilHits = errors.New("No hits in read results found")
-
 	errNoTraces = errors.New("No trace with that ID found")
 
 	defaultMaxDuration = model.DurationAsMicroseconds(time.Hour * 24)
@@ -232,8 +230,8 @@ func (s *SpanReader) multiRead(traceIDs []string, startTime, endTime time.Time) 
 
 	var traces []*model.Trace
 	for _, result := range results.Responses {
-		if result.Hits == nil {
-			return nil, errNilHits
+		if result.Hits == nil || len(result.Hits.Hits) == 0 {
+			continue
 		}
 		spans, err := s.collectSpans(result.Hits.Hits)
 		if err != nil {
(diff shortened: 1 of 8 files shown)


plugin/storage/es/spanstore/reader.go
```diff
@@ -235,6 +235,9 @@ func (s *SpanReader) multiRead(traceIDs []string, startTime, endTime time.Time)
 		if result.Hits == nil {
 			return nil, errNilHits
 		}
+		if len(result.Hits.Hits) == 0 {
```
yurishkuro: * under what conditions will this happen? (maybe add a comment)
  * why should this be handled differently than `result.Hits == nil`?

NeoCN: when trace id is invalid, we got response json as follows:
```json
{
   "responses":[
      {
         "took":2,
         "timed_out":false,
         "_shards":{
            "total":15,
            "successful":15,
            "failed":0
         },
         "hits":{
            "total":0,
            "max_score":null,
            "hits":[]
         },
         "status":200
      }
   ]
}
```
and then the response json is unmarshalled to the `elastic.MultiSearchResult` struct, you will notice that the snippet
```json
"hits":{
    "total":0,
    "max_score":null,
    "hits":[]
},
```
will not get nil `result.Hits`

yurishkuro: What if we combine both `if`s into this instead:
```
if result.Hits == nil || len(result.Hits.Hits) == 0 {
    continue
}
```
I think this will work better with the calling function that will detect empty cumulative result and respond with ErrNotFound. The private `errNilHits` should be removed altogether.

NeoCN: OK, I will do it

yurishkuro:
@NeoCN there is a test that needs to be fixed, and [all commits need to be signed](https://github.com/jaegertracing/jaeger/blob/master/CONTRIBUTING.md#sign-your-work) (not just the last one).  If you have multiple commits in the branch/PR, you can either amend them, or squash them into one.

Let me know if you want me to make these changes instead.

NeoCN:
@yurishkuro I will make a new pr, closing this one now.

PR #599, opened 2017-12-17
Ensure Cassandra bucket number is not negative

yurishkuro:
Span hash is used for cassandra buckets, but since it's signed it can produce negative numbers.

Related to #587, replaces #595

Resolves https://github.com/jaegertracing/spark-dependencies/issues/23


pavolloffay:
anyways I wanted to update my PR with this change

yurishkuro:
ah ok, wasn't sure. I'll merge this since is't a very small change that evidently solves the spark job issue, but I am not sure about the other PR that changes the DB model.

pavolloffay:
it seems that storage tests are not sufficient. They were passing even though, query service could not get spans.

PR #601, opened 2017-12-18
skip nil and zero length hits, fix #556

### LINKED ISSUE #556: when trace id is invalid, jaeger-ui send this request forever

NeoCN:
Focus on the flowing multiread function code snippet
```golang
	var traces []*model.Trace
	for _, result := range results.Responses {
		if result.Hits == nil {
			return nil, errNilHits
		}
		spans, err := s.collectSpans(result.Hits.Hits)
		if err != nil {
			return nil, err
		}
		traces = append(traces, &model.Trace{Spans: spans})
	}
	return traces, nil
```

when trace id is invalid or not in the es store, we got response json as follows:
```json
{
   "responses":[
      {
         "took":2,
         "timed_out":false,
         "_shards":{
            "total":15,
            "successful":15,
            "failed":0
         },
         "hits":{
            "total":0,
            "max_score":null,
            "hits":[]
         },
         "status":200
      }
   ]
}
```
and then the response json is unmarshalled to the `elastic.MultiSearchResult` struct, you will notice that the snippet
```json
"hits":{
    "total":0,
    "max_score":null,
    "hits":[]
},
```
will not get nil `result.Hits`, but a zero length `result.Hits.Hits`.

When `len(result.Hits.Hits) == 0`,  `collectSpans` will return `[]*model.Span` with zero length and no error. 

In this case, got result traces as
```
[]*model.Trace{
    &model.Trace{
        Spans: {
        },
        Warnings: nil,
    },
}
```
this results that `len(traces) == 0` not triggered, and then no `errNoTraces ` is returned.

As explained above, neither nil `result.Hits` nor zero length `result.Hits.Hits` should do the following `collectSpans`, just skip them.

### DIFF
--- a/plugin/storage/es/spanstore/reader.go
+++ b/plugin/storage/es/spanstore/reader.go
@@ -72,8 +72,6 @@ var (
 	// ErrUnableToFindTraceIDAggregation occurs when an aggregation query for TraceIDs fail.
 	ErrUnableToFindTraceIDAggregation = errors.New("Could not find aggregation of traceIDs")
 
-	errNilHits = errors.New("No hits in read results found")
-
 	errNoTraces = errors.New("No trace with that ID found")
 
 	defaultMaxDuration = model.DurationAsMicroseconds(time.Hour * 24)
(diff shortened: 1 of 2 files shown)


yurishkuro:
Thanks!

Fixes #556

PR #602, opened 2017-12-19
Extend query API to allow passing tags as JSON string

yurishkuro:

Part of #550


black-adder:
can we just deprecate the `tag` parameter? Only accepting json seems ideal.

yurishkuro:
> can we just remove the tag parameter? Only accepting json seems ideal.

we can, but not in this PR since it will break the build until UI changes are done.

PR #604, opened 2017-12-20
Support archive storage in the query-service

### LINKED ISSUE #603: Enable trace archiving in query service
#57 created the foundation for archiving functionality, but it's not enabled in the `cmd/query/main`

Related UI issue https://github.com/jaegertracing/jaeger-ui/issues/7

yurishkuro:
Users often ask for this feature. The trace retention period is usually short, and often people would put a link to the trace into some issue tracker to improve service performance, but by the time someone starts working on the ticket the trace has already expired.

Resolves #603


yurishkuro:
ready for review

cmd/query/main.go
```diff
@@ -174,3 +179,26 @@ func registerStaticHandler(r *mux.Router, logger *zap.Logger, qOpts *app.QueryOp
 		logger.Info("Static handler is not registered")
 	}
 }
+
+func archiveOptions(storageFactory istorage.Factory, logger *zap.Logger) []app.HandlerOption {
+	reader, err := storageFactory.CreateSpanReader()
+	if err == istorage.ErrArchiveStorageNotConfigured || err == istorage.ErrArchiveStorageNotSupported {
+		return nil
```
vprithvi: Wouldn't this fail silently when `ErrArchiveStorageNotConfigured` is false; meaning it's configured; and when `ErrArchiveStorageNotSupported` is true?
Shouldn't we return the `ErrArchiveStorageNotSupported` when someone mistakenly configures archive storage on a storage that doesn't support archiving?

yurishkuro: it's not possible to configure archiving on a storage that doesn't support it. For example, with Cassandra you need to pass `--cassandra-archive.enabled=true`.

plugin/storage/cassandra/factory.go
```diff
@@ -86,3 +103,19 @@ func (f *Factory) CreateSpanWriter() (spanstore.Writer, error) {
...
+	}
+	return cSpanStore.NewSpanReader(f.archiveSession, f.metricsFactory, f.logger), nil
+}
+
+// CreateArchiveSpanWriter implements storage.ArchiveFactory
+func (f *Factory) CreateArchiveSpanWriter() (spanstore.Writer, error) {
+	if f.archiveSession == nil {
+		return nil, storage.ErrArchiveStorageNotConfigured
+	}
+	return cSpanStore.NewSpanWriter(f.archiveSession, f.Options.SpanStoreWriteCacheTTL, f.metricsFactory, f.logger), nil
```
vprithvi: I'm a bit confused, is `f.Options.SpanStoreWriteCacheTTL` shared between both the archive and primary span writers?

plugin/storage/cassandra/factory.go
```diff
@@ -38,13 +44,14 @@ type Factory struct {
 
 	primaryConfig  config.SessionBuilder
 	primarySession cassandra.Session
-	// archiveSession cassandra.Session TODO
+	archiveConfig  config.SessionBuilder
```
vprithvi: I feel that the `archiveConfig` can be set to a higher default consistency level than the `primaryConfig`. What do you think?

plugin/storage/factory.go
```diff
@@ -131,3 +131,29 @@ func (f *Factory) InitFromViper(v *viper.Viper) {
 		}
 	}
 }
+
+// CreateArchiveSpanReader implements storage.ArchiveFactory
+func (f *Factory) CreateArchiveSpanReader() (spanstore.Reader, error) {
+	factory, ok := f.factories[f.SpanStorageType]
```
black-adder: this means that the span store and archive span store both have to be using the same storage type? I think that's a fair assumption

plugin/storage/cassandra/factory.go
yurishkuro: at the moment, yes. The retention period of archive storage is expected to be much longer that for primary, so the short cache TTL for the primary is fine for archive too. Strictly speaking we don't even need this cache for archive, but whatever.

plugin/storage/cassandra/factory.go
yurishkuro: it's fully configurable by user. Given how marginal this feature is overall, I don't think introducing custom defaults per storage class is worth the complexity in the code.

plugin/storage/factory.go
yurishkuro: it can be changed in the future by introducing another env var.

plugin/storage/cassandra/factory_test.go
```diff
@@ -25,10 +25,12 @@ import (
 	"github.com/jaegertracing/jaeger/pkg/cassandra"
 	"github.com/jaegertracing/jaeger/pkg/cassandra/mocks"
 	"github.com/jaegertracing/jaeger/pkg/config"
+
```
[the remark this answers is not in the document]
yurishkuro: ppppphhhhh

yurishkuro:
> do you wanna document this anywhere?

yes, once the UI support is added to actually archive traces.

PR #606, opened 2017-12-21
Adjuster fixes.

asp24:
For some app, it is hard to determine IP tag value. That's why for some cases clock-skew adjustment done wrong. In the PR multiple tags support was added.

### DIFF
--- a/model/adjuster/clockskew.go
+++ b/model/adjuster/clockskew.go
@@ -15,9 +15,7 @@
 package adjuster
 
 import (
-	"encoding/binary"
 	"fmt"
-	"net"
 	"time"
 
 	"github.com/jaegertracing/jaeger/model"
@@ -53,6 +51,10 @@ const (
 	warningFormatInvalidParentID = "invalid parent span IDs=%s; skipping clock skew adjustment"
 )
 
+var (
+	hostKeyTags = []string{"jaeger.hostname", "ip"}
+)
+
 type clockSkewAdjuster struct {
 	trace *model.Trace
 	spans map[model.SpanID]*node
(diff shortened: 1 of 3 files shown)


model/adjuster/clockskew.go
```diff
@@ -53,6 +51,10 @@ const (
 	warningFormatInvalidParentID = "invalid parent span IDs=%s; skipping clock skew adjustment"
 )
 
+var (
+	hostKeyTags = []string{"jaeger.hostname", "ip"}
```
yurishkuro: also should add `"hostname"`, it's used in [Go client](https://github.com/jaegertracing/jaeger-client-go/blob/master/constants.go#L36)

model/adjuster/clockskew_test.go
```diff
@@ -197,26 +197,28 @@ func TestClockSkewAdjuster(t *testing.T) {
...
-		{tag: model.Binary("ip", []byte{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0xff, 0xff, 1, 2, 3, 4}), hostKey: "1.2.3.4"},
-		{tag: model.Binary("ip", []byte{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 2, 3, 4}), hostKey: "::102:304"},
-		{tag: model.Binary("ip", []byte{1, 2, 3, 4, 5}), hostKey: ""},
-		{tag: model.Float64("ip", 123.4), hostKey: ""},
+		{tags: []model.KeyValue{model.String("ip", "1.2.3.4")}, hostKey: "1.2.3.4"},
+		{tags: []model.KeyValue{model.String("ipv4", "1.2.3.4")}, hostKey: ""},
+		{
+			tags: []model.KeyValue{
+				model.String("ip", "1.2.3.4"),
+				model.String("jaeger.hostname", "localhost"),
```
yurishkuro: shouldn't `ip` be given priority in this case?

yurishkuro:
I don't think looking for a common host tag is a good idea. Different services may be instrumented with different versions of Jaeger clients, in different languages. Ideally we'd want them all to converge on the same way of representing the host identity, but even if we did this in the latest versions of all clients it will take a long time for all services in a large org to upgrade.

The host identity is used to ensure that the timestamp adjustments from spans from the same host are the same. While this may look like it might apply to different services, in practice it's more likely that when service A calls service B the instances of those two services will be running on different hosts. So the problem is really avoiding ts adjustments from spans _in the same service_, which is a problem that does not require a common host tag because those spans will all come from the same version of the client lib, therefore any stable algorithm of selecting a host tag will do, such as what you had in the first commit by simply iterating through an array of known host tags (the order in the array implies a priority).

yurishkuro:
Per offline discussion, my proposal is:
  * we could build a smarter comparison, e.g.have a struct `Address { ip, hostname }` and a function
`IsSame(a1, a2) { return a1.ip == a2.ip || a1.hostname == a2.hostname }`
  * it still misses the equivalence case when ipv4 and ipv6 map to the same address, but it could also be taken into account, e.g. `if v4 := ipv6.IPv4(); v4 != nil { return v4 }`

yurishkuro:
Closing old inactive PR

PR #614, opened 2017-12-23
Add adjuster that removes bad span references

### LINKED ISSUE #613: Some spans have invalid span references
The Go client sometimes sends invalid references (issue https://github.com/jaegertracing/jaeger-client-go/issues/242) `{"refType":"CHILD_OF","traceIdLow":0,"traceIdHigh":0,"spanId":0}`.

The backend could use a sanitizer to prevent these references from getting into storage or an adjuster before returning them to the client.

yurishkuro:
Fixes #613

### DIFF
--- a/model/adjuster/bad_span_references.go
+++ b/model/adjuster/bad_span_references.go
@@ -1,0 +1,61 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


model/adjuster/bad_span_references.go
```diff
@@ -0,0 +1,61 @@
...
+
+package adjuster
+
+import (
+	"fmt"
+
+	"github.com/jaegertracing/jaeger/model"
+)
+
+// SpanReferences creates am adjuster that removes invalid span references, e.g. with traceID==0
```
black-adder: s/am/an

model/adjuster/bad_span_references.go
```diff
@@ -0,0 +1,61 @@
...
+		for _, span := range trace.Spans {
+			adjuster.adjust(span)
+		}
+		return trace, nil
+	})
+}
+
+type spanReferenceAdjuster struct{}
+
+func (s *spanReferenceAdjuster) adjust(span *model.Span) *model.Span {
```
black-adder: nit: I don't actually see any reason for these functions to be defined as pointer methods

PR #615, opened 2017-12-23
Add support for retrieving unadjusted/raw spans

### LINKED ISSUE #617: [query] Allow retrieving raw/unadjusted traces
For debugging purposes it is often useful to look at the original JSON of the trace as it is stored in the database. This can be done by supporting an extra parameter in the trace API `/api/traces/{traceID}/?raw=true`.

Caveat: there trace is still going to go through a series of transformation: dbmodel -> domain model -> ui/JSON model, but those transformations are more or less lossless.

yurishkuro:
Fixes #617


yurishkuro:
> Do you mind updating the PR description as to why this is useful for future me when I forget?

Added an actual issue (I thought we had it already). Once you try it you'll never forget it!

Issue #213, opened 2017-06-16
Support different service discovery systems


yurishkuro:
Depends on go-kit push model for discovery: https://github.com/go-kit/kit/pull/492

  * [ ] upgrade to 0.5 version of go-kit (not released yet)
  * [ ] decommission internal discovery API and replace it with go-kit
  * [ ] add command line options for configuring different service discovery systems
  * [ ] integration tests

yurishkuro:
A few thoughts since this issue has been opened.

The direct integration with service discovery is quickly becoming an anti-pattern given the rise of service meshes and proxies that abstract away that infrastructural concern. Even at Uber internally our agents are communicating to collectors via a service mesh proxy

If direct integration is still desired for performance reasons, I think integration of numerous SD systems directly into the Jaeger binary is the wrong approach, as it bloats the binary size on one hand, and on the other it pushes us more into dependencies hell. I think the solution here is a plugins architecture (#422) that decouples the main binary from the dependencies of individual SD implementations. An interesting observation here is that unlike storage plugins, the SD plugins have very low QPS and therefore are much more suitable for the sidecar plugin model where the main binary communicates to the plugin over, say, gRPC streams. Of course, the sidecar plugin model is actually a special case of in-process plugin model, so we still need to start there.

pavolloffay:
@yurishkuro do we want to still provide this functionality?

joshushua:
To be clear, no plug-in is available for Service Discovery? And no direct method of Service Discovery? So the recommended approach is to imitate Uber's use of a service mesh proxy?

yurishkuro:
since 1.9 you can use gRPC to communicate between agent and collector, which is understood by many LB solutions.

pavolloffay:
@yurishkuro can we close this with resolution "won't do"? 

We also point to this issue from docs

yurishkuro:
Closing per 
https://github.com/jaegertracing/jaeger/issues/213#issuecomment-354338201

PR #624, opened 2017-12-28
Use go-kit service discovery

black-adder:
Addresses https://github.com/jaegertracing/jaeger/issues/213 partially.

In order to maintain backwards compatibility for PeerListManager until Jaeger 2.0, I had to add a PeerListManagerV2 which embeds PeerListManager. There might be a better way to do this.

I'll add tests for PeerListManagerV2 if we're satisfied with the approach (probably a lot of copy pasta).


yurishkuro:
As we are unlikely to proceed in this direction, I am going to close & part this for now.

PR #630, opened 2018-01-03
Simplify install-glide command

### LINKED ISSUE #629: Glide issue is fixed now, we don't need the warning in Makefile any more.
While playing with the examples/hotrod demo, I received a warning of using glide 0.12.3 instead of glide 0.13 because of this issue: https://github.com/Masterminds/glide/issues/745 . Now this is fixed, and works well on my machine:

```
$ glide --version
glide version v0.13.1
```

kxxoling:
Since original bug is fixed in glide 0.13

Resolves #629

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,9 +93,7 @@ lint:
 
 .PHONY: install-glide
 install-glide:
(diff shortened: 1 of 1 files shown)


yurishkuro:
Thanks!

PR #631, opened 2018-01-03
Refactor healthcheck pkg, fix race condition in tests

yurishkuro:

### DIFF
--- a/pkg/healthcheck/handler.go
+++ b/pkg/healthcheck/handler.go
@@ -15,83 +15,130 @@
 package healthcheck
 
 import (
+	"context"
 	"net"
 	"net/http"
 	"strconv"
+	"sync/atomic"
 
 	"go.uber.org/zap"
 
 	"github.com/jaegertracing/jaeger/pkg/version"
 )
 
-// State represents the current health check state
-type State struct {
-	state   int
-	logger  *zap.Logger
-	onClose chan struct{}
+// Status represents the state of the service.
+type Status int
+
+const (
+	// Unavailable indicates the service is not able to handle requests
+	Unavailable Status = iota
+	// Ready indicates the service is ready to handle requests
+	Ready
+	// Broken indicates that the healthcheck itself is broken, not serving HTTP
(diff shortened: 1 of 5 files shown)


pkg/healthcheck/handler.go
```diff
@@ -15,83 +15,130 @@
...
+}
 
-	s := &http.Server{Handler: handler}
+// Serve starts HTTP server on the specified port.
+func (hc *HealthCheck) Serve(port int) (*HealthCheck, error) {
 	portStr := ":" + strconv.Itoa(port)
 	l, err := net.Listen("tcp", portStr)
 	if err != nil {
-		logger.Error("failed to listen", zap.Error(err))
+		hc.logger.Error("failed to listen", zap.Error(err))
```
vprithvi: Could you add more context similar to line 99, sth like Health Check server failed to listen, or so?

pkg/healthcheck/handler_test.go
```diff
@@ -16,105 +16,53 @@ package healthcheck_test
...
+	. "github.com/jaegertracing/jaeger/pkg/healthcheck"
+	"github.com/jaegertracing/jaeger/pkg/testutils"
 )
 
-func TestProperInitialState(t *testing.T) {
-	logger, _ := zap.NewDevelopment()
-	s, err := healthcheck.NewState(http.StatusServiceUnavailable, logger)
-	if err != nil {
-		t.Error("Could not start the health check server.", zap.Error(err))
+func TestStatusString(t *testing.T) {
```
vprithvi: What is the purpose of this test?

pkg/healthcheck/handler_test.go
```diff
@@ -16,105 +16,53 @@ package healthcheck_test
...
+	tests := map[Status]string{
+		Unavailable: "unavailable",
+		Ready:       "ready",
+		Status(-1):  "unknown",
```
vprithvi: Why does this not test `Broken` as well?

PR #632, opened 2018-01-03
Make agent tests more robust

yurishkuro:
1. The test was using default HTTP port instead of ephemeral port, which lead to unexpected failures if you happen to run some other jaeger components locally
1. The coverage was unstable because there was no waiting for the http server to shutdown and execute the error handler

### DIFF
--- a/cmd/agent/app/agent.go
+++ b/cmd/agent/app/agent.go
@@ -18,6 +18,7 @@ import (
 	"io"
 	"net"
 	"net/http"
+	"sync/atomic"
 
 	"go.uber.org/zap"
 
@@ -28,6 +29,7 @@ import (
 type Agent struct {
 	processors []processors.Processor
 	httpServer *http.Server
+	httpAddr   atomic.Value // string, set once agent starts listening
 	logger     *zap.Logger
 	closer     io.Closer
 }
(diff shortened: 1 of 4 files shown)


cmd/agent/app/agent_test.go
```diff
@@ -18,9 +18,11 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"strings"
 	"testing"
 	"time"
 
+	"github.com/jaegertracing/jaeger/pkg/testutils"
```
black-adder: import ordering is whack

cmd/agent/app/agent_test.go
```diff
@@ -93,4 +106,12 @@ func TestAgentStartStop(t *testing.T) {
 
 	agent.Stop()
 	assert.NoError(t, <-ch)
+
+	for i := 0; i < 1000; i++ {
+		if strings.Contains(logBuf.String(), "agent's http server exiting") {
```
black-adder: DRY the expected log?

cmd/agent/app/agent.go
```diff
@@ -53,18 +57,25 @@ func (a *Agent) Run() error {
 	if err != nil {
 		return err
 	}
+	a.httpAddr.Store(listener.Addr().String())
 	a.closer = listener
 	go func() {
 		if err := a.httpServer.Serve(listener); err != nil {
 			a.logger.Error("http server failure", zap.Error(err))
 		}
+		a.logger.Info("agent's http server exiting")
```
black-adder: wouldn't using a WaitGroup be "cleaner" than doing a log? unless you think this log adds a lot of value outside of code coverage reasons

yurishkuro: I think log is more useful / functional. WG implies a semantic guarantee of blocking Close(), which is not required here (aside from unit tests).

cmd/agent/app/agent_test.go
yurishkuro: it is dry. If you're referring to L116, then it doesn't matter what that line says (I'll change it).

PR #644, opened 2018-01-11
Implement credit store with unit test

isaachier:


cmd/agent/app/throttling/credit_store.go
```diff
@@ -0,0 +1,208 @@
...
+		balance = c.perOperationBalance[operationName]
+	} else {
+		if len(ca.perOperationRateLimiter) < ca.options.MaxOperations {
+			c.perOperationBalance[operationName] = balance
+		} else {
+			balance = c.defaultBalance
+		}
+	}
+
+	credits := ca.Withdraw(operationName, balance)
```
black-adder: maybe it's just me but I think this interface would be much simpler if rather than passing in the balance that the client already has, we pass in what it wants (ie ClientMaxBalance - balance) and if it's 0, we can just short circuit here. It makes creditAccruer more simpler (it shouldn't have to know about ca.options.ClientMaxBalance).

cmd/agent/app/throttling/credit_store.go
```diff
@@ -0,0 +1,208 @@
...
+type client struct {
+	id                  string
+	perOperationBalance map[string]float64
+	defaultBalance      float64
+	updateTime          time.Time
+}
+
+func (c *client) Spend(operationName string, credits float64) bool {
+	balance, ok := c.perOperationBalance[operationName]
+	if !ok {
```
black-adder: what's the reason we cant create a new perOperationBalance?

cmd/agent/app/throttling/credit_store.go
```diff
@@ -0,0 +1,208 @@
...
+	interval := now.Sub(t.lastTick)
+	t.lastTick = now
+	diff := interval.Seconds() * t.creditsPerSecond
+	t.balance = math.Min(t.balance+diff, t.maxBalance)
+	result := math.Min(t.balance, maxDrain)
+	t.balance -= result
+	return result
+}
+
+type client struct {
```
black-adder: this is gonna suck but in the future^TM, we're going to have to support multiple services from the same client. However, given how this is an internal piece of work, I think we can make the changes when required.

cmd/agent/app/throttling/credit_store.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+	s.timeNow = time.Now
+	return s
+}
+
+func (s *creditStore) Withdraw(clientID string, serviceName string, operationName string) float64 {
+	c := s.findOrCreateClient(clientID)
+	ca := s.findOrCreateCreditAccruer(serviceName)
+
+	// To keep client consistent with creditAccruer, we need to see if creditAccruer
```
black-adder: I don't see a big reason to keep these two consistent. I think it's fine for the client to track an operation without it having an entry in the creditAccruer. (ie most clients before we roll this out everywhere). This can be simplified to:

```
balance = c.getBalance() <- create a per operation balance if needed
ca.Widthdraw(operationName, clientMaxBalance) <- if operation isn't being tracked, create a new one or use default
```

cmd/agent/app/throttling/credit_store.go
isaachier: So client here means client to Jaeger agent from within service. How can there be more than one service/application running a Jaeger client?

cmd/agent/app/throttling/credit_store.go
isaachier: I'll explain below.

cmd/agent/app/throttling/credit_store.go
```diff
@@ -0,0 +1,208 @@
...
+
+	credits := rateLimiter.Drain(ca.options.ClientMaxBalance - balance)
+	return credits
+}
+
+type creditStore struct {
+	creditAccruers map[string]*creditAccruer
+	clients        map[string]*client
+	defaultOptions CreditAccruerOptions
+	lifetime       time.Duration
```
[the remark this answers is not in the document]
isaachier: Ya thought about that too. Will change.

cmd/agent/app/throttling/credit_store.go
isaachier: I guess the TTL makes this a non-issue, but my concern was the growth of keys for no apparent reason. Also, it seems sensible for me for clients to have the correct idea of what is going on in the rate limiter, not to expect per operation for every new operation. I can change it, but I'll try to see if I can simplify the logic and leave the behavior the same.

cmd/agent/app/throttling/credit_store.go
isaachier: Interesting idea. Will consider for next set of changes.

cmd/agent/app/throttling/credit_store.go
```diff
@@ -0,0 +1,208 @@
...
+
+func (s *creditStore) PurgeExpired() {
+	for _, ca := range s.creditAccruers {
+		if s.timeNow().Sub(ca.updateTime) >= s.lifetime {
+			delete(s.creditAccruers, ca.serviceName)
+		}
+	}
+	for _, client := range s.clients {
+		if s.timeNow().Sub(client.updateTime) >= s.lifetime {
+			delete(s.clients, client.id)
```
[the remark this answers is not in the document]
isaachier: Lol. The problem is where to hold that dead balance. The only place I can think of is in the `creditAccruer` outside of the rate limiter, and that adds more complexity. If you think it is worth it, I can explore the idea of making a fill method for the `tokenBucket` instead.

cmd/agent/app/throttling/credit_store.go
```diff
@@ -0,0 +1,208 @@
...
+
+type tokenBucket struct {
+	creditsPerSecond float64
+	balance          float64
+	maxBalance       float64
+	lastTick         time.Time
+	timeNow          func() time.Time // For testing
+}
+
+func (t *tokenBucket) Drain(maxDrain float64) float64 {
```
yurishkuro: The name Drain is out of place. It's either Withdraw or Spend.

cmd/agent/app/throttling/credit_store.go
```diff
@@ -0,0 +1,208 @@
...
+// limitations under the License.
+
+package throttling
+
+import (
+	"math"
+	"time"
+)
+
+type tokenBucket struct {
```
yurishkuro: please document all structs in this file

cmd/agent/app/throttling/credit_store.go
isaachier: I guess `Withdraw` makes sense here. I wasn't thinking about context when I wrote this (seeing as it was originally in another file). Will change.

cmd/agent/app/throttling/credit_store.go
isaachier: Definitely

cmd/agent/app/throttling/credit_store.go
```diff
@@ -0,0 +1,208 @@
...
+		} else {
+			rateLimiter = ca.defaultRateLimiter
+		}
+	}
+
+	credits := rateLimiter.Drain(ca.options.ClientMaxBalance - balance)
+	return credits
+}
+
+type creditStore struct {
```
yurishkuro: documentation. I don't follow the business meaning of this struct and the two maps it keeps.

isaachier: Will do ASAP

cmd/agent/app/throttling/credit_store.go
```diff
@@ -0,0 +1,213 @@
...
+		balance:          options.MaxBalance,
+		maxBalance:       options.MaxBalance,
+		timeNow:          timeNow,
+		lastTick:         timeNow(),
+	}
+}
+
+// Withdraw deducts as much of the total balance possible without exceeding
+// maxWithdrawal.
+func (t *tokenBucket) Withdraw(maxWithdrawal float64) float64 {
```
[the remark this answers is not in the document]
isaachier: I think it is better to combine this with a lock later than to add a member that isn't strictly required upfront. Consider the standard Go containers that do the same thing. I would add here if you can convince me it makes sense here vs. in the calling function/struct.

cmd/agent/app/throttling/credit_store.go
```diff
@@ -0,0 +1,213 @@
...
+	// perOperationBalance maintains per-operation balances of withdrawn credits
+	perOperationBalance map[string]float64
+	// updateTime is the last time client was updated
+	updateTime time.Time
+}
+
+// Spend depletes the balance from the client for the given operation. The usage
+// would be when a client submits spans for operation X, it is spending those
+// credits it requested for operation X.
+func (c *client) Spend(operationName string, credits float64) bool {
```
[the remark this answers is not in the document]
isaachier: IDK exactly the semantics of Go public symbols, but the entire type is private, so I thought it wouldn't matter. Rough equivalent is public method of private nested class in Java.

cmd/agent/app/throttling/credit_store.go
```diff
@@ -0,0 +1,213 @@
...
+		clients:        map[string]*client{},
+		options:        options,
+		ttl:            ttl,
+	}
+	s.timeNow = time.Now
+	return s
+}
+
+func (s *creditStore) Withdraw(clientID string, serviceName string, operationName string) float64 {
+	c := s.findOrCreateClient(clientID)
```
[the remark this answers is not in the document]
isaachier: Ya lol then again I have noticed Go doesn't really stress good variable names. One letter names aren't my style in general, but seemed to fit for Go. But in this case, I'll agree `c` and `ca` is confusing.

cmd/agent/app/throttling/credit_store.go
```diff
@@ -0,0 +1,213 @@
...
+	ca := s.findOrCreateCreditAccruer(serviceName)
+	balance := c.perOperationBalance[operationName]
```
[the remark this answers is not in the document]
isaachier: Makes sense

cmd/agent/app/throttling/credit_store.go
```diff
@@ -0,0 +1,213 @@
...
+		}
+		s.creditAccruers[serviceName] = ca
+	}
+	return ca
+}
+
+func (s *creditStore) findOrCreateClient(id string) *client {
+	c, ok := s.clients[id]
+	if !ok {
+		c = &client{
```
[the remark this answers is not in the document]
isaachier: Ya sort of avoided constructors on private types, especially when only used once, but makes sense.

cmd/agent/app/throttling/credit_store.go
```diff
@@ -0,0 +1,213 @@
...
+			id:                  id,
+			updateTime:          s.timeNow(),
+			perOperationBalance: map[string]float64{},
+		}
+		s.clients[id] = c
+	}
+	return c
+}
+
+func (s *creditStore) PurgeExpired() {
```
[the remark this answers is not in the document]
isaachier: Same reason I didn't put a mutex here. Go does thread multiplexing for you so starting a new goroutine doesn't cost anything, but I believe it slightly complicates the agent unnecessarily. Concurrency at multiple levels complicates code in my experience (consider the client race condition https://github.com/jaegertracing/jaeger-client-go/pull/240).

cmd/agent/app/throttling/override_map.go
```diff
@@ -0,0 +1,70 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package throttling
+
+// OverrideMap is an interface describing a configuration map with a fixed
+// number of override entries, falling back on a default value.
+type OverrideMap interface {
```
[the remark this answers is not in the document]
isaachier: Most of the annoying if statements are gone now. :)

cmd/agent/app/throttling/credit_store.go
```diff
@@ -0,0 +1,213 @@
...
+	c.updateTime = s.timeNow()
+	return c.Spend(operationName, credits)
+}
+
+func (s *creditStore) findOrCreateCreditAccruer(serviceName string) *creditAccruer {
+	ca, ok := s.creditAccruers[serviceName]
+	if !ok {
+		defaultRateLimiter := newTokenBucket(s.options, s.timeNow)
+		perOperationRateLimiter := NewOverrideMap(s.options.MaxOperations, defaultRateLimiter)
+		ca = &creditAccruer{
```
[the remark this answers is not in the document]
isaachier: Sure

cmd/agent/app/throttling/credit_store.go
```diff
@@ -0,0 +1,261 @@
...
+	}
+}
+
+// Spend depletes the balance from the client for the given operation. The usage
+// would be when a client submits spans for operation X, it is spending those
+// credits it requested for operation X.
+func (c *client) Spend(operationName string, credits float64) bool {
+	balance := c.perOperationBalance[operationName]
+	if credits > balance {
+		return false
```
yurishkuro: this is unhappy path, but simply returning seems wrong. The client submitted the span, so it already spent the credits, we're just trying to record that fact. If the available balance is less than spent credits, then (a) we either have an error in the protocol of a dishonest client, so we need to report that somewhere, and (b) it's still better to reduce the balance to 0.

cmd/agent/app/throttling/credit_store.go
```diff
@@ -0,0 +1,261 @@
...
+
+// CreditAccruerOptions provides values to be used with a creditAccruer object
+type CreditAccruerOptions struct {
+	MaxOperations    int
+	CreditsPerSecond float64
+	MaxBalance       float64
+	ClientMaxBalance float64
+}
+
+type creditAccruer struct {
```
yurishkuro: documentation

cmd/agent/app/throttling/credit_store.go
```diff
@@ -0,0 +1,261 @@
...
+}
+
+// client represents a Jaeger-instrumented caller. Every Jaeger client
+// corresponds to an instrumented service, but it may be one of many instances
+// of that service. Therefore, there is a one-to-many relationship between
+// services and clients. The clients must request credits from the service's
+// creditAccruer and they all share the same underlying tokenBucket.
+// For example, if one client requests a withdrawal and receives all the service
+// credits, the next client withdrawal request will be rejected until the
+// credits refill.
```
yurishkuro: great doc

cmd/agent/app/throttling/credit_store.go
```diff
@@ -0,0 +1,261 @@
...
+	return &creditAccruer{
+		options:                 options,
+		serviceName:             serviceName,
+		updateTime:              timeNow(),
+		timeNow:                 timeNow,
+		perOperationRateLimiter: perOperationRateLimiter,
+	}
+}
+
+// Withdraw credits from the creditAccruer for the given operationName with an
```
yurishkuro: Withdraw deducts the credits...

cmd/agent/app/throttling/credit_store.go
```diff
@@ -0,0 +1,261 @@
...
+type CreditStore interface {
+	io.Closer
+	// Withdraw deducts as many credits from the service on behalf of the client
+	// without exceeding MaxClientBalance. The client is identified by a unique
+	// clientID string. The credits are deducted from the specific operation on
+	// the service if present, otherwise from the default credit pool.
+	Withdraw(clientID string, serviceName string, operationName string) float64
+	// Spend credits already allocated to this client on a given operation.
+	Spend(clientID string, operationName string, credits float64) bool
+}
```
yurishkuro: we don't need to introduce interface at this point, just make the struct public.

cmd/agent/app/throttling/credit_store.go
```diff
@@ -0,0 +1,261 @@
...
+
+// CreditStore manages the relationship between service instance clients and
+// backend storage credits. Each client must withdraw credits in order to submit
+// trace data to the Jaeger agent. These credits are generated on behalf of the
+// service, of which there can be many instances, and thus many clients.
+type CreditStore interface {
+	io.Closer
+	// Withdraw deducts as many credits from the service on behalf of the client
+	// without exceeding MaxClientBalance. The client is identified by a unique
```
yurishkuro: nit: MaxClientBalance is not part of the interface, so unclear what it means

cmd/agent/app/throttling/credit_store.go
```diff
@@ -0,0 +1,261 @@
...
+	// clientID string. The credits are deducted from the specific operation on
+	// the service if present, otherwise from the default credit pool.
+	Withdraw(clientID string, serviceName string, operationName string) float64
+	// Spend credits already allocated to this client on a given operation.
+	Spend(clientID string, operationName string, credits float64) bool
```
yurishkuro: need to rethink Withdraw and Spend terminology, the pair doesn't quite make sense.

Conceptually the first operation the client does is to reserve a certain amount of credits, such that they are not available to other clients, but not yet spent by this client. Then it spends from the reserved amount. So maybe Reserve/Spend.

The lack of serviceName in the Spend function looks a bit odd. Maybe clientID should be the second argument in the first method, because service->client  is 1-*.

cmd/agent/app/throttling/credit_store.go
```diff
@@ -0,0 +1,261 @@
...
+		clients:        map[string]*client{},
+		options:        options,
+		ttl:            ttl,
+	}
+	s.timeNow = time.Now
+	s.ticker = time.NewTicker(10 * time.Second)
+	s.done = make(chan bool)
+	s.waitGroup.Add(1)
+	go func() {
+		defer s.waitGroup.Done()
```
yurishkuro: ticket should be defined here, and defer-closed. The top level struct doesn't need to know about it.

cmd/agent/app/throttling/credit_store.go
```diff
@@ -0,0 +1,261 @@
...
+	updateTime              time.Time
+	timeNow                 func() time.Time // For testing
+}
+
+func newCreditAccruer(serviceName string, options CreditAccruerOptions, timeNow func() time.Time) *creditAccruer {
+	defaultRateLimiter := newTokenBucket(options, timeNow)
+	perOperationRateLimiter := NewOverrideMap(options.MaxOperations, defaultRateLimiter)
+	return &creditAccruer{
+		options:                 options,
+		serviceName:             serviceName,
```
yurishkuro: why do we need service name in this struct?

cmd/agent/app/throttling/credit_store.go
```diff
@@ -0,0 +1,261 @@
...
+		}
+	}()
+	return s
+}
+
+func (s *creditStore) Withdraw(clientID string, serviceName string, operationName string) float64 {
+	s.Lock()
+	defer s.Unlock()
+	c := s.findOrCreateClient(clientID)
+	accruer := s.findOrCreateCreditAccruer(serviceName)
```
yurishkuro: nit: "credit accruer" doesn't sound like a good term in this file, since it does not explain how, for example, it's different from credit store. It would be nice to minimize the number of terms. Isn't it just a collection of token buckets per operation?

cmd/agent/app/throttling/override_map.go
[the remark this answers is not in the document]
yurishkuro: only introduce interfaces when you actually need them. Right now it can simply stay as a private struct.

cmd/agent/app/throttling/credit_store.go
```diff
@@ -0,0 +1,261 @@
...
+	c.perOperationBalance[operationName] = balance
+	return true
+}
+
+// CreditAccruerOptions provides values to be used with a creditAccruer object
+type CreditAccruerOptions struct {
+	MaxOperations    int
+	CreditsPerSecond float64
+	MaxBalance       float64
+	ClientMaxBalance float64
```
yurishkuro: - what is the difference between these two max-balance values?
- we will vary likely need different limits and refill rates per service, is this design extensible?

cmd/agent/app/throttling/credit_store.go
isaachier: Right. My concern here is what to do when our expectations are not met. This is where the C programmer in me would write `assert(credits <= balance)`, but Go clearly got rid of `assert` for a reason. In any case, I agree that setting the balance to zero makes more sense, so I'll do that. Also, will add some logging/reporting for the error.

cmd/agent/app/throttling/credit_store.go
isaachier: Good idea

cmd/agent/app/throttling/credit_store.go
isaachier: Thank you

cmd/agent/app/throttling/credit_store.go
isaachier: Good catch

cmd/agent/app/throttling/credit_store.go
isaachier: OK. I'm not sure why I did this in retrospect.

cmd/agent/app/throttling/credit_store.go
isaachier: Right, going to move to struct anyway. Maybe can clarify better there.

cmd/agent/app/throttling/credit_store.go
isaachier: Originally, I had done `Withdraw` and `Deposit`, but realized `Deposit` would imply it went back to the service credit pool. Therefore, switched `Deposit` to `Spend`, but `Reserve`/`Spend` or `Acquire`/`Release` might make more sense.

Regarding the second issue, I even considered adding a `serviceName` argument that would remain unused until I found a use for it, just because the consistency bothers me here too. I'll try your suggestion to see if it looks cleaner.

cmd/agent/app/throttling/credit_store.go
isaachier: Sure. Out of curiosity, is there a better approach here? A `Ticker`, `channel`, and `WaitGroup` seems like overkill for such a common operation. Even `Mutex` and `ConditionVariable` looks easier to implement.

cmd/agent/app/throttling/credit_store.go
isaachier: Good question. Will remove.

cmd/agent/app/throttling/credit_store.go
isaachier: True. I was considering using a financial analogy throughout to simplify things. The earlier design used the term credit accruer, so I kept it, but I'm thinking this might work better:

* `CreditBank` to replace `CreditStore`
* `account` to replace `creditAccruer`

Then at least it is easy to explain that several clients can share an account in a bank. Let me know if you prefer using existing terms or the analogy I just described (or another analogy).

cmd/agent/app/throttling/credit_store.go
isaachier: I'll add this to the documentation, but `MaxBalance` describes the maximum for the token bucket, and `ClientMaxBalance` refers to the maximum a client can attain before it must spend the credits (`Withdraw` returns `0` for these clients).

The reason I use a `struct` instead of using constants was to allow for extensibility. In the future, we can use a default for most services, and add overrides for specific services that use custom values.

cmd/agent/app/throttling/override_map.go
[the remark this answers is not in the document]
isaachier: Fair enough.

cmd/agent/app/throttling/credit_store.go
yurishkuro: did you need the wait group?  All it does is block the main function until the go routine is running. Some times its useful, not sure if it is in this case. The struct certainly doesn't need to know about the wait group, it can be just a local var.

Re ticker/channel, given that this is a timer loop I don't know if another way if implementing it. And since you're already using select for the ticker channel, the `done` channel makes perfect sense.

cmd/agent/app/throttling/credit_store.go
yurishkuro: How about Throttler as the main class? Throttling is its primary business function for us, and the "credits" are an implementation detail.

I like `account`.

cmd/agent/app/throttling/credit_store.go
isaachier: I guess it isn't strictly necessary but guarantees `Close` is actually finishes before proceeding. Can see this being useful in general, if not for this specific case. I guess this is more of a question for the Go language developers. After a bit of research, it seems the Go API for condition variables is intentionally limited to avoid my alternative approach.

cmd/agent/app/throttling/credit_store.go
isaachier: Cool will do that

cmd/agent/app/throttling/override_map.go
```diff
@@ -0,0 +1,62 @@
...
+type overrideMap struct {
+	maxOverrides int
+	overrides    map[string]interface{}
+	defaultValue interface{}
+}
+
+// newOverrideMap creates a new OverrideMap with maxOverrides overrides and a
+// fallback value of defaultValue.
+func newOverrideMap(maxOverrides int, defaultValue interface{}) *overrideMap {
+	o := &overrideMap{
```
black-adder: nit: return can be inlined

cmd/agent/app/throttling/throttler.go
```diff
@@ -0,0 +1,294 @@
...
+func (a *account) Withdraw(operationName string, maxWithdrawal float64) float64 {
+	var rateLimiter *tokenBucket
+	if a.perOperationRateLimiter.Has(operationName) ||
+		a.perOperationRateLimiter.IsFull() {
+		rateLimiter = a.perOperationRateLimiter.Get(operationName).(*tokenBucket)
+	} else {
+		rateLimiter = newTokenBucket(a.options, a.timeNow)
+		a.perOperationRateLimiter.Set(operationName, rateLimiter)
+	}
+	credits := rateLimiter.Withdraw(maxWithdrawal)
```
black-adder: nit: inline return

cmd/agent/app/throttling/throttler.go
```diff
@@ -0,0 +1,294 @@
...
+		accounts: map[string]*account{},
+		clients:  map[string]*client{},
+		options:  options,
+		timeNow:  time.Now,
+	}
+	t.ticker = time.NewTicker(t.options.PurgeInterval)
+	t.done = make(chan bool)
+	t.waitGroup.Add(1)
+	go func() {
+		defer t.waitGroup.Done()
```
black-adder: ticker could be started and killed within the go routine.

cmd/agent/app/throttling/throttler.go
```diff
@@ -0,0 +1,294 @@
...
+	if err != nil {
+		err = errors.New(err.Error() + ", serviceName " + serviceName)
+	}
+	return err
+}
+
+// Close closes the throttler and stops all background goroutines.
+func (t *Throttler) Close() error {
+	t.done <- true
+	close(t.done)
```
black-adder: afaik, `close(t.done)` will cause the for loop to exit, the `t.done <- true` isn't required.

cmd/agent/app/throttling/throttler.go
```diff
@@ -0,0 +1,294 @@
...
+// without exceeding MaxClientBalance. The client is identified by a unique
+// clientID string. The credits are deducted from the specific operation on
+// the service if present, otherwise from the default credit pool.
+func (t *Throttler) Withdraw(serviceName string, clientID string, operationName string) float64 {
+	t.Lock()
+	defer t.Unlock()
+	c := t.findOrCreateClient(clientID)
+	a := t.findOrCreateAccount(serviceName)
+	balance := c.perOperationBalance[operationName]
+	maxWithdrawal := t.options.ClientMaxBalance - balance
```
black-adder: what are your plans for when the configs for a specific service are updated and the ClientMaxBalance < balance, maybe a max(0, ClientMaxBalance - balance) here for future proofing?

cmd/agent/app/throttling/throttler.go
isaachier: Good point.

cmd/agent/app/throttling/override_map.go
isaachier: Sure

cmd/agent/app/throttling/throttler.go
isaachier: Sure

cmd/agent/app/throttling/throttler.go
isaachier: Not even going to bother killing it, just going to make it a local variable.

cmd/agent/app/throttling/throttler.go
isaachier: I thought about it, but considered it an invalid case. Again, great place for an assertion if it existed. I can change it for now.

PR #656, opened 2018-01-18
Use elasticsearch bulk API

### LINKED ISSUE #439: Elasticsearch indexing performance issues
Currently ES store does not use bulk API and issues a request to ES for every span. This is not very efficient, but in combination with transaction log flushing leads to performance issues.

The [default setting](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules-translog.html#_translog_settings_2) of `index.translog.durability` is `request` which means that there is a fsync after every request. This leads to pretty high IOPS when not using bulk API. 
For example it's the main factor limiting throughput on AWS instances with small EBS drives - that was the reason why I was experiencing 60+ seconds of delay in indexing and losing 50%-75% of spans because the internal queue was always full.

One (simple) solution would be to change `index.translog.durability` to `async` in the index settings and make it flush asynchronously. In case of a node crash few seconds of data would be lost which might be acceptable in practice.
Another option is to start using bulk API. https://github.com/olivere/elastic/wiki/BulkProcessor can be utilized to avoid writing batching logic by hand.

pavolloffay:
Fixes #439 

Add BulkService API for storing documents. It caches requests until a threshold is reached - size, time, number of operations. Then it reports bulk to ES. 

Performance: store 300k spans
Without bulk API ~15min
WIth bulk API: ~40s 

This currently removes metrics, once this is merged I will do PR to include stats https://github.com/olivere/elastic/wiki/BulkProcessor#stats


yurishkuro:
could you elaborate in the description what exactly this change is doing? The SpanWriter API is still accepting one span at a time, so what do we mean by bulk? Does the client now batch the write requests? If so, how does this affect the success/failure metrics emitted?

pavolloffay:
API is collecting spans until a threshold is reached - size, number of operations, time. Then it reports what it has in the buffer.

I have removed metrics for now e.g. span write duration as there is no way how to find out the number. There is a way how to get stats from bulk service, but that is not per index name (e.g. spans, dependencies). I would prefer to add it in a separate PR https://github.com/olivere/elastic/wiki/BulkProcessor#stats

plugin/storage/es/spanstore/writer_test.go
```diff
@@ -235,75 +197,6 @@ func TestSpanIndexName(t *testing.T) {
 	assert.Equal(t, "jaeger-service-1995-04-21", serviceIndexName)
 }
 
-func TestCheckAndCreateIndex(t *testing.T) {
```
yurishkuro: what's the argument for deleting this test?

yurishkuro:
awesome perf bump

pavolloffay:
Bulk API is always enabled. There are flags to configure it e.g. flush interval, size or flush period.

plugin/storage/es/spanstore/writer_test.go
pavolloffay: It does not affect test coverage, I think some of them were repeating stuff

pkg/es/client.go
```diff
@@ -27,6 +27,7 @@ type Client interface {
 	Index() IndexService
 	Search(indices ...string) SearchService
 	MultiSearch() MultiSearchService
+	Close() error
```
pavolloffay: What about embedding io.Closer here?

I would like to close this in in main case of sigterm, however there is present only spanwriter. Do we want to add close to spanwriter too?

yurishkuro: Lately I prefer not adding Close to interfaces. The main often creates services with explicit type, i.e. `*SomeStruct`, in which case it already has access to Close(). In cases when we use an abstract factory pattern and get an interface back, we can still call close via conditional cast:
```go
if closer, ok := someComponent.(io.Closer); ok {
    closer.Close()
}
```

plugin/storage/es/options.go
```diff
@@ -60,12 +64,16 @@ func NewOptions(primaryNamespace string, otherNamespaces ...string) *Options {
...
+				Username:          "",
+				Password:          "",
+				Sniffer:           false,
+				MaxSpanAge:        72 * time.Hour,
+				NumShards:         5,
+				NumReplicas:       1,
+				BulkSize:          5 * 1000 * 1000,
+				BulkWorkers:       1,
+				BulkActions:       1000,
+				BulkFlushInterval: 0,
```
mabn: Maybe set the default to 100ms? 
I think that with 0 spans might not be flushed if someone is just testing and sends less than 1000 spans. Or the added delay will be high at low throughputs.

pavolloffay: 100ms seems quite low to leverage bulk API. If somebody is testing even 1s could be fine, until you open a browser and refresh spans will be there. 

Here are some tests, flush interval: test duration
100ms: 100s
200ms: 80s
300ms: 77s
500ms: ~76s
1s: ~75s 

I would go with at least 200ms

pavolloffay:
It should be ready for review

pkg/es/client.go
pavolloffay: I had to add `close` here to be able close the bulk service it in es span writer.

pkg/es/config/config.go
```diff
@@ -15,43 +15,72 @@
...
+	Username          string
+	Password          string
+	Sniffer           bool          // https://github.com/olivere/elastic/wiki/Sniffing
+	MaxSpanAge        time.Duration `yaml:"max_span_age"` // configures the maximum lookback on span reads
+	NumShards         int64         `yaml:"shards"`
+	NumReplicas       int64         `yaml:"replicas"`
+	BulkSize          int
+	BulkWorkers       int
+	BulkActions       int
+	BulkFlushInterval time.Duration
```
yurishkuro: do we want to add yaml tags for these? Because they are multi-word names, they will look differently otherwise

yurishkuro:
do we cover the insert behavior in the integration tests with real ES?

pkg/es/config/config.go
pavolloffay: where are these tags used?

pkg/es/wrapper.go
```diff
@@ -100,37 +107,39 @@ func (c ESIndicesCreateService) Do(ctx context.Context) (*elastic.IndicesCreateR
...
 // Type calls this function to internal service.
 func (i ESIndexService) Type(typ string) IndexService {
-	return WrapESIndexService(i.indexService.Type(typ))
+	return WrapESIndexService(i.bulkIndexReq.Type(typ), i.bulkService)
 }
 
 // Id calls this function to internal service.
 func (i ESIndexService) Id(id string) IndexService {
-	return WrapESIndexService(i.indexService.Id(id))
+	a := WrapESIndexService(i.bulkIndexReq.Id(id), i.bulkService)
```
black-adder: nit: inline

pkg/es/wrapper.go
```diff
@@ -100,37 +107,39 @@ func (c ESIndicesCreateService) Do(ctx context.Context) (*elastic.IndicesCreateR
...
 // BodyJson calls this function to internal service.
 func (i ESIndexService) BodyJson(body interface{}) IndexService {
-	return WrapESIndexService(i.indexService.BodyJson(body))
+	return WrapESIndexService(i.bulkIndexReq.Doc(body), i.bulkService)
 }
 
-// Do calls this function to internal service.
-func (i ESIndexService) Do(ctx context.Context) (*elastic.IndexResponse, error) {
-	return i.indexService.Do(ctx)
+// Add add the request to bulk service
```
black-adder: s/Add add/Add adds

yurishkuro:
I suggest rebasing from master, there is a fix for flaky travis install

pavolloffay:
I will merge, if there are more comments I will apply them in a separate PR

PR #669, opened 2018-01-30
Klarrio storage

radekg:
PR will extra jaeger storage types. Klarrio specific.

Storage 1: `authorizingproxy`
---
Given a collector with --span-storage.type=authorizingproxy (or whatever it will be called when complete), forward spans to another collector using an embedded agent reporter.
Optionally, forward only if a certain condition is met.


radekg:
UX failure. Apologies, please ignore.

PR #670, opened 2018-01-31
Delete empty test

vprithvi:


vprithvi:
@black-adder I think it does! Perhaps these files should have a comment stating the reason why they exist.

PR #672, opened 2018-02-01
Extend integration tests to memory storage

yurishkuro:

### DIFF
--- a/plugin/storage/integration/integration_test.go
+++ b/plugin/storage/integration/integration_test.go
@@ -18,7 +18,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
-	"os"
 	"strings"
 	"testing"
 	"time"
@@ -38,19 +37,20 @@ const (
 )
 
 type StorageIntegration struct {
+	// TODO make these public
 	logger           *zap.Logger
-	spanWriter       spanstore.Writer
-	spanReader       spanstore.Reader
-	dependencyWriter dependencystore.Writer
-	dependencyReader dependencystore.Reader
+	SpanWriter       spanstore.Writer
+	SpanReader       spanstore.Reader
+	DependencyWriter dependencystore.Writer
+	DependencyReader dependencystore.Reader
 
 	// cleanUp() should ensure that the storage backend is clean before another test.
 	// called either before or after each test, and should be idempotent
-	cleanUp func() error
+	CleanUp func() error
 
 	// refresh() should ensure that the storage backend is up to date before being queried.
 	// called between set-up and queries in each test
-	refresh func() error
+	Refresh func() error
 }
 
 // === SpanStore Integration Tests ===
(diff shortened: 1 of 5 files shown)


plugin/storage/integration/memstore_test.go
```diff
@@ -0,0 +1,60 @@
...
+func (s *MemStorageIntegrationTestSuite) refresh() error {
+	return nil
+}
+
+func (s *MemStorageIntegrationTestSuite) cleanUp() error {
+	return s.initialize()
+}
+
+func TestMemoryStorage(t *testing.T) {
+	// if os.Getenv("STORAGE") != "memory" {
```
black-adder: is this intended?

plugin/storage/integration/integration_test.go
```diff
@@ -277,21 +298,18 @@ func (s *StorageIntegration) IntegrationTestGetDependencies(t *testing.T) {
 			CallCount: uint64(3),
 		},
 	}
-	require.NoError(t, s.dependencyWriter.WriteDependencies(time.Now(), expected))
-	require.NoError(t, s.refresh())
-	actual, err := s.dependencyReader.GetDependencies(time.Now(), 5*time.Minute)
+	require.NotNil(t, s.DependencyWriter, "DependencyWriter is required")
```
black-adder: doesn't the L282 check catch this?

plugin/storage/integration/integration_test.go
```diff
@@ -118,122 +127,128 @@ func (s *StorageIntegration) IntegrationTestGetOperations(t *testing.T) {
 		t.Log("\t Expected:", expected)
 		t.Log("\t Actual  :", actual)
 	}
-	assert.NoError(t, s.cleanUp())
 }
 
-func (s *StorageIntegration) IntegrationTestGetTrace(t *testing.T) {
-	t.Log("Testing GetTrace ...")
+func (s *StorageIntegration) testGetTrace(t *testing.T) {
```
vprithvi: Does it make sense to leave this exposed? 
I feel that it might be useful to have the ability to run a subset of tests when developing new storages

plugin/storage/integration/integration_test.go
```diff
@@ -118,122 +127,128 @@ func (s *StorageIntegration) IntegrationTestGetOperations(t *testing.T) {
...
+	queryTestCases := loadAndParseQueryTestCases(t)
+
+	// Each query test case only specifies matching traces, but does not provide counterexamples.
+	// To improve coverage we get all possible traces and store all of them before running queries.
+	allTraceFixtures := make(map[string]*model.Trace)
+	expectedTracesPerTestCase := make([][]*model.Trace, 0, len(queryTestCases))
+	for _, queryTestCase := range queryTestCases {
+		var expected []*model.Trace
+		for _, traceFixture := range queryTestCase.ExpectedFixtures {
+			trace, ok := allTraceFixtures[traceFixture]
```
black-adder: not sure what this is used for, we don't seem to be updating the cache after creating a new trace

plugin/storage/integration/integration_test.go
```diff
@@ -38,19 +37,20 @@ const (
 )
 
 type StorageIntegration struct {
+	// TODO make these public
```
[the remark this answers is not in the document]
yurishkuro: except for logger

plugin/storage/integration/integration_test.go
yurishkuro: we can revisit, but it seems a storage that doesn't support all storage methods is not super useful

plugin/storage/integration/integration_test.go
yurishkuro: it's used in L172 and then 174 and then 179

plugin/storage/integration/integration_test.go
yurishkuro: yeah, left-over

plugin/storage/integration/memstore_test.go
yurishkuro: left-over

plugin/storage/integration/integration_test.go
black-adder: sorry, I meant allTraceFixtures, doesn't seem like we set anything in it?

Issue #661, opened 2018-01-22
Add GetSamplingStrategy handler to collector


yurishkuro:
The agent proxies all sampling requests to collector, but there is no handler there :-(

PR #674, opened 2018-02-02
Add sampling handler and sampling strategy store

black-adder:
This adds a tchannel sampling handler and a static sampling strategy store that reads sampling strategies from a json file.

I added a factory for this so that when we move adaptive sampling over to open source, we can easily extend the existing factory with adaptive sampling and all its flags (I've counted 15 flags at the time of writing).

Locations of files might be up for debate (should this belong in pkg and plugin?)
Do I even need a separate sampling strategy store? should I just use the sampling handler interface and have a factory for that?

Fixes https://github.com/jaegertracing/jaeger/issues/661


plugin/sampling/strategystore/static/options.go
```diff
@@ -0,0 +1,42 @@
...
+package static
+
+import (
+	"flag"
+
+	"github.com/spf13/viper"
+)
+
+const (
+	samplingStrategies = "sampling.strategies"
```
yurishkuro: I would prefer a bit more explicit name "sampling.strategies-file"

plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -0,0 +1,107 @@
...
+
+// GetSamplingStrategy implements StrategyStore#GetSamplingStrategy.
+func (h *strategyStore) GetSamplingStrategy(serviceName string) (*sampling.SamplingStrategyResponse, error) {
+	if strategy, ok := h.serviceStrategies[serviceName]; ok {
+		return strategy, nil
+	}
+	return h.defaultStrategy, nil
+}
+
+// TODO good candidate for a global util function
```
yurishkuro: how much overlap is in this code with what the client is doing when parsing JSON response from the agent? Maybe we can combine that all into jaeger-lib?

pavolloffay:
@black-adder does this fix https://github.com/jaegertracing/jaeger/issues/661?

plugin/sampling/strategystore/static/strategy_store.go
black-adder: This TODO was originally meant for creating a util function for reading a json file and unmarshalling it. I didn't have as lofty goal of moving the client and this logic into jaeger-lib. I don't think it's similar enough to warrant that.

pkg/sampling/strategystore/constants.go
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) 2018 The Jaeger Authors.
```
yurishkuro: yes, the `pkg` is rather suspicious for this. It is meant for things not directly related to Jaeger. The simplest approach would be to define the storage interface in `cmd/collector/sampling/strategystore/interface.go`

pkg/sampling/strategystore/factory.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/uber/jaeger-lib/metrics"
+	"go.uber.org/zap"
+)
+
+// Factory defines an interface for a factory that can create implementations of different strategy storage components.
+// Implementations are also encouraged to implement plugin.Configurable interface.
+//
+// See also
+//
+// plugin.Configurable
```
yurishkuro: Configurable also has AddFlags(flagSet *flag.FlagSet). Do you actually need to require Configurable methods in the Factory interface?

pkg/sampling/strategystore/constants.go
```diff
@@ -0,0 +1,44 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package strategystore
+
+import (
+	"github.com/jaegertracing/jaeger/thrift-gen/sampling"
+)
+
+const (
```
yurishkuro: is this needed in the "interface" package? Could it not be placed in the static impl?

pkg/sampling/strategystore/strategy.go
```diff
@@ -0,0 +1,34 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package strategystore
+
+// Strategy defines a sampling strategy. Type can be "probabilistic" or "ratelimiting"
+// and Param will represent "sampling probability" and "max traces per second" respectively.
+type Strategy struct {
```
yurishkuro: these types look like private intermediate model used by static impl. Or did you mean the StrategyStorage interface to return these?

plugin/sampling/strategystore/factory_config.go
```diff
@@ -0,0 +1,42 @@
...
+
+// FactoryConfig tells the Factory what sampling type it needs to create.
+type FactoryConfig struct {
+	StrategyStoreType string
+}
+
+// FactoryConfigFromEnv reads the desired sampling type from the SAMPLING_TYPE environment variable. Allowed values:
+//   * `static` - built-in
+//   * `adaptive` - built-in // TODO
+func FactoryConfigFromEnv() FactoryConfig {
```
yurishkuro: are you planning to add changes to collector.main() here?   Right now this function is not used.

pkg/sampling/strategystore/factory.go
black-adder: It doesn't have to be but it makes life easier.

pkg/sampling/strategystore/strategy.go
black-adder: made private

pkg/sampling/strategystore/constants.go
black-adder: done, should I have the static/adaptive implementations in plugin or should I move the whole thing over to cmd/collector/app?

PR #679, opened 2018-02-05
Remove the override of GOMAXPROCS

oiooj:
From Go 1.5, the default setting of `GOMAXPROCS` is the number of
CPUs available. we might as well remove this so people can set
GOMAXPROCS to a custom value using environment variables.

> https://golang.org/doc/go1.5#runtime

Signed-off-by: kun <oiooj@qq.com>


black-adder:
thanks!

PR #680, opened 2018-02-05
Add options to Cassandra to control indexing

vprithvi:
- Add `StoreIndexesOnly` and `StoreWithoutIndexing` options to the cassandra
  storage which controls whether indexes are written along with spans.

- This allows us to deploy, scale, and prioritize  writers independently for
  indexing and storage workloads.

### DIFF
--- a/plugin/storage/cassandra/spanstore/writer.go
+++ b/plugin/storage/cassandra/spanstore/writer.go
@@ -66,9 +66,8 @@ const (
 )
 
 const (
-	indexAndStore = storageMode(iota)
-	indexOnly
-	storeOnly
+	storeFlag = storageMode(1 << iota)
+	indexFlag
 )
 
 type storageMode uint8
@@ -134,19 +133,12 @@ func (s *SpanWriter) Close() error {
 // WriteSpan saves the span into Cassandra
 func (s *SpanWriter) WriteSpan(span *model.Span) error {
 	ds := dbmodel.FromDomain(span)
-	switch s.storageMode {
-	case storeOnly:
+	if s.storageMode&storeFlag == storeFlag {
 		if err := s.writeSpan(span, ds); err != nil {
 			return err
 		}
-	case indexOnly:
-		if err := s.writeIndexes(span, ds); err != nil {
-			return err
-		}
-	default:
-		if err := s.writeSpan(span, ds); err != nil {
-			return err
-		}
+	}
+	if s.storageMode&indexFlag == indexFlag {
 		if err := s.writeIndexes(span, ds); err != nil {
 			return err
 		}
--- a/plugin/storage/cassandra/spanstore/writer.go
+++ b/plugin/storage/cassandra/spanstore/writer.go
@@ -63,10 +63,12 @@ const (
 	defaultNumBuckets = 10
 
 	durationBucketSize = time.Hour
+)
 
-	indexOnly     = storageMode(1)
-	storeOnly     = storageMode(2)
-	indexAndStore = storageMode(3)
+const (
+	indexAndStore = storageMode(iota)
+	indexOnly     = storageMode(iota)
+	storeOnly     = storageMode(iota)
 )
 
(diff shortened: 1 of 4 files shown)


plugin/storage/cassandra/spanstore/writer.go
```diff
@@ -63,8 +63,13 @@ const (
 	defaultNumBuckets = 10
 
 	durationBucketSize = time.Hour
+
+	indexOnly     = storageMode(1)
```
yurishkuro: * use the `iota` pattern
* I would put indexAndStore as the first item (the default mode)

plugin/storage/cassandra/spanstore/writer.go
```diff
@@ -65,6 +65,13 @@ const (
 	durationBucketSize = time.Hour
 )
 
+const (
+	indexAndStore = storageMode(iota)
+	indexOnly     = storageMode(iota)
+	storeOnly     = storageMode(iota)
```
yurishkuro: this is not the iota pattern, see https://dlintw.github.io/gobyexample/public/constants-and-iota.html

yurishkuro: e.g.
```
const (
	Spring = Season(iota)
	Summer
	Autumn
	Winner
)
```

plugin/storage/cassandra/spanstore/writer.go
```diff
@@ -65,6 +65,13 @@ const (
 	durationBucketSize = time.Hour
 )
 
+const (
```
yurishkuro: above section is already "const", don't need to start a new one

plugin/storage/cassandra/spanstore/writer.go
```diff
@@ -125,6 +134,27 @@ func (s *SpanWriter) Close() error {
...
+		if err := s.writeIndexes(span, ds); err != nil {
+			return err
+		}
+	default:
+		if err := s.writeSpan(span, ds); err != nil {
+			return err
+		}
+		if err := s.writeIndexes(span, ds); err != nil {
+			return err
+		}
```
yurishkuro: duplication of code is a sign of design problem. You could use bitmasks instead of iota-based constants

vprithvi: That was what I was doing previously...

plugin/storage/cassandra/spanstore/writer.go
vprithvi: I wanted to reset the index of `iota`, and it seems that it only happens in a new `const` block according to https://golang.org/ref/spec#Iota

plugin/storage/cassandra/spanstore/writer.go
yurishkuro: I like the previous version better, why did you change it? We might have gotten the wires crossed. If your original version of the constants was bitmap-based, then it wasn't clear from the code - I would just declare two constants, `saveSpanFlag 1` and `indexSpanFlag 2`, and add a comment that they can be combined in a bitmap (also name the struct variable `somethingFlags`)

vprithvi: I thought that we wanted the default of indexing and storing to have an ordinal of 0.

yurishkuro: if we're not using the sequential enum then it doesn't matter.

plugin/storage/cassandra/spanstore/writer_test.go
```diff
@@ -38,15 +38,16 @@ type spanWriterTest struct {
 	writer    *SpanWriter
 }
 
-func withSpanWriter(writeCacheTTL time.Duration, fn func(w *spanWriterTest)) {
+func withSpanWriter(writeCacheTTL time.Duration, fn func(w *spanWriterTest), options ...Option,
+) {
```
yurishkuro: same line

PR #686, opened 2018-02-06
Add --es.disable-basic-auth option

bpot:
This is useful when using Amazon's Elasticsearch service with a VPC
endpoint and ip-based access control.

I also considered automatically skipping the `SetBasicAuth` option if username and password were empty but it seems nicer to make it explicit.

### DIFF
--- a/pkg/es/config/config.go
+++ b/pkg/es/config/config.go
@@ -31,6 +31,7 @@ type Configuration struct {
 	Servers           []string
 	Username          string
 	Password          string
+	DisableBasicAuth  bool
 	Sniffer           bool          // https://github.com/olivere/elastic/wiki/Sniffing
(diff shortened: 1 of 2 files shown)


yurishkuro:
I don't think this is needed. If the username or password are blank, the basic auth will already be disabled. E.g. this is what the olivere's driver does
```go
func SetBasicAuth(username, password string) ClientOptionFunc {
	return func(c *Client) error {
		c.basicAuthUsername = username
		c.basicAuthPassword = password
		c.basicAuth = c.basicAuthUsername != "" || c.basicAuthPassword != ""
		return nil
	}
}
```

bpot:
It looks like you're right. I was using docker image 1.0.0 and everything is working with 1.1.0. The problem was probably the lack of ssl certs in 1.0.0 and not basic auth.

PR #693, opened 2018-02-08
model.NewProcess no longer sorts tags

### LINKED ISSUE #682: Add new model span hash function
We ran into an issue where a single span with different tag ordering will result in https://github.com/jaegertracing/jaeger/blob/d16de316a8bfc0ba7a6d5d1b7e43d3c48156d8ae/model/hash.go generating two different hashes. Ideally, even if the tags aren't ordered the same way, the two spans should be identified as equivalent.

vprithvi:
- model.NewProcess sorts tags and claims that sorted process tags
  are the canonical form. This claim is incorrect, and is violated
  in the jaeger to model convertors.

- Because the claim has no merit, and leads to inconsistencies in
  convertors that use model.NewProcess as opposed to &model.Process{},
  the sorting has been removed, and tests updated.

- fixes #682

### DIFF
--- a/model/converter/thrift/zipkin/fixtures/jaeger_02.json
+++ b/model/converter/thrift/zipkin/fixtures/jaeger_02.json
@@ -9,6 +9,10 @@
         "serviceName": "service-x",
         "tags": [
           {
+            "key": "jaeger.version",
+            "vStr": "Go-1.1"
+          },
+          {
             "key": "hostname",
             "vStr": "some-host.com"
           },
@@ -16,10 +20,6 @@
             "key": "ip",
             "vNum": 12345,
             "vType": "int64"
-          },
-          {
-            "key": "jaeger.version",
-            "vStr": "Go-1.1"
           }
         ]
       },
--- a/model/process.go
+++ b/model/process.go
@@ -23,12 +23,8 @@ type Process struct {
 }
 
 // NewProcess creates a new Process for given serviceName and tags.
-// The tags are sorted in place and kept in the the same array/slice,
-// in order to store the Process in a canonical form that is relied
-// upon by the Equal and Hash functions.
 func NewProcess(serviceName string, tags []KeyValue) *Process {
 	typedTags := KeyValues(tags)
-	typedTags.Sort()
 	return &Process{ServiceName: serviceName, Tags: typedTags}
 }
 
--- a/model/process_test.go
+++ b/model/process_test.go
@@ -28,8 +28,8 @@ import (
 
 func TestProcessEqual(t *testing.T) {
 	p1 := model.NewProcess("s1", []model.KeyValue{
-		model.String("x", "y"),
 		model.Int64("a", 1),
+		model.String("x", "y"),
 	})
 	p2 := model.NewProcess("s1", []model.KeyValue{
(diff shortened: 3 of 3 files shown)


yurishkuro:
* process.hashCode depends on the order
* converter to UI model depends on the stable hash code

vprithvi:
@yurishkuro Could you elaborate your position? 

In this case, at least two other people familiar with the code approved this request. 

Your comment is unhelpful because it seems to assume that the author and reviewers didn't consider that the hashcode depends on this order (it is highlighted in the commit message, and in the code comment in the diff).

After your initial comment, you elaborated with the following:

>process.hashCode depends on the order
converter to UI model depends on the stable hash code

These comments also aren't productive because they don't tell us of the nature of these dependencies, or the effects that breaking this dependencies would have.

Further, the comments don't address the purpose of this commit (as documented in the commit message)

yurishkuro:
> converter to UI model depends on the stable hash code

This PR breaks the UI model converter. This can be evident from the fact that the test was modified to accommodate the change. We can add more tests in the UI converter to demonstrate the dependency on the current behavior. So whatever issue this PR is trying to solve, it cannot be solved by breaking other parts.

yurishkuro:
If my earlier comments were not clear, feel free to ask clarifying questions. I was brief specifically because the PR was hastily approved when it was clearly introducing a breaking change, so I needed to block it.

yurishkuro:
test in ui converter: #708

vprithvi:
>This PR breaks the UI model converter. This can be evident from the fact that the test was modified to accommodate the change. 

I'm not entirely clear on where this PR modifies the UI model converter tests (If it did, it was unintentional). Could you link to it?

>it cannot be solved by breaking other parts.

I agree, could you clarify on which parts break, and what the effects of this breakage are?

I see the new test and PR message added in #708, but it does not elaborate on the why there is a dependency on having a sorted order or what happens when this tags are not sorted.

vprithvi:
Could you also clarify this comment:

>process.hashCode depends on the order

I agree that `process.hashCode` depends on the ordering of tags.
I don't understand the context of this comment because no changes are made to `process.hashCode` in this commit.

yurishkuro:
UI converter uses Process.hashcode to dedupe processes in the trace. It expects that hashcode will be the same regardless of the tags order. The only way this is guaranteed now is if NewProcess sorts the tags.

vprithvi:
>The only way this is guaranteed now is if NewProcess sorts the tags.

But `NewProcess` is **NOT** used in the UI model, or the Jaeger model; which means that this _guarantee doesn't hold._

The only place that it is used is when we convert the old thift zipkin model.
 
Instead of having a surprising API of `NewProcess` sorting tags; we could instead do one of the following:

1. Make the hash function independent of tag order as requested in #682 
1. Sort zipkin tags in the zipkin converter before storing them
1. Sort process tags when retrieving them. Perhaps similar to what we do for log fields: https://github.com/jaegertracing/jaeger/blob/d16de316a8bfc0ba7a6d5d1b7e43d3c48156d8ae/model/adjuster/sort_log_fields.go#L30

yurishkuro:
will be addressed in #713

PR #696, opened 2018-02-13
Support scrolling for elasticsearch

### LINKED ISSUE #668: ES storage has hard limit of 10k spans per trace during reading
ES storage queries are limited to 10k spans, so big traces over 10k spans (usually batches) displayed partially and have limited usability. 

See related span store reader code: 
https://github.com/jaegertracing/jaeger/blob/master/plugin/storage/es/spanstore/reader.go#L52

Maybe using ["Scroll" or "Search After"](https://www.elastic.co/guide/en/elasticsearch/reference/6.1/search-request-from-size.html) would help with this.

sramakr:
Resolves #668

This commit allows to support search_after when fetching traces from elasticsearch.
Steps are,
1. Do MultiRead, by sorting startTime and start with startTime of 0
2. If the TotalHits > 10K , 
3. Find the last span in the results 
4. Use the StartTime from the last span
5. Repeat steps from 1 

Signed-off-by: Sriram Ramakrishnan <sramakr@gmail.com>


yurishkuro:
what problem is this solving?

sramakr:
@yurishkuro Solving  https://github.com/jaegertracing/jaeger/issues/668

mabn:
Few remarks:

- I've had performance problems with scrolling in ES 5.x in the past and [search-after](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-search-after.html) was performing much better. Use of scroll API affects indexing - lucene segments are kept longer during scrolling.

- It adds additional count query which could be avoided by using "search-after"

- It reads all the spans for given trace and it's a problem when there are 10mln spans in a trace, it's better to get first 10k

- Right now search result fetches whole traces (not great), but won't it fetch them without limits? So, after this change, it might happen that search results try to fetch json with millions of spans, even though I'm interested in a small trace that was also returned in the response? It would be an issue.

IMO it would be great if it was possible to do some kind of paging on the trace view page. E.g. see 10k spans ordered by time and be able to go to the next/previous 10k, or something like that. Fetching everything will create new problems.

sramakr:
@mabn Thanks for the comments. I should've raised this concern before , you are totally correct. Instead of 10K limit , will it help to increase the limit to something manageable and possibly use search_after?  I am not aware about the UI, and how to do paging from UI. Gotta check more to see how everything will fit together.

yurishkuro:
Building a sensible DAG without knowing all the spans seems pretty hard. Trace with 1mm spans is an interesting theoretical use case but I doubt any existing tracing system would be able to deal with those today. On the other hand, traces with >10k spans are fairly common, so it would be great to solve this for ES (not a problem in Cassandra, btw). 

As for super large traces, I still think the solution is to try to load a complete trace in memory, but employ some form of compression. A lot of span data is repeated, like operation names, tags, etc. Also, in order to construct a DAG most of this data is not needed and the spans can be compressed to just the id/parent + timing + name. Anyway, this is a problem for another day (after I retire and won't care).

If we ignore the possibility of not loading complete trace, is the search_after a reasonable solution? From the help page it seems like it will retrieve just as much data as needed if used to retrieve spans for a given trace ID.

mabn:
I'm only really concerned about the search page - if such 1mm trace is matched by search criteria (and if it's still happening then most likely some of its spans will match) then I'd like to show something rather than explode on memory or wait 30s for the json to download. And it happens in practice :( due to bugs and poor design.

sramakr:
@mabn if spans > 10K works in cassandra, I mean , won't it be a problem in cassandra too? I think really it becomes practical limitation on number of spans we can fetch on the UI.

sramakr:
Yes, I can modify this PR to use search_after and we can have a separate issue to support pagination in the UI.

sramakr:
@mabn @yurishkuro need inputs on this PR when you get a chance. Thanks!

ledor473:
I've ran into the 10k limit today and gave this PR a try!
With a trace that long it's hard to validate if everything is present, but still here's my observations:

Good:
- The trace name went from `<cannot-find-trace-name>` to the proper name
- `Depth` increased

Bad:
- `Totals Spans` remained at 10 000
- `Durations` did not really increased
- The "span graph" seems to be broken

And here's a before/after screenshot of the top of the UI
after
before

ledor473:
It looks much better! It's now showing more than 40k spans:

screen shot 2018-03-07 at 10 49 44 pm

Also when testing it yesterday, the trace was only ~5MB. Now, it's ~20MB.

Issue:
- Span graph is not showing anymore

yurishkuro:
cc @jpkrohling

plugin/storage/es/spanstore/reader.go
```diff
@@ -205,39 +205,80 @@ func (s *SpanReader) FindTraces(traceQuery *spanstore.TraceQueryParameters) ([]*
...
+			spans, err := s.collectSpans(result.Hits.Hits)
+			if err != nil {
+				return nil, err
+			}
+			lastSpan := spans[len(spans)-1]
+			lastSpanTraceID := lastSpan.TraceID.String()
+
+			if traceSpan, ok := tracesMap[lastSpanTraceID]; ok {
+				for _, span := range spans {
+					traceSpan.Spans = append(traceSpan.Spans, span)
```
jpkrohling: Is the order of this map important? If not, we might want to consider doing some parallel processing here, no?

jpkrohling:
@kevinearls, how hard would it be to run the perf tests against this PR?

plugin/storage/es/spanstore/reader.go
```diff
@@ -205,39 +205,80 @@ func (s *SpanReader) FindTraces(traceQuery *spanstore.TraceQueryParameters) ([]*
...
+
+		for i, traceID := range traceIDs {
+			query := elastic.NewTermQuery("traceID", traceID)
+			if val, ok := searchAfterTime[traceID]; ok {
+				nextTime = val
+			}
+			if nextTime == 0 {
+				searchRequests[i] = elastic.NewSearchRequest().IgnoreUnavailable(true).Type("span").Source(elastic.NewSearchSource().Query(query).Size(defaultDocCount).Sort("startTime", true))
+			} else {
+				searchRequests[i] = elastic.NewSearchRequest().IgnoreUnavailable(true).Type("span").Source(elastic.NewSearchSource().Query(query).Size(defaultDocCount).Sort("startTime", true).SearchAfter(nextTime))
```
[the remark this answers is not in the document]
sramakr: yes , I can simplify above code  and do more testing.
'''
searchRequests[i] = elastic.NewSearchRequest().IgnoreUnavailable(true).Type("span").Source(elastic.NewSearchSource().Query(query).Size(defaultDocCount).Sort("startTime", true).SearchAfter(nextTime))
'''

plugin/storage/es/spanstore/reader.go
sramakr: @jpkrohling l don't think so. I was thinking about concurrent processing ,like run the elastic-search query separate and send the results in a channel to construct the map ?

jpkrohling: That's what I had in mind, yes.

sramakr: @jpkrohling did more thinking on this. elastic search query is dependent on the results of the query to get the start time that'll be used for next search after query. I don't think concurrency might work here. That being said, but, I will give more thoughts on this weekend.

pavolloffay:
> @kevinearls, how hard would it be to run the perf tests against this PR?

I think we don't use jaeger-query in perf tests. Tests read the ES directly.

kevinearls:
@pavolloffay we don't use jaeger-query now but I've been working on this.  I hope to have some information tomorrow.

kevinearls:
@pavolloffay  @jpkrohling  I wrote a test where I created 100,000 traces, and then retrieved them multiple times in chunks of 20,000.  When run (just on my laptop with nothing else running) against the Jaeger master, the average time to retrieve 20,000 traces was 5727 milliseconds.

When run using this PR the average dropped to 3909 milliseconds.

This looks to be a decent performance improvement.  I can run something additional if you think its necessary.

jpkrohling:
Would it be possible to run the test a couple of times more, just to make sure the second run didn't perform better because of better temporal conditions? Ideally, this would run on a perf server, to avoid desktop-specific issues, but if you have consistent results after a few runs, I'm happy with it :)

kevinearls:
@jpkrohling I did run multiple times and got fairly consistent results.  Let me see if I can setup something to run on the CNCF CI Jenkins in the next couple of days.

ledor473:
Any update on this issue?

sramakr:
@ledor473 Are you referring to the issue of "Span graph is not showing anymore" ? I didn't get a chance to look into it. Is that an UI issue ?

ledor473:
I meant the overall state of that PR. Seems like @kevinearls conducted the load testing needed.

sramakr:
ok cool @ledor473  I am still waiting for 👍  from the reviewers.

plugin/storage/es/spanstore/reader.go
```diff
@@ -205,39 +205,76 @@ func (s *SpanReader) FindTraces(traceQuery *spanstore.TraceQueryParameters) ([]*
...
 	indices := findIndices(spanIndexPrefix, startTime.Add(-time.Hour), endTime.Add(time.Hour))
 
-	results, err := s.client.MultiSearch().
-		Add(searchRequests...).
-		Index(indices...).
-		Do(s.ctx)
-	if err != nil {
-		return nil, err
-	}
+	var nextTime uint64
```
black-adder: can this be initialized to startTime?

plugin/storage/es/spanstore/reader.go
```diff
@@ -205,39 +205,76 @@ func (s *SpanReader) FindTraces(traceQuery *spanstore.TraceQueryParameters) ([]*
...
+		for i, traceID := range traceIDs {
+			query := elastic.NewTermQuery("traceID", traceID)
+			if val, ok := searchAfterTime[traceID]; ok {
+				nextTime = val
+			}
+			searchRequests[i] = elastic.NewSearchRequest().IgnoreUnavailable(true).Type("span").Source(elastic.NewSearchSource().Query(query).Size(defaultDocCount).Sort("startTime", true).SearchAfter(nextTime))
+		}
+		// set traceIDs to empty
+		traceIDs = nil
+		results, err := s.client.MultiSearch().Add(searchRequests...).Index(indices...).Do(s.ctx)
```
black-adder: Is the 10000 limit for the entire query? ie if we search for 10 traces and the first trace takes up all 10000 spans, will this return a response for only the first trace?

sramakr: no, it's per trace. obviously, there is the downside of if there are 2 traces and 1st trace gets only 100 spans and next like 50K spans, entire query is going to wait to fetch 50,100 spans before it returns.

plugin/storage/es/spanstore/reader.go
sramakr: I am not sure , because nextTime is derived from the results of the query in line 232.

sramakr: Since nextTime is derived from the results of the previous 10K , don't think we can initialize.

black-adder: doesn't this mean that the first ever query always queries from 0 (searchAfterTime isn't initialized) even though we're only interested things after startTime? (I guess since we're looking at a subset of indices, this isn't too bad performance wise)

sramakr: @black-adder Thought about this a bit... I agree , we should initialize next-time to start-time, don't see a problem there. I am going to modify this and update the PR.

black-adder:
Thanks!

PR #700, opened 2018-02-16
Add a flag to control Cassandra consistency level

### LINKED ISSUE #699: Support on collector and query applications for setting Cassandra consistency levels
We are in the process of setting up failure resilient multi-tenant capable environment on OpenShift with Cassandra as storage backend for Jaeger Tracing components. We are deploying Cassandra in a 2 datacenter setup with 2 nodes of Cassandra in each datacenter. We are fine with traffic in between the Cassandra nodes to travel between the two datacenters and are looking at configuring the consistency level of both the Jaeger-Collector and Jaeger-Query applications.

By default the Go Cassandra client libraries selects LOCAL_ONE as the consisteny level, but we would like to select freely between the other consistency levels:

[](https://docs.datastax.com/en/archived/cassandra/2.0/cassandra/dml/dml_config_consistency_c.html)

ie. the LOCAL_QUORUM, ANY etc.

As we can see the Cassandra Go package have this as configuration options when using it in connections to Cassandra - see:

[jaeger/pkg/cassandra/config/config.go]

but the configurator for Jaeger as a client to Cassandra doesn't appear to have that supported - see:

[jaeger/plugin/storage/cassandra/options.go]

Could this Cassandra consistency option be added to the Jaeger applications to support a more production true use of Cassandra. 

Best regards
Lars Milland

yurishkuro:
Resolves #699

```
$ go run ./cmd/collector/main.go -h | grep consistency
      --cassandra-archive.consistency string                 The Cassandra consistency level, e.g. ANY, ONE, TWO, THREE, QUORUM, ALL, LOCAL_QUORUM, EACH_QUORUM, LOCAL_ONE (default LOCAL_ONE)
      --cassandra.consistency string                         The Cassandra consistency level, e.g. ANY, ONE, TWO, THREE, QUORUM, ALL, LOCAL_QUORUM, EACH_QUORUM, LOCAL_ONE (default LOCAL_ONE)
```

cc @LarsMilland

### DIFF
--- a/plugin/storage/cassandra/options.go
+++ b/plugin/storage/cassandra/options.go
@@ -33,6 +33,7 @@ const (
 	suffixServers          = ".servers"
 	suffixPort             = ".port"
 	suffixKeyspace         = ".keyspace"
+	suffixConsistency      = ".consistency"
 	suffixProtoVer         = ".proto-version"
 	suffixSocketKeepAlive  = ".socket-keep-alive"
 	suffixUsername         = ".username"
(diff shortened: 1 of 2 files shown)


plugin/storage/cassandra/options.go
```diff
@@ -147,6 +148,10 @@ func addFlags(flagSet *flag.FlagSet, nsConfig *namespaceConfig) {
 		nsConfig.namespace+suffixKeyspace,
 		nsConfig.Keyspace,
 		"The Cassandra keyspace for Jaeger data")
+	flagSet.String(
+		nsConfig.namespace+suffixConsistency,
+		nsConfig.Consistency,
+		"The Cassandra consistency level, e.g. ANY, ONE, TWO, THREE, QUORUM, ALL, LOCAL_QUORUM, EACH_QUORUM, LOCAL_ONE (default LOCAL_ONE)")
```
pavolloffay: IIRC default is printed automatically from supplied value

PR #702, opened 2018-02-20
Publish hotrod image to docker hub

pavolloffay:
Resolves #204 

Signed-off-by: Pavol Loffay <ploffay@redhat.com>

PR #710, opened 2018-02-22
Pin testify version to 1.2.1

pavolloffay:
See https://github.com/jaegertracing/jaeger/pull/702#issuecomment-367381845


yurishkuro:
+1 to start using a version, but why pin instead of ^1.2.1?

pavolloffay:
No reason, updated

PR #711, opened 2018-02-23
Wait group should add before the goroutine creating

oiooj:
The calls to Add should execute before the statement creating the
goroutine or other event to be waited for.

Signed-off-by: kun <oiooj@qq.com>


black-adder:
Thanks!

PR #712, opened 2018-02-23
Use UTC for model timezone

vprithvi:
- standardize on UTC for time representation
- the wire format for everything is microseconds since epoch, so this change does not effect data transfer over rpcs. 


black-adder:
I'm ok with this but why?

PR #718, opened 2018-02-28
Make sure different tracers in HotROD process use different RNG seeds

yurishkuro:
I ran into a case where trace had spans with the same IDs. The RNG is seeded with time.Now().UnixNano(), which potentially can give identical values over several calls (since the clock precision is not actually nanos).

This is a quick (hacky) workaround. A proper fix needs to be in the Jaeger client.


yurishkuro:
Yes, but maybe there was another cause. I certainly was able to reproduce same-IDs from different tracers scenario.

PR #720, opened 2018-02-28
Add sampling handler to main

black-adder:


cmd/standalone/Dockerfile
```diff
@@ -20,5 +20,6 @@ EXPOSE 16686
 
 COPY ./jaeger-ui-build /go/src/jaeger-ui-build
 COPY ./cmd/standalone/standalone-linux /go/bin/
+COPY ./cmd/standalone/sampling_strategies.json /go/src/config/
 
-CMD ["/go/bin/standalone-linux","--query.static-files=/go/src/jaeger-ui-build/build/"]
+CMD ["/go/bin/standalone-linux","--query.static-files=/go/src/jaeger-ui-build/build/","--sampling.strategies-file=/go/src/config/sampling_strategies.json"]
```
black-adder: The standalone will be provided with a sampling strategies file which defaults the sampling probability to 1

cmd/collector/main.go
```diff
@@ -58,7 +62,10 @@ func main() {
 	if err != nil {
 		log.Fatalf("Cannot initialize storage factory: %v", err)
 	}
-
+	strategyStoreFactory, err := ss.NewFactory(ss.FactoryConfigFromEnv())
```
black-adder: This might be a breaking change. If no Env is provided, we default to using the default probability of 0.001 (all client's remote controlled samplers default to using 0.001 if it can't contact the collector...)

Do we want to have no default and not initialize a handler if Env is not provided and have users manually enable?

yurishkuro: Sorry, how is it "breaking"? Sounds like you're saying the effect of no Env will be the same - 0.001

cmd/collector/main.go
```diff
@@ -200,3 +211,20 @@ func startZipkinHTTPAPI(
 		}
 	}
 }
+
+func initializeSamplingHandler(
+	samplingStrategyStoreFactory *ss.Factory,
+	v *viper.Viper,
+	logger *zap.Logger,
+	metricsFactory metrics.Factory,
```
yurishkuro: can we keep the arg order consistent with `Initialize(metricsFactory, logger)`?

cmd/standalone/Dockerfile
yurishkuro: clever

cmd/standalone/standalone_test.go
```diff
@@ -94,6 +100,24 @@ func getAPITrace(t *testing.T) {
...
 
+func getSamplingStrategy(t *testing.T) {
+	req, err := http.NewRequest("GET", getSamplingStrategyURL, nil)
+	require.NoError(t, err)
+
+	var queryResponse sampling.SamplingStrategyResponse
+	resp, err := httpClient.Do(req)
+	require.NoError(t, err)
+
+	body, _ := ioutil.ReadAll(resp.Body)
```
yurishkuro: require.NoError(t, err)

scripts/travis/build-all-in-one-image.sh
```diff
@@ -11,7 +11,7 @@ make build-all-in-one-linux
 export REPO=jaegertracing/all-in-one
 
 docker build -f cmd/standalone/Dockerfile -t $REPO:latest .
-export CID=$(docker run -d -p 16686:16686 $REPO:latest)
+export CID=$(docker run -d -p 16686:16686 -p 5778:5778 $REPO:latest)
```
yurishkuro: seriously... how did this ever work then?

black-adder: The test uses the query service's instrumentation to create debug traces so it never needed agent's udp ports.

PR #728, opened 2018-03-06
Add a span merger that merges spans read

vprithvi:
- allows for partially written spans to be merged while querying by selecting
  the span with the longest duration.

See https://github.com/jaegertracing/jaeger-client-java/issues/231 for more discussion

### DIFF
--- a/model/adjuster/span_merger.go
+++ b/model/adjuster/span_merger.go
@@ -15,16 +15,11 @@
 package adjuster
 
 import (
-	"sort"
-
 	"github.com/jaegertracing/jaeger/model"
 )
 
 // MergeSpans returns an Adjuster that merges Jaeger spans with the same spanID.
-// It skips merging Zipkin spans. Zipkin spans are defined as spans sharing the spanID and containing
-// both client and server span.kind annotations.
-// As Zipkin spans are always reported in entirety, we can assume that that the span.kind annotations are present.
-// A drawback of this approach is that incomplete traces with duplicate Zipkin spans will be merged unnecessarily
+// Duplicate spans that have conflicting span.kind annotations are not merged.
 //
 // MergeSpans assumes that the duration field in a span is monotonically increasing for a spans with the
 // same spanID. The span with the longest spanID wins, and is selected in entirety.
@@ -41,10 +36,10 @@ func MergeSpans() Adjuster {
 		trace := &model.Trace{}
 		trace.Warnings = input.Warnings
 		for _, spans := range IDToSpans {
-			if isZipkin(spans) {
-				trace.Spans = append(trace.Spans, spans...)
-			} else {
+			if isMergeable(spans) {
 				trace.Spans = append(trace.Spans, mergeSpans(spans))
+			} else {
+				trace.Spans = append(trace.Spans, spans...)
 			}
 		}
 		return trace, nil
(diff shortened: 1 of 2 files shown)


model/adjuster/span_merger.go
```diff
@@ -0,0 +1,88 @@
...
+			IDToSpans[span.SpanID] = []*model.Span{span}
+		}
+	}
+	return IDToSpans
+}
+
+func mergeSpans(spans []*model.Span) *model.Span {
+	// This assumes that the duration field in a span is monotonically increasing
+	// and uses it to break ties between spans.
+	sort.Slice(spans, func(i, j int) bool {
```
black-adder: nit: sorting is O(nlogn), finding the max duration can take O(n) time.

model/adjuster/span_merger.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/jaegertracing/jaeger/model"
+)
+
+// MergeSpans returns an Adjuster that merges Jaeger spans with the same spanID.
+// It skips merging Zipkin spans. Zipkin spans are defined as spans sharing the spanID and containing
+// both client and server span.kind annotations.
+// As Zipkin spans are always reported in entirety, we can assume that that the span.kind annotations are present.
+// A drawback of this approach is that incomplete traces with duplicate Zipkin spans will be merged unnecessarily
+//
+// MergeSpans assumes that the duration field in a span is monotonically increasing for a spans with the
```
black-adder: s/a spans/spans

model/adjuster/span_merger.go
black-adder: ooh, even better, when you do groupByIDs, compare the durations and save only longest duration one there. 2 birds, one stone.

vprithvi: Agree about removing the sort. 
I don't want to do it in `groupByIDs` because it makes the code less readable (because we'd also have to do the `isZipkin` check). It also makes enhancing the `merge` function more involved.

vprithvi:
>If I recall our conversations, we're going to use a "random value" as the hash so that we can save multiple spans with the same id. We need to figure out a way such that this is compatible with: jaegertracing/jaeger-client-java#231.

Similar, the idea is to kill span hash, and use a random number field to store all spans that arrive at the collector. 

>From my last conversation with @phal0r, we discussed having the clients emit a duration of 0 if the spans aren't complete so that we could circumvent having to add a new first class flag to the cassandra model so that query knows that the span is incomplete and can pass on the info to the UI. This PR is kind of at odds with that approach since if we emit multiple incomplete spans with 0 duration, we'll be randomly picking one rather than using the latest.

Hmmm, this is true, we can enhance the merge function to select the span with the most data to make it compatible with that approach. Note that we don't yet add this as part of the default standard adjusters, so we can iterate to make this compatible with an approach like that. 

I created #729 for further discussion about this topic!

model/adjuster/span_merger.go
```diff
@@ -0,0 +1,88 @@
...
+	// This assumes that the duration field in a span is monotonically increasing
+	// and uses it to break ties between spans.
+	sort.Slice(spans, func(i, j int) bool {
+		return spans[i].Duration > spans[j].Duration
+	})
+
+	return spans[0]
+}
+
+func isZipkin(spans []*model.Span) bool {
```
yurishkuro: I don't think this file should be mentioning zipkin. The adjusters run in a fixed order, we can either run the (Zipkin) deduper first, leaving this file with no additional checks to make, or running this adjuster first, letting it merge all spans that have the same span.kind, and then running Zipkin deduper.

yurishkuro:
@vprithvi what is the plan with this PR? Is it still needed?

vprithvi:
This is subsumed by #979

PR #730, opened 2018-03-07
Increase domain for model.TimeAsEpochMicroseconds

vprithvi:
Use `time.Unix()` instead of `time.UnixNano()` when the input exceeds the domain of `time.UnixNano()`

### DIFF
--- a/model/time.go
+++ b/model/time.go
@@ -26,7 +26,12 @@ func EpochMicrosecondsAsTime(ts uint64) time.Time {
 // TimeAsEpochMicroseconds converts time.Time to microseconds since epoch,
 // which is the format the StartTime field is stored in the Span.
 func TimeAsEpochMicroseconds(t time.Time) uint64 {
-	return uint64(t.UnixNano() / 1000)
+	if t.Year() > 1678 && t.Year() < 2262 {
+		// UnixNano's result is undefined if the Unix time in nanoseconds cannot be
+		// represented by an int64(less than year 1678 or greater than year 2262)
+		return uint64(t.UnixNano() / 1000)
+	}
+	return uint64(t.Unix() * 1e6)
 }
 
 // MicrosecondsAsDuration converts duration in microseconds to time.Duration value.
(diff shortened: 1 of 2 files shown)


vprithvi:
On another note, what do you think about also adding warnings when converting representation formats (model to thrift/ thrift to model, etc) when current time and span time differ by more than a week?

model/time.go
```diff
@@ -26,7 +26,12 @@ func EpochMicrosecondsAsTime(ts uint64) time.Time {
 // TimeAsEpochMicroseconds converts time.Time to microseconds since epoch,
 // which is the format the StartTime field is stored in the Span.
 func TimeAsEpochMicroseconds(t time.Time) uint64 {
-	return uint64(t.UnixNano() / 1000)
+	if t.Year() > 1678 && t.Year() < 2262 {
+		// UnixNano's result is undefined if the Unix time in nanoseconds cannot be
+		// represented by an int64(less than year 1678 or greater than year 2262)
+		return uint64(t.UnixNano() / 1000)
+	}
+	return uint64(t.Unix() * 1e6)
```
yurishkuro: not sure I follow - doesn't this just truncate to full seconds? Isn't there a way to retrieve the sub-second digits, maybe without using UnixNanos()?

model/time.go
```diff
@@ -26,7 +26,12 @@ func EpochMicrosecondsAsTime(ts uint64) time.Time {
 // TimeAsEpochMicroseconds converts time.Time to microseconds since epoch,
 // which is the format the StartTime field is stored in the Span.
 func TimeAsEpochMicroseconds(t time.Time) uint64 {
-	return uint64(t.UnixNano() / 1000)
+	if t.Year() > 1678 && t.Year() < 2262 {
```
yurishkuro: `if y := t.Year(); y > 1678 && y < 2262 {}`

pavolloffay:
@vprithvi could you please push the branch to your fork and remove this branch?

PR #741, opened 2018-03-15
Dockerize es-indices-clean script

### LINKED ISSUE #740: Dockerfile for es_indices_clean.sh

pavolloffay:
Resolves #740 


plugin/storage/es/es_indices_clean.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
```
pavolloffay: bash is not in alpine

plugin/storage/es/esCleaner.py
```diff
@@ -1,6 +1,5 @@
 import elasticsearch
 import curator
-import logging
```
pavolloffay: never used

black-adder:
For what purpose? Is usability improved by running the script inside the container?

pavolloffay:
How do you want to run it as cronjob without container? https://github.com/jaegertracing/jaeger-kubernetes/issues/40

pavolloffay:
> the name of the Dockerhub repository might look better as (jaeger-)es-index-cleaner

It uses the same naming convention as other images. 

> the es_indices_clean.sh looks like an anti-pattern.  Do you want to remove scrip altogether? I agree about pip.

isaachier:
The script might be useful for users not on Kubernetes, but they should probably use a real cronjob or Jenkins, so IMO makes sense to delete it.

yurishkuro:
> > the name of the Dockerhub repository might look better as (jaeger-)es-index-cleaner

> It uses the same naming convention as other images.

There isn't really a convention. I I just think "es index cleaner" describes better what the image represents than "es indices clean".

> The script might be useful for users not on Kubernetes, but they should probably use a real cronjob or Jenkins, so IMO makes sense to delete it.

The python script will still be available, I am just saying there's no reason for a shell script as well, especially because it does `pip install`, which is likely to fail depending on user permissions.

pavolloffay:
> There isn't really a convention. I I just think "es index cleaner" describes better what the image represents than "es indices clean". 

that yes, I didn't read it properly. I thought you want to remove jaeger prefix :/. I will rename it

pavolloffay:
@yurishkuro image has been renamed and shell script removed.

PR #748, opened 2018-03-20
Allow overriding base path for UI/API routes; rm --query.prefix

### LINKED ISSUE #745: Provide a means to define a path-prefix for the UI
With jaegertracing/jaeger-ui#198, a path prefix for the UI can be configured dynamically, e.g. the Jaeger UI can be rooted at `/some-prefix/` instead of the root path. 

This is to address jaegertracing/jaeger-ui#42.

In order to configure path-prefix, the query-service needs to replace the value of a `<base>` element in the index.html provided by the UI.

The text to match is:

```
<base href="/" data-inject-target="BASE_URL"/>
```

The portion to replace is the `"/"` value to the `href` attribute.

For instance, if the path-prefix is `"/tracing/east/"`, the resultant text should be:

```
<base href="/tracing/east/" data-inject-target="BASE_URL"/>
```

Note: The `data-inject-target="BASE_URL"` is only there to serve as some indication that the element is manipulated programmatically, i.e. don't edit manually.

The user needs some way to define the path-prefix. Seems like a CLI param or envvar would be good candidates. The query-service would then use the user-provided value to do the replacement.

A quick note on the format of the value: The first and last characters should be `/`. If the first is omitted then the base URL will be relative to whatever the current URL is, e.g. `http://localhost/trace/abc/tracing/east/` instead of `http://localhost/tracing/east/`. If the last char is not a `/` then relative URLs (mostly the static assets referenced in index.html) will not have a separating `/` from the base URL.

### LINKED ISSUE #746: Remove --query.prefix option
First, it is poorly named (should've been something like `api-prefix`), but most importantly if it is set to anything other than `api` it will break the UI, because it only affects the query service API routes, but does not affect how the UI JavaScript refers to them.

cc @tiffon @pavolloffay @jpkrohling

yurishkuro:
Resolves #745 **Provide a means to define a path-prefix for the UI**
Resolves #746 **Remove --query.prefix option**

* [ ] depends on https://github.com/jaegertracing/jaeger-ui/pull/198 merged

### DIFF
--- a/cmd/query/app/fixture/index.html
+++ b/cmd/query/app/fixture/index.html
@@ -1,6 +1,7 @@
 <!DOCTYPE html>
 <html lang="en">
     <meta charset="UTF-8">
+    <base href="/" data-inject-target="BASE_URL"/>
     <title>Test Page</title>
     <!-- JAEGER_CONFIG=DEFAULT_CONFIG; -->
 </html>
--- a/cmd/query/app/static_handler.go
+++ b/cmd/query/app/static_handler.go
@@ -31,11 +31,17 @@ import (
 var (
 	staticRootFiles = []string{"favicon.ico"}
 	configPattern   = regexp.MustCompile("JAEGER_CONFIG *= *DEFAULT_CONFIG;")
+	basePathPattern = regexp.MustCompile(`<base href="/"`)
+	basePathReplace = `<base href="%s/"`
+	errBadBasePath  = errors.New("Base path must start with / but not end with /; valid path: /jaeger/ui")
 )
 
 // RegisterStaticHandler adds handler for static assets to the router.
 func RegisterStaticHandler(r *mux.Router, logger *zap.Logger, qOpts *QueryOptions) {
-	staticHandler, err := NewStaticAssetsHandler(qOpts.StaticAssets, qOpts.UIConfig)
+	staticHandler, err := NewStaticAssetsHandler(qOpts.StaticAssets, StaticAssetsHandlerOptions{
+		BasePath:     qOpts.BasePath,
+		UIConfigPath: qOpts.UIConfig,
+	})
 	if err != nil {
 		logger.Panic("Could not create static assets handler", zap.Error(err))
 	}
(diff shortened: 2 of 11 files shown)


cmd/query/app/fixture/index.html
```diff
@@ -1,6 +1,7 @@
 <!DOCTYPE html>
 <html lang="en">
     <meta charset="UTF-8">
+    <base href="/" data-inject-target="BASE_URL"/>
```
yurishkuro: @tiffon this models the compiled `index.html` from UI assets.

Q: is it possible that babel would reorder the `href` and `data-inject-target` attributes? Because if it does, the search&replace in the query service won't work

cmd/query/app/static_handler.go
```diff
@@ -74,9 +88,20 @@ func NewStaticAssetsHandler(staticAssetsRoot string, uiConfig string) (*StaticAs
 		bytes, _ := json.Marshal(config)
 		configString = fmt.Sprintf("JAEGER_CONFIG = %v", string(bytes))
 	}
+	indexBytes = configPattern.ReplaceAll(indexBytes, []byte(configString+";"))
+	if options.BasePath == "" {
+		options.BasePath = "/"
+	}
+	if options.BasePath != "/" {
+		if !strings.HasPrefix(options.BasePath, "/") || strings.HasSuffix(options.BasePath, "/") {
```
jpkrohling: Is it possible to add a log statement notifying that what people see in the HTML is *not* what will be served to clients?

yurishkuro: could you elaborate why you think this would be useful? We're modifying the internal file of the app, not something that users provide.

yurishkuro:
Docs would be updated in a separate PR in the docs repo.

cmd/query/app/static_handler.go
jpkrohling: Mainly because it's deviating from what is "expected": if I had a problem with the static resources (like the original problem this PR is solving) and were to check the HTML, I would wonder why it's not the same as what my browser is seeing. A log entry at `debug` level would suffice, IMO.

cmd/query/app/fixture/index.html
tiffon: @yurishkuro It should not change the order of the attributes.

cmd/query/app/static_handler.go
yurishkuro: well, still not convinced why someone needs to know what happens internally, e.g. we also override the config, and in both cases we only do that when requested via cli arguments.  

Let's do it separately, because the constructor doesn't take a logger and the unit tests  explicitly verify that nothing is logged, so it's not as trivial a change as it sounds for a feature that seems rather marginal.

PR #758, opened 2018-03-27
Increase precision for Float64Type of KeyValue when cast to string

black-adder:


yurishkuro:
Please, some analysis of the impact, where it is used today, and what problem this is solving for you. Strictly speaking if you have code that depends on string representation of a float you may already be doing something wrong.

black-adder:
too much work

PR #760, opened 2018-03-29
Sorted key/value store (badger) backed storage plugin

burmanm:
This is a storage plugin for Jaeger implemented against a sorted K/V store. Although this version was done against Badger it should work with relatively small changes against any lexicographically sorted K/V store (RocksDB is such a possibility also - but it would require cgo for compiling).

This is WIP, pushed for early feedback. It is missing implementation for Duration index (range index scanning) as well as GetOperations & GetServices interfaces and benchmarketing/more tests of course. Some smaller TODO parts obviously remain as well, some for easier development purposes and some just lacking optimization (not to mention splitting some parts to separate functions).

cc @pavolloffay


black-adder:
Thanks for the contribution! Some preliminary thoughts: Have you seen this ticket? https://github.com/jaegertracing/jaeger/issues/422 we've been considering supporting third party storage plugins via a new plugin framework so that the core jaeger library isn't overly inundated with storage solution specific implementations (new plugins will be supported by the contributor in a separate repo). We haven't come up with the exact framework yet so we can review this PR but in the future, we'd expect to move it over to the plugin framework.

I'll take a look at the actual PR later today.

burmanm:
Yeah, my intention was to solve the #551 issue. Theoretically moving the code to an external plugin shouldn't be an issue - just add some API listeners in front (if that's the solution #422 ends up with). Separate repo could technically allow defining build flags to support multiple different K/V stores I assume.

burmanm:
There's a bug in the index seek process, it returns results in the DESC order, but does not actually fetch TOP N by using a DESC ordering, but instead uses ASC. Will fix in the next commit with some other updates (as well as add test that catches the mistake)

pavolloffay:
@burmanm thanks for the PR. I missed it.

I am wondering whether we could completely replace in-memory with local storage. I will have a closer look next week.

also travis seems to be broken for the last build

pavolloffay:
There was a discussion about this PR on bi-weekly call. 

Rather than introducing a new storage impl can this substitute current in-memory? What are the implications? Can it work without creating any files? If so we could use this instead of in-memory.

burmanm:
By creating files in the tmpfs it is in-memory storage as tmpfs is not persisted to the disk. We could add an option to clean up the tmpfs files on the shutdown if we wanted (pod reboot / machine reboot / etc will do that of course also)

glide.yaml
```diff
@@ -55,3 +55,5 @@ import:
 - package: github.com/go-openapi/validate
 - package: github.com/go-openapi/loads
 - package: github.com/elazarl/go-bindata-assetfs
+- package: github.com/dgraph-io/badger
```
pavolloffay: It needs `glide update` to update the lock file with this change.

burmanm: Sadly I'm not sure what's the proper way to update this project's dependencies. glide up will break Jaeger (for example viper is updated and will no longer compile) as that has no version locked in the glide.yaml

It also updates several other dependencies.

pavolloffay:
There are a lot of magic numbers in different places. Could you please define them as constants?

pavolloffay:
Then when you are creating prefixes could you please extract those into functions?

plugin/storage/badger/options.go
```diff
@@ -0,0 +1,124 @@
...
+	"flag"
+	"time"
+
+	"github.com/spf13/viper"
+)
+
+// Options store storage plugin related configs
+type Options struct {
+	primary *NamespaceConfig
+	others  map[string]*NamespaceConfig
```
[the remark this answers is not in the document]
burmanm: To be constant with the other storage engines. No other reason really.

plugin/storage/badger/options.go
```diff
@@ -0,0 +1,124 @@
...
+
+// NewOptions creates a new Options struct.
+func NewOptions(primaryNamespace string, otherNamespaces ...string) *Options {
+	options := &Options{
+		primary: &NamespaceConfig{
+			namespace:      primaryNamespace,
+			SpanStoreTTL:   time.Hour * 72, // Default is 3 days
+			SyncWrites:     false,          // Performance over consistency
+			Ephemeral:      true,           // Default is ephemeral storage
+			ValueDirectory: "",
```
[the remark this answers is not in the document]
burmanm: I can make as default the application path + ```data/``` for example. I don't think there's a sane default path for files.

plugin/storage/badger/factory.go
```diff
@@ -0,0 +1,121 @@
...
+func (f *Factory) InitFromViper(v *viper.Viper) {
+	f.Options.InitFromViper(v)
+}
+
+// Initialize implements storage.Factory
+func (f *Factory) Initialize(metricsFactory metrics.Factory, logger *zap.Logger) error {
+	opts := badger.DefaultOptions
+
+	if f.Options.primary.Ephemeral {
+		opts.SyncWrites = false
```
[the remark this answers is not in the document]
burmanm: Data is available in any case directly. With tmpfs it makes no difference is sync is true or false (other than extra overhead of fsync call to tmpfs) for durability and there's never a difference when it comes to visibility. Stuff is always written to the page cache in any case, but this controls if fsync to the disk is called after every write or not.

plugin/storage/badger/options.go
```diff
@@ -0,0 +1,124 @@
...
+const (
+	suffixKeyDirectory   = ".directory-key"
+	suffixValueDirectory = ".directory-value"
+	suffixEphemeral      = ".ephemeral"
+	suffixSpanstoreTTL   = ".span-store-ttl"
+	suffixSyncWrite      = ".consistency"
+)
+
+// NewOptions creates a new Options struct.
+func NewOptions(primaryNamespace string, otherNamespaces ...string) *Options {
```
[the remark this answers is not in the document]
burmanm: See previous comment to keep consistency with other storage engines.

plugin/storage/badger/spanstore/cache.go
```diff
@@ -0,0 +1,187 @@
...
+	err := c.store.View(func(txn *badger.Txn) error {
+		opts := badger.DefaultIteratorOptions
+		it := txn.NewIterator(opts)
+		defer it.Close()
+
+		serviceKey := []byte{serviceNameIndexKey}
+
+		// Seek all the services first
+		for it.Seek(serviceKey); it.ValidForPrefix(serviceKey); it.Next() {
+			timestampStartIndex := len(it.Item().Key()) - 24
```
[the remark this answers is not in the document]
burmanm: It's in the WriteSpan. 8 bytes for the timestamp + 16 bytes for the traceId -> 24

plugin/storage/badger/spanstore/writer.go
```diff
@@ -0,0 +1,178 @@
...
+		cache:  c,
+		closer: closer,
+	}
+}
+
+// WriteSpan writes the encoded span as well as creates indexes with defined TTL
+func (w *SpanWriter) WriteSpan(span *model.Span) error {
+
+	// Avoid doing as much as possible inside the transaction boundary, create entries here
+	entriesToStore := make([]*badger.Entry, 0, len(span.Tags)+4+len(span.Process.Tags)+len(span.Logs)*4)
```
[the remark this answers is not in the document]
burmanm: A random guess to avoid memory allocations while inside transaction boundary.

plugin/storage/badger/spanstore/writer.go
```diff
@@ -0,0 +1,178 @@
...
+
+func createTraceKV(span *model.Span) ([]byte, []byte, error) {
+	// TODO Add Hash for Zipkin compatibility?
+
+	// Note, KEY must include startTime for proper sorting order for span-ids
+	// KEY: ti<trace-id><startTime><span-id> VALUE: All the details (json for now) METADATA: Encoding
+	buf := new(bytes.Buffer)
+
+	buf.WriteByte(spanKeyPrefix)
+	binary.Write(buf, binary.BigEndian, span.TraceID.High)
```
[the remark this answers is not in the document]
burmanm: Why would replacing TraceID with Span's hash work? That does not help finding the Trace. The order and grouping is important in this case as we want to search for TraceID and then list its spans - thus they need to be sorted under the same TraceID for better performance. 

Also, the fact that documentation mentions ```- Notice we have span_hash. This exists only for zipkin backwards compat.``` indicates it's not a core strategy for Jaeger.

plugin/storage/badger/spanstore/writer.go
```diff
@@ -0,0 +1,178 @@
...
+	binary.Write(buf, binary.BigEndian, span.TraceID.Low)
+	binary.Write(buf, binary.BigEndian, model.TimeAsEpochMicroseconds(span.StartTime))
```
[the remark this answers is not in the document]
burmanm: This gives the spans a natural sorting inside a single Trace.

plugin/storage/badger/spanstore/writer.go
```diff
@@ -0,0 +1,178 @@
...
+	})
+
+	// Do cache refresh here to release the transaction earlier
+	w.cache.Update(span.Process.ServiceName, span.OperationName)
+
+	return err
+}
+
+func createIndexKey(indexPrefixKey byte, value []byte, startTime time.Time, traceID model.TraceID) []byte {
+	// KEY: indexKey<indexValue><startTime><traceId> (traceId is last 16 bytes of the key)
```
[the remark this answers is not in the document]
burmanm: Commented above (github reordered)

plugin/storage/badger/spanstore/reader.go
```diff
@@ -0,0 +1,457 @@
...
+	return &TraceReader{
+		store: db,
+		cache: c,
+	}
+}
+
+func (r *TraceReader) getTraces(traceIDs []model.TraceID) ([]*model.Trace, error) {
+	// Get by PK
+	traces := make([]*model.Trace, 0, len(traceIDs))
+	prefixes := make([][]byte, 0, len(traceIDs))
```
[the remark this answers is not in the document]
burmanm: Yet that means doing more inside the transaction (which blocks). These are done here to reduce processing (there's more to optimize also) time when transaction is open.

plugin/storage/badger/spanstore/reader.go
```diff
@@ -0,0 +1,457 @@
...
+		opts := badger.DefaultIteratorOptions
+		opts.PrefetchSize = 10 // TraceIDs are not sorted, pointless to prefetch large amount of values
+		it := txn.NewIterator(opts)
+		defer it.Close()
+
+		val := []byte{}
+		for _, prefix := range prefixes {
+			spans := make([]*model.Span, 0, 4) // reduce reallocation requirements by defining some initial length
+
+			d := 0
```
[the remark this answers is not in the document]
burmanm: I'll remove, this and the d++ (can't remember why they're there - some code I've since deleted)

plugin/storage/badger/spanstore/reader.go
```diff
@@ -0,0 +1,457 @@
...
+		prefixes = append(prefixes, createPrimaryKeySeekPrefix(traceID))
+	}
+
+	err := r.store.View(func(txn *badger.Txn) error {
+		opts := badger.DefaultIteratorOptions
+		opts.PrefetchSize = 10 // TraceIDs are not sorted, pointless to prefetch large amount of values
+		it := txn.NewIterator(opts)
+		defer it.Close()
+
+		val := []byte{}
```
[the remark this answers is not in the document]
burmanm: ValueCopy does append([:0], src...) and val never escapes the loop

plugin/storage/badger/spanstore/reader.go
```diff
@@ -0,0 +1,457 @@
...
+	// TODO We need to do a primary key scan and check all the TraceIDs if they match certain time range (full table scan)
+	// Then sort them based on the time
+	return nil, ErrNotSupported
+}
+
+func validateQuery(p *spanstore.TraceQueryParameters) error {
+	if p == nil {
+		return ErrMalformedRequestObject
+	}
+	if p.ServiceName == "" && len(p.Tags) > 0 {
```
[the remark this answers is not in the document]
burmanm: There's no restriction to the amount.

plugin/storage/badger/spanstore/reader.go
```diff
@@ -0,0 +1,457 @@
...
+	ids := make([][][]byte, 0, len(indexSeeks)+1)
+	if len(indexSeeks) > 0 {
+		for i, s := range indexSeeks {
+			indexResults, _ := r.scanIndexKeys(s, query.StartTimeMin, query.StartTimeMax)
+			if err != nil {
+				return nil, err
+			}
+			ids = append(ids, make([][]byte, 0, len(indexResults)))
+			for _, k := range indexResults {
+				ids[i] = append(ids[i], k[len(k)-16:])
```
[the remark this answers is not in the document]
burmanm: Size of TraceID is 16.

glide.yaml
jpkrohling: What I did in a previous PR was to apply this change in a separate branch and then manually update the lock file *only* for the dependency I'm updating. Sucks, but works.

plugin/storage/badger/options.go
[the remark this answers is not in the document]
jpkrohling: How about something inside `/tmp/`? When ephemeral is true, `/tmp` is appropriate.

plugin/storage/badger/factory.go
```diff
@@ -0,0 +1,121 @@
...
+		f.tmpDir = dir
+		opts.Dir = f.tmpDir
+		opts.ValueDir = f.tmpDir
+	} else {
+		opts.SyncWrites = f.Options.primary.SyncWrites
+		opts.Dir = f.Options.primary.KeyDirectory
+		opts.ValueDir = f.Options.primary.ValueDirectory
+	}
+
+	store, err := badger.Open(opts)
```
jpkrohling: Can you log the final state of `opts`? Something like:

```
logger.Info("Badger storage configuration", zap.Any("configuration", opts))
```

plugin/storage/badger/options.go
```diff
@@ -0,0 +1,124 @@
...
+	)
+	flagSet.String(
+		nsConfig.namespace+suffixKeyDirectory,
+		nsConfig.KeyDirectory,
+		"Path to store the keys (indexes), this directory should reside in SSD disk. Set ephmeral to false if you want to define this setting.",
+	)
+	flagSet.String(
+		nsConfig.namespace+suffixValueDirectory,
+		nsConfig.ValueDirectory,
+		"Path to store the values (spans). Set ephmeral to false if you want to define this setting.",
```
jpkrohling: s/ephmeral/ephemeral/

plugin/storage/badger/options.go
```diff
@@ -0,0 +1,124 @@
...
+	)
+	flagSet.Duration(
+		nsConfig.namespace+suffixSpanstoreTTL,
+		nsConfig.SpanStoreTTL,
+		"time.Duration to store the data. Default is 72 hours (3 days).",
+	)
+	flagSet.String(
+		nsConfig.namespace+suffixKeyDirectory,
+		nsConfig.KeyDirectory,
+		"Path to store the keys (indexes), this directory should reside in SSD disk. Set ephmeral to false if you want to define this setting.",
```
jpkrohling: s/ephmeral/ephemeral/

plugin/storage/badger/options.go
```diff
@@ -0,0 +1,124 @@
...
+func addFlags(flagSet *flag.FlagSet, nsConfig *NamespaceConfig) {
+	flagSet.Bool(
+		nsConfig.namespace+suffixEphemeral,
+		nsConfig.Ephemeral,
+		"Mark this storage ephemeral, data is stored in tmpfs (in-memory). Default is true.",
+	)
+	flagSet.Duration(
+		nsConfig.namespace+suffixSpanstoreTTL,
+		nsConfig.SpanStoreTTL,
+		"time.Duration to store the data. Default is 72 hours (3 days).",
```
jpkrohling: The "default" part is added automatically by Viper. This is how it looks like when you run `SPAN_STORAGE_TYPE=badger go run ./cmd/collector/main.go --help`

```
--badger.span-store-ttl duration    time.Duration to store the data. Default is 72 hours (3 days). (default 72h0m0s)
```

I think you can remove the ` time.` as well from `time.Duration to store ...`

plugin/storage/badger/options.go
```diff
@@ -0,0 +1,124 @@
...
+	)
+	flagSet.String(
+		nsConfig.namespace+suffixValueDirectory,
+		nsConfig.ValueDirectory,
+		"Path to store the values (spans). Set ephmeral to false if you want to define this setting.",
+	)
+	flagSet.Bool(
+		nsConfig.namespace+suffixSyncWrite,
+		nsConfig.SyncWrites,
+		"If all writes should be synced immediately. This will greatly reduce write performance and will require fast SSD drives. Default is false.",
```
jpkrohling: > will require fast SSD drives

If I'm not on SSD, are you stopping the process? If not, what does "require" mean here?

plugin/storage/badger/options.go
```diff
@@ -0,0 +1,124 @@
...
+	for _, cfg := range opt.others {
+		addFlags(flagSet, cfg)
+	}
+}
+
+func addFlags(flagSet *flag.FlagSet, nsConfig *NamespaceConfig) {
+	flagSet.Bool(
+		nsConfig.namespace+suffixEphemeral,
+		nsConfig.Ephemeral,
+		"Mark this storage ephemeral, data is stored in tmpfs (in-memory). Default is true.",
```
jpkrohling: The "default" part is added automatically by Viper. This is how it looks like when you run `SPAN_STORAGE_TYPE=badger go run ./cmd/collector/main.go --help`

```
--badger.ephemeral                  Mark this storage ephemeral, data is stored in tmpfs (in-memory). Default is true. (default true)
```

plugin/storage/badger/options.go
```diff
@@ -0,0 +1,124 @@
...
+// InitFromViper initializes Options with properties from viper
+func (opt *Options) InitFromViper(v *viper.Viper) {
+	initFromViper(opt.primary, v)
+	for _, cfg := range opt.others {
+		initFromViper(cfg, v)
+	}
+}
+
+func initFromViper(cfg *NamespaceConfig, v *viper.Viper) {
+	cfg.Ephemeral = v.GetBool(cfg.namespace + suffixEphemeral)
```
jpkrohling: If this is also intended to replace the current in-memory storage, we need a `max-traces` option as well. See #845/#842

objectiser: Do we need a `max-traces` option? or just a strategy for dealing with a full disk?

plugin/storage/badger/options.go
burmanm: I'm not sure of the removal of time.Duration, how would a user know which format? I personally hate  if programs hide this type of information somewhere else and assume some sort of "everyone should know this is a Golang program using Golang standards which are not standards anywhere else"

plugin/storage/badger/options.go
burmanm: Mostly that this backend really wants SSDs and if you're going to sync all the writes, you probably want something with fast write time. I've removed it for now since technically you could use a DRAM backed disk array also for writes (but reads would be slow most likely)

plugin/storage/badger/options.go
jpkrohling: > some sort of "everyone should know this is a Golang program using Golang standards which are not standards anywhere else"

But isn't it exactly what you are doing when you say `time.Duration`? How one would know it's about https://golang.org/pkg/time/#Duration ?

burmanm: Google returns it to me. But I can add link to that to the description.

burmanm:
Rebased. I also fixed a generic issue with the testing framework inside Jaeger which happens to test against pointer locations instead of real data.

plugin/storage/badger/spanstore/reader.go
```diff
@@ -0,0 +1,488 @@
...
+	buf := new(bytes.Buffer)
+	buf.WriteByte(spanKeyPrefix)
+	binary.Write(buf, binary.BigEndian, traceID.High)
+	binary.Write(buf, binary.BigEndian, traceID.Low)
+	return buf.Bytes()
+}
+
+// GetServices fetches the sorted service list that have not expired
+func (r *TraceReader) GetServices() ([]string, error) {
+	return r.cache.GetServices()
```
SwarnimRaj: I noticed an issue with this implementation that in case of initializing badger factory by reloading from a pre-existing directory, the services list does not get populated with services from pre-existing traces even though the traces are accessible by making a direct REST API call.
Probably the cause is that you are relying on the cache which is not pre-populated.

plugin/storage/badger/spanstore/reader.go
```diff
@@ -0,0 +1,488 @@
...
+}
+
+// GetOperations fetches operations in the service and empty slice if service does not exists
+func (r *TraceReader) GetOperations(service string) ([]string, error) {
+	return r.cache.GetOperations(service)
```
SwarnimRaj: I noticed an issue with this implementation that in case of initializing badger factory by reloading from a pre-existing directory, the operations list does not get populated with services from pre-existing traces even though the traces are accessible by making a direct REST API call.
Probably the cause is that you are relying on the cache which is not pre-populated.

burmanm: NewCacheStore() calls prefillCaches()

burmanm:
@SwarnimRaj Here's a very rudimentary unit test that shows how to load an existing file: https://gist.github.com/burmanm/04021a32d8fb728792eb49ab0044dae8

Now, as for backup & restore, there's a separate tool for that (which should take care of LSM syncing). I have not considered such use case and for that reason there's no instructions for it either. But it can be done by following the instructions here: https://github.com/dgraph-io/badger#database-backup

plugin/storage/badger/spanstore/reader.go
burmanm: The unit test that does persisting (mentioned in other comment) also seems to indicate the caches are prefilled correctly.

I should add this type unit tests to the suite.

SwarnimRaj: I took a backup of Badger db and restored it to a location that was passed as key and value dir to the Jaeger standalone run. I noticed the loadServices() method does find services present in traces of the backup but finds its TTL as 0. This causes the services to be deleted during GetServices() as your logic finds it as expired. I did take a look at your test and somehow TTL in this case is 72 hours ahead of now time so the test passes.
I am not sure why services loaded from a pre-existing Badger db has 0 TTL. Do you know what might be causing this?

For now, commenting the delete line in GetServices() helps populate the services on UI- similarly for GetOperations(). Is there a danger in this approach? Why do you perform the check (v > t) in GetServices?

PS: My use case demands that backed up data should be accessible always.

burmanm: I can't replicate that behavior, in my testing I see:

```
Key: key="\x81service-p\x00\x05pSgE\x89\xbd\x00\x00\x00\x00\x00\x00\x00\x01\x00\x00\x00\x00\x00\x00\x00\x01", version=1, meta=40, 1531137600, time now: 1530878401
Key: key="\x82service-poperation-p\x00\x05pSgE\x89\xbd\x00\x00\x00\x00\x00\x00\x00\x01\x00\x00\x00\x00\x00\x00\x00\x01", version=1, meta=40, 1531137600, time now: 1530878401
```

That is, the services do expire later than the current time (as they should). The reason for that ``v > t`` is that we don't return ghost services or operations. That would be bad in terms of usability if one has a lot of services / operations that keep changing. 

If you need backups that do not expire ever, then I guess TTL is the wrong approach for you. What you'd want is something closer to a snapshot + purge of snapshotted data (with no TTL, such that you would manually take care of retention policies).

Certainly doable, but maybe it is a bit of feature creep for this PR?

burmanm:
Rebased and squashed all the existing commits (rebasing was becoming a pain with large amount of commits).

burmanm:
That's not actually an error (I'm not expecting it to cleanup every time), but an error message I haven't filtered out. That segmentation violation is more interesting, could you paste that log (golang bug?).

SwarnimRaj:
@burmanm The segmentation violation stack trace-
{"level":"error","ts":1531462378.2343898,"caller":"badger/factory.go:151","msg":"ValueLogGC run failed with ","error":"Value log GC attempt didn't result in any cleanup","errorVerbose":"Value log GC attempt didn't result in any cleanup\ngithub.com/jaegertracing/jaeger/vendor/github.com/dgraph-io/badger.init\n\t/Users/swarnim/go/src/github.com/jaegertracing/jaeger/vendor/github.com/dgraph-io/badger/errors.go:64\ngithub.com/jaegertracing/jaeger/plugin/storage/badger.init\n\t<autogenerated>:1\ngithub.com/jaegertracing/jaeger/plugin/storage.init\n\t<autogenerated>:1\ngithub.com/jaegertracing/jaeger/cmd/env.init\n\t<autogenerated>:1\nmain.init\n\t<autogenerated>:1\nruntime.main\n\t/usr/local/Cellar/go/1.10.3/libexec/src/runtime/proc.go:186\nruntime.goexit\n\t/usr/local/Cellar/go/1.10.3/libexec/src/runtime/asm_amd64.s:2361","stacktrace":"github.com/jaegertracing/jaeger/plugin/storage/badger.(*Factory).maintenance\n\t/Users/swarnim/go/src/github.com/jaegertracing/jaeger/plugin/storage/badger/factory.go:151"}
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xb04fa0]

goroutine 18023 [running]:
expvar.(*Int).Set(...)
        /usr/local/Cellar/go/1.10.3/libexec/src/expvar/expvar.go:65
github.com/jaegertracing/jaeger/plugin/storage/badger.(*Factory).diskStatisticsUpdate(0xc420338090)
        /Users/swarnim/go/src/github.com/jaegertracing/jaeger/plugin/storage/badger/stats_linux.go:17 +0x100
github.com/jaegertracing/jaeger/plugin/storage/badger.(*Factory).maintenance(0xc420338090)
        /Users/swarnim/go/src/github.com/jaegertracing/jaeger/plugin/storage/badger/factory.go:153 +0xa5
created by github.com/jaegertracing/jaeger/plugin/storage/badger.(*Factory).Initialize
        /Users/swarnim/go/src/github.com/jaegertracing/jaeger/plugin/storage/badger/factory.go:107 +0x1bf

I see that the file paths in stacktrace are pointing to a location on my local laptop whereas the binary was being executed on a linux server that does not have these files checked out. Could the segmentation violation be happening due to that because in my local environment, it never causes such segmentation violation despite giving the non-error error message.


Could you also comment your views on the live badger backup and log rotation questions.

burmanm:
The segmentation fault should have been fixed in the commit a5858cd. There's a way to create a backup of Badger while it is running, but that tool might not be able to do it. There's however a DB.Backup() call that could do it on the running one. But that would require changes elsewhere in the Jaeger to expose at this point. Or maybe the tool could use some work to avoid that limitaiton (there's no technical reason why it couldn't be done, just would require some merging of disk LSM tree and WAL by the backup tool)

The log rotation is slightly more difficult since badger isn't using time based compactions in the LSM tree.  There is a setting to set the file size for each ValueLog file, but I don't know how it would behave in your case (the key itself would be in another tree and point to a non-existing value at that point). You could test this by setting in factory.go Initialization() the opts.MaxTableSize and opts.ValueLogFileSize

Cassandra backend would theoretically have the exact features you need since TimeWindowCompactionStrategy can be configured in the expected way.

burmanm:
@abbi-gaurav I'm not really sure what's the delay here. I don't think there's anything missing in the scope of this PR.

plugin/storage/badger/dependencystore/storage.go
```diff
@@ -0,0 +1,90 @@
...
+const (
+	dependencyKeyPrefix byte = 0xC0 // Dependency PKs have first two bits set to 1
+)
+
+// DependencyStore handles all queries and insertions to Cassandra dependencies
+type DependencyStore struct {
+	// session                  cassandra.Session
+	// dependencyDataFrequency  time.Duration
+	// dependenciesTableMetrics *casMetrics.Table
+	// logger                   *zap.Logger
```
yurishkuro: what is the plan with the commented out code?

model/sort.go
```diff
@@ -18,6 +18,24 @@ import (
...
+	} else if s[i].High > s[j].High {
+		return false
+	}
+	return s[i].Low < s[j].Low
+}
+
+// SortTraceIDs sorts a list of TraceIDs
+func SortTraceIDs(traceIDs []*TraceID) {
+	sort.Sort(byTraceID(traceIDs))
+}
```
yurishkuro: This does not seem general purpose enough to be in the `model`.

plugin/storage/badger/factory.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+const (
+	defaultTickerInterval time.Duration = 5 * time.Minute
+)
+
+// NewFactory creates a new Factory.
+func NewFactory() *Factory {
+	if ValueLogSpaceAvailable == nil {
+		ValueLogSpaceAvailable = expvar.NewInt("badger_value_log_bytes_available")
```
yurishkuro: why not initialize these in the declaration section?

plugin/storage/badger/factory.go
```diff
@@ -0,0 +1,179 @@
...
+	if f.Options.primary.Ephemeral {
+		opts.SyncWrites = false
+		// Error from TempDir is ignored to satisfy Codecov
+		dir, _ := ioutil.TempDir("", "badger")
+		f.tmpDir = dir
+		opts.Dir = f.tmpDir
+		opts.ValueDir = f.tmpDir
+
+		f.Options.primary.KeyDirectory = f.tmpDir
+		f.Options.primary.ValueDirectory = f.tmpDir
```
yurishkuro: should these all be subdirectories?  Reusing the same one makes assumption about naming that individual components will use.

plugin/storage/badger/factory.go
```diff
@@ -0,0 +1,179 @@
...
+		opts.ValueDir = f.Options.primary.ValueDirectory
+	}
+
+	store, err := badger.Open(opts)
+	if err != nil {
+		return err
+	}
+	f.store = store
+
+	// Error ignored to satisfy Codecov
```
yurishkuro: The error comes from the internal logic, is there no way to simulate it? If we're always ignoring it, is there a reason to even return it?

plugin/storage/badger/factory.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+// CreateDependencyReader implements storage.Factory
+func (f *Factory) CreateDependencyReader() (dependencystore.Reader, error) {
+	return nil, nil
+}
+
+// Close Implements io.Closer and closes the underlying storage
+func (f *Factory) Close() error {
+	f.maintenanceTicker.Stop()
```
yurishkuro: this won't make the `maintenance()` func exit since the ticker channel is never closed.

plugin/storage/badger/factory.go
```diff
@@ -0,0 +1,179 @@
...
+
+	return err
+}
+
+// Maintenance starts a background maintenance job for the badger K/V store, such as ValueLogGC
+func (f *Factory) maintenance() {
+	_, previousSize := f.store.Size()
+	for range f.maintenanceTicker.C {
+		_, vlogSize := f.store.Size()
+		if (previousSize + 1<<30) < vlogSize {
```
yurishkuro: what is the significance of `1<<30`? should it be a named constant or even configurable?

plugin/storage/badger/factory.go
```diff
@@ -0,0 +1,179 @@
...
+			previousSize = vlogSize
+			var err error
+			for err == nil {
+				err = f.store.RunValueLogGC(0.5)
```
yurishkuro: what is `0.5`?

plugin/storage/badger/options.go
```diff
@@ -0,0 +1,146 @@
...
+			SpanStoreTTL:   time.Hour * 72, // Default is 3 days
+			SyncWrites:     false,          // Performance over durability
+			Ephemeral:      true,           // Default is ephemeral storage
+			ValueDirectory: defaultDataDir + defaultValueDir,
+			KeyDirectory:   defaultDataDir + defaultKeysDir,
+		},
+		others: make(map[string]*NamespaceConfig, len(otherNamespaces)),
+	}
+
+	// Commented out to satisfy Codecov
```
yurishkuro: ?

plugin/storage/badger/options.go
```diff
@@ -0,0 +1,146 @@
...
+func getCurrentExecutableDir() string {
+	// We ignore the error, this will fail later when trying to start the store
+	exec, _ := os.Executable()
+	return filepath.Dir(exec)
+}
+
+// AddFlags adds flags for Options
+func (opt *Options) AddFlags(flagSet *flag.FlagSet) {
+	addFlags(flagSet, opt.primary)
+	// Commented out to satisfy Codecov
```
yurishkuro: Noticing a pattern. If other namespaces are not implemented, we should just remove them.

plugin/storage/badger/spanstore/reader.go
```diff
@@ -0,0 +1,488 @@
...
+	"time"
+
+	"github.com/dgraph-io/badger"
+	"github.com/golang/protobuf/proto"
+
+	"github.com/jaegertracing/jaeger/model"
+	"github.com/jaegertracing/jaeger/storage/spanstore"
+)
+
+// All of this is replicated from the ES and Cassandra storage parts.. they should be refactored to a common place
```
yurishkuro: what is this comment referring to?

plugin/storage/badger/spanstore/reader.go
```diff
@@ -0,0 +1,488 @@
...
+				// Add value to the span store (decode from JSON / defined encoding first)
+				// These are in the correct order because of the sorted nature
+				item := it.Item()
+				val, err := item.ValueCopy(val)
+				if err != nil {
+					return err
+				}
+
+				sp := model.Span{}
+				switch item.UserMeta() & 0x0F {
```
yurishkuro: What is 0x0F ?

plugin/storage/badger/spanstore/reader.go
```diff
@@ -0,0 +1,488 @@
...
+			}
+			trace := &model.Trace{
+				Spans: spans,
+			}
+			traces = append(traces, trace)
+		}
+		return nil
+	})
+
+	// TODO Do the Unmarshal here so we can release the transaction earlier (we would pay with extra allocations..)
```
yurishkuro: is this still needed?

pavolloffay:
I think this could be updated to use new protobuf model

yurishkuro:
I think it already uses proto: https://github.com/jaegertracing/jaeger/pull/760/files#diff-a70dbac78fcc916b6a16ea1703bbc6d2R182

plugin/storage/badger/factory.go
burmanm: The idea behind these being under different directories is to allow a mix of SSD + HDD combination. The Keys are recommended to be stored to a SSD for better performance, but the values could reside in the HDD also. There's a different access pattern / compaction process in the badger.

burmanm: The idea behind these being under different directories is to allow a mix of SSD + HDD combination. The Keys are recommended to be stored to a SSD for better performance, but the values could reside in the HDD also. There's a different access pattern / compaction process in the badger.

plugin/storage/badger/factory.go
burmanm: The error could come from corrupted storage / other issues in database files. I think these should be returned, but making badger fail to load LSM tree for tests is quite beyond the scope of Jaeger I would say. But codecov does not allow returning it (I'm sure user would be happy to know that something went wrong).

plugin/storage/badger/factory.go
burmanm: It's a discard ratio, as in if half of the file can be reclaimed it will be rewritten. The 0.5 was the recommendation/sane default in badger's code: https://github.com/dgraph-io/badger/blob/fbb27786246d20213f31bad8bf33f1504837029f/db.go#L988

All the values are compromises so I chose this as a safe one.

plugin/storage/badger/spanstore/reader.go
burmanm: I only want the last 4 bits of the UserMeta byte to be able to use switch for the encoding type. If there was no 0x0F, then that switch statement wouldn't match

plugin/storage/badger/spanstore/reader.go
burmanm: Not really, it's not a bottleneck at the moment.

model/sort.go
burmanm: If there's ``SortTraces``, ``SortTrace``, ``SortSpan`` then why wouldn't ``SortTraceID`` be proper? Feels logical to me.

plugin/storage/badger/dependencystore/storage.go
burmanm: Forgotten comments only, the original wish was not to include dependencystore in the first PR. Is this still the case?

plugin/storage/integration/fixtures/traces/multispottag_opname_trace.json
```diff
@@ -4,7 +4,6 @@
       "traceId": "AAAAAAAAAAAAAAAAAAAAFw==",
       "spanId": "AAAAAAAAAAQ=",
       "operationName": "query17-operation",
-      "references": [],
```
yurishkuro: is there a dependency in this PR on these changes?  If so, could they be made separately in a different PR?

plugin/storage/badger/factory.go
burmanm: They were originally, but I have to say I can't remember which case caused them to behave incorrectly in that scenario. Since I'm unable to replicate that issue anymore (maybe some integration test was updated in the rebases etc) I'll move them back.

plugin/storage/badger/factory.go
burmanm: I'll remove it at this point and make the timer configurable. Could revisit the idea if TWCS equivalent is added to badger.

manishrjain:
Hey guys, author of Badger here. Funny I've been using Jaeger over the past couple of weeks to find and fix Jepsen test failures in Dgraph. So glad to see this PR.

464 issue was fixed quite a while ago. Defrag is intended to be offline tool, so won't be of much use. Let me know how I can help review/test/fix this PR.

yurishkuro:
@manishrjain do tell! Sounds like a good material for a blog post.

manishrjain:
Hey guys, any thoughts on if and when this change could land into Jaeger? We're currently using the default storage, i.e. memory, and it's using up a lot of it, and performing quite slow when looking for traces.

otisg:
Just stumbled upon this issue.  We used Badger in one of our agents at Sematext.  It ate too much memory for our taste and we ended up switching to https://github.com/peterbourgon/diskv and have been very happy with it.  HTH.

burmanm:
@otisg diskv is not hugely useful in this context as it's a "hash-ordered" datastore. It uses btree package to get ordering index if one wants to search keys. But in thatcase the btree is actually the only useful data structure for this use-case, making diskv's hash table completely useless.

LSM-store by itself is very good solution for this use-case, the only real thing missing from badger is a compaction strategy based on time (such as Cassandra's TimeWindowCompactionStrategy). That would improve the performance (reduce I/O & CPU) of the compaction stage and help with purge of older data.

manishrjain:
@burmann Can you file a feature request for Badger, and I can take a look into how that'd be useful and potentially adding something like that.

So far I'm not clear about the issues that you are running into with Badger.

manishrjain:
Hi, @burmanm , @pavolloffay , @SwarnimRaj , @jpkrohling , and other related parties:

We're looking for a tracer tool to make part of our Dgraph k8s configuration, so all our users can debug query latency issues. We're debating between writing our own on top of Badger or assigning resources to help Jaeger support Badger.

I see that work on your side has already been mostly done, but given this PR has been pending for almost 8 months now doesn't inspire confidence. So, I'd like to put together a meeting to understand the timeline here to integrate Badger in Jaeger; and decide upon our course of action.

Would it be possible to have a video chat with the right people involved in this PR? It might help move things along, or at least understand what are priorities from Jaeger POV. My email id is below.

Cheers,
Manish
manish@dgraph.io

yurishkuro:
@manishrjain this is mostly ready to be merged, just needs a rebase & green build.

Feel free to join our [bi-weekly status call][status-call], next one is on Dec 28. Or poke me on ys@uber.com

[status-call]: https://github.com/jaegertracing/jaeger/blob/master/README.md#project-status-bi-weekly-meeting

burmanm:
For some reason I can't reply to your last comment about ``references: []``, in any case it was about a test failure since test data did not follow the matching prototype requirements (omitempty). It's probably just JSON tests though, not protobuf (and maybe current master / trunk does not have this issue - I haven't checked).

I'll go through the rest and fix them in the week starting 2nd of January (I'm away the Christmas time).

plugin/storage/badger/factory.go
```diff
@@ -0,0 +1,171 @@
...
+
+var (
+	// ValueLogSpaceAvailable returns the amount of space left on the value log mount point in bytes
+	ValueLogSpaceAvailable = expvar.NewInt("badger_value_log_bytes_available")
+	// KeyLogSpaceAvailable returns the amount of space left on the key log mount point in bytes
+	KeyLogSpaceAvailable = expvar.NewInt("badger_key_log_bytes_available")
+	// LastMaintenanceRun stores the timestamp of the previous maintenanceRun
+	LastMaintenanceRun = expvar.NewInt("badger_storage_maintenance_last_run")
+	// LastValueLogCleaned stores the timestamp of the previous ValueLogGC run
+	LastValueLogCleaned = expvar.NewInt("badger_storage_valueloggc_last_run")
```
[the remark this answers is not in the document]
burmanm: badger itself exposes metrics (which are not replicated here) using expvar. If these are not exposed by default, then can I somehow encapsulate expvar available metrics through MetricsFactory?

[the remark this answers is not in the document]
burmanm: Prometheus does have expvar Collector, but I didn't see it being called as an option in the MetricsFactory. So should it be implemented there or are they on purpose ignored by default?

plugin/storage/badger/factory_test.go
```diff
@@ -0,0 +1,104 @@
...
+	"os"
+	"testing"
+	"time"
+
+	assert "github.com/stretchr/testify/require"
+	"github.com/uber/jaeger-lib/metrics"
+	"go.uber.org/zap"
+
+	"github.com/jaegertracing/jaeger/pkg/config"
+)
```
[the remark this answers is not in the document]
burmanm: I'm not entirely sure how would this help? The temp dir is different for each instance, thus call to another Close wouldn't help at all? Each call to ioutil.TempDir creates a random number which is in the directory name.

I made the change now in the Close() that even if f.store.Close() fails, we will still delete the tempdir first instead of just returning the error.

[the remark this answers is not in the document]
yurishkuro: sorry, I simply meant to add that line as a validation that the factory implements io.Closer, because it is how the Close method is used by the meta-factory, it tries to cast the objects to io.Closer. If this factory changes its Close method signature, e.g. to not return an error, the cast will fail and Close() won't be called.

plugin/storage/badger/factory.go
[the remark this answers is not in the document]
yurishkuro: I think we're talking about two different things here. The above code declares metrics emitted by the Jaeger code. Everywhere in Jaeger we're using MetricsFactory for that, and the user can choose which metrics backend they want, including expvar or prometheus. So there's no strong reason to hardcode just the expvar implementation for those.

Separately, you're saying that Badger internals are exposing expvar metrics too. This one is slightly more difficult, because iirc we don't currently enable expvar http handler if the metrics backend is not expvar. Maybe we should always enable the /debug/expvar endpoint even if prometheus is the main metrics selection. Alternatively, if prom client has an adapter that can expose expvar via prom scraping endpoint, we could use that.

burmanm:
Travis fails seem to have nothing to do with this PR

yurishkuro:
weird, looks like maybe a bug in golint, it fails with different nonsensical errors
```
./cmd/collector/app/sanitizer/service_name_sanitizer.go is in package sanitizer, not reporter
```
and on another run
```
./cmd/ingester/app/processor/span_processor.go is in package processor, not sanitizer
```

yurishkuro:
Build is fixed. Looks like `maintenance() in factory/go` has a few lines not covered by tests:

https://codecov.io/gh/jaegertracing/jaeger/compare/6bc0c122bfca8e737a747826ae60a22a306d7019...74f33e549f495e59b33c028fb0c5ad62dcc83879/diff#D16-164

burmanm:
One of those badger's internal errors again, means because of Codecov we should ignore it and let the user "guess".

Or make internal parts public so they can be adjusted in the test, like closing the underlying DB and then calling maintenance on it. Which would bad design just to satisfy Codecov.

yurishkuro:
@burmanm add `package <name>` to the empty_test files, and run `make fmt`. I could've done it, but you missed sign-off on the latest commit.

burmanm:
Right, so codecov can't understand tests from another package if empty_test.go is added and empty_test.go is required by another verification.

yurishkuro:
Rerunning the build. I was under impression that it only had a few lines of error handling not tested, but the codecov was showing a massive drop to 93% overall, and only 20% of the diff coverage.

plugin/storage/badger/factory_test.go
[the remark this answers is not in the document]
yurishkuro: I found it to be a good practice to be explicit about interfaces you need to implement, rather than relying on indirect testing of the fact. I.e. this in the test file sets the expectation unambiguously:

```go
var _ io.Closer = new(myType)
```

yurishkuro:
@burmanm a few minor things I want to clean up. A couple questions:

* Is it really necessary to support JSON encoding of data? I see it's not implemented, causing codecov drop, incidentally, and I don't see much value in it in the first place.
* The dependencies store has a separate implementation of reading spans. Perhaps it would make more sense to move that into spanstore and expose something like LoadSpansForTimeRange on the reader struct and make depstore depend on it?

burmanm:
No, the JSON support was done before the proto was implemented to Jaeger. Thus it's probably only historical stuff. Although I'm not sure what you mean with the "not implemented" ? Both reader & writer implement it, but to use it one needs to make a change to the source code instead of command line property. Thus, only the configuration part is not implemented. Maybe I should implement that one instead, if someone wants to debug their system with some external tool by just reading the badger's db-file. 

I'll refactor the dependency reader.

burmanm:
Changed the dependencyreader to use spanstore instead of directly reading from the DB. The negative side is that there's increased memory usage as the []model.Trace must be kept in the memory while loading. Previously each model.Trace could have been thrown away by escape analysis.

In any case, there wasn't any ``context`` delivered to the dependencyreader unlike to the SpanReader/SpanWriter, so I used context.Background() to fetch data from spanstore to dependencystore.

jpkrohling:
Once the merge conflict is solved, can this be merged, or is there still something pending?

yurishkuro:
I think we mostly need to address #1389.

yurishkuro:
So the build is green, and the code coverage decreases are only on error checks. I looked around for any tools that allow ignoring certain lines from code coverage, didn't find any.

manishrjain:
Amazing! Can't wait to use it.

pd40:
I am looking forward to using this as well!

Forgive me if this has already been covered, but will the deployment instructions be updated to include badger as part of this change?

https://github.com/jaegertracing/documentation/blob/master/content/docs/1.11/deployment.md

yurishkuro:
It can already be used it with `jaegertracing/all-in-one` latest. It won't work with any other deployment modes though, because badger runs in-process and cannot be shared by collector & query.

sherwoodzern:
I read that this alternative storage option is available; however, the documentation still only mentions ElasticSearch, Cassandra, and Kafka. I'm trying to understand how I would configure this latest option. I want to have all traces written to a non-ephemeral storage. Does this option provide this capability without having to install ElasticSearch, Cassandra, or Kafka? Any examples of deployment using this approach discussed in this thread?

yurishkuro:
Storage options are documented in the [Deployment](https://www.jaegertracing.io/docs/latest/deployment/) section of the docs

image

sherwoodzern:
Thank you for the response regarding the storage options. I want to confirm that within my Kubernetes deployment, I can create a PVC and specify that storage to be used by Badger. This will require that I deploy a Badger container and would assume that I will deploy as a daemonset; is this correct? I then specify the storage endpoint when configuring Jaeger with badger. In this manner, I should get all of the traces stored in a non-ephemeral storage.

yurishkuro:
@sherwoodzern as explained in the documentation:

> Badger is an **embedded** local storage, **only available with all-in-one distribution**.

You cannot use it with normal production (i.e. scalable) installation of Jaeger, because the storage is embedded in a single process that includes both collection and UI components.

PR #761, opened 2018-04-02
Updated Italian font for "Strongly typed" in readme file.

Shreya1771:
Updated Italian font for "Strongly typed"  in readme file.


shailesh:
Hi @black-adder I thought if this is typo as it should be "  Support strongly typed _span tags_ and _structured logs_ " instead of " Support strongly typed span _tags_ and _structured logs_ " just for highlighting the span tags.

Shreya1771:
Hi @black-adder  I thought if this is typo as it should be " Support strongly typed span tags and structured logs " instead of " Support strongly typed span tags and structured logs " just for highlighting the span tags.

yurishkuro:
I think the original form is better. Spans contain both tags AND logs, so why would we group the word "span" with "logs" only?

PR #763, opened 2018-04-03
[WIP] Influx metrics reporter

mabn:
This adds influx metrics reporter, it uses gokit wrapper added in https://github.com/jaegertracing/jaeger-lib/pull/23

I had to add a bunch of flags because (unlike prometheus/expvar) influx reporter requires at least url, credentials and database name. I also wanted to be able to add some predefined tags to be able to indicate the environment name.

### DIFF
--- a/cmd/agent/app/builder.go
+++ b/cmd/agent/app/builder.go
@@ -113,12 +113,12 @@ func (b *Builder) createMainReporter(mFactory metrics.Factory, logger *zap.Logge
 	return b.CreateReporter(mFactory, logger)
 }
 
-func (b *Builder) getMetricsFactory() (metrics.Factory, error) {
+func (b *Builder) getMetricsFactory(logger *zap.Logger) (metrics.Factory, error) {
 	if b.metricsFactory != nil {
 		return b.metricsFactory, nil
 	}
 
-	baseFactory, err := b.Metrics.CreateMetricsFactory("jaeger")
+	baseFactory, err := b.Metrics.CreateMetricsFactory("jaeger", logger)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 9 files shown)


jpkrohling:
I wonder if we shouldn't have the same approach to the metrics backend as we are planning for the storage, in that we'd define an interface and delegate the implementation to plugins.

mabn:
@jpkrohling What should I do with this PR? Is it likely to get included? Or is the plugin approach the way to go?

yurishkuro:
Not "plugin" in the sense of storage plugins (which could be in a different process, for example), but the factory pattern used for storage https://github.com/jaegertracing/jaeger/blob/master/storage/factory.go

The main thing in that factory pattern is that CLI options do not show up unless a specific factory type is enabled.

Alternatively, is this heavy direct integration with influxDb the only way to report metrics to it? Isn't it possible to keep using Prometheus metrics in Jaeger and run a separate scraper that pulls the metrics from the process and converts then into influxDb?

jpkrohling:
+1 to what @yurishkuro wrote. It's clear by now that we'll not go with [`go plugins`](https://golang.org/pkg/plugin/) and this is what I had when I wrote the comment back in April.

mabn:
I guess I can check how it would look with prometheus reporting to influx. IIRC I tried this a year ago and something was off with histograms.

mabn:
I checked [telegraf](https://github.com/influxdata/telegraf) - it has an input plugin for Prometheus and an output plugin for Influx. Unfortunately it does not support histograms - Prometheus histogram buckets are sent to influx as-is (and fields of a metric). And it's not possible to make use of them (calculate e.g. 99th percentile) because there isn't anything in Influx which could behave like `histogram_quantile()` from Prometheus.

I didn't try to run Prometheus with Influx storage but I'm sure it would be the same case.

This basically means that latency percentiles wouldn't be available and that is not cool.

yurishkuro:
> I didn't try to run Prometheus with Influx storage but I'm sure it would be the same case.

Why do you think so?  I expect that Prometheus stores latencies in the same format as it captures them, as a group of buckets with counts (which should be fine for Influx), and the percentiles are calculated by Prometheus server.

yurishkuro:
@mabn I am going to close this old PR, since it's unlikely that we want to support direct integration w/ InfluxDB for metrics. In the future we will likely upgrade Jaeger internals to use OpenTelemetry, and there probably will be some standard integration of OpenTelemetry with InfluxDB so that we won't need anything special in Jaeger.

PR #764, opened 2018-04-04
Add prettyPrint param to API endpoints

### LINKED ISSUE #756: query-service - Format "Trace JSON"
> Format "Trace JSON" output so the output is nicely readable.

From jaegertracing/jaeger-ui#199

yurishkuro:
Fixes #756 

cc @tiffon

### DIFF
--- a/cmd/query/app/handler_test.go
+++ b/cmd/query/app/handler_test.go
@@ -167,7 +167,6 @@ func TestPrettyPrint(t *testing.T) {
 			defer server.Close()
 
 			out := get(server.URL + testCase.param)
-			println(out)
 			assert.Equal(t, testCase.output, out)
 		})
 	}


cmd/query/app/handler_test.go
```diff
@@ -139,6 +139,24 @@ func TestGetTraceSuccess(t *testing.T) {
...
+	get := func(params string) string {
+		res, err := http.Get(server.URL + "/api/traces/123456" + params)
+		require.NoError(t, err)
+		body, err := ioutil.ReadAll(res.Body)
+		require.NoError(t, err)
+		return string(body)
+	}
+	raw := get("")
+	pretty := get("?prettyPrint=1")
+	assert.True(t, 2*len(raw) < len(pretty), "formattet response must be longer")
```
vprithvi: Could we make this test more robust by comparing the output to an expected string instead of simply checking that the formatted string is longer?

Issue #765, opened 2018-04-04
Agent for Windows


grounded042:
With the new [jaeger-client-csharp](https://github.com/jaegertracing/jaeger-client-csharp) there will be those wanting to run the Jaeger Agent on Windows where their services are hosted. I've tested building the agent with `GOOS=windows` and it works on Windows. 

Would we be able to publish the binary produced for Windows? I think having a Windows binary available for download would greatly help the adoption of the C# client and those on Windows. I can submit a PR with the Makefile targets to build the agent for Windows.

yurishkuro:
Would be nice. There have been requests for binary distros for Linux too, in addition to docker images. The main thing we're missing is the deploy stage in Travis, so a PR to add that would be most welcome.

yurishkuro:
@grounded042 thanks for the pull request. The binaries are now part of the release, check out https://github.com/jaegertracing/jaeger/releases/tag/v1.4.1

PR #766, opened 2018-04-04
add make and travis options for building for windows

grounded042:
This is for #765. It will build executables for windows and in the travis job it will publish the windows and linux binaries with the GitHub release when the build is tagged. Out of the box the standalone binary and the query binary won't work as they need the UI files location specified.

Also, before this is merged someone with access to the repo will have to add a secure key to the travis deploy so the files can be sent along to GitHub.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,9 @@ matrix:
     - HOTROD=true
   - go: 1.9
     env:
+    - LINUX=true
+    - go: 1.9
+    env:
     - WINDOWS=true
   - go: 1.9
     env:
@@ -60,6 +63,7 @@ script:
   - if [ "$DOCKER" == true ]; then bash ./scripts/travis/build-docker-images.sh ; else echo 'skipping docker images'; fi
   - if [ "$ES_INTEGRATION_TEST" == true ]; then bash ./scripts/travis/es-integration-test.sh ; else echo 'skipping elastic search integration test'; fi
   - if [ "$HOTROD" == true ]; then bash ./scripts/travis/hotrod-integration-test.sh ; else echo 'skipping hotrod example'; fi
+  - if [ "$LINUX" == true ]; make build-linux ; else echo 'skipping linux'; fi
   - if [ "$WINDOWS" == true ]; make build-windows ; else echo 'skipping windows'; fi
   - if [ "$DARWIN" == true ]; make build-darwin ; else echo 'skipping darwin'; fi
 
@@ -68,6 +72,9 @@ after_success:
 
 after_failure:
   - if [ "$CROSSDOCK" == true ]; then make crossdock-logs ; else echo 'skipping crossdock'; fi
+
+before_deploy:
+  - bash ./scripts/travis/package-deploy.sh
 
 deploy:
   provider: releases
@@ -86,3 +93,4 @@ deploy:
   on:
     tags: true
     repo: jaegertracing/jaeger
+    branch: master
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,9 @@ matrix:
   - go: 1.9
     env:
     - WINDOWS=true
+  - go: 1.9
+    env:
+    - DARWIN=true
 
 services:
   - docker
@@ -58,6 +61,7 @@ script:
   - if [ "$ES_INTEGRATION_TEST" == true ]; then bash ./scripts/travis/es-integration-test.sh ; else echo 'skipping elastic search integration test'; fi
   - if [ "$HOTROD" == true ]; then bash ./scripts/travis/hotrod-integration-test.sh ; else echo 'skipping hotrod example'; fi
   - if [ "$WINDOWS" == true ]; then bash ./scripts/travis/build-windows.sh ; else echo 'skipping windows'; fi
+  - if [ "$DARWIN" == true ]; make build-darwin ; else echo 'skipping darwin'; fi
 
 after_success:
   - if [ "$COVERAGE" == true ]; then travis_retry goveralls -coverprofile=cover.out -service=travis-ci || true ; else echo 'skipping coverage'; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,9 @@ matrix:
   - go: 1.9
     env:
     - HOTROD=true
+  - go: 1.9
+    env:
+    - WINDOWS=true
 
 services:
   - docker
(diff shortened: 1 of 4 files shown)


scripts/travis/build-windows.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
yurishkuro: is this script necessary? I.e. why not just call `make build-windows` from .travis?

Makefile
```diff
@@ -147,18 +147,34 @@ build_ui:
...
+build-all-in-one-windows: build_ui
+	CGO_ENABLED=0 GOOS=windows installsuffix=cgo go build -o ./cmd/standalone/standalone-windows.exe $(BUILD_INFO) ./cmd/standalone/main.go
+
 .PHONY: build-agent-linux
 build-agent-linux:
 	CGO_ENABLED=0 GOOS=linux installsuffix=cgo go build -o ./cmd/agent/agent-linux $(BUILD_INFO) ./cmd/agent/main.go
 
+.PHONY: build-agent-windows
+build-agent-windows:
+	CGO_ENABLED=0 GOOS=windows installsuffix=cgo go build -o ./cmd/agent/agent-windows.exe $(BUILD_INFO) ./cmd/agent/main.go
```
yurishkuro: Any reason not to call this `agent.exe`? Is there any other platform that uses `.exe` extension?

.travis.yml
```diff
@@ -54,9 +57,28 @@ script:
...
   - if [ "$HOTROD" == true ]; then bash ./scripts/travis/hotrod-integration-test.sh ; else echo 'skipping hotrod example'; fi
+  - if [ "$WINDOWS" == true ]; then bash ./scripts/travis/build-windows.sh ; else echo 'skipping windows'; fi
 
 after_success:
   - if [ "$COVERAGE" == true ]; then travis_retry goveralls -coverprofile=cover.out -service=travis-ci || true ; else echo 'skipping coverage'; fi
 
 after_failure:
   - if [ "$CROSSDOCK" == true ]; then make crossdock-logs ; else echo 'skipping crossdock'; fi
+
+deploy:
```
yurishkuro: I am not familiar with travis deploy workflow - where do these files end up?  And do we need a different naming conventions? E.g. https://prometheus.io/download/:
```
prometheus-2.2.1.darwin-amd64.tar.gz
prometheus-2.2.1.linux-amd64.tar.gz
prometheus-2.2.1.windows-amd64.tar.gz
```

Also, since these are archive files, do we want to bundle all binaries into one archive?

yurishkuro:
> Out of the box the standalone binary and the query binary won't work as they need the UI files location specified.

fwiw we should fix it using binary assets

scripts/travis/build-windows.sh
grounded042: Yup, I can move it to just running `make build-windows` I was following convention and created a script file. I'll change that up.

Makefile
grounded042: Just following convention since the agent for linux is called `agent-linux`. I'll change it up to be `agent.exe`.

.travis.yml
black-adder: also, we should change these commands to only build and deploy these binaries on landing the PR rather than every commit (unless this already does this)

Makefile
grounded042: @yurishkuro what if we named the different binaries for windows along the lines of `jaeger.<product>.exe` and dropped the `-windows` on all of them?

yurishkuro: my preference would be to have identical names but group them into dirs/archives by target, e.g.
```
jaeger-1.4.0-darwin-amd64.tar.gz
  - agent
  - collector
  - etc
jaeger-1.4.0-linux-amd64.tar.gz
  - agent
  - collector
  - etc
jaeger-1.4.0-windows-amd64.tar.gz
  - agent.exe
  - collector.exe
  - etc
```

.travis.yml
jpkrohling: > We could also add `branch: master` to further pin down when it deploys.

[I don't think that's what we want](https://docs.travis-ci.com/user/deployment/releases/):

> GitHub Releases uses git tags. If the build commit does not have any tags, one will be created in the form of `untagged-*`, where `*` is a random hex string.

scripts/travis/build-windows.sh
jpkrohling: Let's be nice to the next person that will touch this feature and keep it consistent with the other parts of the build :)

.travis.yml
grounded042: @jpkrohling I think that's only if you have `on.branch: true`. I was thinking that if you had multiple `on` options it would only deploy if all conditions were met. I'll look into it.

Makefile
grounded042: I like it. I'll work on getting that setup @yurishkuro.

.travis.yml
grounded042: I verified `When all conditions specified in the 'on:' section are met, your build will deploy.` So if we have `on.tags: true` and `on.branch: master` it will only deploy when it's a commit on master that has a tag already.

scripts/travis/build-windows.sh
grounded042: I'm confused by the two comments - should I use `make build-windows` or call the script?

jpkrohling: I would say that the script is desirable, for consistency's sake. But @yurishkuro has the final word on this.

yurishkuro: Personally I prefer minimalist approach, of the script doesn't add anything then why have it? 

But since it's here already, it's fine to leave it.

grounded042:
I pushed up some new commits:
- Added a script to package files into archives with a version number like @yurishkuro asked for above.
- Added a darwin build option for travis
- Added a linux build option for travis - this was needed because we don't build all the linux binaries all at once and would make archiving everything together a little complicated

.travis.yml
```diff
@@ -54,9 +63,34 @@ script:
   - if [ "$DOCKER" == true ]; then bash ./scripts/travis/build-docker-images.sh ; else echo 'skipping docker images'; fi
   - if [ "$ES_INTEGRATION_TEST" == true ]; then bash ./scripts/travis/es-integration-test.sh ; else echo 'skipping elastic search integration test'; fi
   - if [ "$HOTROD" == true ]; then bash ./scripts/travis/hotrod-integration-test.sh ; else echo 'skipping hotrod example'; fi
+  - if [ "$LINUX" == true ]; then make build-linux ; else echo 'skipping linux'; fi
+  - if [ "$WINDOWS" == true ]; then make build-windows ; else echo 'skipping windows'; fi
+  - if [ "$DARWIN" == true ]; then make build-darwin ; else echo 'skipping darwin'; fi
```
yurishkuro: overall this looks good, but this 3 additions concern me as they are going to significantly increase Travis build times. I'd rather have a single DEPLOY=true step that would build files for all targets, which would be a lot faster since we don't have to wait for Travis VM/container to initialize.

Also, there's now a huge amount of duplication in the make file, I think it could be simplified a lot through variables, similar to the deploy.sh file, e.g.
```
linux-binaries: 
    make build-agent build-collector ... GOOS=darwin

build-agent:
    CGO_ENABLED=0 GOOS=$(GOOS) ...
```

yurishkuro: in fact, DEPLOY=true doesn't even need to be a new matrix entry, we can co-locate it with the DOCKER=true entry (since it requires linux binaries anyway).

grounded042: Thanks for the input @yurishkuro! I'll have time closer to the middle of this week to make these changes.

grounded042:
@yurishkuro I've made the changes we talked about above - decreased the duplication a good bit. I left `build-all-in-one-linux` in the makefile since it is used for building the all in one image and I didn't want to leak setting the `GOOS` var outside of make.

Is there any other changes needed on this? I think I covered all the feedback. The one thing that will need to be done before the merge is setting the deploy key for travis to push the artifacts to GitHub - this will need to be done by someone who has access to publish release artifacts.

PR #769, opened 2018-04-10
Log errors from HotROD servers

yurishkuro:

### DIFF
--- a/examples/hotrod/cmd/driver.go
+++ b/examples/hotrod/cmd/driver.go
@@ -32,7 +32,8 @@ var driverCmd = &cobra.Command{
 	Short: "Starts Driver service",
 	Long:  `Starts Driver service.`,
 	RunE: func(cmd *cobra.Command, args []string) error {
-		logger := log.NewFactory(logger.With(zap.String("service", "driver")))
+		zapLogger := logger.With(zap.String("service", "driver"))
+		logger := log.NewFactory(zapLogger)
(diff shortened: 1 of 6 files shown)


examples/hotrod/cmd/driver.go
```diff
@@ -32,15 +32,16 @@ var driverCmd = &cobra.Command{
...
+		logger := log.NewFactory(zapLogger)
 		server := driver.NewServer(
 			net.JoinHostPort(driverOptions.serverInterface, strconv.Itoa(driverOptions.serverPort)),
 			tracing.Init("driver", metricsFactory.Namespace("driver", nil), logger, jAgentHostPort),
 			metricsFactory,
 			logger,
 			jAgentHostPort,
 		)
-		return server.Run()
+		return logError(zapLogger, server.Run())
```
isaachier: Is it worth logging here and returning the same error? Will this lead to duplicate logs up the call stack? Maybe wrapping within another error makes more sense.

yurishkuro: it does lead to dup logs, but that's better than not getting logs at all, which would be the case since the `all` command calls all servers except `frontend` via `go ...`, so returned errors go nowhere.

isaachier: Right. I meant we could improve the code to wrap the errors each time, but this is all sample code so I don't really mind.

PR #770, opened 2018-04-10
Use different "good commit message" reference and inline it

yurishkuro:
For some reason we were pointing to a different blog post from what I thought.


yurishkuro:
cc @pavolloffay @objectiser @jpkrohling @black-adder

PR #772, opened 2018-04-12
Change dbmodel.TagFilter to take span as argument

yurishkuro:
Because sometimes the filtering decision depends not only on the tags, but on the span itself, e.g. debug bit, or the service name.


black-adder:
Chaining becomes a bit funky but I think we can get it to work

black-adder:
Alternative, keep the interface as is but have db.Model span implement it

yurishkuro:
> Alternative, keep the interface as is but have db.Model span implement it

I need to control the implementation of the filter. I cannot change the implementation of the span when doing the filtering. So how would that help? Also, filters are static.

PR #776, opened 2018-04-19
Change metrics naming scheme

jpkrohling:
Tasks pending:

- [x] Use a proper version for `jaeger-lib`
- [x] Agree on keeping/removing `jaeger_collector_spans_recd` (see https://github.com/jaegertracing/jaeger/pull/776#issuecomment-383946237)

### DIFF
--- a/cmd/agent/app/builder.go
+++ b/cmd/agent/app/builder.go
@@ -117,7 +117,13 @@ func (b *Builder) getMetricsFactory() (metrics.Factory, error) {
 	if b.metricsFactory != nil {
 		return b.metricsFactory, nil
 	}
-	return b.Metrics.CreateMetricsFactory("jaeger_agent")
+
+	baseFactory, err := b.Metrics.CreateMetricsFactory("jaeger")
+	if err != nil {
+		return nil, err
+	}
+
+	return baseFactory.Namespace("agent", nil), nil
 }
 
 // CreateAgent creates the Agent
--- a/cmd/collector/main.go
+++ b/cmd/collector/main.go
@@ -90,13 +90,13 @@ func main() {
 			builderOpts := new(builder.CollectorOptions).InitFromViper(v)
 
 			mBldr := new(pMetrics.Builder).InitFromViper(v)
-			metricsFactory, err := mBldr.CreateMetricsFactory("jaeger-collector")
+			baseFactory, err := mBldr.CreateMetricsFactory("jaeger")
 			if err != nil {
 				logger.Fatal("Cannot create metrics factory.", zap.Error(err))
 			}
 
 			storageFactory.InitFromViper(v)
-			if err := storageFactory.Initialize(metricsFactory, logger); err != nil {
+			if err := storageFactory.Initialize(baseFactory, logger); err != nil {
 				logger.Fatal("Failed to init storage factory", zap.Error(err))
 			}
 			spanWriter, err := storageFactory.CreateSpanWriter()
@@ -104,6 +104,7 @@ func main() {
 				logger.Fatal("Failed to create span writer", zap.Error(err))
 			}
 
+			metricsFactory := baseFactory.Namespace("collector", nil)
 			handlerBuilder, err := builder.NewSpanHandlerBuilder(
 				builderOpts,
 				spanWriter,
--- a/cmd/standalone/main.go
+++ b/cmd/standalone/main.go
@@ -96,7 +96,7 @@ func main() {
 			}
 
 			mBldr := new(pMetrics.Builder).InitFromViper(v)
-			metricsFactory, err := mBldr.CreateMetricsFactory("jaeger-standalone")
+			metricsFactory, err := mBldr.CreateMetricsFactory("jaeger")
 			if err != nil {
 				return errors.Wrap(err, "Cannot create metrics factory")
 			}
@@ -166,7 +166,7 @@ func startAgent(
 	logger *zap.Logger,
 	baseFactory metrics.Factory,
 ) {
-	metricsFactory := baseFactory.Namespace("jaeger-agent", nil)
+	metricsFactory := baseFactory.Namespace("agent", nil)
 
 	if len(b.CollectorHostPorts) == 0 {
 		b.CollectorHostPorts = append(b.CollectorHostPorts, fmt.Sprintf("127.0.0.1:%d", cOpts.CollectorPort))
@@ -190,7 +190,7 @@ func startCollector(
 	samplingHandler sampling.Handler,
 	hc *healthcheck.HealthCheck,
 ) {
-	metricsFactory := baseFactory.Namespace("jaeger-collector", nil)
+	metricsFactory := baseFactory.Namespace("collector", nil)
 
 	spanBuilder, err := collector.NewSpanHandlerBuilder(
 		cOpts,
@@ -269,7 +269,7 @@ func startQuery(
 			Param: 1.0,
 		},
 		RPCMetrics: true,
-	}.New("jaeger-query", jaegerClientConfig.Metrics(baseFactory))
+	}.New("jaeger-query", jaegerClientConfig.Metrics(baseFactory.Namespace("client", nil)))
 	if err != nil {
 		logger.Fatal("Failed to initialize tracer", zap.Error(err))
 	}
--- a/pkg/cassandra/metrics/table.go
+++ b/pkg/cassandra/metrics/table.go
@@ -33,7 +33,7 @@ type Table struct {
 // NewTable takes a metrics scope and creates a table metrics struct
 func NewTable(factory metrics.Factory, tableName string) *Table {
 	t := storageMetrics.WriteMetrics{}
-	metrics.Init(&t, factory.Namespace(tableName, nil), nil)
+	metrics.Init(&t, factory.Namespace("", map[string]string{"table": tableName}), nil)
 	return &Table{t}
 }
 
--- a/plugin/storage/cassandra/spanstore/writer.go
+++ b/plugin/storage/cassandra/spanstore/writer.go
@@ -104,18 +104,18 @@ func NewSpanWriter(
 ) *SpanWriter {
 	serviceNamesStorage := NewServiceNamesStorage(session, writeCacheTTL, metricsFactory, logger)
 	operationNamesStorage := NewOperationNamesStorage(session, writeCacheTTL, metricsFactory, logger)
-	tagIndexSkipped := metricsFactory.Counter("tagIndexSkipped", nil)
+	tagIndexSkipped := metricsFactory.Counter("tag_index_skipped", nil)
 	opts := applyOptions(options...)
 	return &SpanWriter{
 		session:              session,
 		serviceNamesWriter:   serviceNamesStorage.Write,
 		operationNamesWriter: operationNamesStorage.Write,
 		writerMetrics: spanWriterMetrics{
-			traces:                casMetrics.NewTable(metricsFactory, "Traces"),
-			tagIndex:              casMetrics.NewTable(metricsFactory, "TagIndex"),
-			serviceNameIndex:      casMetrics.NewTable(metricsFactory, "ServiceNameIndex"),
-			serviceOperationIndex: casMetrics.NewTable(metricsFactory, "ServiceOperationIndex"),
-			durationIndex:         casMetrics.NewTable(metricsFactory, "DurationIndex"),
+			traces:                casMetrics.NewTable(metricsFactory, "traces"),
+			tagIndex:              casMetrics.NewTable(metricsFactory, "tag_index"),
+			serviceNameIndex:      casMetrics.NewTable(metricsFactory, "service_name_index"),
+			serviceOperationIndex: casMetrics.NewTable(metricsFactory, "service_operation_index"),
+			durationIndex:         casMetrics.NewTable(metricsFactory, "duration_index"),
 		},
 		logger:          logger,
 		tagIndexSkipped: tagIndexSkipped,
--- a/cmd/collector/app/metrics.go
+++ b/cmd/collector/app/metrics.go
@@ -60,23 +60,21 @@ type metricsBySvc struct {
 
 // CountsBySpanType measures received, rejected, and receivedByService metrics for a format type
 type CountsBySpanType struct {
-	// Received is the actual number of spans received from upstream
-	Received metrics.Counter
-	// Rejected is the number of spans we rejected (usually due to blacklisting)
-	Rejected metrics.Counter
 	// ReceivedBySvc maintain by-service metrics for a format type
 	ReceivedBySvc metricsBySvc
+	// RejectedBySvc is the number of spans we rejected (usually due to blacklisting) by-service
+	RejectedBySvc metricsBySvc
 }
 
 // NewSpanProcessorMetrics returns a SpanProcessorMetrics
 func NewSpanProcessorMetrics(serviceMetrics metrics.Factory, hostMetrics metrics.Factory, otherFormatTypes []string) *SpanProcessorMetrics {
 	spanCounts := map[string]CountsBySpanType{
-		ZipkinFormatType:  newCountsBySpanType(serviceMetrics.Namespace(ZipkinFormatType, nil)),
-		JaegerFormatType:  newCountsBySpanType(serviceMetrics.Namespace(JaegerFormatType, nil)),
-		UnknownFormatType: newCountsBySpanType(serviceMetrics.Namespace(UnknownFormatType, nil)),
+		ZipkinFormatType:  newCountsBySpanType(serviceMetrics.Namespace("", map[string]string{"format": ZipkinFormatType})),
+		JaegerFormatType:  newCountsBySpanType(serviceMetrics.Namespace("", map[string]string{"format": JaegerFormatType})),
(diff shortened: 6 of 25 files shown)


jpkrohling:
@simonpasquier, would you be able to review the metric names? Please, see the links in my previous comment.

jpkrohling:
Travis [is failing because](https://travis-ci.org/jaegertracing/jaeger/jobs/368573548#L1751) the dependent change is not merged yet:

```
FAIL	github.com/jaegertracing/jaeger/pkg/metrics	0.023s
```

On my local machine, when I symlink jaeger-lib in `vendor` with my local clone, this passes:

```
ok  	github.com/jaegertracing/jaeger/pkg/metrics	1.019s
```

plugin/storage/cassandra/spanstore/writer.go
```diff
@@ -104,18 +104,18 @@ func NewSpanWriter(
...
 		session:              session,
 		serviceNamesWriter:   serviceNamesStorage.Write,
 		operationNamesWriter: operationNamesStorage.Write,
 		writerMetrics: spanWriterMetrics{
-			traces:                casMetrics.NewTable(metricsFactory, "Traces"),
-			tagIndex:              casMetrics.NewTable(metricsFactory, "TagIndex"),
-			serviceNameIndex:      casMetrics.NewTable(metricsFactory, "ServiceNameIndex"),
-			serviceOperationIndex: casMetrics.NewTable(metricsFactory, "ServiceOperationIndex"),
-			durationIndex:         casMetrics.NewTable(metricsFactory, "DurationIndex"),
+			traces:                casMetrics.NewTable(metricsFactory, "traces"),
```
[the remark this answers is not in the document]
jpkrohling: How about we leave this decision to a follow-up issue/PR?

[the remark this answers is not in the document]
jpkrohling: Created #778 for this

plugin/storage/cassandra/spanstore/writer.go
```diff
@@ -104,18 +104,18 @@ func NewSpanWriter(
...
+			tagIndex:              casMetrics.NewTable(metricsFactory, "tag_ndex"),
```
jpkrohling: `s/tag_ndex/tag_index`

Credit goes to @burmanm  :)

plugin/storage/cassandra/spanstore/writer.go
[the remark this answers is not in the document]
yurishkuro: as discussed on the call, I would rather make the change here because this PR is already a breaking change since it renames the metrics, so I would go further and fix them for good by putting table name as a tag instead of appending to the metric name. I commented on #778 how it can be (easily) done.

jpkrohling:
These are the metrics I have that are "by_svc":

```
jaeger_standalone_jaeger_collector_jaeger_spans_by_svc_customer 10
jaeger_standalone_jaeger_collector_jaeger_traces_by_svc_customer 5
jaeger_standalone_jaeger_collector_spans_saved_by_svc_customer 10
jaeger_standalone_jaeger_collector_traces_saved_by_svc_customer 5
```

Just so that I'm clear, you mean to convert the above into:

```
jaeger_standalone_jaeger_collector_jaeger_spans_by_svc{service="customer"} 10
jaeger_standalone_jaeger_collector_jaeger_traces_by_svc{service="customer"} 5
jaeger_standalone_jaeger_collector_spans_saved_by_svc{service="customer"} 10
jaeger_standalone_jaeger_collector_traces_saved_by_svc{service="customer"} 5
``

(Gist updated to include metrics generated on demand when spans are received: https://gist.github.com/jpkrohling/e0823550b66c382bca54138caca824ec#file-metrics-master-inmemory-customerservice-txt)

**Comment by objectiser** [comment]
@jpkrohling Can we reduce the number of times 'jaeger' is used in the metric name to 1 :) e.g.
```
jaeger_standalone_collector_spans_by_svc{service="customer"} 10
```

**Comment by mabn** [comment]
@jpkrohling yes, that's what I meant.

File: cmd/standalone/main.go [hunk 3]
```diff
@@ -91,7 +91,7 @@ func main() {
 			}
 
 			mBldr := new(pMetrics.Builder).InitFromViper(v)
-			metricsFactory, err := mBldr.CreateMetricsFactory("jaeger-standalone")
+			metricsFactory, err := mBldr.CreateMetricsFactory("standalone")
```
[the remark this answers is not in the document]
jpkrohling: Yes, that was my intention but I pushed without testing first :/

jpkrohling:
This PR has been updated to include the changes suggested by @mabn, @objectiser and #778 and is ready for review. The [related PR for jaeger-lib](https://github.com/jaegertracing/jaeger-lib/pull/43) has also been updated.

The [Gist](https://gist.github.com/jpkrohling/e0823550b66c382bca54138caca824ec) has also been updated to show the state of metrics before and after this change when deployed using Cassandra as the backing storage.

objectiser:
Other possible suggestions:

`jaeger_standalone_cassandra...` -> `jaeger_storage_cassandra...`

`jaeger_standalone_collector` - not sure I understand the reason for 'standalone' - can't it just be `jaeger_collector...`

`jaeger_standalone_jaeger...` - these seem to be all client related to possibly `jaeger_client_...` rather than having standalone and jaeger twice.

`jaeger_standalone_agent_` - just `jaeger_agent_`

jpkrohling:
> `jaeger_standalone_collector` - not sure I understand the reason for 'standalone' - can't it just be `jaeger_collector`...

Makes sense, especially to keep the metrics generated by the `collector` binary to be consistent with the ones from the `standalone` binary.

jpkrohling:
Updated based on the last suggestion from @objectiser. There's still a set of metrics with the name starting as `jaeger_jaeger_`, from the Jaeger client. As the client does not provide a way to override its metrics namespace, `jaeger` gets duplicated.

jpkrohling:
I think the latest Gist is already very close to what @yurishkuro is proposing. Here's an overview of the prefixes we have on that Gist. Note that there are four files in that Gist, three with the output from this PR and one with the base from v1.3.

* `jaeger_agent_`
* `jaeger_cassandra`
* `jaeger_client_jaeger_`
* `jaeger_collector_`

I just checked how the metrics are reported when starting the individual components and there were a few mismatches, fixed in the latest commit.

mabn:
One more thing about the duplication @objectiser mentioned. There are two kinds "received" and "saved". For received right now there are:

```
jaeger_collector_jaeger_spans_recd 19
jaeger_collector_jaeger_spans_rejected 0
jaeger_collector_jaeger_spans_by_svc{service="customer"}
jaeger_collector_jaeger_debug_spans_by_svc{service="customer"}
jaeger_collector_jaeger_traces_by_svc{service="query"}
```

The second "jaeger" is the format name. So merge them into just three metrics:
```
jaeger_collector_spans_recd{format="jaeger", service="customer", debug=false}
jaeger_collector_spans_rejected{format="jaeger", service="customer", debug=false}
jaeger_collector_traces_recd{ format="jaeger", service="customer"} # maybe debug tag too?
```

I guess it's a similar story with the corresponding "saved-by-svc" metrics.

jpkrohling:
@mabn: I agree about the format, but not sure about the other suggestion. Here's what we would have by applying only the `format` suggestion:

```
jaeger_collector_spans_recd{format="jaeger"} 36
jaeger_collector_spans_by_svc{format="jaeger",service="customer"} 18
jaeger_collector_spans_by_svc{format="jaeger",service="preference"} 18
```

Your suggestion, as I understand, is to have this:

```
jaeger_collector_spans_recd{format="jaeger",service="customer"} 18
jaeger_collector_spans_recd{format="jaeger",service="preference"} 18
```
(perhaps with one more label, `debug`).

While it wouldn't be hard to come up with a Prometheus query to get all `jaeger_collector_spans_recd{format="jaeger"}`, it would probably be more convenient to leave what we have today.

@simonpasquier: is there a best practice for cases like this?

jpkrohling:
As I will not be able to work on this PR until at least next Monday (Apr 30th) and potentially until May 7th (the week after KubeCon), I just added a couple of tasks to the first comment in this PR, so that we know what else there is to be done here.

@mabn, @yurishkuro, @simonpasquier: feel free to edit (if you can) or request a new item to be added to the list.

simonpasquier:
Personally I've found it less confusing when an application doesn't expose aggregated metrics that I can get from other metrics with a bit of PromQL.

If you only have:
```
jaeger_collector_spans_by_svc{format="jaeger",service="customer"} 18
jaeger_collector_spans_by_svc{format="jaeger",service="preference"} 18
```

Then getting the equivalent of `jaeger_collector_spans_recd{format="jaeger"}` is just `sum(jaeger_collector_spans_by_svc{format="jaeger"})`.

jpkrohling:
PR updated with the changes suggested by @mabn. I applied the same logic to the "rejected" counter (but couldn't simulate this, to get it showing as metric).

The Gist has been updated as well. Here's a short overview:

```
jaeger_collector_spans_by_svc{format="jaeger",service="customer"} 6
jaeger_collector_spans_by_svc{format="jaeger",service="preference"} 6
jaeger_collector_spans_saved_by_svc{service="customer"} 6
jaeger_collector_spans_saved_by_svc{service="preference"} 6
jaeger_collector_traces_by_svc{format="jaeger",service="customer"} 3
jaeger_collector_traces_saved_by_svc{service="customer"} 3
```

jpkrohling:
@mabn, @objectiser, @yurishkuro: is there something else you'd like to see in this PR before it gets merged? The only thing left (see issue description) is to get a proper version for `jaeger-lib`.

pkg/cassandra/metrics/table.go
```diff
@@ -33,7 +33,7 @@ type Table struct {
 // NewTable takes a metrics scope and creates a table metrics struct
 func NewTable(factory metrics.Factory, tableName string) *Table {
 	t := storageMetrics.WriteMetrics{}
-	metrics.Init(&t, factory.Namespace(tableName, nil), nil)
+	metrics.Init(&t, factory.Namespace("", map[string]string{"table": tableName}), nil)
```
black-adder: not sure about this one, this allows users to aggregate metrics across different c* tables. For attempts and inserts (total c* throughput) it makes sense but allowing users to aggregate latencies across different tables doesn't seem right (workload will be different per table). I guess giving the ability to aggregate doesn't imply people will aggregate and technically people using non-tag based metrics systems could already aggregate in the previous rendition. TLDR; I've convinced myself this is fine

jpkrohling: > TLDR; I've convinced myself this is fine

Glad to hear :D

objectiser:
I think the metrics and tags are fine - but prefer the names without "by_svc" - as this implies the metric is per service, whereas that is just one of the tags. Maybe reuse the previous names (the ones without the service tags, e.g. `jaeger_collector_spans_recd`), but now with the tags?

jpkrohling:
> Maybe reuse the previous names (the ones without the service tags, e.g. jaeger_collector_spans_recd), but now with the tags?

+1. Is anyone against it?

yurishkuro:
it's not possible to reuse the same metric name with different set of labels (only different label values)

objectiser:
It was my understanding that there would only now be one use of that metric name. But now with a service tag.

Or are you thinking about queries on the metric server that spans before and after this change?

yurishkuro:
I meant that we probably already have `jaeger_collector_spans_recd` as a cumulative metric, which has no labels.

benley:
I've put together a set of Prometheus metric relabel rules that transform metrics from jaeger 1.3 (probably also 1.4.1, I haven't checked yet) to the new schema proposed in this PR, if this is useful to anybody else:

```yaml
- action: replace
  regex: jaeger_collector:(?:.+:)?[^:]+_by_svc:([^:]+)
  replacement: $1
  source_labels: ["__name__"]
  target_label: service
- action: replace
  regex: (jaeger_collector:(?:.+:)?[^:]+_by_svc):[^:]+
  replacement: $1
  source_labels: ["__name__"]
  target_label: __name__
- action: replace
  regex: jaeger_collector:([^:]+):((?:debug_)?(?:spans|traces)(?:_recd|_rejected|_saved)?(?:_by_svc)?)
  replacement: $1
  source_labels: ["__name__"]
  target_label: format
- action: replace
  regex: jaeger_collector:([^:]+):((?:debug_)?(?:spans|traces)(?:_recd|_rejected|_saved)?(?:_by_svc)?)
  replacement: jaeger_collector:$2
  source_labels: ["__name__"]
  target_label: __name__
- regex: jaeger_collector:([^:]+)
  replacement: jaeger_collector_$1
  source_labels: ["__name__"]
  target_label: __name__
```

jpkrohling:
> I meant that we probably already have jaeger_collector_spans_recd as a cumulative metric, which has no labels.

We had, until the commit titled ["Change rejected to rejected-by-svc, remove recd metrics"](https://github.com/jaegertracing/jaeger/pull/776/commits/4a492524a2158de2805d24ada6816a1ad299e777). This was removed based on the suggestion by @mabn and part of the task list on the PR description for a couple weeks now :)

I just added another commit based on @objectiser's request, but with 'received' instead of 'recd' (as we have also 'rejected')

jpkrohling:
If there are no objections, this is now ready to be merged. The only thing missing is to change the dependency to use a proper version, but it can be done in a follow-up commit.

cmd/collector/main.go
```diff
@@ -90,20 +90,21 @@ func main() {
 			builderOpts := new(builder.CollectorOptions).InitFromViper(v)
 
 			mBldr := new(pMetrics.Builder).InitFromViper(v)
-			metricsFactory, err := mBldr.CreateMetricsFactory("jaeger-collector")
+			baseFactory, err := mBldr.CreateMetricsFactory("jaeger")
```
yurishkuro: why are we removing `-collector` suffix? Wouldn't it create potential clashes when used in standalone? And even if not, once these metrics are scraped by a single Prometheus instance, isn't it better to be able to tell apart which binary emitted the metric?

yurishkuro: why are we removing the `-collector` suffix here? Isn't it better to be able to tell apart which binary emitted the metric? The storage metrics won't be grouped by collector name anymore.

cmd/agent/app/builder.go
```diff
@@ -117,7 +117,7 @@ func (b *Builder) getMetricsFactory() (metrics.Factory, error) {
 	if b.metricsFactory != nil {
 		return b.metricsFactory, nil
 	}
-	return b.Metrics.CreateMetricsFactory("jaeger_agent")
+	return b.Metrics.CreateMetricsFactory("jaeger-agent")
```
yurishkuro: what's the reasoning here? Maybe this should be `b.Metrics.CreateMetricsFactory("jaeger").Namespace("agent",nil)`? That would be more consistent with the way metrics are constructed in the other binaries

cmd/collector/main.go
jpkrohling: > why are we removing the -collector suffix here?

It's just so that the `jaeger` prefix is used for the storage metrics without the component's name. `collector` is added later as a namespace to the collector-specific metrics.

> Isn't it better to be able to tell apart which binary emitted the metric?

I think the agreement was to have per-component metrics, instead of per-binary, so that metrics generated by the collector via `standalone-linux` would be consistent with the metrics generated by collector via `collector-linux`:

(before)
`jaeger_standalone_jaeger_collector_in_queue_latency_bucket{host="caju",le="0.005"} 0`

vs

(after, across all binaries)
`jaeger_collector_in_queue_latency_bucket{host="caju",le="0.005"} 0`

> The storage metrics won't be grouped by collector name anymore.

You mean, `jaeger_cassandra_inserts{table="service_names"}` would be used by `collector`, `query` and `standalone`? This would potentially create three different metric names for the same thing: `jaeger_standalone_cassandra_inserts`, `jaeger_collector_cassandra_inserts`, `jaeger_query_cassandra_inserts`. I personally prefer to have one single metrics, but I do see the point of having one per *component* (not binary). Just not sure how we'd tell the components apart when using the standalone.

objectiser: If knowing the source binary is important, then it could be added as a label instead?

yurishkuro: I am really surprised Prometheus community has no guidelines around this. Can't imagine any minimally complex system not running into the same exact issues with multiple namespaces and hierarchies.

jpkrohling: > I am really surprised Prometheus community has no guidelines around this. Can't imagine any minimally complex system not running into the same exact issues with multiple namespaces and hierarchies.

The Prometheus scraper can add arbitrary labels, so, this kind of information would arguably be added there, along with some metadata not known to the "microservice", like the Geo, DC, Rack, and so on. Our "binary" name could also be added at scrape time, for cases where it has more value than a consistent set of metrics across all binaries.

https://prometheus.io/docs/prometheus/latest/configuration/configuration/#%3Cstatic_config%3E

How about we merge it with the current state, publish a blog post about this change + how to add arbitrary labels, and ask for community feedback? If we hear that having this as part of the metrics we emit, we add it as a label.

yurishkuro: Maybe we should discuss on Friday. We are accumulating a set of breaking changes for 1.5: metrics, docker commands, and Cassandra schema.

Issue #777, opened 2018-04-19
make storage-integration-test fails due to issues in CompareTrace() but error is forgotten


burmanm:
While writing the K/V/Badger storage backend, I enabled the storage-integration-test for that one also. That led me to silly errors, where byte-by-byte equal spans are not actually equal.

It does:

```go
if !assert.EqualValues(t, expected, actual) {
```

Where ``expected`` & ``actual`` are of type ``*model.Trace``. This leads to stretch/testify's doing a call to reflect.DeepEqual() and that fails. I copied over the code from reflect to check where this fails and what I get is:

```go
	case reflect.Struct:
		for i, n := 0, v1.NumField(); i < n; i++ {
			if !deepValueEqual(v1.Field(i), v2.Field(i), visited, depth+1) {
```

This returns false, with the following output:

```go
fmt.Printf("deepValueEqual v1.Field() != v2.Field() inside, %v != %v Struct\n", v1.Field(i), v2.Field(i))
```

Printing out ``deepValueEqual v1.Field() != v2.Field() inside, 0 != 0 Struct``. For more reason, if we do:

```go
fmt.Printf("first %v == second %v: %v\n", v1.Field(i), v2.Field(i), v1.Field(i) == v2.Field(i))
```

We get: ``first 0 == second 0: false``. The reflect.DeepEqual unfortunately hits an unexported field at this point making things difficult, but it does:

```go
	if v.kind() == Interface {
		if v.NumMethod() == 0 {
			return *(*interface{})(v.ptr)
		}
```

For both v1.Field(i) and v2.Field(i) and then compares the result. Thus, we end up with uint64(0) is not equal to uint64(0) and the Trace-comparison fails. This obviously works with the memory backend as it compares the same backing pointers, but these fail if we do json.Unmarshal / Marshal combination in between.

All I could say is that .. this type of assert is not going to work. Or can someone figure out the reason (I'm not this familiar with the golang runtime)? Doing ``fmt.Printf("Found mistake %v != %v\n", *expected, *actual)`` gives two byte-slices, but they are equal so I don't really see any mistake in the real code.

Why this hasn't been caught is simple also, the code in integration_test.go calls:

```go
	if !assert.EqualValues(t, expected, actual) {
		for _, err := range pretty.Diff(expected, actual) {
			t.Log(err)
		}
		out, err := json.Marshal(actual)
		assert.NoError(t, err)
		t.Logf("Actual trace: %s", string(out))
	}
```

[...]

PR #781, opened 2018-04-23
Fail test if the expected and actual traces differ (#777)

burmanm:
Use pretty.Diff(expected, actual) instead of reflect.DeepEqual(expected, actual) to check for equality between given and received trace. Also, if there is a difference, set the test to fail instead of only logging the failed trace.

And prettify the storage-integration-test output

### DIFF
--- a/plugin/storage/integration/domain_trace_compare_test.go
+++ b/plugin/storage/integration/domain_trace_compare_test.go
@@ -50,13 +50,15 @@ func CompareTraces(t *testing.T, expected *model.Trace, actual *model.Trace) {
 	model.SortTrace(expected)
 	model.SortTrace(actual)
 	checkSize(t, expected, actual)
-	if !assert.EqualValues(t, expected, actual) {
-		for _, err := range pretty.Diff(expected, actual) {
-			t.Log(err)
+
+	if diff := pretty.Diff(expected, actual); len(diff) > 0 {
+		for _, d := range diff {
+			t.Logf("Expected and actual differ: %v\n", d)
 		}
 		out, err := json.Marshal(actual)
 		assert.NoError(t, err)
 		t.Logf("Actual trace: %s", string(out))
+		t.Fail()
 	}
 }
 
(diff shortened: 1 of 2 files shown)


plugin/storage/integration/domain_trace_compare_test.go
```diff
@@ -50,13 +50,14 @@ func CompareTraces(t *testing.T, expected *model.Trace, actual *model.Trace) {
 	model.SortTrace(expected)
 	model.SortTrace(actual)
 	checkSize(t, expected, actual)
-	if !assert.EqualValues(t, expected, actual) {
-		for _, err := range pretty.Diff(expected, actual) {
-			t.Log(err)
-		}
+
+	diff := pretty.Diff(expected, actual)
+	if len(diff) > 0 {
```
yurishkuro: nit: combine into one line: `if diff := pretty.Diff(expected, actual); len(diff) > 0 {`

plugin/storage/integration/domain_trace_compare_test.go
```diff
@@ -50,13 +50,14 @@ func CompareTraces(t *testing.T, expected *model.Trace, actual *model.Trace) {
...
+		t.Logf("Expected and actual differ: %v\n", diff)
```
yurishkuro: when there are many differences, this doesn't print them in easily readable way. I prefer the loop as before.

burmanm:
Yep, modified based on the review.

yurishkuro:
looks like the builds are failing due to #786

plugin/storage/integration/domain_trace_compare_test.go
```diff
@@ -50,13 +50,15 @@ func CompareTraces(t *testing.T, expected *model.Trace, actual *model.Trace) {
 	model.SortTrace(expected)
 	model.SortTrace(actual)
 	checkSize(t, expected, actual)
-	if !assert.EqualValues(t, expected, actual) {
-		for _, err := range pretty.Diff(expected, actual) {
-			t.Log(err)
+
+	if diff := pretty.Diff(expected, actual); len(diff) > 0 {
+		for _, d := range diff {
+			t.Logf("Expected and actual differ: %v\n", d)
```
yurishkuro: nit: `\n` is not necessary, but ok

yurishkuro:
thanks @burmanm !

Issue #775, opened 2018-04-16
Add healthcheck to all-in-one


pavolloffay:

eundoosong:
I simply made test code for this issue by using existing healthcheck code.
https://github.com/eundoosong/jaeger/commit/5673f8bd6d2f995cad98e0287b80a84367ada47d
If it is what you intend, it would be great for me to create PR.

PR #784, opened 2018-04-23
Add healthcheck to standalone

eundoosong:
healthcheck is added in the same way as collect and query since it is missed in standalone.
Related issue : https://github.com/jaegertracing/jaeger/issues/775

Signed-off-by: Eundoo Song <eundoo.song@gmail.com>

### DIFF
--- a/cmd/collector/app/builder/builder_flags.go
+++ b/cmd/collector/app/builder/builder_flags.go
@@ -23,13 +23,16 @@ import (
 )
 
 const (
-	collectorQueueSize           = "collector.queue-size"
-	collectorNumWorkers          = "collector.num-workers"
-	collectorWriteCacheTTL       = "collector.write-cache-ttl"
-	collectorPort                = "collector.port"
-	collectorHTTPPort            = "collector.http-port"
-	collectorZipkinHTTPort       = "collector.zipkin.http-port"
-	collectorHealthCheckHTTPPort = "collector.health-check-http-port"
+	collectorQueueSize     = "collector.queue-size"
+	collectorNumWorkers    = "collector.num-workers"
+	collectorWriteCacheTTL = "collector.write-cache-ttl"
+	collectorPort          = "collector.port"
+	collectorHTTPPort      = "collector.http-port"
+	collectorZipkinHTTPort = "collector.zipkin.http-port"
+	// CollectorHealthCheckHTTPPort is the flag name for health check
+	CollectorHealthCheckHTTPPort = "collector.health-check-http-port"
+	// DefaultHealthCheckHTTPPort is the default HTTP Port for health check
+	DefaultHealthCheckHTTPPort = 14269
 )
 
 // CollectorOptions holds configuration for collector
@@ -55,7 +58,7 @@ func AddFlags(flags *flag.FlagSet) {
 	flags.Int(collectorPort, 14267, "The tchannel port for the collector service")
 	flags.Int(collectorHTTPPort, 14268, "The http port for the collector service")
 	flags.Int(collectorZipkinHTTPort, 0, "The http port for the Zipkin collector service e.g. 9411")
-	flags.Int(collectorHealthCheckHTTPPort, 14269, "The http port for the health check service")
+	flags.Int(CollectorHealthCheckHTTPPort, DefaultHealthCheckHTTPPort, "The http port for the health check service")
 }
 
 // InitFromViper initializes CollectorOptions with properties from viper
@@ -65,6 +68,6 @@ func (cOpts *CollectorOptions) InitFromViper(v *viper.Viper) *CollectorOptions {
 	cOpts.CollectorPort = v.GetInt(collectorPort)
 	cOpts.CollectorHTTPPort = v.GetInt(collectorHTTPPort)
 	cOpts.CollectorZipkinHTTPPort = v.GetInt(collectorZipkinHTTPort)
-	cOpts.CollectorHealthCheckHTTPPort = v.GetInt(collectorHealthCheckHTTPPort)
+	cOpts.CollectorHealthCheckHTTPPort = v.GetInt(CollectorHealthCheckHTTPPort)
 	return cOpts
 }
--- a/cmd/collector/app/builder/builder_flags.go
+++ b/cmd/collector/app/builder/builder_flags.go
@@ -29,10 +29,8 @@ const (
 	collectorPort          = "collector.port"
 	collectorHTTPPort      = "collector.http-port"
 	collectorZipkinHTTPort = "collector.zipkin.http-port"
-	// CollectorHealthCheckHTTPPort is the flag name for health check
-	CollectorHealthCheckHTTPPort = "collector.health-check-http-port"
-	// DefaultHealthCheckHTTPPort is the default HTTP Port for health check
-	DefaultHealthCheckHTTPPort = 14269
+	// CollectorDefaultHealthCheckHTTPPort is the default HTTP Port for health check
+	CollectorDefaultHealthCheckHTTPPort = 14269
 )
 
 // CollectorOptions holds configuration for collector
@@ -47,8 +45,6 @@ type CollectorOptions struct {
 	CollectorHTTPPort int
 	// CollectorZipkinHTTPPort is the port that the Zipkin collector service listens in on for http requests
 	CollectorZipkinHTTPPort int
-	// CollectorHealthCheckHTTPPort is the port that the health check service listens in on for http requests
-	CollectorHealthCheckHTTPPort int
 }
 
 // AddFlags adds flags for CollectorOptions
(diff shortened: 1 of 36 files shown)


yurishkuro:
So this is actually adding two independent health check endpoints, for collector and query.

Could you describe in the PR why this is needed, how it can be used? Our objective is to only add features that are actually useful to people (and documented!).

Thanks.

eundoosong:
I think it would be better for people who want to use health check in the same way as separated one.
Because it is more consistent rather than providing another end point.
Plus, help message already shows each healthcheck flags of collector and query.
Documentation should be as different PR in documentation repo.

Thanks.

yurishkuro:
> I think it would be better for people who want to use health check in the same way as separated one.

But how would you use it? E.g. our internal deployment system and discovery system (part of routing) only support one health endpoint per service.

I am not suggesting a separate endpoint (we serve metrics from one port only, healthcheck would be on the same port), but I think there should be only one health status for the binary, not two that are potentially conflicting.

jpkrohling:
> I think there should be only one health status for the binary, not two that are potentially conflicting.

+1, from an operator's perspective, it makes more sense to have one endpoint that reports the health of the system in general than one health check endpoint for each component.

eundoosong:
> we serve metrics from one port only, healthcheck would be on the same port

I don'get this, does this mean healthcheck is using the same port as metrics?

I agree that one endpoint of the health check per service(binary) makes more sense.
Thanks for pointing out.
So, if you agree with adding one healthcheck in standalone, the following should be considered.
- Add a healthcheck for standalone. (with new port number)
- Add a flag like "standalone.health-check-http-port"
- Hide existing collector(--collector.health-check-http-port) and query(--query.health-check-http-port) flags 

I think the implementation will not be many changes, but I'm wondering which port number can be used.

yurishkuro:
You don't need a new port, all-in-one already opens several HTTP ports, you can just add an endpoint to one of them. E.g. we define ports for the `/metrics` endpoint here https://www.jaegertracing.io/docs/monitoring/. In all-in-one the `/metrics` endpoint is served from query service API port (https://github.com/jaegertracing/jaeger/pull/784/files#diff-053312057ab6b9282b572674adc5b7f6R298), so it makes the  most sense to me to add the `/health` endpoint to the same port.

eundoosong:
Your suggestion looks good. But since health check is using its own HTTP server(with ServeMux) and the other service(collector, query) is using Gorilla Mux, it's a bit difficult for health check to use without change in health check. I did implement (https://github.com/eundoosong/jaeger/commit/5d16806e0f7cf6f1a08b6c3a2b33c18de7f2db26) but IMO it doesn't look better in term of design.

So I decided collector port for health check(port num is 14269) is used for standalone health check port.
Please review.

cmd/standalone/main.go
```diff
@@ -144,14 +159,28 @@ func main() {
 		queryApp.AddFlags,
 		pMetrics.AddFlags,
 		strategyStoreFactory.AddFlags,
+		func(flagSet *flag.FlagSet) {
+			flagSet.Int(healthCheckHTTPPort, defaultHealthCheckPort,
+				"The http port for the health check services")
+		},
```
eundoosong: Lambda function is used because there is only one flag existed in standalone.

cmd/standalone/main.go
```diff
@@ -144,14 +159,28 @@ func main() {
...
 	)
+	hideQueryCollectorFlags(command)
 
 	if err := command.Execute(); err != nil {
 		fmt.Println(err.Error())
 		os.Exit(1)
 	}
 }
 
+func hideQueryCollectorFlags(command *cobra.Command) {
```
eundoosong: This hides query and collector health check flags.

cmd/collector/app/builder/builder_flags.go
```diff
@@ -23,13 +23,16 @@ import (
...
-	collectorZipkinHTTPort       = "collector.zipkin.http-port"
-	collectorHealthCheckHTTPPort = "collector.health-check-http-port"
+	collectorQueueSize     = "collector.queue-size"
+	collectorNumWorkers    = "collector.num-workers"
+	collectorWriteCacheTTL = "collector.write-cache-ttl"
+	collectorPort          = "collector.port"
+	collectorHTTPPort      = "collector.http-port"
+	collectorZipkinHTTPort = "collector.zipkin.http-port"
+	// CollectorHealthCheckHTTPPort is the flag name for health check
+	CollectorHealthCheckHTTPPort = "collector.health-check-http-port"
```
yurishkuro: I think a better approach would be to have a single global flag for all binaries, similar to how we do log levels or metrics.

eundoosong: Questions,
- Does this mean we need to take that flag out from collector flags and make a global flag?
>  similar to how we do log levels or metrics.
- Are you talking about BasicOptions(cmd/builder/builder_options.go)? How about SharedFlags(cmd/flags/..)?

yurishkuro: 1 - yes
2 - I meant shared flags, the BasicOption is independent of cli flags

cmd/flags/flags.go
```diff
@@ -53,21 +55,29 @@ func TryLoadConfigFile(v *viper.Viper) error {
...
 
+type healthCheck struct {
+	Port int
+}
+
 // AddFlags adds flags for SharedFlags
 func AddFlags(flagSet *flag.FlagSet) {
 	flagSet.String(spanStorageType, "", fmt.Sprintf("Deprecated; please use %s environment variable", storage.SpanStorageTypeEnvVar))
 	flagSet.String(logLevel, "info", "Minimal allowed log Level. For more levels see https://github.com/uber-go/zap")
+	flagSet.Int(healthCheckHTTPPort, 0, "The http port for the health check service")
```
eundoosong: The default port is removed since it's difficult to pass in the current interface. So I let NewHealthCheck set it when each component initializes health check.
Do we need to add a default port in help message again? Maybe like below..
```go
flagSet.Int(healthCheckHTTPPort, 0, "The http port for the health check service. For default port, see https://https://www.jaegertracing.io/XXXXX")
```

cmd/standalone/main.go
```diff
@@ -119,8 +124,9 @@ func main() {
 			qOpts := new(queryApp.QueryOptions).InitFromViper(v)
 
 			startAgent(aOpts, cOpts, logger, metricsFactory)
-			startCollector(cOpts, spanWriter, logger, metricsFactory, samplingHandler)
-			startQuery(qOpts, spanReader, dependencyReader, logger, metricsFactory, mBldr)
+			startCollector(cOpts, spanWriter, logger, metricsFactory, samplingHandler, hc)
+			startQuery(qOpts, spanReader, dependencyReader, logger, metricsFactory, mBldr, hc)
+			hc.Ready()
```
yurishkuro: I don't necessarily know how to fix it, but there is a race condition here where collector or query background goroutines can fail to start the server and set HC to Unavailable, but this statement will override it.

cmd/standalone/main.go
```diff
@@ -119,8 +124,13 @@ func main() {
...
 			startAgent(aOpts, cOpts, logger, metricsFactory)
-			startCollector(cOpts, spanWriter, logger, metricsFactory, samplingHandler)
-			startQuery(qOpts, spanReader, dependencyReader, logger, metricsFactory, mBldr)
+			collectorServer := startCollector(cOpts, spanWriter, logger, metricsFactory, samplingHandler, hc)
+			queryServer := startQuery(qOpts, spanReader, dependencyReader, logger, metricsFactory, mBldr, hc)
+			logger.Info("Starting jaeger-collector HTTP server", zap.Int("http-port", cOpts.CollectorHTTPPort))
+			logger.Info("Starting jaeger-query HTTP server", zap.Int("port", qOpts.Port))
+			go collectorServer()
+			go queryServer()
+			hc.Ready()
```
eundoosong: Move goroutines to main from each start function, and start them in main then set Ready.
IMO, this will prevent the race issue. 

For 100% guarantees, can we move hc.Ready before goroutines like this? 
```go
hc.Ready()
go collectorServer()
go queryServer()
```

yurishkuro: I don't think returning functions like this provides any additional guarantees. I think it's fine to leave it as it was, i.e. call `startCollector()`, it might fail mid-way and crash the program, or it can spawn goroutine for ListenAndServe, which can fail as well and crush the program later. So there will be a possible time interval where HC==Ready, and then a crash, but I don't think it's critical for all-in-one.

Sorry I led you on the wrong path with the last comment. The fact that the goroutines call log.Fatal on error sort of take care of things.

eundoosong: I wanted hc.Ready() to be called before collector/query goroutine call hc.Set(healthcheck.Unavailable) in ListenAndServe error.
I think this would prevent overriding Unavailable by hc.Ready() as you mentioned above coment.

Anyway, if the issue in all-in-one is not critical and fatal log on error is enough, I'll revert.

yurishkuro: Yes, I don't think it's critical sice even if the override happens the process will exit anyway.

eundoosong:
Reverted.  
Ready to be merged, or is there anything else missing?

PR #792, opened 2018-04-27
Upgrade to  go 1.10

vprithvi:
- remove redundant newline in Fprintln arg list to fix build

### DIFF
--- a/cmd/env/command.go
+++ b/cmd/env/command.go
@@ -40,8 +40,7 @@ their names to upper case and replacing punctuation with underscores. For exampl
       --cassandra.connections-per-host    CASSANDRA_CONNECTIONS_PER_HOST
       --metrics-backend                   METRICS_BACKEND
 
-The following configuration options are only available via environment variables:
-`)
+The following configuration options are only available via environment variables:`+"\n")
 			fs := new(pflag.FlagSet)
 			fs.String(storage.SpanStorageTypeEnvVar, "cassandra", "The type of backend (cassandra, elasticsearch, memory) used for trace storage.")
 			fs.String(storage.DependencyStorageTypeEnvVar, "${SPAN_STORAGE}", "The type of backend used for service dependencies storage.")
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,24 +7,24 @@ dist: trusty
 
 matrix:
   include:
-  - go: 1.9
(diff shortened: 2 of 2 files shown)


cmd/env/command.go
```diff
@@ -40,8 +40,7 @@ their names to upper case and replacing punctuation with underscores. For exampl
       --cassandra.connections-per-host    CASSANDRA_CONNECTIONS_PER_HOST
       --metrics-backend                   METRICS_BACKEND
 
-The following configuration options are only available via environment variables:
-`)
+The following configuration options are only available via environment variables:`)
```
yurishkuro: why? The formatting is better with a blank line.
```
$ go run ./cmd/query/main.go env

their names to upper case and replacing punctuation with underscores. For example,

      SPAN_STORAGE_TYPE string         The type of backend (cassandra, elasticsearch, memory) used for trace storage. (default "cassandra")
```

yurishkuro: then I would add `+ "\n"` before paren. Not having new-line makes display less readable

cmd/env/command.go
```diff
@@ -40,8 +40,7 @@ their names to upper case and replacing punctuation with underscores. For exampl
       --cassandra.connections-per-host    CASSANDRA_CONNECTIONS_PER_HOST
       --metrics-backend                   METRICS_BACKEND
 
-The following configuration options are only available via environment variables:
-`)
+The following configuration options are only available via environment variables:\n`)
```
yurishkuro: are you sure `\n` is actually interpreted when in back-tick string?

vprithvi: Good question, I don't expect it to be, but let me verify/ replace with something that preserves formatting.

yurishkuro: just run the cmd and see the output: `go run ./cmd/query/main.go env`

eundoosong: \n will be interpreted as the literal backslash and character n since it is raw string literal. 
https://golang.org/ref/spec#String_literals
I would suggest inserting fmt.Fprint(cmd.OutOrStdout(), "\n") after Fprintln, if it has to be changed.

yurishkuro: Why the complication? My original suggestion was to concatenate with +"\n"

Issue #790, opened 2018-04-26
Remove SASI index on dependency column family


vprithvi:
While most SASI indexes were removed as part of #80, the one in the dependencies column family still exists, leading to problems when using older versions of cassandra that don't support SASI indexes, or using [alternate storage like ScyllaDB ](https://github.com/jaegertracing/jaeger/pull/201#discussion_r121197772).

We can update the `dependency` schema to not have SASI indexes, and provide a migration script from the old schema to the new schema. 

We should also ensure that https://github.com/jaegertracing/spark-dependencies works with the new schema.

vprithvi:
> I suggest adding a "source" column to the schema, potentially to represent data coming from different sources

I like this idea, but I don't think it should be part of the migration,  I created https://github.com/jaegertracing/jaeger/issues/791 to capture this.

yurishkuro:
I am not suggesting we implement all of the relevant business logic, but if we are already making a breaking schema change, why not include an extra field?

vprithvi:
>but if we are already making a breaking schema change, why not include an extra field?

Because it's unrelated to this change, and is unusable without the business logic. Why shouldn't the change be done along with the business logic?

vprithvi:
I'm thinking that changing the data model to include a date bucket while making the time stamp as a clustering key would enable us to maintain the current query patterns while removing the SASI index.

The schema looks something like this:
https://github.com/jaegertracing/jaeger/blob/d52969b253ec0d8c3d750071c7ccdc5525de3e3a/plugin/storage/cassandra/schema/v001.cql.tmpl#L191-L196

While the write path is largely unaffected, reads become a bit more involved, as we need to compute buckets that we want to retrieve dependencies from. 

We also need to update the spark dependencies job.

The migration path seems to be the following:

1. Stop dependencies job
2. Use the Cassandra [`COPY`](https://docs.datastax.com/en/cql/3.3/cql/cql_reference/cqlshCopy.html) command to export to a CSV file (which works for tables with less than 2 million rows)
3. Delete existing dependencies column family
4. Create dependencies column family with new schema
5. Massage the CSV file into the new format and load into new schema
6. Run new version of dependencies job which writes to new schema

@yurishkuro @black-adder @jpkrohling @pavolloffay wdyt?

PR #793, opened 2018-04-27
[cassandra] Remove SASI index from dependencies table

vprithvi:
- remove SASI indexes so that users may use older Cassandra versions or ScyllaDB
- provide a migration script for current users

 ref #790  

### DIFF
--- a/plugin/storage/cassandra/dependencystore/storage.go
+++ b/plugin/storage/cassandra/dependencystore/storage.go
@@ -15,6 +15,8 @@
 package dependencystore
 
 import (
+	"fmt"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
@@ -27,8 +29,10 @@ import (
 )
 
 const (
-	depsInsertStmt = "INSERT INTO dependencies(ts, ts_index, dependencies) VALUES (?, ?, ?)"
-	depsSelectStmt = "SELECT ts, dependencies FROM dependencies WHERE ts_index >= ? AND ts_index < ?"
+	depsInsertStmt = "INSERT INTO dependencies(ts, date_bucket, dependencies) VALUES (?, ?, ?)"
+	depsSelectFmt  = "SELECT ts, dependencies FROM dependencies WHERE date_bucket IN (%s) AND ts >= ? AND ts < ?"
+	dateFmt        = "20060102"
+	day            = 24 * time.Hour
 )
 
 // DependencyStore handles all queries and insertions to Cassandra dependencies
@@ -64,13 +68,14 @@ func (s *DependencyStore) WriteDependencies(ts time.Time, dependencies []model.D
 			CallCount: int64(d.CallCount),
 		}
 	}
-	query := s.session.Query(depsInsertStmt, ts, ts, deps)
+	query := s.session.Query(depsInsertStmt, ts, ts.Format(dateFmt), deps)
 	return s.dependenciesTableMetrics.Exec(query, s.logger)
 }
 
 // GetDependencies returns all interservice dependencies
 func (s *DependencyStore) GetDependencies(endTs time.Time, lookback time.Duration) ([]model.DependencyLink, error) {
-	query := s.session.Query(depsSelectStmt, endTs.Add(-1*lookback), endTs)
+	startTs := endTs.Add(-1 * lookback)
+	query := s.session.Query(getDepSelectString(startTs, endTs), startTs, endTs)
 	iter := query.Consistency(cassandra.One).Iter()
 
 	var mDependency []model.DependencyLink
@@ -91,6 +96,14 @@ func (s *DependencyStore) GetDependencies(endTs time.Time, lookback time.Duratio
 		return nil, errors.Wrap(err, "Error reading dependencies from storage")
 	}
 	return mDependency, nil
+}
+
+func getDepSelectString(startTs time.Time, endTs time.Time) string {
+	var dateBuckets []string
+	for ts := startTs.Truncate(day); ts.Before(endTs); ts = ts.Add(day) {
+		dateBuckets = append(dateBuckets, ts.Format(dateFmt))
+	}
+	return fmt.Sprintf(depsSelectFmt, strings.Join(dateBuckets, ","))
 }
 
 func (s *DependencyStore) timeIntervalToPoints(endTs time.Time, lookback time.Duration) []time.Time {
--- a/plugin/storage/cassandra/schema/migration/v001tov002.sh
+++ b/plugin/storage/cassandra/schema/migration/v001tov002.sh
@@ -1,0 +1,78 @@
+#!/usr/bin/env bash
+
+function usage {
+    >&2 echo "Error: $1"
+    >&2 echo ""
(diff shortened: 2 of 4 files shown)


plugin/storage/cassandra/schema/v001.cql.tmpl
```diff
@@ -188,14 +188,13 @@ CREATE TYPE IF NOT EXISTS ${keyspace}.dependency (
...
     dependencies list<frozen<dependency>>,
-    PRIMARY KEY (ts)
-)
-    WITH compaction = {
+    PRIMARY KEY (bucket, ts)
+) WITH CLUSTERING ORDER BY (ts DESC)
+    AND compaction = {
         'min_threshold': '4',
         'max_threshold': '32',
         'class': 'org.apache.cassandra.db.compaction.SizeTieredCompactionStrategy'
```
yurishkuro: actually delete SASI below

plugin/storage/cassandra/schema/v001.cql.tmpl
```diff
@@ -188,14 +188,13 @@ CREATE TYPE IF NOT EXISTS ${keyspace}.dependency (
 );
 
 -- compaction strategy is intentionally different as compared to other tables due to the size of dependencies data
--- note we have to write ts twice (once as ts_index). This is because we cannot make a SASI index on the primary key
 CREATE TABLE IF NOT EXISTS ${keyspace}.dependencies (
-    ts          timestamp,
-    ts_index    timestamp,
+    ts           timestamp,
+    date_bucket  text
```
yurishkuro: shouldn't this be `int`?

yurishkuro: it's customary to list PK fields at the top

plugin/storage/cassandra/schema/v001.cql.tmpl
```diff
@@ -188,14 +188,13 @@ CREATE TYPE IF NOT EXISTS ${keyspace}.dependency (
...
     dependencies list<frozen<dependency>>,
-    PRIMARY KEY (ts)
-)
-    WITH compaction = {
+    PRIMARY KEY (bucket, ts)
```
yurishkuro: s/bucket/data_bucket

plugin/storage/cassandra/schema/v001.cql.tmpl
vprithvi: I initially had buckets in a different format, this can be `int`, or `bigint` if we anticipate smaller buckets in the future.

plugin/storage/cassandra/schema/migration/v001tov002.sh
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env bash
+
+function usage {
+    >&2 echo "Error: $1"
```
isaachier: Any reason not to use a heredoc?

plugin/storage/cassandra/dependencystore/storage.go
```diff
@@ -93,6 +98,14 @@ func (s *DependencyStore) GetDependencies(endTs time.Time, lookback time.Duratio
 	return mDependency, nil
 }
 
+func getDepSelectString(startTs time.Time, endTs time.Time) string {
```
isaachier: Cassandra has no date range support?

plugin/storage/cassandra/schema/migration/v001tov002.sh
```diff
@@ -0,0 +1,78 @@
...
+    >&2 echo "The following parameters can be set via environment:"
+    >&2 echo "  KEYSPACE           - keyspace "
+    >&2 echo ""
+    exit 1
+}
+
+confirm() {
+    read -r -p "${1:-Are you sure? [y/N]} " response
+    case "$response" in
+        [yY][eE][sS]|[yY])
```
isaachier: You can probably just do `if [[ "$(echo $response | tr 'a-z' 'A-Z')" =~ "Y(ES)?"]]`.

black-adder:
Changes look fine, what's our strategy for rolling this out? We going to wait until a major release? If not, we probably want this to be backwards compatible.

vprithvi:
>Changes look fine, what's our strategy for rolling this out? We going to wait until a major release? If not, we probably want this to be backwards compatible.

The strategy for rolling out is for people to stop the spark dependency job, which stops writes to the dependency table, run the migration script, and start a new version of the spark job. (I'm writing this up in a readme)

plugin/storage/cassandra/schema/migration/v001tov002.sh
vprithvi: I don't think that would work

plugin/storage/cassandra/schema/migration/v001tov002.sh
vprithvi: Good question, let me try it out

plugin/storage/cassandra/dependencystore/storage.go
vprithvi: Cassandra does not support range queries on partition keys

plugin/storage/cassandra/schema/migration/v001tov002.sh
isaachier: I'm not saying it's necessarily easier, but why wouldn't it work?

plugin/storage/cassandra/dependencystore/storage.go
isaachier: OK that explains it.

plugin/storage/cassandra/schema/migration/v001tov002.sh
vprithvi: I don't know, I didn't spend time debugging it
```
bash-3.2$ response=Y && if [[ "$(echo $response | tr 'a-z' 'A-Z')" =~ "Y(ES)?" ]]; then echo "passed"; fi
bash-3.2$
```

isaachier: Right. I think bash regex might just be awful.

vprithvi:
I had an offline discussion with @black-adder and decided that it makes sense to add a new Keyspace to accomplish some backward compatibility. 

The new dependencies table will live in the new keyspace, so that jaeger-query instances that are not updated can still access old dependencies at the older location.

yurishkuro:
I think I'd need more info. The main concern is data compatibility, not running outdated query - upgrading components is the easy part.

vprithvi:
>I think I'd need more info.

The idea is the we will create a new Keyspace as follows:
`jaeger_analytics_$dc_$version`

This keyspace contains the new dependency table. 

The migration path looks like the following:
1. Run a script to create the new keyspace and schema
2. Run the updated spark-dependencies job (which writes to the new schema)
3. Run a data migration script to transfer old dependencies from the old table to the new table
4. Update jaeger-query
5. Kill the old spark-dependencies job
6. Delete the old dependencies table in the old keyspace

The main benefit offered here is that dependencies will be available throughout the update/migration process because the old spark-dependencies job still writes in the old data format.

vprithvi:
Superseded by #1328

PR #796, opened 2018-05-01
Add metric for debug traces

black-adder:
Currently, we support metrics for the total number of debug spans but another useful metric to have is the debug traces (services that initiate the debug trace). Since debug traces aren't downsampled (this isn't technically true here in OSS...) having metrics to observe which service is creating more debug traces than necessary is useful.

### DIFF
--- a/cmd/collector/app/metrics.go
+++ b/cmd/collector/app/metrics.go
@@ -24,7 +24,7 @@ import (
 
 const (
 	maxServiceNames = 2000
-	otherServices = "other-services"
+	otherServices   = "other-services"
 )
 
 // SpanProcessorMetrics contains all the necessary metrics for the SpanProcessor
@@ -48,10 +48,10 @@ type SpanProcessorMetrics struct { //TODO - initialize metrics in the traditiona
 }
 
 type countsBySvc struct {
-	counts      map[string]metrics.Counter // counters per service
-	debugCounts map[string]metrics.Counter // debug counters per service
-	factory     metrics.Factory
-	lock        *sync.Mutex
+	counts          map[string]metrics.Counter // counters per service
+	debugCounts     map[string]metrics.Counter // debug counters per service
+	factory         metrics.Factory
+	lock            *sync.Mutex
 	maxServiceNames int
 }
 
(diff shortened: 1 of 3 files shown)


cmd/collector/app/metrics.go
```diff
@@ -96,19 +100,16 @@ func NewSpanProcessorMetrics(serviceMetrics metrics.Factory, hostMetrics metrics
...
 		spans: countsBySvc{
-			counts:  make(map[string]metrics.Counter),
-			factory: factory.Namespace("spans."+category, nil),
-			lock:    &sync.Mutex{},
-		},
-		debugSpans: countsBySvc{
-			counts:  make(map[string]metrics.Counter),
-			factory: factory.Namespace("debug-spans."+category, nil),
-			lock:    &sync.Mutex{},
+			counts:      make(map[string]metrics.Counter),
```
black-adder: I refactored this slightly. Rather than having debugSpans and debugTraces as separate metrics, they should be part of the spans and traces metrics but with a debug tag. This way it's easier to aggregate. I guess technically this is a breaking change?

While not in scope of this PR, we could break up the taggify the metrics further by splitting out service as a tag.

yurishkuro:
change looks good, but I am concerned how this would overlap with #776

black-adder:
he who land first, wins

cmd/collector/app/metrics.go
```diff
@@ -172,16 +161,28 @@ func (m metricsBySvc) countTracesByServiceName(serviceName string) {
...
+		counter = getOrCreateCounter(m.debugCounts, func() metrics.Counter {
+			return m.factory.Counter(serviceName, map[string]string{"debug": "true"})
+		})
+	} else {
+		counter = getOrCreateCounter(m.counts, func() metrics.Counter {
+			return m.factory.Counter(serviceName, map[string]string{"debug": "false"})
+		})
 	}
 	m.lock.Unlock()
 	if counter != nil {
```
yurishkuro: we should have a fallback "all other services" name when over limit.

black-adder:
I'll address changes and keep this PR open, I'll coordinate with @jpkrohling so that these two changes can land at the same time.

black-adder:
Ill redo this change on top of his branch and we can land it there

cmd/collector/app/metrics.go
```diff
@@ -169,19 +166,30 @@ func (m metricsBySvc) countTracesByServiceName(serviceName string) {
...
-		m.counts[serviceName] = c
-		counter = c
+	if isDebug {
+		counter = getOrCreateCounter(m.debugCounts, func() metrics.Counter {
+			return m.factory.Counter("", map[string]string{"service": serviceName, "debug": "true"})
+		})
+	} else {
+		counter = getOrCreateCounter(m.counts, func() metrics.Counter {
+			return m.factory.Counter("", map[string]string{"service": serviceName, "debug": "false"})
+		})
```
yurishkuro: this is highly in efficient. The overall function is "count", one would expect it to be quick and zero-allocations, but you are creating all these lambdas that will require memory allocations.

black-adder: went from `1067 ns/op	     231 B/op	      11 allocs/op` to `947 ns/op	     165 B/op	       9 allocs/op`

cmd/collector/app/metrics.go
```diff
@@ -169,19 +166,27 @@ func (m metricsBySvc) countTracesByServiceName(serviceName string) {
...
+func (m *countsBySvc) countByServiceName(serviceName string, isDebug bool) {
 	serviceName = NormalizeServiceName(serviceName)
+	if isDebug {
+		m.x(serviceName, "true", m.debugCounts)
+	} else {
+		m.x(serviceName, "false", m.counts)
+	}
+}
+
+func (m *countsBySvc) x(serviceName, debugStr string, counts map[string]metrics.Counter) {
```
yurishkuro: why do you need an extra function with such a great name? just set two local vars above and continue in the same function.

black-adder: duuude

cmd/collector/app/metrics.go
yurishkuro: maybe if you benchmark the unhappy path, but try with the same service name in a hot loop

black-adder:
https://coveralls.io/jobs/36863391 <- 100% coverage. I'll investigate why overalls is being a pain but landing first.

Issue #609, opened 2017-12-21
Package static assets inside the query-service binary


yurishkuro:
Currently we copy static assets directory to the docker image separately and point the query service to it using cli switch.

Some people swear by https://github.com/jteeuwen/go-bindata or its extension https://github.com/elazarl/go-bindata-assetfs.

This will be especially useful if we support non-Docker packaging (#426).

pavolloffay:
Here is a nice list of alternatives https://github.com/shurcooL/vfsgen#alternatives

I plan to use one for hotrod to fix https://github.com/jaegertracing/jaeger/pull/694#issuecomment-366755805

eundoosong:
This feature looks interesting and I found hotrod already merged with go-bindata-assetfs (https://github.com/jaegertracing/jaeger/pull/702)
I'll have a look for contribution. (would be good start of contribution as a begginer in jaeger backend :) )

PR #815, opened 2018-05-13
Update Dockerfiles to reference executable via ENTRYPOINT

### LINKED ISSUE #802: Allow running containers without overriding command line
Currently Dockerfiles are using a mix of ENTRYPOINT and CMD. If we change the ENTRYPOINT to always point to the executable, and CMD to point to the default arguments (like `all` for hotrod image), then the containers can be run as executables with additional parameters overriding the CMD, but not overriding ENTRYPOINT.

E.g. hotrod's Dockerfile should have
```
ENTRYPOINT ["/go/bin/hotrod-linux"]
CMD ["all"]
```

while all other binaries should have just the ENTRYPOINT pointing to the binary.

The main benefit is that when users run the image and pass additional arguments, they don't need to know the name of the binary inside the image.

zdicesare:
In all Dockerfiles, reference the executable in ENTRYPOINT to allow
consumers to override default parameters without needing to specify
the executable path. Standard defaults move to ENTRYPOINT too for the same reason.

Signed-off-by: Zachary DiCesare <zachary.v.dicesare@gmail.com>

-  Resolves #802 

- In all Dockerfiles, reference the executable path and standard default arguments in ENTRYPOINT 
to allow consumers to override defaults without needing to specify them again.

- Move hotrod default argument to CMD


yurishkuro:
build shows errors like this
```
jaeger-collector_1      | Error: unknown command "/go/bin/collector-linux" for "jaeger-collector"
```

zdicesare:
Hey, sorry about that, I'm still looking into the cause.

zdicesare:
This is ready for review. I'll add a PR to the documentation repository as well

cmd/query/Dockerfile
```diff
@@ -8,4 +8,5 @@ COPY --from=certs /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certifica
 EXPOSE 16686
 COPY query-linux /go/bin/
 ADD jaeger-ui-build /go/jaeger-ui/
-CMD ["/go/bin/query-linux", "--query.static-files=/go/jaeger-ui/"]
+ENTRYPOINT ["/go/bin/query-linux"]
+CMD ["--query.static-files=/go/jaeger-ui/"]
```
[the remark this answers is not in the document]
vprithvi: To answer my own question: It seems that we need to use the `--entrypoint` option to override the entry point while anything after the image in the docker command overrides `cmd`

vprithvi:
lgtm, will merge when the documentation PR is up

cmd/query/Dockerfile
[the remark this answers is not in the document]
vprithvi: I was discussing with @yurishkuro about this, and agree with him in that moving "--query.static-files=/go/jaeger-ui" as an argument to the `ENTRYPOINT` command is better. 

This is because overriding `CMD` in `docker run` overrides all the command parameters forcing users to pass in  "--query.static-files=/go/jaeger-ui/". 

By putting this in the `ENTRYPOINT`,  other overrides become easier, and people may still override the static file location by using the `-d` option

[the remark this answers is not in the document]
zdicesare: That sounds reasonable to me, however in the case that a user did override `query.static-files`, the resulting command would look like:

`/go/bin/query-linux --query.static-files=/go/jaeger-ui/ --query.static-files=/my/desired/path`

which seems a little weird at first glance, but it works because `flags` sets the last value it encounters. Are we ok relying on this?

Also, it seems like the same change would make sense for the defaults in `cmd/standalone/Dockerfile`.

I'm very new to the Jaeger repo, so thank you for the patience here!

zdicesare:
Updated this PR w/ feedback, but want to make sure we're ok with the new usage

Documentation PR at https://github.com/jaegertracing/documentation/pull/74

yurishkuro:
`/go/bin/query-linux --query.static-files=/go/jaeger-ui/ --query.static-files=/my/desired/path`

That's just how it would look internally, right? Don't think it matters as long as it works. Eventually we should fix #609 and not require this option to begin with.

cmd/standalone/Dockerfile
```diff
@@ -22,4 +22,4 @@ COPY ./jaeger-ui-build /go/src/jaeger-ui-build
 COPY ./cmd/standalone/standalone-linux /go/bin/
 COPY ./cmd/standalone/sampling_strategies.json /go/src/config/
 
-CMD ["/go/bin/standalone-linux","--query.static-files=/go/src/jaeger-ui-build/build/","--sampling.strategies-file=/go/src/config/sampling_strategies.json"]
+ENTRYPOINT ["/go/bin/standalone-linux", "--query.static-files=/go/src/jaeger-ui-build/build/", "--sampling.strategies-file=/go/src/config/sampling_strategies.json"]
```
jpkrohling: While I think the static files would "never" change for the Docker image (or would they?), the strategies file are most likely to change in some scenarios. As it's easier to override a config file than a flag on an entrypoint directive, it would be best to use a configuration file instead of passing values via flags, like:

```dockerfile
COPY ./cmd/standalone/standalone.yaml /go/src/config/
ENTRYPOINT ["/go/bin/standalone-linux", "--config-file=/go/src/config/standalone.yaml"]
```

`./cmd/standalone/standalone.yaml` doesn't exist yet, but could look like this:

```yaml
    query:
      static-files: /go/jaeger-ui/
    sampling:
      strategies-file: /go/src/config/sampling_strategies.json
```

The other Dockerfiles should also follow the same pattern.

And finally, I would personally prefer `/conf/` instead of `/go/src/config/`, and `/bin/` instead of `/go/bin/`, unless there's a specific reason to keep `go` in the path name. 

cc @yurishkuro, @pavolloffay

pavolloffay: I don't have a strong opinion on directories. 

I have noticed that some images put binaries into `/bin` or `/usr/share` and configuration into `/etc` so in our case `/etc/jaeger`

jpkrohling: Those images are probably following the LSB (Linux Standard Base), which is a standard for this kind of thing on the Linux world. I'd be OK with that, if we do indeed place the config files at `/etc/jaeger/standalone.yaml` instead of `/etc/jaeger.yaml`.

pavolloffay: I meant `/etc/jaeger` as a dir for whatever configuration we need.

pavolloffay: But as I said I don't have a strong opinion. I am not sure what benefits putting binaries into /bin has as we use `from scratch`. One way or another following standard Linux conventions is not a bad thing :).

jpkrohling: > this also entails a code change in each binary to support pulling config values from a passed YAML file

No, this is already handled by Viper

> As there will now only ever be one argument anyways?

Good point. Imagine that someone just typed `docker run jaegertracing/all-in-one --collector.queue-size=10` . What is final expected command? 

`/go/bin/standalone-linux --config-file=/go/src/config/standalone.yaml --collector.queue-size=10`

or

`/go/bin/standalone-linux --collector.queue-size=10`

@pavolloffay what would Viper do in this case? What happens if the config file has "collector.queue-size=11"?

zdicesare: I had interpreted the suggested change as eliminating all other flags, so that passing in `--collector.queue-size=10` instead of specifying it in the config file would be an error.

Also,  overriding the config file will require a volume mount specified in `docker run`, right?

pavolloffay: The precedence is here https://github.com/spf13/viper#why-viper

we should default to the lowest one - definitely not flag because it overrides everything.

yurishkuro: > `ENTRYPOINT ["/go/bin/standalone-linux", "--config-file=/go/src/config/standalone.yaml"]`

I am not convinced of the value of this. Example shows only two entries in the config, and I think the ui-assets should be simply built into the binary, while strategies by default has to be empty, there's no point in having it in the config.

zdicesare:
Additional PRs at:
https://github.com/jaegertracing/jaeger-openshift/pull/81
https://github.com/jaegertracing/jaeger-kubernetes/pull/87

Command line params to the jaeger containers in https://github.com/kubernetes/charts/tree/master/incubator/jaeger are all empty, and I did not see any references to the executables in that repo but definitely could use a confirmation on that.

cmd/standalone/Dockerfile
jpkrohling: If there are no options that need to be passed, then there's no reason to have a config file. 

If there are CLI options to be passed (as is the case for the current state of this PR), then it's easier for a consumer to override a config file than override the entrypoint directive.

jpkrohling:
> Command line params to the jaeger containers in https://github.com/kubernetes/charts/tree/master/incubator/jaeger are all empty

It would then need only a version bump, to make use of the new feature. 

> but definitely could use a confirmation on that.

Also looks empty to me, but the best person to confirm would be @dvonthenen

cmd/standalone/Dockerfile
zdicesare: I may be a step behind you here- but doesn't the consumer overriding a config file via an argument to the container require a volume mount? In that case it seems easier to override the entrypoint with additional command line arguments.

dvonthenen:
@jpkrohling @zdicesare Yup, command line params are empty. All configuration is done via ENV variables. A bump in the version should take care of it.

cmd/standalone/Dockerfile
yurishkuro: The embedding of UI assets is done in #918. Once it's merged I suggest we change this one to
```
ENTRYPOINT ["/go/bin/standalone-linux"]
CMD ["--sampling.strategies-file=/go/src/config/sampling_strategies.json"]
```
I now see the value of using the config as the lowest priority default value, but I propose we leave it for another PR, as it won't be such a breaking change as this PR.

jpkrohling: Or rather:

```dockerfile
ENTRYPOINT ["/go/bin/standalone-linux"]
CMD ["--sampling.strategies-file=/etc/jaeger/sampling_strategies.json"]
```

https://github.com/jaegertracing/jaeger/pull/918/files#diff-da47d3e14c88bf43b29700d0fa2b7a27R22

jpkrohling: @jkandasa scenarios to test:

- Override the CMD with a custom option (`--log-level`, for instance)
- Override the `/etc/jaeger/sampling_strategies.json` file

cmd/query/Dockerfile
```diff
@@ -8,4 +8,4 @@ COPY --from=certs /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certifica
 EXPOSE 16686
 COPY query-linux /go/bin/
 ADD jaeger-ui-build /go/jaeger-ui/
-CMD ["/go/bin/query-linux", "--query.static-files=/go/jaeger-ui/"]
+ENTRYPOINT ["/go/bin/query-linux", "--query.static-files=/go/jaeger-ui/"]
```
jpkrohling: After #918, the `--query.static-files` option is optional and the assets are embedded into the binary, so, remove this from the entrypoint.

crossdock/docker-compose.yml
```diff
@@ -80,4 +80,4 @@ services:
 
     jaeger-query:
 #    override to disable static files
-        command: ["/go/bin/query-linux", "--query.static-files=", "--cassandra.keyspace=jaeger_v1_dc1", "--cassandra.servers=cassandra"]
+        command: ["--query.static-files=", "--cassandra.keyspace=jaeger_v1_dc1", "--cassandra.servers=cassandra"]
```
jpkrohling: The first option can be removed

zdicesare:
Thank you for updating this @yurishkuro, I can work on updating the other PRs with new images when this is out, including creating a version bump PR in https://github.com/kubernetes/charts/tree/master/incubator/jaeger

PR #823, opened 2018-05-15
Remove unnecessary parameter

vprithvi:

- Reduce complexity by removing unnecessary parameter

- Remove dependency-store-data-frequency

I also verified that the jaegertracing/documentation, openshift, kubernetes doesn't refer to this parameter


vprithvi:
>surprising. Is it somehow hardcoded elsewhere? I seem to recall there was a need for this var.

@yurishkuro I wasn't able to locate any uses

Issue #803, opened 2018-05-04
jaeger-agent: `--log-level` flag not recognized


cboggs:
The 1.4.1 release of jaeger agent doesn't appear to include the `--log-level` flag as indicated [here](https://www.jaegertracing.io/docs/monitoring/#Logging) and [here](https://github.com/jaegertracing/jaeger/pull/514).

Attempt results:
```
> docker run -it jaegertracing/jaeger-agent:1.4.1 /go/bin/agent-linux --log-level=warn
Error: unknown flag: --log-level
Usage:
  jaeger-agent [flags]
  jaeger-agent [command]

Available Commands:
  help        Help about any command
  version     Print the version

Flags:
      --collector.host-port string                            comma-separated string representing host:ports of a static list of collectors to connect to directly (e.g. when not using service discovery)
      --config-file string                                    Configuration file in JSON, TOML, YAML, HCL, or Java properties formats (default none). See spf13/viper for precedence.
      --discovery.min-peers int                               if using service discovery, the min number of connections to maintain to the backend (default 3)
  -h, --help                                                  help for jaeger-agent
      --http-server.host-port string                          host:port of the http server (e.g. for /sampling point and /baggage endpoint) (default ":5778")
      --metrics-backend string                                Defines which metrics backend to use for metrics reporting: expvar, prometheus, none (default "prometheus")
      --metrics-http-route string                             Defines the route of HTTP endpoint for metrics backends that support scraping (default "/metrics")
      --processor.jaeger-binary.server-host-port string       host:port for the UDP server (default ":6832")
      --processor.jaeger-binary.server-max-packet-size int    max packet size for the UDP server (default 65000)
      --processor.jaeger-binary.server-queue-size int         length of the queue for the UDP server (default 1000)
      --processor.jaeger-binary.workers int                   how many workers the processor should run (default 10)
      --processor.jaeger-compact.server-host-port string      host:port for the UDP server (default ":6831")
      --processor.jaeger-compact.server-max-packet-size int   max packet size for the UDP server (default 65000)
      --processor.jaeger-compact.server-queue-size int        length of the queue for the UDP server (default 1000)
      --processor.jaeger-compact.workers int                  how many workers the processor should run (default 10)
      --processor.zipkin-compact.server-host-port string      host:port for the UDP server (default ":5775")
      --processor.zipkin-compact.server-max-packet-size int   max packet size for the UDP server (default 65000)
[...]

PR #828, opened 2018-05-17
Support log-level flag on agent

black-adder:

- resolves https://github.com/jaegertracing/jaeger/issues/803

- Expose the --log-level flag but only the --log-level flag and not the rest of the shared flags

### DIFF
--- a/cmd/agent/main.go
+++ b/cmd/agent/main.go
@@ -73,6 +73,7 @@ func main() {
 		v,
 		command,
 		flags.AddConfigFileFlag,
+		flags.AddLoggingFlag,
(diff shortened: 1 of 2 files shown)


cmd/agent/main.go
```diff
@@ -73,6 +73,7 @@ func main() {
 		v,
 		command,
 		flags.AddConfigFileFlag,
+		flags.AddLoggingFlag,
```
yurishkuro: you're certainly adding it to the cmd line, but is it being used by main to control logging?

black-adder: yup, on L45

PR #830, opened 2018-05-17
Add 'gas' for security problems scanning

### LINKED ISSUE #455: Adopt static code analysis with security verification
Adopt a static code analysis that can also verify common security problems. One possible candidate is https://github.com/alecthomas/gometalinter . 

This is part of #404 .

jpkrohling:

- Closes #455
- Items `static_analysis_common_vulnerabilities` and `static_analysis_fixed` from the CII Best Practices

- Added `make gas`
- Annotated some of the issues to tell `gas` that those are OK
- Removed unused code, flagged by `gas`

## Others
- I tried adding `gometalinter` and activate only `gas`, but got a different output, so, I think it might make sense to cherry-pick the linters and configure them individually
- There's a `make lint` target, but doesn't seem to be used anywhere. And returns quite a bunch of messages :) So, I'm naming the env var `LINT`, but ignoring `make lint` for now. Hopefully, we'll use it later.


plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -59,7 +59,7 @@ func loadStrategies(strategiesFile string) (*strategies, error) {
 	if strategiesFile == "" {
 		return nil, nil
 	}
-	bytes, err := ioutil.ReadFile(strategiesFile)
+	bytes, err := ioutil.ReadFile(strategiesFile) /* nolint #nosec , this comes from an admin, not user */
```
yurishkuro: why does this require linter directives?

.travis.yml
```diff
@@ -11,6 +11,7 @@ matrix:
     env:
     - TESTS=true
     - COVERAGE=true
+    - LINT=true
```
yurishkuro: what's the reason to separate this from the main `make lint` target? Does it run much slower?

jpkrohling: Interesting, I thought `make lint` wasn't being used on purpose, because it fails with:

```
$ make lint
go vet ./cmd/... ./model/... ./pkg/... ./plugin/... ./crossdock/... ./storage/... .
cmd/query/app/static_handler_test.go:121: possible formatting directive in Error call
exit status 1
make: *** [Makefile:114: lint] Error 1
```

This failure does not show up in Travis though. I just changed this PR to fix the lint failure above and to get `make lint` to also call the target `gas`

plugin/sampling/strategystore/static/strategy_store.go
jpkrohling: The theory behind it is that `strategiesFile` can be manipulated by an end user, so that a file like `/etc/shadow` could be read. I tracked all instances of this case (G304, IIRC) and they all come from the CLI, meaning that they are set by an admin, and not by an end user.

Makefile
```diff
@@ -128,6 +133,10 @@ install-go-bindata:
 	go get github.com/jteeuwen/go-bindata/...
 	go get github.com/elazarl/go-bindata-assetfs/...
 
+.PHONY: install-gas
+install-gas:
+	go get github.com/GoASTScanner/gas/cmd/gas/...
```
yurishkuro: I am still not clear why we need the extra makefile targets instead of rolling it up into the main `install` and `lint` (extra targets increase noise / entropy)

jpkrohling: `install` is quite clean, doing only a `glide install`. It does delegate the installation of the dependent tool (`glide`) to a separate target, so, there's a precedent :-)

About `gas` getting called directly from `lint`, I guess it's a matter of taste. When working on fixing the warnings provided by `gas`, I find it easier to run it in isolation as `make gas`.

Issue #773, opened 2018-04-15
Use Protobuf and gRPC for internal communications


yurishkuro:
## Problem

At the moment we have the following data models handled by the backend, often with two-way transformers between them:
  * Domain model in `/model` used by all internal APIs and business logic
    * JSON serialization of the domain model is used in many unit and integration tests
  * `jaeger.thrift` as the main on-the-wire representation
  * `/model/json` is JSON used for communicating with the UI
  * `/model/json` with embedded Process used by ElasticSearch storage plugin
  * `zipkin.thrift` and Zipkin JSON accepted as input data models as well

We also have limitations on routing traffic between agents and collectors because they use TChannel as the communication protocol.

### Converge on a Protobuf 3 Model

By converging on a single Protobuf model we can consolidate some of these models, reduce the number of converters, and unblock / enable several other work streams:
  * The JSON format exposed by the query service can be based on standard [Protobuf JSON Mapping](https://developers.google.com/protocol-buffers/docs/proto3#json) as Jaeger's canonical JSON model (issue https://github.com/jaegertracing/jaeger/issues/706).
  * The collector can accept the same JSON format, which will unblock the work on pure Javascript client (issue https://github.com/jaegertracing/jaeger-client-javascript/issues/1)
  * The Protobuf model can be used to generate official documentation for the API, for gRPC and REST endpoints (issue https://github.com/jaegertracing/jaeger/issues/456).
  * The domain model can be replaced with code generated from Protobuf IDL (already done in https://github.com/jaegertracing/jaeger/issues/856). The canonical JSON representation can be used for unit / integration tests (also done).
  * The UI can be upgraded to use canonical JSON model, eliminating the need for `/model/json` (the ElasticSearch plugin will need to retain the model and converters, which should be migrated into the es module)

### Use gRPC with Protobuf model

* Since gRPC is based on HTTP/2, it has much wider support than TChannel, e.g. for routing via proxies
* gRPC can support TLS, addressing issues https://github.com/jaegertracing/jaeger/issues/458 and https://github.com/jaegertracing/jaeger/issues/404
* gRPC can be used to to serve both binary and REST endpoints, and generate documentation and Swagger API

### Phase 1 - MVP

Goal: deprecate TChannel.

* [ ] Support gRPC/protobuf endpoints for core backend communications
  * [x] Collector service (receiving spans)
  * [ ] Query service
  * [x] Configuration services in the collector (sampling, throttling, baggage restrictions)
[...]

PR #831, opened 2018-05-18
Remove ParentSpanID from domain model

### LINKED ISSUE #824: What to do with model.Span.ParentSpanID
As part of #773 I am trying to replace the domain model with gogoprotobuf-generated classes. It's been going well so far, the last stumbling block is the ParentSpanID field that currently does not exist in protobuf model. The reasoning for not including that field in the new model was to simplify the handling of span references, as we have seen bugs and confusion stemming from the duality of representation where `parentSpanID` is equivalent to a child-of span reference, while the reference itself may or may not be present in the `references` array (and in some cases duplicate references are present due to the same confusion/bugs).

There is also one use case that the current model does not capture. When a span is created with a list of predecessors (span references), all Jaeger clients pick one of those predecessors as the "primary parent" from which the trace ID and flags are inherited. In practice we don't have much instrumentation today that would even pass multiple references.

## Option 1 - remove `ParentSpanID`, use `References` field only

This is the current implementation in the `gogoproto-model` branch.

Pros:
  * unified handling of predecessors

Cons:
  * requires memory allocation for the slice in 99% of spans
  * does not capture the primary parent from which trace ID is inherited

## Option 2 - replace `ParentSpanID` with `ParentSpanRef`

Pros:
  * no extra memory allocations for 99% of spans
  * captures the primary parent

Cons:
  * bifurcates processing of primary parent from other predecessors

Comments:
  * this option is similar to how OpenCensus separates "parent span id" from ["links"][1]. In Census links are are directional and can be added to both sides of the edge.
  * the reason to store complete SpanRef instead of just parent span ID is to capture type of reference

[1]: https://github.com/census-instrumentation/opencensus-proto/blob/c5d48acbb2cb32687ca09177a313f5fcd47103fe/opencensus/proto/trace/trace.proto#L195

@jaegertracing/jaeger-maintainers

yurishkuro:
Resolves #824

This implements option 1 from #824 by removing ParentSpanID from the domain model and only using the References array. `ParentSpanID` is replaced with an accessor `ParentSpanID()` that returns the span ID of the first `child-of` reference for the same trace ID.

Once unfortunate discovery: both Cassandra and Elasticsearch models kept the ParentID field in their respective models. It is no longer populated in the domain->db model conversions, but still has to be taken into account in the dbmodel->domain conversion for backwards compatibility with existing data. I don't know of a better solution. 

A side effect of that is that we won't be able to use gogoprotobuf-generated model (from #773) to generate the JSON for Elasticsearch. The plan would be to move `model/converter/json` code into `plugin/storage/es/dbmodel`, similar to `plugin/storage/cassandra/dbmodel`, while the UI JSON model will be produced directly from protobuf model.

---
This change is [Reviewable](https://reviewable.io/reviews/jaegertracing/jaeger/831)

### DIFF
--- a/model/span.go
+++ b/model/span.go
@@ -66,17 +66,6 @@ func (s *Span) Hash(w io.Writer) (err error) {
 	return enc.Encode(s)
 }
 
-// ParentSpanID returns the ID of a parent span if it exists.
-// It searches the references for a reference pointing to the same trace ID.
-func (s *Span) ParentSpanID() SpanID {
-	for i := range s.References {
-		if s.References[i].TraceID == s.TraceID {
-			return s.References[i].SpanID
-		}
-	}
-	return SpanID(0)
-}
-
 // HasSpanKind returns true if the span has a `span.kind` tag set to `kind`.
 func (s *Span) HasSpanKind(kind ext.SpanKindEnum) bool {
 	if tag, ok := KeyValues(s.Tags).FindByKey(string(ext.SpanKind)); ok {
@@ -105,7 +94,20 @@ func (s *Span) NormalizeTimestamps() {
 	}
 }
 
-// ReplaceParentID replaces span ID in the span reference that is considered a parent span.
+// ParentSpanID returns ID of a parent span if it exists.
+// It searches for the first child-of reference pointing to the same trace ID.
+func (s *Span) ParentSpanID() SpanID {
+	for i := range s.References {
+		ref := &s.References[i]
+		if ref.TraceID == s.TraceID && ref.RefType == ChildOf {
+			return ref.SpanID
+		}
+	}
+	return SpanID(0)
+}
+
+// ReplaceParentID replaces span ID in the parent span reference.
+// See also ParentSpanID.
 func (s *Span) ReplaceParentID(newParentID SpanID) {
 	oldParentID := s.ParentSpanID()
 	for i := range s.References {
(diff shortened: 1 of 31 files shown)


model/span.go
```diff
@@ -95,6 +94,31 @@ func (s *Span) NormalizeTimestamps() {
 	}
 }
 
+// ParentSpanID returns ID of a parent span if it exists.
+// It searches for the first child-of reference pointing to the same trace ID.
+func (s *Span) ParentSpanID() SpanID {
+	for i := range s.References {
+		ref := &s.References[i]
+		if ref.TraceID == s.TraceID && ref.RefType == ChildOf {
```
black-adder: i've always wondered, why do we need traceid to be part of spanref? what use case is there when a span has a spanref to a different traceid?

model/span.go
```diff
@@ -95,6 +94,31 @@ func (s *Span) NormalizeTimestamps() {
 	}
 }
 
+// ParentSpanID returns ID of a parent span if it exists.
+// It searches for the first child-of reference pointing to the same trace ID.
+func (s *Span) ParentSpanID() SpanID {
+	for i := range s.References {
```
black-adder: why not iterate over the reference directly? ie `for _, ref := `

yurishkuro: each iteration will have to copy 4 words into the on-stack var

model/span.go
yurishkuro: meta-traces; multi-trace joins in buffered writes

PR #837, opened 2018-05-18
[WIP] Start work on sqlite plugin for Jaeger

fzakaria:
- Added beggining support for sqlite3 to Jaeger as a storage plugin.
- Library to use https://github.com/mattn/go-sqlite3 however requires
cgo
- Interfaces filled out but mostly default to nil

Signed-off-by: Farid Zakaria <farid.m.zakaria@gmail.com>

For _proof of concept_ deployments of Jaeger, it would be great to have something in between memory (too simple) and Cassandra/ElasticSearch (too much overhead). 
Sqlite seems to fit the bill nicely here as it provides a high throughput durable storage. The work here can potentially be leveraged to other SQL compatible systems as well through Go's sql package.
(If we find commonality and use ANSI SQL types)

- Add support for a new sqlite3 storage plugin.

### DIFF
--- a/plugin/storage/factory.go
+++ b/plugin/storage/factory.go
@@ -26,6 +26,7 @@ import (
 	"github.com/jaegertracing/jaeger/plugin/storage/cassandra"
 	"github.com/jaegertracing/jaeger/plugin/storage/es"
 	"github.com/jaegertracing/jaeger/plugin/storage/memory"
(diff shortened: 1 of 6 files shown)


yurishkuro:
@fzakaria I am going to close it since it's been outstanding for over a year. We do have a reliable way to use all-in-one with Badger for disk-backed storage.

Issue #818, opened 2018-05-15
Support archiving traces with ES storage


nziebart:
## Requirement - what kind of business use case are you trying to solve?

We use the ES backend, and would like to be able to archive traces

## Problem - what in Jaeger blocks you from solving the requirement?

Archiving is only supported by the Cassandra storage plugin

## Proposal - what do you suggest to solve the problem or improve the existing situation?

I briefly looked at the implementation of archiving. It looks like most of the logic is built on things that already existed, and that it would be fairly straightforward to add an ES implementation. Two options come to mind:
1) Allow configuring a second ES cluster for archiving (the same way we do for Cassandra)
2) Use a separate index for archive traces. The `esCleaner.py` script would need to ignore this index.

## Any open questions to address

It seems like we'd need a different way of dividing up the indexes to support this if we go with option (1) above. Currently we create an index for each day, which probably doesn't make sense for archiving. Option (2) above would solve this inherently.

yurishkuro:
I think this needs both 1 and 2 simultaneously. Probably requires #799 and #628 to be implemented first (per cluster), and then add ability to specify archiving cluster.

nziebart:
Is there a specific motivation for wanting a separate cluster for archiving, rather than just a separate keysapce/index?

pavolloffay:
@yurishkuro Are the archived traces shown on `/search` page? it seems that an archived trace can be only accessed directly via `/trace/id` endpoint.

pavolloffay:
If the above is true we don't want to create index for service names, therefore e will have to make changes to span writer and reader. I also assume that we will create only one archive index per deployment. 

To be able to support multiple tenants the index prefix will be just put in front of the archive index e.g. `tenant:jaeger-span-archive` or `tenant:jaeger-span-1970-01-01` :)

pavolloffay:
We should also think about retention for the archive index. We could do it per time like proposed in #628 (day, month, year) or just allow to use a different index name e.g. `jaeger-span-archive-2. It might be also doable with prefix.

Note that it's not expected to delete data from an index in ES.

pavolloffay:
^^ cc @jaegertracing/elasticsearch

masteinhauser:
Just as a counter-point, we would not use the native Jaeger archive at all.

[...]

PR #840, opened 2018-05-19
Allow configurable elasticsearch index date format

### LINKED ISSUE #628: Make default time span of ES index configurable
Right now ES storage rotates indices every 24hrs because of this code in es span writer:
```go
}
```

It would be better to have a CLI flag to configure the time span of the index, and truncate `span.StartTime` to that granularity.

pavelnikolov:
- Resolves #628

- This PR allows overriding the Elasticsearch index date format. There is a new flag when using `SPAN_STORAGE_TYPE=elasticsearch`:

```
--es.index.date-format string       The date format of the Elasticsearch index names suffix, using Golang date format. https://golang.org/pkg/time/#example_Time_Format (default "2006-01-02")
```

### DIFF
--- a/plugin/storage/es/spanstore/writer.go
+++ b/plugin/storage/es/spanstore/writer.go
@@ -50,6 +50,7 @@ type serviceWriter func(string, *jModel.Span)
 type SpanWriter struct {
 	ctx           context.Context
 	client        es.Client
+	dateFormat    string // used for elasticsearch rolling index name suffix
 	logger        *zap.Logger
 	writerMetrics spanWriterMetrics // TODO: build functions to wrap around each Do fn
 	indexCache    cache.Cache
@@ -76,6 +77,7 @@ type Span struct {
 // NewSpanWriter creates a new SpanWriter for use
 func NewSpanWriter(
 	client es.Client,
+	dateFormat string,
 	logger *zap.Logger,
 	metricsFactory metrics.Factory,
 	numShards int64,
@@ -89,9 +91,10 @@ func NewSpanWriter(
 	// TODO: Configurable TTL
 	serviceOperationStorage := NewServiceOperationStorage(ctx, client, metricsFactory, logger, time.Hour*12)
 	return &SpanWriter{
-		ctx:    ctx,
-		client: client,
-		logger: logger,
+		ctx:        ctx,
+		client:     client,
+		dateFormat: dateFormat,
+		logger:     logger,
 		writerMetrics: spanWriterMetrics{
 			indexCreate: storageMetrics.NewWriteMetrics(metricsFactory, "IndexCreate"),
 		},
@@ -109,7 +112,7 @@ func NewSpanWriter(
 
 // WriteSpan writes a span and its corresponding service:operation in ElasticSearch
 func (s *SpanWriter) WriteSpan(span *model.Span) error {
-	spanIndexName, serviceIndexName := indexNames(span)
+	spanIndexName, serviceIndexName := indexNames(span, s.dateFormat)
 	// Convert model.Span into json.Span
 	jsonSpan := json.FromDomainEmbedProcess(span)
 
@@ -129,8 +132,8 @@ func (s *SpanWriter) Close() error {
 	return s.client.Close()
 }
 
-func indexNames(span *model.Span) (string, string) {
-	spanDate := span.StartTime.UTC().Format("2006-01-02")
+func indexNames(span *model.Span, dateFormat string) (string, string) {
+	spanDate := span.StartTime.UTC().Format(dateFormat)
 	return spanIndexPrefix + spanDate, serviceIndexPrefix + spanDate
 }
 
--- a/plugin/storage/es/spanstore/writer_test.go
+++ b/plugin/storage/es/spanstore/writer_test.go
@@ -48,12 +48,18 @@ func withSpanWriter(fn func(w *spanWriterTest)) {
 		client:    client,
 		logger:    logger,
 		logBuffer: logBuffer,
-		writer:    NewSpanWriter(client, logger, metricsFactory, 0, 0),
+		writer:    NewSpanWriter(client, "2006-01-02", logger, metricsFactory, 0, 0),
 	}
 	fn(w)
 }
 
 var _ spanstore.Writer = &SpanWriter{} // check API conformance
+
+func TestWithDateFormat(t *testing.T) {
+	withSpanWriter(func(w *spanWriterTest) {
+		assert.Equal(t, "2006-01-02", w.writer.dateFormat, "Span writer index date format")
+	})
+}
 
 func TestClientClose(t *testing.T) {
 	withSpanWriter(func(w *spanWriterTest) {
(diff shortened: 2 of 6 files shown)


yurishkuro:
overall looks good, but a couple of comments:
  * does this PR require a corresponding change to the cron job script that cleans stale indices?
  * this change does not actually address the issue in #628, which was more about the frequency of index rotation than the format of the index name. In other words, the change would be to add a rotation frequency parameter, expressed as Duration, and then truncate the timestamps to that duration (https://golang.org/pkg/time/#Time.Truncate) before passing it to the date format function.
  * since this PR is already changing the name of the index, we might want to allow changing the rest of the name (prefix) as well, which would help #818

pavelnikolov:
1. I have no idea about the cron-job. I'll have to get familiar with its code.
2. Definitely using `time.Truncate` gives you ultimate control over the duration of the rolling indices. I was thinking of a very basic configuration:
2006-01-02T15:04:05.999999999Z07:00
  - _monthly index_: `2006-01`
  - _daily index_: `2006-01-02`
  - _hourly index_:  `2006-01-02-15`
Unfortunately, we can't, for example, have new index every 4 hours or a weekly index. But I was hoping that the above 3 would cover most cases. Do you insist on having a `time.Duration` parameter and using `time.Truncate`? If yes, I believe this needs to be pushed in a separate PR.
3. I'll add an option to update the prefix as well.

yurishkuro:
The cleaner script is here https://github.com/jaegertracing/jaeger/blob/master/plugin/storage/es/esCleaner.py

pavelnikolov:
Do you think that the dependencies job would be affected as well? https://github.com/jaegertracing/spark-dependencies/blob/master/jaeger-spark-dependencies-elasticsearch/src/main/java/io/jaegertracing/spark/dependencies/elastic/ElasticsearchDependenciesJob.java#L151

plugin/storage/es/spanstore/writer.go
```diff
@@ -76,6 +77,7 @@ type Span struct {
 // NewSpanWriter creates a new SpanWriter for use
 func NewSpanWriter(
 	client es.Client,
+	dateFormat string,
```
yurishkuro: since we're breaking the API, let's use a struct instead
```
type SpanWriterOptions struct {
 	logger *zap.Logger
 	dateFormat string
 	numShards int64
 	numReplicas int64
}

func NewSpanWriter(client es.Client, options SpanWriterOptions) {}
```

It will be easier to extend if we want to make index name configurable

plugin/storage/es/spanstore/writer_test.go
```diff
@@ -200,9 +206,37 @@ func TestSpanIndexName(t *testing.T) {
...
-	spanIndexName, serviceIndexName := indexNames(span)
-	assert.Equal(t, "jaeger-span-1995-04-21", spanIndexName)
-	assert.Equal(t, "jaeger-service-1995-04-21", serviceIndexName)
+
+	data := []struct {
+		caption string
+
+		dateFormat string
+
+		expectedSpanIndexName    string
```
yurishkuro: blank lines not needed

yurishkuro:
Yes, Spark job will be affected as well. But it would have to be a different PR.

pavelnikolov:
I’ll check soon. I’ve been too busy lately with other things.

pavolloffay:
I don't think we should go with this approach - configurable date format option. The biggest downside is the limitation on retention configuration:
* monthly index: 2006-01
* daily index: 2006-01-02
* hourly index: 2006-01-02-15
* yearly 2006

 - not possible to do weekly or N days/hours retention. 

It's also harder to configure and ES cleaner would not look nice either - re date format in python is different.



I think we should go with truncation (`span.StartTime.Truncate(d)`) with minimal value of 1 day.

The reader would use the `es.index.trunc-time` to get the list of indexes based on `es.max-span-age`.

```java
// Returns the array of indices that we need to query, based on query params
func (s *SpanReader) indicesForTimeRange(indexName string, startTime time.Time, endTime time.Time) []string {
	var indices []string
	firstIndex := indexWithDate(indexName, startTime, s.dateFormat)
	currentIndex := indexWithDate(indexName, endTime, s.dateFormat)
                // use trunc time
		endTime = endTime.Add(-s.truncTime * time.Hour)
		currentIndex = indexWithDate(indexName, endTime, s.dateFormat)
	}
	return append(indices, firstIndex)
}

jaegerIndices := s.indicesForTimeRange(s.serviceIndexPrefix, currentTime.Add(-s.maxSpanAge), currentTime)
```

pavolloffay:

yurishkuro:
Was there a specific reason someone asked specifically for date format rather than truncation?

pavolloffay:
from https://github.com/jaegertracing/jaeger/issues/628#issuecomment-388271428

> At the moment all time based indices that we have in our ES clusters are named like <index-name>_YYYY.MM.DD. And we have a single cronjob which deletes indices with data older than some configured number of days. Jaeger indices are the only exception to this naming convention and we have to run/maintain a separate cronjob just for Jaeger. Not a big deal, but it's a little annoying : )

I don't know how @pavelnikolov deletes indices. but IIRC curator should be format agnostic the creation date is retrieved from index properties.

masteinhauser:
> Was there a specific reason someone asked specifically for date format rather than truncation?

As long as the effect is the same: a span is written to the expected index, the means likely do not matter _as long as the span's original timestamp is retained._

> IIRC curator should be format agnostic the creation date is retrieved from index properties.

At least in later versions of elasticsearch-curator package, this is correct.

We are actually moving _away from_ jaeger's time-based index naming entirely and choosing to use the [Elasticsearch Rollover API](https://www.elastic.co/guide/en/elasticsearch/reference/6.5/indices-rollover-index.html) as it scales more predictably for us. Example: 25GB per-shard, automatically, which is ideal for our write/read workloads (both human, and automated anomaly detection)
This is also what we use with our FluentD configuration in Kubernetes for per-team centralized logging.

> not possible to do weekly or N days/hours retention.

At least in all of our usage, this largely doens't make sense in both our high-usage and low-usage indexes. The [Elasticsearch Rollover API](https://www.elastic.co/guide/en/elasticsearch/reference/6.5/indices-rollover-index.html) is a far better solution, from Elasticsearch's perspective, than sharding on time-based indices.

yurishkuro:
Rollover API looks pretty interesting, thanks for pointing it out. Perhaps it can also be used to avoid iterating through indices at query time.

pavolloffay:
> Perhaps it can also be used to avoid iterating through indices at query time.

I think the reader would have to use a different index - alias https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-aliases.html. Once the write index is rolled over the old index has to be moved to read alias. Perhaps we could use this to manage retention in archive index?

pavolloffay:
@pavelnikolov I will close this PR for two reasons:
* it does not seem as the right approach to configure retention - not flexible enough. Better is to use rollover API or truncate timestamp on configurable duration.
* it will complicate the index configuration on index cleaner and spark job - they don't use golang time format.

PR #856, opened 2018-06-01
Replace domain model with Protobuf/gogo-generated model 

yurishkuro:
For #773 

There are a lot of changes to the JSON files that contain the representation of previous domain model. Because gogo/protpbuf/jsonpb renders the model slightly differently, they had to be changed. However, I kept the model/json fixtures intact everywhere, so that this PR does not have any impact on the UI. In other words we can continue developing the gRPC & REST API using the new model, and once that's available the UI can be upgraded to the new model, then the `model/json` can be moved inside Elasticsearch storage plugin (I figured it's safer not to change that format and keep the converters, but scope them to ES only).

TODO:
* [x] the durations are rendered as strings, like "5000ns". We haven't reached the final consensus about what we want to do with those, working proposal - microseconds as float values (to allow for nanos).
* [x] because we need various extension methods on the domain types, the auto-generated `jaeger.pb.go` file is placed under `model`, but it has a lot less coverage. Makefile can grep its lines out of the coverage summary.
* [x] a bunch of code was commented out, need to clean it up


model/keyvalue.go
```diff
@@ -19,45 +19,42 @@ import (
...
-	VType ValueType `json:"vType"`
-	VStr  string    `json:"vStr,omitempty"`
-	VNum  int64     `json:"vNum,omitempty"`
-	VBlob []byte    `json:"vBlob,omitempty"`
-}
+// type KeyValue struct {
+// 	Key   string    `json:"key"`
+// 	VType ValueType `json:"vType"`
+// 	VStr  string    `json:"vStr,omitempty"`
+// 	VNum  int64     `json:"vNum,omitempty"`
```
yurishkuro: the KeyValue struct is now less memory efficient because `vNum` or split into `vBool/vInt64/vFloat64`

examples/proto/main.go
```diff
@@ -0,0 +1,134 @@
...
+	flag.Parse()
+	addr := fmt.Sprintf("localhost:%d", *gRPCPort)
+	lis, err := net.Listen("tcp", addr)
+	if err != nil {
+		log.Fatalln("Failed to listen:", err)
+	}
+	s := grpc.NewServer(
+	// grpc.Creds(credentials.NewServerTLSFromCert(&insecure.Cert)),
+	// grpc.UnaryInterceptor(grpc_validator.UnaryServerInterceptor()),
+	// grpc.StreamInterceptor(grpc_validator.StreamServerInterceptor()),
```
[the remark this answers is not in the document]
yurishkuro: example is WIP, I may remove it from the final PR, just experimenting with service JSON from grpc endpoint.

model/proto/jaeger.proto
```diff
@@ -0,0 +1,212 @@
...
+
+message SpanRef {
+  TraceID traceID = 1 [
+    (gogoproto.nullable) = false
+  ];
+  uint64 spanID = 2 [
+    (gogoproto.nullable) = false,
+    (gogoproto.customtype) = "SpanID" // alias to uint64
+  ];
+  SpanRefType refType = 3;
```
Falco20019: Not compliant to naming scheme: ref_type

model/proto/jaeger.proto
```diff
@@ -0,0 +1,212 @@
...
+}
+
+message Process {
+  string service_name = 1;
+  repeated KeyValue tags = 2 [
+    (gogoproto.nullable) = false
+  ];
+}
+
+message Span {
```
Falco20019: Here it's all mixy with the naming :) trace_id, span_id, start_time and process_id. operation_name is the only outlier here, but the correct one.

model/proto/jaeger.proto
```diff
@@ -0,0 +1,212 @@
...
+    ];
+    uint64 spanID = 2 [
+      (gogoproto.nullable) = false,
+      (gogoproto.customtype) = "SpanID"
+    ];
+}
+
+message Trace {
+  message ProcessMapping {
+      string processID = 1;
```
Falco20019: Not compliant to naming scheme: process_id

model/proto/jaeger.proto
```diff
@@ -0,0 +1,212 @@
...
+      Process process = 2 [
+        (gogoproto.nullable) = false
+      ];
+  }
+  repeated Span spans = 1;
+  repeated ProcessMapping processMap = 2 [
```
Falco20019: Not compliant to naming scheme: process_map

model/proto/jaeger.proto
```diff
@@ -0,0 +1,212 @@
...
+    (gogoproto.stdduration) = true,
+    (gogoproto.nullable) = false
+  ];
+  repeated KeyValue tags = 8 [
+    (gogoproto.nullable) = false
+  ];
+  repeated Log logs = 9 [
+    (gogoproto.nullable) = false
+  ];
+  Process process = 10;
```
Falco20019: `process` and `processID` are XOR, so it should be a oneof. Has no impact on JSON but makes using them easier.

isaachier:
If we end up using this in the clients, we need a better way to bundle the new protobuf dependencies. The so-called "well-known types" are not included in the default protobuf package. Installing Go packages is never ideal for C# or any other language. We should provide them ourselves seeing as that is what all these Go packages are doing anyway.

Falco20019:
Is gogoproto a separate compiler that is compatbile to regular proto or is it a incompatible fork? In C# and Java, the well-knowns are included in the base package. The protos are only in the `Google.Protobuf.Tools` as part of the build chain. But `Google.Protobuf`, the main library has implementations and also the protoc-csharp-plugin does skip generation of code for them.

Very interesting that that's not the case for all languages. They sould be at least part of the build chain.

model/proto/jaeger.proto
```diff
@@ -0,0 +1,212 @@
...
+  uint64 low = 1;
+  uint64 high = 2;
+}
+
+// For comparison of how KeyValue looks when using `oneof`.
+// Both definitions result in the same message size.
+// message KeyValue2 {
+//   string key = 1;
+//   oneof value {
+//     string vStr = 2;
```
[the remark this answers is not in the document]
yurishkuro: Where is the naming scheme defined?  I only see examples in snake case.

model/proto/jaeger.proto
```diff
@@ -0,0 +1,212 @@
...
+
+enum ValueType {
+  STRING  = 0;
+  BOOL    = 1;
+  INT64   = 2;
+  FLOAT64 = 3;
+  BINARY  = 4;
+};
+
+message KeyValue {
```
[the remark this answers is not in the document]
yurishkuro: Oneof is terrible in Go in both usability and performance. Why do you find them easier to use? You need a switch statement either way, and in most cases you need an additional cast.

model/proto/jaeger.proto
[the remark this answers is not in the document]
Falco20019: See the link in the documentation. Everything from Google uses snake_case (with underscores). vStr is camelCase.

yurishkuro:
@Falco20019 appreciate the review. Left a couple questions for you. 

I don't mind using snake_case in protobuf there's an established convention. I just need to see how that would affect the json, I'd rather avoid major changes there if we have a chance of using it with the UI.

model/proto/jaeger.proto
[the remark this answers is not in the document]
Falco20019: In C# you have the switch in both cases, either doing your own enum or using the oneof, and you get the enum generated and maintained by the code generator. Performance-wise it's the same as every other non-oneof field (also on the wire), just that the code sets you a second field which enum was selected, which you would have set yourself either way.

What casts do you need in Go? This is how it looks in C#: https://developers.google.com/protocol-buffers/docs/reference/csharp-generated#oneof

You get an additional enum `AvatarOneofCase`, you get the property `AvatarCase` which is baiscally your `vType`, you get `ImageUrl` and `ImageData` as if they would have been regular fields (no difference at all).

Falco20019:
I would assume that you can use options to tell the JSON name when using grpc-gateway. Haven't used it though, only the C# implementation of the JSON serializer...

Just had a look at the [go oneof implementation](https://developers.google.com/protocol-buffers/docs/reference/go-generated#oneof). Looks like it generates a separate struct for each field, which is, I assume, what you meant with worse performance. From the usability and readability perspective, it seems not too much worse I think. You would have to do the switch either way (using oneof or an enum) and since it generates the getter functions, you can access it just like you would have before. Could it be, that gogoproto is just doing this differently?

model/proto/jaeger.proto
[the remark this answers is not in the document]
yurishkuro: In Go the oneof is modeled as an interface field (example: https://github.com/gogo/protobuf/issues/168), which means it requires memory allocation even for primitive values, with a fairly ugly syntax like 
```go
kv := KeyValue{
    Key:   "key",
    Value: Value_Int{Int: 123},
}
```
and on reading you need to switch on the type (I was wrong, it does not require additional cast).

The memory allocations are the main turn-off of oneof for me, especially for such frequently used thing as KeyValue type. For ProcessID/ProcessMap I could see us using a oneof, but again the syntax is quite ugly.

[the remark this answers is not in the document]
Falco20019: I agree, in C# it looks like this:
```csharp
var kv = new KeyValue{
    Key = "key",
    Int = 123,
}
```
For reading:
```csharp
switch(kv.ValueCase) {
  case ValueOneofCase.Int:
    var int = kv.Int;
...
}
```

No extra types or anything :) Would be interesting how other languages handle this.

model/proto/jaeger.proto
```diff
@@ -0,0 +1,212 @@
...
+  option (gogoproto.equal) = true;
+  option (gogoproto.compare) = true;
+
+  string    key      = 1;
+  ValueType vType    = 2;
```
vprithvi: Do we need `valueType`? Could we simply check which field is set when deserializing?

isaachier: It probably depends on the language. Have we considered using the Google well known protobuf type `Any`? I assume that is an optimized format.

Falco20019:
Any serializes as byte array with type string. So could be a lot bigger for small data like int

isaachier:
I wonder if the builtin types would need a url. Maybe it is optimized.

isaachier:
Sorry I guess I should have said [Value](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Value). Our spans can just contain an instance of [Struct](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Struct) and that would work as well.

List of the well-known types: https://developers.google.com/protocol-buffers/docs/reference/google.protobuf.

isaachier:
@Falco20019 Go tries to avoid unsafe language constructs, so there are no unions. Hence the use of multiple struct fields. However, I surprised gogoprotobof didn't generate pointers to the field types. Maybe the premise is that allocation is more expensive than unused space? IDK.

isaachier:
Also, Google's support for Go using Protobuf is questionable at best. The standard Protobuf package leaves a lot to be desired. That's the whole reason gogoprotobuf, a fork of the original Go Protobuf package, has become more popular than the original. If you look at the Protobuf code, it is clear that the real performance optimizations occur in C++ clients (best example is arena allocation).

isaachier:
Never mind, I see the actual code is more complex to mimic a union. Sort of resembles the pointer solution.

model/proto/jaeger.proto
yurishkuro: fixed

model/proto/jaeger.proto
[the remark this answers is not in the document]
yurishkuro: fixed

yurishkuro:
Too many comments about `oneof`, which doesn't concern me too much. Since it results in very inefficient Go types, I am not planning to use it.

I changed the naming to snake_case, one negative side effect is that traceID/spanID in JSON are now camelCase traceId/spanId (more UI impact). But at least it's canonical proto JSON format.

The bigger open question for me is using `bytes` for IDs, which will change the representation from our custom rendering as hex strings to base64 strings. But if we want to keep standard proto JSON format, I don't see what else we can do. We can add some syntactic sugar in the UI to accept both base64 and hex strings, but it's certainly confusing that on-the-wire representation in trace context is hex, but JSON is base64.

isaachier:
IMHO it is worth reviewing the options available to generate efficient code for `oneof` seeing as it is axiomatic to Protobuf and is more efficient in pretty much every other language.

I don't think using bytes makes sense for trace ID because the type is fixed. Bytes makes sense for binary strings.

yurishkuro:
@isaachier not following you.

What do you propose about oneof?

What do you mean by "bytes ... are fixed"?  I was going to change Span like this:
```
message Span {
  bytes trace_id = 1 [
    (gogoproto.nullable) = false,
    (gogoproto.customname) = "TraceID",
    (gogoproto.customtype) = "TraceID"
  ];
  bytes span_id = 2 [
    (gogoproto.nullable) = false,
    (gogoproto.customname) = "SpanID",
    (gogoproto.customtype) = "SpanID"
  ];
```

and 

```go
// TraceID is a random 128bit identifier for a trace
type TraceID [16]byte

// SpanID is a random 64bit identifier for a span.
type SpanID [8]byte
```

This avoids extra memory allocations. Unfortunately, it also requires me to re-implement marshaling methods that are otherwise available for `bytes` type - the PB methods are easy, just copying the bytes, the JSON methods would have to use base64.

isaachier:
I see the issue on the oneof types and it looks like you are correct. I wish there were another way to handle this.

About the byte layout, that's a lot of gogoprotobuf customization. Have you tried generating any other language using this Protobuf file. I wonder if it will work.

isaachier:
Wow very interesting article here about the differences between Go's protobuf package and gogoprotobuf: https://jbrandhorst.com/post/gogoproto/.

> A few days ago I asked whether Go protobufs were used primarily in Go <-> Go communication, or Go <-> other language communication and it seemed that many users are in pure Go. I suspect that this fact is a source of contention between golang/protobuf, gogo/protobuf, and the users in the Go community. Each party have slightly different motivations:
> 
> golang/protobuf lies within the larger spectrum of language specific proto implementations (e.g., C++, Java, C#, JavaScript, etc) and follows the guidance of the proto team. The proto team is heavily concerned with uniformity of behavior across all languages, and less so about making protos great in any one particular language.
> 
> gogo/protobuf is very Go centric and provides many user-tweakable knobs to adjust the generated struct. Naturally, gogo/protobuf appeals heavily to the Go community.

yurishkuro:
I haven't tried the other languages, will give it a try, but they are supposed to simply ignore gogo options. All the options I used affect the go generated code, not how the buffers are serialized.

isaachier:
I agree in theory but had difficulties in practice with the gogo protobuf annotations. The article I quoted implies gogo protobuf isn't interested in supporting more than Go.

Falco20019:
We just have to make sure, that it's not getting a burden to use with other language just because gogoproto helps you with annotations. For example using `[16]byte` and `[8]byte` as alias helps you, but it does not help anyone else. Point given, it doesn't matter, since for every language it's just a byte array and you only have information about the length at compile time that every other language just get at runtime on deserialization. So I'm fine with using bytes.

But if it would make usability in other languages harder, I would like to discuss against it.

yurishkuro:
*Update*: I changed trace/spanID fields to be `bytes`, backed by custom types in Go. Unfortunately, that meant I had to implement proto & json serialization manually. To ensure that it's the same as the default  serialization I created a new file `jaeger_test.proto` that contains a copy of `SpanRef` message without any gogo options. This test file is compiled with plain `protoc -go_out` (without gogo), and there is a test `TestTraceSpanIDMarshalProto` in `model/ids_test.go` that performs proto and JSON marshaling/unmarshaling to ensure  that the outputs are identical (would be cool to fuzz that test).

At this point the only point of debate remains the `oneof`. My intention is to keep the model as is (i.e not using `oneof`), for the reasons I've already explained. The semantic cleanliness of the model with `oneof` does not compensate the ugliness it creates in the domain model. It's possible to use custom types, but it would require re-implementing serialization, which would be rather large, around 100-200 lines of ugly protobuf-encoding code.

Outstanding: a large corpus of fixtures in the integration tests need trace/spanID values replaced with base64 strings, I need to write some script to do it in bulk. All other tests seem to be ok.

isaachier:
I'm fine with all that. Any chance to try in another language.

yurishkuro:
> Any chance to try in another language.

I tried generating for Java, Python, and C++. Can't say if the actual classes make sense, or if they'd even compile, but `protoc` itself did not complain.

yurishkuro:
Ready for review

isaachier:
Somewhat separate issue, but if we use this format, would it be possible to have the backend accept both Protobuf and JSON? That way a new client can choose to use Protobuf as a dependency or just to use pure JSON.

isaachier:
Also, is there any reason not to move `jaeger.proto` to the [jaeger-idl](https://github.com/jaegertracing/jaeger-idl) repo?

Makefile
```diff
@@ -291,3 +293,64 @@ generate-mocks: install-mockery
...
+
+	# Workaround for https://github.com/grpc-ecosystem/grpc-gateway/issues/229.
+	sed -i '' "s/empty.Empty/types.Empty/g" model/jaeger.pb.gw.go
+
+	protoc \
+		-I model/proto \
+		--go_out=$$GOPATH/src/github.com/jaegertracing/jaeger/model/prototest/ \
+		model/proto/jaeger_test.proto
+
+proto-install:
```
isaachier: `.PHONY`?

Makefile
```diff
@@ -291,3 +293,64 @@ generate-mocks: install-mockery
...
+	go install \
+		./vendor/github.com/gogo/protobuf/protoc-gen-gogo \
+		./vendor/github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway \
+		./vendor/github.com/grpc-ecosystem/grpc-gateway/protoc-gen-swagger
```
isaachier: Get this on my Debian desktop when trying this command:

```
$ make proto-install
go install \
	./vendor/github.com/gogo/protobuf/protoc-gen-gogo \
	./vendor/github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway \
	./vendor/github.com/grpc-ecosystem/grpc-gateway/protoc-gen-swagger
vendor/github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway/main.go:18:2: cannot find package "github.com/golang/glog" in any of:
	/home/ihier/proj/go/src/github.com/jaegertracing/jaeger/vendor/github.com/golang/glog (vendor tree)
	/usr/local/go/src/github.com/golang/glog (from $GOROOT)
	/home/ihier/proj/go/src/github.com/golang/glog (from $GOPATH)
make: *** [Makefile:352: proto-install] Error 1
```

isaachier: After finally getting it to work and generating Python code, I get this clearly invalid import:

```
from github.com.gogo.protobuf.gogoproto import gogo_pb2 as github_dot_com_dot_gogo_dot_protobuf_dot_gogop
roto_dot_gogo__pb2
```

isaachier: Same with C++:
```
#include "github.com/grpc-ecosystem/grpc-gateway/protoc-gen-swagger/options/annotations.pb.h"
#include "github.com/gogo/protobuf/gogoproto/gogo.pb.h"
```

yurishkuro:
@isaachier 

> Somewhat separate issue, but if we use this format, would it be possible to have the backend accept both Protobuf and JSON? That way a new client can choose to use Protobuf as a dependency or just to use pure JSON.

That is the intent.

> Also, is there any reason not to move jaeger.proto to the jaeger-idl repo?

In the future, too early for it now, introduces unnecessary dev friction. I would like to get a gRPC channel between agent & collector working first before moving proto files to idl.

Makefile
yurishkuro: we may have to strip them post-generation. E.g. in python these two extra imports are not actually used.

isaachier:
As long as this works for Go, LGTM. Other languages can be solved later or use JSON.

PR #857, opened 2018-06-01
Switch to codecov

yurishkuro:


.travis.yml
```diff
@@ -58,7 +58,8 @@ script:
   - if [ "$DEPLOY" == true ]; then make build-binaries-linux build-binaries-windows build-binaries-darwin ; else echo 'skipping linux'; fi
 
 after_success:
-  - if [ "$COVERAGE" == true ]; then travis_retry goveralls -coverprofile=cover.out -service=travis-ci || true ; else echo 'skipping coverage'; fi
+  - if [ "$COVERAGE" == true ]; then mv cover.out coverage.txt ; else echo 'skipping coverage'; fi
+  - if [ "$COVERAGE" == true ]; then bash <(curl -s https://codecov.io/bash) ; else echo 'skipping coverage'; fi
```
isaachier: Why two separate lines instead of one large block?

yurishkuro: this way it works, experimenting with a different syntax is 10min wait each time, I don't have patience for it.

isaachier: :) ok nbd anyway

Issue #274, opened 2017-07-16
Kafka for trace storage


vprithvi:
I feel that Kafka could be a good storage backend because the primary way of discovering traces is over time windows that are short. 

### Basic idea
Have `jaeger-agent` or `jaeger-collector` emit spans into Kafka with the `traceID` as key. Instead of using Cassandra/Elasticsearch as an index, simply operate on the data itself. 

### What charecteristics of Kafka apply?
- Durable, high [read/write throughput](https://engineering.linkedin.com/kafka/benchmarking-apache-kafka-2-million-writes-second-three-cheap-machines) 
- Easy(?) to scale
 
### What charecteristics of traces/spans apply?
- Immutability 

### What is the data model? 
The data model could be a thrift serialized representation of the span.

### How do we combine spans into traces?
Kafka guarantees that all messages for a given key end up on a single partition. Combining spans into traces can be done on consumers by using Kafka's session windows based on traceID during query time. 

This assumes that all spans with a given window fit in memory, which depends on window size and data volume. We might be able to improve upon the session window implementation by making it aware of the service name and operation name, allowing us to quickly discard unwanted spans. 

I think that the viability of Kafka as a backend storage hinges on whether we are able to do this operation quickly for a particular service name. If we aren't, we would need to rely on a streaming solution like Flink, Samza, etc which greatly increase complexity for common use cases. 

### How would search work?
Once we've combined spans into traces, we could use [Lucene-Luwak](https://github.com/flaxsearch/luwak) for searches.

Because we are reading from Kafka, we can easily tell users how far we are into the search process, and how much data remains in the time window. 
We would also be able to keep expanding the time window until we hit a match. 
We would be able to support `tail -f` style searches for debugging, etc. 
Search performance would also be linear wrt the time windows selected, assuming a constant data volume. 

### How about aggregations?
We could use a similar approach to search, and maintain a running percentile, etc. 

### How does adaptive sampling work?
Adaptive sampling parameters can be computed by a kafka consumer, and propagated to kafka producers using kafka itself. We could use the same mechanism for baggage whitelisting and other configuration changes. 

See also: https://github.com/uber/jaeger/issues/212, which argues for not having kafka streaming directly out of Jaeger clients.

jpkrohling:
Can this be closed, or are there still things to be done related to this issue?

black-adder:
Just waiting for https://github.com/jaegertracing/jaeger/pull/880 so that the kafka writing can be enabled

[...]

PR #862, opened 2018-06-08
Add Kafka as a Storage Plugin

davit-y:
- Discussed in #274 

- Allows spans to be produced to Kafka
- Toggled by changing storage type to Kafka


plugin/storage/kafka/marshaller.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+}
+
+// Marshall encodes a span as a thrift byte array
+func (h *thriftMarshaller) Marshal(span *model.Span) ([]byte, error) {
+	thriftSpan := jaeger.FromDomainSpan(span)
+
+	thriftSpan.Write(h.protocol)
+
+	return h.memBuffer.Bytes(), nil
```
yurishkuro: - does membuffer need to be reset between writes?
- can this be called concurrently?

plugin/storage/kafka/factory.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/Shopify/sarama"
+	"github.com/spf13/viper"
+	"github.com/uber/jaeger-lib/metrics"
+	"go.uber.org/zap"
+
+	"github.com/jaegertracing/jaeger/storage/dependencystore"
+	"github.com/jaegertracing/jaeger/storage/spanstore"
+)
+
+const configPrefix = "kafka"
```
yurishkuro: why is it defined here, but used in `options.go`?

plugin/storage/kafka/options.go
```diff
@@ -0,0 +1,52 @@
...
+type Options struct {
+	brokers      []string
+	brokersInput string // Viper does not support lists so a comma-separated string is used as input
+	topic        string
+}
+
+// AddFlags adds flags for Options
+func (opt *Options) AddFlags(flagSet *flag.FlagSet) {
+	flagSet.String(
+		configPrefix+"-"+optionBrokers,
```
yurishkuro: why not add dash to the prefix constant for better readability?

plugin/storage/kafka/writer.go
```diff
@@ -0,0 +1,84 @@
...
+type SpanWriter struct {
+	metrics    spanWriterMetrics
+	producer   sarama.AsyncProducer
+	marshaller Marshaller
+	topic      string
+}
+
+// NewSpanWriter initiates and returns a new kafka spanwriter
+func NewSpanWriter(producer sarama.AsyncProducer, topic string, factory metrics.Factory) *SpanWriter {
+	writeMetrics := spanWriterMetrics{
```
yurishkuro: let's use reflection-based initializer

plugin/storage/kafka/writer.go
```diff
@@ -0,0 +1,84 @@
...
+// SpanWriter writes spans to kafka. Implements spanstore.Writer
+type SpanWriter struct {
+	metrics    spanWriterMetrics
+	producer   sarama.AsyncProducer
+	marshaller Marshaller
+	topic      string
+}
+
+// NewSpanWriter initiates and returns a new kafka spanwriter
+func NewSpanWriter(producer sarama.AsyncProducer, topic string, factory metrics.Factory) *SpanWriter {
```
yurishkuro: shouldn't marshaller be passed to the constructor?

plugin/storage/kafka/writer.go
```diff
@@ -0,0 +1,84 @@
...
+
+type spanWriterMetrics struct {
+	SpansWrittenSuccess metrics.Counter
+	SpansWrittenFailure metrics.Counter
+}
+
+// SpanWriter writes spans to kafka. Implements spanstore.Writer
+type SpanWriter struct {
+	metrics    spanWriterMetrics
+	producer   sarama.AsyncProducer
```
yurishkuro: is this an interface or struct?  we should depend on interface, otherwise you cannot test this file

plugin/storage/kafka/writer.go
```diff
@@ -0,0 +1,84 @@
...
+	// in the background as efficiently as possible
+	w.producer.Input() <- &sarama.ProducerMessage{
+		Topic: w.topic,
+		Key:   sarama.StringEncoder(span.TraceID.String()),
+		Value: sarama.ByteEncoder(spanBytes),
+	}
+	return nil
+}
+
+// Close closes SpanWriter by closing producer
```
yurishkuro: `// Close shuts down SpanWriter by closing the Kafka producer, which closes its channels causing go-routines to exit.`

There may be a race condition here if producer closes the input channel but someone calls WriteSpan after that. May need to keep an atomic flag to stop writes after Close()

plugin/storage/kafka/writer.go
davit-y: It's an interface. https://github.com/Shopify/sarama/blob/35324cf48e33d8260e1c7c18854465a904ade249/async_producer.go#L19

plugin/storage/kafka/options.go
```diff
@@ -0,0 +1,53 @@
...
+	brokersInput string // Viper does not support lists so a comma-separated string is used as input
+	topic        string
+}
+
+// AddFlags adds flags for Options
+func (opt *Options) AddFlags(flagSet *flag.FlagSet) {
+	flagSet.String(
+		configPrefix+suffixBrokers,
+		"127.0.0.1:9092",
+		"The comma-separated list of kafka brokers. i.e. 127.0.0.1:9092")
```
black-adder: nit: provide an actual comma-separated list as an example

plugin/storage/kafka/writer.go
vprithvi: If the input channel is closed, shouldn't the send to the closed channel panic? 
I think we should revisit how releasing resources work, and do something similar to the following:
- Set healthcheck to unavailable
- Close tchannel and http servers
- Close span writers

davit-y:
I had a misconfigured git config locally, so I needed to squash my commits to pass DCO sign-off checker.

yurishkuro:
- find a coming X before your changes started
- git reset --soft X
- all your changes are now pending and staged
- git commit -s
- git push --force

davit-y:
Thanks Yuri, that's actually exactly what I ended up doing since my fork was behind origin/master.

plugin/storage/kafka/writer_test.go
```diff
@@ -0,0 +1,183 @@
...
+
+func withSpanWriter(t *testing.T, fn func(span *model.Span, w *spanWriterTest)) {
+	serviceMetrics := metrics.NewLocalFactory(100 * time.Millisecond)
+	saramaConfig := sarama.NewConfig()
+	saramaConfig.Producer.Return.Successes = true
+	producer := saramaMocks.NewAsyncProducer(t, saramaConfig)
+	marshaller := &mocks.Marshaller{}
+	marshaller.On("Marshal", mock.AnythingOfType("*model.Span")).Return([]byte{}, nil)
+
+	tags := model.KeyValues{
```
black-adder: move the tags under the writerTest

plugin/storage/kafka/writer_test.go
```diff
@@ -0,0 +1,183 @@
...
+		model.String("someStringTagKey", "someStringTagValue"),
+	}
+	writerTest := &spanWriterTest{
+		producer:       producer,
+		marshaller:     marshaller,
+		metricsFactory: serviceMetrics,
+		writer:         NewSpanWriter(producer, marshaller, "someTopic", serviceMetrics),
+	}
+
+	span := &model.Span{
```
black-adder: is there anything special about this span? Can we move it out into a var at the top of the file and share it with the marshaller test?

plugin/storage/kafka/writer_test.go
```diff
@@ -0,0 +1,183 @@
...
+			Tags:        tags,
+		},
+	}
+	fn(span, writerTest)
+}
+
+func TestKafkaWriter(t *testing.T) {
+	withSpanWriter(t, func(span *model.Span, w *spanWriterTest) {
+
+		w.producer.ExpectInputAndSucceed()
```
black-adder: mocks done right

glide.yaml
```diff
@@ -46,7 +46,7 @@ import:
 - package: github.com/spf13/cobra
   version: 0.0.1
 - package: github.com/spf13/viper
-  version: ^1
+  version: 1.0.0
```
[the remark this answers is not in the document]
davit-y: Last week the newest version of viper was not compiling due to a dependency issue, no longer necessary.

plugin/storage/kafka/factory_test.go
```diff
@@ -0,0 +1,65 @@
...
+	return nil, m.err
+}
+
+func TestMemoryStorageFactory(t *testing.T) {
+	f := NewFactory()
+	v, command := config.Viperize(f.AddFlags)
+	command.ParseFlags([]string{})
+	f.InitFromViper(v)
+
+	f.config = &mockProducerBuilder{
```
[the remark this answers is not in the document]
davit-y: Initialization failure error on line 58 and flag testing in options_test

glide.yaml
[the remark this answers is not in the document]
jpkrohling: This probably belongs to its own PR then.

plugin/storage/kafka/factory.go
```diff
@@ -0,0 +1,84 @@
...
+// InitFromViper implements plugin.Configurable
+func (f *Factory) InitFromViper(v *viper.Viper) {
+	f.options.InitFromViper(v)
+	f.config = &f.options.config
+}
+
+// Initialize implements storage.Factory
+func (f *Factory) Initialize(metricsFactory metrics.Factory, logger *zap.Logger) error {
+	f.metricsFactory, f.logger = metricsFactory, logger
+	p, err := f.config.NewProducer()
```
jpkrohling: Could you log the contents of `f.config`? Like:

```go
logger.Info("Kafka storage configuration", zap.Any("configuration", f.config))
```

plugin/storage/kafka/writer.go
black-adder: I'm gonna allow this to land and punt on this. This rarely is a concern given how we do deploys internally anyway. And who does graceful shutdowns anyway.

glide.yaml
```diff
@@ -55,3 +55,7 @@ import:
 - package: github.com/go-openapi/validate
 - package: github.com/go-openapi/loads
 - package: github.com/elazarl/go-bindata-assetfs
+- package: github.com/Shopify/sarama
+  version: ^1.16.0
+- package: golang.org/x/sys
```
black-adder: There was some dependency strangeness that I debugged with Davit but couldn't figure out so we had to pin this. I'm gonna land this to unblock and I'll address this separately.

black-adder:
Looks like things are broken: https://github.com/golang/go/issues/25872

yurishkuro:
Someone posted a workaround - https://github.com/web-platform-tests/wpt.fyi/commit/cf5af40bf4de40a0a71f5b010d295c4f9aaca757, which probably can be simpler in our case if we simply pin x/tools to some version.

plugin/storage/kafka/marshaller.go
```diff
@@ -0,0 +1,45 @@
...
+	tProtocolFactory *thrift.TBinaryProtocolFactory
+}
+
+func newThriftMarshaller() *thriftMarshaller {
+	return &thriftMarshaller{thrift.NewTBinaryProtocolFactoryDefault()}
+}
+
+// Marshall encodes a span as a thrift byte array
+func (h *thriftMarshaller) Marshal(span *model.Span) ([]byte, error) {
+	thriftSpan := jaeger.FromDomainSpan(span)
```
black-adder: @yurishkuro @vprithvi @davit-y brought up a very good point that the thrift model doesn't have the process in the definition (https://github.com/jaegertracing/jaeger-idl/blob/master/thrift/jaeger.thrift#L51) (our internal kafka model does). We have a couple of options,
1) Add process to the thrift model
2) Pass a batch with 1 span instead
3) Write a json marshaller for now

I've told Davit to go ahead with option 3 for now until we come to a consensus on how to move forward.

yurishkuro: my preference is to land my proto PR and use that. You can marshal domain model directly into `[]byte`.

plugin/storage/kafka/factory.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+// CreateSpanWriter implements storage.Factory
+func (f *Factory) CreateSpanWriter() (spanstore.Writer, error) {
+	return NewSpanWriter(f.producer, f.marshaller, f.options.topic, f.metricsFactory), nil
+}
+
+// CreateDependencyReader implements storage.Factory
+func (f *Factory) CreateDependencyReader() (dependencystore.Reader, error) {
+	return nil, errors.New("kafka storage is write-only")
```
jpkrohling: Pardon my ignorance here, but does it mean that we won't be seeing any traces on the UI when Kafka is the storage being used? Shouldn't it then be a "secondary writer", instead of "main storage"?

black-adder: Yes I think that's fair. Internally we're currently using kafka as the primary storage and we have another downstream ingestor service which takes the kafka spans and inserts them into cassandra and another separate service which handles the indexing in cassandra. This way, if cassandra is overburdened, the indexer will skip indexing to ensure that the ingestor can at least write the span. We don't expect all users of Jaeger to be doing this (the load might not require this elaborate setup) so they should be able to dual write from the collector to cassandra and kafka (for the data pipeline).

Currently, we don't have a good way of configuring multiple storages via flags (I guess you can pass in comma separated list instead and we'd generate two storages for you). That being said, I think it can be all addressed in a separate PR.

plugin/storage/kafka/marshaller.go
yurishkuro: proto model was merged

yurishkuro: modeel/ids_test.go shows example of to/from `[]byte` marshaling

PR #863, opened 2018-06-08
Close span writer in standalone

pavolloffay:


cmd/standalone/main.go
```diff
@@ -130,6 +131,13 @@ func main() {
 
 			select {
 			case <-signalsChannel:
+				if closer, ok := spanWriter.(io.Closer); ok {
```
black-adder: what about spanReader, dependencyReader, etc. might be easier to register everything that implements the io.Closer interface during initialization and iterate over them and close them

pavolloffay: readers don't implement closer. This is just port from collector main.

PR #865, opened 2018-06-11
Process IP problem/inconsistency (#821)

### LINKED ISSUE #821: Process IP problem/inconsistency
View the process IP address in a human-readable format

## Current setup
- ES Datastore
- NodeJS client 3.9.2dev0
- Java client 0.24.0

As of now, the IP from NodeJS is displayed in Query as `-1408172028`

Here's the value in ES directly:
```
            "process": {
            "serviceName": "something-service",
            "tags": [
              {
                "key": "jaeger.version",
                "type": "string",
                "value": "Node-3.9.2dev0"
              },
              {
                "key": "jaeger.hostname",
                "type": "string",
                "value": "7fdc5f65b378"
              },
              {
                "key": "ip",
                "type": "float64",
                "value": "-1408172028"
              }
            ]
```

### Option 1:
NodeJS could report the IP as a String like the Java client. 
The value in ES would be:
```

              {
                "key": "ip",
                "type": "string",
                "value": "10.0.0.1"
              }
```

### Option 2:
Modify the Java client to report the IP in the same format as NodeJS and convert the value to a human-readable format in Query

olimpias:
Signed-off-by: cem turker <cemturker92@gmail.com>

 Resolves #821

- Process IP problem/inconsistency #821

-  Added support to process float 64 type entered ip addresses.

### DIFF
--- a/model/adjuster/ip_tag.go
+++ b/model/adjuster/ip_tag.go
@@ -34,14 +34,20 @@ func IPTagAdjuster() Adjuster {
 
 	adjustTags := func(tags model.KeyValues) {
 		for i, tag := range tags {
-			if tag.VType != model.Int64Type {
-				continue
+			var value uint32
+			switch tag.VType {
+				case model.Int64Type:
+					value = uint32(tag.Int64())
+				case model.Float64Type:
+					value = uint32(tag.Float64())
+				default:
+					continue
 			}
 			if _, ok := ipTagsToCorrect[tag.Key]; !ok {
 				continue
 			}
 			var buf [4]byte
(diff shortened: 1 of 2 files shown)


jpkrohling:
@olimpias thanks for your contribution! Would you please expand on the problem this PR is solving? Where exactly are you seeing problems? When would you receive an IP address as a float?

In the meantime, you might want also to check Travis failures. Mainly:

https://travis-ci.org/jaegertracing/jaeger/jobs/390629957#L742-L745
```
./model/adjuster/ip_tag_test.go
./model/adjuster/ip_tag.go
make: *** [lint] Error 1
```

black-adder:
thanks!

PR #868, opened 2018-06-14
Use protobuf marshaller

davit-y:
https://github.com/jaegertracing/jaeger/pull/862#discussion_r195198455
- Standard JSON encoding is not optimal for sending spans
- Thrift model doesn't have the process in the definition, therefore encoding causes loss of data when writing

- JSON Marshaller now uses protobuf to encode spans
- Thrift Marshaller has been replaced by a ProtobufMarshaller

### DIFF
--- a/plugin/storage/kafka/marshaller.go
+++ b/plugin/storage/kafka/marshaller.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 
 	"github.com/gogo/protobuf/jsonpb"
+	"github.com/gogo/protobuf/proto"
 
 	"github.com/jaegertracing/jaeger/model"
 )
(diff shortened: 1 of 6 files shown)


yurishkuro:
why not use the actual binary protobuf encoding?

plugin/storage/kafka/marshaller.go
```diff
@@ -15,44 +15,32 @@
...
 
 // Marshaller encodes a span into a byte array to be sent to Kafka
 type Marshaller interface {
 	Marshal(*model.Span) ([]byte, error)
 }
 
-type thriftMarshaller struct {
-	tProtocolFactory *thrift.TBinaryProtocolFactory
+type jsonMarshaller struct {
+	pbMarshaller jsonpb.Marshaler
```
yurishkuro: what are the considerations for caching the marshaller?  Is it more efficient to reuse the same one? Is it thread-safe?

plugin/storage/kafka/options.go
```diff
@@ -24,24 +24,18 @@ import (
 )
 
 const (
-	configPrefix   = "kafka"
-	suffixBrokers  = ".brokers"
-	suffixTopic    = ".topic"
-	suffixEncoding = ".encoding"
```
yurishkuro: not sure why you're removing the encoding param - we may only support one value at this time, but so what...

plugin/storage/kafka/marshaller.go
davit-y: Yes, the main consideration was efficiency. The struct fields are all config related and are not written to when marshalling so it is thread safe.

plugin/storage/kafka/marshaller.go
```diff
@@ -15,44 +15,41 @@
...
 }
 
-type jsonMarshaller struct{}
+type jsonMarshaller struct {
+	pbMarshaller *jsonpb.Marshaler
+}
 
 func newJSONMarshaller() *jsonMarshaller {
-	return &jsonMarshaller{}
+	return &jsonMarshaller{&jsonpb.Marshaler{}}
```
yurishkuro: Nit: This should be multi-line with named field syntax.

PR #869, opened 2018-06-14
Remove golang.org/x/sys pin

black-adder:
- In the previous PR (https://github.com/jaegertracing/jaeger/pull/862), we had to pin a repo because the builds were failing. We were unable to ascertain why it was failing. However, now that I've removed it, the builds are passing. 🤷‍♂️


isaachier:
Does this potentially relate to yesterday's Golang bug: https://github.com/golang/go/issues/25872?

PR #873, opened 2018-06-15
Add Kafka Integration Test

davit-y:
- Kafka writer did not have an integration test

### DIFF
--- a/plugin/storage/integration/kafka_test.go
+++ b/plugin/storage/integration/kafka_test.go
@@ -43,14 +43,14 @@ type KafkaIntegrationTestSuite struct {
 func (s *KafkaIntegrationTestSuite) initialize() error {
 	logger, _ := testutils.NewLogger()
 	s.logger = logger
+	// A new topic is generated per execution to avoid data overlap
 	topic := "jaeger-kafka-integration-test-" + strconv.FormatInt(time.Now().UnixNano(), 10)
 
 	f := kafka.NewFactory()
 	v, command := config.Viperize(f.AddFlags)
 	command.ParseFlags([]string{"--kafka.topic=" + topic, "--kafka.brokers=" + defaultLocalKafkaBroker})
 	f.InitFromViper(v)
-	err := f.Initialize(metrics.NullFactory, s.logger)
-	if err != nil {
+	if err := f.Initialize(metrics.NullFactory, s.logger); err != nil {
 		return err
 	}
 	spanWriter, err := f.CreateSpanWriter()
--- a/plugin/storage/integration/kafka_test.go
+++ b/plugin/storage/integration/kafka_test.go
@@ -1,0 +1,141 @@
+// Copyright (c) 2018 The Jaeger Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package integration
+
+import (
+	"encoding/json"
+	"os"
+	"strconv"
+	"testing"
+	"time"
+
+	"github.com/Shopify/sarama"
+	"github.com/stretchr/testify/require"
+	"github.com/uber/jaeger-lib/metrics"
+	"go.uber.org/zap"
+
+	"github.com/jaegertracing/jaeger/model"
+	"github.com/jaegertracing/jaeger/pkg/config"
+	"github.com/jaegertracing/jaeger/pkg/testutils"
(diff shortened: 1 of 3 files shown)


plugin/storage/integration/kafka_test.go
```diff
@@ -0,0 +1,141 @@
...
+func (r *spanReader) GetTrace(traceID model.TraceID) (*model.Trace, error) {
+	result := &model.Trace{}
+	var err error
+	doneCh := make(chan struct{})
+	go func() {
+		for {
+			select {
+			case msg := <-r.consumer.Messages():
+				newSpan := model.Span{}
+				if err = json.Unmarshal(msg.Value, &newSpan); err != nil {
```
davit-y: I will replace this line with the one below as soon as the protobuf marshalling PR is merged into master

plugin/storage/integration/kafka_test.go
```diff
@@ -0,0 +1,141 @@
...
+				//if err = jsonpb.Unmarshal(bytes.NewReader(msg.Value), &newSpan); err != nil {
+					r.logger.Error("protobuf unmarshaling error", zap.Error(err))
```
davit-y: I am explicitly logging here because the returned err is never logged upstream. Please let me know if I have incorrectly logged the error.

plugin/storage/integration/kafka_test.go
```diff
@@ -0,0 +1,141 @@
...
+	var err error
+	doneCh := make(chan struct{})
+	go func() {
+		for {
+			select {
+			case msg := <-r.consumer.Messages():
+				newSpan := model.Span{}
+				if err = json.Unmarshal(msg.Value, &newSpan); err != nil {
+					//if err = jsonpb.Unmarshal(bytes.NewReader(msg.Value), &newSpan); err != nil {
+					r.logger.Error("protobuf unmarshaling error", zap.Error(err))
```
davit-y: I am explicitly logging here because the returned err is never logged upstream. Please let me know if I have incorrectly logged the error.

plugin/storage/integration/kafka_test.go
```diff
@@ -0,0 +1,141 @@
...
+const defaultLocalKafkaBroker = "127.0.0.1:9092"
+
+type KafkaIntegrationTestSuite struct {
+	StorageIntegration
+}
+
+func (s *KafkaIntegrationTestSuite) initialize() error {
+	logger, _ := testutils.NewLogger()
+	s.logger = logger
+	topic := "jaeger-kafka-integration-test-" + strconv.FormatInt(time.Now().UnixNano(), 10)
```
black-adder: write a quick comment about why this is neccesary

plugin/storage/integration/kafka_test.go
```diff
@@ -0,0 +1,141 @@
...
+
+	f := kafka.NewFactory()
+	v, command := config.Viperize(f.AddFlags)
+	command.ParseFlags([]string{"--kafka.topic=" + topic, "--kafka.brokers=" + defaultLocalKafkaBroker})
+	f.InitFromViper(v)
+	err := f.Initialize(metrics.NullFactory, s.logger)
```
black-adder: inline with the next line

scripts/travis/kafka-integration-test.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
yurishkuro: Should there be a change to Travis file to run this as part of CI?

plugin/storage/integration/kafka_test.go
```diff
@@ -0,0 +1,141 @@
...
+		return err
+	}
+	spanReader, err := createSpanReader(topic)
+	if err != nil {
+		return err
+	}
+	s.SpanWriter = spanWriter
+	s.SpanReader = spanReader
+	s.Refresh = func() error { return nil }
+	s.CleanUp = func() error { return nil }
```
yurishkuro: Wouldn't there be overlap of testing data without clean up? Why not use a unique topic name between tests?

plugin/storage/integration/kafka_test.go
```diff
@@ -0,0 +1,141 @@
...
+	consumer sarama.PartitionConsumer
+}
+
+func createSpanReader(topic string) (spanstore.Reader, error) {
+	logger, _ := testutils.NewLogger()
+	c, err := sarama.NewConsumer([]string{defaultLocalKafkaBroker}, nil)
+	if err != nil {
+		return nil, err
+	}
+	pc, err := c.ConsumePartition(topic, 0, sarama.OffsetOldest)
```
yurishkuro: Wicked that sarama has such method :-)

plugin/storage/integration/kafka_test.go
davit-y: I appended nanoseconds in Unix time when generating the topic.

yurishkuro: But you only do it once, not for each test.

scripts/travis/kafka-integration-test.sh
```diff
@@ -0,0 +1,9 @@
...
+
+set -e
+
+docker pull spotify/kafka
```
yurishkuro: Maybe pin the version? What if the image is upgraded and our driver is not compatible.

PR #874, opened 2018-06-16
gofmt

whilei:
Just ran

	gofmt -w .

on the project root. That's all.

> https://blog.golang.org/go-fmt-your-code


black-adder:
Looks like our existing `make fmt` command missed this since https://github.com/jaegertracing/jaeger/blob/master/Makefile#L6 excludes the `example` directory

PR #876, opened 2018-06-19
Add cassandra index filter

black-adder:

- There are certain times when you want to skip indexing a certain span. ie for debug spans, the most common use case is that the user will use the traceid directly to find the trace, there's no need to index it by duration, etc.

- Add a new IndexFilter which checks the span and determines whether it should be indexed or not. Defaults to indexing everything

### DIFF
--- a/plugin/storage/cassandra/spanstore/dbmodel/index_filter.go
+++ b/plugin/storage/cassandra/spanstore/dbmodel/index_filter.go
@@ -1,0 +1,39 @@
+// Copyright (c) 2018 The Jaeger Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package dbmodel
+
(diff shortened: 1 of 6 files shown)


plugin/storage/cassandra/spanstore/dbmodel/index_filter.go
```diff
@@ -0,0 +1,39 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package dbmodel
+
+// IndexFilter filters out any spans that should not be indexed.
+type IndexFilter interface {
+	IndexByDuration(span *Span) bool
+	IndexByService(span *Span) bool
+	IndexByOperation(span *Span) bool
```
isaachier: This is OK, just not a huge fan of the hard coded function names. Constants can accomplish the same thing dynamically:
```
const (
    durationIndex = iota
    serviceIndex
    operationIndex
)

type IndexFilter func(span *Span, index int)
```

yurishkuro: fair point

plugin/storage/cassandra/spanstore/writer_test.go
```diff
@@ -314,6 +314,40 @@ func TestStorageMode_IndexOnly(t *testing.T) {
 	}, StoreIndexesOnly())
 }
 
+type indexFilter struct{}
```
jpkrohling: This filter will filter out everything, right? Would it be more appropriate to name it `noNameFilter{}`?

isaachier:
@jpkrohling this mechanism could be used to implement that feature as well.

```
type debugFilter struct
func (debugFilter) IndexByX(span *Span) bool {
    return !span.Context().(jaeger.SpanContext).IsDebug()
}
```

black-adder:
@jpkrohling later PR, I promise

PR #880, opened 2018-06-22
Ability to use multiple storage types

### LINKED ISSUE #274: Kafka for trace storage

davit-y:
- Resolves #274
- Continuation of #872

- Adding ability to specify multiple storage types
- All listed storage types are used for writing
- In the case of multiple storage types, the first is used for Reading and by the ArchiveSpanWriter


plugin/storage/factory.go
```diff
@@ -151,9 +162,13 @@ func (f *Factory) CreateArchiveSpanReader() (spanstore.Reader, error) {
 
 // CreateArchiveSpanWriter implements storage.ArchiveFactory
 func (f *Factory) CreateArchiveSpanWriter() (spanstore.Writer, error) {
-	factory, ok := f.factories[f.SpanStorageType]
+	if len(f.SpanWriterTypes) > 1 {
+		fmt.Fprintf(f.log, "WARNING: Since there are multiple span storage types,"+
```
yurishkuro: factory should have access to logger. Don't use printf.

plugin/storage/factory_config.go
```diff
@@ -22,9 +22,12 @@ import (
 )
 
 const (
-	// SpanStorageTypeEnvVar is the name of the env var that defines the type of backend used for span storage.
+	// SpanStorageTypeEnvVar is the name of the env var that defines the type of backend used for span writing.
 	SpanStorageTypeEnvVar = "SPAN_STORAGE_TYPE"
 
+	// SpanReaderTypeEnvVar is the optional env var that defines the type of backend used for span reading in the case of multi-storage.
+	SpanReaderTypeEnvVar = "SPAN_READER_TYPE"
```
yurishkuro: have you tried without introducing this new config? What's not working without it?

plugin/storage/factory_config.go
```diff
@@ -33,12 +36,14 @@ const (
 
 // FactoryConfig tells the Factory which types of backends it needs to create for different storage types.
 type FactoryConfig struct {
-	SpanStorageType         string
+	SpanWriterTypes         []string
+	SpanReaderType          string
 	DependenciesStorageType string
+	log                     io.Writer
```
yurishkuro: ?  

we use `*zap.Logger` for logging.

plugin/storage/factory_config.go
davit-y: Since we can't have a CompositeReader, in the case of multiple storage types we must decide on 1 storage to read from. At the moment, the first listed storage type is used and a warning is emitted, but the extra flag allows users to specify the storage so that they don't have to be concerned with the order of their inputted storage types.

yurishkuro: > but the extra flag allows users to specify the storage so that they don't have to be concerned with the order of their inputted storage types.

I suggest removing this optional behavior - the same can be achieved by reordering items in the main storage type var without additional code.

plugin/storage/factory_config.go
```diff
@@ -47,39 +52,48 @@ type FactoryConfig struct {
...
 		if !strings.HasPrefix(token, spanStorageFlag) {
 			continue
 		}
-		fmt.Fprintf(
-			log,
-			"WARNING: found deprecated command line option %s, please use environment variable %s instead\n",
-			token,
-			SpanStorageTypeEnvVar,
-		)
+		logger.Warn("found deprecated command line option " + token +
```
yurishkuro: sorry, I didn't realize we already use this io.Writer. Let's continue using it instead of creating redundant logger in all mains.

plugin/storage/factory_config.go
black-adder: > the same can be achieved by reordering items in the main storage type var without additional code.

I'm not a fan of this non user friendly behavior where the user has to know the ordering of the flags, the new variable makes it explicit.

black-adder: gonna be a pedant and not forward this conversation in anyway; the ingester will read from kafka and write to kafka/cassandra, this won't be just a marginal use case. As a user, I'd prefer a second explicit flag.

plugin/storage/factory_config.go
```diff
@@ -56,13 +57,19 @@ func FactoryConfigFromEnvAndCLI(args []string, log io.Writer) FactoryConfig {
...
 	}
-	depStoreType := os.Getenv(DependencyStorageTypeEnvVar)
-	if depStoreType == "" {
-		depStoreType = spanStorageType
+	spanWriterTypes := strings.Split(spanStorageType, ",")
+	if len(spanWriterTypes) > 1 {
+		fmt.Fprintf(log, "WARNING: Since multiple storage types have been listed,"+
+			"the first span storage type listed (%s) will be used for reading and archiving.", spanWriterTypes[0])
+	}
+	depStorageType := os.Getenv(DependencyStorageTypeEnvVar)
```
jpkrohling: Should this come from viper?

davit-y: All the main.go files initialize viper after setting up this initial factory.

plugin/storage/factory_config.go
```diff
@@ -56,13 +57,19 @@ func FactoryConfigFromEnvAndCLI(args []string, log io.Writer) FactoryConfig {
 	if spanStorageType == "" {
 		spanStorageType = cassandraStorageType
 	}
-	depStoreType := os.Getenv(DependencyStorageTypeEnvVar)
-	if depStoreType == "" {
-		depStoreType = spanStorageType
+	spanWriterTypes := strings.Split(spanStorageType, ",")
+	if len(spanWriterTypes) > 1 {
+		fmt.Fprintf(log, "WARNING: Since multiple storage types have been listed,"+
```
vprithvi: It seems like this warning could be easily missed. 
Is it difficult to require the user to explicitly specify a read and archive storage type when multiple storage types are specified?

plugin/storage/factory_config.go
jpkrohling: If the factories can be created within the Cobra's command, it can also receive a `zap.Logger`, avoiding the usage of `fmt.Fprintf`

yurishkuro: The reason is to avoid each storage type adding it's flags to cli

jpkrohling: That affects only the contents of `--help`, right?

davit-y:
#880 had commits at some point that partially solve this issue, but the creation of the ingester will create further complications so we decided to only do the basic set of requirements for the Kafka writer for now.

PR #881, opened 2018-06-22
Make new proto RefType backwards compatible in cassandra

black-adder:
- The recent PR that changed the domain model to use proto added this change: https://github.com/jaegertracing/jaeger/pull/856/files#diff-96d29428fedf19815cb3671bba06767eL33. The issue is that previously we were storing "child-of" instead of "CHILD_OF" and doing a simple strings.Upper() isn't going to help.

With this fix, cassandra dbmodel will use it's own version of RefTypes so it's not affected by any outside changes.

-

### DIFF
--- a/plugin/storage/cassandra/spanstore/dbmodel/converter.go
+++ b/plugin/storage/cassandra/spanstore/dbmodel/converter.go
@@ -16,8 +16,21 @@ package dbmodel
 
 import (
 	"errors"
+	"fmt"
 
 	"github.com/jaegertracing/jaeger/model"
+)
+
+var (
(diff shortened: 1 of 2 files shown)


plugin/storage/cassandra/spanstore/dbmodel/converter.go
```diff
@@ -16,10 +16,23 @@ package dbmodel
 
 import (
 	"errors"
+	"fmt"
 
 	"github.com/jaegertracing/jaeger/model"
 )
 
+var (
+	dbToProtoRefMap = map[string]model.SpanRefType{
```
yurishkuro: s/proto/domain/

Issue #896, opened 2018-06-29
badly formatted logs on stderr


tmszdmsk:
## Requirement - what kind of business use case are you trying to solve?

Logs printed on stderr are not properly formatted. It slows debugging problems with `jaeger-collector`

## Problem - what in Jaeger blocks you from solving the requirement?

Our `jaeger-collector` is deployed on Kubernetes. Logs from stdout/stderr are automatically collected and pushed into ELK stack. Because of badly formatted JSON Logstash is unable to parse the logs.

example (a little bit anonymised) `kubectl logs` output from `jaeger-collector` pod.
```
{\"traceID\":\"11a4ea9bddefe82\",\"spanID\":\"1e9c810f2c768990\",\"flags\":1,\"operationName\”:\”***\”,\”references\":[{\"refType\":\"CHILD_OF\",\"traceID\":\"11a4ea9bddefe82\",\"spanID\":\"11a4ea9bddefe82\"}],\"startTime\":1530235009818889,\"duration\":884,\"tags\":[{\"key\":\"span.kind\",\"type\":\"string\",\"value\":\"client\"},{\"key\”:\”***\”,\”type\":\"string\",\"value\”:\”***\”},{\”key\":\"backend.name\",\"type\":\"string\",\"value\":\"***\"},{\"key\":\"http.method\",\"type\":\"string\",\"value\":\"GET\"},{\"key\":\"http.url\",\"type\":\"string\",\"value\":\"***\"},{\"key\":\"http.host\",\"type\":\"string\",\"value\":\"***\"},{\"key\":\"http.status_code\",\"type\":\"int64\",\"value\":\"200\"}],\"logs\":[],\"processID\":\"\",\"process\":{\"serviceName\":\"***\",\"tags\":[{\"key\":\"jaeger.version\",\"type\":\"string\",\"value\":\"Go-2.9.0\"},{\"key\":\"hostname\",\"type\":\"string\",\"value\":\"***\"},{\"key\":\"ip\",\"type\":\"string\",\"value\":\"***\"}]},\"warnings\":null,\"startTimeMillis\":1530235009818}\n
```
As you can see the JSON is escaped so that it is not interpreted as a JSON.

jaeger-collector v1.5.0

## Proposal - what do you suggest to solve the problem or improve the existing situation?

Properly formatted JSON on `stderr`

## Any open questions to address
N/A

jpkrohling:
[...]

PR #902, opened 2018-06-30
ES storage: logging number of total and failed requests

tmszdmsk:
Plz, review it as I am not golang native ;)

- #896 

- for more debug info, number of failed and total requests in bulk in case of error

### DIFF
--- a/pkg/es/config/config.go
+++ b/pkg/es/config/config.go
@@ -79,8 +79,14 @@ func (c *Configuration) NewClient(logger *zap.Logger, metricsFactory metrics.Fac
 			sm.Emit(err, duration)
 
 			if err != nil {
+				failed := 0
+				if response.Failed() != nil {
+					failed = len(response.Failed())
+				}
+				total := len(requests)
 				logger.Error("Elasticsearch could not process bulk request", zap.Error(err),
-					zap.Any("response", response))
+					zap.Any("response", response), zap.Int("failed_number", failed),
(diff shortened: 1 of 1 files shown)


tmszdmsk:
hm, I've noticed that we are already logging response. Should we remove it also?

pkg/es/config/config.go
```diff
@@ -79,8 +79,14 @@ func (c *Configuration) NewClient(logger *zap.Logger, metricsFactory metrics.Fac
 			sm.Emit(err, duration)
 
 			if err != nil {
+				failed := 0
+				if response.Failed() != nil {
+					failed = len(response.Failed())
+				}
```
yurishkuro: `failed := len(response.Failed())` is sufficient, since `len(nil) == 0`

pkg/es/config/config.go
```diff
@@ -79,8 +79,14 @@ func (c *Configuration) NewClient(logger *zap.Logger, metricsFactory metrics.Fac
...
+				total := len(requests)
 				logger.Error("Elasticsearch could not process bulk request", zap.Error(err),
-					zap.Any("response", response))
+					zap.Any("response", response), zap.Int("failed_number", failed),
```
yurishkuro: - let's put one arg per line, including error
- rename to request_count and failed_count
- I would order the fields: request_count, failed_count, error, response

PR #903, opened 2018-06-30
WIP: Sample gRPC server

yurishkuro:
blocked on #927


examples/proto/main.go
```diff
@@ -0,0 +1,135 @@
...
+	lis, err := net.Listen("tcp", addr)
+	if err != nil {
+		log.Fatalln("Failed to listen:", err)
+	}
+	s := grpc.NewServer(
+	// grpc.Creds(credentials.NewServerTLSFromCert(&insecure.Cert)),
+	// grpc.UnaryInterceptor(grpc_validator.UnaryServerInterceptor()),
+	// grpc.StreamInterceptor(grpc_validator.StreamServerInterceptor()),
+	)
+	api_v2.RegisterQueryServiceServer(s, server.New())
```
jpkrohling: Just wanted to mention that it's probably a good idea to have each service on its own port, so that different components can have different constraints. Like, I might want to expose the Query Service to the public internet but not the Collector.

pavolloffay:
done in #1165

Issue #897, opened 2018-06-29
Jaeger-collector stops storing traces


tmszdmsk:
## Requirement - what kind of business use case are you trying to solve?

I want to be sure `jaeger-collector` collects and persists data

## Problem - what in Jaeger blocks you from solving the requirement?

We've started using Jaeger in production 5 days ago. Since we started sending traces from 60+ services it stopped working twice. I mean by that that traces that are sent to this particular collector are no longer available in `jaeger-query`. We have 4 collectors (on prod & dev). All use the same Elasticsearch cluster to store their traces. Still, only those handling production traffic break. Dev traces go through.

Despite not very readable logs on `stderr` (#896) I was able to fish out some stacktrace which may be helpful. Appearance of logs on `stderr` correlates with this collector not working. There are no logs on `stderr` or `stdout` if it operates normally.
```
"stacktrace":"github.com/jaegertracing/jaeger/pkg/es/config.(*Configuration).NewClient.func2\n\t/home/travis/gopath/src/github.com/jaegertracing/jaeger/pkg/es/config/config.go:90\ngithub.com/jaegertracing/jaeger/vendor/gopkg.in/olivere/elastic%2ev5.(*bulkWorker).commit\n\t/home/travis/gopath/src/github.com/jaegertracing/jaeger/vendor/gopkg.in/olivere/elastic.v5/bulk_processor.go:506\ngithub.com/jaegertracing/jaeger/vendor/gopkg.in/olivere/elastic%2ev5.(*bulkWorker).work\n\t/home/travis/gopath/src/github.com/jaegertracing/jaeger/vendor/gopkg.in/olivere/elastic.v5/bulk_processor.go:442"
```

jaeger-collector v1.5.0 deployed on Kubernetes cluster

Pod restart resolves the issue for some time

## Proposal - what do you suggest to solve the problem or improve the existing situation?

N/A

## Any open questions to address

N/A

jpkrohling:
@jkandasa, are we using Elasticsearch for the soak tests? For how long are we running those?

tmszdmsk:
this might add some more insight: https://github.com/jaegertracing/jaeger/issues/896#issuecomment-401335249

tmszdmsk:
After some digging it looks like the culprit is 
```
github.com/jaegertracing/jaeger/pkg/es/config.(*Configuration).NewClient.func2
	/home/travis/gopath/src/github.com/jaegertracing/jaeger/pkg/es/config/config.go:90
[...]

Issue #779, opened 2018-04-20
Collector (ES Storage) error when trying to store trace with long span names


bharat-p:
Observed following exception in Collector logs using Elasticsearch as backend when span names are really long:

```
{"level":"error","ts":1524250693.569798,"caller":"config/config.go:90","msg":"Elasticsearch could not process bulk request","error":"elastic: Error 400 (Bad Request): Validation Failed: 1: id is too long, must be no longer than 512 bytes but was: 541; [type=action_request_validation_exception]","response":null,"requests":"{\"index\":{\"_id\":\"jaeger-test|start-515-vdadclohlswrzxxbdmsahpfhxcqjleonjkymoiqaixglisojzkkiajlrhdgmtwxmttoezjjkypzpisvzkeuysgurctzbzmhhbcgnnbgzswnwbwlazjznggagnktuvdbsskolpdmmfygajbuquqadvhipybykvwifkabdjgpxtjoggbeggprfzvaahmapprdpeveynztdqtjbrbsdpptrecqhsisnznlaocytmfcxvuteguihobetxaxrgxlkqjydrxrywsholbenemazkexfhlxzrdtogzvvgtgvtloclkalhkordtoxbynqcdtxfinofpmndblncheupynxjwgplmdddoppsjlltxkvtbbjszaogvhrhrehkdebhbrucrfwkhuglifmfjvrpyrysptsaahocxfkfalbdgdyldyeshluxtoiayrvopbuzgguhqczibibuhxmluxhygsztfqpqffcgdkaidkbiriawdfgpjlopbemaksvlhpdnynstlkqmyn-end\",\"_index\":\"jaeger-service-2018-04-20\",\"_type\":\"service\"}}
[...]

PR #905, opened 2018-07-02
don't exceed ES _id length limit

tmszdmsk:
- #779 
- #897 

- let's store a hash of the serviceId so that we don't exceed length limit of `_id`


plugin/storage/es/spanstore/service_operation.go
```diff
@@ -16,6 +16,7 @@ package spanstore
 
 import (
 	"context"
+	"crypto/sha256"
```
yurishkuro: there's no need for crypto hash, see https://github.com/jaegertracing/jaeger/blob/d16de316a8bfc0ba7a6d5d1b7e43d3c48156d8ae/model/hash.go#L30-L36

plugin/storage/es/spanstore/service_operation.go
```diff
@@ -75,9 +76,10 @@ func (s *ServiceOperationStorage) Write(indexName string, jsonSpan *jModel.Span)
 		OperationName: jsonSpan.OperationName,
 	}
 	serviceID := fmt.Sprintf("%s|%s", service.ServiceName, service.OperationName)
+	hashedID := fmt.Sprintf("%x", sha256.Sum256([]byte(serviceID)))
 	cacheKey := fmt.Sprintf("%s:%s", indexName, serviceID)
```
yurishkuro: I don't see why we cannot use the hash for both document ID and cacheKey. Let's just have a single value.

harnash:
@yurishkuro thanks for the comments. I hopefully addressed them all in 10fcf67

plugin/storage/es/spanstore/service_operation_test.go
```diff
@@ -31,9 +32,17 @@ func TestWriteService(t *testing.T) {
 		indexService := &mocks.IndexService{}
 
 		indexName := "jaeger-1995-04-21"
+		service := Service{
+			ServiceName:   "service",
+			OperationName: "operation",
+		}
+
+		serviceHash, err := model.HashCode(service)
```
yurishkuro: this introduces unnecessary public method to Service type. Instead I would simply add something like this:
```
func (s Service) hashCode() string {
    h := fnv.New64a()
    h.Write(s.ServiceName)
    h.Write(s. OperationName)
    return fmt.Sprintf("%x", h.Sum64())
}
```
```

File: plugin/storage/es/spanstore/writer.go [hunk 4]
```diff
@@ -73,6 +75,13 @@ type Span struct {
 	StartTimeMillis uint64 `json:"startTimeMillis"`
 }
 
+func (s Service) hashCode() string {
+	h := fnv.New64a()
```
black-adder: can this be cached?

plugin/storage/es/spanstore/service_operation_test.go
```diff
@@ -31,9 +31,15 @@ func TestWriteService(t *testing.T) {
...
+		service := Service{
+			ServiceName:   "service",
+			OperationName: "operation",
+		}
+		serviceHash := service.hashCode()
+
 		indexService.On("Index", stringMatcher(indexName)).Return(indexService)
 		indexService.On("Type", stringMatcher(serviceType)).Return(indexService)
-		indexService.On("Id", stringMatcher("service|operation")).Return(indexService)
+		indexService.On("Id", stringMatcher(serviceHash)).Return(indexService)
```
black-adder: I'm not a big fan of using the same hash function here to test that the service was hashed, I'd prefer if you generated this hash once and compare against the actual hash string.

harnash: I've removed calls to `hashCode()` and put static hash in those tests.

plugin/storage/es/spanstore/writer.go
harnash: @black-adder It could be but it will introduce issue with concurrency. Every simultaneous call to this function would potentially cause race condition and would render unpredictable hash results. I would need to introduce some kind of semaphore and mark this part as critical section. Do you think it is worth?

harnash: @black-adder so my reasoning is like this.
1. You store cached hasher object somewhere (most likely in the writer.go) and use it whenever someone calls `hashCode()`
2. First process (1) calls `hashCode()`, fetches (creates if not cached) hasher object
3. (1) Pushes data: `OperationName` and `ServiceName` to the hasher
4. Second process (2) calls `hashCode()` which fetches the same (cached) hasher object
5. (2) pushes its `OperationName` and `ServiceName` to the same hasher
6. (1) calculates checksum by calling `Sum64()` on the same hasher (which now holds data from two possibly different instances of `Service`)
7. (2) does the calculations too and gets the same (wrong) hash as (1)

I did actually run into this issue while fixing the tests :-) 

Even if I introduce `hash.Reset()` just after `Sum64()` or before data is pushed it the hasher I still have no guarantees that other process is not doing some operation on the cached object. Putting Write semaphore would probably do the trick but I think it is a bit overkill in this case.

I hope this helps.

PR #909, opened 2018-07-03
Require 100% coverage

yurishkuro:
Our codecov.io status no longer blocks PRs with less than 100% code coverage. Trying to fix that.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,12 +1,15 @@
-coverage:
-  range: 100..100
-  round: down
-  precision: 2
-
 codecov:
   notify:
     require_ci_to_pass: yes
+  strict_yaml_branch: master  # only use the latest copy on master branch
 
-# see https://docs.codecov.io/docs/ignoring-paths
-ignore:
-  - "model.pb*.go"
+coverage:
+  precision: 2
+  round: down
+  range: "100...100"
+
+  status:
+    project: yes
+    patch: yes
+    changes: no
+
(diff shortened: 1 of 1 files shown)


.codecov.yml
```diff
@@ -1,12 +1,15 @@
...
-
 codecov:
   notify:
     require_ci_to_pass: yes
+  strict_yaml_branch: master  # only use the latest copy on master branch
+
+coverage:
+  precision: 2
+  round: down
+  range: "100...100"
```
vprithvi: This is only used for displaying coverage

.codecov.yml
```diff
@@ -1,12 +1,15 @@
...
+
+  status:
+    project: yes
+    patch: yes
+    changes: no
```
vprithvi: afaik you'll need to add a `target: 100%` to fail the build

vprithvi: Apologies, see https://docs.codecov.io/docs/commit-status#section-target

yurishkuro: how's this?

yurishkuro:
@vprithvi  So this build is still not failing, even though codecov/project — 99.96%

It might be because `strict_yaml_branch: master  # only use the latest copy on master branch`, not sure.

Actually, this file is invalid, doh...
```
$ curl --data-binary @.codecov.yml https://codecov.io/validate
mapping values are not allowed here
  in "<string>", line 13, column 14:
          default:
```

PR #911, opened 2018-07-05
connCheckTimeout as flag

Henrod:
Adds a option to set connCheckTimeout. Now the value is hardcoded at 250ms. 

Added a flag to accept a new timeout for connCheckTimeout (--discovery.conn-check-timeout) that receives a duration. Defaults to 250ms.


black-adder:
thanks!

PR #912, opened 2018-07-05
Add TunnelConnection param to cassandra config for local development …

whistlinwilly:
…over SSH tunnel

- gocql autodiscovery breaks connecting to cassandra over an SSH tunnel when developing locally, this PR adds a configurable flag to disable all relevant gocql autodiscovery features.

- New bool config parameter and associated options set in cluster config


pkg/cassandra/config/config.go
```diff
@@ -142,6 +143,11 @@ func (c *Configuration) NewCluster() *gocql.ClusterConfig {
 			EnableHostVerification: c.TLS.EnableHostVerification,
 		}
 	}
+	// If tunneling connection to C*, disable cluster autodiscovery features.
+	if c.TunnelConnection {
```
black-adder: This is a bit bespoke, I'd rather make public DisableInitialHostLookup and IgnorePeerAddr or rename this variable as `DisableAutodiscovery`. Let's go with the latter.

pkg/cassandra/config/config.go
```diff
@@ -27,17 +27,18 @@ import (
...
+	Keyspace             string        `validate:"nonzero"`
+	ConnectionsPerHost   int           `validate:"min=1" yaml:"connections_per_host"`
+	Timeout              time.Duration `validate:"min=500"`
+	SocketKeepAlive      time.Duration `validate:"min=0" yaml:"socket_keep_alive"`
+	MaxRetryAttempts     int           `validate:"min=0" yaml:"max_retry_attempt"`
+	ProtoVersion         int           `yaml:"proto_version"`
+	Consistency          string        `yaml:"consistency"`
+	Port                 int           `yaml:"port"`
+	Authenticator        Authenticator `yaml:"authenticator"`
+	DisableAutodiscovery bool          `yaml:"disable_autodiscovery"`
```
yurishkuro: DisableAutoDiscovery

yurishkuro: and `disable_auto_discovery`

PR #915, opened 2018-07-08
Improve HotROD demo

yurishkuro:
* Allow controlling optimizations from cmd line
* Suppress logging of stack traces unless fatal
* Add `hotrod` namespace to all metrics
* Support HTTP or UDP senders

On the topic of senders, the Go client, unfortunately, does not support JAEGER_ENDPOINT env variable like Java client does, and the current CLI argument is called `--jaeger-agent.host-port`, even though it now accepts HTTP endpoint name.

### DIFF
--- a/examples/hotrod/cmd/root.go
+++ b/examples/hotrod/cmd/root.go
@@ -24,6 +24,7 @@ import (
 	jexpvar "github.com/uber/jaeger-lib/metrics/expvar"
 	jprom "github.com/uber/jaeger-lib/metrics/prometheus"
 	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
 )
 
 var (
@@ -35,7 +36,7 @@ var (
 
 // RootCmd represents the base command when called without any subcommands
 var RootCmd = &cobra.Command{
-	Use:   "jaeger-demo",
+	Use:   "examples-hotrod",
 	Short: "HotR.O.D. - A tracing demo application",
 	Long:  `HotR.O.D. - A tracing demo application.`,
 }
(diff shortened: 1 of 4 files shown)


examples/hotrod/cmd/root.go
```diff
@@ -51,23 +58,38 @@ func Execute() {
 
 func init() {
 	RootCmd.PersistentFlags().StringVarP(&metricsBackend, "metrics", "m", "expvar", "Metrics backend (expvar|prometheus)")
-	RootCmd.PersistentFlags().StringVarP(&jAgentHostPort, "jaeger-agent.host-port", "a", "0.0.0.0:6831", "String representing jaeger-agent host:port")
+	RootCmd.PersistentFlags().StringVarP(&jAgentHostPort, "jaeger-agent.host-port", "a", "0.0.0.0:6831", "String representing jaeger-agent UDP host:port")
```
black-adder: should this also say that http host:port is accepted as well?

yurishkuro: we could say that, my main hesitation was that the property is called `jaeger-agent.host-port`, whereas an HTTP endpoint would be running in the collector, not the agent.

black-adder: doh, in that case, should we just add a new flag for jaeger-client.reporting-host-port? and say the old one is deprecated? it is a demo app, not sure if many would care.

yurishkuro: for now I added http to the description. I would rather not create another flag, as I think there's a better way - #920

examples/hotrod/services/config/config.go
```diff
@@ -33,18 +34,19 @@ var (
 	// MySQLGetDelayStdDev is standard deviation
 	MySQLGetDelayStdDev = MySQLGetDelay / 10
 
-	// RouteWorkerPoolSize is the size of the worker pool used to query `route` service
-	RouteWorkerPoolSize = 3
+	// MySQLMutexDisabled controls whether there is a mutex guarding db query execution.
+	// When not disabled it simulates a misconfigured connection pool of size 1.
+	MySQLMutexDisabled = false
```
jpkrohling: Double negatives are confusing :-) How about `MySQLMutexEnabled = true` ?

yurishkuro: This isn't double negative, but more importantly disabled plays nicer with false as the default value.

PR #916, opened 2018-07-08
Detect HTTP payload format from Content-Type

yurishkuro:
Since Zipkin HTTP endpoint has been moved to another port, the only format that Collector HTTP handler recognizes is Thrift, which makes the `?format=` URL parameter redundant since the payload type can be inferred from the `Content-Type` header.

This change removes the `format` URL parameter and only looks at the `Content-Type`.

### DIFF
--- a/cmd/collector/app/http_handler.go
+++ b/cmd/collector/app/http_handler.go
@@ -18,7 +18,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
-	"strings"
 	"time"
 
 	"github.com/apache/thrift/lib/go/thrift"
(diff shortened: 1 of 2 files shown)


cmd/collector/app/http_handler.go
```diff
@@ -29,11 +28,17 @@ import (
 )
 
 const (
-	formatParam = "format"
 	// UnableToReadBodyErrFormat is an error message for invalid requests
 	UnableToReadBodyErrFormat = "Unable to process request body: %v"
 )
 
+var (
```
jpkrohling: For my own improvement: why the parenthesis?

yurishkuro: No strong reason, it looks consistent with const block as declarative section.

PR #917, opened 2018-07-08
Include HotROD binary in the distributions

yurishkuro:
Along with #915 this allows running the whole HotROD demo without recompiling the service, and without Docker.

Also, replaces go-bindata with rakyl/statik, which is simpler to use.


Makefile
```diff
@@ -208,15 +207,15 @@ docker: build_ui docker-no-ui
 
 .PHONY: build-binaries-linux
 build-binaries-linux:
-	GOOS=linux $(MAKE) build-agent build-collector build-query build-all-in-one
+	GOOS=linux $(MAKE) build-agent build-collector build-query build-all-in-one build-examples
```
black-adder: can these commands be DRY'd? looks like the only different is the `GOOS`

Makefile
```diff
@@ -157,19 +157,18 @@ install-glide:
...
-	rm ./examples/hotrod/services/frontend/bindata.go
-	CGO_ENABLED=0 GOOS=linux installsuffix=cgo go build -o ./examples/hotrod/hotrod-linux ./examples/hotrod/main.go
+build-examples: install-statik
+	(cd examples/hotrod/services/frontend/ && statik -f --src web_assets)
+	CGO_ENABLED=0 installsuffix=cgo go build -o ./examples/hotrod/hotrod-$(GOOS) ./examples/hotrod/main.go
 
 .PHONE: docker-hotrod
-docker-hotrod: build-examples
+docker-hotrod:
+	GOOS=linux $(MAKE) build-examples
```
jpkrohling: This sounds a bit odd. What was wrong with the previous approach?

yurishkuro: previously build-examples target didn't work for multiple architectures. Now it is in line with other `build-{binary}` targets.

jpkrohling: I see. And because it's Docker, `GOOS=linux` is appropriate. Makes sense, thanks for the explanation.

PR #918, opened 2018-07-08
Package static assets inside the query-service binary

### LINKED ISSUE #609: Package static assets inside the query-service binary

yurishkuro:
* Resolves #609 
* It is still possible to pass the static assets directory path via cmd line switch, but now it is optional

The Makefile target `build_ui` runs an additional command to package NPM-generated assets into a single Go file. The resulting file is huge (13Mb), so it is `.gitignore`-ed to prevent accidental check-ins. Instead, another placeholder file is generated and checked-in that packages a dummy index.html. When the build tag `ui` is provided, the real assets are used, cf.

```bash
$ go run ./cmd/standalone/main.go
$ go run -tags ui ./cmd/standalone/main.go
```

Another way to run with the real UI is by passing the assets path via `--query.static-files` flag:

```bash
$ go run ./cmd/standalone/main.go --query.static-files jaeger-ui/build
```


cmd/query/app/static_handler.go
```diff
@@ -68,15 +67,13 @@ type StaticAssetsHandlerOptions struct {
...
-		staticAssetsRoot = staticAssetsRoot + "/"
+	assetsFS, _ := fs.New()
+	if staticAssetsRoot != "" {
+		assetsFS = http.Dir(staticAssetsRoot)
 	}
-	indexBytes, err := ioutil.ReadFile(staticAssetsRoot + "index.html")
+	indexBytes, err := loadIndexHTML(assetsFS.Open)
 	if err != nil {
-		return nil, errors.Wrap(err, "Cannot read UI static assets")
+		return nil, errors.Wrap(err, "Cannot load index.html")
```
black-adder: loadIndexHTML seems to return errors that are sufficient, don't think the wrap is necessary.

cmd/query/Dockerfile
```diff
@@ -7,5 +7,6 @@ COPY --from=certs /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certifica
 
 EXPOSE 16686
 COPY query-linux /go/bin/
-ADD jaeger-ui-build /go/jaeger-ui/
-CMD ["/go/bin/query-linux", "--query.static-files=/go/jaeger-ui/"]
+
+# TODO replace this with ENTRYPOINT
```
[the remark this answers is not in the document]
yurishkuro: I did at first, but it would be a breaking change (existing docker-compose, helm/kubernetes configs won't work), so I think it's better to do for all binaries at once (#815).

yurishkuro:
> I'd rather have it now 

ok, done and updated the description.

scripts/travis/package-deploy.sh
```diff
@@ -62,12 +60,6 @@ rm -rf deploy $DEPLOY_STAGING_DIR
 mkdir deploy
 mkdir $DEPLOY_STAGING_DIR
 
-INDEX_HTML=jaeger-ui-build/build/index.html
-if [ ! -f $INDEX_HTML ]; then
-    echo "Cannot find UI assets, e.g. $INDEX_HTML. Aborting."
-    exit 1
-fi
-
```
yurishkuro: no longer needed since compile will fail if the ui packaged file is missing

PR #926, opened 2018-07-10
Fix deploy step

### LINKED ISSUE #919: Version 1.6 countdown
I would like to include the following PRs before we release 1.6:

* [x] Include HotROD binary in the distributions #917 
* [x] Improve HotROD demo #915
* [x] Package static assets inside the query-service binary #918
* [x] Update Dockerfiles to reference executable via ENTRYPOINT #815 
* [x] ES 6.x index race condition https://github.com/jaegertracing/jaeger/pull/921

yurishkuro:
Resolves #919


yurishkuro:
hard to test outside of master, but doesn't seem to fail anymore on parsing:
```
0.01s$ if [[ "$DEPLOY" == true && "$BRANCH" == "master" ]]; then make build-all-platforms ; else echo 'skipping linux'; fi
skipping linux
The command "if [[ "$DEPLOY" == true && "$BRANCH" == "master" ]]; then make build-all-platforms ; else echo 'skipping linux'; fi" exited with 0.
```

PR #930, opened 2018-07-12
Change Composite Writer to Sequential Writer

davit-y:
- #929 

- As part of the necessary changes to add the Ingester, the Composite Writer will fail at the first error when writing spans as opposed to continuing with the next storage type.
- Consider the case where the second storage type is not idempotent, and the first storage type fails to process a span. With a Composite Writer, on every retry, the second storage will have redundant spans written to it. With a Sequential Writer, the second storage type won't be attempted until the write to the first storage type succeeds.

### DIFF
--- a/storage/spanstore/composite.go
+++ b/storage/spanstore/composite.go
@@ -16,7 +16,6 @@ package spanstore
 
 import (
 	"github.com/jaegertracing/jaeger/model"
-	"github.com/jaegertracing/jaeger/pkg/multierror"
 )
 
 // CompositeWriter is a span Writer that tries to save spans into several underlying span Writers
@@ -31,13 +30,12 @@ func NewCompositeWriter(spanWriters ...Writer) *CompositeWriter {
 	}
 }
 
-// WriteSpan calls WriteSpan on each span writer. It will sum up failures, it is not transactional
+// WriteSpan calls WriteSpan on each span writer. It will fail at the first error
 func (c *CompositeWriter) WriteSpan(span *model.Span) error {
-	var errors []error
 	for _, writer := range c.spanWriters {
 		if err := writer.WriteSpan(span); err != nil {
-			errors = append(errors, err)
+			return err
 		}
 	}
-	return multierror.Wrap(errors)
+	return nil
 }
--- a/storage/spanstore/composite.go
+++ b/storage/spanstore/composite.go
@@ -18,20 +18,20 @@ import (
 	"github.com/jaegertracing/jaeger/model"
 )
 
-// CompositeWriter is a span Writer that tries to save spans into several underlying span Writers
-type CompositeWriter struct {
+// SequentialWriter is a span Writer that tries to save spans into several underlying span Writers
+type SequentialWriter struct {
 	spanWriters []Writer
 }
 
-// NewCompositeWriter creates a CompositeWriter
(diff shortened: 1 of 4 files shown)


storage/spanstore/composite.go
```diff
@@ -16,28 +16,26 @@ package spanstore
 
 import (
 	"github.com/jaegertracing/jaeger/model"
-	"github.com/jaegertracing/jaeger/pkg/multierror"
 )
 
-// CompositeWriter is a span Writer that tries to save spans into several underlying span Writers
-type CompositeWriter struct {
+// SequentialWriter is a span Writer that tries to save spans into several underlying span Writers
```
yurishkuro: Expression error semantics here. End sentences with periods.

storage/spanstore/composite_test.go
```diff
@@ -40,16 +39,16 @@ func (n *noopWriteSpanStore) WriteSpan(span *model.Span) error {
 }
 
 func TestCompositeWriteSpanStoreSuccess(t *testing.T) {
```
yurishkuro: Make test names match the type name

storage/spanstore/composite_test.go
```diff
@@ -40,16 +39,16 @@ func (n *noopWriteSpanStore) WriteSpan(span *model.Span) error {
...
-	c := NewCompositeWriter(&noopWriteSpanStore{}, &noopWriteSpanStore{})
+	c := NewSequentialWriter(&noopWriteSpanStore{}, &noopWriteSpanStore{})
 	assert.NoError(t, c.WriteSpan(nil))
 }
 
 func TestCompositeWriteSpanStoreSecondFailure(t *testing.T) {
-	c := NewCompositeWriter(&errProneWriteSpanStore{}, &errProneWriteSpanStore{})
-	assert.EqualError(t, c.WriteSpan(nil), fmt.Sprintf("[%s, %s]", errIWillAlwaysFail, errIWillAlwaysFail))
+	c := NewSequentialWriter(&errProneWriteSpanStore{}, &errProneWriteSpanStore{})
+	assert.EqualError(t, c.WriteSpan(nil), errIWillAlwaysFail.Error())
```
yurishkuro: I don't think you can correctly test this way because you can't tell the difference between which writer returns the error. I suggest to change noop writer to take an optional err parameter that it always returns, and do a data driven test where you pass different permutations of 3 noop writers with and without errors and check for the right error in the end.

jpkrohling:
> Consider the case where the second storage type is not idempotent, and the first storage type fails to process a span. With a Composite Writer, on every retry, the second storage will have redundant spans written to it. With a Sequential Writer, the second storage type won't be attempted until the write to the first storage type succeeds.

What about the second of three writers fail?

vprithvi:
>This way you'll preserve the old opportunistic behavior.

Do we require this opportunistic behavior?

I like the name `SequentialWriter` or even `SequentialFailFastWriter` better because it is more descriptive. While `CompositeWriter` is implemented sequentially, the name itself doesn't allude to that behavior (A writer with the same name may write parallely for e.g.)

davit-y:
Changing the CompositeWriter was found to be unnecessary to complete the Ingester. In addition, the use case of a SequentialWriter is not yet well defined

PR #934, opened 2018-07-13
Cassandra reconnect interval

### LINKED ISSUE #767: Jaeger fragile in the face of cassandra failures
We've been observing Jaeger components - specifically the collectors and query - fail to recover after Cassandra has any sort of outage. While Cassandra reliability is clearly not a topic for here ;), Jaeger's resilience to issues is.

What we see happen is this a stuck process which is only logging one thing even hours after the issue is fixed... ```gocql: no hosts available in the pool```

What we'd like to see happen is a recovery after a few minutes without manual intervention. This is related to, but distinct from, #562

nyanshak:
- Resolves #767 

- Adds support for configuring gocql's `ReconnectInterval` setting. This will make gocql attempt to reconnect to downed Cassandra nodes every interval.

### DIFF
--- a/pkg/cassandra/config/config.go
+++ b/pkg/cassandra/config/config.go
@@ -31,6 +31,7 @@ type Configuration struct {
 	Keyspace             string        `validate:"nonzero"`
 	ConnectionsPerHost   int           `validate:"min=1" yaml:"connections_per_host"`
 	Timeout              time.Duration `validate:"min=500"`
+	ReconnectInterval    time.Duration `validate:"min=500" yaml:"reconnect_interval"`
 	SocketKeepAlive      time.Duration `validate:"min=0" yaml:"socket_keep_alive"`
 	MaxRetryAttempts     int           `validate:"min=0" yaml:"max_retry_attempt"`
 	ProtoVersion         int           `yaml:"proto_version"`
@@ -74,6 +75,9 @@ func (c *Configuration) ApplyDefaults(source *Configuration) {
 	if c.Timeout == 0 {
 		c.Timeout = source.Timeout
 	}
+	if c.ReconnectInterval == 0 {
+		c.ReconnectInterval = source.ReconnectInterval
+	}
 	if c.Port == 0 {
 		c.Port = source.Port
 	}
@@ -109,6 +113,7 @@ func (c *Configuration) NewCluster() *gocql.ClusterConfig {
 	cluster.Keyspace = c.Keyspace
 	cluster.NumConns = c.ConnectionsPerHost
 	cluster.Timeout = c.Timeout
+	cluster.ReconnectInterval = c.ReconnectInterval
 	cluster.SocketKeepalive = c.SocketKeepAlive
 	if c.ProtoVersion > 0 {
 		cluster.ProtoVersion = c.ProtoVersion
--- a/plugin/storage/cassandra/options.go
+++ b/plugin/storage/cassandra/options.go
@@ -26,24 +26,25 @@ import (
 
 const (
 	// session settings
-	suffixEnabled          = ".enabled"
-	suffixConnPerHost      = ".connections-per-host"
-	suffixMaxRetryAttempts = ".max-retry-attempts"
-	suffixTimeout          = ".timeout"
-	suffixServers          = ".servers"
-	suffixPort             = ".port"
-	suffixKeyspace         = ".keyspace"
-	suffixConsistency      = ".consistency"
-	suffixProtoVer         = ".proto-version"
-	suffixSocketKeepAlive  = ".socket-keep-alive"
-	suffixUsername         = ".username"
-	suffixPassword         = ".password"
-	suffixTLS              = ".tls"
-	suffixCert             = ".tls.cert"
-	suffixKey              = ".tls.key"
-	suffixCA               = ".tls.ca"
-	suffixServerName       = ".tls.server-name"
-	suffixVerifyHost       = ".tls.verify-host"
+	suffixEnabled           = ".enabled"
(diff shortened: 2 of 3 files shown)


examples/hotrod/services/frontend/statik/statik.go
```diff
@@ -7,6 +7,6 @@ import (
...
...93z`L\x85)\xa9e\x99+\xe6\xa6\xae\xc2\xf0w\x8d\xdfF\xe0\xf56\x1e\xf5\xbc\xaf\xed$qe\xe2uC\x96\x89\x9ck\x94~\xf7\xf7\xfe\xd7MF7E\x1b\x03.mh\xa9N\xd1\x86\xae}\x0c\xdap\xcdu\xd2nx\xa2v\xd9\xbfo\xeb\xd1[PLQ\x07	MS\x9a\xa27\x02\xcf\xa01B\xc9\xd8{\x1a{oo\x1d\xac\x9a\xb7q\xb3\x03\xf0\xe0\xb43%\x8d\xca1\xccU\xea\xb7\x966>&8W\x1a!\x86)\xb5\x18Ju\xeb\xd7\xacW!]\xd0;\xdf\x8bx\x1b\x8f\x7fn\x1a\xbe6\xbfF\xd6\x03\xef\xefRI\x835y\xa7\xe8\xf6Z,\xad\xc5\xd1z\xb1WS\x0b\xb4\x99\xe2#\xf0\xce\xdf^{\x0d\xc9T\x8c\xa11#\xd8\x94\x88\x0b\xd5\x1eX\xbc\xb3W\x96\xda\xcat7\xe1q\xae\xd3\xb9\xadc\xbb\xeb\xb9{\xb6\x01;\x1d\x1bF\xdd\x93\xeb~\x8c\x9f\xf5/\xb54|{}\xba\x11_\x97S\xd3\x12^\xfb\xb2?N&\x0el-=\xd5b\xd9\x84a\x1c%\x13\xa0Z\x8b\xa5+\x1d!\xa1\x96Y\xdb\xea\x81\x07m\x19mg\xa8I^N-J\xb6jx~\x0d+h\xf2\xe2\xfa\xdd+\xccW\xafu\xe9\xc1E\xea\xa1{\xd2q\x9f\xfaz\xda\\J\xe3\xa8\xf9\xc3\xf4\xbf\x00\x00\x00\xff\xffPK\x07\x08\xb4\x93_{g\x05\x00\x008\x0d\x00\x00PK\x01\x02\x14\x03\x14\x00\x08\x00\x08\x00\xe7\x88\x83J\xb4\x93_{g\x05\x00\x008\x0d\x00\x00\n\x00	\x00\x00\x00\x00\x00\x00\x00\x00\x00\xa4\x81\x00\x00\x00\x00index.htmlUT\x05\x00\x01B\x81\xe2XPK\x05\x06\x00\x00\x00\x00\x01\x00\x01\x00A\x00\x00\x00\xa8\x05\x00\x00\x00\x00"
```
yurishkuro: Q: why is this changing? Can you exclude this change from the PR?

plugin/storage/cassandra/options.go
```diff
@@ -130,6 +132,10 @@ func addFlags(flagSet *flag.FlagSet, nsConfig *namespaceConfig) {
 		nsConfig.namespace+suffixTimeout,
 		nsConfig.Timeout,
 		"Timeout used for queries")
+	flagSet.Duration(
+		nsConfig.namespace+suffixReconnectInterval,
+		nsConfig.ReconnectInterval,
+		"Reconnect interval to retry connecting to downed hosts")
```
yurishkuro: Q: what would be the behavior if this is set to 0?

nyanshak: Current (master) jaeger behavior is equivalent to setting to 0. If hosts are marked as down, gocql will not try to reconnect without this set.

plugin/storage/cassandra/options.go
```diff
@@ -204,6 +210,7 @@ func (cfg *namespaceConfig) initFromViper(v *viper.Viper) {
 	cfg.ConnectionsPerHost = v.GetInt(cfg.namespace + suffixConnPerHost)
 	cfg.MaxRetryAttempts = v.GetInt(cfg.namespace + suffixMaxRetryAttempts)
 	cfg.Timeout = v.GetDuration(cfg.namespace + suffixTimeout)
+	cfg.ReconnectInterval = v.GetDuration(cfg.namespace + suffixReconnectInterval)
```
yurishkuro: please add a test

pkg/cassandra/config/config.go
```diff
@@ -31,6 +31,7 @@ type Configuration struct {
 	Keyspace             string        `validate:"nonzero"`
 	ConnectionsPerHost   int           `validate:"min=1" yaml:"connections_per_host"`
 	Timeout              time.Duration `validate:"min=500"`
+	ReconnectInterval    time.Duration `validate:"min=500" yaml:"reconnect_interval"`
```
vprithvi: I'm not convinced that the minimum of 500ns is a good value. Could you elaborate on how this was chosen?

(I think we need to revisit the `Timeout` as well separately)

nyanshak: I'm not sure that I really understand the `validate:"min=500"` section. It seems that it wasn't actually enforced in my tests. (As in, I would not pass any value, resulting in 0, which is less than 500 and it would pass validation. Unless that's a special case?)

I copied `Timeout` value for this.

vprithvi: Hmmm, it looks like these validations would only be applied if this configuration file was validated with something like https://github.com/go-validator/validator

I'll make a separate ticket to address this

PR #937, opened 2018-07-17
Replaces agent infinite block with agent.Stop()

danehans:
- The `select {}` statement to place `agent` in an infinite loop is uncommon and causes confusion for new contributors.

- The `agent.Stop()` method exists, so let's use it. Calling `defer` with the `agent.Stop()` method will stop `agent` only when the `agent.Start` go routine stops/fails and is easier for new contributors to understand.

Signed-off-by: Daneyon Hansen <danehans@cisco.com>

### DIFF
--- a/cmd/agent/main.go
+++ b/cmd/agent/main.go
@@ -63,7 +63,8 @@ func main() {
 			if err := agent.Run(); err != nil {
 				return errors.Wrap(err, "Failed to run the agent")
 			}
-			select {}
+			defer agent.Stop()
+			return nil
 		},
 	}
 


sergeyklay:
@black-adder Aha! https://travis-ci.org/jaegertracing/jaeger/jobs/405032533#L3365
The same crossdock error I faced yesterday in the #933

cmd/agent/main.go
```diff
@@ -63,7 +63,8 @@ func main() {
 			if err := agent.Run(); err != nil {
 				return errors.Wrap(err, "Failed to run the agent")
 			}
-			select {}
+			defer agent.Stop()
+			return nil
```
yurishkuro: I think this will exit the app immediately, won't it?

yurishkuro:
> The `select {}` statement to place agent in an infinite loop is uncommon and causes confusion for new contributors.

It's not that uncommon. But in the other executables we trapped the interrupts:

https://github.com/jaegertracing/jaeger/blob/7919cd98d30b858808941d72db31015a0317bd2d/cmd/query/main.go#L143-L147

danehans:
@yurishkuro OK. I tested locally and the agent does exit immediately. Although it would be nice to defer the agent.Stop(), it's not required. I will close out this PR. Thank you.

PR #940, opened 2018-07-19
Move over internal ingester code to OSS

vprithvi:
- See #929 

- This commit moves over code used internally at Uber to GitHub so that @davit-y may continue the work of adding initialization code. 
- The only changes done were moving import paths to GitHub, adding License headers and glide dependencies.


yurishkuro:
> Rename mock_saramaConsumer.go to cmd/ingester/app/consumer/mock_saramaConsumer_test.go

What is this commit trying to achieve?

vprithvi:
>What is this commit trying to achieve?

I wasn't able to find a easy way to exclude a single file from a package from coverage. Renaming with a `_test` works here because the mock is only used from tests. 

The mock is not in a separate mocks package because the consumer interface it is mocking was declared private.

vprithvi:
Build fails because of #945

PR #942, opened 2018-07-19
Ingester Consumer

davit-y:
- #929 

- Export consumer to be initialized and started in a main file 
- Add ConsumerBuilder to pkg/kafka/config
- Split pkg/kafka/config into Consumer and Producer


cmd/ingester/app/consumer/consumer.go
```diff
@@ -24,47 +23,84 @@ import (
...
 }
 
-// SaramaConsumer is an interface to features of Sarama that we use
-type SaramaConsumer interface {
-	Partitions() <-chan sc.PartitionConsumer
-	MarkPartitionOffset(topic string, partition int32, offset int64, metadata string)
-	io.Closer
+// New is a constructor for a Consumer
+func New(params Params) (*Consumer, error) {
+	saramaConsumer, err := params.ConsumerBuilder.NewConsumer()
```
black-adder: does `params.NewConsumer()` not work?

pkg/kafka/config/consumer/config.go
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) 2018 The Jaeger Authors.
```
black-adder: mb, I meant `pkg/kafka/config/consumer.go` and `pkg/kafka/config/producer.go`

black-adder: actually, keep as is, I prefer `producer.Configuration` over `config.ProducerConfiguration`

davit-y: Sounds good. I actually did it the first way, then noticed that it would be inconsistent with ES and Cassandra and changed it.

vprithvi: I don't think we need the path of this file to be `pkg/kafka/config/consumer/config.go`; having this be `pkg/kafka/consumer/config.go` conveys the same information

pkg/kafka/config/producer/config.go
```diff
@@ -12,22 +12,22 @@
...
 // limitations under the License.
 
-package config
+package producer
 
 import (
 	"github.com/Shopify/sarama"
 )
 
+// Builder builds a new kafka producer
```
vprithvi: Why move this from `26-30` to here? (I feel that this unnecessarily makes the code review bigger)

plugin/storage/kafka/factory.go
```diff
@@ -35,9 +35,9 @@ type Factory struct {
 	metricsFactory metrics.Factory
 	logger         *zap.Logger
 
-	config     config.ProducerBuilder
 	producer   sarama.AsyncProducer
 	marshaller Marshaller
+	producer.Builder
```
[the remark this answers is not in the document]
davit-y: Won had commented about splitting the config file into consumer and producer packages, would it be better to put everything into config.go in this PR and creating a new PR for the split?

pkg/kafka/config/consumer/config.go
davit-y: I'll make the change once the producer PR #957 is merged

cmd/ingester/app/consumer/consumer.go
```diff
@@ -24,47 +23,72 @@ import (
...
-
-			case <-c.close:
-				c.isClosed.Done()
-				return
-			}
+	go c.mainLoop()
+}
+
+// Close closes the Consumer and underlying sarama consumer
+func (c *Consumer) Close() error {
```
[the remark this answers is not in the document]
davit-y: in the producer PR you recommend this folder structure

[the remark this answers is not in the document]
vprithvi: Oh, I mean moving `Close` from L106-L110 to here.

Issue #850, opened 2018-05-29
Agent metrics are not tagged


yurishkuro:
Post #776 some metrics are still path based, these are the ones I found:

```
jaeger_agent_tc_reporter_jaeger_batch_size 0
jaeger_agent_tc_reporter_jaeger_batches_failures 0
jaeger_agent_tc_reporter_jaeger_batches_submitted 0
jaeger_agent_tc_reporter_jaeger_spans_failures 0
jaeger_agent_tc_reporter_jaeger_spans_submitted 0
jaeger_agent_tc_reporter_zipkin_batch_size 0
jaeger_agent_tc_reporter_zipkin_batches_failures 0
jaeger_agent_tc_reporter_zipkin_batches_submitted 0
jaeger_agent_tc_reporter_zipkin_spans_failures 0
jaeger_agent_tc_reporter_zipkin_spans_submitted 0
```

The `jaeger`/`zipkin` should be pulled into `format` tag

jpkrohling:
I believe the client is also not following the same idea.

yurishkuro:
client seems ok to me, but please open another ticket with examples if not

eundoosong:
@yurishkuro  So, above metrics should look like this?
```
jaeger_agent_tc_reporter_batch_size{format="jaeger"}
```
I think a few lines of codes are required to be changed to fix this.
If it's ok I'll PR.

yurishkuro:
agreed

Issue #927, opened 2018-07-10
Refactor Agent configuration


yurishkuro:
## Requirement - what kind of business use case are you trying to solve?

Enable gRPC communications between agent and collector (main issue #773).

## Problem - what in Jaeger blocks you from solving the requirement?

The agent configuration `app.Builder` directly embeds TChannel-based reporter configuration `app.reporter.tchannel.Builder`, rather than making it pluggable like `--processor.{format}.***` configurations. This makes the following flags into top-level options, even though they may not necessarily make sense to non-TChannel reporters:
```
--collector.host-port
--discovery.conn-check-timeout
--discovery.min-peers
```

## Proposal - what do you suggest to solve the problem or improve the existing situation?

Create a nested hierarchy of reporters, e.g. to support the following config:
```yaml
reporters:
    - kind: tchannel
      # tchannel-specific parameters, e.g.
      collector:
        hostPort: ...
      discovery:
        minPeers: 10
    - kind: grps
      # grpc-specific parameters
```

## Any open questions to address

This will be a breaking change since the command line will be affected. We could potentially keep support for the old three options as "deprecated".

isaachier:
My only issue with gRPC is how we can use a streaming protocol that we have no good way of instrumenting. From what I understand, there is no good hook for adding a context to a streaming request. If we have a solution to this, I would feel fine with this proposal.

yurishkuro:
This issue is not about gRPC (#773 is), but about refactoring agent to allow different output transports.

isaachier:
I know it's more like I feel bad endorsing gRPC as a transport mechanism if clients cannot use it correctly.

isaachier:
OK so then this is totally a non-issue.

danehans:
@yurishkuro thanks for the feedback. I've started digging in. If I get over my head, I'll send an update and work on a different issue. As you mention, I'll provide a summary of the changes before proceeding.

pavolloffay:
So new flags are 
```
--reporter.tchannel.collector.host-port string              comma-separated string representing host:ports of a static list of collectors to connect to directly (e.g. when not using service discovery)
      --reporter.tchannel.discovery.conn-check-timeout duration   sets the timeout used when establishing new connections (default 250ms)
      --reporter.tchannel.discovery.min-peers int                 if using service discovery, the min number of connections to maintain to the backend (default 3)
```
[...]

PR #950, opened 2018-07-23
Use tags in agent metrics

eundoosong:
Signed-off-by: Eundoo Song <eundoo.song@gmail.com>

- Change  path-based metrics to new format in agent.
- Resolves: #850 

- jaeger and zipkin are pulled into format tag.
- Gist : https://gist.github.com/eundoosong/8d3e06a259b60ce0ff346799dd3f7f22
```
jaeger_agent_tc_reporter_batch_size{format="jaeger"} 0
jaeger_agent_tc_reporter_batch_size{format="zipkin"} 0
jaeger_agent_tc_reporter_batches_failures{format="jaeger"} 0
jaeger_agent_tc_reporter_batches_failures{format="zipkin"} 0
jaeger_agent_tc_reporter_batches_submitted{format="jaeger"} 0
jaeger_agent_tc_reporter_batches_submitted{format="zipkin"} 0
jaeger_agent_tc_reporter_spans_failures{format="jaeger"} 0
jaeger_agent_tc_reporter_spans_failures{format="zipkin"} 0
jaeger_agent_tc_reporter_spans_submitted{format="jaeger"} 0
jaeger_agent_tc_reporter_spans_submitted{format="zipkin"} 0
```

### DIFF
--- a/cmd/agent/app/reporter/tchannel/reporter.go
+++ b/cmd/agent/app/reporter/tchannel/reporter.go
@@ -34,19 +34,19 @@ const (
 
 type batchMetrics struct {
 	// Number of successful batch submissions to collector
-	BatchesSubmitted metrics.Counter `metric:"tc-reporter.batches.submitted"`
+	BatchesSubmitted metrics.Counter `metric:"batches.submitted"`
 
 	// Number of failed batch submissions to collector
-	BatchesFailures metrics.Counter `metric:"tc-reporter.batches.failures"`
+	BatchesFailures metrics.Counter `metric:"batches.failures"`
 
 	// Number of spans in a batch submitted to collector
-	BatchSize metrics.Gauge `metric:"tc-reporter.batch_size"`
+	BatchSize metrics.Gauge `metric:"batch_size"`
 
 	// Number of successful span submissions to collector
-	SpansSubmitted metrics.Counter `metric:"tc-reporter.spans.submitted"`
+	SpansSubmitted metrics.Counter `metric:"spans.submitted"`
 
 	// Number of failed span submissions to collector
-	SpansFailures metrics.Counter `metric:"tc-reporter.spans.failures"`
+	SpansFailures metrics.Counter `metric:"spans.failures"`
 }
 
 // Reporter forwards received spans to central collector tier over TChannel.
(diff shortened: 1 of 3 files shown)


cmd/agent/app/reporter/tchannel/reporter.go
```diff
@@ -34,19 +34,19 @@ const (
 
 type batchMetrics struct {
 	// Number of successful batch submissions to collector
-	BatchesSubmitted metrics.Counter `metric:"batches.submitted"`
+	BatchesSubmitted metrics.Counter `metric:"tc-reporter.batches.submitted"`
```
yurishkuro: 1. there is no need to add tc-reporter explicitly like that, since you can pass that string as the first argument to Namespace in L76 (see point 3 below).
2. since the metrics are changing anyway, maybe we should rename them to use underscores, as we did with other backend metrics?
3. I think historically `tc` was used to refer to `tcollector`, which was the TChannel name of the collector service (now it's `jaeger-collector`). When we implement #927, there will be a different type of reporter, e.g. `grpc-reporter`. Also, in some cases we internally run the agent with two reporters (using OtherReporters configuration of the builder), I don't even know if they distinguish their metrics. So my suggestion, at minimum, to rename this to `tchannel-reporter`, to make it more obvious what we're talking about, but we also need to figure out how metrics will work after #927. 

cc @black-adder

eundoosong: > 2. since the metrics are changing anyway, maybe we should rename them to use underscores, as we did with other backend metrics?

I can change it using underscore like `tchannel_reporter`. Just curious, the reason I kept using `-` is normalizer in jaeger-lib already replaces them. Don't we need it?
https://github.com/jaegertracing/jaeger-lib/blob/master/metrics/prometheus/factory.go#L109

yurishkuro: actually I see that collector metrics are still using dots as separators, so never mind about understores

SeonghoByeon:
- Move tc-reporter to Namespace as first argument.
- Rename to tchannel-reporter

eundoosong:
- Move tc-reporter to Namespace as first argument.
- Rename to tchannel-reporter

yurishkuro:
Thanks!

PR #952, opened 2018-07-24
Ingester Main

davit-y:
- #929 

- Add options.go to configure ingester using viper
- Add main.go to initialize and start Ingester

### DIFF
--- a/cmd/ingester/app/builder/builder.go
+++ b/cmd/ingester/app/builder/builder.go
@@ -15,6 +15,7 @@
 package builder
 
 import (
+	"fmt"
 	"strings"
 
 	"github.com/spf13/viper"
@@ -56,7 +57,7 @@ const (
 	// DefaultParallelism is the default parallelism for the span processor
 	DefaultParallelism = 1000
 	// DefaultEncoding is the default span encoding
-	DefaultEncoding = EncodingJSON
+	DefaultEncoding = EncodingProto
 )
 
 // Builder stores the configuration options for the Ingester
@@ -71,8 +72,11 @@ func (b *Builder) CreateConsumer(logger *zap.Logger, metricsFactory metrics.Fact
 	var unmarshaller kafka.Unmarshaller
 	if b.Encoding == EncodingJSON {
 		unmarshaller = kafka.NewJSONUnmarshaller()
-	} else {
+	} else if b.Encoding == EncodingProto {
 		unmarshaller = kafka.NewProtobufUnmarshaller()
+	} else {
+		return nil, fmt.Errorf(`encoding '%s' not recognised, use one of ("%s" or "%s")`,
+			b.Encoding, EncodingProto, EncodingJSON)
 	}
 
 	spParams := processor.SpanProcessorParams{
--- a/cmd/ingester/app/builder/builder.go
+++ b/cmd/ingester/app/builder/builder.go
@@ -1,0 +1,123 @@
+// Copyright (c) 2018 The Jaeger Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package builder
+
+import (
+	"strings"
+
+	"github.com/spf13/viper"
+	"github.com/uber/jaeger-lib/metrics"
+	"go.uber.org/zap"
+
+	"github.com/jaegertracing/jaeger/cmd/ingester/app/consumer"
+	"github.com/jaegertracing/jaeger/cmd/ingester/app/processor"
+	kafkaConsumer "github.com/jaegertracing/jaeger/pkg/kafka/consumer"
+	"github.com/jaegertracing/jaeger/plugin/storage/kafka"
+	"github.com/jaegertracing/jaeger/storage/spanstore"
+)
+
+const (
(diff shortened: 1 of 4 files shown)


cmd/ingester/app/options.go
```diff
@@ -0,0 +1,98 @@
...
+
+	"github.com/spf13/viper"
+
+	"github.com/jaegertracing/jaeger/pkg/kafka/consumer"
+	"github.com/jaegertracing/jaeger/plugin/storage/kafka"
+)
+
+const (
+	configPrefix      = "ingester"
+	suffixBrokers     = ".brokers"
```
black-adder: hmmm, I'm not sure about this. These configs shouldn't belong on the ingester itself, but should be encapsulated inside the storage factory. Do we have to do this like this since we haven't come up with a better story for the storage factory and flags?

davit-y: We talked in person, but I'll recap for those who may be curious. Early in the design process there was a discussion about wether the Ingester should operate on an abstract queue or for Kafka. We decided it will be designed for kafka, but be independent of the write storage. Therefore there isn't an issue with having kafka specific config on the Ingester as long as the write storage is handled similarly to the other jaeger applications.

cmd/ingester/main.go
```diff
@@ -0,0 +1,160 @@
...
+
+			storageFactory.InitFromViper(v)
+			if err := storageFactory.Initialize(baseFactory, logger); err != nil {
+				logger.Fatal("Failed to init storage factory", zap.Error(err))
+			}
+			spanWriter, err := storageFactory.CreateSpanWriter()
+			if err != nil {
+				logger.Fatal("Failed to create span writer", zap.Error(err))
+			}
+			unmarshaller := app.UnmarshallerFromType(options.Encoding)
```
black-adder: I can see us copy pasting a lot of this kafka initialization logic when we OSS the indexer (is that part of the plan btw). Do you think it'll make sense to create something like https://github.com/jaegertracing/jaeger/blob/master/pkg/cassandra/config/config.go to handle all the kafka bootstrapping rather than dumping it all here in main?

black-adder: hmm, let me rephrase a bit, internally, we don't want to write this same main logic if we can help it. Ideally, we create a builder for the ingester similar to the agent builder https://github.com/jaegertracing/jaeger/blob/master/cmd/agent/app/builder.go and internally, we can just run builder.build() instead of copy pasting this same code.

cmd/ingester/app/options.go
```diff
@@ -0,0 +1,98 @@
...
+	suffixTopic       = ".topic"
+	suffixGroupID     = ".group-id"
+	suffixParallelism = ".parallelism"
+	suffixEncoding    = ".encoding"
+
+	encodingJSON  = "json"
+	encodingProto = "protobuf"
+
+	defaultBroker      = "127.0.0.1:9092"
+	defaultTopic       = "jaeger-ingester-spans"
```
vprithvi: Is this the default topic used by the `jaeger-collector`'s kafka writers to emit spans?

cmd/ingester/main.go
```diff
@@ -0,0 +1,168 @@
...
+				logger.Fatal("Unable to set up consumer", zap.Error(err))
+			}
+			kafkaConsumer.Start()
+
+			hc.Ready()
+			select {
+			case <-signalsChannel:
+				err := kafkaConsumer.Close()
+				if err != nil {
+					logger.Error("Failed to close span writer", zap.Error(err))
```
vprithvi: Does closing the consumer also close the span writer?

cmd/ingester/main.go
```diff
@@ -0,0 +1,176 @@
...
+				if err != nil {
+					logger.Error("Failed to close consumer", zap.Error(err))
+				}
+				if closer, ok := spanWriter.(io.Closer); ok {
+					err := closer.Close()
+					if err != nil {
+						logger.Error("Failed to close span writer", zap.Error(err))
+					}
+				}
+				logger.Info("Jaeger Ingester is finishing")
```
vprithvi: nit: I'd also add a log line on L101 stating "Beginning shutdown of Jaeger Ingester" or something similar.

davit-y:
Redoing using a builder to allow for easy initialization of an ingester

cmd/ingester/main.go
davit-y: Done

cmd/ingester/app/builder/builder.go
```diff
@@ -0,0 +1,123 @@
...
+	// DefaultBroker is the default kafka broker
+	DefaultBroker = "127.0.0.1:9092"
+	// DefaultTopic is the default kafka topic
+	DefaultTopic = "jaeger-spans"
+	// DefaultGroupID is the default consumer Group ID
+	DefaultGroupID = "jaeger-ingester"
+	// DefaultParallelism is the default parallelism for the span processor
+	DefaultParallelism = 1000
+	// DefaultEncoding is the default span encoding
+	DefaultEncoding = EncodingJSON
```
vprithvi: I'm curious, is this what we emit from the collectors? (Not protobuf?)

cmd/ingester/app/builder/builder.go
```diff
@@ -0,0 +1,123 @@
...
+	Parallelism int
+	Encoding    string
+}
+
+// CreateConsumer creates a new span consumer for the ingester
+func (b *Builder) CreateConsumer(logger *zap.Logger, metricsFactory metrics.Factory, spanWriter spanstore.Writer) (*consumer.Consumer, error) {
+	var unmarshaller kafka.Unmarshaller
+	if b.Encoding == EncodingJSON {
+		unmarshaller = kafka.NewJSONUnmarshaller()
+	} else {
```
vprithvi: I'm not a fan of this approach because if someone passes in an arbitrary encoding, they would get a protobuf marshaller. I recommend checking for the type and returning an error if nothing matches.

cmd/ingester/app/builder/builder.go
davit-y: My mistake, default should be protobuf

PR #956, opened 2018-07-24
Update query handler to read from archive reader when multiple traceI…

black-adder:
…Ds are passed as parameters

- When multiple traceIDs are passed to the query service, it only reads the traces from the reader, not the archiveReader.


cmd/query/app/handler.go
```diff
@@ -399,24 +399,32 @@ func (aH *APIHandler) withTraceFromReader(
...
+		aH.handleError(w, err, http.StatusNotFound)
+		return
 	}
 	if aH.handleError(w, err, http.StatusInternalServerError) {
 		return
 	}
 	process(trace)
 }
 
+func trace(
```
black-adder: every combination of `getTrace`, `getTraceFromReaders`, etc. was taken. Uh `getTraceFromReadersInternal`?

PR #964, opened 2018-07-27
Parse media types in the Content-Type header correctly

### LINKED ISSUE #963: Zipkin HTTP Endpoint Does Not Correctly Parse Content-Type Header

Post Zipkin traces from an Android client (okhttp).

The Zipkin HTTP handler [incorrectly uses the full text of the `Content-Type` header](https://github.com/jaegertracing/jaeger/blob/53955e348810c9d119bda87fc99951dd73603319/cmd/collector/app/zipkin/http_handler.go#L84-L90) as a means of ascertaining whether the request uses `application/json`. A valid `Content-Type` header may be a [`media-type`](https://tools.ietf.org/html/rfc7231#section-3.1.1.1), a la `application/json; charset=utf8` which would fail the direct comparison in the linked code.

Adjust `Content-Type` detection in the various endpoints to correctly parse a `media-type` following the [specification](https://tools.ietf.org/html/rfc7231#section-3.1.1.1).

netshade:
- Resolves #963 , in which `Content-Type` extraction was using the raw value of the `Content-Type` header, rather than extracting it from the containing Media Type.

- Imports the `mime` module into the HTTP Thrift endpoint as well as the Zipkin endpoints and resolves the `Content-Type` using the `mime` package's [`ParseMediaType`](https://golang.org/pkg/mime/#ParseMediaType) method.

### DIFF
--- a/cmd/collector/app/http_handler.go
+++ b/cmd/collector/app/http_handler.go
@@ -67,7 +67,12 @@ func (aH *APIHandler) saveSpan(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	contentType, _, _ := mime.ParseMediaType(r.Header.Get("Content-Type"))
+	contentType, _, err := mime.ParseMediaType(r.Header.Get("Content-Type"))
+
+	if err != nil {
+		http.Error(w, fmt.Sprintf("Cannot parse content type: %v", err), http.StatusBadRequest)
+		return
+	}
 
 	if _, ok := acceptedThriftFormats[contentType]; !ok {
 		http.Error(w, fmt.Sprintf("Unsupported content type: %v", contentType), http.StatusBadRequest)
--- a/cmd/collector/app/http_handler.go
+++ b/cmd/collector/app/http_handler.go
@@ -17,6 +17,7 @@ package app
 import (
 	"fmt"
 	"io/ioutil"
+	"mime"
 	"net/http"
 	"time"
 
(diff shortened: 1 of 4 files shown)


netshade:
Of note: I do not write Go professionally ( or much at all ) and could easily be glossing over important things. More critical eye likely warranted.

cmd/collector/app/http_handler.go
```diff
@@ -66,7 +67,8 @@ func (aH *APIHandler) saveSpan(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	contentType := r.Header.Get("Content-Type")
+	contentType, _, _ := mime.ParseMediaType(r.Header.Get("Content-Type"))
```
yurishkuro: one issue here (and in all similar statements)  is that the parsing error is swallowed, and will usually cause a later, less intelligible, error like `Unsupported content type: ` (i.e. nothing at the end). The preferred way to handle this is
```go
contentType, _, err := mime.ParseMediaType(r.Header.Get("Content-Type"))
if err != nil {
		http.Error(w, fmt.Sprintf("Cannot parse content type: %v", err), http.StatusBadRequest)
		return
}
```
it will require more failure tests to maintain the code coverage.

netshade: Cool, so to you, it's valuable that parse errors here are surfaced as discrete new errors for Jaeger? Like, you want it such that the caller gets notified how their content type is wrong? ( Rather than just, "it's wrong" )

netshade: ( Because the latter seems already accounted for in the existing code )

netshade: nm, re-read your comment and that's exactly what you want. Cool, I'll add.

yurishkuro: Every time the code swallows an error, someone gets burnt by it, and a fairy dies. It's always better to pass the full error to the user than to tell them "you f-ed up, no comment".

netshade:
@yurishkuro Updated

PR #969, opened 2018-07-31
Fix hard-coded ES spanType

roxit:
Replace "span" with spanType when reading from ES.

Signed-off-by: Shanshi Shi <shishanshi@gmail.com>

ES 6.0 allows only one document type per index, and my organization's ES platform forces "_type" to be "logs".

In order to write to our ES, I had to fork jaeger and changed `spanType` & `serviceType` in `github.com/jaegertracing/jaeger/plugin/storage/es/spanstore/writer.go` to "logs".

After the tweak, jaeger-collector was able to write to our ES, but, jaeger-query still cannot find any traces.

It seems `github.com/jaegertracing/jaeger/plugin/storage/es/spanstore/reader.go` is using the hard-coded `.Type("span")` to query ES instead of `.Type(spanType)`


pavolloffay:
@roxit thanks for the fix!

pavolloffay:
I could not build this on travis due to https://travis-ci.org/jaegertracing/jaeger/jobs/418663898#L1031. 

I will merge rebased PR https://github.com/jaegertracing/jaeger/pull/1012

Issue #365, opened 2017-09-01
Adaptive Sampling


yurishkuro:
## Problem

The most common way of using Jaeger client libraries is with probabilistic sampling which makes a determination if a new trace should be sampled or not. Sampling is necessary to control the amount of tracing data reaching the storage backend. There are two issues with the current approach:
  1. The individual microservices have little insight into what the appropriate sampling rate should be. For example, 0.001 probability (one trace per second per service instance) might seem reasonable, but if the fanout in some downstream services is very high it might flood the tracing backend.
  1. Sampling rates are defined on a per-service basis. If a service has two endpoints with vastly different throughputs, then its sampling rate will be driven on the high QPS endpoint, which may leave the low QPS endpoint never sampled. For example, if the QPS of the endpoints is different by a factor of 100, and the probability is set to 0.001, then the low QPS traffic will have only 1 in 100,000 chance to be sampled.

## Proposed Solution

The adaptive sampling is a solution that addresses these issues by:
  1. Assigning sampling probabilities on a service + endpoint basis rather than just the service
  1. Using a lower bound rate limiter to ensure that all endpoints are sampled with a certain minimal rate
  1. Observing the impact of sampling rates on the overall number of traces sampled from a service and dynamically adjusting the per-endpoint sampling rates to meet certain target rates.

## Status

Pending open-source of the backend functionality. Client work is done.

* [x] sampling data store (Cassandra only) #973 
* [x] adaptive sampling calculator #1179
* [ ] wire up the components from collector/main

robdefeo:
Any idea when the backend functionality will be opensourced?

billowqiu:
"Adaptive Sampling" is now ok in backend?

yurishkuro:
it's coming soon, @black-adder just finished rolling it out internally to all services, so it just needs a bit clean-up (from any internal deps) to move to open source.

billowqiu:
thks @yurishkuro , i am investigating the jaeger and zipkin。

trtg:
@yurishkuro  any progress on this being released?

yurishkuro:
question to @black-adder , "he is a-cooking something up"

sergeyklay:
@black-adder any news?

black-adder:
sorry all, I just started to move the pieces over, hopefully we'll have the whole thing in OSS this week.

agxp:
[...]

PR #973, opened 2018-08-07
Start moving components of adaptive sampling to OSS

black-adder:
This PR is the start of moving adaptive sampling (#365) over to OSS. This is moving over calculation_strategy and leader_election code.


plugin/sampling/internal/calculationstrategy/interface.go
```diff
@@ -0,0 +1,28 @@
...
+
+package calculationstrategy
+
+// ProbabilityCalculator calculates the new probability given the current and target QPS
+type ProbabilityCalculator interface {
+	Calculate(targetQPS, curQPS, prevProbability float64) (newProbability float64)
+}
+
+// Calculate wraps a function of appropriate signature and makes a ProbabilityCalculator from it.
+type Calculate func(targetQPS, curQPS, prevProbability float64) (newProbability float64)
```
yurishkuro: std lib calls this XyzFunc

pkg/testutils/logger.go
```diff
@@ -88,3 +89,17 @@ func (b *Buffer) Write(p []byte) (int, error) {
 	defer b.Unlock()
 	return b.Buffer.Write(p)
 }
+
+// LogMatcher is a helper func that returns true if the subStr appears more than 'occurrences' times in the logs.
+var LogMatcher = func(occurrences int, subStr string, logs []string) bool {
```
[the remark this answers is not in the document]
yurishkuro: ping

plugin/sampling/internal/calculationstrategy/interface.go
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) 2018 The Jaeger Authors.
```
yurishkuro: what is the thinking for this thing to be under `plugins`?  I can see swappable storage to be under plugins, but this seems like the core impl. I would keep it inside the collector app.

plugin/sampling/internal/leaderelection/leader_election.go
```diff
@@ -0,0 +1,117 @@
...
+	"time"
+
+	"go.uber.org/atomic"
+	"go.uber.org/zap"
+
+	dl "github.com/jaegertracing/jaeger/pkg/distributedlock"
+)
+
+var (
+	acquireLockErrMsg = "Failed to acquire lock"
```
yurishkuro: it's a const

plugin/sampling/internal/leaderelection/leader_election.go
```diff
@@ -0,0 +1,117 @@
...
+
+// ElectionParticipant partakes in leader election to become leader.
+type ElectionParticipant interface {
+	IsLeader() bool
+}
+
+type electionParticipant struct {
+	ElectionParticipantOptions
+	lock         dl.Lock
+	isLeader     *atomic.Bool
```
yurishkuro: why a pointer?

plugin/sampling/internal/calculationstrategy/interface.go
black-adder: This is core only to the adaptive sampling strategystore. Don't feel the collector app is the right place for this

plugin/sampling/internal/leaderelection/leader_election.go
black-adder: `atomic.NewBool(false)` returns `*atomic.Bool`

pkg/testutils/logger.go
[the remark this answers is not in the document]
black-adder: passing in T makes this hard to test since if we try to test the failure case, the test itself will fail (unless I can mock T which seems overkill). I'll update this func to return `(bool, errMsg)`

plugin/sampling/internal/calculationstrategy/interface.go
black-adder: gonna go ahead and land, we can figure out where these files should live later

PR #977, opened 2018-08-09
Add error logging for individual ES bulk reposes

pavolloffay:
It's important to check every bulk response because
root error can be nil even when some response failed.

### DIFF
--- a/pkg/es/config/config.go
+++ b/pkg/es/config/config.go
@@ -26,7 +26,7 @@ import (
 
 	"github.com/jaegertracing/jaeger/pkg/es"
 	storageMetrics "github.com/jaegertracing/jaeger/storage/spanstore/metrics"
-)
+	)
 
 // Configuration describes the configuration properties needed to connect to an ElasticSearch cluster
 type Configuration struct {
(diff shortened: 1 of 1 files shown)


pkg/es/config/config.go
```diff
@@ -75,18 +75,28 @@ func (c *Configuration) NewClient(logger *zap.Logger, metricsFactory metrics.Fac
...
+					for key, val := range it {
+						if val.Error != nil {
+							logger.Error("Elasticsearch part of bulk request failed", zap.String("map-key",key),
+								zap.Reflect("response", val))
+						}
+					}
+				}
 			}
+
+			sm.Emit(err, time.Since(start.(time.Time)))
```
pavolloffay: Just fixing identation

pkg/es/config/config.go
```diff
@@ -75,9 +75,19 @@ func (c *Configuration) NewClient(logger *zap.Logger, metricsFactory metrics.Fac
...
+					for key, val := range it {
+						if val.Error != nil {
+							logger.Error("Elasticsearch part of bulk request failed", zap.String("map-key", key),
+								zap.Reflect("response", val))
+						}
+					}
+				}
+			}
 
+			sm.Emit(err, time.Since(start.(time.Time)))
```
black-adder: nit: I don't think the error logging should be a part of the measured latency

pavolloffay: error logging is not part of the measured latency. If the error is present it updates different metric -LatencyErr

pkg/es/config/config.go
```diff
@@ -75,9 +75,19 @@ func (c *Configuration) NewClient(logger *zap.Logger, metricsFactory metrics.Fac
...
 			m.Delete(id)
 
-			duration := time.Since(start.(time.Time))
-			sm.Emit(err, duration)
+			// log individual errors, note that err might be false and these errors still present
+			if response.Errors {
+				for _, it := range response.Items {
+					for key, val := range it {
+						if val.Error != nil {
+							logger.Error("Elasticsearch part of bulk request failed", zap.String("map-key", key),
```
yurishkuro: 7 levels of indentation - could be a record for Go :-)  This code needs refactoring: move `After(func` to a named function, and move error logging to a function too.

Issue #729, opened 2018-03-06
Add the ability to store/retrieve partial spans


vprithvi:
For the first version of this feature, we shall assume the following
-  clients can flush and report partial spans; but not span deltas
-  `duration` is monotonically increasing
-  the final span has the longest duration

For e.g., a client can do the following 
**time = 1s**
`Report Span[traceID=1, spanID=2, duration=10s, operationName="someOperation"]` 
 
**time = 2s**
`Report Span[traceID=1, spanID=2, duration=20s, operationName="someOperation"]` 

**time = 3s**
`Report Span[traceID=1, spanID=2, duration=40s, operationName="someOperation", tags=...]`

To support this, the backend would need the ability to resolve merge conflicts on spans. For V1, this means simply selecting the longest span. 

On Cassandra, the jaeger-collector uses the model spanhash, guaranteeing that all partial spans are stored:
https://github.com/jaegertracing/jaeger/blob/412baf60db217d6f41d7fdca04358069362b29ca/plugin/storage/cassandra/spanstore/dbmodel/converter.go#L45

https://github.com/jaegertracing/jaeger/blob/fcbd21078ab7ed03682964c187452778242d3423/model/span.go#L63-L66

On ElasticSearch, we use the index api, which performs upserts. 
https://github.com/jaegertracing/jaeger/blob/e52ecffbf69a572593504ea9acc4ff65854a3e9a/plugin/storage/es/spanstore/writer.go#L179-L182

To make matters more interesting, note that Jaeger supports storing Zipkin spans that share the same spanID for client and server spans. Jaeger adjusts these spanIDs during query time as seen here: 
https://github.com/jaegertracing/jaeger/blob/412baf60db217d6f41d7fdca04358069362b29ca/model/adjuster/span_id_deduper.go#L23-L31

Another point to note is that `jaeger-collector` makes no guarantee that spans are stored in the order that they are received. 

## Enhancements
1.  Communicate that jaeger-query is serving incomplete spans to the user. One approach is to do something similar to https://github.com/jaegertracing/jaeger-ui/issues/132.  We might have to enhance the Span model to store this information
1. Allow for jaeger-clients to report that the span it is sending is the final span. This allows us to remove the assumption about monotonically increasing duration. 
1. Add support for span deltas, which might allow for clients to reduce state maintained. (Might be at odds with the previous point)
[...]

PR #979, opened 2018-08-09
Incomplete span support

phal0r:
- This PR supports sending intermediate spans as suggested in #729 
- this PR is build upon PR #728 

- changed all jaeger models to accept a new key `incomplete`
- generated new thrift types
- zipkin spans are always converted incomplete: false
- backwards compatible with client, that do not support this feature (incomplete will be false)
- changed elasticsearch mapping to support inomplete attribute
- changed cassandra schema and reader/writer to support incomplete attribute


pavolloffay:
@phal0r it needs to be rebased

phal0r:
@pavolloffay 
Yeah, will do that and it seems that I need to generate the model now from protobuf definitions. I will look into that aswell

jaeger-ui-build/build/asset-manifest.json
```diff
@@ -0,0 +1,15 @@
+{
```
yurishkuro: we don't check in jaeger-ui-build

model/model.pb.go
```diff
@@ -1,36 +1,16 @@
...
 import proto "github.com/gogo/protobuf/proto"
 import golang_proto "github.com/golang/protobuf/proto"
 import fmt "fmt"
 import math "math"
-import _ "github.com/gogo/protobuf/types"
-import _ "github.com/gogo/protobuf/types"
 import _ "github.com/gogo/googleapis/google/api"
-import _ "github.com/grpc-ecosystem/grpc-gateway/protoc-gen-swagger/options"
 import _ "github.com/gogo/protobuf/gogoproto"
+import _ "github.com/gogo/protobuf/types"
```
yurishkuro: it looks like you're using different version of the generator, there should not be so many changes if you only add one field

yurishkuro: I am getting these diffs too, need to track them down. I see gogo version was changed in glide.lock in 933efb3a

isaachier:
@yurishkuro this is what I extracted from the model schema to use in clients: https://github.com/isaachier/jaeger-client-c/tree/master/idl. It simplifies the model a lot by avoiding Go-specific tooling and annotations.

yurishkuro:
@isaachier I don't know what you mean by that. The gogo-specific annotations are necessary to make the generated classes perform well in Go. They should not have any impact on non-Go generators.

yurishkuro:
@phal0r try rebasing off #984 (merged to master), and run `make proto-install proto`

phal0r:
@yurishkuro 
Thanks for the fast help. I rebased the branch and tried this:
```
glide install
make proto-install
make proto
```

But I still get the wrong output with the reference to `InternalMessage `. What version of `protoc` are you using? I have `3.6.1` installed. Maybe this is wrong, too?

yurishkuro:
Hm, I used `protoc-3.5.1-osx-x86_64/bin/protoc`. I will try with 3.6 and post an update.

phal0r:
@yurishkuro 
Yes, after deletion it is working as expected now, thanks! 

I can fix the failing tests now.

phal0r:
Ok, everything is resolved now:
- ui build is deleted
- tests are passing
- proto files are correct

@yurishkuro What about the DCO? Does this refer to the username set in my git client? And is this a problem? I signed them off correctly.

isaachier:
@phal0r this relates to not signing each and every commit. Yes it is annoying... but the only clear workaround is to squash your commits into one and use that.

Try these commands:

```
git rebase master
git branch incomplete_span_support_backup
git checkout master
git branch -D incomplete_span_support
git checkout -b incomplete_span_support
git merge --ff-only --squash incomplete_span_support_backup
git commit -s -m "Implement incomplete span support"
git push --force --set-upstream origin incomplete_span_support
```

isaachier:
Actually, might be a different issue. The message is `Expected "phal0r <martin@atroo.de>", but got "Martin Foerster <martin@atroo.de>".`.

yurishkuro:
I've seen this before, but not sure the real cause. I suspect it's some mismatch between your local git profile and your Github profile, 'cause how else would DCO bot know what "should" be the right signature.

phal0r:
Tests are running and the DCO issue is resolved.

phal0r:
@yurishkuro 
What is the procedure for this Pull Request? Do you have a roadmap, when you want to merge it? I am just curious about the timeframe for merging or eventual changes, that may need to be incorporated.

model/adjuster/span_merger.go
```diff
@@ -0,0 +1,84 @@
...
+
+import (
+	"github.com/jaegertracing/jaeger/model"
+)
+
+// MergeSpans returns an Adjuster that merges Jaeger spans with the same spanID.
+// Duplicate spans that have conflicting span.kind annotations are not merged.
+//
+// MergeSpans assumes that the duration field in a span is monotonically increasing for a spans with the
+// same spanID. The span with the longest spanID wins, and is selected in entirety.
```
vprithvi: s/spanID/duration

model/converter/thrift/jaeger/to_domain.go
```diff
@@ -74,9 +74,17 @@ func (td toDomain) transformSpan(jSpan *jaeger.Span, mProcess *model.Process) *m
...
 		Process:       mProcess,
+		Incomplete:    td.checkIncomplete(jSpan.Incomplete),
 	}
 }
 
+func (td toDomain) checkIncomplete(incomplete *bool) bool {
+	if incomplete == nil {
+		return false
+	}
+	return *incomplete
```
vprithvi: nit: could be simplified to `return incomplete != nil && *incomplete`

model/proto/model.proto
```diff
@@ -145,6 +145,9 @@ message Span {
     (gogoproto.customname) = "ProcessID"
   ];
   repeated string warnings = 12;
+  bool incomplete = 13 [
+    (gogoproto.nullable) = true
```
vprithvi: I believe we can avoid dealing with bool pointers if we set nullable to `false`.  
Could you elaborate on why this needs to be nullable?

thrift-gen/jaeger/ttypes.go
```diff
@@ -878,6 +879,7 @@ type Span struct {
 	Duration      int64      `thrift:"duration,9,required" json:"duration"`
 	Tags          []*Tag     `thrift:"tags,10" json:"tags,omitempty"`
 	Logs          []*Log     `thrift:"logs,11" json:"logs,omitempty"`
+	Incomplete    *bool      `thrift:"incomplete,12" json:"incomplete,omitempty"`
```
vprithvi: What is this being generated from? I don't see corresponding `jaeger.thrift` changes. 

Also, why does thrift generate a pointer to a bool?

phal0r: You are right. The Pull Request to the IDL Repo is missing. Will do that. 

Thrift generates a pointer for all attributes, that are optional. It's the same for Tags, Logs and References. Since the collector should also accept spans from clients, that don't support the incomplete span feature, it must be optional imho.

model/adjuster/span_merger.go
phal0r: This is a change from Pull Request #728 

Shouldn't it be changed there then?

model/converter/thrift/jaeger/to_domain.go
phal0r: Will do that.

model/proto/model.proto
phal0r: Yeah, I think I misread this from stackoverflow: https://stackoverflow.com/questions/42622015/how-to-define-an-optional-field-in-protobuf-3/42634681

It is mentioned:
`In proto3, all fields are "optional" (in that it is not an error if the sender fails to set them)`

I can remove the nullable as it makes not much sense regarding this. But I think for the generated thrift it will still be a boolean pointer.

model/proto/model.proto
```diff
@@ -145,6 +145,9 @@ message Span {
     (gogoproto.customname) = "ProcessID"
   ];
   repeated string warnings = 12;
+  bool incomplete = 13 [
```
yurishkuro: please add documentation for the new field (even though the rest of the file doesn't, but it's TBD)

yurishkuro:
Q: the span merger comment indicates that spans are merged based on duration. Then what does the additional "incomplete" field provide?

phal0r:
@yurishkuro 
That's correct, but the information if the `latest` span is also the `final` span is missing. At the moment the sent span determines the end of the unit of work, whereas sending a lot intermediate spans just indicates progress, but one does not know, when (and if) the unit of work actually ended. This is where the `incomplete` flag comes into play to give clear indication of the end.

thrift-gen/jaeger/ttypes.go
phal0r: @vprithvi 
I created the pull request for the idl repo: https://github.com/jaegertracing/jaeger-idl/pull/50

Since I have no control over the thrift generated files, I will mark this as resolved. Please let me know, if there is still anything to be changed. Thanks!

phal0r:
Everything is resolved from my side. https://github.com/jaegertracing/jaeger-idl/pull/50 for the IDL update is resolved aswell.

yurishkuro:
IDL has been merged

yurishkuro:
> cmd/standalone/standalone-linux

some of the binaries are included in this pr, please remove them

model/adjuster/span_merger.go
```diff
@@ -0,0 +1,84 @@
...
+import (
+	"github.com/jaegertracing/jaeger/model"
+)
+
+// MergeSpans returns an Adjuster that merges Jaeger spans with the same spanID.
+// Duplicate spans that have conflicting span.kind annotations are not merged.
+//
+// MergeSpans assumes that the duration field in a span is monotonically increasing for a span with the
+// same spanID. The span with the longest duration wins, and is selected in entirety.
+// TODO: Granular merging of spans
```
yurishkuro: what is this referring to?

yurishkuro:
@phal0r the Merger seems to be exclusively based on duration field, not on the new `incomplete` flag. Could you please elaborate on what the plan is?

model/adjuster/span_merger.go
phal0r: This refers to your suggestion on span merging rules: https://github.com/jaegertracing/jaeger/issues/729#issuecomment-370947209

The current implementation actually is not merging the spans, but just picking the span with the longest duration since the assumption is, that no deltas are sent, but always full spans.

This makes client implementations easier, since they don't need to care about any state, that may have been flushed already. Disadvantage is, that it needs more storage, because tags and logs will be duplicated for every intermediate span in the storage layer.

In general, granular merging only makes sense, if the client implementations are smart enough to send deltas in order to save network traffic and storage, but I am not sure, if this is not overcomplicating client logic, so I am not sure, if this will ever be needed.

phal0r: Ah, I didn't realize, that Spans don't have getters for their properties. So actually just flushing span data and only keeping spanId and traceId is actually not that hard

phal0r:
The incomplete flag is now used to select the final span. If no final span was received yet, the latest span is selected. Merging is done for spanRefs, tags, logs and warnings.

phal0r:
@yurishkuro 
Can you review the changes. The code is selecting the final span and merging tags, warnings, logs and references

model/adjuster/span_merger.go
```diff
@@ -0,0 +1,116 @@
...
+	return IDToSpans
+}
+
+func mergeSpans(spans []*model.Span) *model.Span {
+	finalSpan := spans[len(spans)-1]
+	//we track the index of the final span to cover the case
+	//that we didn't receive the final span yet and don't want
+	//to duplicate the members of the last span due the fallback
+	//of using the last span as carrier when no final span is stored yet
+	completeSpanIndex := len(spans) - 1
```
yurishkuro: final or complete - let's pick one name, not two

model/adjuster/span_merger.go
```diff
@@ -0,0 +1,116 @@
...
+	logs := []model.Log{}
+	references := []model.SpanRef{}
+	for i, span := range spans {
+		//merge refs, tags, logs and warnings of all spans
+		//take simple values from lastSpan
+		if span.GetIncomplete() && completeSpanIndex != i {
+			references = append(references, span.GetReferences()...)
+			tags = append(tags, span.GetTags()...)
+			logs = append(logs, span.GetLogs()...)
+			warnings = append(warnings, span.GetWarnings()...)
```
yurishkuro: so this would just duplicate everything as many times as the incomplete spans were flushed. We need to dedupe all of these data entries. Or, alternatively, just ignore everything but the final span.

model/converter/json/fixtures/es_01-actual.json
```diff
@@ -0,0 +1,91 @@
+{
```
yurishkuro: this should not be checked in, it's created by the unit test for debugging (when there is a mismatch)

model/converter/json/fixtures/ui_01.json
```diff
@@ -70,7 +71,8 @@
       ],
       "logs": [],
       "processID": "p1",
-      "warnings": null
+      "warnings": null,
+      "incomplete": false
```
yurishkuro: why do we need to add this? Isn't false going to be the default value anyway?

model/proto/model.proto
```diff
@@ -134,6 +134,8 @@ message Span {
     (gogoproto.customname) = "ProcessID"
   ];
   repeated string warnings = 12;
+  //indicates whether this is the final span or an intermediate span
```
yurishkuro: `// indicates that the span has not been finished yet and only contains partial data`

phal0r:
@yurishkuro 
I will integrate the changes. One question to your workflow: How do you manage local build artefacts, that should not be commited to git? We usually use .gitignore for that, but the files (the linux binary and the generated json from the tests) are not part of it.

model/adjuster/span_merger.go
yurishkuro: I am reluctant to rely on the client library doing the exact right thing for this to work. Yes, if the client lib clears all collections (refs, logs, & tags) after each flush, then the simple merge is fine, but what if it doesn't?

model/adjuster/span_merger.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+	if len(warnings) > 0 {
+		finalSpan.Warnings = append(warnings, finalSpan.GetWarnings()...)
+	}
+
+	return finalSpan
+}
+
+func isMergeable(spans []*model.Span) bool {
+	// Checks that span.kind annotations are consistent, i.e all spans contain server/client or no span kind annotations
```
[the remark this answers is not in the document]
phal0r: I think this checks for the special case, that zipkin spans uses the same spanId for server and client spans in a rpc call. Those will be grouped by the grouping logic of the merger, but will be ignored when actually merging the spans. Maybe the comment is misleading here. This is derived from https://github.com/jaegertracing/jaeger/pull/728, so @vprithvi should advise what to do. I would say this check is necessary to not accidentely merge zipkin spans.

model/converter/json/fixtures/es_01-actual.json
phal0r: Yes, will remove it and add it to gitingore

model/converter/json/fixtures/ui_01.json
phal0r: Yes, but the fixtures are read as bytes and compared on byte level with the output of an encoded trace. They are not transformed to an actual trace. This is why I need to add it.

model/proto/model.proto
phal0r: Yeah, sounds better ;)

model/adjuster/span_merger.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+func groupByIDs(spans []*model.Span) map[model.SpanID][]*model.Span {
+	IDToSpans := make(map[model.SpanID][]*model.Span)
+	for _, span := range spans {
+		if spans, ok := IDToSpans[span.SpanID]; ok {
+			IDToSpans[span.SpanID] = append(spans, span)
+		} else {
+			IDToSpans[span.SpanID] = []*model.Span{span}
+		}
```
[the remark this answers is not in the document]
phal0r: ok

model/adjuster/span_merger.go
phal0r: Yes, consistency is key. Will change.

model/adjuster/span_merger.go
phal0r: Yes, agreed, but what if we pick the final span and the client would implement it the other way around, thus flushing the span and only keeping spanId, traceId and startTime? We would never see any tags. that are flushed intermediately. 

So, I think there should be a spec for client implementors on how to correctly flush intermediate spans. IMHO the client needs to handle intermediate spans correctly, doesn't matter, which path we choose or do you think there is a one-size-fits-all solution for the merger?

model/adjuster/span_merger_test.go
```diff
@@ -0,0 +1,345 @@
...
+						Warnings: []string{"First Warning", "Second Warning", "Third Warning"},
+					},
+					{
+						SpanID: model.SpanID(2),
+					},
+				},
+			},
+		},
+		{
+			name: "duplicate Zipkin spans: don't merge",
```
yurishkuro: I would prefer this to be (a) a separate test, and (b) use the actual Zipkin format (e.g. zipkin.thrift) as the input, with the full chain of conversion & adjusters, especially including the deduper.

model/adjuster/span_merger.go
```diff
@@ -0,0 +1,112 @@
+// Copyright (c) 2018 The Jaeger Authors.
```
yurishkuro: there should be no UI merge in this PR - probably needs a rebase off the master.

IurmanJ:
@phal0r @yurishkuro 
Any update on this PR? Also, from what I understood, it would introduce an "incomplete" status for a span that would allow sending chunks of data before it ends. However, I've got a use-case where I'd need to "attach" some data (tags, for instance) to a span, even if this one is already finished (and so, complete). So, even if the (entire) span is already stored in the DB, I'd like to add some tags to it. Do you think it could be easily added? Actually, I don't think that this PR would allow this, hence my question.

yurishkuro:
@IurmanJ this is already possible, although not all clients support that. The backend already deals with multiple records having the same span ID, by merging them at query time into a single span. Some Jaeger clients (Go, C++) support so-called SELF span reference, which allows you to create a span with a known span ID, so when that span is finished and stored in the backend, it can enrich the existing span(s).

IurmanJ:
@yurishkuro Nice, I didn't know this feature was supported. Could you point me to an example? I suppose that it is not implemented in OpenTelemetry, right? In that case, maybe I could handle the request myself, so is there a link to the API for the self span reference? Can you confirm that it is possible for my use-case: I receive a span ID (nothing else) and I want to attach tags to it, whether it is "complete" or not.

yurishkuro:
See https://github.com/jaegertracing/jaeger-client-go#selfref. But you need to be careful with the timestamps and provide them explicitly to start/finish, otherwise the backend will min/max them across all span records.

IurmanJ:
@yurishkuro Thx! However, I have a doubt about the span context as I don't really have one. Can I just craft it with { traceID, spanID } and pass it to SelfRef, without knowing what the parentID is?

> But you need to be careful with the timestamps and provide them explicitly to start/finish, otherwise the backend will min/max them across all span records.

What exactly do you mean by "explicitly to start/finish"? Not sure I understand what you mean.

yurishkuro:
Yes, you can create span context manually, without parent ID.

You need to pass the right timestamps to start/finish functions.

IurmanJ:
> You need to pass the right timestamps to start/finish functions.

Sorry, I wasn't clear. I was more asking *what* the timestamps value should be, as you mentioned "explicitly" and now "right timestamps". Also, I don't see any parameters for start/stop functions, I think I'm not looking at the right place. Could you point me those functions in the repo?

cypris:
@phal0r @yurishkuro any updates? Having this feature would solve a big headache I have right now, and the only alternative I have at the moment is to write a bunch of custom stuff that won’t be as good as just using Jaeger to do my tracing for long running spans

phal0r:
> @phal0r @yurishkuro any updates? Having this feature would solve a big headache I have right now, and the only alternative I have at the moment is to write a bunch of custom stuff that won’t be as good as just using Jaeger to do my tracing for long running spans

@cypris 
I did not work on this change since 01/2019. @yurishkuro mentioned in this thread, that it is totally possible to send the same span multiple times and add more data to it. Jaeger takes care to merge them to one single span, when querying it. The only notable change of this pull request would be a "native" partial flag, which marks spans as "in progress". This could be achieved with custom attributes aswell. Currently I do not intend to continue working on this request.

jpkrohling:
Closing due to inactivity.

schickling:
Is there any kind of successor to this in terms of functionality?

jpkrohling:
Not that I'm aware of.

PR #980, opened 2018-08-09
WIP: Change ES tag schema to support tag search in Kibana

### LINKED ISSUE #906: Span format is not well suited for ES

```
{
  "_type": "span",
  "_version": 1,
  "_score": null,
  "_source": {
    "spanID": "274c0d9d4fc76848",
    "flags": 1,
    "operationName": "/",
    "references": [
      {
        "refType": "CHILD_OF",
        "spanID": "f229796cda43df60"
      }
    ],
    "duration": 1798,
    "tags": [
      {
        "key": "component",
        "type": "string",
        "value": "nginx"
      },
      {
        "key": "nginx.worker_pid",
        "type": "string",
        "value": "10767"
      },
      {
        "key": "peer.address",
        "type": "string",
        "value": "10.244.4.100:37542"
      },
      {
        "key": "http.method",
        "type": "string",
        "value": "POST"
      },
      {
        "key": "http.url",
        "type": "string",
        "value": "YYY"
      },
      {
        "key": "http.host",
        "type": "string",
        "value": "XXX"
      },
      {
        "key": "http.status_code",
        "type": "int64",
        "value": "204"
      },
      {
        "key": "http.status_line",
        "type": "string",
        "value": "204 No Content"
      }
    ],
    "logs": [],
    "processID": "",
    "process": {
      "tags": [
        {
          "key": "jaeger.version",
          "type": "string",
          "value": "C++-0.2.0"
        },
        {
          "key": "hostname",
          "type": "string",
          "value": "vega"
        },
        {
          "key": "ip",
          "type": "string",
          "value": "127.0.0.1"
        }
      ]
    },
    "warnings": null,
  },
  "fields": {
    "startTimeMillis": [
      "2018-07-02T23:56:03.255Z"
    ]
  },
  "sort": [
    1530575763255
  ]
}
```

pavolloffay:
Resolves #906 

Example of the index:
```
tags: {
  "http:method": "GET",
}
```

Significant changes
*  defines span and process tags as object datatype https://www.elastic.co/guide/en/elasticsearch/reference/current/object.html
* dots in tag keys are replaced with `:`

Limitations:
* by default ~1000 unique keys per index `index.mapping.total_fields.limit` https://www.elastic.co/guide/en/elasticsearch/reference/master/mapping.html#mapping-limit-settings. The count includes also meta fields https://github.com/elastic/elasticsearch/issues/24096
* issues if tag key contains `:`

#### TODO

- [x] performance comparison. 
- [ ]  fix `index.mapping.total_fields.limit` - we could count unique tag names, wait for failed PUT and then store following tags differently - array/nested object
- [ ] store type of the tag

### DIFF
--- a/model/json/model.go
+++ b/model/json/model.go
@@ -61,19 +61,20 @@ type Trace struct {
 // When converting to ES model, ProcessID and Warnings should be omitted. Even if
 // included, ES with dynamic settings off will automatically ignore unneeded fields.
 type Span struct {
-	TraceID       TraceID     `json:"traceID"`
-	SpanID        SpanID      `json:"spanID"`
-	ParentSpanID  SpanID      `json:"parentSpanID,omitempty"` // deprecated
-	Flags         uint32      `json:"flags,omitempty"`
-	OperationName string      `json:"operationName"`
-	References    []Reference `json:"references"`
-	StartTime     uint64      `json:"startTime"` // microseconds since Unix epoch
-	Duration      uint64      `json:"duration"`  // microseconds
-	Tags          []KeyValue  `json:"tags"`
-	Logs          []Log       `json:"logs"`
-	ProcessID     ProcessID   `json:"processID,omitempty"`
-	Process       *Process    `json:"process,omitempty"`
-	Warnings      []string    `json:"warnings"`
+	TraceID       TraceID           `json:"traceID"`
+	SpanID        SpanID            `json:"spanID"`
+	ParentSpanID  SpanID            `json:"parentSpanID,omitempty"` // deprecated
+	Flags         uint32            `json:"flags,omitempty"`
+	OperationName string            `json:"operationName"`
+	References    []Reference       `json:"references"`
+	StartTime     uint64            `json:"startTime"` // microseconds since Unix epoch
+	Duration      uint64            `json:"duration"`  // microseconds
+	Tags          []KeyValue        `json:"tags"`
+	TagsMap       map[string]string `json:"tagsMap,omitempty"`
+	Logs          []Log             `json:"logs"`
+	ProcessID     ProcessID         `json:"processID,omitempty"`
+	Process       *Process          `json:"process,omitempty"`
+	Warnings      []string          `json:"warnings"`
 }
 
 // Reference is a reference from one span to another
@@ -85,8 +86,9 @@ type Reference struct {
 
 // Process is the process emitting a set of spans
 type Process struct {
-	ServiceName string     `json:"serviceName"`
-	Tags        []KeyValue `json:"tags"`
+	ServiceName string            `json:"serviceName"`
+	Tags        []KeyValue        `json:"tags"`
+	TagsMap     map[string]string `json:"tagsMap,omitempty"`
 }
 
 // Log is a log emitted in a span
--- a/plugin/storage/es/spanstore/reader.go
+++ b/plugin/storage/es/spanstore/reader.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
@@ -43,8 +44,8 @@ const (
 	startTimeField     = "startTime"
 	serviceNameField   = "process.serviceName"
 	operationNameField = "operationName"
-	tagsField          = "tags"
-	processTagsField   = "process.tags"
+	tagsField          = "tagsMap"
+	processTagsField   = "process.tagsMap"
 	logFieldsField     = "logs.fields"
 	tagKeyField        = "key"
 	tagValueField      = "value"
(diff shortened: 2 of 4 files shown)


pavolloffay:
We might choose different replacing character. Kibana uses `:` as `=` https://www.elastic.co/guide/en/elasticsearch/reference/5.6/query-dsl-query-string-query.html#_field_names

pavolloffay:
This mapping allows storing ~476 unique tag keys with the default `index.mapping.total_fields.limit` set to 1000. There are some metafields `_index, _id` (12) and since we use dynamic mapping each field creates `.raw` field hence I was able to create on 476 unique tags https://github.com/elastic/elasticsearch/issues/24096#issuecomment-324950606. 

@mabn @kacper-jackiewicz @Monnoroch do you know if raw field can be disabled? Or any possible workaround?

pavolloffay:
Performance results https://github.com/pavolloffay/jaeger-perf-tests

reported 300k spans
`QUERY_FROM=jaeger-query JAEGER_QUERY_ASYNC=true  NUMBER_OF_SPANS=50000 NUM_OF_TRACERS=6 JAEGER_QUERY_LIMIT=50000 mvn clean package exec:java`
 https://pastebin.com/7tKiQnuC

Query time of 
`http://localhost:16686/api/traces?service=perf-test-thread-0&limit=50000&lookback=1h&tags={"fooo.bar*?%http.d6cconald":"hehuhoh$?ij","fooo.ba2sar":"true","fooo.ba4342r":"1","fooo.bar1":"fobarhax*+??","fooo.bar*?%http.do**2nald":"goobarRAXbaz","fooo.bar*?%http.don(a44ld":"goobarRAXbaz","fooo.ba24r*?%":"hehe"}`

is about  13363.90 milliseconds. The number of tags in query does not affect the time. The query time is mostly affected byr limit parameter. With limit 20 the query takes: mean = 17.88 milliseconds.

plugin/storage/es/spanstore/reader.go
```diff
@@ -441,10 +454,12 @@ func (s *SpanReader) buildOperationNameQuery(operationName string) elastic.Query
 }
 
 func (s *SpanReader) buildTagQuery(k string, v string) elastic.Query {
-	queries := make([]elastic.Query, len(tagFieldList))
-	for i := range queries {
-		queries[i] = s.buildNestedQuery(tagFieldList[i], k, v)
+	queries := make([]elastic.Query, len(objectTagFieldList)+1)
+	kd := strings.Replace(k, ".", ":", -1)
```
Monnoroch: This replace is duplicated in many places. Any way it can be extracted as a helper method with a nice name?

plugin/storage/es/spanstore/schema.go
```diff
@@ -139,6 +142,9 @@ var (
 				}
 			}
 		},
+		"tagsMap": {
```
Monnoroch: Your example in the description says:
```
tags: {
  "http:method": "GET",
}
```

However the code says "tagsMap".

model/json/model.go
```diff
@@ -85,8 +86,9 @@ type Reference struct {
 
 // Process is the process emitting a set of spans
 type Process struct {
-	ServiceName string     `json:"serviceName"`
-	Tags        []KeyValue `json:"tags"`
+	ServiceName string            `json:"serviceName"`
+	Tags        []KeyValue        `json:"tags"`
+	TagsMap     map[string]string `json:"tagsMap,omitempty"`
```
Monnoroch: I'm not a maintainer, but if it was me, I would have a separate "span" model that is independent of the backend and then I would map it to a backend specific model in the backend specific package. It does not look right to me that you have to modify the common model just to change ES representation.

plugin/storage/es/spanstore/writer.go
```diff
@@ -129,10 +129,26 @@ func (s *SpanWriter) WriteSpan(span *model.Span) error {
...
+	jsonSpan.Process.Tags = nil
 	s.writeSpan(spanIndexName, jsonSpan)
 	return nil
 }
 
+func (s *SpanWriter) tagsToMap(kvs []model.KeyValue) map[string]string {
+	tags := map[string]string{}
+	for _, tag := range kvs {
+		if strings.Contains(tag.Key, ":") {
+			s.logger.Warn("Tag key contains \\':\\', at the query time it will be transformed to \\'.\\'")
```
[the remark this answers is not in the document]
Monnoroch: Sorry, in that case the implementation makes sense apart from my question below. The only thing I would like is to have a way to configure the whitelisted/blacklisted tags somehow. The scenario I have in mind is you first would whitelist all of them and then if your service grows enough to have more and more flags you can blacklist some of them, and eventually move to a whitelist if there are too many of them. Though I realize that it's not trivial to reconfigure in a way that will not break indexing. Or maybe it is and just needs to be confirmed? I would think that ES was designed with logs that change over time in mind.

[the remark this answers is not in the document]
pavolloffay: Once you reach the limit you are not able to store more data in the given index. You could perhaps change the mapping and reindex but probably nobody would want to do that. 

Blacklisting does not make much sense for me. If you are above the limit it might be very inconvenient to list all blacklisted tags. 

I want something that works OOTB for all users especially with high cardinality tags. The other thing is that kibana is not primarily supported system for us.

PR #984, opened 2018-08-13
Pin proto / gogo versions for repeatable builds

yurishkuro:
Undo some upgrades introduced in #940. A short term fix to #985.


vprithvi:
I think this test is unstable, I was able to get it to fail locally without the protobuf changes.

isaachier:
According to @vprithvi we should reconsider #985.

PR #986, opened 2018-08-13
Fix logging in the integration tests

yurishkuro:
Pulling these changes out of #984.

### DIFF
--- a/plugin/storage/integration/integration_test.go
+++ b/plugin/storage/integration/integration_test.go
@@ -29,7 +29,6 @@ import (
 	"github.com/gogo/protobuf/proto"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	"go.uber.org/zap"
 
 	"github.com/jaegertracing/jaeger/model"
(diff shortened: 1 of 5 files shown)


plugin/storage/integration/integration_test.go
```diff
@@ -179,20 +173,21 @@ func (s *StorageIntegration) testGetTrace(t *testing.T) {
...
 	var actual *model.Trace
-	for i := 0; i < iterations; i++ {
-		s.logger.Info(fmt.Sprintf(waitForBackendComment, i+1, iterations))
+	found := s.waitForCondition(t, func(t *testing.T) bool {
 		var err error
 		actual, err = s.SpanReader.GetTrace(expectedTraceID)
-		if err == nil && len(actual.Spans) == len(expected.Spans) {
-			break
+		if err != nil {
+			t.Log(err)
```
black-adder: why does this test get special treatment? the other tests aren't logging the errors

yurishkuro: maybe they all should, but this is specifically the one that was sporadically failing yesterday for Kafka, so I want to get the logging into it for the future.

PR #987, opened 2018-08-13
Shutdown sarama consumer cleanly

vprithvi:
### Cleanly shutdown Sarama parent consumer

Shutdown all partitions before shutting down the sarama consumer. This sidesteps   https://github.com/bsm/sarama-cluster/issues/255 and ensures that the shutdown completes in a reasonable timeframe. 

### Wait for PartitionConsumer shutdown before consuming messages

A given host could be reassigned the same partition quickly. In this case, we wait for proper shutdown and draining of the messages queue to prevent committing offsets for more recent messages before processing older ones. 

While this solution works on one host, it doesn't scale to multiple hosts because `sarama-cluster` does not provide an API to complete actions after a rebalance is triggered, but before subscriptions are released. This is being addressed as part of https://github.com/Shopify/sarama/pull/1099. 

### Use Sarama's PartitionConsumer mock instead of relying on our own because it is richer and well tested. 

### DIFF
--- a/cmd/ingester/app/consumer/consumer.go
+++ b/cmd/ingester/app/consumer/consumer.go
@@ -15,8 +15,6 @@
 package consumer
 
 import (
-	"sync"
-
 	"github.com/Shopify/sarama"
 	sc "github.com/bsm/sarama-cluster"
 	"github.com/uber/jaeger-lib/metrics"
@@ -24,6 +22,7 @@ import (
 
 	"github.com/jaegertracing/jaeger/cmd/ingester/app/processor"
 	"github.com/jaegertracing/jaeger/pkg/kafka/consumer"
+	"sync"
 )
 
 // Params are the parameters of a Consumer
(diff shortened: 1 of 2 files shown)


cmd/ingester/app/consumer/consumer.go
```diff
@@ -42,55 +42,55 @@ type Consumer struct {
...
 
 // Start begins consuming messages in a go routine
 func (c *Consumer) Start() {
-	c.isClosed.Add(1)
-	c.logger.Info("Starting main loop")
-	go c.mainLoop()
+	go func() {
+		c.logger.Info("Starting main loop")
+		for pc := range c.internalConsumer.Partitions() {
+			if p, ok := c.partitionIDToState[pc.Partition()]; ok {
```
black-adder: I'd add a comment here to describe why this is needed.

cmd/ingester/app/consumer/consumer.go
```diff
@@ -42,55 +42,55 @@ type Consumer struct {
...
+				p.state.Wait()
+				delete(c.partitionIDToState, pc.Partition())
```
black-adder: technically not needed as the line below overwrites it, but being explicit is good too

cmd/ingester/app/consumer/consumer.go
```diff
@@ -109,6 +109,7 @@ func (c *Consumer) handleMessages(pc sc.PartitionConsumer) {
 
 		msgProcessor.Process(&saramaMessageWrapper{msg})
 	}
+	c.logger.Debug("Finished handling messages", zap.Int32("partition", pc.Partition()))
```
black-adder: This can be Info?

cmd/ingester/app/consumer/consumer.go
```diff
@@ -42,55 +42,55 @@ type Consumer struct {
 	internalConsumer consumer.Consumer
 	processorFactory ProcessorFactory
 
-	close    chan struct{}
-	isClosed sync.WaitGroup
+	partitionIDToState map[int32]*consumerState
+}
+
+type consumerState struct {
+	state             sync.WaitGroup
```
black-adder: not a big fan of this name

cmd/ingester/app/consumer/consumer_test.go
```diff
@@ -15,157 +15,187 @@
...
+	mc.ExpectErrorsDrainedOnClose()
+
+	saramaPartitionConsumer, e := saramaConsumer.ConsumePartition(topic, partition, msgOffset)
+	require.NoError(t, e)
+
+	undertest := newConsumer(localFactory, topic, &pmocks.SpanProcessor{}, newSaramaClusterConsumer(saramaPartitionConsumer))
+
+	undertest.Start()
+	mc.YieldError(errors.New("Daisy, Daisy"))
+	time.Sleep(100 * time.Millisecond)
```
black-adder: any other condition we can wait on?

vprithvi: We can wait on the metric changing, but that would involve a sleep/poll as well. Processing errors only prints out a log and increments a counter, so not sure we can eliminate this sleep entirely in tests.

black-adder: can we keep checking the metric to see if it's incremented?

vprithvi: We can poll the metric assertion until it's true, but do we need to? (I don't think we do such polling on other tests)

PR #1003, opened 2018-08-17
Use reliable bad address to guarantee failure

### LINKED ISSUE #1002: MockCollector test bug

On my desktop (Arch Linux/Manjaro), I consistently get the following test error:
```
Error Trace:	mock_collector_test.go:162
			Error:      	An error is expected but got nil.
			Test:       	TestMockTCollectorErrors
			Messages:   	error because of bad address
```

The test expects binding to 127.0.0 to fail. After some digging with strace I discovered that somehow the system extends the supposedly broken address of 127.0.0 to be 127.0.0.0. Then, strace shows this:
`bind(3, {sa_family=AF_INET, sin_port=htons(0), sin_addr=inet_addr("127.0.0.0")}, 16) = 0`.

Remove the assertion that requires this bind to fail seeing as it is not portable/correct.

Find out why this is OK in Travis but not my version of Go (1.10.3) and Linux (Linux 4.18.0-1-MANJARO #1 SMP PREEMPT x86_64 GNU/Linux).

isaachier:
-  Resolves #1002.
- Use `256` to overflow IP address.

### DIFF
--- a/cmd/agent/app/testutils/mock_collector_test.go
+++ b/cmd/agent/app/testutils/mock_collector_test.go
@@ -158,6 +158,6 @@ func TestMockTCollectorErrors(t *testing.T) {
 	_, err := startMockTCollector("", "127.0.0.1:0")
 	assert.Error(t, err, "error because of empty service name")
 
-	_, err = startMockTCollector("test", "127.0.0:0")
+	_, err = startMockTCollector("test", "256.0.0:0")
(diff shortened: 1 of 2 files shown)


isaachier:
Kafka integration test is breaking build here. @davit-y @vprithvi any ideas?

vprithvi:
@isaachier none yet, created #1004 to track this.

PR #1005, opened 2018-08-17
Use netcat to detect when Kafka is up

### LINKED ISSUE #1004: Fix flaky Kafka integration test
**TestKafkaStorage/GetTrace** fails and causes the build to fail. Update the test to be reliable. 

```
--- FAIL: TestKafkaStorage (6.18s)
    --- FAIL: TestKafkaStorage/GetTrace (3.02s)
    	integration_test.go:99: Waiting for storage backend to update documents, iteration 1 out of 30
    	integration_test.go:180: Trace was not found
    	. 
        .
        .
    	integration_test.go:99: Waiting for storage backend to update documents, iteration 30 out of 30
    	integration_test.go:180: Trace was not found
    	integration_test.go:180: Trace was not found
    	assertions.go:256: 
    			Error Trace:	integration_test.go:184
    			            				integration_test.go:377
    			Error:      	Should be true
    			Test:       	TestKafkaStorage/GetTrace
    	assertions.go:256: 
    			Error Trace:	domain_trace_compare_test.go:50
    			            				integration_test.go:185
    			            				integration_test.go:377
    			Error:      	Expected value not to be nil.
    			Test:       	TestKafkaStorage/GetTrace
```

isaachier:
- Fixes #1004 (hopefully). 

- Use netcat to wait until Kafka is up before beginning integration test.

### DIFF
--- a/scripts/travis/kafka-integration-test.sh
+++ b/scripts/travis/kafka-integration-test.sh
@@ -4,6 +4,14 @@ set -e
 
 docker pull spotify/kafka
 CID=$(docker run -d -p 2181:2181 -p 9092:9092 --env ADVERTISED_HOST=localhost --env ADVERTISED_PORT=9092 --rm spotify/kafka)
+
+# Guarantees no matter what happens, docker will remove the instance at the end.
+trap 'docker rm -f $CID 2>/dev/null' EXIT INT TERM
+
 export STORAGE=kafka
+while true; do
+    if nc -z localhost 9092; then
+        break
+    fi
+done
 make storage-integration-test
-docker kill $CID


scripts/travis/kafka-integration-test.sh
```diff
@@ -4,6 +4,14 @@ set -e
 
 docker pull spotify/kafka
 CID=$(docker run -d -p 2181:2181 -p 9092:9092 --env ADVERTISED_HOST=localhost --env ADVERTISED_PORT=9092 --rm spotify/kafka)
+
+# Guarantees no matter what happens, docker will remove the instance at the end.
+trap 'docker rm -f $CID 2>/dev/null' EXIT INT TERM
+
 export STORAGE=kafka
+while true; do
+    if nc -z localhost 9092; then
```
jpkrohling: Does it fail immediately if the port is not up, or does it block for a few ms/seconds?

yurishkuro: Looks like it blocks forever until port is available. The build still failed though.

isaachier: It claims to "Specifies that nc should just scan for listening daemons, without sending any data to them." (https://linux.die.net/man/1/nc). I was checking source to see details, but will get back to you on that. Meanwhile, will investigate build issue.

isaachier: Build didn't fail. @black-adder canceled every test but the kafka test and it has succeeded at least once. He said he would retry it a few more times to guarantee it works.

isaachier: From a practical viewpoint, when I run `nc -z localhost 9092` on my MacBook, and Kafka is not running, the command returns within 10 ms (using `time` on 100 calls in a row gives about 500 ms).

black-adder:
I ran the kafka integration test 7 times in a row and all passed. I think this is solid.

Issue #1017, opened 2018-08-23
Use json/dbmodel for ES dependencies


pavolloffay:
At the moment ES dependency storage uses `model.Dependencies`. It should use a separate model.

PR #1018, opened 2018-08-23
Allow storing tags as object fields in ES - better kibana support

### LINKED ISSUE #906: Span format is not well suited for ES

pavolloffay:
Resolves #906 
Supersedes #980, #982 

* Store configured or all tags as object fields. For example: `{"tag": {"http@method": "GET", "span@kind": "server"}}`. 
* Binary tags are always stored as nested objects - there is no way how to reproduce the type - it would be always a string.
* The default behaviour does not use tag fields. It's a subject to a change. By default we could store only standard tags as object field - no breaking change
* with default `index.mapping.total_fields.limit"` it is possible to store ~**940 unique span tag keys**. **If the limit is reached no new unique tag can be stored** - ES throws an exception. _**It basically means span will not be stored in the current index**_
* tag with float value `1.0` will be always returned as int `123` https://github.com/olivere/elastic/issues/891

Introduced 3 flags:
```bash
      --es.tag-as-fields                                      Store all span and process tags as object fields. If true .tag-file-path is ignored.
      --es.tag-file-path string                               Optional path to a file containing tag keys which will be stored as object fields. Each key should be on a separate line.
      --es.tag-key-de-dot-char string                         The character used to replace dots (".") in tag keys stored as object fields. (default "@")

```

Other changes
* copied json model to `plugin/storage/es/spanwriter/dbmodel`

Todos after this is merged:
* move `Service` struct from writer to dbmodel package
* https://github.com/jaegertracing/jaeger/issues/1017

Adding screenshot from kibana. The search pattern is `tag.span@kind:server AND process.tag.ip:10.33.144.223`
screenshot of kibana


model/json/model.go
```diff
@@ -85,8 +86,9 @@ type Reference struct {
 
 // Process is the process emitting a set of spans
 type Process struct {
-	ServiceName string     `json:"serviceName"`
-	Tags        []KeyValue `json:"tags"`
+	ServiceName string                 `json:"serviceName"`
+	Tags        []KeyValue             `json:"tags"`
+	TagsMap     map[string]interface{} `json:"tagsMap,omitempty"`
```
pavolloffay: Please comment if you know a better name.

plugin/storage/es/spanstore/dbmodel/to_domain.go
```diff
@@ -126,10 +126,40 @@ func (td toDomain) convertKeyValues(tags []json.KeyValue) ([]model.KeyValue, err
...
 }
 
+func addTagsMap(tagsMap map[string]interface{}) []model.KeyValue {
+	kvs := make([]model.KeyValue, len(tagsMap))
+	i := 0
+	for k, v := range tagsMap {
+		dKey := strings.Replace(k, ":", ".", -1)
+		tag := model.KeyValue{Key: dKey}
+		// TODO v.(int/64/32) didn't work it is always a float
+		if pInt, err := strconv.ParseInt(fmt.Sprintf("%v", v), 10, 64); err == nil {
```
pavolloffay: ES always returns float64 as a number - even if it was stored as int. This works but it's not that pretty.

plugin/storage/es/spanstore/reader.go
```diff
@@ -444,10 +456,17 @@ func (s *SpanReader) buildOperationNameQuery(operationName string) elastic.Query
...
+	kd := strings.Replace(k, ".", ":", -1)
+	for i := range objectTagFieldList {
+		queries[i] = s.buildObjectQuery(objectTagFieldList[i], kd, v)
+	}
+	for i := range nestedTagFieldList {
+		queries[i+len(objectTagFieldList)] = s.buildNestedQuery(nestedTagFieldList[i], k, v)
 	}
+
+	// TODO we could query tagsMap only if the tag is configured
+	// but configuration can change over time
```
pavolloffay: I mean what if previous deployment was storing tags as object fields. If the new deployment is not using tags as fields query would not find old tags.

model/json/model.go
objectiser: Ideally the name used in the kibana query would be `tags`, but I guess then that may cause backward compatibility issues?
Another alternative would be to use `tag` instead of `tagsMap` - so when referenced it would be `tag.orderId` (for example).

pavolloffay: Yes we cannot use `tags` as it is used for legacy tags (as nested objects).

+1 on `tag`.

pavolloffay:
This is ready for review

plugin/storage/es/options.go
```diff
@@ -75,6 +78,7 @@ func NewOptions(primaryNamespace string, otherNamespaces ...string) *Options {
 				BulkWorkers:       1,
 				BulkActions:       1000,
 				BulkFlushInterval: time.Millisecond * 200,
+				TagDeDotChar:      "@",
```
objectiser: What was the issue with using ':'?

From my ES past, I vaguely remember there being able to mark a field as not_analyzed so it was indexed but using the actual value without being processed. I see that this has been replaced by the concept of keyword in 5.x. Doesn't this avoid the issue of using '.' in the tag name?

plugin/storage/es/options.go
```diff
@@ -37,6 +37,9 @@ const (
 	suffixBulkActions       = ".bulk.actions"
 	suffixBulkFlushInterval = ".bulk.flush-interval"
 	suffixIndexPrefix       = ".index-prefix"
+	suffixTagsFile          = ".tag-file-path"
+	suffixAllTagsAsObject   = ".tag-as-fields"
+	suffixTagDeDotChar      = ".tag-key-de-dot-char"
```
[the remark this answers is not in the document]
pavolloffay: This is usually called de dot https://www.elastic.co/blog/introducing-the-de_dot-filter

plugin/storage/es/options.go
pavolloffay: > What was the issue with using ':'?

It would have to be escaped. It's used in kibana query `tag.key:foo`. To avoid dots in field name we would have to set `enabled:false` on the field.

pavolloffay:
I have published docker images in https://hub.docker.com/u/pavolloffay/ with tag PR1018. For example `pavolloffay:jaeger-collector:PR1018`. Or use `docker run --net=host -e SPAN_STORAGE_TYPE=elasticsearch -e ES_TAG_AS_FIELDS=true  pavolloffay/all-in-one:PR1018` to simply run it (it expects ES to be running on the host).

plugin/storage/es/spanstore/dbmodel/from_domain.go
```diff
@@ -0,0 +1,137 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
```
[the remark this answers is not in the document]
pavolloffay: what is wrong with it? It's automatically handled by `make fmt`

plugin/storage/es/spanstore/dbmodel/model.go
```diff
@@ -0,0 +1,102 @@
...
+	// BinaryType indicates an arbitrary byte array stored in KeyValue
+	BinaryType ValueType = "binary"
+)
+
+// Span is a span denoting a piece of work in some infrastructure
+// When converting to UI model, ParentSpanID and Process should be dereferenced into
+// References and ProcessID, respectively.
+// When converting to ES model, ProcessID and Warnings should be omitted. Even if
+// included, ES with dynamic settings off will automatically ignore unneeded fields.
+type Span struct {
```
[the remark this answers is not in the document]
pavolloffay: yes it's in es/dbmodel package. I think the comment is not accurate. I will fix it

plugin/storage/es/spanstore/dbmodel/from_domain.go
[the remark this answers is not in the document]
jpkrohling: s/2017 Uber Technologies, Inc/2018 The Jaeger Authors/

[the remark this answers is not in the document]
pavolloffay: The `updadeLicense` script seems to be correct. Actually these files were copied (with changes) from domain package. Do you want to change the license?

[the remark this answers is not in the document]
jpkrohling: If they are just copies, then no, keep the old copyright.

plugin/storage/es/spanstore/dbmodel/from_domain_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+		var span model.Span
+		require.NoError(t, jsonpb.Unmarshal(bytes.NewReader(domainStr), &span))
+		converter := NewFromDomain(false, nil, ":")
+		embeddedSpan := converter.FromDomainEmbedProcess(&span)
+
+		var expectedSpan Span
+		require.NoError(t, json.Unmarshal(jsonStr, &expectedSpan))
+
+		testJSONEncoding(t, i, jsonStr, embeddedSpan)
```
jpkrohling: Could you wrap each test in a t.Run() with an appropriate name? Like, "TestFromDomainEmbedProcess#fixture%d"

plugin/storage/es/spanstore/dbmodel/to_domain.go
```diff
@@ -12,27 +12,35 @@
...
-func SpanToDomain(span *json.Span) (*model.Span, error) {
-	return toDomain{}.spanToDomain(span)
+// NewToDomain creates ToDomain
+func NewToDomain(deDotChar string) ToDomain {
+	return toDomain{deDotChar: deDotChar}
 }
 
-type toDomain struct{}
+// ToDomain is used to convert Span to model.Span
+type ToDomain interface {
```
jpkrohling: Same question as before: is this specific to ES? Couldn't this be made at a higher level, to be used by other storage backends? Otherwise, what's the point of changing from a struct to an interface?

plugin/storage/es/spanstore/dbmodel/to_domain.go
```diff
@@ -126,10 +134,40 @@ func (td toDomain) convertKeyValues(tags []json.KeyValue) ([]model.KeyValue, err
 	return retMe, nil
 }
 
+func addTagsMap(tagsMap map[string]interface{}, deDotChar string) []model.KeyValue {
+	kvs := make([]model.KeyValue, len(tagsMap))
+	i := 0
+	for k, v := range tagsMap {
+		dKey := strings.Replace(k, deDotChar, ".", -1)
+		tag := model.KeyValue{Key: dKey}
+		// TODO v.(int/64/32) didn't work it is always a float
```
jpkrohling: What's the TODO here? Sounds more like a regular comment to me :)

plugin/storage/es/spanstore/dbmodel/to_domain.go
```diff
@@ -126,10 +134,40 @@ func (td toDomain) convertKeyValues(tags []json.KeyValue) ([]model.KeyValue, err
...
+		if pInt, err := strconv.ParseInt(fmt.Sprintf("%v", v), 10, 64); err == nil {
+			tag.VType = model.Int64Type
+			tag.VInt64 = pInt
+		} else if val, ok := v.(float64); ok {
```
jpkrohling: Isn't it more idiomatic in Go to use a switch/case with the type?

```golang
switch v.(type) {
case float64:
  tag.VType = model.Float64Type
  tag.VFloat64 = val
  ...
}
```

plugin/storage/es/spanstore/dbmodel/to_domain.go
```diff
@@ -126,10 +134,40 @@ func (td toDomain) convertKeyValues(tags []json.KeyValue) ([]model.KeyValue, err
...
+			tag.VType = model.BoolType
+			tag.VBool = val
+		} else if val, ok := v.(string); ok {
+			tag.VType = model.StringType
+			tag.VStr = val
+			// it's always as string
+		} else if val, ok := v.([]byte); ok {
+			tag.VType = model.BinaryType
+			tag.VBinary = val
+		}
```
jpkrohling: In theory, it can also be something else. How about adding a log statement for that case?

plugin/storage/es/spanstore/dbmodel/to_domain_test.go
```diff
@@ -288,3 +288,34 @@ func TestFailureBadParentSpanID(t *testing.T) {
...
+		t.Logf("Actual trace: %s", string(out))
+	}
+}
+func TestTagsMap(t *testing.T) {
+	tagsMap := map[string]interface{}{}
+	tagsMap["bool:bool"] = true
+	tagsMap["int.int"] = int64(1)
+	tagsMap["float"] = float64(1.1)
+	tagsMap["str"] = "foo"
+	tagsMap["binary"] = []byte("baz")
```
jpkrohling: Add a case for a negative test as well, like by passing a custom type (or just a `interface{}` type)

plugin/storage/es/spanstore/reader.go
```diff
@@ -90,26 +94,39 @@ type SpanReader struct {
...
+	MetricsFactory          metrics.Factory
+	serviceOperationStorage *ServiceOperationStorage
+	IndexPrefix             string
+	TagDeDotChar            string
 }
 
 // NewSpanReader returns a new SpanReader with a metrics.
-func NewSpanReader(client es.Client, logger *zap.Logger, maxLookback time.Duration, metricsFactory metrics.Factory, indexPrefix string) spanstore.Reader {
-	return storageMetrics.NewReadMetricsDecorator(newSpanReader(client, logger, maxLookback, indexPrefix), metricsFactory)
+func NewSpanReader(p SpanReaderParams) spanstore.Reader {
```
jpkrohling: Is this pattern being used elsewhere (passing a "params" object to a constructor)?

plugin/storage/es/spanstore/reader.go
```diff
@@ -444,10 +469,17 @@ func (s *SpanReader) buildOperationNameQuery(operationName string) elastic.Query
...
-	queries := make([]elastic.Query, len(tagFieldList))
-	for i := range queries {
-		queries[i] = s.buildNestedQuery(tagFieldList[i], k, v)
+	queries := make([]elastic.Query, len(nestedTagFieldList)+len(objectTagFieldList))
+	kd := strings.Replace(k, ".", ":", -1)
+	for i := range objectTagFieldList {
+		queries[i] = s.buildObjectQuery(objectTagFieldList[i], kd, v)
 	}
+	for i := range nestedTagFieldList {
+		queries[i+len(objectTagFieldList)] = s.buildNestedQuery(nestedTagFieldList[i], k, v)
```
jpkrohling: You don't need to compute `len(objectTagFieldList)` for every `nestedTagFieldList`

plugin/storage/es/spanstore/reader.go
```diff
@@ -444,10 +469,17 @@ func (s *SpanReader) buildOperationNameQuery(operationName string) elastic.Query
...
+	}
+
+	// TODO we could query tag only if the tag is configured
```
jpkrohling: This doesn't sounds like a "TODO".

plugin/storage/es/spanstore/reader_test.go
```diff
@@ -89,7 +90,12 @@ func withSpanReader(fn func(r *spanReaderTest)) {
 		client:    client,
 		logger:    logger,
 		logBuffer: logBuffer,
-		reader:    newSpanReader(client, logger, 72*time.Hour, ""),
+		reader: newSpanReader(SpanReaderParams{
```
jpkrohling: The previous notation is far more readable, IMO.

plugin/storage/es/spanstore/writer_test.go
```diff
@@ -48,7 +48,7 @@ func withSpanWriter(fn func(w *spanWriterTest)) {
 		client:    client,
 		logger:    logger,
 		logBuffer: logBuffer,
-		writer:    NewSpanWriter(client, logger, metricsFactory, 0, 0, ""),
+		writer:    NewSpanWriter(SpanWriterParams{Client: client, Logger: logger, MetricsFactory: metricsFactory}),
```
jpkrohling: One good side about the "params" object for constructors is that it make it act as a named parameter. Bonus points if default values can be omitted, as its the case here.

But I'm still not sure it's the right PR to introduce this pattern, if it's not being used elsewhere already (I don't remember seeing it, but might be wrong).

plugin/storage/es/spanstore/writer_test.go
```diff
@@ -292,21 +293,69 @@ func TestWriteSpanInternalError(t *testing.T) {
...
+					Type:  dbmodel.StringType,
+					Value: "baz",
+				}}}},
+		},
+	}
+
+	s := &model.Span{Tags: []model.KeyValue{{Key: "foo", VStr: "bar"}},
+		Process: &model.Process{Tags: []model.KeyValue{{Key: "bar", VStr: "baz"}}}}
+	for _, test := range testCases {
+		mSpan := test.writer.spanConverter.FromDomainEmbedProcess(s)
```
jpkrohling: Wrap every iteration with its own t.Run() with an appropriate name. In case of a failure, it should be easy to identify what's the combination that failed.

plugin/storage/integration/domain_trace_compare_test.go
```diff
@@ -34,11 +34,18 @@ func CompareSliceOfTraces(t *testing.T, expected []*model.Trace, actual []*model
 	}
 	if !assert.EqualValues(t, expected, actual) {
 		for _, err := range pretty.Diff(expected, actual) {
+			//fmt.Println("Actual")
```
jpkrohling: Will this be removed before merging?

plugin/storage/integration/elasticsearch_test.go
```diff
@@ -107,14 +124,22 @@ func healthCheck() error {
...
 		t.Fatal(err)
 	}
 	s := &ESStorageIntegration{}
-	require.NoError(t, s.initializeES())
+	require.NoError(t, s.initializeES(allTagsAsFields))
 	s.IntegrationTestAll(t)
 }
+
+func TestElasticsearchStorage(t *testing.T) {
+	testElasticsearchStorage(t, false)
```
jpkrohling: Why not a single test with a combination matrix, like you did with `TestNewSpanTags` ?

plugin/storage/es/spanstore/dbmodel/to_domain.go
pavolloffay: The biggest advantage is keeping the struct private and exposing only the interface with builder function. It makes it clear how it is constructed.

plugin/storage/es/spanstore/reader_test.go
pavolloffay: The param struct was requested in https://github.com/jaegertracing/jaeger/pull/1009#discussion_r211639597

plugin/storage/integration/elasticsearch_test.go
pavolloffay: I think this gives a nicer output, specially if there are multiple dimensions. With yet another test matrix the test output looks like 
```
--- PASS: TestElasticsearchStorage/tagsAsFeilds=true/FindTraces/Multi-spot_Tags_+_Duration_range (0.01s)
```

plugin/storage/es/spanstore/reader.go
pavolloffay: For example here https://github.com/jaegertracing/jaeger/blob/master/cmd/ingester/app/consumer/consumer.go#L54. Search on params returns more examples

plugin/storage/es/spanstore/writer_test.go
pavolloffay: https://github.com/jaegertracing/jaeger/blob/master/cmd/ingester/app/consumer/consumer.go#L54. Search on params

pavolloffay:
@Monnoroch @golonzovsky could you please review/run it?

yurishkuro:
@pavolloffay as I understand it, this is backwards compatible?

Regarding the CLI flags, proposal for minor adjustment:
```
      --es.tags-as-fields.always                                      Store all span and process tags as object fields. If true .tag-file-path is ignored.
      --es.tags-as-fields.config-file string                          Optional path to a file containing tag keys which will be stored as object fields. Each key should be on a separate line.
      --es.tags-as-fields.dot-replacement string                      The character used to replace dots (".") in tag keys stored as object fields. (default "@")
```

This way if people are using config files instead of CLI then all the new options are in the same section.

@black-adder can you review this?

pavolloffay:
It is backward compatible. The proposed flags look fine except I would remove `.always` from the first one. Instead I wanted to add `all` but even without that it makes sense.

plugin/storage/es/spanstore/dbmodel/model.go
```diff
@@ -0,0 +1,93 @@
...
+	// Span adds a StartTimeMillis field to the standard JSON span.
+	// ElasticSearch does not support a UNIX Epoch timestamp in microseconds,
+	// so Jaeger maps StartTime to a 'long' type. This extra StartTimeMillis field
+	// works around this issue, enabling timerange queries.
+	StartTimeMillis uint64                 `json:"startTimeMillis"`
+	Duration        uint64                 `json:"duration"` // microseconds
+	Tags            []KeyValue             `json:"tags"`
+	Tag             map[string]interface{} `json:"tag,omitempty"`
+	Logs            []Log                  `json:"logs"`
+	Process         *Process               `json:"process,omitempty"`
```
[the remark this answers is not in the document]
pavolloffay: good catch, we can make it mandatory it was just copied from json package.

plugin/storage/es/spanstore/dbmodel/model.go
```diff
@@ -0,0 +1,93 @@
...
+	References    []Reference `json:"references"`
+	StartTime     uint64      `json:"startTime"` // microseconds since Unix epoch
+	// Span adds a StartTimeMillis field to the standard JSON span.
+	// ElasticSearch does not support a UNIX Epoch timestamp in microseconds,
+	// so Jaeger maps StartTime to a 'long' type. This extra StartTimeMillis field
+	// works around this issue, enabling timerange queries.
+	StartTimeMillis uint64                 `json:"startTimeMillis"`
+	Duration        uint64                 `json:"duration"` // microseconds
+	Tags            []KeyValue             `json:"tags"`
+	Tag             map[string]interface{} `json:"tag,omitempty"`
```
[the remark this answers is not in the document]
pavolloffay: This affects how the json is stored to ES. I would like to keep `tags` and `tag` close to each other.

plugin/storage/es/spanstore/dbmodel/from_domain.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/jaegertracing/jaeger/model"
+)
+
+// NewFromDomain creates FromDomain used to convert model span to db span
+func NewFromDomain(allTagsAsObject bool, objectTags map[string]bool, deDotChar string) FromDomain {
+	return FromDomain{allTagsAsFields: allTagsAsObject, tagKeysAsFields: objectTags, deDotChar: deDotChar}
+}
+
+// FromDomain is used to convert model span to db span
+type FromDomain struct {
```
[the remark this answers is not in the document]
pavolloffay: We can but the signature would change to `FromDomain(span, allTagsAsObject, objectTags, dotReplacement)`

in reader/writer struct I like to keep only a configured interface/struct for conversion.

I can change `FromDomain` to an interface like I had it before and make the struct private. @jpkrohling found the interface redundant.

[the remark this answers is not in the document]
yurishkuro: Fair enough, if you pre-instantiate it, then keep the struct. Yes, the interface would be an anti-pattern in Go.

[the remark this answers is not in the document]
pavolloffay: Just for curiosity why not to expose only interface and builder func?

pavolloffay:
can somebody re-review/approve?

kacper-jackiewicz:
@pavolloffay 
I am just testing Jaeger with your new images. Everything seems to work fine on Kibana side(searches/visualizations). Just one remark on Jaeger-Query side. Searching for tags works only if we use new format so for ex. **http@method="GET"**, but the tags are shown with old dotted format, which might be misleading for users.

pavolloffay:
@kacper-jackiewicz thanks for running it. It's a bug query side should use the original key format.

pavolloffay:
Can somebody re-review so it can be merged? 

cc) @jpkrohling @yurishkuro @black-adder

yurishkuro:
> I have published docker images in https://hub.docker.com/u/pavolloffay/ with tag PR1018. 

@pavolloffay can you ask someone to test with those images? Ideally existing users, as well as those who asked for the new functionality

masteinhauser:
> > I have published docker images in https://hub.docker.com/u/pavolloffay/ with tag PR1018.
> 
> @pavolloffay can you ask someone to test with those images? Ideally existing users, as well as those who asked for the new functionality

Hi, how can I help? We have been using the new Elasticsearch backend for ~100 days now with a fairly significant amount of span volume and lots of additional tags being added. This feature is incredibly relevant for us.

pavolloffay:
@masteinhauser or anybody else could you please test this PR? 

Images with this functionality are available in `pavolloffay/` with tag `PR1018` https://hub.docker.com/u/pavolloffay/.

pavolloffay:
@masteinhauser thanks for trying it out! I guess we can merge it and release it.

plugin/storage/es/factory.go
```diff
@@ -74,16 +78,58 @@ func (f *Factory) Initialize(metricsFactory metrics.Factory, logger *zap.Logger)
 // CreateSpanReader implements storage.Factory
 func (f *Factory) CreateSpanReader() (spanstore.Reader, error) {
 	cfg := f.primaryConfig
-	return esSpanStore.NewSpanReader(f.primaryClient, f.logger, cfg.GetMaxSpanAge(), f.metricsFactory, cfg.GetIndexPrefix()), nil
+	return esSpanStore.NewSpanReader(esSpanStore.SpanReaderParams{
+		Client:            f.primaryClient,
+		Logger:            f.logger,
+		MetricsFactory:    f.metricsFactory,
+		MaxLookback:       cfg.GetMaxSpanAge(),
```
yurishkuro: @pavolloffay why did we rename it? It's confusing that config params do not match the storage params.

pavolloffay: I don't think we renamed anything in this PR

The flag name is `--es.max-span-age` and configuration `cfg.GetMaxSpanAge()`. `MaxLookback` is only used in reader maybe also writer.

yurishkuro: I think we should change everything to maxSpanAge

pavolloffay: I can change it, there are also other refactoring opportunities

Issue #1022, opened 2018-08-27
Kafka storage doesn't report any error


ledor473:
## Requirement - what kind of business use case are you trying to solve?
Operate Jaeger Collector that reports to Elasticsearch and Kafka.

## Problem - what in Jaeger blocks you from solving the requirement?
When there's an error with Kafka (missing topic in my case), there are no errors in the logs (even when the log level is set to debug) to help troubleshoot it.

The only sign of a problem is that the following metrics`jaeger_kafka_spans_written{status="failure"}` increase over time, but without much information.
The problem seems to come from the fact that such error are only returned when the messages are flushed to Kafka but it's done asynchronously.

## Proposal - what do you suggest to solve the problem or improve the existing situation?
There's already logging in place when an error is returned by `func (w *SpanWriter) WriteSpan(span *model.Span) error`, but it's not working for Kafka (the error is reported _later_).
I think we should log the errors from the Sarama error channel when we increase the failure metric.

PR #1023, opened 2018-08-27
Log every error from Sarama error channel

ledor473:
This should resolves https://github.com/jaegertracing/jaeger/issues/1022

Here's an example of the logging in action:
```json
{"level":"error","ts":1535383169.5811155,"caller":"kafka/writer.go:59","msg":"kafka server: Request was for a topic or partition that does not exist on this broker.","stacktrace":"github.com/jaegertracing/jaeger/plugin/storage/kafka.NewSpanWriter.func2\n\t/go/src/github.com/jaegertracing/jaeger/plugin/storage/kafka/writer.go:59"}
{"level":"error","ts":1535383170.4639068,"caller":"kafka/writer.go:59","msg":"kafka server: Request was for a topic or partition that does not exist on this broker.","stacktrace":"github.com/jaegertracing/jaeger/plugin/storage/kafka.NewSpanWriter.func2\n\t/go/src/github.com/jaegertracing/jaeger/plugin/storage/kafka/writer.go:59"}
{"level":"error","ts":1535383171.324256,"caller":"kafka/writer.go:59","msg":"kafka server: Request was for a topic or partition that does not exist on this broker.","stacktrace":"github.com/jaegertracing/jaeger/plugin/storage/kafka.NewSpanWriter.func2\n\t/go/src/github.com/jaegertracing/jaeger/plugin/storage/kafka/writer.go:59"}
{"level":"error","ts":1535383171.3243344,"caller":"kafka/writer.go:59","msg":"circuit breaker is open","stacktrace":"github.com/jaegertracing/jaeger/plugin/storage/kafka.NewSpanWriter.func2\n\t/go/src/github.com/jaegertracing/jaeger/plugin/storage/kafka/writer.go:59"}
```


plugin/storage/kafka/writer.go
```diff
@@ -47,7 +55,8 @@ func NewSpanWriter(producer sarama.AsyncProducer, marshaller Marshaller, topic s
 		}
 	}()
 	go func() {
-		for range producer.Errors() {
+		for e := range producer.Errors() {
+			logger.Error(e.Err.Error())
```
jpkrohling: Does it make sense to log only once? In your specific example, if the topic is missing, how many times will it be logged?

yurishkuro: Zap actually supports throttling of logs, so rather than building app-specific logic we could expose full Zap configuration.

ledor473: It seems to log the error I've posted in the description for each submitted span.
Not sure about he throttling but I do see this pretty often:
```json
{"level":"error","ts":1535385948.863829,"caller":"kafka/writer.go:59","msg":"circuit breaker is open","stacktrace":"github.com/jaegertracing/jaeger/plugin/storage/kafka.NewSpanWriter.func2\n\t/go/src/github.com/jaegertracing/jaeger/plugin/storage/kafka/writer.go:59"}
```

PR #1024, opened 2018-08-27
Add an empty statik package

vprithvi:

Fixes `dep` error when using `jaegertracing/jaeger` as a dependency:

```
"github.com/jaegertracing/jaeger/cmd/query/app" transitively (through 1 packages) imports "github.com/jaegertracing/jaeger/cmd/query/app/ui/actual/statik", which contains malformed code: no package exists at "github.com/jaegertracing/jaeger/cmd/query/app/ui/actual/statik"
```

- Add an empty statik package with an empty `statik.go` file


cmd/query/app/ui/actual/statik/statik.go
```diff
@@ -0,0 +1,2 @@
+// Package statik is reserved for code generated by statik
```
yurishkuro: better to name the file `doc.go`

vprithvi: Will do!

PR #1029, opened 2018-08-29
Fix crossdock and Travis issues

### LINKED ISSUE #1028: Crossdock and Travis are buggy

~~Ever since #815, the crossdock tests have been broken.~~ This should have been caught immediately, but an error-handling issues in `.travis.yml` allowed the failed test to be ignored.

The existing crossdock config and Travis CI setup have multiple issues:

- DNS config is not set up using `dns_search: .`, which sporadically causes errors when the a service never initializes, when in fact it has initialized, but the DNS config fails.
- ~~Command format (from #815) does not use the executable as the first argument, which is invalid.~~
- Using locally tagged images without specifying `:latest` causes build failures associated with outdated versions of the image.
- When using `if`/`then` in `.travis.yml`, use `test || exit 1` to guarantee test succeeds. Otherwise, the `then` command succeeds regardless of exit code from `test`.

Fix the above issues.

isaachier:

- Fixes #1028.

- Fixes problems listed in #1028.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -262,10 +262,6 @@ build-crossdock-ui-placeholder:
 build-crossdock: build-crossdock-ui-placeholder docker-no-ui
 	make crossdock
 
-.PHONY: build-crossdock-fresh
-build-crossdock-fresh: build-crossdock-linux
-	make crossdock-fresh
-
 .PHONY: install-tools
 install-tools:
 	go get github.com/wadey/gocovmerge
--- a/cmd/collector/Dockerfile
+++ b/cmd/collector/Dockerfile
@@ -7,4 +7,4 @@ COPY --from=certs /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certifica
 
 EXPOSE 14267
 COPY collector-linux /go/bin/
-ENTRYPOINT ["/go/bin/collector-linux"]
+ENTRYPOINT ["/go/bin/collector-linux", "--cassandra.keyspace=jaeger_v1_dc1", "--cassandra.servers=cassandra", "--collector.zipkin.http-port=9411"]
--- a/cmd/collector/Dockerfile
+++ b/cmd/collector/Dockerfile
@@ -7,4 +7,4 @@ COPY --from=certs /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certifica
 
 EXPOSE 14267
 COPY collector-linux /go/bin/
-ENTRYPOINT ["/go/bin/collector-linux", "--cassandra.keyspace=jaeger_v1_dc1", "--cassandra.servers=cassandra", "--collector.zipkin.http-port=9411"]
+ENTRYPOINT ["/go/bin/collector-linux"]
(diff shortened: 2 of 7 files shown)


cmd/collector/Dockerfile
```diff
@@ -7,4 +7,4 @@ COPY --from=certs /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certifica
 
 EXPOSE 14267
 COPY collector-linux /go/bin/
-ENTRYPOINT ["/go/bin/collector-linux"]
+ENTRYPOINT ["/go/bin/collector-linux", "--cassandra.keyspace=jaeger_v1_dc1", "--cassandra.servers=cassandra", "--collector.zipkin.http-port=9411"]
```
black-adder: We might need another solution. I pulled this branch and tried to build and run the collector container with non-cassandra memory store:
```
wjang:jaeger won$ docker run -e SPAN_STORAGE_TYPE=memory jaegertracing/jaeger-collector:latest
unknown flag: --cassandra.keyspace
Error: unknown flag: --cassandra.keyspace
```

crossdock/docker-compose.yml
```diff
@@ -3,6 +3,7 @@ version: '2'
 services:
     crossdock:
         image: crossdock/crossdock
+        dns_search: .  # Ensures unified DNS config.
```
vprithvi: Could you elaborate on how this ensures unified DNS config?
Isn't  `.` an invalid IP Address or host name?

Makefile
```diff
@@ -262,10 +262,6 @@ build-crossdock-ui-placeholder:
 build-crossdock: build-crossdock-ui-placeholder docker-no-ui
 	make crossdock
 
-.PHONY: build-crossdock-fresh
```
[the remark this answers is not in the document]
isaachier: If it was, it wasn't updated when `crossdock-fresh` was deleted from `crossdock/rules.mk`. That might be my own fault.

crossdock/docker-compose.yml
isaachier: I wish I understood this. I can try to get some more info and will post here.

cmd/collector/Dockerfile
isaachier: OK I see the problem. The solution is probably using the original `command` override but adding the program as the first argument.

crossdock/docker-compose.yml
isaachier: OK have a better solution and explained in comment.

vprithvi: What is the problem being solved?

isaachier: Well using `8.8.8.8` showed me that using a real DNS search host can mess up the test. The docker-compose hostnames won't be resolved correctly. The motive for using `dns_search: .` is described [here](https://github.com/crossdock/crossdock/pull/64). I will add a link in the comment.

isaachier:
I'm gonna close this. Looks like I was dead wrong about the `if`/`then` thing too. Just my MacBook being buggy. If anyone else experiences the DNS issue, I can fix that then.

PR #1032, opened 2018-08-30
Cache offset commits

vprithvi:
Only mark offsets if there is a change in offsets

### DIFF
--- a/cmd/ingester/app/consumer/offset/manager.go
+++ b/cmd/ingester/app/consumer/offset/manager.go
@@ -41,6 +41,7 @@ type Manager struct {
 	markOffsetFunction  MarkOffset
 	offsetCommitCount   metrics.Counter
 	lastCommittedOffset metrics.Gauge
+	minOffset           int64
 	list                *ConcurrentList
 	close               chan struct{}
 	isClosed            sync.WaitGroup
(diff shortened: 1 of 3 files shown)


cmd/ingester/app/consumer/offset/manager.go
```diff
@@ -69,13 +71,17 @@ func (m *Manager) MarkOffset(offset int64) {
...
 	go func() {
+		lastCommittedOffset := m.minOffset
 		for {
 			select {
 			case <-time.After(resetInterval):
 				offset := m.list.setToHighestContiguous()
-				m.offsetCommitCount.Inc(1)
-				m.lastCommittedOffset.Update(offset)
-				m.markOffsetFunction(offset)
+				if lastCommittedOffset != offset {
```
black-adder: is this so that we can create an alert on this if the counter isn't increased for a certain period?

yurishkuro: would it make sense to check `lastCommittedOffset > offset` and do something?

vprithvi: I'm not sure that I understand - The `lastCommitedOffset` would always be less than or equal to `offset` because offset is monotonically increasing, and we only modify `lastCommitedOffset` in L83

yurishkuro: I still think it improves the readability of the code if it says `if lastOffset < newOffset { do something }` - the intent is very clear. But when it says `!=`, it raises the question "so you don't care in which direction it changed?" Now you have to explain "oh but it couldn't have gone the other way because ..." - but if it **did** go the other way (due to a bug), would you still want to commit a smaller value of newOffset?

Issue #1031, opened 2018-08-29
Publish binaries for Kafka ingesters


yurishkuro:
Docker and raw executables

PR #1033, opened 2018-08-30
Support building just the ingester

zoidbergwill:
Wanting to build just the ingester

-  Add `Makefile` job to just build the ingester
- Ignore the ingester binaries in the `.gitignore`
- Whitespace that my editor wanted at the end of the `Makefile` ( I can revert it)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,8 @@ cmd/agent/agent
 cmd/agent/agent-linux
 cmd/collector/collector
 cmd/collector/collector-linux
+cmd/ingester/ingester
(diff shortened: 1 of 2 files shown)


vprithvi:
Thanks for your contribution! Could you sign off the DCO as detailed [here](https://probot.github.io/apps/dco/). 

This looks good to me. While you are at it, could you add the ingester target to the `build-platform-binaries` targets as well?

pavolloffay:
thanks for the PR @zoidbergwill. It seems that this is already done in master and #1031 is resolved.

Just doing before winter cleaning.

Issue #1026, opened 2018-08-29
agent: tchannel timeout should be configurable


gouthamve:
## Requirement - what kind of business use case are you trying to solve?

We run a central jaeger collector and all our different clusters send spans over stunnel to that central collector. Now this means sometimes there are latency blips and it takes more than a second. 

## Problem - what in Jaeger blocks you from solving the requirement?

Currently the timeout is hardcoded to `1s` here: https://github.com/jaegertracing/jaeger/blob/master/cmd/agent/app/reporter/tchannel/reporter.go#L123

## Proposal - what do you suggest to solve the problem or improve the existing situation?

It should ideally be configurable via a flag.

gouthamve:
Umm, @black-adder I found this: https://github.com/jaegertracing/jaeger/issues/927 after opening the issue. I think this is a sub issue in that.

As mentioned there, I'm not sure this is trivial to tackle as it involves some refactoring and without clear instructions and context, new comers don't know what to refactor :) Having said that, if you're open to making just this timeout a flag, I'll take a stab this weekend if no-one else does by then.

black-adder:
Good find; on paper looked so innocently simple.

gouthamve:
Closed in #1034

PR #1034, opened 2018-08-30
[agent] Make timeout for reporting configurable

gouthamve:
So, this is less a fix for #1026 but more to kick off the discussion around: https://github.com/jaegertracing/jaeger/issues/927

While the original proposal has:
```
reporters:
    - kind: tchannel
      # tchannel-specific parameters, e.g.
      collector:
        hostPort: ...
      discovery:
        minPeers: 10
    - kind: grps
      # grpc-specific parameters
```

I kind of like:
```
tchannel:
# tchannel-config

grpc:
# grpc-config
```

I took a stab at implementing my version and turns out it's super straight forward and is very little changed code. But if the preference is to have the originally proposed format, I'll make the changes. But now I'm curious, can we 2 different `tchannel` reporters?

Note: This will be a breaking change.

### DIFF
--- a/cmd/agent/app/reporter/tchannel/flags.go
+++ b/cmd/agent/app/reporter/tchannel/flags.go
@@ -30,6 +30,7 @@ const (
 	hostPort                  = "host-port"
 	discoveryMinPeers         = "discovery.min-peers"
 	discoveryConnCheckTimeout = "discovery.conn-check-timeout"
+	reportTimeout             = "report-timeout"
 )
 
 // AddFlags adds flags for Builder.
@@ -46,6 +47,10 @@ func AddFlags(flags *flag.FlagSet) {
 		tchannelPrefix+discoveryConnCheckTimeout,
 		defaultConnCheckTimeout,
 		"sets the timeout used when establishing new connections")
+	flags.Duration(
+		tchannelPrefix+reportTimeout,
+		time.Second,
+		"sets the timeout used when reporting spans")
 	// TODO remove deprecated in 1.9
 	flags.String(
 		collectorHostPort,
@@ -81,5 +86,6 @@ func (b *Builder) InitFromViper(v *viper.Viper, logger *zap.Logger) *Builder {
 	}
 	b.DiscoveryMinPeers = v.GetInt(tchannelPrefix + discoveryMinPeers)
 	b.ConnCheckTimeout = v.GetDuration(tchannelPrefix + discoveryConnCheckTimeout)
+	b.ReportTimeout = v.GetDuration(tchannelPrefix + reportTimeout)
 	return b
 }
--- a/cmd/agent/app/reporter/tchannel/flags.go
+++ b/cmd/agent/app/reporter/tchannel/flags.go
@@ -84,8 +84,13 @@ func (b *Builder) InitFromViper(v *viper.Viper, logger *zap.Logger) *Builder {
 	if len(v.GetString(tchannelPrefix+hostPort)) > 0 {
 		b.CollectorHostPorts = strings.Split(v.GetString(tchannelPrefix+hostPort), ",")
 	}
-	b.DiscoveryMinPeers = v.GetInt(tchannelPrefix + discoveryMinPeers)
-	b.ConnCheckTimeout = v.GetDuration(tchannelPrefix + discoveryConnCheckTimeout)
+
+	if value := v.GetInt(tchannelPrefix + discoveryMinPeers); value != defaultMinPeers {
+		b.DiscoveryMinPeers = value
+	}
+	if value := v.GetDuration(tchannelPrefix + discoveryConnCheckTimeout); value != defaultConnCheckTimeout {
+		b.ConnCheckTimeout = value
+	}
 	b.ReportTimeout = v.GetDuration(tchannelPrefix + reportTimeout)
 	return b
 }
--- a/cmd/agent/app/processors/thrift_processor_test.go
+++ b/cmd/agent/app/processors/thrift_processor_test.go
@@ -77,7 +77,7 @@ func createProcessor(t *testing.T, mFactory metrics.Factory, tFactory thrift.TPr
 
 func initCollectorAndReporter(t *testing.T) (*metrics.LocalFactory, *testutils.MockTCollector, reporter.Reporter) {
 	metricsFactory, collector := testutils.InitMockCollector(t)
-	reporter := reporter.WrapWithMetrics(tchreporter.New("jaeger-collector", collector.Channel, nil, zap.NewNop()), metricsFactory)
+	reporter := reporter.WrapWithMetrics(tchreporter.New("jaeger-collector", collector.Channel, time.Second, nil, zap.NewNop()), metricsFactory)
 	return metricsFactory, collector, reporter
 }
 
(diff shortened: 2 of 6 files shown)


yurishkuro:
Thanks for looking into this one.

> can we 2 different tchannel reporters

We have some deployments at Uber where we're using two reporters of the same type (well, only tchannel reporter is supported anyway today) to send the data to two locations. However, it is done using the `otherReporters` feature: https://github.com/jaegertracing/jaeger/blob/0c866820b7127c191749796dd976f90f22ff9db3/cmd/agent/app/builder.go#L103-L106

I believe we have a custom config file format that we parse in a custom main. We debated that we could achieve that functionality by forwarding spans to a dedicated splitter, but that seems like more complicated deployment vs. a small code change in the agent. Ideally, we would like to preserve that forking functionality in the agent.

Have you looked into going with the list approach as proposed?  Another alternative is to use nested union-like config, instead of `kind: tchannel` qualifier, e.g.

```
  reporters:
    - collectorHostPort: ...
      tchannel:
        # tchannel-specific parameters
        discovery:
          minPeers: 10
    - collectorHostPort: ...
      grpc:
        # grpc-specific parameters
```

gouthamve:
I think this `nested union-like config` approach better :) I'll make the changes.

yurishkuro:
The tricky part is to decide how cli flags will support a list. If that proves to be too ugly, we could draw a line and say we only support one reporter by default. At Uber we can continue using the custom config and main for multiple reporters.

gouthamve:
Yeah, spent an hour+ trying to make the flags and config match up and not fail tests, couldn't get it to work.

Prefer if there is only reporter being supported.

yurishkuro:
let's go with just one reporter, as long as programmatically we can add more (as we can today). Could you post how the CLI / config will look ?

gouthamve:
```
➜  jaeger git:(add-agent-timeout) ✗ ./agent --help
Jaeger agent is a daemon program that runs on every host and receives tracing data submitted by Jaeger client libraries.

.....

Flags:
      --collector.host-port string                            comma-separated string representing host:ports of a static list of collectors to connect to directly (e.g. when not using service discovery)
      --collector.timeout duration                            sets the timeout used when establishing reporting spans (default 1s)
      --config-file string                                    Configuration file in JSON, TOML, YAML, HCL, or Java properties formats (default none). See spf13/viper for precedence.
      --discovery.conn-check-timeout duration                 sets the timeout used when establishing new connections (default 250ms)
      --discovery.min-peers int                               if using service discovery, the min number of connections to maintain to the backend (default 3)
  -h, --help                                                  help for jaeger-agent
      --http-server.host-port string                          host:port of the http server (e.g. for /sampling point and /baggage endpoint) (default ":5778")
.....
````

So, now this just adds a simple `--collector.timeout` flag. But I can make the flags to be: `--collector.tchannel.*` if you want.

This is how the config will look like: https://github.com/jaegertracing/jaeger/blob/66ed190edfb16710bfbe4a54f6f9dd79f60f2268/cmd/agent/app/builder_test.go#L55-L61

**Comment by pavolloffay** [comment]
The next week I will have time to look at #927

**Comment by gouthamve** [comment]
I'm running this in prod for the past 10 days and everything works great, we used to drop a ton of batches due to timeouts, but right now those occurrences are super rare when running with `5s` timeout.

**Comment by pavolloffay** [comment]
@gouthamve sorry for a longer delay. If i understand it correctly tchannel flags are prefixed with `tchannel`. 

If we cannot make it work like it's in https://github.com/jaegertracing/jaeger/pull/1034#issuecomment-417985429. I think this could go in as there are no other options. cc @yurishkuro

**Comment by pavolloffay** [comment]
@gouthamve https://github.com/jaegertracing/jaeger/pull/1092 addresses agent refactoring. Once that is in this will have to be rebased.

**Comment by pavolloffay** [comment]
@gouthamve could you please rebase this on the current master? We have refactored agent flags now it should be easy to add new ones.

**Comment by gouthamve** [comment]
@pavolloffay Rebased :) I'll run this internally and report on how it performs.

**Comment by gouthamve** [comment]
https://github.com/jaegertracing/jaeger/blob/master/cmd/agent/app/reporter/tchannel/flags.go#L82-L83

would always overwrite this: 
https://github.com/jaegertracing/jaeger/blob/master/cmd/agent/app/reporter/tchannel/flags.go#L70-L77
which means people using the deprecated flags are essentially getting the wrong values.

Fix is in the last commit. It wasn't caught in the last commit as flag values from the previous iteration are carried forward to the next iteration.

**Comment by gouthamve** [comment]
Hmm, the test failure looks like a flake.

**Comment by pavolloffay** [comment]
@gouthamve are you also fixing https://github.com/jaegertracing/jaeger/issues/1192 in this PR?

**Comment by gouthamve** [comment]
Yes, I need to re-init the flags for my tests to pass

File: cmd/agent/app/reporter/tchannel/flags.go [hunk 1]
```diff
@@ -30,6 +30,8 @@ const (
 	hostPort                  = "host-port"
 	discoveryMinPeers         = "discovery.min-peers"
 	discoveryConnCheckTimeout = "discovery.conn-check-timeout"
+
```
yurishkuro: nit: remove blank line

cmd/agent/app/reporter/tchannel/flags.go
```diff
@@ -79,7 +85,13 @@ func (b *Builder) InitFromViper(v *viper.Viper, logger *zap.Logger) *Builder {
 	if len(v.GetString(tchannelPrefix+hostPort)) > 0 {
 		b.CollectorHostPorts = strings.Split(v.GetString(tchannelPrefix+hostPort), ",")
 	}
-	b.DiscoveryMinPeers = v.GetInt(tchannelPrefix + discoveryMinPeers)
-	b.ConnCheckTimeout = v.GetDuration(tchannelPrefix + discoveryConnCheckTimeout)
+
+	if v.GetInt(tchannelPrefix+discoveryMinPeers) != defaultMinPeers {
+		b.DiscoveryMinPeers = v.GetInt(tchannelPrefix + discoveryMinPeers)
+	}
```
yurishkuro:
```
	if value := v.GetInt(tchannelPrefix+discoveryMinPeers); value != defaultMinPeers {
		b.DiscoveryMinPeers = value
	}
```

cmd/agent/app/reporter/tchannel/flags.go
```diff
@@ -79,7 +85,13 @@ func (b *Builder) InitFromViper(v *viper.Viper, logger *zap.Logger) *Builder {
...
+	if v.GetDuration(tchannelPrefix+discoveryConnCheckTimeout) != defaultConnCheckTimeout {
+		b.ConnCheckTimeout = v.GetDuration(tchannelPrefix + discoveryConnCheckTimeout)
+	}
```
yurishkuro: Strictly speaking this does not solve the overriding problem if the manually provided value was the same as the default value, but it's still an improvement. The key thing is that we just need to remove the legacy flags soon.

cmd/agent/app/reporter/tchannel/flags.go
```diff
@@ -30,6 +30,8 @@ const (
 	hostPort                  = "host-port"
```
yurishkuro: > `collectorHostPort         = "collector.host-port"`

I think this was renamed to remove "collector" part. Are you missing a re-base?

pavolloffay: `collector` was removed only from the new flags - `reporter.grpc.collector.host-port`

yurishkuro: ok. This still needs a rebase, lots of merge issues.

gouthamve:
Rebased and feedback incorporated

PR #1040, opened 2018-09-02
Fix Go 1.11 incompatibilities

### LINKED ISSUE #1039: Compilation error with Go 1.11

A compilation error occurs when running `make` on the main project using Go 1.11.

The forwarding of vararg arguments to `printf` wrapper functions without explicit conversion to a slice. As the release notes of Go 1.11 point out:

> Additionally, go vet has become more robust when format-checking printf wrappers.

~~Use explicit casting to convert varargs to a slice for use in `printf` wrappers.~~
Avoid use of varargs as a slice. Expand varargs when forwarding to another function.

isaachier:

- Fixes #1039.

- Use casting to suppress error.

### DIFF
--- a/pkg/recoveryhandler/zap.go
+++ b/pkg/recoveryhandler/zap.go
@@ -30,7 +30,7 @@ type zapRecoveryWrapper struct {
 // Println logs an error message with the given fields
 func (z zapRecoveryWrapper) Println(fields ...interface{}) {
 	// if you think i'm going to check the type of each of the fields and then logger with fields, you're crazy.
-	z.logger.Error(fmt.Sprintln(fields))
+	z.logger.Error(fmt.Sprintln([]interface{}(fields)))
 }
 
 // NewRecoveryHandler returns an http.Handler that recovers on panics
--- a/pkg/recoveryhandler/zap.go
+++ b/pkg/recoveryhandler/zap.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2017 Uber Technologies, Inc.
+// Copyright (c) 2017-2018 Uber Technologies, Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


plugin/storage/cassandra/samplingstore/storage_test.go
```diff
@@ -420,8 +420,22 @@ func TestStringToProbabilitiesAndQPS(t *testing.T) {
 	probabilities := s.stringToProbabilitiesAndQPS(testStr)
 
 	assert.Len(t, probabilities, 2)
-	assert.Equal(t, map[string]*model.ProbabilityAndQPS{"GET": {0.001, 63.2}, "PUT": {0.002, 0.0}}, probabilities["svc1"])
-	assert.Equal(t, map[string]*model.ProbabilityAndQPS{"GET": {0.5, 34.2}}, probabilities["svc2"])
+	assert.Equal(t, map[string]*model.ProbabilityAndQPS{
```
isaachier: One of the tools for Go 1.11 doesn't like structs without field names.

pkg/recoveryhandler/zap.go
```diff
@@ -30,7 +30,7 @@ type zapRecoveryWrapper struct {
 // Println logs an error message with the given fields
 func (z zapRecoveryWrapper) Println(fields ...interface{}) {
 	// if you think i'm going to check the type of each of the fields and then logger with fields, you're crazy.
-	z.logger.Error(fmt.Sprintln(fields))
+	z.logger.Error(fmt.Sprintln([]interface{}(fields)))
```
yurishkuro: don't we want `fields...` instead here?

isaachier: Well, the existing test case won't work out that way. It expects `Println(1, 2, 3)` to log as `message: [1, 2, 3]`, not `message: 1, 2, 3`. I'm just wondering why one of the `zap` convenience wrappers weren't used here. Might replace.

yurishkuro: looking at gorilla's interface:
https://github.com/gorilla/handlers/blob/7e0847f9db758cdebd26c149d0ae9d5d0b9c98ce/recovery.go#L10

I don't think it implies that if you call `logger(1, 2, 3)` that you would get `[1, 2, 3]` output. Rather, it should be equivalent to calling `println(1, 2, 3)`, which will print `1 2 3`.

yurishkuro: basically, our implementation of the adapter was incorrect

PR #1042, opened 2018-09-04
gRPC post spans

isaachier:
Add `PostSpans` functionality to gRPC branch.

### DIFF
--- a/cmd/collector/app/grpc_handler.go
+++ b/cmd/collector/app/grpc_handler.go
@@ -16,7 +16,6 @@ package app
 
 import (
 	"context"
-	"fmt"
 	"time"
 
 	"go.uber.org/zap"
@@ -31,7 +30,7 @@ type GRPCHandler struct {
 	spanProcessor SpanProcessor
 }
 
-// NewGRPCHandler registers routes for this handler on the given router
+// NewGRPCHandler registers routes for this handler on the given router.
 func NewGRPCHandler(logger *zap.Logger, spanProcessor SpanProcessor) *GRPCHandler {
 	return &GRPCHandler{
 		logger:        logger,
@@ -41,16 +40,24 @@ func NewGRPCHandler(logger *zap.Logger, spanProcessor SpanProcessor) *GRPCHandle
 
 // PostSpans implements gRPC CollectorService.
 func (g *GRPCHandler) PostSpans(ctx context.Context, r *api_v2.PostSpansRequest) (*api_v2.PostSpansResponse, error) {
-	// TODO
-	fmt.Printf("PostSpans(%+v)\n", *r)
-	for _, s := range r.Batch.Spans {
-		println(s.OperationName)
+	oks, err := g.spanProcessor.ProcessSpans(r.GetBatch().Spans, JaegerFormatType)
+	if err != nil {
+		g.logger.Error("cannot process spans", zap.Error(err))
+		return nil, err
 	}
-	return &api_v2.PostSpansResponse{Ok: true}, nil
+	success := true
+	for _, ok := range oks {
+		if !ok {
+			success = false
+			break
+		}
+	}
+	return &api_v2.PostSpansResponse{Ok: success}, nil
 }
 
 // GetTrace gets trace
 func (g *GRPCHandler) GetTrace(ctx context.Context, req *api_v2.GetTraceRequest) (*api_v2.GetTraceResponse, error) {
+	// TODO
 	return &api_v2.GetTraceResponse{
 		Trace: &model.Trace{
 			Spans: []*model.Span{
--- a/cmd/collector/app/grpc_handler_test.go
+++ b/cmd/collector/app/grpc_handler_test.go
@@ -82,7 +82,7 @@ func TestPostSpans(t *testing.T) {
 	r, err := client.PostSpans(ctx, &api_v2.PostSpansRequest{
 		Batch: model.Batch{
 			Spans: []*model.Span{
-				&model.Span{
+				{
 					OperationName: "test-operation",
 				},
 			},
@@ -104,7 +104,7 @@ func TestPostSpansWithError(t *testing.T) {
 	r, err := client.PostSpans(ctx, &api_v2.PostSpansRequest{
 		Batch: model.Batch{
 			Spans: []*model.Span{
-				&model.Span{
+				{
 					OperationName: "fake-operation",
 				},
 			},
--- a/cmd/collector/app/grpc_handler_test.go
+++ b/cmd/collector/app/grpc_handler_test.go
@@ -20,7 +20,6 @@ import (
 	"net"
 	"sync"
 	"testing"
-	"time"
 
 	"github.com/stretchr/testify/require"
 	"go.uber.org/zap"
@@ -77,9 +76,7 @@ func TestPostSpans(t *testing.T) {
 	defer server.Stop()
 	client, conn := newClient(t, addr)
 	defer conn.Close()
-	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
-	defer cancel()
-	r, err := client.PostSpans(ctx, &api_v2.PostSpansRequest{
+	r, err := client.PostSpans(context.Background(), &api_v2.PostSpansRequest{
 		Batch: model.Batch{
 			Spans: []*model.Span{
 				{
@@ -99,9 +96,7 @@ func TestPostSpansWithError(t *testing.T) {
 	defer server.Stop()
 	client, conn := newClient(t, addr)
 	defer conn.Close()
-	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
-	defer cancel()
-	r, err := client.PostSpans(ctx, &api_v2.PostSpansRequest{
+	r, err := client.PostSpans(context.Background(), &api_v2.PostSpansRequest{
 		Batch: model.Batch{
 			Spans: []*model.Span{
 				{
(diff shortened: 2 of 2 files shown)


cmd/collector/app/grpc_handler_test.go
```diff
@@ -0,0 +1,117 @@
...
+	server := grpc.NewServer()
+	processor := &mockSpanProcessor{err: err}
+	logger, err := zap.NewDevelopment()
+	require.NoError(t, err)
+	handler := NewGRPCHandler(logger, processor)
+	api_v2.RegisterCollectorServiceServer(server, handler)
+	lis, err := net.Listen("tcp", "localhost:0")
+	require.NoError(t, err)
+	go func() {
+		err := server.Serve(lis)
```
black-adder: doesn't this introduce a potential race condition where the client is initialized before the server is ready?

cmd/collector/app/grpc_handler_test.go
```diff
@@ -0,0 +1,117 @@
...
+	require.NoError(t, err)
+	return api_v2.NewCollectorServiceClient(conn), conn
+}
+
+func TestPostSpans(t *testing.T) {
+	server, processor, addr := initializeGRPCTestServer(t, nil)
+	defer server.Stop()
+	client, conn := newClient(t, addr)
+	defer conn.Close()
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
```
black-adder: can't we just use `context.Background()`?

cmd/collector/app/grpc_handler_test.go
```diff
@@ -0,0 +1,117 @@
...
+type mockSpanProcessor struct {
+	err   error
+	mux   sync.Mutex
+	spans []*model.Span
+}
+
+func (p *mockSpanProcessor) ProcessSpans(spans []*model.Span, spanFormat string) ([]bool, error) {
+	p.mux.Lock()
+	defer p.mux.Unlock()
+	p.spans = append(p.spans, spans...)
```
black-adder: is this all necessary? It looks like your tests call `ProcessSpans` just once, the extra complexity (mutex, keeping track of spans, etc) doesn't look required.

cmd/collector/app/grpc_handler_test.go
```diff
@@ -0,0 +1,117 @@
...
+			Spans: []*model.Span{
+				{
+					OperationName: "test-operation",
+				},
+			},
+		},
+	})
+	require.NoError(t, err)
+	require.False(t, r.GetOk())
+	require.Len(t, processor.getSpans(), 1)
```
black-adder: so when you do this, you're essentially testing a function on the mock, kinda weird.

cmd/collector/app/grpc_handler_test.go
```diff
@@ -0,0 +1,117 @@
...
+	return oks, p.err
+}
+
+func (p *mockSpanProcessor) getSpans() []*model.Span {
+	p.mux.Lock()
+	defer p.mux.Unlock()
+	return p.spans
+}
+
+func initializeGRPCTestServer(t *testing.T, err error) (*grpc.Server, *mockSpanProcessor, net.Addr) {
```
yurishkuro: Nit: expectedError

cmd/collector/app/grpc_handler.go
```diff
@@ -41,16 +40,24 @@ func NewGRPCHandler(logger *zap.Logger, spanProcessor SpanProcessor) *GRPCHandle
...
+	}
+	success := true
+	for _, ok := range oks {
+		if !ok {
+			success = false
+			break
+		}
 	}
-	return &api_v2.PostSpansResponse{Ok: true}, nil
+	return &api_v2.PostSpansResponse{Ok: success}, nil
```
yurishkuro: We may want to rethink this API. What are we trying to communicate back to the caller with this Boolean flag? At least (accepted, rejected) counts would make more sense. When can the processor return not ok? Should it be returning specific errors instead?

cmd/collector/app/grpc_handler_test.go
yurishkuro: I think it's needed since read and write are in different goroutines

cmd/collector/app/grpc_handler_test.go
```diff
@@ -0,0 +1,117 @@
...
+		},
+	})
+	require.NoError(t, err)
+	require.False(t, r.GetOk())
+	require.Len(t, processor.getSpans(), 1)
+}
+
+func TestPostSpansWithError(t *testing.T) {
+	processorErr := errors.New("test-error")
+	server, processor, addr := initializeGRPCTestServer(t, processorErr)
```
yurishkuro: We should simplify this. You can return a test server struct with fields instead of multiple values. And we can rewrite this in FP style withServer(params, func) to hide all the client/server complexity from the test.

cmd/collector/app/grpc_handler_test.go
yurishkuro: Maybe timeout helps with the race when client starts before server. Although, the conn car above, is it already open?

cmd/collector/app/grpc_handler_test.go
yurishkuro: It tests sending spans over the wire, no?

cmd/collector/app/grpc_handler_test.go
isaachier: Good point.

cmd/collector/app/grpc_handler_test.go
isaachier: Fair enough. I think this originally helped me when I had an issue with client hanging forever. Now that it's fixed, can remove this.

cmd/collector/app/grpc_handler_test.go
isaachier: Yes. Luckily for me, the only code I had to write here was essentially copied from the equivalent Thrift test. It does locking the same way IIRC.

cmd/collector/app/grpc_handler_test.go
isaachier: Ya again, Thrift handler does this.

cmd/collector/app/grpc_handler_test.go
isaachier: Good idea.

cmd/collector/app/grpc_handler.go
isaachier: I agree that the API as it stands isn't exactly perfect. If anything, a boolean vector (or even bit vector) might make more sense here. Count works too if we don't care which spans failed. An optional error message might be better than the boolean response.

cmd/collector/app/grpc_handler_test.go
isaachier: Sure. Again, sort of copied this stuff. Can easily clean this up a bit.

cmd/collector/app/grpc_handler_test.go
isaachier: Actually, I believe that this is somewhat unavoidable. The issue is that `Serve` will block indefinitely. Using a `sync.WaitGroup` would help, if I could call the `Done` method before the the server blocks, but there is no good place to do that.

cmd/collector/app/grpc_handler.go
isaachier: Any thoughts @yurishkuro?

cmd/collector/app/grpc_handler.go
```diff
@@ -41,16 +40,24 @@ func NewGRPCHandler(logger *zap.Logger, spanProcessor SpanProcessor) *GRPCHandle
 
 // PostSpans implements gRPC CollectorService.
 func (g *GRPCHandler) PostSpans(ctx context.Context, r *api_v2.PostSpansRequest) (*api_v2.PostSpansResponse, error) {
-	// TODO
-	fmt.Printf("PostSpans(%+v)\n", *r)
-	for _, s := range r.Batch.Spans {
-		println(s.OperationName)
+	oks, err := g.spanProcessor.ProcessSpans(r.GetBatch().Spans, JaegerFormatType)
```
yurishkuro: I think this is loosing Batch.Process data.

cmd/collector/app/grpc_handler.go
yurishkuro: The question is: what is the sender going to do with the errors?

cmd/collector/app/grpc_handler_test.go
yurishkuro: I mean if spans go over the wire then we're not just testing mock, we're testing the transmission.

cmd/collector/app/grpc_handler.go
isaachier: So the spans have pointers to the process. I'm not sure when that back-reference is set. If it is already set, there should not be a problem.

cmd/collector/app/grpc_handler_test.go
isaachier: Yes exactly.

cmd/collector/app/grpc_handler.go
yurishkuro: if we want to support retries we should be using gRPC error code for that. Can you check what openCensus is doing in this case?

cmd/collector/app/grpc_handler.go
yurishkuro: by convention when a bunch of spans arrive in a Batch and logically they are all from the same Process then Span.Process will be nil, but Batch.Process will be defined. This is done to minimize the amount of data sent over the wire. So you need to loop through spans and copy the Process pointer from Batch if Span.Process == nil

pavolloffay:
done in #1165

PR #1046, opened 2018-09-04
Add gosimple

### LINKED ISSUE #1045: Run gosimple tool to as part of lint task

Enforce simple and readable code by running a simplicity checker (gosimple) as part of the lint job.

gosimple is not used to check for unnecessarily complex code.

Use gosimple in lint task.

isaachier:
- Resolves #1045.

- Add gosimple to lint task. Apply existing suggestions to code.

### DIFF
--- a/cmd/ingester/main.go
+++ b/cmd/ingester/main.go
@@ -92,7 +92,7 @@ func main() {
 
 			hc.Ready()
 			<-signalsChannel
-			logger.Info("Jaeger Ingester is starting to close")
+			logger.Info("Shutting down")
 			err = consumer.Close()
 			if err != nil {
 				logger.Error("Failed to close consumer", zap.Error(err))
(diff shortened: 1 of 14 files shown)


cmd/ingester/main.go
```diff
@@ -91,21 +91,19 @@ func main() {
 			consumer.Start()
 
 			hc.Ready()
-			select {
-			case <-signalsChannel:
-				logger.Info("Jaeger Ingester is starting to close")
-				err := consumer.Close()
+			<-signalsChannel
+			logger.Info("Jaeger Ingester is starting to close")
```
yurishkuro: nit: s/close/shutdown

isaachier: Sure. I can change the message too. Was just focusing on the issues gosimple pointed out.

isaachier: As a follow up, can we get rid of service name here? I don't see how this is more descriptive when the service logs are always kept separate. Even crossdock prefixes service names to logs.

yurishkuro: I thought the same but didn't want to look too pedantic. Yes the name is unnecessary.

PR #1050, opened 2018-09-05
WIP: Add plugins support to the Collector

### LINKED ISSUE #422: How to support plugins

ledor473:
This PR provides the foundation to support plugins in Jaeger using native Go plugins (`pkg/plugin`). Resolves #422 

Supported plugins:
- Storage (`spanstore` + `dependencystore`)
- Span Processor (`Sanitizer`, `PreProcessor`, `PreSave`, `SpanFilter`)

The current PR is working and this comment https://github.com/jaegertracing/jaeger/pull/1050#issuecomment-456126756 includes an example of a storage plugin that seems to work so far.

## What is missing
- Configuration file to order the plugins
- Unit/Integration tests
- More examples
- Documentation
- Support for that in the Dockerfile

## How to run it
1. Compile the plugins
```
cd examples/collector/plugins/
go build  -buildmode=plugin sanitizer_logger.go
go build  -buildmode=plugin pre_processor_logger.go
go build  -buildmode=plugin pre_save_logger.go
go build  -buildmode=plugin span_filter_logger.go
```
2. Run the collector with the proper flag
```
SPAN_STORAGE_TYPE=memory LOG_LEVEL=debug PLUGINS_DIRECTORY=`pwd`/examples/collector/plugins/ go run cmd/collector/main.go
```
3. Send traces and look at the logs, it should display something like:
```
PreProcessSpans...  4 spans
SpanFilter... TraceID=66333d0d4a3452c SpanID=ef9dbde43426f82f OperationName=truncated
SpanFilter... TraceID=66333d0d4a3452c SpanID=ef9dbde43426f82f OperationName=truncated
SpanFilter... TraceID=66333d0d4a3452c SpanID=7c8453653cab6cbf OperationName=truncated
SpanFilter... TraceID=66333d0d4a3452c SpanID=7c8453653cab6cbf OperationName=truncated
SpanFilter... TraceID=66333d0d4a3452c SpanID=8de1edfd89f8e32e OperationName=truncated
SpanFilter... TraceID=66333d0d4a3452c SpanID=8de1edfd89f8e32e OperationName=truncated
SpanFilter... TraceID=66333d0d4a3452c SpanID=66333d0d4a3452c OperationName=truncated
SpanFilter... TraceID=66333d0d4a3452c SpanID=66333d0d4a3452c OperationName=truncated
Sanitizer... TraceID=66333d0d4a3452c SpanID=ef9dbde43426f82f OperationName=truncated
Sanitizer... TraceID=66333d0d4a3452c SpanID=7c8453653cab6cbf OperationName=truncated
Sanitizer... TraceID=66333d0d4a3452c SpanID=66333d0d4a3452c OperationName=truncated
Sanitizer... TraceID=66333d0d4a3452c SpanID=8de1edfd89f8e32e OperationName=truncated
PreSave... TraceID=66333d0d4a3452c SpanID=66333d0d4a3452c OperationName=truncated
PreSave... TraceID=66333d0d4a3452c SpanID=8de1edfd89f8e32e OperationName=truncated
PreSave... TraceID=66333d0d4a3452c SpanID=ef9dbde43426f82f OperationName=truncated
PreSave... TraceID=66333d0d4a3452c SpanID=7c8453653cab6cbf OperationName=truncated
```

## Gotchas
- Mac& Linux only
- If you want to debug on macOS using delve, you need to compile Go from source to get this fix https://github.com/golang/go/commit/9c833831b2bdaa465349194797cf3894cb85f9c4
  - If needed, something like this could be added to allow someone to debug the rest of the codebase


yurishkuro:
Yes afaik

ledor473:
Yes I think the handling of dependencies in a plugin is not fully working as the issues mentioned [here](https://github.com/jaegertracing/jaeger/issues/422#issuecomment-360954600) are still open. 

The main reason I started with the filters/processors is that a storage implementation would likely require additional libraries. With the current bugs, I'm not sure if it's even feasible to implemement a storage plugins without using hashicorp/go-plugin...

That being said, I'm not sure neither if using hashicorp/go-plugin for filters/processors would be a good idea performance wise.

I could work on the following changes:
- Move to a more generic approach
- Plugins access to the CLI
- Plugins access to the logger
- Ordering of the plugins

But if the dependency issue is a deal breaker for filters/processors (as it is for storage), I would start exploring hashicorp/go-plugins instead.

ledor473:
It turns out much harder than expected to do the proposed changes but I feel (even if it's not fully done yet) that it's a much cleaner implementation!

So here's what I added in the last commits:
- Moved the core features to `pkg/plugin` and the rest to `cmd/collector/app/plugin`
- Moved the `--collector.plugins-directory` to the `PLUGINS_DIRECTORY` environment variable because we need to be able to load the plugins **before** using Viper
- Configurable plugins via CLI args:
```
SPAN_STORAGE_TYPE=memory LOG_LEVEL=info PLUGINS_DIRECTORY="/Users/ledor473/go/src/github.com/jaegertracing/jaeger/examples/collector/plugins/" go run cmd/collector/main.go -h

Usage:
...

Available Commands:
...

Flags:
...
      --pre-save-log-interval int        How often should the PreSave Logger should print (default 30)
...
```
- Ability to provide the logger to the plugins:
```
{"level":"info","ts":1537905084.1992269,"caller":"plugins/pre_save_logger.go:58","msg":"PreSave Logger","rate":0,"count":0,"interval":3}
```
- Ability to add metrics (untested)

Things to do:
- Discuss a way to have a logger **before** the application logger is created (something like a "bootstrap logger" that would be discarded afterward?). This is to replace the usage to os.Stderr
- Plugin ordering

yurishkuro:
> Discuss a way to have a logger before the application logger is created

do we need to?  the storage factory facade is already able to pass loggers to the underlying factories.

ledor473:
So the logger problem is mainly because it's getting created here (in `RunE`) https://github.com/jaegertracing/jaeger/pull/1050/files#diff-03e42ee625fab9eaa63d9d9daa235f8eR90
But I would like to use it here: https://github.com/jaegertracing/jaeger/pull/1050/files#diff-03e42ee625fab9eaa63d9d9daa235f8eR202

The usage is mainly in that method: https://github.com/jaegertracing/jaeger/pull/1050/files#diff-e6ce03f3742f3a6c8b3e443d8e1a53daR58
The debug logs indicate plugin detection and attempt to find symbol.
It's also used in case there's a symbol that is found but is not of the proper type.

A similar thing that would benefit from having a bootstrap logger is this method: https://github.com/jaegertracing/jaeger/pull/1050/files#diff-03e42ee625fab9eaa63d9d9daa235f8eR62 that currently use `os.Stderr` directly

yurishkuro:
maybe punt on the logger for now, use io.Writer same as that last example.

ledor473:
So I was able to make it run. The plugin doesn't do much right now so it's 100% proven but I still wanted to update the PR with more information of what the build looks like.

All the commands will be run inside a Docker, here's my exact run command:
```
docker run \
-v $GOPATH/src/github.com/ledor473:/go/src/github.com/ledor473 \
-v $GOPATH/src/github.com/jaegertracing:/go/src/github.com/jaegertracing \
-it golang:alpine sh
```

Some generic installations/variables to make the build work:
```
apk add --update make git gcc build-base
export GOPATH=/go
export GOOS=linux 
```

Build Jaeger as we already do:
```
cd /go/src/github.com/jaegertracing/jaeger/
make install 
make build-collector
```

Now this gets trickier:
```
# Install plugins deps in vendor
cd /go/src/github.com/ledor473/jaeger-storage-dynamodb/
glide install

# Go back in Jaeger repo
cd /go/github.com/jaegertracing/jaeger/

# Copy all the vendors from the plugins into Jaeger's vendor (could be an rsync or something that just copy the difference from one side)
cp -r /go/src/github.com/ledor473/jaeger-storage-dynamodb/vendor/ ./

# We also have to copy all Go files of the plugins, but for now it's only 1 file
cp /go/src/github.com/ledor473/jaeger-storage-dynamodb/storage_factory.go ./jaeger-storage-dynamodb.go

# Compile the plugin into a .so file
go build -buildmode=plugin ./jaeger-storage-dynamodb.go
```

With that done, you can run it as follow:
```
SPAN_STORAGE_TYPE=plugin PLUGINS_DIRECTORY=`pwd` LOG_LEVEL=debug ./cmd/collector/collector-linux -h
{"level":"info","ts":1543955396.7755852,"caller":"pluginloader/factory.go:122","msg":"Collector plugin found","path":"/go/src/github.com/jaegertracing/jaeger/jaeger-storage-dynamodb.so"}
Jaeger collector receives traces from Jaeger agents and runs them through a processing pipeline.

Usage:
  jaeger-collector [flags]
  jaeger-collector [command]

Available Commands:
...
```

yurishkuro:
Have you tried pulling Jaeger as a vendored dependency of the plugin repo? Because all these machinations remind me of the classic transitive vendor problem caused by relying on a dependency from GOPATH that has its own vendor dir. In our internal builds at Ubet we always build against an empty GOPATH, to avoid these issues and have the dep manager fully responsible for deps.

ledor473:
@olivierboucher even though I've worked on this, I think I prefer the gRPC approach anyway.

Seems like the recent results you posted are promising too!

ledor473:
@yurishkuro So this is looking pretty good! 
In order to try it out, I've copy-pasted the unmerged Badger implementation (from @burmanm) and turned it into a plugin: https://github.com/ledor473/jaeger-storage-badger

The only real modification (except from moving to the main package and changing the repo) was to add this file to properly export the plugin: https://github.com/ledor473/jaeger-storage-badger/blob/master/plugin.go

With the same recipe as the last comment, I now get the following help:
```
$ SPAN_STORAGE_TYPE=plugin PLUGINS_DIRECTORY="`pwd`" LOG_LEVEL=debug go run cmd/collector/main.go -h
{"level":"info","ts":1548087122.6742558,"caller":"pluginloader/factory.go:122","msg":"Collector plugin found","path":"/Users/ledor473/go/src/github.com/jaegertracing/jaeger/jaeger-storage-badger.so"}
{"level":"info","ts":1548087122.751847,"caller":"pluginloader/factory.go:73","msg":"Plugin was successfully loaded","plugin":"jaeger-storage-badger.so","symbol":"Configurable"}
Jaeger collector receives traces from Jaeger agents and runs them through a processing pipeline.

....

Flags:
      --badger.consistency                     If all writes should be synced immediately. This will greatly reduce write performance.
      --badger.directory-key string            Path to store the keys (indexes), this directory should reside in SSD disk. Set ephemeral to false if you want to define this setting. (default "/var/folders/vs/znnjv22531n2cn1gs4txh4nc0000gn/T/go-build211534391/b001/exe/data/keys")
      --badger.directory-value string          Path to store the values (spans). Set ephemeral to false if you want to define this setting. (default "/var/folders/vs/znnjv22531n2cn1gs4txh4nc0000gn/T/go-build211534391/b001/exe/data/values")
      --badger.ephemeral                       Mark this storage ephemeral, data is stored in tmpfs. (default true)
      --badger.maintenance-interval duration   How often the maintenance thread for values is ran. Format is time.Duration (https://golang.org/pkg/time/#Duration) (default 5m0s)
      --badger.span-store-ttl duration         How long to store the data. Format is time.Duration (https://golang.org/pkg/time/#Duration) (default 72h0m0s)
      --collector.grpc-port int                (experimental) The gRPC port for the collector service (default 14250)
      --collector.http-port int                The HTTP port for the collector service (default 14268)
      --collector.num-workers int              The number of workers pulling items from the queue (default 50)
      --collector.plugins-directory string     The directory used to dynamically load collector plugins
      --collector.port int                     The TChannel port for the collector service (default 14267)
      --collector.queue-size int               The queue size of the collector (default 2000)
      --collector.zipkin.http-port int         The HTTP port for the Zipkin collector service e.g. 9411
      --config-file string                     Configuration file in JSON, TOML, YAML, HCL, or Java properties formats (default none). See spf13/viper for precedence.
      --health-check-http-port int             The http port for the health check service (default 14269)
  -h, --help                                   help for jaeger-collector
      --log-level string                       Minimal allowed log Level. For more levels see https://github.com/uber-go/zap (default "info")
      --metrics-backend string                 Defines which metrics backend to use for metrics reporting: expvar, prometheus, none (default "prometheus")
      --metrics-http-route string              Defines the route of HTTP endpoint for metrics backends that support scraping (default "/metrics")
      --sampling.strategies-file string        The path for the sampling strategies file in JSON format. See sampling documentation to see format of the file
      --span-storage.type string               Deprecated; please use SPAN_STORAGE_TYPE environment variable. Run this binary with "env" command for help.
```

And it's not just the help that is working:
```
SPAN_STORAGE_TYPE=plugin PLUGINS_DIRECTORY="`pwd`" go run cmd/collector/main.go
{"level":"info","ts":1548087222.572639,"caller":"pluginloader/factory.go:122","msg":"Collector plugin found","path":"/Users/ledor473/go/src/github.com/jaegertracing/jaeger/jaeger-storage-badger.so"}
{"level":"info","ts":1548087222.646497,"caller":"pluginloader/factory.go:73","msg":"Plugin was successfully loaded","plugin":"jaeger-storage-badger.so","symbol":"Configurable"}
{"level":"info","ts":1548087222.647129,"caller":"healthcheck/handler.go:99","msg":"Health Check server started","http-port":14269,"status":"unavailable"}
{"level":"info","ts":1548087222.647357,"caller":"pluginloader/factory.go:73","msg":"Plugin was successfully loaded","plugin":"jaeger-storage-badger.so","symbol":"Configurable"}
{"level":"info","ts":1548087222.647667,"caller":"pluginloader/factory.go:73","msg":"Plugin was successfully loaded","plugin":"jaeger-storage-badger.so","symbol":"StorageFactory"}
2019/01/21 11:13:42 Replaying from value pointer: {Fid:0 Len:0 Offset:0}
2019/01/21 11:13:42 Iterating file id: 0
2019/01/21 11:13:42 Iteration took: 16.367µs
{"level":"info","ts":1548087222.72284,"caller":"jaeger-storage-badger/factory.go:120","msg":"Badger storage configuration","configuration":{"Dir":"/var/folders/vs/znnjv22531n2cn1gs4txh4nc0000gn/T/badger011673015","ValueDir":"/var/folders/vs/znnjv22531n2cn1gs4txh4nc0000gn/T/badger011673015","SyncWrites":false,"TableLoadingMode":1,"ValueLogLoadingMode":2,"NumVersionsToKeep":1,"MaxTableSize":67108864,"LevelSizeMultiplier":10,"MaxLevels":7,"ValueThreshold":32,"NumMemtables":5,"NumLevelZeroTables":5,"NumLevelZeroTablesStall":10,"LevelOneSize":268435456,"ValueLogFileSize":1073741823,"ValueLogMaxEntries":1000000,"NumCompactors":3,"DoNotCompact":false,"ReadOnly":false,"Truncate":false}}
{"level":"info","ts":1548087222.723698,"caller":"static/strategy_store.go:79","msg":"No sampling strategies provided, using defaults"}
{"level":"info","ts":1548087222.723899,"caller":"collector/main.go:157","msg":"Starting jaeger-collector TChannel server","port":14267}
{"level":"info","ts":1548087222.723972,"caller":"grpcserver/grpc_server.go:64","msg":"Starting jaeger-collector gRPC server","grpc-port":"14250"}
{"level":"info","ts":1548087222.724033,"caller":"collector/main.go:171","msg":"Registering metrics handler with HTTP server","route":"/metrics"}
{"level":"info","ts":1548087222.724065,"caller":"collector/main.go:180","msg":"Starting jaeger-collector HTTP server","http-port":14268}
{"level":"info","ts":1548087222.7240808,"caller":"healthcheck/handler.go:133","msg":"Health Check state change","status":"ready"}
```

ledor473:
Closing as gRPC plugin is merged now

PR #1080, opened 2018-09-23
Feature/imports

GNURub:
Following pattern for grouping imports.


yurishkuro:
I am not so sure this is a good change. Some of it is ok, but changing to `()` style for a single import is superfluous, and changing imports in the auto-generated files like mocks / thrift is counterproductive.

Issue #1072, opened 2018-09-19
Report Kafka Offset lag in seconds


vprithvi:
## Requirement - what kind of business use case are you trying to solve?

Currently, offset lag is reported in number of messages which is difficult to contextualize without also looking at number of incoming messages and rate of processing

## Proposal - what do you suggest to solve the problem or improve the existing situation?

Compute offset lag in minutes and emit metric

vprithvi:
@yurishkuro Indeed! I'll have a diff up in a day or so. Sadly, we can't benefit from this internally yet.

PR #1084, opened 2018-09-25
Add metric for offset lag in ns

vprithvi:
- #1072 

- Report offset lags in ns if the Kafka broker format has this information.

### DIFF
--- a/cmd/ingester/app/consumer/consumer.go
+++ b/cmd/ingester/app/consumer/consumer.go
@@ -16,6 +16,7 @@ package consumer
 
 import (
 	"sync"
+	"time"
 
 	"github.com/Shopify/sarama"
 	sc "github.com/bsm/sarama-cluster"
@@ -43,6 +44,7 @@ type Consumer struct {
 	processorFactory ProcessorFactory
 
 	partitionIDToState map[int32]*consumerState
+	timeSince          func(time.Time) time.Duration
 }
 
 type consumerState struct {
@@ -58,6 +60,7 @@ func New(params Params) (*Consumer, error) {
 		internalConsumer:   params.InternalConsumer,
 		processorFactory:   params.ProcessorFactory,
 		partitionIDToState: make(map[int32]*consumerState),
+		timeSince:          time.Since,
 	}, nil
 }
 
@@ -103,8 +106,8 @@ func (c *Consumer) handleMessages(pc sc.PartitionConsumer) {
 		c.logger.Debug("Got msg", zap.Any("msg", msg))
 		msgMetrics.counter.Inc(1)
 		msgMetrics.offsetGauge.Update(msg.Offset)
-		if !msg.BlockTimestamp.IsZero() {
-			msgMetrics.lagGaugeNs.Update(time.Since(msg.BlockTimestamp).Nanoseconds())
+		if !msg.BlockTimestamp.IsZero() { // This timestamp is zero for Kafka broker (or the broker data format) < v0.10.2
+			msgMetrics.lagGaugeNs.Update(c.timeSince(msg.BlockTimestamp).Nanoseconds())
 		}
 		msgMetrics.lagGauge.Update(pc.HighWaterMarkOffset() - msg.Offset - 1)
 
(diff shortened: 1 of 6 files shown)


crossdock/services/tracehandler_test.go
```diff
@@ -259,7 +259,7 @@ func TestCreateTracesLoop(t *testing.T) {
 	defer server.Close()
 
 	handler := &TraceHandler{
-		logger: zap.NewNop(),
+		logger:                   zap.NewNop(),
```
vprithvi: Not sure why make fmt moved these...

black-adder:
Perhaps we could use a timer instead of a gauge? With timers, you can use histograms to visualize the data (mean, p90, p99, max, etc.). With a gauge, m3 or prom would aggregate the gauges over a 1 sec(?) interval and just output the max/average?

vprithvi:
>Perhaps we could use a timer instead of a gauge? With timers, you can use histograms to visualize the data (mean, p90, p99, max, etc.). With a gauge, m3 or prom would aggregate the gauges over a 1 sec(?) interval and just output the max/average?

I'm concerned about the value of having those accurate metrics because offset lag is computed in an approximate way - the block offset from Kafka is not set per message but per message block and is dependent on Kafka broker timestamps. Having timers might give a greater sense of accuracy compared to gauges.

cmd/ingester/app/consumer/consumer.go
```diff
@@ -103,6 +106,9 @@ func (c *Consumer) handleMessages(pc sc.PartitionConsumer) {
 		c.logger.Debug("Got msg", zap.Any("msg", msg))
 		msgMetrics.counter.Inc(1)
 		msgMetrics.offsetGauge.Update(msg.Offset)
+		if !msg.BlockTimestamp.IsZero() { // This timestamp is zero for Kafka broker (or the broker data format) < v0.10.2
+			msgMetrics.lagGaugeNs.Update(c.timeSince(msg.BlockTimestamp).Nanoseconds())
```
yurishkuro: Why gauge? Out metrics API supports timers natively, and probably accept `time.Duration`.

cmd/ingester/app/consumer/consumer_metrics.go
```diff
@@ -36,6 +37,7 @@ func (c *Consumer) newMsgMetrics(partition int32) msgMetrics {
 		counter:     f.Counter("messages", nil),
 		offsetGauge: f.Gauge("current-offset", nil),
 		lagGauge:    f.Gauge("offset-lag", nil),
+		lagGaugeNs:  f.Gauge("offset-lag-ns", nil),
```
yurishkuro: `-ns` is implementation specific. Need a different name. Also, "offset" isn't applicable for this measure. I think simply "lag" will work.

cmd/ingester/app/consumer/consumer_test.go
```diff
@@ -86,14 +86,16 @@ func newConsumer(
 	factory metrics.Factory,
 	topic string,
 	processor processor.SpanProcessor,
-	consumer consumer.Consumer) *Consumer {
+	consumer consumer.Consumer,
+	timeSince func(time.Time) time.Duration) *Consumer {
```
yurishkuro:
```
    timeSince func(time.Time) time.Duration,
) *Consumer {
```

cmd/ingester/app/consumer/consumer.go
vprithvi: @yurishkuro, @black-adder bought this up earlier to which I had replied

> I'm concerned about the value of having those accurate metrics because offset lag is computed in an approximate way - the block offset from Kafka is not set per message but per message block and is dependent on Kafka broker timestamps. Having timers might give a greater sense of accuracy compared to gauges.

cmd/ingester/app/consumer/consumer_metrics.go
vprithvi: I agree with what you say with the name of the metric, `lag` works. I agree that `-ns` is implementation specific, perhaps this is a reason to use a `Timer` because `Timer` metrics have a notion of measuring in units of time, as opposed to gauges which may refer to anything.

cmd/ingester/app/consumer/consumer.go
yurishkuro: Call it batch lag, I think it addresses the ambiguity much better than using guage instead of timer.

pavolloffay:
@vprithvi could you please push the branch to your fork and remove this branch?

Issue #1052, opened 2018-09-05
Remove dependency on sarama-cluster (Kafka driver)


yurishkuro:
* partitions not picked up on rebalance: https://github.com/bsm/sarama-cluster/issues/235
* partitions are can be held even when no work is being done by client: https://github.com/bsm/sarama-cluster/issues/261
* potential deadlock when rebalancing: https://github.com/bsm/sarama-cluster/issues/255#issuecomment-413967695

yurishkuro:
Transferring #1090 by @vprithvi here:

> Use high level consumer groups from Shopify/sarama#1099 instead of sarama-cluster.
> This is a rewrite of high level consumer functionality and can side step the issues described in #1052

marqc:
Instead of sarama You can take a look at https://github.com/confluentinc/confluent-kafka-go
It's a light wrapper for C implementation of kafka client (librdkafka) and AFAIK is more stable/reliable than shopify/sarama.

vprithvi:
@marqc Thanks for the suggestion - we haven't yet explored it because we weren't familiar with how the build and deployment process changes when depending on a native bindings.

yurishkuro:
My vote is against libraries that depend on native bindings. Not only do they complicate the build in CI, they increase the barrier to entry for new contributors.

isaachier:
Ironically, given my C++ background, I happen to agree. Also, `go get` is completely useless for any Go package using non-Go build dependencies.

clock21am:
is anyone working on this issue, i have bit experience working with Kafka and sarama can take this task if this is okay ?

abhishek-j16:
Is anyone actively working on this? We still see the issue with Jaeger version 1.56(latest as of today). 

We see that rebalancing is not trigerred upon change in partitions, eventually causing data loss.

PR #1087, opened 2018-09-26
Deadlock detector hack for Kafka driver instability

vprithvi:

- Mitigate effects of #1052 as a temporary solution

- Adds a deadlock watcher per partition that monitors message consumption rate per minute. If the rate of consumption is zero, it triggers a rebalance by sending a signal to close the partition. If the partition close is unsuccessful, the instance is killed.

### DIFF
--- a/cmd/ingester/app/consumer/consumer.go
+++ b/cmd/ingester/app/consumer/consumer.go
@@ -43,8 +43,10 @@ type Consumer struct {
 	internalConsumer consumer.Consumer
 	processorFactory ProcessorFactory
 
-	deadlockDetectorFactory deadlockDetectorFactory
-	partitionIDToState      map[int32]*consumerState
+	deadlockDetectorFactory      deadlockDetectorFactory
+	allPartitionDeadlockDetector *deadlockDetector
+
+	partitionIDToState map[int32]*consumerState
 }
 
 type consumerState struct {
@@ -54,13 +56,15 @@ type consumerState struct {
 
 // New is a constructor for a Consumer
 func New(params Params) (*Consumer, error) {
+	deadlockDetectorFactory := newDeadlockDetectorFactory(params.Factory, params.Logger, time.Minute)
 	return &Consumer{
-		metricsFactory:          params.Factory,
-		logger:                  params.Logger,
-		internalConsumer:        params.InternalConsumer,
-		processorFactory:        params.ProcessorFactory,
-		deadlockDetectorFactory: newDeadlockDetectorFactory(params.Factory, params.Logger, time.Minute),
-		partitionIDToState:      make(map[int32]*consumerState),
+		metricsFactory:               params.Factory,
+		logger:                       params.Logger,
+		internalConsumer:             params.InternalConsumer,
+		processorFactory:             params.ProcessorFactory,
+		deadlockDetectorFactory:      deadlockDetectorFactory,
+		allPartitionDeadlockDetector: deadlockDetectorFactory.startMonitoring(),
+		partitionIDToState:           make(map[int32]*consumerState),
 	}, nil
 }
 
@@ -90,6 +94,7 @@ func (c *Consumer) Close() error {
 		c.closePartition(p.partitionConsumer)
 		p.wg.Wait()
 	}
+	c.allPartitionDeadlockDetector.close()
 	c.logger.Info("Closing parent consumer")
 	return c.internalConsumer.Close()
 }
@@ -104,8 +109,8 @@ func (c *Consumer) handleMessages(pc sc.PartitionConsumer) {
 
 	var msgProcessor processor.SpanProcessor
 
-	deadlockDetectorFactory := c.deadlockDetectorFactory.startMonitoringForPartition(pc.Partition())
-	defer deadlockDetectorFactory.close()
+	deadlockDetector := c.deadlockDetectorFactory.startMonitoringForPartition(pc.Partition())
+	defer deadlockDetector.close()
 
 	for {
 		select {
@@ -118,7 +123,8 @@ func (c *Consumer) handleMessages(pc sc.PartitionConsumer) {
 			msgMetrics.counter.Inc(1)
 			msgMetrics.offsetGauge.Update(msg.Offset)
 			msgMetrics.lagGauge.Update(pc.HighWaterMarkOffset() - msg.Offset - 1)
-			deadlockDetectorFactory.incrementMsgCount()
+			deadlockDetector.incrementMsgCount()
+			c.allPartitionDeadlockDetector.incrementMsgCount()
 
 			if msgProcessor == nil {
 				msgProcessor = c.processorFactory.new(pc.Partition(), msg.Offset-1)
@@ -127,7 +133,7 @@ func (c *Consumer) handleMessages(pc sc.PartitionConsumer) {
 
 			msgProcessor.Process(&saramaMessageWrapper{msg})
 
-		case <-deadlockDetectorFactory.getClosePartition():
+		case <-deadlockDetector.getClosePartition():
 			c.logger.Info("Closing partition due to inactivity", zap.Int32("partition", pc.Partition()))
 			return
 		}
--- a/cmd/ingester/app/consumer/consumer.go
+++ b/cmd/ingester/app/consumer/consumer.go
@@ -43,8 +43,8 @@ type Consumer struct {
 	internalConsumer consumer.Consumer
 	processorFactory ProcessorFactory
 
-	seppukuFactory     seppukuFactory
-	partitionIDToState map[int32]*consumerState
+	deadlockDetectorFactory deadlockDetectorFactory
+	partitionIDToState      map[int32]*consumerState
 }
 
 type consumerState struct {
@@ -55,12 +55,12 @@ type consumerState struct {
 // New is a constructor for a Consumer
 func New(params Params) (*Consumer, error) {
 	return &Consumer{
-		metricsFactory:     params.Factory,
-		logger:             params.Logger,
-		internalConsumer:   params.InternalConsumer,
-		processorFactory:   params.ProcessorFactory,
-		seppukuFactory:     newSeppukuFactory(params.Factory, params.Logger, time.Minute),
-		partitionIDToState: make(map[int32]*consumerState),
+		metricsFactory:          params.Factory,
+		logger:                  params.Logger,
+		internalConsumer:        params.InternalConsumer,
+		processorFactory:        params.ProcessorFactory,
+		deadlockDetectorFactory: newDeadlockDetectorFactory(params.Factory, params.Logger, time.Minute),
+		partitionIDToState:      make(map[int32]*consumerState),
 	}, nil
 }
 
@@ -104,8 +104,8 @@ func (c *Consumer) handleMessages(pc sc.PartitionConsumer) {
 
 	var msgProcessor processor.SpanProcessor
 
-	seppuku := c.seppukuFactory.startMonitoringForPartition(pc.Partition())
-	defer seppuku.close()
+	deadlockDetectorFactory := c.deadlockDetectorFactory.startMonitoringForPartition(pc.Partition())
+	defer deadlockDetectorFactory.close()
 
 	for {
 		select {
@@ -118,7 +118,7 @@ func (c *Consumer) handleMessages(pc sc.PartitionConsumer) {
 			msgMetrics.counter.Inc(1)
 			msgMetrics.offsetGauge.Update(msg.Offset)
 			msgMetrics.lagGauge.Update(pc.HighWaterMarkOffset() - msg.Offset - 1)
-			seppuku.incrementMsgCount()
+			deadlockDetectorFactory.incrementMsgCount()
 
 			if msgProcessor == nil {
 				msgProcessor = c.processorFactory.new(pc.Partition(), msg.Offset-1)
@@ -127,7 +127,7 @@ func (c *Consumer) handleMessages(pc sc.PartitionConsumer) {
 
 			msgProcessor.Process(&saramaMessageWrapper{msg})
 
-		case <-seppuku.getClosePartition():
+		case <-deadlockDetectorFactory.getClosePartition():
 			c.logger.Info("Closing partition due to inactivity", zap.Int32("partition", pc.Partition()))
 			return
 		}
--- a/cmd/ingester/app/consumer/consumer.go
+++ b/cmd/ingester/app/consumer/consumer.go
@@ -16,6 +16,7 @@ package consumer
 
 import (
 	"sync"
+	"time"
 
 	"github.com/Shopify/sarama"
 	sc "github.com/bsm/sarama-cluster"
@@ -42,6 +43,7 @@ type Consumer struct {
 	internalConsumer consumer.Consumer
 	processorFactory ProcessorFactory
 
+	seppukuFactory     seppukuFactory
 	partitionIDToState map[int32]*consumerState
 }
 
@@ -57,6 +59,7 @@ func New(params Params) (*Consumer, error) {
 		logger:             params.Logger,
 		internalConsumer:   params.InternalConsumer,
 		processorFactory:   params.ProcessorFactory,
+		seppukuFactory:     newSeppukuFactory(params.Factory, params.Logger, time.Minute),
 		partitionIDToState: make(map[int32]*consumerState),
 	}, nil
 }
@@ -73,6 +76,7 @@ func (c *Consumer) Start() {
 				// to the cleanup process not completing
 				p.wg.Wait()
 			}
+			c.newPartitionMetrics(pc.Partition()).startCounter.Inc(1)
 			c.partitionIDToState[pc.Partition()] = &consumerState{partitionConsumer: pc}
 			go c.handleMessages(pc)
 			go c.handleErrors(pc.Partition(), pc.Errors())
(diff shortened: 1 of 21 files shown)


cmd/ingester/app/consumer/consumer.go
```diff
@@ -97,27 +101,43 @@ func (c *Consumer) handleMessages(pc sc.PartitionConsumer) {
...
+			}
+			c.logger.Debug("Got msg", zap.Any("msg", msg))
+			msgMetrics.counter.Inc(1)
+			msgMetrics.offsetGauge.Update(msg.Offset)
+			msgMetrics.lagGauge.Update(pc.HighWaterMarkOffset() - msg.Offset - 1)
+			seppuku.incrementMsgCount()
+
+			if msgProcessor == nil {
+				msgProcessor = c.processorFactory.new(pc.Partition(), msg.Offset-1)
+				defer msgProcessor.Close()
```
yurishkuro: I find it strange that we init the processor inside the loop and defer close. What's the reason for not doing it in L105?

cmd/ingester/app/consumer/sepukku.go
```diff
@@ -0,0 +1,107 @@
...
+		logger:         logger,
+		interval:       interval,
+		panicFunc: func(partition int32) {
+			factory.Counter("seppuku", map[string]string{"partition": strconv.Itoa(int(partition))}).Inc(1)
+			time.Sleep(time.Second) // Allow time to flush metric
+
+			buf := make([]byte, 1<<20)
+			logger.Panic("No messages processed in the last check interval",
+				zap.Int32("partition", partition),
+				zap.String("stack", string(buf[:runtime.Stack(buf, true)])))
```
yurishkuro: in hotrod zap logger automatically prints the stack. What's different here?

cmd/ingester/app/consumer/sepukku.go
```diff
@@ -0,0 +1,107 @@
...
+// If the message send isn't processed within the next check interval, a seppuku is attempted.This hack relies on a
+// container management system (k8s, aurora, marathon, etc) to reschedule
+// the dead instance.
+//
+// This hack protects jaeger-ingester from issues described in  https://github.com/jaegertracing/jaeger/issues/1052
+//
+func (s *seppukuFactory) startMonitoringForPartition(partition int32) *seppukuWorker {
+	var msgConsumed uint64
+	w := &seppukuWorker{
+		msgConsumed:    &msgConsumed,
```
yurishkuro: no reason to allocate a pointer, you can still use address of a struct field with atomic.

cmd/ingester/app/consumer/sepukku.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	go func() {
+		for range w.ticker.C {
+			if atomic.LoadUint64(w.msgConsumed) == 0 {
+				select {
+				case w.closePartition <- struct{}{}:
+					s.logger.Warn("Signalling partition close due to inactivity", zap.Int32("partition", partition))
+				default:
+					// If closePartition is blocked, attempt seppuku
```
yurishkuro: // If closePartition is blocked, the consumer may have deadlocked -> kill the process

cmd/ingester/app/consumer/sepukku.go
```diff
@@ -0,0 +1,107 @@
...
+
+	return w
+}
+
+func (w *seppukuWorker) getClosePartition() chan struct{} {
+	return w.closePartition
+}
+
+func (w *seppukuWorker) close() {
+	w.ticker.Stop()
```
yurishkuro: this could inadvertently kill the process after rebalance, but not sure how to avoid it

cmd/ingester/app/consumer/sepukku.go
vprithvi: zap only prints the current go routine - This prints out all go routines.

cmd/ingester/app/consumer/consumer.go
vprithvi: It is because sarama-cluster doesn't allow us to retrieve the message offset without reading the first message.

cmd/ingester/app/consumer/sepukku.go
vprithvi: True, I had assumed that the pointer allocation had a similar overhead to an field allocation. I didn't want to use `&` on every access.

cmd/ingester/app/consumer/sepukku.go
yurishkuro: actually, because Ticker.close() does not close the channel, the race condition I was thinking about won't happen on rebalance, but you will leak the goroutine

vprithvi: This is true, I can use a separate channel to close this if you feel strongly about it

yurishkuro: yes, let's fix this, a goroutine leak is not good

cmd/ingester/app/consumer/consumer.go
```diff
@@ -52,12 +56,15 @@ type consumerState struct {
 
 // New is a constructor for a Consumer
 func New(params Params) (*Consumer, error) {
+	deadlockDetectorFactory := newDeadlockDetectorFactory(params.Factory, params.Logger, time.Minute)
```
yurishkuro: separate issue: s/params.Factory/params.MetricsFactory/

cmd/ingester/app/consumer/consumer.go
```diff
@@ -97,27 +106,44 @@ func (c *Consumer) handleMessages(pc sc.PartitionConsumer) {
...
+
+			if msgProcessor == nil {
+				msgProcessor = c.processorFactory.new(pc.Partition(), msg.Offset-1)
+				defer msgProcessor.Close()
+			}
+
+			msgProcessor.Process(&saramaMessageWrapper{msg})
 
-		msgProcessor.Process(&saramaMessageWrapper{msg})
+		case <-deadlockDetector.getClosePartition():
```
yurishkuro: s/getClosePartition/closePartitionChannel/

cmd/ingester/app/consumer/consumer.go
```diff
@@ -97,27 +106,44 @@ func (c *Consumer) handleMessages(pc sc.PartitionConsumer) {
...
+			return
+		}
 	}
-	c.logger.Info("Finished handling messages", zap.Int32("partition", pc.Partition()))
 }
 
 func (c *Consumer) closePartition(partitionConsumer sc.PartitionConsumer) {
 	c.logger.Info("Closing partition consumer", zap.Int32("partition", partitionConsumer.Partition()))
 	partitionConsumer.Close() // blocks until messages channel is drained
+	c.newPartitionMetrics(partitionConsumer.Partition()).closeCounter.Inc(1)
```
yurishkuro: why do we call `new` here? If it internally caches the metrics, then `s/newPartitionMetrics/newPartitionMetrics`

cmd/ingester/app/consumer/consumer_metrics.go
```diff
@@ -30,8 +30,17 @@ type errMetrics struct {
 	errCounter metrics.Counter
 }
 
+type partitionMetrics struct {
+	startCounter metrics.Counter
+	closeCounter metrics.Counter
+}
+
+func (c *Consumer) getNamespace(partition int32) metrics.Factory {
```
yurishkuro: s/getNamespace/metricsFactoryForPartition/

yurishkuro: NB: we don't use "get" in Go.

cmd/ingester/app/consumer/consumer.go
```diff
@@ -97,27 +106,44 @@ func (c *Consumer) handleMessages(pc sc.PartitionConsumer) {
...
+			if !ok {
+				c.logger.Info("Message channel closed. ", zap.Int32("partition", pc.Partition()))
+				return
+			}
+			c.logger.Debug("Got msg", zap.Any("msg", msg))
+			msgMetrics.counter.Inc(1)
+			msgMetrics.offsetGauge.Update(msg.Offset)
+			msgMetrics.lagGauge.Update(pc.HighWaterMarkOffset() - msg.Offset - 1)
+			deadlockDetector.incrementMsgCount()
+			c.allPartitionDeadlockDetector.incrementMsgCount()
```
yurishkuro: I don't follow what the purpose of this allPartitionDeadlockDetector is. It looks like it's only used to increment this counter - why do we even need it? We can always sum the time series to get total counter.

cmd/ingester/app/consumer/deadlock_detector.go
```diff
@@ -0,0 +1,127 @@
...
+	w := &deadlockDetector{
+		msgConsumed:    &msgConsumed,
+		ticker:         time.NewTicker(s.interval),
+		closePartition: make(chan struct{}, 1),
+		done:           make(chan struct{}),
+		logger:         s.logger,
+	}
+
+	go func() {
+		for range w.ticker.C {
```
yurishkuro: this is an odd pattern. If you close(done), nothing will happen till the ticker fires. You probably want unconditional for{} and then select across all channels, including the ticker.

cmd/ingester/app/consumer/deadlock_detector.go
```diff
@@ -0,0 +1,127 @@
...
+func (s *deadlockDetectorFactory) startMonitoring() *deadlockDetector {
+	return s.startMonitoringForPartition(-1)
+}
+
+func (w *deadlockDetector) getClosePartition() chan struct{} {
+	return w.closePartition
+}
+
+func (w *deadlockDetector) close() {
+	w.logger.Info("Closing deadlock detector")
```
yurishkuro: nit: better to capture partition # in the detector and include it in the log stmt

cmd/ingester/app/consumer/deadlock_detector.go
```diff
@@ -0,0 +1,127 @@
...
+// container management system (k8s, aurora, marathon, etc) to reschedule
+// the dead instance.
+//
+// This hack protects jaeger-ingester from issues described in  https://github.com/jaegertracing/jaeger/issues/1052
+//
+func (s *deadlockDetectorFactory) startMonitoringForPartition(partition int32) *deadlockDetector {
+	var msgConsumed uint64
+	w := &deadlockDetector{
+		msgConsumed:    &msgConsumed,
+		ticker:         time.NewTicker(s.interval),
```
yurishkuro: nit: you do not need to leak the ticker. Create it inside goroutine as local var with defer close.

cmd/ingester/app/consumer/deadlock_detector.go
```diff
@@ -0,0 +1,127 @@
...
+// closed by sending a message on the closePartition channel.
+//
+// Closing the partition should result in a rebalance, which alleviates the condition. This means that rebalances can
+// happen frequently if there is no traffic on the Kafka topic. This shouldn't affect normal operations.
+//
+// If the message send isn't processed within the next check interval, a panic is issued.This hack relies on a
+// container management system (k8s, aurora, marathon, etc) to reschedule
+// the dead instance.
+//
+// This hack protects jaeger-ingester from issues described in  https://github.com/jaegertracing/jaeger/issues/1052
```
yurishkuro: may want to move this to the struct comment.

cmd/ingester/app/consumer/consumer.go
```diff
@@ -42,6 +43,9 @@ type Consumer struct {
 	internalConsumer consumer.Consumer
 	processorFactory ProcessorFactory
 
+	deadlockDetectorFactory      deadlockDetectorFactory
+	allPartitionDeadlockDetector *deadlockDetector
```
yurishkuro: nit: allPartitions...

cmd/ingester/app/consumer/consumer.go
```diff
@@ -42,6 +43,9 @@ type Consumer struct {
 	internalConsumer consumer.Consumer
 	processorFactory ProcessorFactory
 
+	deadlockDetectorFactory      deadlockDetectorFactory
```
yurishkuro: I think it would be cleaner and easier to understand if you had a top-level `deadlockDetector`, which can create `partitionDeadlockDetector` as needed (implementation-wide the former may contain the latter for pId=-1). So factory is only used once to create the top-level detector, and the factory does not need to be stored in the Consumer.

cmd/ingester/app/consumer/deadlock_detector.go
```diff
@@ -0,0 +1,127 @@
...
+	logger         *zap.Logger
+	interval       time.Duration
+	panicFunc      func(int32)
+}
+
+type deadlockDetector struct {
+	msgConsumed    *uint64
+	ticker         *time.Ticker
+	logger         *zap.Logger
+	closePartition chan struct{}
```
yurishkuro: I assume this does not apply to all-partitions detector?

vprithvi: It does not

cmd/ingester/app/consumer/consumer.go
vprithvi: Doesn't this mean that the top level `deadlockDetector` also has the responsibilities of the Factor? (That being said, I think it might be cleaner design)

cmd/ingester/app/consumer/consumer.go
vprithvi: I'll address this separately - it'll only add noise to this PR

cmd/ingester/app/consumer/deadlock_detector.go
vprithvi: 👍

cmd/ingester/app/consumer/deadlock_detector.go
vprithvi: 👍

cmd/ingester/app/consumer/consumer.go
vprithvi: 👍

cmd/ingester/app/consumer/consumer.go
vprithvi: 👍

cmd/ingester/app/consumer/deadlock_detector.go
```diff
@@ -0,0 +1,177 @@
...
+		for {
+			select {
+			case <-w.done:
+				s.logger.Info("Closing ticker routine", zap.Int32("partition", partition))
+				return
+			case <-ticker.C:
+				if atomic.LoadUint64(w.msgConsumed) == 0 {
+					select {
+					case w.closePartition <- struct{}{}:
+						s.logger.Warn("Signalling partition close due to inactivity", zap.Int32("partition", partition))
```
yurishkuro: should we have metric here?

cmd/ingester/app/consumer/deadlock_detector.go
```diff
@@ -0,0 +1,177 @@
...
+
+type allPartitionsDeadlockDetector struct {
+	msgConsumed *uint64
+	logger      *zap.Logger
+	done        chan struct{}
+}
+
+func newDeadlockDetector(factory metrics.Factory, logger *zap.Logger, interval time.Duration) deadlockDetector {
+	panicFunc := func(partition int32) {
+		factory.Counter("deadlockdetector.panic-issued", map[string]string{"partition": strconv.Itoa(int(partition))}).Inc(1)
```
yurishkuro: assumes that factory is reentrant for the same metric name, which is not guaranteed

cmd/ingester/app/consumer/consumer.go
```diff
@@ -97,27 +105,43 @@ func (c *Consumer) handleMessages(pc sc.PartitionConsumer) {
...
+			c.logger.Info("Closing partition due to inactivity", zap.Int32("partition", pc.Partition()))
+			return
+		}
 	}
-	c.logger.Info("Finished handling messages", zap.Int32("partition", pc.Partition()))
 }
 
 func (c *Consumer) closePartition(partitionConsumer sc.PartitionConsumer) {
 	c.logger.Info("Closing partition consumer", zap.Int32("partition", partitionConsumer.Partition()))
 	partitionConsumer.Close() // blocks until messages channel is drained
```
yurishkuro: > // blocks until messages channel is drained

what would be draining them?

cmd/ingester/app/consumer/consumer_metrics.go
```diff
@@ -40,7 +49,12 @@ func (c *Consumer) newMsgMetrics(partition int32) msgMetrics {
...
-	f := c.metricsFactory.Namespace("sarama-consumer", map[string]string{"partition": strconv.Itoa(int(partition))})
-	return errMetrics{errCounter: f.Counter("errors", nil)}
+	return errMetrics{errCounter: c.namespace(partition).Counter("errors", nil)}
+}
 
+func (c *Consumer) partitionMetrics(partition int32) partitionMetrics {
+	f := c.namespace(partition)
+	return partitionMetrics{
+		closeCounter: f.Counter("partition-close", nil),
+		startCounter: f.Counter("partition-start", nil)}
```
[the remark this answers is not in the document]
vprithvi: I added a test to [jaeger-lib/metrics](https://github.com/jaegertracing/jaeger-lib/pull/50/files) which shows that calling `Counter` mutiple times with the same tags does not panic for expvar, prometheus and tally.

cmd/ingester/app/consumer/deadlock_detector.go
```diff
@@ -0,0 +1,177 @@
...
+	allPartitionsDeadlockDetector *allPartitionsDeadlockDetector
+}
+
+type allPartitionsDeadlockDetector struct {
+	msgConsumed *uint64
+	logger      *zap.Logger
+	done        chan struct{}
+}
+
+func newDeadlockDetector(factory metrics.Factory, logger *zap.Logger, interval time.Duration) deadlockDetector {
```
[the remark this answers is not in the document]
vprithvi: Ah, this is a result of some laziness and accepting the IDE generated name - I'll update these

cmd/ingester/app/consumer/consumer.go
vprithvi: The goroutine from `handleMessages` would be draining them

cmd/ingester/app/consumer/consumer_metrics.go
```diff
@@ -30,8 +30,17 @@ type errMetrics struct {
 	errCounter metrics.Counter
 }
 
+type partitionMetrics struct {
```
yurishkuro: nit: we should clean-up these metrics. All 3 structs are per-partition, could we not combine them into one? Then we won't have all these small functions on the Consumer.

cmd/ingester/app/consumer/deadlock_detector.go
```diff
@@ -0,0 +1,181 @@
...
+		closePartition: make(chan struct{}, 1),
+		done:           make(chan struct{}),
+		logger:         s.logger,
+
+		incrementAllPartitionMsgCount: func() {
+			s.allPartitionsDeadlockDetector.incrementMsgCount()
+		},
+	}
+
+	go func() {
```
yurishkuro: nit: rather large & deep function, should not be inlined.

cmd/ingester/app/consumer/consumer_metrics.go
vprithvi: I agree - I'll do it as a separate commit

PR #1092, opened 2018-09-28
Refactor agent configuration

pavolloffay:
Resolves: #927 

* agent tchannel flags moved to `reporter.tchannel.*`
* agent reporters have to be specified explicitly `b.WithReporters` - if none added create fails
* agent client configuration manager have to be specified explicitly `b.WithClientConfigManager` - if none added create fails

### DIFF
--- a/cmd/agent/app/builder.go
+++ b/cmd/agent/app/builder.go
@@ -126,10 +126,6 @@ func (b *Builder) CreateAgent(logger *zap.Logger) (*Agent, error) {
 	if err != nil {
 		return nil, errors.Wrap(err, "cannot create metrics factory")
 	}
-
-	if err != nil {
-		return nil, err
-	}
 	r, err := b.getReporter(logger)
 	if err != nil {
 		return nil, err
--- a/cmd/agent/app/builder.go
+++ b/cmd/agent/app/builder.go
@@ -17,18 +17,15 @@ package app
 import (
 	"fmt"
 	"net/http"
-	"time"
 
 	"github.com/apache/thrift/lib/go/thrift"
 	"github.com/pkg/errors"
 	"github.com/uber/jaeger-lib/metrics"
-	"github.com/uber/tchannel-go"
 	"go.uber.org/zap"
 
 	"github.com/jaegertracing/jaeger/cmd/agent/app/httpserver"
 	"github.com/jaegertracing/jaeger/cmd/agent/app/processors"
 	"github.com/jaegertracing/jaeger/cmd/agent/app/reporter"
-	tchreporter "github.com/jaegertracing/jaeger/cmd/agent/app/reporter/tchannel"
 	"github.com/jaegertracing/jaeger/cmd/agent/app/servers"
 	"github.com/jaegertracing/jaeger/cmd/agent/app/servers/thriftudp"
 	jmetrics "github.com/jaegertracing/jaeger/pkg/metrics"
@@ -37,11 +34,9 @@ import (
 )
 
 const (
-	defaultQueueSize        = 1000
-	defaultMaxPacketSize    = 65000
-	defaultServerWorkers    = 10
-	defaultMinPeers         = 3
-	defaultConnCheckTimeout = 250 * time.Millisecond
+	defaultQueueSize     = 1000
+	defaultMaxPacketSize = 65000
+	defaultServerWorkers = 10
 
 	defaultHTTPServerHostPort = ":5778"
 
@@ -73,10 +68,9 @@ type Builder struct {
 	HTTPServer HTTPServerConfiguration  `yaml:"httpServer"`
 	Metrics    jmetrics.Builder         `yaml:"metrics"`
 
-	tchreporter.Builder `yaml:",inline"`
-
-	otherReporters []reporter.Reporter
+	reporters      []reporter.Reporter
 	metricsFactory metrics.Factory
+	configManager  httpserver.ClientConfigManager
 }
 
 // ProcessorConfiguration holds config for a processor that receives spans from Server
@@ -99,9 +93,9 @@ type HTTPServerConfiguration struct {
 	HostPort string `yaml:"hostPort" validate:"nonzero"`
 }
 
-// WithReporter adds auxiliary reporters.
-func (b *Builder) WithReporter(r reporter.Reporter) *Builder {
-	b.otherReporters = append(b.otherReporters, r)
+// WithReporters adds auxiliary reporters.
+func (b *Builder) WithReporters(r ...reporter.Reporter) *Builder {
+	b.reporters = append(b.reporters, r...)
 	return b
 }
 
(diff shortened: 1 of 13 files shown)


cmd/agent/app/builder.go
```diff
@@ -100,8 +98,11 @@ type HTTPServerConfiguration struct {
 }
 
 // WithReporter adds auxiliary reporters.
-func (b *Builder) WithReporter(r reporter.Reporter) *Builder {
-	b.otherReporters = append(b.otherReporters, r)
+func (b *Builder) WithReporter(r reporter.ReporterBuilder) *Builder {
```
jpkrohling: For multiple builders, I think this is meant to be called more than once, right?

```go
b.WithReporter(reporterBuilderA).WithReporter(reporterBuilderB)
```

Wouldn't it be better to accept multiple builders at once? Like:

```go
b.WithReporters(reporterBuilderA, reporterBuilderB)
```

cmd/agent/app/reporter/config.go
```diff
@@ -0,0 +1,42 @@
...
+
+func (s SpecificConfiguration) Configuration() Configuration {
+	return s
+}
+
+type ConfigurationProvider interface {
+	Configuration() Configuration
+}
+
+func a() {
```
jpkrohling: huh?

cmd/agent/app/reporter/tchannel/reporter.go
```diff
@@ -57,6 +57,7 @@ type Reporter struct {
 	peerListMgr    *peerlistmgr.PeerListManager
 	batchesMetrics map[string]batchMetrics
 	logger         *zap.Logger
+	serviceName string
```
jpkrohling: This looks odd... Is it missing a `go fmt`?

cmd/all-in-one/main.go
```diff
@@ -175,9 +176,16 @@ func startAgent(
 	baseFactory metrics.Factory,
 ) {
 	metricsFactory := baseFactory.Namespace("agent", nil)
-
-	if len(b.CollectorHostPorts) == 0 {
-		b.CollectorHostPorts = append(b.CollectorHostPorts, fmt.Sprintf("127.0.0.1:%d", cOpts.CollectorPort))
+	fmt.Println("all main")
```
jpkrohling: I assume all those prints will eventually go away, right?

cmd/all-in-one/main.go
```diff
@@ -175,9 +176,16 @@ func startAgent(
...
+	fmt.Printf("%v", b.ReporterBuilders)
+	fmt.Println(len(b.ReporterBuilders))
+	for _, rb := range b.ReporterBuilders {
+		if tb, ok := rb.(agentRepTchan.Builder); ok {
+			fmt.Println("adding reporter\n\n")
+			if len(tb.CollectorHostPorts) == 0 {
+				tb.CollectorHostPorts = append(tb.CollectorHostPorts, fmt.Sprintf("127.0.0.1:%d", cOpts.CollectorPort))
+			}
+		}
```
jpkrohling: There's no other possible reporter builder yet, but you never know the future... Perhaps it would be better to add a default case where, to log a warning in in case the type isn't recognized.

cmd/agent/app/builder.go
```diff
@@ -100,8 +98,11 @@ type HTTPServerConfiguration struct {
...
+	if b.ReporterBuilders == nil {
+		b.ReporterBuilders = []reporter.ReporterBuilder{}
+	}
```
yurishkuro: not needed, append() works just fine with nil

cmd/agent/app/builder.go
yurishkuro: +1, can make it a vararg

cmd/agent/app/reporter/tchannel/reporter.go
pavolloffay: it's wip without formatting and tests

pavolloffay:
@yurishkuro @jpkrohling PR updated. I think all concerns were addressed.

cmd/agent/main.go
```diff
@@ -50,9 +52,19 @@ func main() {
 
 			builder := &app.Builder{}
 			builder.InitFromViper(v)
+			tchanRep := tchannel.NewBuilder().InitFromViper(v)
+			mFactory, err := builder.GetMetricsFactory()
+			if err != nil {
+				logger.Fatal("Could not create metrics", zap.Error(err))
+			}
+			r, err := tchanRep.CreateReporter(mFactory, logger)
```
yurishkuro: I wonder if it would make more sense to encapsulate "tchannel collector proxy" that will either implement or at least expose both Reporter and ClientConfigManager APIs. It would make the code in main() simpler (and code itself can be unit tested), and will be easier to switch for gRPC implementation.

Right now `httpserver.NewCollectorProxy()` below is leaking tchannel detail outside of tchannel package.

docker-compose/jaeger-docker-compose.yml
```diff
@@ -25,7 +25,7 @@ services:
 
     jaeger-agent:
       image: jaegertracing/jaeger-agent
-      command: ["--collector.host-port=jaeger-collector:14267"]
+      command: ["--reporter.tchannel.collector.host-port=jaeger-collector:14267"]
```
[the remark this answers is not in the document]
pavolloffay: I will try to depreciate in a way that values from old flags would be used and logged a warning. 

I don't want to keep old flags just as noop. Then I think it's better to fail at init time to avoid misconfiguration.

cmd/agent/app/builder.go
```diff
@@ -184,13 +178,25 @@ func (b *Builder) GetProcessors(rep reporter.Reporter, mFactory metrics.Factory,
...
+func (b *Builder) WithClientConfigManager(manager httpserver.ClientConfigManager) *Builder {
+	b.configManager = manager
+	return b
+}
+
 // GetHTTPServer creates an HTTP server that provides sampling strategies and baggage restrictions to client libraries.
-func (c HTTPServerConfiguration) GetHTTPServer(svc string, channel *tchannel.Channel, mFactory metrics.Factory) *http.Server {
-	mgr := httpserver.NewCollectorProxy(svc, channel, mFactory)
+func (c HTTPServerConfiguration) getHTTPServer(manager httpserver.ClientConfigManager, mFactory metrics.Factory, mBuilder *jmetrics.Builder) (*http.Server, error) {
+	if manager == nil {
```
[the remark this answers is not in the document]
pavolloffay: I have added these checks to better understand requirements. manager and reporter is mandatory. We can also change the builder to `build(reporters, manager)`

cmd/agent/main.go
```diff
@@ -50,9 +52,19 @@ func main() {
 
 			builder := &app.Builder{}
 			builder.InitFromViper(v)
+			tchanRep := tchannel.NewBuilder().InitFromViper(v)
+			mFactory, err := builder.GetMetricsFactory()
```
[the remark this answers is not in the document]
pavolloffay: That is cached

cmd/agent/main.go
```diff
@@ -50,9 +52,19 @@ func main() {
 
 			builder := &app.Builder{}
 			builder.InitFromViper(v)
+			tchanRep := tchannel.NewBuilder().InitFromViper(v)
```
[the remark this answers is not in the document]
pavolloffay: Not sure if I follow. Do you mean doing something similar as `SPAN_STORAGE_TYPE`?

pavolloffay:
@yurishkuro could you please have a look?

yurishkuro:
I'm trying to build out internal versions with this branch.

cmd/agent/app/builder.go
```diff
@@ -99,63 +97,36 @@ type HTTPServerConfiguration struct {
 	HostPort string `yaml:"hostPort" validate:"nonzero"`
 }
 
-// WithReporter adds auxiliary reporters.
-func (b *Builder) WithReporter(r reporter.Reporter) *Builder {
```
yurishkuro: Ran in to an issue here. This needs to stay. We internally used another "reporter" that was providing debugging capabilities.

I don't think it makes sense to support plural WithCollectorProxy(), because (a) they will never be used for sampling config, so the CollectorProxy interface is too wide, and (b) having just WithReporter() makes more sense anyway given that we don't know the purpose of the reporter and cannot make assumptions about it being implemented as a proxy.

pavolloffay: Can you please describe why do you use multiple reporters? 

Note that `CollectorProxy` is added in `build(primaryProxy, m, l)` and also as `withCollectorProxy`. Sampling and baggage manager is used only from the `primaryProxy`. I think this coupling makes sense bc manager is not optional and uses the same transport as reporter. Do you think we will use grpc reporter and tchannel for manager? It still could use this interface - the mains would look simpler.

pavolloffay: I have changed this method signature to `withReporter(reporter.Reporter)` but keeping the build with collector proxy. Does it work for you?

yurishkuro: yep, that's what I meant. We have two cases for additional reports: (1) cloning the traffic to another DC via another tchannel reporter, and (b) sending all received spans to a fake reporter (based on https://github.com/uber-go/gwr) that runs in process and allows attaching listeners at runtime, either for debugging purposes, or for getting a stream of spans for some other type of analysis (conceptually similar to z-pages in OpenCensus).

yurishkuro:
I was able to build our internal agent with these changes, but haven't had a change to test it, esp. loading of the old configs.

cmd/agent/app/builder_test.go
```diff
@@ -51,14 +56,6 @@ processors:
...
 httpServer:
     hostPort: 4.4.4.4:5778
-
-collectorHostPorts:
-    - 127.0.0.1:14267
-    - 127.0.0.1:14268
-    - 127.0.0.1:14269
-
-collectorServiceName: some-collector-service
-minPeers: 4
```
yurishkuro: just to confirm - this is no longer supported? I thought there was a fallback with the options that might make the old config still work.

pavolloffay: old flags are supported and deprecated https://github.com/jaegertracing/jaeger/pull/1092/files#diff-1fc7689d387f4303d6e478ca01841286R29

pavolloffay:
```
Jaeger agent is a daemon program that runs on every host and receives tracing data submitted by Jaeger client libraries.

Usage:
  jaeger-agent [flags]
  jaeger-agent [command]

Available Commands:
  help        Help about any command
  version     Print the version

Flags:
      --collector.host-port string                                Deprecated; comma-separated string representing host:ports of a static list of collectors to connect to directly (e.g. when not using service discovery)
      --config-file string                                        Configuration file in JSON, TOML, YAML, HCL, or Java properties formats (default none). See spf13/viper for precedence.
      --discovery.conn-check-timeout duration                     Deprecated; sets the timeout used when establishing new connections (default 250ms)
      --discovery.min-peers int                                   Deprecated; if using service discovery, the min number of connections to maintain to the backend (default 3)
  -h, --help                                                      help for jaeger-agent
      --http-server.host-port string                              host:port of the http server (e.g. for /sampling point and /baggage endpoint) (default ":5778")
      --log-level string                                          Minimal allowed log Level. For more levels see https://github.com/uber-go/zap (default "info")
      --metrics-backend string                                    Defines which metrics backend to use for metrics reporting: expvar, prometheus, none (default "prometheus")
      --metrics-http-route string                                 Defines the route of HTTP endpoint for metrics backends that support scraping (default "/metrics")
      --processor.jaeger-binary.server-host-port string           host:port for the UDP server (default ":6832")
      --processor.jaeger-binary.server-max-packet-size int        max packet size for the UDP server (default 65000)
      --processor.jaeger-binary.server-queue-size int             length of the queue for the UDP server (default 1000)
      --processor.jaeger-binary.workers int                       how many workers the processor should run (default 10)
      --processor.jaeger-compact.server-host-port string          host:port for the UDP server (default ":6831")
      --processor.jaeger-compact.server-max-packet-size int       max packet size for the UDP server (default 65000)
      --processor.jaeger-compact.server-queue-size int            length of the queue for the UDP server (default 1000)
      --processor.jaeger-compact.workers int                      how many workers the processor should run (default 10)
      --processor.zipkin-compact.server-host-port string          host:port for the UDP server (default ":5775")
      --processor.zipkin-compact.server-max-packet-size int       max packet size for the UDP server (default 65000)
      --processor.zipkin-compact.server-queue-size int            length of the queue for the UDP server (default 1000)
      --processor.zipkin-compact.workers int                      how many workers the processor should run (default 10)
      --reporter.tchannel.collector.host-port string              comma-separated string representing host:ports of a static list of collectors to connect to directly (e.g. when not using service discovery)

Use "jaeger-agent [command] --help" for more information about a command.

```

pavolloffay:
@yurishkuro is there anything more what should be done?

yurishkuro:
We're testing it internally to make sure we can still do multi-reporter configs. Will update soon.

cc @isaachier

pavolloffay:
Any news @isaachier or @yurishkuro ?

PR #1098, opened 2018-09-30
Replace glide novendor, fix gosec lint to exclude crossdock

### LINKED ISSUE #1097: gosec fails on crossdock

Cannot submit new pull requests to GitHub.

gosec fails on crossdock source code. Causes Travis CI errors (https://travis-ci.org/jaegertracing/jaeger/jobs/434740163).

Example:

```
[/Users/isaachier/gocode/src/github.com/jaegertracing/jaeger/crossdock/main.go:92] - G107: Potential HTTP request made with variable url (Confidence: MEDIUM, Severity: MEDIUM)
  > http.Get(healthURL)
```

Instead of excluding error G107, which should be fine in normal Jaeger source code, we should exclude the crossdock directory from unnecessary linting.

isaachier:
- Resolves #1097.

- Exclude crossdock from gosec
- Replace `glide novendor` with custom bash script

### DIFF
--- a/scripts/list-packages.sh
+++ b/scripts/list-packages.sh
@@ -5,9 +5,14 @@ set -euo pipefail
 function listPackages() {
 	local d
 	local dirs
-	dirs=$(find -E . -mindepth 1 -maxdepth 1 -type d -not -regex '\./(vendor|thrift-gen|swagger-gen|examples|scripts)' | sed 's|^./||g')
+	dirs=$(find . -mindepth 1 -maxdepth 1 -type d \
+		-not -path './vendor' \
+		-not -path './thrift-gen' \
+		-not -path './swagger-gen' \
+		-not -path './examples' \
+		-not -path './scripts')
 	for d in $dirs; do
-		find "$d" -name '*.go' -type f -exec echo "./$d/..." \; -quit
+		find "$d" -name '*.go' -type f -exec echo "$d/..." \; -quit
 	done
 }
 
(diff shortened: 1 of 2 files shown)


isaachier:
Test needs to be rerun.

scripts/list-packages.sh
```diff
@@ -0,0 +1,19 @@
...
+		-not -path './thrift-gen' \
+		-not -path './swagger-gen' \
+		-not -path './examples' \
+		-not -path './scripts')
+	for d in $dirs; do
+		find "$d" -name '*.go' -type f -exec echo "$d/..." \; -quit
+	done
+}
+
+listPackages "$1"
```
black-adder: is this variable being used in the function?

scripts/list-packages.sh
```diff
@@ -0,0 +1,19 @@
...
+	local d
+	local dirs
+	dirs=$(find . -mindepth 1 -maxdepth 1 -type d \
+		-not -path './vendor' \
+		-not -path './thrift-gen' \
+		-not -path './swagger-gen' \
+		-not -path './examples' \
+		-not -path './scripts')
+	for d in $dirs; do
+		find "$d" -name '*.go' -type f -exec echo "$d/..." \; -quit
```
black-adder: why is the `/...` needed after the filename?

isaachier: It's what glide does. More importantly, I believe it includes all subpackages.

scripts/list-packages.sh
isaachier: My bad. Should have used this instead of current directory in when invoking `find`.

PR #1101, opened 2018-10-04
Add rate limiting to Kafka ingester

### LINKED ISSUE #1100: Add a rate limiter to jaeger-ingester

none

isaachier:
- Resolves #1100.

- Add rate-limiting channel to Kafka ingester.
- Fix typo in consumer.


cmd/ingester/app/consumer/consumer.go
```diff
@@ -52,12 +56,18 @@ type consumerState struct {
 
 // New is a constructor for a Consumer
 func New(params Params) (*Consumer, error) {
+	var rateLimiter *time.Ticker
```
vprithvi: Could you encapsulate the rate limiter in it's own struct?

cmd/ingester/app/consumer/consumer_test.go
```diff
@@ -86,9 +86,17 @@ func newConsumer(
 	factory metrics.Factory,
 	topic string,
 	processor processor.SpanProcessor,
-	consumer consumer.Consumer) *Consumer {
+	consumer consumer.Consumer,
+	maxReadsPerSecond float64) *Consumer {
 
 	logger, _ := zap.NewDevelopment()
+
+	var rateLimiter *time.Ticker
```
vprithvi: Could you add more test cases? 
For e.g., 
If the rate limiter is configured to 0, does it allow any requests?
If the rate limiter is configured to 10qps, does it allow 10 requests?
If the rate limiter is configured to 1qps, and it receives 10 requests, does it allow the 10 requests in 10 seconds?

isaachier: Ya definitely. I agree this deserves direct testing.

cmd/ingester/app/consumer/consumer.go
isaachier: Yes

isaachier:
Something is up with the ES test case.

cmd/ingester/app/flags.go
```diff
@@ -54,13 +56,16 @@ const (
 	DefaultParallelism = 1000
 	// DefaultEncoding is the default span encoding
 	DefaultEncoding = EncodingProto
+	// DefaultMaxReadsPerSecond is the default max reads per second for the span processor
+	DefaultMaxReadsPerSecond = 0.0
```
black-adder: semantically this is wrong; if you pass in 0, you'd expect the consumer to do no work but this value does the exact opposite. Create a new const `DisableRateLimit = 0.0` in rate_limiter.go and use that here instead.

yurishkuro: you can also consider value 0 to be "undefined" and thus not causing any rate limit. Certainly from the CLI API point of view, you should not need to pass any flags to run without a rate limit, and if you do want a rate limit you pass a non-0 value.

cmd/ingester/app/flags.go
```diff
@@ -85,6 +90,10 @@ func AddFlags(flagSet *flag.FlagSet) {
 		ConfigPrefix+SuffixEncoding,
 		DefaultEncoding,
 		fmt.Sprintf(`The encoding of spans ("%s" or "%s") consumed from kafka`, EncodingProto, EncodingJSON))
+	flagSet.String(
```
yurishkuro: why is it a string and not a number?

cmd/ingester/app/consumer/consumer.go
```diff
@@ -100,6 +105,7 @@ func (c *Consumer) handleMessages(pc sc.PartitionConsumer) {
 	var msgProcessor processor.SpanProcessor
 
 	for msg := range pc.Messages() {
+		c.rateLimiter.await()
```
yurishkuro: Alternatively you could wrap `pc.Messages()` channel into another channel that limits the rate of produced messages.  Then your reading code wouldn't be exposed to the rate limiting API, and you can bypass wrapping the channel if the rate parameter is 0.

cmd/ingester/app/flags.go
isaachier: I was using `-1` for this, but used `0` instead seeing as Go default initialization makes this hard to get wrong. I do not mind using another value if you'd like. Not sure what makes the most sense here.

cmd/ingester/app/flags.go
isaachier: That's a great question but I based this snippet on the parallelism flag above (also strangely a string). I can use `Float64` instead.

cmd/ingester/app/consumer/consumer.go
isaachier: So I thought about that. I thought it might be a performance issue to add another channel. Better idea would be to benchmark both. Will get back to you.

cmd/ingester/app/consumer/rate_limiter.go
```diff
@@ -0,0 +1,46 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package consumer
+
+import (
+	"time"
+)
+
+// Rate limiter using ticker explained here: https://gobyexample.com/rate-limiting.
```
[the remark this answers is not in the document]
isaachier: Ya I wanted to give credit but I guess I should add a summary in case that happens.

cmd/ingester/app/consumer/rate_limiter.go
```diff
@@ -0,0 +1,46 @@
...
+
+func newRateLimiter(creditsPerSecond float64) *rateLimiter {
+	if creditsPerSecond <= 0 {
+		return &rateLimiter{}
+	}
+	interval := time.Nanosecond * time.Duration(float64(time.Second.Nanoseconds())/creditsPerSecond)
+	return &rateLimiter{ticker: time.NewTicker(interval)}
+}
+
+func (r *rateLimiter) await() {
```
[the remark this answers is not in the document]
isaachier: Right. I checked Guava's rate limiter to find a name and saw `acquire` (https://github.com/google/guava/blob/master/guava/src/com/google/common/util/concurrent/RateLimiter.java). My own issue was that `acquire` implies `release`, but I guess I'll follow Guava here.

cmd/ingester/app/consumer/rate_limiter_test.go
```diff
@@ -0,0 +1,41 @@
...
+
+	var m sync.Mutex
+	counter := 0
+	for i := 0; i < iterations; i++ {
+		rateLimiter.await()
+		m.Lock()
+		counter++
+		m.Unlock()
+	}
+	require.Equal(t, counter, iterations)
```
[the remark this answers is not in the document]
isaachier: The issue I actually saw was that the rate limiting **did not work** on each iteration, but only as a totality. For example, I had tested that each tick was at least 10 ms apart, and had outright assertion failures because the actual tick came at ~9.64 ms, etc. My guess is the clock is not the same, or the ticker guarantees are much more lax than I had thought.

cmd/ingester/app/consumer/consumer.go
```diff
@@ -99,7 +103,18 @@ func (c *Consumer) handleMessages(pc sc.PartitionConsumer) {
 	msgMetrics := c.newMsgMetrics(pc.Partition())
 	var msgProcessor processor.SpanProcessor
 
-	for msg := range pc.Messages() {
+	// "Cast" channel from chan *sarama.ConsumerMessage to chan interface{}.
```
vprithvi: Could you elaborate which problem we are solving here? 
The casting seems to happen on L117, and L107 is creating a new channel to copy all messages.

cmd/ingester/app/consumer/consumer.go
```diff
@@ -99,7 +103,18 @@ func (c *Consumer) handleMessages(pc sc.PartitionConsumer) {
...
+	msgChannel := make(chan interface{})
+	go func() {
+		defer close(msgChannel)
+		for msg := range pc.Messages() {
+			msgChannel <- msg
+		}
+	}()
+
+	rateLimiter := newRateLimiter(msgChannel, c.maxReadsPerSecond)
+	for v := range rateLimiter.C {
```
vprithvi: In theory, wouldn't this perform worse than the previous variant which called `ratelimiter.await` in this loop?

isaachier: In theory, yes. This is somewhat hard to determine, but I can attempt a benchmark of old vs. new version in the case where no rate limiting is applied (compares await with immediate return against wrapping inner channel). Otherwise, we will just be benchmarking the rate limit, which is useless.

cmd/ingester/app/consumer/consumer.go
isaachier: Definitely will add more detail to comments. A channel such as `chan *sarama.ConsumerMessage` cannot be cast to `chan interface{}`, so I need to forward from the source to the destination channel.

cmd/ingester/app/consumer/rate_limiter.go
```diff
@@ -0,0 +1,108 @@
...
+
+package consumer
+
+import (
+	"sync"
+	"time"
+)
+
+// rateLimiter wraps a channel to provide rate limiting.
+type rateLimiter struct {
```
vprithvi: IMHO, this rate limiter seems pretty complex in that it requires a bunch of channels, a wait group, and a Ticker.

Seeing that this rate limiter is only being used from a single threaded context, and that we are most interested in the stable behavior of the rate limiter,  what advantages does it have over an simpler implementation like one described in https://stackoverflow.com/questions/667508/whats-a-good-rate-limiting-algorithm ?

isaachier:
As I began this channel wrapping, I realized it would probably be worse than a simple rate limiter. I'm glad I went through with it, if only to develop my own understand of channels. Nonetheless, I think I'll revert this latest change in favor of the simpler API.

cmd/ingester/app/consumer/rate_limiter_test.go
```diff
@@ -57,12 +57,12 @@ func TestRateLimiterNoLimit(t *testing.T) {
 	r := newRateLimiter(creditsPerSecond)
 	defer r.Stop()
 
-	const maxInterval = time.Microsecond
+	const maxInterval = time.Millisecond
 	acquireStart := time.Now()
 	for i := 0; i < iterations; i++ {
```
yurishkuro: I don't see any references to `r` in this loop - what is it testing?

isaachier: Good catch. Guess I missed the most important line here.

cmd/ingester/app/consumer/rate_limiter.go
```diff
@@ -0,0 +1,78 @@
...
+	done        chan struct{}
+}
+
+// newRateLimiter constructs a rate limiter with the provided creditsPerSecond.
+// If creditsPerSecond is less than zero, no rate limiting will be
+// provided, and all calls to this object will return immediately. If
+// creditsPerSecond is exactly zero, calls will block indefinitely or until Stop
+// is called.
+func newRateLimiter(creditsPerSecond float64) *rateLimiter {
+	var r *rateLimiter
```
black-adder: I think we can simplify this a bit. Initialize empty tickChannel and ticker at the top and go through the elseif craziness. At the bottom, initialize the rateLimiter. Should reduce the amount of `done:        make(chan struct{})` everywhere. Also, I prefer switch statement here instead of if else (just personal preference).

cmd/ingester/app/consumer/rate_limiter.go
```diff
@@ -0,0 +1,78 @@
...
+	if creditsPerSecond < 0 {
```
black-adder: define -1 and 0 as consts somewhere. Ala
```
disableRateLimiting = -1
noProgress = 0 <- awful name, think of something better
```

cmd/ingester/app/builder/builder.go
```diff
@@ -70,10 +70,11 @@ func CreateConsumer(logger *zap.Logger, metricsFactory metrics.Factory, spanWrit
...
 	consumerParams := consumer.Params{
-		InternalConsumer: saramaConsumer,
-		ProcessorFactory: *processorFactory,
-		Factory:          metricsFactory,
-		Logger:           logger,
+		InternalConsumer:  saramaConsumer,
+		ProcessorFactory:  *processorFactory,
+		Factory:           metricsFactory,
+		Logger:            logger,
+		MaxReadsPerSecond: options.MaxReadsPerSecond,
```
black-adder: ah, I think I've made a mistake when scoping this. This static config means we'd have to redeploy the ingester/indexer everytime we want to throttle vs using a dynamic config value. I think we have a way to plug in dynamic config somewhere else in this repo, let's copy that here.

cmd/ingester/app/consumer/rate_limiter.go
isaachier: Sure. I forgot Go allowed switches with conditions instead of values.

cmd/ingester/app/builder/builder.go
isaachier: Yes. We discussed this, yet I focused on the rate limiter and hadn't thought about refactoring the config. This might be harder to do than it sounds... I have to check other places we did this.

cmd/ingester/app/consumer/rate_limiter_test.go
isaachier: Fixed.

cmd/ingester/app/consumer/rate_limiter.go
vprithvi: I still feel that this general approach is more complex than using a simple token bucket without involving channels. 
What benefits do we gain by doing this? 

See https://github.com/uber-go/ratelimit/blob/c15da02342779cb6dc027fc95ee2277787698f36/ratelimit.go for an example - IMHO it is more readable.

yurishkuro: I tend to agree. That rate limiter just sleeps until the next token is available. Is that worse than using channels for signaling?

cmd/ingester/app/consumer/rate_limiter_test.go
```diff
@@ -0,0 +1,94 @@
...
+	acquireStart := time.Now()
+	for i := 0; i < iterations; i++ {
+		acquireEnd := time.Now()
+		interval := acquireEnd.Sub(acquireStart)
+		require.Truef(t, interval <= maxInterval, "Expected no significant delay before first receive, actual delay = %v, max expected delay = %v", interval, maxInterval)
+		acquireStart = time.Now()
+	}
+}
+
+func TestRateLimiterInfiniteLimit(t *testing.T) {
```
vprithvi: Could you also add a benchmark comparing a empty loop to a loop with a ratelimiter with creditsPerSec set to `math.MaxFloat64`.  I'm curious about the overhead introduced and wanted to verify that it runs > 10k qps.

cmd/ingester/app/consumer/rate_limiter.go
isaachier: I will admit the rate limiter here is also less accurate. The ticker may drop ticks if we aren't receiving from the tick channel frequently enough. My issue with redoing the token bucket is that we should ideally just use the one in the client, and potentially migrate it to jaeger-lib. This is the second time I have reimplemented a token bucket in the project (throttling was first).

isaachier: There is a slightly different semantic, so I guess "less accurate" is not a good way of putting it. The token bucket allows for bursts if enough credits have accumulated. A ticker will not allow for any bursts.

isaachier: Oh man... I already did that in https://github.com/uber/jaeger-lib/pull/35.

vprithvi: This is excellent - could we reuse that ratelimiter?

isaachier: If we weren't expecting a blocking rate limiter, I would have just reused that. The blocking involves calculating a interval and waiting for a timer. The existing rate limiter does not expose its credit balance, making this more difficult.

cmd/ingester/app/consumer/rate_limiter.go
isaachier: Done.

cmd/ingester/app/consumer/rate_limiter.go
vprithvi: Is it very difficult to extend the ratelimiter in `jaeger-lib` to do this?

isaachier:
Hey @vprithvi I have rebased onto latest master and I see the new deadlock detector uses a ticker to calculate throughput, ensuring no consumer is stuck. However, this causes the consumer unit test to fail when rate limiting is applied with the deadlock detector running. Rate limiting inherently changes the throughput. So should I add something to the deadlock detector to allow for rate limiting?

yurishkuro:
why should they interfere?  rate limiting is meant to slow things down, but not stop them completely, while deadlock detector detects complete lack of throughput.

yurishkuro:
Yes, it is exactly the throughput monitor, but it's time horizon should be orders of magnitude larger than the throughput limits of the rate limiter. I.e. the former - no messages in 1min, the latter - up to 1000 msg/sec. So I don't see a fundamental conflict here. You may have clashes in the tests if tests bring those time limits much closer to each other, but they can still be adjusted to be different orders of magnitude.

isaachier:
Right. I'll look at the test case that panics and see what I can do.

isaachier:
Never mind, I was being stupid. The rate limiter change made negative `maxReadsPerSecond` become undefined behavior. I think it was blocking while filling a negative balance. Fixing now.

isaachier:
Any ideas why TestElasticsearchStorage consistently fails in Travis?

cmd/ingester/app/consumer/consumer.go
```diff
@@ -110,8 +114,16 @@ func (c *Consumer) handleMessages(pc sc.PartitionConsumer) {
 
 	deadlockDetector := c.deadlockDetector.startMonitoringForPartition(pc.Partition())
 	defer deadlockDetector.close()
-
+	var rateLimiter *rateLimiter
+	if c.maxReadsPerSecond > 0 {
```
vprithvi: Why allow the `rateLimiter` to be `nil`?
If we consolidate the logic for when `maxReadsPerSecond <= 0` into the rateLImiter, then we can remove the `if` in L124

cmd/ingester/app/consumer/rate_limiter.go
```diff
@@ -0,0 +1,105 @@
...
+// limitations under the License.
+
+package consumer
+
+import (
+	"math"
+	"time"
+)
+
+const cost = 1.0
```
vprithvi: What is the purpose of this?

cmd/ingester/app/consumer/rate_limiter.go
```diff
@@ -0,0 +1,105 @@
...
+
+// Stop stops any ongoing Acquire operation, which exits immediately.
+func (r *rateLimiter) Stop() {
+	close(r.event)
+}
+
+// Update replaces the current rate limiter configuration in a thread-safe
+// manner. This function will panic if Stop has already been called on the
+// rateLimiter.
+func (r *rateLimiter) Update(creditsPerSecond, maxBalance float64) {
```
vprithvi: Where is this used?

cmd/ingester/app/consumer/rate_limiter.go
```diff
@@ -0,0 +1,105 @@
...
+// Acquire causes rateLimiter to spend a credit and return immediately,
+// or block until a credit becomes available. Returns true if credits were
+// acquired. Returns false if Stop was called before credits were acquired,
+// interrupting the Acquire call.
+// Note: Acquire has no corresponding release operation.
+// N.B. This rate limiter implementation is not thread-safe. Do not call Acquire
+// from more than one goroutine.
+func (r *rateLimiter) Acquire() bool {
+	for r.updateBalance() < cost {
+		timer := time.NewTimer(r.calculateWait())
```
vprithvi: What is the benefit of having `Acquire` being interruptible? 
Creating a new timer whenever the balance is less than the cost is quite convoluted - an alternative is to sleep for the calculated wait time.

cmd/ingester/app/consumer/rate_limiter.go
isaachier: Avoiding the use of a magic number throughout. Honestly, I believe 1 is pretty understandable, but we do something very similar in the Go client so I decided to follow that precedent.

cmd/ingester/app/consumer/rate_limiter.go
isaachier: We are considering (at least internally) the idea of dynamically changing this value if/when we need it.

cmd/ingester/app/consumer/consumer.go
vprithvi: I see - IMHO a cleaner way of solving this would be to use a no-op rateLimiter implementation that is picked when the user doesn't require rate limiting

cmd/ingester/app/consumer/rate_limiter.go
vprithvi: I think we should add it when we have a concrete use case - why does `Update` need to be thread safe when `Acquire` isn't?

cmd/ingester/app/consumer/rate_limiter.go
isaachier: Or I could always use `time.AfterFunc`. I'm not sure it makes sense to implement a blocking rate limiter without the ability to cancel/interrupt any blocking calls. Sounds like it might lead to unexpected hangs in an application down the line. Also, this allows us to handle update channels etc. while we sleep.

cmd/ingester/app/consumer/consumer.go
isaachier: Good idea!

cmd/ingester/app/consumer/rate_limiter.go
isaachier: Thread-safe might be a little exaggerated. I'll correct it to be `May be called concurrently with Acquire`.

In broad terms, I'm trying to use Go's concurrency primitives instead of the sync package. The main thinking here is that Go's condition variables have no `WaitUntil` method, so I cannot easily implement an interruptible wait without falling back on a timer. At that point, I might as well use the timer channel instead of the locking.

cmd/ingester/app/consumer/rate_limiter.go
vprithvi: My bad - I misread

cmd/agent/app/httpserver/collector_proxy_test.go
```diff
@@ -22,7 +22,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"github.com/uber/jaeger-lib/metrics"
-	mTestutils "github.com/uber/jaeger-lib/metrics/testutils"
+	"github.com/uber/jaeger-lib/metrics/metricstest"
```
yurishkuro: can we do this in a separate PR?

yurishkuro: also, why do we need to do this at all? The new jaeger-lib is not released as a version afaik.

isaachier: Fair enough. I opened issue #1140 with PR #1141. It would make sense to do this eventually. Meanwhile, I had originally assumed I was using the rate limiter from jaeger-lib, which prompted me to upgrade the library here.

isaachier:
Redoing this in writer instead.

Issue #920, opened 2018-07-08
Make Go client support JAEGER_ENDPOINT env var and use it in hotrod


yurishkuro:
Per https://github.com/jaegertracing/jaeger/pull/915#discussion_r200855019

I think a cleaner approach is to change Go client to respect JAEGER_ENDPOINT env var, similar to Java client, which would allow configuring either UDP or HTTP reporter from env vars only, without adding extra flags to the HotROD binary.

- [x] Go client to respect JAEGER_ENDPOINT
- [ ] Update HotROD to remove extra flags

danehans:
@yurishkuro I would like to work on this issue.

yurishkuro:
@danehans go for it! Thanks!

eundoosong:
Hello, is there any plan to release the next version of go-client including https://github.com/jaegertracing/jaeger-client-go/pull/323 ?

Because I've been working on `Update HotROD to remove extra flags` and almost done.
But the latest version(v2.14.0) of go client doesn't have JAEGER_ENDPOINT.

black-adder:
@eundoosong 2.15 is out

PR #1115, opened 2018-10-12
Support tracer env based initialization in hotrod

eundoosong:
- Support tracer env based initialization in hotrod
- Resolves: #920 

- Remove jAgentHostPort(jaeger-agent.host-port) flag.
- Replace config.Configuration() by config.FromEnv()
- Move metricsFactory.Namespace inside Init()

### DIFF
--- a/examples/hotrod/README.md
+++ b/examples/hotrod/README.md
@@ -49,10 +49,11 @@ go run ./main.go all
 docker run \
   --rm \
   --link jaeger \
+  --env JAEGER_AGENT_HOST=jaeger \
+  --env JAEGER_AGENT_PORT=6831 \
   -p8080-8083:8080-8083 \
   jaegertracing/example-hotrod:1.6 \
-  all \
-  --jaeger-agent.host-port=jaeger:6831
+  all
 ```
 
 Then open http://127.0.0.1:8080
--- a/examples/hotrod/README.md
+++ b/examples/hotrod/README.md
@@ -49,11 +49,10 @@ go run ./main.go all
 docker run \
   --rm \
   --link jaeger \
-  --env JAEGER_AGENT_HOST='jaeger' \
-  --env JAEGER_AGENT_PORT=6831 \
   -p8080-8083:8080-8083 \
-  jaegertracing/example-hotrod:1.8 \
-  all
+  jaegertracing/example-hotrod:1.6 \
+  all \
+  --jaeger-agent.host-port=jaeger:6831
 ```
 
 Then open http://127.0.0.1:8080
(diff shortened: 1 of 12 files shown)


eundoosong:
Looks like the build error is caused by change in glide.lock.
https://travis-ci.org/jaegertracing/jaeger/jobs/440583267
I just did `glide update` and seems the dependancies related to `ingester` were updated as well.
glide.lock is generated from glide.yaml. Is there any way to update go-client only?

glide.lock
```diff
@@ -233,7 +233,7 @@ imports:
 - name: github.com/uber-go/atomic
   version: 8474b86a5a6f79c443ce4b2992817ff32cf208b8
 - name: github.com/uber/jaeger-client-go
-  version: b043381d944715b469fd6b37addfd30145ca1758
+  version: 1a782e2da844727691fef1757c72eb190c2909f0
```
pavolloffay: Can you update the verson in glide.yml?

pavolloffay:
readme needs to be updated too

pavolloffay:
glide update will probably update other deps too. But you didn't add go client bump in glide.yml

eundoosong:
@pavolloffay 
Yes I think readme needs to be updated.
When is example-hotrod docker image tagged? I need to specify the verision(probably 1.8?).

glide.lock
eundoosong: Yes I did, but still error from ingester.

examples/hotrod/README.md
```diff
@@ -49,10 +49,11 @@ go run ./main.go all
 docker run \
   --rm \
   --link jaeger \
+  --env JAEGER_AGENT_HOST=jaeger \
+  --env JAEGER_AGENT_PORT=6831 \
   -p8080-8083:8080-8083 \
   jaegertracing/example-hotrod:1.6 \
```
eundoosong: 1.6 version is old, so it needs to be updated to accept JAEGER_AGENT_XXX env.
Should I remove version tag so that latest is pulled?

pavolloffay: This has to be updated. latest is fine or 1.8

eundoosong: Changed to latest

black-adder:
let's pin sarama and sarama-cluster to 1.16.0 and 2.1.13 respectively, looks like there was a breaking change.

black-adder:
More specifically https://github.com/bsm/sarama-cluster/blame/f7d869db45fd8d74bf62c02a9d17637f945668d0/partitions.go#L23 was added to the interface in the newest version of sarama-cluster

eundoosong:
Pin sarama and sarama-cluster to 1.16.0 and 2.1.13 is done.
It's ready for review.

black-adder:
@eundoosong thanks!

Issue #1054, opened 2018-09-07
Jaeger operational observability


vprithvi:
## Requirement - what kind of business use case are you trying to solve?
Verifying whether all components of Jaeger have been deployed successfully without needing to run special applications, etc. 

I would like confirmation of Jaeger agent and client health, version, reachability and it's ability to send spans. 

## Problem - what in Jaeger blocks you from solving the requirement?
- none

## Proposal - what do you suggest to solve the problem or improve the existing situation?
Have a page in the UI that allows users to see the following:

- Hosts where Jaeger agents are running, their configuration values, version, and the last time the agent interacted with Jaeger Collector
- Service names of services reporting to individual agents, their configuration values, version, and the last time a span was sent from that service
- Effective sampling rate per service

This page is read only,  maintains no history and only shows this information in realtime. 

## Any open questions to address
Determine how to get this data effectively with minimal changes. One approach may be to have Jaeger agents send spans when they start up and shut down.

jpkrohling:
I think the main source of information should be to the collector. A new module (say, "admin") would then query all known collectors and aggregate this information when needed.

Ideally, this admin module would have an extra management endpoint, to get notified when new collectors are added/removed from the cluster.

pavolloffay:
Is it related to https://github.com/jaegertracing/jaeger/issues/789?

yurishkuro:
Isn't there a way to solve this with existing Observability tools instead of building a bespoke solution?

vprithvi:
@yurishkuro What are you suggesting?

yurishkuro:
Expose metrics so that all these signals can be observed via existing tools. Maybe provide base Grafana dashboards.

vprithvi:
I agree certain things can be served by Grafana, but I'm not convinced it is a good solution. 

Let's take the problem of determining which hosts jaeger-agents run on and what versions these agents are on. What metrics would we emit to surface this?

vprithvi:
>These attributes are often provided by metrics collection pipeline. 

I'm not sure - re-reading the ticket, my intent was to determine whether all components are wired up correctly and are connected. 

In this context, I believe the question we are answering is whether there is a jaeger-agent with a compatible version range connected to jaeger-collector, and which host/etc that it is running on. 

I'm opposed to solely relying on metrics because of the following:
[...]

PR #1124, opened 2018-10-17
[agent] Added ping endpoint to agent for connectivity check

### LINKED ISSUE #1007: Provide a positive feedback about client->agent->collector connectivity
Right now, it's hard to find out if a span reported by the client will arrive at the UI at all. To figure out if the connectivity is working, we ask users to get the client to talk to the collector directly, via HTTP.

The agent has a set of endpoints on port 5778 for sampling and baggage managers. The initial proposal would be to add also a "ping" endpoint that clients can call when they are bootstrapped. Clients can then log a message if they can't reach this endpoint.

A similar endpoint can be added to the collector, to be called either by the client or by the agent.

A more elaborate setup would be to send a specially crafted span to the agent and the agent would then provide the client with a positive confirmation that it has seen this span. As the agent is mostly stateless, this approach would cause a bigger impact, so, I think we should start with the simple solution first and move to this one if needed.

annanay25:
Fixes #1007 

- It is hard to know for sure that a trace generated at a client will reach the collector. 

- This PR starts out by adding a ping endpoint to the agent that the client can check with before sending traces. Future commits will be aimed at extending this connectivity check to the collector.

Signed-off-by: Annanay Agarwal <agarwal.ansh@yahoo.co.in>

### DIFF
--- a/cmd/agent/app/httpserver/server.go
+++ b/cmd/agent/app/httpserver/server.go
@@ -40,6 +40,9 @@ func NewHTTPServer(hostPort string, manager ClientConfigManager, mFactory metric
 	mux.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
 		handler.serveSamplingHTTP(w, r, true /* thriftEnums092 */)
 	})
+	mux.HandleFunc("/ping", func(w http.ResponseWriter, r *http.Request) {
+		handler.servePingHTTP(w, r)
+	})
 	mux.HandleFunc("/sampling", func(w http.ResponseWriter, r *http.Request) {
 		handler.serveSamplingHTTP(w, r, false /* thriftEnums092 */)
 	})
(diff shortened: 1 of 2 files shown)


pavolloffay:
We could use healthcheck for this which is already implemented.

jpkrohling:
I think this is probably related to #1054. Have you talked to @vprithvi about it?

jpkrohling:
By the way: while I appreciate your effort in this PR, I believe this is not the best way to implement this. The reason is that clients might be able to do TCP/HTTP connections to the agent, but UDP packets might get dropped. Ideally, an especially crafted span would be generated and *then* the client could make a TCP/HTTP connection to the agent to confirm it received.

Note that this was discussed in the past, which is why I asked if you touched base with @vprithvi before :)

annanay25:
@jpkrohling @pavolloffay Thank you for your reviews. 

> Ideally, an especially crafted span would be generated and then the client could make a TCP/HTTP connection to the agent to confirm it received.  

I thought there was a discussion that since the agent is stateless, it would be bulkier to implement this.  

I will catch up with @vprithvi on this. :)

jpkrohling:
> It would be very inefficient to verify every batch with an additional HTTP request

@isaachier [mentioned on #1007 a good way to solve this](https://github.com/jaegertracing/jaeger/issues/1007#issuecomment-414692802), by marking such spans as "debug" spans and keeping a count per client (perhaps using a credit-based system, to avoid bad clients from abusing this feature).

Issue #797, opened 2018-05-02
Simplify test coverage scripts


vprithvi:
We moved to Go 1.10 in #792, in which `go test` coverpkg has been enhanced to accept multiple packages.

>The go test -coverpkg flag now interprets its argument as a comma-separated list of patterns to match against the dependencies of each test, not as a list of packages to load anew. For example, go test -coverpkg=all is now a meaningful way to run a test with coverage enabled for the test package and all its dependencies. Also, the go test -coverprofile option is now supported when running multiple tests.

This means that we can simplify the logic in [cover.sh](https://github.com/jaegertracing/jaeger/blob/master/scripts/cover.sh) and potentially make the build faster.

yurishkuro:
I did a test run, `go test ./...` ran for 58 second, `make test` ran for 67 seconds. Not a significant difference. To use the combined test run we would need to refactor how code coverage is generated, i.e. remove the packages with `.nocover` from the coverage output (or get rid of `.nocover` completely since we have relaxed 100% coverage rules and can probably tolerate some uncovered packages).

yurishkuro:
Another alternative by @bogdandrutu (https://github.com/open-telemetry/opentelemetry-collector/issues/671#issuecomment-605212324) is to use https://github.com/ory/go-acc, which also allows cross-package coverage to be included.

PankhudiB:
shall I use go-acc and add make task ? or is someone else working on this ?

PankhudiB:
This is the output of go-acc command i just ran on the repo: 

ok  	github.com/jaegertracing/jaeger	0.011s	coverage: [no statements] [no tests to run]
?   	github.com/jaegertracing/jaeger/cmd/agent	[no test files]
ok  	github.com/jaegertracing/jaeger/cmd/agent/app	4.457s	coverage: 4.1
ok  	github.com/jaegertracing/jaeger/cmd/agent/app/configmanager	0.048s	coverage: 1.5
ok  	github.com/jaegertracing/jaeger/cmd/agent/app/configmanager/grpc	0.108s	coverage: 2.7
ok  	github.com/jaegertracing/jaeger/cmd/agent/app/customtransport	0.040s	coverage: 100.0
ok  	github.com/jaegertracing/jaeger/cmd/agent/app/httpserver	0.053s	coverage: 1.3
ok  	github.com/jaegertracing/jaeger/cmd/agent/app/processors	3.874s	coverage: 14.9
[...]

PR #1127, opened 2018-10-18
Use Golang built-in coverage test suites.

ariefrahmansyah:
- https://github.com/jaegertracing/jaeger/issues/797

- Simplify coverage step by using Golang built-in coverage test suites


Makefile
```diff
@@ -1,13 +1,11 @@
...
+TOP_PKGS := $(shell go list ./... | \
 	grep -v \
-		-e ./thrift-gen/... \
-		-e ./swagger-gen/... \
-		-e ./examples/... \
-		-e ./scripts/...\
+		-e examples \
+		-e scripts \
+		-e swagger-gen \
+		-e thrift-gen\
```
yurishkuro: this is not equivalent, glide only returns top-level pkgs
```
$ glide novendor
./swagger-gen/...
./crossdock/...
./cmd/...
./plugin/...
./storage/...
./thrift-gen/...
./examples/...
./model/...
./pkg/...
.
```

ariefrahmansyah: Well, they are returning different result. But once they go to `go test $TOP_PKGS`, the test results are same.
There are two options I have in mind:
1. Keep using glide novendor
2. Rename `TOP_PKGS` to `TEST_PKGS`

yurishkuro: I certainly want to stop using glide for that, I'm just concerned that there are 8 occurrences of `TOP_PKGS` in the Makefile, so we need to ensure that changing what it refers to is still equivalent for all other built steps using it.

ariefrahmansyah: Got it. I understand your concern.

`TOP_PKGS` is used by 4 Go tools: `go list`, `go test`, `go vet`, and `golint`. And 2 other tools: `gosec` and `gosimple`. All of Go tools have the same behavior on handling packages argument. I also read `gosec` and `gosimple` documentation and the way they handle packages argument is same as Go tools.

Reading documentation is not enough, so I created a simple script to compare them. Please check it here: https://gist.github.com/ariefrahmansyah/89cecbd7f933f80a9258c0028c92a2f3

I'm sure that migrating from `glide novendor` to `go list` is fine 🙂 

By the way, another thing to discuss, maybe we can merge `TOP_PKGS` to `ALL_PKGS` 🤔

yurishkuro: > By the way, another thing to discuss, maybe we can merge TOP_PKGS to ALL_PKGS

yes, I think it would make sense, I never liked the split, it was only there because some tools worked differently.

yurishkuro:
@ariefrahmansyah do you want to try again? 

My main concern is how to deal with packages where we explicitly have `.nocover` marker
```
$ find . -name .nocover
./cmd/flags/.nocover
./cmd/agent/app/httpserver/thrift-0.9.2/.nocover
./cmd/agent/.nocover
./pkg/version/.nocover
./pkg/kafka/producer/.nocover
./pkg/cassandra/config/.nocover
./pkg/cassandra/gocql/testutils/.nocover
./pkg/cassandra/gocql/.nocover
./pkg/es/config/.nocover
./pkg/es/.nocover
```

ariefrahmansyah:
@yurishkuro hi, yes i will find a time this week to try this again. thanks

PR #1128, opened 2018-10-18
[Storage] Emit spans for elastic storage backend

### LINKED ISSUE #1043: Enable tracing of ES queries
Similar to https://github.com/jaegertracing/jaeger/pull/1038 it'll be nice if we emit spans for ES span reader

Context is already passed in to ES span reader, all we'd have to do is generate spans.

Could we refactor anything from https://github.com/jaegertracing/jaeger/pull/1038 and use it for ES?

annanay25:
First draft, aims to fix #1043 

- Context is passed, but spans are not emitted for the ES backend.

- Create spans in calls to ES backend.

### DIFF
--- a/plugin/storage/es/spanstore/reader.go
+++ b/plugin/storage/es/spanstore/reader.go
@@ -514,9 +514,6 @@ func (s *SpanReader) buildObjectQuery(field string, k string, v string) elastic.
 }
 
 func logErrorToSpan(span opentracing.Span, err error) {
-	if err == nil {
-		return
-	}
 	ottag.Error.Set(span, true)
 	span.LogFields(otlog.Error(err))
 }
--- a/plugin/storage/es/spanstore/reader.go
+++ b/plugin/storage/es/spanstore/reader.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strconv"
 	"time"
 
 	"github.com/opentracing/opentracing-go"
@@ -136,7 +137,7 @@ func (s *SpanReader) GetTrace(ctx context.Context, traceID model.TraceID) (*mode
 	span, ctx := startSpanForQuery(ctx, "GetTrace", traceIDField)
 	defer span.Finish()
 	currentTime := time.Now()
-	traces, err := s.multiRead(span, []string{traceID.String()}, currentTime.Add(-s.maxSpanAge), currentTime)
+	traces, err := s.multiRead(ctx, []string{traceID.String()}, currentTime.Add(-s.maxSpanAge), currentTime)
 	if err != nil {
 		return nil, err
 	}
@@ -219,6 +220,9 @@ func bucketToStringArray(buckets []*elastic.AggregationBucketKeyItem) ([]string,
 // FindTraces retrieves traces that match the traceQuery
 func (s *SpanReader) FindTraces(ctx context.Context, traceQuery *spanstore.TraceQueryParameters) ([]*model.Trace, error) {
 	span, ctx := startSpanForQuery(ctx, "FindTraces", traceQuery.ServiceName)
+	for k, v := range traceQuery.Tags {
+		span.LogFields(otlog.String("tag.key", k), otlog.String("tag.value", v))
+	}
 	defer span.Finish()
 
 	if err := validateQuery(traceQuery); err != nil {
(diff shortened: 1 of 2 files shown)


plugin/storage/es/spanstore/reader.go
```diff
@@ -209,20 +218,23 @@ func bucketToStringArray(buckets []*elastic.AggregationBucketKeyItem) ([]string,
...
+	uniqueTraceIDs, err := s.findTraceIDs(span, traceQuery)
 	if err != nil {
 		return nil, err
 	}
-	return s.multiRead(uniqueTraceIDs, traceQuery.StartTimeMin, traceQuery.StartTimeMax)
+	return s.multiRead(span, uniqueTraceIDs, traceQuery.StartTimeMin, traceQuery.StartTimeMax)
 }
 
-func (s *SpanReader) multiRead(traceIDs []string, startTime, endTime time.Time) ([]*model.Trace, error) {
+func (s *SpanReader) multiRead(span opentracing.Span, traceIDs []string, startTime, endTime time.Time) ([]*model.Trace, error) {
```
yurishkuro: It is non-idiomatic to pass spans around. Can you pass Context instead?

annanay25: Sure. 

Reason I did it was that I was working with #1038 as a reference. Span was passed to a specific function to log any errors that occurred. 
https://github.com/jaegertracing/jaeger/pull/1038/files#diff-74fbf0ea62bbce4e040d3499e9728171R386

Anyway, since `GetTrace` has just one ES-querying-function, it is OK to treat the span of `GetTrace` as representative for `multiRead`?

yurishkuro: My bad then, I was lazy. We can always extract the span from the context.

Regarding multi-read, if having more instrumentation can be helpful in debugging, then it's better to keep extra data.

annanay25: Okay, sure! Working on this. :)

plugin/storage/es/spanstore/reader.go
```diff
@@ -209,20 +219,32 @@ func bucketToStringArray(buckets []*elastic.AggregationBucketKeyItem) ([]string,
...
-	return s.multiRead(uniqueTraceIDs, traceQuery.StartTimeMin, traceQuery.StartTimeMax)
+	return s.multiRead(ctx, uniqueTraceIDs, traceQuery.StartTimeMin, traceQuery.StartTimeMax)
 }
 
-func (s *SpanReader) multiRead(traceIDs []string, startTime, endTime time.Time) ([]*model.Trace, error) {
+func (s *SpanReader) multiRead(ctx context.Context, traceIDs []string, startTime, endTime time.Time) ([]*model.Trace, error) {
+
+	childSpan, _ := opentracing.StartSpanFromContext(ctx, "multiRead")
+	for k, v := range traceIDs {
+		childSpan.LogFields(otlog.String("tag.num", strconv.Itoa(k)), otlog.String("tag.value", v))
```
yurishkuro: why tag.num? I think you want a single log field `trace_ids` with value being the array of trace IDs.

plugin/storage/es/spanstore/reader.go
```diff
@@ -183,13 +189,17 @@ func (s *SpanReader) indicesForTimeRange(indexName string, startTime time.Time,
...
+	span, ctx := startSpanForQuery(ctx, "GetServices", serviceNameField)
+	defer span.Finish()
 	currentTime := time.Now()
 	jaegerIndices := s.indicesForTimeRange(s.serviceIndexPrefix, currentTime.Add(-s.maxSpanAge), currentTime)
 	return s.serviceOperationStorage.getServices(jaegerIndices)
 }
 
 // GetOperations returns all operations for a specific service traced by Jaeger
 func (s *SpanReader) GetOperations(ctx context.Context, service string) ([]string, error) {
+	span, ctx := startSpanForQuery(ctx, "GetOperations", operationNameField)
```
yurishkuro: I don't think this makes much sense in this form. In case of Cassandra storage, there was actually a CQL query that was being logger, but here you're just using constants that are already implied by the operation name. I think you can simply use `opentracing.StartSpanFromContext(ctx, "operation")`.

Since ES doesn't really have a "query", what might be useful instead is logging the index names. But that's optional.

yurishkuro: Basically the spans should be capturing parameters that may be helpful in debugging.

plugin/storage/es/spanstore/reader.go
annanay25: Right, will change this.

plugin/storage/es/spanstore/reader.go
annanay25: Yes, sorry for the confusion, will clear this out. I didn't realise that `startSpanForQuery` was a wrapper you made for the particular use case until much later. I thought this was the norm.

jpkrohling:
I'm restarting the CI for this PR, but I suspect the failure might actually be related to this change.

annanay25:
@jpkrohling the CI test has passed, could you review the changes?

yurishkuro:
pleas ensure full test coverage

annanay25:
Patch updated. @yurishkuro

yurishkuro:
Thank you, @annanay25 !  

If you are using Jaeger in your organization, please consider commenting on https://github.com/jaegertracing/jaeger/issues/207.

annanay25:
@yurishkuro thanks! I'll look into the survey soon.

PR #1136, opened 2018-10-23
Log jaeger-collector tchannel port

### LINKED ISSUE #1071: Print all open ports

Every binary should print all open ports on startup. This makes it easy to find the port and connect. The log message should contain info about the port - e.g. what it is used for.

mindaugasrukas:
Signed-off-by: Mindaugas Rukas <momomg@gmail.com>

Related to #1071

* Log jaeger-collector tchannel port
* Update jaeger-collector HTTP server log message


pavolloffay:
Would you like to also add it to all-in-one main?

yurishkuro:
all-in-one can be done in another PR

Issue #678, opened 2018-02-05
Using Client Certificate for Elastic Search Authentication


reynaldiwijaya:
Is there any plan to accommodate using client certificate for authenticating jaeger to ES? Our elastic search deployment is using search guard which only enable client certificate now, and we would like to keep it that way. ES is deployed in Kubernetes v1.9, and we would like to deploy Jaeger there as well.

If there are none, do you have any suggestion of workaround for this problem?

yurishkuro:
if the Go driver supports it, we can add the options similar to how they were added for Cassandra:
```
      --cassandra.tls                                        Enable TLS
      --cassandra.tls.ca string                              Path to TLS CA file
      --cassandra.tls.cert string                            Path to TLS certificate file
      --cassandra.tls.key string                             Path to TLS key file
      --cassandra.tls.server-name string                     Override the TLS server name
      --cassandra.tls.verify-host                            Enable (or disable) host key verification (default true)
```

reynaldiwijaya:
oh yeah thanks for update. Should we let this issue opened (in case the client updates it)? Please feel free to close it as well

Sneko:
Hi @reynaldiwijaya! How did you manage that in a proper way please 😃 ?

Sneko:
@yurishkuro I just saw that you are using the olivere/elastic client in Go.

I just checked their docs and found it's possible:
https://github.com/olivere/elastic/issues/636
https://github.com/olivere/elastic/issues/738

It would be really helpful using ES cluster through https 😀 

Do you think it could be on the roadmap? I'm not used to Golang but if I could help by testing... it would be a pleasure 😃

sneko:
Hi @yurishkuro any news about having this on your roadmap?

Thanks :)

isaachier:
Honestly, seeing as we don't use ES internally, don't get too excited about seeing this on a roadmap any time soon.

sneko:
@isaachier oh ok that's too bad 😢 

Anyway, thanks for this tool 😀  !

yurishkuro:
If the driver supports it, the change is fairy trivial, doesn't have to be done by official maintainers.

reynaldiwijaya:
@sneko hi im using cassandra at the end, works like a charm 👍

PR #1139, opened 2018-10-25
Fixed "Using Client Certificate for Elastic Search Authentication"

clyang82:
Signed-off-by: Chun Lin Yang <clyang@cn.ibm.com>

Resolves https://github.com/jaegertracing/jaeger/issues/678
- Go driver support it with HTTPClient. so follow this way to support client certificate for ES


pkg/es/config/config.go
```diff
@@ -188,10 +202,65 @@ func (c *Configuration) GetTagDotReplacement() string {
...
+	if err != nil {
+		return nil, err
+	}
+	certificates := x509.NewCertPool()
+	certificates.AppendCertsFromPEM(caCert)
+	return certificates, nil
+}
+
+// LoadPrivateKeyFrom is used to load the private certificate and key for TLS
+func (c *Configuration) LoadPrivateKeyFrom() (*tls.Certificate, error) {
```
jpkrohling: Perhaps remove the "From" from the function name?

pkg/es/config/config.go
```diff
@@ -188,10 +202,65 @@ func (c *Configuration) GetTagDotReplacement() string {
...
-func (c *Configuration) GetConfigs() []elastic.ClientOptionFunc {
+func (c *Configuration) GetConfigs(logger *zap.Logger) []elastic.ClientOptionFunc {
 	options := make([]elastic.ClientOptionFunc, 3)
+	if c.TLS.Enabled {
+		tlsConfig, err := c.CreateTLSConfig(logger)
+		if err != nil {
+			return nil
+		}
+		httpClient := &http.Client{
+			Timeout: c.Timeout,
```
jpkrohling: Seems like this is something extra, but I think it's ok to have it here.

pkg/es/config/config.go
```diff
@@ -188,10 +202,65 @@ func (c *Configuration) GetTagDotReplacement() string {
...
+			Transport: &http.Transport{
+				TLSClientConfig: tlsConfig,
+			},
+		}
+		options[0] = elastic.SetHttpClient(httpClient)
+		options[1] = elastic.SetURL(c.Servers...)
+		options[2] = elastic.SetSniff(c.Sniffer)
+		return options
+	}
```
yurishkuro: can we instead use `options = append(options, ...)` and have a single return? I don't like that this duplicates SetURL/SetSniff, no reason to fork the logic this way.

pkg/es/config/config.go
```diff
@@ -188,10 +202,65 @@ func (c *Configuration) GetTagDotReplacement() string {
...
 	options[0] = elastic.SetURL(c.Servers...)
 	options[1] = elastic.SetBasicAuth(c.Username, c.Password)
 	options[2] = elastic.SetSniff(c.Sniffer)
 	return options
 }
+
+// CreateTLSConfig creates TLS Configuration to connect with ES Cluster.
+func (c *Configuration) CreateTLSConfig(logger *zap.Logger) (*tls.Config, error) {
```
yurishkuro: I think all these new functions can be private. Also, can they be on the TLS config object instead of the top-level Configuration type?

pkg/es/config/config.go
```diff
@@ -188,10 +202,65 @@ func (c *Configuration) GetTagDotReplacement() string {
...
+	rootCerts, err := c.LoadCertificatesFrom()
+	if err != nil {
+		logger.Fatal("Couldn't load root certificate", zap.Error(err))
```
yurishkuro: what's the reason to log fatals instead of returning errors?

yurishkuro:
Does this address the same issue as #1036 ?

pkg/es/config/config.go
clyang82: If cannot load the certificate or key, the pod cannot be started normally. that is why using Fatal instead of Error.

pkg/es/config/config.go
clyang82: Thanks for your comments. I will update

pkg/es/config/config.go
clyang82: Yes. Agree with your point. I will make changes accordingly.

clyang82:
@yurishkuro it is similar with #1036. both need pass httpclient to elasticsearch client. but AWS Identity and Access Management (IAM) is a web service that helps you securely control access to AWS resources. that need to call AWS api to do authentication, that means need to pass a aws http client instance to elasticsearch client.

clyang82:
@jpkrohling & @yurishkuro Could you please help to merge my pull request if you have no further comments? Thanks.

pkg/es/config/config.go
pavolloffay: This is only being applied when TLS is enabled. Can you make it available for every case?

pavolloffay:
@clyang82 do you want to fix it in this PR or in a follow-upPR?

pkg/es/config/config.go
```diff
@@ -188,10 +202,62 @@ func (c *Configuration) GetTagDotReplacement() string {
 }
 
 // GetConfigs wraps the configs to feed to the ElasticSearch client init
-func (c *Configuration) GetConfigs() []elastic.ClientOptionFunc {
-	options := make([]elastic.ClientOptionFunc, 3)
-	options[0] = elastic.SetURL(c.Servers...)
-	options[1] = elastic.SetBasicAuth(c.Username, c.Password)
-	options[2] = elastic.SetSniff(c.Sniffer)
+func (c *Configuration) GetConfigs(logger *zap.Logger) []elastic.ClientOptionFunc {
+	options := make([]elastic.ClientOptionFunc, 0)
```
yurishkuro: nit: `var options []elastic.ClientOptionFunc`

pkg/es/config/config.go
```diff
@@ -188,10 +202,62 @@ func (c *Configuration) GetTagDotReplacement() string {
...
+	options = append(options, elastic.SetURL(c.Servers...), elastic.SetSniff(c.Sniffer))
```
yurishkuro: can combine
```
options := []elastic.ClientOptionFunc{ elastic.SetURL(c.Servers...), elastic.SetSniff(c.Sniffer) }
```

pkg/es/config/config.go
```diff
@@ -188,10 +202,62 @@ func (c *Configuration) GetTagDotReplacement() string {
...
 	return options
 }
+
+// createTLSConfig creates TLS Configuration to connect with ES Cluster.
+func (tlsConfig *TLSConfig) createTLSConfig(logger *zap.Logger) (*tls.Config, error) {
+	rootCerts, err := tlsConfig.loadCertificate()
+	if err != nil {
+		logger.Fatal("Couldn't load root certificate", zap.Error(err))
+	}
+	if len(tlsConfig.CertPath) > 0 && len(tlsConfig.KeyPath) > 0 {
```
yurishkuro: why is this check needed?  If it fails you return nil, so shouldn't it be an error to set `c.TLS.Enabled` and not provide the other 3 params?

pkg/es/config/config.go
```diff
@@ -188,10 +202,62 @@ func (c *Configuration) GetTagDotReplacement() string {
...
+	if err != nil {
+		return nil, err
+	}
+	certificates := x509.NewCertPool()
+	certificates.AppendCertsFromPEM(caCert)
+	return certificates, nil
+}
+
+// loadPrivateKeyFrom is used to load the private certificate and key for TLS
+func (tlsConfig *TLSConfig) loadPrivateKeyFrom() (*tls.Certificate, error) {
```
yurishkuro: nit: remove "From" suffix

yurishkuro:
> it is similar with #1036. both need pass httpclient to elasticsearch client. but AWS Identity and Access Management (IAM) is a web service that helps you securely control access to AWS resources. that need to call AWS api to do authentication, that means need to pass a aws http client instance to elasticsearch client.

Is it necessary to use IAM with AWS ES? Can your approach be used and have some other script retrieve the credentials from IAM? If we can use your approach then I think it's better than #1036 since it does not introduce additional dependencies aside from std lib.

wesleyk:
> > it is similar with #1036. both need pass httpclient to elasticsearch client. but AWS Identity and Access Management (IAM) is a web service that helps you securely control access to AWS resources. that need to call AWS api to do authentication, that means need to pass a aws http client instance to elasticsearch client.
> 
> Is it necessary to use IAM with AWS ES? Can your approach be used and have some other script retrieve the credentials from IAM? If we can use your approach then I think it's better than #1036 since it does not introduce additional dependencies aside from std lib.

These changes are a bit orthogonal from using IAM credentials IMO, as in the IAM case you need to provide a custom http client.

That being said, the client cert approach for authentication may be a reasonable tradeoff over IAM. Looking more into that now.

pkg/es/config/config.go
clyang82: @pavolloffay Yes. This is only being applied for httpclient which is being used by TLS case. I cannot find where to add for every case. we use `BulkProcessorService` which allows to easily process bulk requests. There is no interface to set timeout. Could you point me out more specifically? Thanks.

pkg/es/config/config.go
clyang82: Thanks

pkg/es/config/config.go
clyang82: OK

pkg/es/config/config.go
pavolloffay: So the driver does not use the client when tls is disabled?

Something like this would not work? 
```
// GetConfigs wraps the configs to feed to the ElasticSearch client init
func (c *Configuration) GetConfigs(logger *zap.Logger) []elastic.ClientOptionFunc {
	options := []elastic.ClientOptionFunc{elastic.SetURL(c.Servers...), elastic.SetSniff(c.Sniffer)}
	httpClient := &http.Client{
		Timeout: c.Timeout,
	}
	if c.TLS.Enabled {
		ctlsConfig, err := c.TLS.createTLSConfig(logger)
		if err != nil {
			return nil
		}
		httpClient.Transport = &http.Transport{
			TLSClientConfig:ctlsConfig,
		}
	} else {
	}
	options = append(options, elastic.SetHttpClient(httpClient))
	return options
}
```

yurishkuro:
Relevant blog post: http://labs.vistarmedia.com/2018/10/31/deploying-jaeger-with-cloudformation-via-bazel.html

pavolloffay:
@jpkrohling We could link the blog post in the newsletter.

clyang82:
@pavolloffay the latest code changes have passed my testing. Thanks.

pavolloffay:
If this approach is orthogonal to IAM we could go forward and merge.

pkg/es/config/config.go
```diff
@@ -64,7 +78,7 @@ func (c *Configuration) NewClient(logger *zap.Logger, metricsFactory metrics.Fac
 	if len(c.Servers) < 1 {
 		return nil, errors.New("No servers specified")
 	}
-	rawClient, err := elastic.NewClient(c.GetConfigs()...)
+	rawClient, err := elastic.NewClient(c.GetConfigs(logger)...)
```
yurishkuro: can we do this without logging?
```go
options, err := c.getConfigOptions() // note name change, no logger
if err != nil {
    return nil, err
}
rawClient, err := elastic.NewClient(options...)
```

pkg/es/config/config.go
```diff
@@ -188,10 +202,61 @@ func (c *Configuration) GetTagDotReplacement() string {
...
+	}
+	clientPrivateKey, err := tlsConfig.loadPrivateKey()
+	if err != nil {
+		logger.Fatal("Couldn't setup client authentication", zap.Error(err))
+		return nil, err
+	}
+	return &tls.Config{
+		RootCAs:      rootCerts,
+		Certificates: []tls.Certificate{*clientPrivateKey},
+	}, err
```
yurishkuro: `}, nil`

clyang82:
@jpkrohling & @yurishkuro Could you please help to merge my pull request? Thanks.

plugin/storage/es/options.go
```diff
@@ -119,6 +124,10 @@ func addFlags(flagSet *flag.FlagSet, nsConfig *namespaceConfig) {
 		nsConfig.namespace+suffixServerURLs,
 		nsConfig.servers,
 		"The comma-separated list of ElasticSearch servers, must be full url i.e. http://localhost:9200")
+	flagSet.Duration(
+		nsConfig.namespace+suffixTimeout,
+		nsConfig.Timeout,
```
yurishkuro: I don't see where a default value for this is set. I assume it's not becoming a required flag, correct? Maybe expand in the help string what happens if not set.

yurishkuro:
Looks good to me, thanks! Had just one nit, but it can be done in another PR.

plugin/storage/es/options.go
clyang82: @yurishkuro Yes. It is not a required flag. the default value is 0. A Timeout of zero means no timeout. so how about change the help string from `Timeout used for queries` to `Timeout used for queries. A Timeout of zero means no timeout.`

yurishkuro: +1

PR #1141, opened 2018-10-25
Use new jaeger-lib metricstest

### LINKED ISSUE #1140: jaeger-lib metricstest
Upgrade jaeger-lib.

jaeger-lib made a breaking change in its package structure. It uses `metricstest` instead of `testutils` for metric mocks, etc.

Replace references with to `testutils` with `metricstest`.

isaachier:
- Resolves #1140. 

- Replace old jaeger-lib API with new jaeger-lib API.

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -17,9 +17,7 @@ import:
   subpackages:
   - transport
 - package: github.com/uber/jaeger-lib
-  repo: https://github.com/isaachier/jaeger-lib
-  version: 4aa908b951cc7516c61cf001ddca1118d02d243e
-  #version: ^1.5.0
+  version: e0a812eccf5a8ac943db957a823bed43b346d654
 - package: github.com/uber/tchannel-go
   version: v1.1.0
   subpackages:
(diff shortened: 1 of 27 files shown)


glide.yaml
```diff
@@ -17,7 +17,9 @@ import:
   subpackages:
   - transport
 - package: github.com/uber/jaeger-lib
-  version: ^1.5.0
+  repo: https://github.com/isaachier/jaeger-lib
```
black-adder: `/isaachier/jaeger-lib`?

isaachier: My bad. Forgot I was using a personal fork.

black-adder:
CI failed with `Failed to set version on github.com/uber/jaeger-lib to 4aa908b951cc7516c61cf001ddca1118d02d243e: Unable to update checked out version: exit status 128`

isaachier:
Interesting... This isn't pressing. I'll fix later.

yurishkuro:
technically we could upgrade to pinned version, but I would rather wait till 2.0 release of jaeger-lib (https://github.com/jaegertracing/jaeger-lib/issues/47)

PR #1147, opened 2018-10-26
Add rate limiter option to Cassandra writer (WIP)

### LINKED ISSUE #1100: Add a rate limiter to jaeger-ingester

isaachier:
- Resolves #1100.

- Add rate limiter to Cassandra writer so we do not overload Cassandra with write requests.

### DIFF
--- a/plugin/storage/cassandra/factory.go
+++ b/plugin/storage/cassandra/factory.go
@@ -28,6 +28,7 @@ import (
 	"github.com/jaegertracing/jaeger/storage"
 	"github.com/jaegertracing/jaeger/storage/dependencystore"
 	"github.com/jaegertracing/jaeger/storage/spanstore"
+	"github.com/jaegertracing/jaeger/storage/spanstore/ratelimit"
 )
 
 const (
@@ -101,7 +102,13 @@ func (f *Factory) CreateSpanReader() (spanstore.Reader, error) {
 
 // CreateSpanWriter implements storage.Factory
 func (f *Factory) CreateSpanWriter() (spanstore.Writer, error) {
-	return cSpanStore.NewSpanWriter(f.primarySession, f.Options.SpanStoreWriteCacheTTL, f.primaryMetricsFactory, f.logger), nil
+	writer := cSpanStore.NewSpanWriter(f.primarySession, f.Options.SpanStoreWriteCacheTTL, f.primaryMetricsFactory, f.logger)
+
+	if f.Options.writesPerSecond == 0 {
+		return writer, nil
+	}
+
+	return ratelimit.NewRateLimitedWriter(writer, f.Options.writesPerSecond)
 }
 
 // CreateDependencyReader implements storage.Factory
--- a/plugin/storage/cassandra/factory_test.go
+++ b/plugin/storage/cassandra/factory_test.go
@@ -27,6 +27,7 @@ import (
 	"github.com/jaegertracing/jaeger/pkg/config"
 	"github.com/jaegertracing/jaeger/pkg/testutils"
 	"github.com/jaegertracing/jaeger/storage"
+	"github.com/jaegertracing/jaeger/storage/spanstore/ratelimit"
 )
 
 var _ storage.Factory = new(Factory)
@@ -86,4 +87,24 @@ func TestCassandraFactory(t *testing.T) {
 
 	_, err = f.CreateArchiveSpanWriter()
 	assert.NoError(t, err)
+}
+
+func TestCassandraFactoryWriterRateLimit(t *testing.T) {
+	logger, _ := testutils.NewLogger()
+	f := NewFactory()
+	v, command := config.Viperize(f.AddFlags)
+	command.ParseFlags([]string{
+		"--cassandra.writes-per-second=10",
+	})
+	f.InitFromViper(v)
+
+	// Use mock to avoid errors in unit test. See details in previous test case.
+	f.primaryConfig = &mockSessionBuilder{}
+	f.archiveConfig = nil
+	assert.NoError(t, f.Initialize(metrics.NullFactory, logger))
+
+	writer, err := f.CreateSpanWriter()
+	assert.NoError(t, err)
+	dummyRateLimitedWriter, _ := ratelimit.NewRateLimitedWriter(nil, 1)
+	assert.IsType(t, dummyRateLimitedWriter, writer)
 }
--- a/plugin/storage/cassandra/options.go
+++ b/plugin/storage/cassandra/options.go
@@ -48,6 +48,7 @@ const (
 
 	// common storage settings
 	suffixSpanStoreWriteCacheTTL = ".span-store-write-cache-ttl"
+	suffixWritesPerSecond        = ".writes-per-second"
 )
 
 // Options contains various type of Cassandra configs and provides the ability
@@ -57,6 +58,7 @@ type Options struct {
 	primary                *namespaceConfig
 	others                 map[string]*namespaceConfig
 	SpanStoreWriteCacheTTL time.Duration
+	writesPerSecond        int
 }
 
 // the Servers field in config.Configuration is a list, which we cannot represent with flags.
@@ -111,6 +113,9 @@ func (opt *Options) AddFlags(flagSet *flag.FlagSet) {
 	flagSet.Duration(opt.primary.namespace+suffixSpanStoreWriteCacheTTL,
 		opt.SpanStoreWriteCacheTTL,
 		"The duration to wait before rewriting an existing service or operation name")
+	flagSet.Int(opt.primary.namespace+suffixWritesPerSecond,
+		opt.writesPerSecond,
+		"Optional upper limit on the number of span writes per second")
 }
 
 func addFlags(flagSet *flag.FlagSet, nsConfig *namespaceConfig) {
@@ -201,6 +206,7 @@ func (opt *Options) InitFromViper(v *viper.Viper) {
 		cfg.initFromViper(v)
 	}
 	opt.SpanStoreWriteCacheTTL = v.GetDuration(opt.primary.namespace + suffixSpanStoreWriteCacheTTL)
+	opt.writesPerSecond = v.GetInt(opt.primary.namespace + suffixWritesPerSecond)
 }
 
 func (cfg *namespaceConfig) initFromViper(v *viper.Viper) {
--- a/plugin/storage/cassandra/options_test.go
+++ b/plugin/storage/cassandra/options_test.go
@@ -58,6 +58,7 @@ func TestOptionsWithFlags(t *testing.T) {
 		"--cas.consistency=ONE",
 		"--cas.proto-version=3",
 		"--cas.socket-keep-alive=42s",
+		"--cas.writes-per-second=10",
 		// enable aux with a couple overrides
 		"--cas-aux.enabled=true",
 		"--cas-aux.keyspace=jaeger-archive",
@@ -69,6 +70,7 @@ func TestOptionsWithFlags(t *testing.T) {
 	assert.Equal(t, "jaeger", primary.Keyspace)
 	assert.Equal(t, []string{"1.1.1.1", "2.2.2.2"}, primary.Servers)
 	assert.Equal(t, "ONE", primary.Consistency)
+	assert.Equal(t, 10, opts.writesPerSecond)
 
 	aux := opts.Get("cas-aux")
 	require.NotNil(t, aux)
--- a/storage/spanstore/ratelimit/rate_limit.go
+++ b/storage/spanstore/ratelimit/rate_limit.go
@@ -1,0 +1,51 @@
+// Copyright (c) 2018 Uber Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ratelimit
+
+import (
+	"errors"
+
+	rlImpl "go.uber.org/ratelimit"
+
+	"github.com/jaegertracing/jaeger/model"
+	"github.com/jaegertracing/jaeger/storage/spanstore"
+)
+
+var errInvalidRate = errors.New("rate must be a positive integer")
+
+type rateLimitedWriter struct {
+	writer  spanstore.Writer
+	limiter rlImpl.Limiter
+}
+
+// NewRateLimitedWriter decorates a Writer with a rate limiter in order to limit
+// the number of writes per second.
+func NewRateLimitedWriter(writer spanstore.Writer, rate int) (spanstore.Writer, error) {
+	if rate <= 0 {
+		return nil, errInvalidRate
+	}
+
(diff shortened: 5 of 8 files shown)


isaachier:
> * What are the advantages of scoping this to Cassandra as opposed to creating a Writer decorator that can be used for all writers?

I had considered that as well. I think it makes sense. The only issue is that config tends to be database specific, so the main change will probably still be in the Cassandra config (unless you'd like me to add to all database implementations immediately).

> * What are the disadvantages of using https://github.com/uber-go/ratelimit in lieu of writing our own implementation?

I'd love to use uber-go/ratelimit, but they only support integer rates, not floating point rates (not sure why). If I can get a PR through to them, that would be the best solution. Then again, it is probably a breaking change for their API.

vprithvi:
>I'd love to use uber-go/ratelimit, but they only support integer rates, not floating point rates

What makes floating point rates indispensable for us?

yurishkuro:
I don't envision us having rate per ingester of less than one msg per second, so integer rates seem fine

isaachier:
OK cool. I can remove the rate limiter too.

isaachier:
Actually, ratelimit is broken. Here is the error:

`vendor/github.com/uber-go/ratelimit/ratelimit.go:27:2: use of internal package github.com/jaegertracing/jaeger/vendor/go.uber.org/ratelimit/internal/clock not allowed`

Looks like the package is not really maintained based on the history and lack of versions.

isaachier:
Can someone please rerun test. ES unit test failed.

black-adder:
> ratelimit is broken

I find that hard to believe given that it's being used by multiple other teams at Uber. Looks like the issue is more a vendoring issue.

storage/spanstore/decorator/rate_limit.go
```diff
@@ -0,0 +1,83 @@
...
+type rateLimitedWriter struct {
+	w spanstore.Writer
+	l utils.RateLimiter
+}
+
+// WriteSpan wraps the write span method of the inner writer, but invokes the
+// rate limiter before each write.
+func (r *rateLimitedWriter) WriteSpan(s *model.Span) error {
+	const cost = 1
+	for ok, waitTime := r.l.CheckCredit(cost); !ok; ok, waitTime = r.l.CheckCredit(cost) {
```
black-adder: Won't 
```
for ok, waitTime := r.l.CheckCredit(cost); !ok {
   time.Sleep(waitTime)
```
work? Not sure why the second call is needed

plugin/storage/cassandra/options.go
```diff
@@ -48,6 +48,8 @@ const (
 
 	// common storage settings
 	suffixSpanStoreWriteCacheTTL = ".span-store-write-cache-ttl"
+	suffixWritesPerSecond        = ".writer-rate-limit.operations-per-second"
```
black-adder: I'm ok with this for now but I think these configs should be storage type agnostic (ie ES can also use the rate limiter but we shouldn't have to pollute ES storage configs with `.writer-rate-limit.operations-per-second` and `.writer-rate-limit.max-burst`)

isaachier: I wish the configs generally were less storage specific. I'm not sure if we share any config at the moment.

storage/spanstore/decorator/rate_limit.go
isaachier: Well let's imagine Java/C++/wtvr other language:

```
for (bool ok = r.l.CheckCredit(cost); !ok; ok = r.l.CheckCredit(cost)) {
    // ...
}
```

First statement is a declaration/initialization statement. Last statement repeats each time the loop is repeated. Not 100% sure, but I imagine Go works the same way.

isaachier: Ya, this explains it: https://tour.golang.org/flowcontrol/1.

plugin/storage/cassandra/options.go
```diff
@@ -70,6 +72,16 @@ type namespaceConfig struct {
 	Enabled   bool
 }
 
+// rateLimitConfig defines common arguments to a rate limiter.
+type rateLimitConfig struct {
+	operationsPerSecond float64
+	maxBurst            float64
```
vprithvi: Is this parameter required?

isaachier: Meant to delete this. Thanks for the reminder.

plugin/storage/cassandra/factory.go
```diff
@@ -101,7 +102,17 @@ func (f *Factory) CreateSpanReader() (spanstore.Reader, error) {
 
 // CreateSpanWriter implements storage.Factory
 func (f *Factory) CreateSpanWriter() (spanstore.Writer, error) {
-	return cSpanStore.NewSpanWriter(f.primarySession, f.Options.SpanStoreWriteCacheTTL, f.primaryMetricsFactory, f.logger), nil
+	writer := cSpanStore.NewSpanWriter(f.primarySession, f.Options.SpanStoreWriteCacheTTL, f.primaryMetricsFactory, f.logger)
+
+	// If rate is zero, then just use regular writer.
+	if f.Options.writesPerSecond == 0 {
```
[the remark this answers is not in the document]
isaachier: Good idea. I wasn't sure if a passthrough would hurt performance due to the indirection.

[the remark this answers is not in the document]
isaachier: Also, @black-adder if it makes sense to use `0` to mean no rate limiting. Semantically, `0` represents always blocking.

storage/spanstore/decorator/rate_limit_test.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+func TestRateLimitedWriterInvalidWritesPerSecond(t *testing.T) {
+	writer := &mockWriter{}
+	decoratedWriter, err := NewRateLimitedWriter(writer, -1)
+	require.Error(t, err)
+	require.Nil(t, decoratedWriter)
+}
+
+func TestRateLimitedWriterUnlimitedWritesPerSecond(t *testing.T) {
```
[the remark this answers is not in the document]
isaachier: Ya. I didn't want to use time alone seeing how flakey that can be. Will try mocking soon.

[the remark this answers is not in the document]
isaachier: Not a huge fan of type testing, but tried it out in latest revision.

plugin/storage/cassandra/factory.go
```diff
@@ -28,6 +28,7 @@ import (
 	"github.com/jaegertracing/jaeger/storage"
 	"github.com/jaegertracing/jaeger/storage/dependencystore"
 	"github.com/jaegertracing/jaeger/storage/spanstore"
+	"github.com/jaegertracing/jaeger/storage/spanstore/decorator"
```
yurishkuro: why the "decorator" package? Metrics is also a decorator, it's in its own package. I suggest `spanstore/ratelimited` instead

plugin/storage/cassandra/factory.go
```diff
@@ -101,7 +102,12 @@ func (f *Factory) CreateSpanReader() (spanstore.Reader, error) {
 
 // CreateSpanWriter implements storage.Factory
 func (f *Factory) CreateSpanWriter() (spanstore.Writer, error) {
-	return cSpanStore.NewSpanWriter(f.primarySession, f.Options.SpanStoreWriteCacheTTL, f.primaryMetricsFactory, f.logger), nil
+	writer := cSpanStore.NewSpanWriter(f.primarySession, f.Options.SpanStoreWriteCacheTTL, f.primaryMetricsFactory, f.logger)
+
+	return decorator.NewRateLimitedWriter(
+		writer,
+		f.Options.writesPerSecond,
```
yurishkuro: if f.Options.writesPerSecond == 0, do we need to wrap?

storage/spanstore/decorator/rate_limit.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+	return &rateLimitedWriter{
+		w: writer,
+		l: rateLimiter,
+	}, nil
+}
+
+type rateLimitedWriter struct {
+	w spanstore.Writer
+	l ratelimit.Limiter
```
yurishkuro: Please use full field names. One letter names are ok for local variables when there is context.

plugin/storage/cassandra/factory.go
isaachier: Lol see what @p.r suggested earlier. We are wrapping with noop rate limiter in that case.

yurishkuro: if we have an if statement, I'd rather have it here and not add two more layers of indirection. Then in your rate limited writer, return an error if rate < 1.

isaachier: Oh man. I want to wait on @vprithvi's response because this is literally what I was wondering above.

> Good idea. I wasn't sure if a passthrough would hurt performance due to the indirection.

https://github.com/jaegertracing/jaeger/pull/1147#discussion_r229048135

vprithvi: I have a preference towards having this logic be in the rate limiter because it means that it doesn't need to be in every storage implementation. However, it's not a strong preference.

yurishkuro: That's a separate concern though, the config and application of rate limiter is currently in the cassandra factory, but we could easily pull it up to the higher level (meta) factory (I just don't think we need to do it right now).  This keeps the rate limiter simpler - you can only instantiate it with a positive rate.

isaachier: I'm wondering when we care about performance of indirection, etc. But I'll go back to the original if statement here.

isaachier:
ES integration test again!

jpkrohling:
Job restarted

plugin/storage/cassandra/factory_test.go
```diff
@@ -87,3 +87,21 @@ func TestCassandraFactory(t *testing.T) {
...
+	})
+	f.InitFromViper(v)
+
+	// Use mock to avoid errors in unit test. See details in previous test case.
+	f.primaryConfig = &mockSessionBuilder{}
+	f.archiveConfig = nil
+	assert.NoError(t, f.Initialize(metrics.NullFactory, logger))
+
+	_, err := f.CreateSpanWriter()
+	assert.NoError(t, err)
```
yurishkuro: This test will pass regardless of the rate limiting setting. Maybe simply test for type, since any real test would involve sleeping and would take a long time to run.

plugin/storage/cassandra/options.go
```diff
@@ -111,6 +113,9 @@ func (opt *Options) AddFlags(flagSet *flag.FlagSet) {
 	flagSet.Duration(opt.primary.namespace+suffixSpanStoreWriteCacheTTL,
 		opt.SpanStoreWriteCacheTTL,
 		"The duration to wait before rewriting an existing service or operation name")
+	flagSet.Int(opt.primary.namespace+suffixWritesPerSecond,
+		opt.writesPerSecond,
+		"The number of writes per second using rate limiter")
```
yurishkuro: Optional upper limit on the number of span writes per second.

plugin/storage/cassandra/options.go
```diff
@@ -57,6 +58,7 @@ type Options struct {
 	primary                *namespaceConfig
 	others                 map[string]*namespaceConfig
 	SpanStoreWriteCacheTTL time.Duration
+	writesPerSecond        int
```
yurishkuro: Shouldn't this be public? And qualified with `SpanStore` prefix.

plugin/storage/cassandra/options_test.go
```diff
@@ -82,4 +83,6 @@ func TestOptionsWithFlags(t *testing.T) {
 	assert.Equal(t, "", aux.Consistency, "aux storage does not inherit consistency from primary")
 	assert.Equal(t, 3, aux.ProtoVersion)
 	assert.Equal(t, 42*time.Second, aux.SocketKeepAlive)
+
+	assert.Equal(t, 10, opts.writesPerSecond)
```
yurishkuro: move to L73

storage/spanstore/ratelimit/rate_limit.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"errors"
+
+	rlImpl "go.uber.org/ratelimit"
+
+	"github.com/jaegertracing/jaeger/model"
+	"github.com/jaegertracing/jaeger/storage/spanstore"
+)
+
+var errInvalidRate = errors.New("rate must be a non-zero positive integer")
```
yurishkuro: positive already implies non-zero

storage/spanstore/ratelimit/rate_limit.go
```diff
@@ -16,36 +16,54 @@ package ratelimit
...
-type rateLimitedWriter struct {
-	writer  spanstore.Writer
-	limiter rlImpl.Limiter
-}
-
 // WriteSpan wraps the write span method of the inner writer, but invokes the
 // rate limiter before each write.
 func (r *rateLimitedWriter) WriteSpan(s *model.Span) error {
-	r.limiter.Take()
+	const threshold = 100 * time.Microsecond
```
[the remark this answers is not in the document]
isaachier: I happen to agree wholeheartedly. Team suggested I add a metric with number of delayed spans. The current rate limiter API makes this essentially impossible. IMO the real issue is that the rate limiter API has limited use in its current form. I believe it would have made more sense to return the duration slept as opposed to the timestamp of the wake time.

storage/spanstore/ratelimit/rate_limit.go
```diff
@@ -16,36 +16,54 @@ package ratelimit
...
+func NewRateLimitedWriter(
+	writer spanstore.Writer,
+	rate int,
+	metricsFactory metrics.Factory,
+) (spanstore.Writer, error) {
 	if rate <= 0 {
 		return nil, errInvalidRate
 	}
 
+	m := metricsFactory.Namespace("rate-limited-writer", nil)
```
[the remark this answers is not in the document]
isaachier: I copied the format from the `storage/spanstore/metrics/writer_metrics.go` file, but I cannot guarantee that file is correct.

isaachier:
I just added metrics here while testing. I understand the concern regarding scope creep, and will move it to another PR soon.

plugin/storage/cassandra/factory_test.go
isaachier: Fair point.

plugin/storage/cassandra/options.go
isaachier: That's better.

plugin/storage/cassandra/options.go
isaachier: IDK why SpanStoreWriteCacheTTL is public either. There is no serialization here. I think it might be a workaround to pass options to another package that need to read `SpanStoreWriteCacheTTL`, but doesn't make much sense given that passing the `int` is just as effective.

yurishkuro: how are we using this internally? SpanStoreWriteCacheTTL is externally settable option.

isaachier: Both of these are initialized by command-line flags, so neither should be public. I did a code search and could not find a reference to SpanStoreWriteCacheTTL (not even in our internal Uber repo). Seems to be vestigial from some point where we needed access from our internal repo into this part of the OSS.

yurishkuro: We don't use cli flags internally. If we need to set these two fields how would we do it without them being public? Did you try your branch without our internal main?

isaachier:
```
$ git grep -l SpanStoreWriteCacheTTL
plugin/storage/cassandra/factory.go
plugin/storage/cassandra/options.go
```

yurishkuro:
I think this is solved differently by #1353

PR #1153, opened 2018-10-30
Allow jaeger client configuration in query service

### LINKED ISSUE #1044: Allow configuration of jaeger client in the query service
The jaeger query service allows for tracing the query service itself. In the all-in-one container, the [default](https://github.com/jaegertracing/jaeger/blob/master/cmd/query/main.go#L83) initialization of the client works since the agent/collector reside on localhost. However, when the query service is deployed individually, the collector and agent are not guaranteed to be running locally or maybe running on non-default ports depending on the setup.

The query service exposes no flags or environment variables to configure the tracer in the query service.

The golang tracer allows configuration via [environment variables](https://github.com/jaegertracing/jaeger-client-go/blob/252d853b2a4f3cfc98bb89c3a8bb3d3aa50ed4d6/config/config_env.go#L53). I think all we'd have to do is to replace [New](https://github.com/jaegertracing/jaeger/blob/master/cmd/query/main.go#L89) with [FromEnv](https://github.com/jaegertracing/jaeger-client-go/blob/252d853b2a4f3cfc98bb89c3a8bb3d3aa50ed4d6/config/config_env.go#L53).

subashiniganesh:
Resolves #1044 

The query service shall now be configured using the environmental variables to configure the tracer.

### DIFF
--- a/cmd/query/main.go
+++ b/cmd/query/main.go
@@ -83,13 +83,12 @@ func main() {
 				logger.Fatal("Cannot create metrics factory.", zap.Error(err))
 			}
 
-			tracer, closer, err := jaegerClientConfig.Configuration{
-				Sampler: &jaegerClientConfig.SamplerConfig{
-					Type:  "probabilistic",
-					Param: 1.0,
-				},
-				RPCMetrics: true,
-			}.New(
+			cfg, err := jaegerClientConfig.FromEnv()
+			if err != nil {
+				logger.Fatal("Cannot parse Jaeger env vars", zap.Error(err))
+			}
+
+			tracer, closer, err := cfg.New(
 				"jaeger-query",
 				jaegerClientConfig.Metrics(baseFactory.Namespace("client", nil)),
 				jaegerClientConfig.Logger(jaegerClientZapLog.NewLogger(logger)),


cmd/query/main.go
```diff
@@ -83,13 +83,12 @@ func main() {
...
 			}
 
-			tracer, closer, err := jaegerClientConfig.Configuration{
-				Sampler: &jaegerClientConfig.SamplerConfig{
-					Type:  "probabilistic",
-					Param: 1.0,
-				},
-				RPCMetrics: true,
-			}.New(
+			cfg, err := jaegerClientConfig.FromEnv()
```
yurishkuro: I think this would effectively disable tracing in the query service since it's typically a low QPS service and the default 0.001 probability will kick in. I would rather have a default configuration like above that can be overwritten via env variables.

subashiniganesh: Sure. That really makes sense. So, I shall check if the above fields are nil and if so will set the default configuration.

subashiniganesh: Does this apply for both the Sampler and RPCMetrics?

yurishkuro: I don't think doing it here is the best solution. I would rather allow Configuration class to be created with default values, and then allow something like InitFromEnv() which will take care of overriding configuration parameters as needed. cc @jpkrohling

jpkrohling: Agree with @yurishkuro: ideally, we should keep the same behavior for the end user as we have now, just with the added possibility of customizing the behavior when and where needed.

subashiniganesh: Thanks for the comment @jpkrohling . If my understanding is correct, within the FromEnv(), there should be IntFromEnv() which gets the default configuration set that is then overridden by the environmental variables.
May I implement this flow?

jpkrohling: @yurishkuro do you mean something like this?

```golang
func InitFromEnv(Configuration defaults) (*Configuration, error) {
...
}
```

yurishkuro: my proposal requires changes to the Jaeger Go client

jpkrohling: @subashiniganesh I also got a bit confused by @yurishkuro's comment, but it doesn't make sense (to me) to have `NewDefaultConfiguration()` in the Jaeger Go Client. So, the changes required are:

1. Add `func NewDefaultConfiguration() Configuration {}` to an appropriate place. I would suggest `pkg/config`, but take a look and see if there's a better place for it. This code would basically do what the [current `cmd/query/main.go` is doing](https://github.com/jaegertracing/jaeger/blob/master/cmd/query/main.go#L86-L96) regarding the client config.
1. Add `func (c *Configuration) InitFromEnv() error {}` to the [Jaeger Go Client](https://github.com/jaegertracing/jaeger-client-go), under the `config` package, which will override the config options from `c` with the ones from the env vars[1]. Take a look at the `config_env.go` source file: you can probably reuse a lot of that, avoiding duplication.
1. Change the `cmd/query/main.go` like you did in this PR, but calling `config.NewDefaultConfiguration().InitFromEnv()` instead of `jaegerClientConfig.FromEnv()`

yurishkuro: NewDefaultConfig function, hopefully, should not be necessary since the empty struct should work as the default config.

jpkrohling: > the empty struct should work as the default config.

This is what's confusing me :) The default config for the query component is not the same as the default config for the Go client (the query config sets the sampling to const 1).

yurishkuro: Then three steps, default config from client (just a struct), override sampling, then override from env. The first two steps are already done in query svc, just need the client change.

annanay25:
Hi @subashiniganesh, Do you have time to continue working on this PR? :)

jpkrohling:
I'm closing this, as there hasn't been movement for almost two years now.

Issue #1126, opened 2018-10-18
Add a metric for total number of partitions held by an individual ingester


vprithvi:
## Requirement - what kind of business use case are you trying to solve?
Determine the split of partitions across ingesters.

## Problem - what in Jaeger blocks you from solving the requirement?
None.

## Proposal - what do you suggest to solve the problem or improve the existing situation?
Add a counter to measure this.

PR #1154, opened 2018-10-30
Add a metric for number of partitions held

vprithvi:
ref #1126


yurishkuro:
@vprithvi saying "ref #1126" does not auto-close the ticket on merge, please use Fixes or Resolves.

PR #1158, opened 2018-10-31
Add wildcardSearch ability to elasticsearch (#1148)

### LINKED ISSUE #1148: Allow wildcard service search for elasticsearch storage

We have a need to collate traces across different services that share a specific tag value. The traces are initiated by separate HTTP requests and therefore appear as separate traces under separate services. We would like to search by a tag across all services with all matches displaying regardless of service name. 

The jaeger-ui and the jaeger-query modules require service as a parameter. 

The jaeger-ui should include an 'All' or 'Any' option in the services dropdown. 

image

The jaeger-query module should handle a request containing the value `service=all` or `service=any` by creating a new wildcard query. The `elastic.NewWildcardQuery` is supported by the `bool.Must` query builder. 

I have a working example with tests, but would like to discuss prior to opening a pull request.

ryansos:
Resolves #1148 

Signed-off-by: Ryan Dwyer <rdwyer@rapidsos.com>

- allows users to search by wildcard against the serviceName in elasticsearch

- adds a config variable WildcardSearch (bool) which defaults to false
- this will require changes to the jaeger-ui repo (pr to follow)

### DIFF
--- a/cmd/query/app/query_parser.go
+++ b/cmd/query/app/query_parser.go
@@ -29,8 +29,7 @@ import (
 )
 
 const (
-	defaultQueryLimit     = 100
-	defaultAnyServiceName = "*"
+	defaultQueryLimit = 100
 
 	operationParam      = "operation"
 	tagParam            = "tag"
@@ -48,7 +47,6 @@ const (
 var (
 	errCannotQueryTagAndDuration = fmt.Errorf("Cannot query for tags when '%s' is specified", minDurationParam)
 	errMaxDurationGreaterThanMin = fmt.Errorf("'%s' should be greater than '%s'", maxDurationParam, minDurationParam)
-	errAnyServiceNameInvalid     = fmt.Errorf("Any Service Name parameter must be '%s'", defaultAnyServiceName)
 
 	// ErrServiceParameterRequired occurs when no service name is defined
 	ErrServiceParameterRequired = fmt.Errorf("Parameter '%s' is required", serviceParam)
@@ -70,6 +68,7 @@ type traceQueryParameters struct {
 //     query ::= param | param '&' query
 //     param ::= service | operation | limit | start | end | minDuration | maxDuration | tag | tags
 //     service ::= 'service=' strValue
+//     anyServiceName ::= 'anyServiceName=' strValue
 //     operation ::= 'operation=' strValue
 //     limit ::= 'limit=' intValue
 //     start ::= 'start=' intValue in unix microseconds
@@ -82,7 +81,12 @@ type traceQueryParameters struct {
 //     tags :== 'tags=' jsonMap
 func (p *queryParser) parse(r *http.Request) (*traceQueryParameters, error) {
 	service := r.FormValue(serviceParam)
-	anyServiceName := r.FormValue(anyServiceNameParam)
+
+	anyServiceName, err := p.parseBool(anyServiceNameParam, r)
+	if err != nil {
+		return nil, err
+	}
+
 	operation := r.FormValue(operationParam)
 
 	startTime, err := p.parseTime(startTimeParam, r)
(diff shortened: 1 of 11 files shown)


pavolloffay:
Yes it is look at `es-integration-test.sh`

```
docker run --rm -it -p 9200:9200 -e "http.host=0.0.0.0" -e "transport.host=127.0.0.1" docker.elastic.co/elasticsearch/elasticsearch:5.4.0
STORAGE=elasticsearch make storage-integration-test
```

pavolloffay:
Wondering whether the ES failure is related to this PR or #1108

ryansos:
> Wondering whether the ES failure is related to this PR or #1108

interesting - it appears to have passed.

pavolloffay:
yah, I think the failure is related to #1108

cmd/query/app/query_parser.go
```diff
@@ -29,23 +29,26 @@ import (
...
-	tagsParam        = "tags"
-	startTimeParam   = "start"
-	limitParam       = "limit"
-	minDurationParam = "minDuration"
-	maxDurationParam = "maxDuration"
-	serviceParam     = "service"
-	endTimeParam     = "end"
-	prettyPrintParam = "prettyPrint"
+	defaultQueryLimit     = 100
+	defaultAnyServiceName = "*"
```
yurishkuro: I think this makes more sense as a boolean flag, similar to prettyPrint.
```go
	if prettyPrint := r.FormValue(prettyPrintParam); prettyPrint != "" && prettyPrint != "false" {}
```

cmd/query/app/query_parser.go
```diff
@@ -79,6 +82,7 @@ type traceQueryParameters struct {
 //     tags :== 'tags=' jsonMap
 func (p *queryParser) parse(r *http.Request) (*traceQueryParameters, error) {
 	service := r.FormValue(serviceParam)
+	anyServiceName := r.FormValue(anyServiceNameParam)
```
yurishkuro: please add to the comments of this method

storage/spanstore/interface.go
```diff
@@ -42,12 +42,13 @@ type Reader interface {
...
-	ServiceName   string
-	OperationName string
-	Tags          map[string]string
-	StartTimeMin  time.Time
-	StartTimeMax  time.Time
-	DurationMin   time.Duration
-	DurationMax   time.Duration
-	NumTraces     int
+	ServiceName    string
+	AnyServiceName string
```
yurishkuro: please make it Boolean

pkg/es/config/config.go
```diff
@@ -57,6 +58,7 @@ type ClientBuilder interface {
 	GetTagsFilePath() string
 	GetAllTagsAsFields() bool
 	GetTagDotReplacement() string
+	GetWildcardSearch() bool
```
yurishkuro: s/GetWildcardSearch/IsWildcardSearch/  (GetAllTagsAsFields was badly named)

plugin/storage/es/options.go
```diff
@@ -41,6 +41,7 @@ const (
 	suffixTagsAsFieldsAll   = suffixTagsAsFields + ".all"
 	suffixTagsFile          = suffixTagsAsFields + ".config-file"
 	suffixTagDeDotChar      = suffixTagsAsFields + ".dot-replacement"
+	suffixWildcardSearch    = ".wildcard-search"
```
yurishkuro: this name seems overly broad. Are you really supporting wildcard search for everything, or just for the service name?

plugin/storage/es/spanstore/reader.go
```diff
@@ -421,12 +425,17 @@ func (s *SpanReader) buildFindTraceIDsQuery(traceQuery *spanstore.TraceQueryPara
 	startTimeQuery := s.buildStartTimeQuery(traceQuery.StartTimeMin, traceQuery.StartTimeMax)
 	boolQuery.Must(startTimeQuery)
 
-	//add process.serviceName query
-	if traceQuery.ServiceName != "" {
+	//add process.serviceName query, only if AnyServiceName is blank
+	if traceQuery.ServiceName != "" && !traceQuery.AnyServiceName {
```
yurishkuro: I would prefer to return an error when both `traceQuery.ServiceName` and `traceQuery.AnyServiceName` are non-empty

plugin/storage/es/spanstore/reader.go
```diff
@@ -459,6 +468,10 @@ func (s *SpanReader) buildServiceNameQuery(serviceName string) elastic.Query {
 	return elastic.NewMatchQuery(serviceNameField, serviceName)
 }
 
+func (s *SpanReader) buildAnyServiceNameQuery() elastic.Query {
+	return elastic.NewWildcardQuery(serviceNameField, defaultAnyServiceName)
```
yurishkuro: I suggest removing `defaultAnyServiceName` constant and using `*` explicitly here.

plugin/storage/factory_config.go
```diff
@@ -36,6 +36,7 @@ type FactoryConfig struct {
 	SpanWriterTypes         []string
 	SpanReaderType          string
 	DependenciesStorageType string
+	WildcardSearch          bool
```
yurishkuro: is this needed at the generic factory level?

jpkrohling:
@ryansos, are you still interested in this one?

PR #1178, opened 2018-11-13
Fix metrics handler registration in agent

### LINKED ISSUE #1177: Metrics handler in agent 1.8 is not registered

pavolloffay:
Resolves #1177 

I have removed metrics registration from agent. now following the same approach as in query, collector and allinone.


pavolloffay:
I think we have to release 1.8.1 once this is merged

PR #1182, opened 2018-11-14
Rename tcollector proxy metric in agent

pavolloffay:
Related to https://github.com/jaegertracing/jaeger/issues/773#issuecomment-436602664

This is not something required but it seems wrong to have tag `tcollector-proxy` when the proxy can use any protocol.

### DIFF
--- a/cmd/agent/app/httpserver/server.go
+++ b/cmd/agent/app/httpserver/server.go
@@ -71,7 +71,7 @@ type httpHandler struct {
 		BadRequest metrics.Counter `metric:"http-server.errors" tags:"status=4xx,source=all"`
 
 		// Number of collector proxy failures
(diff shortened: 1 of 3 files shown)


cmd/agent/app/httpserver/server.go
```diff
@@ -71,7 +71,7 @@ type httpHandler struct {
 		BadRequest metrics.Counter `metric:"http-server.errors" tags:"status=4xx,source=all"`
 
 		// Number of collector proxy failures
-		TCollectorProxyFailures metrics.Counter `metric:"http-server.errors" tags:"status=5xx,source=tcollector-proxy"`
+		TCollectorProxyFailures metrics.Counter `metric:"http-server.errors" tags:"status=5xx,source=collector-proxy"`
```
yurishkuro: s/TCollectorProxyFailures/CollectorProxyFailures/ as well

PR #1185, opened 2018-11-15
Remove unused sanitizers

pavolloffay:
I am not sure if these sanitizers should be used in handlers or can be removed. There are not used at the moment.


yurishkuro:
We use both of these in our internal build. The service name sanitizer is kind of useless without a source of valid service names (we connect it to service registry). The UTF8 one is more useful, we created it because Cassandra could crash when some incorrectly encoded UTF8 strings are placed into span tags (or even service/operation names), however it's not clear if it should really be enabled by default (e.g. it depends on data quality and storage backend sensitivity).

pavolloffay:
Thank for the explanation I am closing this then.

Issue #1191, opened 2018-11-16
Something is broken in crossdock


yurishkuro:
The client builds started to fail where they previously succeeded, e.g. this is the rebuild of master branch of Python client, it was green previously: https://travis-ci.org/jaegertracing/jaeger-client-python/jobs/449920101

This may be related to gRPC changes if we started defaulting to different ports, or because we changed the CLI params.

c @pavolloffay

pavolloffay:
I have restarted the build and it failed again... I am not sure if it's bc of changed cli 54/55 tests passed.

pavolloffay:
java CI also failed, funny that it worked locally

pavolloffay:
The second build in java passed. Is there a way that something is broken in python client?

isaachier:
Seems to be timeout and/or some networking bug. Let me check.

pavolloffay:
getting xdock logs in python is also broken, it seems like bash syntax issue.

isaachier:
OK I see we use alpine in several places. I have had significant DNS bug issues with alpine and crossdock. This could potentially explain issues with elasticsearch as well. See here for more details: https://github.com/gliderlabs/docker-alpine/issues/255.

pavolloffay:
What I see localy the address of `go` service gets translated some public IP.

isaachier:
That might be fine. I've had issues where crossdock image "x" looks for crossdock image "y" by doing some invalid lookup on DNS for a literal "y." It should not be doing that for a local crossdock connection.

yurishkuro:
I fixed crossdock logging in the Python client and I'm seeing this:
```
test_driver_1       | 2018-11-22T15:14:28.307Z	INFO	services/query.go:74	Retrieved trace from query	{"body": "{\"data\":[],\"total\":0,\"limit\":0,\"offset\":0,\"errors\":null}", "url": "http://jaeger-query:16686/api/traces?end=1542899668000000&operation=60bed31f1cb9e6c6&service=crossdock-python&tag=33afd37941439089%3A10b92d0169a39144"}
jaeger-query_1      | {"level":"info","ts":1542899631.441772,"caller":"healthcheck/handler.go:99","msg":"Health Check server started","http-port":16687,"status":"unavailable"}
jaeger-query_1      | 2018/11/22 15:13:51 gocql: unable to dial control conn 172.18.0.3: dial tcp 172.18.0.3:9042: connect: connection refused
[...]

PR #1196, opened 2018-11-16
Fix ES tests by dropping alpine for debian (WIP)

isaachier:

- Might fix part of #1191.

- Use debian instead of alpine.


isaachier:
Other usages of alpine seem to be just file copying. Might have to look at Jaeger client repos.

pavolloffay:
I don't think that we run es-index-cleaner docker in tests.

isaachier:
Could be. Still want to experiment with this a bit more.

PR #1198, opened 2018-11-19
Reorganize config manager packages in agent

### LINKED ISSUE #1194: Refactor httpserver package in agent
The `httpserver` package should contain manager with metrics and two subpackages: tchannel and grpc.

pavolloffay:
Resolves #1194 


pavolloffay:
@jpkrohling would you like to have a look?

cmd/agent/app/reporter/grpc/collector_proxy_test.go
```diff
@@ -80,3 +81,15 @@ func TestMultipleCollectors(t *testing.T) {
...
 	require.Nil(t, proxy.Close())
 }
+
+func initializeGRPCTestServer(t *testing.T, beforeServe func(server *grpc.Server)) (*grpc.Server, net.Addr) {
+	server := grpc.NewServer()
+	lis, err := net.Listen("tcp", "localhost:0")
+	require.NoError(t, err)
+	beforeServe(server)
+	go func() {
+		err := server.Serve(lis)
```
jpkrohling: You can just use `server.Serve(list)` directly as a parameter on the next line: `require.NoError(t, server.Serve(list))`

cmd/agent/app/reporter/grpc/collector_proxy.go
```diff
@@ -25,6 +25,7 @@ import (
 	"google.golang.org/grpc/resolver/manual"
 
 	"github.com/jaegertracing/jaeger/cmd/agent/app/httpserver"
+	mGrpc "github.com/jaegertracing/jaeger/cmd/agent/app/httpserver/grpc"
```
jpkrohling: Just curious, what does `m` stands for? Why not `a` (agent)? (no need to change, really just curious)

cmd/agent/app/httpserver/tchannel/manager.go
```diff
@@ -12,14 +12,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package httpserver
+package tchannel
```
jpkrohling: I'm not a big fan of the new package structure: without looking at the code, I would expect that the packages `cmd/agent/app/httpserver/[tchannel|grpc]` would contain code that ties something about gRPC or TChannel with the HTTP server that the agent starts, when in reality, all these packages have are specific implementations of the `ClientConfigManager` interface.

Perhaps the managers deserve their own packages? `cmd/agent/app/clientconfigmanager/[tchannel|grpc]`? Perhaps they should even be outside of the `cmd` package.

cmd/agent/app/reporter/grpc/collector_proxy.go
pavolloffay: managerGrpc

cmd/agent/app/httpserver/tchannel/manager.go
pavolloffay: The manager interface is used only in httpserver package Therefore it was originally colocated there I am fine scoping it under httpserver.

> Perhaps they should even be outside of the cmd package.

This applies also to reporter and probably other subpackages. It would make sense to move them  if they could be reused in other cmd. This is specific to the agent only.

pavolloffay:
@jpkrohling I have moved  the configmanager to a separate package. Is there anything more?

jpkrohling:
> @jpkrohling I have moved the configmanager to a separate package. Is there anything more?

No, that was it. Looks great :-)

PR #1204, opened 2018-11-21
Migrate query cache to open source

### LINKED ISSUE #1203: Implement storage cache for query service

Currently, the query service always hits Cassandra for data. We can easily cache some data in memory, assuming it is not volatile (e.g. service names that change infrequently).

The current architecture has no caching.

Implement a basic in-memory cache for static data. @black-adder and @yurishkuro have already implemented this in the internal Uber Jaeger repository.

isaachier:
- Resolves #1203.  

- Migrates @black-adder's and @yurishkuro's changes from Uber internal repository to open source.

### DIFF
--- a/cmd/query/app/flags.go
+++ b/cmd/query/app/flags.go
@@ -16,15 +16,17 @@ package app
 
 import (
 	"flag"
+	"time"
 
 	"github.com/spf13/viper"
 )
 
 const (
-	queryPort        = "query.port"
-	queryBasePath    = "query.base-path"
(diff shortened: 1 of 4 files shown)


yurishkuro:
this is not a general purpose cache, it's only for service names and operation names.

isaachier:
I mention that briefly, but the API doesn't specify much more than that it is a wrapper to cache data. BTW I don't even think it caches operation names right now. Just caches service names.

yurishkuro:
since there's no ask for it in oss, I would drop it.

PR #1207, opened 2018-11-22
WIP - Fix crossdock by waiting for Cassandra when needed

### LINKED ISSUE #1191: Something is broken in crossdock

isaachier:

- Hopefully fixes #1191.

- Increase timeout to 3 minutes.
- Test port with netcat instead of using cqlsh immediately.


docker-compose/jaeger-docker-compose.yml
```diff
@@ -3,23 +3,27 @@ version: '2'
 services:
     jaeger-collector:
       image: jaegertracing/jaeger-collector
-      command: ["--cassandra.keyspace=jaeger_v1_dc1", "--cassandra.servers=cassandra", "--collector.zipkin.http-port=9411"]
+      command: ["/bin/wait-for-cassandra.sh", "/go/bin/collector-linux", "--cassandra.keyspace=jaeger_v1_dc1", "--cassandra.servers=cassandra", "--collector.zipkin.http-port=9411"]
```
yurishkuro: > `"--cassandra.keyspace=jaeger_v1_dc1"

I realize this was there before, but it's a Bad Idea (tm). If this is billed as a reusable docker-compose file, it should be using env vars for such things.

plugin/storage/cassandra/schema/docker.sh
```diff
@@ -1,33 +0,0 @@
-#!/bin/bash
-#
-# This script is used in the Docker image jaegertracing/jaeger-cassandra-schema
-# that allows installing Jaeger keyspace and schema without installing cqlsh.
```
yurishkuro: this comment is critical. I would prefer not to delete this file, and I don't think it's necessary to do given my objections above about combining C server with C-schema image. The latter is a public API of Jaeger release.

docker-compose/jaeger-docker-compose.yml
```diff
@@ -3,26 +3,30 @@ version: '2'
...
-      command: ["--cassandra.keyspace=jaeger_v1_dc1", "--cassandra.servers=cassandra", "--collector.zipkin.http-port=9411"]
+      entrypoint: ["/bin/bash", "/bin/wait-for-cassandra.sh", "/go/bin/collector-linux", "--cassandra.keyspace=jaeger_v1_dc1", "--cassandra.servers=cassandra", "--collector.zipkin.http-port=9411"]
       ports:
         - "14269"
         - "14268:14268"
         - "14267"
         - "14250"
         - "9411:9411"
+      volumes:
+        - ./scripts/wait-for-cassandra.sh:/bin/wait-for-cassandra.sh
```
yurishkuro: I wonder how this would work for people reusing this docker-compose file elsewhere, since `scripts/wait-for-cassandra.sh` is not part of the images.

@pavolloffay @jpkrohling do we know who uses this docker-compose file?

docker-compose/service-base/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM debian:jessie
+RUN apt-get update
+RUN apt-get install -y bash ca-certificates netcat python-pip
+RUN pip install cqlsh
```
yurishkuro: -1 - main Jaeger images have nothing to do with cqlsh

yurishkuro:
I tagged this as WIP and am going to stop reviewing. There are many different changes going on here, fine if you want to experiment, but for official PRs let's agree what you're trying to achieve (the whole thing started with "increase timeout"), whether we want such changes (e.g. making the schema image actually start Cassandra - 👎 ), and then do them separately.

plugin/storage/cassandra/schema/wait-for-cassandra.sh
```diff
@@ -0,0 +1,48 @@
...
+		echo "Cassandra is still not up at ${CQLSH_HOST}. Waiting 1 second."
+		sleep 1s
+		((total_wait++))
+	done
+
+	while (( total_wait < CASSANDRA_WAIT_TIMEOUT ))
+	do
+	  if ${CQLSH} "${CQLSH_SSL}" -e "describe keyspaces"; then
+		break
+		fi
```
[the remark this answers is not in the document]
isaachier: I might have mixed tabs and spaces or something. Let me fix this.

plugin/storage/cassandra/Dockerfile
```diff
@@ -1,6 +1,7 @@
...
-COPY schema/* /cassandra-schema/
-
-ENV CQLSH_HOST=cassandra
-ENTRYPOINT ["/cassandra-schema/docker.sh"]
+ENV DATACENTER=dc1 KEYSPACE=jaeger_v1_dc1 MODE=prod
+COPY schema/create.sh /
+COPY schema/v001.cql.tmpl /
+COPY schema/entrypoint-wrapper.sh /
+ENTRYPOINT ["/entrypoint-wrapper.sh"]
+CMD ["cassandra", "-f"]
```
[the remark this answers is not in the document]
isaachier: The reason I did this was to avoid extra complexity Cassandra image. I might be able to do this in an entirely separate image (cassandra-schema), so I'll revisit that idea.

plugin/storage/cassandra/Dockerfile
```diff
@@ -1,6 +1,7 @@
 FROM cassandra:3.11
-
-COPY schema/* /cassandra-schema/
-
-ENV CQLSH_HOST=cassandra
-ENTRYPOINT ["/cassandra-schema/docker.sh"]
+ENV DATACENTER=dc1 KEYSPACE=jaeger_v1_dc1 MODE=prod
```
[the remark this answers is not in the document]
isaachier: Good point. I should just leave them empty and allow users to override as needed.

plugin/storage/cassandra/schema/entrypoint-wrapper.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+#
+# Based on https://stackoverflow.com/a/46037377/1930331.
+
+function run_scripts() {
+	until echo "DROP KEYSPACE IF EXISTS ${KEYSPACE};" | cqlsh localhost; do
```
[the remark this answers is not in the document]
isaachier: Correct. This is only for the cassandra-schema image upon initialization. But I'll use `describe ...` as I do in `wait-for-cassandra.sh`.

docker-compose/jaeger-docker-compose.yml
isaachier: Understood. IDK how reusable this compose setup is other than for testing. I guess I understand it could be a basic setup for standalone distribution.

plugin/storage/cassandra/schema/docker.sh
isaachier: Got it, will fix as stated above.

docker-compose/jaeger-docker-compose.yml
isaachier: I was wondering about that as well. I know the clients use this file as part of their crossdock builds. Might have to embed the script into my new `jaegerrtacing/jaeger-service-base` image.

isaachier:
Waiting for https://github.com/jaegertracing/jaeger-client-go/pull/349 then can submit latest changes and see if combination of newest client with these fixes resolves the issue.

isaachier:
I think there are agent failures due to gRPC. See this crossdock log:

```
{"level":"error","ts":1543327192.891047,"caller":"grpc/reporter.go:70","msg":"Could not send spans over gRPC","error":"rpc error: code = Unavailable desc = all SubConns are in TransientFailure, latest connection error: connection error: desc = \"transport: Error while dialing dial tcp 172.27.0.10:14250: connect: connection refused\"","stacktrace":"github.com/jaegertracing/jaeger/cmd/agent/app/reporter/grpc.(*Reporter).send\n\t/home/ihier/proj/go/src/github.com/jaegertracing/jaeger/cmd/agent/app/reporter/grpc/reporter.go:70\ngithub.com/jaegertracing/jaeger/cmd/agent/app/reporter/grpc.(*Reporter).EmitBatch\n\t/home/ihier/proj/go/src/github.com/jaegertracing/jaeger/cmd/agent/app/reporter/grpc/reporter.go:50\ngithub.com/jaegertracing/jaeger/cmd/agent/app/reporter.(*MetricsReporter).EmitBatch\n\t/home/ihier/proj/go/src/github.com/jaegertracing/jaeger/cmd/agent/app/reporter/metrics.go:77\ngithub.com/jaegertracing/jaeger/thrift-gen/agent.(*agentProcessorEmitBatch).Process\n\t/home/ihier/proj/go/src/github.com/jaegertracing/jaeger/thrift-gen/agent/agent.go:197\ngithub.com/jaegertracing/jaeger/thrift-gen/agent.(*AgentProcessor).Process\n\t/home/ihier/proj/go/src/github.com/jaegertracing/jaeger/thrift-gen/agent/agent.go:152\ngithub.com/jaegertracing/jaeger/cmd/agent/app/processors.(*ThriftProcessor).processBuffer\n\t/home/ihier/proj/go/src/github.com/jaegertracing/jaeger/cmd/agent/app/processors/thrift_processor.go:115"}
```

@pavolloffay any ideas?

yurishkuro:
That was my suspicion when I opened the ticket, since this started happening after grpc endpoint got merged, but the crossdock logs we're not showing.

Do we need to declare the new port somewhere?

isaachier:
Well this should solve crossdock flakiness overall. Probably need another fix for the gRPC stuff. I'm not 100% sure what the issue is. `connection refused` could be that the collector is not up yet. But exposing the port is a good idea.

yurishkuro:
> `connection refused` could be that the collector is not up yet. 

the compose file says to restart on failures, so that should only be a transient issue. But if collector is down because the storage is not ready, then 😢

isaachier:
It seems to be primarily to be a race condition between starting the server and when it is ready for connections. I've had this issue with gRPC in the past. Essentially, the setup looks like this:

```
server := ...
go func() {
    server.serve()
}
// supposedly serving requests at this point
```

The problem is that even a wait group does not help if the serve function blocks indefinitely.

```
```
var wg sync.WaitGroup
server := ...
wg.Add(1)
go func() {
    wg.Done()  // Cannot wait after serve call or will end up blocking forever.
    server.serve()
}
wg.Wait()
// supposedly serving requests at this point
```

That doesn't provide a much better guarantee IMO.

**Comment by yurishkuro** [comment]
I am not sure why that would be a problem - eventually it should still start, and the agent should keep restarting until it is able to connect.

Does grpc server object provide a status method? if so, we could integrate it with the health checker in main.

**Comment by isaachier** [comment]
You are correct about retrying. The problem is the agent only succeeds after attempting to write a trace to the collector, and subsequently drops the trace. The issue relates to the `TRANSIENT_FAILURE` status in gRPC request. We should potentially check for that status and retry the request instead of dropping it. I'm doing more research and will keep you posted.

**Comment by isaachier** [comment]
Got it! Now the only question is if I add this to every gRPC call.

File: crossdock/docker-compose.yml [hunk 9]
```diff
@@ -3,6 +3,7 @@ version: '2'
 services:
     crossdock:
         image: crossdock/crossdock
+        dns_search: .
```
isaachier: I have attempted this change in #1029. All crossdock examples show this `dns_search` line. The reasoning is unclear. Alpine has/had DNS issues that might explain this. It definitely helped me locally when crossdock failed to connect to `go:8080`. Here is the commit from crossdock related to this line: https://github.com/crossdock/crossdock/pull/64. Please let me know if you are strongly opposed to adding this.

yurishkuro: what does it do?

isaachier:
For the record, I consistently see this in crossdock too: `WARN  18:39:19 Unable to lock JVM memory (ENOMEM). This can result in part of the JVM being swapped out, especially with mmapped I/O enabled. Increase RLIMIT_MEMLOCK or run Cassandra as root.`

crossdock/docker-compose.yml
isaachier: From this golang issue (https://github.com/golang/go/issues/15419), it keeps Go from trying to resolve local container names to external URLs. Here is a brief quote from that issue:

> The way how pure Go DNS resolver resolves names caused an error in my program. When I set "ndots:" to 0 and resolved "mysql", I wanted to resolve "mysql." first, but instead, pure Go DNS resolver resolved "mysql.searchdomain." first. As it happened, the domain name"mysql.searchdomain." was valid in my DNS server, so I got a wrong IP of mysql.

yurishkuro: thanks - sure, let's keep it. I am still not clear where this PR is going, since it has too many things (and the shared base image with Cassandra included is a non-starter).

isaachier: Well the PR overall is to add a true "wait for Cassandra" to the crossdock tests. For months, I have found crossdock to be flakey. Now gRPC made it worse. I can split the fixes out into separate pull requests, but I believe fixing these issues are worthwhile.

Regarding the use of a base image, I'm not sure why we didn't do this in the first place. It just simplifies building two nearly identical images (jaeger-collector and jaeger-query) from a standard source. The reason for using Cassandra is that it is the only reliable way to guarantee Cassandra is ready for operation.

An alternative approach would be using a memory store like we've discussed for a while, but that hasn't happened yet.

isaachier:
Closing in favor of a solution to #1221.

crossdock/docker-compose.yml
yurishkuro: * I am fine with "wait for cassandra" approach for crossdock
  * longer term we should avoid using Cassandra for crossdock in the first place (#1221)
* I am **not** fine with including Cassandra dependency into the default (production) images
* I am fine with using a shared base image, as long as it's not going to create more friction, i.e. if it can be completely local during the build and not pushed to Docker Hub. But that base image should have no traces of Cassandra.

Between "wait for Cassandra", using "dns_search", and defining base image, there are at least 3 independent PRs.

isaachier: Right I think avoiding Cassandra is the simplest way to fix most of these issues (DNS stuff still needs another fix if it comes up again).

PR #1212, opened 2018-11-24
WIP: update uber/jaeger-lib to master

gouthamve:
This is a don't merge PR. 

I understand that this will only make it in once `uber/jaeger-lib@v2.0` is released. But I spent the better part of 2hrs changing the APIs for the breaking changes, and I'm hoping it'll be useful for the person updating the dependency later :)


yurishkuro:
done in #1282

PR #1214, opened 2018-11-25
WIP - gRPC Plugin framework

olivierboucher:
- Issue #422 POC

- Plugin based storage backend


plugin/storage/grpc/shared/mapping.go
```diff
@@ -0,0 +1,422 @@
...
+package shared
+
+import (
+	"time"
+
+	"github.com/jaegertracing/jaeger/model"
+	"github.com/jaegertracing/jaeger/plugin/storage/grpc/proto"
+)
+
+func DependencyLinkSliceFromProto(p []*proto.DependencyLink) []model.DependencyLink {
```
yurishkuro: I would like to avoid the need for this file. The whole point of generating domain model from proto IDL was to be able to reuse those types in the grpc plugin interfaces

yurishkuro:
Great start. My main comment is about avoiding duplicating the trace model.

plugin/storage/grpc/proto/model.proto
```diff
@@ -0,0 +1,90 @@
+syntax = "proto3";
```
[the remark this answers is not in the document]
olivierboucher: Yes it was

plugin/storage/grpc/plugin.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+// WithConfiguration creates a new plugin given the provided configuration
+func WithConfiguration(configuration config.Configuration) (*Store, error) {
+	client := plugin.NewClient(&plugin.ClientConfig{
+		HandshakeConfig: shared.Handshake,
+		VersionedPlugins: map[int]plugin.PluginSet{
+			18: shared.PluginMap,
+		},
+		Cmd:              exec.Command("sh", "-c", configuration.PluginBinary),
```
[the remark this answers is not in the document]
olivierboucher: We do not, good point

cmd/memory-grpc-plugin/main.go
```diff
@@ -0,0 +1,36 @@
...
+
+	"github.com/jaegertracing/jaeger/plugin/storage/grpc/shared"
+	"github.com/jaegertracing/jaeger/plugin/storage/memory"
+)
+
+func main() {
+	plugin.Serve(&plugin.ServeConfig{
+		HandshakeConfig: shared.Handshake,
+		VersionedPlugins: map[int]plugin.PluginSet{
+			18: map[string]plugin.Plugin{
```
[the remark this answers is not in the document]
olivierboucher: Meant for 1.8 (current version). Should we use something else?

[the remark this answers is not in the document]
yurishkuro: I don't fully understand the semantics here. Is this the version of the plugin API?

chvck:
Hi @olivierboucher, we're keen to use the work that you have started here, thanks! Do you need any assistance in pushing it to a point where it can be considered complete?

olivierboucher:
@chvck I have been a bit busy lately but I still plan on finishing this. I might need some help with the proto generation in order to use gogoproto and remove the mapping code

glide.yaml
```diff
@@ -76,3 +76,4 @@ import:
 - package: golang.org/x/sys
   subpackages:
   - unix
+- package: github.com/hashicorp/go-plugin
```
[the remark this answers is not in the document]
olivierboucher: No there is none

cmd/memory-grpc-plugin/main.go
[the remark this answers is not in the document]
olivierboucher: It is the version of the plugin API. I changed it to 1

plugin/storage/grpc/options.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"flag"
+
+	"github.com/spf13/viper"
+
+	"github.com/jaegertracing/jaeger/pkg/grpc/config"
+)
+
+const pluginBinary = "grpc-plugin.binary"
```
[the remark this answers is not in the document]
olivierboucher: What would you propose? I'm afraid to come up with something thats too long

olivierboucher:
@yurishkuro 

I am trying to add support for configuration and I may need advice.

I am doing the following to read from the viper context:

```
const pluginArgsPrefix = "grpc-storage-plugin.arg."

func (opt *Options) InitFromViper(v *viper.Viper) {
	opt.Configuration.PluginBinary = v.GetString(pluginBinary)

	for _, f := range v.AllKeys() {
		if strings.HasPrefix(f, pluginArgsPrefix) {
			opt.Configuration.PluginArgs = append(opt.Configuration.PluginArgs, v.GetString(f))
		}
	}
}
```

However I am confused as what to add in the `AddFlags` method since the flags are dynamic. Should I do nothing? Not sure how to document this.

My goal is for people to use the args like this:

`grpc-storage-plugin.arg.plugin-specific-arg=1`

and have `plugin-specific-arg=1` passed to their plugin.

Not sure if this conversation should be moved to an issue either.

Thanks

olivierboucher:
@yurishkuro 

Makes sense ! I just committed the implementation. 

I think it's time to benchmark and write tests against the new code. I'm not familiar with code benchmarking, any method you would suggest? Is there existing code somewhere ?

Thanks

isaachier:
This is where #1221 might come in. If you can figure out crossdock, try adding an integration test. It is definitely tricky so I'll see if I have some free time to help. Benchmarking can be as simple as measuring the latency in requests and perhaps throughput. Not sure how formal the process needs to be.

yurishkuro:
Well, crossdock is more for correctness integration testing, we never used it for performance.

For benchmarking storage I don't think the integration tests are important. We just need a data generator pushing data into SpanWriter interface, one running in-process, another running via grpc-plugin.

yurishkuro:
We have a trace generator in the internal repo that I used to stress test the agents back in the days. Let me see if I can move it to oss.

olivierboucher:
I know it's not top quality data but I re-used the createTrace method from the integration tests and called it in a loop. I compared memory vs grpc-memory-plugin and it's about the same numbers.

Keep me posted regarding that trace generator

yurishkuro:
trace generator also creates fairly simple traces, it's purpose is to generate load of on agents or collectors.

https://github.com/jaegertracing/jaeger/pull/1245

olivierboucher:
I did not, what I did is probably meaningless. I will manage to checkout the tracegen and run proper tests

olivierboucher:
I ran the tracegen tool and those are the results:

```
2018-12-12T15:25:45.913-0500    INFO    tracegen/main.go:55     Initialized global tracer
2018-12-12T15:26:15.917-0500    INFO    tracegen/worker.go:93   Worker 3 generated 519026 traces        {"worker": 3}
2018-12-12T15:26:15.917-0500    INFO    tracegen/worker.go:93   Worker 0 generated 516428 traces        {"worker": 0}
2018-12-12T15:26:15.917-0500    INFO    tracegen/worker.go:93   Worker 1 generated 515793 traces        {"worker": 1}
2018-12-12T15:26:15.917-0500    INFO    tracegen/worker.go:93   Worker 2 generated 517417 traces        {"worker": 2}
2018-12-12T15:26:15.917-0500    INFO    tracegen/main.go:59     Waiting 1.5sec for metrics to flush
olivierboucher in ~/go/src/github.com/jaegertracing/jaeger on plugin-framework *> (production:default)
$ tracegen --duration 30s --workers 4
2018-12-12T15:27:18.570-0500    INFO    tracegen/main.go:55     Initialized global tracer
2018-12-12T15:27:48.572-0500    INFO    tracegen/worker.go:93   Worker 1 generated 729523 traces        {"worker": 1}
2018-12-12T15:27:48.572-0500    INFO    tracegen/worker.go:93   Worker 2 generated 732114 traces        {"worker": 2}
2018-12-12T15:27:48.572-0500    INFO    tracegen/worker.go:93   Worker 0 generated 732004 traces        {"worker": 0}
2018-12-12T15:27:48.572-0500    INFO    tracegen/worker.go:93   Worker 3 generated 730584 traces        {"worker": 3}
2018-12-12T15:27:48.572-0500    INFO    tracegen/main.go:59     Waiting 1.5sec for metrics to flush
```

First is using the grpc plugin, second is using memory. It looks like it has a huge performance impact but I think that was expected. Is this tolerable ?

olivierboucher:
@yurishkuro 

You're correct I did ran tracegen against the 2 backends.

Can you help me by telling me which metrics are meaningful?

I still have all the prometheus metrics stored locally.

yurishkuro:
These are the metrics that should be coming out of any storage

https://github.com/jaegertracing/jaeger/blob/50429cb55aee9280e39621a0b5aae3c2afe5d190/storage/spanstore/metrics/write_metrics.go#L24-L30

Try looking for something like these:
```
jaeger_collector_save_latency_bucket{host="hostname",le="0.005"} 518383
jaeger_collector_save_latency_bucket{host="hostname",le="0.01"} 518424
jaeger_collector_save_latency_bucket{host="hostname",le="0.025"} 518536
jaeger_collector_save_latency_bucket{host="hostname",le="0.05"} 518563
jaeger_collector_save_latency_bucket{host="hostname",le="0.1"} 518563
jaeger_collector_save_latency_bucket{host="hostname",le="0.25"} 518563
jaeger_collector_save_latency_bucket{host="hostname",le="0.5"} 518563
jaeger_collector_save_latency_bucket{host="hostname",le="1"} 518563
jaeger_collector_save_latency_bucket{host="hostname",le="2.5"} 518563
jaeger_collector_save_latency_bucket{host="hostname",le="5"} 518563
jaeger_collector_save_latency_bucket{host="hostname",le="10"} 518563
jaeger_collector_save_latency_bucket{host="hostname",le="+Inf"} 518563
jaeger_collector_save_latency_sum{host="hostname"} 26.363240858001458
jaeger_collector_save_latency_count{host="hostname"} 518563
```

olivierboucher:
I have limited testing capabilities on my MBP (16GB ram fills up quickly)

Here is a graph, my prometheus instance scrape rate is tuned to 2s for this test:

image

First I ran grpc, then memory. From my understanding there only seem to be a ~1ms difference which leaves me perplex

yurishkuro:
I just realized that even though I split the tracegen into a library function separate from main, it still only works by generating traces via jaeger-client, it cannot generate spans in the domain model. It's probably not the best one to use here.

Really, the only benchmark we could have is generating somewhat realistic span and writing it to two SpanWriters:
  * generator -> SpanWriter (InMemory or maybe even noop storage)
  * generator -> SpanWriter (grpc Plugin) -> child process (same storage as in first item)

This way we just measure the CPU/memory overhead per span saved (can be done with Go benchmark, `testing.B`). We know there will be overhead, the benchmark will tell us how bad it is.

olivierboucher:
Not sure I want to commit the actual code, but here's how it went:

```
$ go test -bench=.
goos: darwin
goarch: amd64
pkg: github.com/jaegertracing/jaeger/cmd/noop-grpc-plugin
BenchmarkNoopSpanWriter-8               2000000000               0.28 ns/op
BenchmarkGRPCNoopSpanWriter-8              20000             99580 ns/op
PASS
ok      github.com/jaegertracing/jaeger/cmd/noop-grpc-plugin    2.909s
```

The test can be found here:

https://gist.github.com/olivierboucher/eeb0e4618b1d744dc7510fd3b80f1a0d

Could you help me define what a "realistic" span is? Right now the test uses empty spans

yurishkuro:
by "realistic span" I mean something that comes from real apps, like the hotrod, including all the string and timing fields, a span reference, couple of tags and logs.

So your result, if I am reading it correctly, shows ~100 microseconds overhead. The actual save requests to e.g. Cassandra have millisecond latency.

olivierboucher:
So that's good right? I will modify the spans to see how it impacts the timing because of the extra serialization. Do you want me to commit everything once it's done?

olivierboucher:
@yurishkuro 

I have an issue with the tests since I add to move `TraceID` to `model.proto`.

These 2 methods defined in `ids.go` are now longer present and are defined by protobuf:

```
// MarshalTo converts trace ID into a binary representation. Called by protobuf serialization.
func (t *TraceID) MarshalTo(data []byte) (n int, err error) {
	var b [16]byte
	binary.BigEndian.PutUint64(b[:8], uint64(t.High))
	binary.BigEndian.PutUint64(b[8:], uint64(t.Low))
	return marshalBytes(data, b[:])
}

// Unmarshal inflates this trace ID from binary representation. Called by protobuf serialization.
func (t *TraceID) Unmarshal(data []byte) error {
	if len(data) < 16 {
		return fmt.Errorf("buffer is too short")
	}
	t.High = binary.BigEndian.Uint64(data[:8])
	t.Low = binary.BigEndian.Uint64(data[8:])
	return nil
}
```

It breaks all the fixtures used by tests.

model/proto/model.proto
```diff
@@ -137,8 +137,12 @@ message Span {
 }
 
 message TraceID {
-  uint64 low = 1;
-  uint64 high = 2;
+  uint64 low = 1 [
+    (gogoproto.jsontag) = "lo"
```
yurishkuro: ?

olivierboucher: It was previously 

```
// TraceID is a random 128bit identifier for a trace
type TraceID struct {
	Low  uint64 `json:"lo"`
	High uint64 `json:"hi"`
}
```

So I figured doing this would fix the issue but it's not the case

isaachier: I'm taking a look. Just a sec.

model/proto/model.proto
```diff
@@ -76,7 +76,7 @@ enum SpanRefType {
 };
 
 message SpanRef {
-  bytes trace_id = 1 [
+  TraceID trace_id = 1 [
```
yurishkuro: please do not change proto model, it is fixed.

olivierboucher:
I think I understand what to do in order to make it work. I'll define trace ids as bytes just like it is done in `model.proto` and add the gogo annotations. Not sure how this will translate for other languages but let's focus on go for now

isaachier:
I believe the Protobuf compiles for any language (despite gogo-specific annotations), as long as the gogoproto Protobuf files are available for inclusion.

olivierboucher:
I fixed the issue by reverting changes to TraceID **but** I had to move `DependencyLink` to `model.proto` in order for the plugin interface to implement `dependencystore.Reader`. It does not break anything as far as test goes. Is there any problem with that?

yurishkuro:
> I fixed the issue by reverting changes to TraceID but I had to move DependencyLink to model.proto in order for the plugin interface to implement dependencystore.Reader. It does not break anything as far as test goes. Is there any problem with that?

That should be fine, although we may want to do that as a separate PR. Also, there seem to be any changes in the generated file, something is off, most likely a different version of the generator.

olivierboucher:
> That should be fine, although we may want to do that as a separate PR. Also, there seem to be any changes in the generated file, something is off, most likely a different version of the generator.

What versions of protoc/protoc-gen-go should I run locally? Maybe this is off, I'm running 3.6.1.

So if we want to do this in 2 separate PR I would have to do the DependencyLink one first because this one depends on it.

yurishkuro:
Yes, we can do proto change as a separate (first) PR.

I don't think the version of protoc matters much, it's the version of gogo that's important. The `Makefile` has a target `proto-install` that installs the gogo generators from the `/vendor` directory.

yurishkuro:
this seems to be a blocking issue #1258

olivierboucher:
I fixed this issue adding the prune settings in Gopkg.toml. I will make a PR for those changes

chvck:
I'm not sure if this is the right place for this but it might be worth consideration. I've had a quick look at implementing a plugin using this and it seems nice, I have an issue trying to create a plugin from outside of jaeger due to vendoring though. In my own package if I do 

`	plugin.Serve(&plugin.ServeConfig{
		HandshakeConfig: plugin.HandshakeConfig{
			MagicCookieKey:   "STORAGE_PLUGIN",
			MagicCookieValue: "jaeger",
		},
		VersionedPlugins: map[int]plugin.PluginSet{
			1: map[string]plugin.Plugin{
				shared.StoragePluginIdentifier: &shared.StorageGRPCPlugin{
					Impl: &store,
				},
			},
		},
		GRPCServer: plugin.DefaultGRPCServer,
	})`

then I see

`cannot use shared.StorageGRPCPlugin literal (type *shared.StorageGRPCPlugin) as type "github.com/hashicorp/go-plugin".Plugin in map value:
	*shared.StorageGRPCPlugin does not implement "github.com/hashicorp/go-plugin".Plugin (wrong type for Client method)
		have Client(*"github.com/jaegertracing/jaeger/vendor/github.com/hashicorp/go-plugin".MuxBroker, *rpc.Client) (interface {}, error)
		want Client(*"github.com/hashicorp/go-plugin".MuxBroker, *rpc.Client) (interface {}, error)`

I could be doing this wrong/jumping the gun but I thought that it was worth bringing it up just in case.

yurishkuro:
@chvck this often happens if your code is pulling jaegertracing/jaeger from GOPATH instead of a flat vendor dir inside your repo.

plugin/storage/grpc/proto/storage.proto
```diff
@@ -0,0 +1,201 @@
...
+
+message EmptyResponse {
+
+}
+
+message StoragePluginError {
+    string message = 1;
+}
+
+service StoragePlugin {
```
yurishkuro: We should rethink this naming. Once #760 is merged, it can only be used with all-in-one because collector and query service still need to share the in-memory storage. It would be nice to build a standalone "remote storage" component that can be run as a separate service and accessed by collector/query via gRPC API. *This* file is that API, but it wouldn't be an API of just the harshicorp-plugin, it's an abstract gRPC Storage API .

plugin/storage/grpc/proto/storage.proto
```diff
@@ -0,0 +1,201 @@
...
+    // dependencystore/Writer
+//    rpc WriteDependencies(WriteDependenciesRequest) returns (WriteDependenciesResponse);
+    // dependencystore/Reader
+    rpc GetDependencies(GetDependenciesRequest) returns (GetDependenciesResponse);
```
yurishkuro: Is there a strong reason for merging all different types of storage into a single proto `service`? Could we not have the same component implement different services?

plugin/storage/grpc/plugin.go
```diff
@@ -0,0 +1,58 @@
...
+import (
+	"context"
+	"github.com/jaegertracing/jaeger/model"
+	"github.com/jaegertracing/jaeger/pkg/grpc/config"
+	"github.com/jaegertracing/jaeger/plugin/storage/grpc/shared"
+	"github.com/jaegertracing/jaeger/storage/spanstore"
+	"time"
+)
+
+type Store struct {
```
yurishkuro: what is the purpose of this struct? Given its 1-1 delegation pattern, couldn't `shared.StoragePlugin` type used directly in place of wherever this `Store` type was meant to use?

chvck:
@olivierboucher - do you need any assistance in completing this work?

Also, just to raise this for awareness as I'm not sure it's an issue for this PR or should be addressed once this is in but when I request ~200+ traces I'm hitting gRPC message size limits.

`Unhandled Rejection (Error): HTTP Error: grpc error: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (4467848 vs. 4194304)`

annanay25:
@chvck could you kindly open an issue with your description? We would investigate it independently of this PR :)

chvck:
@yurishkuro am I correct in thinking that a rough plan here would be:
1) Split the proto storage service into something that more closely matches current implementations. 
2) Split out the proto changes and made a new PR from those.
3) Update the Go plugin to address comments in this PR.

Sound about right? Also, what would be the best way for me to go about doing this? I think that I'm going to have to fork and create new PRs even for just updating the Go plugin part.

@annanay25 I'd be happy to create an issue but does it make sense outside of this PR? I think that it's specific to the span data volume returned over the grpc service implemented in the PR.

yurishkuro:
@chvck yes, although I'm not sure about the difference between 1 and 2.

jpkrohling:
What's the state of this PR?

chvck:
> What's the state of this PR?

It seems to be stale now. I've just created https://github.com/jaegertracing/jaeger/pull/1461 which builds on top of the work in this PR.

yurishkuro:
superseded by #1461

PR #1216, opened 2018-11-26
[WIP] [HotROD] Add optional fix to make driver.FindNearest call redis.GetDriver concurrently

tiffon:

The HotROD demo can be optimized by calling `redis.GetDriver` concurrently.

Adds a CLI flag (`-C`) to call `redis.GetDriver` concurrently.

-----

_This is my first foray into Go in some time, lmk if things look awry._

### DIFF
--- a/examples/hotrod/cmd/root.go
+++ b/examples/hotrod/cmd/root.go
@@ -34,9 +34,10 @@ var (
 	logger         *zap.Logger
 	metricsFactory metrics.Factory
 
-	fixDBConnDelay         time.Duration
-	fixDBConnDisableMutex  bool
-	fixRouteWorkerPoolSize int
+	fixDBConnDelay            time.Duration
+	fixDBConnDisableMutex     bool
+	fixGetDriversConcurrently bool
+	fixRouteWorkerPoolSize    int
 
 	customerPort int
 	driverPort   int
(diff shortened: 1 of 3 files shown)


examples/hotrod/cmd/root.go
```diff
@@ -64,6 +65,7 @@ func init() {
 	RootCmd.PersistentFlags().StringVarP(&metricsBackend, "metrics", "m", "expvar", "Metrics backend (expvar|prometheus)")
 	RootCmd.PersistentFlags().DurationVarP(&fixDBConnDelay, "fix-db-query-delay", "D", 300*time.Millisecond, "Average lagency of MySQL DB query")
 	RootCmd.PersistentFlags().BoolVarP(&fixDBConnDisableMutex, "fix-disable-db-conn-mutex", "M", false, "Disables the mutex guarding db connection")
+	RootCmd.PersistentFlags().BoolVarP(&fixGetDriversConcurrently, "fix-get-driver-concurrent", "C", false, "Enable concurrency in driver::findNearest")
```
yurishkuro: I don't think it should be a boolean, rather an integer (default 1), exactly like `fixRouteWorkerPoolSize`

examples/hotrod/services/driver/server.go
```diff
@@ -72,31 +77,57 @@ func (s *Server) Run() error {
...
-// FindNearest implements Thrift interface TChanDriver
+// FindNearest implements Thrift interface TChanDriver. config.GetDriversConcurrently adds concurrency.
 func (s *Server) FindNearest(ctx thrift.Context, location string) ([]*driver.DriverLocation, error) {
 	s.logger.For(ctx).Info("Searching for nearby drivers", zap.String("location", location))
 	driverIDs := s.redis.FindDriverIDs(ctx, location)
-
 	retMe := make([]*driver.DriverLocation, len(driverIDs))
-	for i, driverID := range driverIDs {
-		var drv Driver
+	var topErr error
```
yurishkuro: there seems to be a lot of unnecessary complexity here. See how `fixRouteWorkerPoolSize` is used, this should be exactly the same, using an executor pool.

tiffon: Got it. Will change. Thanks.

examples/hotrod/cmd/root.go
tiffon: SGTM. Will change.

tiffon:
Closing as I haven't progressed this in some time.

PR #1220, opened 2018-11-27
Fix gRPC test case

### LINKED ISSUE #1219: gRPC test case fails on macos

Make test cases pass on all Go-compatible platforms.

Test consistently fails on macos. Seems to rely on Linux behavior.

Use of port 1 is not inherently problematic, when privileges are correct. Use a truly invalid port to test address bind failure.

isaachier:

- Resolves #1219.

- Use port -1 and change expected error.

### DIFF
--- a/cmd/collector/app/grpcserver/grpc_server.go
+++ b/cmd/collector/app/grpcserver/grpc_server.go
@@ -50,11 +50,11 @@ func StartGRPCCollector(
 
 	api_v2.RegisterCollectorServiceServer(server, handler)
 	api_v2.RegisterSamplingManagerServer(server, sampling.NewGRPCHandler(samplingStrategy))
-	starServer(server, lis, logger, serveErr)
+	startServer(server, lis, logger, serveErr)
 	return lis.Addr(), nil
 }
 
-func starServer(server *grpc.Server, lis net.Listener, logger *zap.Logger, serveErr func(error)) {
+func startServer(server *grpc.Server, lis net.Listener, logger *zap.Logger, serveErr func(error)) {
 	var port string
 	if tcpAddr, ok := lis.Addr().(*net.TCPAddr); ok {
 		port = strconv.Itoa(tcpAddr.Port)
--- a/cmd/collector/app/grpcserver/grpc_server_test.go
+++ b/cmd/collector/app/grpcserver/grpc_server_test.go
@@ -38,19 +38,18 @@ func TestFailToListen(t *testing.T) {
 	l, _ := zap.NewDevelopment()
 	handler := app.NewGRPCHandler(l, &mockSpanProcessor{})
 	server := grpc.NewServer()
-	addr, err := StartGRPCCollector(1, server, handler, &mockSamplingStore{}, l, func(e error) {
+	_, err := StartGRPCCollector(-1, server, handler, &mockSamplingStore{}, l, func(e error) {
 	})
-	require.Nil(t, addr)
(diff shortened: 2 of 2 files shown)


cmd/collector/app/grpcserver/grpc_server.go
```diff
@@ -50,11 +50,11 @@ func StartGRPCCollector(
 
 	api_v2.RegisterCollectorServiceServer(server, handler)
 	api_v2.RegisterSamplingManagerServer(server, sampling.NewGRPCHandler(samplingStrategy))
-	starServer(server, lis, logger, serveErr)
+	startServer(server, lis, logger, serveErr)
```
isaachier: Fixed typo while I was here.

cmd/collector/app/grpcserver/grpc_server_test.go
```diff
@@ -38,19 +38,18 @@ func TestFailToListen(t *testing.T) {
...
-	assert.EqualError(t, err, "Failed to listen on gRPC port: listen tcp :1: bind: permission denied")
+	assert.EqualError(t, err, "Failed to listen on gRPC port: listen tcp: address -1: invalid port")
 }
 
 func TestFailServe(t *testing.T) {
 	lis := bufconn.Listen(0)
 	lis.Close()
 	core, logs := observer.New(zap.NewAtomicLevelAt(zapcore.ErrorLevel))
-	wg := &sync.WaitGroup{}
+	var wg sync.WaitGroup
```
isaachier: No need to allocate pointer here.

cmd/collector/app/grpcserver/grpc_server_test.go
```diff
@@ -38,19 +38,19 @@ func TestFailToListen(t *testing.T) {
 	l, _ := zap.NewDevelopment()
 	handler := app.NewGRPCHandler(l, &mockSpanProcessor{})
 	server := grpc.NewServer()
-	addr, err := StartGRPCCollector(1, server, handler, &mockSamplingStore{}, l, func(e error) {
+	addr, err := StartGRPCCollector(-1, server, handler, &mockSamplingStore{}, l, func(e error) {
```
yurishkuro: can you elaborate? hard to understand without context

yurishkuro: also, please define a var `const port = -1` instead of using a constant (but I still don't understand what -1 means)

isaachier: You mean elaborate the reason for using `-1` instead of `1` in a comment in the code? Or just here? I explained a bit in the issue and PR text.

cmd/collector/app/grpcserver/grpc_server_test.go
```diff
@@ -38,19 +38,20 @@ func TestFailToListen(t *testing.T) {
 	l, _ := zap.NewDevelopment()
 	handler := app.NewGRPCHandler(l, &mockSpanProcessor{})
 	server := grpc.NewServer()
-	addr, err := StartGRPCCollector(1, server, handler, &mockSamplingStore{}, l, func(e error) {
+	const port = -1 // Invalid port
```
yurishkuro: code as documentation: `const invalidPort = -1`

isaachier: Lol I was thinking this as I submitted the commit. Will update.

yurishkuro:
unit tests passed, merging

Issue #1229, opened 2018-11-28
Crossdock gRPC issues


isaachier:
## Requirement - what kind of business use case are you trying to solve?

I would like the crossdock tests to pass.

## Problem - what in Jaeger blocks you from solving the requirement?

gRPC communication between agent and collector fails for first one or two test traces due to `TRANSIENT_FAILURE` status in connection. This causes crossdock tests to fail consistently.

## Proposal - what do you suggest to solve the problem or improve the existing situation?

- @yurishkuro suggests maybe using TChannel in crossdock for now until we decide how to handle the `TRANSIENT_FAILURE` issue.
- Pass `grpc.FailFast(false)` to disable the failure in the RPC between the agent and collector. The connection will wait for connection to become ready. More details here: https://github.com/grpc/grpc/blob/master/doc/wait-for-ready.md.

## Any open questions to address

How do we want to use fail fast/wait for ready behavior in general?

yurishkuro:
Unfortunately, #1230 [did not directly help the clients](https://github.com/jaegertracing/jaeger-client-python/blob/master/crossdock/rules.mk#L3) since they still pull the yml file from docker-compose dir instead of crossdock dir. I wonder if we should use a symlink from crossdock to docker-compose after the gRPC issue is fixed

isaachier:
Maybe this will help: https://docs.docker.com/compose/extends/#multiple-compose-files.

pavolloffay:
Do we have a to failed travis job?

isaachier:
Hard to find them because we usually restart them. Even when crossdock does fail, at least half the time the issue is #1221.

yurishkuro:
Pretty much all client builds (eg Python) are failing on master

pavolloffay:
IIRC during the call somebody mentioned that the issue happens because collector's healthcheck signals ready before grpc port if open.

We could change `StartGrpcServer` to block until clients can connect or return a channel to signal that its ready state. https://github.com/jaegertracing/jaeger/blob/master/cmd/collector/app/grpcserver/grpc_server.go#L35

yurishkuro:
How would we do that, by querying the server from the collector itself?

Fundamentally, I think we're missing some resiliency features in the agent, ie that it loses the data even for intermittent hickup in the collector connection.

isaachier:
[...]

PR #1230, opened 2018-11-28
Temporarily switch back to tchannel in crossdock

yurishkuro:
Cf https://github.com/jaegertracing/jaeger/issues/1229

cc @isaachier

### DIFF
--- a/crossdock/jaeger-docker-compose.yml
+++ b/crossdock/jaeger-docker-compose.yml
@@ -1,0 +1,48 @@
+version: '2'
+
+services:
+    jaeger-collector:
+      image: jaegertracing/jaeger-collector
+      command: ["--cassandra.keyspace=jaeger_v1_dc1", "--cassandra.servers=cassandra", "--collector.zipkin.http-port=9411"]
+      ports:
+        - "14269"
+        - "14268:14268"
+        - "14267"
+        - "14250"
+        - "9411:9411"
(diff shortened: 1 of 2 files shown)


crossdock/jaeger-docker-compose.yml
```diff
@@ -0,0 +1,48 @@
+version: '2'
```
isaachier: Why use a new file?

crossdock/rules.mk
```diff
@@ -1,5 +1,9 @@
 XDOCK_YAML=crossdock/docker-compose.yml
-JAEGER_COMPOSE_YAML=docker-compose/jaeger-docker-compose.yml
+
+# FIXME temporarily switching back to tchannel
+# See https://github.com/jaegertracing/jaeger/issues/1229
+# JAEGER_COMPOSE_YAML=docker-compose/jaeger-docker-compose.yml
+JAEGER_COMPOSE_YAML=crossdock/jaeger-docker-compose.yml
```
isaachier: Same question.

crossdock/jaeger-docker-compose.yml
yurishkuro: because I only want to affect crossdock.

isaachier: I see your point. I like the approach.

Issue #1190, opened 2018-11-15
[ingester] Crash if Kafka is specified as storage


yurishkuro:
Reported by @ledor473 - https://github.com/jaegertracing/jaeger/pull/1094#issuecomment-439221399

I've just realized that if you run the Ingester with `SPAN_STORAGE_TYPE=elasticsearch,kafka`, the Ingester crash with the following:
```
WARNING: multiple span storage types have been specified. Only the first type (elasticsearch) will be used for reading and archiving.

flag redefined: kafka.brokers
panic: flag redefined: kafka.brokers

goroutine 1 [running]:
flag.(*FlagSet).Var(0xc000080d80, 0xefd9a0, 0xc00018d080, 0xe04f6e, 0xd, 0xe2cb58, 0x4b)
	/home/travis/.gimme/versions/go1.11.1.linux.amd64/src/flag/flag.go:805 +0x529
flag.(*FlagSet).StringVar(0xc000080d80, 0xc00018d080, 0xe04f6e, 0xd, 0xe05426, 0xe, 0xe2cb58, 0x4b)
	/home/travis/.gimme/versions/go1.11.1.linux.amd64/src/flag/flag.go:708 +0x8a
flag.(*FlagSet).String(0xc000080d80, 0xe04f6e, 0xd, 0xe05426, 0xe, 0xe2cb58, 0x4b, 0xe00e66)
	/home/travis/.gimme/versions/go1.11.1.linux.amd64/src/flag/flag.go:721 +0x8b
github.com/jaegertracing/jaeger/cmd/ingester/app.AddFlags(0xc000080d80)
	/home/travis/gopath/src/github.com/jaegertracing/jaeger/cmd/ingester/app/flags.go:84 +0x6e
github.com/jaegertracing/jaeger/pkg/config.AddFlags(0xc00015a0f0, 0xc000186240, 0xc000159f28, 0x5, 0x5, 0xc00003a042, 0xd)
	/home/travis/gopath/src/github.com/jaegertracing/jaeger/pkg/config/config.go:36 +0x56
main.main()
	/home/travis/gopath/src/github.com/jaegertracing/jaeger/cmd/ingester/main.go:138 +0x59f
```

I'm not sure how, but I feel we should add a validation on the storage type to prevent the use to `KAFKA` in the Ingester. Unless we want to support Kafka -> Kafka with the Ingester, but that feels wrong

yurishkuro:
[...]

PR #1231, opened 2018-12-01
Only support the Kafka storage in Collector

ledor473:
- Solves https://github.com/jaegertracing/jaeger/issues/1190

- Add an unsupported list to the storage factory
- Make Kafka as unsupported in Ingester, Query, all-in-one (because Ingester is not started anyway)

### DIFF
--- a/cmd/all-in-one/main.go
+++ b/cmd/all-in-one/main.go
@@ -68,6 +68,8 @@ import (
 	zc "github.com/jaegertracing/jaeger/thrift-gen/zipkincore"
 )
 
+var unsupportedTypes = []string{storage.KafkaStorageType}
+
 // all-in-one/main is a standalone full-stack jaeger backend, backed by a memory store
 func main() {
 	var signalsChannel = make(chan os.Signal)
@@ -77,7 +79,7 @@ func main() {
 		os.Setenv(storage.SpanStorageTypeEnvVar, "memory") // other storage types default to SpanStorage
 	}
 	storageFactory, err := storage.NewFactory(storage.FactoryConfigFromEnvAndCLI(os.Args, os.Stderr),
-		[]string{storage.KafkaStorageType})
+		storage.Options{UnsupportedTypes: unsupportedTypes})
 	if err != nil {
 		log.Fatalf("Cannot initialize storage factory: %v", err)
 	}
--- a/cmd/all-in-one/main.go
+++ b/cmd/all-in-one/main.go
@@ -76,7 +76,8 @@ func main() {
 	if os.Getenv(storage.SpanStorageTypeEnvVar) == "" {
 		os.Setenv(storage.SpanStorageTypeEnvVar, "memory") // other storage types default to SpanStorage
 	}
-	storageFactory, err := storage.NewFactory(storage.FactoryConfigFromEnvAndCLI(os.Args, os.Stderr))
+	storageFactory, err := storage.NewFactory(storage.FactoryConfigFromEnvAndCLI(os.Args, os.Stderr),
+		[]string{storage.KafkaStorageType})
 	if err != nil {
 		log.Fatalf("Cannot initialize storage factory: %v", err)
 	}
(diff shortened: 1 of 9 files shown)


cmd/all-in-one/main.go
```diff
@@ -76,7 +76,8 @@ func main() {
 	if os.Getenv(storage.SpanStorageTypeEnvVar) == "" {
 		os.Setenv(storage.SpanStorageTypeEnvVar, "memory") // other storage types default to SpanStorage
 	}
-	storageFactory, err := storage.NewFactory(storage.FactoryConfigFromEnvAndCLI(os.Args, os.Stderr))
+	storageFactory, err := storage.NewFactory(storage.FactoryConfigFromEnvAndCLI(os.Args, os.Stderr),
+		[]string{storage.KafkaStorageType})
```
yurishkuro: if we want to introduce optional params, we should switch to passing an Options struct instead of positional args.

https://github.com/jaegertracing/jaeger-lib/blob/3fa9aa04b65a266a1bbe0a5c8ac1408281b2a64a/metrics/factory.go#L31

ledor473: I choose to make it an optional parameter as the [FactoryConfig] is currently only for config from the Cobra, but I could indeed add an Options struct inside the same package and use that!

[FactoryConfig]: https://github.com/jaegertracing/jaeger/blob/master/plugin/storage/factory_config.go#L34-L39

yurishkuro:
The design of this fix doesn't sit well with me. Right now the details of the storage are fully abstracted from the binaries, but here we're introducing some very specific & peculiar dependency that they have to _declare Kafka as unsuitable storage_.

The reason Kafka doesn't work for all-in-one is because it does not implement the SpanReader. The reason it does not work for ingester is because the flags are clashing (aside from the flags issue, however, there's no strong reason why ingester wouldn't do Kafka->Kafka, e.g. as a form of replicating data to another topic). If we somehow solved the clash of flags issue, then I assume all-in-one would've exited with a better error message like "reader not implemented", and we wouldn't need to do any changes. Alternatively, if we did want to handle it more explicitly, then the storage factory could expose metadata methods like `CanRead(), CanWrite()`, and all-in-one would check both and fail for Kafka storage whose factory would return false from `CanRead()`.

ledor473:
>there's no strong reason why ingester wouldn't do Kafka->Kafka, e.g. as a form of replicating data to another topic)
@yurishkuro if we want to support that, maybe we could have a set of flags for the consumer and another set of flags for the producer. Something like this:
```
--kafka.consumer.brokers
--kafka.consumer.topic
--kafka.consumer.encoding
--kafka.consumer.group-id
```
And:
```
--kafka.producer.brokers
--kafka.producer.topic
--kafka.producer.encoding
```

This way you can both replicate from one topic to another, but we can also replicate to one cluster to another cluster. This would also get around the flags being declared twice.

ledor473:
@yurishkuro I was waiting for a confirmation on the latest suggestion (renaming the flags to avoid the exception and to allow the use of multiples brokers/topics).

I might have missed the 👍 on it, but yes I plan to fix this PR.

yurishkuro:
cool. 

I think GitHub doesn't notify of reactions via +1 button, I'll make sure to comment explicitly in the future.

ledor473:
@yurishkuro the change is mostly done in another branch but I see that we do not pass the `logger` to `InitFromViper` (https://github.com/jaegertracing/jaeger/blob/master/plugin/storage/kafka/options.go#L75-L81).

I see 2 options to log the current flag as deprecated: 
1. I change that method in the `Configurable` interface to provide a logger and use `logger.warn`
    - Similar to this: https://github.com/jaegertracing/jaeger/blob/3f77534ea6c9cac637203d378588164c0e264790/cmd/agent/app/reporter/tchannel/flags.go#L70-L74

2. I use `fmt.Printf`
    - Similar to this: https://github.com/jaegertracing/jaeger/blob/3f77534ea6c9cac637203d378588164c0e264790/plugin/storage/factory_config.go#L88-L93

yurishkuro:
I would go with option 2, much fewer changes this way

ledor473:
https://github.com/jaegertracing/jaeger/pull/1360 is merged, closing this one

PR #1240, opened 2018-12-06
Dep migration

isaachier:
- Resumes work from #610.

- Title says it all.


isaachier:
@vprithvi @yurishkuro any thoughts on this change?

Makefile
```diff
@@ -153,13 +156,13 @@ lint: lint-gosec
 	@./scripts/import-order-cleanup.sh stdout > $(IMPORT_LOG)
 	@[ ! -s "$(FMT_LOG)" -a ! -s "$(IMPORT_LOG)" ] || (echo "Go fmt, license check, or import ordering failures, run 'make fmt'" | cat - $(FMT_LOG) && false)
 
-.PHONY: install-glide
```
jpkrohling: Could we keep this for a while, as an alias to `install-dep` + WARN message stating that it's deprecated?

isaachier: Sure good idea.

isaachier:
@jpkrohling I hear the reasoning to move to Go modules. This PR addresses an Uber-specific issue, where we must use dep or glide, but do not support Go modules.

Makefile
```diff
@@ -1,15 +1,18 @@
 PROJECT_ROOT=github.com/jaegertracing/jaeger
 # TOP_PKGS is used with 'go test'
-# TODO: try to do this without glide, since it may not be installed initially
-TOP_PKGS := $(shell glide novendor | \
+TOP_PKGS := $(shell go list ./... | \
```
yurishkuro: so:
```
$ time go list ./...
...
go list ./...  1.51s user 2.29s system 105% cpu 3.595 total
```
Can we do without it altogether? Do we still need TOP_PKGS? If we do, can we use something lighter than `go list`?

Makefile
```diff
@@ -155,11 +158,15 @@ lint: lint-gosec
 
 .PHONY: install-glide
 install-glide:
-	@which glide > /dev/null || go get github.com/Masterminds/glide
+	@echo "Jaeger has migrated to dep, please run 'make install-dep'" 1>&2
```
yurishkuro: this target should be removed

Gopkg.toml
```diff
@@ -0,0 +1,118 @@
...
+# [[override]]
+#   name = "github.com/x/y"
+#   version = "2.4.0"
+#
+# [prune]
+#   non-go = false
+#   go-tests = true
+#   unused-packages = true
+
+
```
yurishkuro: I am not sure this file is equivalent to glide.yaml. We can debate separately if we do or do not want to pin certain things, but for the purpose of this migration I would prefer no changes to the dependencies.

Specifically, `dep` defaults to `^`, i.e. `version = "1.16.0"` in `dep` is equivalent to `"^1.16.0"` in glide. In order to do an actual pin, we need to use `=`, e.g. `version = "=1.16.0"`

yurishkuro:
> Retain install-glide but print warning  …

why?

Makefile
yurishkuro: he said to fall back to dep, which this doesn't. Also, this looks like it's always been an internal target, not something used outside of make, so I don't see much value in keeping just the name.

Gopkg.toml
yurishkuro: I don't think it did, none of the projects are pinned, while many were pinned in glide.

isaachier: OK but why did we pin these. I imagine this will end up being a nuisance for internal dep to handle constraints.

yurishkuro: we don't pin all of them, but we do pin some, esp. I remember having issues with gogo versions and had to work out a consistent working set. Like I said, we should have a discussion about not-pinning, but this migration should not be introducing changes to dependencies, imo.

isaachier: It is funny you say that because internally gogo/protobuf seems to be the packages causing the issue I have experienced.

Gopkg.toml
```diff
@@ -0,0 +1,118 @@
...
+  name = "github.com/prometheus/client_golang"
+  version = "=0.8.0"
+
+[[constraint]]
+  name = "github.com/spf13/cobra"
+  version = "0.0.1"
+
+[[constraint]]
+  name = "github.com/spf13/pflag"
+  version = "1.0.0"
```
yurishkuro: nit: maybe explicitly use `^` when we're not pinning?

Makefile
```diff
@@ -92,7 +81,7 @@ clean:
 
 .PHONY: test
 test: go-gen
-	bash -c "set -e; set -o pipefail; $(GOTEST) $(TOP_PKGS) | $(COLORIZE)"
+	bash -c "set -e; set -o pipefail; $(GOTEST) ./... | $(COLORIZE)"
```
yurishkuro: 🎈

Makefile
```diff
@@ -112,7 +101,7 @@ all-srcs:
 cover: nocover
 	@echo pre-compiling tests
 	@time go test -i $(ALL_PKGS)
-	@./scripts/cover.sh $(shell go list $(TOP_PKGS))
+	@./scripts/cover.sh $(shell go list ./...)
```
yurishkuro: I think we need to get rid of this script, it's no longer needed. See https://github.com/opentracing/opentracing-go/pull/202

But can do in another PR (add a TODO)

yurishkuro:
please rebase

jpkrohling:
> > Retain install-glide but print warning  …
> 
> why?

The idea is to give a chance to people who might be using this target. This *seems* internal, but it's possible that people might be building Jaeger on their own, applying custom patches to it. It would be polite to give them a heads-up before breaking their scripts.

PR #1245, opened 2018-12-12
Add tracegen

yurishkuro:
Small utility used to generate continuous stream of simple spans, useful for stress testing.


internal/tracegen/config.go
```diff
@@ -0,0 +1,79 @@
...
+	Pause    time.Duration
+	Duration time.Duration
+}
+
+// Flags registers config flags.
+func (c *Config) Flags(fs *flag.FlagSet) {
+	fs.IntVar(&c.Workers, "workers", 1, "Number of workers (goroutines) to run")
+	fs.IntVar(&c.Traces, "traces", 1, "Number of traces to generate in each worker (ignored if duration is provided")
+	fs.BoolVar(&c.Marshall, "marshall", false, "Whether to marshall trace context via HTTP headers")
+	fs.BoolVar(&c.Debug, "debug", false, "Whether to set DEBUG flag on the spans to prevent downsampling")
```
black-adder: downsampling is not a thing in OSS

internal/tracegen/worker.go
```diff
@@ -0,0 +1,95 @@
...
+		ext.PeerHostIPv4.Set(child, fakeIP)
+		ext.PeerService.Set(child, "tracegen-client")
+
+		time.Sleep(w.pause)
+
+		if w.pause == 0 {
+			child.Finish()
+			sp.Finish()
+		} else {
+			opt := opentracing.FinishOptions{FinishTime: time.Now().Add(123 * time.Microsecond)}
```
black-adder: what's the magic number?

internal/tracegen/worker.go
```diff
@@ -0,0 +1,95 @@
...
+
+	"github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go/ext"
+	"go.uber.org/zap"
+)
+
+type worker struct {
+	id       int             // worker id
+	traces   int             // how many traces the worker has to generate (only when duration==0)
+	marshall bool            // whether the worker needs to marshall trace context via HTTP headers
```
black-adder: not sure why this is necessary for a tracegen tool

internal/tracegen/config.go
```diff
@@ -0,0 +1,79 @@
...
+	"fmt"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"go.uber.org/zap"
+)
+
+// Config describes the test scenario.
+type Config struct {
```
tiffon: `Config` is also generate the spans, via `.run`, and it seems strange for a config to have a `Config#run` method. Maybe the name `Generator` would work?

internal/tracegen/config.go
isaachier: Perhaps "force sampling" makes more sense?

internal/tracegen/worker.go
isaachier: BTW marshall is used in various places in Jaeger code and I believe it is a typo. Marshal is the correct spelling AFAIK.

internal/tracegen/.nocover
```diff
@@ -0,0 +1 @@
+Test utility
```
yurishkuro: per offline discussion, this is worth merging even though it doesn't have tests (it is itself a test utility).

internal/tracegen/config.go
yurishkuro: Run is now a standalone method that takes Config

Issue #1241, opened 2018-12-06
Add `FindTraceIDs` to the spanstore.Reader interface


vprithvi:
## Requirement - what kind of business use case are you trying to solve?
Have an caching layer which retrieves full traces from cache. 

Finding traces matching a criterion involves finding matching traceIDs and then retrieving them. These are separate concerns for both ES and C*. 
I would like the interface to maintain these as separate concerns so that a caching layer may make decisions on where to retrieve traces from. 

## Problem - what in Jaeger blocks you from solving the requirement?
The current interface, 
`FindTraces(ctx context.Context, query *TraceQueryParameters) ([]*model.Trace, error)`
is pretty coarse and doesn't allow for easily inserting a cache. 

## Proposal - what do you suggest to solve the problem or improve the existing situation?
Add a `FindTraceIDs(ctx context.Context, query *TraceQueryParameters) ([]model.TraceID, error)` to the spanstore.Reader interface

yurishkuro:
+1, always wanted it.

Don't quite agree with your stated use case, but where it's definitely useful is when splitting span storage (e.g. a KV store like Cassandra) and span indexing (e.g. Pinot).

One issue with adding it to the Reader is how the query service is going to use it. E.g. if it always does Find first, followed by Get, then it introduces unnecessary 2nd query to the database even when the request could've been resolved in one query. However, by having FindTraceIDs in the interface allows us to implement a decorator Reader that delegates Find to one storage and delegates Get to another.

vprithvi:
>Having FindTraceIDs in the interface allows us to implement a decorator Reader that implements Get by calling Find on one storage and then Get on another.

Precisely - this is exactly what I want to do!

PR #1246, opened 2018-12-12
Add `FindTraceID` to the spanstore interface

vprithvi:
See #1241 

- Currently only implemented for C*
- Update the reader mock and pass through context everywhere


cmd/query/app/handler_archive_test.go
```diff
@@ -29,13 +29,13 @@ import (
 
 func TestGetArchivedTrace_NotFound(t *testing.T) {
 	mockReader := &spanstoremocks.Reader{}
-	mockReader.On("GetTrace", mock.AnythingOfType("model.TraceID")).
+	mockReader.On("GetTrace", mock.AnythingOfType("*context.valueCtx"), mock.AnythingOfType("model.TraceID")).
```
yurishkuro: What is `context.valueCtx`? Shouldn't it be `context.Context`?

storage/spanstore/metrics/decorator_test.go
```diff
@@ -33,13 +33,13 @@ func TestSuccessfulUnderlyingCalls(t *testing.T) {
 
 	mockReader := mocks.Reader{}
 	mrs := NewReadMetricsDecorator(&mockReader, mf)
-	mockReader.On("GetServices").Return([]string{}, nil)
+	mockReader.On("GetServices", context.Background()).Return([]string{}, nil)
```
yurishkuro: how did this ever work before?

cmd/query/app/handler_archive_test.go
vprithvi: It's an implementation of `context.Context`, I used the implementation because mock.AnythingOfType doesn't appear to work with interfaces. (See https://github.com/stretchr/testify/issues/519)

storage/spanstore/metrics/decorator_test.go
vprithvi: My hypothesis is that the field is ret reflectively and the prior absence of this field in both the `On` invocation and the generated code failed silently. (I don't understand how the mock could satisfy the reader interface though.)

Issue #1249, opened 2018-12-13
Improve error reporting in the API


yurishkuro:
The message says "and 1 failure" but does not specify what it was.

![image](https://user-images.githubusercontent.com/3523016/49959795-97edbb80-fedc-11e8-91f2-213895835eb1.png)

vprithvi:
I looked into this - the query object only contains the string and not the failure. It seems that Cassandra logs might have the failure. 

I'll put out a diff to log the query

yurishkuro:
string would be enough if it actually contained the error. We're currently swallowing the error message somewhere.

vprithvi:
>We're currently swallowing the error message somewhere.

The error message is "Operation failed - received 0 responses and 1 failures"

PR #1250, opened 2018-12-13
Add C* query to error logs

vprithvi:
- Determine which C* caused a read to fail

- Add query details to the log. Gocql should log these as `[query statement=%q values=%+v consistency=%s]`


vprithvi:
Relates to #1249

vprithvi:
>isn't the problem here that the error isn't being percolated all the way back to the UI?

I don't think so - the error message from C*, as seen in #1249 is `Operation failed - received 0 responses and 1 failures`. As far as I can tell, there is no additional information on the query object in  the version of gocql we are using.

black-adder:
> > isn't the problem here that the error isn't being percolated all the way back to the UI?
> 
> I don't think so - the error message from C*, as seen in #1249 is `Operation failed - received 0 responses and 1 failures`. As far as I can tell, there is no additional information on the query object in the version of gocql we are using.

Odd, no where in the gocql codebase can I see how the error `Operation failed - received 0 responses and 1 failures` could be generated. That aside, do we want to wrap the error that we return from the cassandra writer with the error message in the gocql query object?

vprithvi:
>Odd, no where in the gocql codebase can I see how the error Operation failed - received 0 responses and 1 failures could be generated.
This is because it is generated on Cassandra and set to jaeger-query via the gocql error object. `gocql` is simply the conduit.

plugin/storage/cassandra/spanstore/reader.go
```diff
@@ -410,7 +410,8 @@ func (s *SpanReader) executeQuery(span opentracing.Span, query cassandra.Query,
 	tableMetrics.Emit(err, time.Since(start))
 	if err != nil {
 		logErrorToSpan(span, err)
-		s.logger.Error("Failed to exec query", zap.Error(err))
+		span.LogFields(otlog.String("query", query.String()))
+		s.logger.Error("Failed to exec query", zap.Error(err), zap.String("query", query.String()))
```
yurishkuro: `zap.String("query", query.String())` => `zap.Stringer("query", query)`

Issue #1200, opened 2018-11-19
Ingester panics if partition is reassigned


lw346:
## Requirement - what kind of business use case are you trying to solve?

Ingester should not fall over when partition is reassigned to another consumer.

## Problem - what in Jaeger blocks you from solving the requirement?

There is an existing `jaeger-spans` queue (using protobuf encoder), with 6 partitions across 6 kafka nodes.

When running with multiple ingesters (v1.8.0), each ingester in turn will crash with logs that look like the following:

```{"level":"info","ts":1542640177.67382,"caller":"healthcheck/handler.go:99","msg":"Health Check server started","http-port":14270,"status":"unavailable"}
{"level":"info","ts":1542640177.7420416,"caller":"cassandra/factory.go:92","msg":"Cassandra archive storage configuration is empty, skipping"}
{"level":"info","ts":1542640177.7505202,"caller":"ingester/main.go:100","msg":"Registering metrics handler with HTTP server","route":"/metrics"}
{"level":"info","ts":1542640177.7506182,"caller":"ingester/main.go:106","msg":"Starting HTTP server","http-port":14271}
{"level":"info","ts":1542640177.7506611,"caller":"healthcheck/handler.go:133","msg":"Health Check state change","status":"ready"}
{"level":"info","ts":1542640177.7506845,"caller":"consumer/consumer.go:76","msg":"Starting main loop"}
{"level":"info","ts":1542640221.4679706,"caller":"consumer/consumer.go:154","msg":"Starting error handler","partition":4}
{"level":"info","ts":1542640221.4680126,"caller":"consumer/consumer.go:154","msg":"Starting error handler","partition":5}
{"level":"info","ts":1542640221.4679477,"caller":"consumer/consumer.go:105","msg":"Starting message handler","partition":5}
{"level":"info","ts":1542640221.4680393,"caller":"consumer/consumer.go:105","msg":"Starting message handler","partition":4}
{"level":"info","ts":1542640221.4846787,"caller":"consumer/processor_factory.go:62","msg":"Creating new processors","partition":5}
{"level":"info","ts":1542640221.4894123,"caller":"consumer/processor_factory.go:62","msg":"Creating new processors","partition":4}
{"level":"info","ts":1542640227.298487,"caller":"consumer/consumer.go:163","msg":"Finished handling errors","partition":5}
{"level":"info","ts":1542640227.3106964,"caller":"consumer/consumer.go:163","msg":"Finished handling errors","partition":4}
[...]

PR #1254, opened 2018-12-14
[WIP] Change active partitions counter to gauge

yurishkuro:
To address #1200, but needs more work

### DIFF
--- a/cmd/ingester/app/consumer/consumer.go
+++ b/cmd/ingester/app/consumer/consumer.go
@@ -36,7 +36,7 @@ type Params struct {
 	DeadlockCheckInterval time.Duration
 }
 
-// Consumer uses sarama to consume and handle messages from kafka
+// Consumer uses Sarama to consume and handle messages from kafka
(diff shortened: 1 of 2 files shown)


cmd/ingester/app/consumer/consumer.go
```diff
@@ -85,7 +85,8 @@ func (c *Consumer) Start() {
 				p.wg.Wait()
 			}
 			c.partitionIDToState[pc.Partition()] = &consumerState{partitionConsumer: pc}
-			c.partitionIDToState[pc.Partition()].wg.Add(2)
+			c.partitionIDToState[pc.Partition()].wg.Add(2) // we will be adding 2 goroutines
+			c.partitionsHeld.Update(int64(len(c.partitionIDToState)))
```
yurishkuro: @vprithvi this wouldn't quite work because we never seem to delete things from this map.

yurishkuro:
being done in #1485

PR #1256, opened 2018-12-16
Add Zipkin Thrift as kafka ingestion format

geobeau:
- Ingesting Zipkin thrift span from Kafka: This is useful for a seamless migration from Zipkin. It is mentioned in #171

- Add an unmarshaller to process spans in zipkin thrift using what was used for collector zipkin API

### DIFF
--- a/plugin/storage/kafka/unmarshaller.go
+++ b/plugin/storage/kafka/unmarshaller.go
@@ -21,9 +21,9 @@ import (
 	"github.com/gogo/protobuf/jsonpb"
 	"github.com/gogo/protobuf/proto"
 
-	"github.com/jaegertracing/jaeger/thrift-gen/zipkincore"
-	"github.com/jaegertracing/jaeger/model/converter/thrift/zipkin"
 	"github.com/jaegertracing/jaeger/model"
+	"github.com/jaegertracing/jaeger/model/converter/thrift/zipkin"
+	"github.com/jaegertracing/jaeger/thrift-gen/zipkincore"
 )
 
 // Unmarshaller decodes a byte array to a span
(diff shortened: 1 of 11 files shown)


plugin/storage/kafka/unmarshaller.go
```diff
@@ -57,3 +60,34 @@ func (h *JSONUnmarshaller) Unmarshal(msg []byte) (*model.Span, error) {
...
+	zs := &zipkincore.Span{}
+	if err = zs.Read(transport); err != nil {
+		return nil, err
+	}
+	mSpan, err := zipkin.ToDomainSpan(zs)
+
+	if err != nil {
+		return nil, err
+	}
+	return mSpan[0], err
```
geobeau: @yurishkuro I am not sure it's some good code, any advices to avoid using [0]. I thought about changing the interface to return []model.Span and handle the array in the calling function.

geobeau:
Formating done, now going for tests

plugin/storage/kafka/unmarshaller.go
yurishkuro: I think the ingester is currently built with the assumption that a single Kafka message contains just a single span. Even in this case a single Zipkin span can sometimes be transformed to two Jaeger spans, so it would be better to extend the ingester to support the notion of many spans per message, at least after unmarshaling. However, I would not recommend doing it in this PR, it should be a separate PR as it might get quite large. For now I would suggest logging a warning.

There are good reasons to expect a single span per Kafka message, it makes the streaming jobs easier to write.

geobeau: Thanks

geobeau:
Yes, I tested it mostly that way. It has not run for a long continuous time thought. And I am not sure we are using a clean version of Zipkin. I should probably build a small PoC using docker images to test better

PR #1260, opened 2018-12-17
Fix tools pruned by dep

isaachier:
Signed-off-by: Isaac Hier <isaachier@gmail.com>

## Which problem is this PR solving?
- Resolves #1258.

## Short description of the changes
- Use [required](https://golang.github.io/dep/docs/Gopkg.toml.html#required) list to maintain tool dependencies.

PR #1263, opened 2018-12-25
[WIP] - Added missing imports and prune options for make proto to work

olivierboucher:

Could not get `make proto` to work without theses changes even though #1260 was merged.

- Adds missing imports
- Adds missing prune settings for imported proto files to be present

I will eventually commit the proto changes related to #1214  since we wanted to have a second PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -323,9 +323,11 @@ PROTO_INCLUDES := \
 	-I model/proto \
 	-I vendor/github.com/grpc-ecosystem/grpc-gateway \
 	-I vendor/github.com/gogo/googleapis \
+	-I vendor/github.com/gogo/protobuf/protobuf \
 	-I vendor/github.com/gogo/protobuf
 # Remapping of std types to gogo types (must not contain spaces)
 PROTO_GOGO_MAPPINGS := $(shell echo \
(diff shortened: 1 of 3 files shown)


yurishkuro:
@olivierboucher when you say "could not make `make proto` work", could you provide more details?

I just ran it and it seems working fine, with the exception of a single test file that got changed (#1266).

yurishkuro:
I see #1264, so I guess `google/protobuf/descriptor.proto: File not found` is the issue?

Makefile
```diff
@@ -323,9 +323,11 @@ PROTO_INCLUDES := \
 	-I model/proto \
 	-I vendor/github.com/grpc-ecosystem/grpc-gateway \
 	-I vendor/github.com/gogo/googleapis \
+	-I vendor/github.com/gogo/protobuf/protobuf \
 	-I vendor/github.com/gogo/protobuf
 # Remapping of std types to gogo types (must not contain spaces)
 PROTO_GOGO_MAPPINGS := $(shell echo \
+        Mgoogle/protobuf/descriptor.proto=github.com/gogo/protobuf/types, \
```
yurishkuro: whitespace issue

pavolloffay:
I think this is already done, make proto works (at least for me locally)

PR #1270, opened 2018-12-29
ADD http reporter in agent

XiaoZYang:
## Feature Background
We build our tracing system based on `jaeger` in our company, but we leverage the cloud service to collect Spans instead of maintaining the component **Collector** in the self-built data center. However, the service provided by the cloud vendor supports HTTP endpoint only, and there are other services running on the cloud which means we cannot simply migrate to another vendor's service. So we decide to extend the agent.

- The PR enables the agent to report Spans to the collector

The changes are including:
  - implementing the interface of `Reporter`
 - implementing the interface of `ClientConfigManager`
- modifying the config of reporter flags
- modifying the start script of the agent, i.e. `cmd/agent/main.go`

### DIFF
--- a/cmd/agent/app/configmanager/http/manager.go
+++ b/cmd/agent/app/configmanager/http/manager.go
@@ -1,0 +1,27 @@
+package http
+
+import (
+	"errors"
+
+	"github.com/jaegertracing/jaeger/thrift-gen/baggage"
+	"github.com/jaegertracing/jaeger/thrift-gen/sampling"
+)
+
+// SamplingManager returns sampling decisions from collector over HTTP.
+type SamplingManager struct {
+}
+
+// NewConfigManager creates HTTP sampling manager.
+func NewConfigManager(endpoint string) *SamplingManager {
+	return &SamplingManager{}
+}
+
+// GetSamplingStrategy returns sampling strategies from collector.
+func (s *SamplingManager) GetSamplingStrategy(serviceName string) (*sampling.SamplingStrategyResponse, error) {
+	return nil, errors.New("sampling strategy not implemented")
+}
+
+// GetBaggageRestrictions returns baggage restrictions from collector.
+func (s *SamplingManager) GetBaggageRestrictions(serviceName string) ([]*baggage.BaggageRestriction, error) {
+	return nil, errors.New("baggage not implemented")
+}
--- a/cmd/agent/app/configmanager/http/manager_test.go
+++ b/cmd/agent/app/configmanager/http/manager_test.go
@@ -1,0 +1,17 @@
+package http
+
+import (
+	"testing"
+)
+
+func TestNewConfigManager(t *testing.T) {
+	// TODO(xiaozongyang): add unit test for function NewConfigManager
+}
+
+func TestGetSamplingStrategy(t *testing.T) {
+	// TODO(xiaozongyang): add unit test for method GetSamplingStrategy
+}
+
+func TestGetBaggageRetrictions(t *testing.T) {
+	// TODO(xiaozongyang): add unit test for method GetBaggageRestrictions
+}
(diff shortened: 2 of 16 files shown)


yurishkuro:
@XiaoZYang thank you for the PR.  Before reviewing, I would like to understand your use case better.

* To confirm: the cloud provider is running a collector service that accepts Jaeger Thrift data over HTTP?
* To confirm: I assume they only support HTTP because it can be used with a TLS, while TChannel cannot?

If my assumptions above are correct and there are no other reasons to use HTTP, how about the new gRPC endpoint that Jaeger is going to support soon (the code is already in master)? It would address the TLS question, it is HTTP under the hood, and it is our strategic direction to replace TChannel/Thrift with gRPC/Protobuf.

XiaoZYang:
@yurishkuro thanks for you replying.
1. It's confirmed that the cloud provider is running a collector service that accepts Jaeger Thrift data over HTTP. I've communicated with the consult of the cloud provider and tried it myself, their service accepts Jaeger Thrift data or Zipkin Thrift data over HTTP only, gRPC and TChannel are unsupported now.
2. IMO, the reason why they only support HTTP may be that it's easy for Jaeger client and Zipkin client to integrate the cloud service, most use cases are all services with instrumented libs running on their cloud platform, so security may not be an critical issue.

yurishkuro:
@XiaoZYang 

> the reason why they only support HTTP may be that it's easy for Jaeger client and Zipkin client to integrate the cloud service

Agreed, but your PR is not about Jaeger clients (which already support http transport), it's about changing the agent.

> their service accepts Jaeger Thrift data or Zipkin Thrift data over HTTP only, gRPC and TChannel are unsupported now.

Do you know if they are actually running Jaeger collectors or their own software that accepts Jaeger format? What I was proposing is that once we officially release gRPC support for agent->collector communications, your provider can upgrade to the new Jaeger version, and we don't need to build this redundant HTTP reporter into the agent (since from infrastructure support point of view, gRPC is HTTP).

XiaoZYang:
@yurishkuro
> Agreed, but your PR is not about Jaeger clients (which already support http transport), it's about changing the agent.

Oh, I'm sorry about missing some important details about our use cases. 
1. We integrate jaeger with the [java-client](https://github.com/jaegertracing/jaeger-client-java), i.e., our java services report Span to the agent
2. The agent sends Jaeger Span to the collector service provided by the cloud vendor.
We want to deploy a side-car agent along with each bussiness service because we don't want to upgrade the application service when the tracing service upgraded every time. So we try to extend the jaeger agent rather than the jaeger client.

> Do you know if they are actually running Jaeger collectors or their own software that accepts Jaeger format? What I was proposing is that once we officially release gRPC support for agent->collector communications, your provider can upgrade to the new Jaeger version, and we don't need to build this redundant HTTP reporter into the agent (since from infrastructure support point of view, gRPC is HTTP).

Thanks for your suggestions! It seems that they run their own software which is compatible with Jaeger API, I will re-confirm this with the cloud vendor later. Furthermore, about the gRPC, it's on their to-do list but there is not a clear schedule about gRPC support. So we plan to apply tracing cloud service now by HTTP and migrate to gRPC once they can accept gRPC requests.

XiaoZYang:
@yurishkuro 
1. It's confirmed that the vendor writes their own collector service rather than leveraging jaeger collector.
2. 
> You could do what we at Uber do internally: we build our own version of the agent with a custom `main()`, which allows us to wire some of the components from the official jaeger repo differently. This way you can implement this http reporter as your internal component and wire it in from the custom main(), until your vendor supports Jaeger's gRPC API.

Actually, we plan to implement it as you mentioned. If there are other users meet the same problem, maybe we can contribute this to the community. But if it's a rare use case and there is no need to provide HTTP support of agent in jaeger official repository, I'll close this PR.

Thanks for your help and suggestions again @yurishkuro 👍

PR #1274, opened 2019-01-04
Fix nocover target to fail on directories without coverage

yurishkuro:
The `nocover` target was supposed to fail on directories without coverage, but the exit code was ignored due to piping the results to colorize.

* moved colorize inside the script
* verified that the script now fails
* added empty_test.go to all directories that were missing it

NB: perhaps a better alternative approach would be to automatically create dummy test files during the CI build


Makefile
```diff
@@ -110,7 +109,7 @@ cover: nocover
 .PHONY: nocover
 nocover:
 	@echo Verifying that all packages have test files to count in coverage
-	@scripts/check-test-files.sh $(subst github.com/jaegertracing/jaeger/,./,$(ALL_PKGS)) | $(SED) ''/FIXME/s//$(COLOR_FIXME)/''
```
yurishkuro: the pipe was suppressing the exit status

cmd/ingester/app/builder/.nocover
```diff
@@ -0,0 +1 @@
+FIXME - refactor to avoid direct Kafka dependency. Cf. collector's builder, which is testable
```
yurishkuro: @vprithvi fyi

pkg/es/wrapper/wrapper.go
```diff
@@ -12,12 +12,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package es
+package eswrapper
```
yurishkuro: moved these files to `wrapper` package, which makes it more obvious why coverage needs to be suppressed. The parent package is not testable (even though it only has type defs and thus an `empty_test.go` file).

PR #1278, opened 2019-01-09
Fix build

pavolloffay:
I had to regenerate proto to pass the build locally. I have also added `.nocover` to query's statik dir - usually I have compiled UI and the build would fail because of not covered files.


yurishkuro:
@pavolloffay you may be using different dependencies if you had to regenerate proto types. When I try this on master:
```shell
dep ensure
make proto-install
make proto
```
I get no differences from committed versions. The only variable in the above is the version of protoc itself (I have `/Users/yurishkuro/bin/protoc@ -> protoc-3.6.1`), but I don't think that affects the generated code.

pavolloffay:
Yes, it was a problem with my vendor deps. I would like to still exclude statik from coveragehttps://github.com/jaegertracing/jaeger/pull/1279

PR #1279, opened 2019-01-10
Exclude query/statik package from coverage

pavolloffay:
Supersedes #1278 


pavolloffay:
I will merge this as it's just a build configuration. If you have any comments please comment I will fix in follow up PRs.

PR #1283, opened 2019-01-15
Add CLI configurable MaxNumSpans while retrieving spans from ES.

annanay25:

- Follow up PR for #1202 

- Added a CLI configurable option to limit the number of spans fetched in a single query to the ES backend.

### DIFF
--- a/pkg/es/config/config.go
+++ b/pkg/es/config/config.go
@@ -39,6 +39,7 @@ type Configuration struct {
 	Username          string
 	Password          string
 	Sniffer           bool          // https://github.com/olivere/elastic/wiki/Sniffing
+	MaxNumSpans       int           // defines maximum number of spans to fetch from storage per query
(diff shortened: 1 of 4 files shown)


black-adder:
My only question is are we ok with changing behavior that people are used to? If people upgrade jaeger, they'll automatically be hit with the 10000 limit and not know about it. Perhaps if no limit is set, we don't apply a default limit?

annanay25:
@black-adder; this makes sense. 

However, we could mention the reasons for adopting such a limit in the docs (that it could overwhelm the query-service) and mention that it could be overwritten using the flag in the release notes.

yurishkuro:
please add to the changelog.

is the behavior actually different from before? I thought we already had another 10k limit in the code.

yurishkuro:
This is being reverted in https://github.com/jaegertracing/jaeger/pull/4336

PR #1285, opened 2019-01-16
Use DCAwareRoundRobinPolicy as fallback for TokenAwarePolicy

vprithvi:
- Previously, RoundRobinPolicy would attempt cross-dc writes

- When using a multi-datacenter Cassandra deploy, jaeger choses hosts in remote data centers. This is problematic for a variety of reasons
- When configured with only local replicas, remote reads/writes fail entirely
- When the Cassandra cluster is configured with multi zone replication, picking remote data centers results in lower performance

- Use `DCAwareRoundRobin` as a fallback for `TokenAwareHostPolicy`


pkg/cassandra/config/config.go
```diff
@@ -130,7 +131,13 @@ func (c *Configuration) NewCluster() *gocql.ClusterConfig {
 	} else {
 		cluster.Consistency = gocql.ParseConsistency(c.Consistency)
 	}
-	cluster.PoolConfig.HostSelectionPolicy = gocql.TokenAwareHostPolicy(gocql.RoundRobinHostPolicy())
+
+	fallbackHostSelectionPolicy := gocql.RoundRobinHostPolicy()
+	if c.LocalDC != "" {
+		fallbackHostSelectionPolicy = gocql.DCAwareRoundRobinPolicy(c.LocalDC)
+	}
+	cluster.PoolConfig.HostSelectionPolicy = gocql.TokenAwareHostPolicy(fallbackHostSelectionPolicy, gocql.ShuffleReplicas())
```
black-adder: what does `gocql.ShuffleReplicas()` do?

vprithvi: There can be multiple replicas (equal to RF) for a given token. `ShuffleReplicas` selects one of these replicas at random instead of deterministically selecting a replica.

plugin/storage/cassandra/options.go
```diff
@@ -148,6 +149,10 @@ func addFlags(flagSet *flag.FlagSet, nsConfig *namespaceConfig) {
 		nsConfig.namespace+suffixKeyspace,
 		nsConfig.Keyspace,
 		"The Cassandra keyspace for Jaeger data")
+	flagSet.String(
+		nsConfig.namespace+suffixDC,
+		nsConfig.LocalDC,
+		"The Cassandra local data center for DC Aware host selection")
```
yurishkuro: The Cassandra **name** of local data center...

PR #1290, opened 2019-01-18
WIP: Allow configuring es index rotation frequency

### LINKED ISSUE #628: Make default time span of ES index configurable

annanay25:
- Resolves #628 

- Added a CLI flag to configure the format of the ES span index, which in turn governs the frequency of index rotation.

### DIFF
--- a/pkg/es/config/config.go
+++ b/pkg/es/config/config.go
@@ -49,6 +49,7 @@ type Configuration struct {
 	BulkActions       int
 	BulkFlushInterval time.Duration
 	IndexPrefix       string
+	IndexTimeSpan     string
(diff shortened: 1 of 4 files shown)


pavolloffay:
Thanks for the PR, however we don't want to go this direction. The reasons are explained in #840. Instead we will go with rollover API

annanay25:
This change is [Reviewable](https://reviewable.io/reviews/jaegertracing/jaeger/1290)

PR #1298, opened 2019-01-24
Fix duplicated span when querying Elasticsearch  [#1297]

### LINKED ISSUE #1297: querying data repeatedly makes duplicated span on jaeger-ui
When using the  search function of jaeger-query, if the data to be queried is very large, the method multiRead in "github.com/jaegertracing/jaeger/plugin/storage/es/spanstore/reader.go" will query the storage more than once. There is a bug in this method that causes queried data may be repeated probably.
[https://github.com/jaegertracing/jaeger/blob/master/plugin/storage/es/spanstore/reader.go](url)

default
The array of searchRequests is been maked  Outside the for loop，the length of it equals the slice of traceIds.  When  it query second time in the for loop，the length of traceIds will be changed，but the length of searchRequests  will not. This bug may cause repeated traceId be queried.like this:
20190124-200601 espace
There is #duplicated span in this trace

Maybe  we can initial searchRequests in for loop，the length of it will equals the slice of traceIds Always.

Volcannozzz:
Resolves #1297

### DIFF
--- a/plugin/storage/es/spanstore/reader.go
+++ b/plugin/storage/es/spanstore/reader.go
@@ -278,7 +278,7 @@ func (s *SpanReader) multiRead(ctx context.Context, traceIDs []string, startTime
 	if len(traceIDs) == 0 {
 		return []*model.Trace{}, nil
 	}
-	searchRequests := make([]*elastic.SearchRequest, len(traceIDs))
+	
 
 	var traces []*model.Trace
 	// Add an hour in both directions so that traces that straddle two indexes are retrieved.
@@ -293,7 +293,7 @@ func (s *SpanReader) multiRead(ctx context.Context, traceIDs []string, startTime
 		if len(traceIDs) == 0 {
 			break
 		}
-
+		searchRequests := make([]*elastic.SearchRequest, len(traceIDs))
 		for i, traceID := range traceIDs {
 			query := elastic.NewTermQuery("traceID", traceID)
 			if val, ok := searchAfterTime[traceID]; ok {


plugin/storage/es/spanstore/reader.go
```diff
@@ -293,7 +293,7 @@ func (s *SpanReader) multiRead(ctx context.Context, traceIDs []string, startTime
 		if len(traceIDs) == 0 {
 			break
 		}
-
+		searchRequests := make([]*elastic.SearchRequest, len(traceIDs))
```
black-adder: this isn't the most efficient fix, ideally we alloc just once and play around with the indices rather than alloc each iteration. We can leave a TODO for now.

black-adder:
please see https://github.com/jaegertracing/jaeger/blob/master/CONTRIBUTING_GUIDELINES.md#certificate-of-origin---sign-your-work to pass the DCO check

plugin/storage/es/spanstore/reader.go
Volcannozzz: Good suggestion and we(@crazycomputer) have fixed it .We will test it in our project tommorrow and if success,i will make a standard PR again.

Volcannozzz: > this isn't the most efficient fix, ideally we alloc just once and play around with the indices rather than alloc each iteration. We can leave a TODO for now.

Good suggestion and we(@crazycomputer) have fixed it .We will test it in our project tommorrow and if success,i will make a standard PR again.

Volcannozzz: > this isn't the most efficient fix, ideally we alloc just once and play around with the indices rather than alloc each iteration. We can leave a TODO for now.

Good suggestion.We(@crazycomputer) have fixed it and will test it in our project tommorrow.If success,i will make a standard PR again.

yurishkuro:
Please fix the PR title to reflect the change being made (what it fixes goes into description). PR title typically goes into the final commit comment, and should follow the guidelines in CONTRIBUTING.

pavolloffay:
done in #1677 

thanks @Volcannozzz

PR #1301, opened 2019-01-28
let cassandra servers contain whitespace in config

### LINKED ISSUE #1300: Strip whitespace from value for servers in configmap
## Requirement 
Listing the IP addresses of each node of a Cassandra cluster.

In the **configmap.yml** file, the key-value pair of _servers: XX.XX.XXX.XX_ for one IP entry works fine. But when providing more than one entry, like _servers: XX.XX.XXX.XX, XX.XXX.XX.XXX, XX.XXX.XXX.XX_, the Collector's *gocql* reveals an error for the second IP: _dns error: lookup XX.XXX.XX.XXX: no such host_ and reveals the same error for the third IP.

When I remove the whitespace between the comma and the next IP, the thing works fine and the Collector successfully connects to the Cassandra cluster.

Strip white spaces from the value, so that humans can still use whitespaces when writing and reading these values.

Will you please include this type of Cassandra example in your documentation? Something beyond the all-in-one example which is too simple for production scenarios. I couldn't find anything about clusters and raw IP addresses for the cluster.

karlpokus:
- Resolves #1300 

- Let cassandra hosts contain whitespace in config


black-adder:
Cheers!

jpkrohling:
@karlpokus how do you feel about sending your second PR to this project? #1302 seems easy now that you did this one, doesn't it ;-)

karlpokus:
@jpkrohling Sure. I'll take a look tomorrow.

Issue #1302, opened 2019-01-28
Strip whitespace from value for ES servers and kafka brokers


black-adder:
## Requirement - what kind of business use case are you trying to solve?
See https://github.com/jaegertracing/jaeger/issues/1300 for ES servers and kafka brokers

## Problem - what in Jaeger blocks you from solving the requirement?
https://github.com/jaegertracing/jaeger/blob/bfa7b785e2dbb7d4fa55f2a51915b095a2fa928c/plugin/storage/es/options.go#L221 and https://github.com/jaegertracing/jaeger/blob/bfa7b785e2dbb7d4fa55f2a51915b095a2fa928c/plugin/storage/kafka/options.go#L77

## Proposal - what do you suggest to solve the problem or improve the existing situation?
Same solution as in #1301 

## Any open questions to address
Perhaps come up with a general solution for this? Maybe `x/config/viper.go` that wraps viper but with an extra function `GetStrings` that handles the removal of whitespaces? Overkill?

jpkrohling:
> Perhaps come up with a general solution for this?

You mean, to be applied automatically to all properties, or as a helper function? There might be cases where spaces should be part of the value, but we could apply this to properties where the input is well-defined, like IP addresses/hostnames.

black-adder:
a helper function so as to not to copy `stripWhitespace` function everywhere.

annanay25:
@black-adder was this issue fixed with #1305?

pavolloffay:
Yeah I think it was, please reopen if it still has to be solved.

PR #1306, opened 2019-01-29
add shared helper for stripping whitespace of storage host strings

karlpokus:
Signed-off-by: Karl Pokus <karl.pokus@gmail.com>

- #1302 

- shared helper for stripping whitespace of storage host strings

Not sure of the quality of the solution. It works and the tests pass. Didn't have time to wrap the viper lib with the helper like was suggested in #1302.


black-adder:
Thanks! I went ahead and landed the other change; in the future I'd like to add a function to viper but for now, some copy paste is fine.

PR #1307, opened 2019-01-29
Support gRPC for query service

annanay25:
- Initial work to add GRPC support for the `query-service`.

- Use grpc-gateway to multiplex HTTP and GRPC on the same port
- Create GRPC server and handler for `GetTrace()`


cmd/query/app/grpc_handler.go
```diff
@@ -0,0 +1,82 @@
...
+	"google.golang.org/grpc"
+
+	"github.com/jaegertracing/jaeger/proto-gen/api_v2"
+	"github.com/jaegertracing/jaeger/storage/dependencystore"
+	"github.com/jaegertracing/jaeger/storage/spanstore"
+)
+
+// GRPCHandler implements the GRPC endpoint of the query service.
+type GRPCHandler struct {
+	spanReader        spanstore.Reader
```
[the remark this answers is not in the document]
yurishkuro: +1

cmd/query/main.go
```diff
@@ -141,11 +144,36 @@ func main() {
...
 			recoveryHandler := recoveryhandler.NewRecoveryHandler(logger, true)
 
+			// Create GRPC Server.
+			grpcServer := grpc.NewServer()
+
+			// Add logging.
+			grpclog.SetLoggerV2(grpclog.NewLoggerV2(ioutil.Discard, os.Stderr, os.Stderr))
+
+			// Register GRPC Handler.
+			spanReaderGRPC, err := storageFactory.CreateSpanReader()
```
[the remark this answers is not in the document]
annanay25: The PR is here - #1312 

Will rebase this once merged.

cmd/query/main.go
```diff
@@ -141,11 +144,36 @@ func main() {
...
+			if err != nil {
+				logger.Fatal("Failed to create span reader", zap.Error(err))
+			}
+			spanReaderGRPC = storageMetrics.NewReadMetricsDecorator(spanReaderGRPC, baseFactory.Namespace(metrics.NSOptions{Name: "query-grpc", Tags: nil}))
+			dependencyReaderGRPC, err := storageFactory.CreateDependencyReader()
```
annanay25: Should we be creating different instances of the `spanReader` and `dependencyReader` for use with GRPC? I thought to do so in the case that a user might want to use both HTTP and GRPC endpoints at the same time (possibly HTTP for the UI and a custom GRPC querier), but I'm not so sure that's required.

model/proto/api_v2.proto
```diff
@@ -64,6 +64,22 @@ service CollectorService {
     }
 }
 
+message GetTraceRequest {
+    string Id = 1;
```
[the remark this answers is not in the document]
annanay25: We can accept a `model.TraceID`, I'll push the change. Thanks!

cmd/query/main.go
yurishkuro: Should not be creating any storage components, they are already passed to QueryService.

model/proto/api_v2.proto
```diff
@@ -64,6 +64,22 @@ service CollectorService {
...
+}
+
+message GetTraceResponse {
+    Trace trace = 1;
+}
```
yurishkuro: check the discussion in https://github.com/jaegertracing/jaeger/pull/1323. We have an internal grpc implementation of GetTrace, and our users have run into the issue of large traces exceeding grpc default message size. So I think we need to go with streaming APIs for methods that can return large data sets.

I am not sure what it would do to grpc_gateway, how does it map streaming calls to HTTP?

Also, it's been a while since the last time I checked, but if gRPC for JavaScript in the browser has advanced enough, then we won't even need grpc_gateway.

annanay25: > I am not sure what it would do to grpc_gateway, how does it map streaming calls to HTTP?

Seems like they have an open issue for sending chunked JSON responses - grpc-ecosystem/grpc-gateway/issues/562

> Also, it's been a while since the last time I checked, but if gRPC for JavaScript in the browser has advanced enough, then we won't even need grpc_gateway.

Apart from streaming-support (which is on their [roadmap](https://github.com/grpc/grpc-web/blob/master/ROADMAP.md#streaming-friendly-transport-implementation)) what other features would we need in the grpc-web library?

annanay25: Update: Went through the grpc-web readme again, and it mentions that server-side streaming is already supported. This makes it a candidate to be considered for the query UI.

yurishkuro: Great! We can always put a blocking/sync facade in front of grpc service if streaming to web is not supported, but since it's there already, we should go with streaming for the main API.

annanay25:
I would like a second review round @yurishkuro 

Changes in the last commit - 
- Dropped `grpc-gateway`
- Used cmux to multiplex HTTP and GRPC
- Used stream response for `GetTraceRequest`

Todo - 
- Add all endpoints including `FindTraceIDs` etc.
- Add tests.

model/proto/api_v2.proto
```diff
@@ -64,6 +64,28 @@ service CollectorService {
...
+message GetTraceResponseStream {
+  repeated jaeger.api_v2.Span spans = 1 [
+    (gogoproto.nullable) = false
+  ];
+}
+
+service QueryService {
+    rpc GetTrace(GetTraceRequest) returns (stream GetTraceResponseStream) {
+        option (google.api.http) = {
+            post: "/api/v2/traces/{trace_id}"
```
annanay25: This doesn't seem right to me. The `trace_id` should either be part of the body like so - 

```
service QueryService {
    rpc GetTrace(GetTraceRequest) returns (stream GetTraceResponseStream) {
        option (google.api.http) = {
            post: "/api/v2/traces"
            body: "*"
        };
    }
}
```

...or, we could make it a string parameter.

```
message GetTraceRequest {
  string trace_id = 1 [
    (gogoproto.nullable) = false,
  ];
}
```

yurishkuro: if it is `bytes` as in the main API, then
  * how would swagger file describe the parameter?
  * how would it look on the wire?

I don't have a strong opinion about post vs. get here. Get seems more REST-ful.

annanay25: Changing it to bytes and generating swagger documentation gives - 
```
.
.
    "/api/v2/traces": {
      "post": {
        "operationId": "GetTrace",
        "responses": {
          "200": {
            "description": "(streaming responses)",
            "schema": {
              "$ref": "#/definitions/api_v2GetTraceResponseStream"
            }
          }
        },
        "parameters": [
          {
            "name": "body",
            "in": "body",
            "required": true,
            "schema": {
              "$ref": "#/definitions/api_v2GetTraceRequest"
            }
          }
        ],
        "tags": [
          "QueryService"
        ]
      }
    }
  },
.
.
  "definitions": {
    "api_v2GetTraceRequest": {
      "type": "object",
      "properties": {
        "trace_id": {
          "type": "string",
          "format": "byte"
        }
      }
    },
.
.
```

annanay25: https://swagger.io/docs/specification/data-models/data-types/#string

`"type": "string","format": "byte"` would mean base64 encoded traceID on the wire

yurishkuro: I think it’s the same for ids in the returned span JSON, so sounds reasonable?

annanay25: +1

> I don't have a strong opinion about post vs. get here. Get seems more REST-ful.

You're right, I should change this to `get`.

model/proto/api_v2/query.proto
```diff
@@ -0,0 +1,71 @@
...
+  ];
+}
+
+message GetTraceResponseStream {
+  repeated jaeger.api_v2.Span spans = 1 [
+    (gogoproto.nullable) = false
+  ];
+}
+
+service QueryService {
```
yurishkuro: Note: while the APIs don't have to be identical, we should try to keep them very similar

https://github.com/jaegertracing/jaeger/pull/1323

annanay25: Okay so I can just implement this API? - 
https://github.com/jaegertracing/jaeger/blob/65a499a247e0e7d0f977f1898100e469975746ac/cmd/query/app/http_handler.go#L109-L120

model/proto/api_v2/query.proto
```diff
@@ -0,0 +1,136 @@
...
+    (gogoproto.customname) = "TraceId"
+  ];
+}
+
+message ArchiveTraceResponse {
+}
+
+// FIXME: Pass all TraceQueryParameters in SearchRequest?
+message SearchRequest {
+  string query = 1 [
```
annanay25: Question: Should parameters be parsed from a `string query` or should the `SearchRequest` have all TraceQueryParameters as in - https://github.com/jaegertracing/jaeger/blob/65a499a247e0e7d0f977f1898100e469975746ac/storage/spanstore/interface.go#L45-L54

I guess the former because protobuf supports just basic data types and not time/maps etc?

model/proto/api_v2/query.proto
yurishkuro: no, the original HTTP interface is not particularly relevant. I am referring to #1323 that defines the gRPC interface for storage plugins. It has very large overlap with the query API, so I would like them to be as close as possible, including the name of the methods and types.

model/proto/api_v2/query.proto
yurishkuro: it should be structured, as in #1323

annanay25: Thanks! I've made the changes.

model/proto/api_v2/query.proto
```diff
@@ -0,0 +1,154 @@
...
+
+message GetTraceRequest {
+  bytes trace_id = 1 [
+    (gogoproto.nullable) = false,
+    (gogoproto.customtype) = "github.com/jaegertracing/jaeger/model.TraceID",
+    (gogoproto.customname) = "TraceId"
+  ];
+}
+
+message GetTraceResponseStream {
```
yurishkuro: SpansResponseChunk https://github.com/jaegertracing/jaeger/pull/1323/files#diff-ac4433b4ef3a403411d1d1b5141eb9ccR98

model/proto/api_v2/query.proto
```diff
@@ -0,0 +1,154 @@
...
+        };
+    }
+
+    rpc ArchiveTrace(ArchiveTraceRequest) returns (ArchiveTraceResponse) {
+        option (google.api.http) = {
+            post: "/traces/{trace_id}"
+        };
+    }
+
+    rpc Search(FindTracesRequest) returns (FindTracesResponse) {
```
yurishkuro: `rpc FindTraces(FindTracesRequest) returns (stream SpansResponseChunk);`

https://github.com/jaegertracing/jaeger/pull/1323/files#diff-ac4433b4ef3a403411d1d1b5141eb9ccR126

model/proto/api_v2/query.proto
annanay25: Changed.

model/proto/api_v2/query.proto
annanay25: Made the changes. Thanks!

model/proto/api_v2/query.proto
```diff
@@ -0,0 +1,171 @@
...
+  repeated string operations = 1;
+}
+
+message GetDependenciesRequest {
+  google.protobuf.Timestamp start_time = 1 [
+    (gogoproto.stdtime) = true,
+    (gogoproto.nullable) = false
+  ];
+  google.protobuf.Duration lookback = 2 [
+    (gogoproto.stdtime) = true,
```
annanay25: A small mismatch with the plugin API here, as I think `google.protobuf.Duration lookback` better suits `querysvc` APIs. However, can change if required. @yurishkuro

yurishkuro: I think we intentionally changed it to start_time / end_time, as it has less cognitive overhead (no need to understand yet another type Duration).

cmd/query/main.go
```diff
@@ -116,18 +121,53 @@ func main() {
...
 
+			// Create HTTP Server
+			httpServer := &http.Server{
+				Handler: recoveryHandler(compressHandler),
+			}
+
+			// Create GRPC Server.
+			grpcServer := grpc.NewServer()
+
+			grpclog.SetLoggerV2(grpclog.NewLoggerV2(ioutil.Discard, os.Stderr, os.Stderr))
```
yurishkuro: this is already done in `svc.Start(v)`

cmd/query/main.go
```diff
@@ -116,18 +121,53 @@ func main() {
...
+				cmux.HTTP2HeaderField("content-type", "application/grpc"),
+				cmux.HTTP2HeaderField("content-type", "application/grpc+proto"))
+			httpL := s.Match(cmux.Any())
+
 			go func() {
 				logger.Info("Starting HTTP server", zap.Int("port", queryOpts.Port))
-				if err := http.ListenAndServe(portStr, recoveryHandler(compressHandler)); err != nil {
-					logger.Fatal("Could not launch service", zap.Error(err))
-				}
+				httpServer.Serve(httpL)
```
yurishkuro: any reason not to handle error as in the previous code?

cmd/query/main.go
```diff
@@ -116,18 +121,53 @@ func main() {
...
 				svc.HC().Set(healthcheck.Unavailable)
 			}()
 
+			go func() {
+				logger.Info("Starting GRPC server", zap.Int("port", queryOpts.Port))
+				grpcServer.Serve(grpcL)
```
yurishkuro: error handling?

cmd/query/main.go
```diff
@@ -116,18 +121,53 @@ func main() {
...
+				svc.HC().Set(healthcheck.Unavailable)
+			}()
+
+			// Start cmux server.
+			s.Serve()
```
yurishkuro: is this blocking operation? what about error handling?

model/proto/api_v2/query.proto
```diff
@@ -0,0 +1,171 @@
...
+service QueryService {
+    rpc GetTrace(GetTraceRequest) returns (SpansResponseChunk) {
+        option (google.api.http) = {
+            get: "/traces/{trace_id}"
+        };
+    }
+
+    rpc ArchiveTrace(ArchiveTraceRequest) returns (ArchiveTraceResponse) {
+        option (google.api.http) = {
+            post: "/traces/{trace_id}"
```
yurishkuro: /archive/

model/proto/api_v2/query.proto
```diff
@@ -0,0 +1,171 @@
...
+        };
+    }
+
+    rpc FindTraces(FindTracesRequest) returns (stream SpansResponseChunk) {
+        option (google.api.http) = {
+            get: "/search"
```
yurishkuro: how is the request encoded in case of GET?

model/prototest/model_test.pb.go
```diff
@@ -3,11 +3,9 @@
 
 package prototest
 
-import (
```
yurishkuro: this indicates that you're using old plugins. Run `make proto-install` then `make proto`

annanay25:
Thanks for the review @yurishkuro. I've addressed comments and updated.

annanay25:
Thanks for the changes @yurishkuro! 

About tests, should I use a mock query service (as in the current commit) or an instance of query service with mock span readers / dependency readers?

yurishkuro:
mock QA is fine

annanay25:
@yurishkuro could you please review this? I think the last remaining lines from coverage are stream sends, which I'm not sure how to test.

yurishkuro:
🎉 

There is a bit of a clean-up I'd like to do in the test, I think it's leaking goroutines, but didn't want to block the PR on that.

cmd/query/app/grpc_handler_test.go
```diff
@@ -0,0 +1,481 @@
...
+	"github.com/jaegertracing/jaeger/cmd/query/app/querysvc"
+	"github.com/jaegertracing/jaeger/model"
+	"github.com/jaegertracing/jaeger/proto-gen/api_v2"
+	depsmocks "github.com/jaegertracing/jaeger/storage/dependencystore/mocks"
+	"github.com/jaegertracing/jaeger/storage/spanstore"
+	spanstoremocks "github.com/jaegertracing/jaeger/storage/spanstore/mocks"
+)
+
+var (
+	grpcServerPort    = ":14251"
```
yurishkuro: we should be using :0 port

cmd/query/app/grpc_handler_test.go
```diff
@@ -0,0 +1,481 @@
...
+func newGRPCServer(t *testing.T, q *querysvc.QueryService, logger *zap.Logger, tracer opentracing.Tracer) (*grpc.Server, net.Addr) {
+	lis, _ := net.Listen("tcp", grpcServerPort)
+	grpcServer := grpc.NewServer()
+	grpcHandler := NewGRPCHandler(*q, logger, tracer)
+	api_v2.RegisterQueryServiceServer(grpcServer, grpcHandler)
+
+	go func() {
+		err := grpcServer.Serve(lis)
+		require.NoError(t, err)
+	}()
```
yurishkuro: goroutine leak?  Or will `defer server.Stop()` in the tests stop it?

cmd/query/app/grpc_handler_test.go
```diff
@@ -0,0 +1,481 @@
...
+	q, readStorage, dependencyStorage := newQueryService(querysvc.QueryServiceOptions{})
+	logger := zap.NewNop()
+	tracer := opentracing.NoopTracer{}
+
+	grpcServer, addr := newGRPCServer(t, q, logger, tracer)
+
+	return grpcServer, addr, readStorage, dependencyStorage
+}
+
+func initializeTestServerGRPCWithOptions(t *testing.T) (*grpc.Server, net.Addr, *spanstoremocks.Reader, *depsmocks.Reader, *spanstoremocks.Reader, *spanstoremocks.Writer) {
```
yurishkuro: a) any reason not to merge this with the above function?
b) instead of returning many objects it's better to return a single struct
c) to avoid server life cycle management, we typically do this
```go
func withGRPCServer(t *testing.T, actualTest func(server *server)) {
    server := ... create ...
    defer server.close() // other life cycle mgmt
    actualTest(server)
}
```

cmd/query/app/grpc_handler_test.go
```diff
@@ -0,0 +1,481 @@
...
+}
+
+func TestGetTraceSuccessGRPC(t *testing.T) {
+	server, addr, readMock, _ := initializeTestServerGRPC(t)
+	defer server.Stop()
+	readMock.On("GetTrace", mock.AnythingOfType("*context.valueCtx"), mock.AnythingOfType("model.TraceID")).
+		Return(mockTrace, nil).Once()
+
+	client, conn := newGRPCClient(t, addr)
+	defer conn.Close()
```
yurishkuro: similar to above, `withServerAndClient(t *testing.T, func actualTest(server *server, client *client))`

this way it would remove a lot of boilerplate code from the tests

cmd/query/main.go
```diff
@@ -116,14 +119,58 @@ func main() {
...
+
+			// Create cmux server.
+			// cmux will reverse-proxy between HTTP and GRPC backends.
+			s := cmux.New(conn)
+
+			// Add GRPC and HTTP listeners.
+			grpcL := s.Match(
+				cmux.HTTP2HeaderField("content-type", "application/grpc"),
+				cmux.HTTP2HeaderField("content-type", "application/grpc+proto"))
+			httpL := s.Match(cmux.Any())
```
yurishkuro: Booked #1482 to refactor this out of main

cmd/query/app/grpc_handler_test.go
yurishkuro: Booked #1483

annanay25:
Thanks @yurishkuro! 🎊 

I've been traveling but should get a chance to look at this soon :)

PR #1311, opened 2019-02-01
allow use of secure gRPC connections between agent and collector

### LINKED ISSUE #1310: only insecure gRPC connection from agent to collector possible
related to: https://github.com/jaegertracing/jaeger/issues/458

In our environment jaeger is deployed in k8s behind an nginx-ingress. The nginx-ingress only supports http based traffic (expect when using TCP port mappings which comes with some caveats...). Thats why we setup the agents (some running outside of k8s) to connect to the collector by gRPC over our nginx-ingress. This works as desired as long as we don't use an encrypted (SSL/TLS) connection. Unfortunately for my company this is not an option for production use, as we require a secured connection.

Currently the agent does not support gRPC with TLS as the `grpc.Dial` method is called with the option `WithInsecure()` (see [collector_proxy.go#L55](https://github.com/jaegertracing/jaeger/blob/master/cmd/agent/app/reporter/grpc/collector_proxy.go#L55)/[collector_proxy.go#L52](https://github.com/jaegertracing/jaeger/blob/master/cmd/agent/app/reporter/grpc/collector_proxy.go#L52)). As of this we're currently not able to use jaeger in production.

Make it possible to choose whether the agent connects over an insecure or secure connection.

I pached the agent myself to test if the agent would work in our setup:
_error handling and sourrounding code omitted, implemented in the `NewCollectorProxy` method in [collector_proxy.go](https://github.com/jaegertracing/jaeger/blob/master/cmd/agent/app/reporter/grpc/collector_proxy.go)_
```golang
pool, _:= x509.SystemCertPool()
creds := credentials.NewClientTLSFromCert(pool, "")
conn, _ = grpc.Dial(r.Scheme()+":///round_robin", grpc.WithBalancerName(roundrobin.Name), grpc.WithTransportCredentials(creds))
```
With this small change I was able to get a connection to the collector over TLS.
- Agent: Implement a flag to enable/disable insecure/secure connection
- Agent: Flag to pass a public key file to use? Or simply use the system trusts as in my example?
- Collector: TLS implementation for people who don't use edge termination to setup a secured connection

At first I can implement this in the agent if there is intrest in it. But probably it would be also good to natively support TLS on the server side (collector) for people who don't use edge-termination but want to secure the communication (also happy to help out here if desired).

ghouscht:
# Which problem is this PR solving?

Resolves #1310

Implementation for a secure gRPC connection between the agent and the collector.

### DIFF
--- a/cmd/agent/app/reporter/grpc/flags.go
+++ b/cmd/agent/app/reporter/grpc/flags.go
@@ -22,10 +22,13 @@ import (
 )
 
 const (
-	gRPCPrefix        = "reporter.grpc."
-	collectorHostPort = gRPCPrefix + "host-port"
-	retry             = gRPCPrefix + "retry.max"
-	defaultMaxRetry   = 3
+	gRPCPrefix             = "reporter.grpc."
+	collectorHostPort      = gRPCPrefix + "host-port"
+	retry                  = gRPCPrefix + "retry.max"
+	defaultMaxRetry        = 3
+	collectorTLS           = gRPCPrefix + "tls"
+	collectorTLSCA         = gRPCPrefix + "tls.ca"
+	collectorTLSServerName = gRPCPrefix + "tls.server-name"
 )
 
 // Options Struct to hold configurations
@@ -33,12 +36,18 @@ type Options struct {
 	// CollectorHostPort is list of host:port Jaeger Collectors.
 	CollectorHostPort []string
 	MaxRetry          uint
+	TLS               bool
+	TLSCA             string
+	TLSServerName     string
 }
 
 // AddFlags adds flags for Options.
 func AddFlags(flags *flag.FlagSet) {
 	flags.String(collectorHostPort, "", "Comma-separated string representing host:port of a static list of collectors to connect to directly.")
 	flags.Uint(retry, defaultMaxRetry, "Sets the maximum number of retries for a call.")
+	flags.Bool(collectorTLS, false, "Enable TLS.")
+	flags.String(collectorTLSCA, "", "Path to a TLS CA file. (default use the systems truststore)")
+	flags.String(collectorTLSServerName, "", "Override the TLS server name.")
 }
 
 // InitFromViper initializes Options with properties retrieved from Viper.
@@ -48,5 +57,8 @@ func (o *Options) InitFromViper(v *viper.Viper) *Options {
 		o.CollectorHostPort = strings.Split(hostPorts, ",")
 	}
 	o.MaxRetry = uint(v.GetInt(retry))
+	o.TLS = v.GetBool(collectorTLS)
+	o.TLSCA = v.GetString(collectorTLSCA)
+	o.TLSServerName = v.GetString(collectorTLSServerName)
 	return o
 }
--- a/cmd/agent/app/flags_test.go
+++ b/cmd/agent/app/flags_test.go
@@ -24,7 +24,7 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestBingFlags(t *testing.T) {
+func TestBindFlags(t *testing.T) {
 	v := viper.New()
 	b := &Builder{}
 	command := cobra.Command{}
--- a/cmd/agent/app/reporter/flags_test.go
+++ b/cmd/agent/app/reporter/flags_test.go
@@ -24,7 +24,7 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestBingFlags(t *testing.T) {
+func TestBindFlags(t *testing.T) {
 	v := viper.New()
 	command := cobra.Command{}
 	flags := &flag.FlagSet{}
--- a/cmd/agent/app/reporter/grpc/collector_proxy.go
+++ b/cmd/agent/app/reporter/grpc/collector_proxy.go
@@ -15,13 +15,14 @@
 package grpc
 
 import (
+	"crypto/x509"
 	"errors"
 
-	"github.com/grpc-ecosystem/go-grpc-middleware/retry"
 	"github.com/uber/jaeger-lib/metrics"
 	"go.uber.org/zap"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/balancer/roundrobin"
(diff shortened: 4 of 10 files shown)


ghouscht:
Looks like an unrelated test is now failing in CI: `TestSaramaConsumerWrapper_start_Messages`. Can't reproduce it on my machine, is this test known to be flaky/broken?

yurishkuro:
possibly, I restarted the build to see.

yurishkuro:
Confirmed flaky test - #1314

yurishkuro:
@ghouscht what is your action plan with this PR? Are you going to add the equivalent settings to the collector?

ghouscht:
Yes, today I wrote some very basic unit tests for the agent part (just flags and ProxyBuilder). 
Will have a look at the collector code tomorrow to add those settings there to. The only thing left then, is some sort of e2e test, but don't know at the moment how to do that...

ghouscht:
Code for the collector part now also finished. Tested it on my local machine and seems to work well. 

things left TODO:
- agent: rewrite unit test `TestProxyBuilder`with custom CA (open for ideas) possible idea: write a cert to a tempfile and use this file to test the `NewCollectorProxy` method with a real CA or is the test like it is now ok?
- collector: do we need a custom error message if the user starts the collector with `--collector.grpc.tls` but does not provide a cert and key? Or just stay with the current error which is like: `"Could not start gRPC collector","error":"failed to load TLS keys: open : no such file or directory"` (not ideal in my opinion)?
- E2E test? how?

yurishkuro:
loading certs from a file in /tmp might be a good way to unit test the initialization paths

ghouscht:
Added [pkg/testutils/certificate.go](https://github.com/jaegertracing/jaeger/pull/1311/files#diff-55ee2fc4b2ba1290248a7b20168aec0d), maybe a bit overkill at the moment. But this method could be used too for an e2e/integration test where a cert and key is needed.

Do you think there is anything left which needs to be done? E2E test or tests for the initialization paths of the collector in [cmd/collector/main.go](https://github.com/jaegertracing/jaeger/blob/master/cmd/collector/main.go) (currently no tests at all for this part of the code)?

pkg/testutils/certificate.go
```diff
@@ -0,0 +1,99 @@
+// Copyright (c) 2019 The Jaeger Authors.
```
yurishkuro: It doesn't feel like we should have this code (which is also hard to test). Go already comes with `crypto/tls/generate_cert.go` that we can run (there's also installable version https://github.com/odeke-em/ssg).

yurishkuro:
we do have integration tests running binaries, `crossdock/docker-compose.yml`, but it's hard to extend. In fact, we still haven't switched it to use gRPC.

pkg/testutils/certificate.go
ghouscht: sure, removed that part of the PR

ghouscht:
Cleaned up the PR and removed the certificate stuff. `TestSaramaConsumerWrapper_start_Messages` is failing again... (this time with a different error I think so).

benley:
You'll probably want to add a ssl trusted roots bundle to the jaeger-agent docker image, similar to what already exists in [jaeger-collector's Dockerfile](https://github.com/jaegertracing/jaeger/blob/c4bb6b524f6be69b3268757f0c2b7c43e4ec7b4b/cmd/collector/Dockerfile#L1-L6):

```diff
--- a/cmd/agent/Dockerfile
+++ b/cmd/agent/Dockerfile
@@ -1,5 +1,10 @@
+FROM alpine:latest as certs
+RUN apk add --update --no-cache ca-certificates
+
 FROM scratch
 
+COPY --from=certs /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
+
 EXPOSE 5775/udp 6831/udp 6832/udp 5778
 COPY agent-linux /go/bin/
-ENTRYPOINT ["/go/bin/agent-linux"]
+ENTRYPOINT ["/go/bin/agent-linux", "--reporter.grpc.tls.ca=/etc/ssl/certs/ca-certificates.crt"]
```

ghouscht:
@yurishkuro how to proceed now with this PR?

cmd/agent/app/reporter/grpc/flags.go
```diff
@@ -22,19 +22,28 @@ import (
...
 	CollectorHostPort []string
+	TLS               bool
+	TLSCA             string
+	TLSServerName     string
 }
 
 // AddFlags adds flags for Options.
 func AddFlags(flags *flag.FlagSet) {
 	flags.String(collectorHostPort, "", "(experimental) Comma-separated string representing host:port of a static list of collectors to connect to directly.")
+	flags.Bool(collectorTLS, false, "(experimental) Enable TLS.")
```
pavolloffay: We have removed experimental flags from grpc

ghouscht: Ok, will rebase my PR to master and change this in the code..

ghouscht:
Looks like an unrelated test is failing again? What about coverage, is it ok like this?

pavolloffay:
The coverage should remain at 100%.

ghouscht:
Ok, catching the first miss [here](https://codecov.io/gh/jaegertracing/jaeger/pull/1311/diff?src=pr&el=tree#D5-53) could be hard, or at least I don't know how to do this, any ideas?

The second miss [here](https://codecov.io/gh/jaegertracing/jaeger/pull/1311/diff?src=pr&el=tree#D5-60) shouldn't be a problem, will do this in a few minutes.

yurishkuro:
@ghouscht I fixed coverage by adding one more test (#1391). I can get it approved and merged tomorrow, or you can merge it into your branch, and then we can merge this PR.

NB: in the future I recommend not committing directly to master in your fork, but creating a branch (as I did in #1391). Had you done that, I would've been able to push these extra commits directly into your branch and update this PR automatically.

ghouscht:
I'm out  of the office today. Will merge the changes into my branch tomorrow and squash the commits together.

yurishkuro:
merged as #1391

PR #1312, opened 2019-02-01
Separate query-service functionality from http handler

annanay25:

- Prereq for #1307 

- Separate common utils (span reader/writer, logger, tracer) of the query-service from the APIHandler definition. This will be used by the gRPC handler as well.


cmd/query/app/query_service.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/opentracing/opentracing-go"
+	"go.uber.org/zap"
+
+	"github.com/jaegertracing/jaeger/model/adjuster"
+	"github.com/jaegertracing/jaeger/storage/dependencystore"
+	"github.com/jaegertracing/jaeger/storage/spanstore"
+)
+
+// QueryService contains span utils required by the query-service.
+type QueryService struct {
```
yurishkuro: This is a good start, but the idea is to encapsulate everything into this type, not keep accessing its private members from the handlers. To help with that I would recommend moving it to its own package `app/querysvc`.

cmd/query/app/handler.go
```diff
@@ -107,14 +103,14 @@ func NewAPIHandler(spanReader spanstore.Reader, dependencyReader dependencystore
 	if aH.apiPrefix == "" {
 		aH.apiPrefix = defaultAPIPrefix
 	}
-	if aH.adjuster == nil {
-		aH.adjuster = adjuster.Sequence(StandardAdjusters...)
+	if aH.queryService.adjuster == nil {
```
yurishkuro: this defaulting should happen in the constructor of the QueryService

cmd/query/app/query_service.go
```diff
@@ -0,0 +1,35 @@
...
+	spanReader        spanstore.Reader
+	archiveSpanReader spanstore.Reader
+	archiveSpanWriter spanstore.Writer
+	dependencyReader  dependencystore.Reader
+	adjuster          adjuster.Adjuster
+	logger            *zap.Logger
+	tracer            opentracing.Tracer
```
yurishkuro: similar to handler constructor, most of these fields are optional. I suggest creating a QueryServiceOptions struct for that, so that the main constructor looks like this (if error never happens then don't need it):
```
func NewQueryService(
    spanReader spanstore.Reader, 
    dependencyReader dependencystore.Reader, 
    options QueryServiceOptions
) (*QueryService, error) {
}
```

cmd/query/app/querysvc/query_service_options.go
```diff
@@ -0,0 +1,49 @@
...
+// QueryServiceOption is a function that sets some option on the QueryService
+type QueryServiceOption func(qsvc *QueryService)
+
+// QueryServiceOptions is a factory for all available QueryServiceOptions
+var QueryServiceOptions queryServiceOptions
+
+type queryServiceOptions struct{}
+
+// Adjusters creates a QueryServiceOption that initializes the sequence of Adjusters on the QueryService.
+func (queryServiceOptions) Adjusters(adjusters ...adjuster.Adjuster) QueryServiceOption {
```
yurishkuro: Sorry, I should've been more explicit - let's not use the functional options pattern, we're trying to move away from it, just define the additional fields in QueryServiceOptions struct.

cmd/query/app/http_handler.go
```diff
@@ -78,23 +78,18 @@ func NewRouter() *mux.Router {
...
+	queryService *querysvc.QueryService
+	queryParser  queryParser
+	basePath     string
+	apiPrefix    string
+	logger       *zap.Logger
+	tracer       opentracing.Tracer
 }
 
 // NewAPIHandler returns an APIHandler
 func NewAPIHandler(spanReader spanstore.Reader, dependencyReader dependencystore.Reader, options ...HandlerOption) *APIHandler {
```
yurishkuro: I suggest accepting QueryService instead of creating it internally, otherwise you don't have a way to apply its additional configuration options.

cmd/query/app/http_handler.go
```diff
@@ -362,96 +346,61 @@ func (aH *APIHandler) getTraceFromReaders(
 	w http.ResponseWriter,
```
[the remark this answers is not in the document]
annanay25: Yes, will make this change.

cmd/query/app/http_handler.go
```diff
@@ -362,96 +346,61 @@ func (aH *APIHandler) getTraceFromReaders(
...
-			Data:   []string{}, // doens't matter, just want an empty array
-			Errors: []structuredError{},
-		}
-		aH.writeJSON(w, r, &structuredRes)
-	})
+
+	// QueryService.ArchiveTrace can now archive this traceID.
+	err := aH.queryService.ArchiveTrace(r.Context(), &traceID)
+	if err != nil {
+		aH.handleError(w, err, http.StatusInternalServerError)
```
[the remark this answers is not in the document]
annanay25: Will do.

cmd/query/app/querysvc/query_service.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+// Implement the adjuster.Adjuster interface
+
+// Adjust implements adjuster.Adjuster.Adjust
+func (qs QueryService) Adjust(trace *model.Trace) (*model.Trace, error) {
+	return qs.options.adjuster.Adjust(trace)
+}
+
+// Implement the dependencystore.Reader interface
```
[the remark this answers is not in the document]
annanay25: Initially I thought to implement the reader/writer interface so that the `QueryService` object could be passed around as a `spanstore.Reader` / `spanstore.Writer` object in the APIHandler. But I realise that we're trying to abstract the details of the reader/writer away from the APIHandler, so should I change the functions to pass a QueryService object instead?

cmd/query/app/querysvc/query_service_options.go
```diff
@@ -0,0 +1,27 @@
...
+
+package querysvc
+
+import (
+	"github.com/jaegertracing/jaeger/model/adjuster"
+	"github.com/jaegertracing/jaeger/storage/spanstore"
+)
+
+// QueryServiceOptions is a factory for all available QueryServiceOptions
+type QueryServiceOptions struct {
```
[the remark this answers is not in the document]
annanay25: Okay, will do.

cmd/query/app/querysvc/query_service.go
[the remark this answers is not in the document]
yurishkuro: APIHandler already has a reference to QueryService, so you don't need to pass anything, right?

If there's no strong reason to have QueryService implementing Reader/Writer, I would avoid mentioning it, so that in the future these can become de-coupled. E.g. we may need some extra argument to the QueryService, but not to  the storage APIs.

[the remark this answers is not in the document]
annanay25: You're right. I've made all the changes. 

Just one last thing - As part of the `handler_test` we have a mock implementation of the reader/writer interfaces [here](https://github.com/jaegertracing/jaeger/blob/master/cmd/query/app/handler_test.go#L136). This needs to be replaced with a mock QueryService implementation?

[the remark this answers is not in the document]
yurishkuro: why not use the main QueryService?  In many cases it just delegates to the underlying reader/writer anyway, mocking it doesn't seem that useful.

annanay25:
A couple of things I came across as I'm fixing the test cases after refactoring.

1. Test cases like - 
https://github.com/jaegertracing/jaeger/blob/d40ea3430b0172e3b3e7b0861bc10566e1d0d8ca/cmd/query/app/handler_deps_test.go#L309-L314

... are going to change to - 

```
func TestGetDependenciesSuccess(t *testing.T) {
	server, qs := initializeTestServer()
	defer server.Close()
	expectedDependencies := []model.DependencyLink{{Parent: "killer", Child: "queen", CallCount: 12}}
	endTs := time.Unix(0, 1476374248550*millisToNanosMultiplier)
```
And so we either need:
```  
// Expose DependencyReader directly  
qs.DependencyReader.On("GetDependencies", endTs, defaultDependencyLookbackDuration).Return(expectedDependencies, nil).Times(1)
```
OR
```
// Expose dependencyReader through QueryService.getDepReader()
qs.getDepReader().On("GetDependencies", endTs, defaultDependencyLookbackDuration).Return(expectedDependencies, nil).Times(1)
```
.. (if we don't go with implementing a mock for QueryService)

2. Since we're moving away from functional arguments, we'll now need constructors for `QueryServiceOptions` with signatures like -
```
// QueryServiceOptions has optional members of QueryService
type QueryServiceOptions struct {
	archiveSpanReader spanstore.Reader
	archiveSpanWriter spanstore.Writer
	adjuster          adjuster.Adjuster
}

// NewQueryServiceWithReader creates a QueryServiceOption with an archive reader.
func NewQueryServiceWithReader(archiveSpanReader spanstore.Reader) QueryServiceOptions

// NewQueryServiceWithWriter creates a QueryServiceOption with an archive writer.
func NewQueryServiceWithWriter(archiveSpanWriter spanstore.Writer) QueryServiceOptions 

.
.
```

@yurishkuro could you please review?

yurishkuro:
On the first question, initTestServer is the one that creates and returns the mocks for storage. It should continue doing that. The tests are testing the handler endpoints, they don't care if internally the init function constructs QueryService. It seems like I'm missing something that's confusing you.

On the second question, you need to make the fields of the options struct public, since they are part of the public API of the package.

cmd/query/app/http_handler.go
```diff
@@ -352,106 +339,62 @@ func (aH *APIHandler) parseTraceID(w http.ResponseWriter, r *http.Request) (mode
...
-		structuredRes := structuredResponse{
-			Data:   []string{}, // doens't matter, just want an empty array
-			Errors: []structuredError{},
-		}
-		aH.writeJSON(w, r, &structuredRes)
-	})
+
+	// QueryService.ArchiveTrace can now archive this traceID.
+	err := aH.queryService.ArchiveTrace(r.Context(), traceID)
+	if aH.handleError(w, err, http.StatusInternalServerError) {
```
yurishkuro: you may need to special case the "Trace not found" error and return 404

cmd/query/app/querysvc/query_service.go
```diff
@@ -0,0 +1,121 @@
...
+	for _, span := range trace.Spans {
+		err := qs.options.ArchiveSpanWriter.WriteSpan(span)
+		if err != nil {
+			writeErrors = append(writeErrors, err)
+		}
+	}
+	return multierror.Wrap(writeErrors)
+}
+
+// Adjust implements adjuster.Adjuster.Adjust
```
yurishkuro: `// Adjust applies adjusters to the trace.`

annanay25:
@yurishkuro One test is failing because the `querysvc` directory doesn't have a test file. Not sure what to do here.

cmd/query/app/utils.go
```diff
@@ -0,0 +1,53 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package app
+
+import (
+	"go.uber.org/zap"
+
+	"github.com/jaegertracing/jaeger/cmd/query/app/querysvc"
+	istorage "github.com/jaegertracing/jaeger/storage"
```
yurishkuro: why the alias?

cmd/query/app/utils.go
```diff
@@ -0,0 +1,53 @@
+// Copyright (c) 2019 The Jaeger Authors.
```
yurishkuro: utils.go -> archive_options.go

cmd/query/app/utils.go
yurishkuro: NB: by moving it here it would need to be tested. May not be worth  it, at least in this PR

annanay25:
I've added an empty test for now because we have all the mock-tests in place for the reader/writer handlers anyway. Could you re-run travis on this @yurishkuro? The build seems to be failing with an unrelated error.

yurishkuro:
@annanay25 there is a drop in code coverage, which shouldn't happen with plain refactoring. Could you find a way to bring coverage back?

annanay25:
Looking into this.

cmd/query/app/querysvc/query_service_test.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+func initializeTestService() (*QueryService, *spanstoremocks.Reader, *depsmocks.Reader) {
+	readStorage := &spanstoremocks.Reader{}
+	dependencyStorage := &depsmocks.Reader{}
+	qs := NewQueryService(readStorage, dependencyStorage, QueryServiceOptions{})
+	return qs, readStorage, dependencyStorage
+}
+
+// Test QueryService.GetTrace()
```
annanay25: Similarly, need to add tests for all QueryService member functions. Will push these tests by tomorrow, @yurishkuro.

annanay25:
@yurishkuro Build is green :)

annanay25:
@yurishkuro can we go ahead with this?

cmd/query/app/http_handler_test.go
```diff
@@ -97,17 +99,23 @@ func initializeTestServerWithHandler(options ...HandlerOption) (*httptest.Server
...
 	dependencyStorage := &depsmocks.Reader{}
+	qs := querysvc.NewQueryService(readStorage, dependencyStorage, queryOptions)
 	r := NewRouter()
-	handler := NewAPIHandler(readStorage, dependencyStorage, options...)
+	handler := NewAPIHandler(qs, options...)
 	handler.RegisterRoutes(r)
 	return httptest.NewServer(r), readStorage, dependencyStorage, handler
 }
 
+func initializeTestServerWithQueryOptions(queryOptions querysvc.QueryServiceOptions, options ...HandlerOption) (*httptest.Server, *spanstoremocks.Reader, *depsmocks.Reader) {
```
yurishkuro: this is identical to initializeTestServerWithOptions, please remove

yurishkuro:
🎉 🎈

annanay25:
Thanks Yuri! :smile:

Issue #1317, opened 2019-02-06
Add username/password support to esCleaner.py


gregoryfranklin:
The supplied esCleaner.py script is unable to run against elasticsearch clusters that require authentication.  This means we need to implement our own cleanup script or the elasticsearch cluster fills up.

I propose adding support for ES_USERNAME and ES_PASSWORD environment variables.  This would be consistent with the secrets used by jaeger-operator.

PR #1318, opened 2019-02-06
Make it possible for esCleaner.py to use a username/password

gregoryfranklin:
Resolves https://github.com/jaegertracing/jaeger/issues/1317

Add optional ES_USERNAME and ES_PASSWORD environment variables in order that esCleaner.py can authenticate against elasticsearch clusters.

### DIFF
--- a/plugin/storage/es/esCleaner.py
+++ b/plugin/storage/es/esCleaner.py
@@ -15,7 +15,13 @@ def main():
         print('ARCHIVE ... specifies whether to remove archive indices. Use true or false')
         sys.exit(1)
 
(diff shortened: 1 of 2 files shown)


black-adder:
thanks!

plugin/storage/es/esCleaner.py
```diff
@@ -15,7 +15,13 @@ def main():
         print('ARCHIVE ... specifies whether to remove archive indices. Use true or false')
         sys.exit(1)
 
-    client = elasticsearch.Elasticsearch(sys.argv[2:])
+    username = os.getenv("ES_USERNAME")
```
pavolloffay: These should be added into the help print

PR #1319, opened 2019-02-07
Load elasticsearch auth token from file

pavolloffay:
In k8s the auth token (serviceaccount) is injected to pod. 

### DIFF
--- a/plugin/storage/es/factory.go
+++ b/plugin/storage/es/factory.go
@@ -76,11 +76,13 @@ func (f *Factory) Initialize(metricsFactory metrics.Factory, logger *zap.Logger)
 
 	primaryClient, err := f.primaryConfig.NewClient(logger, metricsFactory)
 	if err != nil {
+		f.logger.Error("failed to create primary Elasticsearch client", zap.Error(err))
 		return err
 	}
 	f.primaryClient = primaryClient
 	archiveClient, err := f.archiveConfig.NewClient(logger, metricsFactory)
 	if err != nil {
+		f.logger.Error("failed to create archive Elasticsearch client", zap.Error(err))
 		return err
 	}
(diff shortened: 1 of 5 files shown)


plugin/storage/es/options.go
```diff
@@ -119,6 +120,10 @@ func addFlags(flagSet *flag.FlagSet, nsConfig *namespaceConfig) {
 		nsConfig.namespace+suffixPassword,
 		nsConfig.Password,
 		"The password required by ElasticSearch")
+	flagSet.String(
+		nsConfig.namespace+suffixTokenPath,
+		nsConfig.TokenFilePath,
+		"Path to a file containing bearer token. This flag also uses "+suffixCA+" if it is specified")
```
jpkrohling: `suffixCA` or `suffixTokenPath`? I think it might be implicit, but the token is the whole content of the file, right? Perhaps we could make it more explicit?

pavolloffay: `suffixCA`, it's not a bad thing to have more explanatory docs. I didn't know that we need root CA at all when using token.

yurishkuro: I wouldn't use the constant in the help message. It starts with a dot and will look weird. Just say CA.

plugin/storage/es/factory.go
```diff
@@ -76,11 +76,13 @@ func (f *Factory) Initialize(metricsFactory metrics.Factory, logger *zap.Logger)
 
 	primaryClient, err := f.primaryConfig.NewClient(logger, metricsFactory)
 	if err != nil {
+		f.logger.Error("failed to create primary Elasticsearch client", zap.Error(err))
```
yurishkuro: Logging should happen in main. We may want to call errors.Wrap() here

PR #1323, opened 2019-02-08
Add gRPC plugin proto changes

chvck:
* Issue #422, part of review #1214 

- Split proto and Makefile changes out from previous review to enable further progress there.


plugin/storage/grpc/proto/storage.proto
```diff
@@ -0,0 +1,199 @@
+syntax = "proto3";
+
+package jaeger.api_v2;
```
chvck: Unsure that this is correct.

plugin/storage/grpc/proto/storage.proto
```diff
@@ -0,0 +1,188 @@
...
+    // spanstore/Writer
+    rpc WriteSpan(WriteSpanRequest) returns (WriteSpanResponse);
+}
+
+service GrpcSpanReaderStoragePlugin {
+    // spanstore/Reader
+    rpc GetTrace(GetTraceRequest) returns (GetTraceResponse);
+    rpc GetServices(GetServicesRequest) returns (GetServicesResponse);
+    rpc GetOperations(GetOperationsRequest) returns (GetOperationsResponse);
+    rpc FindTraces(FindTracesRequest) returns (stream Trace);
```
chvck: Returning a stream here solves the issue surfaced in the other PR where grpc message limits were being hit.

yurishkuro: hm, not sure it completely solves those issues unless we allow multiple Trace entries in the batch for the same trace. A single trace can still easily be larger than the msg limit. 

Since this is a storage API, I would rather go with returning `stream FindTracesResponseChunk` where each chunk contains `[]Span`

plugin/storage/grpc/proto/storage.proto
```diff
@@ -0,0 +1,188 @@
...
+                                (gogoproto.nullable) = false,
+                                (gogoproto.customtype) = "github.com/jaegertracing/jaeger/model.TraceID",
+                                (gogoproto.customname) = "TraceIDs"
+                                ];
+}
+
+message FindTraceIDsResponse {
+    oneof response {
+        FindTraceIDsSuccess success = 1;
+        GrpcStoragePluginError error = 2;
```
yurishkuro: couldn't we instead use [grpc Status](https://jbrandhorst.com/post/grpc-errors/)? It already includes an error message.

plugin/storage/grpc/proto/storage.proto
```diff
@@ -0,0 +1,188 @@
...
+    rpc GetDependencies(GetDependenciesRequest) returns (GetDependenciesResponse);
+}
+
+// TODO(olivierboucher): make theses available once the POC goes through
+// samplingstore/Store
+//    rpc InsertThroughput(InsertThroughputRequest) returns (InsertThroughputResponse);
+//    rpc InsertProbabilitiesAndQPS(InsertProbabilitiesAndQPSRequest) returns (InsertProbabilitiesAndQPSResponse);
+//    rpc GetThroughput(GetThroughputRequest) returns (GetThroughputResponse);
+//    rpc GetProbabilitiesAndQPS(GetProbabilitiesAndQPSRequest) returns (GetProbabilitiesAndQPSResponse);
+//    rpc GetLatestProbabilities(GetLatestProbabilitiesRequest) returns (GetLatestProbabilitiesResponse);
```
yurishkuro: let's remove this

plugin/storage/grpc/proto/storage.proto
```diff
@@ -0,0 +1,188 @@
...
+message GrpcStoragePluginError {
+    string message = 1;
+}
+
+service GrpcSpanWriterStoragePlugin {
+    // spanstore/Writer
+    rpc WriteSpan(WriteSpanRequest) returns (WriteSpanResponse);
+}
+
+service GrpcSpanReaderStoragePlugin {
```
yurishkuro: I am not sure we need Grpc and StoragePlugin prefix/suffix, they will be obvious from the package name.

plugin/storage/grpc/proto/storage.proto
```diff
@@ -0,0 +1,188 @@
...
+    // spanstore/Reader
+    rpc GetTrace(GetTraceRequest) returns (GetTraceResponse);
+    rpc GetServices(GetServicesRequest) returns (GetServicesResponse);
+    rpc GetOperations(GetOperationsRequest) returns (GetOperationsResponse);
+    rpc FindTraces(FindTracesRequest) returns (stream Trace);
+    rpc FindTraceIDs(FindTraceIDsRequest) returns (FindTraceIDsResponse);
+}
+
+service GrpcDependenciesStoragePlugin {
```
yurishkuro: any reason we don't want this split into reader and writer, similar to span storage?

yurishkuro:
to resolve Gopkg.lock conflict you may want to merge master, `git checkout --theirs Gopkg.lock`, then rerun `dep ensure --update`

plugin/storage/grpc/proto/storage.proto
chvck: Happy to do that. Do we consider dependency store writer as out of scope for this changeset?

plugin/storage/grpc/proto/storage.proto
chvck: Out of curiosity what sort of size (and number of spans) would you expect to see the largest of traces reach?

yurishkuro: we have seen traces in production with 100,000 spans. At average span size of around 500 bytes, that's 50M. Even 10k spans would add up to 5Mb.

I think eventually we'd want to make the chunk size configurable, but I would start with 1000 or even 100 spans.

plugin/storage/grpc/proto/storage.proto
yurishkuro: we can skip it for now.

plugin/storage/grpc/proto/storage.proto
chvck: Something to note here is that the gogo plugin of Status is go 1.11+ (https://github.com/gogo/status)

plugin/storage/grpc/proto/storage.proto
```diff
@@ -0,0 +1,178 @@
...
+
+package jaeger.storage.v1;
+
+option go_package = "api_v1";
+
+import "gogoproto/gogo.proto";
+import "google/protobuf/timestamp.proto";
+import "google/protobuf/duration.proto";
+import "google/protobuf/empty.proto";
+import "google/rpc/status.proto";
```
chvck: This is a go 1.11+ requirement (https://github.com/gogo/status)

plugin/storage/grpc/proto/storage.proto
```diff
@@ -0,0 +1,178 @@
...
+}
+
+message WriteSpanRequest {
+    jaeger.api_v2.Span span = 1;
+}
+
+message WriteSpanResponse {
+    oneof response {
+        google.protobuf.Empty success = 1;
+        google.rpc.Status error = 2;
```
yurishkuro: Sorry, this is not what I meant. This Status object is transported by gRPC natively, we don't need to use it in our response payloads.

chvck: Are you saying to basically remove the *Response messages and simply return *Success from the rpc functions? I've just had a play doing this and it makes the jaeger plugin code that talks to the generated proto code much cleaner.

plugin/storage/grpc/proto/storage.proto
```diff
@@ -0,0 +1,135 @@
...
+      (gogoproto.stdtime) = true,
+      (gogoproto.nullable) = false
+    ];
+    google.protobuf.Duration lookback = 2 [
+      (gogoproto.stdduration) = true,
+      (gogoproto.nullable) = false
+    ];
+}
+
+message GetDependenciesSuccess {
```
yurishkuro: GetDependenciesResponse ?  Here and in all other responses.

plugin/storage/grpc/proto/storage.proto
```diff
@@ -0,0 +1,135 @@
...
+}
+
+message FindTracesRequest {
+    TraceQueryParameters query = 1;
+}
+
+message FindTracesResponseChunk {
+    repeated bytes spans = 1  [
+      (gogoproto.nullable) = false,
+      (gogoproto.customtype) = "github.com/jaegertracing/jaeger/model.Span"
```
yurishkuro: I don't think this is the right dependency. The type should be Span from `model.proto`, the fact that it is backed by some specific Go type is irrelevant. Otherwise how could one implement a plugin in Java?

plugin/storage/grpc/proto/storage.proto
```diff
@@ -0,0 +1,135 @@
...
+option (gogoproto.marshaler_all) = true;
+// Enable custom Unmarshal method.
+option (gogoproto.unmarshaler_all) = true;
+// Enable custom Size method (Required by Marshal and Unmarshal).
+option (gogoproto.sizer_all) = true;
+// Enable registration with golang/protobuf for the grpc-gateway.
+option (gogoproto.goproto_registration) = true;
+
+message GetDependenciesRequest {
+    google.protobuf.Timestamp end_timestamp = 1 [
```
yurishkuro: minor, but I think we should clean up the API and use `start_time` and `end_time` (don't need to call them `timestamp`, it's just the type of the value, not the meaning of it).

plugin/storage/grpc/proto/storage.proto
```diff
@@ -0,0 +1,135 @@
...
+    repeated bytes trace_id = 1 [
+      (gogoproto.nullable) = false,
+      (gogoproto.customtype) = "github.com/jaegertracing/jaeger/model.TraceID",
+      (gogoproto.customname) = "TraceIDs"
+    ];
+}
+
+service SpanWriterPlugin {
+    // spanstore/Writer
+    rpc WriteSpan(WriteSpanRequest) returns (google.protobuf.Empty);
```
yurishkuro: let's define an empty WriteSpanResponse instead of `google.protobuf.Empty`, it is better for extensibility.

yurishkuro:
looks good overall

yurishkuro:
please resolve conflicts, the DependencyLink type has been extended in 1.10

plugin/storage/grpc/proto/storage.proto
```diff
@@ -22,7 +22,7 @@ option (gogoproto.sizer_all) = true;
 option (gogoproto.goproto_registration) = true;
 
 message GetDependenciesRequest {
-    google.protobuf.Timestamp end_timestamp = 1 [
+    google.protobuf.Timestamp end_time = 1 [
```
yurishkuro: I meant replace `lookback` with `start_time` (and swap the order). The Query Service API in the future will be converted to similar gRPC API, but it will need to work over JSON. It's less cognitive load to have two timestamp-formatted fields in JSON than to have a lookback in nanoseconds.

plugin/storage/grpc/proto/storage.proto
```diff
@@ -43,6 +43,10 @@ message WriteSpanRequest {
     jaeger.api_v2.Span span = 1;
 }
 
+message WriteSpanResponse {
+
```
yurishkuro: add comment `// empty; extensible in the future`

plugin/storage/grpc/proto/storage.proto
```diff
@@ -32,7 +32,7 @@ message GetDependenciesRequest {
     ];
 }
 
-message GetDependenciesSuccess {
+message GetDependenciesResponse {
     repeated bytes dependencies = 1 [
```
yurishkuro: DependencyLink?

plugin/storage/grpc/proto/storage.proto
```diff
@@ -117,19 +120,19 @@ message FindTraceIDsSuccess {
...
-    rpc WriteSpan(WriteSpanRequest) returns (google.protobuf.Empty);
+    rpc WriteSpan(WriteSpanRequest) returns (WriteSpanResponse);
 }
 
 service SpanReaderPlugin {
     // spanstore/Reader
-    rpc GetTrace(GetTraceRequest) returns (GetTraceSuccess);
-    rpc GetServices(GetServicesRequest) returns (GetServicesSuccess);
-    rpc GetOperations(GetOperationsRequest) returns (GetOperationsSuccess);
+    rpc GetTrace(GetTraceRequest) returns (GetTraceResponse);
```
yurishkuro: I suggest implementing this similar to FindTraces, as a stream of chunks, because even a single trace could be larger than the default message limit.

plugin/storage/grpc/proto/storage.proto
```diff
@@ -0,0 +1,138 @@
+syntax = "proto3";
+
+package jaeger.storage.v1;
+
+option go_package = "api_v1";
```
yurishkuro: maybe `storage_v1` would be better?

Makefile
```diff
@@ -393,11 +396,18 @@ proto:
 		--swagger_out=$(PWD)/proto-gen/openapi/ \
 		model/proto/api_v2.proto
 
+	protoc \
+		$(PROTO_INCLUDES) \
+		-I plugin/storage/grpc/proto \
+				--gogo_out=plugins=grpc,$(PROTO_GOGO_MAPPINGS):$(PWD)/plugin/storage/grpc/proto \
```
yurishkuro: I would prefer folder name to match the package, e.g. `plugin/storage/grpc/proto/storage_v1`

model/proto/model.proto
```diff
@@ -158,3 +158,9 @@ message Batch {
       (gogoproto.nullable) = true
     ];
 }
+
+message DependencyLink {
+  string parent = 1;
+  string child = 2;
+  uint64 call_count = 3;
```
yurishkuro: missing the new `source` field

plugin/storage/grpc/proto/storage.proto
```diff
@@ -0,0 +1,138 @@
...
+      (gogoproto.nullable) = false
+    ];
+}
+
+message FindTraceIDsRequest {
+    TraceQueryParameters query = 1;
+}
+
+message FindTraceIDsResponse {
+    repeated bytes trace_id = 1 [
```
yurishkuro: trace_ids

model/proto/model.proto
chvck: Trying to work out the correct way to pull this in, are you expecting to see it as a bytes field type with a customtype hooking out to DependencyLinkSource?

yurishkuro: I see what you mean - I think we should change the current code to remove DependencyLinkSource and use a plain string. The idea was that external sources can submit links data, and there's no reason to have an enum in the code representing those sources, it's enough to have just the constant for "jaeger" as the default source type.

plugin/storage/grpc/proto/storage.proto
chvck: Something that I'm unsure about and may need consideration is that having a Trace as a stream of Span means that the ProcessMap and Warnings fields are lost on the Trace.

yurishkuro: That's true, but these two concepts only apply to the current JSON API used by the UI, they don't apply to the storage API (even though the storage API returns `model.Trace`, it never actually populates ProcessMap or warnings, that's done post-retrieval).

model/proto/model.proto
```diff
@@ -158,3 +158,13 @@ message Batch {
...
+  string source = 4 [
+    (gogoproto.customtype) = "DependencyLinkSource",
```
chvck: I thought that this would be the best of both worlds re: the conversation above.

yurishkuro:
build fails with
```
./plugin/storage/grpc/proto/storage_v1/storage.pb.go
plugin/storage/grpc/proto/storage_v1/storage.pb.go
```

we need to add this file to exclusions somewhere in Makefile

model/proto/model.proto
yurishkuro: well, if we do this, I would recommend adding a simple marshaling/unmarshaling test, similar to `model/ids_test.go`

yurishkuro: Actually, not just for this, but also for importing `model.Span`. I found it not particularly useful just adding gogoproto annotations without testing that serialization works as expected

chvck: I'm not following what you mean by "importing `model.Span`". In the plugin proto file? I don't believe that we actually use gogoproto annotations for importing Span there?

yurishkuro: I mean this:
```
message GetTraceRequest {
    bytes trace_id = 1 [
      (gogoproto.nullable) = false,
    ];
}
```
From my previous experience  with gogoproto, even though `protoc` may compile it, you won't know if it will serialize correctly until you try it in unit tests.

yurishkuro: and similar for spans

yurishkuro:
@chvck please `git commit -s --amend` the last commit

yurishkuro:
Some other commit must be missing sign-off. The simplest solution in this case is to squash all commits and sign the result.

plugin/storage/grpc/proto/storage_v1/storage_test.go
```diff
@@ -0,0 +1,137 @@
...
+			// test unmarshal
+			var ref1u storage_v1.GetTraceRequest
+			err = testCase.unmarshal(d2, &ref1u)
+			require.NoError(t, err)
+			assert.Equal(t, ref1, ref1u)
+		})
+	}
+}
+
+func TestFindTraceIDsResponseMarshalProto(t *testing.T) {
```
yurishkuro: other than serializing a different type, what does this test validate?  The embedded type is still `TraceID` as in the test above. It would be more useful to test serialization of `SpansResponceChunk`, because it uses an imported type, which is different from custom type `TraceID` which is just `bytes` in proto.

plugin/storage/grpc/proto/storage_v1/storage_test.go
```diff
@@ -0,0 +1,137 @@
...
+		marshal   func(proto.Message) ([]byte, error)
+		unmarshal func([]byte, proto.Message) error
+		expected  string
+	}{
+		{
+			name:      "protobuf",
+			marshal:   proto.Marshal,
+			unmarshal: proto.Unmarshal,
+		},
+		{
```
yurishkuro: don't think we need JSON

proto-gen/openapi/api_v2.swagger.json
```diff
@@ -20,7 +20,7 @@
         "operationId": "GetSamplingStrategy",
         "responses": {
           "200": {
-            "description": "",
+            "description": "A successful response.",
```
yurishkuro: I think this needs rebase from master

plugin/storage/grpc/proto/storage_v1/storage_test.go
chvck: I don't really see an effective way to test `SpansResponseChunk`, creating a type in the _test proto file would contain the same import (just missing the notnull annotation). All that I can think to do is marshal and then unmarshal a `SpansResponseChunk` and check that before == after.

yurishkuro: yes, to-from is what I meant, not the types of tests done with _test.proto (those are only useful for custom types with custom marshaling methods).

chvck: OK I see, interestingly the `timestamp` field on `model.Span` seems to lose the `loc` field value when marshaled/unmarshaled. I'm trying to work out if that's a gogoproto issue.

yurishkuro: That may be expected. The timestamps we capture are absolute (UTC), the locale is not supposed to be preserved. We have `span.NormalizeTimestamps()` helper to deal with that.

yurishkuro:
Please make sure to fix the DCO/sign-off issue, we can't merge the PR otherwise.

chvck:
> Please make sure to fix the DCO/sign-off issue, we can't merge the PR otherwise.

Squash and then merge/rebase onto master is throwing up some strange conflicts. Mind if I update this branch by branching master again and then cherry-picking the commits specific to this change?

yurishkuro:
You probably don't need to rebase, just squash & merge master. One other option could be to squash only commits starting from the one where the sign-off was missed, e.g. `git reset --soft C` where C is the commit before the one where you missed the sign-off. Then you can commit all subsequent commits as one.

chvck:
I'm unsure why `make proto` is changing files in `proto-gen`, it doesn't seem to do that locally on my machine.

yurishkuro:
Are you using the same protocol? Did you run "make proto-install" locally before generating?

chvck:
> Are you using the same protocol? Did you run "make proto-install" locally before generating?

Thanks, that was the step I was missing. The plugin proto generated files are generating without the license header until `make fmt` is run which is why the build is flagging those as they've been included in the codebase with the license header. As these are generated files is it safe to leave the header off them?

yurishkuro:
We avoid changing generated files, even with tools (eg https://github.com/jaegertracing/jaeger/blob/master/thrift-gen/jaeger/constants.go). We need to add them to exclusion lists in the Makefile so that go fmt and lint don't run on them.

chvck:
Hi @yurishkuro before I fix this up by merging master in, are there any changes here that you're waiting on me for?

Makefile
```diff
@@ -385,11 +390,23 @@ proto:
 		--swagger_out=allow_merge=true:$(PWD)/proto-gen/openapi/ \
 		model/proto/api_v2/*.proto
 
+	$(PROTOC) \
+		$(PROTO_INCLUDES) \
+		-I plugin/storage/grpc/proto \
+				--gogo_out=plugins=grpc,$(PROTO_GOGO_MAPPINGS):$(PWD)/plugin/storage/grpc/proto/storage_v1 \
```
yurishkuro: in order to avoid code coverage drop, we should place the generated files under `proto-gen/storage_v1/`

Makefile
```diff
@@ -385,11 +390,23 @@ proto:
...
+				plugin/storage/grpc/proto/storage.proto
```
yurishkuro: please fix indentation in these two lines to align with L395

PR #1324, opened 2019-02-08
Require a slightly newer version of grpc-go

benley:
This includes the fix for https://github.com/grpc/grpc-go/issues/1928 (https://github.com/grpc/grpc-go/pull/2045), which makes jaeger-collector usable behind nginx (and probably other grpc/http2 implementations) as load balancers for its grpc endpoint.

More info: https://trac.nginx.org/nginx/ticket/1538

Require grpc-go >=1.13 instead of >=1.11


yurishkuro:
Thanks!

PR #1328, opened 2019-02-11
Remove SASI indices

### LINKED ISSUE #790: Remove SASI index on dependency column family

black-adder:

A lot of Copy Pasta from #793. Resolves #790.

Migration Path:
1) Run `v001tov002part1.sh` which will copy dependencies into a csv, update the `dependency UDT`, create a new `dependenciesv2` table, and write dependencies from the csv into the `dependenciesv2` table.
2) Run the collector and query services with the cassandra flag `dependency-sasi-disabled=true` which will update jaeger to read from the new `dependenciesv2` table.
3) Update [spark job](https://github.com/jaegertracing/spark-dependencies) to write to the new `dependenciesv2` table (this is a TODO).
4) Run `v001tov002part2.sh` which will delete the old dependency table and the SASI index.

Users who wish to continue to use the SASI indices don't have to do anything as the cassandra flag `dependency-sasi-disabled` will default to false. Ideally, users will migrate on their own timeline and we can remove support for the old table in a major version release.

- [ ] Fix crossdock
- [ ] Actually test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,17 @@ Changes by Version
 
 ##### Breaking Changes
 
+- Remove cassandra SASI indices [#1328](https://github.com/jaegertracing/jaeger/pull/1328)
+
+Migration Path:
+
+1. Run `plugin/storage/cassandra/schema/migration/v001tov002part1.sh` which will copy dependencies into a csv, update the `dependency UDT`, create a new `dependencies_v2` table, and write dependencies from the csv into the `dependencies_v2` table.
+2. Run the collector and query services with the cassandra flag `sasi-disabled=true` which will update jaeger to read from the new `dependencies_v2` table.
+3. Update [spark job](https://github.com/jaegertracing/spark-dependencies) to write to the new `dependencies_v2` table.
+4. Run `plugin/storage/cassandra/schema/migration/v001tov002part2.sh` which will DELETE the old dependency table and the SASI index.
+
+Users who wish to continue to use the SASI indices don't have to do anything as the cassandra flag `sasi-disabled` will default to false. Users may migrate on their own timeline however new features will be built solely on the `dependencies_v2` table.
+
 ##### New Features
 
 ##### Bug fixes, Minor Improvements
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,11 +11,11 @@ Changes by Version
 Migration Path:
 
 1. Run `plugin/storage/cassandra/schema/migration/v001tov002part1.sh` which will copy dependencies into a csv, update the `dependency UDT`, create a new `dependencies_v2` table, and write dependencies from the csv into the `dependencies_v2` table.
-2. Run the collector and query services with the cassandra flag `enable-dependencies-v2=true` which will update jaeger to write and read to and from the new `dependencies_v2` table.
+2. Run the collector and query services with the cassandra flag `cassandra.enable-dependencies-v2=true` which will instruct jaeger to write and read to and from the new `dependencies_v2` table.
 3. Update [spark job](https://github.com/jaegertracing/spark-dependencies) to write to the new `dependencies_v2` table.
 4. Run `plugin/storage/cassandra/schema/migration/v001tov002part2.sh` which will DELETE the old dependency table and the SASI index.
 
-Users who wish to continue to use the v1 table don't have to do anything as the cassandra flag `enable-dependencies-v2` will default to false. Users may migrate on their own timeline however new features will be built solely on the `dependencies_v2` table. In the future, we will remove support for v1 completely.
+Users who wish to continue to use the v1 table don't have to do anything as the cassandra flag `cassandra.enable-dependencies-v2` will default to false. Users may migrate on their own timeline however new features will be built solely on the `dependencies_v2` table. In the future, we will remove support for v1 completely.
 
 ##### Breaking Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,18 +6,18 @@ Changes by Version
 
 #### Backend Changes
 
-##### Breaking Changes
-
 - Remove cassandra SASI indices [#1328](https://github.com/jaegertracing/jaeger/pull/1328)
 
 Migration Path:
 
 1. Run `plugin/storage/cassandra/schema/migration/v001tov002part1.sh` which will copy dependencies into a csv, update the `dependency UDT`, create a new `dependencies_v2` table, and write dependencies from the csv into the `dependencies_v2` table.
-2. Run the collector and query services with the cassandra flag `sasi-disabled=true` which will update jaeger to read from the new `dependencies_v2` table.
+2. Run the collector and query services with the cassandra flag `enable-dependencies-v2=true` which will update jaeger to write and read to and from the new `dependencies_v2` table.
 3. Update [spark job](https://github.com/jaegertracing/spark-dependencies) to write to the new `dependencies_v2` table.
 4. Run `plugin/storage/cassandra/schema/migration/v001tov002part2.sh` which will DELETE the old dependency table and the SASI index.
 
-Users who wish to continue to use the SASI indices don't have to do anything as the cassandra flag `sasi-disabled` will default to false. Users may migrate on their own timeline however new features will be built solely on the `dependencies_v2` table.
+Users who wish to continue to use the v1 table don't have to do anything as the cassandra flag `enable-dependencies-v2` will default to false. Users may migrate on their own timeline however new features will be built solely on the `dependencies_v2` table. In the future, we will remove support for v1 completely.
+
+##### Breaking Changes
 
 ##### New Features
 
--- a/model/dependencies.go
+++ b/model/dependencies.go
@@ -30,8 +30,8 @@ type DependencyLink struct {
 	Source    DependencyLinkSource `json:"source"`
 }
 
-// Sanitize sanitizes the DependencyLink.
-func (d DependencyLink) Sanitize() DependencyLink {
+// ApplyDefaults applies defaults to the DependencyLink.
+func (d DependencyLink) ApplyDefaults() DependencyLink {
 	if d.Source == "" {
 		d.Source = JaegerDependencyLinkSource
 	}
--- a/model/dependencies.go
+++ b/model/dependencies.go
@@ -14,9 +14,26 @@
 
 package model
 
+// DependencyLinkSource is the source of data used to generate the dependencies.
+type DependencyLinkSource string
+
+const (
+	// JaegerDependencyLinkSource describes a dependency diagram that was generated from Jaeger traces.
+	JaegerDependencyLinkSource = DependencyLinkSource("jaeger")
+)
+
 // DependencyLink shows dependencies between services
 type DependencyLink struct {
-	Parent    string `json:"parent"`
-	Child     string `json:"child"`
-	CallCount uint64 `json:"callCount"`
+	Parent    string               `json:"parent"`
+	Child     string               `json:"child"`
+	CallCount uint64               `json:"callCount"`
+	Source    DependencyLinkSource `json:"source"`
+}
+
+// Sanitize sanitizes the DependencyLink.
+func (d DependencyLink) Sanitize() DependencyLink {
+	if d.Source == "" {
+		d.Source = JaegerDependencyLinkSource
+	}
+	return d
 }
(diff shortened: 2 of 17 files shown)


plugin/storage/cassandra/dependencystore/storage.go
```diff
@@ -26,60 +28,96 @@ import (
...
-	query := s.session.Query(depsInsertStmt, ts, ts, deps)
 	return s.dependenciesTableMetrics.Exec(query, s.logger)
 }
 
 // GetDependencies returns all interservice dependencies
 func (s *DependencyStore) GetDependencies(endTs time.Time, lookback time.Duration) ([]model.DependencyLink, error) {
-	query := s.session.Query(depsSelectStmt, endTs.Add(-1*lookback), endTs)
+	startTs := endTs.Add(-1 * lookback)
+	var query cassandra.Query
+	switch s.indexMode {
```
vprithvi: Would `GetDependencies` be simplified by omitting this switch? Instead we might read from SASIDisabled first, and if that fails we can read from SASIEnabled?
When migrating, this would mean that users don't need to redeploy query after moving over the dependency table

black-adder: I'm not a fan of adding the extra roundtrip time. Additionally, if the user has opted in to migrating to the new table, redeploying the query service will not be the most strenuous thing.

plugin/storage/cassandra/dependencystore/storage.go
```diff
@@ -26,60 +28,96 @@ import (
...
 	logger                   *zap.Logger
+	indexMode                IndexMode
 }
 
 // NewDependencyStore returns a DependencyStore
 func NewDependencyStore(
 	session cassandra.Session,
 	metricsFactory metrics.Factory,
 	logger *zap.Logger,
+	indexMode IndexMode,
```
vprithvi: Is there anything preventing someone from passing in an undefined `IndexMode`, like IndexMode(10) for e.g.?

plugin/storage/cassandra/factory.go
```diff
@@ -106,7 +106,14 @@ func (f *Factory) CreateSpanWriter() (spanstore.Writer, error) {
 
 // CreateDependencyReader implements storage.Factory
 func (f *Factory) CreateDependencyReader() (dependencystore.Reader, error) {
-	return cDepStore.NewDependencyStore(f.primarySession, f.primaryMetricsFactory, f.logger), nil
+	var (
+		sasiDisabled = f.Options.GetPrimary().DependencySASIDisabled
```
vprithvi: I think the following reads simpler:
 
```
indexMode := cDepStore.SASIEnabled
if f.Options.GetPrimary().DependencySASIDisabled {
indexMode = cDepStore.SASIDisabled
}
```

pkg/cassandra/config/config.go
```diff
@@ -27,20 +27,21 @@ import (
...
+	Timeout                time.Duration `validate:"min=500"`
+	ReconnectInterval      time.Duration `validate:"min=500" yaml:"reconnect_interval"`
+	SocketKeepAlive        time.Duration `validate:"min=0" yaml:"socket_keep_alive"`
+	MaxRetryAttempts       int           `validate:"min=0" yaml:"max_retry_attempt"`
+	ProtoVersion           int           `yaml:"proto_version"`
+	Consistency            string        `yaml:"consistency"`
+	Port                   int           `yaml:"port"`
+	Authenticator          Authenticator `yaml:"authenticator"`
+	DisableAutoDiscovery   bool          `yaml:"disable_auto_discovery"`
+	DependencySASIDisabled bool          `yaml:"dependency_sasi_disabled"`
```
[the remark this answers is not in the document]
black-adder: i'd rather it be more explicit but I can simplify

plugin/storage/cassandra/dependencystore/storage.go
```diff
@@ -26,60 +28,96 @@ import (
...
 	deps := make([]Dependency, len(dependencies))
 	for i, d := range dependencies {
-		deps[i] = Dependency{
+		dep := Dependency{
 			Parent:    d.Parent,
 			Child:     d.Child,
 			CallCount: int64(d.CallCount),
 		}
+		if s.indexMode == SASIEnabled {
+			dep.Source = string(d.Source)
```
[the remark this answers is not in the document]
black-adder: yes, if you attempt to add the source field to a schema that doesn't support it, c* will error. And good catch, it should be for disabled

plugin/storage/cassandra/dependencystore/storage.go
black-adder: ill validate here

plugin/storage/cassandra/dependencystore/storage.go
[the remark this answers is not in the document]
vprithvi: Yes - but that doesn't answer my question; the query string for SASIDisabled doesn't include this field anyways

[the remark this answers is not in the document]
black-adder: i'll test it and let you know

plugin/storage/cassandra/dependencystore/model_test.go
```diff
@@ -24,19 +24,21 @@ import (
 
 func TestDependencyUDT(t *testing.T) {
 	dependency := &Dependency{
-		Parent:    "goo",
-		Child:     "gle",
+		Parent:    "bi",
+		Child:     "ng",
```
yurishkuro: quit raising entropy

CHANGELOG.md
```diff
@@ -8,6 +8,17 @@ Changes by Version
 
 ##### Breaking Changes
```
yurishkuro: I am guessing this is technically a non-breaking change. It will become breaking once we start defaulting to _v2

plugin/storage/cassandra/dependencystore/storage.go
```diff
@@ -26,60 +26,110 @@ import (
...
+// IsValid returns true if the IndexMode is a valid one.
+func (i IndexMode) IsValid() bool {
+	return i < end
+}
+
 const (
-	depsInsertStmt = "INSERT INTO dependencies(ts, ts_index, dependencies) VALUES (?, ?, ?)"
-	depsSelectStmt = "SELECT ts, dependencies FROM dependencies WHERE ts_index >= ? AND ts_index < ?"
+	// SASIEnabled is used when the dependency table is SASI indexed.
+	SASIEnabled IndexMode = iota
```
yurishkuro: would it not make sense to simply refer to this as `v2` throughout, including constants, variables, the CLI flags? SASI then becomes a side effect that the user doesn't really need to know about.

plugin/storage/cassandra/dependencystore/storage.go
```diff
@@ -26,60 +26,110 @@ import (
...
 func (s *DependencyStore) WriteDependencies(ts time.Time, dependencies []model.DependencyLink) error {
 	deps := make([]Dependency, len(dependencies))
 	for i, d := range dependencies {
-		deps[i] = Dependency{
+		dep := Dependency{
 			Parent:    d.Parent,
 			Child:     d.Child,
 			CallCount: int64(d.CallCount),
 		}
+		if s.indexMode == SASIDisabled {
```
yurishkuro: another argument to refer to this change as "dependencies v2" rather than tying to SASI.

CHANGELOG.md
```diff
@@ -6,6 +6,17 @@ Changes by Version
 
 #### Backend Changes
 
+- Remove cassandra SASI indices [#1328](https://github.com/jaegertracing/jaeger/pull/1328)
+
+Migration Path:
+
+1. Run `plugin/storage/cassandra/schema/migration/v001tov002part1.sh` which will copy dependencies into a csv, update the `dependency UDT`, create a new `dependencies_v2` table, and write dependencies from the csv into the `dependencies_v2` table.
+2. Run the collector and query services with the cassandra flag `enable-dependencies-v2=true` which will update jaeger to write and read to and from the new `dependencies_v2` table.
```
yurishkuro: > cassandra flag `enable-dependencies-v2=true`

please specify the exact flag name

> which will update jaeger to write

which will instruct jaeger to write

CHANGELOG.md
```diff
@@ -6,6 +6,17 @@ Changes by Version
...
+3. Update [spark job](https://github.com/jaegertracing/spark-dependencies) to write to the new `dependencies_v2` table.
```
yurishkuro: this should rather say "update Spark job to version N"

plugin/storage/cassandra/dependencystore/model.go
```diff
@@ -36,6 +37,8 @@ func (d *Dependency) MarshalUDT(name string, info gocql.TypeInfo) ([]byte, error
 		return gocql.Marshal(info, d.Child)
 	case "call_count":
 		return gocql.Marshal(info, d.CallCount)
+	case "source":
```
yurishkuro: what happens if you run this code against the schema where UDT was not upgraded? Will gocql simply not invoke this function for "source"?

model/dependencies.go
```diff
@@ -14,9 +14,26 @@
...
-	Child     string `json:"child"`
-	CallCount uint64 `json:"callCount"`
+	Parent    string               `json:"parent"`
+	Child     string               `json:"child"`
+	CallCount uint64               `json:"callCount"`
+	Source    DependencyLinkSource `json:"source"`
+}
+
+// Sanitize sanitizes the DependencyLink.
+func (d DependencyLink) Sanitize() DependencyLink {
```
yurishkuro: ApplyDefaults

plugin/storage/cassandra/schema/migration/v001tov002part1.sh
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/env bash
+
```
yurishkuro: `set -euo pipefail`

plugin/storage/cassandra/schema/migration/v001tov002part1.sh
```diff
@@ -0,0 +1,77 @@
...
+    exit 1
+fi
+
+if [ ${row_count} -ne $(wc -l dependencies.csv | cut -f 1 -d ' ') ]; then
+    echo "Number of rows in file is not equal to number of rows in cassandra"
+    exit 1
+fi
+
+
+while IFS="," read ts dependency; do bucket=`date +"%Y%m%d" -d "$ts"`; echo "$bucket,$ts,$dependency"; done < dependencies.csv > dependencies_datebucket.csv
```
yurishkuro: can you write it in normal multi-line syntax?

plugin/storage/cassandra/schema/migration/v001tov002part2.sh
```diff
@@ -0,0 +1,43 @@
...
+    usage "invalid characters in KEYSPACE=$keyspace parameter, please use letters, digits or underscores"
+fi
+
+
+row_count=$(cqlsh -e "select count(*) from $keyspace.dependencies;"|head -4|tail -1| tr -d ' ')
+
+echo "About to delete $row_count rows."
+confirm
+
+cqlsh -e "DROP INDEX IF EXISTS $keyspace.ts_index"
```
yurishkuro: `$keyspace.ts_index`

naming 🤦‍♂️

CHANGELOG.md
black-adder: I'll punt on this until I've actually made the change in spark.

plugin/storage/cassandra/dependencystore/model.go
black-adder: yes

plugin/storage/cassandra/dependencystore/storage.go
```diff
@@ -26,29 +26,59 @@ import (
 	casMetrics "github.com/jaegertracing/jaeger/pkg/cassandra/metrics"
 )
 
+// Version determines which version of the dependencies table to use.
+type Version int
+
+// IsValid returns true if the Version is a valid one.
+func (i Version) IsValid() bool {
+	return i < end
```
yurishkuro: I would prefer a less confusing name than 'end', e.g. 'versionEnumEnd'

yurishkuro: also, `Version(-1)` is valid according to this function

PR #1329, opened 2019-02-12
Use roundRobin balancing even if only one hostname

benley:
- When using GRPC between jaeger-agent and jaeger-collector, one must currently enumerate each backend on the commandline in order to get any load balancing.  GRPC can do better with its pluggable resolver modules.

- If you run jaeger-agent with `--reporter.grpc.host-port dns:///jaeger-collector-hostname:14250` (note the `dns:///` part!), the GRPC client resolves the given hostname to one or more IPs and can load balance among the backend addresses it finds.  It even looks up and obeys SRV records, if they exist.  This change makes jaeger-agent use round-robin load balancing in that situation, rather than the default of `pick_first`.  If there is only one backend, or the user doesn't explicitly put `dns:///` before the hostname, this will have no effect.

See https://www.marwan.io/blog/grpc-dns-load-balancing for more details.

Question: should this `dns` resolver style be the default behaviour? That can be done easily as well.


jpkrohling:
In time: you'd get bonus points if you could add this to the documentation and/or to the help options.

benley:
> In time: you'd get bonus points if you could add this to the documentation and/or to the help options.

How does this look? https://github.com/jaegertracing/documentation/pull/212

yurishkuro:
@benley awesome!

did you see any tickets that might be closed by this? I'm pretty sure we had these discussions.

Also, would you like to add Postmates to #207 / ADOPTERS.md ?

benley:
> did you see any tickets that might be closed by this?

https://github.com/jaegertracing/jaeger/issues/213 is closely related but I'm not sure if this really closes it. We're adding some much-needed functionality for grpc balancing but DNS service discovery isn't going to be as reliable as, say, Zookeeper or etcd or Kubernetes API watches for noticing when the set of backends changes.

> Also, would you like to add Postmates to #207 / ADOPTERS.md ?

Sure, I will do that!

PR #1336, opened 2019-02-12
Update Sarama to 1.20.1

### LINKED ISSUE #1335: Ingester: Message was too large, server rejected it to avoid allocation error.
Use Kafka in between Jaeger Collector and Jaeger Ingester

We see a great number of errors like the following (running v1.9.0):
```
{"level":"error","ts":1550003610.8029132,"caller":"kafka/writer.go:59","msg":"kafka server: Message was too large, server rejected it to avoid allocation error.","stacktrace":"github.com/jaegertracing/jaeger/plugin/storage/kafka.NewSpanWriter.func2\n\t/home/travis/gopath/src/github.com/jaegertracing/jaeger/plugin/storage/kafka/writer.go:59"}
```

Also, the topic was created with the default `max.message.bytes` value: `1000012` and the broker has a bigger `max.messages.bytes` then the topic.

At first I thought the issue was that we received a single big span, but even after switching from `json` to `protobuf` we are still seeing it (I presume `protobuf` is slightly more compact than `json`).

I started digging and I think the problem is actually with Sarama: https://github.com/Shopify/sarama/issues/1140

This seems to be fixed by this: https://github.com/Shopify/sarama/commit/4c93f1f002740d2d8604ba4f82d2c87a4d49a1ba

1. Upgrade to Sarama to 1.20.1 (currently at 1.20.0)
2. Ideally, I would like to expose the `MaxMessageBytes` config and the `CompressionCodec` as well
3. Expose more stats from Sarama (these would be helpful https://github.com/Shopify/sarama/blob/46cf3e2cf1acef7876068f66cf69ec31aad2d0b2/produce_request.go#L78-L79)

- I plan to send a PR for 1 and another one for 2 and 3

ledor473:
- Resolves #1335 

Update Sarama to latest version that includes a related fix as per: https://github.com/Shopify/sarama/releases

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -121,4 +121,4 @@ required = [
 
 [[constraint]]
   name = "github.com/Shopify/sarama"
-  version = "1.20.0"
+  version = "1.20.1"


yurishkuro:
Resolves #1335

Issue #1331, opened 2019-02-12
Failed to access Cassandra storage backend which is connected right after its reboot


iori-yja:
## Requirement - what kind of business use case are you trying to solve?
I put a cassandra server and jaeger-collector & -query all together in one server. After the server's reboot, the system is wanted to boot up gracefully and automatically.

### current behavior
After the reboot, cassandra boots up after initialization work besides jaeger-collector and -query boots up very quickly. By this time gap, jaeger containers boot & exit for a while until cassandra woke up. After storage server's boot, jaeger containers succeeds to connect the server.

## Problem - what in Jaeger blocks you from solving the requirement?

Even though -collector and -query eventually succeeded to connect to the storage, they fall into a _broken_ state.
_broken_ here means that the jaeger-collector and -query couldn't retrieve the data due to `gocql: no hosts available in the pool`, even though the `storageFactory` has succeeded to initialize. Restarting the containers can solve this behavior for now. (Reproducibility = always)

## Proposal - what do you suggest to solve the problem or improve the existing situation?

1. The most cheapest workaround is to exit the process on "no hosts available" error.
1. Graceful workaround is to fix the server status check correctly.
1. Another graceful check is to exit on `archiveOptions()` checks' failure.

## Log

Here is a part of the long log outputs which starts from the last failed attempt of storage client initialization in the restarting loop. 

```
{"level":"info","ts":1549948198.8485098,"caller":"healthcheck/handler.go:99","msg":"Health Check server started","http-port":16687,"status":"unavailable"}
2019/02/12 05:09:58 gocql: unable to dial control conn 172.17.0.1: dial tcp 172.17.0.1:9042: connect: connection refused
{"level":"fatal","ts":1549948198.8652992,"caller":"query/main.go:107","msg":"Failed to init storage factory","error":"gocql: unable to create session: control: unable to connect to initial hosts: dial tcp 172.17.0.1:9042: connect: connection refused","stacktrace":"main.main.func1\n\t/home/travis/gopath/src/github.com/
jaegertracing/jaeger/cmd/query/main.go:107\ngithub.com/jaegertracing/jaeger/vendor/github.com/spf13/cobra.(*Command).execute\n\t/home/travis/gopath/src/github.com/jaegertracing/jaeger/vendor/github.com/spf13/cobra/command.go:762\ngithub.com/jaegertracing/jaeger/vendor/github.com/spf13/cobra.(*Command).ExecuteC\n\t/hom
[...]

PR #1338, opened 2019-02-13
WIP: Unhealthy on storage blackout

iori-yja:
The health status doesn't tell the healthiness of the server after initialization.
https://github.com/jaegertracing/jaeger/issues/1331#issuecomment-462670392

### Goal
- a part of the server can declare its _unhealthy_ state
- if one of essential parts declares unhealthy, then server is unhealthy
- configurable what is the essential part


yurishkuro:
I am not sure this approach works. In particular, in Cassandra it only sets check-passed once the Session is created. But if Cassandra is unavailable, the factory will simply return an error and abort the application, so what does the health check actually provide?

I think the main change here is not so much connecting storage status to the health check, but actually _having storage status_. Right now the status has only one value: ready. If not ready - app dies.

iori-yja:
What does "abort the application" mean? Actually, it don't die after the first connection succeeded. Particularly, if the application connected to a cassandra server _right after_ it has got available, the data gets never be available because connection pool is empty (or gone).

Possibly it can be fixed by adding another check routine in the factory, though...

iori-yja:
> I am not sure this approach works. In particular, in Cassandra it only sets check-passed once the Session is created. But if Cassandra is unavailable, the factory will simply return an error and abort the application, so what does the health check actually provide?

I understand this point. I am working on propagating `StatusReporter` to reader/writer part of store (eg. spanstore), but this is a little bit messy.
Another approach is to make it just die after expiring `max-retry-attempts` which sounds more easy. But there must be pros/cons.

> If not ready - app dies.
I am going to add an option to specify which part is required to be "ready". If "Initialization" and "Storage" are required, storage must be ready and healthy for the application to be ready. If it is only "Initialization", the behavior is same as now.

iori-yja:
Following DCO guide ruined the history....

iori-yja:
Sorry, I said "never", but it turned out not true. Leaving long time (sorry, I've failed to measure) can solve the problem. So the problem only left here is the health status reports ok before the database gets ready.

plugin/storage/cassandra/factory.go
```diff
@@ -71,10 +73,11 @@ func (f *Factory) InitFromViper(v *viper.Viper) {
...
 // Initialize implements storage.Factory
-func (f *Factory) Initialize(metricsFactory metrics.Factory, logger *zap.Logger) error {
+func (f *Factory) Initialize(metricsFactory metrics.Factory, logger *zap.Logger, reporter healthcheck.StatusReporter) error {
 	f.primaryMetricsFactory = metricsFactory.Namespace(metrics.NSOptions{Name: "cassandra", Tags: nil})
 	f.archiveMetricsFactory = metricsFactory.Namespace(metrics.NSOptions{Name: "cassandra-archive", Tags: nil})
 	f.logger = logger
+	f.StatusReporter = reporter
 
 	primarySession, err := f.primaryConfig.NewSession()
 	if err != nil {
```
yurishkuro: if Cassandra is not ready, this will cause the application to exit

iori-yja: Sorry, I took a while to move back on this.
 
I think my previous explanation was bad.
Actually, the symptom is triggered by cassandra restart. In that case, the session pool become empty after the initialization. In addition, gocql doesn't deal with the reconnection, which makes the jaeger services unavailable after that.

jpkrohling:
@iori-yja are you still interested in this one?

PR #1341, opened 2019-02-13
Use Zipkin annotations if the timestamp is zero

geobeau:

When the span.Timestamp is 0, jaeger write the span with the timestamp to zero even when we have more informations in the annotations.

If the timestamp is zero, we determine if the span is a client or a server and extract timestamp and information from there

### DIFF
--- a/model/converter/thrift/zipkin/to_domain.go
+++ b/model/converter/thrift/zipkin/to_domain.go
@@ -200,20 +200,19 @@ func (td toDomain) getStartTimeAndDuration(zSpan *zipkincore.Span) (int64, int64
 		if cs != nil {
 			timestamp = cs.Timestamp
 			cr := td.findAnnotation(zSpan, zipkincore.CLIENT_RECV)
-			if cr != nil {
+			if cr != nil && duration == 0 {
 				duration = cr.Timestamp - cs.Timestamp
 			}
 		} else if sr != nil {
 			timestamp = sr.Timestamp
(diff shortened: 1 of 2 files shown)


model/converter/thrift/zipkin/to_domain.go
```diff
@@ -188,6 +190,31 @@ func (td toDomain) getFlags(zSpan *zipkincore.Span) model.Flags {
...
+func (td toDomain) getStartTimeAndDuration(zSpan *zipkincore.Span) (int64, int64) {
+	timestamp := zSpan.GetTimestamp()
+	duration := zSpan.GetDuration()
+	if timestamp == 0 {
+		cs := td.findAnnotation(zSpan, zipkincore.CLIENT_SEND)
+		sr := td.findAnnotation(zSpan, zipkincore.SERVER_RECV)
+		if cs != nil {
+			timestamp = cs.Timestamp
+			cr := td.findAnnotation(zSpan, zipkincore.CLIENT_RECV)
+			if cr != nil {
```
yurishkuro: should this perhaps also check for `duration == 0`?

geobeau: Done

yurishkuro:
@geobeau thanks for the PR?  Was there a related ticket, or just the PR?

geobeau:
No, I encountered this issue so I fixed it directly. Should I have opened one?

Thanks to both of you 👍

yurishkuro:
no, PR is fine, thanks.

PR #1345, opened 2019-02-14
Change http to https for security links

trungnvfet:
- Fix http link to https link for secure. 

- Modify http to https in the links of project. Have some http links does not redirect into https.


yurishkuro:
I'll merge without DCO since these are not source code files

PR #1350, opened 2019-02-14
adding a wrapping writer that automatic drop spans before writing wit…

### LINKED ISSUE #1100: Add a rate limiter to jaeger-ingester

guanw:

* Partially resolves #1100 

* Instead of doing ratelimiting based on the volume, simply adding DownSamplingWriter to wrap other writers so user could specify a float ratio number between 0 ~ 1.0 to drop certain percentage of spans.

Ratio parameter is specified by passing `AUTOMATE_DROP_PERCENTAGE` as env variable.

### DIFF
--- a/storage/spanstore/automate_dropping_writer.go
+++ b/storage/spanstore/automate_dropping_writer.go
@@ -15,6 +15,7 @@ type AutomateDroppingWriter struct {
 
 // NewAutomateDroppingWriter creates a AutomateDroppingWriter
 func NewAutomateDroppingWriter(spanWriter Writer, percentage float64) *AutomateDroppingWriter {
+	// span with TraceID.Low below this threshold won't be written
 	threshold := uint64(percentage * float64(math.MaxUint64))
 	return &AutomateDroppingWriter{
 		spanWriter: spanWriter,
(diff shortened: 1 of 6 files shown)


yurishkuro:
I would rename to DownSamplingWriter

storage/spanstore/automate_dropping_writer.go
```diff
@@ -0,0 +1,33 @@
...
+	return &AutomateDroppingWriter{
+		spanWriter: spanWriter,
+		threshold:  threshold,
+	}
+}
+
+// WriteSpan calls WriteSpan on wrapped span writer.
+func (c *AutomateDroppingWriter) WriteSpan(span *model.Span) error {
+	// TraceID consists of two random generated uint64, we use one of them to decide if span will be written
+	if c.threshold != 0 && span.TraceID.Low < c.threshold {
```
yurishkuro: No, let's not do that. This type of logic is too easy to write and thus is being used elsewhere, in particular in the clients that try to minimize computations.

A better approach is to compute a hash of the trace ID with some fixed salt and compare the hash against the ratio. Btw that kind of logic can be extracted into a util type.

guanw: We are not doing encryption/decryption. Is salt really necessary here?

yurishkuro: yes, because many components in the collection chain might be doing downsampling, theoretically using the same algorithm. Using a salt provides a different distribution of the hash values.

PR #1353, opened 2019-02-15
Add a Downsampling writer that drop a percentage of spans

### LINKED ISSUE #1100: Add a rate limiter to jaeger-ingester

guanw:
Accidentally removed my forked jaeger project so have to create this new pull request.. 

* Partially resolves #1100 

* Instead of doing ratelimiting based on the volume, simply adding DownSamplingWriter to wrap other writers so user could specify a float ratio number between 0 ~ 1.0 to drop certain percentage of spans.

Ratio parameter is specified by passing `DOWN_SAMPLING_RATIO` as env variable.


plugin/storage/factory.go
```diff
@@ -118,9 +118,15 @@ func (f *Factory) CreateSpanWriter() (spanstore.Writer, error) {
 		writers = append(writers, writer)
 	}
 	if len(f.SpanWriterTypes) == 1 {
```
yurishkuro: a) you can DRY it by first defining `topWrite` as either `writers[0]` or new sequential, and only then applying downsampler
b) shouldn't downsampler apply conditionally depending on the configuration (DownSamplingRatio)?

plugin/storage/factory_config.go
```diff
@@ -29,13 +30,21 @@ const (
 	DependencyStorageTypeEnvVar = "DEPENDENCY_STORAGE_TYPE"
 
 	spanStorageFlag = "--span-storage.type"
+
+	// DownSamplingRatio is the name of the env var that defines the ratio of spans we would drop
+	DownSamplingRatio = "DOWN_SAMPLING_RATIO"
```
yurishkuro: the name / description is not clear. Does the value of 0.1 mean we _drop_ 10% or that we _keep_ 10%?

yurishkuro: NB: I would suggest to be consistent with the regular probabilistic sampler here, i.e. 0.1 means we only _keep_ 10%. But the comment & CLI flag descriptions should be more clear.

plugin/storage/factory_config.go
```diff
@@ -69,11 +78,22 @@ func FactoryConfigFromEnvAndCLI(args []string, log io.Writer) FactoryConfig {
 	if depStorageType == "" {
 		depStorageType = spanWriterTypes[0]
 	}
+	ratioEnvVariable := os.Getenv(DownSamplingRatio)
+	// Default ratio to drop spans is 0
```
yurishkuro: see, this is the opposite of the normal sampling logic.

plugin/storage/factory_config.go
```diff
@@ -69,11 +78,22 @@ func FactoryConfigFromEnvAndCLI(args []string, log io.Writer) FactoryConfig {
...
+	downSamplingRatio, err := strconv.ParseFloat(ratioEnvVariable, 64)
+	if err != nil {
+		downSamplingRatio = 0
+	}
+	// Default salt is empty string
+	downSamplingHashSalt := os.Getenv(DownSamplingHashSalt)
```
yurishkuro: provide a default.

storage/spanstore/downsampling_writer.go
```diff
@@ -0,0 +1,65 @@
...
+		hashSalt:   downSamplingOptions.HashSalt,
+	}
+}
+
+// WriteSpan calls WriteSpan on wrapped span writer.
+func (c *DownSamplingWriter) WriteSpan(span *model.Span) error {
+	if c.threshold == 0 {
+		return c.spanWriter.WriteSpan(span)
+	}
+	hashVal := HashBytes(c.fnvHash, []byte(c.hashSalt+span.TraceID.String()))
```
yurishkuro: try without memory allocations here. TraceID is binary, no need to convert to string.

storage/spanstore/utils.go
```diff
@@ -0,0 +1,17 @@
+package spanstore
+
+import (
+	"hash"
+	"math"
+)
+
+// HashBytes returns the uint64 hash value of bytes slice
+func HashBytes(h hash.Hash64, bytes []byte) uint64 {
```
yurishkuro: no need to make this a public function or a separate file, is there?

storage/spanstore/utils.go
```diff
@@ -0,0 +1,17 @@
...
+	h.Reset()
```
yurishkuro: this is not thread-safe

yurishkuro: you may want to use a sync.Pool to avoid using lock or allocating too many hash objects.

plugin/storage/factory_config.go
```diff
@@ -69,11 +78,22 @@ func FactoryConfigFromEnvAndCLI(args []string, log io.Writer) FactoryConfig {
 	if depStorageType == "" {
 		depStorageType = spanWriterTypes[0]
 	}
+	ratioEnvVariable := os.Getenv(DownSamplingRatio)
+	// Default ratio to drop spans is 0
+	downSamplingRatio, err := strconv.ParseFloat(ratioEnvVariable, 64)
+	if err != nil {
```
black-adder: we should log the error

storage/spanstore/downsampling_writer.go
```diff
@@ -0,0 +1,65 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
```
black-adder: 2019

storage/spanstore/downsampling_writer.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	"hash"
+	"hash/fnv"
+	"math"
+
+	"github.com/jaegertracing/jaeger/model"
+)
+
+// DownSamplingWriter is a span Writer that tries to drop spans with
```
black-adder: `DownSamplingWriter is a span Writer that drops spans with a predefined downSamplingRatio.`

storage/spanstore/downsampling_writer_test.go
```diff
@@ -0,0 +1,78 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
```
black-adder: 2019, did you copy paste this? It shouldn't be necessary, the make fmt step should add the correct license for you

plugin/storage/factory_config.go
```diff
@@ -29,13 +30,21 @@ const (
 	DependencyStorageTypeEnvVar = "DEPENDENCY_STORAGE_TYPE"
 
 	spanStorageFlag = "--span-storage.type"
+
+	// DownSamplingRatio is the name of the env var that defines the ratio of spans we would drop
```
black-adder: https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences let's try to follow golang conventions for comments. We should end everything with a period.

storage/spanstore/downsampling_writer.go
```diff
@@ -0,0 +1,65 @@
...
+// WriteSpan calls WriteSpan on wrapped span writer.
+func (c *DownSamplingWriter) WriteSpan(span *model.Span) error {
+	if c.threshold == 0 {
+		return c.spanWriter.WriteSpan(span)
+	}
+	hashVal := HashBytes(c.fnvHash, []byte(c.hashSalt+span.TraceID.String()))
+	if hashVal < c.threshold {
+		// Downsampling writer prevents writing span when hashVal falls
+		// in the range of (0, threshold)
+		return nil
```
black-adder: we should emit a metric

storage/spanstore/utils.go
```diff
@@ -0,0 +1,17 @@
+package spanstore
```
black-adder: this file is missing license.

black-adder: not a big fan of the `utils.go` filename, it's a lazy catch all. We can call this hash.go

storage/spanstore/utils.go
black-adder: Let's not increase API scope, this function can be private until we need it to be public.

storage/spanstore/utils.go
```diff
@@ -0,0 +1,17 @@
...
+	"hash"
+	"math"
+)
+
+// HashBytes returns the uint64 hash value of bytes slice
+func HashBytes(h hash.Hash64, bytes []byte) uint64 {
+	h.Reset()
+	_, err := h.Write(bytes)
+	if err != nil {
+		// No downsampling when there's error hashing
```
black-adder: metric please

storage/spanstore/utils.go
black-adder: Is this thread safe?

storage/spanstore/downsampling_writer_test.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+	downSamplingOptions := spanstore.DownSamplingOptions{
+		Ratio:    1,
+		HashSalt: "jaeger-test",
+	}
+	c := NewDownSamplingWriter(&noopWriteSpanStore{}, downSamplingOptions)
+	assert.NoError(t, c.WriteSpan(span))
+}
+
+func TestDownSamplingWriter_HashBytes(t *testing.T) {
```
black-adder: personally I don't think this test is necessary, I trust that `New64a` works as intended

storage/spanstore/downsampling_writer_test.go
```diff
@@ -0,0 +1,78 @@
...
+			High: uint64(1),
+		},
+	}
+
+	h := fnv.New64a()
+	hash1 := HashBytes(h, []byte(span.TraceID.String()))
+	hash2 := HashBytes(h, []byte(span.TraceID.String()))
+	assert.Equal(t, hash1, hash2)
+}
+func TestDownSamplingWriter_RandomHash(t *testing.T) {
```
black-adder: This test feels flaky, maybe change this to a benchmark instead?

plugin/storage/factory.go
guanw: a) will refactor
b) it won't apply if the ratio passed is 1, this is decided within the DownSamplerWriter. We will pass 1.0 as default ratio which means no down sampling at all.

black-adder: also, if we want to dynamically adjust the sampling rate, it's better to always create a DownSamplerWriter and handle the logic there.

plugin/storage/factory_config.go
guanw: default is empty string, this case will be covered by err checking from `strconv.ParseFloat`

storage/spanstore/utils.go
guanw: I notice `composite.go` and `composite_test.go` belong to different packages so I'm copying that logic to downsampling_writer.go and downsampling_writer_test.go. Even if I move the hash function inside DownSamplingWriter, test file won't have access to it if name starts with lowercase? So I guess uppercase is necessary? Please advise.

black-adder: not sure I follow, in golang, private objects are package scoped. Since both `downsampling_writer.go` and `downsampling_writer_test.go` are in the same package, `hashBytes` being private is fine.

guanw: do you mean package name should not be `spanstore_test` for `composite_test.go` ?

plugin/storage/factory_config.go
guanw: I don't see a logger being passed here, is there an example for how we do that?

storage/spanstore/downsampling_writer_test.go
guanw: ic, i will remove those license then.

storage/spanstore/utils.go
guanw: this is because I'm using different package name for downsampling_writer_test.go and downsampling_writer.go. Let me know if that's necessary or not.

storage/spanstore/downsampling_writer_test.go
guanw: Will remove

storage/spanstore/downsampling_writer_test.go
guanw: ok, btw, are we also committing benchmark file to oss jaeger? I don't see any,

storage/spanstore/utils.go
black-adder: AFAIK you use the `_test` suffix in the package name if you ONLY want to test the public interfaces. I don't really see a point for `downsampling_writer.go`

storage/spanstore/downsampling_writer_test.go
black-adder: we should start getting into the habit

plugin/storage/factory_config.go
black-adder: look above on L71

storage/spanstore/downsampling_writer.go
yurishkuro: +1

storage/spanstore/utils.go
guanw: Same question as logger, doesn't seem like metricsFactory is being passed down from consumer?

storage/spanstore/utils.go
guanw: Got it.

storage/spanstore/downsampling_writer_test.go
guanw: @yurishkuro Can you confirm if I should commit the benchmark file here?

yurishkuro: if you wish

guanw:
@yurishkuro I tried running some benchmark against sync.Pool, something like this. 
image

Against another version where I just use mutex for locking, it seems the simple mutex beats sync.Pool both in time and memory. I think using sync.Pool here is not necessary since we don't need to do `fnv.New64a()` everytime hashing: Just need to do `fnv.reset()` to reset it which doesn't involve memory allocation thus sync.Pool won't help here? So i prefer we can just use mutex here.

The benchmark with mutex for running 100 goroutines where each goroutine WriteSpan 10000 times cost 210547ns to finish and 63310Bytes allocation which seem fine to me.
image

storage/spanstore/utils.go
black-adder: also, do you have to call `Reset` every time? According to the wiki: `Start with an initial hash value of FNV offset basis. For each byte in the input, multiply hash by the FNV prime, then XOR it with the byte from the input.` so it looks like you can keep writing rather than resetting.

plugin/storage/factory_config.go
```diff
@@ -29,13 +30,21 @@ const (
...
+	DownSamplingRatio = "DOWN_SAMPLING_RATIO"
+
+	// DownSamplingHashSalt is the name of the env var that stores the hash salt for down sampling
+	DownSamplingHashSalt = "DOWN_SAMPLING_HASH_SALT"
```
vprithvi: nit: downsampling is usually spelled as a single word, so we can replace `DownSampling` with `Downsampling`.

yurishkuro:
You *should* use reset() even with the pool. The point of the pool is to avoid memory allocations.

But in any case, we have an issue with the overall design where we run many go-routines working independently on saving spans, but we have these synchronization points like your mutex that block all writers for little reason. Unfortunately, fixing this requires significant re-design where we create writers per goroutine.

Given that both the explicit mutex and sync.Pool introduce a critical section, the main benefit of the pool should be in avoiding memory allocations while unblocking other goroutines invoking the hash. If you're not getting the benefits of the sync pool with your benchmark, I suspect the benchmark.

black-adder:
@guanw can you add your benchmarks so we can see? In your benchmark you maybe creating threads but not actually waiting for them to run before the benchmark resolves.

@yurishkuro `You _should_ use reset()` why?

black-adder:
If it's some crap about how one day aliens will invade and the underlying implementation will in the future require you to call reset before every call so we should prepare for it now...

yurishkuro:
> @yurishkuro `You _should_ use reset()` why?

as alternative to allocating a new buffer

black-adder:
> > @yurishkuro `You _should_ use reset()` why?
> 
> as alternative to allocating a new buffer

The underlying implementation of the hash is:

```
func (s *sum64a) Write(data []byte) (int, error) {
	hash := *s
	for _, c := range data {
		hash ^= sum64a(c)
		hash *= prime64
	}
	*s = hash
	return len(data), nil
}
```
where
```
type (
	sum64a  uint64
)
```

The underlying hash function doesn't use a buffer for representation. All reset does is:
```
func (s *sum64a) Reset()  { *s = 14695981039346656037 }
```

guanw:
I tried moving the fnv hash logic directly to our code and get rid of the sync.pool. It reduces the benchmark runtime 5x, I think we could just go with that?

guanw:
actually ignore my last comment. I'm still going with sync.Pool.

storage/spanstore/downsampling_writer_benchmark_test.go
```diff
@@ -0,0 +1,78 @@
...
+	b.ReportAllocs()
+
+	for j := 0; j < b.N; j++ {
+		for n := 0; n < numGoroutines; n++ {
+
+			go func(span *model.Span) {
+				for i := 0; i < writeTimes; i++ {
+					c.WriteSpan(span)
+				}
+				defer wg.Done()
```
black-adder: defer is unnecessary

storage/spanstore/downsampling_writer_benchmark_test.go
```diff
@@ -0,0 +1,78 @@
...
+	numGoroutines := 10000
+	writeTimes := 10
+	wg.Add(numGoroutines * b.N)
+	b.ResetTimer()
+	b.ReportAllocs()
+
+	for j := 0; j < b.N; j++ {
+		for n := 0; n < numGoroutines; n++ {
+
+			go func(span *model.Span) {
```
black-adder: you don't have to pass in span here

storage/spanstore/downsampling_writer_benchmark_test.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+type noopWriteSpanStore struct{}
+
+func (n *noopWriteSpanStore) WriteSpan(span *model.Span) error {
+	return nil
+}
+
+func BenchmarkDownSamplingWriter_WriteSpan(b *testing.B) {
```
black-adder: Can you leave in the code about the outcome of the benchmark?

plugin/storage/factory.go
```diff
@@ -117,10 +118,17 @@ func (f *Factory) CreateSpanWriter() (spanstore.Writer, error) {
 		}
 		writers = append(writers, writer)
 	}
+	var topWriter spanstore.Writer
```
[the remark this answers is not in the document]
guanw: seems added automatically.

storage/spanstore/downsampling_writer.go
```diff
@@ -0,0 +1,128 @@
...
+// DownSamplingOptions contains the options for constructing a DownSamplingWriter.
+type DownSamplingOptions struct {
+	Ratio          float64
+	HashSalt       string
+	MetricsFactory metrics.Factory
+}
+
+// NewDownSamplingWriter creates a DownSamplingWriter.
+func NewDownSamplingWriter(spanWriter Writer, downSamplingOptions DownSamplingOptions) *DownSamplingWriter {
+	threshold := uint64(downSamplingOptions.Ratio * float64(math.MaxUint64))
```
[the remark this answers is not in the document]
guanw: validation was done in factory_config.go, unless you think it's necessary we do extra validation here?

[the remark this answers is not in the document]
guanw: will add validation to factory_config.go.

storage/spanstore/downsampling_writer.go
```diff
@@ -0,0 +1,128 @@
...
+		// Drops spans when hashVal falls beyond computed threshold.
+		return nil
+	}
+	return ds.spanWriter.WriteSpan(span)
+}
+
+//hashBytes returns the uint64 hash value of bytes slice.
+func (ds *DownSamplingWriter) hashBytes(bytes []byte) uint64 {
+	h := ds.instance.pool.Get().(hash.Hash64)
+	defer ds.instance.pool.Put(h)
```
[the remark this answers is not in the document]
guanw: Say I put `ds.instance.pool.Put(h)` right before `return h.Sum64()`.what if h.Sum64() was fired in goroutine1 where goroutine2 just did a h.Reset(), wouldn't this cause the hash result of goroutine1 to be unreliable?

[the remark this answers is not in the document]
black-adder:
```
sum := h.Sum64()
ds.instance.pool.Put(h)
return sum
```

storage/spanstore/downsampling_writer.go
```diff
@@ -0,0 +1,128 @@
...
+	hashSaltBytes := []byte(ds.hashSalt)
+	// traceID marshal 16 bytes
+	traceIDBytes := make([]byte, 16)
+	length, err := span.TraceID.MarshalTo(traceIDBytes)
+	if err != nil || length != 16 {
+		// No Downsamping when there's error marshaling.
+		ds.metrics.MarshalingErrors.Inc(1)
+		return ds.spanWriter.WriteSpan(span)
+	}
+	hashVal := ds.hashBytes(append(hashSaltBytes, traceIDBytes...))
```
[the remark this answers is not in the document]
guanw: Will do

storage/spanstore/downsampling_writer.go
```diff
@@ -73,39 +74,33 @@ type DownSamplingOptions struct {
...
-		metrics:    downsamplingMetrics,
+		spanWriter:   spanWriter,
+		threshold:    threshold,
+		hashSalt:     hashSaltBytes,
+		poolInstance: getHashPoolInstance(len(hashSaltBytes) + traceIDByteSize),
 	}
 }
 
-// WriteSpan calls WriteSpan on wrapped span writer.
+// WriteSpan calls WriteSpan2 on wrapped span writer.
```
black-adder: typo

storage/spanstore/downsampling_writer.go
```diff
@@ -115,14 +110,12 @@ func (ds *DownSamplingWriter) WriteSpan(span *model.Span) error {
 
 //hashBytes returns the uint64 hash value of bytes slice.
 func (ds *DownSamplingWriter) hashBytes(bytes []byte) uint64 {
-	h := ds.instance.pool.Get().(hash.Hash64)
-	defer ds.instance.pool.Put(h)
+	h := ds.poolInstance.hashpool.Get().(hash.Hash64)
+	ds.poolInstance.hashpool.Put(h)
```
black-adder: unnecessary

storage/spanstore/downsampling_writer.go
```diff
@@ -115,14 +110,12 @@ func (ds *DownSamplingWriter) WriteSpan(span *model.Span) error {
...
 	h.Reset()
-	_, err := h.Write(bytes)
-	if err != nil {
-		// No downsampling when there's error hashing.
-		ds.metrics.HashingErrors.Inc(1)
-		return math.MaxUint64
-	}
-	return h.Sum64()
+	//Currently fnv.Write() doesn't throw any error so metrics is not necessary here.
```
black-adder: nit: space between `//` and `Currently`

black-adder: also, linter might complain that we are swallowing the error.

storage/spanstore/downsampling_writer_benchmark_test.go
```diff
@@ -18,6 +18,9 @@ func (n *noopWriteSpanStore) WriteSpan(span *model.Span) error {
 	return nil
 }
 
+//Benchmark result:
+//BenchmarkDownSamplingWriter_WriteSpan-12 without bytepool 50	  33352181 ns/op	18981065 B/op	  219750 allocs/op
+//BenchmarkDownSamplingWriter_WriteSpan2-12   with bytepool 100	  28070245 ns/op	16137975 B/op	  127494 allocs/op
```
black-adder: What is writespan2 here? is it with the pooling? That's still feels like a lot of allocations, we might want to pprof this. Once we generated the []bytes in the pool, we shouldn't be seeing 0 allocations.

storage/spanstore/downsampling_writer.go
black-adder: ping, this metric will be nice

storage/spanstore/downsampling_writer_benchmark_test.go
guanw: Let me try that shortly.

storage/spanstore/downsampling_writer.go
```diff
@@ -0,0 +1,121 @@
...
+	// No downSampling when threshold equals maxuint64
+	if ds.threshold == math.MaxUint64 {
+		return ds.spanWriter.WriteSpan(span)
+	}
+	byteSlice := ds.poolInstance.bytepool.Get().([]byte)
+	copy(byteSlice[:len(ds.hashSalt)], ds.hashSalt)
+	// traceID marshal to 16 bytes.
+	// Currently MarshalTo will only return err if size of traceIDBytes is smaller than 16
+	// Since we force traceIDBytes to be size of 16 metrics is not necessary here.
+	span.TraceID.MarshalTo(byteSlice[len(ds.hashSalt):])
```
vprithvi: I don't think the comment adds much value

storage/spanstore/downsampling_writer.go
```diff
@@ -0,0 +1,121 @@
...
+		threshold:    threshold,
+		hashSalt:     hashSaltBytes,
+		poolInstance: getHashPoolInstance(len(hashSaltBytes) + traceIDByteSize),
+	}
+}
+
+// WriteSpan calls WriteSpan2 on wrapped span writer.
+func (ds *DownSamplingWriter) WriteSpan(span *model.Span) error {
+	// No downSampling when threshold equals maxuint64
+	if ds.threshold == math.MaxUint64 {
```
vprithvi: Is this necessary? We can validate that the range of the threshold for downsamplingWriter are (0,1) to avoid this

storage/spanstore/downsampling_writer.go
vprithvi: +1

plugin/storage/factory_config.go
```diff
@@ -29,13 +30,24 @@ const (
 	DependencyStorageTypeEnvVar = "DEPENDENCY_STORAGE_TYPE"
 
 	spanStorageFlag = "--span-storage.type"
+
+	// DownsamplingRatio is the name of the env var that defines the ratio of spans for down sampling
+	// Users are expected to pass ratio ranging from 0 to 1.0
+	// e.g ratio = 0.3 means we are keeping 30% of spans and dropping 70% of spans.
+	DownsamplingRatio = "DOWN_SAMPLING_RATIO"
```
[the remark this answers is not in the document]
yurishkuro: +1. Let's treat "downsampling" as a single word everywhere.

plugin/storage/factory_config.go
```diff
@@ -69,11 +81,41 @@ func FactoryConfigFromEnvAndCLI(args []string, log io.Writer) FactoryConfig {
 	if depStorageType == "" {
 		depStorageType = spanWriterTypes[0]
 	}
+	ratioEnvVariable := os.Getenv(DownsamplingRatio)
```
yurishkuro: Can we read this via normal CLI options? spanStorageType is a special case because it informs which factory implementation to choose before parsing CLI flags. But downsampler params can be read in InitFromViper()

plugin/storage/factory_config.go
```diff
@@ -69,11 +81,41 @@ func FactoryConfigFromEnvAndCLI(args []string, log io.Writer) FactoryConfig {
...
+	var downSamplingRatio float64
+	if ratioEnvVariable == "" {
+		fmt.Fprintf(log,
+			"WARNING: DOWN_SAMPLING_RATIO is not specified. DownSamplingWriter will not be applied.",
```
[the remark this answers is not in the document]
yurishkuro: +1, no need to log

plugin/storage/factory.go
```diff
@@ -117,10 +118,16 @@ func (f *Factory) CreateSpanWriter() (spanstore.Writer, error) {
...
 	}
+	var spanWriter spanstore.Writer
 	if len(f.SpanWriterTypes) == 1 {
-		return writers[0], nil
+		spanWriter = writers[0]
+	} else {
+		spanWriter = spanstore.NewCompositeWriter(writers...)
 	}
-	return spanstore.NewCompositeWriter(writers...), nil
+	return spanstore.NewDownSamplingWriter(spanWriter, spanstore.DownSamplingOptions{
```
yurishkuro: why do we always apply downsampler? if f.DownSamplingRatio == 1 it only creates unnecessary overhead. Rather than checking for the same condition there _on every span_, we can check it here _once_ and skip the downsampler.

storage/spanstore/downsampling_writer_benchmark_test.go
```diff
@@ -0,0 +1,81 @@
...
+	numGoroutines := 10000
+	writeTimes := 10
+	wg.Add(numGoroutines * b.N)
+	b.ResetTimer()
+	b.ReportAllocs()
+
+	for j := 0; j < b.N; j++ {
+		for n := 0; n < numGoroutines; n++ {
+
+			go func() {
```
yurishkuro: this probably allocates memory every time. Try moving outside of the loop and assigning to a variable, e.g. `f := func()...`, then inside the loop only call `f()`

yurishkuro: actually, you should avoid creating go-routing inside the loop after you called `b.ResetTimer()`. You can start then all before `b.ResetTimer()`, and have them all block on reading from a channel. Then close the channel to unblock them all.

storage/spanstore/downsampling_writer.go
```diff
@@ -0,0 +1,121 @@
...
+		// Drops spans when hashVal falls beyond computed threshold.
+		return nil
+	}
+	return ds.spanWriter.WriteSpan(span)
+}
+
+//hashBytes returns the uint64 hash value of bytes slice.
+func (ds *DownSamplingWriter) hashBytes(bytes []byte) uint64 {
+	h := ds.poolInstance.hashpool.Get().(hash.Hash64)
+	ds.poolInstance.hashpool.Put(h)
```
yurishkuro: why do you return it ti the pool immediately?

storage/spanstore/downsampling_writer.go
```diff
@@ -0,0 +1,121 @@
...
+	hashpool *sync.Pool
+	bytepool *sync.Pool
+}
+
+var instance *poolSingleton
+var once sync.Once
+
+// Initialize the singleton if it doesn't exist already.
+func getHashPoolInstance(length int) *poolSingleton {
+	once.Do(func() {
```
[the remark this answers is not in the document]
guanw: this singleton is just here to make sure we are factory_test.go will pass. If not using it, *sync.Pool will be recreated with different value and result in some tests failure.

storage/spanstore/downsampling_writer_benchmark_test.go
guanw: I've removed the goroutines entirely from benchmark test for now.

plugin/storage/factory.go
guanw: i thought we discussed about applying downsampling dynamically in the future, so we should always apply it?

storage/spanstore/downsampling_writer.go
guanw: i'm thinking without it the code will still spend extra time updating byte array & computing hash even when ratio = 1 which is unnecessary?

plugin/storage/factory.go
yurishkuro: Maybe we can add it in the future then. Right now we don't have any dynamic control plane to benefit from such setup.

storage/spanstore/downsampling_writer.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+// Wrap the sync.Pool object within a singleton struct.
+type poolSingleton struct {
+	hashpool *sync.Pool
+	bytepool *sync.Pool
+}
+
+var instance *poolSingleton
+var once sync.Once
```
yurishkuro: Pools should be fields of the factory, not globals. Pointers are not needed.

storage/spanstore/downsampling_writer.go
```diff
@@ -0,0 +1,132 @@
...
+	spanWriter   Writer
+	threshold    uint64
+	hashSalt     []byte
+	poolInstance *poolSingleton
+	metrics      *downSamplingWriterMetrics
+}
+
+// downSamplingWriterMetrics is metrics that keeping track of total number of dropped spans
+type downSamplingWriterMetrics struct {
+	SpansDropped metrics.Counter `metric:"spans_dropped"`
```
yurishkuro: It is better to have parity in metrics, if you have spans_dropped, then also add spans_accepted

storage/spanstore/downsampling_writer.go
```diff
@@ -0,0 +1,132 @@
...
+type poolSingleton struct {
+	hashpool *sync.Pool
+	bytepool *sync.Pool
+}
+
+var instance *poolSingleton
+var once sync.Once
+
+// Initialize the singleton if it doesn't exist already.
+func getHashPoolInstance(length int) *poolSingleton {
```
yurishkuro: It is customary to have constructor follow the declaration. Helper functions go later.

storage/spanstore/downsampling_writer.go
```diff
@@ -0,0 +1,132 @@
...
+		spanWriter:   spanWriter,
+		threshold:    threshold,
+		hashSalt:     hashSaltBytes,
+		poolInstance: getHashPoolInstance(len(hashSaltBytes) + traceIDByteSize),
+		metrics:      writeMetrics,
+	}
+}
+
+// WriteSpan calls WriteSpan on wrapped span writer.
+func (ds *DownSamplingWriter) WriteSpan(span *model.Span) error {
```
yurishkuro: This function could use refactoring. It would be more readable as a high level logic
```
If shouldDownsample(span.TraceID) {
    metrics.dropped.inc()
    return
}
metrics.allowed.inc()
return writer.Write()
```

storage/spanstore/downsampling_writer_benchmark_test.go
```diff
@@ -0,0 +1,82 @@
...
+	b.ReportAllocs()
+
+	for it := 0; it < b.N; it++ {
+		for i := 0; i < 10000; i++ {
+			c.hashBytes(ba)
+		}
+	}
+}
+
+func TestDownSamplingWriter_RandomHash(t *testing.T) {
```
yurishkuro: This is a test, not benchmark

guanw: @black-adder mentioned this test as flaky so i put it here.

plugin/storage/factory.go
guanw: Sounds good.

storage/spanstore/downsampling_writer.go
guanw: Not sure I got the first part. Here instance is lowercase so it won't be exposed to other files. In factory.go the default is to disable DownsamplingWriter entirely so doesn't make sense to create pool there if we are not using downsamplingWriter at all?

I will get rid of the pointer.

storage/spanstore/downsampling_writer.go
guanw: sg.

storage/spanstore/downsampling_writer.go
guanw: sg.

storage/spanstore/downsampling_writer_benchmark_test.go
```diff
@@ -0,0 +1,82 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/uber/jaeger-lib/metrics"
+
+	"github.com/jaegertracing/jaeger/model"
+)
+
+// Benchmark result:
+// BenchmarkDownSamplingWriter_WriteSpan-12    	    2000	    783766 ns/op	       1 B/op	       0 allocs/op
```
black-adder: sweet

storage/spanstore/downsampling_writer_benchmark_test.go
```diff
@@ -0,0 +1,82 @@
...
+			c.WriteSpan(span)
+		}
+
+	}
+}
+
+// Benchmark result:
+// BenchmarkDownSamplingWriter_HashBytes-12    	    5000	    381558 ns/op	       0 B/op	       0 allocs/op
+func BenchmarkDownSamplingWriter_HashBytes(b *testing.B) {
+
```
black-adder: rm newline

storage/spanstore/downsampling_writer_benchmark_test.go
black-adder: I suggested converting the test to a benchmark, not just moving the test over to a benchmark file.

storage/spanstore/downsampling_writer.go
```diff
@@ -0,0 +1,132 @@
...
+	// No downSampling when threshold equals maxuint64
+	if ds.threshold == math.MaxUint64 {
+		return ds.spanWriter.WriteSpan(span)
+	}
+
+	byteSlice := ds.poolInstance.bytepool.Get().(*[]byte)
+	copy((*byteSlice)[:len(ds.hashSalt)], ds.hashSalt)
```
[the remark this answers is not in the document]
guanw: sg

storage/spanstore/downsampling_writer_benchmark_test.go
guanw: my bad. Will refactor shortly

guanw:
@yurishkuro @black-adder @vprithvi Any more comments on this one?

plugin/storage/factory_test.go
```diff
@@ -19,6 +19,8 @@ import (
 	"flag"
 	"testing"
 
+	"github.com/jaegertracing/jaeger/pkg/config"
```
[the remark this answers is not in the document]
black-adder: there must've been a miscommunication here, this is doing the exact opposite of what I asked.
```
golang

third party

github.com/jaegertracing/jaeger
```
should be the order of the imports

plugin/storage/factory.go
```diff
@@ -37,15 +37,24 @@ const (
 	elasticsearchStorageType = "elasticsearch"
 	memoryStorageType        = "memory"
 	kafkaStorageType         = "kafka"
+
+	// DownsamplingRatio defines command line param for DownsamplingRatio
+	DownsamplingRatio = "downsampling.ratio"
```
black-adder: these consts don't need to be public, they're only used in this file.

plugin/storage/factory.go
```diff
@@ -117,10 +127,21 @@ func (f *Factory) CreateSpanWriter() (spanstore.Writer, error) {
...
 	}
+	var spanWriter spanstore.Writer
 	if len(f.SpanWriterTypes) == 1 {
-		return writers[0], nil
+		spanWriter = writers[0]
+	} else {
+		spanWriter = spanstore.NewCompositeWriter(writers...)
 	}
-	return spanstore.NewCompositeWriter(writers...), nil
+	// Turn off DownsamplingWriter entirely if ratio == 1.0
```
black-adder: nit: end sentence with period

storage/spanstore/downsampling_writer.go
```diff
@@ -0,0 +1,133 @@
...
+
+package spanstore
+
+import (
+	"hash"
+	"hash/fnv"
+	"math"
+	"sync"
+
+	"github.com/jaegertracing/jaeger/model"
```
black-adder: import ordering is incorrect

plugin/storage/factory_config.go
```diff
@@ -69,6 +71,7 @@ func FactoryConfigFromEnvAndCLI(args []string, log io.Writer) FactoryConfig {
 	if depStorageType == "" {
 		depStorageType = spanWriterTypes[0]
 	}
+
```
black-adder: unecessary

storage/spanstore/downsampling_writer.go
```diff
@@ -0,0 +1,133 @@
...
+)
+
+var (
+	instance poolSingleton
+	once     sync.Once
+)
+
+// Wrap the sync.Pool object within a singleton struct.
+type poolSingleton struct {
+	hashpool *sync.Pool
```
black-adder: hashPool and byteBool

storage/spanstore/downsampling_writer.go
```diff
@@ -0,0 +1,133 @@
...
+	bytepool *sync.Pool
+}
+
+// downSamplingWriterMetrics is metrics that keeping track of total number of dropped spans
```
black-adder: downSamplingWriterMetrics keeps track of total number of dropped spans.

storage/spanstore/downsampling_writer.go
```diff
@@ -0,0 +1,133 @@
...
+// NewDownSamplingWriter creates a DownSamplingWriter.
+func NewDownSamplingWriter(spanWriter Writer, downSamplingOptions DownSamplingOptions) *DownSamplingWriter {
+	threshold := uint64(downSamplingOptions.Ratio * float64(math.MaxUint64))
+	writeMetrics := &downSamplingWriterMetrics{}
+	metrics.Init(writeMetrics, downSamplingOptions.MetricsFactory, nil)
+	hashSaltBytes := []byte(downSamplingOptions.HashSalt)
+	poolInstance := getHashPoolInstance(len(hashSaltBytes) + traceIDByteSize)
+	byteSlice := poolInstance.bytepool.Get().(*[]byte)
+	copy((*byteSlice)[:len(hashSaltBytes)], hashSaltBytes)
+	poolInstance.bytepool.Put(byteSlice)
```
black-adder: this will only do it for one entry in the pool, you need to do it for every single entry on initialization. Pass in `hashSaltBytes` into `getHashPoolInstance` and update to:
```
				New: func() interface{} {
					byteArray := make([]byte, length)
                                          // TODO write hashSaltBtyes here 
					return &byteArray
				},
```

storage/spanstore/downsampling_writer_test.go
```diff
@@ -0,0 +1,76 @@
...
+	assert.NoError(t, c.WriteSpan(span))
+
+	downSamplingOptions.Ratio = 0
+	c = NewDownSamplingWriter(&noopWriteSpanStore{}, downSamplingOptions)
+	assert.NoError(t, c.WriteSpan(span))
+
+	downSamplingOptions.Ratio = 0.8
+	c = NewDownSamplingWriter(&noopWriteSpanStore{}, downSamplingOptions)
+	assert.NoError(t, c.WriteSpan(span))
+
```
black-adder: rm newline

storage/spanstore/downsampling_writer_test.go
```diff
@@ -0,0 +1,76 @@
...
+		High: uint64(1),
+	}
+	span := &model.Span{
+		TraceID: trace,
+	}
+	traceIDBytes := make([]byte, 16)
+	span.TraceID.MarshalTo(traceIDBytes)
+	// Same traceID should always be hashed to same uint64 in DownSamplingWriter
+	assert.Equal(t, c.hashBytes(traceIDBytes), c.hashBytes(traceIDBytes))
+
```
black-adder: rm newline

storage/spanstore/downsampling_writer_benchmark_test.go
```diff
@@ -0,0 +1,85 @@
...
+				},
+			}
+			hash := c.hashBytes([]byte(span.TraceID.String()))
+			if hash < ratioThreshold {
+				countSmallerThanRatio++
+			}
+		}
+		fmt.Printf("Random hash ratio %f should be close to 0.5, inspect the implementation of hashBytes if not\n", math.Abs(float64(countSmallerThanRatio)/1000000))
+	}
+
```
black-adder: rm newline

storage/spanstore/downsampling_writer_benchmark_test.go
```diff
@@ -0,0 +1,85 @@
...
+		HashSalt: "jaeger-test",
+	})
+
+	b.ResetTimer()
+	b.ReportAllocs()
+	for it := 0; it < b.N; it++ {
+		for writes := 0; writes < 10000; writes++ {
+			c.WriteSpan(span)
+		}
+
```
black-adder: rm newline

storage/spanstore/downsampling_writer_benchmark_test.go
```diff
@@ -0,0 +1,85 @@
...
+			c.hashBytes(ba)
+		}
+	}
+}
+
+func BenchmarkDownSamplingWriter_RandomHash(b *testing.B) {
+	ratioThreshold := uint64(math.MaxUint64 / 2)
+	countSmallerThanRatio := 0
+	downSamplingOptions := DownSamplingOptions{
+		Ratio:          1,
```
black-adder: shouldn't this be 0.5?

storage/spanstore/downsampling_writer_benchmark_test.go
```diff
@@ -0,0 +1,85 @@
...
+		HashSalt:       "jaeger-test",
+		MetricsFactory: metrics.NullFactory,
+	}
+	c := NewDownSamplingWriter(&noopWriteSpanStore{}, downSamplingOptions)
+	for it := 0; it < b.N; it++ {
+		countSmallerThanRatio = 0
+		for i := 0; i < 1000000; i++ {
```
black-adder: 1000000 is used here and on L82, define it at the top of this function

black-adder:
[travis](https://travis-ci.org/jaegertracing/jaeger/jobs/498783229) is failing with: `Go fmt, license check, or import ordering failures, run 'make fmt'`

storage/spanstore/downsampling_writer_benchmark_test.go
guanw: doesn't matter, this is just checking if the random hash falls over/below the ratioThreshold which is half the max uint64. The percentage should be 0.5.

guanw:
@yurishkuro any more comments on this?

plugin/storage/factory.go
```diff
@@ -139,6 +160,21 @@ func (f *Factory) AddFlags(flagSet *flag.FlagSet) {
...
+func addFlags(flagSet *flag.FlagSet) {
+	flagSet.Float64(
+		downsamplingRatio,
+		defaultDownsamplingRatio,
+		"Downsampling ratio defines the ratio of spans for downsampling. Number between 0 ~ 1.0. Default is 1.0 with no downsampling. Values not in the range of 0 ~ 1.0 will be set to default. e.g ratio = 0.3 means we are keeping 30% of spans and dropping 70% of spans.",
+	)
+	flagSet.String(
+		downsamplingHashSalt,
+		defaultDownsamplingHashSalt,
+		"downsamplingHashSalt defines the hash salt for downsampling. Default is empty string.",
```
yurishkuro: simplify comment (viper shows default automatically, no need to explain):

Salt used when hashing trace id for downsampling.

plugin/storage/factory.go
```diff
@@ -37,15 +37,24 @@ const (
 	elasticsearchStorageType = "elasticsearch"
 	memoryStorageType        = "memory"
 	kafkaStorageType         = "kafka"
+
+	// downsamplingRatio defines command line param for DownsamplingRatio.
+	downsamplingRatio = "downsampling.ratio"
+	// downsamplingHashSalt defines command line param for DownsamplingHashSalt.
+	downsamplingHashSalt = "downsampling.hashsalt"
+	// defaultDownsamplingRatio is the default downsampling ratio.
```
yurishkuro: comment does not explain anything, can remove it

plugin/storage/factory.go
```diff
@@ -37,15 +37,24 @@ const (
...
+	defaultDownsamplingRatio = 1.0
+	// defaultDownsamplingHashSalt is the default downsampling hashsalt.
```
yurishkuro: comment does not explain anything, can remove it

storage/spanstore/downsampling_writer.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+	ds.metrics.SpansAccepted.Inc(1)
+	return ds.spanWriter.WriteSpan(span)
+}
+
+func (ds *DownSamplingWriter) shouldDownsample(span *model.Span) bool {
+	byteSlice := ds.poolInstance.bytePool.Get().(*[]byte)
+	// Currently MarshalTo will only return err if size of traceIDBytes is smaller than 16
+	// Since we force traceIDBytes to be size of 16 metrics is not necessary here.
+	span.TraceID.MarshalTo((*byteSlice)[ds.lengthOfSalt:])
```
yurishkuro: `_ = ...` to illustrate that error is ignored

storage/spanstore/downsampling_writer.go
```diff
@@ -0,0 +1,135 @@
...
+	SpansAccepted metrics.Counter `metric:"spans_accepted"`
+}
+
+// DownSamplingWriter is a span Writer that drops spans with a predefined downSamplingRatio.
+type DownSamplingWriter struct {
+	spanWriter   Writer
+	threshold    uint64
+	lengthOfSalt int
+	poolInstance poolSingleton
+	metrics      *downSamplingWriterMetrics
```
yurishkuro: don't need pointer here, assign by value

storage/spanstore/downsampling_writer.go
```diff
@@ -0,0 +1,135 @@
...
+)
+
+const (
+	traceIDByteSize = 16
+)
+
+var (
+	instance poolSingleton
+	once     sync.Once
+)
```
yurishkuro: these are not needed

storage/spanstore/downsampling_writer.go
```diff
@@ -0,0 +1,135 @@
...
+
+// Wrap the sync.Pool object within a singleton struct.
+type poolSingleton struct {
```
yurishkuro: I would rename this to `type hasher struct` and keep only one pool of these structs, since you **always** need both hash and byte buffer. A single pool will reduce synchronization

storage/spanstore/downsampling_writer.go
```diff
@@ -0,0 +1,135 @@
...
+// DownSamplingOptions contains the options for constructing a DownSamplingWriter.
+type DownSamplingOptions struct {
+	Ratio          float64
+	HashSalt       string
+	MetricsFactory metrics.Factory
+}
+
+// Initialize the singleton if it doesn't exist already.
+func getHashPoolInstance(length int, hashSaltBytes []byte) poolSingleton {
+	once.Do(func() {
```
yurishkuro: there's no need to use `once`. Simply instantiate the pool in NewDownSamplingWriter

storage/spanstore/downsampling_writer.go
```diff
@@ -0,0 +1,135 @@
...
+)
+
+// Wrap the sync.Pool object within a singleton struct.
+type poolSingleton struct {
+	hashPool *sync.Pool
+	bytePool *sync.Pool
+}
+
+// downSamplingWriterMetrics keeping track of total number of dropped spans and accepted spans
+type downSamplingWriterMetrics struct {
```
yurishkuro: downSampling -> downsampling (one word) throughout this file)

storage/spanstore/downsampling_writer.go
```diff
@@ -0,0 +1,135 @@
...
+	// Currently MarshalTo will only return err if size of traceIDBytes is smaller than 16
+	// Since we force traceIDBytes to be size of 16 metrics is not necessary here.
+	span.TraceID.MarshalTo((*byteSlice)[ds.lengthOfSalt:])
+	hashVal := ds.hashBytes(*byteSlice)
+	ds.poolInstance.bytePool.Put(byteSlice)
+	return hashVal >= ds.threshold
+}
+
+// hashBytes returns the uint64 hash value of byte slice.
+func (ds *DownSamplingWriter) hashBytes(bytes []byte) uint64 {
```
yurishkuro: once you combine hash and buffer into `hasher` struct, you can define this method on that struct instead of on the Writer

storage/spanstore/downsampling_writer.go
```diff
@@ -0,0 +1,135 @@
...
+	h := ds.poolInstance.hashPool.Get().(hash.Hash64)
+	h.Reset()
+	// Currently fnv.Write() doesn't throw any error so metrics is not necessary here.
```
yurishkuro: is this about metrics or error handling?  When ignoring errors, indicate it explicitly `_ = func()`

guanw: The issue was that if we do a 
```
if err != nil {}
```
the code inside this if block will never be executed thus code coverage won't be 100%. I will add the placeholder then.

guanw:
@yurishkuro updated.

storage/spanstore/downsampling_writer.go
yurishkuro: That's why you use anonymous variable `_`, you assign to it but don't check it afterwards. This way it's obvious to someone reading the code that this happens, but if you simply ignore the function result the reader has no idea that there _was_ a result.

guanw:
@yurishkuro Please let me know if you have any more comments.

storage/spanstore/downsampling_writer.go
```diff
@@ -0,0 +1,116 @@
...
+	SpansDropped  metrics.Counter `metric:"spans_dropped"`
+	SpansAccepted metrics.Counter `metric:"spans_accepted"`
+}
+
+// DownSamplingWriter is a span Writer that drops spans with a predefined downSamplingRatio.
+type DownSamplingWriter struct {
+	spanWriter   Writer
+	threshold    uint64
+	lengthOfSalt int
+	pool         *sync.Pool
```
yurishkuro: hasherPool

storage/spanstore/downsampling_writer.go
```diff
@@ -0,0 +1,116 @@
...
+	metrics      *downsamplingWriterMetrics
```
yurishkuro: don't need pointer

storage/spanstore/downsampling_writer.go
```diff
@@ -0,0 +1,116 @@
...
+			byteArray := make([]byte, len(hashSaltBytes)+traceIDByteSize)
+			copy(byteArray[:len(hashSaltBytes)], hashSaltBytes)
+			return &hasher{
+				fnv:       fnv.New64a(),
+				byteArray: byteArray,
+			}
+		},
+	}
+
+	return &DownSamplingWriter{
```
yurishkuro: DownSamplingWriter -> DownsamplingWriter

storage/spanstore/downsampling_writer.go
```diff
@@ -0,0 +1,116 @@
...
+)
+
+const (
+	traceIDByteSize = 16
+)
+
+// hasher includes data we want to put in sync.Pool
+type hasher struct {
+	fnv       hash.Hash64
+	byteArray []byte
```
yurishkuro: s/byteArray/buffer/  - don't name vars after their type

storage/spanstore/downsampling_writer.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/jaegertracing/jaeger/model"
+)
+
+const (
+	traceIDByteSize = 16
+)
+
+// hasher includes data we want to put in sync.Pool
+type hasher struct {
+	fnv       hash.Hash64
```
yurishkuro: fvn is implementation detail. logical function of the variable is "hash"

storage/spanstore/downsampling_writer.go
```diff
@@ -0,0 +1,116 @@
...
+// NewDownSamplingWriter creates a DownSamplingWriter.
+func NewDownSamplingWriter(spanWriter Writer, downSamplingOptions DownSamplingOptions) *DownSamplingWriter {
+	threshold := uint64(downSamplingOptions.Ratio * float64(math.MaxUint64))
+	writeMetrics := &downsamplingWriterMetrics{}
+	metrics.Init(writeMetrics, downSamplingOptions.MetricsFactory, nil)
+	hashSaltBytes := []byte(downSamplingOptions.HashSalt)
+	pool := &sync.Pool{
+		New: func() interface{} {
+			byteArray := make([]byte, len(hashSaltBytes)+traceIDByteSize)
+			copy(byteArray[:len(hashSaltBytes)], hashSaltBytes)
```
yurishkuro: I think the following will work just as well:
```
copy(byteArray, downSamplingOptions.HashSalt)
```

storage/spanstore/downsampling_writer.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+	ds.metrics.SpansAccepted.Inc(1)
+	return ds.spanWriter.WriteSpan(span)
+}
+
+func (ds *DownSamplingWriter) shouldDownsample(span *model.Span) bool {
+	hasherInstance := ds.pool.Get().(*hasher)
+	// Currently MarshalTo will only return err if size of traceIDBytes is smaller than 16
+	// Since we force traceIDBytes to be size of 16 metrics is not necessary here.
+	byteSlice := hasherInstance.byteArray
```
yurishkuro: hasherInstance already has access to the buffer internally, no need to pass it in

storage/spanstore/downsampling_writer.go
```diff
@@ -0,0 +1,116 @@
...
+		threshold:    threshold,
+		pool:         pool,
+		metrics:      writeMetrics,
+		lengthOfSalt: len(hashSaltBytes),
+	}
+}
+
+// WriteSpan calls WriteSpan on wrapped span writer.
+func (ds *DownSamplingWriter) WriteSpan(span *model.Span) error {
+	if ds.shouldDownsample(span) {
```
yurishkuro: I think this is opposite of what "downsampling" means. Dictionary: "Downsampling is the process of reducing the sampling rate of a signal". So if we have a digital signal with 100 samples / sec, the downsampled signal may have 10 samples / sec. So "downsampling" should be treated the same as "sampling": what we _keep_, not what we _discard_. This is how we defined the ratio as well, 0.3 means we keep 30%.

Therefore, this statement should be `if !ds.shouldDownsample(span) {`

storage/spanstore/downsampling_writer.go
```diff
@@ -0,0 +1,116 @@
...
+func (ds *DownSamplingWriter) shouldDownsample(span *model.Span) bool {
+	hasherInstance := ds.pool.Get().(*hasher)
+	// Currently MarshalTo will only return err if size of traceIDBytes is smaller than 16
+	// Since we force traceIDBytes to be size of 16 metrics is not necessary here.
+	byteSlice := hasherInstance.byteArray
+	// Currently we are always passing byte slice with size of 16 which won't trigger the underlining error
+	_, _ = span.TraceID.MarshalTo((byteSlice)[ds.lengthOfSalt:])
+	hashVal := hasherInstance.hashBytes(byteSlice)
+	ds.pool.Put(hasherInstance)
+	return hashVal >= ds.threshold
```
yurishkuro: `hashVal <= ds.threshold`

storage/spanstore/downsampling_writer.go
guanw: Just curious, how do we decide if metrics or more general struct should have a pointer or not? I saw a lot of examples in internal version with pointer.

guanw: @yurishkuro comment on this?

yurishkuro: There's no hard rule, but from efficiency pov, when you don't need a pointer, don't create one.

plugin/storage/factory.go
```diff
@@ -148,6 +182,16 @@ func (f *Factory) InitFromViper(v *viper.Viper) {
...
 		}
 	}
+	f.initDownsamplingFromViper(v)
+}
+
+func (f *Factory) initDownsamplingFromViper(v *viper.Viper) {
+	f.FactoryConfig.DownsamplingRatio = v.GetFloat64(downsamplingRatio)
+	if f.FactoryConfig.DownsamplingRatio < 0 || f.FactoryConfig.DownsamplingRatio > 1 {
+		// Values not in the range of 0 ~ 1.0 will be set to default.
+		f.FactoryConfig.DownsamplingRatio = 1.0
```
yurishkuro: nit: ` = defaultDownsamplingRatio` ?

plugin/storage/factory_test.go
```diff
@@ -129,11 +132,34 @@ func TestCreate(t *testing.T) {
...
+	spanWriter := new(spanStoreMocks.Writer)
+	mock.On("CreateSpanWriter").Return(spanWriter, nil)
+
+	f.DownsamplingRatio = 0.5
+	m := metrics.NullFactory
+	l := zap.NewNop()
+	mock.On("Initialize", m, l).Return(nil)
+
+	f.Initialize(m, l)
+	_, err = f.CreateSpanWriter()
```
yurishkuro: Don't you want to check that the returned writer is the DownsamplingWriter?

yurishkuro: Also, it would make sense to test that when DownsamplingRatio == 1 then the mock `spanWriter` is returned (i.e. data driven test with 2 param values).

storage/spanstore/downsampling_writer.go
```diff
@@ -0,0 +1,114 @@
...
+	"math"
+	"sync"
+
+	"github.com/uber/jaeger-lib/metrics"
+
+	"github.com/jaegertracing/jaeger/model"
+)
+
+const (
+	traceIDByteSize = 16
```
yurishkuro: shouldn't hardcode this here. The `model/ids.go` file has this:
```
// Size returns the size of this datum in protobuf. It is always 16 bytes.
func (t *TraceID) Size() int {
        return 16
}
```

You can create a fake trace id and ask it for Size(): 
```
var (
    traceIDByteSize := model.NewTraceID(0, 0).Size()`
)
```

storage/spanstore/downsampling_writer.go
```diff
@@ -0,0 +1,114 @@
...
+	HashSalt       string
+	MetricsFactory metrics.Factory
+}
+
+// NewDownsamplingWriter creates a DownsamplingWriter.
+func NewDownsamplingWriter(spanWriter Writer, downsamplingOptions DownsamplingOptions) *DownsamplingWriter {
+	threshold := uint64(downsamplingOptions.Ratio * float64(math.MaxUint64))
+	writeMetrics := &downsamplingWriterMetrics{}
+	metrics.Init(writeMetrics, downsamplingOptions.MetricsFactory, nil)
+	hashSaltBytes := []byte(downsamplingOptions.HashSalt)
```
yurishkuro: I would prefer to define a private constant for defaultSalt & initialize it to some random string, and use that when `downsamplingOptions.HashSalt  == ""`. This would avoid the possibility that some other trace id hashing solution elsewhere would also run with empty string as a salt.

storage/spanstore/downsampling_writer.go
```diff
@@ -0,0 +1,114 @@
...
+	pool := &sync.Pool{
+		New: func() interface{} {
+			byteArray := make([]byte, len(hashSaltBytes)+traceIDByteSize)
```
yurishkuro: byteArray -> buffer

storage/spanstore/downsampling_writer.go
```diff
@@ -0,0 +1,114 @@
...
+	return hashVal <= ds.threshold
+}
+
+// hashBytes returns the uint64 hash value of byte slice.
+func (h *hasher) hashBytes() uint64 {
+	h.hash.Reset()
+	// Currently fnv.Write() implementation doesn't throw any error so metric is not necessary here.
+	_, _ = h.hash.Write(h.buffer)
+	sum := h.hash.Sum64()
+	return sum
```
yurishkuro: inline: `return h.hash.Sum64()`

storage/spanstore/downsampling_writer_test.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/jaegertracing/jaeger/model"
+)
+
+type noopWriteSpanStore struct{}
+
+func (n *noopWriteSpanStore) WriteSpan(span *model.Span) error {
+	return nil
+}
+func TestDownSamplingWriter_WriteSpan(t *testing.T) {
```
yurishkuro: this test is not testing anything, it only executes the writer. It should be checking that (a) span was or was not passed to the underlying writer, and (b) that the right metric was incremented.

storage/spanstore/downsampling_writer_test.go
```diff
@@ -0,0 +1,74 @@
...
+	downsamplingOptions.Ratio = 0
+	c = NewDownsamplingWriter(&noopWriteSpanStore{}, downsamplingOptions)
+	assert.NoError(t, c.WriteSpan(span))
+
+	downsamplingOptions.Ratio = 0.8
+	c = NewDownsamplingWriter(&noopWriteSpanStore{}, downsamplingOptions)
+	assert.NoError(t, c.WriteSpan(span))
+}
+
+func TestDownSamplingWriter_hashBytes(t *testing.T) {
```
yurishkuro: not sure why we need this test, the hasher would be exercised by the above test.

guanw: this is to make sure `h.hash.Reset()` works and same traceID will always hash to the same value.

yurishkuro: +1, maybe add a comment

storage/spanstore/downsampling_writer.go
guanw: If you mean same non-empty salt passed from CreateSpanWriter to DownsamplingWriter, I would pass it in `factory.go`. 

Else we would end up with different salt for different ingesters which won't guarantee that we always downsampling/ not downsampling spans from same trace (if they land on different ingesters)?

yurishkuro: why would it be different if it's a hardcoded const value, only used when no salt is provided externally?

guanw: i think i misunderstood your meaning. Updated the code and passed a hardcoded const value from `factory.go` for `downsamplingWriter`.

guanw: why not just pass the const from factory.go? This way if downsamplingWriter is used elsewhere we could just pass different const string.

plugin/storage/factory.go
yurishkuro: it's still assigning 1.0 instead of default

storage/spanstore/downsampling_writer.go
yurishkuro: because right now your default in the factory is also `""`. What I am suggesting does not prevent using the writer with a different salt:
```
const defaultHashSalt  = "Bender is great!"

func NewDownsamplingWriter(spanWriter Writer, downsamplingOptions DownsamplingOptions) *DownsamplingWriter {
    salt := downsamplingOptions. HashSalt
    if salt == "" {
        salt = defaultHashSalt
    }
    // the rest of constructor
}
```

PR #1364, opened 2019-02-20
Discover dependencies table version automatically

### LINKED ISSUE #1363: Dependencies V2 not friendly for new users

Make dependencies v2 more new user friendly.

#1328 added a new schema mapping for dependencies_v2 with is by default used by cassandra-create-schema scripts, however the v2 table is not by default used by jaeger (you have to pass in a flag). This means that new users will not be able to query dependencies by default.

Do a DESC query on the cassandra jaeger keyspace on startup and see if the dependencies_v2 table is available. If so, use v2, else use v1. 

How will this work with the flag? Should we just go ahead and deprecate it?

black-adder:

Resolves #1363

- [ ] Deprecate `EnableDependenciesV2` flag

### DIFF
--- a/plugin/storage/cassandra/dependencystore/bootstrap.go
+++ b/plugin/storage/cassandra/dependencystore/bootstrap.go
@@ -1,0 +1,28 @@
+// Copyright (c) 2019 Uber Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package dependencystore
+
+import (
+	"github.com/jaegertracing/jaeger/pkg/cassandra"
+)
+
+// GetDependencyVersion attempts to determine the version of the dependencies table.
+// TODO: Remove this once we've migrated to V2 permanently.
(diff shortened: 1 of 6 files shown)


plugin/storage/cassandra/dependencystore/bootstrap.go
```diff
@@ -0,0 +1,28 @@
...
+package dependencystore
+
+import (
+	"github.com/jaegertracing/jaeger/pkg/cassandra"
+)
+
+// GetDependencyVersion attempts to determine the version of the dependencies table.
+// TODO: Remove this once we've migrated to V2 permanently.
+func GetDependencyVersion(s cassandra.Session) Version {
+	if err := s.Query("SELECT ts from dependencies_v2 limit 1;").Exec(); err != nil {
```
yurishkuro: what kind of error is returned? just curious

black-adder: `InvalidRequest: Error from server: code=2200 [Invalid query] message="unconfigured table dependencies"`

pavolloffay:
@black-adder I have raised this also because of https://github.com/jaegertracing/spark-dependencies/issues/58. There can be other components which expect schema for v1 dependencies. We have failing CI jobs now in opearator bc dep. job tries to write to old dependnecies table.

in spark job we should do the same - automatically detect the table

plugin/storage/cassandra/dependencystore/bootstrap.go
```diff
@@ -0,0 +1,28 @@
...
+// limitations under the License.
+
+package dependencystore
+
+import (
+	"github.com/jaegertracing/jaeger/pkg/cassandra"
+)
+
+// GetDependencyVersion attempts to determine the version of the dependencies table.
+// TODO: Remove this once we've migrated to V2 permanently.
```
pavolloffay: Add a pointer to issue

pavolloffay:
@black-adder can this be merged?

PR #1372, opened 2019-02-22
Use CQLSH_HOST in final call to cqlsh as well

funny-falcon:
- docker script doesn't use CQLSH_HOST variable for final call to cqlsh
I suppose, it makes impossible to use docker image on other host that is not part of cassandra cluster.

- add CQLSH_HOST variable in place of host argument

### DIFF
--- a/plugin/storage/cassandra/schema/docker.sh
+++ b/plugin/storage/cassandra/schema/docker.sh
@@ -31,4 +31,4 @@ done
 
 echo "Generating the schema for the keyspace ${KEYSPACE} and datacenter ${DATACENTER}"
 
-MODE="${MODE}" DATACENTER="${DATACENTER}" KEYSPACE="${KEYSPACE}" /cassandra-schema/create.sh "${TEMPLATE}" | ${CQLSH} ${CQLSH_SSL}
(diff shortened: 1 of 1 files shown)


funny-falcon:
Oh, I was just writting that it was wrong change :-(
I've missed that cqlsh accepts environment variable CQLSH_HOST.

My thoughts:
- default CQLSH_HOST is "cassandra"
- default CQLSH_HOST were not exported
- previously default CQLSH_HOST were not passed to final command, but it were passed to check command.

Therefore:
- if CQLSH_HOST were defined from outside (as a docker `-e CQLSH_HOST` argument), it worked correctly in both places (because it were passed as an argument to check command, and passed as a environment variable to final command)
- if CQLSH_HOST were not defined from outside, check command accepted default value "cassandra" as a host argument, and, probably, it failed, making whole script fail.

Therefore, correct change should be removing both usage of CQLSH_HOST as arguments.

I sorry that I sent this pull request without enough investigation.

jpkrohling:
I kinda wondered whether `cqlsh` shouldn't read `CQLSH_HOST`, but given that we have a similar usage a few lines above (line 18), I think this is appropriate.

The default value `cassandra` is consistent with the docs we have, but it's really intended to be set by users, as there's no right default value we could use.

PR #1384, opened 2019-02-26
Make grpc reporter default and add retry

### LINKED ISSUE #1218: Use gRPC between collector and agent as default
We should decide when to use gRPC as default between agent and collector.

The first release with gRPC will be 1.9.0. If all goes well could we switch to the default in 1.10?

pavolloffay:
Resolves #1218 #1229 

### DIFF
--- a/cmd/agent/app/reporter/grpc/collector_proxy.go
+++ b/cmd/agent/app/reporter/grpc/collector_proxy.go
@@ -42,7 +42,7 @@ func NewCollectorProxy(o *Options, mFactory metrics.Factory, logger *zap.Logger)
 	if len(o.CollectorHostPort) == 0 {
 		return nil, errors.New("could not create collector proxy, address is missing")
 	}
-	var conn *grpc.ClientConn
+	var target string
 	if len(o.CollectorHostPort) > 1 {
 		r, _ := manual.GenerateAndRegisterManualResolver()
 		var resolvedAddrs []resolver.Address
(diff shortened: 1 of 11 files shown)


cmd/agent/app/reporter/grpc/collector_proxy.go
```diff
@@ -49,10 +50,12 @@ func NewCollectorProxy(o *Options, mFactory metrics.Factory, logger *zap.Logger)
...
 		}
 		r.InitialAddrs(resolvedAddrs)
-		conn, _ = grpc.Dial(r.Scheme()+":///round_robin", grpc.WithInsecure(), grpc.WithBalancerName(roundrobin.Name))
+		conn, _ = grpc.Dial(r.Scheme()+":///round_robin", grpc.WithInsecure(), grpc.WithBalancerName(roundrobin.Name),
+			grpc.WithUnaryInterceptor(grpc_retry.UnaryClientInterceptor(grpc_retry.WithMax(o.MaxRetry))))
 	} else {
 		// It does not return error if the collector is not running
-		conn, _ = grpc.Dial(o.CollectorHostPort[0], grpc.WithInsecure(), grpc.WithBalancerName(roundrobin.Name))
+		conn, _ = grpc.Dial(o.CollectorHostPort[0], grpc.WithInsecure(), grpc.WithBalancerName(roundrobin.Name),
+			grpc.WithUnaryInterceptor(grpc_retry.UnaryClientInterceptor(grpc_retry.WithMax(o.MaxRetry))))
```
yurishkuro: Can we DRY this so that only the actual differences are inside the if branches?

yurishkuro: Also, let's format it one option per line.

cmd/agent/app/reporter/grpc/reporter.go
```diff
@@ -29,6 +29,10 @@ import (
 	"github.com/jaegertracing/jaeger/thrift-gen/zipkincore"
 )
 
+const (
+	defaultMaxRetry = 3
+)
```
yurishkuro: Can we move this to flags.go where it is used?

cmd/agent/app/reporter/grpc/collector_proxy.go
```diff
@@ -41,19 +42,23 @@ func NewCollectorProxy(o *Options, mFactory metrics.Factory, logger *zap.Logger)
...
 	} else {
-		// It does not return error if the collector is not running
-		conn, _ = grpc.Dial(o.CollectorHostPort[0], grpc.WithInsecure(), grpc.WithBalancerName(roundrobin.Name))
+		target = o.CollectorHostPort[0]
 	}
+	// It does not return error if the collector is not running
+	conn, _ := grpc.Dial(target,
+		grpc.WithInsecure(),
+		grpc.WithBalancerName(roundrobin.Name),
+		grpc.WithUnaryInterceptor(grpc_retry.UnaryClientInterceptor(grpc_retry.WithMax(o.MaxRetry))))
```
vprithvi: How does this retry work? Is it possible to configure it with exponential backoff, or some backoff for e.g.?

pavolloffay: By default it is using `BackoffLinearWithJitter`. Which can be changed.

https://github.com/grpc-ecosystem/go-grpc-middleware/blob/master/retry/options.go#L26

pavolloffay:
Travis passed, I will try to restart it and hope :)

pavolloffay:
I have restarted the build 4 times, it passed. I will merge this to see how it behaves in other repositories

PR #1385, opened 2019-02-26
Ignore DS_Store in file .gitignore

truongnh1992:
Signed-off-by: Nguyen Hai Truong <truongnh@vn.fujitsu.com>

- 

-


yurishkuro:
I suggest you add it to your `~/.gitconfig`, we don't need to track it in the repo.

jpkrohling:
I'm closing, but I think we should consider using a more complete gitignore, like the one generated from [gitignore.io](https://www.gitignore.io/)

PR #1391, opened 2019-02-28
Support secure gRPC channel between agent and collector

yurishkuro:
Extends PR #1311 (which was in ghouscht:master, hard to add to directly)


yossishm:
Hi,
A Q?
On which Jaeger Collector/Agent version is this issue resolved?

I configured a tls on p 14250 between Agent and Collector and Using a wireshark container sniffing I can see the spans which means they aren't encrypted (Did the same TLS checkup on Collector to elastic and data was encrypted).

Attaching sniffing snapshot.

Jaeger-Sniffing


Thanks,
Y

PR #1396, opened 2019-03-03
Adds support for agent level tag

### LINKED ISSUE #833: Support for Agent level tag
I think it would be nice to be able to add tag at the Agent level. It would be similar to adding the tag at the Tracer level via `JAEGER_TAGS` environment variable

Here's some example of the tags we like to be able to inject into each traces:
- AWS Instance metadata (similar to https://github.com/takus/fluent-plugin-ec2-metadata
- Kubernetes metadata (similar to https://fluentbit.io/documentation/0.13/filter/kubernetes.html)
- IP address (if you run the Agent and the application with the Tracer in different container, they will likely have different IP address)

I don't there's anything blocking this, it's a feature request

Ability to provide Jaeger Agent with tags at startup (via something similar to JAEGER_TAGS)

- Should these tags be stored and displayed like the Tracer tags? (under the `Process` section in the Query UI)

_edited to remove the plugin proposition and to reflect the latest state of the request_

annanay25:

- Resolves #833 

- Uses a `--jaeger.tags` flag to enable tag addition (Ex: `--jaeger.tags=key=value`). This can be used with env variables.


cmd/agent/app/reporter/flags.go
```diff
@@ -35,15 +37,18 @@ type Type string
 // Options holds generic reporter configuration.
 type Options struct {
 	ReporterType Type
+	AgentTags    Type
 }
 
 // AddFlags adds flags for Options.
 func AddFlags(flags *flag.FlagSet) {
 	flags.String(reporterType, string(GRPC), fmt.Sprintf("Reporter type to use e.g. %s, %s", string(GRPC), string(TCHANNEL)))
+	flags.String(agentTags, "", fmt.Sprintf("Add agent-based tags. Ex: --jaeger.tags key1=value1,${envVar:defaultValue}"))
```
yurishkuro: what does `,${envVar:defaultValue}` mean?

cmd/agent/app/reporter/flags.go
```diff
@@ -27,6 +27,8 @@ const (
 	TCHANNEL Type = "tchannel"
 	// GRPC is name of gRPC reporter.
 	GRPC Type = "grpc"
+	// Agent tags
```
yurishkuro: move next to `reporter.type` flag

cmd/agent/app/reporter/flags.go
```diff
@@ -35,15 +37,18 @@ type Type string
 // Options holds generic reporter configuration.
 type Options struct {
 	ReporterType Type
+	AgentTags    Type
```
yurishkuro: I think this should be map[string]string

cmd/agent/app/reporter/grpc/reporter.go
```diff
@@ -34,20 +34,22 @@ type Reporter struct {
...
 }
 
 // NewReporter creates gRPC reporter.
-func NewReporter(conn *grpc.ClientConn, logger *zap.Logger) *Reporter {
+func NewReporter(conn *grpc.ClientConn, agentTags string, logger *zap.Logger) *Reporter {
 	return &Reporter{
 		collector: api_v2.NewCollectorServiceClient(conn),
 		logger:    logger,
 		sanitizer: zipkin2.NewChainedSanitizer(zipkin2.StandardSanitizers...),
+		agentTags: parseAgentTags(agentTags),
```
yurishkuro: mixing concern: parsing of tags string should be done wherever CLI flags are parsed. Reporter should take `map[string]string` or even better `[]model.Tag`

cmd/agent/app/reporter/grpc/reporter.go
```diff
@@ -34,20 +34,22 @@ type Reporter struct {
...
 	}
 }
 
 // EmitBatch implements EmitBatch() of Reporter
 func (r *Reporter) EmitBatch(b *thrift.Batch) error {
-	return r.send(jConverter.ToDomain(b.Spans, nil), jConverter.ToDomainProcess(b.Process))
+	return r.send(addAgentTags(jConverter.ToDomain(b.Spans, nil), r.agentTags), jConverter.ToDomainProcess(b.Process))
```
yurishkuro: this statement is becoming hard to read, I recommend splitting into local vars

cmd/agent/app/reporter/grpc/reporter.go
```diff
@@ -59,7 +61,7 @@ func (r *Reporter) EmitZipkinBatch(zSpans []*zipkincore.Span) error {
 	if err != nil {
 		return err
 	}
-	return r.send(trace.Spans, nil)
+	return r.send(addAgentTags(trace.Spans, r.agentTags), nil)
```
yurishkuro: seems not DRY, can do this in one place, in send()

cmd/agent/app/reporter/grpc/reporter.go
```diff
@@ -71,3 +73,40 @@ func (r *Reporter) send(spans []*model.Span, process *model.Process) error {
 	}
 	return err
 }
+
+// addAgentTags appends jaeger tags for the agent to every span it sends to the collector.
+func addAgentTags(spans []*model.Span, agentTags []model.KeyValue) []*model.Span {
+	for _, span := range spans {
+		for _, tag := range agentTags{
+			span.Tags = append(span.Tags, tag)
```
yurishkuro: top-level question for the PR: should these be added to Span tags or Process tags?

cmd/agent/app/reporter/flags.go
annanay25: Copied over from https://github.com/jaegertracing/jaeger-client-go/blob/c58709fe19d793cdd40ad8bfa6e645a40a2eccf8/config/config_env.go#L195-L199 -

```
// parseTags parses the given string into a collection of Tags.
// Spec for this value:
// - comma separated list of key=value
// - value can be specified using the notation ${envVar:defaultValue}, where `envVar`
// is an environment variable and `defaultValue` is the value to use in case the env var is not set
```

cmd/agent/app/reporter/flags.go
annanay25: Will do.

cmd/agent/app/reporter/flags.go
annanay25: +1 to `[]model.Tag` as per your comment below.

cmd/agent/app/reporter/grpc/reporter.go
annanay25: Will do.

cmd/agent/app/reporter/grpc/reporter.go
annanay25: Will do.

cmd/agent/app/reporter/grpc/reporter.go
```diff
@@ -71,3 +73,40 @@ func (r *Reporter) send(spans []*model.Span, process *model.Process) error {
...
+		}
+	}
+	return spans
+}
+
+// Parsing logic borrowed from jaegertracing/jaeger-client-go
+func parseAgentTags(agentTags string) []model.KeyValue {
```
pavolloffay: move also this to flags.go

cmd/agent/app/reporter/tchannel/reporter.go
```diff
@@ -106,3 +114,52 @@ func (r *Reporter) submitAndReport(submissionFunc func(ctx thrift.Context) error
...
+	for _, span := range spans {
+		for _, tag := range agentTags{
+			span.Tags = append(span.Tags, &tag)
+		}
+	}
+	return spans
+}
+
+// Parsing logic borrowed from jaegertracing/jaeger-client-go
+func parseAgentTags(agentTags string) []jaeger.Tag {
```
pavolloffay: There is also `func parseAgentTags(agentTags string) []model.KeyValue {`.

Could we just have one version `func parseAgentTags(agentTags string) map[string][string]` or even return a list of tuples if we support multiple tags with the same keys and move it to flags.go.The map would be passed to reporter and at init time converted to list of appropriate model tags.

annanay25: Done.

cmd/agent/app/reporter/grpc/reporter.go
annanay25: Done.

cmd/agent/app/reporter/grpc/reporter.go
pavolloffay: It might depend. For example in k8s we support deploying agent as a side-car but also as daemon set (1 agent per k8s node). 

I think we would need an additional flag to be able to add it to process tags

cmd/agent/app/reporter/tchannel/reporter.go
```diff
@@ -106,3 +112,38 @@ func (r *Reporter) submitAndReport(submissionFunc func(ctx thrift.Context) error
...
 	return r.serviceName
 }
+
+func addAgentTagsToZipkinBatch(spans []*zipkincore.Span, agentTags []jaeger.Tag) []*zipkincore.Span {
+	for _, span := range spans {
+		for _, tag := range agentTags{
+			span.BinaryAnnotations = append(span.BinaryAnnotations, &zipkincore.BinaryAnnotation{
+				Key: tag.Key,
+				Value: []byte(*tag.VStr),
+				AnnotationType: 6, // static value set to string type for now.
```
pavolloffay: nit: could be this defined as a constant?

cmd/agent/app/reporter/tchannel/reporter.go
```diff
@@ -106,3 +112,38 @@ func (r *Reporter) submitAndReport(submissionFunc func(ctx thrift.Context) error
...
+func addAgentTags(spans []*jaeger.Span, agentTags []jaeger.Tag) []*jaeger.Span {
+	for _, span := range spans {
+		for _, tag := range agentTags{
+			span.Tags = append(span.Tags, &tag)
+		}
+	}
+	return spans
+}
+
+func makeJaegerTag(agentTags map[string]string) []jaeger.Tag {
```
pavolloffay: make the name plural

cmd/agent/app/reporter/tchannel/reporter.go
```diff
@@ -106,3 +112,38 @@ func (r *Reporter) submitAndReport(submissionFunc func(ctx thrift.Context) error
...
+	tags := make([]jaeger.Tag, 0)
+	for k, v := range agentTags {
+		tag := jaeger.Tag{
+			Key: k,
+			VStr: &v,
```
pavolloffay: Set the type````

cmd/agent/app/reporter/tchannel/reporter.go
```diff
@@ -106,3 +112,38 @@ func (r *Reporter) submitAndReport(submissionFunc func(ctx thrift.Context) error
...
+	for _, span := range spans {
+		for _, tag := range agentTags{
+			span.Tags = append(span.Tags, &tag)
+		}
+	}
+	return spans
+}
+
+func makeJaegerTag(agentTags map[string]string) []jaeger.Tag {
+	tags := make([]jaeger.Tag, 0)
```
pavolloffay: the size is known at init time

cmd/agent/app/reporter/tchannel/reporter.go
```diff
@@ -106,3 +112,38 @@ func (r *Reporter) submitAndReport(submissionFunc func(ctx thrift.Context) error
...
+				Key: tag.Key,
+				Value: []byte(*tag.VStr),
+				AnnotationType: 6, // static value set to string type for now.
+			})
+		}
+	}
+	return spans
+}
+
+func addAgentTags(spans []*jaeger.Span, agentTags []jaeger.Tag) []*jaeger.Span {
```
pavolloffay: nit: rename it e.g. `addTags(spans, tags)`...

cmd/agent/app/reporter/grpc/reporter.go
annanay25: > I think we would need an additional flag to be able to add it to process tags

Is that the consensus?

annanay25:
Thanks for the reviews. Addressed comments. Just the top level question remains.

cmd/agent/app/reporter/flags.go
```diff
@@ -17,12 +17,17 @@ package reporter
...
+	"strings"
 
 	"github.com/spf13/viper"
 )
 
 const (
+	// Whether to use grpc or tchannel reporter.
 	reporterType = "reporter.type"
+	// Agent tags
+	agentTags = "jaeger.tag"
```
yurishkuro: `"jaeger.tags"`

cmd/agent/app/reporter/grpc/reporter.go
```diff
@@ -71,3 +74,24 @@ func (r *Reporter) send(spans []*model.Span, process *model.Process) error {
 	}
 	return err
 }
+
+// addTags appends jaeger tags for the agent to every span it sends to the collector.
+func addTags(spans []*model.Span, agentTags []model.KeyValue) []*model.Span {
+	for _, span := range spans {
+		span.Tags = append(span.Tags, agentTags...)
```
yurishkuro: why span.Tags instead of span.Process.tags?  I think a typical use case here would be adding some metadata that's constant for all spans from a given application

cmd/agent/app/reporter/grpc/reporter.go
```diff
@@ -71,3 +74,24 @@ func (r *Reporter) send(spans []*model.Span, process *model.Process) error {
...
+	}
+	return spans
+}
+
+func makeModelKeyValue(agentTags map[string]string) []model.KeyValue {
+	tags := make([]model.KeyValue, len(agentTags))
+	for k, v := range agentTags {
+		tag := model.KeyValue{
```
yurishkuro: please use constructor function instead: model.String(k, v)

cmd/agent/app/reporter/tchannel/reporter.go
```diff
@@ -26,6 +26,10 @@ import (
 	"github.com/jaegertracing/jaeger/thrift-gen/zipkincore"
 )
 
+const (
+	annotationTypeStr = 6
```
yurishkuro: there's already a constant zipkincore.AnnotationType_STRING

annanay25: Changed.

annanay25:
Thanks for the review @yurishkuro!

I've added tests only for the grpc reporter for now, but do let me know if TChannel reporter tests are also required.

yurishkuro:
If we're not adding tchannel tests, we shouldn't change tchannel code either, otherwise coverage will go down (which I can't see now because the bloody crossdock test failed again).

annanay25:
Codecov is at 73%, let me remove all changes to tchannel reporter.

annanay25:
Crossdock again 😅

yurishkuro:
the main addition seems to be untested

image

cmd/agent/app/reporter/flags_test.go
```diff
@@ -34,10 +54,15 @@ func TestBindFlags(t *testing.T) {
...
+		"--jaeger.tags=key=value,envVar1=${envKey1:defaultVal1},envVar2=${envKey2:defaultVal2}",
 	})
 	require.NoError(t, err)
 
 	b := &Options{}
+	os.Setenv("envKey1","envVal1")
 	b.InitFromViper(v)
+
 	assert.Equal(t, Type("grpc"), b.ReporterType)
+	assert.Equal(t, parseAgentTags("key=value,envVar1=${envKey1:defaultVal1},envVar2=${envKey2:defaultVal2}"), b.AgentTags)
```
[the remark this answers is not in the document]
annanay25: Done.

cmd/agent/app/reporter/flags.go
```diff
@@ -35,15 +40,44 @@ type Type string
 // Options holds generic reporter configuration.
 type Options struct {
 	ReporterType Type
+	AgentTags    map[string]string
 }
 
 // AddFlags adds flags for Options.
 func AddFlags(flags *flag.FlagSet) {
 	flags.String(reporterType, string(GRPC), fmt.Sprintf("Reporter type to use e.g. %s, %s", string(GRPC), string(TCHANNEL)))
+	flags.String(agentTags, "", fmt.Sprintf("Add agent-based tags. Ex: --jaeger.tags=key1=value1,${envVar:defaultValue}"))
```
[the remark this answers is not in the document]
annanay25: Done.

cmd/agent/app/reporter/grpc/reporter.go
```diff
@@ -71,3 +74,27 @@ func (r *Reporter) send(spans []*model.Span, process *model.Process) error {
 	}
 	return err
 }
+
+// addTags appends jaeger tags for the agent to every span it sends to the collector.
+func addProcessTags(spans []*model.Span, agentTags []model.KeyValue) []*model.Span {
+	for _, span := range spans {
+		if len(agentTags) > 0 {
+			if span.Process == nil {
+				span.Process = &model.Process{Tags: agentTags}
```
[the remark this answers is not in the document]
annanay25: Done.

cmd/agent/app/reporter/grpc/reporter_test.go
```diff
@@ -71,7 +72,45 @@ func TestReporter_EmitZipkinBatch(t *testing.T) {
...
+		if test.err != "" {
+			assert.EqualError(t, err, test.err)
+		} else {
+			assert.Equal(t, 1, len(handler.requests))
+			assert.Equal(t, test.expected, handler.requests[0].GetBatch())
+		}
+	}
+}
+
+func TestReporter_EmitZipkinBatch_WithAgentTags(t *testing.T) {
```
yurishkuro: A lot of repetition. I would suggest testing the function that merges agent tags into spans/process directly rather than going through client/server setup as in the test above.

cmd/agent/app/reporter/grpc/reporter.go
```diff
@@ -71,3 +74,27 @@ func (r *Reporter) send(spans []*model.Span, process *model.Process) error {
 	}
 	return err
 }
+
+// addTags appends jaeger tags for the agent to every span it sends to the collector.
+func addProcessTags(spans []*model.Span, process *model.Process, agentTags []model.KeyValue) ([]*model.Span, *model.Process) {
+	if process != nil {
+		process.Tags = append(process.Tags, agentTags...)
+		return spans, process
```
yurishkuro: don't need to return (see below)

cmd/agent/app/reporter/grpc/reporter.go
```diff
@@ -71,3 +74,27 @@ func (r *Reporter) send(spans []*model.Span, process *model.Process) error {
...
+	}
+	for _, span := range spans {
+		if len(agentTags) > 0 {
```
yurishkuro: This check can be at the top of the function (return immediately if no agent tags).

Instead, here the check  should be `if span.Process != nil`

cmd/agent/app/reporter/flags.go
```diff
@@ -35,15 +40,45 @@ type Type string
...
 type Options struct {
 	ReporterType Type
+	AgentTags    map[string]string
 }
 
 // AddFlags adds flags for Options.
 func AddFlags(flags *flag.FlagSet) {
 	flags.String(reporterType, string(GRPC), fmt.Sprintf("Reporter type to use e.g. %s, %s", string(GRPC), string(TCHANNEL)))
+	flags.String(agentTags, "", fmt.Sprintf(`One or more tags to be added to the Process tags of all spans passing through this agent.
+		Ex: --jaeger.tags=key1=value1,key2=${envVar:defaultValue}`))
```
yurishkuro: I would remove `--jaeger.tags=` prefix, limit example to the sample value string `key1=value1,key2=${envVar:defaultValue}`

cmd/agent/app/reporter/grpc/reporter.go
```diff
@@ -71,3 +74,27 @@ func (r *Reporter) send(spans []*model.Span, process *model.Process) error {
...
+			span.Process.Tags = append(span.Process.Tags, agentTags...)
+		}
+	}
+	return spans, nil
+}
+
+func makeModelKeyValue(agentTags map[string]string) []model.KeyValue {
+	tags := []model.KeyValue{}
```
yurishkuro: `make([]model.KeyValue, 0, len(agentTags))`

cmd/agent/app/reporter/grpc/reporter_test.go
annanay25: My bad, fixed.

cmd/agent/app/reporter/grpc/reporter.go
annanay25: This would involve an iterator like so -
```
func makeModelKeyValue(agentTags map[string]string) []model.KeyValue {
	tags := make([]model.KeyValue, 0, len(agentTags))
	i := 0
	for k, v := range agentTags {
		tag := model.String(k, v)
		tags[i] = tag
		i++
	}

	return tags
}
``` 

Should I change to this if it looks better?

yurishkuro: it doesn't change the loop, because the slice is allocated with 0 initial length, so append() works the same way (but without extra allocations).

cmd/agent/app/reporter/grpc/reporter.go
```diff
@@ -71,3 +74,30 @@ func (r *Reporter) send(spans []*model.Span, process *model.Process) error {
...
+	}
+	if process != nil {
+		process.Tags = append(process.Tags, agentTags...)
+		return spans, process
+	}
+	for _, span := range spans {
+		// span.Process will never be nil -
+		// Zipkin batch will always populate span.Process
+		span.Process.Tags = append(span.Process.Tags, agentTags...)
+	}
```
yurishkuro: I don't think we should be hardcoding the assumption that `process` and `span.Process` are mutually exclusive. There are cases when a single application generates spans on behalf of different services (e.g. a routing sidecar). It is possible that a single submitted Batch has `batch.Process` defined, but also some individual `span.Process` are not nil. There's no reason to exclude this use case from this function, just need to (a) not return in L85, and (b) check that `span.Process != nil` before L90.

annanay25: Thanks for the explanation, the use-case was unclear to me, this was very helpful :)

yurishkuro: it's the general principle: don't assume the clients will behave in the expected way, try to make sense of all possible inputs

annanay25:
Thanks! :)

PR #1417, opened 2019-03-10
Fix some misspells in comment

JoeWrightss:
"compatibilty" to "compatibility" in line 733 and 745.


yurishkuro:
We don't want to manually change auto-generated files. If you you like to see this through, we need to update the thrift file in jaeger-idl repo, and then regenerate the code here. But I don't think it's worth it for a spelling mistake in the comment.

PR #1421, opened 2019-03-12
use RawMessage in check of value for being json

funny-falcon:
Optimizes a bit "isJSON" check for tag value.

RawMessage replaces `map[string]interface{}` as a target for json.Unmarshal.
Unmarshalling checks JSON validity even when unmarshal to RawMessage.


vprithvi:
I'm a bit unclear - is this quicker/more efficient? Could you detail how/why?

funny-falcon:
Yes, it is quicker. RawMessage is just `[]byte` and json is put into as a slice of bytes in a single allocation.

It is much more efficient than filling `map[string]interface{}`, because map allocates a lot by itself, and then every key is allocation of string and every value is also allocation (because current golang version could not store anything except pointer without allocation). Obviously, nested maps also will allocate (empty map weightsmore than 100bytes afaik).

yurishkuro:
Thanks!

PR #1427, opened 2019-03-16
New layout for proto definitions and generated files

annanay25:
## Which problem is this PR solving?
- Part of #904 

## Short description of the changes
- Split `api_v2.proto` into respective services - `collector` and `sampling`(manager)
- Generate grpc stubs and gateway code from these files
- Generate unified `swagger.json` from the multiple proto definitions

PR #1445, opened 2019-03-26
Ignore autogenerated files from import-order-cleanup

annanay25:

- Travis complains about unformatted proto generated files

- Add list of proto generated files to ignored files in import-order-cleanup.


annanay25:
Seeing this issue in #1427

annanay25:
`TestSaramaConsumerWrapper_start_Messages` seems to have failed.

annanay25:
Actually `make fmt` will not touch any autogenerated files - 
https://github.com/jaegertracing/jaeger/blob/6bbe8dc8ff0fc1a97f87f84413b461f6268c0f2f/Makefile#L4-L20
What we actually need is to remove the fmt check for autogenerated files.

PR #1454, opened 2019-04-03
Fix HotRod job pool termination (#1453)

### LINKED ISSUE #1453: HotRod: pool: Stop all the workers by closing the stop channel
Instead of sending a signal on the stop chan, closing the stop chan is the correct way to stop all the workers; Reading from a closed chan is always successful.

Close the `stop` channel instead of sending values to it.

mfrw:
- Instead of sending a signal on the stop chan, closing the stop chan is the correct way to stop all the workers; Reading from a closed chan is always successful. (Resolves #1453)

- Close the stop channel instead of sending values to it.


examples/hotrod/pkg/pool/pool.go
```diff
@@ -49,5 +49,7 @@ func (p *Pool) Execute(job func()) {
 
 // Stop halts all the workers
 func (p *Pool) Stop() {
-	p.stop <- struct{}{}
+	if p.stop != nil {
```
yurishkuro: why is the nil check necessary?

mfrw: @yurishkuro sorry for the late reply,
IMHO, this is a bug. Assuming you have more than 1 worker, calling stop only stops 1 of the workers and not all, while as closing the chan stops all the workers. Please let me know if you still want me to rename the PR.

yurishkuro: good point

PR #1446, opened 2019-03-26
Add inbound transport as label to collector metrics

guanw:
## Which problem is this PR solving?
Resolves #1444 

## Short description of the changes
- Adding metrics on grpc/http/tchannel endpoints for collector

PR #1458, opened 2019-04-03
Refactor handlers to separate transport and span format concerns 

yurishkuro:
implements https://github.com/jaegertracing/jaeger/pull/1446#discussion_r271811943


cmd/collector/app/tchannel_handler.go
```diff
@@ -0,0 +1,52 @@
...
+package app
+
+import (
+	"github.com/uber/tchannel-go/thrift"
+
+	"github.com/jaegertracing/jaeger/thrift-gen/jaeger"
+	"github.com/jaegertracing/jaeger/thrift-gen/zipkincore"
+)
+
+type tchannelHandler struct {
```
guanw: Maybe we can add an interface for tchannel handler, that way it's more clear this handler is satisfying the thrift definition?

``` 
type tchanHandler struct {
	zipkinSpansHandler ZipkinSpansHandler
	jaegerSpansHandler JaegerBatchesHandler
}
```

yurishkuro: it already implements existing interfaces, no benefit of defining more. I added comments and the API checks to the test file,

guanw:
Actually can we make separate folders (e.g transport/representation/processor) to separate handlers? The current "flat" structure doesn't help people understand handler layers.

yurishkuro:
We could, but the packages will be very small, one file, it's not the recommended practice in Go to have so small packages.

cmd/collector/app/span_processor.go
```diff
@@ -26,6 +26,18 @@ import (
 	"github.com/jaegertracing/jaeger/storage/spanstore"
 )
 
+// ProcessSpansOptions additional options passed to processor along with the spans.
+type ProcessSpansOptions struct {
```
jpkrohling: What would happen if neither `SpanFormat` nor `InboundTransport` are set?

cmd/collector/app/thrift_span_handler.go
```diff
@@ -35,22 +34,21 @@ const (
 	UnknownFormatType = "unknown"
 )
 
+// SubmitBatchOptions are passed to Submit methods of the handlers.
+type SubmitBatchOptions struct {
+	InboundTransport string
```
jpkrohling: Same as above: what are the implications of having the `InboundTransport` not being specified?

guanw: I have a followup story to take care of that. I think if `InBoundTransport` is not set, we would simply skip setting another dimension for span counter metrics.

yurishkuro: We shouldn't skip, but assign an "undefined" value so that the total counter is still correct.

guanw: > We shouldn't skip, but assign an "undefined" value so that the total counter is still correct.

hmmm. What I meant was something like the following:
```
    tags := map[string]string{"svc": serviceName, "debug": debugStr}
    if endpointType != "" {
	  tags["transport"] = endpointType
    }
    c := m.factory.Counter(metrics.Options{Name: m.category, Tags: tags})
```
so if endpointType is empty we still count it just not having a transport tag. Is adding `undefined` tag mandatory here for metrics to work correctly?

yurishkuro: Well, first, we don't create counter objects with labels in real time, because it's expensive and we always prefer to pre-create them when possible and cache. Second, when creating a counter we must define a set of labels (tags) it has. Some metrics backends may not like the value of the label to be blank. Why not be explicit about the name then?
```
    tags := map[string]string{"svc": serviceName, "debug": debugStr,  "transport": "undefined"}
    if endpointType != "" {
	  tags["transport"] = endpointType
    }
```

guanw: gotcha

cmd/collector/app/grpc_handler.go
```diff
@@ -43,7 +43,10 @@ func (g *GRPCHandler) PostSpans(ctx context.Context, r *api_v2.PostSpansRequest)
 			span.Process = r.Batch.Process
 		}
 	}
-	_, err := g.spanProcessor.ProcessSpans(r.GetBatch().Spans, JaegerFormatType)
+	_, err := g.spanProcessor.ProcessSpans(r.GetBatch().Spans, ProcessSpansOptions{
+		InboundTransport: "grpc", // TODO do we have a constant?
```
vprithvi: Could we create constants if they don't exist?

yurishkuro: Jude is doing it in the next PR. Right now these strings are not used anywhere.

guanw: Yes, I have the followup story to address that.

PR #1460, opened 2019-04-04
Set default zap sampling config to nil

### LINKED ISSUE #160: Default logging shouldn't be sampled
Logs are sampled, and any more than ~100 logs lines per second would be dropped. 
Change this so that all logs are output.

psinghal20:
- Resolves #160. Sets the default config behavior to no sampling for zap

- Changes the default sampling config for the zap logger to nil


yurishkuro:
thanks!

PR #1465, opened 2019-04-08
Return info msg for `/health` endpoint

### LINKED ISSUE #1450: Return `uptime` or something similar for `/health` endpoint
Currently, the `/health` endpoint returns a `204` when it becomes `Ready`, which isn't a great indicator of whether the service is healthy or not.
```
// New creates a HealthCheck with the specified initial state.
func New() *HealthCheck {
	hc := &HealthCheck{
		state: int32(Unavailable),
		mapping: map[Status]int{
			Unavailable: http.StatusServiceUnavailable,
			Ready:       http.StatusNoContent,
		},
		logger: zap.NewNop(),
	}
	return hc
}
```
However, it would be nice to return a `200` and something like:
`{"ok":true,"status":"up"}` or even `{"uptime":1h20m}`, etc.

stefanvassilev:
- Return 200 OK for status `ready`
- Return Json containing time `startedAt` and `upTime` statistics

Resolves #1450

### DIFF
--- a/crossdock/main.go
+++ b/crossdock/main.go
@@ -105,7 +105,7 @@ func (h *clientHandler) isInitialized() bool {
 func httpHealthCheck(logger *zap.Logger, service, healthURL string) {
 	for i := 0; i < 240; i++ {
 		res, err := http.Get(healthURL)
-		if err == nil && res.StatusCode == 204 {
+		if err == nil && res.StatusCode == 200 {
 			logger.Info("Health check successful", zap.String("service", service))
 			return
 		}
--- a/crossdock/main.go
+++ b/crossdock/main.go
@@ -102,10 +102,14 @@ func (h *clientHandler) isInitialized() bool {
 	return atomic.LoadUint64(&h.initialized) != 0
 }
 
+func is2xxStatusCode(statusCode int) bool {
+	return statusCode >= 200 && statusCode <= 299
+}
+
 func httpHealthCheck(logger *zap.Logger, service, healthURL string) {
 	for i := 0; i < 240; i++ {
 		res, err := http.Get(healthURL)
-		if err == nil && res.StatusCode == 200 {
+		if err == nil && is2xxStatusCode(res.StatusCode) {
 			logger.Info("Health check successful", zap.String("service", service))
 			return
 		}
--- a/pkg/healthcheck/handler.go
+++ b/pkg/healthcheck/handler.go
@@ -72,10 +72,7 @@ type HealthCheck struct {
 // New creates a HealthCheck with the specified initial state.
 func New() *HealthCheck {
 	hc := &HealthCheck{
-		state: int32(Unavailable),
-		upTimeStats: upTimeStats{
-			StartedAt: time.Now(),
-		},
+		state:  int32(Unavailable),
 		logger: zap.NewNop(),
 		mapping: map[Status]healthCheckResponse{
 			Unavailable: {
@@ -84,7 +81,7 @@ func New() *HealthCheck {
 			},
 			Ready: {
 				statusCode: http.StatusOK,
-				StatusMsg:  "up",
+				StatusMsg:  "Server available",
 			},
 		},
 		server: nil,
@@ -110,7 +107,6 @@ func (hc *HealthCheck) Handler() http.Handler {
 }
 
 func (hc *HealthCheck) createRespBody(resp healthCheckResponse) []byte {
-
 	timeStats := hc.upTimeStats
 	if resp.statusCode == http.StatusOK {
 		timeStats.UpTime = upTime(timeStats.StartedAt)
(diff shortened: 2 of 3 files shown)


pkg/healthcheck/handler.go
```diff
@@ -46,23 +48,46 @@ func (s Status) String() string {
...
 		},
 		logger: zap.NewNop(),
+		mapping: map[Status]healthCheckResponse{
+			Unavailable: {
+				statusCode: http.StatusServiceUnavailable,
+				StatusMsg:  "Server not available",
+			},
+			Ready: {
+				statusCode: http.StatusOK,
+				StatusMsg:  "up",
```
black-adder: perhaps `Server available`

pkg/healthcheck/internal_test.go
```diff
@@ -34,5 +37,30 @@ func TestHttpCall(t *testing.T) {
...
-	assert.Equal(t, http.StatusNoContent, resp.StatusCode)
+	assert.Equal(t, http.StatusOK, resp.StatusCode)
+
+	hr := getHealthCheckResponse(t, resp, )
+
+	assert.Equal(t, "up", hr.StatusMsg)
+	assert.Equal(t, hc.upTimeStats.UpTime, hr.upTimeStats.UpTime)
+	assert.True(t, hc.upTimeStats.StartedAt.Equal(hr.upTimeStats.StartedAt))
+
+	time.Sleep(300)
```
black-adder: this will add an extra 5 minutes for each test run which is overkill

pkg/healthcheck/handler.go
```diff
@@ -75,12 +100,33 @@ func (hc *HealthCheck) SetLogger(logger *zap.Logger) {
...
+		resp := hc.mapping[hc.Get()]
+		w.WriteHeader(resp.statusCode)
+
+		w.Header().Set("Content-Type", "application/json")
+		w.Write(hc.createRespBody(resp))
 	})
 }
 
+func (hc *HealthCheck) createRespBody(resp healthCheckResponse) []byte {
+
```
black-adder: rm newline

pkg/healthcheck/handler.go
```diff
@@ -46,23 +48,46 @@ func (s Status) String() string {
...
 
 // New creates a HealthCheck with the specified initial state.
 func New() *HealthCheck {
 	hc := &HealthCheck{
 		state: int32(Unavailable),
-		mapping: map[Status]int{
-			Unavailable: http.StatusServiceUnavailable,
-			Ready:       http.StatusNoContent,
+		upTimeStats: upTimeStats{
+			StartedAt: time.Now(),
```
black-adder: technically, shouldn't this be set when the sever is first set to ready?

pkg/healthcheck/internal_test.go
stefanvassilev: Those are `Nanosecond`. I wanted to check the case where some time's passed and the service went down, that the uptime does not get incremented.

crossdock/main.go
```diff
@@ -105,7 +105,7 @@ func (h *clientHandler) isInitialized() bool {
 func httpHealthCheck(logger *zap.Logger, service, healthURL string) {
 	for i := 0; i < 240; i++ {
 		res, err := http.Get(healthURL)
-		if err == nil && res.StatusCode == 204 {
+		if err == nil && res.StatusCode == 200 {
```
jpkrohling: I'd say that the most appropriate is to check if the response is a 2xx class. That's what monitoring solutions do.

pkg/healthcheck/handler.go
```diff
@@ -94,5 +136,8 @@ func (hc *HealthCheck) Get() Status {
 
 // Ready is a shortcut for Set(Ready) (kept for backwards compatibility)
 func (hc *HealthCheck) Ready() {
+	hc.upTimeStats = upTimeStats{
+		StartedAt: time.Now(),
```
jpkrohling: The [process' start time](https://godoc.org/github.com/shirou/gopsutil/process#Process.CreateTime) is recorded as a metric:

```
# HELP process_start_time_seconds Start time of the process since unix epoch in seconds.
# TYPE process_start_time_seconds gauge
process_start_time_seconds 1.55480578686e+09
```

Should we add this one as well?

stefanvassilev: Should I add this as well?

jpkrohling: I meant the other way around, to add the "Started At" as a metric.

stefanvassilev: Where should I add it?

jpkrohling: Looks like the handler isn't receiving a metrics object. Perhaps it's better to open a follow-up issue for now.

stefanvassilev: Perhaps a follow-up issue of what should be done would be more helpful, as on my local build the `/metrics` endpoint does not return any `process` information. :)

yurishkuro:
Thanks for the PR!

PR #1469, opened 2019-04-11
Ignore return value

code98:
Ignore useless return values

This is the first time I have submitted source code to jaeger, and I will submit my code often in the future. 

thanks.


yurishkuro:
cannot merge with DCO check failing

Issue #1438, opened 2019-03-21
ClickHouse as a storage backend


sboisson:
[ClickHouse](https://github.com/yandex/ClickHouse), an open-source column-oriented DBMS designed initially for real-time analytics and mostly write-once/read-many big-data use cases, can be used as a very efficient log and trace storage.

Meta issue: #638 Additional storage backends

bzon:
Is anyone working on this? I or my team can maybe give a shot at this.

Slach:
as i now noone working on this

Slach:
@bzon i can join you as tester

bzon:
> @bzon i can join you as tester

Sure!

sboisson:
Happy to see someone working on this :) 
Might be able to join as tester

yurishkuro:
@bzon would be good if you post an architecture here, specifically how you would lay out the data, ingestion, etc. To my knowledge, clickhouse requires batched writes, and it may even be up to you to decide which node to send the writes to, so there are many questions. It may require some benchmarking to find the optimal design.

bzon:
@yurishkuro at the moment, we have zero knowledge with the internals of jaeger. I think that benchmarking should be the first step to see if this integration is feasible. And with that said, the first requirement should be creating the right table schema.

sboisson:
This webinar could be interesting: [A Practical Introduction to Handling Log Data in ClickHouse](https://www.altinity.com/webinarspage/2019/12/17/a-practical-introduction-to-handling-log-data-in-clickhouse)

bobrik:
I took a stab at it (very early WIP):

* https://github.com/jaegertracing/jaeger/compare/master...bobrik:ivan/clickhouse

This is the schema I used:

* Index table for fast searches (haven't measured if indices are useful yet)

```sql
CREATE TABLE jaeger_index (
  timestamp DateTime64(6),
  traceID FixedString(16),
  service LowCardinality(String),
  operation LowCardinality(String),
  durationUs UInt64,
  tags Nested(
    key LowCardinality(String),
    valueString LowCardinality(String),
    valueBool UInt8,
    valueInt Int64,
    valueFloat Float64
  ),
  INDEX tags_strings (tags.key, tags.valueString) TYPE set(0) GRANULARITY 64,
  INDEX tags_ints (tags.key, tags.valueInt) TYPE set(0) GRANULARITY 64
[...]

PR #1475, opened 2019-04-16
Speed up SpanReader::findTraceIDs() for Elasticsearch

bobrik:
We have a service that generates two spans per trace, around 4K spans per second. It's very slow to query for the latest traces in this service.

This is the default query that is currently generated:

```json
{
  "aggregations": {
    "traceIDs": {
      "aggregations": {
        "startTime": {
          "max": {
            "field": "startTime"
          }
        }
      },
      "terms": {
        "field": "traceID",
        "order": [
          {
            "startTime": "desc"
          }
        ],
        "size": 20
      }
    }
  },
  "query": {
    "bool": {
      "must": [
        {
          "range": {
            "startTime": {
              "from": 1555437540000000,
              "include_lower": true,
              "include_upper": true,
              "to": 1555437600000000
            }
          }
        },
        {
          "match": {
            "process.serviceName": {
              "query": "nginx-ssl"
            }
          }
        }
      ]
    }
  },
  "size": 0
}
```

This takes pretty much forever if there are many spans to aggregate:

```json
{
  "took": 38372,
  "timed_out": false,
  "_shards": {
    "total": 5,
    "successful": 5,
    "skipped": 0,
    "failed": 0
  },
  "hits": {
    "total": 331761,
    "max_score": 0,
    "hits": []
  },
  "other": "stuff"
}
```

Presumably, because ES takes full window, then sorts as many buckets
as there are traces.

Instead, we can pull the last X spans, and find unique traceIDs there:

```json
{
  "query": {
    "bool": {
      "must": [
        {
          "range": {
            "startTime": {
              "from": 1555437540000000,
              "include_lower": true,
              "include_upper": true,
              "to": 1555437600000000
            }
          }
        },
        {
          "match": {
            "process.serviceName": {
              "query": "nginx-ssl"
            }
          }
        }
      ]
    }
  },
  "size": 100,
  "sort": [
    {
      "startTime": "desc"
    }
  ]
}
```

This is a whole lot faster:

```json
{
  "took": 26,
  "timed_out": false,
  "_shards": {
    "total": 5,
    "successful": 5,
    "skipped": 0,
    "failed": 0
  },
  "hits": {
    "total": 331761,
    "max_score": null,
    "hits": [
      "100 hits here"
    ]
  },
  "other": "stuff"
}
```

So this is 38372 / 26 = 1475x faster, which is pretty nice. Generally speedup is under 1000x, though.

### DIFF
--- a/pkg/es/client.go
+++ b/pkg/es/client.go
@@ -66,6 +66,7 @@ type SearchService interface {
 	Aggregation(name string, aggregation elastic.Aggregation) SearchService
 	IgnoreUnavailable(ignoreUnavailable bool) SearchService
 	Query(query elastic.Query) SearchService
+	Source(source interface{}) SearchService
 	Do(ctx context.Context) (*elastic.SearchResult, error)
 }
 
--- a/pkg/es/mocks/SearchService.go
+++ b/pkg/es/mocks/SearchService.go
@@ -14,7 +14,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-
 package mocks
 
 import (
@@ -93,6 +92,22 @@ func (_m *SearchService) Query(query elastic.Query) es.SearchService {
 	var r0 es.SearchService
 	if rf, ok := ret.Get(0).(func(elastic.Query) es.SearchService); ok {
 		r0 = rf(query)
+	} else {
+		if ret.Get(0) != nil {
+			r0 = ret.Get(0).(es.SearchService)
+		}
+	}
+
+	return r0
+}
+
+// Sort provides a mock function with given fields: source
+func (_m *SearchService) Source(source interface{}) es.SearchService {
+	ret := _m.Called(source)
+
+	var r0 es.SearchService
+	if rf, ok := ret.Get(0).(func(interface{}) es.SearchService); ok {
+		r0 = rf(source)
 	} else {
 		if ret.Get(0) != nil {
 			r0 = ret.Get(0).(es.SearchService)
--- a/pkg/es/wrapper/wrapper.go
+++ b/pkg/es/wrapper/wrapper.go
@@ -210,6 +210,11 @@ func (s SearchServiceWrapper) Query(query elastic.Query) es.SearchService {
 	return WrapESSearchService(s.searchService.Query(query))
 }
 
+// Source calls this function to internal service.
+func (s SearchServiceWrapper) Source(source interface{}) es.SearchService {
+	return WrapESSearchService(s.searchService.Source(source))
+}
+
 // Do calls this function to internal service.
 func (s SearchServiceWrapper) Do(ctx context.Context) (*elastic.SearchResult, error) {
 	return s.searchService.Do(ctx)
--- a/plugin/storage/es/spanstore/reader.go
+++ b/plugin/storage/es/spanstore/reader.go
@@ -96,6 +96,7 @@ type SpanReader struct {
 	// The age of the oldest service/operation we will look for. Because indices in ElasticSearch are by day,
 	// this will be rounded down to UTC 00:00 of that day.
 	maxSpanAge              time.Duration
+	maxNumSpans             int
 	serviceOperationStorage *ServiceOperationStorage
 	spanIndexPrefix         string
 	serviceIndexPrefix      string
@@ -125,6 +126,7 @@ func NewSpanReader(p SpanReaderParams) *SpanReader {
 		client:                  p.Client,
 		logger:                  p.Logger,
 		maxSpanAge:              p.MaxSpanAge,
+		maxNumSpans:             p.MaxNumSpans,
 		serviceOperationStorage: NewServiceOperationStorage(p.Client, p.Logger, 0), // the decorator takes care of metrics
 		spanIndexPrefix:         indexNames(p.IndexPrefix, spanIndex),
 		serviceIndexPrefix:      indexNames(p.IndexPrefix, serviceIndex),
@@ -377,6 +379,7 @@ func (s *SpanReader) multiRead(ctx context.Context, traceIDs []model.TraceID, st
 	for _, trace := range tracesMap {
 		traces = append(traces, trace)
 	}
+
 	return traces, nil
 }
 
@@ -418,7 +421,7 @@ func (s *SpanReader) findTraceIDs(ctx context.Context, traceQuery *spanstore.Tra
 	defer childSpan.Finish()
 	//  Below is the JSON body to our HTTP GET request to ElasticSearch. This function creates this.
 	// {
-	//      "size": 0,
+	//      "size": 10000,
 	//      "query": {
 	//        "bool": {
 	//          "must": [
@@ -467,46 +470,67 @@ func (s *SpanReader) findTraceIDs(ctx context.Context, traceQuery *spanstore.Tra
 	//          ]
 	//        }
 	//      },
-	//      "aggs": { "traceIDs" : { "terms" : {"size": 100,"field": "traceID" }}}
+	//      "track_total_hits": false
 	//  }
-	aggregation := s.buildTraceIDAggregation(traceQuery.NumTraces)
 	boolQuery := s.buildFindTraceIDsQuery(traceQuery)
 
 	jaegerIndices := s.timeRangeIndices(s.spanIndexPrefix, traceQuery.StartTimeMin, traceQuery.StartTimeMax)
 
+	// cap size at 10000 to keep elasticsearch happy
+	size := s.maxNumSpans * traceQuery.NumTraces
+	if size > 10000 {
+		size = 10000
+	}
+
+	source, err := elastic.NewSearchSource().
+		Query(boolQuery).
+		Size(size).
+		Sort("startTime", true).
+		TrackTotalHits(false).
+		Source()
+
+	if err != nil {
+		return nil, errors.Wrap(err, "Search source creation failed")
+	}
+
 	searchService := s.client.Search(jaegerIndices...).
-		Size(0). // set to 0 because we don't want actual documents.
-		Aggregation(traceIDAggregation, aggregation).
 		IgnoreUnavailable(true).
-		Query(boolQuery)
+		Source(source)
 
 	searchResult, err := searchService.Do(s.ctx)
 	if err != nil {
 		return nil, errors.Wrap(err, "Search service failed")
 	}
-	if searchResult.Aggregations == nil {
+
+	if searchResult.Hits == nil || searchResult.Hits.TotalHits == 0 {
 		return []string{}, nil
 	}
-	bucket, found := searchResult.Aggregations.Terms(traceIDAggregation)
-	if !found {
-		return nil, ErrUnableToFindTraceIDAggregation
+
+	traceIDs := map[string]struct{}{}
+
+	var aux struct {
+		TraceID string `json:"traceID"`
 	}
 
-	traceIDBuckets := bucket.Buckets
-	return bucketToStringArray(traceIDBuckets)
-}
+	for _, hit := range searchResult.Hits.Hits {
+		if err = json.Unmarshal(*hit.Source, &aux); err != nil {
+			return nil, errors.Wrap(err, "elasticsearch result is invalid")
+		}
 
-func (s *SpanReader) buildTraceIDAggregation(numOfTraces int) elastic.Aggregation {
-	return elastic.NewTermsAggregation().
-		Size(numOfTraces).
-		Field(traceIDField).
-		Order(startTimeField, false).
-		SubAggregation(startTimeField, s.buildTraceIDSubAggregation())
-}
+		traceIDs[aux.TraceID] = struct{}{}
+	}
 
-func (s *SpanReader) buildTraceIDSubAggregation() elastic.Aggregation {
-	return elastic.NewMaxAggregation().
-		Field(startTimeField)
+	uniqueTraceIDs := []string{}
+
+	for traceID := range traceIDs {
+		uniqueTraceIDs = append(uniqueTraceIDs, traceID)
+	}
+
+	if len(uniqueTraceIDs) > traceQuery.NumTraces {
+		return uniqueTraceIDs[0:traceQuery.NumTraces], nil
+	}
+
+	return uniqueTraceIDs, nil
 }
 
 func (s *SpanReader) buildFindTraceIDsQuery(traceQuery *spanstore.TraceQueryParameters) elastic.Query {
--- a/plugin/storage/es/spanstore/reader_test.go
+++ b/plugin/storage/es/spanstore/reader_test.go
@@ -530,8 +530,6 @@ func returnSearchFunc(typ string, r *spanReaderTest) ([]string, error) {
 		return r.reader.GetServices(context.Background())
 	} else if typ == operationsAggregation {
 		return r.reader.GetOperations(context.Background(), "someService")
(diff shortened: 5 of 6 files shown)


bobrik:
Ha, this issue is #1475 just like speedup is 1475x 😦

vprithvi:
Thanks for looking into this! It could be pretty sweet!

I'm curious about what would happen if there are only few traces with a large number of spans per service. 

For e.g., let's assume that there are 10 unique traces with 10k spans all for the same service - if the new query is executed with a size of 10, would it retrieve all 10 traceIDs?

bobrik:
It will only return 10 unique traceIDs across the latest 1000 (value of `es.max-num-spans`) spans that match the query, which may as well be just one traceID. I'd personally take this tradeoff against waiting for tens of seconds for the query to finish.

I'm looking into test failures.

vprithvi:
I understand your use case, but I'm not a fan of trading completeness for speed. 

I'm wondering whether we can trigger the strategy of searching by spans and looking at traceIDs only for certain search patterns. For instance, only applying it if the search interval is less than a minute.

Or even better, are there strategies for search that are both complete and quick?

bobrik:
> I understand your use case, but I'm not a fan of trading completeness for speed.

The tradeoff is "working ui" vs "not working ui" for me, unfortunately.

> I'm wondering whether we can trigger the strategy of searching by spans and looking at traceIDs only for certain search patterns. For instance, only applying it if the search interval is less than a minute.

The larger the window, the cheaper it is to search by spans. Aggregation looks at all matched docs, while search only looks at the latest ones.

> Or even better, are there strategies for search that are both complete and quick?

I created a topic in elasticsearch forum to see if there's a proper way to do this:

* https://discuss.elastic.co/t/fast-way-to-find-recent-unique-fields/177351

ledor473:
@bobrik I tried to replicate that slow query in our environment but it ran pretty fast:
```json
{
    "took": 177,
    "timed_out": false,
    "_shards": {
        "total": 5,
        "successful": 5,
        "skipped": 0,
        "failed": 0
    },
    "hits": {
        "total": 1087815,
        "max_score": 0,
        "hits": []
    },
    "aggregations": {
        "traceIDs": {
            "doc_count_error_upper_bound": -1,
            "sum_other_doc_count": 1087569,
            "buckets": [.....]
        }
    }
}
```

I tried to pick something that would match the `hits.total` value in your issue (`331 761`) but ended finding something that is 3x what is mentioned (`1 087 815`). 

Obviously, this is not a perfect comparison, but I still wanted to try to replicate it and report back 👍

NijiadeIX:
> @bobrik I tried to replicate that slow query in our environment but it ran pretty fast:
> 
> ```json
> {
>     "took": 177,
>     "timed_out": false,
>     "_shards": {
>         "total": 5,
>         "successful": 5,
>         "skipped": 0,
>         "failed": 0
>     },
>     "hits": {
>         "total": 1087815,
>         "max_score": 0,
>         "hits": []
>     },
>     "aggregations": {
>         "traceIDs": {
>             "doc_count_error_upper_bound": -1,
>             "sum_other_doc_count": 1087569,
>             "buckets": [.....]
>         }
>     }
> }
> ```
> 
> I tried to pick something that would match the `hits.total` value in your issue (`331 761`) but ended finding something that is 3x what is mentioned (`1 087 815`).
> 
> Obviously, this is not a perfect comparison, but I still wanted to try to replicate it and report back 👍

May I know something detail  about the config of your environment?

ledor473:
@NijiadeIX I've shared number in a talk. You can get the slides from that page: https://kccnceu19.sched.com/event/MPbm/deploy-scale-and-extend-jaeger-louis-etienne-dorval-ticketmaster

Look at slide 14 and 37

bobrik:
@NijiadeIX this is how much data we have on daily basis (33B spans / 2TB of data per replica):

```
$ curl 'https://foo.bar/_cat/indices/jaeger-span-*?v&s=index&h=index,pri,docs.count,store.size'
index                  pri  docs.count store.size
jaeger-span-2019-09-11  38 26600483894      3.2tb
jaeger-span-2019-09-12  38 33379018750        4tb
jaeger-span-2019-09-13  38 33201947406        4tb
jaeger-span-2019-09-14  38 31263193630      3.8tb
jaeger-span-2019-09-15  38 31037438456      3.7tb
jaeger-span-2019-09-16  38 32592635484      3.9tb
jaeger-span-2019-09-17  38 32561140996      3.9tb
jaeger-span-2019-09-18  38 28266696032      3.5tb
```

wqtty:
We have also experienced very slow query performance, we altered the index setting, added 
`"eager_global_ordinals": true,` to  traceID field, it looks like below

`
"traceID": {
          "type": "keyword",
          "eager_global_ordinals": true,
          "ignore_above": 256
        }`

pavolloffay:
@wqtty how much data do you ingest? What is your data retention? 

With that setting do you experience increased disk, CPU overhead at ingestion time? 

> Note: this will, however, be shifting the cost of building global ordinals from search-time to refresh-time. For many use cases, this is a price that’s acceptable to pay for the improvements at search time.
https://medium.com/driven-by-code/elasticsearch-global-ordinals-31df2806391f

wqtty:
> @wqtty how much data do you ingest? What is your data retention?
> 
> With that setting do you experience increased disk, CPU overhead at ingestion time?
> 
> > Note: this will, however, be shifting the cost of building global ordinals from search-time to refresh-time. For many use cases, this is a price that’s acceptable to pay for the improvements at search time.
> > https://medium.com/driven-by-code/elasticsearch-global-ordinals-31df2806391f

we are having about 14k spans per second, and keep them in es for only 2 days, meaning we are having a jaeger-span index about 2.3T every day.

> With that setting do you experience increased disk, CPU overhead at ingestion time?
Sorry, I didn't catch you.

pavolloffay:
If you experience increased CPU load or increased disk space allocation after you applied that setting.

wqtty:
> If you experience increased CPU load or increased disk space allocation after you applied that setting.

not very obvious

Stono:
Our elasticsearch/jaeger setup grinds to a halt as the day goes on!  Being able to control the max number of spans in the query result as a config option feels like something many folks (including me) would opt into.

xiedaijin1018:
I query the data of today can be very slow:
```
{
  "took" : 6864,
  "timed_out" : false,
  "_shards" : {
    "total" : 5,
    "successful" : 5,
    "skipped" : 0,
    "failed" : 0
  },
  "hits" : {
    "total" : 95482,
    "max_score" : 0.0,
    "hits" : [ ]
  },
  "aggregations" : {
    "traceIDs" : {
      "doc_count_error_upper_bound" : -1,
      "sum_other_doc_count" : 95342,
      "buckets" : [...]
    }
  }
}
```

But I query the data of the yesterday can be very fast:
```
{
  "took" : 40,
  "timed_out" : false,
  "_shards" : {
    "total" : 5,
    "successful" : 5,
    "skipped" : 0,
    "failed" : 0
  },
  "hits" : {
    "total" : 98014,
    "max_score" : 0.0,
    "hits" : [ ]
  },
  "aggregations" : {
    "traceIDs" : {
      "doc_count_error_upper_bound" : -1,
      "sum_other_doc_count" : 97875,
      "buckets" : [...]
    }
  }
}
```

pavolloffay:
Could somebody submit a new PR or rebase this one?

Stono:
FYI re my previous comment, we've pretty much solved our performance problems by adding `"eager_global_ordinals": true` to the `tradeId` field, we currently ingest about 3.3billion spans a day and find performance pretty good

RashmiRam:
I suppose we rely on aggregations to get unique traceIds. Is there any way to distinguish the root span alone so that we could filter based on certain field to get all root spans and their traceIDs? I saw `references` will be empty array for  root spans. But again, it is a nested field. We could rely on `parentSpanId` field. But, It saw it is deprecated. Is there any possibility that we could add a new field or leverage existing field to find out root spans alone so that we can avoid this aggregation totally?

bobrik:
@pavolloffay, I can rebase, but it will still fail a test for reading past 10k spans and I don't really have much time on my hands to work around that.

Internally we've moved on from this a bit further, which complicates things.

Our index is [sorted](https://www.elastic.co/guide/en/elasticsearch/reference/master/index-modules-index-sorting.html), so that queries can read data in order and terminate early. This required us to get rid of indexing for nested docs, which means no searching for tags in logs.

Overall there's a massive improvement in terms of docs (we used to run out of docs per segment because of nested docs!), and memory footprint:


```
index                  pri  docs.count store.size segments.memory segments.fixed_bitset_memory fielddata.memory_size memory.total
# Before (cluster one)
jaeger-span-2020-05-20  38 81604373583      8.9tb          17.1gb                         38gb                    0b       17.3gb
# After (cluster two)
jaeger-span-2020-05-20  38  8406585575      4.8tb         192.1mb                           0b                    0b      192.1mb
```

Plus we're looking into Clickhouse as the backend #1438.

bobrik:
@RashmiRam, the aggregation is looking for traceIDs that match the tags. You can't look only at root spans, since they don't have tags of children.

RashmiRam:
Aah. I somehow missed it. Thanks @bobrik

sivatarunp:
@Stono Can I know what is your refresh interval and jvm size. I have tried using `eager_global_ordinals` but found no improvement after that, And our rate is very less when compared to yours.

mehta-ankit:
@pavolloffay @wqtty @bobrik 
I wanted to share big improvements we saw after adding `"eager_global_ordinals": true` to `traceID` field as mentioned in this comment.
https://github.com/jaegertracing/jaeger/pull/1475#issuecomment-566479808
Along with this change we also modified `refresh_interval` for the indices to `60s` (default is 1 s for Elasticsearch).

Improvements we saw were following:
1) For non search endpoints (all endpoints except "/api/traces" that jaeger query makes using jaeger_rpc_request_latency_bucket metric) we saw 65-75% improvement for 90th percentile of latencies.
2) Same metric value for search endpoint (/api/traces) we saw 85% improvement for average latencies. (since highest bucket for this metric is 10s, earlier we could not even plot 90th percentile as it would flat line at 10s value. Hence we used average's 😄 )

Since setting `eager_global_ordinals` to true moves the cost of building global ordinals to refresh phase from search refresh interval helped with overall stability i believe.

We saw heap usage increase by 7-9% after this change.

Our daily index's primary store size is 200-300GB (with default settings of 5 primary shards and 1 replica). We retain data for 7 days and on average we keep 1.5-2TB of data in elasticsearch.

Let me know if you all have any questions.

sivatarunp:
> @pavolloffay @wqtty @bobrik
> I wanted to share big improvements we saw after adding `"eager_global_ordinals": true` to `traceID` field as mentioned in this comment.
> [#1475 (comment)](https://github.com/jaegertracing/jaeger/pull/1475#issuecomment-566479808)
> Along with this change we also modified `refresh_interval` for the indices to `60s` (default is 1 s for Elasticsearch).
> 
> Improvements we saw were following:
> 
> 1. For non search endpoints (all endpoints except "/api/traces" that jaeger query makes using jaeger_rpc_request_latency_bucket metric) we saw 65-75% improvement for 90th percentile of latencies.
> 2. Same metric value for search endpoint (/api/traces) we saw 85% improvement for average latencies. (since highest bucket for this metric is 10s, earlier we could not even plot 90th percentile as it would flat line at 10s value. Hence we used average's 😄 )
> 
> Since setting `eager_global_ordinals` to true moves the cost of building global ordinals to refresh phase from search refresh interval helped with overall stability i believe.
> 
> We saw heap usage increase by 7-9% after this change.
> 
> Our daily index's primary store size is 200-300GB (with default settings of 5 primary shards and 1 replica). We retain data for 7 days and on average we keep 1.5-2TB of data in elasticsearch.
> 
> Let me know if you all have any questions.

@mehta-ankit can we know your cardinality of traceID field, and what is the value of latency you have observed. For us setting the global_ordinals to true didn't help . Our refresh interval is also around 60s.

mehta-ankit:
@sivatarunp I don't have the value for cardinality of traceID field. I am assuming i ll have to use cardinality aggregation to figure this out.

Avg Latencies for `/api/traces` endpoint have been around 500ms-1s in last 1 hour (its still early for us to hit high enough traffic on search right now). Of course it is dependent on params passed around from jaeger query (results limit, lookback value etc).

Question, 

1. how did you set this value ? Did you set it on all your daily indices ?
We use a cronjob to set this value on each daily index when its created. Otherwise values like refresh interval and this one can be overwritten(https://github.com/jaegertracing/jaeger-operator/issues/1077#issuecomment-639488747)

sivatarunp:
@mehta-ankit  you can have a rough estimation of cardinality by traces/s metric as well. For us for lookback value of 1hr and results_limit to 20, Its taking huge time around 25-30s. Hence the worry
 This is on an index which has around 40million queryable docs in the time range.
As far as to answer your question. We use es-aliases. We have the setting update in the jaeger-span-template. So each rollover will have the index created with this setting enabled

sivatarunp:
> @sivatarunp ~I don't have the value for cardinality of traceID field. I am assuming i ll have to use cardinality aggregation to figure this out.~
> 
> Looks like i figured out the cardinality. It seems to be 105 million for a single day's index.
> 
> ```
> "aggregations" : {
>     "traceid_count" : {
>       "value" : 105056553
>     }
>   }
> ```
> 
> Used the following query
> 
> ```
> curl -H "Content-Type: application/json" -X GET "http://localhost:9200/<index_name>/_search?size=0&pretty" -d'{     
>    "aggs": {
>          "traceid_count": {
>              "cardinality": {
>                  "field": "traceID"
>              }
>          }
>      }
> }'
> ```
> 
> Avg Latencies for `/api/traces` endpoint have been around 500ms-1s in last 1 hour (its still early for us to hit high enough traffic on search right now). Of course it is dependent on params passed around from jaeger query (results limit, lookback value etc).
> 
> Question,
> 
> 1. how did you set this value ? Did you set it on all your daily indices ?
>    We use a cronjob to set this value on each daily index when its created. Otherwise values like refresh interval and this one can be overwritten([jaegertracing/jaeger-operator#1077 (comment)](https://github.com/jaegertracing/jaeger-operator/issues/1077#issuecomment-639488747))
Thanks for the info @mehta-ankit 👍 🙏 
Our cardinality is around 130million per day. We have observed the high response times (25-30s) in cases where the querying service has high docs (40m)per shard. This was our test setup, where we ingested more docs of a single service to observe the pattern.  We observed better response times as the no:of docs/shard is reduced. Would you mind sharing few details like the AWS instance types which you were in use for the current setup

jpkrohling:
Closing due to inactivity.

Stono:
Hey for other people coming to this thread, i've summarised all the things I changed to make jaeger -> elasticsearch super performant _for our use case_.  Hope it might be of use to others. https://karlstoney.com/speeding-up-jaeger-on-elasticsearch

yurishkuro:
@Stono great post! You didn't always mention the downsides of enabling various settings, so it's a bit hard to judge, but wouldn't some of these changes make sense permanently in the default index template created by Jaeger? Or do you feel that it's hardly ever a good idea to rely on the default template?

Stono:
That's a really good point to be fair, I'll update it with the negatives tomorrow (which tbh are negligible for us, slightly more CPU during ingest).

I think it's hard for jaeger offer a best for all template our the box.  The default one is pretty good tbh, but I was pushing the scale.  When you start doing that sort of volume really I think you'll end up bespoke tweaking to your use case, and hardware.

Issue #1373, opened 2019-02-22
Gocql may panic when writing with RF=2, CL=LOCAL_ONE


vprithvi:
When writing, we sometimes see panics from gocql with the following stack trace:

```
panic: first replica is not the primary replica for the token

goroutine 97267 [running]:
infra/jaeger/vendor/github.com/gocql/gocql.(*networkTopology).replicaMap(0xc00d162028, 0xc0004a8a00, 0x96, 0x96, 0xc005682000, 0x9600, 0x9d55, 0x0)
	infra/jaeger/vendor/github.com/gocql/gocql/topology.go:201 +0x131e
infra/jaeger/vendor/github.com/gocql/gocql.(*tokenAwareHostPolicy).KeyspaceChanged(0xc0010d6100, 0xc00d7ce5c0, 0xe, 0xc00d7ce5b0, 0x7)
	infra/jaeger/vendor/github.com/gocql/gocql/policies.go:435 +0x3c1
infra/jaeger/vendor/github.com/gocql/gocql.(*Session).handleKeyspaceChange(0xc0010e4000, 0xc00d7ce5c0, 0xe, 0xc00d7ce5b0, 0x7)
	infra/jaeger/vendor/github.com/gocql/gocql/events.go:129 +0x77
infra/jaeger/vendor/github.com/gocql/gocql.(*Session).handleSchemaEvent(0xc0010e4000, 0xc008604710, 0x1, 0x1)
	infra/jaeger/vendor/github.com/gocql/gocql/events.go:114 +0x2f3
infra/jaeger/vendor/github.com/gocql/gocql.(*Session).handleSchemaEvent-fm(0xc008604710, 0x1, 0x1)
	infra/jaeger/vendor/github.com/gocql/gocql/session.go:120 +0x48
created by infra/jaeger/vendor/github.com/gocql/gocql.(*eventDebouncer).flush
	infra/jaeger/vendor/github.com/gocql/gocql/events.go:64 +0x5f
```

This happens quite infrequently, even when there is nothing of interest (nodes restarting/leaving/joining) happening on the Cassandra cluster.

vprithvi:
Perhaps the latest gocql fixes this? See https://github.com/gocql/gocql/pull/1261 for context

guanw:
The log error complains there's keyspace changes. This is due to we were adding/updating/deleting keyspaces in another data center. Should we let gocql folks know this?

PR #1480, opened 2019-04-19
add option to disable schema change events for gocql

guanw:
- This change add the option to enable/disable schema events so #1373 could be avoided.

- Sometimes gocql will panic which cause components using Cassandra to crash if schema change events (e.g Removing Keyspace in a multi-dc jaeger setup), We can disable this attribute to avoid the catastrophic crash.

### DIFF
--- a/pkg/cassandra/config/config.go
+++ b/pkg/cassandra/config/config.go
@@ -133,12 +133,7 @@ func (c *Configuration) NewCluster() *gocql.ClusterConfig {
 		cluster.Consistency = gocql.ParseConsistency(c.Consistency)
 	}
 
-	fallbackHostSelectionPolicy := gocql.RoundRobinHostPolicy()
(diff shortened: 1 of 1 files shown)


vprithvi:
relates to https://github.com/gocql/gocql/issues/1274

I'm not sure about the implications of disabling this - How does this impact normal operations when there are no schema changes?

guanw:
> relates to [gocql/gocql#1274](https://github.com/gocql/gocql/issues/1274)
> 
> I'm not sure about the implications of disabling this - How does this impact normal operations when there are no schema changes?

hmmm. Good question, I think the keyspace change will trigger `KeyspaceChanged` in gocql/policies.go and not update the metadata map. Should we reach out to gocql to understand exactly what disabling the events gets us?

The default value should be false which still enables the events so current codebase won't be affected.

PR #1493, opened 2019-04-24
Add span format tag for jaeger-collector

### LINKED ISSUE #1490: Add span format tag for each span

We want to make sure our internal users of jaeger always use the latest jaeger format (thrift-"jaeger", grpc-"proto")to emit spans.

We have flink job to analysis all the spans coming into kafka, and one of the analysis could be SpanFormatCheck, if jaeger collector always embed a span format check, we can use that to tell users that your format is legacy or not.

Currently jaeger span does not contain information regarding what kind of format the span is: zipkin or jaeger.

Having jaeger collector embed internal.span.format tag with value to be "zipkin"/"jaeger"/"proto" when converting those format to domain object.

guo0693:
Signed-off-by: jung <jung@uber.com>

-  Resolves #1490 
- include span format tag for every processed span in the collector


guo0693:
Screen Shot 2019-04-24 at 16 58 40

Screenshot of running hotRod locally with the changes.

Issue #1497, opened 2019-04-25
Allow custom load balancing for gRPC [agent] 


guanw:
## Requirement - what kind of business use case are you trying to solve?

Allow external discovery service to integrate with grpc by creating a grpc resolver and using notifier.

## Problem - what in Jaeger blocks you from solving the requirement?

None

## Proposal - what do you suggest to solve the problem or improve the existing situation?

Add a resolver component that implements resolver.Resolver interface and integrate with discovery.notifier interface.

## Any open questions to address

None

PR #1498, opened 2019-04-25
Add grpc_resolver using external discovery service

guanw:

- #1497 

- Add grpc resolver which listens to available host/port from discovery.notifier and updates list of collector servers for agent to connect.


cmd/agent/app/reporter/grpc/builder.go
```diff
@@ -39,8 +40,8 @@ type ConnBuilder struct {
 	TLSCA         string
 	TLSServerName string
 
-	notifier discovery.Notifier
-	resolver resolver.Builder
+	subsetSize int
```
yurishkuro: subsetSize name is non-descriptive

cmd/agent/app/reporter/grpc/builder.go
```diff
@@ -39,8 +40,8 @@ type ConnBuilder struct {
...
+	notifier   discovery.Notifier
```
yurishkuro: why not make these two fields public? And drop WithDiscoveryNotifier() function, it doesn't add anything.

cmd/agent/app/reporter/grpc/builder.go
```diff
@@ -83,7 +84,11 @@ func (b *ConnBuilder) CreateConnection(logger *zap.Logger) (*grpc.ClientConn, er
 	}
 
 	if b.notifier != nil {
-		return nil, errors.New("not implemented")
+		logger.Info("Using external discovery service with roundrobin load balancer")
+		grpcResolver := grpcresolver.New(b.notifier, logger, b.subsetSize)
+		resolver.Register(grpcResolver)
```
yurishkuro: is this setting something global? Can we do it without affecting global state?

cmd/agent/app/reporter/grpc/flags.go
```diff
@@ -22,13 +22,14 @@ import (
...
-	collectorTLSCA         = gRPCPrefix + "tls.ca"
-	collectorTLSServerName = gRPCPrefix + "tls.server-name"
+	gRPCPrefix                   = "reporter.grpc."
+	collectorHostPort            = gRPCPrefix + "host-port"
+	retry                        = gRPCPrefix + "retry.max"
+	defaultMaxRetry              = 3
+	collectorTLS                 = gRPCPrefix + "tls"
+	collectorTLSCA               = gRPCPrefix + "tls.ca"
+	collectorTLSServerName       = gRPCPrefix + "tls.server-name"
+	collectorConnectionsPerAgent = gRPCPrefix + "connections.per.host"
```
yurishkuro: please use naming consistent with existing tchannel options
```
```

pkg/discovery/grpcresolver/grpc_resolver.go
```diff
@@ -0,0 +1,126 @@
...
+	salt       string
+}
+
+// New intialize a new grpc resolver with notifier
+func New(
+	notifier discovery.Notifier,
+	logger *zap.Logger,
+	subsetSize int,
+) *Resolver {
+	rand.Seed(time.Now().UTC().UnixNano())
```
yurishkuro: don't use global state, create your own rand instance

pkg/discovery/grpcresolver/grpc_resolver.go
```diff
@@ -0,0 +1,126 @@
...
+
+	r := &Resolver{
+		notifier:   notifier,
+		discoCh:    make(chan []string, 100), // TODO should this number be configurable? What if the number of collectors exceed 100 from discovery service
```
yurishkuro: the messages on this change are a full set of instances in a single slice, so the number of collectors has no relation to the depth of this queue. Let's keep 100 for now.

pkg/discovery/grpcresolver/grpc_resolver.go
```diff
@@ -0,0 +1,126 @@
...
+		stopCh:     make(chan struct{}),
+		logger:     logger,
+		subsetSize: subsetSize,
+		salt:       strconv.FormatInt(rand.Int63(), 10),
+		scheme:     strconv.FormatInt(time.Now().UnixNano(), 36),
+	}
+	// TODO not sure if there's an equivalent way for grpc to maintain connection like what tchannel did?
+
+	// Register the resolver with grpc so it's available for grpc.Dial
+	resolver.Register(r)
```
yurishkuro: looks like another global state

pkg/discovery/grpcresolver/grpc_resolver.go
```diff
@@ -0,0 +1,126 @@
...
+	return r, nil
+}
+
+// Scheme returns the test scheme.
+func (r *Resolver) Scheme() string {
+	return r.scheme
+}
+
+// ResolveNow is a noop for Resolver.
+func (*Resolver) ResolveNow(o resolver.ResolveNowOption) {}
```
yurishkuro: if you accept Discoverer then you can implement this too. That's how it was done for tchannel.

If you don't implement this method, is there going to be some hiccup on startup where the grpc connections waits for an update from uns instead of querying it proactively?

pkg/discovery/grpcresolver/grpc_resolver.go
```diff
@@ -0,0 +1,126 @@
...
+	// Register the resolver with grpc so it's available for grpc.Dial
+	resolver.Register(r)
+
+	// Register the discoCh channel with notifier so it continues to fetch a list of host/port
+	notifier.Register(r.discoCh)
+	return r
+}
+
+// Build returns itself for Resolver, because it's both a builder and a resolver.
+func (r *Resolver) Build(target resolver.Target, cc resolver.ClientConn, opts resolver.BuildOption) (resolver.Resolver, error) {
```
yurishkuro: is there a need to separate New and Build into separate functions?

pkg/discovery/grpcresolver/grpc_resolver.go
```diff
@@ -0,0 +1,126 @@
...
+// ResolveNow is a noop for Resolver.
+func (*Resolver) ResolveNow(o resolver.ResolveNowOption) {}
+
+func (r *Resolver) watcher() {
+	for {
+		select {
+		case latestHostPorts := <-r.discoCh:
+			r.mu.Lock()
+			defer r.mu.Unlock()
+			r.logger.Info("gRPC naming.Watcher Receives updates", zap.Strings("hostPorts", latestHostPorts))
```
yurishkuro: Receives -> received

pkg/discovery/grpcresolver/grpc_resolver.go
```diff
@@ -0,0 +1,126 @@
...
+
+// Close closes both discoCh and stopCh
+func (r *Resolver) Close() {
+	if r.discoCh != nil {
+		r.notifier.Unregister(r.discoCh)
+		close(r.discoCh)
+		r.discoCh = nil
+	}
+
+	if r.stopCh != nil {
```
yurishkuro: I'm always skeptical about having these nil checks - are you trying to make Close() reentrant? What's the use case for that?

pkg/discovery/grpcresolver/grpc_resolver.go
```diff
@@ -0,0 +1,126 @@
...
+
+package grpcresolver
+
+import (
+	"math/rand"
+	"strconv"
+	"sync"
+	"time"
+
+	"github.com/tysontate/rendezvous"
```
yurishkuro: do we need this dependency? I think I described the algorithm on the previous ticket, it just needs hash and sort from stdlib.

cmd/agent/app/reporter/grpc/builder.go
guanw: it's actually already set within `grpcresolver.New`, I will remove it here in followup commit.

pkg/discovery/grpcresolver/grpc_resolver.go
guanw: Build will be fired by grpc-go `resolver_conn_wrapper.go` and `resolver.ClientConn` will be passed from there which will be used to keep track of host addresses.

pkg/discovery/grpcresolver/grpc_resolver.go
guanw: I haven't really tested what you said but it could be right. Though I'm not sure when this method is fired (i.e is it fired after `resolver.ClientConn` is set so that we can update connection state).

cmd/agent/app/reporter/grpc/builder.go
guanw: sg. Was just following our pattern with tchannel builder.

pkg/discovery/grpcresolver/grpc_resolver.go
guanw: yeah essentially the package is doing the same thing. Will replace it in next commit.

guanw:
saw this error twice with travis
image

Local build "make test-ci" works fine. Is there a way to configure memory limit on travis?

cmd/agent/app/reporter/grpc/flags.go
```diff
@@ -22,13 +22,14 @@ import (
...
-	collectorTLSCA         = gRPCPrefix + "tls.ca"
-	collectorTLSServerName = gRPCPrefix + "tls.server-name"
+	gRPCPrefix                   = "reporter.grpc."
+	collectorHostPort            = gRPCPrefix + "host-port"
+	retry                        = gRPCPrefix + "retry.max"
+	defaultMaxRetry              = 3
+	collectorTLS                 = gRPCPrefix + "tls"
+	collectorTLSCA               = gRPCPrefix + "tls.ca"
+	collectorTLSServerName       = gRPCPrefix + "tls.server-name"
+	collectorConnectionsPerAgent = gRPCPrefix + "connections-per-host"
```
yurishkuro: where is "connections-per-host" name coming from? It's not what tchannel uses, we need to be consistent

pkg/discovery/grpcresolver/grpc_resolver.go
yurishkuro: because it makes it difficult to tests in unit tests.

cmd/agent/app/reporter/grpc/flags.go
guanw: Updated

pkg/discovery/grpcresolver/grpc_resolver.go
yurishkuro: You need global because you're registering your new resolver with round-robin balancer _by name_:
```go
	grpcResolver := grpcresolver.New(b.Notifier, b.Discoverer, logger, b.DiscoveryMinPeers)
	dialOptions = append(dialOptions, grpc.WithBalancerName(roundrobin.Name))
	dialTarget = grpcResolver.Scheme() + ":///round_robin"
```

Couldn't you just pass it to RRLB?

```go
	grpcResolver := grpcresolver.New(b.Notifier, b.Discoverer, logger, b.DiscoveryMinPeers)
	balancer  := grpc.RoundRobin(grpcResolver)
	dialOptions = append(dialOptions, grpc.WithBalancer(balancer))
```

pkg/discovery/grpcresolver/grpc_resolver.go
```diff
@@ -0,0 +1,155 @@
...
+
+// ResolveNow is a noop for Resolver.
+func (r *Resolver) ResolveNow(o resolver.ResolveNowOption) {}
+
+func (r *Resolver) watcher() {
+	defer r.wg.Done()
+	for latestHostPorts := range r.discoCh {
+		r.mu.Lock()
+		r.logger.Info("Received updates from notifier", zap.Strings("hostPorts", latestHostPorts))
+		r.cc.UpdateState(resolver.State{Addresses: generateAddresses(rendezvousHash(latestHostPorts, r.salt, r.hasher, r.discoveryMinPeers))})
```
yurishkuro: - pull generateAddresses() call into  a variable
- if you make it a member function, you don't need to pass parameters that are already accessible from `r.`

pkg/discovery/grpcresolver/grpc_resolver.go
```diff
@@ -0,0 +1,155 @@
...
+func (r *Resolver) Build(target resolver.Target, cc resolver.ClientConn, opts resolver.BuildOption) (resolver.Resolver, error) {
+	r.cc = cc
+
+	// Update conn states if proactively updates already work
+	instances, err := r.discoverer.Instances()
+	if err != nil {
+		return nil, err
+	}
+	r.cc.UpdateState(resolver.State{Addresses: generateAddresses(instances)})
+	r.wg.Add(1)
```
yurishkuro: what is this for?

pkg/discovery/grpcresolver/grpc_resolver.go
```diff
@@ -0,0 +1,155 @@
...
+	cc                resolver.ClientConn
+	notifier          discovery.Notifier
+	discoverer        discovery.Discoverer
+	logger            *zap.Logger
+	discoCh           chan []string // used to receive notifications
+	discoveryMinPeers int
+	mu                sync.Mutex
+	salt              []byte
+	hasher            hash.Hash32
+	wg                sync.WaitGroup
```
yurishkuro: document what this is used for

pkg/discovery/grpcresolver/grpc_resolver.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+	sort.Sort(hosts)
+	addressesPerHost := make([]string, discoveryMinPeers)
+	for i := 0; i < discoveryMinPeers; i++ {
+		addressesPerHost[i] = hosts[i].address
+	}
+	return addressesPerHost
+}
+
+func hashAddr(hasher hash.Hash32, node, key []byte) uint32 {
```
yurishkuro: what is the meaning of a key?

pkg/discovery/grpcresolver/grpc_resolver.go
guanw: Regarding using the `grpc.Roundrobin` and `grpc.WithBalancer`, I believe those were deprecated. 

image

image

pkg/discovery/grpcresolver/grpc_resolver.go
guanw: waitGroup is to make sure we are closing the resolver after we've exited the `watcher` goroutine.

pkg/discovery/grpcresolver/grpc_resolver.go
guanw: the saltKey, will update in next commit.

cmd/agent/app/reporter/grpc/flags.go
```diff
@@ -38,6 +39,7 @@ func AddFlags(flags *flag.FlagSet) {
 	flags.Bool(collectorTLS, false, "Enable TLS.")
 	flags.String(collectorTLSCA, "", "Path to a TLS CA file. (default use the systems truststore)")
 	flags.String(collectorTLSServerName, "", "Override the TLS server name.")
+	flags.Int(discoveryMinPeers, 3, "Override number of connections to collectors per agent")
```
yurishkuro: Max number of collectors to which the agent will try to connect at any given time

pkg/discovery/grpcresolver/grpc_resolver.go
yurishkuro: Per https://github.com/jaegertracing/jaeger/pull/1498#discussion_r279375929 looks like we don't have much choice, grpc is forcing the use of global state.

pkg/discovery/grpcresolver/grpc_resolver.go
yurishkuro: I don't follow. What would happen if you remove this wg?

pkg/discovery/grpcresolver/grpc_resolver.go
```diff
@@ -0,0 +1,158 @@
...
+	go r.watcher()
+	return r, nil
+}
+
+// Scheme returns the test scheme.
+func (r *Resolver) Scheme() string {
+	return r.scheme
+}
+
+// ResolveNow is a noop for Resolver.
```
yurishkuro: explain why. As I understand, our Notifier may have the semantics (confirm?) that it ALWAYS gives a callback at least once upon registering a listener channel. Or is it because we're using Discoverer to init the addresses?

Anyway, this needs documentation/explanation.

pkg/discovery/grpcresolver/grpc_resolver.go
```diff
@@ -0,0 +1,158 @@
...
+func (r *Resolver) ResolveNow(o resolver.ResolveNowOption) {}
+
+func (r *Resolver) watcher() {
+	defer r.wg.Done()
+	for latestHostPorts := range r.discoCh {
+		r.mu.Lock()
```
yurishkuro: what are you protecting against? Doesn't look like this function mutates any shared state.

pkg/discovery/grpcresolver/grpc_resolver.go
```diff
@@ -0,0 +1,158 @@
...
+		r.logger.Info("Received updates from notifier", zap.Strings("hostPorts", latestHostPorts))
+		updatedAddresses := generateAddresses(r.rendezvousHash(latestHostPorts))
```
yurishkuro: in L97 you call generateAddresses() without rendezvousHash, is that expected? Could this be encapsulated into
```go
func (r *Resolver) updateAddresses(hostPorts []string) {
    topN  := r.rendezvousHash(hostPorts)
    addresses := generateAddresses(topN)
    r.cc.UpdateState(resolver.State{Addresses: addresses})
}
```

pkg/discovery/grpcresolver/grpc_resolver.go
```diff
@@ -0,0 +1,158 @@
...
+	sort.Sort(hosts)
+	addressesPerHost := make([]string, r.discoveryMinPeers)
+	for i := 0; i < r.discoveryMinPeers; i++ {
+		addressesPerHost[i] = hosts[i].address
+	}
+	return addressesPerHost
+}
+
+func hashAddr(hasher hash.Hash32, node, saltKey []byte) uint32 {
+	hasher.Reset()
```
yurishkuro: Looks like hasher is your only shared state, however  I am not sure it's worth pre-allocating, since `rendezvousHash()` already allocates other arrays, and presumably discovery changes do not happen that often to  worry about gc pressure from that function. So I'd suggest creating new hash every time in `rendezvousHash()` (reusing it across addresses), and doing away with the mutex.

pkg/discovery/grpcresolver/grpc_resolver_test.go
```diff
@@ -0,0 +1,161 @@
...
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"go.uber.org/zap"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/balancer/roundrobin"
+	"google.golang.org/grpc/peer"
+	"google.golang.org/grpc/resolver"
+	testpb "google.golang.org/grpc/test/grpc_testing"
```
yurishkuro: nit:  testpb -> grpctest

pkg/discovery/grpcresolver/grpc_resolver_test.go
```diff
@@ -0,0 +1,161 @@
...
+		s.Stop()
+	}
+}
+
+func startTestServers(count int) (_ *test, err error) {
+	t := &test{}
+
+	defer func() {
+		if err != nil {
+			t.cleanup()
```
yurishkuro: Why don't we need to clean-up if there's no error?

pkg/discovery/grpcresolver/grpc_resolver_test.go
```diff
@@ -0,0 +1,161 @@
...
+		}
+	}()
+	for i := 0; i < count; i++ {
+		lis, err := net.Listen("tcp", "localhost:0")
+		if err != nil {
+			return nil, fmt.Errorf("failed to listen %v", err)
```
yurishkuro: it is customary to pass `t *testing.T` to such setup functions, and use normal asserts, e.g. `require.NoError(t, err)`. This way the caller doesn't need to handle errors, which makes the code simpler.

pkg/discovery/grpcresolver/grpc_resolver_test.go
```diff
@@ -0,0 +1,161 @@
...
+
+	notifier := &discovery.Dispatcher{}
+	//discoverer := discovery.FixedDiscoverer(test.addresses)
+	discoverer := discovery.FixedDiscoverer{}
+	re := New(notifier, discoverer, zap.NewNop(), backendCount)
+	assert.NoError(t, err)
+
+	cc, err := grpc.Dial(re.Scheme()+":///round_robin", grpc.WithInsecure(), grpc.WithBalancerName(roundrobin.Name))
+	if err != nil {
+		t.Fatalf("failed to dial: %v", err)
```
yurishkuro: `require.NoError(t, err)` instead of if()

pkg/discovery/grpcresolver/grpc_resolver_test.go
```diff
@@ -0,0 +1,161 @@
...
+	test, err := startTestServers(backendCount)
+	if err != nil {
+		t.Fatalf("failed to start servers: %v", err)
+	}
+	defer test.cleanup()
+
+	notifier := &discovery.Dispatcher{}
+	//discoverer := discovery.FixedDiscoverer(test.addresses)
+	discoverer := discovery.FixedDiscoverer{}
+	re := New(notifier, discoverer, zap.NewNop(), backendCount)
```
yurishkuro: re -> res ('re'  is regexp)

pkg/discovery/grpcresolver/grpc_resolver_test.go
```diff
@@ -0,0 +1,161 @@
...
+			}
+			if p.Addr.String() == test.addresses[si] {
+				connected = true
+				break
+			}
+			time.Sleep(time.Millisecond)
+		}
+		if !connected {
+			t.Fatalf("Connection to %v was still not up", test.addresses[si])
+		}
```
yurishkuro: `require.True(t, connected , "Connection to %v was still not up", test.addresses[si])`

pkg/discovery/grpcresolver/grpc_resolver_test.go
```diff
@@ -0,0 +1,161 @@
...
+	if _, err := testc.EmptyCall(context.Background(), &testpb.Empty{}, grpc.Peer(&p)); err != nil {
+		t.Fatalf("EmptyCall() = _, %v, want _, <nil>", err)
+	}
+	previousAddr := p
+	firstAddr := p
+	for i := 0; i < backendCount; i++ {
+		if _, err := testc.EmptyCall(context.Background(), &testpb.Empty{}, grpc.Peer(&p)); err != nil {
+			t.Fatalf("EmptyCall() = _, %v, want _, <nil>", err)
+		}
+		if previousAddr == p {
```
yurishkuro: if I understand, `grpc.Peer(&p)` gives you back the address. I think  the following would be more robust and easy to understand test:
```
addrs := make(map[string]struct{})
for i := 0; i < backendCount; i++ {
    err := testc.EmptyCall(context.Background(), &testpb.Empty{}, grpc.Peer(&p))
    require.NoErr(t, err)
    addrs[p] = struct{}{}
}
assert.Len(t, addrs, backendCount, "must call each of the servers once")
```

yurishkuro:
btw, why do you need to change `Gopkg.lock`?

guanw:
> btw, why do you need to change `Gopkg.lock`?

hmmm. There seems to be discrepency error in travis yesterday but now that I checked again it was gone somehow. Will revert it back.

pkg/discovery/grpcresolver/grpc_resolver.go
guanw: So in grpc-go `resolver.go` they just mentioned this as not necessary. I saw this was used in `dns_resolver.go` to send signals to force updating. But for our case, we are already firing `r.cc.UpdateState` everytime we receive new instances i don't think this is necessary?

pkg/discovery/grpcresolver/grpc_resolver.go
```diff
@@ -0,0 +1,155 @@
...
+	notifier          discovery.Notifier
+	discoverer        discovery.Discoverer
+	logger            *zap.Logger
+	discoCh           chan []string // used to receive notifications
+	discoveryMinPeers int
+	salt              []byte
+	hasher            hash.Hash32
+
+	//wg will wait for watcher to exit before firing .Close() to close resolver
+	wg sync.WaitGroup
```
yurishkuro: comment not accurate: Close() will block until the watcher loop is finished.

a better name is wg -> closing

guanw: Updated.

pkg/discovery/grpcresolver/grpc_resolver.go
```diff
@@ -41,8 +41,8 @@ type Resolver struct {
 	salt              []byte
 	hasher            hash.Hash32
 
-	//wg will wait for watcher to exit before firing .Close() to close resolver
-	wg sync.WaitGroup
+	//closing will wait for watcher to exit before firing .Close() to close resolver
```
yurishkuro: `// used to block Close() until the watcher goroutine exits its loop

pkg/discovery/grpcresolver/grpc_resolver.go
```diff
@@ -128,24 +124,39 @@ func (r *Resolver) rendezvousHash(addresses []string) []string {
...
-func hashAddr(hasher hash.Hash32, node, saltKey []byte) uint32 {
+func min(a, b int) int {
+	if a < b {
+		return a
+	}
+	return b
+}
+
+func hashAddr(node, saltKey []byte) uint32 {
+	hasher := fnv.New32()
```
yurishkuro: I was suggesting to create it inside rendezvousHash(), so that it's reused for multiple addresses, but not kept across discovery updates

pkg/discovery/discoverer.go
```diff
@@ -27,3 +29,11 @@ type FixedDiscoverer []string
 
 // Instances implements Discoverer.
 func (d FixedDiscoverer) Instances() ([]string, error) { return d, nil }
+
+// ErrorDiscoverer yields a discoverer that returns error
+type ErrorDiscoverer []string
```
yurishkuro: a) I don't think this is a very useful type for public package. If you only use it in tests then define it there.
b) I think you want it to be `type ErrorDiscoverer  error` so that it can `return nil, d` instead of hardcoding error msg

pkg/discovery/grpcresolver/grpc_resolver.go
```diff
@@ -0,0 +1,167 @@
...
+	r := &Resolver{
+		notifier:          notifier,
+		discoverer:        discoverer,
+		discoCh:           make(chan []string, 100),
+		logger:            logger,
+		discoveryMinPeers: discoveryMinPeers,
+		salt:              []byte(strconv.FormatInt(random.Int63(), 10)), // random salt for rendezvousHash
+		scheme:            strconv.FormatInt(seed, 36),                   // make random scheme which will be used when registering
+	}
+	// TODO not sure if there's an equivalent way for grpc to maintain connection like what tchannel did?
```
yurishkuro: does this require follow-up?

pkg/discovery/grpcresolver/grpc_resolver.go
yurishkuro: still needs to be explained in the comment.

pkg/discovery/grpcresolver/grpc_resolver.go
```diff
@@ -0,0 +1,167 @@
...
+	if err != nil {
+		return nil, err
+	}
+	r.updateAddresses(instances)
+	r.closing.Add(1)
+	go r.watcher()
+	return r, nil
+}
+
+// Scheme returns the test scheme.
```
yurishkuro: not the "test" scheme.

guanw:
@yurishkuro Any further comments?

pkg/discovery/grpcresolver/grpc_resolver_test.go
yurishkuro: res

pkg/discovery/grpcresolver/grpc_resolver_test.go
```diff
@@ -0,0 +1,132 @@
...
+	for si := 0; si < backendCount; si++ {
+		connected := false
+		for i := 0; i < 100; i++ {
+			_, err := testc.EmptyCall(context.Background(), &grpctest.Empty{}, grpc.Peer(&p))
+			assert.NoError(t, err)
+			if p.Addr.String() == test.addresses[si] {
+				connected = true
+				break
+			}
+			time.Sleep(time.Millisecond)
```
yurishkuro: I recommend 10ms, and loop to 1000, because in Travis this could fail due to CPU contention

pkg/discovery/grpcresolver/grpc_resolver_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+func TestErrorDiscoverer(t *testing.T) {
+	notifier := &discovery.Dispatcher{}
+	discoverer := discovery.ErrorDiscoverer{}
+	r := New(notifier, discoverer, zap.NewNop(), 2)
+	_, err := r.Build(resolver.Target{}, nil, resolver.BuildOption{})
+	assert.Error(t, err)
+}
+
+func TestGRPCResolverRoundRobin(t *testing.T) {
```
yurishkuro: this is not testing minPeers setting. I suggest refactoring this test to run twice, or even 3 times, with minPeers <. ==. > len(servers) (but avoid restarting servers, just create different clients with respective LBs)

pkg/discovery/grpcresolver/grpc_resolver_test.go
```diff
@@ -0,0 +1,132 @@
...
+	cc, err := grpc.Dial(re.Scheme()+":///round_robin", grpc.WithInsecure(), grpc.WithBalancerName(roundrobin.Name))
+	assert.NoError(t, err, "could not dial using resolver's scheme")
+	defer cc.Close()
+	testc := grpctest.NewTestServiceClient(cc)
+
+	notifier.Notify(test.addresses)
+
+	var p peer.Peer
+	// Make sure connections to all servers are up.
+	for si := 0; si < backendCount; si++ {
```
yurishkuro: it would be more natural to have this loop inside startTestServers() function

pkg/discovery/grpcresolver/grpc_resolver_test.go
```diff
@@ -0,0 +1,132 @@
...
+		connected := false
+		for i := 0; i < 100; i++ {
+			_, err := testc.EmptyCall(context.Background(), &grpctest.Empty{}, grpc.Peer(&p))
+			assert.NoError(t, err)
```
yurishkuro: if the server is not ready, wouldn't invoking EmptyCall() return an error? If so, doing an assert on it might cause the test to fail, even though connection may succeed on the next iteration.

pkg/discovery/grpcresolver/grpc_resolver_test.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+func (t *test) cleanup() {
+	for _, s := range t.servers {
+		s.Stop()
+	}
+}
+
+// ErrorDiscoverer yields a discoverer that returns error
+type ErrorDiscoverer []string
```
yurishkuro: it's not used in this file, only in builder_test.go

pkg/discovery/grpcresolver/grpc_resolver_test.go
```diff
@@ -0,0 +1,163 @@
...
+
+	test := startTestServers(t, backendCount)
+	defer test.cleanup()
+
+	notifier := &discovery.Dispatcher{}
+	discoverer := discovery.FixedDiscoverer{}
+
+	for _, minPeers := range []int{3, 5, 7} {
+		res := New(notifier, discoverer, zap.NewNop(), minPeers)
+		defer resolver.UnregisterForTesting(res.Scheme())
```
yurishkuro: I usually try to avoid calling defer in  loops or if statements, because the scope of `defer` is the whole function, not the enclosing block. You can use subtests for this, e.g.
```
t.Run(fmt.Sprintf("minPeers=%d', minPeers), func(t *testing.T) {
    // test content
})
```

pkg/discovery/grpcresolver/grpc_resolver_test.go
```diff
@@ -0,0 +1,163 @@
...
+		assert.NoError(t, err, "could not dial using resolver's scheme")
+
+		testc := grpctest.NewTestServiceClient(cc)
+
+		notifier.Notify(test.addresses)
+
+		connections := minPeers
+		if connections > backendCount {
+			connections = backendCount
+		}
```
yurishkuro: it is best to avoid "logic" in the tests. You can make this more explicit, e.g.
```
tests := []struct{
    minPeers int
    connections int
} {
  {3, 3}, {5, 5}, {7, 5},
}
for _, test := range tests {
    t.Run(...)
}
```

pkg/discovery/grpcresolver/grpc_resolver_test.go
```diff
@@ -0,0 +1,163 @@
...
+
+		makeSureConnectionsUp(t, connections, testc)
```
yurishkuro: this should be called with `backendCount`, not `connections`. Or even better - it should be outside of the loop, since you're not restarting servers (which is why I suggested making it part of startTestServers, since the expectation of that function is that "servers are ready"). And you don't need to use testc with the load balancer for this health check.

guanw:
@yurishkuro I'm not clear why we don't need to use testc for the health check. If we don't do a initial load balancing and make sure every server has been reached at least once, what's the point of having `makeSureConnectionsUp` function?

yurishkuro:
> I'm not clear why we don't need to use testc for the health check.

we do need to use a connection, but it can be a different connection constructed without load balancer, since you are giving it explicit peer.

guanw:
> > > I'm not clear why we don't need to use testc for the health check.
> > 
> > 
> > we do need to use a connection, but it can be a different connection constructed without load balancer, since you are giving it explicit peer.
> 
> Are you referring to `testc.EmptyCall`? I think we are populating peer instead of "giving". The peers will only be decided by resolver and `minPeers` we passed to it.

@yurishkuro Any comment on my last reply?

yurishkuro:
testc  is created with your custom load balancer. A plain connection without load balancer is sufficient for server liveness checks. Connection with LB is only needed in the actual unit test.

guanw:
> testc is created with your custom load balancer. A plain connection without load balancer is sufficient for server liveness checks. Connection with LB is only needed in the actual unit test.

I'm not certain that's going to work. I tested the resolver without roundrobin balancer(i.e without `grpc.WithBalancerName(roundrobin.Name)`) the testc will just call the same peer everytime. A plain connection without load balancer is sufficient to check if server is live, but without every peer being called at least once, calling roundrobin at second step could still miss some peers(tested and I could see for example {minPeers 5, connections 5} it could end up calling 4 out of 5 peers instead of 5 out of 5).

Also taking a look at `TestBackendsRoundRobin` in `https://github.com/grpc/grpc-go/blob/master/balancer/roundrobin/roundrobin_test.go`, I still think build connection for all server peers is necessary everytime we build a new grpc client connection.

yurishkuro:
You have two rounds of network calls:
  * in the first, you ping each server directly (by given the connection a Peer) to check it's up. You  don't need load balancing on the connection in this case. Therefore, this whole logic can be encapsulated in the startServers() function
  * in the second round you're actually unit testing the LB, so you start a new connection that uses the LB and run unit test with it

guanw:
> You have two rounds of network calls:
> 
> * in the first, you ping each server directly (by giving the connection a Peer) to check it's up. You  don't need load balancing on the connection in this case. Therefore, this whole logic can be encapsulated in the startServers() function
> * in the second round you're actually unit testing the LB, so you start a new connection that uses the LB and run unit test with it

But still, load balancing at step 2 does not necessarily work even if the first step works. Cause now they are from two different connections?

yurishkuro:
Step 1 has nothing to do with step 2, it's preparation of servers and a check that they are alive. If step 2 fails after that (using a different connection) then it's either a bug in the LB or in the test, and makes it even more sense to use *different* connection vs. step 1.

guanw:
> Step 1 has nothing to do with step 2, it's preparation of servers and a check that they are alive. If step 2 fails after that (using a different connection) then it's either a bug in the LB or in the test, and makes it even more sense to use _different_ connection vs. step 1.

I tested against the original grpc-go roundrobin test they make by swapping the "making connection up" part with directly calling peers, it would fail. I posted the question on their github page and will continue on this once they replied.

yurishkuro:
can  you link that issue here?

guanw:
https://github.com/grpc/grpc-go/issues/2808

yurishkuro:
Ah, the explanation in that ticket makes sense - if you create a new connection for the LB test, you'd still need to ensure that it is connected to each peer before you can execute the test. We should add this to the comments in the test.

guanw:
sg. Will add some comment in.

pkg/discovery/grpcresolver/grpc_resolver_test.go
```diff
@@ -0,0 +1,178 @@
...
+
+	testInstances := startTestServers(t, backendCount)
+	defer testInstances.cleanup()
+
+	notifier := &discovery.Dispatcher{}
+	discoverer := discovery.FixedDiscoverer{}
+
+	tests := []struct {
+		minPeers    int
+		connections int
```
yurishkuro: `//  expected number of unique connections to servers`

pkg/discovery/grpcresolver/grpc_resolver_test.go
```diff
@@ -0,0 +1,178 @@
...
+			defer cc.Close()
+
+			testc := grpctest.NewTestServiceClient(cc)
+
+			notifier.Notify(testInstances.addresses)
+
+			connections := test.minPeers
+			if connections > test.connections {
+				connections = test.connections
+			}
```
yurishkuro: I think you're misusing `test.connection`, the whole point of having that field is to not have to do this if statement

pkg/discovery/grpcresolver/grpc_resolver_test.go
```diff
@@ -0,0 +1,178 @@
...
+
+			// This extra step is necessary to make sure servers are ready for roundrobin test.
+			// For detailed explanation, see https://github.com/grpc/grpc-go/issues/2808
+			makeSureConnectionsUp(t, connections, testc)
```
yurishkuro: pass `test.connections`

pkg/discovery/grpcresolver/grpc_resolver_test.go
```diff
@@ -0,0 +1,178 @@
...
+
+			assertRoundRobinCall(t, connections, testc)
```
yurishkuro: pass `test.connections`

pkg/discovery/grpcresolver/grpc_resolver_test.go
```diff
@@ -0,0 +1,178 @@
...
+			testc := grpctest.NewTestServiceClient(cc)
+
+			notifier.Notify(testInstances.addresses)
+
+			connections := test.minPeers
+			if connections > test.connections {
+				connections = test.connections
+			}
+
+			// This extra step is necessary to make sure servers are ready for roundrobin test.
```
yurishkuro: // This step is necessary to ensure that connections to all min-peers are ready, 
// otherwise round-robin may loop only through already connected peers.

pkg/discovery/grpcresolver/grpc_resolver_test.go
```diff
@@ -0,0 +1,178 @@
...
+	defer testInstances.cleanup()
+
+	notifier := &discovery.Dispatcher{}
+	discoverer := discovery.FixedDiscoverer{}
+
+	tests := []struct {
+		minPeers    int
+		connections int
+	}{
+		{3, 5}, {5, 5}, {7, 5},
```
yurishkuro: I think you want the first pair to be {3, 3}

pkg/discovery/grpcresolver/grpc_resolver.go
yurishkuro: Please don't resolve comments if they are not addressed. I don't know what to do with this TODO.

guanw: Tested it on my local setup. Roundrobin will include servers that's brought back automatically. Cheers.

PR #1503, opened 2019-04-26
DRY query servers in all-in-one

yurishkuro:


cmd/all-in-one/main.go
```diff
@@ -317,32 +315,25 @@ func startQuery(
...
+	spanReader spanstore.Reader,
+	depReader dependencystore.Reader,
+	rootFactory metrics.Factory,
+	baseFactory metrics.Factory,
+) *queryApp.Server {
+	spanReader = storageMetrics.NewReadMetricsDecorator(spanReader, baseFactory.Namespace(metrics.NSOptions{Name: "query"}))
+	qs := querysvc.NewQueryService(spanReader, depReader, *queryOpts)
+	server := queryApp.NewServer(svc, qs, qOpts, opentracing.GlobalTracer())
+	if err := server.Start(); err != nil {
+		svc.Logger.Fatal("Could not start jaeger-query service", zap.Error(err))
```
guo0693: If the server fail to start, no healthcheck update, how do we know the query service failed beside checking the log?

cmd/all-in-one/main.go
```diff
@@ -292,16 +299,7 @@ func startZipkinHTTPAPI(
...
-	qOpts *queryApp.QueryOptions,
-	spanReader spanstore.Reader,
-	depReader dependencystore.Reader,
-	logger *zap.Logger,
-	rootFactory metrics.Factory,
-	baseFactory metrics.Factory,
-	hc *healthcheck.HealthCheck,
-	queryOpts querysvc.QueryServiceOptions,
-) {
+func initTracer(metricsFactory metrics.Factory, logger *zap.Logger) io.Closer {
```
guo0693: This works fine, but more intuitively I think initTracer should return a tracer (instead of closer) which provides a close function.

yurishkuro: that's not possible given how jaeger-client API is designed (it returns a separate closer).

cmd/all-in-one/main.go
yurishkuro: Valid question, but not really related to this PR - we'd need to redesign how we handle failures to start in the servers that run in background goroutines. https://github.com/jaegertracing/jaeger/issues/1508

guo0693: OK

cmd/all-in-one/main.go
guo0693: OK

PR #1505, opened 2019-04-27
Enable staticcheck linter and fix warnings

### LINKED ISSUE #1271: Build complains about gosimple
```
$ if [ "$TESTS" == true ]; then make test-ci ; else echo 'skipping tests'; fi
go get -u github.com/rakyll/statik
(cd examples/hotrod/services/frontend/ && statik -f --src web_assets)
CGO_ENABLED=0 installsuffix=cgo go build -o ./examples/hotrod/hotrod- ./examples/hotrod/main.go
gosec -quiet -exclude=G104,G107 ./...
go vet ./...
gosimple ./...
Gosimple has been deprecated. Please use staticcheck instead.
```

yurishkuro:
Replace deprecated gosimple with staticcheck and fix all warnings.

Resolves #1271.

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -39,7 +39,7 @@ required = [
 
 [[constraint]]
   name = "honnef.co/go/tools"
-  version = "=2019.1.1"
+  version = "2019.1.1"
 
 [[constraint]]
   name = "github.com/apache/thrift"
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -39,7 +39,6 @@ required = [
 
 [[constraint]]
   name = "honnef.co/go/tools"
-  # branch = "incr"
   version = "=2019.1.1"
 
 [[constraint]]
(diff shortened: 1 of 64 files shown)


Gopkg.toml
```diff
@@ -36,6 +37,10 @@ required = [
   name = "github.com/securego/gosec"
   version = "1.3.0"
 
+[[constraint]]
+  name = "honnef.co/go/tools"
+  version = "=2019.1.1"
```
vprithvi: Could you comment on why we need to pin this? I prefer similar tools to be unpinned so that we can get the latest checks without manually bumping up.

yurishkuro: Primarily I pinned it because the project is using an odd versioning scheme, which is not a semver. So even if we don't pint it, it would be treated as ^2019.x.x, i.e. will not upgrade next year. But I can unpin, since we don't have an API dependency.

Issue #1506, opened 2019-04-29
jaeger ingester supports customized consumer client id


newly12:
## Requirement - what kind of business use case are you trying to solve?

I'm trying to use Jaeger Ingester to consume tracing messages from Kafka.

## Problem - what in Jaeger blocks you from solving the requirement?

Currently Ingester doesn't support specify client id. Would have to add more customization for ingester.

## Proposal - what do you suggest to solve the problem or improve the existing situation?

na

## Any open questions to address

na

yurishkuro:
Please elaborate what you're trying to achieve by setting a custom client-id. What is the impact of not setting anything?

newly12:
Thanks for reply. 

According to Kafka doc client is is used for tracking purpose
> An id string to pass to the server when making requests. The purpose of this is to be able to track the source of requests beyond just ip/port by allowing a logical application name to be included in server-side request logging.

as well as quota management. https://kafka.apache.org/documentation/#design_quotasgroups
> Quotas can be applied to (user, client-id), user or client-id groups. 

In my case client is is a mandatory value, that must be set to the value that we given in order to connect to our Kafka cluster to consume messages.

PR #1507, opened 2019-04-29
supports customized kafka client id

newly12:
#1506 

supports customized kafka client id in ingester

### DIFF
--- a/cmd/ingester/app/flags.go
+++ b/cmd/ingester/app/flags.go
@@ -68,7 +68,6 @@ const (
 // Options stores the configuration options for the Ingester
 type Options struct {
 	kafkaConsumer.Configuration
-	ClientID         string
 	Parallelism      int
 	Encoding         string
 	DeadlockInterval time.Duration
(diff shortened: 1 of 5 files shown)


plugin/storage/integration/kafka_test.go
```diff
@@ -46,6 +46,7 @@ func (s *KafkaIntegrationTestSuite) initialize() error {
 	s.logger, _ = testutils.NewLogger()
 	const encoding = "json"
 	const groupID = "kafka-integration-test"
+	const ClientID = "kafka-integration-test"
```
vprithvi: nit: this can be private

newly12: Thanks. it's updated.

cmd/ingester/app/flags.go
```diff
@@ -64,6 +68,7 @@ const (
 // Options stores the configuration options for the Ingester
 type Options struct {
 	kafkaConsumer.Configuration
+	ClientID         string
```
vprithvi: I didn't notice this earlier - but we should move this to `kafkaConsumer.Configuration` which is where GroupID/Topic/etc are configured.

newly12: yes it is redundant. it is removed.

vprithvi:
Thanks for the contribution!

PR #1521, opened 2019-05-07
Go modules

freeformz:
- Go Modules is the path forward

- Move from dep to Go Modules,


guanw:
Do you have an issue you want to link this PR to? I think we normally do PR to target at issues with Jaeger.

freeformz:
@guanw The go ecosystem is largely moving to Go Modules. Jaeger moving to go modules keeps the project on that track. This is the basic moving of the code base to modules itself and doesn't tackle anything else related to [Semantic Versioning](https://golang.org/cmd/go/#hdr-Module_compatibility_and_semantic_versioning)

freeformz:
Looks like I need to close this and re-open it because the checks have been modified since I started.

PR #1546, opened 2019-05-16
Move jaeger code base to go modules

freeformz:

Move the jaeger code base to use [Go Modules](https://github.com/golang/go/wiki/Modules) for dependency management.

Replace glide with go modules. Record tooling versions in `tools.go`, adjust Makefile and other build tooling to use go modules.


yurishkuro:
We're not ready to switch to go mod because there are users (including Uber) who import this repo as lib using dep, and dep won't be able to pick the correct versions of transitive dependencies.

We could go to dual more, similar to how in the past we supported both glide and dep simultaneously, for a period of time. If we do that, we'd need the CI to do a build with both tools.

freeformz:
Do you know what time scale Uber has to move things that rely on jaeger to go modules? If they're all public I'd be happy to do the work to move them to go modules as well.

yurishkuro:
I know the work on supporting modules has started, but don't  know the ETA. I'd guess closer to the end of the year. It's internal to Uber.

Issue #1548, opened 2019-05-17
gRPC endpoint is not implemented for baggageRestriction


guanw:
## Requirement - what kind of business use case are you trying to solve?

We are replacing tchannel with grpc now in Uber's internal jaeger system. And one thing that's blocking me to do this is the baggageRestriction. Right now the endpoint is not implemented (check `jaegertracing/jaeger/cmd/agent/app/configmanger/grpc/manager.go GetBaggageRestrictions` for details)

We need to make baggageRestriction available before continue the internal work of replacing tchannel with gRPC.

## Problem - what in Jaeger blocks you from solving the requirement?

Probably none. Just need to follow how baggageRestriction is implemented for tchannel.

## Proposal - what do you suggest to solve the problem or improve the existing situation?

Implement grpc GetBaggageRestrictions.

## Any open questions to address
None

PR #1549, opened 2019-05-18
adding baggageRestriction with grpc

guanw:

- #1548 

- Added proto file and generated file for baggageRestriction
- Also register grpc client for baggageRestriction


guanw:
One thing I'm confused about baggageRestriction is that for tchannel implementation there doesn't seem to be a server on the collector side for oss. Is this intentional or am I missing anything?

PR #1551, opened 2019-05-18
Synchronize elasticsearch index creation

### LINKED ISSUE #1550: Elasticsearch index creation is not serialized
I was looking through tcpdump of an ingester with elasticsearch storage backend and found that index existence is checked a lot of times. Not only that, but index is also created a lot of times.

If I change the `SpanWriter` to log every attempt:

```go
exists, _ := s.client.IndexExists(indexName).Do(s.ctx) // don't need to check the error because the exists variable will be false anyway if there is an error
fmt.Printf("checked if index exists = %v (writer = %p)\n", exists, s)
```

I can see that it's triggered 1000+ times on ingester startup:

```
checked if index exists = true (writer = 0xc000340320)
checked if index exists = true (writer = 0xc000340320)
...
```

My suggestion here is to serialize `SpanWriter::createIndex` with a mutex as there isn't really a good reason to try to do this more than once.

Thoughts?

bobrik:
This makes sure we only create an index once per process and not hundreds of times, like we did with parallel access.

Fixes #1550

Signed-off-by: Ivan Babrou <ibobrik@gmail.com>


plugin/storage/es/spanstore/writer.go
```diff
@@ -151,6 +153,14 @@ func (s *SpanWriter) Close() error {
 
 func (s *SpanWriter) createIndex(indexName string, mapping string, jsonSpan *dbmodel.Span) error {
 	if !keyInCache(indexName, s.indexCache) {
+		s.indexMutex.Lock()
+		defer s.indexMutex.Unlock()
```
yurishkuro: This blocks not only index existence checks, but also all writes, which is not necessary.

Instead of mutex, we can use `sync.Once`, which is faster after the first execution.

bobrik: I don't think `sync.Once` would do, multiple indices are created from here, while `sync.Once` would only allow to create one. Do you have an example of how it would work?

Mutex is also only entered on cache miss, when index is missing.

yurishkuro: you're right, it would need to keep a map of these Once objects per index name, may not be worth it

plugin/storage/es/spanstore/writer.go
```diff
@@ -149,6 +151,14 @@ func (s *SpanWriter) Close() error {
...
+
+		// re-check if index exists in case other goroutine did the job under lock for us
+		if keyInCache(indexName, s.indexCache) {
+			return nil
+		}
```
yurishkuro: any way to add a test for this, e.g. to ensure that s.client.IndexExists() is not called twice (via mock)

PR #1555, opened 2019-05-20
Update version in hotrod example

everett980:
The version in the `examples/hotrod/README.md` is quite old (~10 months I believe) and contains bugs that have been fixed. It seems to me that the example docs should use `latest`. If we don't want to use a dynamic version, this could be a more recent static version (`1.12` in particular).

### DIFF
--- a/examples/hotrod/README.md
+++ b/examples/hotrod/README.md
@@ -29,7 +29,7 @@ docker run \
   --name jaeger \
   -p6831:6831/udp \
   -p16686:16686 \
-  jaegertracing/all-in-one:1.6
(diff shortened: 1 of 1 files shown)


guo0693:
https://vsupalov.com/docker-latest-tag/ something might worth considering when use :latest Tag

pavolloffay:
I am fine with using `latest`, if it is done properly and not overridden by older builds.

yurishkuro:
Yes latest is fine in this case. I wouldn't do it in the blog posts because those need to be repeatable.

PR #1463, opened 2019-04-05
Add CORS handling for Zipkin collector service, fixes #703

JonasVerhofste:
Signed-off-by: Jonas Verhofsté <25819942+JonasVerhofste@users.noreply.github.com>

## Which problem is this PR solving?
- Handling of CORS on the Zipkin collector service, issue #703 

## Short description of the changes
- Add and call the CORS handler
- Add variables for allowed origins and headers

PR #1556, opened 2019-05-21
Change Zipkin CORS origins and headers to comma separated list

JonasVerhofste:
Signed-off-by: Jonas Verhofsté <25819942+JonasVerhofste@users.noreply.github.com>

When I implemented the CORS handling in f9702c4b03903601e752af784b50a66b49c6058f (#1463), I forgot that there can be more than one origin and one header. This adds the functionality of specifying multiple origins and headers as a comma separated list.

### DIFF
--- a/cmd/collector/main.go
+++ b/cmd/collector/main.go
@@ -22,6 +22,7 @@ import (
 	"net/http"
 	"os"
 	"strconv"
+	"strings"
 
 	"github.com/gorilla/mux"
 	"github.com/rs/cors"
@@ -227,10 +228,13 @@ func startZipkinHTTPAPI(
 		r := mux.NewRouter()
 		zHandler.RegisterRoutes(r)
 
+		origins := strings.Split(strings.Replace(allowedOrigins, " ", "", -1), ",")
+		headers := strings.Split(strings.Replace(allowedHeaders, " ", "", -1), ",")
+
 		c := cors.New(cors.Options{
-			AllowedOrigins: []string{allowedOrigins},
+			AllowedOrigins: origins,
 			AllowedMethods: []string{"POST"}, // Allowing only POST, because that's the only handled one
-			AllowedHeaders: []string{allowedHeaders},
+			AllowedHeaders: headers,
 		})
 
 		httpPortStr := ":" + strconv.Itoa(zipkinPort)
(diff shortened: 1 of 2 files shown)


JonasVerhofste:
@yurishkuro Well, it's just an edit of what I added in #1463, where [you told me](https://github.com/jaegertracing/jaeger/pull/1463#issuecomment-480883127) there are no tests?

JonasVerhofste:
I also based myself on how arrays are handled in the storage plugins, but I'm not sure if this is the right way. Can Viper handle arrays like this? For example, in yaml:
```
collector:
  zipkin:
    allowed-origins:
      - origin1
      - origin2
```
As far as I understand the format above is not going to work, but this format would:
```
collector:
  zipkin:
    allowed-origins: "origin1, origin2"
```
I could be wrong of course, haven't gotten the time yet to test it, it's just all just how I interpret the code. Although if I am right, I would suggest taking a look if we could somehow use both (to provide backwards compatibility). The first example is much more human readable than the second one, in my opinion. If I'm wrong, do correct me! Can always have missed something.

yurishkuro:
I don't know about your Viper question, you'd have to try.

My earlier comment about no-testing was in relation to CORS functionality. But we do have tests for parsing command line parameters:
```
$ go test -cover ./cmd/collector/app/builder/
ok  	github.com/jaegertracing/jaeger/cmd/collector/app/builder	0.019s	coverage: 100.0% of statements
```

JonasVerhofste:
> I don't know about your Viper question, you'd have to try.

It seems I was right, only the second one works, arrays don't work. I'll see if I can fix that.



> But we do have tests for parsing command line parameters:

@yurishkuro the only test I can see there is `span_handler_builder_test.go`, as far as I can tell, there's no test for the `main.go` or the `builder_flags.go`, which are the only files I've edited (so far).

cmd/collector/main.go
```diff
@@ -225,10 +226,13 @@ func startZipkinHTTPAPI(
 		r := mux.NewRouter()
 		zHandler.RegisterRoutes(r)
 
+		origins := strings.Split(strings.Replace(allowedOrigins, " ", "", -1), ",")
+		headers := strings.Split(strings.Replace(allowedHeaders, " ", "", -1), ",")
```
yurishkuro: we shouldn't do this in main, it is better to encapsulate this into the builder, similar to this:

https://github.com/jaegertracing/jaeger/blob/4ed618d934a132407b62888969147557b1fcb796/cmd/agent/app/reporter/grpc/flags.go#L47-L50

yurishkuro:
@JonasVerhofste do you plan to finish this one?

JonasVerhofste:
@yurishkuro Sure, but I'm still wondering where I should add those tests. Like I noted before, `cmd/collector/app/builder/builder_flags.go` doesn't have a test file. Whatever I change to the file, the coverage stays the same. Which is probably because only `cmd/collector/app/builder/span_handler_builder.go` has a test-file?

Issue #1011, opened 2018-08-21
Fix Kafka storage tests


pavolloffay:
Kafka storage tests randomly fails https://travis-ci.org/jaegertracing/jaeger/jobs/418644623

```
-- FAIL: TestKafkaStorage (6.21s)
    --- FAIL: TestKafkaStorage/GetTrace (3.02s)
    	integration_test.go:99: Waiting for storage backend to update documents, iteration 1 out of 30
    	integration_test.go:180: Trace was not found
    	integration_test.go:99: Waiting for storage backend to update documents, iteration 2 out of 30
    	integration_test.go:180: Trace was not found
    	integration_test.go:99: Waiting for storage backend to update documents, iteration 3 out of 30
    	integration_test.go:180: Trace was not found
    	integration_test.go:99: Waiting for storage backend to update documents, iteration 4 out of 30
    	integration_test.go:180: Trace was not found
    	integration_test.go:99: Waiting for storage backend to update documents, iteration 5 out of 30
    	integration_test.go:180: Trace was not found
    	integration_test.go:99: Waiting for storage backend to update documents, iteration 6 out of 30
    	integration_test.go:180: Trace was not found
    	integration_test.go:99: Waiting for storage backend to update documents, iteration 7 out of 30
    	integration_test.go:180: Trace was not found
    	integration_test.go:99: Waiting for storage backend to update documents, iteration 8 out of 30
    	integration_test.go:180: Trace was not found
    	integration_test.go:99: Waiting for storage backend to update documents, iteration 9 out of 30
    	integration_test.go:180: Trace was not found
    	integration_test.go:99: Waiting for storage backend to update documents, iteration 10 out of 30
    	integration_test.go:180: Trace was not found
    	integration_test.go:99: Waiting for storage backend to update documents, iteration 11 out of 30
    	integration_test.go:180: Trace was not found
    	integration_test.go:99: Waiting for storage backend to update documents, iteration 12 out of 30
    	integration_test.go:180: Trace was not found
    	integration_test.go:99: Waiting for storage backend to update documents, iteration 13 out of 30
    	integration_test.go:180: Trace was not found
    	integration_test.go:99: Waiting for storage backend to update documents, iteration 14 out of 30
    	integration_test.go:180: Trace was not found
    	integration_test.go:99: Waiting for storage backend to update documents, iteration 15 out of 30
    	integration_test.go:180: Trace was not found
[...]

PR #1563, opened 2019-05-24
Speed up building of all-in-one image by enclosing docker build context

wxdao:
Signed-off-by: wxdao <waxiadao@gmail.com>

Before this PR building of all-in-one image uses the whole project directory as docker build context, while the artifacts needed for the image are all stored under cmd/all-in-one/. Since the former could take up as large as 1Gi after `make build`, most of the building time is consumed on `Sending build context to Docker daemon`.

This PR tweaks Dockerfile in cmd/all-in-one and the build script in scripts/travis to limit docker build context to be cmd/all-in-one, where all the needed artifacts are stored after `make build-all-in-one-linux`.

### DIFF
--- a/cmd/all-in-one/Dockerfile
+++ b/cmd/all-in-one/Dockerfile
@@ -26,8 +26,8 @@ EXPOSE 14250
 # Web HTTP
 EXPOSE 16686
 
-COPY ./cmd/all-in-one/all-in-one-linux /go/bin/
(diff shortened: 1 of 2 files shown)


wxdao:
None of the failed parts havs anything to do with this change. I think it encountered a randomly fail as #1011. What should I do?

PR #1572, opened 2019-05-29
Add docs generation command

### LINKED ISSUE #1537: Generate flags documentation
Cobra allows generating documentation based on commands and flags https://github.com/spf13/cobra#generating-documentation-for-your-command.

This documentation could be made available on our documentation. People often ask what options are we supporting. The operator could also leverage these docs as flags are passed directly into options map in CR.

cc @objectiser

pavolloffay:
Resolves #1537 

docs PR: https://github.com/jaegertracing/documentation/pull/257

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -30,7 +30,8 @@ required = [
     "github.com/golang/protobuf/protoc-gen-go",
     "github.com/gogo/protobuf/protoc-gen-gogo",
     "github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway",
-    "github.com/grpc-ecosystem/grpc-gateway/protoc-gen-swagger"
+    "github.com/grpc-ecosystem/grpc-gateway/protoc-gen-swagger",
+    "github.com/russross/blackfriday"
 ]
 
 [[constraint]]
@@ -48,6 +49,11 @@ required = [
 [[constraint]]
   name = "github.com/bsm/sarama-cluster"
   version = "=2.1.13"
+
+[[constraint]]
+  name = "github.com/russross/blackfriday"
+  version = "=1.5"
+
 
 [[constraint]]
   branch = "master"
--- a/cmd/agent/main.go
+++ b/cmd/agent/main.go
@@ -85,7 +85,7 @@ func main() {
 	}
 
 	command.AddCommand(version.Command())
-	command.AddCommand(docs.Command())
+	command.AddCommand(docs.Command(v))
 
 	config.AddFlags(
 		v,
--- a/cmd/agent/main.go
+++ b/cmd/agent/main.go
@@ -30,6 +30,7 @@ import (
 	"github.com/jaegertracing/jaeger/cmd/agent/app/reporter"
 	"github.com/jaegertracing/jaeger/cmd/agent/app/reporter/grpc"
 	"github.com/jaegertracing/jaeger/cmd/agent/app/reporter/tchannel"
+	"github.com/jaegertracing/jaeger/cmd/docs"
 	"github.com/jaegertracing/jaeger/cmd/flags"
 	"github.com/jaegertracing/jaeger/pkg/config"
 	"github.com/jaegertracing/jaeger/pkg/version"
@@ -84,6 +85,7 @@ func main() {
 	}
 
 	command.AddCommand(version.Command())
+	command.AddCommand(docs.Command())
 
 	config.AddFlags(
 		v,
--- a/cmd/all-in-one/main.go
+++ b/cmd/all-in-one/main.go
@@ -24,13 +24,13 @@ import (
 	"strconv"
 
 	"github.com/gorilla/mux"
-	opentracing "github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 	jaegerClientConfig "github.com/uber/jaeger-client-go/config"
 	jaegerClientZapLog "github.com/uber/jaeger-client-go/log/zap"
 	"github.com/uber/jaeger-lib/metrics"
-	tchannel "github.com/uber/tchannel-go"
+	"github.com/uber/tchannel-go"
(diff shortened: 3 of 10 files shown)


cmd/docs/command.go
```diff
@@ -0,0 +1,35 @@
...
+func Command() *cobra.Command {
+	return &cobra.Command{
+		Use:   "docs",
+		Short: "Generates documentation",
+		Long:  `Generates command and flags documentation`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			for cmd.Parent() != nil {
+				cmd = cmd.Parent()
+			}
+			return doc.GenMarkdownTree(cmd, "./")
```
yurishkuro: ran this locally, it's kind of misleading that there's no output from the command. I would recommend at least some log file "generating .md files in the current directory"

cmd/docs/docs.md
```diff
@@ -0,0 +1,19 @@
+## docs
```
yurishkuro: does this file need to be checked in?

Gopkg.toml
```diff
@@ -30,7 +30,8 @@ required = [
     "github.com/golang/protobuf/protoc-gen-go",
     "github.com/gogo/protobuf/protoc-gen-gogo",
     "github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway",
-    "github.com/grpc-ecosystem/grpc-gateway/protoc-gen-swagger"
+    "github.com/grpc-ecosystem/grpc-gateway/protoc-gen-swagger",
+    "github.com/russross/blackfriday"
```
yurishkuro: why do we need it here?

Gopkg.toml
```diff
@@ -49,6 +50,11 @@ required = [
   name = "github.com/bsm/sarama-cluster"
   version = "=2.1.13"
 
+[[constraint]]
+  name = "github.com/russross/blackfriday"
+  version = "=1.5"
```
yurishkuro: Why do we need this as explicit dependency, instead of transitive via cobra?

cmd/all-in-one/main.go
```diff
@@ -165,6 +167,11 @@ func main() {
 		strategyStoreFactory.AddFlags,
 	)
 
+	//err = doc.GenMarkdownTree(command, "./")
+	//if err != nil {
+	//	log.Fatal(err)
+	//}
```
yurishkuro: plz remove

pavolloffay:
Cobra can also output html and man pages. We could even make it configurable via a flag.

Most importantly we should go with a format which can be embedded to our docs. I am not sure if we can easily embed there htlm. The markdown can be easily embedded see the PR to the website.

Gopkg.toml
pavolloffay: I could not make it work that way

Gopkg.toml
pavolloffay: It's needed by cobra and the automatic transitive resolution didn't work.

pavolloffay:
> actually, I don't think pre-formatted output in .md file is the best format. I see that blackfriday supports HTML output, not sure if it makes a different and/or of cobra can generate HTML.

actually cobra cannot generate html, I have added a flag to generate output in md, rst and man format.

cmd/docs/command.go
```diff
@@ -0,0 +1,81 @@
...
+				cmd = cmd.Parent()
+			}
+			log.Println("Generating documentation in the current working directory")
+			switch v.GetString(formatFlag) {
+			case "md":
+				doc.GenMarkdownTree(cmd, "./")
+			case "man":
+				return man(cmd)
+			case "rst":
+				doc.GenReSTTree(cmd, "./")
```
yurishkuro: Why do some of these return error and others don't?

cmd/docs/command.go
```diff
@@ -0,0 +1,81 @@
...
+		Short: "Generates documentation",
+		Long:  `Generates command and flags documentation`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			for cmd.Parent() != nil {
+				cmd = cmd.Parent()
+			}
+			log.Println("Generating documentation in the current working directory")
+			switch v.GetString(formatFlag) {
+			case "md":
+				doc.GenMarkdownTree(cmd, "./")
```
yurishkuro: Should it also take the output dir param? You gitignored cmd/docs, which seems different.

cmd/docs/command.go
pavolloffay: +1 done

pavolloffay:
@yurishkuro one last thing. Do we like more `jaeger docs` or `jaeger gen docs`. The later is more flexible if we add more gen commands (e.g. auto-completion).

cmd/agent/main.go
```diff
@@ -84,6 +85,7 @@ func main() {
 	}
 
 	command.AddCommand(version.Command())
+	command.AddCommand(docs.Command())
```
yurishkuro: missing `v` arg?

cmd/docs/command.go
```diff
@@ -0,0 +1,85 @@
...
+func Command(v *viper.Viper) *cobra.Command {
+	c := &cobra.Command{
+		Use:   "docs",
+		Short: "Generates documentation",
+		Long:  `Generates command and flags documentation`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			for cmd.Parent() != nil {
+				cmd = cmd.Parent()
+			}
+			log.Println("Generating documentation in the current working directory")
```
yurishkuro: `in ${dir}`, not current dir

cmd/docs/command.go
```diff
@@ -0,0 +1,85 @@
...
+			dir := v.GetString(dirFlag)
+			switch v.GetString(formatFlag) {
+			case "md":
+				return doc.GenMarkdownTree(cmd, dir)
+			case "man":
+				return man(cmd, dir)
+			case "rst":
+				return doc.GenReSTTree(cmd, dir)
```
yurishkuro: let's add
```go
			case "yaml":
				return doc.GenYamlTree(cmd, dir)
```

Issue #1574, opened 2019-05-31
No metrics tag for sampler.type


guanw:
## Requirement - what kind of business use case are you trying to solve?

Within Uber we are trying to remove downsampling in collector completely. To do this, we need some insight into samplers used by different services. The current metrics doesn't provide such thing.

## Problem - what in Jaeger blocks you from solving the requirement?
None

## Proposal - what do you suggest to solve the problem or improve the existing situation?
Add a new tag in metrics.go named `samplerType`

## Any open questions to address
None

PR #1576, opened 2019-05-31
Add sampler type tag to collector ingestion metrics

guanw:
- #1574 

- This pull request adds new tag representing sampler types to metrics


cmd/collector/app/metrics.go
```diff
@@ -237,6 +241,9 @@ func (m *countsBySvc) countByServiceName(serviceName string, isDebug bool) {
 			debugStr = "true"
 		}
 		tags := map[string]string{"svc": serviceName, "debug": debugStr}
+		if samplerType != "" {
```
guanw: Only add tag when `samplerType` is not empty. This ensures only trace counter gets this tag.

cmd/collector/app/metrics.go
```diff
@@ -28,6 +28,8 @@ const (
 
 	// otherServices is the catch-all label when number of services exceeds maxServiceNames
 	otherServices = "other-services"
+
+	samplerTypeKey = "samplerType"
```
black-adder: let's use `sampler_type` since the metrics library we use will lowercase this when it's reported.

model/span.go
```diff
@@ -47,6 +50,15 @@ func (s *Span) HasSpanKind(kind ext.SpanKindEnum) bool {
 	return false
 }
 
+// GetSamplerType returns the sampler type for span
+func (s *Span) GetSamplerType() string {
+	// There's no corresponding opentracing-go tag label corresponding to sampler.type
+	if tag, ok := KeyValues(s.Tags).FindByKey(samplerType); ok {
+		return tag.VStr
```
black-adder: what if this is an empty string? better to return `unknown`

model/span.go
```diff
@@ -47,6 +50,18 @@ func (s *Span) HasSpanKind(kind ext.SpanKindEnum) bool {
 	return false
 }
 
+// GetSamplerType returns the sampler type for span
+func (s *Span) GetSamplerType() string {
+	// There's no corresponding opentracing-go tag label corresponding to sampler.type
+	if tag, ok := KeyValues(s.Tags).FindByKey(samplerType); ok {
+		if tag.VStr == "" {
+			return samplerTypeUnknown
```
black-adder: is this line covered?

guanw: Updated

jpkrohling:
If we have one service, "hotrod", deployed in two hosts/pods, each with a different sampler (const, ratelimiting), we'd then end up with two metrics, right?

What's the impact of this PR on people who are using the previous version of this metric?

black-adder:
> What's the impact of this PR on people who are using the previous version of this metric?

Depending on how users have setup their dashboards, users may see the `service.traces.received` and `service.spans.received` metrics now split by the `sampler_type` label. But if users have metrics setup like: `sum(service.traces.received) by (svc, debug, format)` then they should see no difference as the two new metrics will be summed.

yurishkuro:
Do we need to be changing service.spans.received? Sampler is only defined on the root span and that's what traces.received metric is already capturing.

guanw:
> Do we need to be changing service.spans.received? Sampler is only defined on the root span and that's what traces.received metric is already capturing.

Yeah you're right. The samplerType metric will only be applied to trace.received. For spans.received it will simply pass.

model/span_test.go
```diff
@@ -206,6 +206,16 @@ func TestIsDebug(t *testing.T) {
...
 }
 
+func TestSamplerType(t *testing.T) {
+	span := makeSpan(model.String("sampler.type", "lowerbound"))
+	assert.Equal(t, "lowerbound", span.GetSamplerType())
+	span = makeSpan(model.String("sampler.type", ""))
+	assert.Equal(t, "unknown", span.GetSamplerType())
+	span = makeSpan(model.KeyValue{})
+	assert.Equal(t, "unknown", span.GetSamplerType())
+
```
black-adder: rm whitespace

guanw:
@yurishkuro is there anyway to debug travis ci test failure? `make test-ci` runs fine on my local setup and I couldn't reproduce the error by manually running L11-16 of `build-all-in-one-image.sh` locally either. (As quoted by Won: Best kind of bug)

yurishkuro:
don't know if you can debug it, but the log output suggests a timeout - https://github.com/jaegertracing/jaeger/issues/1583

cmd/collector/app/metrics.go
black-adder: can you add this comment in the code? This feels slightly hacky and I think we can refactor this better (not sure how atm)

cmd/collector/app/metrics.go
```diff
@@ -136,18 +138,26 @@ func newMetricsBySvc(factory metrics.Factory, category string) metricsBySvc {
...
 }
 
-func newCountsBySvc(factory metrics.Factory, category string, maxServiceNames int) countsBySvc {
+func newCountsBySvc(factory metrics.Factory, category string, maxServiceNames int, isTraceMetric bool) countsBySvc {
+	countsTags := map[string]string{"svc": otherServices, "debug": "false"}
+	debugCountsTags := map[string]string{"svc": otherServices, "debug": "true"}
+	if isTraceMetric {
+		// The `samplerTypeKey` tag needs be added to otherServices trace metric counter so prometheus
+		// won't panic because `otherServices` share different tags with normal service.
+		countsTags[samplerTypeKey] = "unknown"
```
black-adder: does this mean "other" services will always be classified with the "unknown" `sampler_type`? I'd prefer to have a counter for each `sampler_type` for "unknown" as well if possible.

cmd/collector/app/metrics.go
```diff
@@ -237,6 +249,9 @@ func (m *countsBySvc) countByServiceName(serviceName string, isDebug bool) {
 			debugStr = "true"
 		}
 		tags := map[string]string{"svc": serviceName, "debug": debugStr}
+		if samplerType != "" {
+			tags[samplerTypeKey] = samplerType
+		}
 		c := m.factory.Counter(metrics.Options{Name: m.category, Tags: tags})
 		counts[serviceName] = c
```
black-adder: we have a problem here where if a service has 2 sampler types (`lowerbound` and `probabilistic` both occur all the time together), we only create a counter once. We have to key this map differently so that we can store counters for multiple `sampler_type`s.

guanw: what do you think if we just put `service_name + sampler_type` as key here.

cmd/collector/app/metrics.go
guanw: Would this break jaeger users's metric dashboard though? I'm fine with the updates but want to make sure impact is minimum for now.

black-adder: _should_ have no impact

cmd/collector/app/metrics.go
black-adder: sure, but do something like `service_name + "$_$" + sampler_type` just in case there is a collision. Also, might want to use a string builder since string concatenation is expensive.

guanw: Updated. Also I separate the `countsBySvc` for trace and span cause I'm using sync.Pool to reduce allocation of stringBuilder for trace whereas span metrics doesn't need that. Figured it made more sense to do so. Please let me know if you disagree.

cmd/collector/app/metrics.go
```diff
@@ -28,6 +29,25 @@ const (
 
 	// otherServices is the catch-all label when number of services exceeds maxServiceNames
 	otherServices = "other-services"
+
+	samplerTypeKey           = "sampler_type"
+	samplerTypeConst         = "const"
+	samplerTypeRemote        = "remote"
```
black-adder: remote is an "abstract" sampler type and it will never be reported.

cmd/collector/app/metrics.go
```diff
@@ -50,7 +70,7 @@ type SpanProcessorMetrics struct {
 	spanCounts    SpanCountsByFormat
 }
 
-type countsBySvc struct {
```
black-adder: let's keep this struct and embed it in spanCountsBySvc and traceCountsBySvc?

cmd/collector/app/metrics.go
```diff
@@ -136,19 +166,43 @@ func newMetricsBySvc(factory metrics.Factory, category string) metricsBySvc {
...
-		},
-		debugCounts: map[string]metrics.Counter{
-			otherServices: factory.Counter(metrics.Options{Name: category, Tags: map[string]string{"svc": otherServices, "debug": "true"}}),
+func newTraceCountsBySvc(factory metrics.Factory, category string, maxServices int) traceCountsBySvc {
+	return traceCountsBySvc{
+		counts:      newTraceCountsOtherServices(factory, category, "false"),
+		debugCounts: newTraceCountsOtherServices(factory, category, "true"),
+		factory:     factory,
+		lock:        &sync.Mutex{},
+		maxServices: maxServices + numOfSamplerTypes, // numOfSamplerType is the offset added to maxServices threshold
```
black-adder: multiply?

guanw: oh, about this, I thought if we keep using the 4000 threshold for `serviceName * numOfSamplerType`, the `numberOfSamplerTypes` added here will just be "compensating" for extra slots occupied by `otherServices[SamplerType]`

cmd/collector/app/metrics.go
guanw: So what's returned for adaptive sampling in `(s *Span) GetSamplerType()`?

black-adder: `probabilistic` or `lowerbound`

Issue #1583, opened 2019-06-03
all-in-one integration test is unstable


yurishkuro:
Build failing https://travis-ci.org/jaegertracing/jaeger/jobs/540902708#L990

```
=== RUN   TestAllInOne
--- FAIL: TestAllInOne (1.01s)
    require.go:794: 
        	Error Trace:	all_in_one_test.go:86
        	            				all_in_one_test.go:60
        	Error:      	Received unexpected error:
        	            	Get http://0.0.0.0:16686/api/traces?service=jaeger-query&tag=jaeger-debug-id:debug: dial tcp 0.0.0.0:16686: connect: connection refused
        	Test:       	TestAllInOne
FAIL
FAIL	github.com/jaegertracing/jaeger/cmd/all-in-one	1.222s
make: *** [integration-test] Error 1
```

Looking at https://github.com/jaegertracing/jaeger/blob/5b36e3bc3ccb3dccfdf42eaa0d8f0bead19dff57/cmd/all-in-one/all_in_one_test.go#L121-L129

it seems that it may not wait long enough for health check to become green. I would extend the loop to 10k iteration (10seconds).

guanw:
Not sure about this.. Since the error was thrown on L86 where it was tryhing to `getAPITrace` instead of `healthCheck`. But I will try that in a bit.

guanw:
Increase the loop count & sleep time don't work. I'm going to try that script on a linux vm environment see if I could reprod there.

guanw:
Turns out this to be an issue with prometheus panic as it requires all categories sharing same structure (in this case, the tag set should also be the same) I made the fix in the original story.

yurishkuro:
are you sure?  I thought I saw the same failure on  master.

guanw:
i think so. The test is passing now on #1576

yurishkuro:
OK, in that case this ticket is still relevant, in the sense that the integration test doesn't check something. For example, if your issue in #1576 was causing all-in-one to crash upon receiving a request, we should've seen it in the test somehow, e.g. by dumping logs from Docker.

yurishkuro:
have not seen this issue for a long time, closing

PR #1587, opened 2019-06-05
dump log if integration test fails

guanw:

- #1583 

- Add additional bash script to dump docker log if docker integration test panic&fail

### DIFF
--- a/scripts/travis/build-all-in-one-image.sh
+++ b/scripts/travis/build-all-in-one-image.sh
@@ -14,9 +14,9 @@ docker build -f cmd/all-in-one/Dockerfile -t $REPO:latest cmd/all-in-one
 export CID=$(docker run -d -p 16686:16686 -p 5778:5778 $REPO:latest)
 make integration-test
 
-if [ -z "$( docker ps | grep jaeger )" ] ; then
-    echo "---- jaeger stopped unexpectedly during test ---------"
-    echo "---- please check the following log for details ----"
+if [ $? -ne 0 ] ; then
+    echo "---- integration test failed unexpectedly ----"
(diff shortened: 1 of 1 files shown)


guanw:
@yurishkuro I use the same logic to uncover the panic error in docker integration test. What do you think?

scripts/travis/build-all-in-one-image.sh
```diff
@@ -13,6 +13,14 @@ export REPO=jaegertracing/all-in-one
 docker build -f cmd/all-in-one/Dockerfile -t $REPO:latest cmd/all-in-one
 export CID=$(docker run -d -p 16686:16686 -p 5778:5778 $REPO:latest)
 make integration-test
+
+if [ -z "$( docker ps | grep jaeger )" ] ; then
```
yurishkuro: I think this condition is too specific. If the test fails, `make` will return a non-zero exit code, and that's the condition to dump the logs.

guanw: Do you mean `export code=$(make integration-test)`?
I ran it locally and the result turns out to be string instead of integer. 

I think this dump would help when golang panics during the test. If the test failed without panic, the log won't be very interesting to see.

yurishkuro:
$(cmd) returns output, not status code. Status is exposed via a special she'll variable, like $? or something.

We should be running with debug logs, so they could be interesting even without panic.

guanw:
> $(cmd) returns output, not status code. Status is exposed via a special she'll variable, like $? or something.
> 
> We should be running with debug logs, so they could be interesting even without panic.

Got it. Updated the condition.

PR #1591, opened 2019-06-07
Add optional client TLS auth

tcolgate:


tcolgate:
The option names in the agent probably aren't optimal.

yurishkuro:
Is it possible to add tests? I think we  already have tests for TLS & certificates in other parts of the code

tcolgate:
yep, the existing tests should make this reasonably easy I think, I'll get started. I'll need to update the docs too.

tcolgate:
@yurishkuro added tests that check the various combinations of agent/client TLS misconfiguration

tcolgate:
I'm not sure why the CI tests failed there, I don't think it's related to my PR.

cmd/agent/app/reporter/grpc/builder_test.go
```diff
@@ -180,6 +189,39 @@ func TestProxyBuilder(t *testing.T) {
 			grpcBuilder: &ConnBuilder{CollectorHostPorts: []string{"localhost:0000"}, TLS: true, TLSCA: "/not/valid"},
 			expectError: true,
 		},
+		{
+			name: "with secure grpc connection and valid TLS Client settings",
+			grpcBuilder: &ConnBuilder{
+				CollectorHostPorts: []string{"localhost:0000"},
+				TLS:                true,
+				TLSCA:              tmpfile.Name(),
+				TLSCert:            "testdata/client.jaeger.io-client.pem",
```
yurishkuro: if we're going with separate  test files, it would be good to remove the temp file logic in L153-L165 and move `certPEM` to a file as well. Looks like it simplifies the test.

cmd/collector/app/builder/builder_flags.go
```diff
@@ -32,6 +32,7 @@ const (
 	collectorGRPCTLS              = "collector.grpc.tls"
 	collectorGRPCCert             = "collector.grpc.tls.cert"
 	collectorGRPCKey              = "collector.grpc.tls.key"
+	collectorGRPCClientCA         = "collector.grpc.tls.client.ca"
```
yurishkuro: This is slightly inconsistent with the flags in the agent, where it's called just `tls.ca`. Is there a logical reason to have that difference?

If not, it would be nice to refactor this whole TLS flags into a shared `pkg/tls` component, that would include defining & parsing flags and instantiating the objects. NB: it's ok if you'd rather not do that, we can do it in a separate PR, I just want to get your thoughts on whether it's a feasible approach. We seem to have quite a lot of these already, e.g. in storage:
```shell
$ grep -r .tls.cert plugin
plugin/storage/cassandra/options.go:	suffixCert                 = ".tls.cert"
plugin/storage/es/options.go:	suffixCert              = ".tls.cert"
```

Even in this PR we're repeating code to load cert/key/etc twice.

tcolgate: There are two CAs involved, for two different purposes. One verifies hosts connected to by outbound hosts, the other checks certs for incoming clients. In this current case, each service only needs to do one of those, so only take one CA (though use it for different purposes). I'd suggest keeping the names distinct in the event that you extend the approach in this PR to other components.

tcolgate:
@yurishkuro any more thoughts on this PR?

freeformz:
Why not take an optional [tls.Config](https://golang.org/pkg/crypto/tls/#Config) when tls configuration is required?

tcolgate:
> Why not take an optional [tls.Config](https://golang.org/pkg/crypto/tls/#Config) when tls configuration is required?

@freeformz I'm not sure what you are suggesting, do you mean have a tls.Config in a user configurable config object? (I don't think that could work, most of that type is not configurable by literals, and none of the fields have the appropirate annotations).

cmd/agent/app/reporter/grpc/flags.go
```diff
@@ -39,6 +41,8 @@ func AddFlags(flags *flag.FlagSet) {
 	flags.Bool(collectorTLS, false, "Enable TLS.")
 	flags.String(collectorTLSCA, "", "Path to a TLS CA file. (default use the systems truststore)")
 	flags.String(collectorTLSServerName, "", "Override the TLS server name.")
+	flags.String(agentCert, "", "Path to a TLS client certificate file.")
+	flags.String(agentKey, "", "Path to a TLS client key file.")
```
yurishkuro: Could you please expand the flag descriptions (starting from `--tls`) to explain in more details what they enable? For example, what `collectorTLS` means after this PR, etc. Pretend someone is reading https://www.jaegertracing.io/docs/1.13/cli/#jaeger-collector-cassandra, is there enough information to understand what's going to happen with various combinations of the flags?

cmd/collector/app/builder/builder_flags.go
```diff
@@ -74,6 +77,7 @@ func AddFlags(flags *flag.FlagSet) {
 	flags.Bool(collectorGRPCTLS, false, "Enable TLS")
 	flags.String(collectorGRPCCert, "", "Path to TLS certificate file")
 	flags.String(collectorGRPCKey, "", "Path to TLS key file")
+	flags.String(collectorGRPCClientCA, "", "Path to TLS certificates for authenticating clients")
 	flags.String(collectorZipkinAllowedOrigins, "*", "Allowed origins for the Zipkin collector service, default accepts all")
 	flags.String(collectorZipkinAllowedHeaders, "content-type", "Allowed headers for the Zipkin collector service, default content-type")
```
yurishkuro: similar request - expand on all tls-flags descriptions?

cmd/collector/main.go
```diff
@@ -195,13 +198,33 @@ func startGRPCServer(
 		if opts.CollectorGRPCCert == "" || opts.CollectorGRPCKey == "" {
 			return nil, fmt.Errorf("you requested TLS but configuration does not include a path to cert and/or key")
 		}
-		creds, err := credentials.NewServerTLSFromFile(
-			opts.CollectorGRPCCert,
-			opts.CollectorGRPCKey,
-		)
+
+		tlsCfg := &tls.Config{
```
yurishkuro: this code should not be in main, but somewhere in the server factory, maybe similar to `grpcserver.StartGRPCCollector()` function. Right now it appears that similar code is repeated in the unit tests. Can you move it to `grpcserver` and reuse in unit tests so that it's also tested? We can't really test main()

tcolgate:
@yurishkuro rebased, squashed, updated args, and moved the TLS setup.

cmd/collector/app/builder/builder_flags.go
```diff
@@ -71,9 +74,10 @@ func AddFlags(flags *flag.FlagSet) {
 	flags.Int(collectorHTTPPort, ports.CollectorHTTP, "The HTTP port for the collector service")
 	flags.Int(collectorGRPCPort, ports.CollectorGRPC, "The gRPC port for the collector service")
 	flags.Int(collectorZipkinHTTPort, 0, "The HTTP port for the Zipkin collector service e.g. 9411")
-	flags.Bool(collectorGRPCTLS, false, "Enable TLS")
-	flags.String(collectorGRPCCert, "", "Path to TLS certificate file")
-	flags.String(collectorGRPCKey, "", "Path to TLS key file")
+	flags.Bool(collectorGRPCTLS, false, "Enable TLS for the gRPC collector port")
+	flags.String(collectorGRPCCert, "", "Path to TLS certificate or the gRPC collector TLS service")
```
yurishkuro: > or the gRPC collector

was it meant to be  "for"?

cmd/collector/app/builder/builder_flags.go
```diff
@@ -53,6 +54,8 @@ type CollectorOptions struct {
 	CollectorGRPCTLS bool
 	// CollectorGRPCCert is the path to a TLS certificate file for the server
 	CollectorGRPCCert string
+	// CollectorGRPCClientCA is the path to a TLS certificate file for // authenticating clients
```
yurishkuro: > for // authenticating clients

remove `//` ?

cmd/collector/app/builder/builder_flags.go
```diff
@@ -71,9 +74,10 @@ func AddFlags(flags *flag.FlagSet) {
 	flags.Int(collectorHTTPPort, ports.CollectorHTTP, "The HTTP port for the collector service")
 	flags.Int(collectorGRPCPort, ports.CollectorGRPC, "The gRPC port for the collector service")
 	flags.Int(collectorZipkinHTTPort, 0, "The HTTP port for the Zipkin collector service e.g. 9411")
-	flags.Bool(collectorGRPCTLS, false, "Enable TLS")
-	flags.String(collectorGRPCCert, "", "Path to TLS certificate file")
-	flags.String(collectorGRPCKey, "", "Path to TLS key file")
+	flags.Bool(collectorGRPCTLS, false, "Enable TLS for the gRPC collector port")
+	flags.String(collectorGRPCCert, "", "Path to TLS certificate or the gRPC collector TLS service")
+	flags.String(collectorGRPCKey, "", "Path to TLS key for the gRPC collector TLS cert")
+	flags.String(collectorGRPCClientCA, "", "Path to a TLS CA to verify certificates presented by clients against (if unset, all clients are permitted)")
```
yurishkuro: remove "against"

tcolgate:
@yurishkuro pushed fixes f

Issue #1624, opened 2019-06-21
Data race between Agent.Run() and Agent.Stop()


tigrannajaryan:
## Requirement - what kind of business use case are you trying to solve?

Agent.Stop() must not fail if called after Agent.Run() returns.

## Problem - what in Jaeger blocks you from solving the requirement?

Agent.Stop() panics or shows a data race during tests.

## Proposal - what do you suggest to solve the problem or improve the existing situation?

Fix the data race. Make sure Stop() waits until effects of Agent.Run() are finished or interrupts them gracefully. Alternatively make Agent.Run() synchronous so that it waits until all starting effects are complete.

The root cause of the race is that Stop() calls WaitGroup.Wait() before Run() calls WaitGroup.Add() for the first time which is prohibited: https://golang.org/pkg/sync/#WaitGroup.Add

## Reproduction

Add the following test to /github.com/jaegertracing/jaeger/cmd/agent/app/agent_test.go:
```go
func TestStartStopRace(t *testing.T) {
	resetDefaultPrometheusRegistry()
	cfg := Builder{
		Processors: []ProcessorConfiguration{
			{
				Model:    jaegerModel,
				Protocol: compactProtocol,
				Server: ServerConfiguration{
					HostPort: "127.0.0.1:0",
				},
			},
		},
	}
	logger, logBuf := testutils.NewLogger()
	mBldr := &jmetrics.Builder{HTTPRoute: "/metrics", Backend: "prometheus"}
	mFactory, err := mBldr.CreateMetricsFactory("jaeger")
	require.NoError(t, err)
	agent, err := cfg.CreateAgent(fakeCollectorProxy{}, logger, mFactory)
	require.NoError(t, err)
	if err := agent.Run(); err != nil {
		t.Errorf("error from agent.Run(): %s", err)
	}

	agent.Stop()

	for i := 0; i < 1000; i++ {
		if strings.Contains(logBuf.String(), "agent's http server exiting") {
			return
		}
		time.Sleep(time.Millisecond)
	}
	t.Fatal("Expecting server exit log")
}
```

and run `go test -v -race github.com/jaegertracing/jaeger/cmd/agent/app`

The output shows a data race:

```
==================
WARNING: DATA RACE
Write at 0x00c00018a8f8 by goroutine 19:
  sync.(*WaitGroup).Wait()
      /usr/local/go/src/internal/race/race.go:41 +0xef
[...]

PR #1625, opened 2019-06-21
Fix data race between Agent.Run() and Agent.Stop()

tigrannajaryan:

We had a data race with Agent.Stop() failing if called immediately
after Agent.Run() returns.

The root cause of the race was that Stop() called WaitGroup.Wait()
before Run() called WaitGroup.Add() for the first time which is
prohibited: https://golang.org/pkg/sync/#WaitGroup.Add

This change moves WaitGroup.Add() to earlier stage which guarantees
that WaitGroup.Wait() will be called after that.

Github issue: https://github.com/jaegertracing/jaeger/issues/1624

Testing done:

Added an automated test which was failing before the bug was fixed
and does not fail any more after the fix.

Also verified that `make test` passes.

Signed-off-by: Tigran Najaryan <tigran@najaryan.net>

### DIFF
--- a/cmd/agent/app/agent_test.go
+++ b/cmd/agent/app/agent_test.go
@@ -147,6 +147,47 @@ func withRunningAgent(t *testing.T, testcase func(string, chan error)) {
 	t.Fatal("Expecting server exit log")
 }
 
+func TestStartStopRace(t *testing.T) {
+	resetDefaultPrometheusRegistry()
+	cfg := Builder{
+		Processors: []ProcessorConfiguration{
+			{
+				Model:    jaegerModel,
+				Protocol: compactProtocol,
+				Server: ServerConfiguration{
+					HostPort: "127.0.0.1:0",
+				},
+			},
+		},
+	}
+	logger, logBuf := testutils.NewLogger()
+	mBldr := &jmetrics.Builder{HTTPRoute: "/metrics", Backend: "prometheus"}
+	mFactory, err := mBldr.CreateMetricsFactory("jaeger")
+	require.NoError(t, err)
+	agent, err := cfg.CreateAgent(fakeCollectorProxy{}, logger, mFactory)
+	require.NoError(t, err)
+
+	// This test attempts to hit the data race bug when Stop() is called
+	// immediately after Run(). We had a bug like that which is now fixed:
+	// https://github.com/jaegertracing/jaeger/issues/1624
+	// Before the bug was fixed this test was failing as expected when
+	// run with -race flag.
+
+	if err := agent.Run(); err != nil {
+		t.Errorf("error from agent.Run(): %s", err)
+	}
+
+	agent.Stop()
+
+	for i := 0; i < 1000; i++ {
+		if strings.Contains(logBuf.String(), "agent's http server exiting") {
+			return
+		}
+		time.Sleep(time.Millisecond)
+	}
+	t.Fatal("Expecting server exit log")
+}
+
 func resetDefaultPrometheusRegistry() {
 	// Create and assign a new Prometheus Registerer/Gatherer for each test
 	registry := prometheus.NewRegistry()
--- a/cmd/agent/app/processors/thrift_processor.go
+++ b/cmd/agent/app/processors/thrift_processor.go
@@ -22,7 +22,7 @@ import (
 	"github.com/uber/jaeger-lib/metrics"
 	"go.uber.org/zap"
 
-	"github.com/jaegertracing/jaeger/cmd/agent/app/customtransports"
+	customtransport "github.com/jaegertracing/jaeger/cmd/agent/app/customtransports"
 	"github.com/jaegertracing/jaeger/cmd/agent/app/servers"
 )
 
@@ -76,12 +76,12 @@ func NewThriftProcessor(
 		numProcessors: numProcessors,
 	}
 	metrics.Init(&res.metrics, mFactory, nil)
+	res.processing.Add(res.numProcessors)
 	return res, nil
 }
 
 // Serve initiates the readers and starts serving traffic
 func (s *ThriftProcessor) Serve() {
-	s.processing.Add(s.numProcessors)
 	for i := 0; i < s.numProcessors; i++ {
 		go s.processBuffer()
 	}
--- a/cmd/agent/app/processors/thrift_processor.go
+++ b/cmd/agent/app/processors/thrift_processor.go
@@ -22,7 +22,7 @@ import (
 	"github.com/uber/jaeger-lib/metrics"
 	"go.uber.org/zap"
 
-	customtransport "github.com/jaegertracing/jaeger/cmd/agent/app/customtransports"
+	"github.com/jaegertracing/jaeger/cmd/agent/app/customtransports"
 	"github.com/jaegertracing/jaeger/cmd/agent/app/servers"
 )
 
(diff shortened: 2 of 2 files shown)


cmd/agent/app/processors/thrift_processor.go
```diff
@@ -22,7 +22,7 @@ import (
 	"github.com/uber/jaeger-lib/metrics"
 	"go.uber.org/zap"
 
-	"github.com/jaegertracing/jaeger/cmd/agent/app/customtransports"
+	customtransport "github.com/jaegertracing/jaeger/cmd/agent/app/customtransports"
```
tigrannajaryan: This line is needed otherwise line 66 will not compile. I am not sure how it was compiling before. I may be missing something.

cmd/agent/app/agent_test.go
```diff
@@ -147,6 +147,47 @@ func withRunningAgent(t *testing.T, testcase func(string, chan error)) {
 	t.Fatal("Expecting server exit log")
 }
 
+func TestStartStopRace(t *testing.T) {
```
jan25: I couldn't reproduce this on my mac https://github.com/open-telemetry/opentelemetry-service/issues/43

cmd/agent/app/processors/thrift_processor.go
```diff
@@ -76,12 +76,12 @@ func NewThriftProcessor(
 		numProcessors: numProcessors,
 	}
 	metrics.Init(&res.metrics, mFactory, nil)
+	res.processing.Add(res.numProcessors)
```
jan25: I have a hunch that deadlock might've been in `func processBuffer` due to not executing `s.processing.Done()`, maybe it should be `defer s.processing.Done()`?

cmd/agent/app/agent_test.go
jan25: I think `wg.Wait()` doesn't block if `count=0`, not sure if i missed something

cmd/agent/app/processors/thrift_processor.go
jan25: Sorry, that won't make sense unless `Server()` was called

cmd/agent/app/agent_test.go
tigrannajaryan: >I couldn't reproduce this on my mac open-telemetry/opentelemetry-service#43

Yes, unfortunately it is a non-deterministic race, no guarantee that it will be reproducible. The test _may_ catch the problem and it does sometimes when running on my machine.

>I think wg.Wait() doesn't block if count=0, not sure if i missed something

The documentation of WaitGroup.Add() says:

>Note that calls with a positive delta that occur when the counter is zero must happen before a Wait. 

This means it is not valid to call Wait with zero counter before Add is called. However this is precisely what happens in our code.

You are right that in production mode Wait() does not block when count=0 however it is considered a mistake and go test race detector catches this mistake, see this comment https://github.com/golang/go/blob/master/src/sync/waitgroup.go#L124-L127

cmd/agent/app/processors/thrift_processor.go
tigrannajaryan: I am not seeing a deadlock, I am observing a data race. Here is the output of `go test -v -race github.com/jaegertracing/jaeger/cmd/agent/app` without my fix applied:

```
==================
WARNING: DATA RACE
Write at 0x00c000346118 by goroutine 19:
  sync.(*WaitGroup).Wait()
      /usr/local/go/src/internal/race/race.go:41 +0xef
  github.com/jaegertracing/jaeger/cmd/agent/app/processors.(*ThriftProcessor).Stop()
      /home/tigran/go/src/github.com/jaegertracing/jaeger/cmd/agent/app/processors/thrift_processor.go:102 +0x11d

Previous read at 0x00c000346118 by goroutine 18:
  sync.(*WaitGroup).Add()
      /usr/local/go/src/internal/race/race.go:37 +0x169
  github.com/jaegertracing/jaeger/cmd/agent/app/processors.(*ThriftProcessor).Serve()
      /home/tigran/go/src/github.com/jaegertracing/jaeger/cmd/agent/app/processors/thrift_processor.go:84 +0x62

Goroutine 19 (running) created at:
  github.com/jaegertracing/jaeger/cmd/agent/app.(*Agent).Stop()
      /home/tigran/go/src/github.com/jaegertracing/jaeger/cmd/agent/app/agent.go:88 +0xa1
  github.com/jaegertracing/jaeger/cmd/agent/app.TestStartStopRace()
      /home/tigran/go/src/github.com/jaegertracing/jaeger/cmd/agent/app/agent_test.go:180 +0x4ca
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:865 +0x163

Goroutine 18 (running) created at:
  github.com/jaegertracing/jaeger/cmd/agent/app.(*Agent).Run()
      /home/tigran/go/src/github.com/jaegertracing/jaeger/cmd/agent/app/agent.go:75 +0x2bf
  github.com/jaegertracing/jaeger/cmd/agent/app.TestStartStopRace()
      /home/tigran/go/src/github.com/jaegertracing/jaeger/cmd/agent/app/agent_test.go:176 +0x442
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:865 +0x163
==================
--- FAIL: TestStartStopRace (0.01s)
    testing.go:809: race detected during execution of test
```

It shows a race between sync.(*WaitGroup).Wait() and sync.(*WaitGroup).Add() which is prohibited by WaitGroup documentation. The documentation says Add() must be called before Wait().

tigrannajaryan: >I have a hunch that deadlock might've been in func processBuffer due to not executing s.processing.Done(), maybe it should be defer s.processing.Done()?

I only see one exit path in processBuffer(), so s.processing.Done() is guaranteed be called when the function returns (I am ignoring the possible panic case, since that is not happening in this particular case).

I agree with you it would be more idiomatic to use `defer s.processing.Done()` which would make it more resilient if in the future additional exit path is added to processBuffer(), but I do not think in this particular case it cases any problems.

freeformz: suggestion....

Suggestion: - Make this
```go
res.processing.Add(res.numProcessors)
go func() {
  s.processBuffer()
  s.processing.Done()
}()
```

Why? It consolidates Add() and Done() close together.

the processBuffer goroutines will immediately block as there is nothing in the server's data channel.

This also nicely removes the ability to accidentally spin up multiple sets of processors for the same ThriftProcessor value that aren't tracked in the waitgroup.

tigrannajaryan: I like the suggestion, it makes WaitGroup operation more visible. It needs to be slightly modified to account for numProcessors:
```go
	res.processing.Add(res.numProcessors)
	for i := 0; i < res.numProcessors; i++ {
		go func() {
			res.processBuffer()
			res.processing.Done()
		}()
	}
```
I'll push an update to this PR.

tigrannajaryan:
Reviewers, please check again.

cmd/agent/app/processors/thrift_processor.go
freeformz: Suggestion: Rename the directory the the package is in or the package itself to match the directory name. Right now the package is named `customtransport`, which the directory is `customertransports`.

tigrannajaryan: Yeah, that would be the right approach, otherwise it breaks a common naming practice with no apparent good reason.

However, I am reluctant to make this change. If any other external code (not in this repo) imports this package renaming it will break that external code. I'd rather let this repo's maintainers decide if that's what they want to do.

I am reverting this change since apparently it compiles fine the way it was, not sure why I thought it doesn't.

tigrannajaryan:
@yurishkuro @pavolloffay can you review this, please?

pavolloffay:
thanks @tigrannajaryan @freeformz !

tigrannajaryan:
@yurishkuro the race was happening when Stop() was called immediately after Run(). It was happening because when Run() was returning there was no guarantee that `res.processing.Add()` would be already called. So a subsequent Stop() call would try to call `s.processing.Wait()` while `res.processing.Add()` was still in progress. This is incorrect and is not allowed by WaitGroup implementation.

This change guarantees that `res.processing.Add()` is finished when Run() returns. Thus when Stop() is called after Run() it is guaranteed that `s.processing.Wait()` will never race with `res.processing.Add()`. This fixes the race between Add() and Wait() of WaitGroup.

PR #1628, opened 2019-06-25
Minor update on contribution guide

openmohan:
- nil

- update on contribution guide

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -28,13 +28,14 @@ Then install dependencies and run the tests:
 
 ```
 git submodule update --init --recursive
+make install-tools
 dep ensure
 make test
 ```
 
 ### Running local build with the UI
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -88,18 +89,19 @@ github.com/jaegertracing/jaeger
   mkdocs.yml                - MkDocs builds the documentation in docs/
 ```
 
-  * Note 1: `pkg` is a collection of utility packages used by the Jaeger components
-    without being specific to its internals. Utility packages are kept separate from
-    the Jaeger core codebase to keep it as small and concise as possible. If some
-    utilities grow larger and their APIs stabilize, they may be moved to their own
-    repository, to facilitate re-use by other projects.
+- Note 1: `pkg` is a collection of utility packages used by the Jaeger components
```
pavolloffay: nit: limit your PR to only changes which are required. `*` is a valid char for list and removing spaces does not change the formatting either.

PR #1653, opened 2019-07-04
[badger] Make default dirs work in Windows

### LINKED ISSUE #1652: Badger can't create default directories under Windows

Under Windows (and Powershell), setting just ``BADGER_EPHEMERAL = false`` and ``SPAN_STORAGE_TYPE = badger`` is not enough. The default directories (start dir + data + key/value) use hardcoded delimiter.

Starting under Windows without setting badger's key and value directory manually.

```
{"level":"fatal","ts":1562251290.8270292,"caller":"all-in-one/main.go:104","msg":"Failed to init storage factory","error":"Error Creating Dir: \"C:\\\\projects\\\\jaeger/data/keys\": mkdir C:\\projects\\jaeger/data/keys: The system cannot find the path specified.","errorVerbose":"mkdir C:\\projects\\jaeger/data/keys: The system cannot find the path specified.\nError Creating Dir: \"C:\\\\projects\\\\jaeger/data/keys\"\ngithub.com/jaegertracing/jaeger/vendor/github.com/dgraph-io/badger/y.Wrapf\n\
```

burmanm:

Resolves #1652 (Windows default directories do not work), but also ensures that the directories are created successfully since badger only creates the last directory - not the missing parent directories. This is why default directory does not work unless ``data`` is created first.

Use os.pathSeparator instead of hardcoded ``/`` and create directories if possible.

### DIFF
--- a/plugin/storage/badger/factory.go
+++ b/plugin/storage/badger/factory.go
@@ -96,6 +96,10 @@ func (f *Factory) Initialize(metricsFactory metrics.Factory, logger *zap.Logger)
 		f.Options.primary.KeyDirectory = f.tmpDir
 		f.Options.primary.ValueDirectory = f.tmpDir
 	} else {
+		// Errors are ignored as they're catched in the Open call
+		initializeDir(f.Options.primary.KeyDirectory)
+		initializeDir(f.Options.primary.ValueDirectory)
+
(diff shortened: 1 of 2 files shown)


burmanm:
I don't think travis test failures are because of changes in this PR

yurishkuro:
restarting

PR #1640, opened 2019-06-27
Configurable kafka protocol version for msg consuming by jaeger ingester

marqc:
## Which problem is this PR solving?

Jaeger ingester silently doesn't consume messages, when kafka server doesn't support kafka protocol in version 0.0.9.

## Short description of the changes

I added new `kafka.consumer.protocol-version` configuration option, that allows specify kafka protocol version to use. Default is "0.9.0.0" which was used by bsm/sarama-cluster as default version (for backwards compatibility).

PR #1658, opened 2019-07-09
Collector kafka producer protocol version config

marqc:
- Configure kafka server version in client. When kafka server has disabled backwards compatibility with older versions it may cause collector failure, because sarama by default uses oldest supported version.

- This is symmetric patch to https://github.com/jaegertracing/jaeger/pull/1640 as suggested by @yurishkuro  https://github.com/jaegertracing/jaeger/pull/1640#issuecomment-509295078

### DIFF
--- a/plugin/storage/kafka/options.go
+++ b/plugin/storage/kafka/options.go
@@ -60,19 +60,19 @@ func (opt *Options) AddFlags(flagSet *flag.FlagSet) {
 	flagSet.String(
 		configPrefix+suffixBrokers,
 		defaultBroker,
-		"(experimental) The comma-separated list of kafka brokers. i.e. '127.0.0.1:9092,0.0.0:1234'")
+		"The comma-separated list of kafka brokers. i.e. '127.0.0.1:9092,0.0.0:1234'")
 	flagSet.String(
 		configPrefix+suffixTopic,
 		defaultTopic,
-		"(experimental) The name of the kafka topic")
+		"The name of the kafka topic")
 	flagSet.String(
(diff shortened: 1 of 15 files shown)


plugin/storage/kafka/options.go
```diff
@@ -64,6 +65,10 @@ func (opt *Options) AddFlags(flagSet *flag.FlagSet) {
 		configPrefix+suffixTopic,
 		defaultTopic,
 		"(experimental) The name of the kafka topic")
+	flagSet.String(
+		configPrefix+suffixProtocolVersion,
+		"",
+		"(experimental) Kafka protocol version - must be supported by kafka server")
```
yurishkuro: why experimental?

marqc: just sticking to convention. all other kafka.producer settings are marked as experimental

```
      --kafka.producer.brokers string                 (experimental) The comma-separated list of kafka brokers. i.e. '127.0.0.1:9092,0.0.0:1234' (default "127.0.0.1:9092")
      --kafka.producer.encoding string                (experimental) Encoding of spans ("json" or "protobuf") sent to kafka. (default "protobuf")
      --kafka.producer.protocol-version string        (experimental) Kafka protocol version - must be supported by kafka server
      --kafka.producer.topic string                   (experimental) The name of the kafka topic (default "jaeger-spans")
```

yurishkuro: oh, that's probably just stale, can you remove (experimental) from all 4?

Issue #1667, opened 2019-07-18
Comsos DB as storage backend


marius-stanescu:
Could you also add Cosmos DB as a storage backend?
Related to #638 and #1105.

jpkrohling:
We have already Elasticsearch, Cassandra, Badger and in-memory in the main repository. To support other storage backends, we have support for storage plugins (#1461). We might be missing the documentation of how to develop one such plugin, but the PR introducing that feature also added an example:

https://github.com/jaegertracing/jaeger/blob/d3ee676bcb9666832a582ccf8b2c8e126b9532a1/examples/memstore-plugin/main.go#L30-L65

Right now, we are not hosting plugins in the main repository, but given enough traction, we can certainly host it at this organization. 

As such, I'm closing this issue, but do let us know in case you develop a plugin for Cosmos DB!

omerlh:
Hey, @marius-stanescu I started to play a bit with the GRPC plugin API in dotnet, should be pretty simple to write a plugin. Is that something you're interested at?

yurishkuro:
I am going to re-open this and link from #638. We can close once we have a clear answer.

I was under impression that CosmosDB has a wire protocol compatible with Cassandra, is it not the case?

yurishkuro:
There are some discussions about CosmosDB here: #1105

yurishkuro:
And here:  #1205

yurishkuro:
Pining @sagaranand015 @nutharsh @486 @lukasmrtvy who previously commented on issues related to CosmosDB - did anyone got it to work?

sagaranand015:
@yurishkuro : Checked the cosmosDb connection with cassandra API and as mentioned in https://github.com/jaegertracing/jaeger/issues/1105, it works by removing the SnappyCompressor in Cassandra connection. Also, the keyspace creation using templates do not work because of the commas(which Cosmos db does not support)

How do you think we should proceed here? A Plugin approach to Cosmos db or a storage backend altogether?

Also, should the compressor be forced or should it be driven by config? (This seems like a change we can make before going to plugin/storage backend development).

yurishkuro:
@sagaranand015 these changes sound very minor. I don't think it makes sense to treat it as a new storage backend, just parameterize the existing code to allow to control for these variations.

sagaranand015:
@yurishkuro :
So, this boils down to 2 action items:
[...]

Issue #1105, opened 2018-10-08
Do not force cassandra protocol's compressor


sylr:
When using cassandra, Jaeger force the use of the `SnappyCompressor` for the cassandra protocol.

https://github.com/jaegertracing/jaeger/blob/4ada2b924bfcd393ff370506b9bab8ced1c72376/pkg/cassandra/config/config.go#L127

Azure Cosmos DB provides a cassandra compatible API but it does not seem to be compatible with SnappyCompressor (see https://github.com/gocql/gocql/issues/1205).

yurishkuro:
I don't remember, what is the compressor is even used for? Compressing on the wire or at rest?

black-adder:
Would you be open to creating a PR that makes this configurable?

sylr:
FYI Azure Cosmos DB does not allow "CREATE TYPE" so it won't work anyway.

486:
@sylr yes Cosmos DB does support CREATE TYPE, we tried it. It is just not listed. 
So I think making it configurable is still a good idea.

xihw:
Hi @sylr / @black-adder  I also confirmed Cosmos DB does support CREATE TYPE now. Would be of great help if you could make it configurable.

sylr:
Sorry guys, we switched to elastic for the data storage and I no longer have time to test CosmosDB.

486:
@black-adder @yurishkuro we hacked on it and made it work with CosmosDB... mostly. 

The point is that CosmosDB in Cassandra mode has some specialties that we need to work around. In the end it would become a distinct backend type rather than a few config switches. 

Before we put more work into it, would you be open to accept a new backend?

It would surely make jaeger very convenient to use on Azure.

xihw:
@486 good to hear that you are able to make it work. Would you mind sharing more information on what work around you did ?

For me it always says: 
{"level":"info","ts":1541018500.2958531,"caller":"healthcheck/handler.go:99","msg":"Health Check server started","http-port":16687,"status":"unavailable"}
[...]

PR #1675, opened 2019-07-22
Added flags for driving cassandra connection compression through config

sagaranand015:
Signed-off-by: Sagar  Anand <sagar.anand015@gmail.com>

- Resolves https://github.com/jaegertracing/jaeger/issues/1667 and https://github.com/jaegertracing/jaeger/issues/1105

- Added flag to drive the Cassandra Connection Compression from config. This will be use to use SnappyCompressor by default or opt out of using the SnappyCompressor

### DIFF
--- a/pkg/cassandra/config/config.go
+++ b/pkg/cassandra/config/config.go
@@ -38,6 +38,7 @@ type Configuration struct {
 	MaxRetryAttempts     int           `validate:"min=0" yaml:"max_retry_attempt"`
 	ProtoVersion         int           `yaml:"proto_version"`
 	Consistency          string        `yaml:"consistency"`
+	EnableCompression    bool          `yaml:"enable-compression"`
 	Port                 int           `yaml:"port"`
 	Authenticator        Authenticator `yaml:"authenticator"`
 	DisableAutoDiscovery bool          `yaml:"disable_auto_discovery"`
@@ -128,7 +129,11 @@ func (c *Configuration) NewCluster() *gocql.ClusterConfig {
 	if c.Port != 0 {
 		cluster.Port = c.Port
 	}
-	cluster.Compressor = gocql.SnappyCompressor{}
+
+	if c.EnableCompression {
+		cluster.Compressor = gocql.SnappyCompressor{}
+	}
+
 	if c.Consistency == "" {
 		cluster.Consistency = gocql.LocalOne
 	} else {
--- a/pkg/cassandra/config/config.go
+++ b/pkg/cassandra/config/config.go
@@ -38,7 +38,7 @@ type Configuration struct {
 	MaxRetryAttempts     int           `validate:"min=0" yaml:"max_retry_attempt"`
 	ProtoVersion         int           `yaml:"proto_version"`
 	Consistency          string        `yaml:"consistency"`
-	EnableCompression    bool          `yaml:"enable-compression"`
+	DisableCompression   bool          `yaml:"disable-compression"`
 	Port                 int           `yaml:"port"`
 	Authenticator        Authenticator `yaml:"authenticator"`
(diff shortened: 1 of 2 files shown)


sagaranand015:
@yurishkuro : Could you review this PR please? 
References: https://github.com/jaegertracing/jaeger/issues/1105 & https://github.com/jaegertracing/jaeger/issues/1667
Thanks

pkg/cassandra/config/config.go
```diff
@@ -38,6 +38,7 @@ type Configuration struct {
 	MaxRetryAttempts     int           `validate:"min=0" yaml:"max_retry_attempt"`
 	ProtoVersion         int           `yaml:"proto_version"`
 	Consistency          string        `yaml:"consistency"`
+	EnableCompression    bool          `yaml:"enable-compression"`
```
yurishkuro: This makes the struct not backwards compatible, because the default value will be false (only initializing from CLI flags would use true as default).

I prefer to call this `DisableCompression`

sagaranand015: @yurishkuro : Sure, disableCompression makes more sense, now that I think about it. Will make the changes. 
Quick question though, the following puts the default value of enableCompression to be true. So, default as true also affects the backend compatibility? 
```suggestion
	flagSet.Bool(
		nsConfig.namespace+suffixEnableCompression,
		true,
		"Enable the use of Compression while connecting to Cassandra API based Storage backend. Uses SnappyCompression by default")
```

yurishkuro: This default only works if flags are used. But the CassandraConfiguration struct can also be used directly, e.g. it's how we do it in our customized binaries at Uber, where configuration is read from yaml files and auto-populated into config structs. So the struct itself should be backwards compatible.

sagaranand015: Thank you for this @yurishkuro 
Made the required changes, please review.

pkg/cassandra/config/config.go
```diff
@@ -128,7 +129,16 @@ func (c *Configuration) NewCluster() *gocql.ClusterConfig {
...
-	cluster.Compressor = gocql.SnappyCompressor{}
+
+	fmt.Println("Disable Compression Flag is: ")
+	fmt.Println(c.DisableCompression)
+	if !c.DisableCompression {
+		cluster.Compressor = gocql.SnappyCompressor{}
+	}
+
+	fmt.Println("Compression is:")
+	fmt.Println(cluster.Compressor)
```
yurishkuro: please remove print statements, we don't do that in Jaeger. I believe at some point we simply log the whole config struct, using the logger, not println.

plugin/storage/cassandra/options.go
```diff
@@ -163,6 +164,10 @@ func addFlags(flagSet *flag.FlagSet, nsConfig *namespaceConfig) {
 		nsConfig.namespace+suffixConsistency,
 		nsConfig.Consistency,
 		"The Cassandra consistency level, e.g. ANY, ONE, TWO, THREE, QUORUM, ALL, LOCAL_QUORUM, EACH_QUORUM, LOCAL_ONE (default LOCAL_ONE)")
+	flagSet.Bool(
+		nsConfig.namespace+suffixDisableCompression,
+		false,
+		"Disables the use of the default Snappy Compression while connecting to the Cassandra Cluster if set to true")
```
yurishkuro: may want to add to description: "This may be useful when using Azure Cosmos DB instead of Cassandra"

pkg/cassandra/config/config.go
sagaranand015: My bad..Removed these in the latest push

sagaranand015:
@yurishkuro : If everything seems ok, could you please merge this?
Thanks

yurishkuro:
All good, I was just waiting for a green build.

PR #1079, opened 2018-09-21
Throttling backend

isaachier:
Signed-off-by: Isaac Hier <ihier@uber.com>

## Which problem is this PR solving?
- Resolves #1078.

## Short description of the changes
- Throttling component in agent.

PR #1676, opened 2019-07-22
Throttling backend

annanay25:
This PR is taking over from #1079 which is the work of @isaachier 

- Resuming work on the throttling backend infrastructure

- Rebased to current master, fixed proto definition layout.


yurishkuro:
@annanay25, just curious, what is your motivation for this work?

I think client-side throttling would be a very useful feature in Jaeger, but the original pr only supported debug traces, not all traffic. It was also, as readme explained, not a real distributed throttler. We needed that feature internally, but in oss we may want to consider a more general solution.

annanay25:
So the primary objective was to have a way to describe connectivity between Jaeger components, relevant issue - #1007. But definitely, the PR can be extended to include all traffic, not just debug spans. 

What would you recommend? @yurishkuro

yurishkuro:
#1007 sounds quite unrelated, nothing to do with throttling (although can share the client->agent registration mechanism, assuming we want to keep it).

For the actual throttling, I would prefer to design a proper solution that works for all sampled traffic, not just debug spans, AND ideally a distributed rate limiting (e.g. like Youtube Doorman), rather than per-agent.

annanay25:
I agree, I'm on-board with implementing a solution that works for all sampled traffic.

PS; just wanted to point out that as per  - https://github.com/jaegertracing/jaeger/issues/1007#issuecomment-414692802 - seems like the consensus was that connectivity could be checked via debug spans which would work on the credit model designed in this PR.

annanay25:
@yurishkuro 
First thought on reading the [design doc](https://github.com/youtube/doorman/blob/master/doc/design.md) of YouTube Doorman - 

A distributed solution like YouTube Doorman would require a central state store to coordinate resource division. The collectors would talk to the state store and run the division algorithms. Would end users be willing to add a dependence (ex: etcd) to use the throttling feature?

jpkrohling:
What's the state of this PR? Looks like it's been without activity for over a year now.

PR #1680, opened 2019-07-24
support jaeger.tags when calling 'GetSamplingStrategy', e.g. return R…

limfriend:
…ATE_LIMITING when jaeger.tags = {region,CA}

SamplingStrategy support for agent level tag
e.g. return RATE_LIMITING when jaeger.tags = {"region","CA"}

transparent  metadata which converted from 'jaeger.tags' to collector service


limfriend:
close for recommit

PR #1685, opened 2019-07-24
Allow token propagation if token type is not specified

rubenvp8510:
- openshift oauth-proxy  (and is possible that other proxies) does not put the authorization header in the form `Authorization: <type> <token> `, they put the token in the form `Authorization: <token>`  and assumed that the token there is a bearer token.

- Allow those kind of tokens to be propagated.

### DIFF
--- a/cmd/query/app/token_propagation_handler.go
+++ b/cmd/query/app/token_propagation_handler.go
@@ -39,8 +39,11 @@ func bearerTokenPropagationHandler(logger *zap.Logger, h http.Handler) http.Hand
 				if headerValue[0] == "Bearer" {
 					token = headerValue[1]
 				}
+			} else if len(headerValue) == 1 {
+				// Tread all value as a token
+				token = authHeaderValue
 			} else {
-				logger.Warn("Invalid authorization header, skipping bearer token propagation")
+				logger.Warn("Invalid authorization header value, skipping token propagation")
 			}
 			h.ServeHTTP(w, r.WithContext(spanstore.ContextWithBearerToken(ctx, token)))
 		} else {
--- a/cmd/query/app/token_propagation_hander_test.go
+++ b/cmd/query/app/token_propagation_hander_test.go
@@ -51,16 +51,19 @@ func Test_bearTokenPropagationHandler(t *testing.T)  {
 	}
 
 	testCases := []struct {
-		name		string
-		sendHeader bool
-		header 		string
-		handler 	func(stop *sync.WaitGroup) http.HandlerFunc
+		name        string
+		sendHeader  bool
+		headerValue string
+		headerName  string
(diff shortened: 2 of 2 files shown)


yurishkuro:
lgtm but I don't think  we want all those log statements

rubenvp8510:
@yurishkuro  ok I can remove it

rubenvp8510:
I remove one, and change to debug others. Is that ok? or do you prefer to get rid of it?

cmd/query/app/token_propagation_handler.go
```diff
@@ -26,21 +27,30 @@ import (
...
+
 		if authHeaderValue != "" {
 			headerValue := strings.Split(authHeaderValue, " ")
 			token := ""
 			if len(headerValue) == 2 {
 				// Make sure we only capture bearer token , not other types like Basic auth.
 				if headerValue[0] == "Bearer" {
 					token = headerValue[1]
+				} else {
+					logger.Debug("Unsupported type of token " + headerValue[0] + " skipping token propagation")
```
yurishkuro: Why do we need to log this?

cmd/query/app/token_propagation_handler.go
```diff
@@ -26,21 +27,30 @@ import (
 func bearerTokenPropagationHandler(logger *zap.Logger, h http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		ctx := r.Context()
+		logger.Info("Propagating bearer token")
```
yurishkuro: Why do we need to log this?

cmd/query/app/token_propagation_handler.go
```diff
@@ -26,21 +27,30 @@ import (
...
+		log.Print(r)
```
yurishkuro: We don't use log package

cmd/query/app/token_propagation_handler.go
```diff
@@ -26,21 +27,30 @@ import (
...
 		authHeaderValue := r.Header.Get("Authorization")
-		// If no Authorization header is present, try with X-Forwarded-Access-Token
+		// If no Authorization headerValue is present, try with X-Forwarded-Access-Token
 		if authHeaderValue == "" {
 			authHeaderValue = r.Header.Get("X-Forwarded-Access-Token")
 		}
+		logger.Info("Token: " + authHeaderValue)
```
yurishkuro: Why do we need to log this?

cmd/query/app/token_propagation_handler.go
rubenvp8510: Ups it was not my intend to have this here.

rubenvp8510:
Done, logs removed, except warning log for unparseable header.

PR #1687, opened 2019-07-25
support jaeger.tags when calling 'GetSamplingStrategy', e.g. return R…

limfriend:
…ATE_LIMITING when jaeger.tags = {region,CA}

Signed-off-by: osfriend <osfriend@163.com>

SamplingStrategy support for agent level tag
e.g. return RATE_LIMITING when jaeger.tags = {"region","CA"}

transparent metadata which converted from 'jaeger.tags' to collector service

### DIFF
--- a/cmd/agent/app/configmanager/grpc/manager.go
+++ b/cmd/agent/app/configmanager/grpc/manager.go
@@ -19,7 +19,7 @@ import (
 	"errors"
 
 	"google.golang.org/grpc"
-
+	"google.golang.org/grpc/metadata"
 	"github.com/jaegertracing/jaeger/model/converter/thrift/jaeger"
 	"github.com/jaegertracing/jaeger/proto-gen/api_v2"
 	"github.com/jaegertracing/jaeger/thrift-gen/baggage"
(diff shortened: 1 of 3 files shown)


limfriend:
@pavolloffay @jpkrohling  could you help to review this PR, I have not idea about CI build failure.

yurishkuro:
Could you please first open a ticket and describe the business use case that you're trying to solve?

limfriend:
@yurishkuro  add issue for this PR, https://github.com/jaegertracing/jaeger/issues/1692

yurishkuro:
Thanks. So it looks like this PR only passes the agent tags as metadata to the collector, it does not actually implement being able to vary sampling strategies in response to those tags.

limfriend:
your are right, It is easy to write code for passing tags to collector . but It is complex to get sampling strategies for composing serviceName and agentTag.  
1. Maybe some agent tags is global for sampling strategies. e.g samplingRate=1 when t 'environment' tag= test.
2  maybe some serviceName  is ignore agent tags for sampling strategies. e.g ‘order’ service is always samplingRate=1， No matter what the  'environment' tag is test or product. 
I have no clean idea about how to config the priority for serviceName and agentTag.  Could you please give me some help?

yurishkuro:
There is a related proposal here https://github.com/jaegertracing/jaeger/issues/365#issuecomment-505913581, which will require representing sampling rules with more granularity than just service-operation as today. That approach is easily generalizable to include agent tags.

limfriend:
yes, it is a good generalizable proposal.  When is it probably done? could you please merge the agent code first, then I extended agentTag Sampling base this  sampling rules.

yurishkuro:
I don't know if @lookfwd made any progress on it. We internally have similar requirements and will likely work on that functionality in September. But we haven't finalized the backend design yet.

I don't want to merge this PR until we have a design for the backend, because it's not guaranteed that we will want to keep the data path in this form. We can keep it open for now.

jpkrohling:
What's the status of this PR?

PR #1690, opened 2019-07-25
Elasticsearch 7 Support

### LINKED ISSUE #1474: Jaeger Ingest for Elasticsearch 7.0 release date
Since 7.0 just came out of Elasticsearch, how long will it typically take for Jaeger Ingest to be updated to 7.0?

gregoryfranklin:
Adds support for Elasticsearch 7.x

Resolves #1474

- Updates github.com/olivere/elastic to 6.2.21
- Removes the deprecated _default_ field from mappings
- Replaces document types with '_doc' as a transition step to removing them entirely
- Sets include_type_name for compatibility between elasticsearch 6.x and 7.x

rest_total_hits_as_int support is also required for compatibility between elasticsearch 6.x and 7.x (waiting for feature to be merged in github.com/olivere/elastic)

### DIFF
--- a/pkg/es/config/config.go
+++ b/pkg/es/config/config.go
@@ -28,7 +28,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/uber/jaeger-lib/metrics"
 	"go.uber.org/zap"
-	"gopkg.in/olivere/elastic.v5"
+	"github.com/olivere/elastic"
 
 	"github.com/jaegertracing/jaeger/pkg/es"
 	"github.com/jaegertracing/jaeger/pkg/es/wrapper"
--- a/pkg/es/wrapper/wrapper.go
+++ b/pkg/es/wrapper/wrapper.go
@@ -17,7 +17,7 @@ package eswrapper
 import (
 	"context"
 
-	"gopkg.in/olivere/elastic.v5"
+	"github.com/olivere/elastic"
 
 	"github.com/jaegertracing/jaeger/pkg/es"
 )
--- a/pkg/es/wrapper/wrapper.go
+++ b/pkg/es/wrapper/wrapper.go
@@ -100,6 +100,11 @@ func (c IndicesCreateServiceWrapper) Body(mapping string) es.IndicesCreateServic
 	return WrapESIndicesCreateService(c.indicesCreateService.Body(mapping))
 }
 
+// IncludeTypeName calls this function to internal service.
+func (c IndicesCreateServiceWrapper) IncludeTypeName(include bool) es.IndicesCreateService {
+	return WrapESIndicesCreateService(c.indicesCreateService.IncludeTypeName(include))
+}
+
 // Do calls this function to internal service.
 func (c IndicesCreateServiceWrapper) Do(ctx context.Context) (*elastic.IndicesCreateResult, error) {
 	return c.indicesCreateService.Do(ctx)
--- a/scripts/travis/es-integration-test.sh
+++ b/scripts/travis/es-integration-test.sh
@@ -2,8 +2,11 @@
 
 set -e
 
-docker pull docker.elastic.co/elasticsearch/elasticsearch:5.6.12
-CID=$(docker run -d -p 9200:9200 -e "http.host=0.0.0.0" -e "transport.host=127.0.0.1" docker.elastic.co/elasticsearch/elasticsearch:5.6.12)
+# Default to elasticsearch 5.6.12 if ES_VERSION is not set
+ES_VERSION=${ES_VERSION:-5.6.12}
+
+docker pull docker.elastic.co/elasticsearch/elasticsearch:${ES_VERSION}
+CID=$(docker run -d -p 9200:9200 -e "http.host=0.0.0.0" -e "transport.host=127.0.0.1" docker.elastic.co/elasticsearch/elasticsearch:${ES_VERSION})
 export STORAGE=elasticsearch
 make storage-integration-test
 docker kill $CID
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -656,6 +656,18 @@
   version = "v1.0.0"
 
 [[projects]]
+  digest = "1:2bf1079dc711508d423f2d1bcb97078f0a015152433b02858e681006fcb1147f"
+  name = "github.com/olivere/elastic"
+  packages = [
+    ".",
+    "config",
+    "uritemplates",
+  ]
+  pruneopts = "UT"
+  revision = "2ce6dee3af14e17dc059746a1de4f983169030e9"
+  version = "v6.2.21"
+
+[[projects]]
   branch = "master"
   digest = "1:bccaead3121ab7964fd80fab704f612e5893fb5a2c581d520ec847ed8cfac27e"
   name = "github.com/opentracing-contrib/go-stdlib"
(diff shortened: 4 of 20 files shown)


pavolloffay:
## Elasticsearch 7 documentation

* https://www.elastic.co/guide/en/elasticsearch/reference/7.x/release-notes-7.0.0.html
* https://www.elastic.co/guide/en/elasticsearch/reference/7.x/breaking-changes-7.0.html
* https://www.elastic.co/guide/en/elasticsearch/reference/current/removal-of-types.html

## I have tested

- [ ] self-provisioned ES in Jaeger operator
- [ ] spark-dependencies
- [ ] Kibana integration with ES7 https://medium.com/jaegertracing/jaeger-elasticsearch-and-kibana-7ecb846137b6 

## Mapping changes:
* Removed "index.mapper.dynamic":false` - https://github.com/elastic/elasticsearch/issues/36711, https://www.elastic.co/guide/en/elasticsearch/reference/current/removal-of-types.html
* Removed `_default_` mapping - https://www.elastic.co/guide/en/elasticsearch/reference/current/removal-of-types.html
* `template` renamed to `index_patterns` - already deprecated in ES6 https://www.elastic.co/guide/en/elasticsearch/reference/6.2/breaking_60_indices_changes.html

## Other changes which affect us

The number of tags and logs is now limited to 10000 by default

> To safeguard against out of memory errors, the number of nested json objects within a single document across all fields has been limited to 10000. This default limit can be changed with the index setting index.mapping.nested_objects.limit.  https://www.elastic.co/guide/en/elasticsearch/reference/7.x/breaking-changes-7.0.html#limit-number-nested-json-objects

scripts/travis/es-integration-test.sh
```diff
@@ -2,10 +2,15 @@
 
 set -e
 
-docker pull docker.elastic.co/elasticsearch/elasticsearch:5.6.12
-CID=$(docker run -d -p 9200:9200 -e "http.host=0.0.0.0" -e "transport.host=127.0.0.1" -e "xpack.security.enabled=false" -e "xpack.monitoring.enabled=false" docker.elastic.co/elasticsearch/elasticsearch:5.6.12)
+run_integration_test() {
+  ES_VERSION=$1
+  docker pull docker.elastic.co/elasticsearch/elasticsearch:${ES_VERSION}
+  CID=$(docker run -d -p 9200:9200 -e "http.host=0.0.0.0" -e "transport.host=127.0.0.1" -e "xpack.security.enabled=false" -e "xpack.monitoring.enabled=false" docker.elastic.co/elasticsearch/elasticsearch:${ES_VERSION})
```
pavolloffay: could you please add `--rm` param?

pkg/es/wrapper/wrapper.go
```diff
@@ -42,12 +48,20 @@ func (c ClientWrapper) IndexExists(index string) es.IndicesExistsService {
 
 // CreateIndex calls this function to internal client.
 func (c ClientWrapper) CreateIndex(index string) es.IndicesCreateService {
-	return WrapESIndicesCreateService(c.client.CreateIndex(index))
+	indicesCreateService := c.client.CreateIndex(index)
+	if c.esVersion == 7 {
+		indicesCreateService = indicesCreateService.IncludeTypeName(true)
```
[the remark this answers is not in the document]
gregoryfranklin: include_type_name is left over from when I was using the same template for all versions.  Now its split into a different template for ES7, it should be possible to remove include_type_name everywhere.

pkg/es/wrapper/wrapper.go
```diff
@@ -58,12 +72,20 @@ func (c ClientWrapper) Index() es.IndexService {
 
 // Search calls this function to internal client.
 func (c ClientWrapper) Search(indices ...string) es.SearchService {
-	return WrapESSearchService(c.client.Search(indices...))
+	searchService := c.client.Search(indices...)
+	if c.esVersion == 7 {
+		searchService = searchService.RestTotalHitsAsInt(true)
```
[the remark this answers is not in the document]
gregoryfranklin: Yes, rest_total_hits_as_int is required whilst using the v6 branch of github.com/olivere/elastic because the result from elasticsearch needs to be unmarshalled into a go struct.
The go struct that comes with github.com/olivere/elastic v6 matches the response from ES5 and ES6 but only matches ES7 if rest_total_hits_as_int is used.

pavolloffay:
@gregoryfranklin the PR needs to be updated and it can go in. @jaegertracing/elasticsearch please also have a look if you are interested in ES7. Or if you could review data migration part I provided in https://github.com/jaegertracing/jaeger/pull/1690#issuecomment-520830821

pavolloffay:
Based on https://github.com/jaegertracing/jaeger/pull/1690#issuecomment-520830821 we could use ES 7 mapping when 6.8 is detected for more seamless update. Users would just deploy with 6.8 and wait until old indices TTLout and are removed and then deploy with ES 7.x.

gregoryfranklin:
I have removed include_type_name.  This also meant removing types from all the search queries.

The token propagation test is currently broken.  The reason for this is that Jaeger will now not start up unless it can reach elasticsearch and determine the ES version, but the token propagation test is not using a real ES server.  I need to think about this a bit more to find a solution.

pavolloffay:
> The token propagation test is currently broken. The reason for this is that Jaeger will now not start up unless it can reach elasticsearch and determine the ES version, but the token propagation test is not using a real ES server. I need to think about this a bit more to find a solution.

We are using a mocked ES server. Just create an endpoint which returns ES version.

pavolloffay:
@gregoryfranklin could you please merge mappings for ES 6 and ES 5 to a single file and just add a new mapping for ES 7. The legacy mapping works well for ES5 and ES6. It seems unnecessary to create duplicate files.

plugin/storage/integration/token_propagation_test.go
```diff
@@ -89,6 +102,19 @@ func TestBearTokenPropagation(t *testing.T) {
 	defer srv.Shutdown(context.Background())
 
 	go createElasticSearchMock(srv, t)
+	// Wait for http server to start
+	time.Sleep(100 * time.Millisecond)
+
+	// Path relative to plugin/storage/integration/token_propagation_test.go
+	cmd := exec.Command("../../../cmd/query/query-linux", "--es.server-urls=http://127.0.0.1:9200", "--es.tls=false", "--query.bearer-token-propagation=true")
```
pavolloffay: Could you please revert this. It's better to run the query from outside the test.

pkg/es/config/config.go
```diff
@@ -154,7 +155,19 @@ func (c *Configuration) NewClient(logger *zap.Logger, metricsFactory metrics.Fac
...
+	// Determine ElasticSearch Version
+	pingResult, _, err := rawClient.Ping(c.Servers[0]).Do(context.Background())
+	if err != nil {
+		return nil, err
+	}
+	esVersion, err := strconv.Atoi(string(pingResult.Version.Number[0]))
+	if err != nil {
+		return nil, err
+	}
+	logger.Sugar().Infof("Detected elasticsearch %d", esVersion)
```
pavolloffay: Change this to structured logging: `logger.Info("Elasticsearch detected", zap.Int("version", esVersion))`

plugin/storage/es/dependencystore/schema.go
```diff
@@ -22,3 +22,17 @@ const dependenciesMapping = `{
       "` + dependencyType + `":{}
    }
 }`
+
+const dependenciesMapping7 = `{
```
pavolloffay: Do we need this? The legacy mapping should work with the ES7. I think you added this after I tested this and it worked without this change. 

If it works please revert this.

plugin/storage/integration/token_propagation_test.go
gregoryfranklin: The query service must be started after the elasticsearch server because the version check is done when the client is initialised (ie on startup).  If you start the query service before running the test, as it was before, the query service exits with an error.

plugin/storage/es/dependencystore/schema.go
gregoryfranklin: The legacy mapping does not work with ES7 because the legacy mapping includes the type name.  Type names are not permitted in ES7 unless you use the include_type_name parameter, which we removed because it is deprecated.

pavolloffay:
I think this looks pretty good, I will do final testing and merge tomorrow.

One thing I don't like is the change to token propagation tests. To fix that we should allow starting Jaeger if the Elasticsearch version cannot be retrieved and just use one template.

PR #1694, opened 2019-07-26
Refactor badger storage dependency parsing for better performance

burmanm:

Currently getting the dependencies window can take a lot of time if there are significant amount of active spans in the badger.

With 1.6 million active spans this took 10 seconds to return on my machine. After this patch, it takes 1.1s on my machine.

* Adds new index for dependency writing (one extra write - but inside same transaction)
* Does not decode the protobufs or touch the values in badger and instead uses key only parsing to find these results.
* Does not store a lot of stuff in the memory, instead it parses them while getting them from badger
* Adds new key to the storage which indicates the used version
* Adds schema migration from version 0 to version 1

### DIFF
--- a/plugin/storage/badger/factory.go
+++ b/plugin/storage/badger/factory.go
@@ -118,6 +118,13 @@ func (f *Factory) Initialize(metricsFactory metrics.Factory, logger *zap.Logger)
 	}
 	f.store = store
 
+	err = badgerStore.SchemaUpdate(store, logger)
+	if err != nil {
+		logger.Error("Failed to update the schema, badger storage failed to start: " + err.Error())
+		store.Close()
+		return err
+	}
+
 	f.cache = badgerStore.NewCacheStore(f.store, f.Options.primary.SpanStoreTTL, true)
 
 	f.metrics.ValueLogSpaceAvailable = metricsFactory.Gauge(metrics.Options{Name: valueLogSpaceAvailableName})
--- a/plugin/storage/badger/factory.go
+++ b/plugin/storage/badger/factory.go
@@ -198,8 +198,6 @@ func (f *Factory) maintenance() {
 			}
 			if err == badger.ErrNoRewrite {
 				f.metrics.LastValueLogCleaned.Update(t.UnixNano())
-			} else {
-				f.logger.Error("Failed to run ValueLogGC", zap.Error(err))
 			}
 
 			f.metrics.LastMaintenanceRun.Update(t.UnixNano())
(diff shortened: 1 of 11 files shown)


plugin/storage/badger/factory.go
```diff
@@ -111,6 +111,13 @@ func (f *Factory) Initialize(metricsFactory metrics.Factory, logger *zap.Logger)
 	}
 	f.store = store
 
+	err = badgerStore.SchemaUpdate(store, logger)
```
jpkrohling: Is it possible to run the badger with collector (rw) in one host and query (read-only) in another? If so, you'd need a way to lock the storage here, so that only one pod will attempt to update the schema.

jpkrohling: Either lock, or a CLI option to to skip the schema migrations.

burmanm: Badger creates a lock when it reads the storage files, thus no two instances of badger can share the files. Also, the updates happen in a transaction -> they either completely succeed or fail (ACID guarantees) - thus preventing any issues with concurrent access in the same process.

So there's no additional locking required. And no, you can't access same storage files from two different hosts as they can't share the memory (which has the current state before it is written to the disk). While I guess WAL polling could be done, the LSM tree's compactions would make it next to impossible. I'm not really aware of any real storage engine that would allow it.

jpkrohling: Thanks for the clarification.

burmanm:
Seems Travis is failing to github timeouts today..

jpkrohling:
This latest failure is actually a real one:

> Go fmt, license check, or import ordering failures, run 'make fmt'

burmanm:
``make fmt`` gives me nothing related to this PR:

```
[michael@nina jaeger]$ git status -s
 M examples/hotrod/services/frontend/gen_assets.go
 M jaeger-ui
[michael@nina jaeger]$ make fmt
./scripts/import-order-cleanup.sh inplace
Running go fmt...
./scripts/updateLicenses.sh
[michael@nina jaeger]$ git status -s
 M examples/hotrod/services/frontend/gen_assets.go
 M jaeger-ui
[michael@nina jaeger]
```

Nor ``make test-ci`` ?

burmanm:
Lets see if rebase fixes Travis

burmanm:
codecov never ceases to amaze me. I'm getting coverage diff rejection because of change in ``processor.go``. Yet, this PR doesn't even have that file.

yurishkuro:
I see most uncovered code coming from storage.go: https://codecov.io/gh/jaegertracing/jaeger/compare/e5f9953251b50643c5c4479cc8941395c75faa76...5f361bd82dbe0f1e18db9f4819ce365510e8d741/diff

jpkrohling:
Closing due to inactivity.

PR #1701, opened 2019-07-30

chandresh-pancholi:
- Improve configuration parameters for Kafka #1359

- This PR adds Kafka producer configs like Compression & CompressionLevel. To get the CompressionLevel, it uses https://cwiki.apache.org/confluence/display/KAFKA/KIP-390%3A+Allow+fine-grained+configuration+for+compression as reference.

### DIFF
--- a/pkg/cassandra/config/config.go
+++ b/pkg/cassandra/config/config.go
@@ -38,6 +38,7 @@ type Configuration struct {
 	MaxRetryAttempts     int           `validate:"min=0" yaml:"max_retry_attempt"`
 	ProtoVersion         int           `yaml:"proto_version"`
 	Consistency          string        `yaml:"consistency"`
+	DisableCompression   bool          `yaml:"disable-compression"`
 	Port                 int           `yaml:"port"`
 	Authenticator        Authenticator `yaml:"authenticator"`
 	DisableAutoDiscovery bool          `yaml:"disable_auto_discovery"`
@@ -128,7 +129,11 @@ func (c *Configuration) NewCluster() *gocql.ClusterConfig {
 	if c.Port != 0 {
 		cluster.Port = c.Port
 	}
-	cluster.Compressor = gocql.SnappyCompressor{}
+
+	if !c.DisableCompression {
+		cluster.Compressor = gocql.SnappyCompressor{}
+	}
+
 	if c.Consistency == "" {
 		cluster.Consistency = gocql.LocalOne
 	} else {
--- a/plugin/storage/cassandra/options.go
+++ b/plugin/storage/cassandra/options.go
@@ -37,6 +37,7 @@ const (
 	suffixKeyspace             = ".keyspace"
 	suffixDC                   = ".local-dc"
 	suffixConsistency          = ".consistency"
+	suffixDisableCompression   = ".disable-compression"
 	suffixProtoVer             = ".proto-version"
 	suffixSocketKeepAlive      = ".socket-keep-alive"
 	suffixUsername             = ".username"
@@ -163,6 +164,10 @@ func addFlags(flagSet *flag.FlagSet, nsConfig *namespaceConfig) {
 		nsConfig.namespace+suffixConsistency,
 		nsConfig.Consistency,
 		"The Cassandra consistency level, e.g. ANY, ONE, TWO, THREE, QUORUM, ALL, LOCAL_QUORUM, EACH_QUORUM, LOCAL_ONE (default LOCAL_ONE)")
+	flagSet.Bool(
+		nsConfig.namespace+suffixDisableCompression,
+		false,
+		"Disables the use of the default Snappy Compression while connecting to the Cassandra Cluster if set to true. This is useful for connecting to Cassandra Clusters(like Azure Cosmos Db with Cassandra API) that do not support SnappyCompression")
 	flagSet.Int(
 		nsConfig.namespace+suffixProtoVer,
 		nsConfig.ProtoVersion,
@@ -243,6 +248,7 @@ func (cfg *namespaceConfig) initFromViper(v *viper.Viper) {
 	cfg.TLS.ServerName = v.GetString(cfg.namespace + suffixServerName)
 	cfg.TLS.EnableHostVerification = v.GetBool(cfg.namespace + suffixVerifyHost)
 	cfg.EnableDependenciesV2 = v.GetBool(cfg.namespace + suffixEnableDependenciesV2)
+	cfg.DisableCompression = v.GetBool(cfg.namespace + suffixDisableCompression)
 }
 
 // GetPrimary returns primary configuration.


plugin/storage/kafka/options.go
```diff
@@ -94,3 +125,34 @@ func (opt *Options) InitFromViper(v *viper.Viper) {
 func stripWhiteSpace(str string) string {
 	return strings.Replace(str, " ", "", -1)
 }
+
+// getCompressionLevel to get compression level from compression type
+func getCompressionLevel(compression, compressionLevel int) (int, error) {
+	switch sarama.CompressionCodec(compression) {
+	case sarama.CompressionGZIP:
+		if compressionLevel == 0 {
+			return 6, nil
```
yurishkuro: what are these magic numbers? Doesn't sarama define constants for them?

plugin/storage/kafka/options.go
```diff
@@ -74,15 +83,37 @@ func (opt *Options) AddFlags(flagSet *flag.FlagSet) {
 		defaultEncoding,
 		fmt.Sprintf(`Encoding of spans ("%s" or "%s") sent to kafka.`, EncodingJSON, EncodingProto),
 	)
+	flagSet.Int(
+		configPrefix+suffixRequiredAcks,
+		defaultRequiredAcks,
+		"(experimental) Required kafka broker acknowledgement. default = 1, no response = 0, wait for local = 1, wait for all = -1",
+	)
+	flagSet.Int(
```
yurishkuro: I would prefer this to be a string field with readable values, rather than cryptic numbers

plugin/storage/kafka/options.go
```diff
@@ -74,15 +83,37 @@ func (opt *Options) AddFlags(flagSet *flag.FlagSet) {
 		defaultEncoding,
 		fmt.Sprintf(`Encoding of spans ("%s" or "%s") sent to kafka.`, EncodingJSON, EncodingProto),
 	)
+	flagSet.Int(
```
yurishkuro: can this be a string field with readable values? e.g. noack, local, all

plugin/storage/kafka/options.go
```diff
@@ -74,15 +83,37 @@ func (opt *Options) AddFlags(flagSet *flag.FlagSet) {
...
+	)
+	flagSet.Int(
+		configPrefix+suffixCompression,
+		defaultCompression,
+		"(experimental) Type of compression to use on messages.  none = 0, gzip = 1, snappy = 2, lz4 = 3, zstd = 4",
+	)
+	flagSet.Int(
+		configPrefix+suffixCompressionLevel,
+		defaultCompressionLevel,
+		"(experimental) Level of compression to use on messages. default= -1000, gzip = 1-9 (default = 6), snappy = none, lz4 = 1-17 (default = 9), zstd = -131072 - 22 (default = 3)",
```
yurishkuro: the numbers make almost no sense. If it's a "level of compression", can we normalize it to a single scale, e.g. 1 to 9 (lowest - highest)?

yurishkuro: why is default -1000?  

also, Viper already  prints the default value in parantheses

plugin/storage/kafka/options.go
```diff
@@ -33,14 +35,21 @@ const (
...
-	suffixProtocolVersion = ".protocol-version"
-	suffixEncoding        = ".encoding"
-	defaultBroker         = "127.0.0.1:9092"
-	defaultTopic          = "jaeger-spans"
-	defaultEncoding       = EncodingProto
+	configPrefix           = "kafka.producer"
+	suffixBrokers          = ".brokers"
+	suffixTopic            = ".topic"
+	suffixEncoding         = ".encoding"
+	suffixRequiredAcks     = ".required.acks"
```
yurishkuro: this should be `required-acks` (with a dash, not a dot, because it's not a hierarchical struct, it's a single term)

plugin/storage/kafka/options.go
chandresh-pancholi: Sarama doesn't define these constants. These numbers have been taken from Kafka. I have mentioned the link in PR.

chandresh-pancholi: Screenshot 2019-07-31 at 1 25 29 PM

plugin/storage/kafka/options.go
chandresh-pancholi: done

plugin/storage/kafka/options.go
chandresh-pancholi: done

plugin/storage/kafka/options.go
chandresh-pancholi: fixed

plugin/storage/kafka/options.go
chandresh-pancholi: fixed

plugin/storage/kafka/options.go
```diff
@@ -33,19 +35,41 @@ const (
...
+	defaultCompressionLevel = 0
+	suffixProtocolVersion   = ".protocol-version"
 )
 
 var (
 	// AllEncodings is a list of all supported encodings.
 	AllEncodings = []string{EncodingJSON, EncodingProto, EncodingZipkinThrift}
+
+	//CompressionModes is a mapping of supported CompressionType to compressionCodec
+	CompressionModes = map[string]sarama.CompressionCodec{
```
yurishkuro: do these constants need to be public? If not, let's use lowercase names.

plugin/storage/kafka/options.go
```diff
@@ -33,19 +35,41 @@ const (
...
+		"none":   sarama.CompressionNone,
+		"gzip":   sarama.CompressionGZIP,
+		"lz4":    sarama.CompressionLZ4,
+		"snappy": sarama.CompressionSnappy,
```
yurishkuro: I would recommend this instead, so that you don't have to `switch` later for default compression level
```go
var compressionModes = map[string]struct {
	compressor   sarama.CompressionCodec
	defaultLevel int
}{
	"none": {
		compressor:   sarama.CompressionNone,
		defaultLevel: 0,
	},
	"gzip": {
		compressor:   sarama.CompressionGZIP,
		defaultLevel: 6,
	},
}
```

or even extend it with min/max

plugin/storage/kafka/options.go
```diff
@@ -74,15 +98,49 @@ func (opt *Options) AddFlags(flagSet *flag.FlagSet) {
 		defaultEncoding,
 		fmt.Sprintf(`Encoding of spans ("%s" or "%s") sent to kafka.`, EncodingJSON, EncodingProto),
 	)
+	flagSet.String(
+		configPrefix+suffixRequiredAcks,
+		defaultRequiredAcks,
+		"(experimental) Required kafka broker acknowledgement. i.e. noack, local, all",
```
yurishkuro: to keep consistent with above:
```
"(experimental) Required kafka broker acknowledgement (noack, local, all)"
```

plugin/storage/kafka/options.go
```diff
@@ -74,15 +98,49 @@ func (opt *Options) AddFlags(flagSet *flag.FlagSet) {
...
+	)
+	flagSet.String(
+		configPrefix+suffixCompression,
+		defaultCompression,
+		"(experimental) Type of compression to use on messages. i.e none, gzip, snappy, lz4, zstd",
```
yurishkuro: to keep consistent with above:
```
"(experimental) Type of compression (none, gzip, snappy, lz4, zstd) to use on messages"
```

plugin/storage/kafka/options.go
```diff
@@ -74,15 +98,49 @@ func (opt *Options) AddFlags(flagSet *flag.FlagSet) {
...
+	)
+	flagSet.Int(
+		configPrefix+suffixCompressionLevel,
+		defaultCompressionLevel,
+		"(experimental) compression level to use on messages. gzip = 1-9 (default = 6), snappy = none, lz4 = 1-17 (default = 9), zstd = -131072 - 22 (default = 3)",
```
yurishkuro: maybe just include the URL where these are defined. Or construct the description from the map[]struct defined above, so that they are always in sync.

plugin/storage/kafka/options.go
yurishkuro: please include the link in the comments. And this code can be largely simplified by using a map[string]struct I suggested above.

yurishkuro:
did you mean to close this?

chandresh-pancholi:
Unfortunately, It got closed by mistake. I am creating another PR.
Apologies for the mishap.

PR #1704, opened 2019-08-01
Export expvar metrics of badger to the metricsFactory

burmanm:

Badger's internal metrics are not available through jaeger's metrics interface. This adds them so they can be monitored.

Copy the metric definitions from expvar and then sync them manually in a background thread. I have a nagging feeling this can't be the best solution though, but I didn't find metricsFactory to have "external collectors" that would fetch the data from there.


plugin/storage/badger/factory.go
```diff
@@ -150,7 +161,8 @@ func (f *Factory) CreateDependencyReader() (dependencystore.Reader, error) {
 
 // Close Implements io.Closer and closes the underlying storage
 func (f *Factory) Close() error {
-	f.maintenanceDone <- true
+	f.maintenanceDone <- true // maintenance close
+	f.maintenanceDone <- true // metrics close
```
jpkrohling: Shouldn't it be `metricsTicker`?

burmanm: No, the ticker is closed when the select ends.

plugin/storage/badger/factory_test.go
```diff
@@ -144,3 +144,33 @@ func TestMaintenanceCodecov(t *testing.T) {
...
+	mFactory := metricstest.NewFactory(0)
+	f.Initialize(mFactory, zap.NewNop())
+	assert.NotNil(t, f.metrics.badgerMetrics)
+	_, found := f.metrics.badgerMetrics["badger_memtable_gets_total"]
+	assert.True(t, found)
+
+	waiter := func(previousValue int64) int64 {
+		sleeps := 0
+		_, gs := mFactory.Snapshot()
+		for gs["adger_memtable_gets_total"] == previousValue && sleeps < 8 {
```
yurishkuro: why "adger"?

yurishkuro:
so badger code does not use any abstraction for metrics, only writes directly to expvar?

plugin/storage/badger/factory_test.go
burmanm: Typo. It always sleeps the maximum with that typo, so the test functions correctly - but might take too long. I'll fix that.

burmanm:
Yes, the badger's metric definitions are here:

https://github.com/dgraph-io/badger/blob/master/y/metrics.go

And they're written like this:

https://github.com/dgraph-io/badger/blob/master/db.go#L1009

plugin/storage/badger/factory.go
yurishkuro: why are we writing twice to the same channel?  If it's only meant to be read-once for exit signal, then I think the usual practice for that is to close the channel.

plugin/storage/badger/factory.go
```diff
@@ -190,3 +205,56 @@ func (f *Factory) maintenance() {
...
+				if strings.HasPrefix(kv.Key, "badger") {
+					if intVal, ok := kv.Value.(*expvar.Int); ok {
+						if g, found := f.metrics.badgerMetrics[kv.Key]; found {
+							g.Update(intVal.Value())
+						}
+					} else if mapVal, ok := kv.Value.(*expvar.Map); ok {
+						mapVal.Do(func(innerKv expvar.KeyValue) {
+							// The metrics we're interested in have only a single inner key (dynamic name)
+							// and we're only interested in its value
+							if intVal, ok := innerKv.Value.(*expvar.Int); ok {
```
yurishkuro: this is catching the metrics declared like this?

```go
NumLSMGets = expvar.NewMap("badger_lsm_level_gets_total")
```

Why does Badger use a map in this case? It looks like it could have more than one entry in the map, but we're collapsing all under the map's name.

plugin/storage/badger/factory.go
```diff
@@ -190,3 +205,56 @@ func (f *Factory) maintenance() {
...
+func (f *Factory) registerBadgerExpvarMetrics(metricsFactory metrics.Factory) {
+	f.metrics.badgerMetrics = make(map[string]metrics.Gauge)
+
+	expvar.Do(func(kv expvar.KeyValue) {
+		if strings.HasPrefix(kv.Key, "badger") {
+			if _, ok := kv.Value.(*expvar.Int); ok {
+				g := metricsFactory.Gauge(metrics.Options{Name: kv.Key})
+				f.metrics.badgerMetrics[kv.Key] = g
+			} else if mapVal, ok := kv.Value.(*expvar.Map); ok {
+				mapVal.Do(func(innerKv expvar.KeyValue) {
```
yurishkuro: wouldn't this map always be empty at this point?

plugin/storage/badger/factory.go
yurishkuro: btw, if multiple values are possible, we could put that value into a tag on the metric.

plugin/storage/badger/factory_test.go
```diff
@@ -144,3 +144,33 @@ func TestMaintenanceCodecov(t *testing.T) {
...
+	f := NewFactory()
+	v, command := config.Viperize(f.AddFlags)
+	command.ParseFlags([]string{
+		"--badger.metrics-update-interval=10ms",
+	})
+	f.InitFromViper(v)
+	mFactory := metricstest.NewFactory(0)
+	f.Initialize(mFactory, zap.NewNop())
+	assert.NotNil(t, f.metrics.badgerMetrics)
+	_, found := f.metrics.badgerMetrics["badger_memtable_gets_total"]
```
yurishkuro: would be good to see a test for the expvar.Map metrics as well.

plugin/storage/badger/factory.go
burmanm: Currently it is storing in the map the directory of the LSM or Value. We have that information in the Factory properties also, but it might indeed make sense to export that information somewhere. Unless there's a place to view properties already in Jaeger?

plugin/storage/badger/factory.go
burmanm: No, since badger.Open() is called before the metrics are fetched. And badger.Open() registers the expvars (it calculates the initial sizes during Open()).

plugin/storage/badger/factory.go
burmanm: With the tags it would look like this:

```
# HELP jaeger_badger_lsm_size_bytes badger_lsm_size_bytes
# TYPE jaeger_badger_lsm_size_bytes gauge
jaeger_badger_lsm_size_bytes{directory="/tmp/badger043433444"} 0
```

yurishkuro: as long as there's a guarantee that map has only one entry, it makes more sense to go without the dir name in the label (since it might be different after restart)

PR #1712, opened 2019-08-06

### LINKED ISSUE #1359: Improve configuration parameters for Kafka

chandresh-pancholi:
- Resolves #1359

- Add RequiredAck, Compressor, and Compression Level params


chandresh-pancholi:
@yurishkuro, I have made all the required changes along with the test. Kindly review.

plugin/storage/kafka/options.go
```diff
@@ -33,19 +35,69 @@ const (
...
+	suffixCompression      = ".compression"
+	suffixCompressionLevel = ".compression.level"
+
+	defaultBroker           = "127.0.0.1:9092"
+	defaultTopic            = "jaeger-spans"
+	defaultEncoding         = EncodingProto
+	defaultRequiredAcks     = "local"
+	defaultCompression      = "none"
+	defaultCompressionLevel = -1000
+	suffixProtocolVersion   = ".protocol-version"
```
yurishkuro: please keep `suffix` constants together

plugin/storage/kafka/options.go
```diff
@@ -94,3 +181,36 @@ func (opt *Options) InitFromViper(v *viper.Viper) {
 func stripWhiteSpace(str string) string {
 	return strings.Replace(str, " ", "", -1)
 }
+
+// getCompressionLevel to get compression level from compression type
+func getCompressionLevel(compressionMode string, compressionLevel int) (int, error) {
+	compressionModeData := compressionModes[compressionMode]
+
+	if compressionLevel == defaultCompressionLevel {
```
yurishkuro: this is confusing, I suggest `if compressionLevel == 0 {`, which makes it obvious that default type's default value is being converted.

plugin/storage/kafka/options.go
```diff
@@ -94,3 +181,36 @@ func (opt *Options) InitFromViper(v *viper.Viper) {
 func stripWhiteSpace(str string) string {
 	return strings.Replace(str, " ", "", -1)
 }
+
+// getCompressionLevel to get compression level from compression type
+func getCompressionLevel(compressionMode string, compressionLevel int) (int, error) {
+	compressionModeData := compressionModes[compressionMode]
```
yurishkuro: in the other functions you use ToLower, should it be consistent?

plugin/storage/kafka/options.go
```diff
@@ -94,3 +181,36 @@ func (opt *Options) InitFromViper(v *viper.Viper) {
...
+
+	if compressionLevel == defaultCompressionLevel {
+		return compressionModeData.defaultCompressionLevel, nil
+	}
+
+	if compressionModeData.minCompressionLevel <= compressionLevel && compressionModeData.maxCompressionLevel >= compressionLevel {
```
yurishkuro: I suggest reversing the condition and returning an error "compression level is not within valid range"

plugin/storage/kafka/options_test.go
```diff
@@ -45,4 +53,144 @@ func TestFlagDefaults(t *testing.T) {
...
+	val, ok = compressionModes["lz4"]
+	assert.Equal(t, true, ok)
+	assert.Equal(t, val.compressor, sarama.CompressionLZ4)
+
+	val, ok = compressionModes["snappy"]
+	assert.Equal(t, true, ok)
+	assert.Equal(t, val.compressor, sarama.CompressionSnappy)
+}
+
+func TestRequiredAcks(t *testing.T) {
```
yurishkuro: I don't think this test is useful, it just tests that map[] works as expected

plugin/storage/kafka/options_test.go
```diff
@@ -45,4 +53,144 @@ func TestFlagDefaults(t *testing.T) {
...
+
+func TestGetRequiredAcksFailures(t *testing.T) {
+	_, err := getRequiredAcks("test")
+	require.Error(t, err)
+
+	_, err = getRequiredAcks("temp")
+	require.Error(t, err)
+}
+
+func TestCompressionModes(t *testing.T) {
```
yurishkuro: I don't think this test is useful, it just tests that map[] works as expected

plugin/storage/kafka/options_test.go
```diff
@@ -45,4 +53,144 @@ func TestFlagDefaults(t *testing.T) {
...
+
+	_, err = getCompressionLevel("lz4", 18)
+	require.Error(t, err)
+
+	_, err = getCompressionLevel("zstd", 25)
+	require.Error(t, err)
+}
+
+func TestGetCompressionModes(t *testing.T) {
+	compressionModes, err := getCompressionModes("gzip")
```
yurishkuro: s/compressionModes/compression/

plugin/storage/kafka/options.go
chandresh-pancholi: This is a snippet from sarama library. this is the reason why I made defaultCompressionLevel = -1000 
```
// CompressionLevelDefault is the constant to use in CompressionLevel
// to have the default compression level for any codec. The value is picked
// that we don't use any existing compression levels.
const CompressionLevelDefault = -1000
```

plugin/storage/kafka/options.go
chandresh-pancholi: fixed

plugin/storage/kafka/options.go
chandresh-pancholi: fixed

plugin/storage/kafka/options.go
chandresh-pancholi: fixed

plugin/storage/kafka/options_test.go
chandresh-pancholi: fixed

plugin/storage/kafka/options_test.go
chandresh-pancholi: fixed

plugin/storage/kafka/options_test.go
chandresh-pancholi: fixed

plugin/storage/kafka/options.go
yurishkuro: that's not the point. The meaning of "default" in sarama is different from what we have here. When dealing with CLI, if you want to use the default value for the respective compressor you simply not specify compression-level flag. Both 0 and -1000 are invalid values for most compressors except for zstd, where BOTH are valid and neither is the default. So I don't see a reason why this bad parameter design from sarama should be carried over to our flags. 0 is the natural null value for int field. It makes no sense to be playing tricks with -1000, which achieve the same results.

plugin/storage/kafka/options.go
```diff
@@ -94,3 +181,37 @@ func (opt *Options) InitFromViper(v *viper.Viper) {
...
+// getCompressionLevel to get compression level from compression type
+func getCompressionLevel(mode string, compressionLevel int) (int, error) {
+	compressionModeData := compressionModes[mode]
+
+	if compressionLevel == defaultCompressionLevel {
+		return compressionModeData.defaultCompressionLevel, nil
+	}
+
+	if compressionModeData.minCompressionLevel > compressionLevel || compressionModeData.maxCompressionLevel < compressionLevel {
+		return 0, fmt.Errorf("compression level is not within valid range")
```
yurishkuro: could make the error more description:
```
fmt.Errorf("compression level %d for '%s' is not within valid range [%d, %d]", level, mode, min, max)
```

plugin/storage/kafka/options.go
```diff
@@ -94,3 +181,37 @@ func (opt *Options) InitFromViper(v *viper.Viper) {
...
+	}
+
+	return compressionLevel, nil
+}
+
+//getCompressionModes maps input modes to sarama CompressionCodec
+func getCompressionModes(mode string) (sarama.CompressionCodec, error) {
```
yurishkuro: s/getCompressionModes/getCompressionMode/ (it returns a singular value, not plural)

plugin/storage/kafka/options_test.go
yurishkuro: fixed?  I'm still seeing `compressionModes` variable, whereas it's a single `compression` value

plugin/storage/kafka/options_test.go
```diff
@@ -45,4 +53,109 @@ func TestFlagDefaults(t *testing.T) {
...
+	assert.Equal(t, sarama.CompressionSnappy, compressionModes)
+
+	compressionModes, err = getCompressionModes("none")
+	require.NoError(t, err)
+	assert.Equal(t, sarama.CompressionNone, compressionModes)
+}
+
+func TestCompressionModesFailures(t *testing.T) {
+	_, err := getCompressionModes("test")
+	require.Error(t, err)
```
yurishkuro: nit: this should be assert, not require (you CAN continue the test)

plugin/storage/kafka/options_test.go
yurishkuro: fixed how? what is it testing? It's just a map lookup.

plugin/storage/kafka/options.go
chandresh-pancholi: I agree with you, i will make the changes

chandresh-pancholi: It's been fixed.

plugin/storage/kafka/options.go
chandresh-pancholi: fixed

plugin/storage/kafka/options_test.go
chandresh-pancholi: My bad. I have removed it now.

plugin/storage/kafka/options.go
chandresh-pancholi: fixed

plugin/storage/kafka/options_test.go
chandresh-pancholi: fixed

chandresh-pancholi:
@yurishkuro Kindly review.

plugin/storage/kafka/options_test.go
```diff
@@ -45,4 +53,92 @@ func TestFlagDefaults(t *testing.T) {
...
+	compressionModes, err = getCompressionMode("none")
+	require.NoError(t, err)
+	assert.Equal(t, sarama.CompressionNone, compressionModes)
+}
+
+func TestCompressionModeFailures(t *testing.T) {
+	_, err := getCompressionMode("test")
+	assert.Error(t, err)
+
+	_, err = getCompressionMode("temp")
```
objectiser: Why testing `test` and `temp` - isn't this one redundant?

plugin/storage/kafka/options_test.go
```diff
@@ -45,4 +53,92 @@ func TestFlagDefaults(t *testing.T) {
...
+	assert.Error(t, err)
+}
+
+func TestRequiredAcksFailures(t *testing.T) {
```
objectiser: What about testing the valid values aswell?

Also same comment as above related to `test` and `temp`?

yurishkuro:
NB: the code coverage drop is only from error clauses.

image

chandresh-pancholi:
@yurishkuro How do you propose to test it? I am not able to figure it out.

plugin/storage/kafka/options_test.go
chandresh-pancholi: fixed it. Also added test for valid values.

plugin/storage/kafka/options_test.go
chandresh-pancholi: fixed it.

plugin/storage/kafka/options.go
```diff
@@ -94,3 +181,37 @@ func (opt *Options) InitFromViper(v *viper.Viper) {
 func stripWhiteSpace(str string) string {
 	return strings.Replace(str, " ", "", -1)
 }
+
+// getCompressionLevel to get compression level from compression type
+func getCompressionLevel(mode string, compressionLevel int) (int, error) {
+	compressionModeData := compressionModes[mode]
+
```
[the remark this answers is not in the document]
chandresh-pancholi: fixed it.

plugin/storage/kafka/options.go
```diff
@@ -33,19 +35,69 @@ const (
...
-	defaultBroker         = "127.0.0.1:9092"
-	defaultTopic          = "jaeger-spans"
-	defaultEncoding       = EncodingProto
+	configPrefix           = "kafka.producer"
+	suffixBrokers          = ".brokers"
+	suffixTopic            = ".topic"
+	suffixEncoding         = ".encoding"
+	suffixRequiredAcks     = ".required-acks"
+	suffixCompression      = ".compression"
+	suffixCompressionLevel = ".compression.level"
```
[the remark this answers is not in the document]
chandresh-pancholi: done

plugin/storage/kafka/options_test.go
```diff
@@ -28,12 +30,18 @@ func TestOptionsWithFlags(t *testing.T) {
 	command.ParseFlags([]string{
 		"--kafka.producer.topic=topic1",
 		"--kafka.producer.brokers=127.0.0.1:9092, 0.0.0:1234",
-		"--kafka.producer.encoding=protobuf"})
+		"--kafka.producer.encoding=protobuf",
+		"--kafka.producer.required-acks=local",
+		"--kafka.producer.compression=gzip",
+		"--kafka.producer.compression.level=6"})
```
objectiser: This should have been changed to `compression-level` - so not sure why the unit tests seems to be passing atm?

chandresh-pancholi: Such a stupid mistake from my side. :(
fixed and pushed it.

plugin/storage/kafka/options_test.go
```diff
@@ -28,12 +31,18 @@ func TestOptionsWithFlags(t *testing.T) {
 	command.ParseFlags([]string{
 		"--kafka.producer.topic=topic1",
 		"--kafka.producer.brokers=127.0.0.1:9092, 0.0.0:1234",
-		"--kafka.producer.encoding=protobuf"})
+		"--kafka.producer.encoding=protobuf",
+		"--kafka.producer.required-acks=local",
+		"--kafka.producer.compression=gzip",
+		"--kafka.producer.compression-level=6"})
```
objectiser: You need to change the value, as 6 is the default for gzip - so regardless of whether this flag is present, the test would currently pass.

chandresh-pancholi: done.

objectiser:
@chandresh-pancholi To address the code coverage, you should create some additional tests, variations of `TestOptionsWithFlags`, where the flags exhibit some error values.

chandresh-pancholi:
@objectiser Thank you so much for the review. I didn't know about the squashing commit. I will keep that in mind in future commits.

chandresh-pancholi:
@objectiser I agree but how to test `log.Fatal`? It's easy to reach to that line but assertion won't help here. What's your suggestion? I think of below code just to increase the code coverage.

```
func TestOptionsWithFlagsWithInvalidRequiredAcks(t *testing.T)  {
	opts := &Options{}
	v, command := config.Viperize(opts.AddFlags)
	command.ParseFlags([]string{
		"--kafka.producer.topic=topic1",
		"--kafka.producer.brokers=127.0.0.1:9092, 0.0.0:1234",
		"--kafka.producer.encoding=protobuf",
		"--kafka.producer.required-acks=test",
		"--kafka.producer.compression=gzip",
		"--kafka.producer.compression-level=7"})
	opts.InitFromViper(v)
}
```

chandresh-pancholi:
is this PR mergable to master?

objectiser:
@chandresh-pancholi Good point - possibly the best way would be to return the error and handle up the stack? From quick look the other storage plugins don't seem to validate the options there.

@yurishkuro Thoughts?

PR #1716, opened 2019-08-07
Add NOTICE

yurishkuro:
cc @objectiser @mattklein123

modeled after 
 * https://github.com/envoyproxy/envoy/blob/master/NOTICE
 * https://github.com/prometheus/prometheus/blob/master/NOTICE

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -1,0 +1,7 @@
+Jaeger, Distributed Tracing Platform.
+
+Copyright 2015-2019 The Jaeger Project Authors
+
+Licensed under Apache License 2.0.  See LICENSE for terms.
+
+Includes software developed at Uber Technologies, Inc. (https://eng.uber.com/).


NOTICE
```diff
@@ -0,0 +1,7 @@
+Jaeger, Distributed Tracing Platform.
+
+Copyright 2015-2019 The Jaeger Project Authors
+
+Licensed under Apache License 2.0.  See LICENSE for terms.
+
+Includes software developed at Uber Technologies, Inc. (https://eng.uber.com/).
```
mattklein123: @caniszczyk WDYT here? When we transferred Envoy we relicensed/copyrighted everything. I just want to make sure we won't run into any issues in the future.

yurishkuro:
@mattklein123 PR #1738 added "The Jaeger Authors" copyright to all files, while keeping the original Uber copyright where it already existed. This is in accordance with the CNCF guidelines https://github.com/cncf/foundation/blob/master/copyright-notices.md

PR #1721, opened 2019-08-08
Fix badger merge-join algorithm to correctly filter indexes

### LINKED ISSUE #1719: Badger index merge is working incorrectly

While doing some additional tests, I managed to create a case where the end result that used multiple index queries was not correct.

Merge-join processing has bugs.

* Missing one advance in the equality compare
* Input for the next round should be the output of previous invocation (in same sorted order)

Working on it, adding also some more tests also to verify more complex index merge scenarios.

burmanm:

Resolves #1719, the index seeks were not correctly merged and filtered. 

Make the merge-join correctly update two indices when encountering equal items. Also, the input must be the output of previous merge. Also, changed ASC to DESC reversing to happen after the top query filtering - thus reducing unnecessary work.

### DIFF
--- a/plugin/storage/badger/spanstore/read_write_test.go
+++ b/plugin/storage/badger/spanstore/read_write_test.go
@@ -161,6 +161,11 @@ func TestIndexSeeks(t *testing.T) {
 							VStr:  fmt.Sprintf("val%d", j),
 							VType: model.StringType,
 						},
+						{
+							Key:   "error",
+							VType: model.BoolType,
+							VBool: true,
+						},
 					},
 				}
 				err := sw.WriteSpan(&s)
@@ -200,6 +205,7 @@ func TestIndexSeeks(t *testing.T) {
 		params.OperationName = "operation-1"
 		tags := make(map[string]string)
 		tags["k11"] = "val0"
+		tags["error"] = "true"
 		params.Tags = tags
 		params.DurationMin = time.Duration(1 * time.Millisecond)
 		// params.DurationMax = time.Duration(1 * time.Hour)
--- a/plugin/storage/badger/spanstore/reader.go
+++ b/plugin/storage/badger/spanstore/reader.go
@@ -346,7 +346,13 @@ func (r *TraceReader) durationQueries(query *spanstore.TraceQueryParameters, ids
 }
 
 func mergeJoinIds(left, right [][]byte) [][]byte {
-	merged := make([][]byte, 0, len(left)) // len(left) or len(right) is the maximum, whichever is smallest
+	// len(left) or len(right) is the maximum, whichever is the smallest
+	allocateSize := len(left)
+	if len(right) < allocateSize {
+		allocateSize = len(right)
+	}
+
+	merged := make([][]byte, 0, allocateSize)
 
 	lMax := len(left) - 1
 	rMax := len(right) - 1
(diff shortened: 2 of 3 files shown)


plugin/storage/badger/spanstore/reader.go
```diff
@@ -346,53 +345,60 @@ func (r *TraceReader) durationQueries(query *spanstore.TraceQueryParameters, ids
...
 		}
-
 	} else {
-		// mergeIntersected should be reversed intersected
-		for i, j := 0, len(intersected)-1; j >= 0; i, j = i+1, j-1 {
-			mergeIntersected = append(mergeIntersected, intersected[j])
-		}
-		intersected = mergeIntersected
+		merged = ids[0]
+
```
pavolloffay: nit: remove line?

plugin/storage/badger/spanstore/reader.go
```diff
@@ -346,53 +345,60 @@ func (r *TraceReader) durationQueries(query *spanstore.TraceQueryParameters, ids
 	return ids
 }
 
+func mergeJoinIds(left, right [][]byte) [][]byte {
+	merged := make([][]byte, 0, len(left)) // len(left) or len(right) is the maximum, whichever is smallest
```
pavolloffay: Why not allocate the smallest and use the size also in the following loop?

plugin/storage/badger/spanstore/reader.go
```diff
@@ -346,53 +345,60 @@ func (r *TraceReader) durationQueries(query *spanstore.TraceQueryParameters, ids
 	return ids
 }
 
+func mergeJoinIds(left, right [][]byte) [][]byte {
```
pavolloffay: rename to `mergeEqualIds` ?

pavolloffay: Are the ids sorted? Maybe that should be documented somewhere.

burmanm: It's mentioned at the beginning of the package. Everything is sorted (it's a sorted K/V).

burmanm: As for the name, it's because the algorithm is called "sort-merge join" and is used in relational databases. Here the sorting phase happens in the DB and the merge phase in this code. It's pretty descriptive in my opinion since if someone wants to improve this method such as doing it parallel or using sharding from multiple badgers there are known algorithms for those variations too (which would underneath use this in any case).

pavolloffay: Thanks for the explanation. Now it rings a bell..

plugin/storage/badger/spanstore/read_write_test.go
```diff
@@ -200,6 +205,7 @@ func TestIndexSeeks(t *testing.T) {
 		params.OperationName = "operation-1"
 		tags := make(map[string]string)
 		tags["k11"] = "val0"
+		tags["error"] = "true"
```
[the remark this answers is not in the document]
burmanm: Devil is in the details. That single line exploits the bug (the test fails with older version) since it adds another index query against the tags.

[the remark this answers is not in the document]
burmanm: As for the id list, it is basically the list of matches for the search query. A form of a posting list (of traceIDs) if thinking in terms of the ES.

In terms of relational database, it's equivalent to something like: ``SELECT id FROM dbo.spans WHERE service = 'invoices'`` 

That is, a single id list is equivalent to that one. Just imagine each id list is one similar query, touching a single index and single value. It doesn't matter if the index is the same or not (so one query could be against service, one against tags index etc).

PR #1723, opened 2019-08-08
Added null guards to 'Process' when processing an incoming span

### LINKED ISSUE #1722: gRPC batch without service name crashes the collector
When developing the Jaeger Span Exporter via gRPC for OpenTelemetry, I got into a situation where I have a batch, but no `Process` object. Sending such a batch to the collector via gRPC results in the server crashing, due to:

```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0xbe67d1]

goroutine 187 [running]:
github.com/jaegertracing/jaeger/cmd/collector/app.metricsBySvc.ReportServiceNameForSpan(0xc0002d00f0, 0xc0002d0120, 0x1b893e0, 0xc0002638b0, 0xc0004a5368, 0xfa0, 0x13c3eb9, 0x8, 0xc0002d09c0, 0xc0002d1260, ...)
	/home/travis/gopath/src/github.com/jaegertracing/jaeger/cmd/collector/app/metrics.go:247 +0x41
github.com/jaegertracing/jaeger/cmd/collector/app.(*spanProcessor).enqueueSpan(0xc00052aea0, 0xc000308000, 0x13c0205, 0x5, 0x13bf162, 0x4, 0x0)
	/home/travis/gopath/src/github.com/jaegertracing/jaeger/cmd/collector/app/span_processor.go:145 +0x1b2
github.com/jaegertracing/jaeger/cmd/collector/app.(*spanProcessor).ProcessSpans(0xc00052aea0, 0xc0004b4038, 0x1, 0x1, 0x13c0205, 0x5, 0x13bf162, 0x4, 0xc0004f7a28, 0x9edddc, ...)
	/home/travis/gopath/src/github.com/jaegertracing/jaeger/cmd/collector/app/span_processor.go:129 +0x112
github.com/jaegertracing/jaeger/cmd/collector/app.(*GRPCHandler).PostSpans(0xc000394200, 0x1b85b60, 0xc00030e840, 0xc00052a5a0, 0xc000394200, 0xc00030e840, 0xc00060dba8)
	/home/travis/gopath/src/github.com/jaegertracing/jaeger/cmd/collector/app/grpc_handler.go:46 +0x1c3
github.com/jaegertracing/jaeger/proto-gen/api_v2._CollectorService_PostSpans_Handler(0x11db5c0, 0xc000394200, 0x1b85b60, 0xc00030e840, 0xc00009c910, 0x0, 0x1b85b60, 0xc00030e840, 0xc0000909c0, 0x58)
	/home/travis/gopath/src/github.com/jaegertracing/jaeger/proto-gen/api_v2/collector.pb.go:200 +0x23e
github.com/jaegertracing/jaeger/vendor/google.golang.org/grpc.(*Server).processUnaryRPC(0xc00021b980, 0x1b92fc0, 0xc000480a80, 0xc0000f4100, 0xc0001f3380, 0x26d7fb0, 0x0, 0x0, 0x0)
	/home/travis/gopath/src/github.com/jaegertracing/jaeger/vendor/google.golang.org/grpc/server.go:972 +0x470
github.com/jaegertracing/jaeger/vendor/google.golang.org/grpc.(*Server).handleStream(0xc00021b980, 0x1b92fc0, 0xc000480a80, 0xc0000f4100, 0x0)
	/home/travis/gopath/src/github.com/jaegertracing/jaeger/vendor/google.golang.org/grpc/server.go:1252 +0xda6
github.com/jaegertracing/jaeger/vendor/google.golang.org/grpc.(*Server).serveStreams.func1.1(0xc0004a4214, 0xc00021b980, 0x1b92fc0, 0xc000480a80, 0xc0000f4100)
	/home/travis/gopath/src/github.com/jaegertracing/jaeger/vendor/google.golang.org/grpc/server.go:691 +0x9f
created by github.com/jaegertracing/jaeger/vendor/google.golang.org/grpc.(*Server).serveStreams.func1
	/home/travis/gopath/src/github.com/jaegertracing/jaeger/vendor/google.golang.org/grpc/server.go:689 +0xa1
```

Of course my OpenTelemetry code has to be fixed, but the collector should never crash because of bad data.

jpkrohling:
- Fixes #1722 by returning an error before attempting to write the span that has no process attached to it. For the metrics collection, the service name is set to `__unknown`. Previously, such problems were not apparent via metrics.

Before this PR, the collector would crash at different points (see #1722) depending on the storage being used. After this PR, the following message is shown in the logs:

```
{"level":"error","ts":1565269296.058833,"caller":"app/span_processor.go:114","msg":"Failed to save span","error":"process is empty for the span","stacktrace":"github.com/jaegertracing/jaeger/cmd/collector/app.(*spanProcessor).saveSpan\n\t/mnt/storage/jpkroehling/Projects/src/github.com/jaegertracing/jaeger/cmd/collector/app/span_processor.go:114\ngithub.com/jaegertracing/jaeger/cmd/collector/app.ChainedProcessSpan.func1\n\t/mnt/storage/jpkroehling/Projects/src/github.com/jaegertracing/jaeger/cmd/collector/app/model_consumer.go:34\ngithub.com/jaegertracing/jaeger/cmd/collector/app.(*spanProcessor).processItemFromQueue\n\t/mnt/storage/jpkroehling/Projects/src/github.com/jaegertracing/jaeger/cmd/collector/app/span_processor.go:139\ngithub.com/jaegertracing/jaeger/cmd/collector/app.NewSpanProcessor.func1\n\t/mnt/storage/jpkroehling/Projects/src/github.com/jaegertracing/jaeger/cmd/collector/app/span_processor.go:68\ngithub.com/jaegertracing/jaeger/pkg/queue.(*BoundedQueue).StartConsumers.func1\n\t/mnt/storage/jpkroehling/Projects/src/github.com/jaegertracing/jaeger/pkg/queue/bounded_queue.go:65"}
```

And this counter is incremented:

```
jaeger_collector_spans_saved_by_svc_total{debug="false",result="err",svc="__unknown"} 1
```

### DIFF
--- a/cmd/collector/app/metrics.go
+++ b/cmd/collector/app/metrics.go
@@ -244,10 +244,13 @@ func (m *SpanProcessorMetrics) GetCountsForFormat(spanFormat SpanFormat, transpo
 // reportServiceNameForSpan determines the name of the service that emitted
 // the span and reports a counter stat.
 func (m metricsBySvc) ReportServiceNameForSpan(span *model.Span) {
-	serviceName := span.Process.ServiceName
-	if serviceName == "" {
-		return
+	var serviceName string
+	if nil == span.Process || len(span.Process.ServiceName) == 0 {
+		serviceName = "__unknown"
+	} else {
+		serviceName = span.Process.ServiceName
 	}
+
 	m.countSpansByServiceName(serviceName, span.Flags.IsDebug())
 	if span.ParentSpanID() == 0 {
 
--- a/cmd/collector/app/span_processor.go
+++ b/cmd/collector/app/span_processor.go
@@ -109,6 +109,12 @@ func (sp *spanProcessor) Stop() {
 }
 
 func (sp *spanProcessor) saveSpan(span *model.Span) {
+	if nil == span.Process {
+		sp.logger.Error("process is empty for the span")
+		sp.metrics.SavedErrBySvc.ReportServiceNameForSpan(span)
+		return
+	}
+
 	startTime := time.Now()
 	if err := sp.spanWriter.WriteSpan(span); err != nil {
 		sp.logger.Error("Failed to save span", zap.Error(err))
--- a/cmd/collector/app/span_processor_test.go
+++ b/cmd/collector/app/span_processor_test.go
@@ -299,3 +299,19 @@ func TestSpanProcessorBusy(t *testing.T) {
 	assert.Error(t, err, "expcting busy error")
 	assert.Nil(t, res)
 }
+
+func TestSpanProcessorWithNilProcess(t *testing.T) {
+	mb := metricstest.NewFactory(time.Hour)
+	serviceMetrics := mb.Namespace(metrics.NSOptions{Name: "service", Tags: nil})
+
+	w := &fakeSpanWriter{}
+	p := NewSpanProcessor(w, Options.ServiceMetrics(serviceMetrics)).(*spanProcessor)
+	defer p.Stop()
+
+	p.saveSpan(&model.Span{})
+
+	expected := []metricstest.ExpectedMetric{{
+		Name: "service.spans.saved-by-svc|debug=false|result=err|svc=__unknown", Value: 1,
+	}}
+	mb.AssertCounterMetrics(t, expected...)
+}


jpkrohling:
Another solution to this is what was suggested by @pavolloffay in the issue, by making sure the `Process` is never nil. We'd then have to come up with a default `serviceName`. For the metric reporting, I'm setting `__unknown`, which could also serve as the default service name in case it's empty.

Given that all clients currently fail in case the service name is empty, I'd argue that the right thing to do when the `Process` is empty is to indeed return an error from `WriteSpan`

pavolloffay:
Does the storage work fine when the service name is nill? If no, we could have a sanitizer which sets the service name to some value.

jpkrohling:
Some storages fail (ES), some work (Memory). It's not really consistent across storage mechanisms.

pavolloffay:
The client libraries also fail if the service name is null. 

In the codebase we do not do nil checks on the arguments. As the storage might also fails when the service name is empty I think maybe a better would be to set it to some value or fail early in the ingestion at one place.

jpkrohling:
In theory, each method should guard itself against bad data ("be conservative in what you do, be liberal in what you accept from others"). We could add a check at the `spanProcessor#saveSpan`, which calls all other `WriteSpan` methods, but that method itself isn't using the `Process` object.

yurishkuro:
I don't think it's necessary to spread this check so much around. Storage implementations have a contract with the caller, Process is a required element in the data model.

We should only check this on the receiving side, as close as possible to the servers, and not infect the rest of the business logic with input validation.

jpkrohling:
I just updated the PR to do the check before the call to `WriteSpan`. This doesn't make the `WriteSpan` methods more robust, but as long as we can guarantee that no such calls are made with `nil` as the `Process` we should be good.

Here's the error message for the current state of this PR when a batch with a nil `Process` is sent:

```
{"level":"error","ts":1565339952.5434508,"caller":"app/span_processor.go:113","msg":"process is empty for the span","stacktrace":"github.com/jaegertracing/jaeger/cmd/collector/app.(*spanProcessor).saveSpan\n\t/mnt/storage/jpkroehling/Projects/src/github.com/jaegertracing/jaeger/cmd/collector/app/span_processor.go:113\ngithub.com/jaegertracing/jaeger/cmd/collector/app.ChainedProcessSpan.func1\n\t/mnt/storage/jpkroehling/Projects/src/github.com/jaegertracing/jaeger/cmd/collector/app/model_consumer.go:34\ngithub.com/jaegertracing/jaeger/cmd/collector/app.(*spanProcessor).processItemFromQueue\n\t/mnt/storage/jpkroehling/Projects/src/github.com/jaegertracing/jaeger/cmd/collector/app/span_processor.go:145\ngithub.com/jaegertracing/jaeger/cmd/collector/app.NewSpanProcessor.func1\n\t/mnt/storage/jpkroehling/Projects/src/github.com/jaegertracing/jaeger/cmd/collector/app/span_processor.go:68\ngithub.com/jaegertracing/jaeger/pkg/queue.(*BoundedQueue).StartConsumers.func1\n\t/mnt/storage/jpkroehling/Projects/src/github.com/jaegertracing/jaeger/pkg/queue/bounded_queue.go:65"}
```

And here are the metrics that are recorded when this happens:

```console
$ curl localhost:14269/metrics 2>/dev/null | grep __unknown 
jaeger_collector_spans_received_total{debug="false",format="proto",svc="__unknown",transport="grpc"} 1
jaeger_collector_spans_saved_by_svc_total{debug="false",result="err",svc="__unknown"} 1
jaeger_collector_traces_received_total{debug="false",format="proto",sampler_type="unknown",svc="__unknown",transport="grpc"} 1
jaeger_collector_traces_saved_by_svc_total{debug="false",result="err",sampler_type="unknown",svc="__unknown"} 1
```

jpkrohling:
@pavolloffay , @yurishkuro , I believe this is now in line with that you suggested. Could you please take a look?

PR #1733, opened 2019-08-15
Place a queue infront of agent reporter to provide some resiliency

### LINKED ISSUE #1430: Jaeger Agent does not buffer spans when the Collector is unavailable

In our environment, the Collector comes up after the Jaeger-Agent, which we deploy as a sidecar in a Kubernetes pod. This means that between the time an app in a pod starts tracing and the time the Collector is up, the traces are lost. 

It seems that the Jaeger-Agent simply drops spans if the Collector is not available. Some of these early spans are important to understand system behaviour. 

At a minimum, I imagine the Jaeger-Agent buffering the spans in-memory (with a configurable amount of memory); if the in-memory limit is reached, drop older traces or drop new ones (also configurable?).

As a slightly mode advanced version, Jeager-Agent could also offer to write the buffered spans to disk.

burmanm:

Fixes #1430 , by buffering incoming spans if the collector can't be reached. 

This wraps the outbound reporter with a queue to which new spans are first pushed to and then pulled by the processing threads. If the collector can't be reached due to an error that's caused by network issues, the processing thread is put to sleep for a exponential time (until the maximum is hit).

The queue can be changed to a different one to provide features such as persistence to processed messages. The process should provide "at-least-once" delivery, not necessarily "only-once". 

Certain things omitted from this PR to get feedback if this was the intended approach, so no configuration options for amount of workers, times or which queue to be used.

### DIFF
--- a/cmd/agent/app/flags.go
+++ b/cmd/agent/app/flags.go
@@ -47,8 +47,8 @@ var defaultProcessors = []struct {
 func AddFlags(flags *flag.FlagSet) {
 	for _, p := range defaultProcessors {
 		prefix := fmt.Sprintf("processor.%s-%s.", p.model, p.protocol)
-		flags.Int(prefix+suffixWorkers, defaultServerWorkers, "how many workers the processor should run, deprecated - use reporter config")
-		flags.Int(prefix+suffixServerQueueSize, defaultQueueSize, "length of the queue for the UDP server, deprecated - use reporter config")
+		flags.Int(prefix+suffixWorkers, 0, "(deprecated) see --reporter.queue.workers")
+		flags.Int(prefix+suffixServerQueueSize, 0, "(deprecated) see --reporter.queue.memory.size")
 		flags.Int(prefix+suffixServerMaxPacketSize, defaultMaxPacketSize, "max packet size for the UDP server")
 		flags.String(prefix+suffixServerHostPort, ":"+strconv.Itoa(p.port), "host:port for the UDP server")
 	}
--- a/cmd/agent/app/flags.go
+++ b/cmd/agent/app/flags.go
@@ -47,8 +47,8 @@ var defaultProcessors = []struct {
 func AddFlags(flags *flag.FlagSet) {
 	for _, p := range defaultProcessors {
 		prefix := fmt.Sprintf("processor.%s-%s.", p.model, p.protocol)
-		flags.Int(prefix+suffixWorkers, defaultServerWorkers, "how many workers the processor should run")
-		flags.Int(prefix+suffixServerQueueSize, defaultQueueSize, "length of the queue for the UDP server")
+		flags.Int(prefix+suffixWorkers, defaultServerWorkers, "how many workers the processor should run, deprecated - use reporter config")
+		flags.Int(prefix+suffixServerQueueSize, defaultQueueSize, "length of the queue for the UDP server, deprecated - use reporter config")
 		flags.Int(prefix+suffixServerMaxPacketSize, defaultMaxPacketSize, "max packet size for the UDP server")
 		flags.String(prefix+suffixServerHostPort, ":"+strconv.Itoa(p.port), "host:port for the UDP server")
 	}
@@ -63,8 +63,6 @@ func (b *Builder) InitFromViper(v *viper.Viper) *Builder {
 	for _, processor := range defaultProcessors {
 		prefix := fmt.Sprintf("processor.%s-%s.", processor.model, processor.protocol)
 		p := &ProcessorConfiguration{Model: processor.model, Protocol: processor.protocol}
-		p.Workers = v.GetInt(prefix + suffixWorkers)
-		p.Server.QueueSize = v.GetInt(prefix + suffixServerQueueSize)
 		p.Server.MaxPacketSize = v.GetInt(prefix + suffixServerMaxPacketSize)
 		p.Server.HostPort = v.GetString(prefix + suffixServerHostPort)
 		b.Processors = append(b.Processors, *p)
--- a/cmd/agent/app/reporter/common/domain_tools.go
+++ b/cmd/agent/app/reporter/common/domain_tools.go
@@ -12,21 +12,22 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package queue
+package common
 
 import "github.com/jaegertracing/jaeger/model"
 
-// NonQueue sends stuff directly without queueing. Useful for testing purposes
-type NonQueue struct {
-	processor func(model.Batch) error
-}
-
-// NewNonQueue returns direct processing "queue"
-func NewNonQueue(processor func(model.Batch) error) *NonQueue {
-	return &NonQueue{processor}
-}
-
-// Enqueue calls processor instead of queueing
-func (n *NonQueue) Enqueue(batch model.Batch) error {
-	return n.processor(batch)
+// AddProcessTags appends jaeger tags for the agent to every span it sends to the collector.
+func AddProcessTags(spans []*model.Span, process *model.Process, agentTags []model.KeyValue) ([]*model.Span, *model.Process) {
+	if len(agentTags) == 0 {
+		return spans, process
+	}
+	if process != nil {
+		process.Tags = append(process.Tags, agentTags...)
+	}
+	for _, span := range spans {
+		if span.Process != nil {
+			span.Process.Tags = append(span.Process.Tags, agentTags...)
+		}
+	}
+	return spans, process
 }
--- a/cmd/agent/app/reporter/flags.go
+++ b/cmd/agent/app/reporter/flags.go
@@ -19,39 +19,68 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	"time"
 
 	"github.com/spf13/viper"
 )
 
 const (
+	prefix = "reporter."
 	// Whether to use grpc or tchannel reporter.
-	reporterType = "reporter.type"
+	reporterType = "type"
+	// Queue implementation to use
+	reporterQueueType = "queue.type"
+	// Bounded-queue size
+	boundedQueueSize = "queue.memory.size"
+	// Max retry interval
+	reporterMaxInterval = "max-retry-interval"
+	// Amount of workers for reporter
+	reporterQueueWorkers = "queue.workers"
 	// Agent tags
 	agentTags = "jaeger.tags"
 	// TCHANNEL is name of tchannel reporter.
 	TCHANNEL Type = "tchannel"
 	// GRPC is name of gRPC reporter.
 	GRPC Type = "grpc"
+
+	// MEMORY is the default reporter queue
+	MEMORY QueueType = "memory"
 )
 
 // Type defines type of reporter.
 type Type string
 
+// QueueType defines the type of queue in reporter
+type QueueType string
+
 // Options holds generic reporter configuration.
 type Options struct {
-	ReporterType Type
-	AgentTags    map[string]string
+	ReporterType             Type
+	QueueType                QueueType
+	ReporterConcurrency      int
+	BoundedQueueSize         int
+	ReporterMaxRetryInterval time.Duration
+	AgentTags                map[string]string
 }
 
 // AddFlags adds flags for Options.
 func AddFlags(flags *flag.FlagSet) {
-	flags.String(reporterType, string(GRPC), fmt.Sprintf("Reporter type to use e.g. %s, %s", string(GRPC), string(TCHANNEL)))
+	flags.String(prefix+reporterType, string(GRPC), fmt.Sprintf("Reporter type to use e.g. %s, %s", string(GRPC), string(TCHANNEL)))
+	flags.String(prefix+reporterQueueType, string(MEMORY), "queue implementation to use in the reporter. Available options: memory")
+	flags.Int(prefix+boundedQueueSize, defaultBoundedQueueSize, "maximum size of bounded in-memory queue")
+	flags.Int(prefix+reporterQueueWorkers, defaultQueueWorkers, "the amount of concurrent reporter connections to use")
+	flags.Duration(prefix+reporterMaxInterval, defaultMaxRetryInterval, "longest period of time to wait before retry. Format is time.Duration (https://golang.org/pkg/time/#Duration).")
+
 	flags.String(agentTags, "", "One or more tags to be added to the Process tags of all spans passing through this agent. Ex: key1=value1,key2=${envVar:defaultValue}")
 }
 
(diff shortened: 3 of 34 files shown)


cmd/agent/app/reporter/queue.go
```diff
@@ -0,0 +1,98 @@
...
+		retryTimerChange:  time.Now(),
+		retryMaxWait:      20 * time.Second,
+	}
+	q.retryTimer = q.retryDefaultSleep
+	q.queue = queue.NewBoundQueue(1000, q.batchProcessor, logger)
+	return q
+}
+
+func (q *QueuedReporter) EmitZipkinBatch(spans []*zipkincore.Span) error {
+	return nil
```
yurishkuro: maybe just delegate to wrapped

cmd/agent/app/reporter/queue.go
```diff
@@ -0,0 +1,98 @@
...
+func WrapWithQueue(reporter Reporter, logger *zap.Logger) *QueuedReporter {
+	q := &QueuedReporter{
+		wrapped:           reporter,
+		logger:            logger,
+		retryDefaultSleep: time.Millisecond * 100,
+		retryTimerChange:  time.Now(),
+		retryMaxWait:      20 * time.Second,
+	}
+	q.retryTimer = q.retryDefaultSleep
+	q.queue = queue.NewBoundQueue(1000, q.batchProcessor, logger)
```
yurishkuro: these params would need to be exposed as configurable

cmd/agent/app/reporter/queue/bounded.go
```diff
@@ -0,0 +1,39 @@
...
+	b := &Bound{
+		queue:  queue.NewBoundedQueue(bufSize, nil),
+		logger: logger,
+	}
+
+	b.queue.StartConsumers(1, func(item interface{}) {
+		// This queue does not have persistence, thus we don't handle transactionality
+		err := processor(item.(*jaeger.Batch))
+		if err != nil {
+			b.logger.Error("Could not transmit span", zap.Error(err))
```
yurishkuro: "could not transmit batch"

cmd/agent/app/reporter/grpc/reporter.go
```diff
@@ -100,3 +102,14 @@ func makeModelKeyValue(agentTags map[string]string) []model.KeyValue {
 
 	return tags
 }
+
+func (r *Reporter) Retryable(err error) bool {
+	state := status.Convert(err)
+	switch state.Code() {
+	case codes.DeadlineExceeded:
+		fallthrough
```
yurishkuro: can we just return? easier to read

cmd/agent/app/reporter/grpc/collector_proxy.go
```diff
@@ -45,7 +45,7 @@ func NewCollectorProxy(builder *ConnBuilder, agentTags map[string]string, mFacto
 	grpcMetrics := mFactory.Namespace(metrics.NSOptions{Name: "", Tags: map[string]string{"protocol": "grpc"}})
 	return &ProxyBuilder{
 		conn:     conn,
-		reporter: reporter.WrapWithMetrics(NewReporter(conn, agentTags, logger), grpcMetrics),
+		reporter: reporter.WrapWithQueue(reporter.WrapWithMetrics(NewReporter(conn, agentTags, logger), grpcMetrics), logger),
```
yurishkuro: We need to rethink how metrics would work, I think this change breaks them. Every time wrapped reporter fails to submit, we'd increment the failed counter, even though the batch may be retried. And vice versa, once we give up retrying, there's no metric at all. I think we need to split failure metrics into (a) how many retries we're going (a sustained elevated level may be cause for alarm), and (b) how many spans/batches we're dropping even after retries.

cmd/agent/app/reporter/grpc/reporter.go
```diff
@@ -100,3 +102,14 @@ func makeModelKeyValue(agentTags map[string]string) []model.KeyValue {
 
 	return tags
 }
+
+func (r *Reporter) Retryable(err error) bool {
```
yurishkuro: Instead of introducing this introspection method to the Reporter interface, I think it would be cleaner to encode this directly into the error returned.
```go
// in the top reporter.go file
type RetryableError interface {
    error
    IsRetryable() bool
}

func IsRetryable(err error) bool {
    if r, ok := err.(RetryableError); ok {
        return r.IsRetryable()
    }
    return false
}

// in grpc reporter
type maybeRetryableGRPCError error

func (r maybeRetryableGRPCError) IsRetryable() bool {
   // this logic
}
```

Then gRPC reporter just returns `maybeRetryableGRPCError(err)`

yurishkuro:
sorry, submitted an incomplete review. But I think it raises two significant questions, so further review could wait.

cmd/agent/app/reporter/queue.go
```diff
@@ -0,0 +1,98 @@
...
+	wrapped Reporter
+	queue   Queue
+	logger  *zap.Logger
+
+	retryMutex sync.Mutex
+
+	retryTimerChange  time.Time
+	retryTimer        time.Duration
+	retryMaxWait      time.Duration
+	retryDefaultSleep time.Duration
```
yurishkuro: I think these could use better names, e.g. 
  * lastRetry
  * currentRetryInterval
  * maxRetryInterval
  * initialRetryInterval

yurishkuro:
Great to see this work happening, btw, thanks!!! 🎉 

Why is it tagged Badger? Right now it's not offloading to disk, but I can see this being extended to do so.

cmd/agent/app/reporter/grpc/collector_proxy.go
burmanm: I can merge the metrics decorator to this one, that would at least allow more accurate reporting.

pavolloffay:
> Why is it tagged Badger?

my bad I have removed it.

burmanm:
As for the amount of workers, it's going to be a bit trickier after this. Considering the offload from queue to gRPC is mostly I/O bound - there's no point restricting this to the amount of CPU cores. Most of the workers in this scenario are going to be waiting for the gRPC transfer. Thus the amount of active workers can easily exceed the amount of available CPU cores.

The other part is actually receiving the data (I/O - although I guess this could be mostly done in the kernel layer as part of the epoll processing) and then just enqueuing the data. For persistent queue this could cause some I/O waiting (hopefully less than the gRPC part) but for in-memory ones this is mostly CPU bound.

Thus limiting them to same count doesn't feel like ideal option. As for the risk of serializing the communication, I don't think the processor part is going to get serialized even if all the workers towards the gRPC server have stalled. The queue would still accept new writes. The selected "1" worker here was just for easier debugging at this point and also to discuss how we might want to configure the amount of workers. 

With the queue moved to this position I'm not sure if it makes sense to have the queue in the processing part anymore though. Although it would probably be close to "SEDA"-architecture, I can't think of much of a benefit (but certainly negatives). Especially since the processing is no longer really blocked (assuming the queue can keep up) and shouldn't be the limiting factor of the thrift-processing.

Of course, we could have used the tbuffered's queue also and just push the retry-information down there. But I think the original ticket wanted to have this as "generic" approach which to me sounded like it should buffer even if the spans didn't come through the tbuffered.

yurishkuro:
> With the queue moved to this position I'm not sure if it makes sense to have the queue in the processing part anymore though.

I tend to agree, therefore interested in how we can change the architecture. Today's 3 queues, one per each inbound model representation, were never ideal anyway.

cmd/agent/app/reporter/grpc/reporter.go
burmanm: There's actually more of them, at least Unknown as well (on some runs I can see that from timeouts). But yeah, I can do return from them as well  although to me that makes it more difficult to read ;) (Go is probably the only language with this type of switch..)

cmd/agent/app/reporter/common/domain_tools.go
```diff
@@ -0,0 +1,33 @@
...
+func AddProcessTags(spans []*model.Span, process *model.Process, agentTags []model.KeyValue) ([]*model.Span, *model.Process) {
+	if len(agentTags) == 0 {
+		return spans, process
+	}
+	if process != nil {
+		process.Tags = append(process.Tags, agentTags...)
+	}
+	for _, span := range spans {
+		if span.Process != nil {
+			span.Process.Tags = append(span.Process.Tags, agentTags...)
```
pavolloffay: Isn't the process the same as above? The batch should contain only one unique process.

cmd/agent/app/reporter/flags.go
```diff
@@ -19,39 +19,71 @@ import (
...
 	// TCHANNEL is name of tchannel reporter.
 	TCHANNEL Type = "tchannel"
 	// GRPC is name of gRPC reporter.
 	GRPC Type = "grpc"
+
+	// MEMORY is the default reporter queue
+	MEMORY QueueType = "memory"
+
+	// DIRECT is the reporter queue for testing (no queue at all)
+	DIRECT QueueType = "direct"
```
pavolloffay: Could be this moved to test file?

cmd/agent/app/flags.go
```diff
@@ -47,8 +47,8 @@ var defaultProcessors = []struct {
 func AddFlags(flags *flag.FlagSet) {
 	for _, p := range defaultProcessors {
 		prefix := fmt.Sprintf("processor.%s-%s.", p.model, p.protocol)
-		flags.Int(prefix+suffixWorkers, defaultServerWorkers, "how many workers the processor should run")
-		flags.Int(prefix+suffixServerQueueSize, defaultQueueSize, "length of the queue for the UDP server")
+		flags.Int(prefix+suffixWorkers, defaultServerWorkers, "how many workers the processor should run, deprecated - use reporter config")
+		flags.Int(prefix+suffixServerQueueSize, defaultQueueSize, "length of the queue for the UDP server, deprecated - use reporter config")
```
pavolloffay: Could this be used to set the new option value in the reporter? `queue.memory.size`

cmd/agent/app/flags.go
```diff
@@ -47,8 +47,8 @@ var defaultProcessors = []struct {
 func AddFlags(flags *flag.FlagSet) {
 	for _, p := range defaultProcessors {
 		prefix := fmt.Sprintf("processor.%s-%s.", p.model, p.protocol)
-		flags.Int(prefix+suffixWorkers, defaultServerWorkers, "how many workers the processor should run")
-		flags.Int(prefix+suffixServerQueueSize, defaultQueueSize, "length of the queue for the UDP server")
+		flags.Int(prefix+suffixWorkers, defaultServerWorkers, "how many workers the processor should run, deprecated - use reporter config")
```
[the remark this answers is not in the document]
pavolloffay: Could this be used to set the new option in `queue.workers` otherwise this can break some deployments.

cmd/agent/app/reporter/queue.go
```diff
@@ -0,0 +1,218 @@
...
+		initialRetryInterval:    time.Millisecond * 100,
+		maxRetryInterval:        opts.ReporterMaxRetryInterval,
+		reporterMetrics:         NewMetricsReporter(forwarder, mFactory),
+		agentTags:               model.KeyValueFromMap(opts.AgentTags),
+		close:                   uint32(0),
+	}
+	q.currentRetryInterval = q.initialRetryInterval
+
+	// Some sanity checks
+	if opts.QueueType == "" {
```
pavolloffay: Does not seem needed? There can be default statement in the switch below.

cmd/agent/app/reporter/queue.go
```diff
@@ -0,0 +1,218 @@
...
+	err := q.wrapped.ForwardBatch(batch)
+	if err != nil {
+		for IsRetryable(err) {
+			if q.closed() {
+				return false, err
+			}
+			q.reporterMetrics.BatchMetrics.BatchesRetries.Inc(1)
+			// Block this processing instance before returning
+			sleepTime := q.backOffTimer()
+			q.logger.Error(fmt.Sprintf("Failed to contact the collector, waiting %s before retry", sleepTime.String()))
```
pavolloffay: We use structural logging in Jaeger. 

e.g. 
```
							logger.Error("Elasticsearch part of bulk request failed", zap.String("map-key", key),

```

cmd/agent/app/reporter/flags.go
```diff
@@ -19,39 +19,71 @@ import (
...
+	BoundedQueueSize         int
+	ReporterMaxRetryInterval time.Duration
+	AgentTags                map[string]string
 }
 
 // AddFlags adds flags for Options.
 func AddFlags(flags *flag.FlagSet) {
-	flags.String(reporterType, string(GRPC), fmt.Sprintf("Reporter type to use e.g. %s, %s", string(GRPC), string(TCHANNEL)))
+	flags.String(prefix+reporterType, string(GRPC), fmt.Sprintf("Reporter type to use e.g. %s, %s", string(GRPC), string(TCHANNEL)))
+	flags.String(prefix+reporterQueueType, string(defaultQueueType), "queue implementation to use in the reporter. Available options: memory")
```
pavolloffay: Do we want to offer also direct e.g. skip the queue?

cmd/agent/app/flags.go
burmanm: I did think about this earlier, but do we want to reuse existing flags to modify new behavior? Could that cause some unexpected user experience?

Although in short term we could remap this to the same one (use the new flag if that has value, otherwise check this one and log something).

cmd/agent/app/reporter/common/domain_tools.go
burmanm: I haven't changed this code, it was just moved. So the behavior and output of the reporter would stay the same as currently.

cmd/agent/app/reporter/flags.go
burmanm: Referring to your comment below, we could still keep it to skip the queue (just remove the "testing only" from the description)

cmd/agent/app/flags.go
pavolloffay: We did the same before for other flags, it prevented us from breaking changes. I think we also generated warning message if the old flag has been used.

cmd/agent/app/reporter/queue.go
burmanm: I thought it would make more sense to have it separated instead of switch's fallthrough. To me it makes different responsibilities of the code separated. Easier to read, nothing else. But YMMV.

pavolloffay:
@burmanm could you please rebase it?

pavolloffay:
Closing per https://github.com/jaegertracing/jaeger/issues/1430#issuecomment-631496478

PR #1745, opened 2019-08-20
Added base Grafana dashboard and Alert rules

jpkrohling:


pavolloffay:
I have two comments:

* Is there a way to verify that these files work with grafana? An automated test would be useful.
* Do these definitions work for other systems than grafana? If no I would consider renaming subdirectory `jaeger-mixins` to `grafana`. It should help with search but also it tells folks not familiar with grafana that we provide something.  `monitoring/jaeger-mixin/dashboards-jaeger.json`

jpkrohling:
> Is there a way to verify that these files work with grafana? An automated test would be useful.

Not easily: a test would require a client sending spans to Jaeger, which exposes metrics that are scraped by Prometheus and displayed by Grafana.

> Do these definitions work for other systems than grafana?

I'm not sure, but I agree that the dashboard file name could be changed to mention Grafana in the name. I'll also update the readme file.

PR #1695, opened 2019-07-26
Support for Zipkin Protobuf spans over HTTP

jan25:
## Which problem is this PR solving?
- Adds support in collector for Zipkin Protobuf spans sent to `/api/v2/spans`
- Fixes https://github.com/jaegertracing/jaeger/issues/1586

## Short description of the changes
- We had JSON support for Zipkin spans sent over HTTP to v2 endpoint, this PR is a mirrored implementation based on [jsonv2](https://github.com/jaegertracing/jaeger/blob/master/cmd/collector/app/zipkin/jsonv2.go) implementation
- Related PRs that should be merged to complete adding zipkin proto support: https://github.com/jaegertracing/xdock-zipkin-brave/pull/13 and https://github.com/jaegertracing/jaeger-idl/pull/56
- <s>Temporarily marking this PR as WIP. Please review/suggest on the implementation</s>

PR #1747, opened 2019-08-20
Refactor zipkin/protov2.go

jan25:
- Related to https://github.com/jaegertracing/jaeger/pull/1695#discussion_r314391550 (from previously merged PR)
- Improve `model/ids.go` to parse `[]byte` into SpanID/TraceID
- Use `model/ids.go` in `zipkin/protov2.go`

### DIFF
--- a/model/ids.go
+++ b/model/ids.go
@@ -173,6 +173,15 @@ func SpanIDFromString(s string) (SpanID, error) {
 	return SpanID(id), nil
 }
 
+// SpanIDFromBytes creates a SpandID from list of bytes
+func SpanIDFromBytes(data []byte) (SpanID, error) {
+	id := SpanID(0)
+	if err := id.Unmarshal(data); err != nil {
+		return SpanID(0), err
+	}
+	return id, nil
+}
+
 // MarshalText is called by encoding/json, which we do not want people to use.
 func (s SpanID) MarshalText() ([]byte, error) {
 	return nil, fmt.Errorf("unsupported method SpanID.MarshalText; please use github.com/gogo/protobuf/jsonpb for marshalling")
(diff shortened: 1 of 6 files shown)


jan25:
crossdock-go test failed, i think its flaky and unrelated to this PR. Is there something we can do to fix that? so CI wouldn't break for any PRs in future

model/ids.go
```diff
@@ -77,6 +77,21 @@ func TraceIDFromString(s string) (TraceID, error) {
...
+	case len(data) == traceIDLongBytesLen:
+		t.High = binary.BigEndian.Uint64(data[:traceIDShortBytesLen])
+		t.Low = binary.BigEndian.Uint64(data[traceIDShortBytesLen:])
+	case len(data) == traceIDShortBytesLen:
+		t.Low = binary.BigEndian.Uint64(data)
+	default:
+		return TraceID{}, fmt.Errorf("invalid length for TraceID")
+	}
+	return t, nil
+}
```
yurishkuro: Since we're adding a new function, I would prefer to reuse it from `Unmarshal(data []byte)`, e.g.
```
func (t *TraceID) Unmarshal(data []byte) error {
    var err error
    *t, err = TraceIDFromBytes(data)
    return err
}
```

Similar for Unmarshal on SpanID

model/ids.go
```diff
@@ -166,6 +181,15 @@ func SpanIDFromString(s string) (SpanID, error) {
 	return SpanID(id), nil
 }
 
+// SpanIDFromBytes creates a SpandID from list of bytes
+func SpanIDFromBytes(data []byte) (SpanID, error) {
```
yurishkuro: to follow on the TraceIDFromBytes, swap it and implement SpanID.Unmarshal in terms of SpanIDFromBytes.

model/ids.go
jan25: Done. And this change would let `func Unmarshal` parse 8byte long TraceID(Higher 8bytes will be 0)

PR #1749, opened 2019-08-20
Make operation_strategies part also be part of default_strategy 

### LINKED ISSUE #1748: Allow default per-operation sampling strategies

The `/health` and `/metrics` endpoint make other traces hard to find.

These traces can be ignored with the fix of #814. This fix allows to set operation strategies on a per-service basis, but with a lot of services it can quickly become unmaintainable to disable spans for the health endpoint for every service.

Allow to specify operation strategies in the default sampling strategy, like so:

```json
{
  "default_strategy": {
    "type": "probabilistic",
    "param": 1,
    "operation_strategies": [
      {
        "operation": "op1",
        "type": "probabilistic",
        "param": 0.2
      },
      {
        "operation": "op2",
        "type": "probabilistic",
        "param": 0.4
      }
    ]
  }
}
```

None

Note: this issue is also very closely related to #1387

rutgerbrf:

Resolves #1748 

Makes operation strategies a default part of strategy so operation strategies can be defined for the default sampling strategy.


plugin/sampling/strategystore/static/strategy.go
```diff
@@ -17,8 +17,9 @@ package static
 // strategy defines a sampling strategy. Type can be "probabilistic" or "ratelimiting"
 // and Param will represent "sampling probability" and "max traces per second" respectively.
 type strategy struct {
-	Type  string  `json:"type"`
-	Param float64 `json:"param"`
+	Type                string               `json:"type"`
+	Param               float64              `json:"param"`
+	OperationStrategies []*operationStrategy `json:"operation_strategies"`
```
yurishkuro: This is not a good encapsulation, a strategy is a "simple" struct. If necessary, the overall config can have a separate service-less per-operation section, but we should not break the notion of a simple strategy.

yurishkuro:
btw, this might be of interest https://github.com/jaegertracing/jaeger/issues/365#issuecomment-505913581

plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -74,17 +75,67 @@ func loadStrategies(strategiesFile string) (*strategies, error) {
...
+
 	for _, s := range strategies.ServiceStrategies {
 		h.serviceStrategies[s.Service] = h.parseServiceStrategies(s)
+
+		// Merge with the default operation strategies, because only merging with
+		// the default strategy has no effect on service strategies (the default strategy
+		// is not merged with and only used as a fallback).
+		opS := h.serviceStrategies[s.Service].OperationSampling
+		if opS == nil {
+			// It has no use to merge, just reference the default settings.
```
yurishkuro: `// Service has no per-operation strategies, so just reference the default settings.`

plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -74,17 +75,67 @@ func loadStrategies(strategiesFile string) (*strategies, error) {
...
+			h.serviceStrategies[s.Service].OperationSampling = h.defaultStrategy.OperationSampling
+			continue
+		}
+		if h.defaultStrategy.OperationSampling == nil ||
+			h.defaultStrategy.OperationSampling.PerOperationStrategies == nil {
+			continue
+		}
```
yurishkuro: this could go before L90

plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -120,6 +165,55 @@ func (h *strategyStore) parseServiceStrategies(strategy *serviceStrategy) *sampl
 	return resp
 }
 
+func (h *strategyStore) parseDefaultStrategy(strategy *defaultStrategy) *sampling.SamplingStrategyResponse {
```
yurishkuro: is this function really needed? Couldn't we just use `parseServiceStrategies`?

plugin/sampling/strategystore/static/strategy.go
```diff
@@ -34,8 +34,14 @@ type serviceStrategy struct {
 	strategy
 }
 
+// defaultStrategy defines the default strategy.
+type defaultStrategy struct {
```
yurishkuro: I think we can simply use `serviceStrategy`.  The `Service` field can be ignored. Then it will match the structure of the store:
```go
type strategyStore struct {
	logger *zap.Logger

	defaultStrategy   *sampling.SamplingStrategyResponse
	serviceStrategies map[string]*sampling.SamplingStrategyResponse
}
```

plugin/sampling/strategystore/static/constants.go
```diff
@@ -35,7 +35,7 @@ const (
 var (
 	// defaultStrategy is the default sampling strategy the Strategy Store will return
 	// if none is provided.
-	defaultStrategy = sampling.SamplingStrategyResponse{
+	defaultStrategyResponse = sampling.SamplingStrategyResponse{
```
yurishkuro: Would it be better to make this into a function? At one point we're doing deepCopy(), instead we can just always return a new object.

plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -74,17 +75,67 @@ func loadStrategies(strategiesFile string) (*strategies, error) {
...
+			h.defaultStrategy.OperationSampling.PerOperationStrategies)
+	}
+}
+
+// mergePerOperationStrategies merges two operation strategies a and b, where a takes precedence over b.
+func mergePerOperationSamplingStrategies(
+	a, b []*sampling.OperationSamplingStrategy,
+) []*sampling.OperationSamplingStrategy {
+	// Guess the size of the slice of the two merged.
+	merged := make([]*sampling.OperationSamplingStrategy, 0, (len(a)+len(b))/4*3)
```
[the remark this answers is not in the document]
rutgerbrf: The idea is that, if a service already defines a strategy for a method, the default strategy should not override this and simply use the strategy defined for the service instead of the default.

rutgerbrf:
@yurishkuro The build seems to be failing where it shouldn't?

yurishkuro:
@rutgerbrf thanks for the PR! 

PTAL at the respective docs change: https://github.com/jaegertracing/documentation/pull/346

Issue #1754, opened 2019-08-22
Cannot connect to query-service via gRPC


yurishkuro:
Reported by @32leaves on Gitter.

## Requirement - what kind of business use case are you trying to solve?

Programmatically retrieve traces via gRPC endpoint on all-in-one.

## Problem - what in Jaeger blocks you from solving the requirement?

https://gist.github.com/32leaves/8d422784ad7cb105ff4b1a36e8726222

gRPC client cannot complete the handshake with the server.

## Proposal - what do you suggest to solve the problem or improve the existing situation?

Looks like a bug. We should have a working example and a unit test.

## Any open questions to address

At Uber we have an internal client library that works, but we're running gRPC server on a separate port from HTTP server, so the issue might be with the cmux.

PR #1756, opened 2019-08-23
Fix gRPC query service cmux

32leaves:
fixes https://github.com/jaegertracing/jaeger/issues/1754

breaking change: https://github.com/grpc/grpc-go/issues/2406
workaround described in:
- https://github.com/soheilhy/cmux/issues/64
- https://github.com/soheilhy/cmux#limitations

### DIFF
--- a/cmd/query/app/server.go
+++ b/cmd/query/app/server.go
@@ -102,9 +102,7 @@ func (s *Server) Start() error {
 	// cmux server acts as a reverse-proxy between HTTP and GRPC backends.
 	cmuxServer := cmux.New(s.conn)
 
-	grpcListener := cmuxServer.Match(
-		cmux.HTTP2HeaderField("content-type", "application/grpc"),
-		cmux.HTTP2HeaderField("content-type", "application/grpc+proto"))
+	grpcListener := cmuxServer.MatchWithWriters(cmux.HTTP2MatchHeaderFieldSendSettings("content-type", "application/grpc"))
 	httpListener := cmuxServer.Match(cmux.Any())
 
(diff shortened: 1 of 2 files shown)


yurishkuro:
Thanks for the research & the fix, @32leaves .

I would like to have a unit test that uses grpc client to hit grpc endpoint via cmux, so that we don't have regressions like these in the future.

yurishkuro:
I don't think this fix is helping, I added a unit test #1758 for connecting to grpc server via cmux, and it's failing both before and after the change to matchers. Also, the behavioral change in https://github.com/grpc/grpc-go/issues/2406 was only released in grpc-go 1.23, but we're currently pinning grpc to 1.21.1

yurishkuro:
cc @annanay25

32leaves:
I just pushed the test I used while working on the fix (didn't see you had already added a test as well - I'm happy to remove mine again). With this fix I can successfully call `GetServices` using the gRPC interface.

annanay25:
Thanks for the PR @32leaves. Travis is failing with - 
```
cmd/query/app/server_test.go:69:97: grpc.WithTimeout is deprecated: use DialContext and context.WithTimeout instead.  (SA1019)
```

We should be able to go forward and land once this is fixed :)  
Ref: #1758

yurishkuro:
My mistake, my PR was using the wrong port.

It's fixed now, I prefer to merge my PR because it includes a bunch of other clean-up. I pulled the main commit from here for the actual fix.

yurishkuro:
superseded by #1758

PR #1758, opened 2019-08-24
Fix gRPC over cmux and add unit tests

### LINKED ISSUE #1754: Cannot connect to query-service via gRPC

yurishkuro:
Resolves #1754

Includes the main commit by @32leaves from #1756

### DIFF
--- a/cmd/query/app/server_test.go
+++ b/cmd/query/app/server_test.go
@@ -60,32 +60,19 @@ func TestServer(t *testing.T) {
 		opentracing.NoopTracer{})
 	assert.NoError(t, server.Start())
 
+	// wait for the server to come up
+	// TODO find a way to not wait full 1s, only as long as needed
 	time.Sleep(1 * time.Second)
 
 	client := newGRPCClient(t, fmt.Sprintf(":%d", ports.QueryHTTP))
 	defer client.conn.Close()
 
-	var queryErr error
-	for i := 0; i < 1; i++ {
-		queryErr = func() error {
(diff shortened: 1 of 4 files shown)


cmd/query/app/server_test.go
```diff
@@ -43,13 +47,40 @@ func TestServer(t *testing.T) {
...
+	dependencyReader := &depsmocks.Reader{}
+
+	querySvc := querysvc.NewQueryService(spanReader, dependencyReader, querysvc.QueryServiceOptions{})
+
 	tracer := opentracing.NoopTracer{}
 
-	server := NewServer(flagsSvc, querySvc, &QueryOptions{Port: ports.QueryAdminHTTP,
-		BearerTokenPropagation: true}, tracer)
+	server := NewServer(flagsSvc, querySvc,
+		&QueryOptions{Port: ports.QueryAdminHTTP, BearerTokenPropagation: true},
```
annanay25: Changing the listening port here to -

```
		&QueryOptions{Port: ports.QueryHTTP, BearerTokenPropagation: true},
```

.. and using `cmuxServer.MatchWithWriters(...)` as suggested by @32leaves seems to work for me. 

We definitely needed an integration test with cmux though, thanks for pointing out!

PR #598, opened 2017-12-16
Add CA certificates to collector/query images

yurishkuro:
Resolves #485 

When connecting query/collector to storage backends that use TLS with certificates signed by public root CAs, it's useful to have those CA certificates in the Docker image.

On the other hand, the certificates can always be mounted from the host filesystem, which is necessary anyway when connecting to servers using self-signed certificates.

I think this is a nice to have so that people don't have to do extra work. For collector it increases the image size from 24.6MB to 25.3MB.

PR #1764, opened 2019-08-27
Add CA certificates to agent image

### LINKED ISSUE #1760: Jaeger Agent container should trust certificates from public CAs by default
We use public certificates to secure our gRPC communicates on the Jaeger collector. These are signed by trusted Certificate Authorities, so the expectation is that the agent would trust them by default.  However, these are not trusted today by the Jaeger agent. 

CA certificates are not loaded into the jaeger-agent container.

Load the cacerts package from alpine similar to how OpenTelemetry service does it.
https://github.com/open-telemetry/opentelemetry-service/blob/master/cmd/otelsvc/Dockerfile#L5

yurishkuro:
Resolves #1760, similar to #598.

@jpkrohling @pavolloffay please review

### DIFF
--- a/cmd/agent/Dockerfile
+++ b/cmd/agent/Dockerfile
@@ -1,4 +1,9 @@
+FROM alpine:latest as certs
+RUN apk add --update --no-cache ca-certificates
+
 FROM scratch
+
+COPY --from=certs /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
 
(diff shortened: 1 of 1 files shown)


jpkrohling:
For the record, the image is just marginally bigger:

Before this change:
```
jaegertracing/jaeger-agent                        latest                                     aac8239a5cf3        54 seconds ago      26.9MB
```

After this change:
```
jaegertracing/jaeger-agent                        latest                                     883a55bf3887        19 seconds ago       27.1MB
```

PR #1766, opened 2019-09-01
WIP - Adds sqlite storage

fcvarela:
- Adds an SQLite storage backend for traces thereby providing an alternative to badger. I realise Jaeger already has a solution for local storage using badger so understand if this contribution isn't welcome.

- The PR adds a new storage backend using SQLite.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -584,6 +584,14 @@
   revision = "1ea4449da9834f4d333f1cc461c374aea217d249"
 
 [[projects]]
+  digest = "1:79e87abf06b873987dee86598950f5b51732ac454d5a5cab6445a14330e6c9e3"
+  name = "github.com/mattn/go-sqlite3"
+  packages = ["."]
+  pruneopts = "UT"
+  revision = "b612a2feea6aa87c6d052d9086572551df06497e"
+  version = "v1.11.0"
+
+[[projects]]
   digest = "1:ff5ebae34cfbf047d505ee150de27e60570e8c394b3b8fdbb720ff6ac71985fc"
   name = "github.com/matttproud/golang_protobuf_extensions"
   packages = ["pbutil"]
@@ -1281,6 +1289,8 @@
     "github.com/bsm/sarama-cluster",
     "github.com/crossdock/crossdock-go",
     "github.com/dgraph-io/badger",
+    "github.com/dgraph-io/badger/options",
+    "github.com/fsnotify/fsnotify",
     "github.com/go-openapi/errors",
     "github.com/go-openapi/loads",
     "github.com/go-openapi/runtime",
@@ -1309,6 +1319,7 @@
     "github.com/hashicorp/go-hclog",
     "github.com/hashicorp/go-plugin",
     "github.com/kr/pretty",
+    "github.com/mattn/go-sqlite3",
     "github.com/olivere/elastic",
     "github.com/opentracing-contrib/go-stdlib/nethttp",
     "github.com/opentracing/opentracing-go",
--- a/plugin/storage/factory.go
+++ b/plugin/storage/factory.go
@@ -30,12 +30,14 @@ import (
 	"github.com/jaegertracing/jaeger/plugin/storage/grpc"
 	"github.com/jaegertracing/jaeger/plugin/storage/kafka"
 	"github.com/jaegertracing/jaeger/plugin/storage/memory"
+	"github.com/jaegertracing/jaeger/plugin/storage/sqlite"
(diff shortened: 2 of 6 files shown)


fcvarela:
I was hoping I'd be able to tag this WIP until I add support for dependencies but it seems I can't use any tags. Please feel free to review but understand this isn't yet complete.

yurishkuro:
Can you elaborate why you want it? Pros/Cons?

fcvarela:
> Can you elaborate why you want it? Pros/Cons?

 While trying out badger I was underwhelmed with its performance and thought it would be interesting to write an alternative backend to benchmark against. Having done that I could see no reason not to share it.

This could also be moved to a separate project as a gRPC plugin but that would defeat the whole purpose of SQLite which is to embed it with the program that uses it.

The cons are that it requires CGO so full integration would mean tweaking some of the makefiles and getting around possible issues w/ Alpine/musl.

burmanm:
I wonder what was your issue with the badger performance? On my machine I can write around ~30 000 spans / s to disk which far exceeds the capabilities of what my other parts of Jaeger can do. 

Although I do notice that your PR does not actually store to the disk:

```go
if _, err := db.Exec("PRAGMA journal_mode = MEMORY;", nil); err != nil {
```

So there's reduced integrity - the badger instance will store it to the disk always (even if application crashes, the WriteSync option will only reduce the amount of fsyncs to the physical disk). Are you testing the same things here?

fcvarela:
>I wonder what was your issue with the badger performance? On my machine I can write around ~30 000 spans / s to disk which far exceeds the capabilities of what my other parts of Jaeger can do.

I wonder if those are example spans or representative of production workloads with tags, logs, etc? 

> Although I do notice that your PR does not actually store to the disk:

That option controls where the transaction journal is kept, not the actual data. The actual value for the option should be user provided according to how they assess the risk/benefit.

To be clear, the current options are for benchmarking purposes and do persist to disk.

burmanm:
From sqlite3 documentation: ``The MEMORY journaling mode stores the rollback journal in volatile RAM``, now journaling log / WAL / transaction log is the key to the integrity of a SQL database. The tables in the disk are not written after every write (as they need to be rebuit) but instead only during the checkpointing. I don't see SQLite doing this process any differently.

It's easy to check. Keep writing data, kill the process and reload. If you don't have all the transactions, you never wrote them to the disk. And this is what happens. It's great for benchmarketing purposes, sure, but only to replace in-memory storage. The WAL is a bottleneck for many DBs as those operations need to be acked from the disk interface before they can be acked back to the client. And they're raced by all the concurrent writers.

But, that aside, looking at the implementation - I noticed you don't write all the data to the SQLite. So the original span is not written anywhere? I guess you could rely on just the tables to rebuild them, but in that case you need to write all the fields (some are missing now). Logs need indexing (and storing) also, as well as duration (unless you sacrifice reading performance for write performance). Also "index spans by service_name, operation_name, start_time" is missing start_time in the actual index. 

And yes, the amount of indexed tags plays a big role in badger (each one is a new write). So with 128 unique tags values per span I get 3k/s with large amount of time spent in the kernel processes. Although, none of this actually stresses my machine so there's clearly a bottleneck in some serial process, disk or something else.

fcvarela:
I had mentioned this earlier, but the WAL mode and journal modes should be provided by the client when the PR is complete. I believe it's up to them to be the judges of risk/benefit of each combination for their use case. This, of course, means exposing the options and documenting them, which I have not done yet.

> But, that aside, looking at the implementation - I noticed you don't write all the data to the SQLite. So the original span is not written anywhere?
I'm not sure I understand, the spans are stored and I can query and see them in the UI. Do you mean actually marshalling the model.Span struct and store it in some field in the db for faster recreation at query time?

> you need to write all the fields (some are missing now).
This is why I tagged the PR "WIP". At the moment, as you mentioned, durations aren't indexed and logs aren't stored. They don't need indexing as there's no way of querying for them.

> Also "index spans by service_name, operation_name, start_time" is missing start_time in the actual index.
Good catch, thanks!

fcvarela:
By the way, this isn't obvious and I should have mentioned it before.

- I'm not claiming this backend is faster than badger. I am writing it to benchmark against badger and submitted the PR in case someone is interested.
- The choice of wal mode and synchronous mode should be exposed to the end user as they will be better informed about which makes more sense to their use case.
- The benchmarks journal in memory but do persist to disk. Losing the journal during a restart is not something I consider an issue. Losing some spans or traces should be statistically insignificant and would happen anyway during a restart of an all-in-one deployment.

jpkrohling:
Closing due to inactivity.

Issue #1781, opened 2019-09-07
Unpin gRPC


yurishkuro:
For some reason #1492 pinned gRPC to a specific release, we need to unpin it. It will require resolving these issues:
```
Detected staticcheck failures:
cmd/agent/app/reporter/grpc/builder.go:133:36: grpc.WithBalancerName is deprecated: use WithDefaultServiceConfig and WithDisableServiceConfig instead.  Will be removed in a future 1.x release.  (SA1019)
pkg/discovery/grpcresolver/grpc_resolver_test.go:146:78: grpc.WithBalancerName is deprecated: use WithDefaultServiceConfig and WithDisableServiceConfig instead.  Will be removed in a future 1.x release.  (SA1019)
make: *** [lint-staticcheck] Error 1
```

guanw:
@yurishkuro I can work on this?

guanw:
`WithDefaultServiceConfig` introduces a json string `serviceConfig` that looks like [this](https://github.com/grpc/grpc/blob/master/doc/service_config.md), should we also introduce a flag for this which defaults to `{"loadBalancingPolicy":"round_robin"}`

guanw:
Also when I comment out `override` section for `google.golang.org/grpc` in toml file and do `dep ensure -update google.golang.org/grpc`, the following error shows (incomplete log):
![image](https://user-images.githubusercontent.com/7311035/64489715-564c3000-d224-11e9-82aa-47f581b96286.png)

Looks like @objectiser introduces the revision. Any idea we are pinning `jaeger-client-go` to `6733ee486c780528f2c8088305e16fdb685134c7` (Can we remove this constraint first)?

yurishkuro:
I already have a PR to remove client pin: #1780

yurishkuro:
> WithDefaultServiceConfig introduces a json string serviceConfig

I assume json is an option, not a requirement, i.e. do we still have programmatic way to configure? I prefer we fix the main issue first, not introduce new configuration options.

guanw:
@yurishkuro See https://github.com/grpc/grpc-go/issues/3003, looks like that's the only recommended way. We can put a `const serviceConfig = <service config json string>` for now to avoid setting flags.

Issue #1713, opened 2019-08-06
Flaky ES test - TestSpanReader_multiRead_followUp_query


yurishkuro:
defined in plugin/storage/es/spanstore/reader_test.go:208

```
=== RUN   TestSpanReader_multiRead_followUp_query
--- FAIL: TestSpanReader_multiRead_followUp_query (0.00s)
panic: 
mock: Unexpected Method Call
-----------------------------
Index(string,string)
		0: "jaeger-span-2019-08-06"
		1: "jaeger-span-2019-08-05"
The closest call I have is: 
Index(mock.AnythingOfTypeArgument)
		0: "string"
Provided 1 arguments, mocked for 2 arguments
Diff: 1: FAIL:  (string=jaeger-span-2019-08-05) != (Missing) [recovered]
	panic: 
mock: Unexpected Method Call
-----------------------------
Index(string,string)
		0: "jaeger-span-2019-08-06"
		1: "jaeger-span-2019-08-05"
The closest call I have is: 
Index(mock.AnythingOfTypeArgument)
		0: "string"
Provided 1 arguments, mocked for 2 arguments
Diff: 1: FAIL:  (string=jaeger-span-2019-08-05) != (Missing)
goroutine 22 [running]:
testing.tRunner.func1(0xc00010a700)
	/home/travis/.gimme/versions/go1.12.1.linux.amd64/src/testing/testing.go:830 +0x69d
panic(0xca69c0, 0xc000135c00)
	/home/travis/.gimme/versions/go1.12.1.linux.amd64/src/runtime/panic.go:522 +0x1b5
github.com/jaegertracing/jaeger/vendor/github.com/stretchr/testify/mock.(*Mock).fail(0xc0000d5180, 0xd9d701, 0x6f, 0xc000139b40, 0x4, 0x4)
	/home/travis/gopath/src/github.com/jaegertracing/jaeger/vendor/github.com/stretchr/testify/mock/mock.go:236 +0x1ad
github.com/jaegertracing/jaeger/vendor/github.com/stretchr/testify/mock.(*Mock).MethodCalled(0xc0000d5180, 0x1394873, 0x5, 0xc00016c2c0, 0x2, 0x2, 0xc000139b00, 0x4, 0x4)
	/home/travis/gopath/src/github.com/jaegertracing/jaeger/vendor/github.com/stretchr/testify/mock/mock.go:358 +0xab5
[...]

PR #1786, opened 2019-09-09
(WIP) Unpin grpc version and use serviceConfig to set loadbalancer 

guanw:

- #1781 

- This PR aims to unpin grpc and replace deprecated code

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -204,15 +204,15 @@
   version = "v0.5.0"
 
 [[projects]]
-  digest = "1:87082ea822adf4db51d1bec2a92c9fe0a38b56729bbcd7d036a36848a9bb2b97"
+  digest = "1:adbc24353f685b2045b5881bb15f48cec5267fc70a5ecaf2cf60b1241a43c83a"
   name = "github.com/go-openapi/analysis"
   packages = [
     ".",
     "internal",
   ]
   pruneopts = "UT"
-  revision = "48f4521ad6c3df9d6192efb04287c4a411e1f806"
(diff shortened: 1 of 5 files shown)


guanw:
I notice there's a new commit detected in git for `idl/` and not sure if I should commit it.

guanw:
@pavolloffay One question regarding the `TestSpanReader_multiRead_followUp_query`, after upgrading lock file this test starts to fail where mocks `Index` method was expecting two string instead of one. Any idea why?

guanw:
@pavolloffay Actually it feels like `TestSpanReader_multiRead_followUp_query` is being flaky since the `s.timeRangeIndices(s.spanIndexPrefix, startTime.Add(-time.Hour), endTime.Add(time.Hour))` could compute slice with either one entry or two based on the time zone and time.Now(). I think you need to add a second `mock.AnythingOfType("string")` to `.On("Index", ...)` mock function.

pavolloffay:
> I notice there's a new commit detected in git for idl/ and not sure if I should commit it.

If it is not related to this change it should be reverted.

pavolloffay:
> @pavolloffay One question regarding the TestSpanReader_multiRead_followUp_query(introduced in #1677 ), after upgrading lock file this test starts to fail where mocks Index method was expecting two string instead of one. Any idea why?

The changes here does not seem to be related to the ES reader. The test is probably flaky. Could oyu please report that in a separate issue?

yurishkuro:
There is already an issue for that test: https://github.com/jaegertracing/jaeger/issues/1713

PR #1800, opened 2019-09-14
cmd/query: init tracer config from env

### LINKED ISSUE #1044: Allow configuration of jaeger client in the query service

holycheater:
Resolves #1044 

Getting a lot of these errors in jaeger-query docker image:
```{"level":"error","ts":1568465347.7179556,"caller":"zap/logger.go:33","msg":"error when flushing the buffer: write udp 127.0.0.1:59514->127.0.0.1:6831: write: connection refused","stacktrace":"github.com/jaegertracing/jaeger/vendor/github.com/uber/jaeger-client-go/log/zap.(*Logger).Error\n\t/home/travis/gopath/src/github.com/jaegertracing/jaeger/vendor/github.com/uber/jaeger-client-go/log/zap/logger.go:33\ngithub.com/jaegertracing/jaeger/vendor/github.com/uber/jaeger-client-go.(*remoteReporter).processQueue.func1\n\t/home/travis/gopath/src/github.com/jaegertracing/jaeger/vendor/github.com/uber/jaeger-client-go/reporter.go:257\ngithub.com/jaegertracing/jaeger/vendor/github.com/uber/jaeger-client-go.(*remoteReporter).processQueue\n\t/home/travis/gopath/src/github.com/jaegertracing/jaeger/vendor/github.com/uber/jaeger-client-go/reporter.go:267"}```

Using jaeger-client-go `FromEnv` configuration

### DIFF
--- a/cmd/query/main.go
+++ b/cmd/query/main.go
@@ -64,14 +64,11 @@ func main() {
 			baseFactory := svc.MetricsFactory.Namespace(metrics.NSOptions{Name: "jaeger"})
 			metricsFactory := baseFactory.Namespace(metrics.NSOptions{Name: "query"})
 
-			tracer, closer, err := jaegerClientConfig.Configuration{
-				ServiceName: "jaeger-query",
-				Sampler: &jaegerClientConfig.SamplerConfig{
-					Type:  "probabilistic",
-					Param: 1.0,
-				},
-				RPCMetrics: true,
-			}.NewTracer(
+			cfg, err := app.TracerConfig()
+			if err != nil {
+				logger.Fatal("Failed to read tracer config from env", zap.Error(err))
+			}
+			closer, err := cfg.InitGlobalTracer("jaeger-query",
 				jaegerClientConfig.Metrics(svc.MetricsFactory),
 				jaegerClientConfig.Logger(jaegerClientZapLog.NewLogger(logger)),
 			)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: cmd/query/main.go [hunk 1]
```diff
@@ -64,22 +64,18 @@ func main() {
...
 
-			tracer, closer, err := jaegerClientConfig.Configuration{
-				ServiceName: "jaeger-query",
-				Sampler: &jaegerClientConfig.SamplerConfig{
-					Type:  "probabilistic",
-					Param: 1.0,
-				},
-				RPCMetrics: true,
-			}.NewTracer(
+			cfg, err := jaegerClientConfig.FromEnv()
```
yurishkuro: Unfortunately, this is not backwards compatible. This really needs to be fixed in the client, so that FromEnv could be called on an already populated config as a for of override.

holycheater: Making changes to client for this single application seems like an overkill
 I've made changes in tracer config initialization, should be backwards-compatible now

yurishkuro:
Sorry, I was probably unclear. I don't think _this_ repository needs to solve it by re-implementing env var lookup. The issue is in the jaeger-client-go: https://github.com/jaegertracing/jaeger-client-go/issues/430

pavolloffay:
Done in https://github.com/jaegertracing/jaeger/pull/1919

Issue #1821, opened 2019-09-28
gRPC plugin framework does not respect --query.bearer-token-propagation flag


radekg:
## Requirement - what kind of business use case are you trying to solve?

I am trying to run elastic search storage back end via GRPC plugin and have the bearer token forwarded to the GRPC plugin server.

## Problem - what in Jaeger blocks you from solving the requirement?

Query UI has an option to forward Bearer token to the storage back end using `--query.bearer-token-propagation=true` flag. However, when using the GRPC storage plugin, the context created with `spanstore.ContextWithBearerToken(ctx, token)` in https://github.com/jaegertracing/jaeger/blob/master/cmd/query/app/token_propagation_handler.go#L48 is not passed to the GRPC plugin.

Currently, the last place in the storage where the context with the `bearer.token` can be reached, is the GRPC client: https://github.com/jaegertracing/jaeger/blob/master/plugin/storage/grpc/shared/grpc_client.go.

This can be tested by adding these lines of code:

```golang
  str, ok := spanstore.GetBearerToken(ctx)
  fmt.Println(fmt.Sprintf(" =====================> CLIENT: The context is: %v ::: %t", str, ok))
```

to any of the following methods:

- `func (c *grpcClient) GetTrace(ctx context.Context, traceID model.TraceID) (*model.Trace, error)`
- `func (c *grpcClient) GetServices(ctx context.Context) ([]string, error)`
- `func (c *grpcClient) GetOperations(ctx context.Context, service string) ([]string, error)`
- `func (c *grpcClient) FindTraces(ctx context.Context, query *spanstore.TraceQueryParameters) ([]*model.Trace, error)`
- `func (c *grpcClient) FindTraceIDs(ctx context.Context, query *spanstore.TraceQueryParameters) ([]model.TraceID, error)`

However, on the server part of the plugin, the context is a new instance because the plugin is running as a separate process.

## Proposal - what do you suggest to solve the problem or improve the existing situation?

To fix this problem, the bearer token has to be passed inside of the GRPC requests to the GRPC server and the `ContextWithBearerToken` must be reconstructed on the other side.

yurishkuro:
Out of curiosity, which storage backend are you using via grpc plugin that requires bearer token?

radekg:
[...]

Issue #1763, opened 2019-08-27
Concurrency model of gRPC plugins


sergiimk:
## Requirement - what kind of business use case are you trying to solve?

We are implementing a custom gRPC-based storage plugin as per [this doc](https://github.com/jaegertracing/jaeger/blob/master/plugin/storage/grpc/README.md).

## Problem - what in Jaeger blocks you from solving the requirement?

Could you please clarify what is the concurrency model of a gRPC plugin in Jaeger. I'm getting following confusing results from my experiments:

* When loading main page Jager will usually call `GetServices` and `GetOperations` methods on a plugin. I can see that these two calls are executed in parallel, so if each call takes ~3s to fetch data from the DB the total load time will be 3s.

* When several clients load main page simultaneously (refreshing main page from many browser tabs at once) the calls to the plugin seems to be linearized, i.e. `GetServices` call made as part of serving data for the first client will block, and only after it returns the `GetServices` call for the second client will start executing.

Having observed the first behavior locally the second case came as a big surprise. Our pluging is concurrency safe, has no mutexes, and able to execute all operations in parallel. I have also verified this behavior with `GetServices/GetOperations` returning a pre-canned responses and adding `time.Sleep()`.

Impact: We already implemented a sophisticated caching logic in our plugin for services and operations, but this head-of-line blocking behavior for requests that cannot be cached can result in a very poor user experience when Jaeger is exposed to multiple users.

PR #1822, opened 2019-09-28
Propagate the bearer token to the gRPC plugin server

radekg:
This PR attempts to resolve the problem described in https://github.com/jaegertracing/jaeger/issues/1821.

### Problem summary

Query UI has an option to forward Bearer token to the storage back end using `--query.bearer-token-propagation=true` flag. However, when using the gRPC storage plugin, the context created with `spanstore.ContextWithBearerToken(ctx, token)` in https://github.com/jaegertracing/jaeger/blob/master/cmd/query/app/token_propagation_handler.go#L48 is not passed to the gRPC plugin.

Currently, the last place in the storage where the context with the `bearer.token` can be reached, is the gRPC client: https://github.com/jaegertracing/jaeger/blob/master/plugin/storage/grpc/shared/grpc_client.go.

This can be tested by adding these lines of code:

```golang
  str, ok := spanstore.GetBearerToken(ctx)
  fmt.Println(fmt.Sprintf(" =====================> CLIENT: The context is: %v ::: %b", str, ok))
```

to any of the following methods:

- `func (c *grpcClient) GetTrace(ctx context.Context, traceID model.TraceID) (*model.Trace, error)`
- `func (c *grpcClient) GetServices(ctx context.Context) ([]string, error)`
- `func (c *grpcClient) GetOperations(ctx context.Context, service string) ([]string, error)`
- `func (c *grpcClient) FindTraces(ctx context.Context, query *spanstore.TraceQueryParameters) ([]*model.Trace, error)`
- `func (c *grpcClient) FindTraceIDs(ctx context.Context, query *spanstore.TraceQueryParameters) ([]model.TraceID, error)`

However, on the server part of the plugin, the context is a new instance because the plugin is running as a separate process.

### Solution summary

This pull request adds support for bearer token forwarding to the gRPC `spanstore.Reader` methods of the gRPC storage. Similarly to the solution from the ES storage provider, only forward the bearer token when the `--query.bearer-token-propagation` flag has been set to `true`.


plugin/storage/grpc/proto/storage.proto
```diff
@@ -66,16 +66,20 @@ message GetTraceRequest {
       (gogoproto.customtype) = "github.com/jaegertracing/jaeger/model.TraceID",
       (gogoproto.customname) = "TraceID"
     ];
+    string bearerToken = 2;
 }
 
-message GetServicesRequest {}
+message GetServicesRequest {
+    string bearerToken = 1;
```
yurishkuro: I don't think it's a good idea to explicitly change the storage API for this cross-cutting concern. gRPC requests can pass metadata similar to HTTP headers, I think it would be a much better place for auth tokens.

plugin/storage/grpc/shared/grpc_client.go
```diff
@@ -29,9 +29,20 @@ import (
 
 // grpcClient implements shared.StoragePlugin and reads/writes spans and dependencies
 type grpcClient struct {
-	readerClient     storage_v1.SpanReaderPluginClient
-	writerClient     storage_v1.SpanWriterPluginClient
-	depsReaderClient storage_v1.DependenciesReaderPluginClient
+	allowTokenFromContext bool
```
yurishkuro: is the config option actually required? If the query-service is configured to pass the bearer token, it would include it in the context, and this code just needs to respect that.

plugin/storage/grpc/shared/grpc_server.go
```diff
@@ -53,7 +64,8 @@ func (s *grpcServer) WriteSpan(ctx context.Context, r *storage_v1.WriteSpanReque
 
 // GetTrace takes a traceID and streams a Trace associated with that traceID
 func (s *grpcServer) GetTrace(r *storage_v1.GetTraceRequest, stream storage_v1.SpanReaderPlugin_GetTraceServer) error {
-	trace, err := s.Impl.SpanReader().GetTrace(stream.Context(), r.TraceID)
+	opCtx := s.getMaybeBearerTokenContext(stream.Context(), r.BearerToken)
```
yurishkuro: this may not be needed per my earlier comments, but if needed then it would be better to rename to a more general method of "adapting" the context from request, so that we could, in the future, support different types of middleware.

plugin/storage/grpc/shared/interface.go
```diff
@@ -34,9 +34,13 @@ var Handshake = plugin.HandshakeConfig{
 	MagicCookieValue: "jaeger",
 }
 
-// PluginMap is the map of plugins we can dispense.
-var PluginMap = map[string]plugin.Plugin{
-	StoragePluginIdentifier: &StorageGRPCPlugin{},
+// GetPluginMap returns a plugin map.
+func GetPluginMap(allowTokenFromContext bool) map[string]plugin.Plugin {
```
yurishkuro: if we can get away without explicit configuration, please undo this

radekg:
Hi @yurishkuro, thank you for a quick review. I have updated the code according to your request / suggestions. Indeed, this is a way cleaner solution than the original one.

plugin/storage/grpc/proto/storage.proto
radekg: Indeed, updated code does not do this anymore.

plugin/storage/grpc/shared/grpc_client.go
```diff
@@ -34,6 +35,18 @@ type grpcClient struct {
 	depsReaderClient storage_v1.DependenciesReaderPluginClient
 }
 
+// updateContextWithBearerToken updates the outgoing context with a bearer token extracted from the source
+func (c *grpcClient) updateContextWithBearerToken(outgoing context.Context, source context.Context) context.Context {
```
radekg: I am not sure what would be a cleaner way to handle this. The problem here is that `FindTraces` and `FindTraceIDs` uses `context.Background()` so this function has to operate on two contexts.

plugin/storage/grpc/shared/grpc_client.go
```diff
@@ -34,6 +35,18 @@ type grpcClient struct {
...
+	bearerToken, hasToken := spanstore.GetBearerToken(source)
+	if hasToken {
+		requestMetadata := metadata.New(map[string]string{
+			"bearerToken": bearerToken,
```
radekg: A suggestion here. There is a place in the code where the property identifying this setting is already defined. It is: https://github.com/jaegertracing/jaeger/blob/master/storage/spanstore/token_propagation.go#L21.
Maybe we could do something like:

```golang
const BearerTokenKey = "bearer.token"
const bearerToken = contextKey("bearer.token")
```

and use the `spanstore.BearerTokenKey` here and then in the gRPC plugin server?

plugin/storage/grpc/shared/grpc_client.go
```diff
@@ -96,7 +109,7 @@ func (c *grpcClient) GetOperations(ctx context.Context, service string) ([]strin
 
 // FindTraces retrieves traces that match the traceQuery
 func (c *grpcClient) FindTraces(ctx context.Context, query *spanstore.TraceQueryParameters) ([]*model.Trace, error) {
-	stream, err := c.readerClient.FindTraces(context.Background(), &storage_v1.FindTracesRequest{
```
yurishkuro: I am not sure why this is using Background instead of passing the `ctx` arg directly. Is there any reasoning in git blame?  If not, I think it's a bug / oversight.

plugin/storage/grpc/shared/grpc_client.go
yurishkuro: +1

plugin/storage/grpc/shared/grpc_client.go
radekg: That’s what I originally also thought. I think this is an oversight for `FindTraceIDs` because the server returns a response for that call: https://github.com/jaegertracing/jaeger/blob/master/plugin/storage/grpc/shared/grpc_server.go#L132. However, `FindTraces` does chunking: https://github.com/jaegertracing/jaeger/blob/master/plugin/storage/grpc/shared/grpc_server.go#L107. Hence the `context.Background()`.

I’m fairly confident `FindTraceIDs` context can be changed, not so sure about `FindTraces`. I could investigate in a separate PR. I’ll have to revisit the gRPC storage anyway because I’ll also hit this: https://github.com/jaegertracing/jaeger/issues/1763

What do you think, @yurishkuro?

plugin/storage/grpc/shared/grpc_client.go
radekg: Will do. Final question: should I squash before it is merged?

yurishkuro: No need, we squash on merge

plugin/storage/grpc/shared/grpc_client.go
yurishkuro: Not sure I follow how chunking is related to using Background().

radekg: I’ve seen hanging plugins when doing bidirectional calls. But here is no bidirectional communication. I’ve got plenty confused here. Code adapted to use the original context.

plugin/storage/grpc/shared/grpc_client.go
jpkrohling: We do want a way to tell our components that the token shouldn't be propagated.

yurishkuro: There is already configuration to that effect in the query service, which controls token handoff from query to storage. I don't think yet another flag for storage-storagePlugin is needed.

storage/spanstore/token_propagation.go
```diff
@@ -18,6 +18,8 @@ import "context"
 
 type contextKey string
 
+// BearerTokenKey is the string literal used internally in the implementation of this context.
+const BearerTokenKey = "bearer.token"
```
yurishkuro: next line can use this new constant

plugin/storage/grpc/shared/grpc_client.go
```diff
@@ -34,6 +35,20 @@ type grpcClient struct {
 	depsReaderClient storage_v1.DependenciesReaderPluginClient
 }
 
+// upgradeContextWithBearerToken turns the context into a gRPC outgoing context with bearer token
+// in the request metadata, if the original context has bearer token attached.
+// Otherwise returns original context.
+func (c *grpcClient) upgradeContextWithBearerToken(ctx context.Context) context.Context {
```
yurishkuro: this function does not need a receiver

jpkrohling:
Looks like there's one or more files that need formatting. The CI job [failed](https://travis-ci.org/jaegertracing/jaeger/jobs/592120509#L543) with:

```
```

radekg:
All fixed and signed off.

yurishkuro:
This code is not covered: 
image

radekg:
Hi @yurishkuro, I'm not sure if the failing tests are related to my changes. I can see cql related fatal errors in the Travis job.

pavolloffay:
thanks @radekg

radekg:
Thank you.

Issue #1858, opened 2019-10-14
Problems upgrading to ES7 (AWS)


dcelasun:
## Requirement - what kind of business use case are you trying to solve?

I'm trying to follow [the documentation](https://www.jaegertracing.io/docs/1.14/deployment/#upgrade-to-elasticsearch-7-x) on upgrading to ES7.

## Problem - what in Jaeger blocks you from solving the requirement?

According to the docs, I should be able the start the collector in ES7 mode while connecting to ES 6.8.

> Elasticsearch 6.8 supports 7.x, 6.x, 5.x compatible mappings. The upgrade has to be done first to ES 6.8, then apply data migration or wait until old daily indices are removed (this requires to start Jaeger with `--es.version=7` to force using ES 7.x mappings for newly created indices).

However, trying to connect to ES 6.8 (AWS managed ES, if that matters) with `--es.version=7` fails with:

```sh
$ ./collector-linux version
{"gitCommit":"fb5505005a21f007792dedbcd2ad49484d1d587e","GitVersion":"v1.14.0","BuildDate":"2019-10-11T16:39:14Z"}

$ ES_SERVER_URLS=https://example.com SPAN_STORAGE_TYPE=elasticsearch ./collector-linux --es.version=7
2019/10/14 10:11:12 maxprocs: Leaving GOMAXPROCS=12: CPU quota undefined
{"level":"info","ts":1571044272.9358723,"caller":"flags/service.go:115","msg":"Mounting metrics handler on admin server","route":"/metrics"}
{"level":"info","ts":1571044272.9359663,"caller":"flags/admin.go:108","msg":"Mounting health check on admin server","route":"/"}
{"level":"info","ts":1571044272.9360156,"caller":"flags/admin.go:114","msg":"Starting admin HTTP server","http-port":14269}
{"level":"info","ts":1571044272.9360237,"caller":"flags/admin.go:100","msg":"Admin server started","http-port":14269,"health-status":"unavailable"}
{"level":"fatal","ts":1571044273.0986753,"caller":"collector/main.go:95","msg":"Failed to create span writer","error":"elastic: Error 400 (Bad Request): Malformed [mappings] section for type [dynamic_templates], should include an inner object describing the mapping [type=illegal_argument_exception]","stacktrace":"<redacted>"}
```

pavolloffay:
Could you please paste the response of ES ping endpoint? `http://es:9200/`

pavolloffay:
[...]

PR #1860, opened 2019-10-14
Use ES 7.x mappings when ES 6.8 detected

jan25:
## Which problem is this PR solving
- Resolves https://github.com/jaegertracing/jaeger-operator/issues/634
- Mappings for ES7.x are compatible with ES6.8. This PR enforces ES7 mappings when ES >=6.8 version is detected


pavolloffay:
Unfortunately, this will not work see https://github.com/jaegertracing/jaeger/issues/1858. 

We will also need an integration test to verify the behavior.

jan25:
Oh. Thanks for linking that issue. I'll do more manual tests and setup a integration test

pavolloffay:
This is not needed. ES7 indices can read indices created by ES6 by default.

pavolloffay:
@jan25 thanks for looking into this!

PR #1872, opened 2019-10-21
Use Elasticsearch 6 in xdock

pavolloffay:
This has been discussed during bi-weekly call. The motivation is to get more reliable and faster xdock tests.  


pavolloffay:
We have a green build. The run time is almost the same. I will restart it to see whether it is stable or not.

PR #1884, opened 2019-10-29
Bump Jaeger dependency on badger to 1.6.0 and resolve breaking changes

andreynkravtsov:
- Need to upgrade badger dependency to 1.6.0

- DefaultOptions is now a function with path default passed in
- Iterator now seems unhappy when you do it.Item() != null and expects to be checked against .Valid()

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -137,7 +137,7 @@ required = [
 
 [[constraint]]
   name = "github.com/dgraph-io/badger"
-  version = "^1.6.0"
+  version = "=1.6.0"
 
 [prune]
   go-tests = true
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -121,7 +121,7 @@ required = [
 
 [[constraint]]
   name = "github.com/olivere/elastic"
-  version = "^6.2.23"
+  version = "^6.2.22"
 
 [[constraint]]
   name = "github.com/gocql/gocql"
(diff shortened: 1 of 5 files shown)


andreynkravtsov:
Ooops, will revert the elastic ver bump

Gopkg.toml
```diff
@@ -137,7 +137,7 @@ required = [
 
 [[constraint]]
   name = "github.com/dgraph-io/badger"
-  version = "=1.5.3"
+  version = "^1.6.0"
```
yurishkuro: Given breaking changes, are you sure we want ^ and not =?

andreynkravtsov: Good call, will fix

plugin/storage/badger/factory.go
```diff
@@ -89,16 +88,14 @@ func (f *Factory) InitFromViper(v *viper.Viper) {
 func (f *Factory) Initialize(metricsFactory metrics.Factory, logger *zap.Logger) error {
 	f.logger = logger
 
-	opts := badger.DefaultOptions
-	opts.TableLoadingMode = options.MemoryMap
```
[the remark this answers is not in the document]
andreynkravtsov: I should have mentioned. In 1.6.0 TableLoadingMode is defaulted to MemoryMap

burmanm:
Yeah, 1.6.0 is slower and has couple of new bugs I didn't want. So I would skip this version (that's why I've never sent a PR).

Issue #1883, opened 2019-10-29
Elasticsearch Cross Cluster Query


allen13:
Would like to add the ability to use the elasticsearch cross cluster query api to query spans across multiple AZs.

I already have some general idea on how to do it but wanted to check here before spending the time working on an MR.

Initial design:

`--es.remote-read-clusters=local,cluster_one,cluster_two`

This will result in the search query GET /jaeger-span-2019-10-28,cluster_one:jaeger-span-2019-10-28,cluster_two:jaeger-span-2019-10-28/_search

es/options.go

```go
flagSet.String(
		nsConfig.namespace+suffixRemoteReadClusters,
		nsConfig.remoteReadClusters,
		"(experimental) The comma-separated list of Elasticsearch remote clusters to read from. Add local to the list to also target the main cluster. i.e. local,cluster1,cluster2")
```

es/spanstore/reader.go

Used a wrapper function to inject the remote clusters. This is the best way I could come up with that has the minimal affect on current code.

```go
func getTimeRangeIndexFn(archive, useReadWriteAliases bool, remoteReadClusters string[]) timeRangeIndexFn {
	if archive {
		var archivePrefix string
		if useReadWriteAliases {
			archivePrefix = archiveReadIndexSuffix
		} else {
			archivePrefix = archiveIndexSuffix
		}
		return addRemoteReadClusters(func(indexName string, startTime time.Time, endTime time.Time) []string {
			return []string{archiveIndex(indexName, archivePrefix)}
		}, remoteReadClusters)
	}

	if useReadWriteAliases {
		return addRemoteReadClusters(func(indices string, startTime time.Time, endTime time.Time) []string {
			return []string{indices + "read"}
		}, remoteReadClusters)
	}

	return addRemoteReadClusters(timeRangeIndices, remoteReadClusters)
}

// Elasticsearch cross cluster query api example: GET /twitter,cluster_one:twitter,cluster_two:twitter/_search
// Add remote cluster prefix to each index
func addRemoteReadClusters(fn timeRangeIndexFn, remoteReadClusters string[]) timeRangeIndexFn {
	return func(indices string, startTime time.Time, endTime time.Time) []string {
		jaegerIndices := fn(indices, startTime, endTime)
		if len(remoteReadClusters) > 0 {
			remoteJaegerIndices := string[]{}
			for remoteCluster := range remoteReadClusters {
				for jaegerIndex := range jaegerIndices {

					if remoteCluster != "local"{
[...]

PR #1892, opened 2019-11-01
Add remote read clusters option to elasticsearch to enable cross-cluster querying

allen13:
Enables elasticsearch [cross-cluster querying](https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-cross-cluster-search.html)

Resolves [#1883](https://github.com/jaegertracing/jaeger/issues/1883)

The cross cluster query api functions by appending a pre-configured remote cluster to an index e.g. cluster_one:index_name. This change simply applies a prefix for each remote cluster and for each index. You will end up with `total_indicies = indicies * remote_clusters`. Also added a special case `local` which will not append a prefix and query the local cluster. 

This works with both standard and rollover indices since it mostly comes down to a simple string manipulation.

### DIFF
--- a/plugin/storage/es/options.go
+++ b/plugin/storage/es/options.go
@@ -32,6 +32,7 @@ const (
 	suffixSniffer             = ".sniffer"
 	suffixTokenPath           = ".token-file"
 	suffixServerURLs          = ".server-urls"
+	suffixRemoteReadClusters  = ".remote-read-clusters"
 	suffixMaxSpanAge          = ".max-span-age"
 	suffixMaxNumSpans         = ".max-num-spans"
 	suffixNumShards           = ".num-shards"
@@ -73,8 +74,9 @@ type Options struct {
 // preparing the actual config.Configuration.
 type namespaceConfig struct {
 	config.Configuration
-	servers   string
-	namespace string
+	servers            string
+	remoteReadClusters string
+	namespace          string
 }
 
 // NewOptions creates a new Options struct.
@@ -99,8 +101,9 @@ func NewOptions(primaryNamespace string, otherNamespaces ...string) *Options {
 				CreateIndexTemplates: true,
 				Version:              0,
 			},
-			servers:   "http://127.0.0.1:9200",
-			namespace: primaryNamespace,
+			servers:            "http://127.0.0.1:9200",
+			remoteReadClusters: "",
+			namespace:          primaryNamespace,
 		},
 		others: make(map[string]*namespaceConfig, len(otherNamespaces)),
 	}
@@ -141,6 +144,11 @@ func addFlags(flagSet *flag.FlagSet, nsConfig *namespaceConfig) {
 		nsConfig.namespace+suffixServerURLs,
 		nsConfig.servers,
 		"The comma-separated list of Elasticsearch servers, must be full url i.e. http://localhost:9200")
+	flagSet.String(
+		nsConfig.namespace+suffixRemoteReadClusters,
+		nsConfig.remoteReadClusters,
+		"Comma-separated list of Elasticsearch remote cluster names for cross-cluster querying."+
+			"See Elasticsearch remote clusters and cross-cluster query api.")
 	flagSet.Duration(
 		nsConfig.namespace+suffixTimeout,
 		nsConfig.Timeout,
@@ -249,6 +257,7 @@ func initFromViper(cfg *namespaceConfig, v *viper.Viper) {
 	cfg.TokenFilePath = v.GetString(cfg.namespace + suffixTokenPath)
 	cfg.Sniffer = v.GetBool(cfg.namespace + suffixSniffer)
 	cfg.servers = stripWhiteSpace(v.GetString(cfg.namespace + suffixServerURLs))
+	cfg.remoteReadClusters = stripWhiteSpace(v.GetString(cfg.namespace + suffixRemoteReadClusters))
 	cfg.MaxSpanAge = v.GetDuration(cfg.namespace + suffixMaxSpanAge)
 	cfg.MaxNumSpans = v.GetInt(cfg.namespace + suffixMaxNumSpans)
 	cfg.NumShards = v.GetInt64(cfg.namespace + suffixNumShards)
@@ -278,6 +287,13 @@ func initFromViper(cfg *namespaceConfig, v *viper.Viper) {
 // GetPrimary returns primary configuration.
 func (opt *Options) GetPrimary() *config.Configuration {
 	opt.primary.Servers = strings.Split(opt.primary.servers, ",")
+
+	if opt.primary.remoteReadClusters == "" {
+		opt.primary.RemoteReadClusters = []string{}
+	} else {
+		opt.primary.RemoteReadClusters = strings.Split(opt.primary.remoteReadClusters, ",")
+	}
+
 	return &opt.primary.Configuration
 }
 
@@ -293,6 +309,13 @@ func (opt *Options) Get(namespace string) *config.Configuration {
 		nsCfg.servers = opt.primary.servers
 	}
 	nsCfg.Servers = strings.Split(nsCfg.servers, ",")
+
+	if nsCfg.remoteReadClusters == "" {
+		nsCfg.RemoteReadClusters = []string{}
+	} else {
+		nsCfg.RemoteReadClusters = strings.Split(nsCfg.remoteReadClusters, ",")
+	}
+
 	return &nsCfg.Configuration
 }
 
--- a/plugin/storage/es/spanstore/reader.go
+++ b/plugin/storage/es/spanstore/reader.go
@@ -114,6 +114,7 @@ type SpanReaderParams struct {
 	TagDotReplacement   string
 	Archive             bool
 	UseReadWriteAliases bool
+	RemoteReadClusters  []string
 }
 
 // NewSpanReader returns a new SpanReader with a metrics.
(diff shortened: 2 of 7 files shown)


plugin/storage/es/options.go
```diff
@@ -293,6 +310,17 @@ func (opt *Options) Get(namespace string) *config.Configuration {
 		nsCfg.servers = opt.primary.servers
 	}
 	nsCfg.Servers = strings.Split(nsCfg.servers, ",")
+
+	if nsCfg.remoteReadClusters == "" {
+		nsCfg.remoteReadClusters = opt.primary.remoteReadClusters
+	}
+
+	if nsCfg.remoteReadClusters == "" {
```
pavolloffay: can be omitted and inner block added to line 314

plugin/storage/es/spanstore/reader.go
```diff
@@ -135,24 +136,51 @@ type timeRangeIndexFn func(indexName string, startTime time.Time, endTime time.T
...
+}
+
+// Elasticsearch cross cluster query api example: GET /twitter,cluster_one:twitter,cluster_two:twitter/_search
+// Add a remote cluster prefix for each cluster and for each index.
+// The special remote cluster "local" leaves the prefix off of it's subset of indexes resulting in a standard local query
+func addRemoteReadClusters(fn timeRangeIndexFn, remoteReadClusters []string) timeRangeIndexFn {
+	return func(indices string, startTime time.Time, endTime time.Time) []string {
+		jaegerIndices := fn(indices, startTime, endTime)
+		if len(remoteReadClusters) > 0 {
+			remoteJaegerIndices := []string{}
```
pavolloffay: We know the size at creation time

plugin/storage/es/spanstore/reader.go
```diff
@@ -135,24 +136,51 @@ type timeRangeIndexFn func(indexName string, startTime time.Time, endTime time.T
...
+			for _, remoteCluster := range remoteReadClusters {
+				for _, jaegerIndex := range jaegerIndices {
+
+					if remoteCluster != "local" {
```
pavolloffay: Can we avoid this? We can require not adding it via flags.

plugin/storage/es/spanstore/reader_test.go
```diff
@@ -146,35 +146,38 @@ func TestSpanReaderIndices(t *testing.T) {
...
 		{params: SpanReaderParams{Client: client, Logger: logger, MetricsFactory: metricsFactory,
 			IndexPrefix: "foo:", Archive: true},
-			index: "foo:" + indexPrefixSeparator + spanIndex + archiveIndexSuffix},
+			indices: []string{"foo:" + indexPrefixSeparator + spanIndex + archiveIndexSuffix}},
 		{params: SpanReaderParams{Client: client, Logger: logger, MetricsFactory: metricsFactory,
 			IndexPrefix: "foo:", Archive: true, UseReadWriteAliases: true},
-			index: "foo:" + indexPrefixSeparator + spanIndex + archiveReadIndexSuffix},
+			indices: []string{"foo:" + indexPrefixSeparator + spanIndex + archiveReadIndexSuffix}},
+		{params: SpanReaderParams{Client: client, Logger: logger, MetricsFactory: metricsFactory,
+			IndexPrefix: "", Archive: false, RemoteReadClusters: []string{"local", "cluster_one"}},
```
pavolloffay: could you add a test with two remote clusters?

pavolloffay:
@allen13 does this switch work for ES 5, 6 and 7?

plugin/storage/es/spanstore/reader.go
allen13: Some options:
 --es.remote-read-clusters-remote-only
--es.remote-read-clusters-no-local

What do you think?

pavolloffay: I might be missing something, but why to flags? The names on the flag will be added to indices. If local is added it will be added. if it is not added it won't be added.  I just want to avoid unnecessary checks in the reader.

allen13: Ahh, I misunderstood. Are you suggesting we always keep the local un-prefixed index by default? I really struggled to figure out how we should toggle on and off the local cluster reference. I've seen use cases where you have an elastic query node that only has remote clusters and no data local.

Current:
  Input: --es.remote-read-clusters=local,cluster_one
  Output: /twitter,cluster_one:twitter
Alternate?:
 Input: --es.remote-read-clusters=cluster_one
 Output: /twitter,cluster_one:twitter

allen13:
@pavolloffay Cross-cluster query support was added in Elasticserch 5.4. https://www.elastic.co/guide/en/elasticsearch/reference/5.5/modules-cross-cluster-search.html

plugin/storage/es/spanstore/reader.go
allen13: Do you want the array to be statically sized?
remoteJaegerIndices := [size]string{}

pavolloffay: yes

plugin/storage/es/spanstore/reader.go
pavolloffay: >  I've seen use cases where you have an elastic query node that only has remote clusters and no data local.

can you provide any references to this use-case? 

Then I guess we cannot go with the following as it will always add local data.
```
Alternate?:
Input: --es.remote-read-clusters=cluster_one
Output: /twitter,cluster_one:twitter
```

plugin/storage/es/options.go
allen13: Done

plugin/storage/es/spanstore/reader.go
allen13: After removing local just appended to the jaegerIndices to simplify things.

plugin/storage/es/spanstore/reader.go
allen13: Thanks for the input! This does make things much simpler! The no local use case can be addressed in the future if anyone ever needs it.

plugin/storage/es/spanstore/reader_test.go
allen13: Done.

plugin/storage/es/options.go
```diff
@@ -278,6 +287,13 @@ func initFromViper(cfg *namespaceConfig, v *viper.Viper) {
 // GetPrimary returns primary configuration.
 func (opt *Options) GetPrimary() *config.Configuration {
 	opt.primary.Servers = strings.Split(opt.primary.servers, ",")
+
+	if opt.primary.remoteReadClusters == "" {
+		opt.primary.RemoteReadClusters = []string{}
```
yurishkuro: this branch is not necessary

plugin/storage/es/options.go
```diff
@@ -293,6 +309,13 @@ func (opt *Options) Get(namespace string) *config.Configuration {
 		nsCfg.servers = opt.primary.servers
 	}
 	nsCfg.Servers = strings.Split(nsCfg.servers, ",")
+
+	if nsCfg.remoteReadClusters == "" {
+		nsCfg.RemoteReadClusters = []string{}
```
yurishkuro: branch not necessary

plugin/storage/es/spanstore/reader.go
```diff
@@ -135,24 +136,45 @@ type timeRangeIndexFn func(indexName string, startTime time.Time, endTime time.T
...
+		}, remoteReadClusters)
+	}
+
+	return addRemoteReadClusters(timeRangeIndices, remoteReadClusters)
+}
+
+// Elasticsearch cross cluster query api example: GET /twitter,cluster_one:twitter,cluster_two:twitter/_search
+// Add a remote cluster prefix for each cluster and for each index and add it to the list of original indices
+func addRemoteReadClusters(fn timeRangeIndexFn, remoteReadClusters []string) timeRangeIndexFn {
+	return func(indices string, startTime time.Time, endTime time.Time) []string {
```
yurishkuro: shouldn't this be short-circuited to return `fn` if `len(remoteReadClusters) == 0`?

plugin/storage/es/spanstore/reader_test.go
```diff
@@ -146,35 +146,38 @@ func TestSpanReaderIndices(t *testing.T) {
...
 			IndexPrefix: "foo:", Archive: true},
-			index: "foo:" + indexPrefixSeparator + spanIndex + archiveIndexSuffix},
+			indices: []string{"foo:" + indexPrefixSeparator + spanIndex + archiveIndexSuffix}},
 		{params: SpanReaderParams{Client: client, Logger: logger, MetricsFactory: metricsFactory,
 			IndexPrefix: "foo:", Archive: true, UseReadWriteAliases: true},
-			index: "foo:" + indexPrefixSeparator + spanIndex + archiveReadIndexSuffix},
+			indices: []string{"foo:" + indexPrefixSeparator + spanIndex + archiveReadIndexSuffix}},
+		{params: SpanReaderParams{Client: client, Logger: logger, MetricsFactory: metricsFactory,
+			IndexPrefix: "", Archive: false, RemoteReadClusters: []string{"cluster_one", "cluster_two"}},
+			indices: []string{spanIndex + dateFormat, "cluster_one:" + spanIndex + dateFormat, "cluster_two:" + spanIndex + dateFormat}},
```
yurishkuro: please make this stmt multi-line, one cluster per line

dgrizzanti:
@allen13 @pavolloffay I'd be really interested in seeing this PR merged. Do you know if anyone is actively looking into this? I'd be happy to try and get this ready otherwise if there is still willingness for the feature.  Thanks!

jpkrohling:
Closing, as I think the linked PR replaced this one.

Issue #1340, opened 2019-02-13
[HotROD] Add a base path option


nikolay:
Deploying HotROD into an existing Kubernetes set up as a test if your Jaeger setup works is a bit problematic due to the hardcoded fix base URL for Ajax calls at `/dispatch`. If there was an option to change the base path of the whole app, it would be much easier to deploy to a variety of test environments.

PR #1894, opened 2019-11-03
Support custom basepath in HotROD

jan25:
- With this PR, HotROD frontend app can be served at custom URL like `localhost:8080/hotrod`
- Resolves https://github.com/jaegertracing/jaeger/issues/1340

- Added a new `basepath` argument to `cmd/root.go`
- Adjusted index.html to use relative path for ajax requests

## TODO
- May need to add this to the docs somewhere?


examples/hotrod/services/frontend/server.go
```diff
@@ -54,20 +56,27 @@ func NewServer(options ConfigOptions, tracer opentracing.Tracer, logger log.Fact
...
-	mux.Handle("/", http.FileServer(s.assetFS))
-	mux.Handle("/dispatch", http.HandlerFunc(s.dispatch))
+	switch {
+	case s.basepath != "/":
+		mux.Handle(s.basepath, http.StripPrefix(s.basepath, http.FileServer(s.assetFS)))
+		mux.Handle(s.basepath+"/dispatch", http.HandlerFunc(s.dispatch))
+	default:
+		mux.Handle("/", http.FileServer(s.assetFS))
+		mux.Handle("/dispatch", http.HandlerFunc(s.dispatch))
+	}
```
albertteoh: switch isn't necessary, could be simplified to:
```
       p := path.Join("/", s.basepath)
       mux.Handle(p, http.StripPrefix(p, http.FileServer(s.assetFS)))
       mux.Handle(path.Join(p, "dispatch"), http.HandlerFunc(s.dispatch))
```

examples/hotrod/services/frontend/server.go
```diff
@@ -54,20 +56,27 @@ func NewServer(options ConfigOptions, tracer opentracing.Tracer, logger log.Fact
...
 		assetFS:  assetFS,
+		basepath: options.Basepath,
 	}
 }
 
 // Run starts the frontend server
 func (s *Server) Run() error {
 	mux := s.createServeMux()
-	s.logger.Bg().Info("Starting", zap.String("address", "http://"+s.hostPort))
+	s.logger.Bg().Info("Starting", zap.String("address", "http://"+s.hostPort+s.basepath))
```
albertteoh: use `path.Join`, i.e. `"http://"+path.Join(s.hostPort, s.basepath)`. Join will prevent duplicate `/`, so a `basepath` of `foo` and `/foo` are both valid. This will make the implementation more robust.

examples/hotrod/cmd/root.go
```diff
@@ -73,6 +75,9 @@ func init() {
 	RootCmd.PersistentFlags().IntVarP(&frontendPort, "frontend-service-port", "f", 8080, "Port for frontend service")
 	RootCmd.PersistentFlags().IntVarP(&routePort, "route-service-port", "r", 8083, "Port for routing service")
 
+	// Flag for serving frontend at custom basepath url
+	RootCmd.PersistentFlags().StringVarP(&basepath, "basepath", "b", "/", "Basepath for frontend service")
```
albertteoh: nit: suggest default of empty string `""`as we shouldn't expect the user to know to prefix with `/` but code should support both. See the suggestions below.

examples/hotrod/services/frontend/web_assets/index.html
```diff
@@ -73,7 +73,12 @@ <h4><em>Rides On Demand</em></h4>
   };
   console.log(headers);
   var before = Date.now();
-  $.ajax('/dispatch?customer=' + customer + '&nonse=' + Math.random(), {
+
+  // Use current URI as basepath for ajax requests
+  var pathPrefix = window.location.pathname;
+  pathPrefix = pathPrefix != "/" ? pathPrefix : '';
+
+  $.ajax(pathPrefix + '/dispatch?customer=' + customer + '&nonse=' + Math.random(), {
```
albertteoh: the demo still works for me without this change. Why is it necessary? (Perhaps I've missed something?)

jan25: without this the browser would always talk to `/dispatch` instead of `/something/dispatch`. Any changes to web_assets need to be compiled using `esc`. Perhaps you missed that?

Check this line https://github.com/jaegertracing/jaeger/blob/master/examples/hotrod/services/frontend/gen_assets.go#L1

jan25:
@albertteoh Thanks for the review! I like your idea about using `path` library, I'll make those changes.

jan25:
Hi @pavolloffay @yurishkuro! could you please review this? This is a simple change :)

examples/hotrod/cmd/root.go
```diff
@@ -73,6 +75,9 @@ func init() {
 	RootCmd.PersistentFlags().IntVarP(&frontendPort, "frontend-service-port", "f", 8080, "Port for frontend service")
 	RootCmd.PersistentFlags().IntVarP(&routePort, "route-service-port", "r", 8083, "Port for routing service")
 
+	// Flag for serving frontend at custom basepath url
+	RootCmd.PersistentFlags().StringVarP(&basepath, "basepath", "b", "", "Basepath for frontend service")
```
pavolloffay: Could you please add a default value `/` it makes the help command more self explanatory.

```
  -b, --basepath string                  Basepath for frontend service (default "/")
```

instead of 
```
  -b, --basepath string                  Basepath for frontend service
```

PR #1900, opened 2019-11-06
Add gzip flag to jaeger-query

### LINKED ISSUE #1899: Add flag allow disable gzip encoding on jaeger-query
### Problem
Some HTTP client cannot process gzip encoding automatically

I can't change the HTTP client behavior

### Proposal 

Add a flag, allow disable gzip encoding

```
--query.gzip    enable gzip compress (default true)
```

Sasasu:
Signed-off-by: lizhaolong (Sasasu) <su@sasasu.me>

- Resolves #1899

- Add a flag to jaeger-query to control gzip is enabled or not

### DIFF
--- a/cmd/query/app/flags.go
+++ b/cmd/query/app/flags.go
@@ -55,7 +55,7 @@ func AddFlags(flagSet *flag.FlagSet) {
 	flagSet.String(queryStaticFiles, "", "The directory path override for the static assets for the UI")
 	flagSet.String(queryUIConfig, "", "The path to the UI configuration file in JSON format")
 	flagSet.Bool(queryTokenPropagation, false, "Allow propagation of bearer token to be used by storage plugins")
-	flagSet.Bool(queryGzip, true, "enable gzip compress")
+	flagSet.Bool(queryGzip, true, "Enable Gzip compression")
 }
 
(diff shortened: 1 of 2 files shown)


cmd/query/app/flags.go
```diff
@@ -52,7 +55,7 @@ func AddFlags(flagSet *flag.FlagSet) {
 	flagSet.String(queryStaticFiles, "", "The directory path override for the static assets for the UI")
 	flagSet.String(queryUIConfig, "", "The path to the UI configuration file in JSON format")
 	flagSet.Bool(queryTokenPropagation, false, "Allow propagation of bearer token to be used by storage plugins")
-
+	flagSet.Bool(queryGzip, true, "enable gzip compress")
```
pavolloffay: * remove the empty line above
* Start the message with a capital letter  - `Gzip compression`

Sasasu: resolved :+1:

yurishkuro:
I am not sure we should be implementing these hacks to support a faulty client that indicates that understand compression via the `Accept-Encoding` header, but then is unable to handle it.

Sasasu:
the client is `ribbon` with `eureka` with `RestTemplate(org.springframework.web.client.RestTemplate)` reported by another Java developer.

PR #1903, opened 2019-11-06
WIP: support for NATS as a collector

The-Alchemist:
Looking for feedback in the areas of go conventions, Jaeger contribution guidelines, etc.

> Requirement - what kind of business use case are you trying to solve?

NATS is a "simple, secure and performant communications system for digital systems, services and devices," and it's part of the CNCF and written in Go, just like jaeger itself.  For many use cases, it can replace Kafka.

Jaeger supports Kafka OOTB, but not any other messaging platforms.  NATS support in Jaeger would allow Jaeger deployments with much lower resource requirements than Kafka while still supporting the use cases that Kafka enables. 

> Problem - what in Jaeger blocks you from solving the requirement?

AFAICT, there is no support for plugins for collectors in Jaeger.  Since Kafka support is built into Jaeger, it seems to make sense to bake NATS support alongside Kafka.

> Proposal - what do you suggest to solve the problem or improve the existing situation?

This PR reuses adds NATS support as a collector while reusing some Kafka functions (like for marshalling/unmarshalling).

> Any open questions to address

* Should NATS Streaming support be built into the NATS collector, or a separate type?
* It's currently using Kafka's marshaling support by exporting those functions.
   Should they be extracted into a common location instead?

P.S. Commits not signed yet, and unit tests aren't working yet.


yurishkuro:
> Looking for feedback in the areas of go conventions, **Jaeger contribution guidelines**, etc.

https://github.com/jaegertracing/jaeger/blob/master/CONTRIBUTING_GUIDELINES.md#making-a-change

**Before making any significant changes, please open an issue**. Each issue should describe the following:

* Requirement - what kind of business use case are you trying to solve?
* Problem - what in Jaeger blocks you from solving the requirement?
* Proposal - what do you suggest to solve the problem or improve the existing situation?
* Any open questions to address

jpkrohling:
Closing due to inactivity.

PR #1904, opened 2019-11-07
Cassandra Tag Index Filters

### LINKED ISSUE #1901: Allow configuration of Cassandra tag indexing

Our Jaeger installation is currently constrained by the number of tag indexes written into Cassandra/Scylla.  We are ingesting ~7.5k spans/second and ~75k tag indexes/second.

We would like a central place to control/reduce the number of tags we index.  This would allow us to increase our sampling rate and capture more spans/second.

The Jaeger Cassandra StorageFactory internally has support for tag filters, but they are not configurable: 
https://github.com/jaegertracing/jaeger/blob/master/plugin/storage/cassandra/spanstore/writer.go#L124

Expose two simple filters that allow for tag indexing control:

`cassandra.tag-index-blacklist`
`cassandra.tag-index-whitelist`

These two filters would be a comma delimited list of tags to white or blacklist by exact match.  They would be mutually exclusive.  

None.

joe-elliott:
- Resolves #1901 

- Added two configuration options for the cassandra storage backend: `tag-index-blacklist`, `tag-index-whitelist`
- Added appropriate TagFilters, Tests


joe-elliott:
At scale the canary ingester is using roughly the same CPU and indexing significantly less tags

image

joe-elliott:
Manually confirmed in our environment that blacklisted tags are not being indexed and other tags are.  Removing draft.

pkg/cassandra/config/config.go
```diff
@@ -45,6 +45,8 @@ type Configuration struct {
 	DisableAutoDiscovery bool          `yaml:"disable_auto_discovery"`
 	EnableDependenciesV2 bool          `yaml:"enable_dependencies_v2"`
 	TLS                  TLS
+	TagIndexWhitelist    []string `yaml:"tag-index-whitelist"`
+	TagIndexBlacklist    []string `yaml:"tag-index-blacklist"`
```
yurishkuro: this seems like the wrong place for these options (and `enable_dependencies_v2`, for that matter), because this struct is meant as a generic Cassandra connection configuration, not the SpanWriter configuration.

joe-elliott: There does not appear to be a generic SpanWriter configuration and these settings are Cassandra specific.  I could make a `SpanWriterConfiguration` struct in the same package?

joe-elliott: Ignore.  Adding it to the cassandra Options struct like `SpanStoreWriteCacheTTL`

plugin/storage/cassandra/options.go
```diff
@@ -116,6 +120,15 @@ func (opt *Options) AddFlags(flagSet *flag.FlagSet) {
 	flagSet.Duration(opt.primary.namespace+suffixSpanStoreWriteCacheTTL,
 		opt.SpanStoreWriteCacheTTL,
 		"The duration to wait before rewriting an existing service or operation name")
+	flagSet.String(
+		opt.primary.namespace+suffixTagIndexBlacklist,
+		opt.tagIndexBlacklist,
+		"The comma-separated list of tags to blacklist")
```
yurishkuro: The comma-separated list of span tags to blacklist from being indexed. All other tags will be indexed. Mutually exclusive with the whitelist option.

plugin/storage/cassandra/options.go
```diff
@@ -116,6 +120,15 @@ func (opt *Options) AddFlags(flagSet *flag.FlagSet) {
...
+	flagSet.String(
+		opt.primary.namespace+suffixTagIndexWhitelist,
+		opt.tagIndexWhitelist,
+		"The comma-separated list of tags to whitelist")
```
yurishkuro: The comma-separated list of span tags to whitelist for being indexed. All other tags will not be indexed. Mutually exclusive with the blacklist option.

plugin/storage/cassandra/spanstore/dbmodel/tag_filter_blacklist.go
```diff
@@ -0,0 +1,61 @@
...
+func (tf BlacklistTagFilter) FilterTags(span *model.Span, tags model.KeyValues) model.KeyValues {
+	return tf.filter(tags)
+}
+
+// FilterLogFields implements TagFilter
+func (tf BlacklistTagFilter) FilterLogFields(span *model.Span, logFields model.KeyValues) model.KeyValues {
+	return tf.filter(logFields)
+}
+
+func (tf BlacklistTagFilter) filter(processTags model.KeyValues) model.KeyValues {
```
yurishkuro: nit: s/processTags/tags/

plugin/storage/cassandra/spanstore/dbmodel/tag_filter_whitelist.go
```diff
@@ -0,0 +1,61 @@
...
+// limitations under the License.
+
+package dbmodel
+
+import (
+	"github.com/jaegertracing/jaeger/model"
+)
+
+// WhitelistTagFilter filters out all tags in its tags slice
+type WhitelistTagFilter struct {
```
yurishkuro: couldn't whitelist/blacklist filter be implemented with a single struct that tags a boolean parameter to change the behavior?

plugin/storage/cassandra/spanstore/dbmodel/tag_filter_blacklist.go
```diff
@@ -0,0 +1,61 @@
+// Copyright (c) 2019 The Jaeger Authors.
+// Copyright (c) 2017 Uber Technologies, Inc.
```
yurishkuro: my directory search doesn't show you working for Uber :-) . This line is not needed.

plugin/storage/cassandra/options.go
```diff
@@ -276,6 +292,24 @@ func (opt *Options) Get(namespace string) *config.Configuration {
 	return &nsCfg.Configuration
 }
 
+// GetTagIndexBlacklist returns the list of blacklisted tags
+func (opt *Options) GetTagIndexBlacklist() []string {
```
yurishkuro: We can drop the `Get` prefix

plugin/storage/cassandra/factory_test.go
```diff
@@ -100,3 +100,68 @@ func TestCassandraFactory(t *testing.T) {
 	_, err = f.CreateArchiveSpanWriter()
 	assert.NoError(t, err)
 }
+
+func TestExclusiveWhiteListBlackList(t *testing.T) {
```
yurishkuro: nit: don't capitalize "List"

plugin/storage/cassandra/options.go
```diff
@@ -276,6 +292,24 @@ func (opt *Options) Get(namespace string) *config.Configuration {
...
+	if len(opt.tagIndexBlacklist) > 0 {
+		return strings.Split(opt.tagIndexBlacklist, ",")
+	}
+
+	return []string{}
```
yurishkuro: `return nil` would also work

plugin/storage/cassandra/options_test.go
```diff
@@ -60,11 +60,14 @@ func TestOptionsWithFlags(t *testing.T) {
...
 		"--cas.proto-version=3",
 		"--cas.socket-keep-alive=42s",
+		"--cas.tag-index-blacklist=blerg, blarg,blorg ",
+		"--cas.tag-index-whitelist=flerg, flarg,florg ",
 		// enable aux with a couple overrides
 		"--cas-aux.enabled=true",
 		"--cas-aux.keyspace=jaeger-archive",
 		"--cas-aux.servers=3.3.3.3, 4.4.4.4",
 		"--cas-aux.enable-dependencies-v2=true",
+		"--cas-aux.tag-index-whitelist=foo,bar",
```
yurishkuro: did you mean to test for this too?

plugin/storage/cassandra/options.go
joe-elliott: Updated both as recommended.

plugin/storage/cassandra/spanstore/dbmodel/tag_filter_blacklist.go
joe-elliott: Updated as recommended.

plugin/storage/cassandra/spanstore/dbmodel/tag_filter_whitelist.go
joe-elliott: Yup.  Combined into `ExactMatchTagFilter`.  In my mind it would be more efficient to have two different structs.  In retrospect that seems silly.

plugin/storage/cassandra/spanstore/dbmodel/tag_filter_blacklist.go
joe-elliott: Ha.  Nope. :)

plugin/storage/cassandra/spanstore/dbmodel/tag_filter_exact_match.go
```diff
@@ -0,0 +1,64 @@
...
+// ExactMatchTagFilter filters out all tags in its tags slice
+type ExactMatchTagFilter struct {
+	tags        map[string]struct{}
+	dropMatches bool
+}
+
+// NewExactMatchTagFilter creates a ExactMatchTagFilter with the provided tags.  Passing
+// dropMatches true will exhibit blacklist behavior.  Passing dropMatches false
+// will exhibit whitelist behavior.
+func NewExactMatchTagFilter(tags []string, dropMatches bool) ExactMatchTagFilter {
```
yurishkuro: nit: I would prefer to keep the factory method names as NewBlacklistTagFilter/NewWhitelistTagFilter, even though they return the same type. In this current form it's hard to read at the call sites. E.g. in 6 months would you remember what this means: `NewExactMatchTagFilter(tags, true)`?

plugin/storage/cassandra/spanstore/dbmodel/tag_filter_exact_match_test.go
```diff
@@ -0,0 +1,121 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/jaegertracing/jaeger/model"
+)
+
+func TestBlacklistFilter(t *testing.T) {
+	// expected
+	tt := [][][]string{
```
yurishkuro: nit: make this more readable by using struct:
```
tests := []struct{
    input []string
    filter []string
    expected []string
}{
  {
    input: ...
    filter: ...
  },
}
```

yurishkuro:
Please make sure the description says `Resolves #1901` rather than just the issue number, to enable auto-close.

plugin/storage/cassandra/spanstore/dbmodel/tag_filter_exact_match_test.go
joe-elliott: Agreed and I appreciate the patience.  I'm relatively new to go.

PR #1910, opened 2019-11-08
Use proxy env vars if they're configured

zoidbergwill:
We use a custom `net/http` Transport for talking to Elastic Search that doesn't currently support `HTTP{,S}_PROXY` environment variables, but Elastic Search might be only accessible via a Proxy for some corporate enterprise reasons or because you're using something like Envoy as a proxy for egress for service discovery.


yurishkuro:
Could you please provide a business case in the PR description?  There's no explanation why this change is needed.

Please make sure the CI is green.

zoidbergwill:
I'll rebase and properly sign off my commits.

PR #1915, opened 2019-11-11
[cassandra] Drop filters for tag indexing by type

### LINKED ISSUE #1905: Allow disabling of log indexing

In order to increase spans throughput to Cassandra we need more control over what is indexed.  Currently all logs are being indexed.  Once https://github.com/jaegertracing/jaeger/issues/1904 is merged we will have more control, but it would be nice to just disable log indexing all together.

There is a partial implementation already: https://github.com/jaegertracing/jaeger/blob/master/plugin/storage/cassandra/spanstore/dbmodel/log_fields_filter.go

It just needs to be exposed as a configuration option.

Add a boolean configuration option to the Cassandra backend `index-logs` that defaults to true.

None

joe-elliott:
- Resolves #1905 

- Moved all index filtering options under `index`
- Added 3 options to drop all tags of a certain type
- Converted log fields filter into a configurable "drop all" filter
- Fixed bug where ChainedTagFilter was not calling FilterLogFields correctly
- Added appropriate tests


joe-elliott:
Manually tested in our environment.  Dropping draft status.

plugin/storage/cassandra/spanstore/dbmodel/tag_filter.go
```diff
@@ -26,6 +26,15 @@ type TagFilter interface {
 	FilterLogFields(span *model.Span, logFields model.KeyValues) model.KeyValues
 }
 
+// NewCompositeFilter returns a TagFilter that implements all of the passed filters in order.
+func NewCompositeFilter(filters ...TagFilter) TagFilter {
```
yurishkuro: why do we need this constructor, if we already have NewChainedTagFilter? It just adds noise.

plugin/storage/cassandra/spanstore/dbmodel/tag_filter_drop_all.go
```diff
@@ -0,0 +1,60 @@
...
+
+// TagFilterDropAll filters all fields of a given type.
+type TagFilterDropAll struct {
+	dropTags        bool
+	dropProcessTags bool
+	dropLogs        bool
+}
+
+// NewTagFilterDropAll return a filter that filters all of the specified type
+func NewTagFilterDropAll(dropTags bool, dropProcessTags bool, dropLogs bool) TagFilterDropAll {
```
yurishkuro: I would return a pointer here and change receivers accordingly. In practice this will be used by callers that will assign the result to something of type TagFilter, which is an interface, meaning that this struct would still escape to the heap in order to be converted to the interface.

plugin/storage/cassandra/spanstore/dbmodel/tag_filter_drop_all_test.go
```diff
@@ -0,0 +1,92 @@
+// Copyright (c) 2019 The Jaeger Authors.
+// Copyright (c) 2017 Uber Technologies, Inc.
```
yurishkuro: rm

plugin/storage/cassandra/spanstore/dbmodel/tag_filter_drop_all.go
```diff
@@ -0,0 +1,60 @@
+// Copyright (c) 2019 The Jaeger Authors.
+// Copyright (c) 2017 Uber Technologies, Inc.
```
yurishkuro: rm

plugin/storage/cassandra/spanstore/dbmodel/tag_filter.go
joe-elliott: I thought that perhaps the functionality to optionally create a ChainedTagFilter depending on the length of the TagFilter slice would be reusable.

Removing.

Issue #1921, opened 2019-11-13
Cassandra & memory storage update: query service to support spanKind when retrieve operations for a given service.


guo0693:
## Requirement - what kind of business use case are you trying to solve?
Related to #1920 

- Cassandra & memory storage update: query service to support spanKind when retrieve operations for a given service.

- Update query service so that the restful & grpc endpoints satisfy requirement in #1920.

yurishkuro:
this ticket is redundant, it's sufficient to have #1920

guo0693:
Done

PR #1924, opened 2019-11-13
Query service to support spanKind when retrieve operations for a given service

guo0693:

-  Solves #1921 

- update query service endpoints to allow optional spanKind param when querying operations
- update cassandra & memory storage plugin to write/read using the extra arg spanKind
- create migration script for the new operation_names table


guo0693:
Going to test with local setup and post screenshots tomorrow. Open for early review

cmd/query/app/querysvc/query_service.go
```diff
@@ -79,8 +80,8 @@ func (qs QueryService) GetServices(ctx context.Context) ([]string, error) {
 }
 
 // GetOperations is the queryService implementation of spanstore.Reader.GetOperations
-func (qs QueryService) GetOperations(ctx context.Context, service string) ([]string, error) {
-	return qs.spanReader.GetOperations(ctx, service)
+func (qs QueryService) GetOperations(ctx context.Context, service string, spanKind string) ([]*storage_v1.OperationMeta, error) {
```
yurishkuro: Please make it to accept parameters in a struct instead of positionally (context should remain separate).

model/json/model.go
```diff
@@ -109,3 +109,8 @@ type DependencyLink struct {
 	Child     string `json:"child"`
 	CallCount uint64 `json:"callCount"`
 }
+
+type OperationMeta struct {
```
yurishkuro: Meta is unnecessary, can call it just Operation.

model/proto/api_v2/query.proto
```diff
@@ -109,10 +109,16 @@ message GetServicesResponse {
 
 message GetOperationsRequest {
   string service = 1;
+  string span_kind = 2;
+}
+
+message OperationMeta {
+    string operation = 1;
```
yurishkuro: s/operation/name

model/json/model.go
```diff
@@ -109,3 +109,8 @@ type DependencyLink struct {
...
+	OperationName string `json:"operationName"`
```
yurishkuro: Just “name” is fine & unambiguous.

plugin/storage/grpc/proto/storage.proto
```diff
@@ -76,10 +76,16 @@ message GetServicesResponse {
 
 message GetOperationsRequest {
     string service = 1;
+    string span_kind = 2;
+}
+
+message OperationMeta {
```
yurishkuro: Operation

plugin/storage/cassandra/schema/v003.cql.tmpl
```diff
@@ -0,0 +1,204 @@
...
+    AND dclocal_read_repair_chance = 0.0
+    AND default_time_to_live = ${trace_ttl}
+    AND speculative_retry = 'NONE'
+    AND gc_grace_seconds = 10800; -- 3 hours of downtime acceptable on nodes
+
+CREATE TABLE IF NOT EXISTS ${keyspace}.operation_names (
+    service_name        text,
+    operation_name      text,
+    span_kind           text,
+    PRIMARY KEY ((service_name), operation_name)
```
vprithvi: The `PRIMARY KEY` needs to be updated to include `span_kind`, otherwise rows are identified only by the `service_name` and `operation_name`, meaning there can only be one row per `service_name`/`operation_name` combination. 

With the current schema, the following sequence of inserts will only contain data from the last write. 
```
INSERT INTO operation_names (service_name, operation_name, span_kind) VALUES ('svc_a', 'y', 'client');
INSERT INTO operation_names (service_name, operation_name, span_kind) VALUES ('svc_a', 'y', 'server');
```

We also have to rethink data migration, because primary keys may not be altered. 
It may make sense to create a new table, and have the code read/write from the new table if it's present, and perform a data migration in parallel.

guo0693: created a new table instead of using the existing one, migration script updated. Tested in my local, the migration script successfully ported data from operation_names to operations table.

guo0693:
Tested locally:

Get operations using the old endpoint:

```~/go/src$ curl 'http://localhost:6686/api/services/gopro/operations' | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   154  100   154    0     0  15945      0 --:--:-- --:--:-- --:--:-- 17111
{
  "data": [
    "UserService::getExtendedUser",
    "UserService::getPaymentProfileViews",
    "UserService::updateUserInfo"
  ],
  "total": 3,
  "limit": 0,
  "offset": 0,
  "errors": null
}
```
Get operations using the new endpoint:
```
~/go/src$ curl 'http://localhost:6686/api/operations?service=gopro' | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   223  100   223    0     0  13234      0 --:--:-- --:--:-- --:--:-- 13937
{
  "data": [
    {
      "name": "UserService::getExtendedUser",
      "spanKind": ""
    },
    {
      "name": "UserService::getPaymentProfileViews",
      "spanKind": ""
    },
    {
      "name": "UserService::updateUserInfo",
      "spanKind": ""
    }
  ],
  "total": 3,
  "limit": 0,
  "offset": 0,
  "errors": null
}
```
Index some more spans so that the spanKind is enriched, get server operations only:
```
~/go/src$ curl 'http://localhost:6686/api/operations?service=gopro&spanKind=server' | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   241  100   241    0     0  31673      0 --:--:-- --:--:-- --:--:-- 34428
{
  "data": [
    {
      "name": "UserService::getExtendedUser",
      "spanKind": "server"
    },
    {
      "name": "UserService::getPaymentProfileViews",
      "spanKind": "server"
    },
    {
      "name": "UserService::updateUserInfo",
      "spanKind": "server"
    }
  ],
  "total": 3,
  "limit": 0,
  "offset": 0,
  "errors": null
}
```

guo0693:
Seems like travis build failed bcz of panic: No messages processed in the last check interval

yurishkuro:
Just to make sure, is the code going to be compatible with the previous db schema? We need to make sure it is.

plugin/storage/cassandra/spanstore/operation_names.go
```diff
@@ -25,23 +26,27 @@ import (
...
+	"github.com/jaegertracing/jaeger/storage/spanstore"
 )
 
 const (
-	insertOperationName = `INSERT INTO operation_names(service_name, operation_name) VALUES (?, ?)`
-	queryOperationNames = `SELECT operation_name FROM operation_names WHERE service_name = ?`
+	operationTableName        = `operations`
+	insertOperationName       = `INSERT INTO ` + operationTableName + ` (service_name, span_kind, operation_name) VALUES (?, ?, ?)`
+	queryOperationNames       = `SELECT span_kind, operation_name FROM ` + operationTableName + ` WHERE service_name = ? `
+	queryOperationNamesByKind = `SELECT span_kind, operation_name FROM ` + operationTableName + ` WHERE service_name = ? AND span_kind = ? `
```
yurishkuro: We need to be backwards compatible with previous DB schema. On startup we can [interrogate table metadata][1] to check if the table has the extra column or not.

[1]: https://stackoverflow.com/questions/38696316/how-to-list-all-cassandra-tables/50163285

guo0693:
> Just to make sure, is the code going to be compatible with the previous db schema? We need to make sure it is.

The code will read from new table. Therefore lib user need to run the migration script to create new table before running the code, so it's not really backward compatible.

Please see @vprithvi 's comment above for the reason why we need separate table.

If we really need to make it backward compatible, we may need to have two APIs on spanReader:

- GetOperations read from old table, used by the old endpoint: get /api/v1/services/{service}/operation

- GetOperationsV2 read from new table, used by new endpoint: get /api/v1/operations?service={}&spanKind={}

plugin/storage/cassandra/spanstore/operation_names.go
guo0693: We have backward compatibility on the api (endpoint) level, which returns []string as operations for the old API, the old UI will work when we deploy new jaeger-query (verified in staging).
Given only UI is the user of the api, i think it's good enough.

Do we have any specific reason that we need to be backward compatible on DB schema level?

yurishkuro:
If we need to add the new field to the clustering key, then there's no way to make the table backwards compatible, since Cassandra won't allow ALTER TABLE with key change. In that case going with `{table}_v2` is appropriate, but we still should allow the code to read from the existing table if v2 is not found. That allows people to separate code upgrade from db schema upgrade.

Is it difficult to write a migration script? Did we do it for dependencies_v2?

If we don't do a migration script and people just run schema upgrade and create an empty v2 table, then it will be empty, which is not great user experience (even though it will quickly get filled up, it may not reflect the previous traces already stored, and we don't know how long people keep the traces around).

guo0693:
Closing this PR as the relevant small PRs are merged

Issue #1737, opened 2019-08-16
Suggestion: Continuous Fuzzing


yevgenypats:
Hi, I'm Yevgeny Pats Founder of [Fuzzit](https://fuzzit.dev) - Continuous fuzzing as a service platform.

We have a free plan for OSS and I would be happy to contribute a PR if that's interesting.
The PR will include the following
- [go-fuzz](https://github.com/dvyukov/go-fuzz) fuzzers (This is generic step not-connected to fuzzit)
- Continuous Fuzzing of master branch which will generate new corpus and look for new crashes
- Regression on every PR that will run the fuzzers through all the generated corpus and fixed crashes from previous step. This will prevent new or old bugs from crippling into master.

You can see our basic example [fuzzitdev/example-go](https://github.com/fuzzitdev/example-go) and you can see an example of "in the wild" integration [google/syzkaller](https://github.com/google/syzkaller).

Let me know if this is something worth working on.

Cheers,
Yevgeny

jpkrohling:
I think we could give it a try, but I'm trying to figure out one thing first: do contributors have to sign up for fuzzit in order to view the reports? I'd rather not force that upon our contributors.

yevgenypats:
@jpkrohling So good question, it depends on what is your security policy.

The regression tests are run locally in the CI (Travis) and if there are any crashes the contributors will see it in the CI log and there is no need to sign-up for fuzzit.

The fuzzing which is essentially an infinite job that runs on fuzzit and alerts only when crashes are found. The most common use-case is to alert the security team or specific people as you usually don't won't to publish crashes before they are fixed. We also have a feature that we will roll-out soon to file issue automatically on Github but then it will be publicly. 

Currently public regression + private crash reporting works for most projects. let me know if this also works for your use-case/project?

jpkrohling:
Thanks for your clarification. 

> We also have a feature that we will roll-out soon to file issue automatically on Github but then it will be publicly.

Are you aware of GitHub Security Advisories? Perhaps they expose an API, allowing your tool to create one such advisory:

https://github.com/jaegertracing/jaeger/security/advisories 

This will be visible only to project maintainers.

> Currently public regression + private crash reporting works for most projects. let me know if this also works for your use-case/project?

I think it makes sense, you have my +1.

yevgenypats:
Yes, I'm aware of the github security advisories. currently they don't provide an API afaik but once they do we will for sure integrate. 

[...]

PR #1925, opened 2019-11-14
Remove fuzzing

pavolloffay:
Related to #1737 


DavidKorczynski:
@pavolloffay and @yurishkuro would you be willing to revisit fuzzing? In particular, we could add Jaeger to OSS-Fuzz and then there should be no problems with your CI? @AdamKorcz did an initial integration here https://github.com/google/oss-fuzz/pull/6892

PR #1929, opened 2019-11-17
Adding warnings field to Cassandra traces table

### LINKED ISSUE #705: Span.Warnings not stored in Cassandra
The `model Span.Warnings` field isn't stored in Cassandra. 

https://github.com/jaegertracing/jaeger/blob/fcbd21078ab7ed03682964c187452778242d3423/model/span.go#L59

Cassandra schema for reference:
https://github.com/jaegertracing/jaeger/blob/ec6378a7a5094ff3019c506163108da29f3f2f99/plugin/storage/cassandra/schema/v001.cql.tmpl#L54-L68

The warnings field contains details on any conversion failures from thrift to the internal model object. I feel that this is important information that should be persisted along with the rest of the Span.

lazyf0z:
Signed-off-by: Aditya Kumar Praharaj <adityakumarpraharaj@gmail.com>

- Resolves #705 - namely adds the field `Warnings` to the `traces` table in jaeger's cassandra keyspace.

- This diff adds the `Warnings` field in `traces` table and renames the subsequent table to `traces_v2`. It accompanies migrations `v003.cql.tmpl` which is the CQL schema files and two subsequent migration scripts from moving to `traces_v2` from `traces`:
  - `migration/v002tov003part1.sh`: Alters existing `traces` table to add the column and creates a new `traces_v2` table as well. Copies existing data from `traces` to `traces_v2`.
  - `migration/v002tov003part2.sh`: Truncates and deletes `traces` table.
- It makes changes to `SpanReader` and `SpanWriter` to actually read from `traces_v2`. This is backwards compatible and falls back to using `traces` if the `traces_v2` table actually doesn't exist (which means the migrations have not been carried out).

## Followup

- Create the same changes for Elasticsearch Backend.
- Propagating and showing these in the UI.

### DIFF
--- a/plugin/storage/cassandra/spanstore/table_version_test.go
+++ b/plugin/storage/cassandra/spanstore/table_version_test.go
@@ -1,0 +1,53 @@
+// Copyright (c) 2019 The Jaeger Authors.
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package spanstore
+
+import (
+	"context"
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/jaegertracing/jaeger/pkg/cassandra/mocks"
+	"github.com/jaegertracing/jaeger/plugin/storage/cassandra/spanstore/dbmodel"
+)
+
+func TestTraceVersion_V1(t *testing.T) {
+	versionQuery := func() *mocks.Query {
+		query := &mocks.Query{}
+		query.On("Exec").Return(errors.New("version read error"))
+		return query
+	}
+
+	mockSession := &mocks.Session{}
+	mockSession.On("Query", stringMatcher(queryTraceTableVersion), matchEverything()).Return(versionQuery())
+	version := getTraceTableVersion(context.Background(), mockSession)
+	assert.EqualValues(t, dbmodel.V1, version)
+}
+
+func TestTraceVersion_V2(t *testing.T) {
+	versionQuery := func() *mocks.Query {
+		query := &mocks.Query{}
+		query.On("Exec").Return(nil)
+		return query
+	}
+
+	mockSession := &mocks.Session{}
+	mockSession.On("Query", stringMatcher(queryTraceTableVersion), matchEverything()).Return(versionQuery())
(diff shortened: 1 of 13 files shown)


yurishkuro:
A few thoughts:

- pr #1924 is already introducing changes to the DB schema, we should combine them to minimize dB migrations
- it is preferable if the code works both with the old and new schema and just automatically  adjusts to the presence or absence of the additional columns. An example of that can be found in the handling of the dependencies table.

lazyf0z:
Hey Yuri! 

Thanks for the comments. Would appreciate if you can clarify some points:

- I saw the issue linked above (#1924), and it seems like the changes are not merged. After those changes are done (or before) - I will make my changes on top of those. Along those same lines:
   - What is the strategy for DB migrations? Should all the subsequent DB migrations since V2 be merged in a single file? 
   - What about handling these in code - does the code need to actually handle all the possible states between migrations? That seems necessary though unsustainable.
- Thanks. I have tried to add that by having both V1 and V2 versions of query statements and automatically detecting which columns to read / update depending on the state of the keyspace at runtime. (Please refer to changes in `cassandra/spanstore/reader.go`, `cassandra/spanstore/table_version.go` and `cassandra/spanstore/writer.go`).

Sorry for so many questions. Just trying to get a general sense of things cause this is my first diff here.

plugin/storage/cassandra/spanstore/table_version_test.go
```diff
@@ -0,0 +1,53 @@
+// Copyright (c) 2019 The Jaeger Authors.
+// Copyright (c) 2017 Uber Technologies, Inc.
```
yurishkuro: don't include this line

yurishkuro:
The PR (#1924) needs some work, but it should happen shortly. It currently does not implement backwards compatibility (e.g. by [interrogating table metadata[[1]).

It is ok to have an additional migration script, but they should all be wired together for "upgrade for next release", AND it's better to combine changes into a single V3 full schema file.

Handling in code - we usually keep compatibility for a 1-2 releases then remove the old code. Initially the code only needs to be compatible with the previous version of the schema.

The `dependencies` table example is not quite right for this change, because that one required full schema change (we were switching from SASI indices to direct lookups). Your change is much less intrusive, it's only adding one optional, non-indexed column, so creating a whole new table for that is overkill, we should simply alter the existing table.

Speaking of schema changes, we're actually planning to implement an alternative format for storing spans in Cassandra, as Protobuf blobs instead of fully typed schema. This should make Cassandra storage much more performant, and would automatically include the extra column (that would warrant `traces_v2` table).  Would you be interested in working on that instead?

[1]: https://stackoverflow.com/questions/38696316/how-to-list-all-cassandra-tables/50163285

lazyf0z:
Yeah that sounds interesting. I would like to take that up as well. Let me try to  refactor this diff for the changes suggested in the meantime.

Sasasu:
> storing spans in Cassandra, as Protobuf blobs instead of fully typed schema

Will Jaeger use a compressed format? Like use zstd to compress after protobuf serialization.

yurishkuro:
Compression should definitely be an option. @vprithvi experimented with trained `zstd` in the past, it easily achieved 8x compression ratio.

jpkrohling:
Closing due to inactivity.

PR #1931, opened 2019-11-18
#1921 1. Add new input/output type for get operations & util function to read span kind tag

guo0693:

- Relates to #1921

Per https://github.com/jaegertracing/jaeger/pull/1924#pullrequestreview-318041879, I am going to open a few small PRs to replace https://github.com/jaegertracing/jaeger/pull/1924

- Add new input/output type for get operatoins
- Add util function to read span kind tag

### DIFF
--- a/model/span.go
+++ b/model/span.go
@@ -27,7 +27,7 @@ const (
 	SampledFlag = Flags(1)
 	// DebugFlag is the bit set in Flags in order to define a span as a debug span
 	DebugFlag = Flags(2)
-	//FirehoseFlag is the bit in Flags in order to define a span as a firehose span
+	// FirehoseFlag is the bit in Flags in order to define a span as a firehose span
 	FirehoseFlag = Flags(8)
 
 	samplerType        = "sampler.type"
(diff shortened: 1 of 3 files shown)


model/span.go
```diff
@@ -30,6 +30,8 @@ const (
 	//FirehoseFlag is the bit in Flags in order to define a span as a firehose span
 	FirehoseFlag = Flags(8)
 
+	SpanKindUnknown = ""
```
vprithvi: Does this need to be exported? 

Also - could you document why the empty string was chosen instead of `unknown`?

model/span.go
```diff
@@ -53,6 +55,14 @@ func (s *Span) HasSpanKind(kind ext.SpanKindEnum) bool {
 	return false
 }
 
+// GetSpanKind returns value of `span.kind` tag, empty string if not present
+func (s *Span) GetSpanKind() string {
```
vprithvi: This seems tangential to the commit message...

model/span.go
guo0693: We decide to use empty string so that the endpoint can be backward compatible: GET /api/v1/operations?serivce={}
If spanKind is not present, the operation index will use empty string to query Cassandra to return operations names of all span kind.

I don't feel it's right to document the reason here in the model as it's related to the API.

Do you think "// Used to indicate the span.kind tag is missing" is enough?

model/span.go
guo0693: commit msg updated

model/span.go
yurishkuro: as far as this API goes, you don't need to have the constant, you can simply return a boolean flag
```
GetSpanKind() (spanKind string, found bool)
```

PR #1933, opened 2019-11-19
Fix typo in license URL

swinslow:
Fixes a typo in the license comments URL in one file.

Signed-off-by: Steve Winslow <swinslow@gmail.com>


swinslow:
Found it during a license scan using a tool called [Fossology](https://fossology.org). It is an automated tool but requires a fair bit of manual effort for reviewing / clearing scan results, so not particularly easy to drop into an automated build process.

I'm currently running and clearing scan results for Jaeger as well as other CNCF projects, and I'll be sharing the results with Jaeger maintainers. Just came across this one during that process and figured it was an easy fix  :)

PR #1934, opened 2019-11-19
Bypassing building UI altogether on non-master PR builds

### LINKED ISSUE #1908: Optimize all-in-one and Docker CI steps
One of the longest parts of all-in-one and Docker steps is building the UI, which is now repeated (after #1909).

We can either combine these two steps, or fix the scripts to skip building the UI when on a PR build.

lazyf0z:
Signed-off-by: Aditya Kumar Praharaj <adityakumarpraharaj@gmail.com>

- Resolves #1908  by bypassing building UI altogether for PR branches.

- This diff adds support for skipping the make build-ui stage which builds the UI when the current branch is a PR-branch (non-master).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -67,7 +67,7 @@ install:
 
 script:
   - export BRANCH=$(if [ "$TRAVIS_PULL_REQUEST" == "false" ]; then echo $TRAVIS_BRANCH; else echo $TRAVIS_PULL_REQUEST_BRANCH; fi)
-  - export SKIP_UI=$(if [ "$BRANCH" == "master" ]; then echo "true"; fi)
+  - export SKIP_UI=$(if [ "$BRANCH" != "master" ]; then echo "true"; fi)
   - if [ "$TESTS" == true ]; then make test-ci ; else echo 'skipping tests'; fi
   - if [ "$ALL_IN_ONE" == true ]; then bash ./scripts/travis/build-all-in-one-image.sh ; else echo 'skipping all_in_one'; fi
   - if [ "$CROSSDOCK" == true ]; then bash ./scripts/travis/build-crossdock.sh ; else echo 'skipping crossdock'; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -71,7 +71,7 @@ script:
   - if [ "$ALL_IN_ONE" == true ]; then bash ./scripts/travis/build-all-in-one-image.sh ; else echo 'skipping all_in_one'; fi
   - if [ "$CROSSDOCK" == true ]; then bash ./scripts/travis/build-crossdock.sh ; else echo 'skipping crossdock'; fi
   - if [ "$DOCKER" == true ]; then bash ./scripts/travis/build-docker-images.sh ; else echo 'skipping build-docker-images'; fi
-  - if [ "$DEPLOY" == true ]; then make build-all-platforms ; else echo 'skipping build-all-platforms'; fi
+  - if [ "$DEPLOY" == true ]; then bash ./scripts/travis/build-all-platform-binaries.sh ; else echo 'skipping build-all-platforms'; fi
   - if [ "$ES_INTEGRATION_TEST" == true ]; then bash ./scripts/travis/es-integration-test.sh ; else echo 'skipping elastic search integration test'; fi
   - if [ "$KAFKA_INTEGRATION_TEST" == true ]; then bash ./scripts/travis/kafka-integration-test.sh ; else echo 'skipping kafka integration test'; fi
   - if [ "$CASSANDRA_INTEGRATION_TEST" == true ]; then bash ./scripts/travis/cassandra-integration-test.sh ; else echo 'skipping cassandra integration test'; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -67,6 +67,7 @@ install:
 
 script:
   - export BRANCH=$(if [ "$TRAVIS_PULL_REQUEST" == "false" ]; then echo $TRAVIS_BRANCH; else echo $TRAVIS_PULL_REQUEST_BRANCH; fi)
+  - export SKIP_UI=$(if [ "$BRANCH" == "master" ]; then echo "true"; fi)
   - if [ "$TESTS" == true ]; then make test-ci ; else echo 'skipping tests'; fi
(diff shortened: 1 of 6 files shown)


lazyf0z:
Will use a different approach. Defining different stages just for this doesn't make sense.

Makefile
```diff
@@ -290,7 +308,7 @@ include crossdock/rules.mk
 # Crossdock tests do not require fully functioning UI, so we skip it to speed up the build.
 .PHONY: build-crossdock-ui-placeholder
 build-crossdock-ui-placeholder:
-	mkdir -p cmd/query/app/ui/actual
+	mkdir -p cmd/query/app/u	i/actual
```
jpkrohling: oops?

scripts/travis/build-all-platform-binaries.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+# Builds all platform binaries.
+set -e
+
+BRANCH=${BRANCH:?'missing BRANCH env var'}
```
jpkrohling: You can probably fallback to `git branch --show-current`.

scripts/travis/build-all-platform-binaries.sh
```diff
@@ -0,0 +1,20 @@
...
+
+set -x
+
+if [[ $BRANCH == "master" ]]; then
```
jpkrohling: It's probably better to have an env var like `SKIP_UI`, which is set to `true` if the branch isn't master. This way, if I run this script locally, I can override the `SKIP_UI` env var even if I'm on a non-master branch.

Makefile
lazyf0z: Thanks, fixed.

scripts/travis/build-all-platform-binaries.sh
lazyf0z: Done.

scripts/travis/build-all-platform-binaries.sh
lazyf0z: Yeah that's better. Made all such main scripts have a `SKIP_UI` option instead.

Makefile
```diff
@@ -211,10 +211,18 @@ build-all-in-one-linux: build-ui
 build-all-in-one: elasticsearch-mappings
 	CGO_ENABLED=0 installsuffix=cgo go build -tags ui -o ./cmd/all-in-one/all-in-one-$(GOOS) $(BUILD_INFO) ./cmd/all-in-one/main.go
 
+.PHONY: build-all-in-one-without-ui
+build-all-in-one-without-ui: elasticsearch-mappings
+	CGO_ENABLED=0 installsuffix=cgo go build -o ./cmd/all-in-one/all-in-one-$(GOOS) $(BUILD_INFO) ./cmd/all-in-one/main.go
```
jpkrohling: Just realized that this target is exactly the same as the `build-all-in-one`. They could perhaps share the same command, just changing the value for the "tags" flag. Like:

    CGO_ENABLED=0 installsuffix=cgo go build -o ./cmd/query/query-linux -tags "${TAGS}" ./cmd/query/main.go

One would set `TAGS` to `ui`, one would keep it empty. This is not blocking the PR, but would be a nice thing to have.

yurishkuro: not just here, but a lot of new `*-without-ui` targets added below, which are identical to existing ones aside from `-tags`. I agree, it's better to control that aspect with a variable rather than replicating targets.

scripts/travis/build-all-in-one-image.sh
```diff
@@ -2,15 +2,24 @@
...
 
-BRANCH=${BRANCH:?'missing BRANCH env var'}
-
-source ~/.nvm/nvm.sh
-nvm use 10
-make build-ui
+if [[ -z "$BRANCH" ]]; then
+  BRANCH=$(git rev-parse --abbrev-ref HEAD)
+  echo "BRANCH env var not defined, using current branch $BRANCH instead ..."
+fi
```
yurishkuro: what is prompting this change?

scripts/travis/build-all-platform-binaries.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
yurishkuro: If we go with `-tags` controlled via env var, this file is probably not necessary.

Overall, it's better to have fewer coupling between .travis.yaml and these shell scripts - if we can test conditions in travis, let's keep it there. We could also use build stages that I believe allow branch/tag/PR checks right in the yaml syntax

scripts/travis/build-all-in-one-image.sh
jpkrohling: Instead of falling back to `missing BRANCH env var`, it's way better to fall back to the actual local branch name.

.travis.yml
```diff
@@ -67,11 +67,12 @@ install:
 
 script:
   - export BRANCH=$(if [ "$TRAVIS_PULL_REQUEST" == "false" ]; then echo $TRAVIS_BRANCH; else echo $TRAVIS_PULL_REQUEST_BRANCH; fi)
+  - export SKIP_UI=$(if [ "$BRANCH" != "master" ]; then echo "true"; fi)
```
pavolloffay: The UI should be also built for `release-x.x` branches.

pavolloffay:
Perhaps we could build UI only in UI repository and publish it as a docker or archive. Then here just move the compiled UI to docker image.

Building the UI is not only slow but also sometimes broken. The UI sometimes updates dependencies or node version which breaks the build here.

yurishkuro:
> Perhaps we could build UI only in UI repository and publish it as a docker or archive.

Not a bad idea. The only concern is that sometimes we may want to bump the UI in the backend to an unreleased version, but it's very rare and in the worst case we could release it with something like `-rc#` tag.

pavolloffay:
We could publish the image with `latest`, release tag and also hash with each commit to master - we do this in this repository https://hub.docker.com/repository/docker/jaegertracing/jaeger-collector-snapshot - although it seems broken I will book a ticket

yurishkuro:
I think it might be easier to publish to github rather than Docker, because we need the static assets bundle available when we compile the binary.

jpkrohling:
What's the status of this PR?

Issue #1920, opened 2019-11-12
Support optional query param "spanKind" when get operations for given service


guo0693:
## Requirement - what kind of business use case are you trying to solve?

We want to filter the operations of a service by span kind, those filtered operations can be used for features e.g. get dependencies graph for service:operation pair from server side (spanKind == "server") 

## Problem - what in Jaeger blocks you from solving the requirement?

Currently we only save operations indexed by service name and we can only query operations by service name. We will need to add extra info: spanKind when saving operations for a service.

## Proposal - what do you suggest to solve the problem or improve the existing situation?

**API changes:**
_Restful:_

- GET /api/services/{serviceName}/operations : 
We will keep this endpoint unchanged for backward compatibility, it will still return all operations as a list of string for a given service.

- GET /api/operations?service={service}&spanKind={spanKind}
We will update this endpoint to return list of operation metadata for a given service whose spanKind matched the query parameter. If spanKind is empty, then return all operations of a service.
Sample response: 
`{
  "data": [
    {
      "operationName": "A",
      "spanKind": "client"
    },
    {
      "operationName": "B",
      "spanKind": "server"
    }
  ],
  "total": 2,
  "limit": 0,
  "offset": 0,
  "errors": null
}`

_GRPC:_ 
rpc GetOperations(GetOperationsRequest) returns (GetOperationsResponse)
Add span_kind to the rpc request and response
`message GetOperationsRequest {
  string service = 1;
  string span_kind = 2;
}

message OperationMeta {
    string operation = 1;
    string span_kind = 2;
}

message GetOperationsResponse {
  repeated OperationMeta operations = 1;
}`

_Interfaces:_
spanstore:
`
type Reader interface {
	...
	// Get operations by service and spanKind, empty spanKind means get operations for all kinds of span
	GetOperations(ctx context.Context, service string, spanKind string) ([]*storage_v1.OperationMeta, error)
	...
}
`
Span reader will need to take extra parameter to query operations by service name and span kind. All the storage plugins needs to be updated so that:

- write operation index with service name and span kind
- query operation index by service name and span kind, return all kinds of operations when span kind is empty.

burmanm:
Couple of questions:

* What happens to old data? It can't be queried this way if the information doesn't exist.
* What defines the kind? The writing client? I don't see any links to the client changes, so is this something coming from for example OpenTelemetry?
* Why not just a tag?

[...]

PR #1937, opened 2019-11-19
Create new cassandra table "operation_names_v2" with "spanKind" column for operation name index

guo0693:

- Relates to #1921, #1920

- add migration script
- read from the latest table if available, otherwise fail back to previous table

### DIFF
--- a/plugin/storage/cassandra/factory_test.go
+++ b/plugin/storage/cassandra/factory_test.go
@@ -91,7 +91,7 @@ func TestCassandraFactory(t *testing.T) {
 	_, err = f.CreateArchiveSpanWriter()
 	assert.EqualError(t, err, "archive storage not configured")
 
-	f.archiveConfig = &mockSessionBuilder{}
+	f.archiveConfig = newMockSessionBuilder(session, nil)
 	assert.NoError(t, f.Initialize(metrics.NullFactory, zap.NewNop()))
 
 	_, err = f.CreateArchiveSpanReader()
--- a/plugin/storage/cassandra/spanstore/operation_names.go
+++ b/plugin/storage/cassandra/spanstore/operation_names.go
@@ -16,6 +16,7 @@
 package spanstore
 
 import (
+	"fmt"
 	"time"
 
 	"github.com/pkg/errors"
@@ -25,23 +26,51 @@ import (
 	"github.com/jaegertracing/jaeger/pkg/cache"
 	"github.com/jaegertracing/jaeger/pkg/cassandra"
 	casMetrics "github.com/jaegertracing/jaeger/pkg/cassandra/metrics"
+	"github.com/jaegertracing/jaeger/storage/spanstore"
 )
 
 const (
-	insertOperationName = `INSERT INTO operation_names(service_name, operation_name) VALUES (?, ?)`
-	queryOperationNames = `SELECT operation_name FROM operation_names WHERE service_name = ?`
+	// LatestVersion latest version of operation_names table schema, increase the version if your table schema changes require code change
+	LatestVersion = 1
+	// TableQueryStmt the query statement used to check if a table exists or not
+	TableQueryStmt = "SELECT * from %s limit 1"
 )
+
+type schemaMeta struct {
+	TableName       string
+	InsertStmt      string
+	QueryByKindStmt string
+	QueryStmt       string
+}
+
+var schemas = []schemaMeta{
+	{
+		TableName:       "operation_names",
+		InsertStmt:      "INSERT INTO %s(service_name, operation_name) VALUES (?, ?)",
+		QueryByKindStmt: "SELECT operation_name FROM %s WHERE service_name = ?",
+		QueryStmt:       "SELECT operation_name FROM %s WHERE service_name = ?",
+	},
+	{
+		TableName:       "operation_names_v2",
+		InsertStmt:      "INSERT INTO %s(service_name, span_kind, operation_name) VALUES (?, ?, ?)",
+		QueryByKindStmt: "SELECT span_kind, operation_name FROM %s WHERE service_name = ? AND span_kind = ?",
+		QueryStmt:       "SELECT span_kind, operation_name FROM %s WHERE service_name = ?",
+	},
+}
 
 // OperationNamesStorage stores known operation names by service.
 type OperationNamesStorage struct {
 	// CQL statements are public so that Cassandra2 storage can override them
-	InsertStmt     string
-	QueryStmt      string
-	session        cassandra.Session
-	writeCacheTTL  time.Duration
-	metrics        *casMetrics.Table
-	operationNames cache.Cache
-	logger         *zap.Logger
+	SchemaVersion   int
+	TableName       string
+	InsertStmt      string
+	QueryStmt       string
+	QueryByKindStmt string
+	session         cassandra.Session
+	writeCacheTTL   time.Duration
+	metrics         *casMetrics.Table
+	operationNames  cache.Cache
+	logger          *zap.Logger
 }
 
 // NewOperationNamesStorage returns a new OperationNamesStorage
@@ -51,13 +80,23 @@ func NewOperationNamesStorage(
 	metricsFactory metrics.Factory,
 	logger *zap.Logger,
 ) *OperationNamesStorage {
+
+	schemaVersion := LatestVersion
+
+	if !tableExist(session, schemas[schemaVersion].TableName) {
+		schemaVersion = schemaVersion - 1
+	}
+
 	return &OperationNamesStorage{
-		session:       session,
-		InsertStmt:    insertOperationName,
-		QueryStmt:     queryOperationNames,
-		metrics:       casMetrics.NewTable(metricsFactory, "operation_names"),
-		writeCacheTTL: writeCacheTTL,
-		logger:        logger,
+		session:         session,
+		TableName:       schemas[schemaVersion].TableName,
+		SchemaVersion:   schemaVersion,
+		InsertStmt:      fmt.Sprintf(schemas[schemaVersion].InsertStmt, schemas[schemaVersion].TableName),
+		QueryByKindStmt: fmt.Sprintf(schemas[schemaVersion].QueryByKindStmt, schemas[schemaVersion].TableName),
+		QueryStmt:       fmt.Sprintf(schemas[schemaVersion].QueryStmt, schemas[schemaVersion].TableName),
+		metrics:         casMetrics.NewTable(metricsFactory, schemas[schemaVersion].TableName),
+		writeCacheTTL:   writeCacheTTL,
+		logger:          logger,
 		operationNames: cache.NewLRUWithOptions(
 			100000,
 			&cache.Options{
@@ -67,12 +106,27 @@ func NewOperationNamesStorage(
 	}
 }
 
+func tableExist(session cassandra.Session, tableName string) bool {
+	query := session.Query(fmt.Sprintf(TableQueryStmt, tableName))
+	err := query.Exec()
+	return err == nil
+}
+
 // Write saves Operation and Service name tuples
 func (s *OperationNamesStorage) Write(serviceName string, operationName string) error {
 	var err error
+	//TODO: take spanKind from args
+	spanKind := ""
 	query := s.session.Query(s.InsertStmt)
-	if inCache := checkWriteCache(serviceName+"|"+operationName, s.operationNames, s.writeCacheTTL); !inCache {
-		q := query.Bind(serviceName, operationName)
+	if inCache := checkWriteCache(serviceName+"|"+spanKind+"|"+operationName, s.operationNames, s.writeCacheTTL); !inCache {
+		var q cassandra.Query
+		switch s.SchemaVersion {
+		case 1:
+			q = query.Bind(serviceName, spanKind, operationName)
+		case 0:
+			q = query.Bind(serviceName, operationName)
+		}
+
 		err2 := s.metrics.Exec(q, s.logger)
 		if err2 != nil {
 			err = err2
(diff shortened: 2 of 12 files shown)


plugin/storage/cassandra/factory_test.go
```diff
@@ -91,7 +91,7 @@ func TestCassandraFactory(t *testing.T) {
 	_, err = f.CreateArchiveSpanWriter()
 	assert.EqualError(t, err, "archive storage not configured")
 
-	f.archiveConfig = &mockSessionBuilder{}
+	f.archiveConfig = newMockSessionBuilder(session, nil)
```
guo0693: Have to pass the mock session so that operation names table check can pass

guo0693:
Tested in local env, the code works with both old and new table

plugin/storage/cassandra/spanstore/operation_names.go
```diff
@@ -25,23 +26,51 @@ import (
 	"github.com/jaegertracing/jaeger/pkg/cache"
 	"github.com/jaegertracing/jaeger/pkg/cassandra"
 	casMetrics "github.com/jaegertracing/jaeger/pkg/cassandra/metrics"
+	"github.com/jaegertracing/jaeger/storage/spanstore"
 )
 
 const (
-	insertOperationName = `INSERT INTO operation_names(service_name, operation_name) VALUES (?, ?)`
-	queryOperationNames = `SELECT operation_name FROM operation_names WHERE service_name = ?`
+	// LatestVersion latest version of operation_names table schema, increase the version if your table schema changes require code change
```
yurishkuro: `LatestVersion latest version` stutters. You can simply drop the words `latest version` and it will read better.

Also, break the comment into two lines, it's too long for no reason, can't even see on GitHub without scrolling.

plugin/storage/cassandra/spanstore/operation_names.go
```diff
@@ -25,23 +26,51 @@ import (
...
+	LatestVersion = 1
```
yurishkuro: The semantics of this are not clear to me, w.r.t. the schemaMeta below. E.g. the new table is called operation_names_v2, but its index in the table is `1`.

What if you do LatestVersion as a string, e.g. `v2`, and change the below table to a `map[strong]schemaMeta`

plugin/storage/cassandra/spanstore/operation_names.go
```diff
@@ -25,23 +26,51 @@ import (
...
-	session        cassandra.Session
-	writeCacheTTL  time.Duration
-	metrics        *casMetrics.Table
-	operationNames cache.Cache
-	logger         *zap.Logger
+	SchemaVersion   int
+	TableName       string
+	InsertStmt      string
+	QueryStmt       string
+	QueryByKindStmt string
```
yurishkuro: * why do these need to be public?
* you already have the struct above containing the same fields, so you can just have a single field `table schemaMeta`

plugin/storage/cassandra/spanstore/operation_names.go
```diff
@@ -25,23 +26,51 @@ import (
...
 const (
-	insertOperationName = `INSERT INTO operation_names(service_name, operation_name) VALUES (?, ?)`
-	queryOperationNames = `SELECT operation_name FROM operation_names WHERE service_name = ?`
+	// LatestVersion latest version of operation_names table schema, increase the version if your table schema changes require code change
+	LatestVersion = 1
+	// TableQueryStmt the query statement used to check if a table exists or not
+	TableQueryStmt = "SELECT * from %s limit 1"
 )
 
+type schemaMeta struct {
```
yurishkuro: I would rename this to tableMeta, because it is currently scoped to a single table.

plugin/storage/cassandra/spanstore/operation_names.go
```diff
@@ -25,23 +26,51 @@ import (
...
+	"github.com/jaegertracing/jaeger/storage/spanstore"
 )
 
 const (
-	insertOperationName = `INSERT INTO operation_names(service_name, operation_name) VALUES (?, ?)`
-	queryOperationNames = `SELECT operation_name FROM operation_names WHERE service_name = ?`
+	// LatestVersion latest version of operation_names table schema, increase the version if your table schema changes require code change
+	LatestVersion = 1
+	// TableQueryStmt the query statement used to check if a table exists or not
+	TableQueryStmt = "SELECT * from %s limit 1"
```
yurishkuro: does it need to be public?

yurishkuro: and the name could be `tableVersionCheckStmt`, since you're not actually querying this table

plugin/storage/cassandra/spanstore/operation_names.go
```diff
@@ -67,12 +106,27 @@ func NewOperationNamesStorage(
...
 	query := s.session.Query(s.InsertStmt)
-	if inCache := checkWriteCache(serviceName+"|"+operationName, s.operationNames, s.writeCacheTTL); !inCache {
-		q := query.Bind(serviceName, operationName)
+	if inCache := checkWriteCache(serviceName+"|"+spanKind+"|"+operationName, s.operationNames, s.writeCacheTTL); !inCache {
+		var q cassandra.Query
+		switch s.SchemaVersion {
+		case 1:
+			q = query.Bind(serviceName, spanKind, operationName)
+		case 0:
+			q = query.Bind(serviceName, operationName)
```
yurishkuro: I would do this switch in the constructor and just capture a function as a struct field:
```
bindQuery func(query cas.Query, service, kind, opName string)
```
Also, it would be better to encapsulate all these differences in the tableMeta object

plugin/storage/cassandra/spanstore/operation_names.go
```diff
@@ -51,13 +80,23 @@ func NewOperationNamesStorage(
...
-		QueryStmt:     queryOperationNames,
-		metrics:       casMetrics.NewTable(metricsFactory, "operation_names"),
-		writeCacheTTL: writeCacheTTL,
-		logger:        logger,
+		session:         session,
+		TableName:       schemas[schemaVersion].TableName,
+		SchemaVersion:   schemaVersion,
+		InsertStmt:      fmt.Sprintf(schemas[schemaVersion].InsertStmt, schemas[schemaVersion].TableName),
+		QueryByKindStmt: fmt.Sprintf(schemas[schemaVersion].QueryByKindStmt, schemas[schemaVersion].TableName),
+		QueryStmt:       fmt.Sprintf(schemas[schemaVersion].QueryStmt, schemas[schemaVersion].TableName),
```
yurishkuro: in line with the other comments, this could be a single field, created like this:
```
table: schemas[schemaVersion].materialize()
```

All sprintf can be done in materialize(), as well as creating the respective insert/query functions.

plugin/storage/cassandra/spanstore/operation_names_test.go
```diff
@@ -30,6 +30,12 @@ import (
 	"github.com/jaegertracing/jaeger/pkg/testutils"
 )
 
+type test struct {
+	ttl           time.Duration
+	schemaVersion int
+	expErr        error
+}
```
yurishkuro: this is defined waaaay to far from the two places where it's used. We normally use inline type definition
```
tests := []struct {
	ttl           time.Duration
	schemaVersion int
	expErr        error
}{
   { ... test 1 params ... },
   { ... test 2 params ... },
}
```

plugin/storage/cassandra/spanstore/operation_names_test.go
```diff
@@ -55,38 +69,49 @@ func withOperationNamesStorage(writeCacheTTL time.Duration, fn func(s *operation
...
 func TestOperationNamesStorageWrite(t *testing.T) {
-	for _, ttl := range []time.Duration{0, time.Minute} {
-		writeCacheTTL := ttl // capture loop var
-		t.Run(fmt.Sprintf("writeCacheTTL=%v", writeCacheTTL), func(t *testing.T) {
-			withOperationNamesStorage(writeCacheTTL, func(s *operationNameStorageTest) {
+	for _, test := range []test{
+		{0, 0, nil},
+		{time.Minute, 0, nil},
+		{0, 1, nil},
+		{time.Minute, 1, nil},
```
yurishkuro: Please use initialization with named fields, it is very hard to understand what `0, 0, nil` is supposed to mean.

Also, add a string field `name` to the struct which describes that the sub-test is testing

plugin/storage/cassandra/spanstore/operation_names_test.go
```diff
@@ -55,38 +69,49 @@ func withOperationNamesStorage(writeCacheTTL time.Duration, fn func(s *operation
...
+	} {
+		writeCacheTTL := test.ttl // capture loop var
```
yurishkuro: You don't need this, there are no go-routines here, the variable `test` is stable already during each iteration.

plugin/storage/cassandra/spanstore/operation_names_test.go
```diff
@@ -55,60 +63,73 @@ func withOperationNamesStorage(writeCacheTTL time.Duration, fn func(s *operation
...
+		name          string
+		ttl           time.Duration
+		schemaVersion string
+	}{
+		{name: "test old schema with 0 ttl", ttl: 0, schemaVersion: previousVersion},
+		{name: "test old schema with 1min ttl", ttl: time.Minute, schemaVersion: previousVersion},
+		{name: "test new schema with 0 ttl", ttl: 0, schemaVersion: latestVersion},
+		{name: "test new schema with 1min ttl", ttl: time.Minute, schemaVersion: latestVersion},
+	} {
+		fmt.Printf(test.name)
```
yurishkuro: use `t.Run(test.name, func() {})`

plugin/storage/cassandra/spanstore/operation_names_test.go
```diff
@@ -118,28 +139,40 @@ func TestOperationNamesStorageGetServices(t *testing.T) {
...
+		name          string
+		schemaVersion string
+		expErr        error
+	}{
+		{name: "test old schema without error", schemaVersion: previousVersion, expErr: nil},
+		{name: "test old schema with scan error", schemaVersion: previousVersion, expErr: scanError},
+		{name: "test new schema without error", schemaVersion: latestVersion, expErr: nil},
+		{name: "test new schema with scan error", schemaVersion: latestVersion, expErr: scanError},
+	} {
+		fmt.Printf(test.name)
```
yurishkuro: t.Run()

plugin/storage/cassandra/spanstore/operation_names_test.go
```diff
@@ -118,28 +139,40 @@ func TestOperationNamesStorageGetServices(t *testing.T) {
...
-				assert.Equal(t, []string{""}, services)
+				if test.schemaVersion == previousVersion {
+					// expect one empty operation result because mock iter.Scan(&placeholder) does not write to `placeholder`
+					assert.Equal(t, []string{""}, services)
+				} else {
+					assert.Equal(t, []string{}, services)
+				}
 			} else {
-				assert.EqualError(t, err, "Error reading operation_names from storage: "+expErr.Error())
+				assert.EqualError(t, err, fmt.Sprintf("Error reading %s from storage: %s", schemas[test.schemaVersion].tableName, test.expErr.Error()))
```
yurishkuro: arguments to assert functions are already treated as printf args, you don't need additional fmt.Sprintf

plugin/storage/cassandra/spanstore/operation_names_test.go
```diff
@@ -118,28 +139,40 @@ func TestOperationNamesStorageGetServices(t *testing.T) {
...
-			if expErr == nil {
+			if test.expErr == nil {
 				assert.NoError(t, err)
-				// expect empty string because mock iter.Scan(&placeholder) does not write to `placeholder`
-				assert.Equal(t, []string{""}, services)
+				if test.schemaVersion == previousVersion {
+					// expect one empty operation result because mock iter.Scan(&placeholder) does not write to `placeholder`
+					assert.Equal(t, []string{""}, services)
+				} else {
+					assert.Equal(t, []string{}, services)
```
yurishkuro: it is not clear to me why in the first case there's an empty string in the array, and in the second case an empty array

plugin/storage/cassandra/spanstore/operation_names.go
```diff
@@ -25,18 +26,65 @@ import (
...
+}
+
+func (t *tableMeta) materialize() {
+	t.insertStmt = fmt.Sprintf(t.insertStmt, t.tableName)
+	t.queryByKindStmt = fmt.Sprintf(t.queryByKindStmt, t.tableName)
+	t.queryStmt = fmt.Sprintf(t.queryStmt, t.tableName)
+}
+
+var schemas = map[string]*tableMeta{
+	"v1": {
```
yurishkuro: you have constants defined for these values

plugin/storage/cassandra/spanstore/operation_names.go
```diff
@@ -83,16 +141,71 @@ func (s *OperationNamesStorage) Write(serviceName string, operationName string)
...
+	err := query.Exec()
+	return err == nil
+}
+
+func getOperationsV1(s *OperationNamesStorage, query *spanstore.OperationQueryParameters) ([]*spanstore.Operation, error) {
+	iter := s.session.Query(s.table.queryStmt, query.ServiceName).Iter()
 
 	var operation string
-	var operations []string
+	var operationNames []string
```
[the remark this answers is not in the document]
guo0693: good catch, I missed this when I rewrite the PRs which is now different from the original PR. Quite a lot of context switching to work on different small PRs.

plugin/storage/cassandra/spanstore/operation_names_test.go
guo0693: I tried, but it failed. 

`func EqualError(t TestingT, theError error, errString string, msgAndArgs ...interface{}) bool`
fmt.Sprintf is used to generate the errString, without fmt.Sprintf, it will treat "schemas[test.schemaVersion].tableName, test.expErr.Error()" as msg args for the output when the assert fail

plugin/storage/cassandra/spanstore/operation_names_test.go
guo0693: I actually fixed that in the second PR, it's due to the initialization of the mock. Going to move it here.

plugin/storage/cassandra/spanstore/operation_names_test.go
```diff
@@ -108,38 +138,49 @@ func TestOperationNamesStorageWrite(t *testing.T) {
...
+		name          string
+		schemaVersion schemaVersion
+		expErr        error
+	}{
+		{name: "test old schema without error", schemaVersion: previousVersion, expErr: nil},
+		{name: "test old schema with scan error", schemaVersion: previousVersion, expErr: scanError},
+		{name: "test new schema without error", schemaVersion: latestVersion, expErr: nil},
+		{name: "test new schema with scan error", schemaVersion: latestVersion, expErr: scanError},
+	} {
+		t.Run(fmt.Sprintf("%s", test.name), func(t *testing.T) {
```
guo0693: I have optimized to use t.Run(test.name, func) in the PR 1921-3. Try to avoid update it in 2nd branch and merge into 3rd branch again.

yurishkuro:
why is the build failing?

guo0693:
> why is the build failing?

Because the t.run is using fmt.Sprintf("%s", test.name), instead of test.name directly. 

I fixed it in the [PR 1921-3](https://github.com/guo0693/jaeger/pull/1/files#diff-2b0bb4e731deb0e139995b85f7b202fdR84) when I merge this branch into 1921-3 last night before travis throw this error.

I don't want to change this branch and solve merge conflict again in 1921-3 since the build will pass when I merge 1921-3 to this branch.

My plan is: 

1. merge 1921-3 to this branch
2. get the build pass
3. merge this branch to master
4. open PR for endpoint & query service changes against master
5. open diff in internal repo to pull the latest master

yurishkuro:
We do not merge non-compiling PRs to master. Why is it so difficult to apply the t.Run to two lines? If it matches another branch, there won't be merge conflicts.

guo0693:
I think in general, when we work on the big feature changes, the plan could be:
1. create a feature branch base on master
2. open staggered PRs against the feature branch (e.g. PR1, PR2 base on PR1, PR3 base on PR2 etc) and have those small staggered PRs reviewed and merged into feature branch. We should not do code changes directly in feature branch.
3. After all staggered PRs are merged, we are at a state the feature branch is complete. Merge feature branch into master, solve potential conflict with master if necessary at this stage.

What do you think?

guo0693:
> We do not merge non-compiling PRs to master. Why is it so difficult to apply the t.Run to two lines? If it matches another branch, there won't be merge conflicts.

We are not going to merge non-compiling PRs to master.
1921-3 is going to be merged to 1921-2 (this feature branch) first. We will then merge 1921-2 to master.

Both 1921-2 and 1921-3 touched the same files, therefore everytime I touch operation_names.go and test, I have to solve merge conflicts in 1921-3. I have repeated that several times yesterday, feel like it's not the right way. Although I am OK to update this PR with two lines changes and do it again in 1921-3.

In general, I want to propose a process for working on the big feature changes:

create a feature branch base on master
open staggered PRs against the feature branch (e.g. PR1, PR2 base on PR1, PR3 base on PR2 etc) and have those small staggered PRs reviewed and merged into feature branch. We should not do code changes directly in feature branch.
After all staggered PRs are merged, we are at a state the feature branch is complete. Merge feature branch into master, solve potential conflict with master if necessary at this stage.
What do you think?

guo0693:
If you are OK with above, can i get a stamp on https://github.com/guo0693/jaeger/pull/1 first? 
I will merge it into this branch. Then you can stamp on this PR. I will then merge it into master.

yurishkuro:
I don't understand the workflow you're proposing. Merging something into this PR will require another top to bottom review, how does it make reviewing easier?

guo0693:
> I don't understand the workflow you're proposing. Merging something into this PR will require another top to bottom review, how does it make reviewing easier?

"We should not do code changes directly in feature branch." Each individual PR is reviewed separately. As long as there is no un-necessary code changes in feature branch (solving merged conflict in the last step is necessary code change), we don't need another top to bottom review.

Here, we need to make sure: 
1. you are ok with 1921-3 PR, and you are ok with 1921-2 PR
2. when we merge 1921-3 PR into 1921-2, there is no code changes.
3. immediately merge 1921-2 PR into master. Do not add any more code changes in 1921-2.

guo0693:
Btw, at the same time, I pushed changes to fix the travis build in this branch, also rebased 1921-3 onto this branch

PR #1942, opened 2019-11-22
Update storage api to support query operations by spanKind

guo0693:
- Relates to jaegertracing#1921

- storage API changes to support new query parameter and return type (across all storage engines)

This PR contains the exact same changes as https://github.com/guo0693/jaeger/pull/1, The only difference is this PR is opened against master branch.


plugin/storage/cassandra/spanstore/reader_test.go
```diff
@@ -368,7 +391,11 @@ func TestSpanReaderFindTraces(t *testing.T) {
 					assert.EqualError(t, err, testCase.expectedError)
 				}
 				for _, expectedLog := range testCase.expectedLogs {
-					assert.True(t, strings.Contains(r.logBuffer.String(), expectedLog), "Log must contain %s, but was %s", expectedLog, r.logBuffer.String())
+					assert.True(t,
```
yurishkuro: there is assert.Contains()

yurishkuro: which also makes the last 3 args unnecessary since it will show the diffs automatically

plugin/storage/cassandra/spanstore/writer_test.go
```diff
@@ -192,13 +194,20 @@ func TestSpanWriter(t *testing.T) {
...
 				w.session.On("Query", stringMatcher(insertTag), matchOnce()).Return(tagsQuery)
 
-				w.session.On("Query", stringMatcher(serviceNameIndex), matchEverything()).Return(serviceNameQuery)
-				w.session.On("Query", stringMatcher(serviceOperationIndex), matchEverything()).Return(serviceOperationNameQuery)
+				w.session.On("Query",
+					stringMatcher(serviceNameIndex),
+					matchEverything()).Return(serviceNameQuery)
+				w.session.On("Query",
+					stringMatcher(serviceOperationIndex),
+					matchEverything()).Return(serviceOperationNameQuery)
```
yurishkuro: (!) while I raised the issue with long lines, I only meant it for new/changed code. It is a very bad practice to reformat the code in the same PR as the main code changes, creates a lot of unnecessary noise.

plugin/storage/es/spanstore/reader_test.go
```diff
@@ -488,12 +488,11 @@ func testGet(typ string, t *testing.T) {
 		searchResult   *elastic.SearchResult
 		searchError    error
 		expectedError  string
-		expectedOutput []string
+		expectedOutput interface{}
 	}{
 		{
-			caption:        typ + " full behavior",
-			searchResult:   &elastic.SearchResult{Aggregations: elastic.Aggregations(goodAggregations)},
-			expectedOutput: []string{"123"},
```
yurishkuro: so this expected output is deleted? seems like it would change the test

plugin/storage/es/spanstore/reader_test.go
```diff
@@ -506,6 +505,16 @@ func testGet(typ string, t *testing.T) {
...
+	if typ == operationsAggregation {
+		testCases[0].expectedOutput = []*spanstore.Operation{
+			{
+				Name:     "123",
+				SpanKind: "",
+			},
+		}
+	} else {
+		testCases[0].expectedOutput = []string{"123"}
+	}
```
yurishkuro: these values should be provided at the initialization of the test struct.

plugin/storage/es/spanstore/service_operation_test.go
```diff
@@ -121,7 +122,8 @@ func TestSpanReader_GetOperationsEmptyIndex(t *testing.T) {
 			Return(&elastic.MultiSearchResult{
 				Responses: []*elastic.SearchResult{},
 			}, nil)
-		services, err := r.reader.GetOperations(context.Background(), "foo")
+		services, err := r.reader.GetOperations(context.Background(),
+			&spanstore.OperationQueryParameters{ServiceName: "foo", SpanKind: ""})
```
yurishkuro: SpanKind: "" is redundant

plugin/storage/grpc/proto/storage.proto
```diff
@@ -76,10 +76,17 @@ message GetServicesResponse {
...
+
+message Operation {
+    string name = 1;
+    string span_kind = 2;
 }
 
 message GetOperationsResponse {
-    repeated string operations = 1;
+    repeated string operations = 1; // deprecated
+    repeated Operation operations_v2 = 2;
```
yurishkuro: it is ok to rename fields, it has no effect on the binary format. I prefer
```
message GetOperationsResponse {
    repeated string operationNames = 1; // deprecated
    repeated Operation operations = 2;
}
```

plugin/storage/grpc/shared/grpc_client.go
```diff
@@ -104,15 +104,24 @@ func (c *grpcClient) GetServices(ctx context.Context) ([]string, error) {
 }
 
 // GetOperations returns the operations of a given service
-func (c *grpcClient) GetOperations(ctx context.Context, service string) ([]string, error) {
+func (c *grpcClient) GetOperations(ctx context.Context, query *spanstore.OperationQueryParameters) (
+	[]*spanstore.Operation, error) {
```
yurishkuro: signature formatting

plugin/storage/grpc/shared/grpc_server.go
```diff
@@ -78,13 +78,24 @@ func (s *grpcServer) GetServices(ctx context.Context, r *storage_v1.GetServicesR
 }
 
 // GetOperations returns the operations of a given service
-func (s *grpcServer) GetOperations(ctx context.Context, r *storage_v1.GetOperationsRequest) (*storage_v1.GetOperationsResponse, error) {
-	operations, err := s.Impl.SpanReader().GetOperations(ctx, r.Service)
+func (s *grpcServer) GetOperations(ctx context.Context, r *storage_v1.GetOperationsRequest) (
+	*storage_v1.GetOperationsResponse, error) {
```
yurishkuro: formatting

plugin/storage/grpc/shared/grpc_server.go
```diff
@@ -78,13 +78,24 @@ func (s *grpcServer) GetServices(ctx context.Context, r *storage_v1.GetServicesR
...
+	operations, err := s.Impl.SpanReader().GetOperations(ctx, &spanstore.OperationQueryParameters{
+		ServiceName: r.Service,
+		SpanKind:    r.SpanKind,
+	})
 	if err != nil {
 		return nil, err
 	}
+	grpcOperation := make([]*storage_v1.Operation, len(operations))
```
yurishkuro: is this backwards compatible if the plugin only returns operation names?

plugin/storage/grpc/shared/grpc_server_test.go
```diff
@@ -87,14 +87,23 @@ func TestGRPCServerGetServices(t *testing.T) {
 
 func TestGRPCServerGetOperations(t *testing.T) {
 	withGRPCServer(func(r *grpcServerTest) {
-		r.impl.spanReader.On("GetOperations", mock.Anything, "service-a").
-			Return([]string{"operation-a"}, nil)
+		r.impl.spanReader.On("GetOperations",
```
yurishkuro: this needs backwards compatibility test

plugin/storage/integration/kafka_test.go
```diff
@@ -119,7 +119,8 @@ func (r *ingester) GetServices(ctx context.Context) ([]string, error) {
 	return nil, nil
 }
 
-func (r *ingester) GetOperations(ctx context.Context, service string) ([]string, error) {
+func (r *ingester) GetOperations(ctx context.Context, query *spanstore.OperationQueryParameters) (
+	[]*spanstore.Operation, error) {
```
yurishkuro: formatting

storage/spanstore/metrics/decorator.go
```diff
@@ -106,9 +106,10 @@ func (m *ReadMetricsDecorator) GetServices(ctx context.Context) ([]string, error
 }
 
 // GetOperations implements spanstore.Reader#GetOperations
-func (m *ReadMetricsDecorator) GetOperations(ctx context.Context, service string) ([]string, error) {
+func (m *ReadMetricsDecorator) GetOperations(ctx context.Context, query *spanstore.OperationQueryParameters) (
+	[]*spanstore.Operation, error) {
```
yurishkuro: formatting

plugin/storage/grpc/shared/grpc_server.go
guo0693: It's backward compatible in that case, since the new return structure will just have empty spanKind.

But I do need to return the response in below structure:
```
storage_v1.GetOperationsResponse{
		Operations:   uniqueNames,
		OperationsV2: grpcOperation,
	}
```
So that old grpc client can still get unique operations names

yurishkuro: but I don't see you reading `operationNames` field. The old plugin will only return that field.

plugin/storage/cassandra/spanstore/operation_names.go
```diff
@@ -126,36 +128,35 @@ func NewOperationNamesStorage(
...
 }
 
 // GetOperations returns all operations for a specific service traced by Jaeger
-func (s *OperationNamesStorage) GetOperations(service string) ([]string, error) {
-	operations, err := s.table.getOperations(s, &spanstore.OperationQueryParameters{
-		ServiceName: service,
+func (s *OperationNamesStorage) GetOperations(
+	query *spanstore.OperationQueryParameters,
+) ([]*spanstore.Operation, error) {
+	return s.table.getOperations(s, &spanstore.OperationQueryParameters{
```
yurishkuro: you can pass `query` to this function directly

plugin/storage/badger/spanstore/reader.go
```diff
@@ -237,8 +237,11 @@ func (r *TraceReader) GetServices(ctx context.Context) ([]string, error) {
 }
 
 // GetOperations fetches operations in the service and empty slice if service does not exists
-func (r *TraceReader) GetOperations(ctx context.Context, service string) ([]string, error) {
-	return r.cache.GetOperations(service)
+func (r *TraceReader) GetOperations(
+	ctx context.Context,
+	query *spanstore.OperationQueryParameters,
+) ([]*spanstore.Operation, error) {
```
yurishkuro: before it's too late, I suggest changing this to return slice of values `[]spanstore.Operation`. There's zero benefit in returning pointers, only extra allocations.

plugin/storage/cassandra/spanstore/dbmodel/operation.go
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) 2019 The Jaeger Authors.
+// Copyright (c) 2017 Uber Technologies, Inc.
```
yurishkuro: this line doesn't belong

plugin/storage/cassandra/spanstore/operation_names.go
```diff
@@ -50,7 +50,8 @@ type tableMeta struct {
 	queryByKindStmt  string
 	queryStmt        string
 	createWriteQuery func(query cassandra.Query, service, kind, opName string) cassandra.Query
-	getOperations    func(s *OperationNamesStorage, query *spanstore.OperationQueryParameters) ([]*spanstore.Operation, error)
+	getOperations    func(s *OperationNamesStorage, query *spanstore.OperationQueryParameters) (
+		[]*spanstore.Operation, error)
```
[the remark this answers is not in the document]
yurishkuro: this should be either multi-line arguments, or a single line signature as before. Currently it looks like some weird Java formatting.

plugin/storage/grpc/shared/grpc_server.go
yurishkuro: actually, never mind, it's fine. The comment applies to grpc client, which is what calls potentially old plugin.

plugin/storage/grpc/shared/grpc_server.go
```diff
@@ -78,13 +78,33 @@ func (s *grpcServer) GetServices(ctx context.Context, r *storage_v1.GetServicesR
...
+		}
+		operationNameSet[operation.Name] = struct{}{}
+	}
+	var uniqueNames []string
+	for name := range operationNameSet {
+		uniqueNames = append(uniqueNames, name)
+	}
 	return &storage_v1.GetOperationsResponse{
-		Operations: operations,
+		Operations:   uniqueNames,
```
yurishkuro: there is no need to return this field. If the plugin uses this version of grpc server pkg, it is already working with client that understands the new field.

plugin/storage/integration/kafka_test.go
```diff
@@ -119,7 +119,9 @@ func (r *ingester) GetServices(ctx context.Context) ([]string, error) {
 	return nil, nil
 }
 
-func (r *ingester) GetOperations(ctx context.Context, service string) ([]string, error) {
+func (r *ingester) GetOperations(
+	ctx context.Context,
+	query *spanstore.OperationQueryParameters) ([]*spanstore.Operation, error) {
```
yurishkuro:
```go
func (r *ingester) GetOperations(
	ctx context.Context,
	query *spanstore.OperationQueryParameters,
) ([]*spanstore.Operation, error) {
```

plugin/storage/memory/memory_test.go
```diff
@@ -224,18 +227,42 @@ func TestStoreGetServices(t *testing.T) {
...
+		assert.NoError(t, store.WriteSpan(childSpan1))
+		assert.NoError(t, store.WriteSpan(childSpan2))
+		assert.NoError(t, store.WriteSpan(childSpan2_1))
+		operations, err := store.GetOperations(context.Background(),
+			&spanstore.OperationQueryParameters{
+				ServiceName: childSpan1.Process.ServiceName,
+				SpanKind:    "server",
+			})
 		assert.NoError(t, err)
 		assert.Len(t, operations, 1)
```
yurishkuro: it's much easier to read if you just used assert.Equal and pass the expected values in a structured form, e.g.
```
expected := []spanstore.Operation{{Name: .., Kind: ...}}
assert.Equal(t, expected, operations)
```

plugin/storage/grpc/shared/grpc_server.go
guo0693: In this PR, all plugins are updated to return []*spanReader.Operation struct{
	Name     string
	SpanKind string
}
For badger & ES, the spanKind will just be empty.

Are you saying it's possible for grpcService to be configured to use old version plugins?

plugin/storage/cassandra/spanstore/operation_names.go
guo0693: Good catch! I got lost during refactors between branches...

plugin/storage/badger/spanstore/reader.go
guo0693: I can also change the parameter so the signature will be:

GetOperations(ctx context.Context, query OperationQueryParameters) ([]Operation, error)

I will not increase the scope of this PR by updating below methods:
```
FindTraces(ctx context.Context, query *TraceQueryParameters) ([]*model.Trace, error)
FindTraceIDs(ctx context.Context, query *TraceQueryParameters) ([]model.TraceID, error)
```

plugin/storage/grpc/shared/grpc_server.go
guo0693: Are there any external grpc client that make send request to the server?

plugin/storage/cassandra/spanstore/operation_names.go
yurishkuro: that's because you're using a convoluted process for stacking diffs, instead of the usual chained branches.

plugin/storage/badger/spanstore/reader.go
yurishkuro: ok

plugin/storage/grpc/shared/grpc_server.go
yurishkuro: you may want to review how plugins work: https://github.com/jaegertracing/jaeger/tree/master/plugin/storage/grpc

The client always runs in jaeger binary. But the plugin implements the server and could be from previous versions of Jaeger, so it will only return operation names.

guo0693: Oh, you mean new grpc client send request to old grpc server. That's interesting use case. In my previous work, we only have issue of old client against new server.

plugin/storage/cassandra/spanstore/operation_names.go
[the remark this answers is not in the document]
guo0693: Yeah, auto wrapping from intellij doesn't work that well for go.

plugin/storage/cassandra/spanstore/operation_names.go
guo0693: This branch is based on branch #1921-2, if that's what you mean by chained branches. 
But we need to merge the PRs in reverse order into chained branches. E.g. PR3 merge to branch 2 then PR2 merge into branch 1, then PR1 into master.
Every time we make changes in branch 1, we will need to propagate changes from branch 1 all the way to branch N.

I am not sure if I should open the PR for endpoint changes against this branch.

yurishkuro:
> But we need to merge the PRs in reverse order into chained branches. E.g. PR3 merge to branch 2 then PR2 merge into branch 1, then PR1 into master.

I've never tried this, but GH allows you  to change the base branch of a PR: https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/changing-the-base-branch-of-a-pull-request

guo0693:
> > But we need to merge the PRs in reverse order into chained branches. E.g. PR3 merge to branch 2 then PR2 merge into branch 1, then PR1 into master.
> 
> I've never tried this, but GH allows you to change the base branch of a PR: https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/changing-the-base-branch-of-a-pull-request

It doesn't work that well for us:
master
branch1
branch2

assume branch1 has several commits, branch2 is based on branch1. After branch1 merged to master, those commits are squashed, when we update PR from branch2 to compare to master, the commit history is different, we will need to merge master into branch2 and solve merge conflicts first, this will apply to all the dependent branches and require extra reviews.

guo0693:
Btw, this is the 3rd PR for the endpoint changes: https://github.com/guo0693/jaeger/pull/2/files

plugin/storage/cassandra/spanstore/reader_test.go
```diff
@@ -73,10 +74,19 @@ func TestSpanReaderGetServices(t *testing.T) {
 
 func TestSpanReaderGetOperations(t *testing.T) {
 	withSpanReader(func(r *spanReaderTest) {
-		r.reader.operationNamesReader = func(string) ([]string, error) { return []string{"operation-a"}, nil }
-		s, err := r.reader.GetOperations(context.Background(), "service-x")
+		r.reader.operationNamesReader =
+			func(parameters spanstore.OperationQueryParameters) ([]spanstore.Operation, error) {
+				return []spanstore.Operation{
```
yurishkuro: nit: you could pull this slice into `expectedOperations` var and reuse it in the assert.

plugin/storage/grpc/shared/grpc_client_test.go
```diff
@@ -119,17 +119,33 @@ func TestGRPCClientGetServices(t *testing.T) {
...
+		assert.Equal(t, []spanstore.Operation{{Name: "operation-a"}}, s)
+	})
+}
+
+func TestGRPCClientGetOperationsV2(t *testing.T) {
+	withGRPCClient(func(r *grpcClientTest) {
+		r.spanReader.On("GetOperations", mock.Anything, &storage_v1.GetOperationsRequest{
+			Service: "service-a",
+		}).Return(&storage_v1.GetOperationsResponse{
+			Operations: []*storage_v1.Operation{{Name: "operation-a"}},
```
yurishkuro: V2 plugin should return the result with SpanKind

plugin/storage/grpc/shared/grpc_server_test.go
```diff
@@ -87,14 +87,25 @@ func TestGRPCServerGetServices(t *testing.T) {
...
+			spanstore.OperationQueryParameters{ServiceName: "service-a"}).
+			Return(expOperations, nil)
 
-		s, err := r.server.GetOperations(context.Background(), &storage_v1.GetOperationsRequest{
+		resp, err := r.server.GetOperations(context.Background(), &storage_v1.GetOperationsRequest{
 			Service: "service-a",
 		})
 		assert.NoError(t, err)
-		assert.Equal(t, &storage_v1.GetOperationsResponse{Operations: []string{"operation-a"}}, s)
+		for i, operation := range resp.Operations {
```
yurishkuro: should also assert the length

plugin/storage/integration/integration_test.go
```diff
@@ -149,14 +149,18 @@ func (s *StorageIntegration) testGetLargeSpan(t *testing.T) {
 func (s *StorageIntegration) testGetOperations(t *testing.T) {
 	defer s.cleanUp(t)
 
-	expected := []string{"example-operation-1", "example-operation-3", "example-operation-4"}
+	expected := []spanstore.Operation{
+		{Name: "example-operation-1"},
+		{Name: "example-operation-3"},
+		{Name: "example-operation-4"}}
```
yurishkuro: hmm, how would this work when some storage impls return SpanKind while others do not?

plugin/storage/memory/memory.go
```diff
@@ -34,7 +34,7 @@ type Store struct {
 	ids        []*model.TraceID
 	traces     map[model.TraceID]*model.Trace
 	services   map[string]struct{}
-	operations map[string]map[string]struct{}
+	operations map[string]map[string]map[string]struct{}
```
yurishkuro: I find it very hard to reason about 3 nested maps. IIRC the `services` is just a set of services, and `operations` is a map of service -> operations set. In the new implementation you want it to be a map of service -> operation+spanKind. So can it be something like `map[string]map[Operation]struct{}`?

plugin/storage/integration/integration_test.go
guo0693: Updated the test to include spanKind, when spanKind is missing, it will be treated as empty string in the Operation struct.

plugin/storage/memory/memory.go
guo0693: Updated

plugin/storage/integration/integration_test.go
```diff
@@ -149,14 +149,18 @@ func (s *StorageIntegration) testGetLargeSpan(t *testing.T) {
 func (s *StorageIntegration) testGetOperations(t *testing.T) {
 	defer s.cleanUp(t)
 
-	expected := []string{"example-operation-1", "example-operation-3", "example-operation-4"}
+	expected := []spanstore.Operation{
+		{Name: "example-operation-1"},
+		{Name: "example-operation-3", SpanKind: "server"},
+		{Name: "example-operation-4", SpanKind: "client"}}
```
yurishkuro: ES tests are failing, as I was expecting (but Badger tests do not, which is unexpected)
```
        integration_test.go:168: 
            	Error Trace:	integration_test.go:168
            	Error:      	Should be true
            	Test:       	TestElasticsearchStorage_AllTagsAsObjectFields/GetOperations
        integration_test.go:169: 	 Expected: [{example-operation-1 } {example-operation-3 server} {example-operation-4 client}]
        integration_test.go:170: 	 Actual  : [{example-operation-1 } {example-operation-3 } {example-operation-4 }]
```

I think you may need to introduce some "capability" flag into StorageIntegration so that the storage backend can declare whether it supports returning spanKind or not, and then the test should perform different expectation asserts depending on that flag.

PR #1943, opened 2019-11-25
Update grpc & http endpoint to query operations by service & spanKind

guo0693:

- Relates to #1921 

- Update grpc & http endpoint to query operations by service & spanKind
- legacy http endpoint will stay unchanged: /services/%s/operations


cmd/query/app/grpc_handler_test.go
```diff
@@ -427,7 +427,8 @@ func TestGetOperationsSuccessGRPC(t *testing.T) {
 		assert.NoError(t, err)
 		assert.Equal(t, len(expectedOperations), len(res.Operations))
 		for i, actualOp := range res.Operations {
-			assert.Equal(t, expectedOperations[i].Name, actualOp)
+			assert.Equal(t, expectedOperations[i].Name, actualOp.Name)
```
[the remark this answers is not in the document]
guo0693: This one is tricky, spanreader returns []spanstore.Operation vs grpc response.Operations []*api_v2.Operation

cmd/query/app/http_handler.go
```diff
@@ -175,12 +190,24 @@ func (aH *APIHandler) getOperations(w http.ResponseWriter, r *http.Request) {
...
+	spanKind := r.FormValue(spanKindParam)
+	operations, err := aH.queryService.GetOperations(
+		r.Context(),
+		spanstore.OperationQueryParameters{ServiceName: service, SpanKind: spanKind},
+	)
+
 	if aH.handleError(w, err, http.StatusInternalServerError) {
 		return
 	}
+	data := make([]*ui.Operation, len(operations))
```
yurishkuro: I think this would work just fine without pointers

cmd/query/app/http_handler.go
```diff
@@ -157,13 +157,28 @@ func (aH *APIHandler) getOperationsLegacy(w http.ResponseWriter, r *http.Request
 	vars := mux.Vars(r)
 	// given how getOperationsLegacy is bound to URL route, serviceParam cannot be empty
 	service, _ := url.QueryUnescape(vars[serviceParam])
-	operations, err := aH.queryService.GetOperations(r.Context(), service)
+	//for backwards compatibility, we will retrieve operations with all span kind
```
yurishkuro: * the accepted practice (even enforced by some linters) is to leave a space after `//`
* all span kinds

cmd/query/app/http_handler.go
```diff
@@ -157,13 +157,28 @@ func (aH *APIHandler) getOperationsLegacy(w http.ResponseWriter, r *http.Request
...
+	operations, err := aH.queryService.GetOperations(r.Context(),
+		spanstore.OperationQueryParameters{
+			ServiceName: service,
+			SpanKind:    "",
```
yurishkuro: `// include all kinds`

cmd/query/app/http_handler.go
```diff
@@ -157,13 +157,28 @@ func (aH *APIHandler) getOperationsLegacy(w http.ResponseWriter, r *http.Request
...
+		})
+
 	if aH.handleError(w, err, http.StatusInternalServerError) {
 		return
 	}
+	//only return unique operation names
```
yurishkuro: space after `//`

cmd/query/app/http_handler_test.go
```diff
@@ -484,16 +484,27 @@ func TestGetOperationsSuccess(t *testing.T) {
...
-	readMock.On("GetOperations", mock.AnythingOfType("*context.valueCtx"),
-		spanstore.OperationQueryParameters{ServiceName: "abc/trifle"}).Return(expectedOperations, nil).Once()
+	readMock.On(
+		"GetOperations",
+		mock.AnythingOfType("*context.valueCtx"),
+		spanstore.OperationQueryParameters{ServiceName: "abc/trifle"},
+	).Return(expectedOperations, nil).Once()
+
+	var response struct {
+		Operations []*ui.Operation   `json:"data"`
```
yurishkuro: pointers not needed

model/proto/api_v2/query.proto
```diff
@@ -109,10 +109,17 @@ message GetServicesResponse {
...
+}
+
+message Operation {
+    string name = 1;
+    string span_kind = 2;
 }
 
 message GetOperationsResponse {
-  repeated string operations = 1;
+  repeated string operationNames = 1; //deprecated
```
yurishkuro: Please include an update to CHANGELOG, as this is a breaking change. I suggest also mentioning grpc plugin changes and the db schema changes / upgrade instructions.

If we are not populating operationNames field, we can rename it to `_`. But it might be better to populate both collections and have a config options to disable populating the names for installations that don't have legacy users and want a bit more efficiency. This way installation can be upgraded to a new version and clients can switch to the new API not in lock-step.

cmd/query/app/http_handler.go
```diff
@@ -175,12 +190,24 @@ func (aH *APIHandler) getOperations(w http.ResponseWriter, r *http.Request) {
...
+	data := make([]*ui.Operation, len(operations))
+	for i, operation := range operations {
+		data[i] = &ui.Operation{
+			Name:     operation.Name,
+			SpanKind: operation.SpanKind,
+		}
+	}
 	structuredRes := structuredResponse{
-		Data:  operations,
+		Data:  data,
```
yurishkuro: please include this as breaking change in the CHANGELOG

model/proto/api_v2/query.proto
guo0693: I would prefer to rename it to _ if grpc server is not widely used, otherwise we will create tech debt for future cleanup (to remove the flag and code that generate operation names from the structs) and this usually ends up stay in code for months to years.

guo0693: Per offline discussion, we will set both operationNames and operations in the grpc response, and remove operationNames after 2+ releases

Issue #943, opened 2018-07-19
Memory-sensitive collector queue size


jpkrohling:
## Requirement - what kind of business use case are you trying to solve?
Avoiding dropping of spans on the collector side when the backing storage isn't really fast, but enough memory is available for a bigger buffer on the collector size

## Problem - what in Jaeger blocks you from solving the requirement?
Under stress conditions, the default queue size of 2000 causes spans to be dropped quite frequently, even though there's enough memory to have a bigger queue size

## Proposal - what do you suggest to solve the problem or improve the existing situation?
The first idea is to have a dynamic collector queue size, with the following characteristics

- [ ] Heuristics to determine the "average" span size in memory (could also be used for the in-memory storage)
- [ ] Flag to specify a memory size for the queue (default: half of the host's memory? 25%?)
- [ ] Adjust the queue size based a simple division of the above
- [ ] Update a metrics gauge with the current queue size
- [ ] Flag to enable usage of this new collector queue mechanism

yurishkuro:
This does not seem related to #355, correct? Maybe retitle to "memory-sensitive queue size" rather than "dynamic".

isaachier:
So there are a bunch of algorithms that have been studied and used for network scheduling (https://en.wikipedia.org/wiki/Network_scheduler). My suggestion is either to reimplement one of them directly, or potentially leverage existing OS configurations to drop packets from spans according to one of those existing strategies.

PR #1949, opened 2019-11-28
Added 'resize' operation to BoundedQueue

jpkrohling:

- This PR is a prerequisite for #943

- This adds a new `Resize(int)` method to the `BoundedQueue`

### DIFF
--- a/pkg/queue/bounded_queue.go
+++ b/pkg/queue/bounded_queue.go
@@ -93,7 +93,7 @@ func (q *BoundedQueue) Produce(item interface{}) bool {
 	// we might have two concurrent backing queues at the moment
 	// their combined size is stored in q.size, and their combined capacity
 	// should match the capacity of the new queue
-	if atomic.LoadInt32(&q.size) >= int32(q.Capacity()) && q.Capacity() > 0 {
+	if q.Size() >= q.Capacity() && q.Capacity() > 0 {
 		// current consumers of the queue (like tests) expect the queue capacity = 0 to work
 		// so, we don't drop items when the capacity is 0
 		q.onDroppedItem(item)
--- a/pkg/queue/bounded_queue.go
+++ b/pkg/queue/bounded_queue.go
@@ -19,6 +19,7 @@ import (
 	"sync"
 	"sync/atomic"
 	"time"
+	"unsafe"
 
 	"github.com/uber/jaeger-lib/metrics"
 )
@@ -29,22 +30,23 @@ import (
 // channels, with a special Reaper goroutine that wakes up when the queue is full and consumers
 // the items from the top of the queue until its size drops back to maxSize
 type BoundedQueue struct {
-	capacity      int
 	size          int32
 	onDroppedItem func(item interface{})
-	items         chan interface{}
+	items         *chan interface{}
 	stopCh        chan struct{}
 	stopWG        sync.WaitGroup
 	stopped       int32
+	consumer      func(item interface{})
+	workers       int
 }
 
 // NewBoundedQueue constructs the new queue of specified capacity, and with an optional
 // callback for dropped items (e.g. useful to emit metrics).
 func NewBoundedQueue(capacity int, onDroppedItem func(item interface{})) *BoundedQueue {
+	queue := make(chan interface{}, capacity)
 	return &BoundedQueue{
-		capacity:      capacity,
 		onDroppedItem: onDroppedItem,
-		items:         make(chan interface{}, capacity),
+		items:         &queue,
 		stopCh:        make(chan struct{}),
 	}
 }
@@ -52,23 +54,31 @@ func NewBoundedQueue(capacity int, onDroppedItem func(item interface{})) *Bounde
 // StartConsumers starts a given number of goroutines consuming items from the queue
 // and passing them into the consumer callback.
 func (q *BoundedQueue) StartConsumers(num int, consumer func(item interface{})) {
+	q.workers = num
+	q.consumer = consumer
 	var startWG sync.WaitGroup
-	for i := 0; i < num; i++ {
+	for i := 0; i < q.workers; i++ {
 		q.stopWG.Add(1)
 		startWG.Add(1)
-		go func() {
+		go func(queue chan interface{}) {
 			startWG.Done()
 			defer q.stopWG.Done()
 			for {
 				select {
-				case item := <-q.items:
-					atomic.AddInt32(&q.size, -1)
-					consumer(item)
+				case item, ok := <-queue:
+					if ok {
+						atomic.AddInt32(&q.size, -1)
+						q.consumer(item)
+					} else {
+						// channel closed, finish worker
+						return
+					}
 				case <-q.stopCh:
+					// the whole queue is closing, finish worker
 					return
 				}
 			}
-		}()
+		}(*q.items)
 	}
 	startWG.Wait()
 }
@@ -79,11 +89,23 @@ func (q *BoundedQueue) Produce(item interface{}) bool {
 		q.onDroppedItem(item)
 		return false
 	}
+
+	// we might have two concurrent backing queues at the moment
+	// their combined size is stored in q.size, and their combined capacity
+	// should match the capacity of the new queue
+	if atomic.LoadInt32(&q.size) >= int32(q.Capacity()) && q.Capacity() > 0 {
+		// current consumers of the queue (like tests) expect the queue capacity = 0 to work
+		// so, we don't drop items when the capacity is 0
+		q.onDroppedItem(item)
+		return false
+	}
+
 	select {
-	case q.items <- item:
+	case *q.items <- item:
 		atomic.AddInt32(&q.size, 1)
 		return true
 	default:
+		// should not happen, as overflows should have been captured earlier
 		if q.onDroppedItem != nil {
 			q.onDroppedItem(item)
 		}
@@ -97,7 +119,7 @@ func (q *BoundedQueue) Stop() {
 	atomic.StoreInt32(&q.stopped, 1) // disable producer
 	close(q.stopCh)
 	q.stopWG.Wait()
-	close(q.items)
+	close(*q.items)
 }
 
 // Size returns the current size of the queue
@@ -107,7 +129,7 @@ func (q *BoundedQueue) Size() int {
 
 // Capacity returns capacity of the queue
 func (q *BoundedQueue) Capacity() int {
-	return q.capacity
+	return cap(*q.items)
 }
 
 // StartLengthReporting starts a timer-based goroutine that periodically reports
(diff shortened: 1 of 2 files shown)


yurishkuro:
how is it supposed to be used? who would call the Resize()? Isn't it possible to just pre-determine the queue size upfront?

jpkrohling:
> how is it supposed to be used? who would call the Resize()? 

A change to the span processor will make it keep track of the number of spans and the total bytes it has seen during the process lifetime. Once enough spans have been recorded, the span processor starts to calculate what's the ideal queue size based on the average number of spans and available memory. It can then decide to trigger a `Resize()` operation.

I have a local branch that does the above, but I thought it would be cleaner to separate those two features into separate PRs.

> Isn't it possible to just pre-determine the queue size upfront?

When we know what's the expected average span size, yes. But I've not heard of any cases where people know in advance what's the average span size, making the queue size hard to calculate.

jpkrohling:
I just added a benchmark to the tests, to make ensure that I'm not bringing performance problems to the queue. Here are the results comparing master with this PR, which shows no performance degradation for current users of the queue (unless I'm making something really wrong with the benchmark):

```
$ git branch --show-current 
master

$ go test -benchmem -run=^$ ./pkg/queue -bench "^(BenchmarkBoundedQueue)$"
goos: linux
goarch: amd64
pkg: github.com/jaegertracing/jaeger/pkg/queue
BenchmarkBoundedQueue-8   	19407266	        52.2 ns/op	       8 B/op	       1 allocs/op
PASS
ok  	github.com/jaegertracing/jaeger/pkg/queue	1.081s

$ git branch --show-current 
Resize-bounded-queue

$ go test -benchmem -run=^$ ./pkg/queue -bench "^(BenchmarkBoundedQueue)$"
goos: linux
goarch: amd64
pkg: github.com/jaegertracing/jaeger/pkg/queue
BenchmarkBoundedQueue-8   	20110519	        51.4 ns/op	       8 B/op	       1 allocs/op
PASS
ok  	github.com/jaegertracing/jaeger/pkg/queue	1.099s
```

yurishkuro:
I am skeptical about relying on the average span size, because the actual size differences can be very large, 2-3 orders of magnitude, e.g a basic span vs. span with a stack trace vs. span with request payload added to the logs. I think if we want to go by memory size, we should go by the actual size of spans in the queue. The proto/domain model already implements Size() method iirc, which could be used as an approximation of the actual memory size, and we would need to re-implement the queue using an explicit buffer rather than a bound channel (might be worth looking around for an existing implementation).

jpkrohling:
> the actual size differences can be very large, 2-3 orders of magnitude, e.g a basic span vs. span with a stack trace vs. span with request payload added to the logs

You are right, but in reality, this wouldn't be a big problem. The queue sizes will typically be big enough to make outliers non-relevant. In [my working branch](https://github.com/jpkrohling/jaeger/tree/943-DynQueueSize) for #943, spans generated by `tracegen` have an average of 1.5KiB, which means a queue size of about 700k items at 1GiB. And even if spans start being bigger all of a sudden, the [periodic background process](https://github.com/jpkrohling/jaeger/blob/bac686546a9cc8bae07c016e9ad05bdfd8dd52d2/cmd/collector/app/span_processor.go#L205-L209) will readjust the queue size as long as the difference is bigger [than a few % points](https://github.com/jpkrohling/jaeger/blob/bac686546a9cc8bae07c016e9ad05bdfd8dd52d2/cmd/collector/app/span_processor.go#L258-L263). Note that the "memory" aspect is an approximation: there shouldn't be a crash if we end up using more memory for a short period of time. And most of the times, the queue should be empty anyway ;-)

I would give this simple approach a try first and see how it works in the real world.

jpkrohling:
@yurishkuro any feedback on this one?

pkg/queue/bounded_queue.go
```diff
@@ -97,17 +105,52 @@ func (q *BoundedQueue) Stop() {
...
+	// if the capacity size hasn't changed, there's no need to resize
+	if cap(q.items) == newCap {
+		return false
+	}
+
+	// we need exclusive rw access to the internal queue
+	q.resizeLock.Lock()
+	defer q.resizeLock.Unlock()
+
+	new := make(chan interface{}, newCap)
```
yurishkuro: * what if newCap is less than currentCap and current size? This will probably deadlock
* please don't use `new` keyword as var name

jpkrohling: Funny, I thought I had added a test for the downscale, but looks like I haven't. Will add one.

yurishkuro:
I wonder if it's possible to avoid the mutex by approaching this differently. Go routines are cheap to start, so rather than copying the messages into another queue, we could start another set of workers reading from a new channel, close the original channel and let those old workers drain it naturally and exit.

jpkrohling:
That could work, let me give it a try.

jpkrohling:
PR updated, to use a new channel for the resized queue.

jpkrohling:
@yurishkuro would you please re-review?

pkg/queue/bounded_queue_test.go
```diff
@@ -156,3 +157,173 @@ func (s *consumerState) assertConsumed(expected map[string]bool) {
...
+	})
+
+	var firstConsumer, secondConsumer, releaseConsumers sync.WaitGroup
+	firstConsumer.Add(1)
+	secondConsumer.Add(1)
+	releaseConsumers.Add(1)
+
+	released, resized := false, false
+	q.StartConsumers(1, func(item interface{}) {
+		if !resized { // we'll have a second consumer once the queue is resized
```
pavolloffay: I didn't find the second consumer

pkg/queue/bounded_queue_test.go
```diff
@@ -156,3 +157,173 @@ func (s *consumerState) assertConsumed(expected map[string]bool) {
...
+	})
+
+	var consumer, releaseConsumers sync.WaitGroup
+	consumer.Add(1)
+	releaseConsumers.Add(1)
+
+	released := false
+	q.StartConsumers(1, func(item interface{}) {
+		// once we release the lock, we might end up with multiple calls to reach this
+		if !released {
```
pavolloffay: Released does not seem to be needed

pkg/queue/bounded_queue_test.go
```diff
@@ -156,3 +157,173 @@ func (s *consumerState) assertConsumed(expected map[string]bool) {
...
+	assert.EqualValues(t, 4, q.Capacity())
+	assert.EqualValues(t, q.Capacity(), q.Size())
+	assert.EqualValues(t, q.Capacity(), len(*q.items))
+
+	assert.True(t, q.Resize(2))
+	assert.False(t, q.Produce("f")) // dropped
+
+	assert.EqualValues(t, 2, q.Capacity())
+	assert.EqualValues(t, 4, q.Size())      // the queue will eventually drain, but it will live for a while over capacity
+	assert.EqualValues(t, 0, len(*q.items)) // the new queue is empty, as the old queue is still full and over capacity
```
pavolloffay: Shouldn't we keep accepting here? There are free spots in the queue.

pkg/queue/bounded_queue.go
```diff
@@ -79,11 +89,23 @@ func (q *BoundedQueue) Produce(item interface{}) bool {
 		q.onDroppedItem(item)
 		return false
 	}
+
+	// we might have two concurrent backing queues at the moment
+	// their combined size is stored in q.size, and their combined capacity
+	// should match the capacity of the new queue
+	if atomic.LoadInt32(&q.size) >= int32(q.Capacity()) && q.Capacity() > 0 {
```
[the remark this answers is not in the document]
jpkrohling: There are consumers of this queue that are not setting the queue size. IIRC, one of them is in the span processor tests. Agree with your earlier suggestion though.

pkg/queue/bounded_queue_test.go
jpkrohling: Around line 200:

https://github.com/jaegertracing/jaeger/blob/a7819b895930fc85727750f2a00cf5b21bc8cbb7/pkg/queue/bounded_queue_test.go#L197-L200

pkg/queue/bounded_queue_test.go
jpkrohling: Why's that? Note that there's quite a bunch of go funcs just waiting for this lock to be released to start calling consumers with other items. If we have *not* released, we mark one step down in the work group. If we have released the consumer, we do not want to count down further, as we'd reach a negative count.

pkg/queue/bounded_queue_test.go
jpkrohling: There are free spots in the individual channel, but if the queue size was set by the user to 3000, it's wrong to have 6000 items in the two queues combined.

pkg/queue/bounded_queue.go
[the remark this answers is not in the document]
pavolloffay: I see there is a test which sets the size to `0`

pkg/queue/bounded_queue_test.go
pavolloffay: I mean you can remove this parameter altogether. It's not changed during the test - only at the end which does not affect tests and the default value satisfy this if statement.

pavolloffay: Removal of this variable does not have any effect on the tests. It seems redundant

jpkrohling: It needs to be there, otherwise the test will intermittently fail due to "negative counts" in the work group.

jpkrohling:
Thanks!

pkg/queue/bounded_queue.go
```diff
@@ -29,46 +30,55 @@ import (
...
+						// channel closed, finish worker
+						return
+					}
 				case <-q.stopCh:
+					// the whole queue is closing, finish worker
 					return
 				}
 			}
-		}()
+		}(*q.items)
```
yurishkuro: Accessing the pointer like this creates an unnecessary race condition where some workers can theoretically start against a different queue. A simpler solution is to dereference the field outside of the loop.

pkg/queue/bounded_queue.go
```diff
@@ -29,46 +30,55 @@ import (
...
 // the items from the top of the queue until its size drops back to maxSize
 type BoundedQueue struct {
-	capacity      int
 	size          int32
 	onDroppedItem func(item interface{})
-	items         chan interface{}
+	items         *chan interface{}
 	stopCh        chan struct{}
 	stopWG        sync.WaitGroup
 	stopped       int32
```
yurishkuro: we should move this field to the top to ensure word alignment

pkg/queue/bounded_queue.go
```diff
@@ -29,46 +30,55 @@ import (
 // channels, with a special Reaper goroutine that wakes up when the queue is full and consumers
 // the items from the top of the queue until its size drops back to maxSize
 type BoundedQueue struct {
-	capacity      int
 	size          int32
 	onDroppedItem func(item interface{})
-	items         chan interface{}
+	items         *chan interface{}
```
yurishkuro: This is being updated with a CAS in one place but accessed directly in other places, which is technically not thread safe. I would rather we used a normal mutex or `atomic.Value`.

pkg/queue/bounded_queue.go
```diff
@@ -79,11 +89,23 @@ func (q *BoundedQueue) Produce(item interface{}) bool {
...
+
+	// we might have two concurrent backing queues at the moment
+	// their combined size is stored in q.size, and their combined capacity
+	// should match the capacity of the new queue
+	if q.Size() >= q.Capacity() && q.Capacity() > 0 {
+		// current consumers of the queue (like tests) expect the queue capacity = 0 to work
+		// so, we don't drop items when the capacity is 0
+		q.onDroppedItem(item)
+		return false
+	}
```
yurishkuro: I don't understand this part. A second queue should never be the concern of Produce(), since it only operates on the _current_ queue. And why do we need capacity=0 for tests?

jpkrohling: > A second queue should never be the concern of Produce()

Correct, the comment is there only to make it explicit that there's no need to check two queues, as the state from both is handled at `q.Size()`. I can remove the comment if it's confusing.

> And why do we need capacity=0 for tests?

When no `Option.QueueSize()` is set and used in the span processor, the default queue size is 0 in tests like `cmd/collector/app/span_processor_test.go`. I changed the test to explicitly set a queue size, and removed the conditional here, as @pavolloffay had similar concerns.

Issue #1657, opened 2019-07-06
Jaeger clients should generate full-length trace/span IDs


yurishkuro:
## Requirement - what kind of business use case are you trying to solve?

As a user, I should be able to use Trace/span IDs encoded in the trace context (e.g. `uber-trace-id` header) or exposed via `SpanContext.toTraceID()` methods to refer to traces in the UI.

## Problem - what in Jaeger blocks you from solving the requirement?

Currently, many Jaeger clients drop leading zeroes from hex representation of trace/span IDs, resulting in inconsistent strings returned in JSON to the UI (relates to #1578). This is done as an optimization of the trace context header length, but it's a silly one since it only affects about 1/16th of all traces.

## Proposal - what do you suggest to solve the problem or improve the existing situation?

Client libraries should always output hex strings of 16 or 32 characters in length, depending on on whether 64bit or 128bit IDs are used, padded with zeros on the left to the desired length.

For backwards compatibility, clients should still be able to parse shorted hex strings.

* [x] Node.js client https://github.com/jaegertracing/jaeger-client-node/issues/391
* [x] Go client https://github.com/jaegertracing/jaeger-client-go/pull/472
* [x] Java client https://github.com/jaegertracing/jaeger-client-java/pull/746 (released in 1.5.0)
* [ ] Python client
* [x] C# client https://github.com/jaegertracing/jaeger-client-csharp/pull/171 (released in 0.4.0)
* [x] C++ client https://github.com/jaegertracing/jaeger-client-cpp/pull/259 (pending 0.7.0 release)

SezalAgrawal:
Is this issue resolved?

yurishkuro:
It is not, see the checklist in the description.

yurishkuro:
Per #3362, we're sunsetting Jaeger clients.

PR #1956, opened 2019-11-30
Do not remove leading zeros in trace ID hex string

### LINKED ISSUE #1578: [jaeger-query] Leading zeros stripped from trace IDs

I would like to view traces in Jaeger UI where the trace ID has leading 0s.

When jaeger-query returns data for traces where there are leading 0s in the trace ID, the leading 0s are stripped from the trace ID, within the data. E.g. for a trace with ID `0123`, all of the spans returned refer to trace ID `123`.

For each trace returned by jaeger-query, this applies to:

* The trace ID on the root of the trace data object
* The trace ID fields in each span
* The trace ID fields in internal references 

The zeros should not be stripped.

yurishkuro:
Resolves #1578

## Motivation

Among randomly generated and uniformly distributed trace IDs, only 1/16th of them start with 0 followed by a significant digit, 1/256th start with two 0s, and so on in decreasing geometric progression. Therefore, trimming the leading 0s is a very modest optimization on the size of the data being transmitted or stored.

However, trimming 0s leads to ambiguities when the IDs are used as correlations with other monitoring systems, such as logging, that treat the IDs as opaque strings and cannot establish the equivalence between padded and unpadded IDs. It is also incompatible with W3C Trace Context and Zipkin B3 format, both of which include all leading 0s, so an application instrumented with OpenTelemetry SDKs may be logging different trace ID strings than application instrumented with Jaeger SDKs (related issue #1657).

Open question: this change may prevent ES backend from reading the old traces if their IDs had leading zeroes, because previously the zeroes were not added to string representation, and now they are. Roughly one in 16 traces will be affected.


yurishkuro:
@pavolloffay please review, and any thoughts on the ES question?  One option is to add a backwards compatibility mode by doing a fallback search with unpadded trace ID strings if full-length strings are not found. It would be an option people can configure.

yurishkuro:
cc @jpkrohling (I think Pavol is on PTO)

pavolloffay:
I will have a look at this one, hod for the moment

pavolloffay:
> Open question: this change may prevent ES backend from reading the old traces if their IDs had leading zeroes, because previously the zeroes were not added to string representation, and now they are. Roughly one in 16 traces will be affected.

Indeed, spans previously stored with with id like `1` are not searchable after this change. We definitely need a way to be backward compatible.

pavolloffay:
Something like this should work https://github.com/yurishkuro/jaeger/pull/1/files it needs some tests though.

@yurishkuro could you please summarize why adding leading zeros is necessary?

yurishkuro:
@pavolloffay I added motivation to description.

pavolloffay:
thanks @yurishkuro. It makes sense. I have updated my PR to your branch.

plugin/storage/es/spanstore/reader.go
```diff
@@ -393,15 +393,37 @@ func (s *SpanReader) multiRead(ctx context.Context, traceIDs []model.TraceID, st
...
+	}
+	return elastic.NewBoolQuery().Should(
+		elastic.NewTermQuery(traceIDField, traceIDStr).Boost(2),
+		elastic.NewTermQuery(traceIDField, legacyTraceID))
+}
+
 func convertTraceIDsStringsToModels(traceIDs []string) ([]model.TraceID, error) {
-	traceIDsModels := make([]model.TraceID, len(traceIDs))
-	for i, ID := range traceIDs {
+	traceIDsMap := map[model.TraceID]bool{}
```
pavolloffay: I forgot to add a TODO comment here similar to one in `buildTraceByIDQuery`

yurishkuro: I'll add it. Please approve if you're otherwise ok with the change.

pavolloffay:
We should list this in breaking changes as it might impact integrations

PR #1963, opened 2019-12-04
Change default rollover conditions to 2 days

pavolloffay:
Change default rollover conditions from 7 days to 2 days.

Given that by default Jaeger uses daily indices and some organizations do not keep data longer than 7 days the default of 7 days seems unreasonable - it might result in a too big index and running curator would immediatelly remove the old index.

 We should be more careful and change the default to 2 days - double the default daily indices. Is anybody from @jaegertracing/elasticsearch running rollover? Can anybody share the experience with this?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ Changes by Version
 
 #### Breaking Changes
 
-* List of service operations can be classified by span kinds ([#1943](https://github.com/jaegertracing/jaeger/pull/1943), [@guo0693](https://github.com/guo0693))
+##### List of service operations can be classified by span kinds ([#1943](https://github.com/jaegertracing/jaeger/pull/1943), [@guo0693](https://github.com/guo0693))
 
     * Endpoint changes:
         * Both Http & gRPC servers now take new optional parameter `spanKind` in addition to `service`. When spanKind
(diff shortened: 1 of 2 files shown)


yurishkuro:
before merging please add a note to CHANGELOG

pavolloffay:
Changelog added. I have reformatted changelog for 1.16 to match older version style.

yurishkuro:
I intentionally modified the formatting for 1.16 to make it easier to read, as well as to fix the incorrect L+1 header levels. I think using the new format is better going forward, and there's no particular reason why it has to match exactly previous versions.

pavolloffay:
I have changed back the heading sizes to bigger ones as you proposed.

yurishkuro:
why not leave it as is?  Each breaking change has its own section with a proper header, makes it easier to read than a long bullet list. Bullet lists are used for short collection of items, not for page+ long descriptions with examples.

pavolloffay:
> Each breaking change has its own section with a proper header, makes it easier to read than a long bullet list.

I think it looks worse and also it mixes styles - some sections are using a list and other header. However headers allow us to reference specific sections which is useful for breaking changes. 

I have changed  breaking changes to headers - however keep the same header structure as for list and include it under backed changes.

PR #1964, opened 2019-12-05
Make rollover init step idempotent

pavolloffay:
Resolves https://github.com/jaegertracing/jaeger-operator/issues/757

### DIFF
--- a/plugin/storage/es/esRollover.py
+++ b/plugin/storage/es/esRollover.py
@@ -108,14 +108,8 @@ def create_index(client, name):
     Create archive index
     """
     print('Creating index {}'.format(name))
-    create = curator.CreateIndex(client=client, name=name)
-    try:
-        create.do_action()
-    except curator.exceptions.FailedExecution as e:
-        if ("index_already_exists_exception" or "resource_already_exists_exception") in str(e):
-            raise e
-        else:
-            print("Index {} already exists".format(name))
+    create = curator.CreateIndex(client=client, name=name, ignore_existing=True)
+    create.do_action()
 
 
 def create_aliases(client, alias_name, archive_index_name):
--- a/plugin/storage/es/esRollover.py
+++ b/plugin/storage/es/esRollover.py
@@ -69,6 +69,10 @@ def main():
 
 def perform_action(action, client, write_alias, read_alias, index_to_rollover, template_name):
     if action == 'init':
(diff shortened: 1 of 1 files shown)


plugin/storage/es/esRollover.py
```diff
@@ -69,6 +69,10 @@ def main():
 
 def perform_action(action, client, write_alias, read_alias, index_to_rollover, template_name):
     if action == 'init':
+        if not is_alias_empty(client, read_alias) or not is_alias_empty(client, write_alias):
+            print("Alias {} or {} is not empty aborting init step".format(read_alias, write_alias))
```
yurishkuro: Shouldn't instead the creation of the index be made conditional? If this script fails after creating a read alias but before write alias there's no way to fix the situation by rerunning it.

pavolloffay: +1. I think that was the initial goal. I have reverted this and fixed catching the exception in the create index func

plugin/storage/es/esRollover.py
```diff
@@ -112,8 +112,10 @@ def create_index(client, name):
     try:
         create.do_action()
     except curator.exceptions.FailedExecution as e:
-        if "index_already_exists_exception" not in str(e) and "resource_already_exists_exception" not in str(e):
+        if ("index_already_exists_exception" or "resource_already_exists_exception") in str(e):
```
yurishkuro: Doesn't OR of two strings simply return the first one?

plugin/storage/es/esRollover.py
```diff
@@ -112,8 +112,10 @@ def create_index(client, name):
...
             raise e
+        else:
+            print("Index {} already exists".format(name))
```
yurishkuro: Aside from a log entry, how is it functionally different from before, when the exception was simply ignored?

pavolloffay: This just logs that the index was not created because it already exists

plugin/storage/es/esRollover.py
pavolloffay: ohh this would not work. My python skills... 

Actually `str(e)` never returns this string. There is a parameter `ignore_existing=True` I would use that instead. If the index exists it logs it.

Index jaeger-span-000001 already exists.

pavolloffay:
@yurishkuro could you please re-review?

PR #1969, opened 2019-12-09
Use a single index with wildcard in Elasticsearch reader

### LINKED ISSUE #1361: Use NumericRangeQuery in  ES queries when rollover is used
Then rollover is used the reader reads from a single read alias pointing to many indices. To support `s.max-span-age`  our `esRollover.py` supports an action to remove old indices from read alias. Instead of requiring to run the script to remove indices from read alias we could change our queries to support time ranges.

See https://github.com/jaegertracing/jaeger/issues/1242#issuecomment-455280650

pavolloffay:
Resolves #1361 

Use a wildcard (`jaeger-span-*`) in queries instead of the concrete list of indices (`jaeger-span-2018-12-24`, `jaeger-span-2018-12-25`...).

The motivation for this change:
* simplified index management and usage - no `--es.max-span-age` parameter
* search screen and trace detail screen will show the same results.
* align with rollover (a single read index) and later deprecate `rollover lookback` option (similar to `--es-max-span-age`) https://medium.com/jaegertracing/using-elasticsearch-rollover-to-manage-indices-8b3d0c77915d

Possible negative impacts:
* Performance - although kibana also uses wildcard pattern therefore time range queries (on span timestamp) should be well optimized.

Docker images to test:
```
pavolloffay/jaeger-query:es-wildcard
pavolloffay/jaeger-collector:es-wildcard
```

* `GetTrace`, `GetServices` and `GetOperatios` used `--es.max-span-age` to compose a list of historical indices for the query - now query all indices via the wildcard
* `FidTraceIDs` and `FindTraces` use times from query parameters to compose the list of historical indices - now query all indices but they use timestamp range query.

### DIFF
--- a/plugin/storage/es/options.go
+++ b/plugin/storage/es/options.go
@@ -86,7 +86,6 @@ func NewOptions(primaryNamespace string, otherNamespaces ...string) *Options {
 				Username:             "",
 				Password:             "",
 				Sniffer:              false,
-				MaxSpanAge:           72 * time.Hour,
 				MaxNumSpans:          10000,
 				NumShards:            5,
 				NumReplicas:          1,
@@ -147,8 +146,8 @@ func addFlags(flagSet *flag.FlagSet, nsConfig *namespaceConfig) {
 		"Timeout used for queries. A Timeout of zero means no timeout")
 	flagSet.Duration(
 		nsConfig.namespace+suffixMaxSpanAge,
-		nsConfig.MaxSpanAge,
-		"The maximum lookback for spans in Elasticsearch")
+		time.Hour*72,
+		"(deprecated) The maximum lookback for spans in Elasticsearch. Now all indices are searched.")
 	flagSet.Int(
 		nsConfig.namespace+suffixMaxNumSpans,
 		nsConfig.MaxNumSpans,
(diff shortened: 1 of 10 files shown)


pavolloffay:
@jaegertracing/elasticsearch we are looking for volunteers to test this functionality especially if the query performance changed. Any feedback is appreciated. The docker images are listed in the PR description.

plugin/storage/es/options.go
```diff
@@ -147,8 +146,8 @@ func addFlags(flagSet *flag.FlagSet, nsConfig *namespaceConfig) {
 		"Timeout used for queries. A Timeout of zero means no timeout")
 	flagSet.Duration(
 		nsConfig.namespace+suffixMaxSpanAge,
-		nsConfig.MaxSpanAge,
-		"The maximum lookback for spans in Elasticsearch")
+		time.Hour*72,
+		"(deprecated) The maximum lookback for spans in Elasticsearch. Now all indices are searched.")
```
annanay25: This seems confusing. Can we add the message `Now all indices are searched` in the parenthesis?

pavolloffay:
> Should we also change the esCleaner.py script as part of this PR and allow users to specify a max TTL for span data?

I am not sure if I follow you. The purpose of `esCleaner.py` script is to clean old data. Elasticsearch does not allow to specify TTL in any form.

annanay25:
@pavolloffay Sorry I didn't frame that properly. 

> Elasticsearch does not allow to specify TTL in any form.

Yes. What I meant was that should we modify the `esCleaner.py` script as part of this PR as well, to include the change made in index naming format, so that users can continue to specify how long they want to store data.

pavolloffay:
There is no change in index name format in this PR. The writer still writes to daily indices.

pavolloffay:
@jkandasa run performance tests to measure query time with this patch. The results prove increased query time by on average 300-400%. Therefore I am closing this PR. 

https://docs.google.com/spreadsheets/d/1rwTzDvnHHhssNdZibhFuHBZDf6njt6b1DrvRiJrkt14/edit?usp=sharing

PR #1970, opened 2019-12-09
Do not truncate tags in Elasticsearch

### LINKED ISSUE #1960: HTTP Error: Converting JSONSpan to domain Span failed
Hello,
Sometimes on some span the ui is replying with a status 500 error

Status | 500
-- | --
Status text | Internal Server Error
URL | /api/traces
Query | end=1575368774721000&limit=20&lookback=1h&maxDuration&minDuration&service=**********&start=1575365174721000
Response body | {   "data": null,   "total": 0,   "limit": 0,   "offset": 0,   "errors": [     {       "code": 500,       "msg": "Converting JSONSpan to domain Span failed: encoding/hex: invalid byte: U+002E '.'"     }   ] }

Screenshot from 2019-12-03 11-41-15

For info we are using elk to store span data

pavolloffay:
Resolves #1960 

`KV.AsString()` truncates binary tag value if it is longer than 256 and adds `...` at the end. This causes failures when reading and decoding the tag value.


plugin/storage/es/spanstore/dbmodel/from_domain_test.go
```diff
@@ -106,3 +107,51 @@ func TestTagMap(t *testing.T) {
 	assert.Equal(t, tagsMap, dbSpan.Tag)
 	assert.Equal(t, tagsMap, dbSpan.Process.Tag)
 }
+
+func TestConvertKeyValueValue(t *testing.T) {
+	longString := `Bender Bending Rodrigues Bender Bending Rodrigues Bender Bending Rodrigues Bender Bending Rodrigues
+	Bender Bending Rodrigues Bender Bending Rodrigues Bender Bending Rodrigues Bender Bending Rodrigues Bender Bending Rodrigues
+	Bender Bending Rodrigues Bender Bending Rodrigues Bender Bending Rodrigues Bender Bending Rodrigues Bender Bending Rodrigues
+	Bender Bending Rodrigues Bender Bending Rodrigues Bender Bending Rodrigues Bender Bending Rodrigues Bender Bending Rodrigues
+	Bender Bending Rodrigues Bender Bending Rodrigues Bender Bending Rodrigues Bender Bending Rodrigues Bender Bending Rodrigues `
```
yurishkuro: this was supposed to be "Bender is Great!" string repeated :-)

plugin/storage/es/spanstore/dbmodel/from_domain_test.go
```diff
@@ -106,3 +107,51 @@ func TestTagMap(t *testing.T) {
...
+	tests := []struct {
+		kv       model.KeyValue
+		expected string
+	}{
+		{
+			kv:       model.KeyValue{VType: model.ValueType_BOOL, VBool: true},
```
yurishkuro: nit: can we use factory methods instead of explicit struct assignments?

yurishkuro:
could you elaborate in the description why this solves the truncation problem? Does `KV.AsString()` truncate values?

yurishkuro:
> Does `KV.AsString()` truncate values?

Should we add, instead, a function there that doesn't truncate?

pavolloffay:
> Should we add, instead, a function there that doesn't truncate?

It is a good idea, I will rename the current func to `AsStringLossy` and add `AsString` which does not truncate.

model/keyvalue.go
```diff
@@ -144,6 +144,27 @@ func (kv *KeyValue) AsString() string {
 	}
 }
 
+// AsString returns a string representation of the value.
+func (kv *KeyValue) AsString() string {
```
yurishkuro: the two functions are nearly identical, can we have them both call into a private helper with a `truncate bool` parameter?

cmd/collector/app/span_processor_test.go
```diff
@@ -337,7 +337,7 @@ func TestSpanProcessorWithCollectorTags(t *testing.T) {
 		var foundTag bool
 		for _, tag := range span.Process.Tags {
 			if tag.GetKey() == k {
-				assert.Equal(t, v, tag.AsString())
+				assert.Equal(t, v, tag.AsStringLossy())
```
yurishkuro:
```suggestion
				assert.Equal(t, v, tag.AsString())
```

model/converter/json/from_domain.go
```diff
@@ -143,7 +143,7 @@ func (fd fromDomain) convertKeyValuesString(keyValues model.KeyValues) []json.Ke
 		out[i] = json.KeyValue{
 			Key:   kv.Key,
 			Type:  json.ValueType(strings.ToLower(kv.VType.String())),
-			Value: kv.AsString(),
+			Value: kv.AsStringLossy(),
```
yurishkuro:
```suggestion
			Value: kv.AsString(),
```

model/keyvalue_test.go
```diff
@@ -130,13 +130,13 @@ func TestKeyValueAsStringAndValue(t *testing.T) {
 	for _, tt := range testCases {
 		testCase := tt // capture loop var
 		t.Run(testCase.str, func(t *testing.T) {
-			assert.Equal(t, testCase.str, testCase.kv.AsString())
+			assert.Equal(t, testCase.str, testCase.kv.AsStringLossy())
```
yurishkuro:
```suggestion
			assert.Equal(t, testCase.str, testCase.kv.AsString())
```

model/keyvalue_test.go
```diff
@@ -130,13 +130,13 @@ func TestKeyValueAsStringAndValue(t *testing.T) {
...
 			assert.Equal(t, testCase.val, testCase.kv.Value())
 		})
 	}
 	t.Run("invalid type", func(t *testing.T) {
 		kv := model.KeyValue{Key: "x", VType: model.ValueType(-1)}
-		assert.Equal(t, "unknown type -1", kv.AsString())
+		assert.Equal(t, "unknown type -1", kv.AsStringLossy())
```
yurishkuro:
```suggestion
		assert.Equal(t, "unknown type -1", kv.AsString())
```

model/span.go
```diff
@@ -48,15 +48,15 @@ func (s *Span) Hash(w io.Writer) (err error) {
 // HasSpanKind returns true if the span has a `span.kind` tag set to `kind`.
 func (s *Span) HasSpanKind(kind ext.SpanKindEnum) bool {
 	if tag, ok := KeyValues(s.Tags).FindByKey(string(ext.SpanKind)); ok {
-		return tag.AsString() == string(kind)
+		return tag.AsStringLossy() == string(kind)
```
yurishkuro:
```suggestion
		return tag.AsString() == string(kind)
```

model/span.go
```diff
@@ -48,15 +48,15 @@ func (s *Span) Hash(w io.Writer) (err error) {
...
 	}
 	return false
 }
 
 // GetSpanKind returns value of `span.kind` tag and whether the tag can be found
 func (s *Span) GetSpanKind() (spanKind string, found bool) {
 	if tag, ok := KeyValues(s.Tags).FindByKey(string(ext.SpanKind)); ok {
-		return tag.AsString(), true
+		return tag.AsStringLossy(), true
```
yurishkuro:
```suggestion
		return tag.AsString(), true
```

plugin/storage/badger/spanstore/writer.go
```diff
@@ -91,16 +91,16 @@ func (w *SpanWriter) WriteSpan(span *model.Span) error {
 	for _, kv := range span.Tags {
 		// Convert everything to string since queries are done that way also
```
yurishkuro: "lossy" is not required in this file

plugin/storage/cassandra/spanstore/dbmodel/unique_tags.go
```diff
@@ -38,7 +38,7 @@ func GetAllUniqueTags(span *model.Span, tagFilter TagFilter) []TagInsertion {
 		uniqueTags = append(uniqueTags, TagInsertion{
 			ServiceName: span.Process.ServiceName,
 			TagKey:      allTags[i].Key,
-			TagValue:    allTags[i].AsString(),
+			TagValue:    allTags[i].AsStringLossy(),
```
yurishkuro:
```suggestion
			TagValue:    allTags[i].AsString(),
```

plugin/storage/memory/memory.go
```diff
@@ -243,7 +243,7 @@ func (m *Store) validTrace(trace *model.Trace, query *spanstore.TraceQueryParame
 
 func findKeyValueMatch(kvs model.KeyValues, key, value string) (model.KeyValue, bool) {
 	for _, kv := range kvs {
-		if kv.Key == key && kv.AsString() == value {
+		if kv.Key == key && kv.AsStringLossy() == value {
```
yurishkuro:
```suggestion
		if kv.Key == key && kv.AsString() == value {
```

model/keyvalue_test.go
pavolloffay: This would not pass, the expected string is trucated

PR #1977, opened 2019-12-13
Added configuration for stale bot and needs-triage

jpkrohling:
- Adds the bots discussed during today's bi-weekly meeting

- Adds the configuration to both the stale bot and the "needs-triage" dev bot

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,4 @@
+daysUntilStale: 60
+daysUntilClose: 7
+onlyLabels:
+  - needs-info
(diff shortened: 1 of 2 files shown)


.github/stale.yml
```diff
@@ -0,0 +1,4 @@
+daysUntilStale: 60
+daysUntilClose: 7
+onlyLabels:
```
yurishkuro: what does onlyLabels do? can we add a comment?

jpkrohling: At the beginning, we are only checking the issues that containing this label.

Issue #1973, opened 2019-12-10
Travis CI arm64 support


jishminor:
## Requirement - what kind of business use case are you trying to solve?

Run and test all Jaeger components on arm64.

## Problem - what in Jaeger blocks you from solving the requirement?

No existing infrastructure to build and test Jaeger components on arm64.

## Proposal - what do you suggest to solve the problem or improve the existing situation?

Travis CI can now run multiarch builds out of the box: https://docs.travis-ci.com/user/multi-cpu-architectures/.

## Any open questions to address

Would there be any interest in modifying the CI configuration to support arm64 builds?

yurishkuro:
I'd be concerned with how much extra load it would add to our CI jobs, which already take quite a while to run for each PR.

jpkrohling:
I think we could consider adding it at least for the unit tests. I think it would cover most of the arch-related issues that we might face.

jishminor:
Pushing the arm64 docker images to dockerhub would also be a great addition.

jpkrohling:
@jishminor are you able to commit to support this when it needs maintenance? If so, send a PR with your proposal and let's see how the CI gets affected.

jishminor:
I have only so far been using an arm64 jaeger agent, which speaks to a x86 jaeger collector. I still need to fully familiarize myself with your build and test infrastructure before I can give my commitment for maintenance support for arm64. Let me do some experimenting and get back to you. I don't for-see many things breaking when swapping out the x86 go runtime for the arm64, besides any OS level package use (syscall interfaces etc).

jpkrohling:
Let us know if you hit any roadblocks!

vielmetti:
@jishminor have you had any success here? If you need arm64 infrastructure to test or develop, reach out to the @worksonarm project via opening an issue at https://github.com/worksonarm/cluster/issues .

jishminor:
@vielmetti I have not yet. I have my own repository for the Jaeger Agent on arm64 and arm, but have not gotten around to playing with the build infrastructure in this repo.

MrXinWang:
Hi @jishminor! I am wondering will you propose a PR to add support for arm64? If not, I have a PR to support it :) Can I propose it? or I wait for yours?

yurishkuro:
ARM64 binaries are already supported (#1656) as part of the release (but not the Docker images).

We don't have the CI resources to do integration testing on all supported platforms. I am closing this.

PR #1982, opened 2019-12-18
Build binaries for Linux on IBM Z / s390x architecture

prankkelkar:
Signed-off-by: Prasanna Kelkar <Prasanna.Kelkar@ibm.com>

-  PR not intended to resolve specific issue but As Travis CI [officially supports](https://blog.travis-ci.com/2019-11-12-multi-cpu-architecture-ibm-power-ibm-z) s390x builds, adding support for same.

- Added s390x specific changes to skip usage of `RACE` flag as it is not supported on s390x architecture.
- Added type casting in `plugin/storage/badger/stats_linux.go` file to resolve `mismatched types int64 and uint32` issue on s390x platform.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,21 +7,13 @@ dist: trusty
 
 matrix:
   include:
-  - arch: s390x
-    go: 1.13.x
+  - go: "1.13.x"
     env:
     - TESTS=true
     - COVERAGE=true
-    addons:
-      apt:
-        packages:
-        - jq
-        - time
-  - arch: amd64
-    go: 1.13.x
+  - go: "1.13.x"
     env:
-    - TESTS=true
-    - COVERAGE=true  
+    - PROTO_GEN_TEST=true
   - go: "1.13.x"
     env:
     - ALL_IN_ONE=true
@@ -62,7 +54,6 @@ env:
 
 install:
   - docker rmi $(docker images -q) || true
-  - if [ "$HOSTTYPE" == "s390x" ]; then mkdir -p $HOME/gopath/bin; fi
   - travis_retry make install-ci
   - if [ "$ALL_IN_ONE" == true ]; then bash ./scripts/travis/install-ui-deps.sh ; fi
   - if [ "$DOCKER" == true ]; then bash ./scripts/travis/install-ui-deps.sh ; fi
@@ -70,8 +61,8 @@ install:
   - |
     if [ "$PROTO_GEN_TEST" == true ]; then
         make proto-install
-        curl -L https://github.com/google/protobuf/releases/download/v3.6.1/protoc-3.6.1-linux-x86_64.zip -o /tmp/protoc.zip;
-        unzip /tmp/protoc.zip -d "$HOME"/protoc   
+        curl -L https://github.com/google/protobuf/releases/download/v3.6.1/protoc-3.6.1-linux-x86_64.zip -o /tmp/protoc.zip
+        unzip /tmp/protoc.zip -d "$HOME"/protoc
     fi
 
 script:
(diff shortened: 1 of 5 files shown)


jpkrohling:
Please, see #1973. There are two main problems:

1. the extra time required to perform a build. Instead of running the whole build in a new arch, only the fast unit-tests should be executed
1. what happens if there's a new problem in the future? Are you able to commit to support this when it needs maintenance? The current maintainers do not have the necessary bandwidth to support arches that aren't used by ourselves...

prankkelkar:
> Please, see #1973. There are two main problems:
> 
> 1. the extra time required to perform a build. Instead of running the whole build in a new arch, only the fast unit-tests should be executed
> 2. what happens if there's a new problem in the future? Are you able to commit to support this when it needs maintenance? The current maintainers do not have the necessary bandwidth to support arches that aren't used by ourselves...

1. I have modified the changes as per you instructions to run only unit tests for s390x architecture.
2. We already build and maintain jaeger on s390x. Our build instructions are published [here](https://github.com/linux-on-ibm-z/docs/wiki/Building-Jaeger)

yurishkuro:
We're currently not running any tests on Windows or MacOS platforms, we only build the binaries for them. Could we do the same for this arch? This would have minimal impact on the CI time.

prankkelkar:
Yes. It will be great if we can have binaries released for s390x architecture which could be downloaded directly. I believe changes in Makefile to add target for s390x in `build-all-platforms` should be enough. Please let me know if any more changes are needed.

> We're currently not running any tests on Windows or MacOS platforms, we only build the binaries for them. Could we do the same for this arch? This would have minimal impact on the CI time.

prankkelkar:
@yurishkuro Could you please guide me with the changes which are needed to release binaries for s390x ?

yurishkuro:
Follow the path of `make build-all-platforms` in the Makefile.

prankkelkar:
@yurishkuro While i was adding linux s390x support i came across the names of the binaries used. for eg. `collector/collector-$(GOOS)`
In the case of s390x as the os name is same i.e Linux, two binaries will be created with the same name as per current code. To resolve this issue I was planning to add $(GOARCH) only to s390x binary name to differentiate it from amd64 linux.
Is there any other naming conventions we have to follow?

yurishkuro:
Could we manage that at the directory name level, and keep the file names as is?

We may also consider not using OS in the binary name, only in the directory.

cc @jpkrohling @pavolloffay @objectiser

jpkrohling:
Minikube adds both to the binaries:

    minikube-linux-amd64
    minikube-windows-amd64.exe
    minikube-darwin-amd64

In our case, it would probably be too noisy, so, the idea of grouping them by OS makes sense.

prankkelkar:
I have currently added following code to handle this scenario
```
.PHONY: build-binaries-s390x
build-binaries-s390x:
	GOOS=linux GOARCH=s390x $(MAKE) build-platform-binaries

.PHONY: build-agent
build-agent:
ifeq ($(GOARCH), s390x)
	CGO_ENABLED=0 installsuffix=cgo go build -o ./cmd/agent/agent-$(GOOS)-$(GOARCH) $(BUILD_INFO) ./cmd/agent/main.go
else
	CGO_ENABLED=0 installsuffix=cgo go build -o ./cmd/agent/agent-$(GOOS) $(BUILD_INFO) ./cmd/agent/main.go
endif
```
This will only change the name of the binary of s390x and rest will remain as it is.
I have also tweaked `./scripts/travis/package-deploy.sh` to handle this change. Should i go ahead to update PR with this approach?

yurishkuro:
sgtm

prankkelkar:
@yurishkuro I have updated the PR to cross compile and release the binaries. Please review the code and let me know if anything needs to be changed.

PR #1983, opened 2019-12-18
Configure Jager Collector to work with Kafka With Basic Authenticatio…

### LINKED ISSUE #1966: Configure Jager Collector to work with Kafka With Basic Authentication
## Requirement - 
Looking to integrate with Kafka which is provided by a platform Service. However, the kafka service provided by the platform only supports basic authentication

## Problem - 
The jager documentation or options did not indicate support for Kafka basic authentication

As a result, we would not be able to integrate our Jaeger solution with existing Kafka Services provided with our platform
## Proposal -
Jaeger to support authentication mechanisms provided by Kafka

Is it available? Is it documentation that I possibly missed? Or is there a reason why it was not implemented?
(Maybe some best practise that I am unaware of)

chandresh-pancholi:
Configure Jager Collector to work with Kafka With Basic Authentication

Resolves #1966

Signed-off-by: chandresh-pancholi <chandreshpancholi007@gmail.com>

- Configure Jager Collector to work with Kafka With Basic Authentication

-  Adding plaintext Kafka basic authentication


pkg/kafka/auth/options.go
```diff
@@ -99,6 +106,17 @@ func addTLSFlags(configPrefix string, flagSet *flag.FlagSet) {
 		"Path to the TLS Key for the Kafka connection")
 }
 
+func addPlainTextFlag(configPrefix string, flagSet *flag.FlagSet) {
+	flagSet.String(
+		configPrefix+plainTextPrefix+suffixPlainTextUserName,
+		defaultPlainTextUserName,
+		"The plaintext Username for SASL/PLAIN authentication")
```
pavolloffay: shouldn't we call it basic like in other flags?

chandresh-pancholi: Kafka and Sarama always mentioned it as Plaintext. Shall I change it to basic?
I can replace "plainText" to "basicAuth"

chandresh-pancholi:
> Hi Chandresh-Pancholi! I was very happy when I saw your pull request regarding kafka basic auth. I want to connect a Kafka enabled Azure Event Hub to jaeger-ingest and jaeger-collector. Unfortunatly there is currently no support for e.g. these properties:
> 
> sasl.mechanism=PLAIN
> security.protocol=SASL_SSL
> sasl.jaas.config=org.apache.kafka.common.security.plain.PlainLoginModule required
> username="$ConnectionString"
> password="{MY.EVENTHUBS.CONNECTION.STRING}";
> 
> Examples for Azure Event Hubs for Kafka connections: https://github.com/Azure/azure-event-hubs-for-kafka/tree/master/tutorials
> 
> It would be nice if you could add that to your PR too cause this would enable a lot of people that are using Azure EH (kafka enabled) to connect jaeger to their solutions. Does it make sense for you?

I never used Azure Event bus. Let me spend some time over the weekend. It would be great if you can create another issue specific to Azure Event bus. Let these changes go live for now.

awesomeinsight:
> > Hi Chandresh-Pancholi! I was very happy when I saw your pull request regarding kafka basic auth. I want to connect a Kafka enabled Azure Event Hub to jaeger-ingest and jaeger-collector. Unfortunatly there is currently no support for e.g. these properties:
> > sasl.mechanism=PLAIN
> > security.protocol=SASL_SSL
> > sasl.jaas.config=org.apache.kafka.common.security.plain.PlainLoginModule required
> > username="$ConnectionString"
> > password="{MY.EVENTHUBS.CONNECTION.STRING}";
> > Examples for Azure Event Hubs for Kafka connections: https://github.com/Azure/azure-event-hubs-for-kafka/tree/master/tutorials
> > It would be nice if you could add that to your PR too cause this would enable a lot of people that are using Azure EH (kafka enabled) to connect jaeger to their solutions. Does it make sense for you?
> 
> I never used Azure Event bus. Let me spend some time over the weekend. It would be great if you can create another issue specific to Azure Event bus. Let these changes go live for now.

Hi chandresh-pancholi! Thanks for your quick response. As requested I created a separate issue for it: https://github.com/jaegertracing/jaeger/issues/1989. I also referenced this one here. Regards and Merry X-Mas :-)

pkg/kafka/auth/options.go
yurishkuro: I'd keep it as plaintext to match Kafka documentation.

pkg/kafka/auth/options.go
```diff
@@ -99,6 +106,17 @@ func addTLSFlags(configPrefix string, flagSet *flag.FlagSet) {
 		"Path to the TLS Key for the Kafka connection")
 }
 
+func addPlainTextFlag(configPrefix string, flagSet *flag.FlagSet) {
```
yurishkuro:
```suggestion
func addPlainTextFlags(configPrefix string, flagSet *flag.FlagSet) {
```

pkg/kafka/auth/plaintext.go
```diff
@@ -0,0 +1,31 @@
...
+// PlainTextConfig describes the configuration properties needed for SASL/PLAIN with kafka
+type PlainTextConfig struct {
+	UserName string
+	Password string
+}
+
+func setPlainTextConfiguration(config *PlainTextConfig, saramaConfig *sarama.Config) {
+	saramaConfig.Net.SASL.Enable = true
+	saramaConfig.Net.SASL.User = config.UserName
+	saramaConfig.Net.SASL.Password = config.Password
```
yurishkuro: this doesn't track. Kafka docs https://docs.confluent.io/current/kafka/authentication_sasl/authentication_sasl_plain.html mention "plaintext" as an alternative to SASL "plain", but here we seem to make them the same.

chandresh-pancholi: Please correct me if I understood it, You are trying to say rather than using "PlainText" we shall use "Plain".

yurishkuro: no, I am saying Kafka docs make clear distinction between those two methods, but in this code we seem to treat them as the same thing: calling it plaintext but using SASL structure to pass it. So which one is it?

chandresh-pancholi: saramaconfig only has SASL struct. below is the code snippet.

```
// SASL based authentication with broker. While there are multiple SASL authentication methods
		// the current implementation is limited to plaintext (SASL/PLAIN) authentication
		SASL struct {
			// Whether or not to use SASL authentication when connecting to the broker
			// (defaults to false).
			Enable bool
			// SASLMechanism is the name of the enabled SASL mechanism.
			// Possible values: OAUTHBEARER, PLAIN (defaults to PLAIN).
			Mechanism SASLMechanism
			// Whether or not to send the Kafka SASL handshake first if enabled
			// (defaults to true). You should only set this to false if you're using
			// a non-Kafka SASL proxy.
			Handshake bool
			//username and password for SASL/PLAIN  or SASL/SCRAM authentication
			User     string
			Password string
```

pkg/kafka/auth/options.go
chandresh-pancholi: fixed.

pkg/kafka/auth/plaintext.go
yurishkuro: well, then this is "plain", not "plaintext", because credentials are still sent over encrypted channel. From the docs:

> `PLAIN` versus `PLAINTEXT` Do not confuse the _SASL mechanism PLAIN_ with _no SSL encryption being called PLAINTEXT_.

chandresh-pancholi: From Confluent slack channel:
`None of the SASL mechanisms encrypt wire data not even GSSAPI. You have to enable TLS for that.`

chandresh-pancholi: https://docs.confluent.io/current/kafka/authentication_ssl.html
> By default, Apache Kafka® communicates in PLAINTEXT, which means that all data is sent in the clear. To encrypt communication, it is recommended to configure all the Confluent Platform components in your deployment to use SSL encryption.

PR #69, opened 2017-03-29
Add standalone Dockerfile that runs agent, collector, and query

badiib:
- UI is not part of this docker instance

PR #1985, opened 2019-12-19
Added support for dynamic queue sizes

### LINKED ISSUE #943: Memory-sensitive collector queue size

jpkrohling:
- Closes #943

* Introduces the new flags: `collector.dyn-queue-size-warmup` and `collector.dyn-queue-size-memory`.
* When the warm-up flag is set, the span processor starts a new background process to recalculate the queue capacity at regular intervals.
* The span processor now keeps track of the number of spans it has seen, as well as the total bytes it processed. That information allows to derive the average spans size that the collector has processed. Along with the memory to allocate, we can come up with the ideal queue size.
* New metrics: `jaeger_collector_queue_capacity` and `jaeger_collector_spans_bytes`.

This commit shouldn't cause performance degradation for instances not using the new options, except for a small memory overhead due to the new span tracker (quantity/size) and the new metrics.

### DIFF
--- a/cmd/collector/app/builder/builder_flags.go
+++ b/cmd/collector/app/builder/builder_flags.go
@@ -27,6 +27,7 @@ import (
 )
 
 const (
+	collectorDynQueueSizeMemory   = "collector.queue-size-memory"
 	collectorQueueSize            = "collector.queue-size"
 	collectorNumWorkers           = "collector.num-workers"
 	collectorPort                 = "collector.port"
@@ -46,6 +47,8 @@ var tlsFlagsConfig = tlscfg.ServerFlagsConfig{
 
 // CollectorOptions holds configuration for collector
 type CollectorOptions struct {
+	// DynQueueSizeMemory determines how much memory to use for the queue
+	DynQueueSizeMemory uint
 	// QueueSize is the size of collector's queue
 	QueueSize int
 	// NumWorkers is the number of internal workers in a collector
@@ -70,6 +73,7 @@ type CollectorOptions struct {
 
 // AddFlags adds flags for CollectorOptions
 func AddFlags(flags *flag.FlagSet) {
+	flags.Uint(collectorDynQueueSizeMemory, 0, "(experimental) The max memory size in MiB to use for the dynamic queue.")
 	flags.Int(collectorQueueSize, app.DefaultQueueSize, "The queue size of the collector")
 	flags.Int(collectorNumWorkers, app.DefaultNumWorkers, "The number of workers pulling items from the queue")
 	flags.Int(collectorPort, ports.CollectorTChannel, "The TChannel port for the collector service")
@@ -84,6 +88,7 @@ func AddFlags(flags *flag.FlagSet) {
 
 // InitFromViper initializes CollectorOptions with properties from viper
 func (cOpts *CollectorOptions) InitFromViper(v *viper.Viper) *CollectorOptions {
+	cOpts.DynQueueSizeMemory = v.GetUint(collectorDynQueueSizeMemory) * 1024 * 1024 // we receive in MiB and store in bytes
 	cOpts.QueueSize = v.GetInt(collectorQueueSize)
 	cOpts.NumWorkers = v.GetInt(collectorNumWorkers)
 	cOpts.CollectorPort = v.GetInt(collectorPort)
--- a/cmd/collector/app/metrics.go
+++ b/cmd/collector/app/metrics.go
@@ -58,9 +58,13 @@ type SpanProcessorMetrics struct {
 	InQueueLatency metrics.Timer
 	// SpansDropped measures the number of spans we discarded because the queue was full
 	SpansDropped metrics.Counter
+	// SpansBytes records how many bytes were processed
+	SpansBytes metrics.Gauge
 	// BatchSize measures the span batch size
 	BatchSize metrics.Gauge // size of span batch
-	// QueueLength measures the size of the internal span queue
+	// QueueCapacity measures the capacity of the internal span queue
+	QueueCapacity metrics.Gauge
+	// QueueLength measures the current number of elements in the internal span queue
 	QueueLength metrics.Gauge
 	// SavedOkBySvc contains span and trace counts by service
 	SavedOkBySvc  metricsBySvc  // spans actually saved
@@ -150,7 +154,9 @@ func NewSpanProcessorMetrics(serviceMetrics metrics.Factory, hostMetrics metrics
 		InQueueLatency: hostMetrics.Timer(metrics.TimerOptions{Name: "in-queue-latency", Tags: nil}),
 		SpansDropped:   hostMetrics.Counter(metrics.Options{Name: "spans.dropped", Tags: nil}),
 		BatchSize:      hostMetrics.Gauge(metrics.Options{Name: "batch-size", Tags: nil}),
+		QueueCapacity:  hostMetrics.Gauge(metrics.Options{Name: "queue-capacity", Tags: nil}),
 		QueueLength:    hostMetrics.Gauge(metrics.Options{Name: "queue-length", Tags: nil}),
+		SpansBytes:     hostMetrics.Gauge(metrics.Options{Name: "spans.bytes", Tags: nil}),
 		SavedOkBySvc:   newMetricsBySvc(serviceMetrics.Namespace(metrics.NSOptions{Name: "", Tags: map[string]string{"result": "ok"}}), "saved-by-svc"),
 		SavedErrBySvc:  newMetricsBySvc(serviceMetrics.Namespace(metrics.NSOptions{Name: "", Tags: map[string]string{"result": "err"}}), "saved-by-svc"),
 		spanCounts:     spanCounts,
--- a/cmd/collector/app/options.go
+++ b/cmd/collector/app/options.go
@@ -31,19 +31,21 @@ const (
 )
 
 type options struct {
-	logger           *zap.Logger
-	serviceMetrics   metrics.Factory
-	hostMetrics      metrics.Factory
-	preProcessSpans  ProcessSpans
-	sanitizer        sanitizer.SanitizeSpan
-	preSave          ProcessSpan
-	spanFilter       FilterSpan
-	numWorkers       int
-	blockingSubmit   bool
-	queueSize        int
-	reportBusy       bool
-	extraFormatTypes []SpanFormat
-	collectorTags    map[string]string
+	logger             *zap.Logger
+	serviceMetrics     metrics.Factory
+	hostMetrics        metrics.Factory
+	preProcessSpans    ProcessSpans
+	sanitizer          sanitizer.SanitizeSpan
+	preSave            ProcessSpan
+	spanFilter         FilterSpan
+	numWorkers         int
+	blockingSubmit     bool
+	queueSize          int
+	dynQueueSizeWarmup uint
+	dynQueueSizeMemory uint
+	reportBusy         bool
+	extraFormatTypes   []SpanFormat
+	collectorTags      map[string]string
 }
 
 // Option is a function that sets some option on StorageBuilder.
@@ -119,6 +121,20 @@ func (options) BlockingSubmit(blockingSubmit bool) Option {
 func (options) QueueSize(queueSize int) Option {
 	return func(b *options) {
 		b.queueSize = queueSize
+	}
+}
+
+// DynQueueSize creates an Option that initializes the queue size
+func (options) DynQueueSizeWarmup(dynQueueSizeWarmup uint) Option {
+	return func(b *options) {
+		b.dynQueueSizeWarmup = dynQueueSizeWarmup
+	}
+}
+
+// DynQueueSize creates an Option that initializes the queue size
+func (options) DynQueueSizeMemory(dynQueueSizeMemory uint) Option {
+	return func(b *options) {
+		b.dynQueueSizeMemory = dynQueueSizeMemory
 	}
 }
 
--- a/cmd/collector/app/options_test.go
+++ b/cmd/collector/app/options_test.go
@@ -39,12 +39,16 @@ func TestAllOptionSet(t *testing.T) {
 		Options.PreProcessSpans(func(spans []*model.Span) {}),
 		Options.Sanitizer(func(span *model.Span) *model.Span { return span }),
 		Options.QueueSize(10),
+		Options.DynQueueSizeWarmup(1000),
+		Options.DynQueueSizeMemory(1024),
 		Options.PreSave(func(span *model.Span) {}),
 		Options.CollectorTags(map[string]string{"extra": "tags"}),
 	)
 	assert.EqualValues(t, 5, opts.numWorkers)
 	assert.EqualValues(t, 10, opts.queueSize)
 	assert.EqualValues(t, map[string]string{"extra": "tags"}, opts.collectorTags)
+	assert.EqualValues(t, 1000, opts.dynQueueSizeWarmup)
+	assert.EqualValues(t, 1024, opts.dynQueueSizeMemory)
 }
 
 func TestNoOptionsSet(t *testing.T) {
@@ -59,4 +63,5 @@ func TestNoOptionsSet(t *testing.T) {
 	assert.True(t, opts.spanFilter(nil))
 	span := model.Span{}
 	assert.EqualValues(t, &span, opts.sanitizer(&span))
+	assert.EqualValues(t, 0, opts.dynQueueSizeWarmup)
 }
--- a/cmd/collector/app/span_processor.go
+++ b/cmd/collector/app/span_processor.go
@@ -16,15 +16,25 @@
 package app
 
 import (
+	"sync"
 	"time"
 
 	tchannel "github.com/uber/tchannel-go"
+	"go.uber.org/atomic"
 	"go.uber.org/zap"
 
 	"github.com/jaegertracing/jaeger/cmd/collector/app/sanitizer"
 	"github.com/jaegertracing/jaeger/model"
 	"github.com/jaegertracing/jaeger/pkg/queue"
 	"github.com/jaegertracing/jaeger/storage/spanstore"
+)
+
+const (
+	// if this proves to be too low, we can increase it
+	maxQueueSize = 1_000_000
+
+	// if the new queue size isn't 20% bigger than the previous one, don't change
+	minRequiredChange = 1.2
 )
 
 // ProcessSpansOptions additional options passed to processor along with the spans.
@@ -40,17 +50,23 @@ type SpanProcessor interface {
 }
 
 type spanProcessor struct {
-	queue           *queue.BoundedQueue
-	metrics         *SpanProcessorMetrics
-	preProcessSpans ProcessSpans
-	filterSpan      FilterSpan             // filter is called before the sanitizer but after preProcessSpans
-	sanitizer       sanitizer.SanitizeSpan // sanitizer is called before processSpan
-	processSpan     ProcessSpan
-	logger          *zap.Logger
-	spanWriter      spanstore.Writer
-	reportBusy      bool
-	numWorkers      int
-	collectorTags   map[string]string
+	queue              *queue.BoundedQueue
+	queueResizeMu      sync.Mutex
+	metrics            *SpanProcessorMetrics
+	preProcessSpans    ProcessSpans
+	filterSpan         FilterSpan             // filter is called before the sanitizer but after preProcessSpans
+	sanitizer          sanitizer.SanitizeSpan // sanitizer is called before processSpan
+	processSpan        ProcessSpan
+	logger             *zap.Logger
+	spanWriter         spanstore.Writer
+	reportBusy         bool
+	numWorkers         int
+	collectorTags      map[string]string
+	dynQueueSizeWarmup uint
+	dynQueueSizeMemory uint
+	bytesProcessed     *atomic.Uint64
+	spansProcessed     *atomic.Uint64
+	stopCh             chan struct{}
 }
 
 type queueItem struct {
@@ -70,7 +86,11 @@ func NewSpanProcessor(
 		sp.processItemFromQueue(value)
 	})
 
-	sp.queue.StartLengthReporting(1*time.Second, sp.metrics.QueueLength)
+	sp.background(1*time.Second, sp.updateGauges)
+
+	if sp.dynQueueSizeWarmup > 0 {
+		sp.background(1*time.Minute, sp.updateQueueSize)
+	}
 
 	return sp
 }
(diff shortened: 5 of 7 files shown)


jpkrohling:
Two extra comments:

1. I removed `CGO_DISABLED=0`, as I need CGO to get the amount of available memory. I couldn't figure out why it has been set this way, and it seems to be introduced by #69 without further comment. If there's a reason to have CGO disabled, let me know and I'll try to come up with another way to obtain the amount of memory available (no promises, as I did some research already and the other options seem worse than the current approach)
1. This can be tested with: 
```console
$ go run -tags ui ./cmd/all-in-one/main.go  --collector.dyn-queue-size-warmup 1000 --memory.max-traces 10000
$ go run ./cmd/tracegen/main.go -workers 10 -duration 5m
```
There should be both a change recorded in the log and in the metric `jaeger_collector_queue_capacity`.

Makefile
```diff
@@ -190,7 +190,7 @@ elasticsearch-mappings:
 .PHONY: build-examples
 build-examples:
 	esc -pkg frontend -o examples/hotrod/services/frontend/gen_assets.go  -prefix examples/hotrod/services/frontend/web_assets examples/hotrod/services/frontend/web_assets
-	CGO_ENABLED=0 installsuffix=cgo go build -o ./examples/hotrod/hotrod-$(GOOS) ./examples/hotrod/main.go
+	installsuffix=cgo go build -o ./examples/hotrod/hotrod-$(GOOS) ./examples/hotrod/main.go
```
jpkrohling: @pavolloffay suggested having these changes in a separate PR.

cmd/collector/app/options.go
```diff
@@ -172,5 +191,15 @@ func (o options) apply(opts ...Option) options {
 	if ret.numWorkers == 0 {
 		ret.numWorkers = DefaultNumWorkers
 	}
+	if ret.dynQueueSizeMemory == 0 {
+		// this seems to be the best way to get the total memory size of the host without resorting to external libraries
```
yurishkuro: I think if this option is zero the functionality should be disabled. If people want to manage queue size by memory, they should be in charge of defining the limit, not some hidden heuristic in Jaeger.

Makefile
yurishkuro: I agree. I also am not sure we need this.

cmd/collector/app/options_test.go
```diff
@@ -39,12 +39,16 @@ func TestAllOptionSet(t *testing.T) {
 		Options.PreProcessSpans(func(spans []*model.Span) {}),
 		Options.Sanitizer(func(span *model.Span) *model.Span { return span }),
 		Options.QueueSize(10),
+		Options.DynQueueSizeWarmup(1_000),
+		Options.DynQueueSizeMemory(1_024),
```
yurishkuro: underscores seem overkill for these small numbers

cmd/collector/app/span_processor.go
```diff
@@ -130,6 +154,11 @@ func (sp *spanProcessor) saveSpan(span *model.Span) {
 	sp.metrics.SaveLatency.Record(time.Since(startTime))
 }
 
+func (sp *spanProcessor) countSpan(span *model.Span) {
+	atomic.AddUint64(&sp.bytesProcessed, uint64(span.Size()))
+	atomic.AddUint64(&sp.spansProcessed, 1)
```
yurishkuro: need to make sure these fields are properly aligned. Consider using uber-go/atomic instead, which we already depend on.

cmd/collector/app/span_processor.go
```diff
@@ -87,27 +100,38 @@ func newSpanProcessor(spanWriter spanstore.Writer, opts ...Option) *spanProcesso
...
+		preProcessSpans:    options.preProcessSpans,
+		filterSpan:         options.spanFilter,
+		sanitizer:          options.sanitizer,
+		reportBusy:         options.reportBusy,
+		numWorkers:         options.numWorkers,
+		spanWriter:         spanWriter,
+		collectorTags:      options.collectorTags,
+		dynQueueSizeWarmup: options.dynQueueSizeWarmup,
+		stopCh:             make(chan struct{}),
+		dynQueueSizeMemory: options.dynQueueSizeMemory,
```
yurishkuro: group options field together

cmd/collector/app/span_processor.go
```diff
@@ -87,27 +100,38 @@ func newSpanProcessor(spanWriter spanstore.Writer, opts ...Option) *spanProcesso
...
+	}
+
+	processSpanFuncs := []ProcessSpan{}
```
yurishkuro: can be initialized with `options.preSave, sp.saveSpan`, instead of adding them later, since they are not optional

cmd/collector/app/span_processor.go
```diff
@@ -70,7 +79,11 @@ func NewSpanProcessor(
 		sp.processItemFromQueue(value)
 	})
 
-	sp.queue.StartLengthReporting(1*time.Second, sp.metrics.QueueLength)
+	sp.StartStateReporting(1 * time.Second)
```
yurishkuro: Don't think this function is needed:
```
sp.background(time.Second, sp.updateGauges)
```

cmd/collector/app/span_processor.go
```diff
@@ -70,7 +79,11 @@ func NewSpanProcessor(
...
+
+	if sp.dynQueueSizeWarmup > 0 {
+		sp.StartDynQueueSizeUpdater(time.Second)
```
yurishkuro:
```go
sp.background(time.Minute, sp.updateQueueSize)
```

also, once a minute is probably fine

cmd/collector/app/span_processor.go
```diff
@@ -177,3 +206,76 @@ func (sp *spanProcessor) enqueueSpan(span *model.Span, originalFormat SpanFormat
...
+}
+
+func (sp *spanProcessor) StartDynQueueSizeUpdater(reportPeriod time.Duration) {
+	sp.background(reportPeriod, func() {
+		sp.updateQueueSize()
+	})
+}
+
+func (sp *spanProcessor) background(reportPeriod time.Duration, callback func()) {
+	ticker := time.NewTicker(reportPeriod)
```
yurishkuro: nit: this can be scoped inside `go func() {}`

cmd/collector/app/span_processor.go
```diff
@@ -177,3 +206,76 @@ func (sp *spanProcessor) enqueueSpan(span *model.Span, originalFormat SpanFormat
...
+		diff = idealQueueSize / current
+	} else {
+		diff = current / idealQueueSize
+	}
+
+	// resizing is a costly operation, we only perform it if we are at least 10% apart from the desired value
+	if diff > 1.1 {
+		s := int(idealQueueSize)
+		sp.logger.Info("Resizing the internal span queue", zap.Int("new-size", s), zap.Uint64("average-span-size-bytes", average))
+		sp.queue.Resize(s)
```
yurishkuro: I think there needs to be a mutex to prevent concurrent resize attempts

cmd/collector/app/span_processor.go
```diff
@@ -177,3 +206,76 @@ func (sp *spanProcessor) enqueueSpan(span *model.Span, originalFormat SpanFormat
...
+	if sp.dynQueueSizeWarmup == 0 {
+		return
+	}
+
+	if sp.spansProcessed < uint64(sp.dynQueueSizeWarmup) {
+		return
+	}
+
+	// first, we get the average size of a span, by dividing the bytes processed by num of spans
+	average := sp.bytesProcessed / sp.spansProcessed
```
yurishkuro: it might make more sense to reset these variables once a resize happens, so that future resizes are based on recent data.

cmd/collector/app/span_processor_test.go
```diff
@@ -345,3 +345,148 @@ func TestSpanProcessorWithCollectorTags(t *testing.T) {
...
+	assert.NotEqual(t, uint64(0), p.bytesProcessed)
+
+	for i := 0; i < 15; i++ {
+		_, g := mb.Snapshot()
+		if g["spans.bytes"] == 0 {
+			time.Sleep(time.Millisecond)
+		} else {
+			assert.Equal(t, p.bytesProcessed, uint64(g["spans.bytes"]))
+			return
+		}
```
yurishkuro: reduce branching:
```go
		if b := g["spans.bytes"]; b > 0 {
			assert.Equal(t, p.bytesProcessed, uint64(b))
			return
		}
		time.Sleep(time.Millisecond)
```

pkg/queue/bounded_queue.go
```diff
@@ -164,6 +166,8 @@ func (q *BoundedQueue) Resize(capacity int) bool {
 	// #nosec
 	swapped := atomic.CompareAndSwapPointer((*unsafe.Pointer)(unsafe.Pointer(&q.items)), unsafe.Pointer(q.items), unsafe.Pointer(&queue))
 	if swapped {
+		atomic.CompareAndSwapInt32(&q.capacity, int32(q.Capacity()), int32(capacity))
```
yurishkuro: queue capacity is determined by the channel in L47, updating this field does not achieve anything.

Makefile
jpkrohling: Extracted as part of #2003

cmd/collector/app/options_test.go
jpkrohling: Changed

cmd/collector/app/options.go
jpkrohling: Agree that people should specify that all the time, but this is in place as a "sensible default" that would be effective only when the warmup value is set. I can remove it and disable the feature with a warning in case only the warmup value is set.

cmd/collector/app/span_processor.go
jpkrohling: Not sure I follow. Do you mean that the gofunc + timer (the current code from `sp.background()`) should be duplicated directly in the other functions (`StartDynQueueSizeUpdater`, `StartStateReporting`)?

cmd/collector/app/span_processor.go
jpkrohling: Would the order matter? I don't think it would, but it's worth asking...

cmd/collector/app/span_processor.go
jpkrohling: Can we add that as something to improve in a future PR? There are some aspects that we should consider for this:

1. the effects of this as a metric (increase, increase, reset, increase, ...)
1. desirable/undesirable effects on quick bursts -- the way we have, quick bursts won't cause unnecessary resizes, which is a good thing if we are under heavy load (less processing), but if the burst is big enough, it might cause the collector to use more memory than it has available

jpkrohling: Created #2004 to track this

cmd/collector/app/span_processor.go
jpkrohling: There should be only one such operation per minute at most, but you are right. Better be safe here.

pkg/queue/bounded_queue.go
jpkrohling: True. I had in mind the case where two channels exist at a given time after a resize operation, but the new channel will indeed have the right capacity value if the previous pointer swap happened successfully.

cmd/collector/app/options.go
yurishkuro: I prefer not to provide "sensible default" for this setting (which would also make the CGO changes unnecessary). Let people figure out what they want to do with the memory.

cmd/collector/app/span_processor.go
jpkrohling: Sorry, I get it now. Updated.

jpkrohling:
Marked as WIP, as it depends on #1997.

jpkrohling:
Perhaps someone else could review this one? @vprithvi ?

cmd/collector/app/span_processor.go
```diff
@@ -177,3 +209,79 @@ func (sp *spanProcessor) enqueueSpan(span *model.Span, originalFormat SpanFormat
...
 	return sp.queue.Produce(item)
 }
+
+func (sp *spanProcessor) StartStateReporting(reportPeriod time.Duration) {
+	sp.background(reportPeriod, sp.updateGauges)
+}
+
+func (sp *spanProcessor) StartDynQueueSizeUpdater(reportPeriod time.Duration) {
+	sp.background(reportPeriod, sp.updateQueueSize)
+}
```
yurishkuro: Why not just call `sp.background(reportPeriod, sp.updateQueueSize)` directly? These two functions are only called once each.

Why do they need to be public?

cmd/collector/app/span_processor.go
```diff
@@ -177,3 +209,79 @@ func (sp *spanProcessor) enqueueSpan(span *model.Span, originalFormat SpanFormat
...
+				return
+			}
+		}
+	}()
+}
+
+func (sp *spanProcessor) updateQueueSize() {
+	if sp.spansProcessed.Load() == 0 {
+		return
+	}
```
yurishkuro: this check is redundant, it is covered by the condition in L250

cmd/collector/app/span_processor.go
```diff
@@ -177,3 +209,79 @@ func (sp *spanProcessor) enqueueSpan(span *model.Span, originalFormat SpanFormat
...
+
+	var diff float64
+	current := float64(sp.queue.Capacity())
+	if idealQueueSize > current {
+		diff = idealQueueSize / current
+	} else {
+		diff = current / idealQueueSize
+	}
+
+	// resizing is a costly operation, we only perform it if we are at least 10% apart from the desired value
```
yurishkuro: 10% seems pretty low tolerance

yurishkuro:
cf #2013

jpkrohling:
> E.g. we start with a circular buffer of some configurable initial capacity, and just count how many bytes are stored in it using Span.Size().

@pavolloffay had the same suggestion last week. We could do some tests, but I think this would need synchronization in a few extra places than just around the two counters from this PR. In the end, it might be worth it, in the name of precision. My original idea though wasn't to provide something that consumes *at most* X bytes, but to provide something with a limit of *approximately* X bytes. Given that the queue would be empty most of the time, this is a good compromise between precision and performance.

In any case, I see you opened a PR with your suggestion, and given that two maintainers are aligned on the idea that the linked list is a better approach, I'm closing this one.

jpkrohling:
Reopening, as discussed in #2013

cmd/collector/app/span_processor.go
jpkrohling: +1, I think this code was bigger in a previous iteration. In its current state, doesn't make sense to have it as a separate function.

cmd/collector/app/span_processor.go
jpkrohling: +1

cmd/collector/app/span_processor.go
jpkrohling: What would be a better value? 20%? Note that this is only calculated after the warmup phase. In any case, I increased it to 20%. We can then see how it works in the real world and adjust accordingly.

cmd/collector/app/metrics.go
```diff
@@ -58,9 +58,13 @@ type SpanProcessorMetrics struct {
 	InQueueLatency metrics.Timer
 	// SpansDropped measures the number of spans we discarded because the queue was full
 	SpansDropped metrics.Counter
+	// SpansBytes records how many bytes were processed
+	SpansBytes metrics.Gauge
```
yurishkuro: From the comment sounds like this should be a counter.

cmd/collector/app/span_processor.go
```diff
@@ -177,3 +209,67 @@ func (sp *spanProcessor) enqueueSpan(span *model.Span, originalFormat SpanFormat
...
+		diff = idealQueueSize / current
+	} else {
+		diff = current / idealQueueSize
+	}
+
+	// resizing is a costly operation, we only perform it if we are at least 20% apart from the desired value
+	if diff > 1.2 {
+		s := int(idealQueueSize)
+		sp.logger.Info("Resizing the internal span queue", zap.Int("new-size", s), zap.Uint64("average-span-size-bytes", average))
+		sp.queueResizeMu.Lock()
```
yurishkuro: I would move the locking higher, before the size calculation starts.

cmd/collector/app/metrics.go
jpkrohling: Sounds like it, but I'm keeping track of this in a local var, and this metric reflects the value of that var.

cmd/collector/app/span_processor.go
jpkrohling: Done

jpkrohling:
> The only concern with merging it ahead of true memory bound queue is that it will introduce the warm-up flag which won't be applicable later. We already have queue length flag, maybe it can serve two purposes? Is is really important for the user to control the warm-up length?

Great question. We took the presence of the warmup value as the hint to activate this feature, and calculated a reasonable default for the queue size in MiB, so that people would only have to set the warmup flag. Now that we removed the default for the queue size in MiB, users have to set it anyway, so, we could activate the feature based on that instead.

I'm OK in leaving the warmup flag out for now, and see if people complain about the fixed warmup value (= queue size).

cmd/collector/app/span_processor_test.go
```diff
@@ -346,3 +347,160 @@ func TestSpanProcessorWithCollectorTags(t *testing.T) {
...
+			assert.Equal(t, p.bytesProcessed.Load(), uint64(g["spans.bytes"]))
+			return
+		}
+		time.Sleep(time.Millisecond)
+	}
+
+	assert.Fail(t, "gauge hasn't been updated within a reasonable amount of time")
+}
+
+func TestUpdateQueueSizeScaleUp(t *testing.T) {
```
albertteoh: Could some of the following tests be DRYer through table-driven tests?

cmd/collector/app/span_processor.go
```diff
@@ -27,6 +29,9 @@ import (
 	"github.com/jaegertracing/jaeger/storage/spanstore"
 )
 
+// if this proves to be too low, we can increase it
+const maxQueueSize = 1_000_000
```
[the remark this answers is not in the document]
jpkrohling: We could allow customization of this value, but I'd do it only if users ask. The queue is supposed to be close to empty most of the time, and we need a hard cap only to ensure we don't get wild values by a bug somewhere.

cmd/collector/app/builder/builder_flags.go
```diff
@@ -46,6 +49,10 @@ var tlsFlagsConfig = tlscfg.ServerFlagsConfig{
 
 // CollectorOptions holds configuration for collector
 type CollectorOptions struct {
+	// DynQueueSizeWarmup determines when the dynamic queue size feature should start. Setting it to 0 disables the feature.
```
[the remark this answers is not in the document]
jpkrohling: I ended up removing this flag.

cmd/collector/app/span_processor.go
```diff
@@ -177,3 +209,67 @@ func (sp *spanProcessor) enqueueSpan(span *model.Span, originalFormat SpanFormat
...
+	var diff float64
+	current := float64(sp.queue.Capacity())
+	if idealQueueSize > current {
+		diff = idealQueueSize / current
+	} else {
+		diff = current / idealQueueSize
+	}
+
+	// resizing is a costly operation, we only perform it if we are at least 20% apart from the desired value
+	if diff > 1.2 {
```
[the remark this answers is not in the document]
jpkrohling: Funny, I find the current way more readable. If others agree with you, I'll change it.

cmd/collector/app/span_processor_test.go
jpkrohling: Good idea!

cmd/collector/app/span_processor_test.go
```diff
@@ -346,3 +347,160 @@ func TestSpanProcessorWithCollectorTags(t *testing.T) {
...
+	p.spansProcessed = atomic.NewUint64(1000)
+	p.bytesProcessed = atomic.NewUint64(10 * 1024 * p.spansProcessed.Load()) // 10KiB per span
+
+	// 1024 ^ 3 / (10 * 1024) = 104857,6
+	// ideal queue size = 104857
+	p.background(10*time.Millisecond, p.updateQueueSize)
+
+	// we wait up to 50 milliseconds
+	for i := 0; i < 5; i++ {
+		if p.queue.Capacity() == 100 {
```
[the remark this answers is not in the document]
jpkrohling: Most of those occurrences were addressed while changing the most of the new tests to be part of a single table-driven test.

cmd/collector/app/span_processor.go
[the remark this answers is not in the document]
yurishkuro: I think raw number is easier to read, but the trade-off is that named constants at the top of the file hint at available parameterization, which is otherwise hard to spot.

cmd/collector/app/builder/builder_flags.go
```diff
@@ -27,6 +27,7 @@ import (
 )
 
 const (
+	collectorDynQueueSizeMemory   = "collector.dyn-queue-size-memory"
```
yurishkuro: The `dyn` prefix doesn't seem relevant.

jpkrohling: Changed

jpkrohling:
Final test:

```
$ go run -tags ui ./cmd/all-in-one/main.go --collector.queue-size-memory 100 --memory.max-traces 10000
...
...
{"level":"info","ts":1579770160.554552,"caller":"app/span_processor.go:130","msg":"Dynamically adjusting the queue size at runtime.","memory-mib":100,"queue-size-warmup":2000}
...
...
```

In another terminal:

```
$ go run ./cmd/tracegen/main.go -workers 10 -duration 5m
```

After a while, this is shown in the collector's logs:

```
{"level":"info","ts":1579770220.5560262,"caller":"app/span_processor.go:272","msg":"Resizing the internal span queue","new-size":335008,"average-span-size-bytes":313}
```

And here are the new metrics:

```
# HELP jaeger_collector_queue_capacity queue-capacity
# TYPE jaeger_collector_queue_capacity gauge
jaeger_collector_queue_capacity{host="caju"} 335008

# HELP jaeger_collector_spans_bytes spans.bytes
# TYPE jaeger_collector_spans_bytes gauge
jaeger_collector_spans_bytes{host="caju"} 1.687953831e+09
```

Issue #1971, opened 2019-12-10
Replicate /sampling endpoint from agent in the collector


yurishkuro:
## Requirement - what kind of business use case are you trying to solve?

When using HTTP reporters in the clients to export tracing data directly to the collectors, users still want to be able to use remote sampler that loads centrally managed sampling strategies.

## Problem - what in Jaeger blocks you from solving the requirement?

The `/sampling` endpoint is only implemented in the agent today.

## Proposal - what do you suggest to solve the problem or improve the existing situation?

Expose the same endpoint from the collectors, e.g. `/api/sampling`.

yurishkuro:
When this is done, update FAQ docs, cf. https://github.com/jaegertracing/documentation/pull/335

namratachaudhary:
Hi
If this is still open, I can take it up.

jpkrohling:
It's yours.

RickyRajinder:
Hi @namratachaudhary I have already submitted a PR for this issue.

yurishkuro:
Open question: can clients be configured with a different sampling URL?

yurishkuro:
Clients support the following config options:
  * JAEGER_ENDPOINT, e.g. `http(s)://collector:port/api/traces`
  * JAEGER_SAMPLER_MANAGER_HOST_PORT - name implies no part is allowed

Possible solutions:
  * We could introduce JAEGER_SAMPLING_ENDPOINT, but that still leaves out other configurations like baggage restrictions. 
  * We could introduce JAEGER_CONFIG_ENDPOINT, but that wouldn't be a real endpoint, but a base address, with `/sampling` or `/baggage` path added by the client as needed.
  * We could introduce JAEGER_API (or something along these lines) that will be the base address used by both span reporter (by adding `/traces`) and sampler (by adding `/sampling`).

Q: is it better to hardcode the final segments of the URL in the clients or make them configurable? Strictly speaking the URL path is part of the contract, e.g. if we introduce a different format of the sampling output it should become `/api/v2/sampling` or something like that.

cc @jpkrohling @pavolloffay @objectiser

pavolloffay:
> Q: is it better to hardcode the final segments of the URL in the clients or make them configurable?

I would lean towards the consistency and require configuring the full URL. It's more flexible and one could deploy it on different path.

yurishkuro:
@pavolloffay so you'd vote for JAEGER_SAMPLING_ENDPOINT?

[...]

Issue #1455, opened 2019-04-03
Cross-dock tests are flaky


jpkrohling:
I'm noticing that `master` is marked as `red` in Travis quite often. Whenever I notice it, I just start again the test that failed, and the culprit is almost always `cross-dock`, with failures like:

```
Executing Matrix...
✗ [endtoend] test_driver→ (services=go) ⇒ Fail: could not retrieve traces from query service
✓ [endtoend] test_driver→ (services=java) ⇒ Pass
✓ [endtoend] test_driver→ (services=node) ⇒ Pass
✓ [endtoend] test_driver→ (services=python) ⇒ Pass
✓ [endtoend] test_driver→ (services=zipkin-brave-json) ⇒ Pass
✓ [endtoend] test_driver→ (services=zipkin-brave-json-v2) ⇒ Pass
✓ [endtoend] test_driver→ (services=zipkin-brave-thrift) ⇒ Pass
6/7 passed (0/7 skipped)
Tests did not pass!
2019/04/02 14:20:48 At least one test failed
grep 'Tests passed!' run-crossdock.log
make[1]: *** [crossdock] Error 1
make[1]: Leaving directory `/home/travis/gopath/src/github.com/jaegertracing/jaeger'
make: *** [build-crossdock] Error 2
The command "if [ "$CROSSDOCK" == true ]; then bash ./scripts/travis/build-crossdock.sh ; else echo 'skipping crossdock'; fi" exited with 2.
```

yurishkuro:
I tried to increase logging recently, but it didn't help much, so still don't know what's causing it, only know that both times this started happening in the past 6mo was strongly correlated with enabling grpc instead of tchannel (non-scientific evidence).

jpkrohling:
I'll reuse this issue to track the otel-crossdock tests, as they are also flaky.

PR #1990, opened 2019-12-21
Add /api/sampling endpoint in collector

RickyRajinder:
- #1971: Expose the /api/sampling endpoint in collector to allow clients to retrieve sampling strategies without using the agent.

- Consolidated agent and collector API handlers into one

### DIFF
--- a/cmd/all-in-one/main.go
+++ b/cmd/all-in-one/main.go
@@ -16,7 +16,6 @@
 package main
 
 import (
-	"errors"
 	"fmt"
 	"io"
 	"log"
@@ -66,7 +65,6 @@ import (
 	"github.com/jaegertracing/jaeger/storage/dependencystore"
 	"github.com/jaegertracing/jaeger/storage/spanstore"
 	storageMetrics "github.com/jaegertracing/jaeger/storage/spanstore/metrics"
-	"github.com/jaegertracing/jaeger/thrift-gen/baggage"
 	jc "github.com/jaegertracing/jaeger/thrift-gen/jaeger"
 	sc "github.com/jaegertracing/jaeger/thrift-gen/sampling"
 	zc "github.com/jaegertracing/jaeger/thrift-gen/zipkincore"
@@ -254,20 +252,24 @@ func startCollector(
 
 	{
 		r := mux.NewRouter()
-		configManager := &collectorConfigManager{samplingStrategy: strategyStore}
-		cfgParams := clientcfgHandler.HTTPHandlerParams{
-			ConfigManager:          configManager,
-			MetricsFactory:         metricsFactory,
-			LegacySamplingEndpoint: false,
-		}
-		apiHandler := collectorApp.NewAPIHandler(jaegerBatchesHandler, *clientcfgHandler.NewHTTPHandler(cfgParams))
+		apiHandler := collectorApp.NewAPIHandler(jaegerBatchesHandler)
 		apiHandler.RegisterRoutes(r)
-		httpPortStr := ":" + strconv.Itoa(cOpts.CollectorHTTPPort)
-		recoveryHandler := recoveryhandler.NewRecoveryHandler(logger, true)
 
+		cfgHandler := clientcfgHandler.NewHTTPHandler(clientcfgHandler.HTTPHandlerParams{
+			ConfigManager: &clientcfgHandler.ConfigManager{
+				SamplingStrategyStore: strategyStore,
+			},
+			MetricsFactory:         metricsFactory,
+			BasePath:               "/api",
+			LegacySamplingEndpoint: false,
+		})
+		cfgHandler.RegisterRoutes(r)
+
+		recoveryHandler := recoveryhandler.NewRecoveryHandler(logger, true)
 		go startZipkinHTTPAPI(logger, cOpts.CollectorZipkinHTTPPort, zipkinSpansHandler, recoveryHandler)
 
-		logger.Info("Starting jaeger-collector HTTP server", zap.Int("http-port", cOpts.CollectorHTTPPort))
+		httpPortStr := ":" + strconv.Itoa(cOpts.CollectorHTTPPort)
+		logger.Info("Starting jaeger-collector HTTP server", zap.String("http-host-port", httpPortStr))
 		go func() {
 			if err := http.ListenAndServe(httpPortStr, recoveryHandler(r)); err != nil {
 				logger.Fatal("Could not launch jaeger-collector HTTP server", zap.Error(err))
@@ -329,18 +331,6 @@ func startQuery(
 		svc.Logger.Fatal("Could not start jaeger-query service", zap.Error(err))
 	}
 	return server
-}
-
-type collectorConfigManager struct {
-	samplingStrategy strategystore.StrategyStore
-}
-
-func (c *collectorConfigManager) GetSamplingStrategy(serviceName string) (*sc.SamplingStrategyResponse, error) {
-	return c.samplingStrategy.GetSamplingStrategy(serviceName)
-}
-
-func (c *collectorConfigManager) GetBaggageRestrictions(serviceName string) ([]*baggage.BaggageRestriction, error) {
-	return nil, errors.New("baggage not implemented")
 }
 
 func initSamplingStrategyStore(
--- a/cmd/all-in-one/main.go
+++ b/cmd/all-in-one/main.go
@@ -16,6 +16,7 @@
 package main
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"log"
@@ -53,6 +54,7 @@ import (
 	"github.com/jaegertracing/jaeger/cmd/flags"
 	queryApp "github.com/jaegertracing/jaeger/cmd/query/app"
 	"github.com/jaegertracing/jaeger/cmd/query/app/querysvc"
+	clientcfgHandler "github.com/jaegertracing/jaeger/pkg/clientcfg/clientcfghttp"
 	"github.com/jaegertracing/jaeger/pkg/config"
 	"github.com/jaegertracing/jaeger/pkg/healthcheck"
 	"github.com/jaegertracing/jaeger/pkg/recoveryhandler"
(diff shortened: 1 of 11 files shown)


cmd/collector/app/http_handler.go
```diff
@@ -24,12 +26,15 @@ import (
 	"github.com/apache/thrift/lib/go/thrift"
 	"github.com/gorilla/mux"
 
+	"github.com/jaeger/cmd/agent/app/configmanager"
 	tJaeger "github.com/jaegertracing/jaeger/thrift-gen/jaeger"
+	"github.com/uber/jaeger-lib/metrics"
```
yurishkuro: this should be in the previous import group as it's not from this repo.

cmd/collector/app/http_handler.go
```diff
@@ -39,23 +44,46 @@ var (
 	}
 )
 
+var (
+	errBadRequest = errors.New("bad request")
+)
+
 // APIHandler handles all HTTP calls to the collector
 type APIHandler struct {
+	manager configmanager.ClientConfigManager
```
yurishkuro:
```suggestion
	configManager configmanager.ClientConfigManager
```

cmd/collector/app/http_handler.go
```diff
@@ -95,3 +123,45 @@ func (aH *APIHandler) SaveSpan(w http.ResponseWriter, r *http.Request) {
 
 	w.WriteHeader(http.StatusAccepted)
 }
+
+func (aH *APIHandler) ServeSampling(w http.ResponseWriter, r *http.Request) {
```
yurishkuro: This looks like a copy-paste of the code from the agent. We need to refactor that code into a reusable piece that can be called from both the agent and collector.

cmd/collector/app/http_handler_test.go
```diff
@@ -28,10 +31,14 @@ import (
 	"github.com/apache/thrift/lib/go/thrift"
 	"github.com/gorilla/mux"
 	"github.com/stretchr/testify/assert"
-	jaegerClient "github.com/uber/jaeger-client-go"
-	"github.com/uber/jaeger-client-go/transport"
+	"github.com/stretchr/testify/require"
 
+	"github.com/jaeger/cmd/agent/app/httpserver"
```
yurishkuro: there is no such package

cmd/agent/app/httpserver/server_test.go
```diff
@@ -26,44 +26,18 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	"github.com/uber/jaeger-lib/metrics/metricstest"
 
 	tSampling092 "github.com/jaegertracing/jaeger/cmd/agent/app/httpserver/thrift-0.9.2"
 	"github.com/jaegertracing/jaeger/thrift-gen/baggage"
 	"github.com/jaegertracing/jaeger/thrift-gen/sampling"
+	"github.com/uber/jaeger-lib/metrics/metricstest"
```
yurishkuro: it was in the right group

RickyRajinder:
@yurishkuro Do I update the content/docs/next-release/faq in the documentation repo?

yurishkuro:
> Do I update the content/docs/next-release/faq in the documentation repo?

yes

RickyRajinder:
@yurishkuro How would an import from the same repo resolve if it is pointing to the remote master that does not contain the branch's changes?

yurishkuro:
Sorry, I don't follow.

RickyRajinder:
@yurishkuro For example, when I import `"github.com/jaegertracing/jaeger/cmd/agent/app/httpserver"` the import points to the remote repo that has been downloaded in `/vendor`, which causes compilation errors. So my question is, how would I properly import the local package from my branch.

yurishkuro:
There's something wrong with your Go setup then. Maybe you cloned this repo in the wrong location, it should be under `$GOPATH/src/github.com/jaegertracing/jaeger/`

RickyRajinder:
@yurishkuro it appears cross-dock is failing and it is not clear from the logs as to why. Could this be related to #1455?

cmd/agent/app/httpserver/server.go
```diff
@@ -36,30 +36,32 @@ var (
...
 	})
 	mux.HandleFunc("/baggageRestrictions", func(w http.ResponseWriter, r *http.Request) {
 		handler.serveBaggageHTTP(w, r)
 	})
 	return &http.Server{Addr: hostPort, Handler: mux}
 }
 
-func newHTTPHandler(manager configmanager.ClientConfigManager, mFactory metrics.Factory) *httpHandler {
-	handler := &httpHandler{manager: manager}
+//NewHTTPHandler creates a new HTTP endpoint handler
```
yurishkuro: please leave space after `//`

cmd/all-in-one/main.go
```diff
@@ -255,7 +259,13 @@ func startCollector(
 
 	{
 		r := mux.NewRouter()
-		apiHandler := collectorApp.NewAPIHandler(jaegerBatchesHandler)
+		grpcBuilder.CollectorHostPorts = append(grpcBuilder.CollectorHostPorts, fmt.Sprintf("127.0.0.1:%d", cOpts.CollectorGRPCPort))
+		cp, err := agentApp.CreateCollectorProxy(repOpts, tchanBuilder, grpcBuilder, logger, metricsFactory)
```
yurishkuro: collector is already running the modules that generate configurations, there is no reason for the new HTTP handler to do another network hop back to collector itself, we should be able to just give the handler an implementation of an interface that provides the data directly

RickyRajinder:
> This PR now creates a lot of new public APIs. I don't think it's necessary. What we need is to pull the HTTP handler from the agent into a shared location, e.g. `/pkg/clientcfg/http` where there's only a single public type `HTTPHandler` that has `RegisterHander(mux)` function. The actual HTTPServer should remain in the agent, and the collector has its own HTTPService (used for /api/traces endpoint). This will also mean the test infrastructure that you made public can remain private in that package.

@yurishkuro if `RegisterHandler(mux)` will go in `/pkg/clientcfg/http/http_handler.go`, for example, then does that mean you want all the handler functions such as `serveSamplingHTTP` and `serveBaggageHTTP` to go there as well?

yurishkuro:
Yes, since they are part of the implementation. And they can remain private

RickyRajinder:
> I did the refactoring as a separate PR #1996, please rebase. It should limit the scope of this change to only registering the endpoints from collector.

So `APIHandler` in `cmd/collector/app/http_handler.go` should contain a `clientcfghttp.HTTPHandler`?

cmd/all-in-one/main.go
```diff
@@ -322,6 +331,18 @@ func startQuery(
 	return server
 }
 
+type collectorConfigManager struct {
+	samplingStrategy strategystore.StrategyStore
```
yurishkuro:
```suggestion
	strategyStore strategystore.StrategyStore
```

cmd/collector/app/http_handler.go
```diff
@@ -42,19 +38,20 @@ var (
...
+func NewAPIHandler(jaegerBatchesHandler JaegerBatchesHandler, cfgHandler clientcfghttp.HTTPHandler) *APIHandler {
 	return &APIHandler{
 		jaegerBatchesHandler: jaegerBatchesHandler,
+		cfgHandler:           cfgHandler,
 	}
 }
 
 // RegisterRoutes registers routes for this handler on the given router
 func (aH *APIHandler) RegisterRoutes(router *mux.Router) {
+	aH.cfgHandler.RegisterRoutes(router)
```
yurishkuro: I think it would be better to mount the endpoints under`/api` prefix, to match `/api/traces`. To do that we need to extend cfgHandler struct to accept a `BasePath` parameter via `HTTPHandlerParams`.

cmd/all-in-one/main.go
```diff
@@ -251,7 +254,13 @@ func startCollector(
...
 	{
 		r := mux.NewRouter()
-		apiHandler := collectorApp.NewAPIHandler(jaegerBatchesHandler)
+		configManager := &collectorConfigManager{samplingStrategy: strategyStore}
+		cfgParams := clientcfgHandler.HTTPHandlerParams{
+			ConfigManager:          configManager,
+			MetricsFactory:         metricsFactory,
+			LegacySamplingEndpoint: false,
+		}
+		apiHandler := collectorApp.NewAPIHandler(jaegerBatchesHandler, *clientcfgHandler.NewHTTPHandler(cfgParams))
```
yurishkuro: It is questionable whether all-in-one needs this, since it's already running the agent and the endpoints are available on the agent's http server. But in order to make all-in-one swappable for collector, let's keep it.

However, I don't think we need to pass cfg handler to APIHandler. We can simply call its RegisterRoutes here (same in the collector).

Nit: inline HTTPHandlerParams with NewHTTPHandler() call.

cmd/collector/main.go
```diff
@@ -241,6 +250,18 @@ func startZipkinHTTPAPI(
...
+type collectorConfigManager struct {
+	samplingStrategy strategystore.StrategyStore
+}
+
+func (c *collectorConfigManager) GetSamplingStrategy(serviceName string) (*sc.SamplingStrategyResponse, error) {
+	return c.samplingStrategy.GetSamplingStrategy(serviceName)
+}
+
+func (c *collectorConfigManager) GetBaggageRestrictions(serviceName string) ([]*baggage.BaggageRestriction, error) {
+	return nil, errors.New("baggage not implemented")
```
yurishkuro:
```suggestion
	return nil, errors.New("baggage restrictions not implemented")
```

cmd/collector/main.go
```diff
@@ -241,6 +250,18 @@ func startZipkinHTTPAPI(
 	}
 }
 
+type collectorConfigManager struct {
+	samplingStrategy strategystore.StrategyStore
```
yurishkuro:
```suggestion
	strategyStore strategystore.StrategyStore
```

PR #2000, opened 2020-01-03
Remove ineffectual context assignment

q384566678:
Signed-off-by: Zhou Hao <zhouhao@cn.fujitsu.com>

- 

- Delete unused variables.


plugin/storage/cassandra/spanstore/reader.go
```diff
@@ -377,7 +377,7 @@ func (s *SpanReader) queryByDuration(ctx context.Context, traceQuery *spanstore.
 
 func (s *SpanReader) queryByServiceNameAndOperation(ctx context.Context, tq *spanstore.TraceQueryParameters) (dbmodel.UniqueTraceIDs, error) {
 	//lint:ignore SA4006 failing to re-assign context is worse than unused variable
```
yurishkuro: in this case the comment should be removed

q384566678: updated, PTAL

PR #2002, opened 2020-01-06
Admin/Query: Log the real port instead of the provided one to enable the use of port 0

ChadiEM:
- When using port 0 with the admin or query port, the Jaeger components print 0 instead of the port actually used. What this PR changes is to print the actual port.

- Uses [`net.SplitHostPort`](https://golang.org/pkg/net/#SplitHostPort) to get the actual port.

### DIFF
--- a/cmd/flags/admin_test.go
+++ b/cmd/flags/admin_test.go
@@ -15,12 +15,17 @@
 package flags
 
 import (
+	"context"
+	"sync"
 	"testing"
 
-	"github.com/jaegertracing/jaeger/pkg/config"
+	"github.com/marusama/cyclicbarrier"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zaptest/observer"
+
+	"github.com/jaegertracing/jaeger/pkg/config"
 )
 
 func TestAdminServerHandlesPortZero(t *testing.T) {
@@ -42,4 +47,33 @@ func TestAdminServerHandlesPortZero(t *testing.T) {
 	onlyEntry := message.All()[0]
 	port := onlyEntry.ContextMap()["http-port"].(int64)
 	assert.Greater(t, port, int64(0))
+}
+
+func TestAdminServerHandlesSimultaneousStartupOnPortZero(t *testing.T) {
+	const parallelInstances = 2
+
+	var wg sync.WaitGroup
+	barrier := cyclicbarrier.New(parallelInstances)
+
+	for i := 0; i < parallelInstances; i++ {
+		wg.Add(1)
+
+		go func() {
+			defer wg.Done()
+
+			adminServer := NewAdminServer(0)
+
+			v, _ := config.Viperize(adminServer.AddFlags)
+			logger := zap.NewNop()
+			adminServer.initFromViper(v, logger)
+
+			if assert.NoError(t, adminServer.Serve(), "Failed to start two Admin servers at port 0") {
+				defer adminServer.Close()
+			}
+
+			require.NoError(t, barrier.Await(context.Background()))
+		}()
+	}
+
+	wg.Wait()
 }
--- a/cmd/flags/admin_test.go
+++ b/cmd/flags/admin_test.go
@@ -1,0 +1,45 @@
+// Copyright (c) 2020 The Jaeger Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package flags
+
+import (
+	"testing"
+
+	"github.com/jaegertracing/jaeger/pkg/config"
+	"github.com/stretchr/testify/assert"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zaptest/observer"
+)
+
+func TestAdminServerHandlesPortZero(t *testing.T) {
+	adminServer := NewAdminServer(0)
+
+	v, _ := config.Viperize(adminServer.AddFlags)
+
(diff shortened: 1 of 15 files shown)


cmd/query/app/server_test.go
```diff
@@ -103,3 +103,23 @@ func TestServerGracefulExit(t *testing.T) {
...
+	server := NewServer(flagsSvc, querySvc, &QueryOptions{Port: 0}, tracer)
+	assert.NoError(t, server.Start())
+	server.Close()
+
+	message := logs.FilterMessage("Query server started")
+	assert.Equal(t, 1, message.Len(), "Expected query started log message.")
+
+	onlyEntry := message.All()[0]
+	port := onlyEntry.ContextMap()["port"].(int)
+	assert.True(t, port > 0,
```
yurishkuro: can use `assert.Greater`

cmd/query/app/server_test.go
```diff
@@ -103,3 +103,23 @@ func TestServerGracefulExit(t *testing.T) {
...
+		"Expected a non-zero port in the logs, got instead: %d", port)
```
yurishkuro: strictly speaking, it is possible that NewServer() function replaces 0 port with the default value, in which case this test will not catch it.

yurishkuro: one way to guarantee that it does is by starting two servers and validating that they both use different ports.

pkg/netutils/port.go
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) 2019 The Jaeger Authors.
```
yurishkuro: 2020!

pkg/netutils/port.go
```diff
@@ -0,0 +1,35 @@
...
+// limitations under the License.
+
+package netutils
+
+import (
+	"net"
+	"strconv"
+)
+
+// GetPort returns the port of an endpoint address
```
yurishkuro: end with period, per Go guidelines.

pkg/netutils/port_test.go
```diff
@@ -0,0 +1,35 @@
...
+func TestPortZero(t *testing.T) {
+	lis, err := net.Listen("tcp", "localhost:0")
+	require.NoError(t, err)
+	defer lis.Close()
+
+	port, err := GetPort(lis.Addr())
+	require.NoError(t, err)
+
+	assert.Greater(t, port, 0,
+		"Expected a non-zero port, got instead: %d", port)
```
yurishkuro: don't need the message, assert will already print something very similar

cmd/query/app/server_test.go
yurishkuro: can we fix this?

cmd/query/app/server_test.go
```diff
@@ -103,3 +103,23 @@ func TestServerGracefulExit(t *testing.T) {
...
+	assert.NoError(t, server.Start())
+	server.Close()
+
+	message := logs.FilterMessage("Query server started")
+	assert.Equal(t, 1, message.Len(), "Expected query started log message.")
+
+	onlyEntry := message.All()[0]
+	port := onlyEntry.ContextMap()["port"].(int64)
+	assert.Greater(t, port, int64(0),
+		"Expected a non-zero port in the logs, got instead: %d", port)
```
yurishkuro: also, message `"Expected a non-zero port in the logs, got instead: %d"` is redundant

pkg/netutils/port_test.go
ChadiEM: yep, makes sense.

cmd/query/app/server_test.go
ChadiEM: > strictly speaking, it is possible that NewServer() function replaces 0 port with the default value, in which case this test will not catch it.

> one way to guarantee that it does is by starting two servers and validating that they both use different ports.


Sure, I will add that test. However, what this PR targets is printing the real port in the log file instead of 0 and not the fact of using port 0 itself.
Using 0 works even before this PR :-)


btw, I also have a gap for testing the admin port log output. In the current state, I do not have a way to set the logger on the `AdminServer`. What I can do is to expose `setLogger()` and use it, then call `Serve()` and check the log in that case. Would that be ok?

yurishkuro: Logger can be passed via InitFromViper. Look for Viperize helper function somewhere in the repo.

ChadiEM: I added tests for launching two instances in parallel. I used a cyclicbarrier to make sure that both servers are synchronized after the Start(), because without it, each goroutine can run on its own and the test behavior will be random. Let me know if you are ok with the thirdparty, or if you have other suggestions. :-)

ChadiEM:
For consistency, I could also apply this change to the [agent](https://github.com/jaegertracing/jaeger/blob/master/cmd/agent/app/agent.go#L70) code (although it uses a different method to get the port with an equivalent result).

ChadiEM:
Is the codecov/patch failure due to the lack of coverage of the error cases in the `GetPort` function?

cmd/flags/admin_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+	message := logs.FilterMessage("Admin server started")
+	assert.Equal(t, 1, message.Len(), "Expected Admin server started log message.")
+
+	onlyEntry := message.All()[0]
+	port := onlyEntry.ContextMap()["http-port"].(int64)
+	assert.Greater(t, port, int64(0))
+}
+
+func TestAdminServerHandlesSimultaneousStartupOnPortZero(t *testing.T) {
```
yurishkuro: I don't think this test is necessary (two admin servers is not a real use case). In fact, my comment about two servers was not to test _two servers themselves_, but to test that when port 0 is passed as a parameter it's not overridden by some fixed default port number, since this is what was the reason for your PR (iirc).

cmd/flags/admin_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+			v, _ := config.Viperize(adminServer.AddFlags)
+			logger := zap.NewNop()
+			adminServer.initFromViper(v, logger)
+
+			if assert.NoError(t, adminServer.Serve(), "Failed to start two Admin servers at port 0") {
+				defer adminServer.Close()
+			}
+
+			require.NoError(t, barrier.Await(context.Background()))
```
yurishkuro: what are trying to achieve with this?

yurishkuro:
Thanks for the PR!

ChadiEM:
Thanks @yurishkuro!

PR #2008, opened 2020-01-07
Follow-up on review comments for the bounded queue

jpkrohling:
- Some comments were left on #1949 after it got merged. This PR addresses some of the comments, and backports the remaining changes from #1985.

- Changed existing tests to always set a queue size
- Use uber-go/atomic for non-pointer types
- Other smaller items based on the review comments from #1949

### DIFF
--- a/pkg/queue/bounded_queue.go
+++ b/pkg/queue/bounded_queue.go
@@ -22,6 +22,7 @@ import (
 	"unsafe"
 
 	"github.com/uber/jaeger-lib/metrics"
+	uatomic "go.uber.org/atomic"
 )
 
 // BoundedQueue implements a producer-consumer exchange similar to a ring buffer queue,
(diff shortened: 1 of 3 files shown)


jpkrohling:
This is now ready for review. The only thing missing based on the review comments is to either use a mutex for touching the queue or wrapping it in `atomic.Value`. The first implementation of the PR had mutexes to change the backing queue and we talked about changing it to use a second queue, so that we could avoid mutexes.

I might be missing something, but I took another look at the code and couldn't find any serious consequences of a concurrent read while the new queue is being set. At most, we'd be getting an updated value for the capacity, which in the worst case will cause a span to be dropped (the other case is that the queues will have one more span than the desired capacity).

If you think it's still worth having a lock, I can certainly add one for the `Resize` operation.

pkg/queue/bounded_queue.go
```diff
@@ -165,10 +170,13 @@ func (q *BoundedQueue) Resize(capacity int) bool {
 	swapped := atomic.CompareAndSwapPointer((*unsafe.Pointer)(unsafe.Pointer(&q.items)), unsafe.Pointer(q.items), unsafe.Pointer(&queue))
 	if swapped {
 		// start a new set of consumers, based on the information given previously
-		q.StartConsumers(q.workers, q.consumer)
+		q.StartConsumers(int(q.workers), q.consumer)
```
yurishkuro: q.workers is already `int`

jpkrohling: Did you spot it using a different linter?

yurishkuro: no, just paying attention

jpkrohling: We should plug your brain in the CI, as our linters didn't catch it.

Issue #2005, opened 2020-01-06
Reporting client data loss stats to Jaeger backend


yurishkuro:
## Requirement - what kind of business use case are you trying to solve?

We want to have accurate global measurements of the potential data loss in Jaeger clients. 

## Problem - what in Jaeger blocks you from solving the requirement?

Although Jaeger clients internally support integrations with metrics backends, not all users actually configure those, and even if they do, the metrics are often namespaced to the application, e.g. by including a tag `service=xyz` or using a custom prefix, which makes it more difficult to observe across the site.

## Proposal - what do you suggest to solve the problem or improve the existing situation?

Allow clients to report certain data loss metrics when submitting span batches to Jaeger backend. The data elements should include:
  * batch sequence number to detect possible dropped packets (especially with UDP transports)
  * count of spans dropped by the client due to internal queue overflow
  * count of spans dropped by the client due to exceeding max packet size (e.g. 65000 bytes cap for UDP transport)

The Jaeger clients (some, at least) are already including a `client-uuid` string tag in the Process, which will allow the agent/collector to monitor those reported stats and emit consistent metrics.

* [x] Thrift IDL change https://github.com/jaegertracing/jaeger-idl/pull/60
* [x] Go client prototype https://github.com/jaegertracing/jaeger-client-go/pull/482
* [x] Agent prototype https://github.com/jaegertracing/jaeger/pull/2010
* [ ] Other clients

yurishkuro:
@jpkrohling @pavolloffay fyi

jpkrohling:
I wouldn't expect that extracting metrics from clients is difficult, especially because applications *should* be exposing their own metrics anyway. Once metrics are collected, it should also be easy to get aggregated values with or without service names using any reasonably modern metrics backend, like Prometheus.

But if you do need this internally, this might be a good indication that other people might benefit from it.

I do have one question, though:

> batch sequence number to detect possible dropped packets (especially with UDP transports)

What do you have in mind for syncing this across all collectors in a cluster?

yurishkuro:
> I wouldn't expect that extracting metrics from clients is difficult

We indeed found this quite challenging, given how heterogeneous Uber's ecosystem is. Some apps, for example, are still using statsd-style metrics that don't even support tags, a lot more apps simply do not configure Jaeger clients with a metrics client, because they are not interested in those metrics. So we want to have a more guaranteed way of collecting critical metrics like data loss (I don't want to replicate all 20 client metrics here).

[...]

PR #2010, opened 2020-01-09
[agent] Process data loss stats from clients

yurishkuro:
- #2005

- Keeps track of stats reported by clients in `jaegerThrift.Batch`, per `client-uuid`
- Emits corresponding metrics
- Cleans-up the per-client-uuid cache periodically

This is how new metrics look:
```
$ curl -s http://localhost:14269/metrics | grep client_stats
# HELP jaeger_agent_client_stats_batches_sent_total Total count of batches sent by clients
# TYPE jaeger_agent_client_stats_batches_sent_total counter
jaeger_agent_client_stats_batches_sent_total 4
# HELP jaeger_agent_client_stats_connected_clients Total count of unique clients sending data to the agent
# TYPE jaeger_agent_client_stats_connected_clients gauge
jaeger_agent_client_stats_connected_clients 1
# HELP jaeger_agent_client_stats_spans_dropped_total Total count of spans dropped by clients
# TYPE jaeger_agent_client_stats_spans_dropped_total counter
jaeger_agent_client_stats_spans_dropped_total{cause="full-queue"} 0
jaeger_agent_client_stats_spans_dropped_total{cause="send-failure"} 0
jaeger_agent_client_stats_spans_dropped_total{cause="too-large"} 0
```

## TODO

* [ ] Pin correct jaeger-client-go and jaeger-idl once their respective PRs are merged
* [ ] Add unit tests for new functionality

### DIFF
--- a/cmd/agent/app/reporter/client_metrics.go
+++ b/cmd/agent/app/reporter/client_metrics.go
@@ -63,8 +63,10 @@ type clientMetrics struct {
 }
 
 type lastReceivedClientStats struct {
-	lock                  sync.Mutex
-	lastUpdated           time.Time
+	lock        sync.Mutex
+	lastUpdated time.Time
+
+	// Thrift stats are reported as signed i64, so keep the type to avoid multiple conversions back and forth.
 	batchSeqNo            int64
 	fullQueueDroppedSpans int64
 	tooLargeDroppedSpans  int64
@@ -190,7 +192,9 @@ func (s *lastReceivedClientStats) update(
 	defer s.lock.Unlock()
 
 	if s.batchSeqNo >= batchSeqNo && !wrapped(s.batchSeqNo, batchSeqNo) {
-		// ignore out of order batches, the metrics will be updated later
+		// Ignore out of order batches. Once we receive a batch with a larger-than-seen number,
+		// it will contain new cumulative counts, which will we use to update the metrics.
+		// That makes the metrics slightly off in time, but accurate in aggregate.
 		return
 	}
 
--- a/cmd/agent/app/reporter/client_metrics.go
+++ b/cmd/agent/app/reporter/client_metrics.go
@@ -109,7 +109,7 @@ func WrapWithClientMetrics(params ClientMetricsReporterParams) *ClientMetricsRep
 		shutdown:      make(chan struct{}),
 		closed:        atomic.NewBool(false),
 	}
-	go r.expireClientMetrics()
+	go r.expireClientMetricsLoop()
 	return r
 }
 
@@ -131,34 +131,37 @@ func (r *ClientMetricsReporter) Close() {
 	}
 }
 
-func (r *ClientMetricsReporter) expireClientMetrics() {
+func (r *ClientMetricsReporter) expireClientMetricsLoop() {
 	ticker := time.NewTicker(r.params.ExpireFrequency)
 	defer ticker.Stop()
 	for {
 		select {
-		case <-ticker.C:
-			t := time.Now()
-			var size int64
-			r.lastReceivedClientStats.Range(func(k, v interface{}) bool {
-				stats := v.(*lastReceivedClientStats)
-				stats.lock.Lock()
-				defer stats.lock.Unlock()
-
-				if !stats.lastUpdated.IsZero() && t.Sub(stats.lastUpdated) > r.params.ExpireTTL {
-					r.lastReceivedClientStats.Delete(k)
-					r.params.Logger.Debug("have not heard from a client for a while, freeing stats",
-						zap.Any("client-uuid", k),
-						zap.Time("last-message", stats.lastUpdated),
-					)
-				}
-				size++
-				return true // keep running through all values in the map
-			})
-			r.clientMetrics.ConnectedClients.Update(size)
+		case now := <-ticker.C:
+			r.expireClientMetrics(now)
 		case <-r.shutdown:
 			return
 		}
 	}
+}
+
+func (r *ClientMetricsReporter) expireClientMetrics(t time.Time) {
+	var size int64
+	r.lastReceivedClientStats.Range(func(k, v interface{}) bool {
+		stats := v.(*lastReceivedClientStats)
+		stats.lock.Lock()
+		defer stats.lock.Unlock()
+
+		if !stats.lastUpdated.IsZero() && t.Sub(stats.lastUpdated) > r.params.ExpireTTL {
+			r.lastReceivedClientStats.Delete(k)
+			r.params.Logger.Debug("have not heard from a client for a while, freeing stats",
+				zap.Any("client-uuid", k),
+				zap.Time("last-message", stats.lastUpdated),
+			)
+		}
+		size++
+		return true // keep running through all values in the map
+	})
+	r.clientMetrics.ConnectedClients.Update(size)
 }
 
 func (r *ClientMetricsReporter) updateClientMetrics(batch *jaeger.Batch) {
@@ -202,8 +205,8 @@ func (s *lastReceivedClientStats) update(
 
 	if stats != nil {
 		metrics.FailedToEmitSpans.Inc(delta(s.failedToEmitSpans, stats.FailedToEmitSpans))
-		metrics.FailedToEmitSpans.Inc(delta(s.tooLargeDroppedSpans, stats.TooLargeDroppedSpans))
-		metrics.FailedToEmitSpans.Inc(delta(s.fullQueueDroppedSpans, stats.FullQueueDroppedSpans))
+		metrics.TooLargeDroppedSpans.Inc(delta(s.tooLargeDroppedSpans, stats.TooLargeDroppedSpans))
+		metrics.FullQueueDroppedSpans.Inc(delta(s.fullQueueDroppedSpans, stats.FullQueueDroppedSpans))
 
 		s.failedToEmitSpans = stats.FailedToEmitSpans
 		s.tooLargeDroppedSpans = stats.TooLargeDroppedSpans
--- a/cmd/agent/app/reporter/client_metrics.go
+++ b/cmd/agent/app/reporter/client_metrics.go
@@ -15,7 +15,6 @@
 package reporter
 
 import (
-	"math"
 	"sync"
 	"time"
 
@@ -28,11 +27,6 @@ import (
 )
 
 const (
-	// If client-reported counters wrap over MaxInt64, we can have old > new.
-	// We will "detect" the wrapping by checking that old is within the tolerance
-	// from MaxInt64 and new is within the tolerance from 0.
-	wrappedCounterTolerance = 10000000
-
 	defaultExpireFrequency = 15 * time.Minute
 	defaultExpireTTL       = time.Hour
 )
@@ -194,19 +188,19 @@ func (s *lastReceivedClientStats) update(
 	s.lock.Lock()
 	defer s.lock.Unlock()
 
-	if s.batchSeqNo >= batchSeqNo && !wrapped(s.batchSeqNo, batchSeqNo) {
+	if s.batchSeqNo >= batchSeqNo {
 		// Ignore out of order batches. Once we receive a batch with a larger-than-seen number,
 		// it will contain new cumulative counts, which will we use to update the metrics.
 		// That makes the metrics slightly off in time, but accurate in aggregate.
 		return
 	}
 
-	metrics.BatchesSent.Inc(delta(s.batchSeqNo, batchSeqNo))
+	metrics.BatchesSent.Inc(batchSeqNo - s.batchSeqNo)
 
 	if stats != nil {
-		metrics.FailedToEmitSpans.Inc(delta(s.failedToEmitSpans, stats.FailedToEmitSpans))
-		metrics.TooLargeDroppedSpans.Inc(delta(s.tooLargeDroppedSpans, stats.TooLargeDroppedSpans))
-		metrics.FullQueueDroppedSpans.Inc(delta(s.fullQueueDroppedSpans, stats.FullQueueDroppedSpans))
+		metrics.FailedToEmitSpans.Inc(stats.FailedToEmitSpans - s.failedToEmitSpans)
+		metrics.TooLargeDroppedSpans.Inc(stats.TooLargeDroppedSpans - s.tooLargeDroppedSpans)
+		metrics.FullQueueDroppedSpans.Inc(stats.FullQueueDroppedSpans - s.fullQueueDroppedSpans)
 
 		s.failedToEmitSpans = stats.FailedToEmitSpans
 		s.tooLargeDroppedSpans = stats.TooLargeDroppedSpans
@@ -215,17 +209,6 @@ func (s *lastReceivedClientStats) update(
 
 	s.lastUpdated = time.Now()
 	s.batchSeqNo = batchSeqNo
-}
-
-func wrapped(old int64, new int64) bool {
-	return (old > math.MaxInt64-wrappedCounterTolerance) && (new < wrappedCounterTolerance)
-}
-
-func delta(old int64, new int64) int64 {
-	if !wrapped(old, new) {
-		return new - old
-	}
-	return new + (math.MaxInt64 - old)
 }
 
 func clientUUID(batch *jaeger.Batch) string {
--- a/cmd/agent/app/reporter/client_metrics.go
+++ b/cmd/agent/app/reporter/client_metrics.go
@@ -20,6 +20,7 @@ import (
 	"time"
 
 	"github.com/uber/jaeger-lib/metrics"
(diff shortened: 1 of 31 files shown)


cmd/agent/app/reporter/client_metrics.go
```diff
@@ -0,0 +1,208 @@
...
+func (s *lastReceivedClientStats) update(
+	batchSeqNo int64,
+	stats *jaeger.ClientStats,
+	metrics *clientMetrics,
+) {
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	if s.batchSeqNo >= batchSeqNo && !wrapped(s.batchSeqNo, batchSeqNo) {
+		// ignore out of order batches, the metrics will be updated later
```
jpkrohling: The batch you are processing has a lower sequence number than a batch you've seen before, so, you aren't updating the metrics later, are you? I think you are just using the metrics from the highest known seq number instead.

cmd/agent/app/reporter/client_metrics.go
```diff
@@ -0,0 +1,208 @@
...
+	TooLargeDroppedSpans metrics.Counter `metric:"spans_dropped" tags:"cause=too-large"`
+
+	// Total count of spans dropped by clients because they failed Thrift encoding or submission.
+	FailedToEmitSpans metrics.Counter `metric:"spans_dropped" tags:"cause=send-failure"`
+}
+
+type lastReceivedClientStats struct {
+	lock                  sync.Mutex
+	lastUpdated           time.Time
+	batchSeqNo            int64
```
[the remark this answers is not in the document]
yurishkuro: no, but I wanted to keep the types in sync with Thrift values, which only supports signed i64.

cmd/agent/app/reporter/client_metrics.go
yurishkuro: I will have unit tests to work through the edge cases.

yurishkuro:
this is ready for review

cmd/agent/app/reporter/client_metrics_test.go
```diff
@@ -0,0 +1,270 @@
...
+				},
+			},
+			{
+				clientUUID: &clientUUID,
+				seqNo:      nPtr(101),
+				expCounters: []metricstest.ExpectedMetric{
+					{Name: prefix + "batches_sent", Value: 101},
+				},
+			},
+			{clientUUID: &clientUUID, seqNo: nPtr(90)},
```
jpkrohling: Can you add the expected counters/gauges to this test? They should be the same from the previous test (seq nr. 101).

cmd/agent/app/reporter/client_metrics_test.go
```diff
@@ -0,0 +1,270 @@
...
+}
+
+func TestClientMetricsReporter_Delta(t *testing.T) {
+	m := int64(math.MaxInt64)
+	tests := []struct {
+		old, new, delta int64
+	}{
+		{old: 5, new: 10, delta: 5},
+		{old: m - 10, new: m - 5, delta: 5},
+		{old: m - 5, new: 10, delta: 15},
```
jpkrohling: Just out of curiosity, do you expect a client to ever reach `math.MaxInt64` without restarts?

cmd/agent/app/reporter/client_metrics.go
```diff
@@ -0,0 +1,245 @@
...
+
+func (r *ClientMetricsReporter) updateClientMetrics(batch *jaeger.Batch) {
+	clientUUID := clientUUID(batch)
+	if clientUUID == "" {
+		return
+	}
+	if batch.SeqNo == nil {
+		return
+	}
+	entry, found := r.lastReceivedClientStats.Load(clientUUID)
```
jpkrohling: A client is always expected to generate a new UUID during the bootstrap, right? Those are not in any way stable IDs.

cmd/agent/app/reporter/client_metrics.go
```diff
@@ -0,0 +1,245 @@
...
+func (s *lastReceivedClientStats) update(
+	batchSeqNo int64,
+	stats *jaeger.ClientStats,
+	metrics *clientMetrics,
+) {
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	if s.batchSeqNo >= batchSeqNo && !wrapped(s.batchSeqNo, batchSeqNo) {
+		// Ignore out of order batches. Once we receive a batch with a larger-than-seen number,
```
jpkrohling: Would probably be a good idea to add a debug log statement here. While it might not bring much benefit on the usual case, it might be valuable when debugging bad clients (like, a client that always sends the same fixed UUID).

cmd/agent/app/reporter/client_metrics_test.go
```diff
@@ -0,0 +1,270 @@
...
+	testClientMetrics(func(tr *clientMetricsTest) {
+		assert.NoError(t, tr.r.EmitZipkinBatch([]*zipkincore.Span{{}}))
+		assert.Len(t, tr.mr.ZipkinSpans(), 1)
+	})
+}
+
+func TestClientMetricsReporter_Jaeger(t *testing.T) {
+	testClientMetrics(func(tr *clientMetricsTest) {
+		blank := ""
+		clientUUID := "foobar"
```
jpkrohling: Are the known clients supposed to send values compliant with the RFC-4122? If so, should the agent require the format to be compliant with that? I can't think of a good reason for a client to send a non-compliant value, other than configuration error...

cmd/agent/app/reporter/metrics_test.go
```diff
@@ -26,24 +26,12 @@ import (
...
-	return r.err
-}
-
 func TestMetricsReporter(t *testing.T) {
 	tests := []struct {
 		expectedCounters []metricstest.ExpectedMetric
 		expectedGauges   []metricstest.ExpectedMetric
 		action           func(reporter Reporter)
-		rep              *noopReporter
+		rep              *alwaysFailReporter
```
jpkrohling: Perhaps it needs a better name? :-)

yurishkuro:
> Once there's a PR ready to be tested on the client side, I'll run an end-to-end manual test.

@jpkrohling the working client version is already included here as dependency (https://github.com/jaegertracing/jaeger-client-go/pull/482), and the functionality can be tested with all-in-one (since query service is self-tracing).

cmd/agent/app/reporter/client_metrics_test.go
yurishkuro: Perhaps you're right, we could ignore that possibility and simplify the code. Napkin math:
  * A service with 10k QPS => 1,000 batches / sec
  * this is about 300,000,000 years to roll over 2^63

cmd/agent/app/reporter/client_metrics.go
yurishkuro: Yes, new client generates new ID. We don't need them to be stable across restarts.

cmd/agent/app/reporter/client_metrics.go
yurishkuro: Out of order batches are very normal when using UDP, because of unpredictability of the scheduler affecting how agent is processing packets. So I don't think logging it is warranted.

cmd/agent/app/reporter/client_metrics_test.go
yurishkuro: I don't think they are sending real UUIDs today (probably just random bits). We just need them to be reasonably unique. As far as this code is concerned, it's just some opaque string.

cmd/agent/app/reporter/metrics_test.go
yurishkuro: why? it's an existing mock struct that was re-implemented twice, I just consolidated them.

cmd/agent/app/reporter/client_metrics.go
jpkrohling: "We don't *support*" would be more appropriate: a client using a stable ID will send new sequence numbers, which will be ignored by the server until the sequence number is higher than a previous connection.

cmd/agent/app/reporter/client_metrics.go
jpkrohling: They are very normal, hence the debug-level logging. Someone trying to figure out why a specific client metric isn't showing up might benefit from a log message here, in case the client is not behaving appropriately. Or are there any costs in zap for debug-level messages when debug isn't enabled?

cmd/agent/app/reporter/client_metrics_test.go
jpkrohling: Alright. If we knew that all current clients were sending real UUIDs, we could assume that anything non-standard would be a misconfigured client.

cmd/agent/app/reporter/metrics_test.go
jpkrohling: `alwaysFailReporter` is misleading: people have to look at the implementation to realize that it won't actually cause any failures, that that the `TestMetricsReporter` isn't (only) testing reporters that fail...

cmd/agent/app/reporter/client_metrics.go
yurishkuro: A general rule for logging in high throughput systems is "don't log things that happen on every request". Out of order batches don't happen on every request, but can easily happen on 50% of requests at high throughput and concurrency. Logging will introduce both runtime overhead (not large, but not zero either), and if debug logging is actually enabled it will flood the logs. At the same time, at low throughput the logging is unlikely to be helpful since out of order batches are also unlikely there.

I'd be ok with adding another metrics counter for out of order batches (but it will be aggregate, not client-specific).

jpkrohling: Sometimes, I miss HotSpot JVM's cleverness ;-) Not sure a metric would help in this case. Let's leave it out and see if there are requests for it.

cmd/agent/app/reporter/client_metrics_test.go
yurishkuro: done

cmd/agent/app/reporter/metrics_test.go
yurishkuro: renamed to mockReporter

cmd/agent/app/reporter/tchannel/collector_proxy.go
```diff
@@ -25,22 +25,28 @@ import (
...
 // ProxyBuilder holds objects communicating with collector
 type ProxyBuilder struct {
-	reporter reporter.Reporter
+	reporter *reporter.ClientMetricsReporter
 	manager  configmanager.ClientConfigManager
 	tchanRep *Reporter
 }
 
 // NewCollectorProxy creates ProxyBuilder
 func NewCollectorProxy(builder *Builder, mFactory metrics.Factory, logger *zap.Logger) (*ProxyBuilder, error) {
```
guo0693: a little bit irrelevant to data loss, we don't need to correct it in this PR. But usually Builder.NewXXX means build a new XXX,  seems more reasonable for NewCollectorProxy to create a CollectorProxy instead of ProxyBuilder.

thrift-gen/jaeger/tchan-jaeger.go
```diff
@@ -43,10 +43,6 @@ func (c *tchanCollectorClient) SubmitBatches(ctx thrift.Context, batches []*Batc
 	}
 	success, err := c.client.Call(ctx, c.thriftService, "submitBatches", &args, &resp)
 	if err == nil && !success {
-		switch {
```
guo0693: Seems generated deletion looks weird. Why no logging?

thrift-gen/baggage/tchan-baggage.go
```diff
@@ -43,10 +43,6 @@ func (c *tchanBaggageRestrictionManagerClient) GetBaggageRestrictions(ctx thrift
 	}
 	success, err := c.client.Call(ctx, c.thriftService, "getBaggageRestrictions", &args, &resp)
 	if err == nil && !success {
-		switch {
```
guo0693: Why no logging?

cmd/agent/app/reporter/tchannel/collector_proxy.go
```diff
@@ -25,22 +25,28 @@ import (
...
+	r2 := reporter.WrapWithClientMetrics(reporter.ClientMetricsReporterParams{
+		Reporter:       r1,
+		Logger:         logger,
+		MetricsFactory: mFactory,
+	})
 	return &ProxyBuilder{
-		tchanRep: r,
-		reporter: reporter.WrapWithMetrics(r, tchannelMetrics),
-		manager:  configmanager.WrapWithMetrics(tchannel.NewConfigManager(r.CollectorServiceName(), r.Channel()), tchannelMetrics),
+		tchanRep: tchanRep,
```
guo0693: Why do we need to export tchanRep? Is tchanRep used any where? If not, we could remove tchanRep from below struct:
`type ProxyBuilder struct {
	reporter reporter.Reporter
	manager  configmanager.ClientConfigManager
	tchanRep *Reporter
}`

cmd/agent/app/reporter/tchannel/collector_proxy.go
yurishkuro: I agree, the naming here has become weird overtime.

cmd/agent/app/reporter/tchannel/collector_proxy.go
yurishkuro: Yes, it's unfortunate side effect of the way constructors are defined. I think the Sampling server depends on the underlying tchannel that is being created here.

thrift-gen/jaeger/tchan-jaeger.go
yurishkuro: I can't help it, it's how the recent generator works. But the new code looks correct - propagating the error up, rather than introducing that odd override.

Also, low-level libraries are not supposed to log.

cmd/agent/app/reporter/grpc/collector_proxy.go
```diff
@@ -39,9 +39,16 @@ func NewCollectorProxy(builder *ConnBuilder, agentTags map[string]string, mFacto
 		return nil, err
 	}
 	grpcMetrics := mFactory.Namespace(metrics.NSOptions{Name: "", Tags: map[string]string{"protocol": "grpc"}})
+	r1 := NewReporter(conn, agentTags, logger)
+	r2 := reporter.WrapWithMetrics(r1, grpcMetrics)
+	r3 := reporter.WrapWithClientMetrics(reporter.ClientMetricsReporterParams{
```
guo0693: Is it possible to have a generic WrappedReporter? Essentially r2 & r3 are wrapped reporter with callbacks around EmitBatch and EmitZipkinBatch. It will be great if we can have chained wrapper, something like:
NewReporter(...).wrapWith().wrapWith().

yurishkuro: - in what sense "generic"?

guo0693: I haven't thought it through, not sure if it's totally possible or not. Some rough idea as below:
type WrappedReporter interface {
	EmitZipkinBatch(spans []*zipkincore.Span) (err error)
	EmitBatch(batch *jaeger.Batch) (err error)
        WrapWith(func onEmitZipkinBatch, func onEmitBatch) (*WrappedReporter, error)
}
We may need to solve the problem whether the callback is called before or after EmitBatch, seems like metrics.go & client_metrics.go do those operations in different order, one before while the other one after.

yurishkuro: what benefits would this provide? The two existing decorators have specific things they are doing before or after calling the underlying, there's little room to generalize there. And WrapWith as part of the interface smells like an anti-pattern due to unnecessary coupling, i.e. each of the two decorators does not need to know if it itself is being decorated further.

guo0693: As I read the code, I felt confused when I first saw metrics.go and client_metrics.go, after spent some time reading, I understood they are reporters wrapped with different call backs.

Regarding detailed implementation of the callbacks, yes they will be very different and no need to be generalized. The generalization is for better readability and extendability I guess, it will be easy to wrap another callback around emit calls in the future instead of implement a new reader if necessary .

I agree, the decorators don't need to know if they will be decorated further or not. I just got the idea from [chained promise from javascript](https://javascript.info/promise-chaining), it will be good if we can support similar feature here. This way, you don't need:
```
r1 := NewReporter(conn, agentTags, logger)
r2 := reporter.WrapWithMetrics(r1, grpcMetrics)
r3 := reporter.WrapWithClientMetrics(...)
```
You just need:
```
NewReporter(...).wrapWith(//r2 functionalities).wrapWith(//r3 functinoalities) 
```

yurishkuro:
I think I addressed @jpkrohling comments, so merging.

PR #2013, opened 2020-01-12
WIP: Add memory-bound queue

yurishkuro:
As alternative to channel-based queue in #1985


pkg/queue/ring_buffer_test.go
```diff
@@ -0,0 +1,122 @@
...
+func TestRingBufferQueue_EnsureCapacity(t *testing.T) {
+	tests := []struct {
+		items    []int // positive to enqueue, negative to dequeue
+		newCap   int
+		expCap   int
+		expItems []int // content of the buffer after EnsureCapacity
+	}{
+		{items: []int{1, 2}, newCap: 4, expCap: 4, expItems: []int{1, 2, 0, 0}},
+		{items: []int{1, 2, 3, -1}, newCap: 4, expCap: 4, expItems: []int{2, 3, 0, 0}},
+		{items: []int{1, 2, 3, -1, -2, 4, 5}, newCap: 4, expCap: 4, expItems: []int{3, 4, 5, 0}},
```
jpkrohling: I'm getting something wrong... This is would I would expect:

```
1, 2, 3, 0 (initial state)
2, 3, 0, 0 (after -1)
0, 0, 0, 0 (after -2)
4, 5, 0, 0 (final state)
```

jpkrohling: Unless `-2` doesn't mean `take two`, but `take one, for the second time`

pkg/queue/ring_buffer.go
```diff
@@ -0,0 +1,95 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package queue
+
+type RingBufferQueue struct {
```
jpkrohling: This also require the consumer to take care of concurrency. Would also be good to have a documentation somewhere showing how the types relate to each other. Like: 

> ConcurrentMemoryBoundedQueue wraps MemoryBoundedQueue, which wraps RingBufferQueue

pkg/queue/mem_bounded_queue.go
```diff
@@ -0,0 +1,127 @@
+package queue
+
+import (
+	"sync"
+	"time"
+
+	"github.com/uber/jaeger-lib/metrics"
+)
+
+type Sizeable interface {
```
pavolloffay: Is there a core API to get the size? Could we use that instead? 

https://golang.org/src/go/types/sizes.go

pkg/queue/ring_buffer.go
pavolloffay: or we can make these package private and export just the concurrent version.

pkg/queue/ring_buffer.go
```diff
@@ -0,0 +1,95 @@
...
+
+func (q *RingBufferQueue) Full() bool {
+	return (q.tail+1)%q.size == q.head
+}
+
+func (q *RingBufferQueue) Enqueue(item interface{}) bool {
+	if q.Full() {
+		return false
+	}
+	if q.head == -1 {
```
pavolloffay: call to `Empty` is more idiomatic

yurishkuro:
@jpkrohling this was a quick prototype, if we think it's worth pursuing I can polish it. The basic ring buffer is ready/tested, but the other two types are only written, I never ran them yet.

jpkrohling:
> if we think it's worth pursuing I can polish it

Is it possible to benchmark it against the bounded queue that was recently changed? If the performance is similar, I think it's well worth pursuing it!

yurishkuro:
The new queue uses a single mutex, those are usually more efficient than Go channels. But the calls to Span.Size(), which are recursive, will introduce an extra overhead (although it will apply to both implementations). I am skeptical that the queue itself will be in any way a limiting factor in the collector's throughput, and micro-benchmarks won't reflect that anyway.

pavolloffay:
Could we have a benchmark for the concurrent version and compare it against the current impl? 

The POC looks good to me, if there is no perf degradation.

jpkrohling:
@yurishkuro will you continue working on this one? I closed the PR that adds the flags (#1985), but just realized that this PR here only adds a new queue type. Should I reopen that and work on the review comments, eventually switching to this new queue, or would you rather have me wait for this one here to be merged first?

yurishkuro:
If someone could take over this pr it would be great, as I will only have time to return to it over the weekend or next week.

The other PR you closed is still needed as it introduces the wiring for the new queue (cli flags, etc).

jpkrohling:
> The other PR you closed is still needed as it introduces the wiring for the new queue (cli flags, etc).

I'll reopen that and address the comments. If this here is merged before that one, I'll adapt that to use this new queue, otherwise, I'll send a follow-up PR once the new queue gets merged.

yurishkuro:
Benchmark results are puzzling: the mutex+cond version is about x3 slower than channels.
```
BenchmarkChannelQueue-12    	 1227919	       968 ns/op
BenchmarkRingBuffer-12      	  477895	      2522 ns/op
BenchmarkChannel-12         	 1796756	       651 ns/op
BenchmarkCond-12            	  557828	      2170 ns/op
```
I added the last two benchmarks which are just raw channel and Cond, without any queueing logic. If I remove the calls to Goshed(), then Cond benchmark becomes slightly faster than channel, although the full queue is still slowed (custom data structure understandably slower than built-in channels).
```
BenchmarkChannelQueue-12    	14336493	        72.7 ns/op
BenchmarkRingBuffer-12      	 3808519	       318 ns/op
BenchmarkChannel-12         	11800252	        96.9 ns/op
BenchmarkCond-12            	14106967	        84.7 ns/op
```
The CPU profiles are very similar (with yield), spending most time in the same Cond wait, yet channel is much faster:

channel:
image

Cond:
image

yurishkuro:
> does this mean the existing approach of using the ChannelQueue is preferred given the benchmark results?

That is precisely the open question. I don't think these benchmarks are conclusive, the methodology is flawed. It makes sense to me that plain Cond is faster than channels, considering that profile shows that channels are using the same primitives underneath. I cannot explain why adding yield makes Cond test 3 times slower. I used yield as a standby for a syscall, which the real collector would be doing in both producer and consumer, but real syscalls are likely to dominate the benchmark.

PR #2020, opened 2020-01-16
Changed default schema creation job

arpitjindal97:
By default, collection and query uses schema `jaeger_v1_test`, so schema creation job should also use this default value as schema and not `jaeger_v1_dc1`.

https://www.jaegertracing.io/docs/1.16/cli/

Here is the related discussion thread, where this change was done in operator.

https://github.com/jaegertracing/jaeger-operator/pull/837


plugin/storage/cassandra/schema/create.sh
```diff
@@ -41,7 +41,7 @@ else
     usage "invalid MODE=$MODE, expecting 'prod' or 'test'"
 fi
 
-keyspace=${KEYSPACE:-"jaeger_v1_${datacenter}"}
+keyspace=${KEYSPACE:-"jaeger_v1_test"}
```
yurishkuro: This will break the existing customization functionality.

plugin/storage/cassandra/schema/create.sh
```diff
@@ -10,7 +10,7 @@ function usage {
     >&2 echo "  DATACENTER         - datacenter name for network topology used in prod (optional in MODE=test)"
     >&2 echo "  TRACE_TTL          - time to live for trace data, in seconds (default: 172800, 2 days)"
     >&2 echo "  DEPENDENCIES_TTL   - time to live for dependencies data, in seconds (default: 0, no TTL)"
-    >&2 echo "  KEYSPACE           - keyspace (default: jaeger_v1_{datacenter})"
+    >&2 echo "  KEYSPACE           - keyspace (default: jaeger_v1_test)"
```
yurishkuro: it already defaults to test per L37.

arpitjindal97: Schema name shoud never be dependent on value of datacenter. I tried executing cassandra-schema-job container without giving any keyspace value. I expect it to create `jaeger_v1_test` because that's the default value used by collector and query

PR #2025, opened 2020-01-16
Allow raw port numbers for UDP servers

yurishkuro:
Currently flags like `--processor.jaeger-binary.server-host-port 5001` result in a failure. I think we can be more flexible since in most cases the host is not required, just the port.

### DIFF
--- a/cmd/agent/app/flags.go
+++ b/cmd/agent/app/flags.go
@@ -19,7 +19,6 @@ import (
 	"flag"
 	"fmt"
 	"strconv"
-	"strings"
 
 	"github.com/spf13/viper"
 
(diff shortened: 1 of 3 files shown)


cmd/agent/app/flags.go
```diff
@@ -66,10 +67,17 @@ func (b *Builder) InitFromViper(v *viper.Viper) *Builder {
...
+		p.Server.HostPort = toHostPort(v.GetString(prefix + suffixServerHostPort))
 		b.Processors = append(b.Processors, *p)
 	}
 
-	b.HTTPServer.HostPort = v.GetString(httpServerHostPort)
+	b.HTTPServer.HostPort = toHostPort(v.GetString(httpServerHostPort))
 	return b
 }
+
+func toHostPort(v string) string {
```
vprithvi: What happens if someone is running on a IPV6 host with the address of the form `2001:0db8:85a3:0000:0000:8a2e:0370:7334` ?

vprithvi: Perhaps we might check that the string starts with a `:` instead?

yurishkuro: so really I was trying to catch a raw number here, so I could instead try Long.parse()

PR #2031, opened 2020-01-24
Improved graceful shutdown - Agent

### LINKED ISSUE #295: Graceful shutdown by properly handling SIGTERM
From https://github.com/uber/jaeger/pull/280 :

> What I think should happen is you tell the health endpoint to start returning Unavailable and then sleep for a couple seconds, and the exit. The main tchannel / http servers will remain in operation during the sleep and allowed to finish service the in-flight requests, while the routing mesh might hopefully detect health check down and stop sending traffic.

jpkrohling:
- Resolves #295;
- The current code seems to go in the direction of graceful shutdown but fails to correctly implement this behavior for all components. For instance, before this PR, the gRPC server would be closed without taking in-flight requests into consideration;
- More thoughtful shutdown order, like in the all-in-one command: the agent is now closed before the collector disappears, preventing messages like the following after `SIGTERM`:

```
WARNING: 2020/02/18 13:27:44 grpc: addrConn.createTransport failed to connect to {127.0.0.1:14250  <nil> 0 <nil>}. Err :connection error: desc = "transport: Error while dialing dial tcp 127.0.0.1:14250: connect: connection refused". Reconnecting...
```

- The first commit shows what would be necessary to get a graceful shutdown for the collector
- The second commit moves most of the startup/shutdown logic for the collector into a "component". I'm not happy with its current state, especially about its location, but it shows that quite a good amount of code could be removed by adopting this idea. The same could be applied to the other components, but I want to run this idea here first, before spending much time on this refactoring.

### DIFF
--- a/cmd/agent/app/agent.go
+++ b/cmd/agent/app/agent.go
@@ -16,10 +16,11 @@
 package app
 
 import (
-	"io"
+	"context"
 	"net"
 	"net/http"
 	"sync/atomic"
+	"time"
 
 	"github.com/gorilla/mux"
 	"go.uber.org/zap"
@@ -33,7 +34,6 @@ type Agent struct {
 	httpServer *http.Server
 	httpAddr   atomic.Value // string, set once agent starts listening
 	logger     *zap.Logger
-	closer     io.Closer
 }
 
 // NewAgent creates the new Agent.
@@ -65,10 +65,9 @@ func (a *Agent) Run() error {
 		return err
 	}
 	a.httpAddr.Store(listener.Addr().String())
-	a.closer = listener
 	go func() {
 		a.logger.Info("Starting jaeger-agent HTTP server", zap.Int("http-port", listener.Addr().(*net.TCPAddr).Port))
-		if err := a.httpServer.Serve(listener); err != nil {
+		if err := a.httpServer.Serve(listener); err != http.ErrServerClosed {
 			a.logger.Error("http server failure", zap.Error(err))
 		}
 		a.logger.Info("agent's http server exiting")
@@ -86,8 +85,16 @@ func (a *Agent) HTTPAddr() string {
 
 // Stop forces all agent go routines to exit.
 func (a *Agent) Stop() {
-	for _, processor := range a.processors {
-		go processor.Stop()
+	// first, close the http server, so that we don't have any more inflight requests
+	a.logger.Info("shutting down agent's HTTP server", zap.String("addr", a.HTTPAddr()))
+	timeout, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	if err := a.httpServer.Shutdown(timeout); err != nil {
+		a.logger.Error("failed to close HTTP server", zap.Error(err))
 	}
-	a.closer.Close()
+	cancel()
+
+	// then, close all processors that are called for the incoming http requests
+	for _, processor := range a.processors {
+		processor.Stop()
+	}
 }
--- a/cmd/agent/app/builder.go
+++ b/cmd/agent/app/builder.go
@@ -68,7 +68,6 @@ var (
 type CollectorProxy interface {
 	GetReporter() reporter.Reporter
 	GetManager() configmanager.ClientConfigManager
-	Close() error
 }
 
 // Builder Struct to hold configurations
--- a/cmd/agent/app/builder.go
+++ b/cmd/agent/app/builder.go
@@ -68,6 +68,7 @@ var (
 type CollectorProxy interface {
 	GetReporter() reporter.Reporter
 	GetManager() configmanager.ClientConfigManager
+	Close() error
 }
 
 // Builder Struct to hold configurations
--- a/cmd/agent/app/builder.go
+++ b/cmd/agent/app/builder.go
@@ -17,6 +17,7 @@ package app
 
 import (
 	"fmt"
+	"io"
 	"net/http"
 	"strconv"
 
@@ -68,6 +69,7 @@ var (
 type CollectorProxy interface {
 	GetReporter() reporter.Reporter
 	GetManager() configmanager.ClientConfigManager
+	io.Closer
 }
 
 // Builder Struct to hold configurations
--- a/cmd/agent/app/reporter/grpc/collector_proxy.go
+++ b/cmd/agent/app/reporter/grpc/collector_proxy.go
@@ -22,7 +22,6 @@ import (
 	"github.com/jaegertracing/jaeger/cmd/agent/app/configmanager"
 	grpcManager "github.com/jaegertracing/jaeger/cmd/agent/app/configmanager/grpc"
 	"github.com/jaegertracing/jaeger/cmd/agent/app/reporter"
-	"github.com/jaegertracing/jaeger/pkg/multierror"
 )
 
 // ProxyBuilder holds objects communicating with collector
@@ -70,11 +69,5 @@ func (b ProxyBuilder) GetManager() configmanager.ClientConfigManager {
 
 // Close closes connections used by proxy.
 func (b ProxyBuilder) Close() error {
-	var errs []error
-
-	if err := b.conn.Close(); err != nil {
-		errs = append(errs, err)
-	}
-
-	return multierror.Wrap(errs)
+	return b.conn.Close()
 }
--- a/cmd/agent/app/reporter/grpc/collector_proxy.go
+++ b/cmd/agent/app/reporter/grpc/collector_proxy.go
@@ -22,6 +22,7 @@ import (
 	"github.com/jaegertracing/jaeger/cmd/agent/app/configmanager"
 	grpcManager "github.com/jaegertracing/jaeger/cmd/agent/app/configmanager/grpc"
 	"github.com/jaegertracing/jaeger/cmd/agent/app/reporter"
+	"github.com/jaegertracing/jaeger/pkg/multierror"
 )
 
 // ProxyBuilder holds objects communicating with collector
@@ -69,6 +70,15 @@ func (b ProxyBuilder) GetManager() configmanager.ClientConfigManager {
 
 // Close closes connections used by proxy.
 func (b ProxyBuilder) Close() error {
-	b.reporter.Close()
-	return b.conn.Close()
+	var errs []error
+
+	if err := b.conn.Close(); err != nil {
+		errs = append(errs, err)
+	}
+
+	if err := b.reporter.Close(); err != nil {
+		errs = append(errs, err)
+	}
+
+	return multierror.Wrap(errs)
 }
--- a/cmd/agent/app/reporter/grpc/collector_proxy.go
+++ b/cmd/agent/app/reporter/grpc/collector_proxy.go
@@ -76,9 +76,5 @@ func (b ProxyBuilder) Close() error {
 		errs = append(errs, err)
 	}
 
-	if err := b.reporter.Close(); err != nil {
-		errs = append(errs, err)
-	}
-
 	return multierror.Wrap(errs)
 }
--- a/cmd/agent/app/reporter/grpc/collector_proxy.go
+++ b/cmd/agent/app/reporter/grpc/collector_proxy.go
@@ -69,5 +69,6 @@ func (b ProxyBuilder) GetManager() configmanager.ClientConfigManager {
 
 // Close closes connections used by proxy.
 func (b ProxyBuilder) Close() error {
+	b.reporter.Close()
 	return b.conn.Close()
 }
--- a/cmd/agent/app/reporter/grpc/reporter.go
+++ b/cmd/agent/app/reporter/grpc/reporter.go
@@ -65,6 +65,11 @@ func (r *Reporter) EmitZipkinBatch(zSpans []*zipkincore.Span) error {
 	return r.send(trace.Spans, nil)
 }
 
+// Close the reporter
+func (r *Reporter) Close() error {
+	return nil
+}
+
 func (r *Reporter) send(spans []*model.Span, process *model.Process) error {
 	spans, process = addProcessTags(spans, process, r.agentTags)
 	batch := model.Batch{Spans: spans, Process: process}
--- a/cmd/agent/app/reporter/grpc/reporter.go
+++ b/cmd/agent/app/reporter/grpc/reporter.go
@@ -65,11 +65,6 @@ func (r *Reporter) EmitZipkinBatch(zSpans []*zipkincore.Span) error {
 	return r.send(trace.Spans, nil)
 }
 
-// Close the reporter
-func (r *Reporter) Close() error {
-	return nil
-}
-
 func (r *Reporter) send(spans []*model.Span, process *model.Process) error {
 	spans, process = addProcessTags(spans, process, r.agentTags)
 	batch := model.Batch{Spans: spans, Process: process}
--- a/cmd/agent/app/reporter/reporter.go
+++ b/cmd/agent/app/reporter/reporter.go
@@ -26,6 +26,7 @@ import (
 type Reporter interface {
 	EmitZipkinBatch(spans []*zipkincore.Span) (err error)
 	EmitBatch(batch *jaeger.Batch) (err error)
+	Close() error
 }
 
 // MultiReporter provides serial span emission to one or more reporters.  If
@@ -55,6 +56,17 @@ func (mr MultiReporter) EmitBatch(batch *jaeger.Batch) error {
 	var errors []error
 	for _, rep := range mr {
 		if err := rep.EmitBatch(batch); err != nil {
+			errors = append(errors, err)
+		}
+	}
+	return multierror.Wrap(errors)
+}
+
+// Close all the underlying reporters
+func (mr MultiReporter) Close() error {
+	var errors []error
+	for _, rep := range mr {
+		if err := rep.Close(); err != nil {
 			errors = append(errors, err)
 		}
 	}
--- a/cmd/agent/app/reporter/reporter.go
+++ b/cmd/agent/app/reporter/reporter.go
@@ -26,7 +26,6 @@ import (
 type Reporter interface {
 	EmitZipkinBatch(spans []*zipkincore.Span) (err error)
 	EmitBatch(batch *jaeger.Batch) (err error)
-	Close() error
 }
 
 // MultiReporter provides serial span emission to one or more reporters.  If
@@ -56,17 +55,6 @@ func (mr MultiReporter) EmitBatch(batch *jaeger.Batch) error {
 	var errors []error
 	for _, rep := range mr {
 		if err := rep.EmitBatch(batch); err != nil {
-			errors = append(errors, err)
-		}
-	}
-	return multierror.Wrap(errors)
-}
-
-// Close all the underlying reporters
-func (mr MultiReporter) Close() error {
-	var errors []error
-	for _, rep := range mr {
-		if err := rep.Close(); err != nil {
 			errors = append(errors, err)
 		}
 	}
--- a/cmd/agent/app/reporter/tchannel/reporter.go
+++ b/cmd/agent/app/reporter/tchannel/reporter.go
@@ -65,6 +65,11 @@ func (r *Reporter) Channel() *tchannel.Channel {
 	return r.channel
 }
 
+// Close the underlying channel
+func (r *Reporter) Close() error {
+	return nil
+}
+
 // EmitZipkinBatch implements EmitZipkinBatch() of Reporter
 func (r *Reporter) EmitZipkinBatch(spans []*zipkincore.Span) error {
 	submissionFunc := func(ctx thrift.Context) error {
--- a/cmd/agent/app/reporter/tchannel/reporter.go
+++ b/cmd/agent/app/reporter/tchannel/reporter.go
@@ -65,11 +65,6 @@ func (r *Reporter) Channel() *tchannel.Channel {
 	return r.channel
 }
 
-// Close the underlying channel
-func (r *Reporter) Close() error {
-	return nil
-}
-
 // EmitZipkinBatch implements EmitZipkinBatch() of Reporter
 func (r *Reporter) EmitZipkinBatch(spans []*zipkincore.Span) error {
 	submissionFunc := func(ctx thrift.Context) error {
--- a/cmd/all-in-one/main.go
+++ b/cmd/all-in-one/main.go
@@ -120,6 +120,7 @@ by default uses only in-memory database.`,
 			cOpts := new(collectorApp.CollectorOptions).InitFromViper(v)
 			qOpts := new(queryApp.QueryOptions).InitFromViper(v, logger)
 
+			// collector
 			c := collectorApp.New(&collectorApp.CollectorParams{
 				ServiceName:    "jaeger-collector",
 				Logger:         logger,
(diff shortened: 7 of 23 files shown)


pkg/collector/component.go
```diff
@@ -0,0 +1,277 @@
...
+	"github.com/jaegertracing/jaeger/storage/spanstore"
+	jc "github.com/jaegertracing/jaeger/thrift-gen/jaeger"
+	sc "github.com/jaegertracing/jaeger/thrift-gen/sampling"
+	zc "github.com/jaegertracing/jaeger/thrift-gen/zipkincore"
+)
+
+var _ (io.Closer) = (*Component)(nil)
+
+// Component returns the collector as a manageable unit of work
+type Component struct {
```
yurishkuro: I would call this `collector.Collector`. Calling it `collector.Component` makes it unnecessarily vague.

pkg/collector/component.go
```diff
@@ -0,0 +1,277 @@
...
+	"github.com/jaegertracing/jaeger/pkg/recoveryhandler"
+	ss "github.com/jaegertracing/jaeger/plugin/sampling/strategystore"
+	"github.com/jaegertracing/jaeger/plugin/storage"
+	"github.com/jaegertracing/jaeger/storage/spanstore"
+	jc "github.com/jaegertracing/jaeger/thrift-gen/jaeger"
+	sc "github.com/jaegertracing/jaeger/thrift-gen/sampling"
+	zc "github.com/jaegertracing/jaeger/thrift-gen/zipkincore"
+)
+
+var _ (io.Closer) = (*Component)(nil)
```
yurishkuro: this should go to test file

pkg/collector/component.go
```diff
@@ -0,0 +1,277 @@
...
+
+// Component returns the collector as a manageable unit of work
+type Component struct {
+	// required to start a new collector
+	serviceName          string
+	logger               *zap.Logger
+	vp                   *viper.Viper
```
yurishkuro: I think this should not be here. It's better to follow the already established pattern:
  * instantiate component
  * register flags with Viper
  * component.InitFromViper()
  * component.Start()

pkg/collector/component.go
```diff
@@ -0,0 +1,277 @@
...
+	hServer    *http.Server
+	zkServer   *http.Server
+	grpcServer *grpc.Server
+	tchServer  *tchannel.Channel
+	spanWriter spanstore.Writer
+	options    *builder.CollectorOptions
+}
+
+// New constructs a new collector component, ready to be started
+func New(serviceName string, logger *zap.Logger, vp *viper.Viper, mf metrics.Factory, sf *storage.Factory, strF *ss.Factory, hCheck *healthcheck.HealthCheck) *Component {
```
yurishkuro: please create a struct CollectorParams instead of using long list of args

pkg/collector/component.go
```diff
@@ -0,0 +1,277 @@
...
+	if err := c.strategyStoreFactory.Initialize(c.metricsFactory, c.logger); err != nil {
+		c.logger.Fatal("Failed to init sampling strategy store factory", zap.Error(err))
+	}
+	strategyStore, err := c.strategyStoreFactory.CreateStrategyStore()
+	if err != nil {
+		c.logger.Fatal("Failed to create sampling strategy store", zap.Error(err))
+	}
+
+	c.tchServer, err = tchannel.NewChannel(c.serviceName, &tchannel.ChannelOptions{})
+	{
```
yurishkuro: this could use a function

pkg/collector/component.go
```diff
@@ -0,0 +1,277 @@
...
+		c.logger.Warn("TChannel has been deprecated and will be removed in a future release")
+		c.tchServer.Serve(listener)
+	}
+
+	c.grpcServer, err = startGRPCServer(builderOpts, grpcHandler, strategyStore, c.logger)
+	if err != nil {
+		c.logger.Fatal("Could not start gRPC collector", zap.Error(err))
+	}
+
+	{
```
yurishkuro: function?

pkg/collector/component.go
```diff
@@ -0,0 +1,277 @@
...
+	if c.zkServer != nil {
+		timeout, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		err := c.zkServer.Shutdown(timeout)
+		if err != nil {
+			c.logger.Error("Failed to stop the Zipkin server", zap.Error(err))
+		}
+		defer cancel()
+	}
+
+	// by now, we shouldn't have any in-flight requests anymore
```
yurishkuro: Could there still be items in the collector queue? Or is it encapsulated behind the "writer" notion?

pkg/collector/component.go
```diff
@@ -0,0 +1,277 @@
...
+
+	return nil
+}
+
+// Options returns the option set that was built for this collector
+func (c *Component) Options() *builder.CollectorOptions {
+	return c.options
+}
+
+func startGRPCServer(
```
yurishkuro: I think these functions can now have the receiver and take fewer parameters

pkg/collector/component.go
jpkrohling: What about the package? I think `pkg/collector` is really wrong, as this depends on the `cmd/collector/builder` and `cmd/collector/app` packages, but I don't have a good idea for a package name for this...

pkg/collector/component.go
jpkrohling: Yeah, I need to clean this up a bit, as I just wanted to do a quick PoC and see check the interest in it. Same applies to most of the other comments in this PR at this state, which is why I'm marking them as "resolved".

pkg/collector/component.go
jpkrohling: Great question, I'll make sure to address this.

pkg/collector/component.go
yurishkuro: We can put it directly under cmd/collector/app, similar to Agent type.

jpkrohling: Alright, will give it a try.

jpkrohling:
Status: this is taking quite some time, as it requires moving other things around as well, to avoid cyclic dependencies between packages. I think I'll split this into two tasks: the graceful shutdown for the collector, and the refactoring for the main.go files.

cmd/agent/app/agent.go
```diff
@@ -86,8 +89,22 @@ func (a *Agent) HTTPAddr() string {
 
 // Stop forces all agent go routines to exit.
 func (a *Agent) Stop() {
+	// first, close the http server, so that we don't have any more inflight requests
+	timeout, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	if err := a.httpServer.Shutdown(timeout); err != nil {
+		a.logger.Error("failed to close HTTP server", zap.Error(err))
+	}
+	defer cancel()
```
yurishkuro:
```suggestion
	cancel()
```
no point deferring

cmd/agent/app/agent.go
```diff
@@ -86,8 +89,22 @@ func (a *Agent) HTTPAddr() string {
...
+
+	// then, close all processors that are called by the incoming http requests
 	for _, processor := range a.processors {
-		go processor.Stop()
+		processor.Stop()
+	}
+
+	// and the reporter that was being used by the processor
+	if a.reporter != nil {
```
yurishkuro: is nil check necessary?

cmd/agent/app/reporter/grpc/reporter.go
```diff
@@ -32,6 +32,7 @@ import (
 
 // Reporter reports data to collector over gRPC.
 type Reporter struct {
+	conn      *grpc.ClientConn
```
yurishkuro: I am not sure about this. If Reporter fully owns the grpc connection, then there is no need to pass it externally, it can create the connection itself. Since we're passing it externally (perhaps because it is shared for another functionality, like running both span reporting and sampling strategies via the same connection), then reporter should not be in charge of closing the connection, it should be done by whoever created the connection.

cmd/agent/app/reporter/tchannel/reporter.go
```diff
@@ -65,6 +65,12 @@ func (r *Reporter) Channel() *tchannel.Channel {
 	return r.channel
 }
 
+// Close the underlying channel
+func (r *Reporter) Close() error {
+	r.channel.Close()
```
yurishkuro: same comment as for grpc connection - the reporter is not the only user of the channel, so it cannot be in charge of closing it.

cmd/collector/app/handler/tchannel_handler_test.go
```diff
@@ -27,12 +28,18 @@ var (
 	// verify API compliance
 	_ jaeger.TChanCollector           = new(TChannelHandler)
 	_ zipkincore.TChanZipkinCollector = new(TChannelHandler)
```
yurishkuro: change these two to use the same syntax

cmd/agent/app/agent.go
jpkrohling: Not in this iteration. Previously, I got a null pointer because I forgot to set the reporter and thought it would be a good idea to not panic in case the agent's reporter wasn't set. I'm removing it though.

cmd/agent/app/reporter/grpc/reporter.go
jpkrohling: This was also something I wasn't happy. The other choice would be to store the proxy as a property of the agent. On a second thought, it does make sense to have it there, so that the agent can close this connection once it's stopped.

cmd/collector/app/handler/thrift_span_handler.go
```diff
@@ -131,6 +137,10 @@ func (h *zipkinSpanHandler) SubmitZipkinBatch(spans []*zipkincore.Span, options
 	return responses, nil
 }
 
+func (h *zipkinSpanHandler) Close() error {
+	return h.modelProcessor.Close()
```
yurishkuro: This looks like we might be closing this twice? It's the ownership issue again - handlers don't create processors, so they should not be closing them.

jpkrohling: You are absolutely right. This wasn't a problem because the processor's `#Close()` was checking whether it had been closed before. I kept this safety check, but extracted the building of the span processor into its own function. The collector now builds it explicitly and passes to the handler builder.

While I don't quite like that the collector calls `builder.BuildProcessor()` and then `builder.BuildHandlers(processor)`, I think it's the best option that makes sense without building another abstraction with its own `#Close()`.

yurishkuro: if Collector builds the processor explicitly, shouldn't it be in charge of closing it as well, rather than having the handlers do it?

jpkrohling: ...

What if I told you that, after all that I wrote before, I say that I actually forgot to remove it from the handler's `#Close()` methods...

Hopefully, it's now fixed in this last commit.

cmd/agent/app/agent.go
```diff
@@ -86,8 +88,20 @@ func (a *Agent) HTTPAddr() string {
...
 // Stop forces all agent go routines to exit.
 func (a *Agent) Stop() {
+	// first, close the http server, so that we don't have any more inflight requests
+	timeout, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	if err := a.httpServer.Shutdown(timeout); err != nil {
+		a.logger.Error("failed to close HTTP server", zap.Error(err))
+	}
+	cancel()
+
+	// then, close all processors that are called by the incoming http requests
```
yurishkuro:
```suggestion
	// then, close all processors that are called for the incoming span batches
```

yurishkuro: processors don't work off HTTP requests

cmd/agent/app/agent.go
```diff
@@ -86,8 +88,20 @@ func (a *Agent) HTTPAddr() string {
 
 // Stop forces all agent go routines to exit.
 func (a *Agent) Stop() {
+	// first, close the http server, so that we don't have any more inflight requests
```
yurishkuro:
```suggestion
	// first, close the http server, so that we don't have any more inflight requests
	a.logger.Info("shutting down agent's HTTP server", zap.String("addr", a.httpAddr.Load()))
```

cmd/agent/app/agent.go
```diff
@@ -29,20 +30,22 @@ import (
...
 	processors []processors.Processor
 	httpServer *http.Server
 	httpAddr   atomic.Value // string, set once agent starts listening
 	logger     *zap.Logger
-	closer     io.Closer
 }
 
 // NewAgent creates the new Agent.
 func NewAgent(
+	proxy CollectorProxy,
```
yurishkuro: Why do we need Agent to be responsible for closing CollectorProxy if it didn't create it? There may be other services using that connection beyond the main agent's services.

cmd/agent/app/builder.go
```diff
@@ -69,6 +69,7 @@ var (
 type CollectorProxy interface {
 	GetReporter() reporter.Reporter
 	GetManager() configmanager.ClientConfigManager
+	Close() error
```
yurishkuro: Pending decision on whether agent should be in charge of closing the proxy. If not, then this Close() method may be unnecessary since whoever created the proxy would have access to `Close()` on the concrete type.

cmd/agent/app/reporter/grpc/collector_proxy.go
```diff
@@ -23,6 +23,7 @@ import (
 	grpcManager "github.com/jaegertracing/jaeger/cmd/agent/app/configmanager/grpc"
 	"github.com/jaegertracing/jaeger/cmd/agent/app/reporter"
 	aReporter "github.com/jaegertracing/jaeger/cmd/agent/app/reporter"
+	"github.com/jaegertracing/jaeger/pkg/multierror"
 )
 
 // ProxyBuilder holds objects communicating with collector
```
yurishkuro: let's rename this to CollectorProxy (but ok to do in another PR)

cmd/collector/app/collector.go
```diff
@@ -169,7 +177,18 @@ func (c *Collector) Close() error {
...
 
-	// by now, we shouldn't have any in-flight requests anymore
+	// by now, we shouldn't have any in-flight requests anymore, close the processor and handlers
+	for _, closer := range []io.Closer{c.spanProcessor, c.zipkinSpansHandler, c.jaegerBatchesHandler, c.grpcHandler} {
+		if closer == nil {
+			continue
+		}
+
+		if err := closer.Close(); err != nil {
+			c.logger.Error("Failed to close span handler.", zap.Error(err))
```
yurishkuro: * start with lowercase
* first item in the list is technically not a span handler

cmd/collector/app/collector.go
```diff
@@ -169,7 +177,18 @@ func (c *Collector) Close() error {
 		defer cancel()
 	}
 
-	// by now, we shouldn't have any in-flight requests anymore
+	// by now, we shouldn't have any in-flight requests anymore, close the processor and handlers
+	for _, closer := range []io.Closer{c.spanProcessor, c.zipkinSpansHandler, c.jaegerBatchesHandler, c.grpcHandler} {
```
yurishkuro: In all three span handlers the Close is no-op. What's the point of adding it?

cmd/collector/app/span_handler_builder.go
```diff
@@ -58,6 +55,15 @@ func (b *SpanHandlerBuilder) BuildHandlers() (
 		Options.DynQueueSizeMemory(b.CollectorOpts.DynQueueSizeMemory),
 	)
 
+}
+
+// BuildHandlers builds span handlers (Zipkin, Jaeger)
+func (b *SpanHandlerBuilder) BuildHandlers(spanProcessor processor.SpanProcessor) (
+	handler.ZipkinSpansHandler,
+	handler.JaegerBatchesHandler,
+	*handler.GRPCHandler,
```
yurishkuro: we may want to consider returning a single `*struct SpanHandlers` with 3 handlers

cmd/collector/app/span_processor.go
```diff
@@ -126,10 +128,18 @@ func newSpanProcessor(spanWriter spanstore.Writer, opts ...Option) *spanProcesso
...
-// Stop halts the span processor and all its go-routines.
+// Stop is an alias to #Close(). Deprecated.
 func (sp *spanProcessor) Stop() {
-	close(sp.stopCh)
-	sp.queue.Stop()
+	sp.Close()
+}
+
+func (sp *spanProcessor) Close() error {
+	if sp.stopped.CAS(false, true) {
```
yurishkuro: Usually this kind of double-close protection is a code smell for a problem elsewhere (such as bad ownership). Do we really need it?

cmd/collector/app/span_processor.go
```diff
@@ -126,10 +128,18 @@ func newSpanProcessor(spanWriter spanstore.Writer, opts ...Option) *spanProcesso
 	return &sp
 }
 
-// Stop halts the span processor and all its go-routines.
+// Stop is an alias to #Close(). Deprecated.
 func (sp *spanProcessor) Stop() {
```
yurishkuro: let's remove it. It's a method on a private receiver anyway, not part of a public API.

yurishkuro:
> The current code seems to go in the direction of graceful shutdown but fails to correctly implement this behavior for all components

Could you improve the description by explaining what was wrong with the current state?

cmd/agent/app/agent.go
jpkrohling: True. It makes the all-in-one a bit more complicated for now, but it is indeed the right thing to do.

cmd/agent/app/builder.go
jpkrohling: What's the cost of leaving this here? Having it as part of the interface forces new implementations to at least think about the `#Close()` operation.

cmd/collector/app/collector.go
jpkrohling: While the case is strong for removing it, as we have three implementations and none needs to close anything, this seems like the type of thing that would require a closer in some implementations. 

I can remove it, but the message that it sends is stronger than the fact that it's currently no-op.

jpkrohling:
> Could you improve the description by explaining what was wrong with the current state?

Done!

jpkrohling:
> E.g. git checkout master cmd/collector

TIL! This PR has been updated to include only the agent changes. I also squashed the commits, as the original commit set doesn't make much sense anymore.

cmd/collector/app/collector.go
yurishkuro: The message is sends is that these handlers might be stateful in some way and require Close to clean-up, which is kind of the opposite of what they are now: stateless proxies that simply forward to collector (while not managing the life cycle of collector proxy). I am in favor of simplicity, rather than adding code that might possibly be needed in the future.

It also complicates our internal build since now we also have to call close on these handlers, to respect their API.

cmd/agent/app/builder.go
yurishkuro: I think having Close() in an interface is pretty often an anti-pattern that indicates a problem elsewhere. But in some cases it is appropriate when the object that holds reference to this type is indeed requires an interface, because it either deals with different implementations or a decorator pattern is used that requires the decorator to accept an interface. In this concrete case I don't have a strong opinion, aside from my general preference of not introducing things until they are actually necessary.

yurishkuro: An example of the "cost" is `cmd/agent/app/builder_test.go` which now needs to add no-op Close().

cmd/agent/app/reporter/reporter.go
```diff
@@ -26,6 +26,7 @@ import (
 type Reporter interface {
 	EmitZipkinBatch(spans []*zipkincore.Span) (err error)
 	EmitBatch(batch *jaeger.Batch) (err error)
+	Close() error
```
yurishkuro: Another example that causes cascading changes in this PR without any apparent benefits, since the two main reporters (grpc and tchannel) have no-op Close().

cmd/all-in-one/main.go
```diff
@@ -130,14 +131,25 @@ by default uses only in-memory database.`,
...
 
-			startAgent(aOpts, repOpts, tchanBuilder, grpcBuilder, cOpts.CollectorGRPCPort, logger, metricsFactory)
+			// agent
+			grpcBuilder.CollectorHostPorts = append(grpcBuilder.CollectorHostPorts, fmt.Sprintf("127.0.0.1:%d", cOpts.CollectorGRPCPort))
+			agentMetricsFactory := metricsFactory.Namespace(metrics.NSOptions{Name: "agent", Tags: nil})
+			cp, err := agentApp.CreateCollectorProxy(repOpts, tchanBuilder, grpcBuilder, logger, agentMetricsFactory)
+			if err != nil {
+				logger.Fatal("Could not create collector proxy", zap.Error(err))
+			}
+			agent := startAgent(cp, aOpts, logger, metricsFactory)
```
yurishkuro: If we had a notion of life cycle (e.g. https://godoc.org/go.uber.org/fx#Lifecycle), this code could've been moved to a shared component, because all the individual objects created here are _only_ exposed to be able to call Close() on some of them.  Our internal builds using Fx extensively, making `main()` 10-lines long, while having all initialization code fully testable (unlike main). But DI in Go is a tricky proposition (e.g. this pearl: https://github.com/uber-go/fx/issues/694).

We don't have to do full DI though, the `svc.RunAndThen` below is already a form of lifecycle management. We could extend it with just OnStart/OnStop hooks.

jpkrohling: DI is indeed one of the things I miss from Java. Let's see how the code here looks in the end, and discuss adopting a better pattern for all components. This might bring benefits if we decide to make the components more easily consumable as a package (like, from [jaeger-opentelemetry-collector](https://github.com/jaegertracing/jaeger-opentelemetry-collector))

cmd/agent/app/builder.go
jpkrohling: Alright, I'll remove it, we can add it back if there's a need in the future.

jpkrohling: I'm adding this back, because both implementations actually create a connection on their own and need to properly close it. The only implementation that does not need a `#Close` is the test.

cmd/agent/app/reporter/grpc/collector_proxy.go
```diff
@@ -69,6 +70,11 @@ func (b ProxyBuilder) GetManager() configmanager.ClientConfigManager {
...
 func (b ProxyBuilder) Close() error {
-	b.reporter.Close()
-	return b.conn.Close()
+	var errs []error
+
+	if err := b.conn.Close(); err != nil {
+		errs = append(errs, err)
+	}
+
+	return multierror.Wrap(errs)
```
yurishkuro: there's only one object being closed here, why multierr?

yurishkuro: there's only one object being closed here, why multierr?

jpkrohling: Oops. Fixed.

cmd/agent/app/reporter/grpc/collector_proxy.go
```diff
@@ -69,6 +69,5 @@ func (b ProxyBuilder) GetManager() configmanager.ClientConfigManager {
 
 // Close closes connections used by proxy.
 func (b ProxyBuilder) Close() error {
-	b.reporter.Close()
```
yurishkuro: why are we removing this?

jpkrohling: See, this is why I think reporters and handlers should be `io.Closer` as well even if most are no-op. We argued in a previous iteration that reporters are stateless and need no closing, but then, we need to remember to close this one specific reporter, which we only know that needs closing because a test fails in CI (passes locally, for some strange reason)... Without checking the struct's definition, we would even think that this could be any implementation of `Reporter`, but it's actually tied to a concrete reporter (`ClientMetricsReporter`).

Anyway, I restored this line.

jpkrohling:
@yurishkuro is this ready to be merged, once 1.17 is out?

PR #2040, opened 2020-01-27
Add support for dynamically reloading sampling strategies

### LINKED ISSUE #1058: collector: reload sampling strategies on file content change

It should be possible to change the sampling strategies without restarting the collector.

I modified the sampling strategies file, only to find that it had no impact.

The collector should ideally watch the file for changes but it's also okay to just periodically re-read the file and figure out if anything changed.

The meta issue for this is here: https://github.com/jaegertracing/jaeger/issues/355 but I think for now we should reload the sampling-strategies on change.

annanay25:
- Add support for dynamically reloading sampling strategies, resolves #1058 

- Create new viper instance to watch changes to the sampling strategies file
- Wrap the sampling strategy store in a `RWMutex` for concurrent access.

### DIFF
--- a/plugin/sampling/strategystore/static/strategy_store.go
+++ b/plugin/sampling/strategystore/static/strategy_store.go
@@ -44,11 +44,10 @@ type strategyStore struct {
 func (h *strategyStore) loadAndParseStrategies(strategiesFile string) error {
 	s, err := loadStrategies(strategiesFile)
 	if err != nil {
-		h.logger.Warn("Using the last saved configuration for sampling strategies.", zap.Error(err))
+		h.logger.Warn("using the last saved configuration for sampling strategies.", zap.Error(err))
 		return err
 	}
 
-	h.logger.Info("Updating sampling strategies file!", zap.Any("Strategies", s))
 	h.parseStrategies(s)
 	return nil
 }
--- a/plugin/sampling/strategystore/static/strategy_store.go
+++ b/plugin/sampling/strategystore/static/strategy_store.go
@@ -125,6 +125,8 @@ func (h *strategyStore) parseStrategies(strategies *strategies) {
 	h.lock.Lock()
 	defer h.lock.Unlock()
 
+	h.defaultStrategy = defaultStrategyResponse()
+
 	if strategies == nil {
 		h.logger.Info("No sampling strategies provided, using defaults")
 		return
--- a/plugin/sampling/strategystore/static/strategy_store.go
+++ b/plugin/sampling/strategystore/static/strategy_store.go
@@ -24,7 +24,6 @@ import (
 
 	"github.com/fsnotify/fsnotify"
 	"github.com/pkg/errors"
-	"github.com/spf13/viper"
 	"go.uber.org/zap"
 
 	ss "github.com/jaegertracing/jaeger/cmd/collector/app/sampling/strategystore"
@@ -33,7 +32,7 @@ import (
 
 type strategyStore struct {
 	// to allow concurrent update of sampling strategies
-	sync.RWMutex
+	lock sync.RWMutex
 
 	logger *zap.Logger
 
@@ -48,38 +47,61 @@ func NewStrategyStore(options Options, logger *zap.Logger) (ss.StrategyStore, er
 		serviceStrategies: make(map[string]*sampling.SamplingStrategyResponse),
 	}
 
-	// new instance of viper to watch sampling strategies file
-	viper.AddConfigPath(options.StrategiesFile)
-	viper.WatchConfig()
-	viper.OnConfigChange(func(e fsnotify.Event) {
-		s, err := loadStrategies(options.StrategiesFile)
-		if err != nil {
-			logger.Fatal("Error while parsing strategies file", zap.Error(err))
-		}
-		h.parseStrategies(s)
-	})
-
+	// Read strategies
 	strategies, err := loadStrategies(options.StrategiesFile)
 	if err != nil {
 		return nil, err
 	}
 	h.parseStrategies(strategies)
+
+	// Watch strategies file for changes.
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		logger.Error("failed to create a new watcher for the sampling strategies file", zap.Error(err))
+	}
+
+	go func() {
+		for {
+			select {
+			case event := <-watcher.Events:
+				if event.Op&fsnotify.Remove == fsnotify.Remove {
+					logger.Warn("the sampling strategies file has been removed, using the last known version")
+					continue
+				}
+
+				s, err := loadStrategies(options.StrategiesFile)
+				if err != nil {
+					logger.Warn("Error while parsing strategies file", zap.Error(err))
+				}
+				h.parseStrategies(s)
+			case err, ok := <-watcher.Errors:
+				if !ok {
+					return
+				}
+				logger.Error("event", zap.Error(err))
+			}
+		}
+	}()
+
+	err = watcher.Add(options.StrategiesFile)
+	if err != nil {
+		logger.Error("error adding watcher to file", zap.String("file", options.StrategiesFile), zap.Error(err))
+	} else {
+		logger.Info("watching", zap.String("file", options.StrategiesFile))
+	}
+
 	return h, nil
 }
 
 // GetSamplingStrategy implements StrategyStore#GetSamplingStrategy.
 func (h *strategyStore) GetSamplingStrategy(serviceName string) (*sampling.SamplingStrategyResponse, error) {
-	h.RLock()
-	strategy, ok := h.serviceStrategies[serviceName]
-	h.RUnlock()
-
-	if !ok {
-		h.RLock()
-		strategy = h.defaultStrategy
-		h.RUnlock()
+	h.lock.RLock()
+	defer h.lock.RUnlock()
+	if strategy, ok := h.serviceStrategies[serviceName]; ok {
+		return strategy, nil
 	}
 
-	return strategy, nil
+	return h.defaultStrategy, nil
 }
 
 // TODO good candidate for a global util function
@@ -99,27 +121,23 @@ func loadStrategies(strategiesFile string) (*strategies, error) {
 }
 
 func (h *strategyStore) parseStrategies(strategies *strategies) {
+	h.lock.Lock()
+	defer h.lock.Unlock()
+
 	if strategies == nil {
 		h.logger.Info("No sampling strategies provided, using defaults")
 		return
 	}
 	if strategies.DefaultStrategy != nil {
-		defaultStrategy := h.parseServiceStrategies(strategies.DefaultStrategy)
-
-		h.Lock()
-		h.defaultStrategy = defaultStrategy
-		h.Unlock()
+		h.defaultStrategy = h.parseServiceStrategies(strategies.DefaultStrategy)
 	}
 
 	merge := true
-	h.RLock()
 	if h.defaultStrategy.OperationSampling == nil ||
 		h.defaultStrategy.OperationSampling.PerOperationStrategies == nil {
 		merge = false
 	}
-	h.RUnlock()
 
-	h.Lock()
 	for _, s := range strategies.ServiceStrategies {
 		h.serviceStrategies[s.Service] = h.parseServiceStrategies(s)
 
@@ -139,7 +157,6 @@ func (h *strategyStore) parseStrategies(strategies *strategies) {
 				h.defaultStrategy.OperationSampling.PerOperationStrategies)
 		}
 	}
-	h.Unlock()
 }
 
 // mergePerOperationStrategies merges two operation strategies a and b, where a takes precedence over b.
--- a/plugin/sampling/strategystore/static/strategy_store.go
+++ b/plugin/sampling/strategystore/static/strategy_store.go
@@ -65,15 +65,28 @@ func NewStrategyStore(options Options, logger *zap.Logger) (ss.StrategyStore, er
 			select {
 			case event := <-watcher.Events:
 				if event.Op&fsnotify.Remove == fsnotify.Remove {
-					logger.Warn("the sampling strategies file has been removed, using the last known version")
+					logger.Warn("the sampling strategies file has been removed")
+
+					watcher.Remove(event.Name)
+					watcher.Add(options.StrategiesFile)
+
+					s, err := loadStrategies(options.StrategiesFile)
+					if err != nil {
+						logger.Warn("Error while parsing strategies file. Using the last saved configuration.", zap.Error(err))
+					} else {
+						logger.Info("Updating sampling strategies file!", zap.Any("Strategies", s))
+						h.parseStrategies(s)
+					}
 					continue
 				}
-
-				s, err := loadStrategies(options.StrategiesFile)
-				if err != nil {
-					logger.Warn("Error while parsing strategies file", zap.Error(err))
-				} else {
-					h.parseStrategies(s)
+				if event.Op&fsnotify.Write == fsnotify.Write {
+					s, err := loadStrategies(options.StrategiesFile)
+					if err != nil {
+						logger.Warn("Error while parsing strategies file. Using the last saved configuration.", zap.Error(err))
+					} else {
+						logger.Info("Updating sampling strategies file!", zap.Any("Strategies", s))
+						h.parseStrategies(s)
+					}
 				}
 			case err, ok := <-watcher.Errors:
 				if !ok {
--- a/plugin/sampling/strategystore/static/strategy_store.go
+++ b/plugin/sampling/strategystore/static/strategy_store.go
@@ -72,8 +72,7 @@ func NewStrategyStore(options Options, logger *zap.Logger) (ss.StrategyStore, er
 
 	go h.runWatcherLoop(watcher, options.StrategiesFile)
 
-	err = watcher.Add(options.StrategiesFile)
-	if err != nil {
+	if err = watcher.Add(options.StrategiesFile); err != nil {
 		logger.Error("error adding watcher to file", zap.String("file", options.StrategiesFile), zap.Error(err))
 	} else {
 		logger.Info("watching", zap.String("file", options.StrategiesFile))
--- a/plugin/sampling/strategystore/static/strategy_store.go
+++ b/plugin/sampling/strategystore/static/strategy_store.go
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"path/filepath"
 	"sync"
 
 	"github.com/fsnotify/fsnotify"
@@ -78,6 +79,14 @@ func NewStrategyStore(options Options, logger *zap.Logger) (ss.StrategyStore, er
 		logger.Info("watching", zap.String("file", options.StrategiesFile))
 	}
 
+	dir := filepath.Dir(options.StrategiesFile)
+	err = watcher.Add(dir)
+	if err != nil {
+		h.logger.Error("error adding watcher to dir", zap.String("dir", dir), zap.Error(err))
+	} else {
+		h.logger.Info("watching", zap.String("dir", dir))
+	}
+
 	return h, nil
 }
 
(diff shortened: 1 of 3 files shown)


plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -41,6 +47,18 @@ func NewStrategyStore(options Options, logger *zap.Logger) (ss.StrategyStore, er
 		logger:            logger,
 		serviceStrategies: make(map[string]*sampling.SamplingStrategyResponse),
 	}
+
+	// new instance of viper to watch sampling strategies file
+	viper.AddConfigPath(options.StrategiesFile)
```
yurishkuro: Why do we need to depend on viper (and especially on its global state) for watching?

Note that we already have a file watcher implemented for UI config.

plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -20,15 +20,21 @@ import (
...
+	"github.com/spf13/viper"
 	"go.uber.org/zap"
 
 	ss "github.com/jaegertracing/jaeger/cmd/collector/app/sampling/strategystore"
 	"github.com/jaegertracing/jaeger/thrift-gen/sampling"
 )
 
 type strategyStore struct {
+	// to allow concurrent update of sampling strategies
+	sync.RWMutex
```
yurishkuro: It's better to define a field, embedding extends the struct API.

plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -41,6 +47,18 @@ func NewStrategyStore(options Options, logger *zap.Logger) (ss.StrategyStore, er
...
 		serviceStrategies: make(map[string]*sampling.SamplingStrategyResponse),
 	}
+
+	// new instance of viper to watch sampling strategies file
+	viper.AddConfigPath(options.StrategiesFile)
+	viper.WatchConfig()
+	viper.OnConfigChange(func(e fsnotify.Event) {
+		s, err := loadStrategies(options.StrategiesFile)
+		if err != nil {
+			logger.Fatal("Error while parsing strategies file", zap.Error(err))
```
yurishkuro: Do we really want to kill the service for this?

plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -51,10 +69,17 @@ func NewStrategyStore(options Options, logger *zap.Logger) (ss.StrategyStore, er
 
 // GetSamplingStrategy implements StrategyStore#GetSamplingStrategy.
 func (h *strategyStore) GetSamplingStrategy(serviceName string) (*sampling.SamplingStrategyResponse, error) {
-	if strategy, ok := h.serviceStrategies[serviceName]; ok {
-		return strategy, nil
+	h.RLock()
```
yurishkuro: I think it's sufficient to lock and defer unlock at the beginning of the function. Same for parse function.

plugin/sampling/strategystore/static/strategy_store.go
annanay25: Swapped for `fsnotify` watcher.

annanay25:
Going through the discussion on #1688 - 
IIUC fsnotify requires the creation of new watchers for every file being monitored if we desire notifications on independent updates (we don't want the UI to reload when the sampling strategies file is changed) - so I don't think there any advantage of creating a "central component" for hot reload. Also, since the reload callbacks are going to be different for each component, we'll need a go-routine running in each of these components anyway.

The obvious con is repetition of code.

plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -41,19 +46,62 @@ func NewStrategyStore(options Options, logger *zap.Logger) (ss.StrategyStore, er
...
+	// Watch strategies file for changes.
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		logger.Error("failed to create a new watcher for the sampling strategies file", zap.Error(err))
+	}
+
+	go func() {
+		for {
+			select {
+			case event := <-watcher.Events:
```
jpkrohling: Could you try this in Kubernetes with a file coming from a config map? Because of the two-level symlinking, I had some problems with the hot reload of the UI config (#1688).

annanay25: Does it make sense to have an e2e test suite in Jaeger to cover these scenarios?

annanay25:
Deployment spec used in test -
```
---
kind: ConfigMap 
apiVersion: v1 
metadata:
  name: jaeger-sampling-strategies
data:
  config: |-
    {
      "service_strategies": [
        {
          "service": "foo",
          "type": "probabilistic",
          "param": 0.8,
          "operation_strategies": [
            {
              "operation": "op1",
              "type": "probabilistic",
              "param": 0.2
            },
            {
              "operation": "op2",
              "type": "probabilistic",
              "param": 0.4
            }
          ]
        },
        {
          "service": "bar",
          "type": "ratelimiting",
          "param": 5
        }
      ],
      "default_strategy": {
        "type": "probabilistic",
        "param": 0.5
      }
    }

---
kind: Pod 
apiVersion: v1 
metadata:
  name: jaeger-collector
spec:
  containers:
    - name: jaeger-collector
      image: annanay25/jaeger-collector:latest
      imagePullPolicy: IfNotPresent
      volumeMounts:
      - name: config-volume
        mountPath: /etc/config
  volumes:
    - name: config-volume
      configMap:
        name: jaeger-sampling-strategies
        items:
          - key: config
            path: config
```

plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -41,19 +46,75 @@ func NewStrategyStore(options Options, logger *zap.Logger) (ss.StrategyStore, er
...
+				if event.Op&fsnotify.Remove == fsnotify.Remove {
+					logger.Warn("the sampling strategies file has been removed")
+
+					watcher.Remove(event.Name)
+					watcher.Add(options.StrategiesFile)
```
yurishkuro: below we're checking for error, but not here?

also, wouldn't this always fail since the file was removed? Perhaps Remove should be just no-op, aside from logging.

plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -41,19 +46,75 @@ func NewStrategyStore(options Options, logger *zap.Logger) (ss.StrategyStore, er
...
 	}
 	h.parseStrategies(strategies)
+
+	// Watch strategies file for changes.
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		logger.Error("failed to create a new watcher for the sampling strategies file", zap.Error(err))
+	}
+
+	go func() {
```
yurishkuro: please move to a function, e.g. `go runWatcherLoop(watcher)`

plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -41,19 +46,75 @@ func NewStrategyStore(options Options, logger *zap.Logger) (ss.StrategyStore, er
...
+		for {
+			select {
+			case event := <-watcher.Events:
+				if event.Op&fsnotify.Remove == fsnotify.Remove {
+					logger.Warn("the sampling strategies file has been removed")
+
+					watcher.Remove(event.Name)
+					watcher.Add(options.StrategiesFile)
+
+					s, err := loadStrategies(options.StrategiesFile)
```
yurishkuro: why do we want to parse a file that was removed?

plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -41,19 +46,75 @@ func NewStrategyStore(options Options, logger *zap.Logger) (ss.StrategyStore, er
...
+				if !ok {
+					return
+				}
+				logger.Error("event", zap.Error(err))
+			}
+		}
+	}()
+
+	err = watcher.Add(options.StrategiesFile)
+	if err != nil {
```
yurishkuro: one line
```
if err := watcher.Add(options.StrategiesFile); err != nil {
}
```

plugin/sampling/strategystore/static/strategy_store.go
annanay25: This is actually a workaround for k8s ConfigMaps. Since it is a simlinked file, fsnotify notifies a remove event when the ConfigMap is updated. I should probably add a comment in the file to explain this behaviour.

plugin/sampling/strategystore/static/strategy_store.go
annanay25: > below we're checking for error, but not here?

Will add error checking.

> also, wouldn't this always fail since the file was removed? 

Please check comment below.

plugin/sampling/strategystore/static/strategy_store.go
jpkrohling: You mean, to test a Kubernetes-specific setup, or the nested symlinks ?

annanay25: A kubernetes-specific setup

jpkrohling: It might then belong to the operator repo, instead of this one.

annanay25:
@yurishkuro @jpkrohling - Can we merge this?

plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -29,31 +32,103 @@ import (
...
 	logger *zap.Logger
 
 	defaultStrategy   *sampling.SamplingStrategyResponse
 	serviceStrategies map[string]*sampling.SamplingStrategyResponse
 }
 
+func (h *strategyStore) loadAndParseStrategies(strategiesFile string) error {
+	s, err := loadStrategies(strategiesFile)
+	if err != nil {
+		h.logger.Warn("Using the last saved configuration for sampling strategies.", zap.Error(err))
```
yurishkuro: start log messages with lower case (as you do in the other files below)

plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -29,31 +32,103 @@ import (
...
+		return err
+	}
+
+	h.logger.Info("Updating sampling strategies file!", zap.Any("Strategies", s))
```
yurishkuro: What is the intention of this log stmt? Are you trying to log the actual loaded strategies?

plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -29,31 +32,103 @@ import (
...
+
+	// Watch strategies file for changes.
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		logger.Error("failed to create a new watcher for the sampling strategies file", zap.Error(err))
+	}
+
+	go h.runWatcherLoop(watcher, options.StrategiesFile)
+
+	if err = watcher.Add(options.StrategiesFile); err != nil {
```
yurishkuro:
```suggestion
	if err := watcher.Add(options.StrategiesFile); err != nil {
```
don't leak variable scope

plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -29,31 +32,103 @@ import (
...
+		logger.Error("error adding watcher to file", zap.String("file", options.StrategiesFile), zap.Error(err))
+	} else {
+		logger.Info("watching", zap.String("file", options.StrategiesFile))
+	}
+
+	dir := filepath.Dir(options.StrategiesFile)
+	err = watcher.Add(dir)
+	if err != nil {
```
yurishkuro:
```suggestion
	if err := watcher.Add(dir); err != nil {
```

plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -29,31 +32,103 @@ import (
...
+		select {
+		case event := <-watcher.Events:
+			if event.Op&fsnotify.Remove == fsnotify.Remove {
+				if event.Name == strategiesFile {
+					h.logger.Warn("the sampling strategies file has been removed")
+
+					// This is a workaround for k8s configmaps. Since k8s loads configmaps as
+					// symlinked files within the containers, changes to the configmap register
+					// as `fsnotify.Remove` events.
+					if err := watcher.Add(strategiesFile); err != nil {
```
yurishkuro: Isn't there a race condition here? If the file was removed, even temporarily, and you call `watcher.Add()` on non-existing file, will it not error out immediately?

plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -29,31 +32,103 @@ import (
...
+						h.logger.Warn("Error adding sampling strategy config file to fsnotify watcher", zap.Error(err))
```
yurishkuro:
```suggestion
						h.logger.Warn("cannot watch sampling strategy config file", zap.Error(err))
```

plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -29,31 +32,103 @@ import (
...
+				continue
+			}
+			if event.Op&fsnotify.Write == fsnotify.Write {
+				h.loadAndParseStrategies(strategiesFile)
+			}
+		case err, ok := <-watcher.Errors:
+			if !ok {
+				return
+			}
+			h.logger.Error("event", zap.Error(err))
```
yurishkuro:
```suggestion
			h.logger.Error("file watcher error", zap.Error(err))
```

plugin/sampling/strategystore/static/strategy_store.go
annanay25: Must have added this for debugging, removing from PR.

plugin/sampling/strategystore/static/strategy_store.go
annanay25: So this line is specifically added for the k8s configmap scenario. It is added to handle successive reloads without restarting the collector (without this line, the collector reloads the strategies only the first time the file was changed). I've verified that it works. 

However, if the config file was manually removed, yes it will error out immediately, but that works for us.

plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -29,31 +32,102 @@ import (
...
+					// as `fsnotify.Remove` events.
+					if err := watcher.Add(strategiesFile); err != nil {
+						h.logger.Warn("cannot watch sampling strategy config file", zap.Error(err))
+					}
+
+					h.loadAndParseStrategies(strategiesFile)
+				}
+				continue
+			}
+			if event.Op&fsnotify.Write == fsnotify.Write {
```
yurishkuro: what about Create event? I think it would've addressed my question above.

plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -29,31 +32,102 @@ import (
...
+	}
+
 	return h, nil
 }
 
+func (h *strategyStore) runWatcherLoop(watcher *fsnotify.Watcher, strategiesFile string) {
+	for {
+		select {
+		case event := <-watcher.Events:
+			if event.Op&fsnotify.Remove == fsnotify.Remove {
```
yurishkuro: Why are these treated as bitmasks? The Op constants are defined as ordinary numbers (https://godoc.org/github.com/fsnotify/fsnotify#Op).

plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -29,31 +32,102 @@ import (
...
+				if event.Name == strategiesFile {
```
yurishkuro: given that we're watching just a single file, when would the name not equal?

plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -29,31 +32,102 @@ import (
...
+
+	go h.runWatcherLoop(watcher, options.StrategiesFile)
+
+	if err := watcher.Add(options.StrategiesFile); err != nil {
+		logger.Error("error adding watcher to file", zap.String("file", options.StrategiesFile), zap.Error(err))
+	} else {
+		logger.Info("watching", zap.String("file", options.StrategiesFile))
+	}
+
+	dir := filepath.Dir(options.StrategiesFile)
```
yurishkuro: why do we need to watch the dir?

plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -29,31 +32,102 @@ import (
...
+		h.logger.Info("watching", zap.String("dir", dir))
+	}
+
 	return h, nil
 }
 
+func (h *strategyStore) runWatcherLoop(watcher *fsnotify.Watcher, strategiesFile string) {
+	for {
+		select {
+		case event := <-watcher.Events:
```
yurishkuro: Could we not simplify this by handling all event? E.g.
```go
case event := <-watcher.Events:
    if event.Op == fsnotify.Remove {
        // resubscribe to handle k8s use case
    }
    h.loadAndParseStrategies(strategiesFile)
    continue
```

annanay25: The current logic is actually very close to this 

1. Remove events could occur in a non k8s environment. In which case we need to handle only write events, hence the following block-
```			
if event.Op&fsnotify.Write == fsnotify.Write {
	h.loadAndParseStrategies(strategiesFile)
}
```

2. If they occur in a k8s environment, we need to reload the file strategies file hence the first part

annanay25:
OK, I think I added the directory logic to get past Codecov. Could you check if the logic at this commit works? https://github.com/jaegertracing/jaeger/pull/2040/commits/31ea29e5d5a0d5380cd258c8ae33658614e6d1d2

PR coverage is 54%, but that's because we're not writing k8s specific test cases.

plugin/sampling/strategystore/static/strategy_store.go
jpkrohling: Just a note: the second case isn't specific to Kubernetes, but it's how Kubernetes mounts those files in the local file system (with two symlinks at different levels). The same would happen on non-Kubernetes environments as well, if people use symlinks on directories pointing to the "current" version of a config map.

annanay25:
Gah, Codecov is at 40%, but the only lines left out are fsnotify errors.

plugin/sampling/strategystore/static/strategy_store.go
yurishkuro: my point is that you want special handing for Remove by re-subscribing, and all events should result in parsing the file. Right now the code ignores several events altogether.

annanay25:
@jpkrohling @yurishkuro - soft ping, can we move forward with this? The test cases for this will have to be written in the operator repo

yurishkuro:
Please address review comments. Only 40% of new code is covered, please bring it above the 95% target.

annanay25:
@yurishkuro - I've already added one test case. Test cases to cover the remaining lines of code will have to be added in the operator repo.

yurishkuro:
Tests in the operator repo do not affect code coverage. Why can’t we unit test?

annanay25:
> Tests in the operator repo do not affect code coverage

Yes, I'm aware. 

[Lines checking for `fsnotify` errors](https://codecov.io/gh/jaegertracing/jaeger/compare/1d8cf5e157d4f559ccca53fc91fe44a181dd7520...8721e11c2628a9e509b6d4901f0fdaa18ac0495a/src/plugin/sampling/strategystore/static/strategy_store.go?before=plugin/sampling/strategystore/static/strategy_store.go#D1-115) are the only ones untested. Not sure how they can be unit tested.

plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -29,31 +32,101 @@ import (
...
 	logger *zap.Logger
 
 	defaultStrategy   *sampling.SamplingStrategyResponse
 	serviceStrategies map[string]*sampling.SamplingStrategyResponse
 }
 
+func (h *strategyStore) loadAndParseStrategies(strategiesFile string) error {
+	s, err := loadStrategies(strategiesFile)
+	if err != nil {
+		h.logger.Warn("using the last saved configuration for sampling strategies.", zap.Error(err))
```
yurishkuro:
```suggestion
		h.logger.Warn("Unable to reload the sampling strategies file. Using the last saved configuration.", zap.Error(err), zap.String("file", strategiesFile))
```

plugin/sampling/strategystore/static/strategy_store.go
yurishkuro: ?

plugin/sampling/strategystore/static/strategy_store.go
yurishkuro: > Right now the code ignores several events altogether.

?

plugin/sampling/strategystore/static/strategy_store.go
yurishkuro: ?

NareshBhanawat:
instead of monitoring file change, why dont we expose an API which can access json payload on sampling strategy?

defool:
Checking file periodically is  another way to reload sampling strategies, which is simpler and more reliable. Can I create a PR to do this? @yurishkuro

yurishkuro:
@defool take a look at #2186

yurishkuro:
#2188 implements a simpler, timer-based reload, instead of watching for file changes. I suggest we consider closing this PR - @annanay25

PR #2044, opened 2020-01-27
Add before you push to the queue to prevent race condition on size

### LINKED ISSUE #2042: Dropping small amount of spans on master

Trying to run master and checking if everything works.

Dropping a small number of spans

Here's more info:
We are currently running Jaeger with about 10K spans/sec on a host. But we noticed that `rate(jaeger_collector_spans_dropped_total[5m])` is `~1` on upgrade to master. Which shouldn't be there, because we have an insane queue capacity (300K).

So I put in some debug logging

```
diff --git a/pkg/queue/bounded_queue.go b/pkg/queue/bounded_queue.go
index 077e7af..bbbabb7 100644
--- a/pkg/queue/bounded_queue.go
+++ b/pkg/queue/bounded_queue.go
@@ -16,6 +16,7 @@
 package queue
 import (
+       "fmt"
        "sync"
        "sync/atomic"
        "time"
@@ -99,8 +100,11 @@ func (q *BoundedQueue) Produce(item interface{}) bool {
        // we might have two concurrent backing queues at the moment
        // their combined size is stored in q.size, and their combined capacity
        // should match the capacity of the new queue
-       if q.Size() >= q.Capacity() {
+       size := q.Size()
+       capacity := q.Capacity()
+       if size >= capacity {
                // note that all items will be dropped if the capacity is 0
+               fmt.Println("size is bigger than capacity! dropping span!", size, capacity)
                q.onDroppedItem(item)
                return false
        }
@@ -111,6 +115,7 @@ func (q *BoundedQueue) Produce(item interface{}) bool {
                return true
        default:
                // should not happen, as overflows should have been captured earlier
+               fmt.Println("dropping span on overflow!")
                if q.onDroppedItem != nil {
                        q.onDroppedItem(item)
                }
```

And saw this in the logs:
```
size is bigger than capacity! dropping span! 4294967295 300000
size is bigger than capacity! dropping span! 4294967295 300000
size is bigger than capacity! dropping span! 4294967295 300000
size is bigger than capacity! dropping span! 4294967295 300000
size is bigger than capacity! dropping span! 4294967295 300000
size is bigger than capacity! dropping span! 4294967295 300000
size is bigger than capacity! dropping span! 4294967295 300000
size is bigger than capacity! dropping span! 4294967295 300000
size is bigger than capacity! dropping span! 4294967295 300000
```

So I see that recently the queue size handling moved to using ubers atomic package. And from int32 to uint32. I think before the size used to go to -1 but then it worked, but now it's broken.

joe-elliott:
Fixes #2042 

Adding to size before we put an item in the channel prevents `q.size.Sub(1)` hitting before the addition.  This prevents the uint from underflowing to max value.

This change fixed the issue described in #2042 in our environment.

cc @gouthamve

### DIFF
--- a/pkg/queue/bounded_queue.go
+++ b/pkg/queue/bounded_queue.go
@@ -105,12 +105,13 @@ func (q *BoundedQueue) Produce(item interface{}) bool {
 		return false
 	}
 
+	q.size.Add(1)
 	select {
 	case *q.items <- item:
-		q.size.Add(1)
 		return true
 	default:
 		// should not happen, as overflows should have been captured earlier
+		q.size.Sub(1)
 		if q.onDroppedItem != nil {
 			q.onDroppedItem(item)
 		}


joe-elliott:
Submodule updated.  Fixing.

pkg/queue/bounded_queue.go
```diff
@@ -105,12 +105,13 @@ func (q *BoundedQueue) Produce(item interface{}) bool {
 		return false
 	}
 
+	q.size.Add(1)
```
jpkrohling: This actually looks like a real bug, we are subtracting in both code branches, but adding in only one. Good catch!

jpkrohling:
Thanks for the PR!

Issue #961, opened 2018-07-26
Make clock skew adjustment transparent


yurishkuro:
## Requirement - what kind of business use case are you trying to solve?

It is currently not obvious in Jaeger UI is the span timestamps were adjusted for clock skew. @tedsuo wrote:

> The issue was that displaying “manipulated” data, however well intentioned, led users to mistrust what they were seeing in some cases. “Is this 'real data' or is this ‘corrected data’?" was becoming a common question when looking at odd display patterns, or patterns that did not match other data sources. Given that we usually could not “correct” most scenarios properly but could only place the spans into a sort of “best guess” situation, we decided it would be better to avoid confusion and always show raw data, with visual annotations to point out where we felt the raw data was suspicious. So the main issue is trust, and retaining the ability to triage raw data. Another approach would be to give users a toggle to turn off the corrections so they can quickly see the raw data.
> 
> As an example, there were situations where clock correction was actually covering up issues, or subtly moving spans in a way which “looked right” but developers knew were impossible. For example, in one case there was a batch processing workload where the clock correction subtly moved things to make it look like the map/reduce phases were slightly overlapping in impossible ways. Other cases were where incorrect instrumentation lead to “incoceivable” child spans, but the source was not clock skew - that child span really was offset in a crazy way!

## Problem - what in Jaeger blocks you from solving the requirement?

Jaeger clock skew adjustment is not recorded anywhere in the adjusted spans. UI does not make it obvious that the timestamps were adjusted.

## Proposal - what do you suggest to solve the problem or improve the existing situation?

* [x] Clock skew adjuster should record the delta in the adjusted span
* [x] Jaeger UI should display a warning in the span details (and/or some other visual cue) that the span was adjusted
* [ ] Jaeger UI should provide a toggle to view the unadjusted trace (the `/api/trace` endpoint already allows retrieving unadjusted trace)

## Any open questions to address

How to do a visual cue that the span is adjusted.

felippe-mendonca:
I had the same problem as exposed [here](https://github.com/jaegertracing/jaeger-client-java/issues/192). It took some time do figure out if the problem was mine or not. Because of that I think that could be interesting to have a mix between the second and third proposal: when displaying adjusted trace, warn the user about that.

cameronbraid:
Is there any way to make the UI run without the clock skew adjuster ? 

It is leading to incorrect representations of spans, shifting spans by hundreds of milliseconds in some cases.

[...]

PR #2052, opened 2020-02-02
Warn about time adjustment in tags

bobrik:
Time adjustment is confusing to many people, as you can see in #961.

This change adds a warning if we do any time adjustments, so that it's at least clear that adjustments happened.


bobrik:
I spent a few hours trying to figure out how on earth can a span for 301 redirect follow can happen before the first request that triggered said redirect finished.

image

The reason was that tracing with futures is hard and redirect following span outlived its parent, which triggered an adjustment. This adjustment may or may not be a good idea, but let's at least be clear it happened, so people can save some time in the future:

image

bobrik:
It definitely looks better as a warning, updated:

image

prongs:
Any traction on this?

PR #2049, opened 2020-01-30
Support regex tags search for Elasticseach backend

annanay25:
Signed-off-by: Annanay <annanayagarwal@gmail.com>

## Which problem is this PR solving?
When using the ES backend, adds the ability to 
- use regex search for the values field. So a query like `foo=bar.*` should match all tags like `{foo:bar}, {foo:barbaz}, {foo:bar1}`

Related to #684 

## Short description of the changes
- Swapped the `NewMatchQuery` method in ES reader for `NewRegexpQuery`

/cc @pavolloffay

PR #2054, opened 2020-02-03
[ES, C*] Check for the presence of a tag

annanay25:
- Adds functionality to check for the presence of a tag, to the ES and C* readers.

- New queries for C* and ES backends.
- Added parameter to the GRPC TraceQueryParameters and updated proto-generated files.
- Added tests.

In the UI, this feature could be implemented as a search-box, input being a comma separated list of tags that need to be present in the result traces.


plugin/storage/cassandra/spanstore/reader.go
```diff
@@ -301,9 +308,30 @@ func (s *SpanReader) findTraceIDs(ctx context.Context, traceQuery *spanstore.Tra
 	if len(traceQuery.Tags) > 0 {
 		return s.queryByTagsAndLogs(ctx, traceQuery)
 	}
+	if len(traceQuery.CheckTagsPresent) > 0 {
+		s.queryCheckTagPresence(ctx, traceQuery)
```
yurishkuro: Return?

annanay25: Oops. Will fix.

annanay25:
I'd like to wait for https://github.com/jaegertracing/jaeger/pull/2049 to be merged before moving forward with this.

jpkrohling:
@annanay25 looks like #2049 is already there ;-)

PR #2056, opened 2020-02-05
Allow the Configuration of Additional Headers on the Jaeger Query HTTP API

### LINKED ISSUE #2039: Add CORS handling to query API

I love jaeger I think it's a great tool,

I want to use jaeger data with [Costume UI](https://github.com/SAP/Tracer).
I think the best approach for me is to use [HTTP JSON (internal) API](https://www.jaegertracing.io/docs/1.13/apis/#http-json-internal)
 

Without support CORS headers (Access-Control-Allow-Origin: *)
 I can't use the query service from UI applications running on different hosts.

joe-elliott:
- Fixes #2039 

- Added a parameter `query.additional-headers` that allows for multiple headers to be added to jaeger query's api
- Configurable from the CLI
  - `--query.additional-headers "Access-Control-Allow-Origin: https://mozilla.org" --query-additional-headers "Access-Control-Expose-Headers: X-My-Custom-Header"`
- Configurable from yaml
  ```
  query:
    additional-headers:
      - "Access-Control-Allow-Origin: https://mozilla.org"
      - "Access-Control-Expose-Headers: X-My-Custom-Header"
  ```
- Added a config.StringSlice object that can be used to pass string slices through pflag/viper

### DIFF
--- a/cmd/query/app/additional_headers_test.go
+++ b/cmd/query/app/additional_headers_test.go
@@ -1,0 +1,50 @@
+// Copyright (c) 2020 The Jaeger Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package app
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestAdditionalHeadersHandler(t *testing.T) {
+
+	additionalHeaders := http.Header{}
+	additionalHeaders.Add("Access-Control-Allow-Origin", "https://mozilla.org")
+	additionalHeaders.Add("Access-Control-Expose-Headers", "X-My-Custom-Header")
+	additionalHeaders.Add("Access-Control-Expose-Headers", "X-Another-Custom-Header")
+	additionalHeaders.Add("Access-Control-Request-Headers", "field1, field2")
+
+	emptyHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Write([]byte{})
+	})
+
+	handler := additionalHeadersHandler(emptyHandler, additionalHeaders)
+	server := httptest.NewServer(handler)
+	defer server.Close()
+
+	req, err := http.NewRequest("GET", server.URL, nil)
+	assert.NoError(t, err)
+
+	resp, err := httpClient.Do(req)
+	assert.NoError(t, err)
+
+	for k, v := range additionalHeaders {
+		assert.Equal(t, v, resp.Header[k])
+	}
+}
--- a/cmd/query/app/additional_headers_test.go
+++ b/cmd/query/app/additional_headers_test.go
@@ -23,7 +23,6 @@ import (
 )
 
 func TestAdditionalHeadersHandler(t *testing.T) {
-
 	additionalHeaders := http.Header{}
 	additionalHeaders.Add("Access-Control-Allow-Origin", "https://mozilla.org")
 	additionalHeaders.Add("Access-Control-Expose-Headers", "X-My-Custom-Header")
--- a/cmd/query/app/additional_headers_test.go
+++ b/cmd/query/app/additional_headers_test.go
@@ -40,7 +40,7 @@ func TestAdditionalHeadersHandler(t *testing.T) {
 	req, err := http.NewRequest("GET", server.URL, nil)
 	assert.NoError(t, err)
 
-	resp, err := httpClient.Do(req)
+	resp, err := server.Client().Do(req)
 	assert.NoError(t, err)
 
 	for k, v := range additionalHeaders {
--- a/cmd/query/app/flags.go
+++ b/cmd/query/app/flags.go
@@ -17,6 +17,7 @@ package app
 
 import (
 	"flag"
+	"net/http"
 
 	"github.com/spf13/viper"
 
@@ -46,7 +47,7 @@ type QueryOptions struct {
 	// BearerTokenPropagation activate/deactivate bearer token propagation to storage
 	BearerTokenPropagation bool
 	// AdditionalHeaders
-	AdditionalHeaders []string
+	AdditionalHeaders http.Header
 }
 
 // AddFlags adds flags for QueryOptions
@@ -66,6 +67,12 @@ func (qOpts *QueryOptions) InitFromViper(v *viper.Viper) *QueryOptions {
 	qOpts.StaticAssets = v.GetString(queryStaticFiles)
 	qOpts.UIConfig = v.GetString(queryUIConfig)
 	qOpts.BearerTokenPropagation = v.GetBool(queryTokenPropagation)
-	qOpts.AdditionalHeaders = v.GetStringSlice(queryAdditionalHeaders)
+	qOpts.AdditionalHeaders = stringSliceAsHeader(v.GetStringSlice(queryAdditionalHeaders))
 	return qOpts
+}
+
+// stringSliceAsHeader parses a string slice and returns a header.  each line
+//
+func stringSliceAsHeader(s []string) http.Header {
+
 }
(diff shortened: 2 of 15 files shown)


joe-elliott:
Functionality confirmed with local testing.  Dropping Draft status.

cmd/query/app/http_handler.go
```diff
@@ -450,6 +451,11 @@ func (aH *APIHandler) writeJSON(w http.ResponseWriter, r *http.Request, response
 		}
 	}
 	resp, _ := marshall(response)
-	w.Header().Set("Content-Type", "application/json")
+
+	header := w.Header()
+	header.Set("Content-Type", "application/json")
+	for key, values := range aH.additionalHeaders {
+		header[key] = values
```
yurishkuro: should this be calling `header.Add()` instead?

cmd/query/app/server.go
```diff
@@ -161,3 +166,24 @@ func (s *Server) Close() {
 	s.httpServer.Close()
 	s.conn.Close()
 }
+
+// stringSliceAsHeader parses a slice of strings and returns a http.Header.
+//  Each string in the slice is expected to be in the format "key: value"
+func stringSliceAsHeader(slice []string, logger *zap.Logger) http.Header {
```
yurishkuro: I think this method should be somewhere in the configuration package that defines the CLI flags. This is more of a "business" functionality that should get the data in the already-parsed format.

cmd/query/app/http_handler.go
joe-elliott: I debated both.   I ended up on this because the code is cleaner and, if any other headers besides `Content-Type` were hardcoded, the user would have the ability to overwrite them.

Either way is fine with me.  If you prefer `.Add()` I can swap it.

cmd/query/app/flags.go
```diff
@@ -43,16 +45,18 @@ type QueryOptions struct {
 	UIConfig string
 	// BearerTokenPropagation activate/deactivate bearer token propagation to storage
 	BearerTokenPropagation bool
+	// AdditionalHeaders
+	AdditionalHeaders []string
 }
 
 // AddFlags adds flags for QueryOptions
 func AddFlags(flagSet *flag.FlagSet) {
+	flagSet.Var(&config.StringSlice{}, queryAdditionalHeaders, `Additional HTTP response headers.  Can be specified multiple times.  Format: "Key: Value"`)
```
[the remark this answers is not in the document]
joe-elliott: Fixed.  Will be in the next push.

cmd/query/app/server.go
joe-elliott: Moved to `flags.go` and changed the type on QueryOptions from `[]string` to `http.Header`.  If this is not what you had in mind let me know!

cmd/query/app/http_handler.go
yurishkuro: Add will support repeated headers, which are valid in HTTP.

joe-elliott: The above supports repeated headers.  `http.Header` is a `map[string][]string`.  `values` is a `[]string` not just a `string`.

cmd/query/app/http_handler.go
```diff
@@ -450,6 +451,11 @@ func (aH *APIHandler) writeJSON(w http.ResponseWriter, r *http.Request, response
 		}
 	}
 	resp, _ := marshall(response)
-	w.Header().Set("Content-Type", "application/json")
+
```
yurishkuro: Q: should this really be scoped only to the API endpoints? I thought it would be a top-level middleware for all of query-service (that's how we did security headers internally).

joe-elliott: So I'm not a CORs expert, but my understanding is that it's mainly used on APIs to give browser information about how the API should be accessible to a script in a webpage.  So, as I was working on this feature I was mainly thinking in terms of the API calls.

I'm a little concerned that applying the same headers to the API calls as the main page would have unintended side effects.  I'm not an expert on the behavior of these headers, but I'd be glad to add that functionality.

yurishkuro: @everett980 @EranOfer do you have thoughts on this?

joe-elliott: If you'd like I can just go ahead and headers to all jaeger-query endpoints.  If it's working internally at Jaeger then it's probably fine.

yurishkuro: Yes, let's do it as a top-level middleware on all of query-service HTTP endpoints. If someone has specific needs later on to restrict to some endpoints only, we can revisit it later.

cmd/query/app/additional_headers_test.go
```diff
@@ -0,0 +1,49 @@
...
+	})
+
+	handler := additionalHeadersHandler(emptyHandler, additionalHeaders)
+	server := httptest.NewServer(handler)
+	defer server.Close()
+
+	req, err := http.NewRequest("GET", server.URL, nil)
+	assert.NoError(t, err)
+
+	resp, err := httpClient.Do(req)
```
yurishkuro:
```suggestion
	resp, err := server.Client().Do(req)
```

cmd/query/app/flags_test.go
```diff
@@ -30,10 +32,32 @@ func TestQueryBuilderFlags(t *testing.T) {
...
+		"--query.additional-headers=access-control-allow-origin:blerg",
+		"--query.additional-headers=whatever:thing",
 	})
-	qOpts := new(QueryOptions).InitFromViper(v)
+	qOpts := new(QueryOptions).InitFromViper(v, zap.NewNop())
 	assert.Equal(t, "/dev/null", qOpts.StaticAssets)
 	assert.Equal(t, "some.json", qOpts.UIConfig)
 	assert.Equal(t, "/jaeger", qOpts.BasePath)
 	assert.Equal(t, 80, qOpts.Port)
+	assert.Equal(t, http.Header{"Access-Control-Allow-Origin": []string{"blerg"}, "Whatever": []string{"thing"}}, qOpts.AdditionalHeaders)
```
yurishkuro: reads better this way:
```suggestion
	assert.Equal(t, http.Header{
	           "Access-Control-Allow-Origin": []string{"blerg"}, 
	           "Whatever": []string{"thing"}
	       }, qOpts.AdditionalHeaders)
```

cmd/query/app/flags_test.go
```diff
@@ -30,10 +32,32 @@ func TestQueryBuilderFlags(t *testing.T) {
...
+}
+
+func TestStringSliceAsHeader(t *testing.T) {
+	headers := []string{"Access-Control-Allow-Origin: https://mozilla.org",
```
yurishkuro:
```suggestion
	headers := []string{
	    "Access-Control-Allow-Origin: https://mozilla.org",
```

cmd/query/app/flags_test.go
```diff
@@ -30,10 +32,32 @@ func TestQueryBuilderFlags(t *testing.T) {
...
+	}
+
+	parsedHeaders := stringSliceAsHeader(headers, zap.NewNop())
+
+	assert.Equal(t, []string{"https://mozilla.org"}, parsedHeaders["Access-Control-Allow-Origin"])
+	assert.Equal(t, []string{"X-My-Custom-Header", "X-Another-Custom-Header"}, parsedHeaders["Access-Control-Expose-Headers"])
+
+	malformedHeaders := append(headers, "this is not a valid header")
+	parsedHeaders = stringSliceAsHeader(malformedHeaders, zap.NewNop())
+	assert.Nil(t, parsedHeaders)
```
yurishkuro: nit: it's better to make `stringSliceAsHeader()` return an error that can be tested, and only log it from InitFromViper, which has the logger.

pkg/config/string_slice.go
```diff
@@ -0,0 +1,44 @@
...
+		return "[]"
+	}
+
+	return `["` + strings.Join(*l, `","`) + `"]`
+}
+
+// Set implements pflag.Value
+func (l *StringSlice) Set(value string) error {
+	*l = append(*l, value)
+	return nil
```
yurishkuro: perhaps this should return an error if there is a comma in the `value`, because it will break the array syntax later.

cmd/query/app/flags_test.go
joe-elliott: done.

pkg/config/string_slice.go
joe-elliott: Commas actually work.  See `TestStringSlice`

```
	f.Set("test3,test4")
	assert.Equal(t, `["test","test2","test3,test4"]`, f.String())
```

yurishkuro:
restarting unit tests to get the coverage numbers (call to codecov timed out)

PR #2060, opened 2020-02-07
Refactor collector mains

jpkrohling:
- Follow up from #2031, refactoring the collector-specific entry points for all-in-one and collector

- Each commit is a refactoring unit and can be analyzed alone: each commit passes the tests from `make test` and both `all-in-one` and `collector` start successfully
- The first 4 commits are in preparation for the fourth, to avoid cyclic imports
- The fifth commit is similar to #2031, but without some of the graceful shutdown parts
- Then, a few changes based on the reviews for #2031 
- The remaining commits are further smaller refactoring, moving the servers to a specific package, plus consolidating the old `grpcserver` into `server/grpc.go`.
- And finally, split the server's networking parts from the serving parts, so that they are more testable. Tests not included.

I think this PR is already big enough on its own, so, I won't continue the refactoring of `all-in-one` as part of this commit.


cmd/collector/app/server/grpc.go
```diff
@@ -0,0 +1,93 @@
...
+		server = grpc.NewServer(grpc.Creds(creds))
+	} else {
+		// server without TLS
+		server = grpc.NewServer()
+	}
+
+	grpcPortStr := ":" + strconv.Itoa(params.Port)
+	listener, err := net.Listen("tcp", grpcPortStr)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to listen on gRPC port")
```
pavolloffay: It might be useful to include the port the message?  The caller knows the port when specified but it might not know it if the default is being used.

cmd/collector/app/server/thrift.go
```diff
@@ -0,0 +1,82 @@
...
+	StrategyStore        strategystore.StrategyStore
+	ServiceName          string
+	Port                 int
+	Logger               *zap.Logger
+}
+
+// StartThriftServer based on the given parameters
+func StartThriftServer(params *ThriftServerParams) (*tchannel.Channel, error) {
+	var tchServer *tchannel.Channel
+	var err error
```
pavolloffay: Do you need to declare an error here?

cmd/collector/app/server/zipkin.go
```diff
@@ -0,0 +1,88 @@
...
+	RecoveryHandler func(http.Handler) http.Handler
+	AllowedOrigins  string
+	AllowedHeaders  string
+	HealthCheck     *healthcheck.HealthCheck
+	Logger          *zap.Logger
+}
+
+// StartZipkinServer based on the given parameters
+func StartZipkinServer(params *ZipkinServerParams) (*http.Server, error) {
+	if params.Port == 0 {
```
pavolloffay: Why is there this check? We normally do not do these checks unless there is a convention in this API that it will no start the server if the port is `0`.

cmd/collector/app/server/zipkin.go
```diff
@@ -0,0 +1,88 @@
...
+
+	server := &http.Server{Addr: httpPortStr}
+	if err := serveZipkin(server, listener, params); err != nil {
+		return nil, err
+	}
+
+	return server, nil
+}
+
+func serveZipkin(server *http.Server, listener net.Listener, params *ZipkinServerParams) error {
```
pavolloffay: error is not being used

cmd/collector/app/collector.go
```diff
@@ -0,0 +1,183 @@
...
+		c.logger.Fatal("Could not start the Zipkin server", zap.Error(err))
+	} else {
+		c.zkServer = zkServer
+	}
+
+	return nil
+}
+
+// Close the component and all its underlying dependencies
+func (c *Collector) Close() error {
```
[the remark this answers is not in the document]
jpkrohling: The return type is part of the [`io.Closer`](https://golang.org/pkg/io/#Closer) interface.

cmd/collector/app/processor/span.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+// SpansOptions additional options passed to processor along with the spans.
+type SpansOptions struct {
+	SpanFormat       SpanFormat
+	InboundTransport InboundTransport
+}
+
+// Span handles model spans
+type Span interface {
```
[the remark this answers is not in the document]
jpkrohling: This one was just moved from a "generic" source file from `cmd/collector/app`. It has been moved here to avoid a cyclic dependency. We could perhaps do some refactoring around the processor part as well in a follow-up PR.

cmd/collector/app/server/zipkin.go
jpkrohling: That's the Zipkin server: we'll not start it if the port hasn't been specified.

cmd/collector/app/server/thrift.go
jpkrohling: Looks like I don't need it anymore

cmd/collector/app/server/zipkin.go
jpkrohling: All other `serve` functions are returning an error, even though only the Thrift one actually returns an error. I would keep it, so that callers can handle it once we decide to unify the signatures under an interface.

cmd/collector/app/server/grpc.go
jpkrohling: The wrapped error already includes it:

```
failed to listen on gRPC port: listen tcp :80: bind: permission denied
```

jpkrohling:
Review addressed + conflicts solved.

cmd/collector/app/collector.go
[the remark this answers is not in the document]
pavolloffay: Then return the error. There are several return assignments in this function and none error being returned.

cmd/collector/app/processor/span.go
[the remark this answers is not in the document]
vprithvi: I think that keeping the old name of `SpanProcessor` (and having a stuttering type name) is fine and more readable in this instance.

cmd/collector/app/collector.go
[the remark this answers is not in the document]
jpkrohling: There's no error that we would want to bubble up from here. For instance, when issuing a `Close()`, we don't want to break the routine when the HTTP server failed to close, as that would result in the Zipkin server not being closed.

cmd/collector/app/processor/span.go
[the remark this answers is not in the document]
jpkrohling: `processor.SpanProcessor#ProcessSpans()` seems quite repetitive, but I agree with reverting this change, especially because it's not the focus of this PR.

jpkrohling:
The rename of `processor.SpanProcessor` has been reverted and the PR has been rebased.

jpkrohling:
Coverage is failing, which is kinda expected, as the refactored code didn't have any tests at all, except for the gRPC server, which was kept.

cmd/collector/app/collector.go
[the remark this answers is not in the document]
pavolloffay: It does not mean that the error has to be returned _immediatelly_. It can be wrapped and returned at the end of the function. Hopefully, this function is not used by many clients...

cmd/collector/app/server/http.go
```diff
@@ -0,0 +1,88 @@
...
+
+	server := &http.Server{Addr: httpPortStr}
+	if err := serveHTTP(server, listener, params); err != nil {
+		return nil, err
+	}
+
+	return server, nil
+}
+
+func serveHTTP(server *http.Server, listener net.Listener, params *HTTPServerParams) error {
```
pavolloffay: Why there is `error` in the signature? It is not being returned.

cmd/collector/app/server/zipkin.go
pavolloffay: I see this is conventions we use for zipkin port flag.

cmd/collector/app/server/zipkin.go
pavolloffay: The API should be clear and return what is needed for a caller to know and handle.

cmd/collector/app/collector.go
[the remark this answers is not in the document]
jpkrohling: Interesting idea. If the HTTP server and Zipkin fails, Zipkin will end up wrapping the HTTP server failure. Not sure that's intuitive from the perspective of the admin looking at the logs.

> Hopefully, this function is not used by many clients...

It's being used by the `main.go` of both the Collector and All In One.

cmd/collector/app/server/zipkin.go
jpkrohling: Alright, removed it. If we refactor it and use a common interface, we can add it back.

cmd/collector/app/server/http.go
jpkrohling: Removed.

cmd/collector/app/processor/span.go
```diff
@@ -0,0 +1,59 @@
+// Copyright (c) 2020 The Jaeger Authors.
```
yurishkuro: "interface" would be better file name

cmd/collector/app/server/grpc.go
```diff
@@ -0,0 +1,93 @@
+// Copyright (c) 2020 The Jaeger Authors.
```
yurishkuro: This file seems to be replacing `cmd/collector/app/grpcserver/grpc_server.go`, which was deleted (git didn't recognize it as a move).

cmd/collector/app/server/grpc.go
```diff
@@ -0,0 +1,93 @@
...
+
+	params.Logger.Info("Starting jaeger-collector gRPC server", zap.Int("grpc-port", params.Port))
+	go func(server *grpc.Server) {
+		if err := server.Serve(listener); err != nil {
+			params.Logger.Error("Could not launch gRPC service", zap.Error(err))
+			if params.OnError != nil {
+				params.OnError(err)
+			}
+		}
+	}(server)
```
yurishkuro: `server` variable is already in scope, no need to pass it as argument

cmd/collector/app/server/grpc.go
```diff
@@ -0,0 +1,93 @@
...
+
+		creds := credentials.NewTLS(tlsCfg)
+		server = grpc.NewServer(grpc.Creds(creds))
+	} else {
+		// server without TLS
+		server = grpc.NewServer()
+	}
+
+	grpcPortStr := ":" + strconv.Itoa(params.Port)
+	listener, err := net.Listen("tcp", grpcPortStr)
```
yurishkuro: Future TODO: similar to query-service that runs both http and grpc servers on the same port, we could do the same for collector.

cmd/collector/app/server/grpc.go
```diff
@@ -0,0 +1,93 @@
...
+	Handler       *handler.GRPCHandler
+	SamplingStore strategystore.StrategyStore
+	Logger        *zap.Logger
+	OnError       func(error)
+}
+
+// StartGRPCServer based on the given parameters
+func StartGRPCServer(params *GRPCServerParams) (*grpc.Server, error) {
+	var server *grpc.Server
+	grpclog.SetLoggerV2(grpclog.NewLoggerV2(ioutil.Discard, os.Stderr, os.Stderr))
```
yurishkuro: hm, odd that we don't do that in query-service, as far as I could find

cmd/collector/app/server/http.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/jaegertracing/jaeger/cmd/collector/app/sampling/strategystore"
+	clientcfgHandler "github.com/jaegertracing/jaeger/pkg/clientcfg/clientcfghttp"
+	"github.com/jaegertracing/jaeger/pkg/healthcheck"
+)
+
+// HTTPServerParams to construct a new Jaeger Collector HTTP Server
+type HTTPServerParams struct {
+	Port            int
+	Handler         handler.JaegerBatchesHandler
+	RecoveryHandler func(http.Handler) http.Handler
```
yurishkuro: why does this need to be exposed?

cmd/collector/app/server/http.go
```diff
@@ -0,0 +1,84 @@
...
+
+	server.Handler = params.RecoveryHandler(r)
+	go func(listener net.Listener, hServer *http.Server) {
+		if err := hServer.Serve(listener); err != nil {
+			if err != http.ErrServerClosed {
+				params.Logger.Fatal("Could not start HTTP collector", zap.Error(err))
+			}
+		}
+		params.HealthCheck.Set(healthcheck.Unavailable)
+	}(listener, server)
```
yurishkuro: variables are already in scope

cmd/collector/main.go
```diff
@@ -85,6 +63,7 @@ func main() {
 			logger := svc.Logger // shortcut
 			baseFactory := svc.MetricsFactory.Namespace(metrics.NSOptions{Name: "jaeger"})
 			metricsFactory := baseFactory.Namespace(metrics.NSOptions{Name: "collector"})
+			strategyStoreFactory.InitFromViper(v)
```
yurishkuro: dup in L77?

cmd/collector/main.go
```diff
@@ -95,82 +74,28 @@ func main() {
...
 
 			svc.RunAndThen(func() {
-				if closer, ok := spanWriter.(io.Closer); ok {
-					server.GracefulStop()
-					err := closer.Close()
-					if err != nil {
-						logger.Error("Failed to close span writer", zap.Error(err))
-					}
-				}
+				c.Close()
```
yurishkuro: doesn't this return an error?

cmd/collector/app/server/grpc.go
jpkrohling: I think this might actually be more suitable for the mains...

cmd/collector/app/server/grpc.go
jpkrohling: You are right. I think I got used to being explicit on this because of the concurrency work I had in another task, and it's good practice to be explicit in such cases, but it doesn't bring any benefits in this case here.

cmd/collector/app/server/http.go
jpkrohling: Not sure I get the question. The `HTTPServerParams` instance is created by the `Collector#Start` in `cmd/collector/app`...

cmd/collector/main.go
jpkrohling: What an eye!

cmd/collector/main.go
jpkrohling: True, I should at least log it.

jpkrohling:
I'll work on @yurishkuro's comments as soon as #2031 is merged, as there's potential for conflicts here.

cmd/collector/app/server/http.go
yurishkuro: the recovery handler is a mechanical thing that can be added silently, it's weird to have it exposed via API like this.

cmd/collector/app/server/grpc.go
jpkrohling: Did you mean to set the copyright headers to `// Copyright (c) 2018 The Jaeger Authors.` here, as it was in the `grpc_server.go` file?

PR #2066, opened 2020-02-10
Clarify that "kafka" is not a real storage backend

yurishkuro:
To address user questions like:

> Hi , I am exploring Jaeger with Kafka as an intermediate storage. I am unable to understand that why do we have Kafka producer as config for Jaeger query component. I believe that Jaeger query would always query the data from either Elastic or Cassandra? Can someone pls clarify? Thanks

### DIFF
--- a/cmd/env/command.go
+++ b/cmd/env/command.go
@@ -39,7 +39,7 @@ The following configuration options are only available via environment variables
 `
 	storageTypeDescription = `The type of backend [%s] used for trace storage.
 Multiple backends can be specified as comma-separated list, e.g. "cassandra,elasticsearch"
-(currently only for writing spans). Note that 'kafka' is not a real storage backend, it is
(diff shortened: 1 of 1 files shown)


cmd/env/command.go
```diff
@@ -36,15 +36,31 @@ command line option                 environment variable
 
 The following configuration options are only available via environment variables:
 %s
+`
+	storageTypeDescription = `The type of backend [%s] used for trace storage.
+Multiple backends can be specified as comma-separated list, e.g. "cassandra,elasticsearch"
+(currently only for writing spans). Note that "kafka" is not a real storage backend, it is
```
vprithvi: I'd rephrase this to say "kafka" is distributed queue and not a complete storage solution and is expected to be used in conjunction with ingester.

yurishkuro: rephrased

Issue #2019, opened 2020-01-16
Switch to go mod


kylesliu:
## Requirement - what kind of business use case are you trying to solve?

Hello everyone, I'm a gopher from China. When using jaeger, I found that there is no go mod . go mod can provide great help to developers in China.

ChadiEM:
@kylesliu is probably talking about [Go Modules](https://github.com/golang/go/wiki/Modules).

pavolloffay:
@yurishkuro are we able to use go mod. I found this PR from the last year https://github.com/jaegertracing/jaeger/pull/1546

yurishkuro:
We can switch this repo to `go mod`, but we will lose the sync of transitive dependencies from jaeger-client-go and jaeger-lib, which are not using `go mod`. Those two are harder to switch because they are not on v1, so would require breaking changes.

pavolloffay:
Do we know which transitive deps of `jaeger-lib` and `jaeger-client-go` are not using go modules?

yurishkuro:
I don't think anyone did such analysis. But it's not what I meant, I meant dependencies like `jaeger -> jaeger-lib -> X`. If `jaeger` is on `go mod`, but `jaeger-lib` is not, then `X` will only be synced at the time of migrating to `go mod`, and after that we won't see updates to X even if `jaeger-lib` bumps it.

pavolloffay:
> We can switch this repo to go mod, but we will lose the sync of transitive dependencies from jaeger-client-go and jaeger-lib, which are not using go mod. Those two are harder to switch because they are not on v1, so would require breaking changes.

Why would it be a breaking change? Does switching of dependency management mean breaking change?

yurishkuro:
Go mod requires /vN directory for libs that are v2+. Making such change to source code makes the libs incompatible with existing consumers that import from root and use other dep mgrs. 

I was originally thinking we could combine switch to go mod in the libs with a move from uber/ to jaegertracing/ path and going back to v1, but that won’t work either because we already have v1 tags in those repos.

PR #2086, opened 2020-02-25
Introduce OpenTelemetry collector

pavolloffay:
Fixes https://github.com/jaegertracing/jaeger-opentelemetry-collector/issues/8


cmd/opentelemetry-collector/main.go
```diff
@@ -0,0 +1,58 @@
...
+		ExeName:  "jaeger-opentelemetry-collector",
+		LongName: "Jaeger OpenTelemetry Collector",
+		// TODO
+		//Version:  version.Version,
+		//GitHash:  version.GitHash,
+	}
+
+	// TODO caused compilation issues due to refactoring in https://github.com/jaegertracing/jaeger/pull/2060
+	// TODO app/HandlerConfig moved to app/handler/HandlerConfig
+	//cmpts, err := defaults.Components()
```
pavolloffay: I could not initialize all default components (importers, processors, exporters) because Jaeger importer imports jaeger/collector/app and there was a breaking change https://github.com/jaegertracing/jaeger/pull/2060

pavolloffay: Once https://github.com/open-telemetry/opentelemetry-collector/pull/570 is merged we should be able to use `defaults.components()`

cmd/opentelemetry-collector/main.go
```diff
@@ -0,0 +1,58 @@
+package main
+
+import (
+	"github.com/open-telemetry/opentelemetry-collector/config"
```
yurishkuro: there should be a corresponding change in Gopkg.toml

pavolloffay:
So importing `opentelemetry-collector` will certainly introduce some _complications_. 

For instance a breaking change in this repo might result in a compilation error (like in https://github.com/jaegertracing/jaeger/pull/2086/files/de7040ace7ce90518b1ff99c32f471a801d6665c#diff-bc7bd0edae17579706fb28fa796dd8cf). 

The opposite approach might be to move Jaeger components here and import them in opentelemetry-collector if we assume that OTEL interfaces will not break (I assume they will break less than Jaeger code).

An alternative solution might be to re-implement jaeger components directly here and use those instead. We could just import a shared Jaeger configuration from OTEL (It would have to be moved to a separate package). I frankly don't like this option.

Also Jaeger uses `go dep` and OTEL go modules. I am not sure if this will cause any problems once we switch Jaeger to go modules https://github.com/jaegertracing/jaeger/pull/1546.

Any thoughts @yurishkuro @tigrannajaryan?

pavolloffay:
So importing `opentelemetry-collector` will certainly introduce some _complications_. 

For instance a breaking change in this repo might result in a compilation error (like in https://github.com/jaegertracing/jaeger/pull/2086/files/de7040ace7ce90518b1ff99c32f471a801d6665c#diff-bc7bd0edae17579706fb28fa796dd8cf). 

The opposite approach might be to move Jaeger components here and import them in opentelemetry-collector if we assume that OTEL interfaces will not break (I assume they will break less than Jaeger code).

An alternative solution might be to re-implement jaeger components directly here and use those instead. We could just import a shared Jaeger configuration from OTEL (It would have to be moved to a separate package). I frankly don't like this option.

Also Jaeger uses `go dep` and OTEL go modules. I am not sure if this will cause any problems once we switch Jaeger to go modules https://github.com/jaegertracing/jaeger/pull/1546.

Any thoughts @yurishkuro @tigrannajaryan?

pavolloffay:
So importing `opentelemetry-collector` will certainly introduce some _complications_. 

For instance a breaking change in this repo might result in a compilation error (like in https://github.com/jaegertracing/jaeger/pull/2086/files/de7040ace7ce90518b1ff99c32f471a801d6665c#diff-bc7bd0edae17579706fb28fa796dd8cf). 

The opposite approach might be to move Jaeger components here and import them in opentelemetry-collector if we assume that OTEL interfaces will not break (I assume they will break less than Jaeger code).

An alternative solution might be to re-implement jaeger components directly here and use those instead. We could just import a shared Jaeger configuration from OTEL (It would have to be moved to a separate package). I frankly don't like this option.

Also Jaeger uses `go dep` and OTEL go modules. I am not sure if this will cause any problems once we switch Jaeger to go modules https://github.com/jaegertracing/jaeger/pull/1546.

Any thoughts @yurishkuro @tigrannajaryan?

pavolloffay:
So importing `opentelemetry-collector` will certainly introduce some _complications_. 

For instance a breaking change in this repo might result in a compilation error (like in https://github.com/jaegertracing/jaeger/pull/2086/files/de7040ace7ce90518b1ff99c32f471a801d6665c#diff-bc7bd0edae17579706fb28fa796dd8cf). 

The opposite approach might be to move Jaeger components here and import them in opentelemetry-collector if we assume that OTEL interfaces will not break (I assume they will break less than Jaeger code).

An alternative solution might be to re-implement jaeger components directly here and use those instead. We could just import a shared Jaeger configuration from OTEL (It would have to be moved to a separate package). I frankly don't like this option.

Also Jaeger uses `go dep` and OTEL go modules. I am not sure if this will cause any problems once we switch Jaeger to go modules https://github.com/jaegertracing/jaeger/pull/1546.

Any thoughts @yurishkuro @tigrannajaryan?

cmd/opentelemetry-collector/main.go
tigrannajaryan: I commented on https://github.com/open-telemetry/opentelemetry-collector/pull/570 and we can merge it quickly if it blocks you.

tigrannajaryan:
> So importing `opentelemetry-collector` will certainly introduce some _complications_.
> 
> For instance a breaking change in this repo might result in a compilation error (like in https://github.com/jaegertracing/jaeger/pull/2086/files/de7040ace7ce90518b1ff99c32f471a801d6665c#diff-bc7bd0edae17579706fb28fa796dd8cf).
> 
> The opposite approach might be to move Jaeger components here and import them in opentelemetry-collector if we assume that OTEL interfaces will not break (I assume they will break less than Jaeger code).

This would mean this repo references opentelemetry-collector and opentelemetry-collector references this repo. I do not know if go modules allow a circular references between repos, but even if it does it will not look nice. I'd rather not do it.

Alternatively we could references Jaeger components located in this repo from opentelemetry-collector-contrib. That would no longer be a circular reference, however it would put Jaeger out of core build, which I would not want to do (among other things we committed to having built-in support for Jaeger in Collector core).

> 
> An alternative solution might be to re-implement jaeger components directly here and use those instead. We could just import a shared Jaeger configuration from OTEL (It would have to be moved to a separate package). I frankly don't like this option.

Yeah, this does not sound like a good solution. Lots of code duplication.

> 
> Also Jaeger uses `go dep` and OTEL go modules. I am not sure if this will cause any problems once we switch Jaeger to go modules #1546.

I have no experience mixing the 2 approaches.

yurishkuro:
@pavolloffay can you enumerate which parts of Jaeger the OTel Collector imports?

pavolloffay:
```
// exporter, receiver
// receiver, translator
	"github.com/jaegertracing/jaeger/thrift-gen/baggage"
	"github.com/jaegertracing/jaeger/thrift-gen/jaeger"
	"github.com/jaegertracing/jaeger/thrift-gen/jaeger"

// receiver
	"github.com/jaegertracing/jaeger/cmd/agent/app/configmanager"
	"github.com/jaegertracing/jaeger/cmd/agent/app/configmanager/grpc"
	"github.com/jaegertracing/jaeger/cmd/agent/app/httpserver"
	"github.com/jaegertracing/jaeger/cmd/agent/app/processors"
	"github.com/jaegertracing/jaeger/cmd/agent/app/reporter"
	"github.com/jaegertracing/jaeger/cmd/agent/app/servers"
	"github.com/jaegertracing/jaeger/cmd/agent/app/servers/thriftudp"
	"github.com/jaegertracing/jaeger/cmd/collector/app/handler"
	"github.com/jaegertracing/jaeger/cmd/collector/app/sampling"
	"github.com/jaegertracing/jaeger/plugin/sampling/strategystore/static"

// translator
```

annanay25:
Is it possible to extract the common jaeger components into packages, say at `github.com/jaegertracing/jaeger/pkg` and have both Jaeger and OTel import these?

yurishkuro:
It is possible, but I don’t like it. It will make jaeger development harder by moving critical parts to another repo.

pavolloffay:
> This would mean this repo references opentelemetry-collector and opentelemetry-collector references this repo. I do not know if go modules allow a circular references between repos, but even if it does it will not look nice. I'd rather not do it.

One way or another there will be a cyclic dependency if we want to have the jaeger-opentelemetry-collector in this repository.
I don't know how it exactly works, maybe the cycle is not allowed only per packages and can be used per repositories.


I think it will be better to host as much as possible Jaeger related code in Jaeger repository. We could have OTEL importer, processor, exporter like API implementation in Jaeger (not direct import of OTEL). Then the actual Jaeger importer, processor, exporter will use this package to implement OTEL components.  This way if we break API in Jaeger it will not result in a compilation error like it did now. OpenTelemetry will be able to break API too as the jaeger importers, processors will be implemented directly in the OTEL.

yurishkuro:
@pavolloffay maybe it’s worth trying these cross-repo imports on a couple toy projects using go mod.

pavolloffay:
If we agree that Jaeger can be moved to `go mod` I can try directly with Jaeger and OTEL. Actually I have started some work already for https://github.com/jaegertracing/jaeger/issues/2019.

tigrannajaryan:
> One way or another there will be a cyclic dependency if we want to have the jaeger-opentelemetry-collector in this repository.
> I don't know how it exactly works, maybe the cycle is not allowed only per packages and can be used per repositories.

If you use go.mod then every go.mod containing directory is considered a separate module (minus subdirectories that contain their own go.mod files). So even though it is a circular dependency between repos, it will no longer be considered a circular dependency between go modules if you structure them appropriately (and I think go tooling should be fine with it, but don't trust my word, I never tried this actual setting).


> 
> I think it will be better to host as much as possible Jaeger related code in Jaeger repository. We could have OTEL importer, processor, exporter like API implementation in Jaeger (not direct import of OTEL). Then the actual Jaeger importer, processor, exporter will use this package to implement OTEL components. This way if we break API in Jaeger it will not result in a compilation error like it did now. OpenTelemetry will be able to break API too as the jaeger importers, processors will be implemented directly in the OTEL.

Makes sense.

pavolloffay:
First we should migrate to go modules https://github.com/jaegertracing/jaeger/pull/2098.

Based on this PR I have tried importing Jaeger in OTEL via override, the build worked and I was able to build Jaeger with imported OTEL (This is the option 2). 

Then there are two approaches we can take:

1. define a separate module for jaeger-otel-collector. This module will import Jaeger and OTEL. So if there are breaking changes in Jaeger we will be able to coordinate fix in OTEL and subsequently build jaeger-otel collector with bumped versions of OTEL and jaeger.

2. Do not define go module for jaeger-otelcollector and just simply import OTEL in jaeger. This is more dangerous and we could run into build problems.  These problems could be minimized by moving most of the Jaeger code from OTEL to Jaeger. Then the Jaeger and OTEL would import this code and use it to implement importers, processors, exporters.

Option 1. seems like a better option to` me. Although, I found several comments on github discouraging this approach especially when the module should serve as a library or API, but we are just trying to use it for the main package mainly. I didn't find any exact reasons why.  I will have to experiment more with this.
Here are some references where go submodules have been discussed 
* https://groups.google.com/forum/#!msg/prometheus-developers/F1Vp0rLk3TQ/F412gRw0CAAJ
* https://github.com/coreos/prometheus-operator/issues/2988#issuecomment-581530000
* https://github.com/golang/go/issues/27056

yurishkuro:
> define a separate module for jaeger-otel-collector

what do you mean by that?

pavolloffay:
Use go submodules. Multiple go modules in a single repository.

cmd/opentelemetry-collector/go.mod
```diff
@@ -0,0 +1,9 @@
+module github.com/jaegertracing/jaeger/cmd/opentelemetry-collector
```
pavolloffay: @yurishkuro this is an example fo go submodule. 

By default Jaeger version would be resolved from OTEL collector, but with `replace` we are able to point it to the current master/head. If we do a breaking change we can pin the version to a version prior the change and collaborate with OTEL to update the code there.

pavolloffay:
@yurishkuro we should make progress on this one. The work in https://github.com/jaegertracing/jaeger-opentelemetry-collector will require changes in the core. For instance changing storage writers to work on the batches rather than on a single spans https://github.com/jaegertracing/jaeger-opentelemetry-collector/issues/15.

yurishkuro:
is there an open question?

pavolloffay:
There aren't any open questions. It just needs to be merged if we agree that this can go in.

A separate go module for the otel collector is the safest path for now. There are still breaking changes happening if there wasn't we could import it in the root package.

There was a discussion regarding the versioning. The jaeger-otel collector will follow the version schema for with jaeger binaries, so with regards to this point it is also fine.

cmd/opentelemetry-collector/main.go
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) 2020 The Jaeger Authors.
```
yurishkuro: what is the binary file above? probably don't want it checked in.

.travis.yml
```diff
@@ -36,6 +36,9 @@ matrix:
   - go: "1.13.x"
     env:
     - HOTROD=true
+  - go: "1.13.x"
```
yurishkuro: Why do we need a separate matrix step for it? I'd rather include it in the main build.

pavolloffay: +1 this was just to quickly test the build. I will remove it

PR #2089, opened 2020-02-26
Add pcc64le to binary builds

redmark-alt:
- 

-

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,8 +19,6 @@ Changes by Version
 
 ### Backend Changes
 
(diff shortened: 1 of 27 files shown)


redmark-alt:
I seem to be stuck in a DCO loop... each time I follow the steps to correct it I get the same issue.

pavolloffay:
@redmark-alt I would suggest squashing all commits to a single commit and then sign it. 

Personally I would use interactive rebase to squash the commits and then `g rebase HEAD~1 && g add -A && g commit -s "message" -s`, or something similar.

PR #2090, opened 2020-02-26
Do not skip service/operation indexing for firehose spans + a couple fixes

yurishkuro:
The transitive dep graphs view uses service & operation index to populate the dropdown. However, when traces are stored in the firehose mode, they are excluded from indexing (to reduce the load on storage), including the service/operation tables, which are actually very inexpensive to write because writes only happen once per 12hr period due to a cache in front of them.

Allow service/operation tables to be written even in firehose mode.

### DIFF
--- a/cmd/agent/app/reporter/client_metrics_test.go
+++ b/cmd/agent/app/reporter/client_metrics_test.go
@@ -256,15 +256,16 @@ func TestClientMetricsReporter_Expire(t *testing.T) {
 			t.Run(fmt.Sprintf("iter%d:gauge=%d,log=%s", i, test.expGauge, test.expLog), func(t *testing.T) {
 				// Expire loop runs every 100us, and removes the client after 5ms.
 				// We check for condition in each test for up to 5ms (10*500us).
+				var gaugeValue int64 = -1
 				for i := 0; i < 10; i++ {
 					_, gauges := tr.mb.Snapshot()
-					if gauges["client_stats.connected_clients"] == int64(test.expGauge) {
+					gaugeValue = gauges["client_stats.connected_clients"]
+					if gaugeValue == int64(test.expGauge) {
 						break
 					}
 					time.Sleep(500 * time.Microsecond)
 				}
-				tr.mb.AssertGaugeMetrics(t,
-					metricstest.ExpectedMetric{Name: "client_stats.connected_clients", Value: test.expGauge})
+				assert.EqualValues(t, test.expGauge, gaugeValue)
 				tr.assertLog(t, test.expLog, clientUUID)
 
 				// sleep between tests long enough to exceed the 5ms TTL.
--- a/plugin/storage/cassandra/schema/docker.sh
+++ b/plugin/storage/cassandra/schema/docker.sh
@@ -3,7 +3,7 @@
 # This script is used in the Docker image jaegertracing/jaeger-cassandra-schema
 # that allows installing Jaeger keyspace and schema without installing cqlsh.
 
-CQLSH=${CQLSH:-"/usr/bin/cqlsh"}
+CQLSH=${CQLSH:-"cqlsh"}
 CQLSH_HOST=${CQLSH_HOST:-"cassandra"}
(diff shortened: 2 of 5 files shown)


plugin/storage/cassandra/spanstore/writer.go
```diff
@@ -181,7 +186,10 @@ func (s *SpanWriter) writeIndexes(span *model.Span, ds *dbmodel.Span) error {
 		// should this be a soft failure?
 		return s.logError(ds, err, "Failed to insert service name and operation name", s.logger)
 	}
+	return nil
+}
 
+func (s *SpanWriter) writeOtherIndexes(span *model.Span, ds *dbmodel.Span) error {
```
vprithvi: Could you use a more descriptive name here?

yurishkuro: like what?

plugin/storage/cassandra/spanstore/writer_test.go
```diff
@@ -352,23 +353,36 @@ func TestStorageMode_IndexOnly_WithFilter(t *testing.T) {
...
 		w.session.AssertExpectations(t)
-		w.session.AssertNotCalled(t, "Query", stringMatcher(serviceOperationIndex))
 		w.session.AssertNotCalled(t, "Query", stringMatcher(serviceNameIndex))
 		w.session.AssertNotCalled(t, "Query", stringMatcher(durationIndex))
+		assert.Equal(t, "planet-express", serviceWritten.Load())
+		assert.Equal(t, dbmodel.Operation{
+			ServiceName:   "planet-express",
+			SpanKind:      "",
+			OperationName: "package-delivery",
+		}, operationWritten.Load())
```
vprithvi: Are these not tested elsewhere? For e.g., if plugin/storage/cassandra/spanstore/writer.go L183 were removed - wouldn't other tests fail?

plugin/storage/cassandra/spanstore/writer.go
vprithvi: Perhaps this could just be `writeIndexes` like earlier, but it calls `writeServiceOperationIndex`? Might simplify the branching in `writer.go` as well.

plugin/storage/cassandra/spanstore/writer_test.go
yurishkuro: I am not testing whether index writing works, but whether it's invoked for firehose span. This is the only test that makes distinction between two types of indices.

cmd/agent/app/reporter/client_metrics_test.go
```diff
@@ -256,15 +256,16 @@ func TestClientMetricsReporter_Expire(t *testing.T) {
...
-					if gauges["client_stats.connected_clients"] == int64(test.expGauge) {
+					gaugeValue = gauges["client_stats.connected_clients"]
+					if gaugeValue == int64(test.expGauge) {
 						break
 					}
 					time.Sleep(500 * time.Microsecond)
 				}
-				tr.mb.AssertGaugeMetrics(t,
-					metricstest.ExpectedMetric{Name: "client_stats.connected_clients", Value: test.expGauge})
+				assert.EqualValues(t, test.expGauge, gaugeValue)
```
yurishkuro: Unrelated change to fix a flaky test. There was a race condition with gauge value being 1 inside the loop and then dropping back to 0 by the time the AssertGaugeMetrics was called.

plugin/storage/cassandra/schema/docker.sh
```diff
@@ -3,7 +3,7 @@
 # This script is used in the Docker image jaegertracing/jaeger-cassandra-schema
 # that allows installing Jaeger keyspace and schema without installing cqlsh.
 
-CQLSH=${CQLSH:-"/usr/bin/cqlsh"}
+CQLSH=${CQLSH:-"/opt/cassandra/bin/cqlsh"}
```
yurishkuro: unrelated change to fix regression in Cassandra tests that were breaking CI.

PR #2098, opened 2020-02-28
Migrate project to go modules

### LINKED ISSUE #2019: Switch to go mod

pavolloffay:
Resolves #2019 
Supersedes: #1546


CONTRIBUTING.md
```diff
@@ -88,7 +84,7 @@ github.com/jaegertracing/jaeger
     jaeger/
     sampling/
     zipkincore/
-  Gopkg.toml                - Dep is the project's dependency manager
+  go.mod                    - Go module file to track dependencies
   mkdocs.yml                - MkDocs builds the documentation in docs/
```
yurishkuro: this should be removed, left-over

Makefile
```diff
@@ -377,8 +367,8 @@ thrift: idl/thrift/jaeger.thrift thrift-image
 	[ -d $(THRIFT_GEN_DIR) ] || mkdir $(THRIFT_GEN_DIR)
 	$(THRIFT) -o /data --gen go:$(THRIFT_GO_ARGS) --out /data/$(THRIFT_GEN_DIR) /data/idl/thrift/agent.thrift
 #	TODO sed is GNU and BSD compatible
-	sed -i.bak 's|"zipkincore"|"$(PROJECT_ROOT)/thrift-gen/zipkincore"|g' $(THRIFT_GEN_DIR)/agent/*.go
-	sed -i.bak 's|"jaeger"|"$(PROJECT_ROOT)/thrift-gen/jaeger"|g' $(THRIFT_GEN_DIR)/agent/*.go
+	sed -i.bak 's|"zipkincore"|"github.com/jaegertracing/jaeger/thrift-gen/zipkincore"|g' $(THRIFT_GEN_DIR)/agent/*.go
+	sed -i.bak 's|"jaeger"|"github.com/jaegertracing/jaeger/thrift-gen/jaeger"|g' $(THRIFT_GEN_DIR)/agent/*.go
```
yurishkuro: can we keep using var instead of explicit path?

Makefile
```diff
@@ -500,10 +491,10 @@ proto:
...
 	go install \
-		./vendor/github.com/golang/protobuf/protoc-gen-go \
-		./vendor/github.com/gogo/protobuf/protoc-gen-gogo \
-		./vendor/github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway \
-		./vendor/github.com/grpc-ecosystem/grpc-gateway/protoc-gen-swagger
+		 github.com/golang/glog \
+		 github.com/golang/protobuf/protoc-gen-go \
+		 github.com/gogo/protobuf/protoc-gen-gogo \
+		 github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway \
+		 github.com/grpc-ecosystem/grpc-gateway/protoc-gen-swagger
```
yurishkuro: How does `go install` work with modules? Is it going to install the version from go.mod?

Perhaps we can wait for @annanay25 to change the gen to use Docker image, then we won't have to do these things here.

examples/hotrod/README.md
```diff
@@ -48,7 +48,6 @@ Jaeger UI can be accessed at http://localhost:16686.
 ```bash
 go get github.com/jaegertracing/jaeger
 cd $GOPATH/src/github.com/jaegertracing/jaeger
```
yurishkuro: these two should probably be `git clone` and `cd jaeger`, rather than relying on GOPATH.

tools.go
```diff
@@ -0,0 +1,34 @@
...
+// limitations under the License.
+
+// +build tools
+
+package jaeger
+
+// https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module
+import (
+	_ "honnef.co/go/tools/cmd/staticcheck"
+
```
yurishkuro: why blank line?

tools.go
```diff
@@ -0,0 +1,34 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// +build tools
+
+package jaeger
+
+// https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module
```
yurishkuro: nice!

Makefile
pavolloffay: > How does go install work with modules? Is it going to install the version from go.mod?

That is my understanding.

> Perhaps we can wait for @annanay25 to change the gen to use Docker image, then we won't have to do these things here.

I don't want to block this PR on that. That can be done later.

tools.go
pavolloffay: Our auto-formatting magic :)

Makefile
pavolloffay: +1 I have changed it back

pavolloffay:
@yurishkuro PR updated based on the comments.

Makefile
yurishkuro: >> Is it going to install the version from go.mod?
> That is my understanding.

Can we find some evidence to that?

kylesliu:
star

Makefile
pavolloffay: For instance version of `github.com/sectioneight/md-to-godoc` is defined in `go.mod`. The never version does not work due to conflicts. I specified the older version explicitly like we had in `dep`.

pavolloffay:
As it is approved and the build is passing I will merge. Feel free to comment after the merge I will apply fixes in subsequent PR(s).

PR #2099, opened 2020-02-28
Build binaries for Linux on IBM Power pcc64le architecture

redmark-alt:
Signed-off-by: Ron Edmark <redmark@us.ibm.com>

- No issue opened for this PR.

- Added support to build binaries on IBM Power ppc64le... similar to the work done here  #1982

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,8 +22,8 @@ ALL_SRC := $(shell find . -name '*.go' \
 # ALL_PKGS is used with 'go cover' and 'golint'
 ALL_PKGS := $(shell go list $(sort $(dir $(ALL_SRC))))
 UNAME := $(shell uname -m)
-#Race flag is not supported on s390x architecture
-ifeq ($(UNAME), s390x)
+#Race flag is not supported on s390x/ppc64le architecture
+ifeq ($(UNAME),$(filter $(UNAME), s390x ppc64le))
 	RACE=
 else
 	RACE=-race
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -22,8 +22,8 @@ ALL_SRC := $(shell find . -name '*.go' \
 # ALL_PKGS is used with 'go cover' and 'golint'
 ALL_PKGS := $(shell go list $(sort $(dir $(ALL_SRC))))
 UNAME := $(shell uname -m)
-#Race flag is not supported on s390x architecture
-ifeq ($(UNAME), s390x)
+#Race flag is not supported on s390x/ppc64le architecture
+ifeq ($(UNAME),$(filter $(UNAME), s390x ppc64le))
```
yurishkuro: could we DRY this more?  Perhaps define a function

redmark-alt: Are you looking for a Makefile function for the RACE variable assignment?

yurishkuro: A function that will encapsulate the fact that we're checking two `uname`s instead of one, so that if we add a 3rd one in the future there'd be just one place to change in the Makefile.

redmark-alt: I'm not a Makefile expert, if you can point me to an example I will review it and apply the changes as required.

yurishkuro: I'm not an expert either. Here's example of functions: https://coderwall.com/p/cezf6g/define-your-own-function-in-a-makefile

At minimum, we could consider moving the list of "special" architectures into a variable so that, again, only one place needs to change.

clnperez: actually -- and sorry for not catching this sooner, the race flag has been supported for Power for a few releases now. I've asked someone else to carry this PR and will have her link to this one when she submits. Thanks @yurishkuro for the review.

PR #2102, opened 2020-02-29
Switch build to use jaegertracing/protobuf image

### LINKED ISSUE #1213: Find a way to ignore gogoproto annotations in the IDL

Use `model/model.proto` to generate code in other languages.

The `model/model.proto` currently contains a lot of gogoproto options needed to make the internal domain model types more efficient. While these annotations are optional for other languages, the file does not compile if the compiler does not have access to the gogoproto IDLs.

There are two ways we can go about solving this problem:

1. Write a script that strips the gogoproto annotations before compiling the IDL in other languages.
1. Create a reusable Docker image for generating code in all languages, which includes all the dependent IDLs, including gogoproto.

annanay25:

- Fixes #1213 

- Swap to use the docker image and generate swagger api docs.


.travis.yml
```diff
@@ -77,7 +75,7 @@ script:
   - if [ "$KAFKA_INTEGRATION_TEST" == true ]; then bash ./scripts/travis/kafka-integration-test.sh ; else echo 'skipping kafka integration test'; fi
   - if [ "$CASSANDRA_INTEGRATION_TEST" == true ]; then bash ./scripts/travis/cassandra-integration-test.sh ; else echo 'skipping cassandra integration test'; fi
   - if [ "$HOTROD" == true ]; then bash ./scripts/travis/hotrod-integration-test.sh ; else echo 'skipping hotrod example'; fi
-  - if [ "$PROTO_GEN_TEST" == true ]; then make proto PROTOC=$HOME/protoc/bin/protoc && git diff --name-status --exit-code ; else echo 'skipping protoc validation'; fi
+  - if [ "$PROTO_GEN_TEST" == true ]; then make proto && git diff --name-status --exit-code ; else echo 'skipping protoc validation'; fi
```
annanay25: `PROTO_GEN_TEST` is passing, but I'm wondering if we should preserve this step.. in order to double check that the docker image is synced with the main repository.

Could you PTAL @yurishkuro?

Makefile
```diff
@@ -472,23 +471,24 @@ proto:
 	$(PROTOC) \
 		$(PROTO_INCLUDES) \
 		--gogo_out=plugins=grpc,$(PROTO_GOGO_MAPPINGS):$(PWD)/proto-gen/ \
+		--swagger_out=allow_merge=true:$(PWD)/proto-gen/openapi/ \
```
yurishkuro: let's not do the swagger

annanay25: Just curious, why not?

yurishkuro: because it gives the impression that it is somehow the officially supported JSON, which is not the case.

.travis.yml
yurishkuro: Yes, we need to preserve this step. It validates that the generated classes are up to date.

Makefile
```diff
@@ -419,14 +421,11 @@ generate-mocks: install-mockery
...
 PROTO_INCLUDES := \
-	-I model/proto \
-	-I idl/proto \
-	-I vendor/github.com/grpc-ecosystem/grpc-gateway \
-	-I vendor/github.com/gogo/googleapis \
-	-I vendor/github.com/gogo/protobuf/protobuf \
-	-I vendor/github.com/gogo/protobuf
+	-Imodel/proto \
+	-Iidl/proto \
+	-I/usr/include/github.com/gogo/protobuf
```
yurishkuro: I'm curious, why is this last import needed? Isn't it already present in the Docker image?

pavolloffay:
@annanay25 can be target `proto-install` removed now?

annanay25:
@pavolloffay - Yes, let me fix that with a follow up PR.

Makefile
annanay25: The Docker image has `-I/usr/include` but I think we needed this relative path as well

PR #2104, opened 2020-03-02
Remove proto-install target from Makefile

annanay25:
- Follow up from #2102 

- Remove `proto-install` step from Makefile and `.travis.yml` as proto-gen now uses new docker image

/cc @pavolloffay 


pavolloffay:
@annanay25 there is a conflict

PR #2112, opened 2020-03-03
Move all tchannel packages to a single top level package

pavolloffay:
Related to #2105 
Supersedes #2109 

The top level package is `./tchannel`. I have kept the original package structure. e.g. `./tchannel/agent/app/reporter/tchannel`.


cmd/agent/app/reporter/flags.go
```diff
@@ -48,19 +46,24 @@ type Options struct {
...
 func AddFlags(flags *flag.FlagSet) {
-	flags.String(reporterType, string(GRPC), fmt.Sprintf("Reporter type to use e.g. %s, %s[%s]", string(GRPC), string(TCHANNEL), "NOTE: Deprecated since 1.16"))
-	if !setupcontext.IsAllInOne() {
-		flags.String(agentTagsDeprecated, "", "(deprecated) see --"+agentTags)
-		flags.String(agentTags, "", "One or more tags to be added to the Process tags of all spans passing through this agent. Ex: key1=value1,key2=${envVar:defaultValue}")
+	AddFlagsWithCustomTypes([]string{})(flags)
+}
+
+// AddFlagsWithCustomTypes adds types to use message.
+func AddFlagsWithCustomTypes(types []string) func(*flag.FlagSet) {
```
vprithvi: Why do we need `types`? 

Also - if `GRPC` is the only supported reporter, we can deprecate the flag so that we can remove it later

pavolloffay: I made it backwars compatible, so in Uber's custom build the type can be added back.

> Also - if GRPC is the only supported reporter, we can deprecate the flag so that we can remove it later

I would leave it for now as it is. It gives us more flexibility to extend.

pavolloffay:
The coverage is failing bc of the new files in `tchannel` package I will not cover those as It will be eventually removed.

cmd/agent/app/reporter/flags.go
```diff
@@ -48,19 +46,24 @@ type Options struct {
...
+	return func(flags *flag.FlagSet) {
+		types = append([]string{string(GRPC)}, types...)
+		flags.String(reporterType, string(GRPC), fmt.Sprintf("Reporter type to use e.g. %s", types))
+		if !setupcontext.IsAllInOne() {
+			flags.String(agentTagsDeprecated, "", "(deprecated) see --"+agentTags)
```
yurishkuro: https://github.com/jaegertracing/jaeger/issues/2113

cmd/agent/main.go
```diff
@@ -57,7 +57,14 @@ func main() {
 			rOpts := new(reporter.Options).InitFromViper(v, logger)
 			tchanBuilder := tchannel.NewBuilder().InitFromViper(v, logger)
 			grpcBuilder := grpc.NewConnBuilder().InitFromViper(v)
-			cp, err := app.CreateCollectorProxy(rOpts, tchanBuilder, grpcBuilder, logger, mFactory)
+			builders := map[reporter.Type]app.ProxyBuilder{}
+			builders[reporter.GRPC] = func(m map[string]string, factory metrics.Factory, logger *zap.Logger) (app.CollectorProxy, error) {
+				return grpc.NewCollectorProxy(grpcBuilder, rOpts.AgentTags, mFactory, logger)
```
yurishkuro: I would prefer to have a function in each package that matches the `ProxyBuilder` signature, rather than using lambdas here.

cmd/agent/main.go
```diff
@@ -90,7 +97,7 @@ func main() {
 		command,
 		svc.AddFlags,
 		app.AddFlags,
-		reporter.AddFlags,
+		reporter.AddFlagsWithCustomTypes([]string{"(deprecated)" + string(tchannel.ReporterType)}),
```
yurishkuro: let's think more about this, I find it quite confusing

tchannel/collector/app/flags.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"flag"
+
+	"github.com/spf13/viper"
+)
+
+const (
+	// CollectorTChannel is the default port for TChannel server for sending spans
+	CollectorTChannel = 14267
+	collectorPort     = "collector.port"
```
yurishkuro: It looks like we're going back to the deprecated flag name? Maybe we can remove the old one, as it's been quite a while since we deprecated it.

cmd/agent/app/builder.go
```diff
@@ -217,25 +215,16 @@ func defaultInt(value int, defaultVal int) int {
 // CreateCollectorProxy creates collector proxy
 func CreateCollectorProxy(
 	opts *reporter.Options,
-	tchanBuilder *tchannel.Builder,
-	grpcBuilder *grpc.ConnBuilder,
+	builders map[reporter.Type]ProxyBuilder,
```
yurishkuro: 👍

tchannel/collector/app/flags.go
pavolloffay: The intent of this PR is to just move tchannel code to a separate package. The component itself will be removed in the follow-up PR along with flags bindings.

cmd/agent/app/reporter/flags.go
pavolloffay: I have added a better description to the signature

cmd/agent/main.go
pavolloffay: I have created `proxy_builders.go` in `agent/app` package. They cannot live in the reporter's package bc it creates cyclic imports 

```
# github.com/jaegertracing/jaeger/cmd/agent/app/processors
FAIL	github.com/jaegertracing/jaeger/cmd/agent/app/processors [setup failed]
import cycle not allowed in test
FAIL
package github.com/jaegertracing/jaeger/cmd/agent/app/processors (test)
	imports github.com/jaegertracing/jaeger/tchannel/agent/app/reporter/tchannel
	imports github.com/jaegertracing/jaeger/cmd/agent/app
	imports github.com/jaegertracing/jaeger/cmd/agent/app/processors



GOROOT=/home/ploffay/bin/go #gosetup
GOPATH=/home/ploffay/projects/golang #gosetup
/home/ploffay/bin/go/bin/go test -c -o /tmp/___go_test_github_com_jaegertracing_jaeger_cmd_agent_app github.com/jaegertracing/jaeger/cmd/agent/app #gosetup
FAIL	github.com/jaegertracing/jaeger/cmd/agent/app [setup failed]
# github.com/jaegertracing/jaeger/cmd/agent/app
FAIL
import cycle not allowed in test
package github.com/jaegertracing/jaeger/cmd/agent/app (test)
	imports github.com/jaegertracing/jaeger/tchannel/agent/app/reporter
	imports github.com/jaegertracing/jaeger/cmd/agent/app

```

yurishkuro: I wasn't suggesting importing the function type, just having functions in tch/grpc packages that match it on the signature.

pavolloffay: There was also a suggestion to pass pass `Options` instead of listing the parameters. That creates cycle too.

yurishkuro: oh, but the struct, I see...

cmd/agent/app/reporter/flags.go
```diff
@@ -46,9 +44,16 @@ type Options struct {
...
+	// Reporters which area added to reporter type help message
+	Reporters []string
+}
+
 // AddFlags adds flags for Options.
-func AddFlags(flags *flag.FlagSet) {
-	flags.String(reporterType, string(GRPC), fmt.Sprintf("Reporter type to use e.g. %s, %s[%s]", string(GRPC), string(TCHANNEL), "NOTE: Deprecated since 1.16"))
+func (f Flags) AddFlags(flags *flag.FlagSet) {
+	reps := append([]string{string(GRPC)}, f.Reporters...)
+	flags.String(reporterType, string(GRPC), fmt.Sprintf("Reporter type to use e.g. %s", reps))
```
yurishkuro: if all this complication just to include tchannel in the possible reporter types in the help string, I would say it's not worth it.

cmd/agent/app/reporter/grpc/collector_proxy.go
```diff
@@ -31,7 +31,7 @@ type ProxyBuilder struct {
 	conn     *grpc.ClientConn
 }
 
-// NewCollectorProxy creates ProxyBuilder
+// NewCollectorProxy creates CollectorProxyBuilder
```
yurishkuro: should revert this

cmd/agent/main.go
```diff
@@ -57,7 +57,14 @@ func main() {
 			rOpts := new(reporter.Options).InitFromViper(v, logger)
 			tchanBuilder := tchannel.NewBuilder().InitFromViper(v, logger)
 			grpcBuilder := grpc.NewConnBuilder().InitFromViper(v)
-			cp, err := app.CreateCollectorProxy(rOpts, tchanBuilder, grpcBuilder, logger, mFactory)
+			builders := map[reporter.Type]app.CollectorProxyBuilder{}
+			builders[reporter.GRPC] = app.GRPCCollectorProxyBuilder(grpcBuilder)
+			builders[tchannel.ReporterType] = app.TCollectorProxyBuilder(tchanBuilder)
```
yurishkuro:
```suggestion
			builders := map[reporter.Type]app.CollectorProxyBuilder{
			    reporter.GRPC: app.GRPCCollectorProxyBuilder(grpcBuilder),
			    tchannel.ReporterType: app.TCollectorProxyBuilder(tchanBuilder),
			}
```
can we do this instead?

tchannel/collector/app/flags.go
yurishkuro: nvm, I though we had another option similar to
```
      --collector.grpc-port int                         The gRPC port for the collector service (default 14250)
```

cmd/all-in-one/main.go
```diff
@@ -130,11 +131,22 @@ by default uses only in-memory database.`,
...
+			tc := tCollector.Collector{}
+			tc.Start("jaeger-collector", tCollectorOpts, logger, c.SpanHandlers(), strategyStore)
 
 			// agent
 			grpcBuilder.CollectorHostPorts = append(grpcBuilder.CollectorHostPorts, fmt.Sprintf("127.0.0.1:%d", cOpts.CollectorGRPCPort))
 			agentMetricsFactory := metricsFactory.Namespace(metrics.NSOptions{Name: "agent", Tags: nil})
-			cp, err := agentApp.CreateCollectorProxy(repOpts, tchanBuilder, grpcBuilder, logger, agentMetricsFactory)
+			builders := map[agentRep.Type]agentApp.CollectorProxyBuilder{}
+			builders[agentRep.GRPC] = agentApp.GRPCCollectorProxyBuilder(grpcBuilder)
+			builders[agentTchanRep.ReporterType] = agentApp.TCollectorProxyBuilder(tchanBuilder)
```
yurishkuro: can we use map initializer syntax?

cmd/all-in-one/main.go
```diff
@@ -130,11 +131,22 @@ by default uses only in-memory database.`,
 				HealthCheck:    svc.HC(),
 			})
 			c.Start(cOpts)
+			tCollectorOpts := new(tCollector.Options).InitFromViper(v)
+			tc := tCollector.Collector{}
+			tc.Start("jaeger-collector", tCollectorOpts, logger, c.SpanHandlers(), strategyStore)
```
yurishkuro: I would do this instead:
```suggestion
			tc := tCollector.New("jaeger-collector", tCollectorOpts, logger, c.SpanHandlers(), strategyStore)
```

cmd/agent/app/reporter/flags.go
pavolloffay: Yes just for the help message, I am removing it

pavolloffay:
@yurishkuro comments addressed and PR updated

yurishkuro:
close  #2109 ?

I am waiting on our RPC team to resolve the conflict with grpc 1.26, then we can pull in this change to test, and then switch to the internal copy of the new tchannel package.

PR #2115, opened 2020-03-05
Remove tchannel between agent and collector

### LINKED ISSUE #2105: Remove tchannel reporter from agent

pavolloffay:
Resolves  #2105 


cmd/agent/main.go
```diff
@@ -55,11 +54,9 @@ func main() {
 				Namespace(metrics.NSOptions{Name: "agent"})
 
 			rOpts := new(reporter.Options).InitFromViper(v, logger)
-			tchanBuilder := tchannel.NewBuilder().InitFromViper(v, logger)
 			grpcBuilder := grpc.NewConnBuilder().InitFromViper(v)
 			builders := map[reporter.Type]app.CollectorProxyBuilder{
```
annanay25: Do we need a map here now? We have just one exporter

cmd/collector/main.go
```diff
@@ -36,7 +36,6 @@ import (
 	ss "github.com/jaegertracing/jaeger/plugin/sampling/strategystore"
 	"github.com/jaegertracing/jaeger/plugin/storage"
 	"github.com/jaegertracing/jaeger/ports"
-	tCollector "github.com/jaegertracing/jaeger/tchannel/collector/app"
```
annanay25: Is this issue also dealing with removing the `tchannel` endpoint from the collector? End user applications will not be able to send spans directly to the collector? This is a breaking change and should be documented

pavolloffay: End applications are not using this port. This is used only by the agent.

cmd/agent/main.go
pavolloffay: It's needed for extensibility. E.g. at uber they will add custom reporters.

CHANGELOG.md
```diff
@@ -6,6 +6,24 @@ Changes by Version
 
 ### Backend Changes
```
yurishkuro: Maybe it should go in breaking changes

PR #2119, opened 2020-03-10
Clock skew config

### LINKED ISSUE #2053: Restrict clock skew adjustment to a certain max delta
Similar to #961, perhaps we should impose a limit on the skew adjustment, e.g. no more than 1s. If the calculated delta exceeds that limit, add a span warning saying that adjustment was not done, and include the proposed delta in the message.

joe-elliott:
- Fixes #2053 

- Added duration parameter `query.max-clock-skew-adjust`
- Wiring to pass into adjusters.ClockSkew()
- Tests to confirm functionality

Note that I consolidated the creation of `QueryServiceOptions` in the `cmd/query/app` package.  I don't necessarily think this is a great place for it, but I struggled to find a better one.  If you have recommendations please let me know.

### DIFF
--- a/cmd/query/app/flags.go
+++ b/cmd/query/app/flags.go
@@ -28,8 +28,11 @@ import (
 	"github.com/spf13/viper"
 	"go.uber.org/zap"
 
+	"github.com/jaegertracing/jaeger/cmd/query/app/querysvc"
+	"github.com/jaegertracing/jaeger/model/adjuster"
 	"github.com/jaegertracing/jaeger/pkg/config"
 	"github.com/jaegertracing/jaeger/ports"
+	"github.com/jaegertracing/jaeger/storage"
 )
 
 const (
@@ -88,6 +91,24 @@ func (qOpts *QueryOptions) InitFromViper(v *viper.Viper, logger *zap.Logger) *Qu
 		qOpts.AdditionalHeaders = headers
 	}
 	return qOpts
+}
+
+// BuildQueryServiceOptions creates a QueryServiceOptions struct with appropriate adjusters and archive config
+func (qOpts *QueryOptions) BuildQueryServiceOptions(storageFactory storage.Factory, logger *zap.Logger) *querysvc.QueryServiceOptions {
+	opts := &querysvc.QueryServiceOptions{}
+	if !opts.InitArchiveStorage(storageFactory, logger) {
+		logger.Info("Archive storage not initialized")
+	}
+
+	opts.Adjuster = adjuster.Sequence([]adjuster.Adjuster{
+		adjuster.SpanIDDeduper(),
+		adjuster.ClockSkew(qOpts.MaxClockSkewAdjust),
+		adjuster.IPTagAdjuster(),
+		adjuster.SortLogFields(),
+		adjuster.SpanReferences(),
+	}...)
+
+	return opts
 }
 
 // stringSliceAsHeader parses a slice of strings and returns a http.Header.
--- a/cmd/query/app/flags.go
+++ b/cmd/query/app/flags.go
@@ -23,6 +23,7 @@ import (
 	"net/http"
 	"net/textproto"
 	"strings"
+	"time"
 
 	"github.com/spf13/viper"
 	"go.uber.org/zap"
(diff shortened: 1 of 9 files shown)


cmd/query/app/flags.go
```diff
@@ -64,6 +68,7 @@ func AddFlags(flagSet *flag.FlagSet) {
 	flagSet.String(queryStaticFiles, "", "The directory path override for the static assets for the UI")
 	flagSet.String(queryUIConfig, "", "The path to the UI configuration file in JSON format")
 	flagSet.Bool(queryTokenPropagation, false, "Allow propagation of bearer token to be used by storage plugins")
+	flagSet.Duration(queryMaxClockSkewAdjust, time.Second, "The maximum duration by which a span will be adjusted in the UI")
```
yurishkuro:
```suggestion
	flagSet.Duration(queryMaxClockSkewAdjust, time.Second, "The maximum delta by which span timestamps may be adjusted in the UI due to clock skew; set to 0s to disable clock skew adjustments")
```

model/adjuster/clockskew.go
```diff
@@ -52,12 +53,14 @@ func ClockSkew() Adjuster {
 const (
 	warningDuplicateSpanID       = "duplicate span IDs; skipping clock skew adjustment"
 	warningFormatInvalidParentID = "invalid parent span IDs=%s; skipping clock skew adjustment"
+	warningMaxSkewAdjustExceeded = "max skew adjust of %v exceeded.  not applying calculated adjustment of %v"
```
yurishkuro:
```suggestion
	warningMaxSkewAdjustExceeded = "max clock skew adjustment delta of %v exceeded; not applying calculated delta of %v"
```

cmd/query/app/util.go
```diff
@@ -14,7 +14,31 @@
...
+	if !opts.InitArchiveStorage(storageFactory, logger) {
+		logger.Info("Archive storage not initialized")
+	}
+
+	opts.Adjuster = adjuster.Sequence([]adjuster.Adjuster{
+		adjuster.SpanIDDeduper(),
+		adjuster.ClockSkew(queryOptions.MaxClockSkewAdjust),
+		adjuster.IPTagAdjuster(),
+		adjuster.SortLogFields(),
+		adjuster.SpanReferences(),
```
yurishkuro: I would rather change `StandardAdjusters` to a function and pass the max delta to it. Otherwise StandardAdjusters becomes an empty promise.

cmd/query/app/util.go
```diff
@@ -14,7 +14,31 @@
...
+import (
+	"github.com/jaegertracing/jaeger/cmd/query/app/querysvc"
+	"github.com/jaegertracing/jaeger/model/adjuster"
+	"github.com/jaegertracing/jaeger/storage"
+	"github.com/jaegertracing/jaeger/storage/spanstore"
+	"go.uber.org/zap"
+)
+
+// BuildQueryServiceOptions creates a QueryServiceOptions struct with appropriate adjusters and archive config
+func BuildQueryServiceOptions(storageFactory storage.Factory, queryOptions *QueryOptions, logger *zap.Logger) *querysvc.QueryServiceOptions {
```
yurishkuro: unless there are circular package import issues, I would define this function on the `*QueryOptions` receiver.

model/adjuster/clockskew.go
```diff
@@ -52,12 +53,14 @@ func ClockSkew() Adjuster {
...
 )
 
 type clockSkewAdjuster struct {
-	trace *model.Trace
-	spans map[model.SpanID]*node
-	roots map[model.SpanID]*node
+	trace         *model.Trace
+	spans         map[model.SpanID]*node
+	roots         map[model.SpanID]*node
+	maxSkewAdjust time.Duration
```
yurishkuro:
```suggestion
	maxDelta time.Duration
```
(we already know it's for skew adjustment from the type name)

model/adjuster/clockskew.go
```diff
@@ -177,10 +180,23 @@ func (a *clockSkewAdjuster) adjustTimestamps(n *node, skew clockSkew) {
 		return
 	}
 
+	if absDuration(skew.delta) > a.maxSkewAdjust {
```
yurishkuro:
```suggestion
        if a.maxSkewAdjust == 0 {
		n.span.Warnings = append(n.span.Warnings, fmt.Sprintf(warningSkewAdjustDisabled, skew.delta))
		return
        }
	if absDuration(skew.delta) > a.maxSkewAdjust {
```
and at the top
```
        warningSkewAdjustDisabled := "clock skew adjustment disabled; not applying calculated delta of %v"
```

cmd/query/app/flags.go
```diff
@@ -23,21 +23,26 @@ import (
...
-	queryUIConfig          = "query.ui-config"
-	queryTokenPropagation  = "query.bearer-token-propagation"
-	queryAdditionalHeaders = "query.additional-headers"
+	queryPort               = "query.port"
+	queryBasePath           = "query.base-path"
+	queryStaticFiles        = "query.static-files"
+	queryUIConfig           = "query.ui-config"
+	queryTokenPropagation   = "query.bearer-token-propagation"
+	queryAdditionalHeaders  = "query.additional-headers"
+	queryMaxClockSkewAdjust = "query.max-clock-skew-adjust"
```
yurishkuro: nit: "adjust" is not a noun
```suggestion
	queryMaxClockSkewAdjust = "query.max-clock-skew-adjustment"
```

yurishkuro:
Can we get code coverage up?

joe-elliott:
Test pushed that should raise coverage.  It's a bit trivial as it just asserts nil and non-nil fields on the QueryServiceOptions struct after building.

PR #2132, opened 2020-03-18
Return appropriate gRPC errors/codes to indicate request status

yurishkuro:
- When using HTTP query API, we get a clear 404 when trace is not found. But gRPC API always returns Unknown gRPC code.

- Return appropriate gRPC errors and codes.

### DIFF
--- a/cmd/query/app/grpc_handler.go
+++ b/cmd/query/app/grpc_handler.go
@@ -19,6 +19,8 @@ import (
 
 	"github.com/opentracing/opentracing-go"
 	"go.uber.org/zap"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 
 	"github.com/jaegertracing/jaeger/cmd/query/app/querysvc"
 	"github.com/jaegertracing/jaeger/model"
@@ -26,9 +28,13 @@ import (
 	"github.com/jaegertracing/jaeger/storage/spanstore"
 )
 
-const maxSpanCountInChunk = 10
+const (
+	maxSpanCountInChunk = 10
 
-// GRPCHandler implements the GRPC endpoint of the query service.
+	msgTraceNotFound = "trace not found"
+)
+
+// GRPCHandler implements the gRPC endpoint of the query service.
 type GRPCHandler struct {
 	queryService *querysvc.QueryService
 	logger       *zap.Logger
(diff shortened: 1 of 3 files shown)


cmd/query/app/grpc_handler.go
```diff
@@ -46,36 +52,36 @@ func NewGRPCHandler(queryService *querysvc.QueryService, logger *zap.Logger, tra
...
 	if err == spanstore.ErrTraceNotFound {
-		g.logger.Error("trace not found", zap.Error(err))
-		return err
+		g.logger.Error(msgTraceNotFound, zap.Error(err))
+		return status.Errorf(codes.NotFound, "%s: %v", msgTraceNotFound, err)
 	}
 	if err != nil {
-		g.logger.Error("Could not fetch spans from backend", zap.Error(err))
-		return err
+		g.logger.Error("failed to fetch spans from the backend", zap.Error(err))
```
vprithvi: Logs may be capitalized and with periods while error strings need not (because they are pulled into stack traces)

cmd/query/app/grpc_handler.go
```diff
@@ -46,36 +52,36 @@ func NewGRPCHandler(queryService *querysvc.QueryService, logger *zap.Logger, tra
...
+		return status.Errorf(codes.Internal, "failed to fetch spans from the backend: %v", err)
 	}
 	return g.sendSpanChunks(trace.Spans, stream.Send)
 }
 
-// ArchiveTrace is the GRPC handler to archive traces.
+// ArchiveTrace is the gRPC handler to archive traces.
```
vprithvi: Is everything consistently using this?

cmd/query/app/grpc_handler.go
yurishkuro: I know, but I felt it's simpler to keep strings consistent between errors and logs

cmd/query/app/grpc_handler.go
yurishkuro: Fixing as I see them. Not important, but less complaints from IDE.

go.mod
```diff
@@ -84,7 +84,7 @@ require (
 	golang.org/x/lint v0.0.0-20200130185559-910be7a94367
 	golang.org/x/net v0.0.0-20200202094626-16171245cfb2
 	golang.org/x/sys v0.0.0-20200217220822-9197077df867
-	golang.org/x/tools v0.0.0-20200227222343-706bc42d1f0d
+	golang.org/x/tools v0.0.0-20200227222343-706bc42d1f0d // indirect
```
yurishkuro: I just ran `go mod tidy`

PR #2137, opened 2020-03-24
WIP: Add unit test for examples/hotrod/pkg/pool/pool.go

Hellcatlk:
Signed-off-by: ZouYu <zouy.fnst@cn.fujitsu.com>

- 
Unit Test
- 
add unit test for examples/hotrod/pkg/pool/pool.go

### DIFF
--- a/examples/hotrod/pkg/pool/pool_test.go
+++ b/examples/hotrod/pkg/pool/pool_test.go
@@ -1,0 +1,115 @@
+// Copyright (c) 2020 The Jaeger Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package pool
+
+import (
+	"sync"
+	"sync/atomic"
+	"testing"
+)
+
+var (
+	n       int32 = 0
+	wg      sync.WaitGroup
+	rwMutex sync.RWMutex
+)
+
+func testJob() {
+	rwMutex.RLock()
+	defer rwMutex.RUnlock()
+	atomic.AddInt32(&n, 1)
+	wg.Done()
+}
+
+func TestExecute(t *testing.T) {
+	tests := []struct {
+		name        string
+		numOfWorker int
+		numOfJob    int
+		expected    int32
+	}{
+		//num of worker == num of job
+		{
+			name:        "worker: 1, job: 1",
+			numOfWorker: 1,
+			numOfJob:    1,
+			expected:    1,
+		},
+		{
+			name:        "worker: 2, job: 2",
+			numOfWorker: 2,
+			numOfJob:    2,
+			expected:    2,
+		},
+		//num of worker < num of job
+		{
+			name:        "worker: 1, job: 3",
+			numOfWorker: 1,
+			numOfJob:    3,
+			expected:    3,
+		},
+		{
+			name:        "worker: 2, job: 4",
+			numOfWorker: 2,
+			numOfJob:    4,
+			expected:    4,
+		},
+		//num of worker > num of job
+		{
+			name:        "worker: 5, job: 4",
+			numOfWorker: 5,
+			numOfJob:    4,
+			expected:    4,
+		},
+		{
+			name:        "worker: 6, job: 2",
+			numOfWorker: 6,
+			numOfJob:    2,
+			expected:    2,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			n = 0
+			blockedJob := 0
+			testPool := New(tt.numOfWorker)
+			defer testPool.Stop()
+
+			for i := 0; i < tt.numOfJob; i++ {
+				if blockedJob == 0 {
+					rwMutex.Lock()
+				}
+				testPool.Execute(testJob)
+				blockedJob++
+				if blockedJob == tt.numOfWorker {
+					wg.Add(blockedJob)
+					rwMutex.Unlock()
+					wg.Wait()
+					blockedJob = 0
+				}
+			}
+
+			if blockedJob != 0 {
+				wg.Add(blockedJob)
+				rwMutex.Unlock()
+				wg.Wait()
+			}
+
+			if n != tt.expected {
+				t.Errorf("expect: %v, got: %v\n", tt.expected, n)
+			}
+		})
+	}
+}


examples/hotrod/pkg/pool/pool_test.go
```diff
@@ -0,0 +1,77 @@
...
+import (
+	"testing"
+	"time"
+)
+
+var n int = 0
+
+func TestExecute(t *testing.T) {
+	testPool := New(1)
+	tests := []struct {
```
yurishkuro: there is no value is making the test data-driven if there is only one test case involved

examples/hotrod/pkg/pool/pool_test.go
```diff
@@ -0,0 +1,77 @@
+// Copyright (c) 2019 The Jaeger Authors.
+// Copyright (c) 2017 Uber Technologies, Inc.
```
yurishkuro:
```suggestion
// Copyright (c) 2020 The Jaeger Authors.
```

examples/hotrod/pkg/pool/pool_test.go
```diff
@@ -0,0 +1,77 @@
...
+			job: func() {
+				n = 1
+			},
+			expect: 1,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			n = 0
+			testPool.Execute(tt.job)
```
yurishkuro: there is a race condition, no guarantee that the job has been executed

examples/hotrod/pkg/pool/pool_test.go
```diff
@@ -0,0 +1,77 @@
...
+				n = 2
+			},
+			unexpect: 2,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			n = 0
+			testPool.Execute(tt.job)
+			testPool.Stop()
```
yurishkuro: race condition

examples/hotrod/pkg/pool/pool_test.go
Hellcatlk: thank's for your reply, i have update the code.

examples/hotrod/pkg/pool/pool_test.go
```diff
@@ -0,0 +1,100 @@
...
+			name:   "2 worker, loop 6",
+			pool:   New(2),
+			job:    testJob,
+			used:   time.Second*3 + 500000000,
+			expect: 6,
+		},
+	}
+	for index, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			wg.Add(index + 1)
```
yurishkuro: This looks like incorrect usage of wg.Add - since you're only adding one job, it should be Add(1).

examples/hotrod/pkg/pool/pool_test.go
```diff
@@ -0,0 +1,100 @@
...
+	"testing"
+	"time"
+)
+
+var n int32 = 0
+var wg sync.WaitGroup
+
+func testJob() {
+	atomic.AddInt32(&n, 1)
+	time.Sleep(time.Second)
```
yurishkuro: why do you need sleep at all, if you're using waitGroup to signal completion?

Hellcatlk: Because i rely on runtime to determine if it is running correctly, sleeping for 1 second makes it easier to calculate the running time.

examples/hotrod/pkg/pool/pool_test.go
Hellcatlk: the job will be executed (index + 1) times

examples/hotrod/pkg/pool/pool_test.go
yurishkuro: I don't follow. WaitGroup allows you to know that a job was executed, using sleep in addition to that does not add any extra benefits. Making tests relying on the timestamps always makes them unstable in CI where scheduling and CPU allocations are unpredictable.

Hellcatlk: OK, I will remove judging the number of workers by runtime

examples/hotrod/pkg/pool/pool_test.go
yurishkuro: This is difficult to reason about, and makes the whole test fragile by creating dependencies on the order of tests. If you need N job runs per test, then you can make N an explicit parameter of the test (a field in the struct).

Also, if you're running jobs in a loop, it's still easier to reason about them if you call wg.Add(1) before each job, rather than wg.Add(N).

examples/hotrod/pkg/pool/pool_test.go
```diff
@@ -0,0 +1,89 @@
...
+			expected: 1,
+		},
+		{
+			name:     "worker: 2, job: 2",
+			pool:     New(2),
+			job:      testJob,
+			expected: 2,
+		},
+		{
+			name:     "worker: 3, job: 3",
```
yurishkuro: what new conditions are being tested with counts 3 and 4 that were not already tested with 1 and 2?

examples/hotrod/pkg/pool/pool_test.go
```diff
@@ -0,0 +1,89 @@
...
+	for index, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			wg.Add(index + 1)
+			n = 0
+			for i := 0; index >= i; i++ {
+				tt.pool.Execute(tt.job)
+			}
+			wg.Wait()
+			tt.pool.Stop()
+			if n != tt.expected{
```
yurishkuro: Note that this is not really testing the concurrent of the executor pool. For example, if you have a pool with 2 goroutines, and you run two jobs, this test cannot verify that both jobs did not run sequentially in a single goroutine rather than being run concurrently.

Hellcatlk: I try to use a  rwlock instead of sleep, but when the number of jobs is greater than the number of workers, the program will be blocked.

examples/hotrod/pkg/pool/pool_test.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			n = 0
+			used := tt.numOfWorker
+			testPool := New(tt.numOfWorker)
+			defer testPool.Stop()
+
+			for i := 0; tt.numOfJob > i; i++ {
+				if tt.numOfWorker == used {
```
yurishkuro: I don't follow this. The waitgroup is a mechanism to notify the test driver that the jobs have been completed, correct? As such, why would the use of WG be conditional on loop iterations?

Hellcatlk: In my test, when rwMutex.Unlock() is completed, rwMutex.Lock() will be executed immediately. So i use wg to let rwMutex.Lock() delay.

yurishkuro: There are tests where numOfWorker < numOfJob. For those tests `used` will become negative - what does that mean? Can you try to simplify the logic?

examples/hotrod/pkg/pool/pool_test.go
```diff
@@ -0,0 +1,92 @@
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			n = 0
+			used := tt.numOfWorker
+			testPool := New(tt.numOfWorker)
+			defer testPool.Stop()
+
+			for i := 0; tt.numOfJob > i; i++ {
```
yurishkuro:
```suggestion
			for i := 0; i < tt.numOfJob; i++ {
```
this is the traditional way if writing loops

PR #2139, opened 2020-03-24
Add Cassandra OTEL exporter

pavolloffay:
Depdens on #2135


cmd/opentelemetry-collector/app/exporter/cassandra/factory.go
```diff
@@ -0,0 +1,75 @@
...
+type OptionsFactory func() *cassandra.Options
+
+// DefaultOptions creates Cassandra options supported by this exporter.
+func DefaultOptions() *cassandra.Options {
+	return cassandra.NewOptions("cassandra")
+}
+
+// Factory is the factory for Jaeger Cassandra exporter.
+type Factory struct {
+	OptionsFactory OptionsFactory
```
yurishkuro: in what scenarios does this need to be a function instead of a struct?

cmd/opentelemetry-collector/app/exporter/cassandra/factory.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// Type gets the type of exporter.
+func (Factory) Type() string {
+	return TypeStr
+}
+
+// CreateDefaultConfig returns default configuration of Factory.
+func (f Factory) CreateDefaultConfig() configmodels.Exporter {
```
yurishkuro: Is the name CreateDefaultConfig required by some external API?  We usually mention it in the comment.

Given that the result of this function depends on externally provided (via OptionsFactory) options, the name "default" is confusing.

annanay25: `CreateDefaultConfig()` is part of the `extension.Factory` interface in the Otel collector. I agree that "default" is confusing here, we should ideally apply user supplied options once the default config is created

cmd/opentelemetry-collector/app/exporter/cassandra/config.go
```diff
@@ -0,0 +1,27 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cassandra
+
+import (
+	"github.com/open-telemetry/opentelemetry-collector/config/configmodels"
+
+	"github.com/jaegertracing/jaeger/plugin/storage/cassandra"
```
annanay25: Also we should consider naming this import `storageCassandra`. Less confusing and also consistent with other files where it has been imported

cmd/opentelemetry-collector/app/exporter/cassandra/factory.go
pavolloffay: We need to initialize the `Options` struct with wiper once the collector service has started.

We can revisit this and instead pass a viper instance and run the initialization in `CreateDefaultConfgi`.

cmd/opentelemetry-collector/app/exporter/cassandra/factory.go
pavolloffay: >  I agree that "default" is confusing here, we should ideally apply user supplied options once the default config is created

OTEL service applies config file changes to the returned instance. We cannot override them by config supplied as legacy jaeger flags. The OTEL config takes precedence over legacy jaeger configuration.

cmd/opentelemetry-collector/app/defaults/defaults_test.go
```diff
@@ -19,16 +19,22 @@ import (
...
 	assert.Equal(t, "jaeger_kafka", factories.Exporters[kafka.TypeStr].Type())
+	assert.Equal(t, "jaeger_cassandra", factories.Exporters[cassandra.TypeStr].Type())
 
 	kafkaFactory := factories.Exporters[kafka.TypeStr]
 	kc := kafkaFactory.CreateDefaultConfig().(*kafka.Config)
 	assert.Equal(t, []string{"127.0.0.1:9092"}, kc.Config.Brokers)
+
+	cassandraFactory := factories.Exporters[cassandra.TypeStr]
+	cc := cassandraFactory.CreateDefaultConfig().(*cassandra.Config)
+	assert.Equal(t, []string{"127.0.0.1"}, cc.Options.GetPrimary().Servers)
```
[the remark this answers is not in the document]
pavolloffay: Double-check that the default values have been applied. I will consider changing `defaults` interface to include cobra command and install Jaeger flags by default.

cmd/opentelemetry-collector/app/exporter/cassandra/config.go
pavolloffay: It is only in one file, where there is a conflict in imports.

pavolloffay:
@yurishkuro I will merge this, feel free to comment and I will fix issues in follow up PR.

PR #2144, opened 2020-03-30
Add support for ES index aliases / rollover to the dependency store (Resolves #2143)

frittentheke:
## Which problem is this PR solving?
Currently the dependency store has no support for index aliases / rollover indices.
Resolves https://github.com/jaegertracing/jaeger/issues/2143#

## Short description of the changes
Use the already existing and used config variable `UseReadWriteAliases` to switch between the current behavior and using non-dated "-read" and "-write" index names. This aligns with the behavior that is already in place for spans and services.

PR #2149, opened 2020-04-02
Move ES dependencies index mapping to JSON template file

frittentheke:
This is a follow-up to PR https://github.com/jaegertracing/jaeger/pull/2144

This PR removes the inline ES index template applied to every newly create dependencies index and replaces it with a template matching dependencies indices by name. Similar to PR https://github.com/jaegertracing/jaeger/pull/1309 which did the same for spans / services


plugin/storage/es/dependencystore/storage.go
```diff
@@ -31,46 +32,61 @@ import (
...
+	useReadWriteAliases   bool
+}
+
+// DependencyStoreParams holds constructor parameters for NewDependencyStore
+type DependencyStoreParams struct {
+	Client              es.Client
+	Logger              *zap.Logger
+	MetricsFactory      metrics.Factory
+	IndexPrefix         string
+	UseReadWriteAliases bool
```
pavolloffay, line 55: Could you please remove this and keep only mappings templates handling in this PR?

frittentheke: I pushed a fresh commit only containing the removal of the indices

plugin/storage/es/dependencystore/storage.go
```diff
@@ -60,17 +60,15 @@ func NewDependencyStore(client es.Client, logger *zap.Logger, indexPrefix string
 // WriteDependencies implements dependencystore.Writer#WriteDependencies.
 func (s *DependencyStore) WriteDependencies(ts time.Time, dependencies []model.DependencyLink) error {
 	indexName := indexWithDate(s.indexPrefix, ts)
-	if err := s.createIndex(indexName); err != nil {
```
pavolloffay, line 63: IIRC this should stay here. The index has to be created

frittentheke: @pavolloffay  I don't think so - you can simply "write" to the index and it will be created implicitly - check out the SpanWriter / ServiceWriter. The sole reason this was here was to create the index with explicit mappings. But with this PR we have a template that is applied to a new, implicitly created index automagically.

pavolloffay: You are right sorry my bad.

frittentheke: No worries - thanks for looking into my changes to thoroughly.

frittentheke:
> The PR looks much cleaner now +1 . Could you please fix the CI? There seem to be some failures
> https://travis-ci.org/github/jaegertracing/jaeger/jobs/670675957#L786

@pavolloffay  I did fix / adjust the tests and also ran the make job to update the statically embedded files (`make elasticsearch-mappings`) - let's see if we get a green light now.

frittentheke:
@pavolloffay the two currently still failing CI builds seems to be out of "my" control as they are related to other components. I already did move up to the current master, no change though.

pavolloffay:
I have restarted the CI

frittentheke:
@pavolloffay  Maybe you could help me a little with the still failing tests?


1) I honestly don't understand the purpose of this test here:
https://travis-ci.org/github/jaegertracing/jaeger/jobs/672665074#L3474
Why is there an error expected to occur?
With this PR the index creation does not happen explicitly anymore, so could we just remove this test?

2) What is this test_driver here doing and why is it timing out on startup?
https://travis-ci.org/github/jaegertracing/jaeger/jobs/672665077#L3877

3) Last but not least in this case there seems to be no Elasticsearch running to apply the template "jaeger-dependencies" to:  https://travis-ci.org/github/jaegertracing/jaeger/jobs/672665079#L3600

But how could that relate to my changes?

pavolloffay:
All the tests have to be fixed before merging. If the logic has changed then also adapt the tests for the new logic.

Sometimes CI is flaky, the best is to run the test locally and see whether it passes or not.

frittentheke:
@pavolloffay maybe that was lost in translation, but I was not arguing to fix the CI, I gladly dive into things more and prove additional commits to get this PR ready to merge.

Just because you wrote so much of that code (and the tests and CI to it) I was hoping you could give me a few hints on why things might be broken.

pavolloffay:
> Just because you wrote so much of that code (and the tests and CI to it) I was hoping you could give me a few hints on why things might be broken.

Point me to the code where you have questions. Generally speaking, we have around 100% test coverage in Jaeger so test really examine each branch.

frittentheke:
> > Just because you wrote so much of that code (and the tests and CI to it) I was hoping you could give me a few hints on why things might be broken.
> 
> Point me to the code where you have questions. Generally speaking, we have around 100% test coverage in Jaeger so test really examine each branch.


@pavolloffay thanks for the offer to help. I actually did ask questions in my previous https://github.com/jaegertracing/jaeger/pull/2149#issuecomment-611187001 . One question (1) was actually about the validity of one test in general (asking if I may remove it altogether).  The other two questions where about the CI setup in general. And quite honestly there are a whole lot of scripts interworking with each other to setup a test environment for the CI to work against. The complexity there (and even the flakiness you mentioned yourself) create somewhat of a burden for new contributors to understand before they can get to debug their actual code / tests.

frittentheke:
@pavolloffay I am right to assume that I should base my PR commits on the last release tag (release-1.17.1) instead of master itself?

Because master seems to be quite active and there are so mange unrelated issues with various tests there when I apply my innocent three commits onto there.

pavolloffay:
The PR should be based on the latest master. There is already one git conflict in this PR. About the tests as I mentioned previously, the test should be adapted to code changes. I often run ES integration tests locally. It shouldn't be too hard.

frittentheke:
@pavolloffay allright things are moving forward - merge should be clean and with commit https://github.com/jaegertracing/jaeger/pull/2149/commits/820921d20ed9222181e0db3ea023138bbd438487 I adjusted the WriteDependencies test according to the changes of this PR -> all green now.

What remains are the other two CI jobs - one about ElasticSearch integration (https://travis-ci.org/github/jaegertracing/jaeger/jobs/674885438#L3648). While that one seems to be related to my changes as it affects ES - there is no real indication / output on why it fails and why this service did not properly start up.

And then there is the crossdock docker compose thing not properly starting up (https://travis-ci.org/github/jaegertracing/jaeger/jobs/674885436#L3734) - this is where your CI things actually are a little more complex to understand ... there is no error whatsoever until that service apparently throws a 503: https://travis-ci.org/github/jaegertracing/jaeger/jobs/674885436#L3734

Do you have any more hints on getting things replicated locally.  As per contribution guidelines (https://github.com/jaegertracing/jaeger/blob/master/CONTRIBUTING.md#pre-requisites) I was able to run the tests locally via `make test`, but apparently the CI goes much further and starts of quite a few things to run integration tests. Apart from reading CI scripts there seems to be no easy way for a newbie to start this up on one's local box.

pavolloffay:
The xdock e2e tests are harder to understand and debug, but they are using ES so assuming I have restarted the job a couple of times it seems to be related to this PR. Let's fix the ES integration and that should also fix the xdock.

The ES storage integration test can be run by 
```
docker run -it --rm -e "ES_JAVA_OPTS=-Xms2g -Xmx2g" -p 9200:9200 -p 9300:9300 -e "http.host=0.0.0.0" -e "discovery.type=single-node" --name=elasticsearch docker.elastic.co/elasticsearch/elasticsearch-oss:6.8.4
STORAGE=elasticsearch make storage-integration-test  
```

Similarly you can run other tests from this script https://github.com/jaegertracing/jaeger/blob/47d20296982d2e01432be46daef1dfc2c365872b/scripts/travis/es-integration-test.sh#L9

frittentheke:
> The xdock e2e tests are harder to understand and debug, but they are using ES so assuming I have restarted the job a couple of times it seems to be related to this PR. Let's fix the ES integration and that should also fix the xdock.
> 
> The ES storage integration test can be run by
> 
> ```
> docker run -it --rm -e "ES_JAVA_OPTS=-Xms2g -Xmx2g" -p 9200:9200 -p 9300:9300 -e "http.host=0.0.0.0" -e "discovery.type=single-node" --name=elasticsearch docker.elastic.co/elasticsearch/elasticsearch-oss:6.8.4
> STORAGE=elasticsearch make storage-integration-test  
> ```
> 
> Similarly you can run other tests from this script
> 
> https://github.com/jaegertracing/jaeger/blob/47d20296982d2e01432be46daef1dfc2c365872b/scripts/travis/es-integration-test.sh#L9


Thanks for the hint @pavolloffay .


Well the core ElasticSearch tests are all green (as they were in Travis-CI), but the issue starts with 

`
bash -c "set -e; set -o pipefail; go test -v -race -tags token_propagation -run TestBearTokenPropagation ./plugin/storage/integration/... | sed ''/PASS/s//PASS/'' | sed ''/FAIL/s//FAIL/''"
`

which  is the called at https://github.com/jaegertracing/jaeger/blob/master/scripts/travis/es-integration-test.sh#L18 .

With script itself is rather silent about itself and there are `make` targets being called to build and run other things fore the TokenPropagation tests ... this honestly is a little complex and hard to debug just by looking at the CI output.

I shall be trying to replicate this locally and adding some debug statements to see where things might fail when this runs in CI.

frittentheke:
@pavolloffay I found (and hopefully) fixed an issue (of course a stupid syntax error in the template file itself) so kudos to CI :-)

When trying to replicate the CI stages locally, I was wondering about the startup of the jaeger-query here: https://github.com/jaegertracing/jaeger/blob/master/scripts/travis/es-integration-test.sh#L25

It misses an Elasticsearch service on port 9200 (the other ones used before were all killed again).
Who or how is that every started?

pavolloffay:
Try it locally, it does not require ES to be up. Then the test use mocked ES instance.

```
SPAN_STORAGE_TYPE=elasticsearch ./cmd/query/query-linux --es.server-urls=http://127.0.0.1:9200 --es.tls=false --es.version=7 --query.bearer-token-propagation=true
```

scripts/travis/es-integration-test.sh
```diff
@@ -22,7 +22,10 @@ make build-crossdock-ui-placeholder
 GOOS=linux make build-query
 
 make test-compile-es-scripts
+CID=$(docker run --rm -d -p 9200:9200 -e "http.host=0.0.0.0" -e "transport.host=127.0.0.1" -e "xpack.security.enabled=false" -e "xpack.monitoring.enabled=false" docker.elastic.co/elasticsearch/elasticsearch:7.3.0)
```
pavolloffay, line 25: Why is this needed?

frittentheke: I was just playing around with things and locally this was required (no ES was running / mocked on port 9200 in order for those test to run against)

Honestly I am a little lost  ... I simply tried the same approach that was done a few lines above in that script in order to make things work.

frittentheke:
> Try it locally, it does not require ES to be up. Then the test use mocked ES instance.
> 
> ```
> SPAN_STORAGE_TYPE=elasticsearch ./cmd/query/query-linux --es.server-urls=http://127.0.0.1:9200 --es.tls=false --es.version=7 --query.bearer-token-propagation=true
> ```

@pavolloffay  host / where is the mock started then? A few lines lines above and as per your instructions there is a single node elasticsearch started as a docker container right before those tests are started.

frittentheke:
@pavolloffay  would you kindly take another look at this. I just pushed commit https://github.com/jaegertracing/jaeger/commit/b6d1c8bcc31c09d2b538270341008b6736efbfc9 containing the discussed missing ES instance to run `TestBearTokenPropagation`.

I believe the ElasticSearch intergration tests are skipped most of the time, which is  why my tests fail, but master work (check out last run: https://travis-ci.org/github/jaegertracing/jaeger/jobs/677609252#L2285)

frittentheke:
This PR has been obsoleted by #2285 - closing this one.

frittentheke:
This PR has been obsoleted by #2285 - closing this one.

PR #2168, opened 2020-04-06
Fix flaky client metrics test

### LINKED ISSUE #2108: Fix flaky test TestClientMetricsReporter_Expire
It seems to be touched in https://github.com/jaegertracing/jaeger/pull/2010

```
--- FAIL: TestClientMetricsReporter_Expire (0.01s)
    --- FAIL: TestClientMetricsReporter_Expire/iter0:gauge=1,log=new_client (0.01s)
        client_metrics_test.go:268: 
            	Error Trace:	client_metrics_test.go:268
            	Error:      	Not equal: 
            	            	expected: int(1)
            	            	actual  : int64(0)
            	Test:       	TestClientMetricsReporter_Expire/iter0:gauge=1,log=new_client
FAIL	github.com/jaegertracing/jaeger/cmd/agent/app/reporter	0.064s
```

pavolloffay:
Resolves #2108 

@yurishkuro could you please review?

It might not be the best fix but it seems to work.

### DIFF
--- a/cmd/agent/app/reporter/client_metrics_test.go
+++ b/cmd/agent/app/reporter/client_metrics_test.go
@@ -263,7 +263,7 @@ func TestClientMetricsReporter_Expire(t *testing.T) {
 					if gaugeValue == int64(test.expGauge) {
 						break
 					}
-					time.Sleep(500 * time.Microsecond)
+					time.Sleep(5 * time.Millisecond)
 				}
 				assert.EqualValues(t, test.expGauge, gaugeValue)
 				tr.assertLog(t, test.expLog, clientUUID)


cmd/agent/app/reporter/client_metrics_test.go
```diff
@@ -263,7 +263,7 @@ func TestClientMetricsReporter_Expire(t *testing.T) {
 					if gaugeValue == int64(test.expGauge) {
 						break
 					}
-					time.Sleep(500 * time.Microsecond)
+					time.Sleep(5 * time.Millisecond)
```
yurishkuro, line 266: Not sure about this fix. The ticket shows failure in the first iteration, meaning the loop did not wait long enough for the new client connection to be detected. But this change is setting the wait time to be equal to the cache TTL, at which point the counter goes back to 0. So this change is probably increasing the likelihood of race condition.

yurishkuro: I would instead try increasing the loop limit from 10 to 20, or wait time from 0.5 to 1 ms.
```suggestion
					time.Sleep(1 * time.Millisecond)
```

yurishkuro:
I implemented the tests differently in #2170

pavolloffay:
Thanks @yurishkuro

PR #2190, opened 2020-04-19
Support elasticsearch index rollover by hour or quarter hour

### LINKED ISSUE #2189: ElasticSearch index gets too large when tracing a large cluster.
ElasticSearch index gets too large when tracing a large cluster. It's more than 1TB at this time and we expect it to grow even more later on. 

A single huge index makes searching pretty slow even when I only need to search the latest spans within an hour. 

Support rollover indices by finer granularity like every hour or even every quarter hour.

sunxiaoguang:
Signed-off-by: Xiaoguang Sun <sunxiaoguang@zhihu.com>

- Resolves #2189

- add new es.rollover-interval and es-archive.rollover-interval flag to specify if new index should be created at daily (default), hourly or even quarterly basis.
- change spanstore to support new rollover intervals
- change esCleanup.py to support new rollover intervals

### DIFF
--- a/pkg/es/config/config.go
+++ b/pkg/es/config/config.go
@@ -38,24 +38,36 @@ import (
 	storageMetrics "github.com/jaegertracing/jaeger/storage/spanstore/metrics"
 )
 
+type RolloverInterval byte
+
+const (
+	// RolloverDaily means creating an new index once a day, this is the default behavior
+	RolloverDaily = iota
+	// RolloverHourly means creating an new index once an hour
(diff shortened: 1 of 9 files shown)


yurishkuro:
Jaeger works with Elastic Rollover, which allows you to set different index rollover policies other than one index per day. Please see https://www.jaegertracing.io/docs/1.17/deployment/#elasticsearch-rollover.

jpkrohling:
Closing due to inactivity.

PR #2191, opened 2020-04-21
Remove agent's hostPort env var from go service in xdock

pavolloffay:


crossdock/docker-compose.yml
```diff
@@ -29,8 +29,6 @@ services:
         image: jaegertracing/xdock-go
         ports:
             - "8080-8082"
-        environment:
-            - AGENT_HOST_PORT=jaeger-agent:6831
```
pavolloffay, line 33: not needed anymore fix has been applied in go service.

PR #2199, opened 2020-04-25
fix StartTimeMin

candiduslynx:
Per https://www.jaegertracing.io/docs/1.17/apis/#service-dependencies-graph-internal lookback is in milliseconds. However, badger plugin fails to account for it.

Fix timing unit.


candiduslynx:
found in code that lookback is actually already converted to duration

Issue #2113, opened 2020-03-03
Tracking and removing deprecated CLI flags


yurishkuro:
We've accumulated quite a few deprecated flags over time. We should start removing them. However, we need to define a backwards compatibility policy. I propose the following:
  * if a flag is deprecated in release N, it can be removed in release N+2
  * when adding a `(deprecated)` prefix to the flags, we should also add an indication of when it will b removed. Given the above rule, we could set a 3 months time interval. So the deprecation message should specify the date, e.g. `(deprecated, will be removed after 2020-03-15 or in release v1.19.0, whichever is later)`.
  * at the top of the file where the flag name is defined, add a constant and a comment, e.g.
```go
// TODO deprecated flag to be removed
healthCheckHTTPPortWarning = "(deprecated, will be removed after 2020-03-15 or in release v1.19.0, whichever is later)"
```
  * use that constant as the prefix to the help text, e.g.
```go
flagSet.Int(healthCheckHTTPPort, 0, healthCheckHTTPPortWarning+" see --"+adminHTTPHostPort)
```
  * when reading the flag into config, log a warning with the same warning
  * take care of deprecated flags in `initFromViper` functions, do not pass them to business functions

yurishkuro:
We can also book tickets for the respective milestone for each deprecated flag, so that we don't forget to remove them in a timely manner.

pavolloffay:
It sounds reasonable. Maybe instead of providing users with a date, it would be better to put there release version.

jpkrohling:
> if a flag is deprecated in release N, it can be removed in release N+2

+1

> when adding a (deprecated) prefix to the flags, we should also add an indication of when it will b removed. Given the above rule, we could set a 3 months time interval. So the deprecation message should specify the date, e.g. (deprecated, will be removed after 2020-03-15 or in release v1.19.0, whichever is later).

I'm afraid this will make the help command too verbose, which impacts the readability for the non-deprecated flags as well. Apart from that, I have nothing against it.

One further action we could be taking is to log a message whenever a deprecated flag is being used, increasing the severity for every release:

1.17.0 - `--some-flag` becomes deprecated, log at `INFO` level
1.18.0 - `--some-flag` is still being used, log at `WARN` level
1.19.0 - `--some-flag` removed from code, log at `ERROR` level
1.20.0 - no more traces of `--some-flag` in the code

yurishkuro:
> Maybe instead of providing users with a date, it would be better to put there release version.

[...]

PR #2212, opened 2020-05-02
Normalize CLI flags to use host:port addresses

### LINKED ISSUE #1447: Normalize all CLI flags for port numbers

ohdearaugustin:
Signed-off-by: Jonas Reindl <j-reindl@live.at>

- Fixes #1447

- Normalize CLI flags to use host:port address format
- Add HostPort Flag to query app as it was left out

### DIFF
--- a/cmd/query/app/flags.go
+++ b/cmd/query/app/flags.go
@@ -38,13 +38,13 @@ import (
 const (
 	queryHostPort           = "query.host-port"
 	queryPort               = "query.port"
+	queryPortWarning        = "(deprecated, will be removed after 2020-08-31 or in release v1.20.0, whichever is later)"
 	queryBasePath           = "query.base-path"
 	queryStaticFiles        = "query.static-files"
 	queryUIConfig           = "query.ui-config"
 	queryTokenPropagation   = "query.bearer-token-propagation"
 	queryAdditionalHeaders  = "query.additional-headers"
 	queryMaxClockSkewAdjust = "query.max-clock-skew-adjustment"
-	queryPortWarning        = "(deprecated, will be removed after 2020-06-30 or in release v1.20.0, whichever is later)"
 )
 
 // QueryOptions holds configuration for query service
--- a/cmd/query/app/flags.go
+++ b/cmd/query/app/flags.go
@@ -51,8 +51,6 @@ const (
 type QueryOptions struct {
 	// HostPort is the host:port address that the query service listens o n
 	HostPort string
-	// Port is the port that the query service listens in on (deprecated, will be removed after 2020-06-30 or in release v1.20.0, whichever is later)
-	Port int
 	// BasePath is the prefix for all UI and API HTTP routes
 	BasePath string
 	// StaticAssets is the path for the static assets for the UI (https://github.com/uber/jaeger-ui)
--- a/cmd/query/app/flags.go
+++ b/cmd/query/app/flags.go
@@ -51,9 +51,9 @@ const (
 type QueryOptions struct {
 	// HostPort is the host:port address that the query service listens o n
 	HostPort string
-	// Port is the port that the query service listens in on ()
+	// Port is the port that the query service listens in on (deprecated, will be removed after 2020-06-30 or in release v1.20.0, whichever is later)
 	Port int
-	// BasePath is the prefix for all UI and API HTTP routes (deprecated, will be removed after 2020-06-30 or in release v1.20.0, whichever is later)
+	// BasePath is the prefix for all UI and API HTTP routes
 	BasePath string
 	// StaticAssets is the path for the static assets for the UI (https://github.com/uber/jaeger-ui)
 	StaticAssets string
@@ -81,7 +81,6 @@ func AddFlags(flagSet *flag.FlagSet) {
 
 // InitFromViper initializes QueryOptions with properties from viper
 func (qOpts *QueryOptions) InitFromViper(v *viper.Viper, logger *zap.Logger) *QueryOptions {
-	// qOpts.Port = v.GetInt(queryPort)
 	qOpts.HostPort = ports.GetAddressFromCLIOptions(v.GetInt(queryPort), v.GetString(queryHostPort))
 	qOpts.BasePath = v.GetString(queryBasePath)
 	qOpts.StaticAssets = v.GetString(queryStaticFiles)
--- a/cmd/query/app/flags.go
+++ b/cmd/query/app/flags.go
@@ -36,6 +36,7 @@ import (
 )
 
 const (
+	queryHostPort           = "query.host-port"
 	queryPort               = "query.port"
 	queryBasePath           = "query.base-path"
 	queryStaticFiles        = "query.static-files"
@@ -43,13 +44,16 @@ const (
 	queryTokenPropagation   = "query.bearer-token-propagation"
 	queryAdditionalHeaders  = "query.additional-headers"
 	queryMaxClockSkewAdjust = "query.max-clock-skew-adjustment"
+	queryPortWarning        = "(deprecated, will be removed after 2020-06-30 or in release v1.20.0, whichever is later)"
 )
 
 // QueryOptions holds configuration for query service
 type QueryOptions struct {
-	// Port is the port that the query service listens in on
+	// HostPort is the host:port address that the query service listens o n
+	HostPort string
+	// Port is the port that the query service listens in on ()
 	Port int
-	// BasePath is the prefix for all UI and API HTTP routes
+	// BasePath is the prefix for all UI and API HTTP routes (deprecated, will be removed after 2020-06-30 or in release v1.20.0, whichever is later)
 	BasePath string
 	// StaticAssets is the path for the static assets for the UI (https://github.com/uber/jaeger-ui)
 	StaticAssets string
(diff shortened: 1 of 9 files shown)


cmd/query/app/flags.go
```diff
@@ -93,6 +99,19 @@ func (qOpts *QueryOptions) InitFromViper(v *viper.Viper, logger *zap.Logger) *Qu
 	return qOpts
 }
 
+// Utility function to get listening address based on port (deprecated flags) or host:port (new flags)
+func getAddressFromCLIOptions(port int, hostPort string) string {
```
yurishkuro, line 103: This function already exists in `./cmd/collector/app/builder_flags.go`, please move it to `./ports/ports.go` as a shared util function `HostPortFromPortOrString(port int, hostPort string)`

cmd/query/app/flags.go
```diff
@@ -76,7 +81,8 @@ func AddFlags(flagSet *flag.FlagSet) {
 
 // InitFromViper initializes QueryOptions with properties from viper
 func (qOpts *QueryOptions) InitFromViper(v *viper.Viper, logger *zap.Logger) *QueryOptions {
-	qOpts.Port = v.GetInt(queryPort)
+	// qOpts.Port = v.GetInt(queryPort)
```
yurishkuro, line 84: remove?

cmd/query/app/flags.go
```diff
@@ -36,20 +36,24 @@ import (
...
+	queryPortWarning        = "(deprecated, will be removed after 2020-06-30 or in release v1.20.0, whichever is later)"
 )
 
 // QueryOptions holds configuration for query service
 type QueryOptions struct {
-	// Port is the port that the query service listens in on
+	// HostPort is the host:port address that the query service listens o n
+	HostPort string
+	// Port is the port that the query service listens in on ()
 	Port int
```
yurishkuro, line 55: I think this field should be removed, only the legacy CLI option should remain

cmd/query/app/flags.go
```diff
@@ -36,20 +36,24 @@ import (
...
-	// BasePath is the prefix for all UI and API HTTP routes
+	// BasePath is the prefix for all UI and API HTTP routes (deprecated, will be removed after 2020-06-30 or in release v1.20.0, whichever is later)
```
yurishkuro, line 56: why is BasePath deprecated?

cmd/query/app/flags_test.go
```diff
@@ -28,7 +28,7 @@ import (
 	spanstore_mocks "github.com/jaegertracing/jaeger/storage/spanstore/mocks"
 )
 
-func TestQueryBuilderFlags(t *testing.T) {
+func TestQueryBuilderFlagsDeprecation(t *testing.T) {
```
yurishkuro, line 31: since this test is call "test deprecation", it should only include deprecated flags. All other flags are already tested in the following test, there's no reason to repeat then here.

cmd/query/app/flags_test.go
```diff
@@ -60,6 +84,28 @@ func TestQueryBuilderBadHeadersFlags(t *testing.T) {
 	assert.Nil(t, qOpts.AdditionalHeaders)
 }
 
+func TestQueryOptionsWithFlags_CheckHostPort(t *testing.T) {
```
yurishkuro, line 87: These tests for different syntax of host-port string would better to into `./ports/ports_test.go`. This file tests interaction with CLI flags.

cmd/query/app/server.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2019 The Jaeger Authors.
+// Copyright (c) 2020 The Jaeger Authors.
```
yurishkuro, line 1: It's not necessary to change the copyright year, but if you do, the correct form would be `2019,2020`

cmd/query/app/server.go
```diff
@@ -105,10 +104,11 @@ func (s *Server) Start() error {
 	if port, err := netutils.GetPort(s.conn.Addr()); err == nil {
 		tcpPort = port
 	}
+	s.svc.Logger.Info("addr", zap.Int("port", tcpPort))
 
 	s.svc.Logger.Info(
 		"Query server started",
-		zap.Int("port", tcpPort))
+		zap.String("addr", s.queryOptions.HostPort))
```
yurishkuro, line 111: keep the port field as well, because starting with `:0` is valid, and the `tcpPort` var gives the actual port number.

same for the next 3 logs

cmd/query/app/server_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2019 The Jaeger Authors.
+// Copyright (c) 2020 The Jaeger Authors.
```
yurishkuro, line 1: 2019,2020

cmd/query/app/ui/placeholder/gen_assets.go
```diff
@@ -213,7 +213,7 @@ var _escData = map[string]*_escFile{
 		name:    "index.html",
 		local:   "cmd/query/app/ui/placeholder/public/index.html",
 		size:    376,
-		modtime: 1547655513,
+		modtime: 1586972220,
```
yurishkuro, line 216: revert this, it's not changing in this PR.

`git checkout master cmd/query/app/ui/placeholder/gen_assets.go`

cmd/query/app/flags.go
ohdearaugustin: This field is necessary so the legacy CLI option is still functional. I will add deprecation notice to it.

cmd/query/app/flags.go
ohdearaugustin: To be honest that was a total brainfart from my side.^^

cmd/query/app/flags.go
ohdearaugustin: refactor as suggested

cmd/query/app/flags.go
ohdearaugustin: removed

cmd/query/app/flags_test.go
ohdearaugustin: refactored as suggested

cmd/query/app/server.go
ohdearaugustin: fixed

cmd/query/app/server.go
ohdearaugustin: refactored as suggested

cmd/query/app/server_test.go
ohdearaugustin: fixxed

cmd/query/app/ui/placeholder/gen_assets.go
ohdearaugustin: reverted

cmd/query/app/flags.go
yurishkuro: Why is it necessary? It is no longer assigned in InitFromViper

cmd/query/app/flags.go
```diff
@@ -36,18 +36,22 @@ import (
...
 const (
+	queryHostPort           = "query.host-port"
 	queryPort               = "query.port"
 	queryBasePath           = "query.base-path"
 	queryStaticFiles        = "query.static-files"
 	queryUIConfig           = "query.ui-config"
 	queryTokenPropagation   = "query.bearer-token-propagation"
 	queryAdditionalHeaders  = "query.additional-headers"
 	queryMaxClockSkewAdjust = "query.max-clock-skew-adjustment"
+	queryPortWarning        = "(deprecated, will be removed after 2020-06-30 or in release v1.20.0, whichever is later)"
```
yurishkuro, line 47: nit: I would place it next to queryPort const

yurishkuro: per #2113, we want to set deprecation to now+3m

cmd/query/app/flags_test.go
yurishkuro: thx

cmd/query/app/flags.go
ohdearaugustin: I will change that. So deprecated at 2020-08-31?

yurishkuro: yes

cmd/query/app/flags_test.go
ohdearaugustin: moved to ./ports/ports_test.go

cmd/query/app/flags.go
ohdearaugustin: I will look into it a bit more. So far the test fail with ``FAIL	github.com/jaegertracing/jaeger/cmd/query/app [build failed]``.

ohdearaugustin: Fixed it, waiting for Travis

PR #2218, opened 2020-05-05
Provide the ability to define the default strategy for one service

### LINKED ISSUE #2171: sampler.param shows defaultSamplingProbability instead of samplingRate
## Apply different probability per service
This issue may be related with [Issue #696](https://github.com/jaegertracing/jaeger-operator/issues/696)

Giving this strategy configuration:
 
```
     {
       "service_strategies": [
         {
           "service": "ServiceA",
           "type": "probabilistic",
           "param": 1.0
         }
       ],
       "default_strategy": {
         "type": "probabilistic",
         "param": 0.2,
         "operation_strategies": [
           {
             "operation": "/health",
             "type": "probabilistic",
             "param": 0.0
           }
         ]
       }
     }
```
The client correctly receives this configuration:
```
{
  "probabilisticSampling": {
    "samplingRate": 1
  },
  "operationSampling": {
    "defaultSamplingProbability": 0.2,
    "defaultLowerBoundTracesPerSecond": 0,
    "perOperationStrategies": [
     {
      "operation": "/health",
      "probabilisticSampling": {
        "samplingRate": 0
       }
     }
   ],
  "defaultUpperBoundTracesPerSecond": 0
  }
}
```

However, the traces for ServiceA have this tag:
```
sampler.param: 0.2  
sampler.type: "probabilistic"
```

The same happens if I change the strategy of serviceA to ratelimit. I don't know if it's just the tag that it's wrong or if it's using the default strategy instead of the correct one.

I'm using this docker image: jaegertracing/all-in-one:1.17.1

defool:
- Fix #2171

- Add a bool flag --sampling.disable-merge-strategies to provide the ability to define the default strategy for one service.

### DIFF
--- a/plugin/sampling/strategystore/static/options.go
+++ b/plugin/sampling/strategystore/static/options.go
@@ -24,6 +24,7 @@ import (
 const (
 	SamplingStrategiesFile           = "sampling.strategies-file"
 	samplingStrategiesReloadInterval = "sampling.strategies-reload-interval"
+	samplingDisableMergeStrategies   = "sampling.disable-merge-strategies"
 )
 
 // Options holds configuration for the static sampling strategy store.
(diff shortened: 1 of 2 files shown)


yurishkuro:
(1) I don't think this change addresses #2171, which was about fixing incorrect behavior rather than disabling merging altogether. Not sure why we need the disabling, as nobody asked for it. Are you thinking of some backwards compatibility support?

(2) If we do fix #2171 in this change, the important part is to add adequate unit tests.

PR #2230, opened 2020-05-07
Enable service default sampling param

### LINKED ISSUE #2171: sampler.param shows defaultSamplingProbability instead of samplingRate

defool:
Signed-off-by: defool <defool@foxmail.com>

-  Fix #2171

- Use `service`'s sampling param event though it has no per-operation strategies.

### DIFF
--- a/plugin/sampling/strategystore/static/strategy_store.go
+++ b/plugin/sampling/strategystore/static/strategy_store.go
@@ -164,7 +164,8 @@ func (h *strategyStore) parseStrategies(strategies *strategies) {
 		// is not merged with and only used as a fallback).
 		opS := newStore.serviceStrategies[s.Service].OperationSampling
 		if opS == nil {
-			if newStore.defaultStrategy.OperationSampling == nil {
+			if newStore.defaultStrategy.OperationSampling == nil ||
+				newStore.serviceStrategies[s.Service].ProbabilisticSampling == nil {
 				continue
 			}
 			// Service has no per-operation strategies, so just reference the default settings and change default samplingRate.
--- a/plugin/sampling/strategystore/static/strategy_store.go
+++ b/plugin/sampling/strategystore/static/strategy_store.go
@@ -164,11 +164,15 @@ func (h *strategyStore) parseStrategies(strategies *strategies) {
 		// is not merged with and only used as a fallback).
 		opS := newStore.serviceStrategies[s.Service].OperationSampling
 		if opS == nil {
-			// Service has no per-operation strategies, so just reference the default settings.
-			newStore.serviceStrategies[s.Service].OperationSampling = newStore.defaultStrategy.OperationSampling
+			if newStore.defaultStrategy.OperationSampling == nil {
+				continue
+			}
+			// Service has no per-operation strategies, so just reference the default settings and change default samplingRate.
+			newOpS := *newStore.defaultStrategy.OperationSampling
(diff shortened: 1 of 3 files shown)


plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -164,11 +164,15 @@ func (h *strategyStore) parseStrategies(strategies *strategies) {
...
 		opS := newStore.serviceStrategies[s.Service].OperationSampling
 		if opS == nil {
-			// Service has no per-operation strategies, so just reference the default settings.
-			newStore.serviceStrategies[s.Service].OperationSampling = newStore.defaultStrategy.OperationSampling
+			if newStore.defaultStrategy.OperationSampling == nil {
+				continue
+			}
+			// Service has no per-operation strategies, so just reference the default settings and change default samplingRate.
+			newOpS := *newStore.defaultStrategy.OperationSampling
+			newOpS.DefaultSamplingProbability = newStore.serviceStrategies[s.Service].ProbabilisticSampling.SamplingRate
```
yurishkuro, line 172: is it possible that `newStore.serviceStrategies[s.Service].ProbabilisticSampling == nil` here?

yurishkuro: For example, in `plugin/sampling/strategystore/static/fixtures/operation_strategies.json` there is
```
    {
      "service": "bar",
      "type": "ratelimiting",
      "param": 5,
      ...
```

yurishkuro: Side node: this type was a mistake introduced in the past:
```
// operationStrategy defines an operation specific sampling strategy.
type operationStrategy struct {
        Operation string `json:"operation"`
        strategy
}
```

because operations can only use probabilistic strategy ([example][]), so this generalized schema allows things that clients can't understand. Similarly, if service-level strategy contains per-operation instructions, then the top-level strategy can _only_ be probabilistic (`default` sampler in the [example][]).

[example]: https://github.com/jaegertracing/jaeger-client-go/blob/4293012d2abb22b7e56d0f30c0335238a4c1a65c/sampler.go#L331

PR #2233, opened 2020-05-08
Remove explicit parsing of OTEL config flag

pavolloffay:

Explicit parsing of the flag is not needed. `GetConfigFile` is called after command run so it should be already intialized.


pavolloffay:
I will merge this because it is a bug that is preventing getting the right config file name. The valid parsing works only if the config flag is first flag after the binary name.

PR #2242, opened 2020-05-12
Do not skip service/operation indexing for firehose spans

yurishkuro:
PR #2090 allowed recording of service and operations in the services/operations tables, but did not index traces by service & operation. This actually made UI even more confusing since the Operations dropdown shows a lot more operations, yet no traces can be found for them.

This change re-enables indexing of firehose traces by service and operation for real. It's a compromise, which brings 3x write amplification for each span (writes from PR #2090 are cached so don't happen often).  However, it significantly improves usability and is still much cheaper than indexing all of the span tags.

### DIFF
--- a/plugin/storage/cassandra/spanstore/writer_test.go
+++ b/plugin/storage/cassandra/spanstore/writer_test.go
@@ -74,6 +74,7 @@ func TestClientClose(t *testing.T) {
 func TestSpanWriter(t *testing.T) {
 	testCases := []struct {
 		caption                        string
+		firehose                       bool
 		mainQueryError                 error
 		tagsQueryError                 error
 		serviceNameQueryError          error
@@ -85,6 +86,10 @@ func TestSpanWriter(t *testing.T) {
 	}{
 		{
 			caption: "main query",
+		},
+		{
+			caption:  "main firehose query",
+			firehose: true,
 		},
 		{
 			caption:        "main query error",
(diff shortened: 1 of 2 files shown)


vprithvi:
Could you add/update tests to ensure that the service name / operation name to traceID indexes are populated?

yurishkuro:
> I feel that perhaps disabling the input fields for tag based search and duration based search when services/operations with firehose spans are selected could lead to better user experience.

This would require much more extensive changes. We can think about it, but I think the current fix is a compromise - yes, you cannot find those traces by tags, but doing that requires a roundtrip to first find the trace, then confirm that it has a tag, and then fail to find by that tag. Where as the current situation is you get operation name in the dropdown, but no traces are found for it - much easier to run into.

plugin/storage/cassandra/spanstore/writer_test.go
```diff
@@ -186,18 +190,21 @@ func TestSpanWriter(t *testing.T) {
...
+				tagsQuery := &mocks.Query{}
+				tagsQuery.On("Exec").Return(testCase.tagsQueryError)
+				tagsQuery.On("String").Return("select from tags")
+
 				durationNoOperationQuery := &mocks.Query{}
 				durationNoOperationQuery.On("Bind", matchEverything()).Return(durationNoOperationQuery)
 				durationNoOperationQuery.On("Exec").Return(testCase.durationNoOperationQueryError)
 				durationNoOperationQuery.On("String").Return("select from duration_index")
 
+				// Define expected order of queries
```
vprithvi, line 202: This comment is rather misleading because one might assume that the expected ordering is enforced on asserts. I think we should remove it. 

> AssertExpectations asserts that everything specified with On and Return was in fact called as expected. Calls may have occurred in any order.

I'm not sure whether this is possible with testify/mock, but gomock provides [InOrder](https://godoc.org/github.com/golang/mock/gomock#InOrder) which does enforce ordering of expectations.

vprithvi:
I agree that it is a compromise and that we should get this in right now. 

I think it's confusing to not communicate these compromise to UI users who don't get any indication that only certain parts of search work for services setting the firehose tag. In fact, even if they end up on an initial firehose trace (via service/operation search), there is no indication that they are viewing a firehose trace and that there are implications wrt search.

yurishkuro:
I booked a ticket for UI change: https://github.com/jaegertracing/jaeger-ui/issues/575

Issue #2250, opened 2020-05-18
Jaeger OTEL Ingester/Collector seems not to save spans to elasticsearch (from kafka)


ning2008wisc:
## Problem - Jaeger OTEL Ingester/Collector seems not to save spans to elasticsearch (from kafka)

I set up an Jaeger OTEL Ingester/Collector to consumer spans from `kafka`, without special processors (e.g. simply use `queued_retry`), then want to save spans to `elasticsearch`. Here is the log of Jaeger OTEL Ingester/Collector, after I send some spans from upstream to the subscribed `kafka` topic (with 2 partitions): 

```
2020-05-18T08:46:33.457Z	INFO	service/service.go:382	Starting Jaeger OpenTelemetry Ingester...	{"Version": "v1.18.0", "GitHash": "75a1d9357a8268ffb447a245388d906073ee7836", "NumCPU": 16}
2020-05-18T08:46:33.457Z	INFO	service/service.go:221	Setting up own telemetry...
2020-05-18T08:46:33.457Z	INFO	service/telemetry.go:94	Serving Prometheus metrics	{"address": "localhost:8888", "legacy_metrics": true, "new_metrics": false, "level": 3, "service.instance.id": ""}
2020-05-18T08:46:33.487Z	INFO	service/service.go:259	Loading configuration...
2020-05-18T08:46:33.500Z	INFO	service/service.go:270	Applying configuration...
2020-05-18T08:46:33.500Z	INFO	service/service.go:291	Starting extensions...
2020-05-18T08:46:33.500Z	INFO	builder/extensions_builder.go:53	Extension is starting...	{"component_kind": "extension", "component_type": "health_check", "component_name": "health_check"}
2020-05-18T08:46:33.500Z	INFO	healthcheckextension/healthcheckextension.go:40	Starting health_check extension	{"component_kind": "extension", "config": {"TypeVal":"health_check","NameVal":"health_check","Port":13133}}
2020-05-18T08:46:33.500Z	INFO	builder/extensions_builder.go:59	Extension started.	{"component_kind": "extension", "component_type": "health_check", "component_name": "health_check"}
2020-05-18T08:46:33.504Z	INFO	config/config.go:173	Elasticsearch detected	{"component_kind": "exporter", "component_type": "jaeger_elasticsearch", "component_name": "jaeger_elasticsearch", "version": 5}
[...]

PR #2251, opened 2020-05-18
Add kafka receiver flags to ingester

pavolloffay:
Related to #2250


pavolloffay:
So far we haven't been testing main classes. The options/builder struct with flags should be tested in the original package.

PR #2254, opened 2020-05-20
Ignore chmod events on UI config watcher.

rubenvp8510:
Signed-off-by: Ruben <ruben.vargas@aluxoft.com>

-  Fixes https://github.com/jaegertracing/jaeger-operator/issues/934 (The issue was filed on jaeger-operator but I think the issue should be transferred to main jaeger project).

- The config file watcher now reacts to `CHMOD` and other events, and should only react when `CREATE` or `WRITE` events occurs. This prevents a lot of log noisy. Specially on kubernetes/openshift environments.

### DIFF
--- a/cmd/query/app/static_handler.go
+++ b/cmd/query/app/static_handler.go
@@ -130,28 +130,21 @@ func (sH *StaticAssetsHandler) configListener(watcher *fsnotify.Watcher) {
 	for {
 		select {
 		case event := <-watcher.Events:
+			// Ignore if the event filename is not the UI configuration or if it is a chmod event.
+			if filepath.Base(event.Name) != filepath.Base(sH.options.UIConfigPath) || event.Op&fsnotify.Chmod == fsnotify.Chmod {
+				continue
+			}
 			if event.Op&fsnotify.Remove == fsnotify.Remove {
-				// this might be related to a file inside the dir, so, just log a warn if this is about the file we care about
-				// otherwise, just ignore the event
-				if event.Name == sH.options.UIConfigPath {
-					sH.options.Logger.Warn("the UI config file has been removed, using the last known version")
-				}
+				sH.options.Logger.Warn("the UI config file has been removed, using the last known version")
 				continue
 			}
-			if event.Op&fsnotify.Write != fsnotify.Write && event.Op&fsnotify.Create != fsnotify.Create {
-				continue
+			// this will catch events for all files inside the same directory, which is OK if we don't have many changes
+			sH.options.Logger.Info("reloading UI config", zap.String("filename", sH.options.UIConfigPath))
+			content, err := loadIndexBytes(sH.assetsFS.Open, sH.options)
+			if err != nil {
+				sH.options.Logger.Error("error while reloading the UI config", zap.Error(err))
 			}
-			if event.Name == sH.options.UIConfigPath {
-				// this will catch events for all files inside the same directory, which is OK if we don't have many changes
-				sH.options.Logger.Info("reloading UI config", zap.String("filename", sH.options.UIConfigPath))
-
-				content, err := loadIndexBytes(sH.assetsFS.Open, sH.options)
-				if err != nil {
-					sH.options.Logger.Error("error while reloading the UI config", zap.Error(err))
-				}
-
-				sH.indexHTML.Store(content)
-			}
+			sH.indexHTML.Store(content)
 		case err, ok := <-watcher.Errors:
 			if !ok {
 				return
--- a/cmd/query/app/static_handler.go
+++ b/cmd/query/app/static_handler.go
@@ -130,8 +130,12 @@ func (sH *StaticAssetsHandler) configListener(watcher *fsnotify.Watcher) {
 	for {
 		select {
 		case event := <-watcher.Events:
-			// Ignore if the event filename is not the UI configuration or if it is a chmod event.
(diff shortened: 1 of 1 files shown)


yurishkuro:
What could be causing so frequent reloads in the original ticket?

cmd/query/app/static_handler.go
```diff
@@ -149,16 +149,17 @@ func (sH *StaticAssetsHandler) watch() {
 					}
 					continue
 				}
+				if event.Op&fsnotify.Write == fsnotify.Write || event.Op&fsnotify.Create == fsnotify.Create {
```
yurishkuro, line 152: 1. reduce nesting by inverting the condition and calling `continue`, as above
2. should this be checking `event.Name` to match the actual config file, rather than reacting to all changes in the directory?
3. another option to reduce logging is to compare the file context before and after, and only log on changed content. This is the approach taken for reloading sampling strategies file

cmd/query/app/static_handler.go
```diff
@@ -149,16 +149,17 @@ func (sH *StaticAssetsHandler) watch() {
 					}
```
yurishkuro, line 149: could we please refactor this nested `go func()` into a top function? There are 5 levels of indentation/nesting here.

rubenvp8510: Sure.

cmd/query/app/static_handler.go
rubenvp8510: 1) Ok, sounds reasonable.

2) I agree , but at the same time I'm not sure why in https://github.com/jaegertracing/jaeger/pull/2254/files#diff-5893b37f4eb854b962cce2c3be922456R180 we watch directory changes too, that line gave me the impression that we want to react to directory changes , but not sure why and I might be misunderstanding something here.

3) Compare file content, we can do that, or we can compute the md5sum or something like that. but Do we really need that? The only scenario which I observed the excessive logging is due `CHMOD` events, of course can be other scenarios and a comparison could prevent future issues with like this one.

rubenvp8510:
> What could be causing so frequent reloads in the original ticket?

As showing here: https://github.com/jaegertracing/jaeger-operator/issues/934#issuecomment-592647439, for some reason that still not clear to me, when the UI configuration is on a kubernetes/openshift `ConfigMap`, there are periodic `CHMOD` events triggered. Not sure if this is due to something internal on kubernetes, but in any case I think we should not reload the UI for those kind of events.

cmd/query/app/static_handler.go
yurishkuro: 2 - I remember something about renaming in k8s, etc. which required watching the directory. And when watching the directory, there could be many other changes happening, although it's still strange that the user experience logs every 0.5s.  By checking that the event is actually for the file name we want we can eliminate all other events in the dir.

3 - we can try with just adding the name check, although doing a full comparison on `(string)content` is trivial, just a couple of lines right in this function.

cmd/query/app/static_handler.go
```diff
@@ -126,30 +126,22 @@ func loadIndexBytes(open func(string) (http.File, error), options StaticAssetsHa
...
-					}
-					continue
+func (sH *StaticAssetsHandler) configListener(watcher *fsnotify.Watcher) {
+	for {
+		select {
+		case event := <-watcher.Events:
+			if event.Op&fsnotify.Remove == fsnotify.Remove {
+				// this might be related to a file inside the dir, so, just log a warn if this is about the file we care about
+				// otherwise, just ignore the event
+				if event.Name == sH.options.UIConfigPath {
```
yurishkuro, line 136: I don't think this is correct comparison. Based on the documentation of `event.Name`, it is the name of the file without the path, whereas `UIConfigPath` may contain the path.

You probably want `filepath.Base(UIConfigPath)`

yurishkuro: also, this check could be done first, before checking for `event.Op`.

rubenvp8510: The documentation said relative path: https://godoc.org/github.com/fsnotify/fsnotify#Event

Anyway, for comparison proposes I get the base of both.

cmd/query/app/static_handler.go
rubenvp8510: Done, I separated the `go func()` into another function. That removes some indentation levels.

cmd/query/app/static_handler.go
rubenvp8510: I did all except the content comparison, I don't think we need it at this point but if you think is good I can add it, as you said is a trivial thing to do.

cmd/query/app/static_handler.go
```diff
@@ -126,6 +126,34 @@ func loadIndexBytes(open func(string) (http.File, error), options StaticAssetsHa
...
 }
 
+func (sH *StaticAssetsHandler) configListener(watcher *fsnotify.Watcher) {
+	for {
+		select {
+		case event := <-watcher.Events:
+			// Ignore if the event filename is not the UI configuration or if it is a chmod event.
+			if filepath.Base(event.Name) != filepath.Base(sH.options.UIConfigPath) || event.Op&fsnotify.Chmod == fsnotify.Chmod {
+				continue
+			}
```
yurishkuro, line 136: no need to combine unrelated checks
```suggestion
			if filepath.Base(event.Name) != filepath.Base(sH.options.UIConfigPath) {
				continue
			}
			if event.Op&fsnotify.Chmod == fsnotify.Chmod {
				continue
			}
```

rubenvp8510: Ah, I just want to have a more compact code. but sure I'll separe it.

Issue #2071, opened 2020-02-16
stdout/stderr sync not implemented for gRPC storage plugins


petermetz:
## Requirement - what kind of business use case are you trying to solve?

Need to propagate logs of a storage plugin gRPC server (implemented not in Go) to the OS stdout+stderr so that logs of the storage plugin can be analyzed.

## Problem - what in Jaeger blocks you from solving the requirement?

`go-plugin` doesn't sync stdout/stderr so there's no way to get the logs of the plugin (logging to file is not feasible in the environment)

## Proposal - what do you suggest to solve the problem or improve the existing situation?

I'm not a 100% sure how to fix it until go-plugin fixes in on their side, but figured maybe a workaround could be added in Jaeger because the relevant go-plugin issue doesn't seem to be getting a fix anytime soon:

https://github.com/hashicorp/go-plugin/issues/93

## Any open questions to address

N/A

yurishkuro:
I don't think we can build anything special in Jaeger, we're fully delegating to go-plugin to start & manage the child process containing the plugin.

See also https://github.com/hashicorp/go-plugin/pull/113#issuecomment-586667462

petermetz:
> I don't think we can build anything special in Jaeger, we're fully delegating to go-plugin to start & manage the child process containing the plugin.
> 
> See also [hashicorp/go-plugin#113 (comment)](https://github.com/hashicorp/go-plugin/pull/113#issuecomment-586667462)

@yurishkuro Makes sense.
Now that https://github.com/hashicorp/go-plugin/pull/135 is done, can the go-plugin version used by Jaeger be updated? It's been a while since we spoke about this so apologies if this was already taken care of in one of the releases since February (in which case this issue can be closed)

yurishkuro:
Jaeger is currently on go-plugin v1.0.1, but https://github.com/hashicorp/go-plugin/pull/135 was released in v1.2, from what I understand. We can try upgrading.

yurishkuro:
closing as we plan to sunset go-plugin based extensions #4647

PR #2261, opened 2020-05-27
Bump go-plugin to 1.3

yurishkuro:
Related to #2071


yurishkuro:
cc @pavolloffay

pavolloffay:
Do the existing plugins have to be rebuild after the plugin version bump?

yurishkuro:
> Do the existing plugins have to be rebuild after the plugin version bump?

I hope not, because there should be no wire API changes in the upgrade.

PR #2263, opened 2020-05-28
Expose option to enable TLS when sniffing an Elasticsearch Cluster

jennynilsen:
Jaeger uses the default scheme set by the olivere client (which is http) when sniffing an Elasticsearch cluster without the option to change it.  This makes it impossible to use sniffing with a TLS Elasticsearch cluster.

The scheme can be set using the SetScheme client option
https://pkg.go.dev/github.com/olivere/elatic/v7\?tab\=doc\#SetScheme

This change exposes that client option as a boolean command line option: --es.sniffer-tls-enabled

Signed-off-by: nilsenj <jennynilsen@rentalcars.com>

In our use case, jaeger-collector needs to connect to Elasticsearch running in Kubernetes from outside the cluster.  By using sniffing, we can use the ingress for the initial node discovery and then send traffic directly to the elasticsearch nodes without it needing to go through the ingress.
In our case, when using ingress only (without sniffing) jaeger represents a substantial proportion of the ingress traffic.  When we switched to using sniffing we were able to reduce the load on our ingress 10x.
If Elasticsearch is running with TLS enabled, then we need a way to tell jaeger-collector to use https when connecting to the discovered nodes.

-

### DIFF
--- a/pkg/es/config/config.go
+++ b/pkg/es/config/config.go
@@ -46,7 +46,7 @@ type Configuration struct {
 	TokenFilePath         string         `mapstructure:"token_file"`
 	AllowTokenFromContext bool           `mapstructure:"-"`
 	Sniffer               bool           `mapstructure:"sniffer"` // https://github.com/olivere/elastic/wiki/Sniffing
-	SnifferTlsEnabled     bool           `mapstructure:"sniffer_tls_enabled"`
+	SnifferTLSEnabled     bool           `mapstructure:"sniffer_tls_enabled"`
 	MaxNumSpans           int            `mapstructure:"-"`                     // defines maximum number of spans to fetch from storage per query
 	MaxSpanAge            time.Duration  `yaml:"max_span_age" mapstructure:"-"` // configures the maximum lookback on span reads
 	NumShards             int64          `yaml:"shards" mapstructure:"num_shards"`
@@ -213,8 +213,8 @@ func (c *Configuration) ApplyDefaults(source *Configuration) {
 	if c.BulkFlushInterval == 0 {
 		c.BulkFlushInterval = source.BulkFlushInterval
 	}
-	if !c.SnifferTlsEnabled {
-		c.SnifferTlsEnabled = source.SnifferTlsEnabled
+	if !c.SnifferTLSEnabled {
+		c.SnifferTLSEnabled = source.SnifferTLSEnabled
 	}
 }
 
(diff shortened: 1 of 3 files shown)


pavolloffay:
What `--es.server-urls` were you using with TLS configuration? We are using TLS with ES in our operator and there wasn't this problem https://github.com/jaegertracing/jaeger-operator/blob/b7ce28f171a416945d881cf7c018280baa4aae01/pkg/storage/elasticsearch.go#L22.

jennynilsen:
The operator does not use sniffing.  It connects the service address which only works from within the cluster.
We want to connect to an Elasticsearch cluster in a different Kubernetes cluster and so --es.server-urls points to the address of an nginx ingress controller.

pkg/es/config/config.go
```diff
@@ -288,6 +292,11 @@ func (c *Configuration) getConfigOptions(logger *zap.Logger) ([]elastic.ClientOp
 		// we don' have a valid token to do the check ad if we don't disable the check the service that
 		// uses this won't start.
 		elastic.SetHealthcheck(!c.AllowTokenFromContext)}
+	if c.SnifferTlsEnabled {
+		options = append(options, elastic.SetScheme("https"))
```
pavolloffay, line 296: Could we derive the scheme from `es.server-urls`? That way we don't have to add new flag.

jennynilsen: Using the scheme from es-server-urls would work for us because both ingress and the elasticsearch nodes are using https.  So, I'm happy to make this change.  However, strictly speaking the two things are independent of each other.  The point of using sniffing is that we can use ingress/loadbalancer for the initial node discovery then connect to the actual elasticsearch nodes for sending traffic.  Are you concerned about somebody that has an https enabled ingress/loadbalancer but uses http for the elasticsearch nodes?

plugin/storage/es/options.go
```diff
@@ -95,6 +96,7 @@ func NewOptions(primaryNamespace string, otherNamespaces ...string) *Options {
 				CreateIndexTemplates: true,
 				Version:              0,
 				Servers:              []string{defaultServerURL},
+				SnifferTlsEnabled:    false,
```
pavolloffay, line 99: can be removed `false` is default

pkg/es/config/config.go
pavolloffay: I thought that people using TLS would normally protect all endpoints. If that is not the case we can merge your patch that adds an additional flag to control TLS just for sniffing. 

The PR looks good I have suggested two edits. Once that is done we can merge it.

pavolloffay:
@jennynilsen would you like to add your company to https://github.com/jaegertracing/jaeger/blob/master/ADOPTERS.md or comment on https://github.com/jaegertracing/jaeger/issues/207?

PR #2274, opened 2020-06-03
Use unique package names for OTEL components

pavolloffay:
Use unique package names to avoid package collisions.


yurishkuro:
travis seems to be having a bad day

PR #2288, opened 2020-06-10
Storage benchmark tests

rubenvp8510:

- This is the first approach to have a framework for storage performance tests.

-  In this first approach I used the storageWriter interface and included a generator of traces. Still need to see what metrics to collect, as now I'm only spans/sec.

@pavolloffay


pavolloffay:
Regarding the trace generator have you also looked at https://github.com/jaegertracing/jaeger/tree/master/cmd/tracegen? Yours seems more configurable, but perhaps we could unify them to a single tool. WDYT @yurishkuro ?

Regarding the test structure. For running the tests on span writer interface I would suggest using go benchmarking `testing.B`. Then for longer running perf tests we probably need to split the data generation and storage into two deployments. This will also allow us to run query tests.

rubenvp8510:
> Regarding the trace generator have you also looked at https://github.com/jaegertracing/jaeger/tree/master/cmd/tracegen? Yours seems more configurable, but perhaps we could unify them to a single tool. WDYT @yurishkuro ?

I took a look at tracegen but it doesn't generate enough random data/tags as far as I see , I agree on unify,

> 
> Regarding the test structure. For running the tests on span writer interface I would suggest using go benchmarking `testing.B`. Then for longer running perf tests we probably need to split the data generation and storage into two deployments. This will also allow us to run query tests.

This work is using the testing.B framework.

rubenvp8510:
@yurishkuro Any thoughts regarding of integrating the trace generator with existing tracegen? and in general any observation/ recommendation for this PR?

yurishkuro:
I would prefer to add new functionality to tracegen over creating a new one. There are not specific critical functionality in tracegen today, we can change it as needed.

jpkrohling:
Closing due to inactivity.

PR #2293, opened 2020-06-16
Add ppc64le support

### LINKED ISSUE #2272: Support ppc64le architecture build
Adding a travis-ci job which runs on ppc64le arch. Any thoughts on this?

Siddhesh-Ghadi:
Signed-off-by: Siddhesh Ghadi <Siddhesh.Ghadi@ibm.com>
- Resolves #2272

- Adds ppc64le support for binaries & runs unit test on ppc64le architecture on travis-ci

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,13 +11,6 @@ matrix:
     env:
     - TESTS=true
     - COVERAGE=true
-    arch: ppc64le
-    addons:
-        apt: jq
-  - go: "1.14.x"
-    env:
-    - TESTS=true
-    - COVERAGE=true
   - go: "1.14.x"
     env:
     - PROTO_GEN_TEST=true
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -7,6 +7,13 @@ dist: trusty
 
 matrix:
   include:
+  - go: "1.14.x"
+    env:
+    - TESTS=true
+    - COVERAGE=true
+    arch: ppc64le
+    addons:
+        apt: jq
```
pavolloffay, line 16: why is this needed?

yurishkuro: why is this whole section needed? We do not run tests per architecture, it's too expensive without significant benefits.

Siddhesh-Ghadi: It is for executing unit test on ppc64le but since it is expensive, I have removed it as requested.

PR #2297, opened 2020-06-21
TLS support for gRPC Query server

### LINKED ISSUE #2248: Support mTLS on query-service gRPC endpoints
The collector's gRPC endpoints already support TLS and mTLS. We need to add similar capability to query service.

jan25:
- Add TLS for gRPC Query server
- Fixes #2248 

- Similar to collector TLS implementation. Made use of `pkg/config/tlscfg` and added new query options

## TODO
~~- Add docs for `--query.grpc.tls.*` options~~ (will go in documentation repo)

### DIFF
--- a/cmd/opentelemetry/cmd/all-in-one/main.go
+++ b/cmd/opentelemetry/cmd/all-in-one/main.go
@@ -200,7 +200,7 @@ func startQuery(v *viper.Viper, logger *zap.Logger, exporter configmodels.Export
 	tracerCloser := initTracer(logger)
 	server, err := queryApp.NewServer(logger, queryService, queryOpts, opentracing.GlobalTracer())
 	if err != nil {
-		logger.Fatal("Could not create jaeger-query service", zap.Error(err))
+		return nil, nil, fmt.Errorf("Could not create jaeger-query service: %w", err)
 	}
 	if err := server.Start(); err != nil {
 		return nil, nil, err
--- a/cmd/opentelemetry/cmd/all-in-one/main.go
+++ b/cmd/opentelemetry/cmd/all-in-one/main.go
@@ -198,7 +198,10 @@ func startQuery(v *viper.Viper, logger *zap.Logger, exporter configmodels.Export
 		*queryServiceOptions)
 
 	tracerCloser := initTracer(logger)
-	server := queryApp.NewServer(logger, queryService, queryOpts, opentracing.GlobalTracer())
+	server, err := queryApp.NewServer(logger, queryService, queryOpts, opentracing.GlobalTracer())
+	if err != nil {
+		logger.Fatal("Could not create jaeger-query service", zap.Error(err))
+	}
 	if err := server.Start(); err != nil {
 		return nil, nil, err
 	}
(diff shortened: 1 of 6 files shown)


yurishkuro:
Please see the build error:
```
vet: cmd/all-in-one/main.go:201:12: cannot initialize 1 variables with 2 values
```

cmd/query/app/server.go
```diff
@@ -47,28 +48,48 @@ type Server struct {
...
+
+	if options.TLS.Enabled {
+		// user requested a server with TLS, setup creds
+		tlsCfg, err := options.TLS.Config()
+		if err != nil {
+			return nil, err
+		}
+
+		creds := credentials.NewTLS(tlsCfg)
+		server = grpc.NewServer(grpc.Creds(creds))
```
yurishkuro, line 84: nit: please define []grpc.Options above and append to it, so that there's only one place where we do
```
server = grpc.NewServer(options...)
```

cmd/opentelemetry/cmd/all-in-one/main.go
```diff
@@ -198,7 +198,10 @@ func startQuery(v *viper.Viper, logger *zap.Logger, exporter configmodels.Export
 		*queryServiceOptions)
 
 	tracerCloser := initTracer(logger)
-	server := queryApp.NewServer(logger, queryService, queryOpts, opentracing.GlobalTracer())
+	server, err := queryApp.NewServer(logger, queryService, queryOpts, opentracing.GlobalTracer())
+	if err != nil {
+		logger.Fatal("Could not create jaeger-query service", zap.Error(err))
```
yurishkuro, line 203: returning the error is more consistent in this function:
```suggestion
		return nil, nil, fmt.Errorf("Could not create jaeger-query service: %w", err)
```

jan25:
Thanks for the review! I adjusted the code based on suggestions. For docs, i'll open a separate PR in documentation repo

cmd/query/app/server.go
```diff
@@ -47,28 +48,47 @@ type Server struct {
...
 	return s.unavailableChannel
 }
 
-func createGRPCServer(querySvc *querysvc.QueryService, logger *zap.Logger, tracer opentracing.Tracer) *grpc.Server {
-	srv := grpc.NewServer()
+func createGRPCServer(querySvc *querysvc.QueryService, options *QueryOptions, logger *zap.Logger, tracer opentracing.Tracer) (*grpc.Server, error) {
+	var grpcOpts []grpc.ServerOption
+
+	if options.TLS.Enabled {
+		// user requested a server with TLS, setup creds
```
yurishkuro, line 77: unnecessary comment

yurishkuro:
@jan25 let me know if you'd like to increase unit test coverage in server.go

image
image

jan25:
Yes, let me see to improve the coverage

jan25:
I see two ways to improve coverage - One is to write `TestCreateServer` that just creates server instance with TLS enabled and tests for nil err. Other is to setup server with TLS, gRPC client and test a RPC call(which may need test certs and keys). Which one would you recommend?

yurishkuro:
It would be nice to have both, but looking at collector code we don't have tests for the client/server TLS. Let's at least add tests for the error cases (like invalid path to a cert file).

In the future we may consolidate starting the gRPC server into a shared package that can be tested for TLS.

yurishkuro:
Thanks, @jan25 🎉

PR #2308, opened 2020-06-25
Remove dependency on tchannel-go

### LINKED ISSUE #2129: Remove TChannel from agent->collector path

pavolloffay:
Resolves #2129 

This PR removes `/tchannel` package and dependency on `tchannel-go` from Jaeger core.

Organizations that are using `tchannel` can copy `tchannel` package to their codebase and initialize tchannel collector and reporter in their mains.

@yurishkuro could you please review this one?


pavolloffay:
Added one more commit to remove tchannel comments from the codebase and remove tchannel reporter metrics.

PR #2314, opened 2020-06-26
Define busy error type in processor

pavolloffay:
Created from https://github.com/jaegertracing/jaeger/pull/2306#discussion_r445591890


cmd/collector/app/handler/grpc_handler.go
```diff
@@ -49,6 +51,9 @@ func (g *GRPCHandler) PostSpans(ctx context.Context, r *api_v2.PostSpansRequest)
 		SpanFormat:       processor.ProtoSpanFormat,
 	})
 	if err != nil {
+		if err == processor.ErrBusy {
+			return nil, status.Errorf(codes.ResourceExhausted, err.Error())
```
pavolloffay, line 55: @yurishkuro it wasn't clear to me what error should we use here https://grpc.io/docs/guides/error/

yurishkuro: I think `GRPC_STATUS_RESOURCE_EXHAUSTED`

PR #2317, opened 2020-06-29
Grpc plugin archive storage support

### LINKED ISSUE #2299: gRPC plugin archive storage support

Existing Elasticsearch and Cassandra as archive storage do not suit me. I want to use grpc plugin as archive storage.

Currently, grpc plugin can't be use as archive storage. 

I propose archive storage support in grpc plugin.

I have following options:

1. Create another namespace `grpc-archive-storage-plugin` in addition to existing `grpc-storage-plugin` and duplicate all options (`binary`, `configuration-file`, `log-level`). This will issue another process in addition to existing `grpc-storage-plugin.binary`. This is most flexible than next option.
2. Implement ArchiveFactory interface in shared.grpcClient with modifying storage.proto in some way...

Which option I should follow?

m8rge:
Signed-off-by: Andrew Putilov <to.merge@gmail.com>

Resolves #2299 

- Added `ArchiveSpanReader` and `ArchiveSpanWriter` to `shared.StoragePlugin` interface.
- Added to `proto/storage.proto` `ArchiveSpanWriterPlugin` and `ArchiveSpanReaderPlugin` services to serve archive storage methods. 
- New rpc `ArchiveSupported() bool` to return is archive storage supported or not.
- Made compatible with old plugins without `ArchiveSpanReaderPlugin` service implemented. Will act same as archive storage not supported by plugin.


m8rge:
codecov/patch fails due /cmd/query/app/static_handler.go file. I didn't modify it.

yurishkuro:
@m8rge ack-ing that I saw the PR, will review over the weekend, unless @pavolloffay gets to it first.

plugin/storage/grpc/README.md
```diff
@@ -62,6 +62,8 @@ A plugin must implement the StoragePlugin interface of:
 type StoragePlugin interface {
    	SpanReader() spanstore.Reader
    	SpanWriter() spanstore.Writer
+	ArchiveSpanReader() shared.ArchiveReader
+	ArchiveSpanWriter() shared.ArchiveWriter
```
yurishkuro, line 66: Extending the interface won't be backwards compatible, contrary to the PR description. Perhaps it would be better to introduce an optional interface 
```
type ArchiveStoragePlugin interface {
   	ArchiveSpanReader() spanstore.Reader
   	ArchiveSpanWriter() spanstore.Writer
}
```
If the plugin does not implement this interface then the corresponding methods on the grpc client (which is internal to Jaeger) can return an error. It is the same pattern as the used by the main `storage.Factory` and `storage.ArchiveFactory`.

plugin/storage/grpc/proto/storage.proto
```diff
@@ -134,6 +134,15 @@ message FindTraceIDsResponse {
...
 }
 
+// empty; extensible in the future
+message ArchiveSupportedRequest {
+
+}
+
+message ArchiveSupportedResponse {
+    bool supported = 1;
+}
```
yurishkuro, line 144: I don't think this is necessary. The internal grpc plugin client can always pretend that `storage.ArchiveFactory` is implemented, but the grpc plugin server will check if the plugin actually implements archive plugin interface, and if not it can return a not-implemented error.

m8rge: To detect whether grpc plugin support archive storage or not I have to call one of methods: `GetArchiveTrace` or `WriteArchiveSpan`. Writing new span every grpc plugin init is bad. I believe calling `GetArchiveTrace` every grpc plugin init is not ideal way either. Thats why I introduce new `ArchiveSupported` method.

yurishkuro: I see, you're trying to match the existing factory behavior and return nil from archive factory methods. Fair enough.

I think it makes sense to introduce a new grpc service:

```
message CapabilitiesRequest { }

message CapabilitiesResponse { 
    bool archiveSpanReader = 1;
    bool archiveSpanWriter = 2;
}

service PluginCapabilities {
    rpc Capabilities(CapabilitiesRequest) returns (CapabilitiesResponse);
}
```

This approach is forward-compatible with new storage interfaces we might want to add in the future. The client will query this service at start-up. Internally, the server can implement this method based on the storage implementing the new interface ArchiveStoragePlugin that I showed above.

plugin/storage/grpc/README.md
yurishkuro: This doesn't match now.

plugin/storage/grpc/config/config.go
```diff
@@ -33,7 +33,7 @@ type Configuration struct {
 }
 
 // Build instantiates a StoragePlugin
-func (c *Configuration) Build() (shared.StoragePlugin, error) {
+func (c *Configuration) Build() (shared.StoragePlugin, shared.ArchiveStoragePlugin, shared.PluginCapabilities, error) {
```
yurishkuro, line 36: please return a struct, e.g. `*PluginServices`, instead of positional values.

plugin/storage/grpc/factory.go
```diff
@@ -88,3 +93,27 @@ func (f *Factory) CreateSpanWriter() (spanstore.Writer, error) {
 func (f *Factory) CreateDependencyReader() (dependencystore.Reader, error) {
 	return f.store.DependencyReader(), nil
 }
+
+// CreateArchiveSpanReader implements storage.ArchiveFactory
+func (f *Factory) CreateArchiveSpanReader() (spanstore.Reader, error) {
+	if f.capabilities == nil {
+		return nil, storage.ErrArchiveStorageNotSupported
+	}
+	capabilities, _ := f.capabilities.Capabilities()
```
yurishkuro, line 102: this is a remote call that can fail, please wrap & bubble up the error

plugin/storage/grpc/shared/grpc_client.go
```diff
@@ -17,6 +17,8 @@ package shared
 import (
 	"context"
 	"fmt"
+	"github.com/jaegertracing/jaeger/plugin/storage/grpc/shared/extra"
+	"google.golang.org/grpc/codes"
```
yurishkuro, line 21: please group imports per convention: stdlib, external, local

plugin/storage/grpc/shared/interface.go
```diff
@@ -46,28 +48,61 @@ type StoragePlugin interface {
...
 }
 
+// ArchiveStoragePlugin is the interface we're exposing as a plugin.
+type ArchiveStoragePlugin interface {
+	ArchiveSpanReader() ArchiveReader
+	ArchiveSpanWriter() ArchiveWriter
+}
+
+// ArchiveReader finds and loads traces and other data from storage.
+type ArchiveReader interface {
```
yurishkuro, line 58: is it necessary to introduce ArchiveReader/ArchiveWriter interfaces? Can we not use just spanstore.Reader/spanstore.Writer?

plugin/storage/grpc/shared/extra/capabilities.go
```diff
@@ -0,0 +1,7 @@
+package extra
```
yurishkuro, line 1: this is used directly from `plugin/storage/grpc/shared/interface.go`. Is there a reason for having this in a separate package?

plugin/storage/grpc/shared/grpc_server.go
```diff
@@ -27,7 +29,9 @@ const spanBatchSize = 1000
 
 // grpcServer implements shared.StoragePlugin and reads/writes spans and dependencies
 type grpcServer struct {
-	Impl StoragePlugin
+	Impl             StoragePlugin
+	ArchiveImpl      ArchiveStoragePlugin
+	CapabilitiesImpl PluginCapabilities
```
yurishkuro, line 34: I was thinking this can be implemented differently: have the single `Impl` that can optionally implement `ArchiveStoragePlugin`, and implement capabilities directly in this server type by checking if `Impl` implements `PluginCapabilities`. This seems like an easier thing to implement for people. For example, the `examples/memstore-plugin/main.go` would not need to change at all.

plugin/storage/grpc/shared/extra/capabilities.go
m8rge: This package was introduced to remove circular imports: `github.com/jaegertracing/jaeger/plugin/storage/grpc/shared` and `github.com/jaegertracing/jaeger/plugin/storage/grpc/shared/mocks`

plugin/storage/grpc/shared/grpc_server.go
m8rge: Do you mean to change `Impl` declaration to `interface{}`? 

In other hands having `Impl StoragePlugin` attribute silently implementing inside numerous other interfaces is not transparent architecture as for me.

plugin/storage/grpc/shared/interface.go
m8rge: I introduce ArchiveReader/ArchiveWriter to have separate method names from spanstore.Reader/spanstore.Writer implemented in grpc plugin. I believe it is necessary to distinguish archive read/write or not.

m8rge:
I'll change README.md last.

m8rge:
@yurishkuro, Could you fill your comments?

plugin/storage/grpc/README.md
yurishkuro: Can we do another interface for Archive?

examples/memstore-plugin/main.go
```diff
@@ -45,7 +46,9 @@ func main() {
 	opts := memory.Options{}
 	opts.InitFromViper(v)
 
-	grpc.Serve(&memoryStore{store: memory.NewStore()})
+	grpc.Serve(&config.PluginServices{
+		Store: &memoryStore{store: memory.NewStore()},
```
yurishkuro, line 50: I think it would be advisable to implement archive/capabilities in the memstore-plugin, as a way of testing the changes in the grpc layer.

plugin/storage/grpc/shared/grpc_client_test.go
```diff
@@ -302,3 +317,138 @@ func TestGRPCClientGetDependencies(t *testing.T) {
...
 	})
 }
+
+func TestGrpcClientWriteArchiveSpan(t *testing.T) {
+	withGRPCClient(func(r *grpcClientTest) {
+		r.archiveWriter.On("WriteArchiveSpan", mock.Anything, &storage_v1.WriteSpanRequest{
+			Span: &mockTraceSpans[0],
+		}).Return(&storage_v1.WriteSpanResponse{}, nil)
+
+		err := r.client.WriteArchiveSpan(&mockTraceSpans[0])
```
yurishkuro, line 327: I think this is a misleading test. Instead of calling WriteArchiveSpan, we should be calling `client.ArchiveWriter().WriteSpan`.

plugin/storage/grpc/shared/archive.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+// FindTraceIDs not used in ArchiveReader
+func (r *ArchiveReader) FindTraceIDs(ctx context.Context, query *spanstore.TraceQueryParameters) ([]model.TraceID, error) {
+	return nil, errors.New("FindTraceIDs not implemented")
+}
+
+// WriteSpan saves the span into Archive Storage
+func (w *ArchiveWriter) WriteSpan(span *model.Span) error {
+	_, err := w.client.WriteArchiveSpan(context.Background(), &storage_v1.WriteSpanRequest{
```
yurishkuro, line 71: the change to add Context to WriteSpan has been merged, please rebase on top of it.

plugin/storage/grpc/shared/archive.go
```diff
@@ -0,0 +1,79 @@
...
+	"errors"
+	"fmt"
+
+	"github.com/jaegertracing/jaeger/model"
+	"github.com/jaegertracing/jaeger/proto-gen/storage_v1"
+	"github.com/jaegertracing/jaeger/storage/spanstore"
+)
+
+// ArchiveReader wraps storage_v1.ArchiveSpanReaderPluginClient into spanstore.Reader
+type ArchiveReader struct {
```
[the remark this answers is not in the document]
m8rge: They're used in `grpcClient` which located in same grpc.shared package.

plugin/storage/grpc/config/config.go
```diff
@@ -67,11 +67,30 @@ func (c *Configuration) Build() (shared.StoragePlugin, error) {
 	if !ok {
 		return nil, fmt.Errorf("unexpected type for plugin \"%s\"", shared.StoragePluginIdentifier)
 	}
+	archiveStoragePlugin, ok := raw.(shared.ArchiveStoragePlugin)
+	if !ok {
+		return nil, fmt.Errorf("unexpected type for plugin \"%s\"", shared.StoragePluginIdentifier)
```
joe-elliott, line 72: We could help debugging by including the type of `raw` in these error messages.  Under normal circumstances this wouldn't fail, but if it did this would be a good clue to finding out why.  Also differentiating the error messages would also be nice `unable to cast %T to shared.ArchiveStoragePlugin` or something

examples/memstore-plugin/main.go
```diff
@@ -45,11 +47,27 @@ func main() {
...
+	})
 }
 
 type memoryStore struct {
-	store *memory.Store
+	store        *memory.Store
+	archiveStore *memory.Store
+}
+
+func (ns *memoryStore) Capabilities() (*extra.Capabilities, error) {
```
joe-elliott, line 66: Instead of making the plugin implement this interface can't we just infer capabilities on whether or not it implements the Archive interfaces.

This could be done automatically in Serve/ServeWithGRPCServer methods in `plugin/storage/grpc/grpc.go`

joe-elliott (commented):
Some minor comments/thoughts.  After this is merged I'm going to look into improving tests per this discussion https://github.com/jaegertracing/jaeger/pull/2455#discussion_r484370971

jpkrohling:
Other than the conflict, what's the status of this PR?

m8rge:
> Other than the conflict, what's the status of this PR?

I'm actively working on comments. Plan to submit changes today.

examples/memstore-plugin/main.go
m8rge: I made this refactor. Now capabilities are calculated from archive implementation: https://github.com/jaegertracing/jaeger/pull/2317/files#diff-7df9e55281a47c4dbf643c4aaa7b6895R168

plugin/storage/grpc/config/config.go
```diff
@@ -65,13 +65,36 @@ func (c *Configuration) Build() (shared.StoragePlugin, error) {
 
 	storagePlugin, ok := raw.(shared.StoragePlugin)
 	if !ok {
-		return nil, fmt.Errorf("unexpected type for plugin \"%s\"", shared.StoragePluginIdentifier)
+		return nil, fmt.Errorf("unable to cast %T to shared.StoragePlugin for plugin \"%s\"",
+			raw, shared.StoragePluginIdentifier)
 	}
+	archiveStoragePlugin, ok := raw.(shared.ArchiveStoragePlugin)
```
yurishkuro, line 71: This looks like we came back to my original proposal of doing a runtime check to determine whether the plugin implements additional interfaces. However, I am puzzled by how this works, given that `raw` is an `interface{}` proxy that's apparently created by the plugin framework to represent the attached plugin binary (which technically may not be even in Go). How does the framework determine that the proxy need to implement these internal interfaces?

yurishkuro (commented):
@m8rge sorry to make you rebase again, but it turns out the CI was not integration-testing the `memstore-plugin`, I just enabled it in #2474. There's a bit of refactoring there to clean up the structures.

go.sum
```diff
@@ -763,6 +766,7 @@ google.golang.org/grpc v1.23.1/go.mod h1:Y5yQAOtifL1yxbo5wqy6BxZv8vAUGQwXBOALyac
 google.golang.org/grpc v1.24.0/go.mod h1:XDChyiUovWa60DnaeDeZmSW86xtLtjtZbwvSiRnRtcA=
 google.golang.org/grpc v1.25.1/go.mod h1:c3i+UQWmh7LiEpx4sFZnkU36qjEYZ0imhYfXVyQciAY=
 google.golang.org/grpc v1.26.0/go.mod h1:qbnxyOmOxrQa7FizSgH+ReBfzJrCY1pSN7KXBS8abTk=
+google.golang.org/grpc v1.28.0-pre/go.mod h1:qbnxyOmOxrQa7FizSgH+ReBfzJrCY1pSN7KXBS8abTk=
```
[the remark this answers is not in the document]
m8rge: Reverted

examples/memstore-plugin/main.go
```diff
@@ -45,11 +46,19 @@ func main() {
 	opts := memory.Options{}
 	opts.InitFromViper(v)
 
-	grpc.Serve(&memoryStore{store: memory.NewStore()})
+	plugin := &memoryStore{
```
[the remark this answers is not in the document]
m8rge: Done

plugin/storage/grpc/config/config.go
m8rge: `raw` is type of `shared.grpcClient` which implements `shared.ArchiveStoragePlugin` and `shared.PluginCapabilities`. Jaeger will call `Capabilities()` method which have following error handling: https://github.com/jaegertracing/jaeger/pull/2317/files#diff-4ad2ef46d844069ff3a195fb9a785ed4R219. If plugin doesn't have `Capabilities` method we fallback to `archive storage not supported` logic.

jpkrohling:
I just restarted the crossdock job.

plugin/storage/grpc/shared/grpc_server_test.go
```diff
@@ -142,7 +158,7 @@ func TestGRPCServerFindTraces(t *testing.T) {
 
 		var traces []*model.Trace
 		var traceID model.TraceID
-		var trace *model.Trace
+		trace := &model.Trace{}
```
yurishkuro, line 161: null ptr as initial value seems safer in the test

plugin/storage/grpc/shared/archive.go
```diff
@@ -0,0 +1,79 @@
...
+	client storage_v1.ArchiveSpanWriterPluginClient
+}
+
+// GetTrace takes a traceID and returns a Trace associated with that traceID from Archive Storage
+func (r *ArchiveReader) GetTrace(ctx context.Context, traceID model.TraceID) (*model.Trace, error) {
+	stream, err := r.client.GetArchiveTrace(upgradeContextWithBearerToken(ctx), &storage_v1.GetTraceRequest{
+		TraceID: traceID,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("plugin error: %w", err)
```
yurishkuro, line 43: I think this may mangle the NotFound error

yurishkuro:
I pushed a bit of clean-up to your branch, hope you don't mind: 9e9bf6a

There's one outstanding issue with GetTrace potentially mangling NotFound. I added a test #2478 to the integration test suite (`make grpc-plugin-storage-integration-test`), but it's not failing in this branch, which is odd.

Otherwise LGTM.

plugin/storage/grpc/shared/grpc_server_test.go
m8rge: I use code from `grpcClient.GetTrace` https://github.com/jaegertracing/jaeger/blob/9e9bf6ad39eab7526b29eebe9fd14c3e6542e5c8/plugin/storage/grpc/shared/grpc_client.go#L91. This code must be equal. We have to fix it in two places if `grpcClient.GetTrace` contains bug.

plugin/storage/grpc/shared/archive.go
m8rge: I copy that code from https://github.com/jaegertracing/jaeger/blob/0a889e8ea7ca887f30eaa814456d86ad97be3f8e/plugin/storage/grpc/shared/grpc_client.go#L69-L75
We have to fix it in two places if it contain a bug.

holycheater: should probably extract grpc status from error and compare it with codes.NotFound in both places

Though, the test seems to be wrong in the original grpc client span-reader `GetTrace` since mock returns `spanstore.ErrTraceNotFound`, but real client will always return grpc status: https://github.com/jaegertracing/jaeger/blob/66f9ca761a26715e05a533fcfd253fb353c4b7e2/plugin/storage/grpc/shared/grpc_client_test.go#L204-L211

m8rge:
I refactored passing "trace not found" error via `status.Errorf(codes.NotFound, spanstore.ErrTraceNotFound.Error())`.

plugin/storage/grpc/shared/grpc_server.go
```diff
@@ -160,3 +167,38 @@ func (s *grpcServer) sendSpans(spans []*model.Span, sendFn func(*storage_v1.Span
...
+		ArchiveSpanWriter: s.ArchiveImpl != nil,
+	}, nil
+}
+
+func (s *grpcServer) GetArchiveTrace(r *storage_v1.GetTraceRequest, stream storage_v1.ArchiveSpanReaderPlugin_GetArchiveTraceServer) error {
+	if s.ArchiveImpl == nil {
+		return status.Error(codes.Unimplemented, "not implemented")
+	}
+	trace, err := s.ArchiveImpl.ArchiveSpanReader().GetTrace(stream.Context(), r.TraceID)
+	if err != nil {
```
yurishkuro, line 183: should this also do this first?
```
	if err == spanstore.ErrTraceNotFound {
		return status.Errorf(codes.NotFound, spanstore.ErrTraceNotFound.Error())
	}
```

m8rge: Missed that. Push fixes

yurishkuro (approved):
LGTM!  🎉  🎉  🎉

yurishkuro:
I merged master because the code coverage was looking wonky, dropping below 95% and showing files that are unrelated to this PR like `cmd/query/app/static_handler.go`.

yurishkuro:
Hm, coverage is still going down, I am trying an experiment #2505 with rebased branch.

yurishkuro:
ok, [this report](https://codecov.io/gh/jaegertracing/jaeger/pull/2505/diff) finally shows what coverage is missing.

@m8rge are there tests you could add to increase it?  And please pull the latest master and rebase your branch, this will allow codecov to do a better job with code coverage .

m8rge:
@yurishkuro how can I trigger codecov? It stuck in "Waiting for status to be reported" status

m8rge:
@yurishkuro Codecov now points to `plugin/storage/grpc/config/config.go` without tests at all. I didn't know how properly test this lines https://github.com/jaegertracing/jaeger/pull/2317/files#annotation_459564037 because it deep into HashiCorp plugin system. Can we just merge this PR?

yurishkuro:
🎉 🎉 🎉 

Thanks for your patience, @m8rge.  Do you mind creating another PR to explain these changes in `plugin/storage/grpc/README.md`?

PR #2320, opened 2020-06-30
Rename ARCH to TARGETARCH for multi platform build by docker buildx

morlay:
- first step to support #1091 

- `TARGETARCH` is the `ARG` need for `buildx` https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope

now could build local by

```bash
make build-ui
make build-query GOOS=linux GOARCH=arm64
make build-all-in-one GOOS=linux GOARCH=arm64
make build-query GOOS=linux GOARCH=amd64
make build-all-in-one GOOS=linux GOARCH=amd64
repo=jaegertracing/all-in-one # all change to custom repo
docker buildx build --platform=linux/amd64,linux/arm64 --push -f cmd/all-in-one/Dockerfile -t $repo:latest cmd/all-in-one
```

demo <https://hub.docker.com/r/querycap/jaegertracing-all-in-one/tags>

In travis, need to setup buildx with qemu <https://github.com/crazy-max/ghaction-docker-buildx>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -205,7 +205,7 @@ build-tracegen:
 .PHONY: docker-hotrod
 docker-hotrod:
 	GOOS=linux $(MAKE) build-examples
-	docker build -t $(DOCKER_NAMESPACE)/example-hotrod:${DOCKER_TAG} ./examples/hotrod --build-arg ARCH=$(GOARCH)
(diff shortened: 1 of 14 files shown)


morlay:
mv all build in docker will be make docker buildx happy.

morlay:
@pavolloffay sure.

Issue #2292, opened 2020-06-15
Multiarch docker images


rur0:
## Requirement - what kind of business use case are you trying to solve?
Ability to utilize multiarch images

## Problem - what in Jaeger blocks you from solving the requirement?
I have a rpi k3s (arm64) cluster that I use to run a web app written in golang, and I would like to use jaeger for tracing on it. I have compiled jaeger-operator and got it running on the cluster, however when it pulls the main jaeger docker image- it errors out with 'exec format'.

## Proposal - what do you suggest to solve the problem or improve the existing situation?
Build and upload multiarch images to official docker repo

yurishkuro:
Are you referring to a single image somehow being multi-arch, or to multiple images? There would be a lot of manual work on creating individual docker hub repo for each one permutation.

pavolloffay:
@rur0 thanks for creating the issue. 

This has been created from https://github.com/jaegertracing/jaeger-operator/issues/1086.

I think he is referring to https://docs.docker.com/buildx/working-with-buildx/. It seems that a single image can have multiple architectures. When the client requests the image from the registry it gets multiple manifests. Then the client chooses the right image.

Other resources: https://www.youtube.com/watch?v=SYJgkkjqd7s

@rur0 would you like to contribute this?

rur0:
@pavolloffay Yes

Jingzhao123:
@MrXinWang

Issue #2116, opened 2020-03-05
Change base image


jpkrohling:
# Requirement - what kind of business use case are you trying to solve?

When an application container has a problem in production, it's often useful to be able to enter the container and execute a few commands, like `curl`, or to copy files from inside the container using commands like `kubectl cp` . Unfortunately, using a scratch image means that there's nothing in the container, breaking the two use cases above. Concretely, I'm debugging a problem with the badger storage and tried to get the files out of the container using `kubectl cp`.

While all those problems can be worked around by either building a custom image with another base image, or by using another image to mount the volumes we want to access, not having a proper base image slows down progress when trying to find the root cause of problems related to Jaeger.

## Proposal - what do you suggest to solve the problem or improve the existing situation?

There are a few images that we could use. The image I have most experience with is Red Hat's UBI, which is used in quite a few operators, including OpenTelemetry's ([minimal](https://github.com/open-telemetry/opentelemetry-operator/blob/4593927867598681b2a13d30012c15ba98b5749c/build/Dockerfile#L1)) and Jaeger's ([regular](https://github.com/jaegertracing/jaeger-operator/blob/deec90def3c674ee3c3656426a1295f04930a3b4/build/Dockerfile#L1)). I know this is a secure and well-maintained image, but we should certainly consider other images as well. I'm open for suggestions here.

Using the ubi-minimal would add about 90MB to the image size, which isn't that much, especially considering that this is mostly relevant for the initial pull.

yurishkuro:
Adding 90Mb means a 6x increase in the current image size. I would rather publish another group of -dev images.

jpkrohling:
It really sounds dramatic when we talk about "6x" increase, but in absolute terms, 90MB isn't even a minute away from most domestic broadband connections, to be shared across all components.

Is there a concrete concern about the image size, other than the time it takes to download it?

yurishkuro:
Jaeger has often been praised for being lean, I think our original 10Mb images contributed to it (not they are 20Mb). I also think that having dev tools in the image is something that only a small % of users need, e.g. when setting up Jaeger for the first time, figuring out networking namespaces, etc.

jpkrohling:
"Lean" as in the container image size, or as in *binary* size? The container isn't placed as a whole in memory, just the binary. The process binary is exactly the same no matter the base image, and the RSS is virtually identical:

```
## scratch
$ ps -q 362569 uxwww
[...]

PR #2325, opened 2020-07-02
Setup docker buildx for all-in-one

morlay:
- buildx in travis.yml to support #2292 

```bash
# file ./scripts/travis/build-all-in-one-image.sh

bash ./scripts/travis/setup-docker-buildx.sh


# multi binary build
make build-all-in-one GOOS=linux GOARCH=amd64
make build-all-in-one GOOS=linux GOARCH=arm64

repo=jaegertracing/all-in-one

# load to local docker for integration_test
docker buildx build --load -f cmd/all-in-one/Dockerfile -t $repo:latest cmd/all-in-one
run_integration_test $repo
# push after testing
# could push multi tags by multi `-t $repo:latest -t $repo:1.18 -t $repo:1.18.1`
# it will load from buildkit cache
docker buildx build --push -f cmd/all-in-one/Dockerfile -t $repo:latest cmd/all-in-one
```

@pavolloffay 
have to change upload-to-docker, need to discuss.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -173,3 +173,41 @@ Before merging a PR make sure:
 
 Merge the PR by using "Squash and merge" option on Github. Avoid creating merge commits.
 After the merge make sure referenced issues were closed.
+
+### Developing guidelines
+
+#### Building OCI Images for multiple arch (linux/arm64, linux/amd64)
+
+[OCI Images](https://github.com/opencontainers/image-spec/blob/master/spec.md) could be built and published by [buildx](https://github.com/docker/buildx), 
+it could be executed for local publish `all-in-one` images via:
+
+```shell
+TRAVIS_SECURE_ENV_VARS=true NAMESPACE=$(whoami) BRANCH=master ./scripts/travis/build-all-in-one-image.sh
+```
+
+more arch support only need to change `--platform=linux/amd64,linux/arm64,linux/s390x`
+
+if we want to execute this in local env, need to setup buildx:
+
+1. install docker cli plugin
+
+```shell
+$ export DOCKER_BUILDKIT=1
+$ docker build --platform=local -o . git://github.com/docker/buildx
+$ mkdir -p ~/.docker/cli-plugins
+$ mv buildx ~/.docker/cli-plugins/docker-buildx
+```
+(via https://github.com/docker/buildx#with-buildx-or-docker-1903, if docker issue, could check https://docs.docker.com/engine/install/linux-postinstall/#troubleshooting)
+
+2. install qemu for multi arch
+
+```shell
+$ docker run --privileged --rm tonistiigi/binfmt --install all
+```
+(via https://github.com/docker/buildx#building-multi-platform-images)
+
+3. create a builder
+
+```shell
+$ docker buildx create --use --name builder
+```
--- a/cmd/all-in-one/Dockerfile
+++ b/cmd/all-in-one/Dockerfile
@@ -1,8 +1,23 @@
-ARG base_image
-ARG debug_image
+ARG TARGET=release
 
-FROM $base_image AS release
-ARG TARGETARCH=amd64
+FROM golang:1.15-alpine as build
+ENV GOPATH /go
+
+RUN apk add --update --no-cache ca-certificates make git && \
+    go get github.com/go-delve/delve/cmd/dlv && \
+    cd /go/src/github.com/go-delve/delve && \
+    make install
+
+FROM golang:1.15-alpine AS base-image-debug
+
+COPY --from=build /go/bin/dlv /go/bin/dlv
+COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
+
+FROM alpine:3.12 AS base-image-release
+
+COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
+
+FROM base-image-${TARGET} as common
 
 # Agent zipkin.thrift compact
 EXPOSE 5775/udp
@@ -25,43 +40,27 @@ EXPOSE 14250
 # Web HTTP
 EXPOSE 16686
 
-COPY all-in-one-linux-$TARGETARCH /go/bin/all-in-one-linux
 COPY sampling_strategies.json /etc/jaeger/
 
 VOLUME ["/tmp"]
+
+FROM common AS release
+
+ARG TARGETARCH
+COPY all-in-one-linux-${TARGETARCH} /go/bin/all-in-one-linux
+
 ENTRYPOINT ["/go/bin/all-in-one-linux"]
 CMD ["--sampling.strategies-file=/etc/jaeger/sampling_strategies.json"]
 
-FROM $debug_image AS debug
-ARG TARGETARCH=amd64
+FROM common AS debug
 
-# Agent zipkin.thrift compact
-EXPOSE 5775/udp
-
-# Agent jaeger.thrift compact
-EXPOSE 6831/udp
-
-# Agent jaeger.thrift binary
-EXPOSE 6832/udp
-
-# Agent config HTTP
-EXPOSE 5778
-
-# Collector HTTP
-EXPOSE 14268
-
-# Collector gRPC
-EXPOSE 14250
-
-# Web HTTP
-EXPOSE 16686
+ARG TARGETARCH
+COPY all-in-one-debug-linux-${TARGETARCH} /go/bin/all-in-one-linux
 
 # Delve
 EXPOSE 12345
 
-COPY all-in-one-debug-linux-$TARGETARCH /go/bin/all-in-one-linux
-COPY sampling_strategies.json /etc/jaeger/
-
-VOLUME ["/tmp"]
 ENTRYPOINT ["/go/bin/dlv", "exec", "/go/bin/all-in-one-linux", "--headless", "--listen=:12345", "--api-version=2", "--accept-multiclient", "--log", "--"]
 CMD ["--sampling.strategies-file=/etc/jaeger/sampling_strategies.json"]
+
+FROM ${TARGET}
--- a/scripts/travis/build-all-in-one-image.sh
+++ b/scripts/travis/build-all-in-one-image.sh
@@ -7,6 +7,9 @@ BRANCH=${BRANCH:?'missing BRANCH env var'}
 # be overrided by passing architecture value to the script:
 # `GOARCH=<target arch> ./scripts/travis/build-all-in-one-image.sh`.
 GOARCH=${GOARCH:-$(go env GOARCH)}
+# local run with `TRAVIS_SECURE_ENV_VARS=true NAMESPACE=$(whoami) BRANCH=master ./scripts/travis/build-all-in-one-image.sh`
+NAMESPACE=${NAMESPACE:-jaegertracing}
+ARCHS="amd64 arm64 s390x"
 
 source ~/.nvm/nvm.sh
 nvm use 10
@@ -26,42 +29,89 @@ run_integration_test() {
   docker kill $CID
 }
 
-upload_to_docker() {
-  # Only push the docker container to Docker Hub for master branch
-  if [[ ("$BRANCH" == "master" || $BRANCH =~ ^v[0-9]+\.[0-9]+\.[0-9]+$) && "$TRAVIS_SECURE_ENV_VARS" == "true" ]]; then
-    echo 'upload to Docker Hub'
-  else
-    echo 'skip docker upload for PR'
-    exit 0
+docker_buildx() {
+  CMD_ROOT=${1}
+  FLAGS=${@:2}
+
+  if [[ "$FLAGS" == *"-push"*  ]]; then
+    if [[ ("$BRANCH" == "master" || $BRANCH =~ ^v[0-9]+\.[0-9]+\.[0-9]+$) && "$TRAVIS_SECURE_ENV_VARS" == "true" ]]; then
+      echo 'upload to Docker Hub'
+    else
+      echo 'skip docker upload for PR'
+      exit 0
+    fi
   fi
-  export REPO=$1
-  bash ./scripts/travis/upload-to-docker.sh
+
+  docker buildx build -f ${CMD_ROOT}/Dockerfile ${FLAGS} ${CMD_ROOT}
 }
 
-make build-all-in-one GOOS=linux GOARCH=$GOARCH
-repo=jaegertracing/all-in-one
-docker build -f cmd/all-in-one/Dockerfile \
-        --target release \
-        --tag $repo:latest cmd/all-in-one \
-        --build-arg base_image=localhost/baseimg:1.0.0-alpine-3.12 \
-        --build-arg debug_image=localhost/debugimg:1.0.0-golang-1.15-alpine \
-        --build-arg TARGETARCH=$GOARCH
-run_integration_test $repo
-upload_to_docker $repo
+image_tags_for() {
+  REPO=${1}
 
-make build-all-in-one-debug GOOS=linux GOARCH=$GOARCH
-repo=jaegertracing/all-in-one-debug
-docker build -f cmd/all-in-one/Dockerfile \
-        --target debug \
-        --tag $repo:latest cmd/all-in-one \
-        --build-arg base_image=localhost/baseimg:1.0.0-alpine-3.12 \
-        --build-arg debug_image=localhost/debugimg:1.0.0-golang-1.15-alpine \
-        --build-arg TARGETARCH=$GOARCH
-run_integration_test $repo
-upload_to_docker $repo
+  major=""
+  minor=""
+  patch=""
 
-make build-otel-all-in-one GOOS=linux GOARCH=$GOARCH
-repo=jaegertracing/opentelemetry-all-in-one
-docker build -f cmd/opentelemetry/cmd/all-in-one/Dockerfile -t $repo:latest cmd/opentelemetry/cmd/all-in-one --build-arg TARGETARCH=$GOARCH
-run_integration_test $repo
-upload_to_docker $repo
+  if [[ "$BRANCH" == "master" ]]; then
+    TAG="latest"
+  elif [[ $BRANCH =~ ^v([0-9]+)\.([0-9]+)\.([0-9]+)$ ]]; then
+    major="${BASH_REMATCH[1]}"
+    minor="${BASH_REMATCH[2]}"
+    patch="${BASH_REMATCH[3]}"
+    TAG=${major}.${minor}.${patch}
+  else
+    TAG="${BRANCH///}"
+  fi
+
+  IMAGE_TAGS="--tag ${REPO}:${TAG}"
+
+  # add major, major.minor and major.minor.patch tags
+  if [[ -n $major ]]; then
+    IMAGE_TAGS="${IMAGE_TAGS} -t ${REPO}:${major}"
+    if [[ -n $minor ]]; then
+       IMAGE_TAGS="${IMAGE_TAGS} -t ${REPO}:${major}.${minor}"
+      if [[ -n $patch ]]; then
+          IMAGE_TAGS="${IMAGE_TAGS} -t ${REPO}:${major}.${minor}.${patch}"
+      fi
+    fi
+  fi
+
+  echo "${IMAGE_TAGS}"
+}
+
+target_platforms() {
+  PLATFORMS=""
+  for arch in ${ARCHS}; do
+    if [ -n "${PLATFORMS}" ]; then
+      PLATFORMS="${PLATFORMS},linux/${arch}"
+    else
+      PLATFORMS="linux/${arch}"
+    fi
+  done
+  echo ${PLATFORMS}
+}
+
+
+for arch in ${ARCHS}; do
+  make build-all-in-one GOOS=linux GOARCH=${arch}
+done
+repo=${NAMESPACE}/all-in-one
+docker_buildx cmd/all-in-one --load --build-arg=TARGET=release --platform=linux/${GOARCH} -t $repo:latest
+run_integration_test ${repo}
+docker_buildx cmd/all-in-one --push --build-arg=TARGET=release --platform=$(target_platforms) $(image_tags_for ${repo})
+
+for arch in ${ARCHS}; do
+  make build-all-in-one-debug GOOS=linux GOARCH=${arch}
+done
+repo=${NAMESPACE}/all-in-one-debug
+docker_buildx cmd/all-in-one --load --build-arg=TARGET=debug --platform=linux/${GOARCH} -t $repo:latest
+run_integration_test ${repo}
+docker_buildx cmd/all-in-one --push --build-arg=TARGET=debug --platform=$(target_platforms) $(image_tags_for ${repo})
+
+for arch in ${ARCHS}; do
+  make build-otel-all-in-one GOOS=linux GOARCH=${arch}
+done
+repo=${NAMESPACE}/opentelemetry-all-in-one
+docker_buildx cmd/opentelemetry/cmd/all-in-one --load --platform=linux/${GOARCH} -t $repo:latest
+run_integration_test ${repo}
+docker_buildx cmd/opentelemetry/cmd/all-in-one --push --platform=$(target_platforms) $(image_tags_for ${repo})
--- a/scripts/travis/setup-docker-buildx.sh
+++ b/scripts/travis/setup-docker-buildx.sh
@@ -1,0 +1,56 @@
+#!/bin/bash
+
+set -eux
+
+# update docker
+# for setup buildx https://travis-ci.community/t/installing-docker-19-03/8077/2
+sudo apt update
+
+sudo systemctl stop docker
+sudo apt install -y docker.io
+
+sudo systemctl unmask docker.service
+sudo systemctl unmask docker.socket
+sudo systemctl start docker
+sudo systemctl status docker.socket
+
+docker version
+
+DOCKER_BUILDX_VERSION=v0.4.2
+
+LOCAL_OS=$(uname -s | tr '[A-Z]' '[a-z]')
+
+case $(uname -m) in
+x86_64)
+  LOCAL_ARCH=amd64
+  ;;
+aarch64)
+  LOCAL_ARCH=arm64
+  ;;
+*)
+  echo "unsupported architecture"
+  exit 1
+  ;;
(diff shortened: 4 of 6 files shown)


pavolloffay:
Does this build images for all architectures `docker buildx build --load -f cmd/all-in-one/Dockerfile -t $repo:latest cmd/all-in-one`?

pavolloffay:
Could you please enable the multi arch build on travis and if possible reuse the existing make targets? Ideally we should not require dependency on buildx when building a single arch locally.

morlay:
`docker buildx build --load -f --platform=linux/arm64,linux/amd64 cmd/all-in-one/Dockerfile -t $repo:latest cmd/all-in-one` will build arm64 & amd64 both.

but now in travis, need to run tests after building. this why `--load` here.

after testings, we could use `--push` to push mutli arch images.

and now, in your scripts, will tag for mutil versions.
and buildx not reuse the `./scripts/travis/upload-to-docker.sh`, i am no ideas should i refactor that file? or just refactor `./scripts/travis/build-all-in-one-image.sh`

morlay:
@pavolloffay i updated `./scripts/travis/build-all-in-one-image.sh` with buildx.

pavolloffay:
@morlay could you please open a separate PR with change from trusty to bionic?

morlay:
@pavolloffay sure

scripts/travis/build-all-in-one-image.sh
```diff
@@ -34,18 +42,54 @@ upload_to_docker() {
     echo 'skip docker upload for PR'
     exit 0
   fi
-  export REPO=$1
-  bash ./scripts/travis/upload-to-docker.sh
+
+  REPO=${1}
+  CMD_ROOT=${1}
+
+  unset major minor patch
```
pavolloffay, line 49: why has been this copied from `upload-to-docker.sh`? Could we just keep it there?

morlay: with buildx, we could not use tag directly.

`docker tag` just tag local image for local arch, when we push that tags, it will break the oci image with multi arch.

pavolloffay:
@jaegertracing/jaeger-maintainers any objections on merging this? We could the multiarch build for only the all-in-one initially.

scripts/travis/build-all-in-one-image.sh
```diff
@@ -34,18 +42,54 @@ upload_to_docker() {
...
+      fi
+    fi
+  fi
+
+  docker_buildx ${CMD_ROOT} ${IMAGE_TAGS} --platform=linux/amd64,linux/arm64  --push
 }
 
-make build-all-in-one GOOS=linux GOARCH=$GOARCH
+GOOS=linux GOARCH=amd64 make build-all-in-one
+GOOS=linux GOARCH=arm64 make build-all-in-one
```
yurishkuro, line 82: nit: these vars are Make vars, not env vars, they can go on the right side.

scripts/travis/build-all-in-one-image.sh
yurishkuro: ok, but  there's a lot being copied here, and would have to be repeated for other Docker images.
Can we not either parameterize `upload-to-docker.sh` or refactor to avoid repetition?

scripts/travis/build-all-in-one-image.sh
```diff
@@ -34,18 +42,54 @@ upload_to_docker() {
     echo 'skip docker upload for PR'
     exit 0
   fi
-  export REPO=$1
-  bash ./scripts/travis/upload-to-docker.sh
+
+  REPO=${1}
+  CMD_ROOT=${1}
```
yurishkuro, line 47: both of these are `$1`?

scripts/travis/build-all-in-one-image.sh
morlay: yep. i reverted it

scripts/travis/build-all-in-one-image.sh
morlay: Sorry. CMD_ROOT should be `$2`. fixed it.

scripts/travis/build-all-in-one-image.sh
pavolloffay: I am fine to go with this and see how it works and refactor other images in a second PR

morlay:
@pavolloffay 

anything need i to do next?

pavolloffay (dismissed):
I will approve it and merge next week. 

thanks @morlay!

davidsbond:
@pavolloffay is there any chance to get this merged in? I'd like to be able to use the `allinone` image on my raspberry pi swarm

morlay:
@davidsbond 
you could use https://hub.docker.com/r/querycapistio/all-in-one/tags before official support.

source here. https://github.com/querycap/istio/tree/master/build/jaegertracing

jpkrohling:
@Mergifyio rebase

jpkrohling (commented):
This looks nice, but we do need this documented somewhere in case we need to build the images locally. Would be nice to get positive confirmation from other reviewers (@yurishkuro and @pavolloffay) that you ran this locally and are able to build those images. The instructions from the jaeger-operator repo can be used as base for this.

Please, wait to work on the other images, as there's another PR that will conflict with this: #2545.

morlay:
@jpkrohling 

local build here could just run `./scripts/travis/build-all-in-one-image.sh`
a sub script to setup buildx included.

jaeger-operator use github workflow for CI , so could use with actions.

conflicts #2545 could merged that pr first. 
i can do rebase and fix conflicts.

jpkrohling:
> local build here could just run ./scripts/travis/build-all-in-one-image.sh
> a sub script to setup buildx included.

True, but we still need it documented somewhere. Not everyone is using Ubuntu, and we've seen that there are things specific to more secured distributions like Fedora that need to be done.

> conflicts #2545 could merged that pr first.
> i can do rebase and fix conflicts.

Thanks!

morlay:
@jpkrohling 

But i am not familiar with Fedora

is the docker offical troubleshooting cover all issues?
https://docs.docker.com/engine/install/linux-postinstall/#troubleshooting

here a checker too.
```
$ curl https://raw.githubusercontent.com/docker/docker/master/contrib/check-config.sh > check-config.sh
$ bash ./check-config.sh
```

jpkrohling:
Having something just like what we have for the operator would be sufficient: 

https://github.com/jaegertracing/jaeger-operator/blob/master/CONTRIBUTING.md#building-oci-images-for-multiple-arch-linuxarm64-linuxamd64

morlay:
documents added. 

but `scripts/travis/build-all-in-one-image.sh` under travis folder. should we need to run local?

CONTRIBUTING.md
```diff
@@ -153,3 +153,49 @@ Before merging a PR make sure:
...
+```
+
+more arch support only need to change `--platform=linux/amd64,linux/arm64`
+
+if we want to execute this in local env, need to setup buildx:
+
+0. ensure docker ready
+
+```
+$ curl https://raw.githubusercontent.com/docker/docker/master/contrib/check-config.sh > check-config.sh
```
jpkrohling, line 174: Can we just go without this? If this is part of the official instructions, then let's just point users there.

jpkrohling:
> but scripts/travis/build-all-in-one-image.sh under travis folder. should we need to run local?

No, it's fine to keep it there.

CONTRIBUTING.md
morlay: sure.

jpkrohling:
>  i can do rebase and fix conflicts.

#2545 has just been merged. Do you want to update this PR?

morlay:
Sure. i will updates this pr tomorrow.

morlay:
@jpkrohling 

Have to remove `docker/*` and merge multi stages into `cmd/all-in-one/Dockerfile`
spliting base image builds have trouble for multi arch build (need to build base image for multi arch too.)

local run (after buildx ready):
```shell
TRAVIS_SECURE_ENV_VARS=true NAMESPACE=$(whoami) BRANCH=v1.20.0 ./scripts/travis/build-all-in-one-image.sh
```

results:
https://hub.docker.com/u/morlay/all-in-one/tags
https://hub.docker.com/u/morlay/all-in-one-debug/tags
https://hub.docker.com/u/morlay/opentelemetry-all-in-one/tags

s390x included following jaeger-operator did. 

btw, Why changed base image of `all-in-one` from `scratch` to `alpine:3.12` ? 
i reverted to `scratch` in this pr too.

CONTRIBUTING.md
```diff
@@ -173,3 +173,40 @@ Before merging a PR make sure:
 
 Merge the PR by using "Squash and merge" option on Github. Avoid creating merge commits.
 After the merge make sure referenced issues were closed.
+
+### Developing guidelines
+
+#### Building [OCI Images](https://github.com/opencontainers/image-spec/blob/master/spec.md) for multiple arch (linux/arm64, linux/amd64)
```
jpkrohling, line 179: nit: can you keep the title without a link, and add the link to "OCI images" in the line 181?

CONTRIBUTING.md
```diff
@@ -173,3 +173,40 @@ Before merging a PR make sure:
...
+
+OCI images could be built and published by [buildx](https://github.com/docker/buildx), it could be executed for local test via:
+
+```
+$ OPERATOR_VERSION=devel ./.ci/publish-images.sh
```
jpkrohling, line 184: This isn't the operator repository ;-) The instructions here need to fit this repo.

plugin/storage/es/mappings/gen_assets.go
```diff
@@ -213,7 +213,7 @@ var _escData = map[string]*_escFile{
 		name:    ".nocover",
 		local:   "plugin/storage/es/mappings/.nocover",
 		size:    43,
-		modtime: 1597437395,
```
jpkrohling, line 216: This change does not belong to this PR.

scripts/travis/build-all-in-one-image.sh
```diff
@@ -26,42 +29,89 @@ run_integration_test() {
...
+    minor="${BASH_REMATCH[2]}"
+    patch="${BASH_REMATCH[3]}"
+    TAG=${major}.${minor}.${patch}
   else
-    echo 'skip docker upload for PR'
-    exit 0
+    TAG="${BRANCH///}"
+  fi
+
+  IMAGE_TAGS="--tag ${REPO}:${TAG}"
```
jpkrohling, line 66: @aebirim, for your change, you might just need to add a new tag here, as well as a `docker login` to all configured repositories.

scripts/travis/build-all-in-one-image.sh
```diff
@@ -26,42 +29,89 @@ run_integration_test() {
...
-upload_to_docker $repo
-
-make build-otel-all-in-one GOOS=linux GOARCH=$GOARCH
-repo=jaegertracing/opentelemetry-all-in-one
-docker build -f cmd/opentelemetry/cmd/all-in-one/Dockerfile -t $repo:latest cmd/opentelemetry/cmd/all-in-one --build-arg TARGETARCH=$GOARCH
-run_integration_test $repo
-upload_to_docker $repo
+
+for arch in ${ARCHS}; do
+  make build-all-in-one GOOS=linux GOARCH=${arch}
```
jpkrohling, line 96: What's the difference in the clock time for this step with all arches vs. only one arch?

scripts/travis/build-all-in-one-image.sh
```diff
@@ -26,42 +29,89 @@ run_integration_test() {
...
+done
+repo=${NAMESPACE}/all-in-one
+docker_buildx cmd/all-in-one --load --build-arg=TARGET=release --platform=linux/${GOARCH} -t $repo:latest
+run_integration_test ${repo}
+docker_buildx cmd/all-in-one --push --build-arg=TARGET=release --platform=$(target_platforms) $(image_tags_for ${repo})
+
+for arch in ${ARCHS}; do
```
jpkrohling, line 103: I see potential for a refactoring here, which we should do in a next PR.

scripts/travis/build-all-in-one-image.sh
morlay: we need to build each go binary for each arch.

the final Dockerfile not includes build stage, just copy binary from the step.

using `for in` here, only want to make ARCHS be easier to configurate

scripts/travis/build-all-in-one-image.sh
morlay: Yes. if other components need to support multi-arch. need to mv to a common shell file.

scripts/travis/build-all-in-one-image.sh
jpkrohling: My question is mainly to determine if we have a penalty in build time for CI. If it takes 5m to build *each* arch, then we'll have to rethink the approach. If it takes 5m for the first plus 1m for each subsequent arch, then it's fine.

morlay: it will not take 5m for each arch. 

1. go build for amd64, go download deps (3m) + go prebuild common (0.5m) + go prebuild for amd64 (0.5m) + go build bin (1m)
2. go build for arm64, go prebuild for arm64 (0.5m) + go build bin (1m)
2. go build for s390x, go prebuild for s390x (0.5m) + go build bin (1m)

if we split, it will take 5m for each arch. this why i upgrade for jaeger-operator.

jpkrohling: Great, thanks for the confirmation. I wasn't sure whether the download step would be shared across steps. If it adds 90s for each subsequent arch, I think it's reasonable.

CONTRIBUTING.md
```diff
@@ -173,3 +173,41 @@ Before merging a PR make sure:
 
 Merge the PR by using "Squash and merge" option on Github. Avoid creating merge commits.
 After the merge make sure referenced issues were closed.
+
+### Developing guidelines
+
+#### Building OCI Images for multiple arch (linux/arm64, linux/amd64)
```
jpkrohling, line 179: This section needs some better formatting and grammar check (@objectiser?), but I'm fine in having this done by someone else or in a future PR.

jpkrohling (approved):
I'll leave it open for the day, to give other maintainers the chance to review.

cmd/all-in-one/Dockerfile
```diff
@@ -1,8 +1,14 @@
...
+ARG TARGET=release
 
-FROM $base_image AS release
-ARG TARGETARCH=amd64
+FROM golang:1.15-alpine AS base-image-debug
+
+FROM alpine:3.12 AS cert
+RUN apk add --update --no-cache ca-certificates
+
+FROM scratch AS base-image-release
```
Ashmita152, line 8: Are we changing back the base image back to scratch ?

jpkrohling: Good catch! No, we shouldn't change it back to scratch.

jpkrohling (commented):
Dismissing the approval, as the base image shouldn't be changed.

cmd/all-in-one/Dockerfile
morlay: i asked. 

Why we need to change to alpine for the release image

jpkrohling: See #2545 for context.

morlay: Sure. 
got info from https://github.com/jaegertracing/jaeger/issues/2116. 
I have put it back.

cmd/all-in-one/Dockerfile
```diff
@@ -1,8 +1,14 @@
-ARG base_image
-ARG debug_image
+ARG TARGET=release
 
-FROM $base_image AS release
-ARG TARGETARCH=amd64
+FROM golang:1.15-alpine AS base-image-debug
```
Ashmita152, line 3: I see you're using golang:1.15-alpine as base image for debug, but it will still have delve missing. I feel like we reverted back changes to this Dockerfile from my PR.

jpkrohling: This PR here should not revert that PR.

cmd/all-in-one/Dockerfile
```diff
@@ -25,43 +31,27 @@ EXPOSE 14250
...
 EXPOSE 12345
 
-COPY all-in-one-debug-linux-$TARGETARCH /go/bin/all-in-one-linux
-COPY sampling_strategies.json /etc/jaeger/
-
-VOLUME ["/tmp"]
 ENTRYPOINT ["/go/bin/dlv", "exec", "/go/bin/all-in-one-linux", "--headless", "--listen=:12345", "--api-version=2", "--accept-multiclient", "--log", "--"]
 CMD ["--sampling.strategies-file=/etc/jaeger/sampling_strategies.json"]
+
+FROM ${TARGET}
```
Ashmita152, line 57: I am zero knowledge of docker buildx, curious what is the meaning of have FROM at the end of Dockerfile ?

morlay: Yes. 

it is not new of buildx.

FROM could be from a stage name.

i learn this from istio Dockerfiles, which is used to switch default and distroless

cmd/all-in-one/Dockerfile
morlay: Sure. i added it.

We could setup multi arch build for base images next

scripts/travis/build-all-in-one-image.sh
aebirim: > @aebirim, for your change, you might just need to add a new tag here, as well as a `docker login` to all configured repositories.

np. @albertteoh @jpkrohling  I'll wait for this script to be merged before adding my quay change

CONTRIBUTING.md
```diff
@@ -173,3 +173,41 @@ Before merging a PR make sure:
...
+
+[OCI Images](https://github.com/opencontainers/image-spec/blob/master/spec.md) could be built and published by [buildx](https://github.com/docker/buildx), 
+it could be executed for local publish `all-in-one` images via:
+
+```shell
+TRAVIS_SECURE_ENV_VARS=true NAMESPACE=$(whoami) BRANCH=master ./scripts/travis/build-all-in-one-image.sh
```
jpkrohling, line 185: Why is this `TRAVIS_SECURE_ENV_VARS=true` needed? Isn't there a make target that could be used instead of the Travis script?

jpkrohling (commented):
@Ashmita152, could you please review and test this as well? We should have no regressions when compared to your previous work.

Ashmita152:
Sure I will give it a try today.

scripts/travis/build-all-in-one-image.sh
```diff
@@ -26,42 +29,89 @@ run_integration_test() {
...
-upload_to_docker $repo
-
-make build-all-in-one-debug GOOS=linux GOARCH=$GOARCH
-repo=jaegertracing/all-in-one-debug
-docker build -f cmd/all-in-one/Dockerfile \
-        --target debug \
-        --tag $repo:latest cmd/all-in-one \
-        --build-arg base_image=localhost/baseimg:1.0.0-alpine-3.12 \
-        --build-arg debug_image=localhost/debugimg:1.0.0-golang-1.15-alpine \
-        --build-arg TARGETARCH=$GOARCH
```
Ashmita152, line 59: @morlay I noticed all occurrences of TARGETARCH is removed from this script but I don't see the place it is passed  as argument to Dockerfile from this script. TARGETARCH is being used in all-in-one's Dockerfile to decide which binary to copy inside container. Is this some arg which buildx populates internally by default ?

scripts/travis/build-all-in-one-image.sh
```diff
@@ -7,6 +7,9 @@ BRANCH=${BRANCH:?'missing BRANCH env var'}
 # be overrided by passing architecture value to the script:
 # `GOARCH=<target arch> ./scripts/travis/build-all-in-one-image.sh`.
 GOARCH=${GOARCH:-$(go env GOARCH)}
+# local run with `TRAVIS_SECURE_ENV_VARS=true NAMESPACE=$(whoami) BRANCH=master ./scripts/travis/build-all-in-one-image.sh`
+NAMESPACE=${NAMESPACE:-jaegertracing}
+ARCHS="amd64 arm64 s390x"
```
Ashmita152, line 12: Does travis support building images for `s390x` architecture ? Doesn't that means it will require base image of this architecture for alpine too ?

scripts/travis/setup-docker-buildx.sh
```diff
@@ -0,0 +1,56 @@
...
+
+# enable buildx
+export DOCKER_CLI_EXPERIMENTAL=enabled
+
+# checkout buildx available
+docker buildx version
+
+# enabled qemu if needed
+if [[ ! $(docker buildx inspect default | grep Platforms) == *arm64* ]]; then
+  docker run --privileged --rm tonistiigi/binfmt --install all
```
Ashmita152, line 52: I noticed you're using `tonistiigi/binfmt` image. I guess this is because it has images for all architecture which we need. Can we use append the tag too `tonistiigi/binfmt:<tag>` Right now it is pulling latest tag and the owner updating the latest image may break Jaeger travis CI.

Ashmita152:
Hi @morlay looks great overall. I've left some minor comments. Would it be possible for you to share the travis run link where you tested your branch.

Ashmita152 (changes requested):

scripts/travis/build-all-in-one-image.sh
morlay: TARGETARCH and other platform args will be passed by `buildx`

https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope

morlay:
@Ashmita152  i tested on local by (need to setup buildx)

```shell
TRAVIS_SECURE_ENV_VARS=true NAMESPACE=$(whoami) BRANCH=v1.20.0 ./scripts/travis/build-all-in-one-image.sh
```

scripts/travis/build-all-in-one-image.sh
morlay: Yes. We could build base images by `buildx` in next step.

golang and alpine both support `s390x` too. so the base images of jaeger could support `s390x`

the magic is `tonistiigi/binfmt` who installs `qemu`

scripts/travis/setup-docker-buildx.sh
morlay: `tonistiigi/binfmt` is not provide the images we need for architecture.

it just install `qemu` as simulator for target arch. 
so we could run arm64 binary on amd64 host. (but it is a very slow way, and we don't use this for compiling)

the images `alpine:3.12`, you could see the arch list in docker hub.

image

This means docker hub store each image for each arch.

When we use `FROM alpine:3.12`, it will pull the matched arch image to local. 

like docker with `--platform` did.
```
docker pull --platform=linux/arm64 alpine:3.12
docker pull --platform=linux/amd64 alpine:3.12
docker pull --platform=linux/s390x alpine:3.12
```

With `buildx`,  the  `FROM alpine:3.12` in Dockerfile, pulled for each arch image too.
And in `RUN apt add git`, `apt` (arm64 binary) need to execute on matched arch (arm64) host.
This why we need `qemu`. with `qemu` we could run `apt` (arm64 binary) on arm64 host. 

And we don't need the tag.  it is a official tool in [buildx doc](https://github.com/docker/buildx#building-multi-platform-images). when this tool is stable, i think it will merge back to docker's namespace.

Ashmita152: I mean you are using `tonistiigi/binfmt` instead of official image `linuxkit/binfmt` Is there any particular reason for that ?

morlay: no special reason, just it is on README of buildx Which is widely used by github actions.

https://github.com/docker/buildx#building-multi-platform-images
image

morlay: May be the version of qemu.

`tonistiigi/binfmt` with qemu@5.0
`linuxkit/binfmt` with qemu@4.2

Ashmita152: Ahh I see, thank you @morlay that answers my question.

CONTRIBUTING.md
morlay: I don't want to change too much in this pr.

We could refactor this in future

Ashmita152:
I tested the all-in-one's Dockerfile changes locally, it worked for me like charm. Thank you @morlay for answering all my questions. I am not sure what will be the right way to test travis part.

morlay:
`./scripts/travis/build-all-in-one-image.sh` called in pr checks, only difference is not push to hub.

jpkrohling:
@yurishkuro, would you like to take a look? The only point that I had was about the usage of travis-specific env vars for local building/publishing, but if @morlay promises to work on that on a follow-up PR, I'm OK with that.

jpkrohling:
ping @yurishkuro

yurishkuro:
It seems to be that there are a lot of changes to the Dockerfiles that revert the previous changes for debug images.

morlay:
@yurishkuro it is not revert. 

Just added stages of base images  to support multi arch for `all-in-one` images, and enhance the build stages to avoid redundant configuration.

We could to do next for support multi arch for all component images.

* [ ] build & push base & debug muti images 
* [ ] refactor Dockerfile of all-in-one and other components.

This pr is since July. before debug images created.
I don't want change the scope for `all-in-one`.

If you couldn't like those changes. We could close this pr.
And re-open a new pr to setup from base images.

yurishkuro:
It most definitely reverts a bunch of stuff. I realize it is unfortunate that the two PRs overlapped in time, but the debug images PR did a lot of refactoring on the images to DRY them, and those changes are being reverted here (for example, all the copying of cert files, which is currently done only once in the base images).

I think we need to rethink how buildx should work with the new refactored setup, simply merging this old PR on top is probably not going to work.

Ashmita152:
I was thinking to pick work on multi-arch images for base images after this PR gets merged. I can pick it before if you think that will help.

morlay:
@Ashmita152 @yurishkuro 

if we want to setup from base images, and for all components images

Stages could be

Stage 0 - switch docker build to docker buildx build

- updates all docker cli 
- --platform
- Dockerfile，remove default value of platform args like TARGETARCH
- Makefile enhance and updates CI scripts

Base images not need to
publish docker.io.
Using local registery works well too.
But will be a extra setup step for local buld.

Stage 1 - Dockerfile enhance

- multi stage for base image switching

scripts/travis/build-all-in-one-image.sh
jpkrohling: This PR was closed in favor of #2611

PR #2327, opened 2020-07-04
fix null pointer in jaeger-spark-dependencies

moolen:
fixes jaegertracing/jaeger-operator#1113

I'm not sure if other fields have the same issue.

### DIFF
--- a/cmd/opentelemetry/app/exporter/elasticsearchexporter/esmodeltranslator/modeltranslator.go
+++ b/cmd/opentelemetry/app/exporter/elasticsearchexporter/esmodeltranslator/modeltranslator.go
@@ -34,8 +34,11 @@ const (
 )
 
 var (
-	errZeroTraceID = errors.New("traceID is zero")
-	errZeroSpanID  = errors.New("spanID is zero")
+	errZeroTraceID     = errors.New("traceID is zero")
+	errZeroSpanID      = errors.New("spanID is zero")
+	emptyLogList       = []dbmodel.Log{}
+	emptyReferenceList = []dbmodel.Reference{}
+	emptyTagList       = []dbmodel.KeyValue{}
 )
 
 // Translator configures translator
@@ -134,7 +137,7 @@ func toTime(nano pdata.TimestampUnixNano) time.Time {
 func references(links pdata.SpanLinkSlice, parentSpanID pdata.SpanID, traceID dbmodel.TraceID) ([]dbmodel.Reference, error) {
 	parentSpanIDSet := len(parentSpanID.Bytes()) != 0
 	if !parentSpanIDSet && links.Len() == 0 {
-		return nil, nil
+		return emptyReferenceList, nil
 	}
 
 	refsCount := links.Len()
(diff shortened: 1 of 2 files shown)


yurishkuro:
Is the expectation that any list must not be null, or just references? For example, `logs()` function also may return nil.

moolen:
It depends on the consumer, i guess. speaking for `jaeger-spark-dependencies`: `tags` and `process` must be non-nil (otherwise we'll see a null pointer exception).

see [SpanDeserializer.java](https://github.com/jaegertracing/spark-dependencies/blob/master/jaeger-spark-dependencies-elasticsearch/src/main/java/io/jaegertracing/spark/dependencies/elastic/json/SpanDeserializer.java#L58) / [ObjectMapper.java](https://github.com/FasterXML/jackson-databind/blob/master/src/main/java/com/fasterxml/jackson/databind/ObjectMapper.java#L1210).


While taking a look at #2295:
> OTEL data model is directly transformed to Jaeger ES model

I think writing `null` into `$storage` is a bug.

@pavolloffay you're much deeper into this, can you confirm this? If so i'll change `tags` and `process` accordingly.

cmd/opentelemetry/app/exporter/elasticsearchexporter/esmodeltranslator/modeltranslator.go
```diff
@@ -134,7 +134,7 @@ func toTime(nano pdata.TimestampUnixNano) time.Time {
 func references(links pdata.SpanLinkSlice, parentSpanID pdata.SpanID, traceID dbmodel.TraceID) ([]dbmodel.Reference, error) {
 	parentSpanIDSet := len(parentSpanID.Bytes()) != 0
 	if !parentSpanIDSet && links.Len() == 0 {
-		return nil, nil
+		return []dbmodel.Reference{}, nil
```
yurishkuro, line 137: we could avoid some memory allocations by using constants in these cases, like
```
const emptyReferenceList = []dbmodel.Reference{}
```
the most impact will be for logs, since it's quite rare to find spans without references or tags.

BTW, I wonder if there is another bug in the translator from OTEL that fails to create a CHILD_OF reference to the parent span ID, which is required in the Jaeger model since we don't store parentSpanID in the database.

moolen: Thanks for your feedback, i made changes accordingly.
I don't see where there could be a bug (...or i don't get it :smile: ). The `CHILD_OF` refs are tested properly as far as i can tell. It either creates a reference or it errors out.

cmd/opentelemetry/app/exporter/elasticsearchexporter/esmodeltranslator/modeltranslator.go
```diff
@@ -131,10 +131,12 @@ func toTime(nano pdata.TimestampUnixNano) time.Time {
 	return time.Unix(0, int64(nano)).UTC()
 }
 
+var emptyReferenceList = []dbmodel.Reference{}
```
yurishkuro, line 134: please group all vars at the top of the file (after constants, if any)
```
var (
    emptyReferenceList = []dbmodel.Reference{}
    emptyTagList = []dbmodel.KeyValue{}
    etc
)
```

Issue #2331, opened 2020-07-07
Buffer size for udp is too small


kbarukhov:
**Describe the bug**
Default buffer size for udp is too small. Because of that some udp-packets are dropped

**What troubleshooting steps did you try?**
We have quick fix for this issue. May be we should merge it?
https://github.com/ozonru/jaeger/blob/1.18-udp-buffer-size/cmd/agent/app/servers/thriftudp/transport.go#L90

yurishkuro:
+1, please do a PR. Could you also provide more context - your setup, under what conditions you notice data loss, how did you arrive to the current buffer size? What is the default buffer size? Is it OS-dependent?

It may make sense to expose this parameter as configurable.

kbarukhov:
With something like 1.2 million spans in second and about 5 Gbits total traffic we saw package dropping. Default buffer size is around 4 kB. Current size was found experimently) I'll test if these syscalls work on windows and do a PR

yurishkuro:
"1.2 million spans/sec" for one agent? Sweet.

mkabischev:
Noo, 1.2kk spans is total throughput for our setup. We have some instances with "heavy" services that send a lot of telemetry and that caused this problem.

PR #2336, opened 2020-07-09
Make UDP socket buffer size configurable

kbarukhov:
Buffer size for udp is too small #2331

Add ability to change udp buffer size via syscall, but it doesn't work on windows


pavolloffay:
@kbarukhov could you please sign off the commits? https://github.com/jaegertracing/jaeger/blob/master/CONTRIBUTING_GUIDELINES.md#missing-sign-offs

cmd/agent/app/flags.go
```diff
@@ -50,6 +51,7 @@ func AddFlags(flags *flag.FlagSet) {
 		flags.Int(prefix+suffixWorkers, defaultServerWorkers, "how many workers the processor should run")
 		flags.Int(prefix+suffixServerQueueSize, defaultQueueSize, "length of the queue for the UDP server")
 		flags.Int(prefix+suffixServerMaxPacketSize, defaultMaxPacketSize, "max packet size for the UDP server")
+		flags.Int(prefix+suffixServerBufferSize, defaultBufferSize, "buffer size for UDP packets")
```
pavolloffay, line 54: Could document units here?

cmd/agent/app/builder.go
```diff
@@ -39,6 +39,7 @@ import (
 const (
 	defaultQueueSize     = 1000
 	defaultMaxPacketSize = 65000
+	defaultBufferSize    = 4 * 1024
```
pavolloffay, line 42: this is declared in `transport.go` could we consume it from there?

mkabischev: I think we should omit default value. If it's not explicitly set then agent must use system value.

cmd/agent/app/builder.go
```diff
@@ -90,6 +90,7 @@ type ProcessorConfiguration struct {
 type ServerConfiguration struct {
 	QueueSize     int    `yaml:"queueSize"`
 	MaxPacketSize int    `yaml:"maxPacketSize"`
+	BufferSize    int    `yaml:"bufferSize"`
```
yurishkuro, line 93: I would prefer to call this SocketBufferSize, it's much more descriptive.

Similarly for the CLI flag.

cmd/agent/app/servers/thriftudp/transport.go
```diff
@@ -71,7 +73,7 @@ func createClient(destAddr, locAddr *net.UDPAddr) (*TUDPTransport, error) {
 // It will listen for incoming udp packets on the specified host/port
 // Example:
 // 	trans, err := thriftudp.NewTUDPClientTransport("localhost:9001")
-func NewTUDPServerTransport(hostPort string) (*TUDPTransport, error) {
+func NewTUDPServerTransport(hostPort string, bufferSize int) (*TUDPTransport, error) {
```
yurishkuro, line 76: instead of changing the API (which has a knock off effect on a lot of code), could we just add a SetSocketBufferSize method to TUDPTransport and call it from where we need it?  Especially since it's optional.

cmd/agent/app/builder_test.go
```diff
@@ -125,7 +125,7 @@ func TestBuilderWithProcessorErrors(t *testing.T) {
 	}{
```
yurishkuro, line 125: Please add a test case to TestBuilderFromConfig where the new parameter is not 0. This should take care of the low code coverage for the diff.

yurishkuro:
hm, the problem is that we calculate code coverage per package only, so having tests in another package that exercise the SetSocketBufferSize in another package does not improve code coverage. Is there  something we can do directly in thriftudp package to exercise that path?

kbarukhov:
> hm, the problem is that we calculate code coverage per package only, so having tests in another package that exercise the SetSocketBufferSize in another package does not improve code coverage. Is there something we can do directly in thriftudp package to exercise that path?

We can do a test for transport, call the SetSocketBufferSize with appropriate size and with the value bigger than maximum. And check if no error and an error occured?

yurishkuro:
sgtm

yurishkuro:
The build is failing on go.sum changing from checked-in version. Do you really need to change go.mod/go.sum in the PR?

Issue #2249, opened 2020-05-17
Support TLS and mTLS in collector and query HTTP servers


yurishkuro:
Add support for TLS and mTLS to collector and query HTTP endpoints, similar to what's already supported in the collector's gRPC endpoints.

rjs211:
Can I be assigned to this issue please?  Thanks.

pavolloffay:
Please take it.

davidcmitchell:
Hey @rjs211 are you also working on the Client to Collector TLS support as well?

rjs211:
@davidcmitchell Thank you.  I havent started working on Client to collector side of things yet.  If it isnt of great urgency, i can start working on it from Monday. Would that be alright?

rjs211:
@davidcmitchell Thank you.  I havent started working on Client to collector side of things yet.  If it isnt of great urgency, i can start working on it from Monday. Would that be alright?

davidcmitchell:
That’s great, thanks!

davidcmitchell:
@yurishkuro @rjs211 I know #2798 is to support TLS for the HTTP endpoint on the Collector server. Is there a plan to support mTLS for Client -> Collector? Happy to assist if you need additional help.

Issue #2329, opened 2020-07-04
Flaky test TestTBufferedServer/processed (metricstest.go)


yurishkuro:
```
=== RUN   TestTBufferedServer/processed
    TestTBufferedServer/processed: metricstest.go:49: 
        	Error Trace:	metricstest.go:49
        	            				metricstest.go:43
        	            				tbuffered_server_test.go:108
        	            				tbuffered_server_test.go:37
        	Error:      	Not equal: 
        	            	expected: int(0)
        	            	actual  : int64(1)
        	Test:       	TestTBufferedServer/processed
        	Messages:   	expected metric name: thrift.udp.server.queue_size, tags: map[]
=== RUN   TestTBufferedServer/dropped
--- FAIL: TestTBufferedServer (0.02s)
    --- FAIL: TestTBufferedServer/processed (0.01s)
```

PR #2337, opened 2020-07-09
TLS  support for HTTP API of Query server

rjs211:
- Adds TLS support for HTTP endpoint of query server
- solves part of issue #2249   
- solves issue #2338  
- Similar to PR #2297 .  
- independent TLS flags are exposed for gRPC and HTTP endpoints, enabling the user to provide different set of key, cert, CA-Cert , etc for each communication channal.

- provides the option of enabling TLS/mTLS in none, either one or both of HTTP and gRPC endpoints.
- forces the user to use dedicated HTTP and gRPC ports if TLS is enabled in any of the endpoints.


cmd/query/app/server_test.go
```diff
@@ -53,7 +53,20 @@ func TestCreateTLSServerError(t *testing.T) {
...
+
+func TestCreateTLSHttpServerError(t *testing.T) {
+	tlsCfg := tlscfg.Options{
+		Enabled:      true,
+		CertPath:     "invalid/path",
+		KeyPath:      "invalid/path",
+		ClientCAPath: "invalid/path",
+	}
+
+	_, err := NewServer(zap.NewNop(), &querysvc.QueryService{},
```
jpkrohling, line 68: Could you also add a positive test, to make sure that the typical path for `queryOptions.TLSHttp.Enabled` is tested?

jpkrohling:
We may want to hold on merging this before #2338 is sorted out.

cmd/query/app/server_test.go
rjs211: I am sorry,  I am new to this.  Can you show me a similar example (preferably one that involves dummy certificat ,key and other TLSConfig)?   Thanks so much.

cmd/query/app/flags.go
```diff
@@ -66,8 +72,10 @@ type QueryOptions struct {
 	UIConfig string
 	// BearerTokenPropagation activate/deactivate bearer token propagation to storage
 	BearerTokenPropagation bool
-	// TLS configures secure transport
-	TLS tlscfg.Options
+	// TLS configures secure transport (Consumer to Query service gRPC APO)
```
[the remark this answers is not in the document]
yurishkuro: also `s/TLS configures/TLSGRPC configures/` - the description starts with the field name

yurishkuro:
Please describe if/how you performed the real integration test that uses certificates. Would be nice to add those to the suite of integration tests in the CI.

cmd/query/app/server_test.go
jpkrohling: You can generate a cert locally and place in a test data directory. You can take a look at these examples: 

- ./pkg/config/tlscfg/testdata
- ./cmd/agent/app/reporter/grpc/testdata

rjs211: `httpServer.serveTLS(...)`  that i used from [here](https://golang.org/pkg/net/http/#ServeTLS)  requires  " If the certificate is signed by a certificate authority, the certFile should be the concatenation of the server's certificate, any intermediates, and the CA's certificate. "    Is this a fair assumption?  It is not the case in the testData provided.   The other alternative is to use tls.listener that wraps around the current listener `httpListener`.  This can use TLSConfig directly.

jpkrohling: I'm pretty sure the certs from the test data are self-signed. For the tests, this is OK and you can use the same CA on both the client and server, so that they both have the same chain of trust.

rjs211: What server name should i use for generating keys?  i have used "localhost" to generate the keys and all requests are made to "localhost".   Can you please tell me the actual server name?  Where can i find the convention ? Thanks.

yurishkuro: we already have test certificates generated in `./pkg/config/tlscfg/testdata/`

rjs211: And what is the server name for those certificates?  TLS client checks the server name...  Else there could be a man-in-the-middle attack.   `./cmd/agent/app/reporter/grpc/testdata` could be a better set of certificates (client and server are separate certificates)  but that has a fixed server name of `server.jaeger.io`.  p.s.  I am checking TLS and mTLS so need both client and server certifs.

yurishkuro: I am not sure. We also have another set of test certificates under `./cmd/agent/app/reporter/grpc/testdata`, and they are used for the actual client/server test in `TestProxyClientTLS -- ./cmd/agent/app/reporter/grpc/builder_test.go`.

@tcolgate do you remember how you generated those certificates? We should have documented it somewhere, either in testdata/README or have a Makefile target.

tcolgate:
I used mkcert I think.

cmd/query/app/server_test.go
jpkrohling: About the server name: with gRPC, you can specify what's the expected server name in the cert, like what we do in the flag `--reporter.grpc.tls.server-name`

rjs211: Yes,  but @yurishkuro agreed on Gitter that it would be better to use one set of certificates and one generic servername associated with the certificate (the servername associated with the server's certificate is tested by the client).  I am waiting for someone to tell me what generic name to use so that i can create a client and server set of Certificates and keys with a common CA.

p.s. the expected server name of  `--reporter.grpc.tls.server-name` is actually `server.jaeger.io`  I traced the add flags call from the builder to the actual addition of flags in `tlscfg` and couldnt find a line of thought as to why it was named as `server.jaeger.io` .  I guess some design choise was made that  I am unable to figure out.  If possible,  can you please explain the naming so that i could follow the same convention in the future?  Thanks in advance.

jpkrohling: When doing an HTTP request, the server will send a TLS cert which includes the subject (ie, who can use the cert). The client then compares the hostname in the HTTP request with the subject in the TLS cert. The flag `server-name` tells gRPC to use the flag's value instead of the value from the HTTP request. This is useful when you make a request to `http://localhost/path` but you know the cert was issued for `server.jaeger.io`.

All in all, `server.jaeger.io` is just a name (an unfortunate one, if you ask me), and it represents a host name including the base domain name (`jaeger.io`, which is registered to an entity foreign to this project). It would probably have been better to name it something like `jaeger-collector.example.com` (see [RFC 2606](https://tools.ietf.org/html/rfc2606) on why `example.com` should be used).

cmd/query/app/testdata/README.md
```diff
@@ -0,0 +1,35 @@
...
+openssl x509 -req -in example-client.csr -CA example-CA-cert.pem -CAkey example-CA-key.pem -CAcreateserial -out example-client-cert.pem
+
+# cleanup
+rm example-CA-cert.srl
+rm example-client.csr
+rm example-server.csr
+```
+
+The server name (common name in the server certificate ) is `example.com`  . (in accordance to [RFC 2006](https://tools.ietf.org/html/rfc2606) )
+The common name of the client is `example-client` (never actually unused).
```
yurishkuro, line 34: did you mean "never actually used"?

cmd/query/app/testdata/README.md
```diff
@@ -0,0 +1,35 @@
+# Example Certificate Authority and Certificate creation for testing
```
yurishkuro, line 1: Thanks for writing this!

cmd/query/app/testdata/example-CA-cert.pem
```diff
@@ -0,0 +1,19 @@
+-----BEGIN CERTIFICATE-----
```
yurishkuro, line 1: is it necessary to create a new set of files? can we just update the already existing files and load them all from a single location?

rjs211: Sure.  That would be great.  I'll do that as a separate PR perhaps.  I ran a bash command to find all the .pem files in the project.  That can be found here:  https://pastebin.ubuntu.com/p/gJ5DmtYZsF/   .  It would be great if you could tell me the ones that could be replaced by the files from the central location (which i think is `./pkg/config/tlscfg/testdata`).  I am not well versed on the purpose of each one.   

1.  Please tell me which files are to be replaced by central certificates (As of now,  I am sure that reporter, query and tlscfg 's testdata can be replaced.)
1.  Please also tell me the path/location where i need to store the keys (   `./pkg/config/tlscfg/testdata`  makes it seem like it is sepcific to tlscfg module.  Which is true for now. so shall i go along with the changes? )     

Thanks.

cmd/query/app/testdata/README.md
rjs211: my bad.  Sorry. will change in the next push.

cmd/query/app/testdata/example-CA-cert.pem
yurishkuro: I think `./pkg/config/tlscfg/testdata` is the ideal location, since it's the shared TLS-related code. Do you want to consolidate pem files in a different PR first and then come back to this one?

yurishkuro: Regarding your list, there are only two locations I see there, `./pkg/config/tlscfg/testdata` and `./cmd/agent/app/reporter/grpc/testdata`. The UI paths are external dependencies.

rjs211: Okay,  Thanks.  Ill make the changes and create another pull request.  Once that is merged,  ill delete thefiles in the query directory. 

Also, I am facing https://github.com/jaegertracing/jaeger/issues/2329 issue during CI build.  I have only made changes to the query package,  This shouldn't have happened.  Any workarounds / details that i missed?  Thanks,

yurishkuro: simply re-running the build usually helps, but we should really fix that one.

rjs211: Is there anyway to trigger rerunning the build without a new push?  Sorry.  I am new to CI as well.  Is an empty commit i.e.  `git commit --allow-empty` the way to go?

rjs211: Closing and opening PR works too.  Thanks.

cmd/query/app/server_test.go
```diff
@@ -53,10 +62,314 @@ func TestCreateTLSServerError(t *testing.T) {
...
+				tlsCfg, _ := serverOptions.TLSHTTP.Config()
+				tlsCfg.Rand = rand.Reader
+				tlsHTTPListener := tls.NewListener(httpListener, tlsCfg)
+				go func() {
+					// err
+					_ = httpServer.Serve(tlsHTTPListener)
+
+				}()
+
+				time.Sleep(10 * time.Millisecond) // wait for server to start serving
```
yurishkuro, line 194: below code seems identical to code in the else part

cmd/query/app/server.go
```diff
@@ -146,13 +161,29 @@ func (s *Server) Start() error {
 
 	go func() {
 		s.logger.Info("Starting HTTP server", zap.Int("port", tcpPort), zap.String("addr", s.queryOptions.HostPort))
+		var err error
```
yurishkuro, line 164: this looks like it would be better as a helper function

cmd/query/app/server_test.go
```diff
@@ -53,10 +62,314 @@ func TestCreateTLSServerError(t *testing.T) {
...
+
+func TestTLSHTTPServer(t *testing.T) {
+	tests := []struct {
+		name             string
+		serverTLS        tlscfg.Options
+		clientTLS        tlscfg.Options
+		expectError      bool
+		expectServerFail bool
+	}{
+		// {
```
yurishkuro, line 90: remember to remove if not needed

cmd/query/app/server_test.go
```diff
@@ -53,10 +62,314 @@ func TestCreateTLSServerError(t *testing.T) {
...
+				require.Error(t, err)
+			} else {
+				require.NoError(t, err)
+			}
+
+			if serverOptions.TLSHTTP.Enabled {
+
+				tlsCfg, _ := serverOptions.TLSHTTP.Config()
+				tlsCfg.Rand = rand.Reader
+				tlsHTTPListener := tls.NewListener(httpListener, tlsCfg)
```
yurishkuro, line 187: this should use the actual business logic, to test it, not replicate it here

cmd/query/app/server_test.go
rjs211: It actually isn't.  the class of functions for TLS works on `*Conn` rather than `Conn`.   i.e.  the syntax is `(*Conn) Close()`  not `(Conn Close())`
The net package works on `Conn`.  i.e.  the syntax is `(Conn) Close()`  

even though TLS's Conn implements net.Conn,  we cannot overload in this case and use the same functions.  so,  hence If and else are necessary.  The variables are similar for better readability.

cmd/query/app/server_test.go
rjs211: I left that as a template for adding future tests if necessary.  Some fields are relevant for client and not for server and vice-versa.  Should i go ahead and remove it?  thanks.

cmd/query/app/server_test.go
rjs211: Sorry,  I dont fully understand what you mean.  

This was the relevant minimum functionality / logic that i derive form the tesing of HTTP Server.  I need to do this  because mTLS with a bad certificate isn't verifyable with `TLS.Dial` (i.e.  no error is given. The error is given in the first read only.   [This](https://github.com/golang/go/issues/28779#issuecomment-442694636) could shed some light on the issue.  So in order to read something that would be accepted with no errors,  I resorted to the logic defined in https://github.com/jaegertracing/jaeger/blob/master/cmd/query/app/http_handler_test.go .  

Could you please elaborate a little bit on the error and what is expected?  thanks.

cmd/query/app/server_test.go
yurishkuro: the struct definition is sufficient, no need for template

cmd/query/app/server.go
```diff
@@ -114,7 +129,26 @@ func createHTTPServer(querySvc *querysvc.QueryService, queryOpts *QueryOptions,
...
+	}, nil
+}
+
+func (s *Server) serveHTTP(httpListener net.Listener) error {
+	var err error
+	if s.queryOptions.TLSHTTP.Enabled {
+		tlsCfg, err1 := s.queryOptions.TLSHTTP.Config()
+		if err1 == nil {
+			tlsCfg.Rand = rand.Reader
+			tlsHTTPListener := tls.NewListener(httpListener, tlsCfg)
```
yurishkuro, line 141: is this going to be a problem due to #2338 ?

cmd/query/app/server_test.go
yurishkuro: I mean that it is largely repeating what `func (s *Server) serveHTTP(httpListener net.Listener)` function does, which is production code, and we should try to use that rather than repeating it in the test.

cmd/query/app/server_test.go
yurishkuro: the only function you use from `conn` is Close, so you could have if-statement creating the `conn` object via tls or net and assign it to something like `io.Closable`, and not repeating the rest of the statements.

cmd/query/app/server.go
rjs211: Here i am only testing the HTTP server component separately. i.e.  to check the `createHTTPServer`method and to check if TLSEnabled actually works .  The problem of #2338 occurs only when we want to test the queryServer which has both HTTPServer and GRPC server.

cmd/query/app/server_test.go
```diff
@@ -53,10 +62,314 @@ func TestCreateTLSServerError(t *testing.T) {
...
+				go func() {
+					// err
+					_ = httpServer.Serve(tlsHTTPListener)
```
yurishkuro, line 190: let's try to assert no error, with exception of ServerClosed

cmd/query/app/server_test.go
rjs211: Aah.  okay.  I wrote this method to test `func createHTTPServer(querySvc *querysvc.QueryService, queryOpts *QueryOptions, tracer opentracing.Tracer, logger *zap.Logger) (*http.Server, error)`  function without the `Server`.  Once #2338 is resolved,  only then i can use `Server` directly with TLS.  Ill get right on it.

cmd/query/app/server_test.go
yurishkuro, line 194: > time.Sleep(10 * time.Millisecond) // wait for server to start serving

sleeps like this usually lead to flaky/unstable tests in CI. If you check other unit tests in this project, the usual approach is to check for condition periodically until it is satisfied or a much larger time period has passed, e.g.
```
for i:=0; i < 100; i++ {
    if conditionSatisfied() {
        break
    }
    sleep 10ms
}
assert condition is satisfied
```

this loop does not increase unit test run time if CPU resources are available, but in CI environment, where CPU is shared and busy, it can wait long enough (1s here) to succeed on subsequent iterations.

cmd/query/app/server_test.go
rjs211: actually i did that before.  but both this subroutine and the outer subroutine both use `t *testing.T`  which lead to race conditions.  Hence i ignored it.  the best we could do is log the message.  Assertion wouldnt work.

cmd/query/app/testdata/example-CA-cert.pem
jpkrohling: You can try `git commit --amend`, it should generate a new hash.

cmd/query/app/server_test.go
rjs211: I used a variable that can store the `X.Close()` method and invoked it outside.

cmd/query/app/flags.go
```diff
@@ -46,11 +46,13 @@ const (
 	queryTokenPropagation   = "query.bearer-token-propagation"
 	queryAdditionalHeaders  = "query.additional-headers"
 	queryMaxClockSkewAdjust = "query.max-clock-skew-adjustment"
+	queryHTTPTLSEnabled     = "query.http.tls.enabled"
+	queryGRPCTLSEnabled     = "query.gprc.tls.enabled"
```
yurishkuro, line 50: why do we need these? `enabled` is already a standard flag in the TLS config https://github.com/jaegertracing/jaeger/blob/master/pkg/config/tlscfg/flags.go#L26

rjs211: We need two one for GRPC one for HTTP.  but even if either one of them are enabled, we need only one set of certificates and keys.  So,  I used two hardCoded flags for each and used `showEnabled:false ` for the common certificate configuration.

rjs211: using enabled of `tlsCfg` would also force two sets of certificate and key.  https://github.com/jaegertracing/jaeger/issues/2338#issuecomment-659955372

yurishkuro: I don't follow. If you want to share certificates and keys between two servers, then shouldn't we have only one `TLS tlscfg.Options` field in the options struct, as we had before? Then it would make sense to have to extra `.enabled` flags to allow switching TLS on each server separately.

rjs211: Sure. I'll do that.

Does the project have any convention for multi-vlaued flags?  Because clientCA of GRPC and HTTP server might be different.  Currently we still have this problem.

rjs211: The user must concatenate all certificates int  a single certificate file.  I guess it would work. Thanks.

rjs211:
If this is okay,  I'll start working on the collector HTTP then?

cmd/query/app/flags.go
tcolgate: FWIW, I'd say that, atleast  when using mutual tls, it's unlikely you'll want the same server certs and CA for https and grpc. The grpc services will generally have certs issued from a common infrastructure CA controlled by the user, whereas user facing https will probably want a cert signed by a public CA, recognized by browsers.

rjs211: Then I think we would need two separate ports for the two different servers.  I am unaware how to use a single port with [cmux](https://godoc.org/github.com/soheilhy/cmux) with two different TLS certificates.

jpkrohling: I'd love to be proven wrong, but I don't think it's even possible. In order to determine the payload type, the headers have to be read, and this can only be done after the TLS handshake. Without knowing which cert to serve, there can't be a TLS handshake.

tcolgate: I'm sure it's probably possible to use the regular TLS cert negotiation selection with cmux, but on the whole I'm really against the whole "do it all on one port" thing, it gets very confusing when debugging.
FWIW, SNI negotiation occurs before the TLS handshake, so you can at least know which cert the client wanted, but the client does have control over that, so it's all a bit dubious at that point.
We're using grpc mtls betweem agents in local clusters to a single remote collector.

rjs211: @jpkrohling  Agreed. So our options are:

1.  Either we have to use one certificate where the server uses only one port for both HTTP and gRPC endpoints.
1.  Use separate port for http and another port fro gRPC.  then,  each will have different listeners and we can use two set of certificates generated by two diffrent CAs.

jpkrohling: @tcolgate given that Jaeger 2.x will be based on OpenTelemetry Collector, you might want to bring your concerns to this issue as well: open-telemetry/opentelemetry-collector#1256. 

This PR here is about the query server, which has been released already with both protocols on the same port. Splitting it now would cause current clients to break, so, I'm afraid we don't have much choice at this point.

tcolgate: @jpkrohling I did consider that. I added the mutual TLS support to jaeger and thanos, and then looked at doing it for otel, but cmux made it awkward, and debugging it did my brain in. (I disliked the cmux setup before that, but this was the first time I had to actually modify a project using it, that wasn't just ripping it out).
  Hopefully someone with more time an patience will do the otel work before I have to look at it again.

tcolgate: @jpkrohling I've added a couple of comments to that bug. I hadn't realised they'd reverted the cmux usage. I failed to get mutual TLS working with cmux, I can't recall the exact details of why, but it looks like the bug that caused them to revert might be the reason.
  If the grpc service only supports functionality that is provided by the UI, then having the same CA/certs is probably reasonable, assuming none of the other jaeger components need grpc to talk to it (other than maybe a cli query, which would be roughly equivalent to browser usage, so could just use a system cert pack).

jpkrohling: @yurishkuro can confirm, but my understanding is that the query gRPC endpoint is (or is going to be) *the* supported way to query the backend. The first client is the UI, but external clients are also expected to exist, so, mTLS is certainly something we'd want to support there.

jpkrohling:
@rjs211 is the code ready for review? Apparently, there are some left overs in the code, and the current state might not reflect the latest discussions.

It would also be good to edit the original PR description to include the final design for this. This will help make it clear what the intentions are.

rjs211:
@jpkrohling  I think it is ready for review.  I added comments for int cmux logic in the code. Hopefully it is understandable.

cmd/query/app/server.go
```diff
@@ -114,16 +125,81 @@ func createHTTPServer(querySvc *querysvc.QueryService, queryOpts *QueryOptions,
...
+	s.conn = conn
+
+	if err != nil {
+		return nil, nil, nil, err
+	}
+	if s.queryOptions.HTTPTLSEnabled != s.queryOptions.GRPCTLSEnabled { // exactly one of HTTP or GRPC has TLS enabled
+		cmux1 = cmux.New(conn)
+
+		if !s.queryOptions.HTTPTLSEnabled { // HTTPTLS disabled GRPCTLS enabled => match http request
+			httpListener = cmux1.Match(cmux.HTTP1Fast())
```
jpkrohling, line 156: If I'm reading this right, using a HTTP2 client will then get `cmux` to serve the request using TLS.

Honestly, I'm starting to think that `cmux` isn't worth the complexity it brings. Debugging an issue around this code is probably going to be a nightmare...

cmd/query/app/server.go
```diff
@@ -114,16 +125,81 @@ func createHTTPServer(querySvc *querysvc.QueryService, queryOpts *QueryOptions,
...
+		} else { // HTTPTLS enabled GRPCTLS disabled
+			grpcListener = cmux1.MatchWithWriters( // HTTPTLS disabled GRPCTLS enabled => match GRPC request
+				cmux.HTTP2MatchHeaderFieldSendSettings("content-type", "application/grpc"),
+				cmux.HTTP2MatchHeaderFieldSendSettings("content-type", "application/grpc+proto"),
+			)
+		}
+		tlsListener := cmux1.Match(cmux.Any()) // unmatched request are routed to TLS Listener
+		tlsListener, err = getTLSListener(tlsListener, s.queryOptions.TLS)
+		if err != nil {
+			return nil, nil, nil, err
```
jpkrohling, line 166: I guess it's worth returning a struct already :)

cmd/query/app/server.go
```diff
@@ -135,24 +211,17 @@ func (s *Server) Start() error {
...
-
-	grpcListener := cmuxServer.MatchWithWriters(
-		cmux.HTTP2MatchHeaderFieldSendSettings("content-type", "application/grpc"),
-		cmux.HTTP2MatchHeaderFieldSendSettings("content-type", "application/grpc+proto"),
-	)
-	httpListener := cmuxServer.Match(cmux.Any())
-
 	go func() {
 		s.logger.Info("Starting HTTP server", zap.Int("port", tcpPort), zap.String("addr", s.queryOptions.HostPort))
+		// s.serveHTTP(httpListener);
```
jpkrohling, line 216: left over?

pkg/config/tlscfg/flags.go
```diff
@@ -70,8 +70,10 @@ func (c ServerFlagsConfig) AddFlags(flags *flag.FlagSet) {
 	}
 	flags.String(c.Prefix+tlsCert, "", "Path to a TLS Certificate file, used to identify this server to clients")
 	flags.String(c.Prefix+tlsKey, "", "Path to a TLS Private Key file, used to identify this server to clients")
-	flags.String(c.Prefix+tlsClientCA, "", "Path to a TLS CA (Certification Authority) file used to verify certificates presented by clients (if unset, all clients are permitted)")
-	flags.String(c.Prefix+tlsClientCAOld, "", "(deprecated) see --"+c.Prefix+tlsClientCA)
+	if c.ShowClientCA {
```
jpkrohling, line 73: Is this related to this PR? Sounds like a reasonable, yet unrelated, change.

yurishkuro:
@jpkrohling your proposal sounds ok, but to maintain the old behavior with a single port we'd still need to use cmux, right?

Doing that split of http/grpc is probably better in a different PR, before this one.

jpkrohling:
> to maintain the old behavior with a single port we'd still need to use cmux, right?

Yes, which seems to work fine for non-TLS connections. People who are experiencing performance issues might want to try splitting the protocols, by setting an explicit `--query.grpc-server.host-port`.

> Doing that split of http/grpc is probably better in a different PR, before this one.

Agree. @rjs211, would you like to work on that one as well?

tcolgate:
cockroachdb found that they couldn't share their UI and grpc on a single port due to gprc client behaviour in the end, which might be a more appropriate reason to avoid this (something to do with the behaviour when sending headers causing the cmux probing to lock up).
It's mention briefly at the end here:
https://www.cockroachlabs.com/blog/a-tale-of-two-ports/

cmd/query/app/server.go
```diff
@@ -62,18 +63,27 @@ func NewServer(logger *zap.Logger, querySvc *querysvc.QueryService, options *Que
 		return nil, err
 	}
 
+	if (options.TLSHTTP.Enabled || options.TLSGRPC.Enabled) && (grpcPort == httpPort) {
+		return nil, errors.New("Server with TLS enabled can not use same host ports for gRPC and HTTP.  Use dedicated HTTP and gRPC host ports instead")
```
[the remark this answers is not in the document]
jpkrohling: The error message should not be capitalized, and you should probably externalize this with a var at the top of the source file like this:

https://github.com/jaegertracing/jaeger/blob/df2582d5331f8b0931a88e0a7c4c32ef788561f0/cmd/opentelemetry/app/internal/reader/es/esspanreader/validate_query.go#L23-L29

cmd/query/app/server_test.go
```diff
@@ -55,10 +77,473 @@ func TestCreateTLSServerError(t *testing.T) {
...
+	}
+
+	require.NoError(t, err)
+	return &grpcClient{
+		QueryServiceClient: api_v2.NewQueryServiceClient(conn),
+		conn:               conn,
+	}
+}
+
+func TestServerGRPCTLS(t *testing.T) {
```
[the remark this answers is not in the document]
jpkrohling: Yes, remove one of them (the less specialized), but leave a comment saying something like: `this is a cross test for the "dedicated ports" use case`

cmd/query/app/flags.go
```diff
@@ -101,11 +111,13 @@ func (qOpts *QueryOptions) InitPortsConfigFromViper(v *viper.Viper, logger *zap.
...
 	qOpts.HostPort = ports.GetAddressFromCLIOptions(v.GetInt(queryPort), v.GetString(queryHostPort))
 
-	qOpts.TLS = tlsFlagsConfig.InitFromViper(v)
+	qOpts.TLSGRPC = tlsGRPCFlagsConfig.InitFromViper(v)
+	qOpts.TLSHTTP = tlsHTTPFlagsConfig.InitFromViper(v)
 
-	// query.host-port is not defined and at least one of query.grpc-server.host-port or query.http-server.host-port is defined.
+	// If either GRPC or HTTP servers use TLS, use dedicated ports.
+	// else-if query.host-port is not defined and at least one of query.grpc-server.host-port or query.http-server.host-port is defined.
 	// User intends to use separate GRPC and HTTP host:port flags
```
jpkrohling, line 119: These four lines of comment need to be rewritten, they are hard to grasp as they are right now. Also, the `if` statement below needs a refactoring: I'd rather have nested or seemingly duplicate code than have a complex statement. Like:

```go
if qOpts.TLSGRPC.Enabled || qOpts.TLSHTTP.Enabled {
    return qOpts
}

if !(v.IsSet(queryHostPort) || v.IsSet(queryPort)) && (v.IsSet(queryHTTPHostPort) || v.IsSet(queryGRPCHostPort)) {
    return qOpts
}
```

jpkrohling (commented):
Please, decompose the if statement to make it easier to read. You might also want to set parts of the statement as vars, like: 

```go
legacyHostPortOptionSet = viper.IsSet(...) || viper.IsSet(...)

if !legacyHostPortOptionSet && ... {
...
}
```

I also need to run a manual test, but other than that, the code looks sane.

jpkrohling:
@rjs211 are you still interested in working on this?

jpkrohling:
When starting the query without TLS cert information for the HTTP port, the server is still started, but it's in an unusable state. Is that on purpose? Or should we fail to start?

```
$ SPAN_STORAGE_TYPE=memory go run ./cmd/query --query.grpc.tls.enabled=true --query.grpc.tls.cert=/home/jpkroehling/Projects/src/github.com/open-telemetry/opentelemetry-collector/local/self-signed.pem --query.grpc.tls.key=/home/jpkroehling/Projects/src/github.com/open-telemetry/opentelemetry-collector/local/self-signed-key.pem --query.http.tls.enabled=true
2020/10/19 12:11:00 maxprocs: Leaving GOMAXPROCS=12: CPU quota undefined
{"level":"info","ts":1603102260.4300442,"caller":"flags/service.go:117","msg":"Mounting metrics handler on admin server","route":"/metrics"}
{"level":"info","ts":1603102260.4300754,"caller":"flags/service.go:123","msg":"Mounting expvar handler on admin server","route":"/debug/vars"}
{"level":"info","ts":1603102260.4301715,"caller":"flags/admin.go:121","msg":"Mounting health check on admin server","route":"/"}
{"level":"info","ts":1603102260.4302065,"caller":"flags/admin.go:127","msg":"Starting admin HTTP server","http-addr":":16687"}
{"level":"info","ts":1603102260.4302154,"caller":"flags/admin.go:113","msg":"Admin server started","http.host-port":"[::]:16687","health-status":"unavailable"}
{"level":"info","ts":1603102260.4327462,"caller":"memory/factory.go:61","msg":"Memory storage initialized","configuration":{"MaxTraces":0}}
{"level":"info","ts":1603102260.432959,"caller":"querysvc/query_service.go:137","msg":"Archive storage not created","reason":"archive storage not supported"}
{"level":"info","ts":1603102260.4329696,"caller":"app/flags.go:159","msg":"Archive storage not initialized"}
{"level":"info","ts":1603102260.4508872,"caller":"app/server.go:175","msg":"Query server started"}
{"level":"info","ts":1603102260.4509492,"caller":"healthcheck/handler.go:128","msg":"Health Check state change","status":"ready"}
{"level":"info","ts":1603102260.4509723,"caller":"app/server.go:237","msg":"Starting HTTP server","port":16686,"addr":":16686"}
{"level":"info","ts":1603102260.4510095,"caller":"app/server.go:256","msg":"Starting GRPC server","port":16685,"addr":":16685"}
{"level":"error","ts":1603102260.4510472,"caller":"app/server.go:248","msg":"Could not start HTTP server","error":"open : no such file or directory","stacktrace":"github.com/jaegertracing/jaeger/cmd/query/app.(*Server).Start.func1\n\t/home/jpkroehling/Projects/src/github.com/jaegertracing/jaeger/cmd/query/app/server.go:248"}
{"level":"info","ts":1603102260.451137,"caller":"healthcheck/handler.go:128","msg":"Health Check state change","status":"unavailable"}
```

Other than that, looks good to me.

rjs211:
That is the old behaviour as well, incase if any error is encountered during server-start.  I didn't make any changes to that.  Should I ?

jpkrohling:
If that's the current behavior, it's OK to keep it for now. We can evaluate a new behavior in a follow-up PR.

jpkrohling (approved):

jpkrohling (dismissed):
LGTM. I'll leave it open for a day, to see if other maintainers want to review it as well.

cmd/query/app/flags.go
```diff
@@ -101,10 +111,16 @@ func (qOpts *QueryOptions) InitPortsConfigFromViper(v *viper.Viper, logger *zap.
 	qOpts.GRPCHostPort = v.GetString(queryGRPCHostPort)
 	qOpts.HostPort = ports.GetAddressFromCLIOptions(v.GetInt(queryPort), v.GetString(queryHostPort))
 
-	qOpts.TLS = tlsFlagsConfig.InitFromViper(v)
+	qOpts.TLSGRPC = tlsGRPCFlagsConfig.InitFromViper(v)
+	qOpts.TLSHTTP = tlsHTTPFlagsConfig.InitFromViper(v)
 
-	// query.host-port is not defined and at least one of query.grpc-server.host-port or query.http-server.host-port is defined.
-	// User intends to use separate GRPC and HTTP host:port flags
+	// If either GRPC or HTTP servers use TLS, use dedicated ports.
```
yurishkuro, line 117: below is a set of pretty complicated conditions, could we have a data driven unit test that illustrates how the various combinations are supposed to be treated?

cmd/query/app/server.go
```diff
@@ -62,18 +67,27 @@ func NewServer(logger *zap.Logger, querySvc *querysvc.QueryService, options *Que
 		return nil, err
 	}
 
+	if (options.TLSHTTP.Enabled || options.TLSGRPC.Enabled) && (grpcPort == httpPort) {
+		return nil, errSharedTLSPort
```
yurishkuro, line 71: errSharedTLSPort is not reused or needs to be pre-allocated for performance. Please inline it.

cmd/query/app/server.go
```diff
@@ -126,9 +142,21 @@ func createHTTPServer(querySvc *querysvc.QueryService, queryOpts *QueryOptions,
...
+		return &http.Server{
+			Handler:   recoveryHandler(handler),
+			TLSConfig: tlsCfg,
+		}, nil
+
+	}
 	return &http.Server{
 		Handler: recoveryHandler(handler),
-	}
+	}, nil
```
yurishkuro, line 159: avoid repetition
```go
server := &http.Server{
    Handler: recoveryHandler(handler),
}
if queryOpts.TLSHTTP.Enabled {
   ...
   server.TLSConfig = tlsConfig
}
return server, nil
```

jpkrohling:
@rjs211 are you still interested in working on this one?

davidcmitchell:
@rjs211 Are you still working on this issue?

rjs211:
I had finished this task and all the test cases had passed. I had received one approval and was waiting for a second one.  I see now that in the meantime the branch is our of date with master.  I'll start working on this again.

jpkrohling:
Is this ready for a review?

rjs211:
Yes.  It is.  There weren't any merge conflicts.  Ready for review.

jpkrohling:
I'll take another look tomorrow! Thanks for working on this one.

jpkrohling:
Sorry, didn't have time to look into this yet, but should be able to look at this at most early next week.

jpkrohling (approved):
Would you like to give a final review, @yurishkuro ?

yurishkuro (approved):

yurishkuro:
@rjs211 thanks!  Could you add another PR that describes these changes in the CHANGELOG? This change is not as straightforward as the PR title implies, due to the tricky handling of separate ports.

rjs211:
@yurishkuro Hello, and apologies for delayed response. My personal laptop was under repair.  I was looking at the entries in https://github.com/jaegertracing/jaeger/blob/master/CHANGELOG.md and they seem to me one liners without much detail.  Could you please show me an example where some logic is explained?

Would writing a documentation help? If so, could you please direct me to the correct place to write the same? Thanks.

yurishkuro:
See, for example, the Breaking Changes section in https://github.com/jaegertracing/jaeger/blob/master/CHANGELOG.md#1200-2020-09-29

We don't need to explain the logic, but what is changing and how those changes affect users.

PR #2349, opened 2020-07-17
Add ability to stop and drain the bounded queue

bogdandrutu:
- Draining the queue when stop is called. This is to ensure that items are not dropped while shutting down.

- Close the items channel first to ensure consumers consume all the items from the queue. then call stop which has no effect on the consumers but will stop the length reporting

### DIFF
--- a/pkg/queue/bounded_queue_test.go
+++ b/pkg/queue/bounded_queue_test.go
@@ -112,6 +112,55 @@ func TestBoundedQueue(t *testing.T) {
 	assert.False(t, q.Produce("x"), "cannot push to closed queue")
 }
 
+func TestBoundedQueue_StopWithDrain(t *testing.T) {
+	q := NewBoundedQueue(1, func(item interface{}) {})
+	assert.Equal(t, 1, q.Capacity())
+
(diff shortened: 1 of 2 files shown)


pkg/queue/bounded_queue_test.go
```diff
@@ -112,6 +112,55 @@ func TestBoundedQueue(t *testing.T) {
...
+	consumerState.assertConsumed(map[string]bool{
+		"a": true,
+	})
+
+	// produce one more item to fill the queue
+	assert.True(t, q.Produce("b"))
+	assert.Equal(t, 1, q.Size())
+
+	go func() {
+		// this is a bit of a race condition because it may unblock the consumer before
```
yurishkuro, line 150: can you use a WaitGroup instead to unblock this after calling StopWithDrain?  We're trying to keep the tests executing as fast as possible, and half a second is a noticeable delay, while may be insufficient on a busy CI server to avoid race condition.

jpkrohling: @bogdandrutu are you still interested in working on this?

PR #2371, opened 2020-08-04
Add --kafka.producer.batch-min-messages collector flag

prymitive:

Currently to control batch size when collector writes to kafka user can set `--kafka.producer.batch-max-messages` flag. This flag is used to configure kafka client library and set upper limit on the number of messages in a single batch, so the meaning of this flag is 'write up to N messages in a single batch'.
Current wording on `--kafka.producer.batch-max-messages` help text is 'Number of message to batch before sending records to Kafka.' which suggests that it's used to set minimal number of messages needed to send a batch to kafka.
For a user who wants to enforce a minimal batch size this means that there's currently no way of doing that, but at the same time there's a flag that is described as a way to do it, while actually doing something else.

This patch adds a `--kafka.producer.batch-min-messages` flag that allows configuring minimal batch size in the kafka client library and updates `--kafka.producer.batch-max-messages` flag description to be more clear on the effect it causes.

### DIFF
--- a/pkg/kafka/producer/config.go
+++ b/pkg/kafka/producer/config.go
@@ -36,6 +36,7 @@ type Configuration struct {
 	ProtocolVersion           string                  `mapstructure:"protocol_version"`
 	BatchLinger               time.Duration           `mapstructure:"batch_linger"`
 	BatchSize                 int                     `mapstructure:"batch_size"`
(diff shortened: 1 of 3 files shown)


yurishkuro:
thank you!

PR #2387, opened 2020-08-13
Separate Ports for GRPC and HTTP requests in Query Server

### LINKED ISSUE #2338: gRPC with TLS doesn't work with cmux in mixed mode
**Describe the bug**
I faced https://github.com/open-telemetry/opentelemetry-collector/issues/1227 some time ago, and while reviewing #2337, I noticed that we have the same pattern in Jaeger:

https://github.com/jaegertracing/jaeger/blob/02d45678ec9079ee91e19ad901ab7592aa4de085/cmd/query/app/server.go#L138-L145

Basically, when gRPC is behind TLS, `cmux` can't determine the headers before doing the TLS handshake, and the TLS handshake happens at the gRPC layer. The end result is that all calls end up going to the HTTP handler when a client is attempting to establish a gRPC TLS connection.

**To Reproduce**
1. Not reproducible at the moment, as there's a bug preventing the TLS flags from being exposed for gRPC in the Query component.

A possible solution is mentioned here in a follow-up issue at OpenTelemetry Collector: https://github.com/open-telemetry/opentelemetry-collector/issues/1256#issuecomment-654975422

### LINKED ISSUE #2377: Separate GRPC and HTTP ports for Query

Required for enabling TLS support for query server in the future. #2249 

There is a need for adding separate ports for grRPC and http component of query server so as to eliminate the use of cmux.   Cmux elimination is required for adding TLS support in the future and maintain performance doing so.  Presence of cmux forces the server to use a single certificate to be used for both gRPC and http requests,  which isn't always favourable, as pointed out by @tcolgate  https://github.com/jaegertracing/jaeger/pull/2337#discussion_r462166145 

Proposed by @jpkrohling  in https://github.com/jaegertracing/jaeger/pull/2337#pullrequestreview-458476438  and reviewed by @yurishkuro 
copying from above:

1.  add a new flag, `--query.grpc-server.host-port` (like the collector)
1.  replace `--query.host-port` with `--query.http-server.host-port` (again, like the collector)
1.  when no specific gRPC or http host-port is set, the current behavior is kept
This way, we keep backwards compatibility with current users, while we have a chance to get out of the `cmux`

declaring and defining of default values for flag: 'query.http-server.host-port'  as `ports.queryHTTP` and flag: 'query.grpc-server.host-port'  as `ports.queryGRPC`

rjs211:
- Resolves #2377 
- Resolves #2338

- separate the HTTP port and gRPC port for query server
- TODO:  define default port for gRPC and http is set


rjs211:
Is there any convention for defining new ports and port Numbers?  (Need to add a default port of gRPC requests) `query.grpc-server.host-port`

jpkrohling:
For ports that are highly related, just use the next integer for the new port, if it's free. Like: `16686` for `query-http` and `16687` for `query-grpc`.

rjs211:
I am going with `16685` for `query-grpc` (new port) and  `16686` for `query-http` (existing port )( `16687` is already in use by `query-admin`)

objectiser:
@rjs211 @jpkrohling Wouldn't this be a breaking change then? Shouldn't the default value for both of the new flags be `16686`, with the logic as suggested by Yuri - so that by default (with no flags) it will use cmux. So if users want them on separate ports, they will then decide which port numbers they want to use.

rjs211:
@objectiser 

The actual assignment is as follows:

1.  The values of  {`query.host-port` , `query.port`, `query.http-server.host-port`, `query.grpc-server.host-port` }   are initalized form the viper (including default values. At the end of point, grpc and http flags have different default values )
1.  if  neither of {`query.host-port` , `query.port`} is explicitly set and atleast one of {`query.http-server.host-port`, `query.grpc-server.host-port`} is set explicitly,  then the current option with different values of http and grpc ports is returned.
1.  else (this includes no ports defined case as well),  both `query.http-server.host-port`  and `query.grpc-server.host-port` is assigned the value of  {`query.host-port` , `query.port`} whichever is defined.

for now, default values of `query.http-server.host-port` or `query.grpc-server.host-port`  is used in case of the second point (when one of them is defined).

This will be more significant when the usage of TLS is enabled for either or both of grpc and http (once http TLS is supported by #2337 ) and the use of two separate ports will be forced on the user (as explained by issue #2377 ).

objectiser:
@rjs211 Understand the reasons for having separate ports when TLS is being configured, but from the sounds of (1):

> The values of {query.host-port , query.port, query.http-server.host-port, query.grpc-server.host-port } are initalized form the viper (including default values. At the end of this point, grpc and http flags have different default values )

the default behaviour (i.e. where no flags are defined) will be that grpc and http query ports are different. Therefore this is a breaking change?

rjs211:
the case of `No ports defined` or `no flags defined`  is (3).  The conditional in (2) would fail if no ports / flags are defined.

jpkrohling:
Looks like we are all on the same page, but just to be clear: we should not have breaking changes.

cmd/query/app/flags_test.go
```diff
@@ -60,6 +61,56 @@ func TestQueryBuilderFlags(t *testing.T) {
...
+	assert.Equal(t, ports.PortToHostPort(ports.QueryGRPC), qOpts.GRPCHostPort)
+
+	assert.Equal(t, http.Header{
+		"Access-Control-Allow-Origin": []string{"blerg"},
+		"Whatever":                    []string{"thing"},
+	}, qOpts.AdditionalHeaders)
+	assert.Equal(t, 10*time.Second, qOpts.MaxClockSkewAdjust)
+}
+
+func TestQueryBuilderFlagsSeparateNoPorts(t *testing.T) {
```
rjs211, line 89: @objectiser I think this test case would verify that the change is not a breaking change.

cmd/query/app/flags.go
```diff
@@ -66,7 +72,7 @@ type QueryOptions struct {
 	UIConfig string
 	// BearerTokenPropagation activate/deactivate bearer token propagation to storage
 	BearerTokenPropagation bool
-	// TLS configures secure transport
+	// TLS configures secure transport`
```
jpkrohling, line 75:
```suggestion
	// TLS configures secure transport
```

jpkrohling, line 75:
```suggestion
	// TLS configures secure transport
```

cmd/query/app/flags.go
```diff
@@ -78,6 +84,8 @@ type QueryOptions struct {
 func AddFlags(flagSet *flag.FlagSet) {
 	flagSet.Var(&config.StringSlice{}, queryAdditionalHeaders, `Additional HTTP response headers.  Can be specified multiple times.  Format: "Key: Value"`)
 	flagSet.String(queryHostPort, ports.PortToHostPort(ports.QueryHTTP), "The host:port (e.g. 127.0.0.1:5555 or :5555) of the query's HTTP server")
+	flagSet.String(queryHTTPHostPort, ports.PortToHostPort(ports.QueryHTTP), "The host:port (e.g. 127.0.0.1:5555 or :5555) of the query's HTTP server")
+	flagSet.String(queryGRPCHostPort, ports.PortToHostPort(ports.QueryGRPC), "The host:port (e.g. 127.0.0.1:5555 or :5555) of the query's HTTP server")
```
jpkrohling, line 88:
```suggestion
	flagSet.String(queryGRPCHostPort, ports.PortToHostPort(ports.QueryGRPC), "The host:port (e.g. 127.0.0.1:5555 or :5555) of the query's gRPC server")
```

cmd/query/app/flags_test.go
```diff
@@ -60,6 +61,56 @@ func TestQueryBuilderFlags(t *testing.T) {
 	assert.Equal(t, 10*time.Second, qOpts.MaxClockSkewAdjust)
 }
 
+func TestQueryBuilderFlagsSeparatePorts(t *testing.T) {
+	v, command := config.Viperize(AddFlags)
+	command.ParseFlags([]string{
```
jpkrohling, line 66: Do you need all the flags for this test? I think only `--query.http-server.host-port` is under test here, right?

cmd/query/app/flags_test.go
jpkrohling: Could you please leave only the flags that are relevant for the test?

cmd/query/app/server_test.go
```diff
@@ -100,17 +100,61 @@ func TestServer(t *testing.T) {
 	assert.Equal(t, healthcheck.Unavailable, flagsSvc.HC().Get())
 }
 
+func TestServer2(t *testing.T) {
```
jpkrohling, line 103: This might deserve a better name. Can you describe what this particular is doing? What's different from it and the previous test? Suggestion: `TestServerWithDedicatedPorts`

cmd/query/app/server_test.go
```diff
@@ -100,17 +100,61 @@ func TestServer(t *testing.T) {
...
+	flagsSvc := flags.NewService(ports.QueryAdminHTTP)
+	flagsSvc.Logger = zap.NewNop()
+	// hostPort := ports.GetAddressFromCLIOptions(ports.QueryHTTP, "")
```
jpkrohling, line 106: You might want to remove this line.

cmd/query/app/server_test.go
```diff
@@ -100,17 +100,61 @@ func TestServer(t *testing.T) {
...
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
+	defer cancel()
+
+	res, err := client.GetServices(ctx, &api_v2.GetServicesRequest{})
+	assert.NoError(t, err)
+	assert.Equal(t, expectedServices, res.Services)
+
+	server.Close()
+	for i := 0; i < 10; i++ {
+		if flagsSvc.HC().Get() == healthcheck.Unavailable {
```
jpkrohling, line 138: I wonder if we have a new flaky test here (I don't remember if the `TestServer` was flaky), but ideally, you'd create a `wg := sync.WaitGroup{}` at the beginning, with `wg.Add(1)`. In the line where you set the health check status, you'd check if it contains the state you are expecting and then call `wg.Done()`. At the end of your test, you'd call `wg.Wait()`. If the test times out, your test failed.

cmd/query/app/server_test.go
```diff
@@ -147,5 +191,6 @@ func TestServerHandlesPortZero(t *testing.T) {
 
 	onlyEntry := message.All()[0]
 	port := onlyEntry.ContextMap()["port"]
+	fmt.Println(port)
```
jpkrohling, line 194: Left over?

jpkrohling (commented):
I'll do a couple of manual tests, and there are a couple of minor things to change, but in general, LGTM.

cmd/query/app/flags.go
```diff
@@ -78,6 +84,8 @@ type QueryOptions struct {
 func AddFlags(flagSet *flag.FlagSet) {
 	flagSet.Var(&config.StringSlice{}, queryAdditionalHeaders, `Additional HTTP response headers.  Can be specified multiple times.  Format: "Key: Value"`)
 	flagSet.String(queryHostPort, ports.PortToHostPort(ports.QueryHTTP), "The host:port (e.g. 127.0.0.1:5555 or :5555) of the query's HTTP server")
```
jpkrohling, line 86: Please, mark this port as deprecated as well.

jpkrohling:
For the manual test, I build a simple Go application that lists the services from a target Query server, and I can confirm that the gRPC changes from this PR works:

```golang
package main

import (
	"context"
	"flag"
	"fmt"

	_ "github.com/jaegertracing/jaeger/proto-gen/api_v2"
	"go.uber.org/zap"
	"google.golang.org/grpc"
)

func main() {
	logger, err := zap.NewDevelopment()
	if err != nil {
		fmt.Println(err)
		return
	}

	hostPort := flag.String("host-port", "localhost:16686", "the host-port to use when connecting to the remote Jaeger Query server")
	flag.Parse()

	fmt.Printf("Using host-port: %s\n", *hostPort)

	conn, err := grpc.Dial(*hostPort, grpc.WithInsecure())
	if err != nil {
		logger.Error("failed to connect to the collector", zap.Error(err))
	}
	defer conn.Close()

	client := api_v2.NewQueryServiceClient(conn)

	req := api_v2.GetServicesRequest{}
	resp, err := client.GetServices(context.Background(), &req)
	if err != nil {
		logger.Error("failed to get list of services", zap.Error(err))
		return
	}

	if len(resp.Services) == 0 {
		fmt.Println("No services at the target server.")
		return
	}

	fmt.Println("Available services at the target server:")
	for _, svc := range resp.Services {
		fmt.Printf("- %s\n", svc)
	}
}
```

ports/ports.go
```diff
@@ -52,6 +54,14 @@ func PortToHostPort(port int) string {
 	return ":" + strconv.Itoa(port)
 }
 
+// HostPortToPort converts the  host:port address string intoto port int
+func HostPortToPort(hostPort string) (int, error) {
+	s := strings.Split(hostPort, ":")
```
jpkrohling, line 59: codecov is saying that those lines aren't covered by tests. Would you please add one test specific to this function?

jpkrohling (approved):
LGTM.

jpkrohling:
Anyone else wants to re-review? @yurishkuro ?

cmd/query/app/server.go
```diff
@@ -42,13 +43,27 @@ type Server struct {
...
+
+	httpPort, err := ports.HostPortToPort(options.HTTPHostPort)
+	if err != nil {
+		return nil, err
+	}
+	grpcPort, err := ports.HostPortToPort(options.GRPCHostPort)
+	if err != nil {
+		return nil, err
+	}
+	separatePorts := (grpcPort != httpPort)
```
albertteoh, line 65: nit: could probably inline this below: `separatePorts: (grpcPort != httpPort)`

cmd/query/app/server.go
```diff
@@ -138,16 +170,46 @@ func (s *Server) Start() error {
...
+	if err != nil {
+		return err
+	}
+
+	var tcpPort int
+	if !s.separatePorts {
+		if port, err := netutils.GetPort(s.conn.Addr()); err == nil {
+			tcpPort = port
+		}
+
```
albertteoh, line 197: nit: please remove blank line, unless if this was intentional

cmd/query/app/flags.go
```diff
@@ -86,14 +95,33 @@ func AddFlags(flagSet *flag.FlagSet) {
...
+	qOpts.HostPort = ports.GetAddressFromCLIOptions(v.GetInt(queryPort), v.GetString(queryHostPort))
+
+	qOpts.TLS = tlsFlagsConfig.InitFromViper(v)
+
+	// query.host-port is not defined and atleast one of query.grpc-server.host-port or query.http-server.host-port is defined
+	// user intends to use the separate flags.
+	if !(v.IsSet(queryHostPort) || v.IsSet(queryPort)) && (v.IsSet(queryHTTPHostPort) || v.IsSet(queryGRPCHostPort)) {
+		return qOpts
+	}
+	logger.Warn(fmt.Sprintf("Use of %s and %s is deprecated.  use %s and %s instead", queryPort, queryHostPort, queryHTTPHostPort, queryGRPCHostPort))
```
albertteoh, line 111: nit: s/  use/ Use/

ports/ports_test.go
```diff
@@ -39,3 +39,15 @@ func TestGetAddressFromCLIOptionOnlyPort(t *testing.T) {
 func TestGetAddressFromCLIOptionOnlyPortWithoutColon(t *testing.T) {
 	assert.Equal(t, ":123", GetAddressFromCLIOptions(0, "123"))
 }
+
+func TestHostPortToPortOnlyPort(t *testing.T) {
+	port, err := HostPortToPort(":42")
+	assert.Nil(t, err)
```
albertteoh, line 45: nit: Consider `assert.NoError(t, err)` to be more explicit (same below).

cmd/query/app/flags.go
```diff
@@ -86,14 +95,33 @@ func AddFlags(flagSet *flag.FlagSet) {
...
+// InitPortsConfigFromViper initializes the port numbers and TLS configuration of ports
+func (qOpts *QueryOptions) InitPortsConfigFromViper(v *viper.Viper, logger *zap.Logger) *QueryOptions {
+	qOpts.HTTPHostPort = v.GetString(queryHTTPHostPort)
+	qOpts.GRPCHostPort = v.GetString(queryGRPCHostPort)
+	qOpts.HostPort = ports.GetAddressFromCLIOptions(v.GetInt(queryPort), v.GetString(queryHostPort))
+
+	qOpts.TLS = tlsFlagsConfig.InitFromViper(v)
+
+	// query.host-port is not defined and atleast one of query.grpc-server.host-port or query.http-server.host-port is defined
+	// user intends to use the separate flags.
```
albertteoh, line 107: suggestion: "User intends to use separate GRPC and HTTP host:port flags."

ports/ports.go
```diff
@@ -52,6 +54,14 @@ func PortToHostPort(port int) string {
 	return ":" + strconv.Itoa(port)
 }
 
+// HostPortToPort converts the  host:port address string intoto port int
+func HostPortToPort(hostPort string) (int, error) {
```
albertteoh, line 58: Could [`netutils.GetPort()`](https://github.com/jaegertracing/jaeger/blob/master/pkg/netutils/port.go#L23) be used instead?

albertteoh (commented):
Looks pretty good to me! Mostly minor comments.

cmd/query/app/flags.go
```diff
@@ -40,6 +40,9 @@ const (
 	queryHostPort           = "query.host-port"
 	queryPort               = "query.port"
 	queryPortWarning        = "(deprecated, will be removed after 2020-08-31 or in release v1.20.0, whichever is later)"
+	queryHOSTPortWarning    = "(deprecated, will be removed after 2020-11-30 or in release v1.20.0, whichever is later)"
```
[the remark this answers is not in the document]
jpkrohling: Actually, this is now wrong again. It should be at least one version after the next one, so that people get enough time to migrate/remove. 

The date is a bit arbitrary: we aim to have one release per month, so, a couple of months in the future means two versions, which matches the "next+1" idea.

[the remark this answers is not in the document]
yurishkuro: The date is not arbitrary either, we agreed on 3 months or v+2, whichever is later, see #2113 (which needs to be moved to CONTRIBUTING)

ports/ports.go
rjs211: Thanks.  This method lead to the `SplitHostPort()` method in the `net` package.

albertteoh:
@rjs211 please increase the patch code coverage.

albertteoh (approved):
Thanks @rjs211! LGTM

jpkrohling:
Wonderful work, @rjs211! Thank you for your contribution!

rjs211:
@jpkrohling @albertteoh @objectiser @yurishkuro Thank you very much for your guidance and patience.  This is my first code contribution to opensource.  Hope to contribute further.

pavolloffay:
This PR introduced warking message in OTEL all-in-one:
```
2020-09-16T16:18:41.492+0200	WARN	app/flags.go:111	Use of query.port and query.host-port is deprecated.  Use query.http-server.host-port and query.grpc-server.host-port instead
github.com/jaegertracing/jaeger/cmd/query/app.(*QueryOptions).InitPortsConfigFromViper
	/home/ploffay/projects/jaegertracing/jaeger/cmd/query/app/flags.go:111
github.com/jaegertracing/jaeger/cmd/query/app.(*QueryOptions).InitFromViper
	/home/ploffay/projects/jaegertracing/jaeger/cmd/query/app/flags.go:120
main.startQuery
	/home/ploffay/projects/jaegertracing/jaeger/cmd/opentelemetry/cmd/all-in-one/main.go:194
main.main
	/home/ploffay/projects/jaegertracing/jaeger/cmd/opentelemetry/cmd/all-in-one/main.go:152
runtime.main
	/home/ploffay/bin/go/src/runtime/proc.go:203
```

rjs211:
I logged the warning intentionally as per my understanding of https://github.com/jaegertracing/jaeger/issues/2377#issuecomment-671122786  .  Should have I not done that?

pavolloffay:
I have submitted a PR to fix it https://github.com/jaegertracing/jaeger/pull/2479

PR #2343, opened 2020-07-15
Use a single shared set of CA, client & server keys/certs for testing

rjs211:
Signed-off-by: rjs211 <srivatsa211@gmail.com>

## Which problem is this PR solving?
- currently, there are two versions of certificates used for for testing.  This is redundant 

## Short description of the changes
- consolidate the certificates needed for any secure client-server channel used during testing.   
- Added README.md for creation of CA, client, server keys and certificates

PR #2390, opened 2020-08-13
Update validity of TLS certificates to 3650 days

rjs211:

- TLS certificate validity expired for example certificates

-  Previously, the default duration of validity (30 days) was used for generating certificates.  This PR increases it to 3650 days or 10 years


rjs211:
#2343 It was my blunder in the first place.  Apologies.

PR #2401, opened 2020-08-19
Fix go vet warnings on Go 1.15

prymitive:
go vet reports 3 issues when using Go 1.15:

```
go vet ./...
crossdock/services/tracehandler_test.go:40:12: conversion from untyped int to TraceID (string) yields a string of one rune, not a string of digits (did you mean fmt.Sprint(x)?)
crossdock/services/tracehandler_test.go:252:90: conversion from untyped int to TraceID (string) yields a string of one rune, not a string of digits (did you mean fmt.Sprint(x)?)
pkg/normalizer/service_name.go:57:28: conversion from int to string yields a string of one rune, not a string of digits (did you mean fmt.Sprint(x)?)
```
Signed-off-by: Łukasz Mierzwa <l.mierzwa@gmail.com>

### DIFF
--- a/pkg/normalizer/service_name.go
+++ b/pkg/normalizer/service_name.go
@@ -54,7 +54,7 @@ func newServiceNameReplacer() *strings.Replacer {
 	oldnew := make([]string, 0, 2*(256-2-10-int('z'-'a'+1)))
 	for i := range mapping {
 		if mapping[i] != byte(i) {
-			oldnew = append(oldnew, string(i), string(mapping[i]))
+			oldnew = append(oldnew, string(rune(i)), string(rune(mapping[i])))
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


jpkrohling (approved):

pkg/normalizer/service_name.go
```diff
@@ -54,7 +55,7 @@ func newServiceNameReplacer() *strings.Replacer {
 	oldnew := make([]string, 0, 2*(256-2-10-int('z'-'a'+1)))
 	for i := range mapping {
 		if mapping[i] != byte(i) {
-			oldnew = append(oldnew, string(i), string(mapping[i]))
+			oldnew = append(oldnew, fmt.Sprintf("%c", i), fmt.Sprintf("%c", mapping[i]))
```
yurishkuro, line 58: I think the recommendation from 1.15 release notes is to use `string(rune(i))`

prymitive: updated, thanks

yurishkuro:
Thanks!

You may want to do another PR bumping Go version in our CI and go.mod

PR #2402, opened 2020-08-19
Apply go fmt

zhouhao3:
- 

- 
I found that [ALL_SRC](https://github.com/jaegertracing/jaeger/blob/master/Makefile#L6-L21) does not include some folders in the Makefile, which results in the format of some files that cannot be detected by make fmt. Not sure if this is acceptable.

Signed-off-by: Zhou Hao <zhouhao@cn.fujitsu.com>

### DIFF
--- a/examples/hotrod/services/driver/client.go
+++ b/examples/hotrod/services/driver/client.go
@@ -20,9 +20,9 @@ import (
 	"time"
 
 	otgrpc "github.com/opentracing-contrib/go-grpc"
-	"google.golang.org/grpc"
 	"github.com/opentracing/opentracing-go"
 	"go.uber.org/zap"
+	"google.golang.org/grpc"
 
 	"github.com/jaegertracing/jaeger/examples/hotrod/pkg/log"
 )
--- a/examples/hotrod/services/driver/server.go
+++ b/examples/hotrod/services/driver/server.go
@@ -34,7 +34,7 @@ type Server struct {
 	tracer   opentracing.Tracer
 	logger   log.Factory
 	redis    *Redis
-	server *grpc.Server
+	server   *grpc.Server
 }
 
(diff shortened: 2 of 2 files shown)


thrift-gen/agent/agent-consts.go
```diff
@@ -3,15 +3,14 @@
...
+import (
 	"bytes"
 	"context"
-	"reflect"
 	"fmt"
 	"github.com/apache/thrift/lib/go/thrift"
 	"github.com/jaegertracing/jaeger/thrift-gen/jaeger"
 	"github.com/jaegertracing/jaeger/thrift-gen/zipkincore"
-
+	"reflect"
```
jpkrohling, line 13: Something went wrong here.

thrift-gen/agent/agent.go
```diff
@@ -3,15 +3,14 @@
...
+import (
 	"bytes"
 	"context"
-	"reflect"
 	"fmt"
 	"github.com/apache/thrift/lib/go/thrift"
 	"github.com/jaegertracing/jaeger/thrift-gen/jaeger"
 	"github.com/jaegertracing/jaeger/thrift-gen/zipkincore"
-
+	"reflect"
```
jpkrohling, line 13: And here :)

jpkrohling (commented):
Shouldn't this PR include some change to the Makefile to include the new paths? I would also like to leave the generated files alone, or to apply a `make fmt` right after generating them. Otherwise, they'll be reverted whenever they are re-generated.

yurishkuro:
We do not want to format or lint auto-generated files.

zhouhao3:
updated, PTAL.

yurishkuro:
This still includes mocks, which are auto-generated.

zhouhao3:
Done.

Issue #2310, opened 2020-06-25
Add TraceID to Logs in HotRod Example Application


ghostsquad:
This is a request to add the TraceID to logs for the HotRod application. This would allow for some cool demos, especially with Grafana 7.0 and it's support for unification of metrics, logs and jaeger tracing.

pavolloffay:
Would you like to submit a patch for it?

ghostsquad:
Ya. This would be a good exercise for me.

PR #2384, opened 2020-08-12
Added span and trace id to hotrod logs

joe-elliott:
## Which problem is this PR solving?
- Fixes #2310

## Short description of the changes
- Adds trace ID and span ID hotrod logs as requested.

Example:
```
2020-08-11T22:20:15.010-0400	INFO	route/client.go:54	Finding route	{"service": "frontend", "component": "route_client", "trace_id": "6b1ab6a4c91f3780", "span_id": "6b1ab6a4c91f3780", "pickup": "193,499", "dropoff": "577,322"}
2020-08-11T22:20:15.010-0400	INFO	route/client.go:54	Finding route	{"service": "frontend", "component": "route_client", "trace_id": "6b1ab6a4c91f3780", "span_id": "6b1ab6a4c91f3780", "pickup": "476,860", "dropoff": "577,322"}
2020-08-11T22:20:15.010-0400	INFO	route/client.go:54	Finding route	{"service": "frontend", "component": "route_client", "trace_id": "6b1ab6a4c91f3780", "span_id": "6b1ab6a4c91f3780", "pickup": "324,354", "dropoff": "577,322"}
2020-08-11T22:20:15.010-0400	INFO	route/server.go:71	HTTP request received	{"service": "route", "trace_id": "6b1ab6a4c91f3780", "span_id": "4a17e4c6aeb157ad", "method": "GET", "url": "/route?dropoff=577%2C322&pickup=193%2C499"}
2020-08-11T22:20:15.011-0400	INFO	route/server.go:71	HTTP request received	{"service": "route", "trace_id": "6b1ab6a4c91f3780", "span_id": "3b22d1cc6f90e8c", "method": "GET", "url": "/route?dropoff=577%2C322&pickup=476%2C860"}
2020-08-11T22:20:15.011-0400	INFO	route/server.go:71	HTTP request received	{"service": "route", "trace_id": "6b1ab6a4c91f3780", "span_id": "12de05f88735056f", "method": "GET", "url": "/route?dropoff=577%2C322&pickup=324%2C354"}
```

I know this was assigned to @ghostsquad, but it's been over a month so I took a stab at it.

PR #2408, opened 2020-08-22
Add Grafana integration example

fktkrt:
- The idea came from https://github.com/jaegertracing/jaeger/issues/2310 to provide a bundled Grafana integration example

- This is a rework of one of the early official Grafana demos and the hotrod compose, having Grafana, Loki and Jaeger packed together to provide easy correlation with metrics, logs and traces in one application.
 - If the effort is welcomed, I can add further details to the README regarding the infrastructure.


examples/grafana-integration/docker-compose.yaml
```diff
@@ -0,0 +1,51 @@
...
+  grafana:
+    image: grafana/grafana:7.1.4
+    ports:
+      - '3000:3000'
+    volumes:
+      - ./datasources.yaml:/etc/grafana/provisioning/datasources/datasources.yaml
+    logging:
+      driver: loki
+      options:
+        loki-url: 'http://localhost:3100/api/prom/push'
```
yurishkuro, line 12: Shouldn't this be `loki:3100`?

examples/grafana-integration/README.md
```diff
@@ -0,0 +1,26 @@
+# Hot R.O.D. - Rides on Demand  - Grafana integration
+
+This is the Hot R.O.D. demo application that consists of the same components as the `examples/hotrod/`, only Grafana and Loki integration is added to this setup, so you can correlate logs and traces in one application.
```
yurishkuro, line 3: It would be useful to explain this in more detail. If someone is new to this whole space, with the main HotROD example we at least have a detailed blog and many videos showing _what to do_ with the example. But here you're assuming substantial familiarity with Grafana and Loki. For example, I think there are some standard Grafana dashboards for Jaeger, could this be included? Or at least a dashboard that shows, ideally, both the HotROD metrics, Loki logs, and traces.

examples/grafana-integration/docker-compose.yaml
joe-elliott: The docker driver runs in the host network namespace.  So localhost is correct.  I just learned this recently for a little demo I put together.

yurishkuro: I am not sure this is universal behavior. I'd say it's clearer to refer to the host name of the service than assume that localhost is correctly connected.

joe-elliott: Right, but "loki" won't resolve in the host network namespace and the driver won't be able to send logs.  The port mapping on the loki service allows the docker logging driver to refer to localhost:3100 and successfully send logs to loki.

This does work consistently in Linux.  Unsure about OSX.

yurishkuro: why wouldn't `loki` resolve, but `jaeger` in L20 does?

joe-elliott: because the logging driver is running on the host, but the containers are running on the docker bridge.

fktkrt: That is correct. 
I was experimenting with different configs, and this came out as the most portable and consistent solution. An other one could be using `http://host.docker.internal:3100`, which can work on OSX and Windows, but only recently (?) on Linux, so I did go with the former one.
I can also confirm that it's working fine in both Linux and in OSX, I believe it should work on Windows as well, but currently I don't have access to any.

examples/grafana-integration/README.md
jpkrohling: +1 for adding a Grafana dashboard to the main `monitoring` directory in the repository, similar to what we have for Jaeger itself:

https://github.com/jaegertracing/jaeger/tree/master/monitoring/jaeger-mixin

fktkrt: Yes, I agree, that's exactly what I meant in the PR description. 
I can think of multiple possible ways to extend/transform this:

- I can remove the Loki integration, this way it's a much more simple solution, basically a clean Grafana demo as the frontend to Jaeger -> the setup is more simple, shorter introduction is needed
- I can give additional details to this README to make it easier to understand the setup AND put together an even more detailed blog post, explaining all aspects thoroughly

As for the dashboards:
As far as I know there isn't any official Jaeger dashboard that could be useful for this scenario (if we would like to focus on the tracing aspect of it), since most of them are for monitoring Jaeger itself currently. These can be added, and should be added, just these serve a bit different purpose in my opinion, so having these under `monitoring/` can make sense. 

Although, I've just found [this one](https://grafana.com/grafana/dashboards/12611), which is something similar what could be useful for this setup, but some changes would be needed to tailor to our needs (if it's a welcomed effort, I can work on it).  
Incorporating a dashboard to filter actual logs (and eventually metrics, since those are not yet implemented in this setup) is a greater task, and it might benefit from a more detailed blog post format, but I am happy to work on those as well.
What do you think?

jpkrohling: I'm all for having a dashboard that serves as the official one, as long as we have people committed to maintain it.

fktkrt: Alright, I can open a PR to have one for the logs then.

In that case we would need a setup to validate and use it, so based on these, I'd say the Loki setup should be included as well. 
Although, another option would be to have a dedicated example for vanilla Grafana (for the sake of simplicity), and one for this more advanced one, where Loki and related dashboards are included.

yurishkuro: as long as the overall setup is simple i.e. `docker-compose up`, I think it makes sense to include Loki integration and have a dashboard that already includes all the relevant panels for HotROD: metrics, logs, and traces.

fktkrt: I like the idea. 
I will add Prometheus to this setup too, then start working on a full-fledged HotROD overview dashboard.

fktkrt: Just pushed a new config w/ Prometheus integration. 
I will need to look into it more thoroughly, but after taking a quick look at the metrics including errors and the error labels, it's not quite trivial to build useful metrics panels. 

At the moment spotting issues seems easier with Loki, since I can see just filter for `{compose_project="grafana-jaeger"} |= "traceID"` then I can jump to the traces immediately (and this will be even better after https://github.com/jaegertracing/jaeger/pull/2384).  

Is there any dashboard/documentation focusing on the HotROD Prometheus metrics currently to fasten up this part?

examples/grafana-integration/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Running
+
+### Run everything via `docker-compose`
+
+#### Prerequisites
+* Download `docker-compose.yml` from https://github.com/jaegertracing/jaeger/blob/master/examples/grafana-integration/docker-compose.yml
+* Download the `datasources.yaml` from the `examples/grafana-integration/` folder
```
yurishkuro, line 11: Considering that there are several files in play here, I would rather recommend that people clone the repository and `cd examples/grafana-integration`.

examples/grafana-integration/docker-compose.yaml
```diff
@@ -0,0 +1,64 @@
...
+  loki:
+    image: grafana/loki:master
+    ports:
+      - '3100:3100'
+    command: -config.file=/etc/loki/local-config.yaml
+    environment:
+      - JAEGER_AGENT_HOST=jaeger
+      - JAEGER_AGENT_PORT=6831
+      - JAEGER_SAMPLER_TYPE=const
+      - JAEGER_SAMPLER_PARAM=1
```
yurishkuro, line 23: This is for tracing loki server itself?  I would add a comment

examples/grafana-integration/README.md
```diff
@@ -0,0 +1,26 @@
+# Hot R.O.D. - Rides on Demand  - Grafana integration
+
+This is the Hot R.O.D. demo application that consists of the same components as the `examples/hotrod/`, only Grafana and Loki integration is added to this setup, so you can correlate logs and traces in one application.
+
+## Running
+
+### Run everything via `docker-compose`
```
yurishkuro, line 7: This sub-section doesn't add anything, I would merge it into previous one as `## Running via docker-compose`

examples/grafana-integration/README.md
fktkrt: Added the first version of the HotROD dashboard, reorganized the file structure, extended the README.
Currently it looks like the redis timeouts don't have traceIDs, and the application metrics are not really showing significant error symptoms. 
I had a discussion with @joe-elliott, one of us will check the redis logging, and later I might will be able to open another PR to have new metrics.
Anyway, it looks like the current setup can be considered finished from my perspective. 
Once we have these metrics, I am planning to write a blogpost introducing this new kind of troubleshooting experience.

examples/grafana-integration/README.md
```diff
@@ -0,0 +1,66 @@
+# Hot R.O.D. - Rides on Demand  - Grafana integration
+
+This is the Hot R.O.D. demo application that consists of the same components as the `examples/hotrod/`, only Grafana, Loki and Prometheus integration is added to this setup, so you can correlate logs, metrics and traces in one application.
```
yurishkuro, line 3:
```suggestion
This example combines the Hot R.O.D. demo application ([examples/hotrod/](../hotrod/)) with Grafana, Loki and Prometheus integration, to demonstrate logs, metrics and traces correlation.
```

examples/grafana-integration/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Running via `docker-compose`
+
+### Prerequisites
+
+* Clone the Jaeger repository `git clone https://github.com/jaegertracing/jaeger.git`, then `cd examples/grafana-integration`
+
+* All services will log to Loki via the loki docker logging plugin.
+To install run this on the host:
```
yurishkuro, line 12:
```suggestion
* Install the Loki logging plugin for Docker:
```
If the plugin has its own page/README, I would link `Loki logging plugin` to it.

examples/grafana-integration/README.md
```diff
@@ -0,0 +1,56 @@
...
+```bash
+docker plugin install \
+grafana/loki-docker-driver:latest \
+--alias loki \
+--grant-all-permissions
+```
+
+### Deploy the services
+
+Pull up the services: `docker-compose up`
```
yurishkuro, line 23:
```suggestion
### Run the services

`docker-compose up` 
```

examples/grafana-integration/README.md
```diff
@@ -0,0 +1,56 @@
...
+--grant-all-permissions
+```
+
+### Deploy the services
+
+Pull up the services: `docker-compose up` 
+
+### Access the services
+* HotROD application at http://localhost:8080
+* Access Grafana UI at http://localhost:3000
```
yurishkuro, line 27:
```suggestion
* Grafana UI at http://localhost:3000
```

examples/grafana-integration/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+### Explore with Loki
+
+Currently the most powerful way to correlate application logs with traces can be performed via Grafana's Explore interface.
+
+After setting the datasource to Loki, all the log labels become available, and can be easily filtered using [Loki's LogQL query language](https://grafana.com/docs/loki/latest/logql/).
+
+For example after selecting the compose project, errors can be filtered with the following expression:
```
yurishkuro, line 35: what is "selecting the compose project"?

examples/grafana-integration/README.md
```diff
@@ -0,0 +1,56 @@
...
+{compose-project="grafana-integration"} |= "error"
+```
+
+which will list the redis timeout events.
+
+### HotROD - Metrics and logs overview dashboard
+
+Since the HotROD application can expose its metrics in Prometheus' format, these can be also used during investigation.
+
+This example includes a dashboard, which contains a log panel, which displays the logs for the selected services in real time. These can be also filtered by a search field.
```
yurishkuro, line 47:
```suggestion
This example includes a dashboard that contains a log panel for the selected services in real time. These can be also filtered by a search field.
```

"filtered by a search field" is a bit cryptic, do you want to provide an example?

examples/grafana-integration/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+There are also panels to display the ratio/percentage of errors in the current timeframe.
+
+Additionally, there are graphs for each service, visualing the rate of the requests and showing latency percentiles.
+
+### Clean up
+
+To stop the containers use `docker-compose -f path-to-yml-file down`
```
yurishkuro, line 55:
```suggestion
`docker-compose down`
```

examples/grafana-integration/datasources.yaml
```diff
@@ -0,0 +1,18 @@
+ApiVersion: 1
```
yurishkuro, line 1: is this the same as examples/grafana-integration/grafana/provisioning/datasources/datasources.yaml?

examples/grafana-integration/grafana/provisioning/dashboards/dashboard.yml
```diff
@@ -0,0 +1,11 @@
+apiVersion: 1
```
yurishkuro, line 1: Could these files be placed under a shorter path `examples/grafana-integration/grafana/`? I think this deep nesting makes the example unnecessarily more complex. If Grafana expects a certain path inside the container, we can always map it in the docker-compose.yaml.

fktkrt: I was thinking the same, although this way the services have their own folder, which makes it easier to tie the config files to their origin/purpose. 
Alternatively, I can link these from the `examples/grafana-integration/`, directly.

examples/grafana-integration/README.md
fktkrt: Added explanations.

examples/grafana-integration/README.md
fktkrt: Added explanation/example.

examples/grafana-integration/grafana/provisioning/dashboards/dashboard.yml
fktkrt: Alright, let's simplify it then.

examples/grafana-integration/datasources.yaml
fktkrt: Yes, it was a duplicate, now it's deleted.

examples/grafana-integration/README.md
fktkrt: Good point, linked.

examples/grafana-integration/datasources.yaml
yurishkuro: I still see the other file, which looks identical

fktkrt:
@yurishkuro All the suggestions are addressed, thank you.
Is there anything else needed?

PR #2409, opened 2020-08-24
Fix Elasticsearch version in ES OTEL writer

### LINKED ISSUE #2407: Does Jaeger OpenTelemetry support Elasticsearch 7?
In current Jaeger OpenTelemetry docker compose example
https://github.com/jaegertracing/jaeger/blob/master/crossdock/jaeger-opentelemetry-docker-compose.yml

it is using Elasticsearch 6:

```yml
image: docker.elastic.co/elasticsearch/elasticsearch-oss:6.8.3
```

I saw it is using OpenTelemetry version images for Jaeger:

- jaegertracing/jaeger-opentelemetry-collector
- jaegertracing/jaeger-opentelemetry-agent

I succeed testing on elasticsearch-oss 6.8.3 and 6.8.12, but failed upgrading to 7.0.0 and 7.9.0, got error:

```json
{
  "data": null,
  "total": 0,
  "limit": 0,
  "offset": 0,
  "errors": [
    {
      "code": 500,
      "msg": "search services failed: elastic: Error 400 (Bad Request): all shards failed [type=search_phase_execution_exception]"
    }
  ]
}
```

I tried to upgrading to Elasticsearch 7 for a while, but no luck.
So just wonder does Jaeger OpenTelemetry currently support Elasticsearch 7? cc author @pavolloffay 

If it does, it would be very helpful to upgrade the example docker-compose.yml to use Elasticsearch 7. Thanks!

pavolloffay:
 
Resolves #2407

### DIFF
--- a/cmd/opentelemetry/app/internal/esclient/client.go
+++ b/cmd/opentelemetry/app/internal/esclient/client.go
@@ -44,6 +44,9 @@ type ElasticsearchClient interface {
 	Search(ctx context.Context, query SearchBody, size int, indices ...string) (*SearchResponse, error)
 	// MultiSearch searches data via /_msearch
 	MultiSearch(ctx context.Context, queries []SearchBody) (*MultiSearchResponse, error)
+
+	// Major version returns major ES version
+	MajorVersion() int
 }
 
 // BulkResponse is a response returned by Elasticsearch Bulk API
@@ -172,20 +175,20 @@ func NewElasticsearchClient(params config.Configuration, logger *zap.Logger) (El
 	if err != nil {
 		return nil, err
 	}
-	if params.GetVersion() == 0 {
+	esVersion := int(params.GetVersion())
+	if esVersion == 0 {
 		esPing := elasticsearchPing{
 			username:     params.Username,
 			password:     params.Password,
 			roundTripper: roundTripper,
 		}
-		esVersion, err := esPing.getVersion(params.Servers[0])
+		esVersion, err = esPing.getVersion(params.Servers[0])
 		if err != nil {
 			return nil, err
 		}
 		logger.Info("Elasticsearch detected", zap.Int("version", esVersion))
-		params.Version = uint(esVersion)
 	}
-	return newElasticsearchClient(int(params.Version), clientConfig{
+	return newElasticsearchClient(esVersion, clientConfig{
 		DiscoverNotesOnStartup: params.Sniffer,
 		Addresses:              params.Servers,
 		Username:               params.Username,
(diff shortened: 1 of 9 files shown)


cmd/opentelemetry/app/internal/esclient/client.go
```diff
@@ -172,20 +175,20 @@ func NewElasticsearchClient(params config.Configuration, logger *zap.Logger) (El
...
 			password:     params.Password,
 			roundTripper: roundTripper,
 		}
-		esVersion, err := esPing.getVersion(params.Servers[0])
+		esVersion, err = esPing.getVersion(params.Servers[0])
 		if err != nil {
 			return nil, err
 		}
 		logger.Info("Elasticsearch detected", zap.Int("version", esVersion))
-		params.Version = uint(esVersion)
```
pavolloffay, line 186: The ES version is derived here in the client factory method and assigned to the config. However the config is passed by value, hence the changes are not reflected after this function. Then the wrong value of the version was used by the called to create index template.

Instead of overriding the config it is cleaner to expose the version on the client.

objectiser (approved):
LGTM, although lint failed. I'll approve if you can fix that.

cmd/opentelemetry/app/internal/esclient/client.go
```diff
@@ -172,20 +175,20 @@ func NewElasticsearchClient(params config.Configuration, logger *zap.Logger) (El
...
-	if params.GetVersion() == 0 {
+	esVersion := int(params.GetVersion())
+	if esVersion == 0 {
 		esPing := elasticsearchPing{
 			username:     params.Username,
 			password:     params.Password,
 			roundTripper: roundTripper,
 		}
-		esVersion, err := esPing.getVersion(params.Servers[0])
+		esVersion, err = esPing.getVersion(params.Servers[0])
```
joe-elliott, line 185: Does it make sense to check length of `.Servers` before dereferencing?  It does appear that right now it's impossible to pass Servers with length 0 into this method, but the code that prevents that is fairly distant from this.

Might be better if this method protects itself.

pavolloffay: +1 I am will submit a follow-up PR to fix that

pavolloffay: https://github.com/jaegertracing/jaeger/pull/2411

apptut:
I have define the --es.version=7, but in es v7.15.2 already not working.

```
HTTP Error: search services failed: elastic: Error 400 (Bad Request): all shards failed [type=search_phase_execution_exception]

{
  "data": null,
  "total": 0,
  "limit": 0,
  "offset": 0,
  "errors": [
    {
      "code": 500,
      "msg": "search services failed: elastic: Error 400 (Bad Request): all shards failed [type=search_phase_execution_exception]"
    }
  ]
}
```

PR #2416, opened 2020-08-25
WIP: Add support for status command

### LINKED ISSUE #2361: Jaeger containers healthcheck support

I would like to make my container depend on the successful start up of jager containers (agent for instance, but applies to all jaeger containers). In AWS ECS deployment is unable to perform a container-level healthcheck on the jaeger containers. So declaring my app to depend on the "HEALTHY" startup of a jaeger container is not possible at this time. Nor is ECS able to determine if the deployment completed successfully since it cannot healthcheck the app at startup.

Jaeger containers are stripped down and do not have a way to execute wget or curl to check the admin port. ECS needs a [command it can execute](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#container_definition_healthcheck) in the container that will exit 0.

Rather than including curl or wget which would bloat the containers needlessly. Add a built-in command to jager binaries to check their respective health (taking a minimal/identical set of environment/arguments to be compatible). The idea being we can very easily just have jaeger check it's own health and exit 0 if successful.

```
jaeger-agent health [--admin.http.host-port] [--config-file]
Checks the local agent admin endpoint to verify health.
```

dekimsey:
- Resolves #2361

- Adds a `status` subcommand to daemons

This was my first (or more like third) attempt at adding a `status` subcommand. My first pass simply hardcoded the parent command's default ports and that wasn't cool.

Major questions I have been unable to answer are:

  - Why does returning an err from cobra.Command generate help output?
  - The current cmd's flagSet doesn't seem to be affecting AdminServer, why?
  - How might this work for opentracing commands?
  - How should argument parsing be refactored? flags.AdminServer has a number of deprecations to deal with.

I haven't been able to get the overriding behavior on the argument handling working, unfortunately I still am not sure how the flags, Cobra, and Viper interact. I'm submitting this WIP to get some eyes/feedback on it.


go.mod
```diff
@@ -6,41 +6,44 @@ require (
 	github.com/AndreasBriese/bbloom v0.0.0-20190825152654-46b345b51c96 // indirect
 	github.com/DataDog/zstd v1.4.4 // indirect
 	github.com/Shopify/sarama v1.22.2-0.20190604114437-cd910a683f9f
+	github.com/VividCortex/gohistogram v1.0.0 // indirect
```
yurishkuro, line 9: I don't think there are any changes in this PR that require changing the dependencies, so please revert this. It's preferable to change dependencies (when needed) in a dedicated PR.

cmd/status/status.go
```diff
@@ -0,0 +1,69 @@
...
+			if err != nil {
+				return fmt.Errorf("no default admin port available for %s: %w", cmd.Name(), err)
+			}
+			return nil
+		},
+		// FIXME: Normally I'd want to return an Error and let the exit handler do the os.Exit.
+		// But it's triggering --help output everytime and I don't know why.
+		Run: func(cmd *cobra.Command, args []string) {
+			adminURL, _ = adminServer.URL()
+			resp, err := http.Get(adminURL)
```
yurishkuro, line 54: I am not sure that this command needs to go through HTTP endpoint. Could we not pass the healthcheck object to it that all `main()`s instantiate?

cmd/status/status.go
```diff
@@ -0,0 +1,69 @@
...
+	adminURL := ""
+	c := &cobra.Command{
+		Use:   "status",
+		Short: "Print the status.",
+		Long:  `Prints admin status information, exit non-zero on any error.`,
+		PreRunE: func(cmd *cobra.Command, args []string) (err error) {
+			rootCmd := cmd
+			for rootCmd.Parent() != nil {
+				rootCmd = rootCmd.Parent()
+			}
```
yurishkuro, line 42: what is the objective of this loop?

cmd/status/status_test.go
```diff
@@ -0,0 +1,42 @@
+// Copyright (c) 2019 The Jaeger Authors.
```
yurishkuro, line 1: 2020

yurishkuro: Was this automatically added?  We  might need to fix the script if so.

plugin/storage/es/mappings/gen_assets.go
```diff
@@ -213,7 +213,7 @@ var _escData = map[string]*_escFile{
 		name:    ".nocover",
 		local:   "plugin/storage/es/mappings/.nocover",
 		size:    43,
-		modtime: 1593482248,
+		modtime: 1595973361,
```
yurishkuro, line 216: no need to change this file

dekimsey: Agreed, these unrelated build components were purged but I I think I reintroduced them accidentally in a fixup. They are not, and would not be expected to me in the final submit.

go.mod
dekimsey: Unintended for sure.

cmd/status/status_test.go
dekimsey: I'm not sure how it was added. I started the branch a few weeks ago following the contributing doc. It is likely I copied it from somewhere.

cmd/status/status.go
dekimsey: Good question. This was left over from the initial commit where I selected the correct part based on the name of the command. So this would give me 'jaeger-agent'. To be fair, it's a bit cargo culty, as I looked at how the docs subcommand (I believe, on mobile can't check) worked and emulated it's behavior.

cmd/status/status.go
dekimsey: The purpose of the command is to provide an easy hook to make status checking possible in something like AWS ECS where only CMD are permitted and the Jaeger containers are running in empty environments (which is how they are distributed).
In essence enabling `docker exec -it jaeger-agent status` to report the currently running service's status. Since it's running as a different PID, I'm not clear how I could pass the healthcheck object to it and get that behavior. I did play around passing the object through, but it didn't expose the listening port and I still needed a url. If you think modifying the admin healthcheck interface is better, I'd be happy to do it. I just couldn't grasp how & when flags are being propagated.

yurishkuro: oh, you're right, never mind.

cmd/status/status.go
```diff
@@ -0,0 +1,69 @@
...
+				return fmt.Errorf("no default admin port available for %s: %w", cmd.Name(), err)
+			}
+			return nil
+		},
+		// FIXME: Normally I'd want to return an Error and let the exit handler do the os.Exit.
+		// But it's triggering --help output everytime and I don't know why.
+		Run: func(cmd *cobra.Command, args []string) {
+			adminURL, _ = adminServer.URL()
+			resp, err := http.Get(adminURL)
+			if err != nil {
```
yurishkuro, line 55: Note: `err != nil` will only happen if HTTP request cannot be made. But if the server is actually running but responding with unhealthy state, then `err == nil` and we need to check the returned status code.

yurishkuro: also, the health endpoint returns JSON like below, I think it would be useful to echo it to stdout:
```
$ curl http://localhost:14269/health
{"status":"Server available","upSince":"2020-08-26T04:05:13.974371436Z","uptime":"13.961416525s"}%
```

cmd/status/status.go
yurishkuro: it doesn't look like it's even used

cmd/status/status.go
```diff
@@ -0,0 +1,69 @@
...
+		Use:   "status",
+		Short: "Print the status.",
+		Long:  `Prints admin status information, exit non-zero on any error.`,
+		PreRunE: func(cmd *cobra.Command, args []string) (err error) {
+			rootCmd := cmd
+			for rootCmd.Parent() != nil {
+				rootCmd = rootCmd.Parent()
+			}
+			// FIXME: arg overrides aren't working, adminServer doesn't seem to see the current flagSet. Why?
+			adminURL, err = adminServer.URL()
```
yurishkuro, line 44: why not do this in the Run function? The `PreRunE` looks unnecessary altogether.

cmd/status/status.go
```diff
@@ -0,0 +1,69 @@
...
+			if err != nil {
+				return fmt.Errorf("no default admin port available for %s: %w", cmd.Name(), err)
+			}
+			return nil
+		},
+		// FIXME: Normally I'd want to return an Error and let the exit handler do the os.Exit.
+		// But it's triggering --help output everytime and I don't know why.
+		Run: func(cmd *cobra.Command, args []string) {
+			adminURL, _ = adminServer.URL()
```
yurishkuro, line 53: I think this would be just `http://localhost:14269/`, but we need `http://localhost:14269/health`

cmd/status/status.go
dekimsey: I was using it to find the flags or try to figure out where they were being set. I don't believe it has much value now or won't by the time it all functions.

cmd/status/status.go
dekimsey: My understanding of [adminServer.serveWithListener](https://github.com/jaegertracing/jaeger/blob/master/cmd/flags/admin.go#L119) is that the healthcheck is on `/`. Am I misreading this?

yurishkuro: ah, ok

cmd/status/status.go
```diff
@@ -0,0 +1,80 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package status
+
+import (
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"log"
```
pavolloffay, line 21: Can we use zap logger? IIRC it should be available when the command is created

cmd/status/status.go
```diff
@@ -0,0 +1,80 @@
...
+	c := &cobra.Command{
+		Use:   "status",
+		Short: "Print the status.",
+		Long:  `Prints admin status information, exit non-zero on any error.`,
+		PreRunE: func(cmd *cobra.Command, args []string) (err error) {
+			rootCmd := cmd
+			for rootCmd.Parent() != nil {
+				rootCmd = rootCmd.Parent()
+			}
+			// FIXME: arg overrides aren't working, adminServer doesn't seem to see the current flagSet. Why?
```
pavolloffay, line 44: maybe the flags are bind to a command. you might need to iterate to root cmd and get it from there? I am not sure though.

dekimsey: I need to poke at this some more. I'm pretty sure I'm holding it wrong, but I'm not an familiar with how flags work between go, cobra, viper, and then sub-commands sprinkeled on top of. Do you happen to know of any documentation on how those are supposed to function (cobra, viper in particular)?

cmd/flags/admin.go
```diff
@@ -151,3 +154,27 @@ func (s *AdminServer) registerPprofHandlers() {
 func (s *AdminServer) Close() error {
 	return s.server.Shutdown(context.Background())
 }
+
+// URL return the admin servers URL
+// FIXME: Parsing the s.adminHostPort should already be done, why do it again? Valid port
+// detection and the like should be handled during the argument parsing step. But I cannot figure out
+// how that is done (or if). Then should we drop any pretence of checking the values?
+func (s *AdminServer) URL() (string, error) {
```
dekimsey, line 162: Regardless of the decision made WRT parsing, `net.SplitHostPort` is probably a better choice than doing it manually.

jpkrohling:
What's the status of this one?

PR #2417, opened 2020-08-26
Allow configuration of ES tag keys directly through config/cli

joe-elliott:
- Fixes #2295 by providing a way to specify ES tag keys directly through the CLI/config file

- Added a new parameter `es.tags-as-fields.include`.  "include" was chosen to leave space later for "exclude", but I'd be willing to change it to whatever.
- Dropped `LoadTagsFromFile` in favor of `TagKeysAsFields`.  `TagKeysAsFields` returns a slice of tags keys that combines both the new cli option and the file.
- Added appropriate tests
- Added missing `.Tags` fields to `ApplyDefaults`.  Unsure if these were left out for a reason.

### DIFF
--- a/pkg/es/config/config.go
+++ b/pkg/es/config/config.go
@@ -221,7 +221,7 @@ func (c *Configuration) ApplyDefaults(source *Configuration) {
 	if !c.SnifferTLSEnabled {
 		c.SnifferTLSEnabled = source.SnifferTLSEnabled
 	}
-	if c.Tags.AllAsFields == false {
+	if !c.Tags.AllAsFields {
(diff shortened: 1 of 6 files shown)


pkg/es/config/config.go
```diff
@@ -77,6 +77,8 @@ type TagsAsFields struct {
 	DotReplacement string `mapstructure:"dot_replacement"`
 	// File path to tag keys which should be stored as object fields
 	File string `mapstructure:"config_file"`
+	// Comma Delimited list of tags to store as object fields
```
pavolloffay, line 80: s/Delimited/delimited

joe-elliott: fixed

PR #2429, opened 2020-08-28
Fix error equals

albertteoh:
Fixes a broken test introduced by an upgrade of go version to 1.15, caused by a [change in the time package](https://github.com/golang/go/commit/201cb046b745f8bb00e3d382290190c74ba7b7e1), specifically in [format.go:1437](https://github.com/golang/go/blob/18ea6f597c031d5f3c5160217db72d80cb62f689/src/time/format.go#L1437).

- Although the practice of asserting on error messages is debatable, the spirit of this PR is to respect the original intention of this test in making a strong assertion on the specific input value which causes the error. Further, the error message should be consistent between OSes, just not between go versions of course.
- Fixes the broken test, which does not expect a quoted value in the error message. Go 1.15 introduces a sensible change of quoting the values to more clearly identify what is being parsed.
- The fix involves modifying the test to accept either quoted (1.15) or unquoted values (1.14 or below) in a regex of the error message and hence, ensures compatibility between both versions.
- Alternatives considered:
  - Using `strings.HasPrefix` which is slightly simpler but slightly reduces the fidelity of the test, losing a comparison on the actual value.
  - Just `assert.Error(t, err)`, which is the simplest yet loses the most fidelity from the original test assertion.

Tested with the following to ensure they pass:
```
# Using go 1.14
$ go clean -testcache
$ go test github.com/jaegertracing/jaeger/cmd/query/app

# Switch to go 1.15
$ go clean -testcache
$ go test github.com/jaegertracing/jaeger/cmd/query/app
```

### DIFF
--- a/cmd/query/app/query_parser_test.go
+++ b/cmd/query/app/query_parser_test.go
@@ -16,7 +16,9 @@
 package app
 
 import (
+	"fmt"
 	"net/http"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


jpkrohling (approved):
LGTM. Ideally, we'd start wrapping the constant parts of the messages into the variable parts, making it easy to compare with `errors.Is`. That said, your PR improves the code base and prepares it for 1.15.

pavolloffay (approved):

cmd/query/app/query_parser_test.go
```diff
@@ -41,8 +44,8 @@ func TestParseTraceQuery(t *testing.T) {
 		{"x?service=service&start=string", errParseInt, nil},
 		{"x?service=service&end=string", errParseInt, nil},
 		{"x?service=service&limit=string", errParseInt, nil},
-		{"x?service=service&start=0&end=0&operation=operation&limit=200&minDuration=20", "cannot not parse minDuration: time: missing unit in duration 20", nil},
-		{"x?service=service&start=0&end=0&operation=operation&limit=200&minDuration=20s&maxDuration=30", "cannot not parse maxDuration: time: missing unit in duration 30", nil},
+		{"x?service=service&start=0&end=0&operation=operation&limit=200&minDuration=20", `cannot not parse minDuration: time: missing unit in duration "?20"?$`, nil},
+		{"x?service=service&start=0&end=0&operation=operation&limit=200&minDuration=20s&maxDuration=30", `cannot not parse maxDuration: time: missing unit in duration "?30"?$`, nil},
```
yurishkuro, line 48: Personally, having been burnt by the changing stdlib error messages in the past, I now try to write tests that only check for a substring or a prefix of the error which is stable and comes from our code, such as `cannot not parse minDuration`.

albertteoh: Yup, I would agree with that. Or, alternatively, wrap the error into a custom error type and assert on the type rather than error string comparisons.

PR #2434, opened 2020-08-30
Add context to dependencies endpoint

yoave23:
- resolves https://github.com/jaegertracing/jaeger/issues/2369 - auth token is not propagated in `/dependencies` endpoint

- add a `context.Context` param to the `GetDependencies` interface and implement accordingly


cmd/query/app/ui/placeholder/gen_assets.go
```diff
@@ -213,7 +213,7 @@ var _escData = map[string]*_escFile{
 		name:    "index.html",
```
yurishkuro, line 213: please exclude this file from the PR, it's not related

plugin/storage/es/dependencystore/storage.go
```diff
@@ -81,13 +81,13 @@ func (s *DependencyStore) writeDependencies(indexName string, ts time.Time, depe
...
 // GetDependencies returns all interservice dependencies
-func (s *DependencyStore) GetDependencies(endTs time.Time, lookback time.Duration) ([]model.DependencyLink, error) {
+func (s *DependencyStore) GetDependencies(ctx context.Context, endTs time.Time, lookback time.Duration) ([]model.DependencyLink, error) {
 	indices := getIndices(s.indexPrefix, endTs, lookback)
 	searchResult, err := s.client.Search(indices...).
 		Size(10000). // the default elasticsearch allowed limit
 		Query(buildTSQuery(endTs, lookback)).
 		IgnoreUnavailable(true).
-		Do(s.ctx)
+		Do(ctx)
```
yurishkuro, line 90: Good catch. Can we delete the field `s.ctx` altogether? I think it was implemented as anti-pattern.

plugin/storage/es/mappings/gen_assets.go
```diff
@@ -213,7 +213,7 @@ var _escData = map[string]*_escFile{
 		name:    ".nocover",
```
yurishkuro, line 213: please exclude this file

plugin/storage/es/dependencystore/storage.go
yoave23: deleted

PR #2436, opened 2020-08-31
Add context.Context to WriteSpan

yurishkuro:


cmd/collector/app/span_processor.go
```diff
@@ -140,7 +141,7 @@ func (sp *spanProcessor) saveSpan(span *model.Span) {
 	}
 
 	startTime := time.Now()
-	if err := sp.spanWriter.WriteSpan(span); err != nil {
+	if err := sp.spanWriter.WriteSpan(context.TODO(), span); err != nil {
```
yurishkuro, line 144: Using TODO here because there is a design issue with spanProcessor, which should be managing the context earlier

cmd/ingester/app/processor/span_processor.go
```diff
@@ -58,9 +59,9 @@ func NewSpanProcessor(params SpanProcessorParams) *KafkaSpanProcessor {
 
 // Process unmarshals and writes a single kafka message
 func (s KafkaSpanProcessor) Process(message Message) error {
-	mSpan, err := s.unmarshaller.Unmarshal(message.Value())
+	span, err := s.unmarshaller.Unmarshal(message.Value())
 	if err != nil {
 		return fmt.Errorf("cannot unmarshall byte array into span: %w", err)
 	}
-	return s.writer.WriteSpan(mSpan)
+	return s.writer.WriteSpan(context.TODO(), span)
```
yurishkuro, line 66: Using TODO here because KafkaSpanProcessor should be managing the context earlier.

cmd/opentelemetry/app/receiver/kafkareceiver/kafka_receiver.go
```diff
@@ -83,11 +83,12 @@ type writer struct {
...
 
 // WriteSpan writes a span to the next consumer.
-func (w writer) WriteSpan(span *model.Span) error {
+func (w writer) WriteSpan(ctx context.Context, span *model.Span) error {
 	batch := model.Batch{
 		Spans:   []*model.Span{span},
 		Process: span.Process,
 	}
 	traces := jaegertranslator.ProtoBatchToInternalTraces(batch)
+	// TODO what does w.ctx represent?
```
yurishkuro, line 92: @pavolloffay can we get rid of `w.ctx`?

storage/spanstore/interface.go
```diff
@@ -23,16 +23,16 @@ import (
 	"github.com/jaegertracing/jaeger/model"
 )
 
-// Writer writes spans to storage.
-type Writer interface {
-	WriteSpan(span *model.Span) error
-}
-
 var (
```
yurishkuro, line 26: constants must come before types

cmd/opentelemetry/app/receiver/kafkareceiver/kafka_receiver.go
pavolloffay: This is because of the metrics, check the line 38, Since the context is now propagated we should modify it here.

pavolloffay (commented):
The context in the OTEL kafka has to be modified.

cmd/opentelemetry/app/receiver/kafkareceiver/kafka_receiver.go
yurishkuro: If I use the `ctx` arg in L93, I don't see where else `w.ctx` would be used. Should `w.ctx` be removed completely then? Or do we need to enrich `ctx` here similar to L38 and L40?

cmd/collector/app/span_processor.go
annanay25: Yep, for complete context propagation context needs to be added to the `processor.SpanProcessor` interface, we can do that in a follow up PR.

plugin/storage/es/spanstore/writer.go
```diff
@@ -132,7 +132,7 @@ func getSpanAndServiceIndexFn(archive, useReadWriteAliases bool, prefix string)
 }
 
 // WriteSpan writes a span and its corresponding service:operation in ElasticSearch
-func (s *SpanWriter) WriteSpan(span *model.Span) error {
+func (s *SpanWriter) WriteSpan(ctx context.Context, span *model.Span) error {
```
annanay25, line 135: Can we mark this as `_ context.Context` if it is not being used right now? 

Also, `SpanWriter` has a `context` field which is probably redundant if `WriteSpan` can be called concurrently, IIRC it can.

cmd/opentelemetry/app/receiver/kafkareceiver/kafka_receiver.go
pavolloffay: yes the `w.ctx` can be removed and instead the context can be modified in the write method.

yurishkuro: ok, I made the changes, PTAL. But I think this needs to be revisited because it looks like a significant performance hit due to multiple memory allocations just to add some static labels. There should be a cheaper way of doing it, e.g. attaching a mutable structure to the Context at the start.

plugin/storage/es/spanstore/writer.go
yurishkuro: +1

cmd/opentelemetry/app/receiver/kafkareceiver/kafka_receiver.go
pavolloffay: Anyways this kafka receiver will go away and we will use the one from OTEL upsteream.

pavolloffay (approved):

PR #2438, opened 2020-08-31
Fix TLS flags settings in jaeger OTEL receiver

pavolloffay:

The GRPC TLS settings were not applied when `--collector.grpc.host-port=`  was not provided. This was causing issues in the operator as it does not set the endpoint.

### DIFF
--- a/cmd/opentelemetry/app/receiver/jaegerreceiver/jaeger_receiver.go
+++ b/cmd/opentelemetry/app/receiver/jaegerreceiver/jaeger_receiver.go
@@ -16,6 +16,7 @@ package jaegerreceiver
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/spf13/viper"
 	"go.opentelemetry.io/collector/component"
@@ -31,6 +32,7 @@ import (
 	grpcRep "github.com/jaegertracing/jaeger/cmd/agent/app/reporter/grpc"
 	collectorApp "github.com/jaegertracing/jaeger/cmd/collector/app"
 	"github.com/jaegertracing/jaeger/plugin/sampling/strategystore/static"
+	"github.com/jaegertracing/jaeger/ports"
 )
 
 // Factory wraps jaegerreceiver.Factory and makes the default config configurable via viper.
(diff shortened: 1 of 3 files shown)


cmd/opentelemetry/app/receiver/jaegerreceiver/jaeger_receiver.go
```diff
@@ -87,15 +89,22 @@ func configureCollector(v *viper.Viper, cfg *jaegerreceiver.Config) {
...
 			},
 		}
-		if cOpts.TLS.CertPath != "" && cOpts.TLS.KeyPath != "" {
-			cfg.GRPC.TLSSetting = &configtls.TLSServerSetting{
-				ClientCAFile: cOpts.TLS.ClientCAPath,
-				TLSSetting: configtls.TLSSetting{
-					KeyFile:  cOpts.TLS.KeyPath,
-					CertFile: cOpts.TLS.CertPath,
+	}
+	if cOpts.TLS.Enabled == true {
```
joe-elliott, line 93: can simplify `if cOpts.TLS.Enabled {`

Should enabling TLS through the Jaeger Collector options also turn on TLS for HTTP below?

pavolloffay:
> Final question. Since GRPC and Thrift HTTP are automatically enabled on the default jaeger collector should we always start the otel collector receivers?

They are started/configured by the `defaultconfig` package. I would not enable them here by default to keep the config in sync with the OTEL upstream. E.g. we might want to provide a distribution with the same default config as OTEL upstream https://github.com/jaegertracing/jaeger/issues/2281#issuecomment-641815507 (e.g the root command would behave like otelcol)

joe-elliott:
Ok, It took some unwinding, but I see what you're referring to.  It looks like the order is:

1. Start with OTEL defaults 
https://github.com/jaegertracing/jaeger/blob/5fb5e8ab5634559299c4955f93f6ab4fc807abbb/cmd/opentelemetry/app/receiver/jaegerreceiver/jaeger_receiver.go#L55

2. Apply Viper (Jaeger) Config Over
https://github.com/jaegertracing/jaeger/blob/5fb5e8ab5634559299c4955f93f6ab4fc807abbb/cmd/opentelemetry/app/receiver/jaegerreceiver/jaeger_receiver.go#L62

3. Fill out Jaeger Defaults
https://github.com/jaegertracing/jaeger/blob/5fb5e8ab5634559299c4955f93f6ab4fc807abbb/cmd/opentelemetry/app/defaultconfig/default_config.go#L127

I'm wondering if steps two and three should be swapped or even just start with a OTEL config that is Jaeger defaults.  For instance in this PR if `cfg.GRPC == nil` we are subbing in the default Jaeger GRPC settings which we then do again in step #3.

Obviously there is a lot of history here and I'm just playing catch up.  I will continue to think about this as I review the code/PRs!

pavolloffay:
> I'm wondering if steps two and three should be swapped. For instance in this PR if cfg.GRPC == nil we are subbing in the default Jaeger GRPC settings which we then do again in step 3.

Maybe it could be swapped or even merged together. Maybe at the default conf creation time we know what componnet (collector, agent...) is being started.

> Is there a default config for OTEL? Last time I checked the operator had to configure the basic pipeline.

There isn't there is also a proposal to have one, but It probably won't be merged. Maybe OTEL will provide a default config file with the distribution.

PR #2442, opened 2020-09-01
Fix mutex unlock in storage exporters

pavolloffay:


annanay25 (approved):
nice catch! :)

PR #2444, opened 2020-09-01
Use queue retry per exporter

pavolloffay:

This PR configures queue retry per exporter and removes the qretry form the default configuration. This is a prefered configuration since the global qretry would retry all exporters in the pipeline. See example ocnfig from OTEL https://github.com/open-telemetry/opentelemetry-collector/blob/master/exporter/otlpexporter/config.go#L26. 

Other notable changes:
* qretry is not defined for in-memory and kafka (we will swith to OTEL kafka impl soon)

### DIFF
--- a/cmd/opentelemetry/app/defaultconfig/default_config.go
+++ b/cmd/opentelemetry/app/defaultconfig/default_config.go
@@ -24,7 +24,6 @@ import (
 	"go.opentelemetry.io/collector/config/configmodels"
 	"go.opentelemetry.io/collector/config/confignet"
 	"go.opentelemetry.io/collector/processor/batchprocessor"
-	"go.opentelemetry.io/collector/processor/queuedprocessor"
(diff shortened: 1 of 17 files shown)


pavolloffay:
@joe-elliott would you like to review?

pavolloffay:
> Why not add this to the jaegerexporter? It seems like we'd want queueing and retries from agent -> collector

It already supports qretry, we just wrapping it and adding support for flags.

pavolloffay:
> In several places you note // TODO: Enable the queued settings by default.. Why not just leave the defaults alone and enable it now?

It is from https://github.com/open-telemetry/opentelemetry-collector/pull/1401/files#r457838535. I have created issue to track it https://github.com/open-telemetry/opentelemetry-collector/issues/1721 

I think we can enable it by default since it will be enabled anyways.

PR #2453, opened 2020-09-03
Configurable ES doc count

### LINKED ISSUE #2437: ES: Configurable query result sizes

We limit the number of results returned from Elasticsearch (ES) from user queries to prevent large queries from negatively impacting the ES cluster. With ES as our Jaeger backing store, we also need to ensure jaeger-query -> ES requests abide by these limits, ideally as a configurable parameter.

Currently, `defaultDocCount` is applied to the ES query's `size` attribute and I was not able to find a means of overriding this value. Examples:

https://github.com/jaegertracing/jaeger/search?q=defaultDocCount&unscoped_q=defaultDocCount

In some cases, this value is hard-coded directly instead of using the const:
https://github.com/jaegertracing/jaeger/blob/3f6f3a736a8feee40939ba316e7481cb1ba45f04/plugin/storage/es/dependencystore/storage.go#L87

Resulting in the need for less desirable workarounds to ensure requests from jaeger-query to ES set the appropriate result size limit.

Introduce a new integer flag, perhaps `es.doc-count` (`ES_DOC_COUNT` env var) for consistency, that represents an override to the 10000 `defaultDocCount` const. Open to suggestions on better parameter names as well. 

- Given 10,000 is applied across [service](https://github.com/jaegertracing/jaeger/blob/5ed586c114152dfdb9f9643b9303fd40d45cc6e5/plugin/storage/es/spanstore/service_operation.go#L106)/[operation](https://github.com/jaegertracing/jaeger/blob/5ed586c114152dfdb9f9643b9303fd40d45cc6e5/plugin/storage/es/spanstore/service_operation.go#L137) aggregations and [dependency store](https://github.com/jaegertracing/jaeger/blob/3f6f3a736a8feee40939ba316e7481cb1ba45f04/plugin/storage/es/dependencystore/storage.go#L87) (and also in OTEL), does it make sense to have the one configurable variable in all these places? Or should we separate, aggregation and dependency store doc counts?

- The `es.max-num-spans` (used [here](https://github.com/jaegertracing/jaeger/search?q=maxnumspans&unscoped_q=maxnumspans)) also defaults to a value of 10,000. Please let me know if there is any desire to consolidate this flag with the proposed `es.doc-count`.

albertteoh:
- Closes #2437

- Add deprecation note for `--es.max-num-spans`. (Will create another issue to decommission this flag for a future release).
- Add new flag `--es.max-doc-count`, that configures query (and aggregation) doc sizes, defaulting to `10_000`. Name based on existing `defaultDocCount` variable for consistency.
- Fix tests which were incorrectly checking for the document size at the query level. 
- Add test to assert on aggregation sizes applied.

### DIFF
--- a/cmd/opentelemetry/app/internal/reader/es/esdependencyreader/dependency_store.go
+++ b/cmd/opentelemetry/app/internal/reader/es/esdependencyreader/dependency_store.go
@@ -36,10 +36,7 @@ const (
 
 	timestampField = "timestamp"
 
-	// default number of documents to return from a query (elasticsearch allowed limit)
-	// see search.max_buckets and index.max_result_window
-	defaultMaxDocCount = 10_000
-	indexDateFormat    = "2006-01-02" // date format for index e.g. 2020-01-20
+	indexDateFormat = "2006-01-02" // date format for index e.g. 2020-01-20
 )
 
 // DependencyStore defines Elasticsearch dependency store.
@@ -47,13 +44,14 @@ type DependencyStore struct {
 	client      esclient.ElasticsearchClient
 	logger      *zap.Logger
 	indexPrefix string
+	maxDocCount int
 }
 
 var _ dependencystore.Reader = (*DependencyStore)(nil)
 var _ dependencystore.Writer = (*DependencyStore)(nil)
 
 // NewDependencyStore creates dependency store.
-func NewDependencyStore(client esclient.ElasticsearchClient, logger *zap.Logger, indexPrefix string) *DependencyStore {
+func NewDependencyStore(client esclient.ElasticsearchClient, logger *zap.Logger, indexPrefix string, maxDocCount int) *DependencyStore {
 	if indexPrefix != "" {
 		indexPrefix += "-"
 	}
@@ -61,6 +59,7 @@ func NewDependencyStore(client esclient.ElasticsearchClient, logger *zap.Logger,
 		client:      client,
 		logger:      logger,
 		indexPrefix: indexPrefix + dependencyIndexBaseName + "-",
+		maxDocCount: maxDocCount,
 	}
 }
 
@@ -84,10 +83,10 @@ func (r *DependencyStore) WriteDependencies(ts time.Time, dependencies []model.D
 
 // GetDependencies implements dependencystore.Reader
 func (r *DependencyStore) GetDependencies(ctx context.Context, endTs time.Time, lookback time.Duration) ([]model.DependencyLink, error) {
-	searchBody := getSearchBody(endTs, lookback)
+	searchBody := getSearchBody(endTs, lookback, r.maxDocCount)
 
 	indices := dailyIndices(r.indexPrefix, endTs, lookback)
-	response, err := r.client.Search(ctx, searchBody,defaultMaxDocCount, indices...)
+	response, err := r.client.Search(ctx, searchBody, r.maxDocCount, indices...)
 	if err != nil {
 		return nil, err
 	}
@@ -106,12 +105,12 @@ func (r *DependencyStore) GetDependencies(ctx context.Context, endTs time.Time, 
 	return dbmodel.ToDomainDependencies(dependencies), nil
 }
 
-func getSearchBody(endTs time.Time, lookback time.Duration) esclient.SearchBody {
+func getSearchBody(endTs time.Time, lookback time.Duration, maxDocCount int) esclient.SearchBody {
 	return esclient.SearchBody{
 		Query: &esclient.Query{
 			RangeQueries: map[string]esclient.RangeQuery{timestampField: {GTE: endTs.Add(-lookback), LTE: endTs}},
 		},
-		Size: defaultMaxDocCount,
+		Size: maxDocCount,
 	}
 }
 
--- a/cmd/opentelemetry/app/internal/reader/es/esdependencyreader/dependency_store.go
+++ b/cmd/opentelemetry/app/internal/reader/es/esdependencyreader/dependency_store.go
@@ -36,7 +36,7 @@ const (
 
 	timestampField = "timestamp"
 
-	// default number of documents to fetch in a query
+	// default number of documents to return from a query (elasticsearch allowed limit)
 	// see search.max_buckets and index.max_result_window
 	defaultMaxDocCount = 10_000
 	indexDateFormat    = "2006-01-02" // date format for index e.g. 2020-01-20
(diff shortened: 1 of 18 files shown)


pavolloffay:
@albertteoh could you please fix the commit history, it seems that the PR contains other commits.

albertteoh:
> @albertteoh could you please fix the commit history, it seems that the PR contains other commits.

Thanks for pointing that out @pavolloffay, I must have messed up the rebase. I've cleaned this up so commit history should only show my changes.

cmd/opentelemetry/app/internal/reader/es/esdependencyreader/dependency_store.go
```diff
@@ -87,7 +87,7 @@ func (r *DependencyStore) GetDependencies(ctx context.Context, endTs time.Time,
 	searchBody := getSearchBody(endTs, lookback)
 
 	indices := dailyIndices(r.indexPrefix, endTs, lookback)
-	response, err := r.client.Search(ctx, searchBody, defaultDocCount, indices...)
+	response, err := r.client.Search(ctx, searchBody,defaultMaxDocCount, indices...)
```
pavolloffay, line 90: Could we make the `defaultMaxDocCount` configurable here as well?

plugin/storage/es/factory.go
```diff
@@ -186,6 +186,11 @@ func createSpanWriter(
 	return writer, nil
 }
 
+func createDependencyReader(client es.Client, logger *zap.Logger, cfg config.ClientBuilder) (dependencystore.Reader, error) {
```
pavolloffay, line 189: Thy introducing this method? It's oneliner, the line 190 can return directly.

pkg/es/config/config.go
```diff
@@ -49,7 +49,7 @@ type Configuration struct {
 	AllowTokenFromContext bool           `mapstructure:"-"`
 	Sniffer               bool           `mapstructure:"sniffer"` // https://github.com/olivere/elastic/wiki/Sniffing
 	SnifferTLSEnabled     bool           `mapstructure:"sniffer_tls_enabled"`
-	MaxNumSpans           int            `mapstructure:"-"`                     // defines maximum number of spans to fetch from storage per query
+	MaxNumSpans           int            `mapstructure:"-"`                     // deprecated: use MaxDocCount instead. Defines maximum number of spans to fetch from storage per query
```
pavolloffay, line 52: This can be renamed to `MaxDocCount` and we need a test case when both flags are specified to show that the old flag works and also that the new one takes precedence.

cmd/opentelemetry/app/internal/reader/es/esdependencyreader/dependency_store.go
albertteoh: Good catch, I missed this one.

pkg/es/config/config.go
albertteoh: Okay, I misunderstood your comment in the Issue discussion, and thought you preferred to leave `MaxNumSpans` untouched while in a deprecated state. I've updated the diff to replace `MaxNumSpans` -> `MaxDocCount` in all places except for the user-facing flag.

plugin/storage/es/factory.go
albertteoh: The intention was to maintain consistency with how SpanWriters/Readers were created as well, but agree it's an unnecessary change.

plugin/storage/es/options.go
```diff
@@ -173,8 +178,8 @@ func addFlags(flagSet *flag.FlagSet, nsConfig *namespaceConfig) {
 		"The maximum lookback for spans in Elasticsearch")
 	flagSet.Int(
 		nsConfig.namespace+suffixMaxNumSpans,
-		nsConfig.MaxNumSpans,
-		"The maximum number of spans to fetch at a time per query in Elasticsearch")
+		nsConfig.MaxDocCount,
+		"(deprecated, will be removed in release v1.21.0. Please use es.max-doc-count). The maximum number of spans to fetch at a time per query in Elasticsearch")
```
pavolloffay, line 182: could you please create an issue for this to remove it?

plugin/storage/es/options_test.go
```diff
@@ -95,3 +95,29 @@ func TestOptionsWithFlags(t *testing.T) {
...
+	testCases := []struct {
+		name            string
+		flags           []string
+		wantMaxDocCount int
+	}{
+		{"neither defined", []string{}, 10_000},
+		{"max-num-spans only", []string{"--es.max-num-spans=1000"}, 1000},
+		{"max-doc-count only", []string{"--es.max-doc-count=1000"}, 1000},
+		{"max-num-spans == max-doc-count", []string{"--es.max-num-spans=1000", "--es.max-doc-count=1000"}, 1000},
+		{"max-num-spans < max-doc-count", []string{"--es.max-num-spans=999", "--es.max-doc-count=1000"}, 999},
```
pavolloffay, line 109: Is always the lower number returned?

pavolloffay (approved):

pavolloffay:
@albertteoh could you please increase the test coverage? https://codecov.io/gh/jaegertracing/jaeger/pull/2453/diff

plugin/storage/es/options.go
```diff
@@ -279,6 +287,18 @@ func initFromViper(cfg *namespaceConfig, v *viper.Viper) {
 	cfg.Enabled = v.GetBool(cfg.namespace + suffixEnabled)
 	cfg.CreateIndexTemplates = v.GetBool(cfg.namespace + suffixCreateIndexTemplate)
 	cfg.Version = uint(v.GetInt(cfg.namespace + suffixVersion))
+
+	maxNumSpans := v.GetInt(cfg.namespace + suffixMaxNumSpans)
+	cfg.MaxDocCount = v.GetInt(cfg.namespace + suffixMaxDocCount)
+
+	if maxNumSpans != 0 {
```
pavolloffay, line 294: nit: there is also `viper.IsSet`. it might help here.

pavolloffay:
One more thing @albertteoh we should add this change to changelog in braking changes since the limit is applied to new set of queries - we should enumerate what queries from the UI perspective might be affected.

albertteoh (commented):
> One more thing @albertteoh we should add this change to changelog in braking changes since the limit is applied to new set of queries - we should enumerate what queries from the UI perspective might be affected.

Good point, I've added a note in the changelog for v1.20 (unreleased). Please let me know if anything in that changelog entry requires modification.

Thanks for reviewing!

pavolloffay (approved):

PR #1814, opened 2019-09-24
Handling of expected error codes coming from grpc storage plugins #1741

chandresh-pancholi:
Signed-off-by: chandresh-pancholi <chandreshpancholi007@gmail.com>

## Which problem is this PR solving?
- #1741 

## Short description of the changes
- Changing the error code

Issue #1741, opened 2019-08-19
Handling of expected error codes coming from grpc storage plugins


sergiimk:
## Requirement - what kind of business use case are you trying to solve?

We are implementing a custom gRPC-based storage plugin as per [this doc](https://github.com/jaegertracing/jaeger/blob/master/plugin/storage/grpc/README.md).

## Problem - what in Jaeger blocks you from solving the requirement?

When searching in jaeger UI for a **non-existing** trace ID the `GetTrace` function in the storage plugin is [expected](https://github.com/jaegertracing/jaeger/blob/master/plugin/storage/memory/memory.go#L157) to return `spanstore.ErrTraceNotFound`. However the error handling on the Jaeger side doesn't seem to handle this error correctly and treats it as a generic internal error:

```
{"level":"error","ts":1566246187.758402,"caller":"app/http_handler.go:409","msg":"HTTP handler, Internal Server Error","error":"grpc stream error: rpc error: code = Unknown desc = trace was not found", ...
```

UI displays:
```
HTTP Error: grpc stream error: rpc error: code = Unknown desc = trace was not found
```

You can easily reproduce this issue by running Jaeger with the [example in-memory storage gRPC plugin](https://github.com/jaegertracing/jaeger/tree/master/examples/memstore-plugin) and searching for a trace.

Impact: cosmetic in this case, but other expected error types returned by gRPC plugin may be affected as well.

Abhi-khandelwal:
Hi @yurishkuro , Can I work on this issue ?

yurishkuro:
@Abhi-khandelwal by all means

chandresh-pancholi:
@sergiimk I will be picking this up as there is no activity on this thread.

chandresh-pancholi:
@sergiimk Could you please help me out here with the reproduction of this issue?

sergiimk:
@chandresh-pancholi you can reproduce this by compiling the [memstore-plugin](https://github.com/jaegertracing/jaeger/tree/master/examples/memstore-plugin) and running Jaeger with it as described [here](https://github.com/jaegertracing/jaeger/blob/master/plugin/storage/grpc/README.md#running-with-a-plugin). Searching for a random traceID in the UI will then return an ugly-looking HTTP error.

The sequence of commands should look roughly like this:
```
$ go build -o memstore-plugin examples/memstore-plugin/main.go
$ SPAN_STORAGE_TYPE="grpc-plugin" \
[...]

PR #2455, opened 2020-09-03
Propagate TraceNotFound error from grpc storage plugins

joe-elliott:
A previous attempt to fix this was made here:  https://github.com/jaegertracing/jaeger/pull/1814
Original issue: https://github.com/jaegertracing/jaeger/issues/1741

However, this is currently not working b/c on this line `ok` is set to true which skips the code that checks to see if the string equals "trace not found".
https://github.com/jaegertracing/jaeger/blob/412022065959c6a45e6250a733b081c387e9fdea/plugin/storage/grpc/shared/grpc_client.go#L80

Tracking it down:
**RecvMsg**
https://github.com/grpc/grpc-go/blob/8630cac324bf02ea0edfba758c2b2f3344af7bf7/stream.go#L749
**recvMsg**
https://github.com/grpc/grpc-go/blob/master/stream.go#L886
**toRPCErr**
https://github.com/grpc/grpc-go/blob/8630cac324bf02ea0edfba758c2b2f3344af7bf7/rpc_util.go#L775

`toRPCErr` calls `status.FromError` which wraps the error in a `GRPCStatus`.  This causes the `FromError` in `grpc_client.go` to recognize the error as `GRPCStatus` and return `true`.

Fix:
- Remove the `status.FromError` check and test the error directly.
- Compare the error message to a `GRPCStatus` wrapped `spanstore.ErrTraceNotFound`

I have confirmed this works manually by building jaeger-query with a custom GRPC plugin.

### DIFF
--- a/plugin/storage/grpc/shared/grpc_client.go
+++ b/plugin/storage/grpc/shared/grpc_client.go
@@ -18,10 +18,10 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"strings"
 	"time"
 
 	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/status"
 
 	"github.com/jaegertracing/jaeger/model"
 	"github.com/jaegertracing/jaeger/proto-gen/storage_v1"
@@ -77,7 +77,8 @@ func (c *grpcClient) GetTrace(ctx context.Context, traceID model.TraceID) (*mode
 	trace := model.Trace{}
 	for received, err := stream.Recv(); err != io.EOF; received, err = stream.Recv() {
 		if err != nil {
-			if strings.Contains(err.Error(), spanstore.ErrTraceNotFound.Error()) {
+			traceNotFoundStatus, _ := status.FromError(spanstore.ErrTraceNotFound)
+			if err.Error() == traceNotFoundStatus.Err().Error() {
 				return nil, spanstore.ErrTraceNotFound
 			}
 			return nil, fmt.Errorf("grpc stream error: %w", err)
--- a/plugin/storage/grpc/shared/grpc_client.go
+++ b/plugin/storage/grpc/shared/grpc_client.go
@@ -18,10 +18,10 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"strings"
 	"time"
 
 	"google.golang.org/grpc/metadata"
-	"google.golang.org/grpc/status"
 
 	"github.com/jaegertracing/jaeger/model"
 	"github.com/jaegertracing/jaeger/proto-gen/storage_v1"
@@ -77,10 +77,8 @@ func (c *grpcClient) GetTrace(ctx context.Context, traceID model.TraceID) (*mode
 	trace := model.Trace{}
 	for received, err := stream.Recv(); err != io.EOF; received, err = stream.Recv() {
 		if err != nil {
-			if e, ok := status.FromError(err); !ok {
-				if e.Message() == spanstore.ErrTraceNotFound.Error() {
-					return nil, spanstore.ErrTraceNotFound
-				}
+			if strings.Contains(err.Error(), spanstore.ErrTraceNotFound.Error()) {
+				return nil, spanstore.ErrTraceNotFound
 			}
 			return nil, fmt.Errorf("grpc stream error: %w", err)
 		}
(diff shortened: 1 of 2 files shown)


plugin/storage/grpc/shared/grpc_client.go
```diff
@@ -77,10 +77,9 @@ func (c *grpcClient) GetTrace(ctx context.Context, traceID model.TraceID) (*mode
 	trace := model.Trace{}
 	for received, err := stream.Recv(); err != io.EOF; received, err = stream.Recv() {
 		if err != nil {
-			if e, ok := status.FromError(err); !ok {
-				if e.Message() == spanstore.ErrTraceNotFound.Error() {
-					return nil, spanstore.ErrTraceNotFound
-				}
+			traceNotFoundStatus, _ := status.FromError(spanstore.ErrTraceNotFound)
+			if err.Error() == traceNotFoundStatus.Err().Error() {
```
yurishkuro, line 81: `traceNotFoundStatus.Err().Error()` prints as:
```
rpc error: code = Unknown desc = trace not found
```
I wonder if there's a better way than comparing strings. Especially suspicious is `code = Unknown`.

yurishkuro (approved):

plugin/storage/grpc/shared/grpc_client_test.go
```diff
@@ -200,9 +201,11 @@ func TestGRPCClientGetTrace_NoTrace(t *testing.T) {
 }
 
 func TestGRPCClientGetTrace_StreamErrorTraceNotFound(t *testing.T) {
+	s, _ := status.FromError(spanstore.ErrTraceNotFound)
+
 	withGRPCClient(func(r *grpcClientTest) {
 		traceClient := new(grpcMocks.SpanReaderPlugin_GetTraceClient)
-		traceClient.On("Recv").Return(nil, spanstore.ErrTraceNotFound)
+		traceClient.On("Recv").Return(nil, s.Err())
```
yurishkuro, line 208: I think we should write a test that goes through client and server, not just mocking the client.

plugin/storage/grpc/shared/grpc_client.go
joe-elliott: Agree, pushed something that i like better.  It's still string comparison, but it is now comparing directly "trace not found".

The original error was created in a different process so I don't think there's any clean way to compare error equality.  I believe string comparison is as good as it gets here.

plugin/storage/grpc/shared/grpc_client.go
```diff
@@ -77,10 +77,9 @@ func (c *grpcClient) GetTrace(ctx context.Context, traceID model.TraceID) (*mode
 	trace := model.Trace{}
 	for received, err := stream.Recv(); err != io.EOF; received, err = stream.Recv() {
 		if err != nil {
-			if e, ok := status.FromError(err); !ok {
```
yurishkuro, line 80: wasn't this if statement mostly correct, except the condition should be `ok` instead of `!ok`?

joe-elliott: Yup, that's true.  I prefer what I have pushed b/c it doesn't matter if `err` is an `error` or a `status.Status` if the message is "trace not found" then it will return the correct error.

We can do the minimal change if it's preferred though.

yurishkuro: the original version (aside from !ok) seems better, because your version is ignoring the error, so `s` may be undefined.

joe-elliott: good point.  I am relying on `status.FromError` to return a non-nil status.  I think this would the best construction

```
if s, _ := status.FromError(err); s != nil {
  // check if s is a trace not found error.
}
```

`ok` just indicates if `FromError` had to create a new `status.Status` or if it was able to retrieve one using the `GRPCStatus()` method.  For our purposes we don't care, we're just going to compare `s.Message()` to "trace not found" either way.

plugin/storage/grpc/shared/grpc_client_test.go
joe-elliott: Agree.  I can't say I have time right now to knock it out.  Perhaps if i find some time in the next week or two I'll dig into it.

yurishkuro (approved):

PR #2456, opened 2020-09-04
Add partial retry capability to OTEL ES exporter.

pavolloffay:
The partial retry as the name suggests ensures that only spans that failed to be stored were retried. The returned error encapsulates `pdata.Traces` that is used by qretry on in the consecutive call to the exporter. The `pdata.Traces`  for retry is created from a struct that is created by ES model translator.

### DIFF
--- a/cmd/opentelemetry/app/exporter/elasticsearchexporter/spanstore.go
+++ b/cmd/opentelemetry/app/exporter/elasticsearchexporter/spanstore.go
@@ -37,6 +37,7 @@ import (
 	"github.com/jaegertracing/jaeger/model"
 	"github.com/jaegertracing/jaeger/pkg/cache"
 	"github.com/jaegertracing/jaeger/pkg/es/config"
+	"github.com/jaegertracing/jaeger/pkg/multierror"
 	"github.com/jaegertracing/jaeger/plugin/storage/es/spanstore/dbmodel"
 )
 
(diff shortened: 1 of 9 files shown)


pavolloffay:
@joe-elliott would you like to have a look at this?

joe-elliott (commented):
So the logic seems generally fine, but I'm concerned about the increased memory/cpu usage under normal operation due to creating the `ConvertedData` structs.

I'm also concerned about failure scenarios.  If a large number of spans start failing for any reason then `bulkItemsToTraces` will start being called extremely rapidly putting additional memory/GC/cpu pressure on this process.  I'm concerned a failing Elasticsearch cluster could then cause a cascading failure into the Jaeger collector/ingester.

So it seems that you primarily are using the `ConvertedData` struct to rebuild the trace to return in the `PartialTracesError` as accurately as possible.  However, I don't think you need all this data.  The data we actually need to store is already in the span (span data + process).  Using the `dbmodel.Span` alone I think we could rebuild enough of a trace to return in `PartialTracesError` that the next time it it came into the exporter it would be sufficient to push it again.  

For instance, the `InstrumentationLibrarySpan.InstrumentationLibrary` data is being kept in order to rebuild the trace, but the Jaeger exporter doesn't do anything with this data.

Basically I'm proposing putting a partial/fake trace back into `PartialTracesError` derived from `dbmodel.Span` only.  We should be able to rebuild the trace accurately enough that when it comes back into the ES exporter we will save the data correctly.

pavolloffay:
> So the logic seems generally fine, but I'm concerned about the increased memory/cpu usage under normal operation due to creating the ConvertedData structs.

The `ConvertedData` struct should be light weight and holds references to already allocated objects (ES model span and structs from pdata).

> I'm also concerned about failure scenarios. If a large number of spans start failing for any reason then bulkItemsToTraces will start being called extremely rapidly putting additional memory/GC/cpu pressure on this process. I'm concerned a failing Elasticsearch cluster could then cause a cascading failure into the Jaeger collector/ingester.

> So it seems that you primarily are using the ConvertedData struct to rebuild the trace to return in the PartialTracesError as accurately as possible. However, I don't think you need all this data. The data we actually need to store is already in the span (span data + process). Using the dbmodel.Span alone I think we could rebuild enough of a trace to return in PartialTracesError that the next time it it came into the exporter it would be sufficient to accurately save it to ES.

Rebuilding `pdata.Traces` from `esmodel.Span` would be more CPU/memory heavy. The idea behind `ConvertedData` is to keep references to already existing objects so in case of failure the remap is as simple as possible (which reminds me that instead of using `CopyTo` in `bulkItemsToTraces` we should use append where possible). 

> For instance, the InstrumentationLibrarySpan.InstrumentationLibrary data is being kept in order to rebuild the trace, but the Jaeger exporter doesn't do anything with this data.

right now it's not being converted to ES span model, but it's a pending feature that should be implemented. We need all original data (resource, instrumentation and span) to correctly reconstruct ES model span.

joe-elliott:
> The ConvertedData struct should be light weight and holds references to already allocated objects (ES model span and structs from pdata).

I see and agree :+1:.

> Rebuilding pdata.Traces from esmodel.Span would be more CPU/memory heavy. The idea behind ConvertedData is to keep references to already existing objects so in case of failure the remap is as simple as possible (which reminds me that instead of using CopyTo in bulkItemsToTraces we should use append where possible).

Was about to comment about the `CopyTo()` reviewing the code.  It is allocating new objects.  Agree on appending.

> right now it's not being converted to ES span model, but it's a pending feature that should be implemented. We need all original data (resource, instrumentation and span) to correctly reconstruct ES model span.

Agree with this as well.

cmd/opentelemetry/app/exporter/elasticsearchexporter/spanstore.go
```diff
@@ -216,11 +223,36 @@ func hashCode(serviceName, operationName string) string {
...
 }
+
+func bulkItemsToTraces(bulkItems []bulkItem) pdata.Traces {
+	traces := pdata.NewTraces()
+	for _, op := range bulkItems {
+		spanData := op.spanData
+		if traces.ResourceSpans().Len() == 0 {
+			traces.ResourceSpans().Resize(1)
+		} else {
+			traces.ResourceSpans().Resize(traces.ResourceSpans().Len())
```
joe-elliott, line 242: Does this do anything?  

Should we just set `traces.ResourceSpans().Resize(len(bulkItems))` before the loop and make each bulk item its own ResourceSpans?  This would be the simplest way to do it.

Otherwise I think you'd need a map of `Resource` => `ResourceSpans` which you could check for every `bulkItem` to see if a `ResourceSpan` for this `Resource` already existed and append to it.

cmd/opentelemetry/app/exporter/elasticsearchexporter/spanstore.go
```diff
@@ -216,11 +223,36 @@ func hashCode(serviceName, operationName string) string {
...
+		}
+		rss := traces.ResourceSpans().At(traces.ResourceSpans().Len() - 1)
+		if !spanData.Resource.IsNil() {
+			spanData.Resource.CopyTo(rss.Resource())
+		}
+		rss.InstrumentationLibrarySpans().Resize(1)
```
joe-elliott, line 248: Similar to above, you could keep some kind of map of Resource/InstrumentationLibrary => InstrumentationLibrarySpans which you'd then check to see if you had an InstrumentationLibrarySpans to append to or if you should just make a new one.

Might be faster to just make one ResourceSpans/InstrumentationLibrarySpans for each bulkitem and forget trying to reuse them.  It's hard to say if spans in a batch will likely come from the same trace and this will be worthwhile or not.   Probably depends on which processors are configured in the pipeline.

cmd/opentelemetry/app/exporter/elasticsearchexporter/spanstore.go
pavolloffay: +1 it's better to allocate at the beginning.

cmd/opentelemetry/app/exporter/elasticsearchexporter/spanstore.go
pavolloffay: We cannot expect that spans in a batch share the same resource&instrumentation.

> Might be faster to just make one ResourceSpans/InstrumentationLibrarySpans for each bulkitem and forget trying to reuse them. It's hard to say if spans in a batch will likely come from the same trace and this will be worthwhile or not. Probably depends on which processors are configured in the pipeline.

Keeping the resource/instrumentation -> spans would be more memory efficient. However, I would ho with this simple approach for now. The instrumentation library has two string fields and resource has a map.

pavolloffay:
@joe-elliott could you please have a look again?

cmd/opentelemetry/app/exporter/elasticsearchexporter/spanstore.go
joe-elliott: Agreed +1

joe-elliott (approved):
`bulkItemsToTraces` is difficult code to write in a performant way particurlarly given the generated shim you're working through to access the proto.

Without going to more extremes (such as the already discussed) I think this is as good as can be done.

PR #2458, opened 2020-09-05
Consolidated OTel ES IndexNameProviders

joe-elliott:
There were two `indexNameProvider`s with very similar code.  One for reading and one for writing.  These have been consolidated into one provider in the `internal/esutil` package

- Consolidated `IndexNameProvider`s and tests


joe-elliott:
@pavolloffay If you have a chance to look at this I would appreciate it.

cmd/opentelemetry/app/internal/esclient/index_name.go
```diff
@@ -12,39 +12,54 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package esspanreader
+package esclient
```
pavolloffay, line 15: It should go to a different package than `esclient`. The `esclient` is dedicated only for the ES client used by Jaeger writer/reader.

pavolloffay (commented):
It looks good but the package should be changed. I am fine to keep it in reader or exporter packages or move it to internal under something else than `esclient`.

cmd/opentelemetry/app/internal/esclient/index_name.go
joe-elliott: moved to  `./internal/esutil`

pavolloffay (approved):

PR #2461, opened 2020-09-09
Regenerate certificates to use SANs instead of Common Name

### LINKED ISSUE #2435: Go 1.15 compatiblity
**Describe the bug**

Even after a couple of fixes (#2401, #2429), there are more issues related to compatibility with Go 1.15 

**To Reproduce**
1. `brew install go@1.15`
2. `make test`

**Expected behavior**
All tests pass

**Screenshots**
These tests fail, possibly more:
```
=== RUN   TestProxyClientTLS/should_pass_with_TLS_client_to_trusted_TLS_server_with_correct_hostname
    builder_test.go:342:
        	Error Trace:	builder_test.go:342
        	Error:      	Received unexpected error:
        	            	rpc error: code = Unavailable desc = connection error: desc = "transport: authentication handshake failed: x509: certificate relies on legacy Common Name field, use SANs or temporarily enable Common Name matching with GODEBUG=x509ignoreCN=0"
        	Test:       	TestProxyClientTLS/should_pass_with_TLS_client_to_trusted_TLS_server_with_correct_hostname
=== RUN   TestProxyClientTLS/should_fail_with_TLS_client_without_cert_to_trusted_TLS_server_requiring_cert
=== RUN   TestProxyClientTLS/should_fail_with_TLS_client_without_cert_to_trusted_TLS_server_requiring_cert_from_a_different_CA
=== RUN   TestProxyClientTLS/should_pass_with_TLS_client_with_cert_to_trusted_TLS_server_requiring_cert
    builder_test.go:342:
        	Error Trace:	builder_test.go:342
        	Error:      	Received unexpected error:
        	            	rpc error: code = Unavailable desc = connection error: desc = "transport: authentication handshake failed: x509: certificate relies on legacy Common Name field, use SANs or temporarily enable Common Name matching with GODEBUG=x509ignoreCN=0"
        	Test:       	TestProxyClientTLS/should_pass_with_TLS_client_with_cert_to_trusted_TLS_server_requiring_cert
--- FAIL: TestProxyClientTLS (0.15s)
    --- PASS: TestProxyClientTLS/should_pass_with_insecure_grpc_connection (0.01s)
    --- PASS: TestProxyClientTLS/should_fail_with_TLS_client_to_non-TLS_server (0.12s)
    --- PASS: TestProxyClientTLS/should_fail_with_TLS_client_to_untrusted_TLS_server (0.00s)
    --- PASS: TestProxyClientTLS/should_fail_with_TLS_client_to_trusted_TLS_server_with_incorrect_hostname (0.00s)
    --- FAIL: TestProxyClientTLS/should_pass_with_TLS_client_to_trusted_TLS_server_with_correct_hostname (0.00s)
    --- PASS: TestProxyClientTLS/should_fail_with_TLS_client_without_cert_to_trusted_TLS_server_requiring_cert (0.00s)
    --- PASS: TestProxyClientTLS/should_fail_with_TLS_client_without_cert_to_trusted_TLS_server_requiring_cert_from_a_different_CA (0.00s)
    --- FAIL: TestProxyClientTLS/should_pass_with_TLS_client_with_cert_to_trusted_TLS_server_requiring_cert (0.00s)
=== RUN   TestMultipleCollectors
--- PASS: TestMultipleCollectors (0.00s)
=== RUN   TestBindFlags
--- PASS: TestBindFlags (0.00s)
=== RUN   TestReporter_EmitZipkinBatch
--- PASS: TestReporter_EmitZipkinBatch (0.00s)
=== RUN   TestReporter_EmitBatch
--- PASS: TestReporter_EmitBatch (0.00s)
=== RUN   TestReporter_SendFailure
--- PASS: TestReporter_SendFailure (0.00s)
=== RUN   TestReporter_AddProcessTags_EmptyTags
--- PASS: TestReporter_AddProcessTags_EmptyTags (0.00s)
=== RUN   TestReporter_AddProcessTags_ZipkinBatch
--- PASS: TestReporter_AddProcessTags_ZipkinBatch (0.00s)
=== RUN   TestReporter_AddProcessTags_JaegerBatch
--- PASS: TestReporter_AddProcessTags_JaegerBatch (0.00s)
=== RUN   TestReporter_MakeModelKeyValue
--- PASS: TestReporter_MakeModelKeyValue (0.00s)
FAIL
FAIL	github.com/jaegertracing/jaeger/cmd/agent/app/reporter/grpc	0.383s
```

cc @albertteoh

albertteoh:

Proposed fix to a failing test caused by Go 1.15 dropping support for Common Names.

Please let me know if there is a better way of generating certs using SANs or if any configurations appear spurious, as this is pretty new to me.

Closes #2435

- Regenerate certificates using Subject Alternative Names as per the recommendation in https://github.com/golang/go/issues/39568#issuecomment-671424481.
- Update README with instructions to generate the current set of certificates.
- Add a copy of existing Travis jobs to be executed against Go 1.15 to avoid introduced regressions.


.travis.yml
```diff
@@ -7,6 +7,44 @@ dist: bionic
 
 matrix:
   include:
+  # Go 1.15
```
albertteoh, line 10: I wasn't able to get matrix expansion working here, but figured it's pretty important to avoid silent regressions on Go 1.15 builds.

Was hoping we could simply add:
```
go:
- 1.14.x
- 1.15.x
```

and remove the references to `1.14.x` in the `matrix.include` list, but it seems `go` can't be expanded into the matrix. 

Suggestions welcome!

pkg/config/tlscfg/testdata/README.md
```diff
@@ -1,35 +1,60 @@
 # Example Certificate Authority and Certificate creation for testing
 
-The following commands were used to create the CA, server and client's certificates and keys
+The following commands were used to create the CA, server and client's certificates and keys. These certificates use the Subject Alternative Name extension rather than the Common Name, which will be unsupported in Go 1.15.
```
yurishkuro, line 3: just curious:
 - couldn't we just move all of this into a script, instead of a run book?
 - is there a flag to for "accept defaults" to make the script even less interactive?

.travis.yml
yurishkuro: Since the build now passes for 1.15, I suggest we just update 1.14 -> 1.15, instead of duplicating (the CI already runs for too long, we don't have capacity to double the number of steps).

We may want to add one extra step to run unit tests & lint with go tip (but I would do it in a separate PR).

albertteoh: Okay, updated to 1.15. 
Where can I learn more about "go tip"?

pkg/config/tlscfg/testdata/README.md
albertteoh: Good point, and yes, there's a way to accept defaults. I've modified this README file into a bash script.

.travis.yml
yurishkuro: I've seen it used elsewhere, e.g. https://github.com/opentracing/opentracing-go/blob/d34af3eaa63c4d08ab54863a4bdd0daa45212e12/.travis.yml#L7

pkg/config/tlscfg/testdata/gen-certs.sh
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env bash
+
```
yurishkuro, line 2: I recommend `set -ex`

pkg/config/tlscfg/testdata/gen-certs.sh
```diff
@@ -0,0 +1,58 @@
...
+# The following commands were used to create the CA, server and client's certificates and keys in this directory used by unit tests.
+# These certificates use the Subject Alternative Name extension rather than the Common Name, which will be unsupported in Go 1.15.
+
+# Generate config files.
+# The server name (under alt_names in the ssl.conf) is `example.com`. (in accordance to [RFC 2006](https://tools.ietf.org/html/rfc2606))
+source gen-ssl-conf.sh example.com ssl.conf
+source gen-ssl-conf.sh wrong.com wrong-ssl.conf
```
yurishkuro, line 9: perhaps instead of writing to the current dir (and accidentally checking in), we could use temp dir
```shell
tmp_dir=$(mktemp -d -t certificates)
clean_up () {
    ARG=$?
    rm -rf $tmp_dir
    exit $ARG
} 
trap clean_up EXIT
```

pkg/config/tlscfg/testdata/gen-certs.sh
```diff
@@ -0,0 +1,58 @@
...
+             -sha256 \
+             -days 3650 \
+             -in example-client.csr \
+             -signkey example-client-key.pem \
+             -out example-client-cert.pem \
+             -extensions req_ext \
+             -CA example-CA-cert.pem \
+             -CAkey example-CA-key.pem \
+             -CAcreateserial \
+             -extfile ssl.conf
```
yurishkuro, line 49: how about writing PEM files into `$tmp_dir` and then copying into the right place in the source tree?

pkg/config/tlscfg/testdata/gen-ssl-conf.sh
```diff
@@ -0,0 +1,34 @@
...
+# Generates the SSL conf files required for generating certificates.
+
+domain_name="$1"
+output_file="$2"
+
+if [[ -z "$domain_name" || -z "$output_file" ]]; then
+    printf "A script to generate SSL configuration files for testing purposes.\n\n"
+    printf "Usage: ssl-conf-gen.sh DOMAIN_NAME OUTPUT_FILE\n\n"
+    printf "Example: ssl-conf-gen.sh example.com ssl.conf\n"
+    return
```
yurishkuro, line 12: should this be `exit -1` instead?

pkg/config/tlscfg/testdata/README.md
```diff
@@ -1,35 +0,0 @@
-# Example Certificate Authority and Certificate creation for testing
```
yurishkuro, line 1: I would keep the README with some basic instruction how to regenerate certificates, preferably via a make command.

pkg/config/tlscfg/testdata/gen-ssl-conf.sh
albertteoh: `return` was intentional because this script is `source`d, which means it runs within the parent shell. 

The benefit of `source`-ing is that the `set -ex` that you suggested above will also apply in this script so commands in this script are printed to STDOUT and will also trigger an early exit if something fails.

If developers want to call this script separately, it can be run with `source gen-ssl-conf.sh <args...>`.

pkg/config/tlscfg/testdata/README.md
albertteoh: README.md and Makefile added.

pkg/config/tlscfg/testdata/gen-certs.sh
albertteoh: Nice, I've learnt something new today!

.travis.yml
albertteoh: Would an additional "tip" build version result in doubling the number of steps (1.15 + tip)?

pkg/config/tlscfg/testdata/gen-certs.sh
```diff
@@ -0,0 +1,95 @@
...
+      rm -rf "$tmp_dir"
+    fi
+    exit $ARG
+}
+trap clean_up EXIT
+
+# Generate config files.
+# The server name (under alt_names in the ssl.conf) is `example.com`. (in accordance to [RFC 2006](https://tools.ietf.org/html/rfc2606))
+source gen-ssl-conf.sh example.com "$tmp_dir/ssl.conf"
+source gen-ssl-conf.sh wrong.com "$tmp_dir/wrong-ssl.conf"
```
yurishkuro, line 45: Why do you need to use `source` instead of just invoking the script?

Also, since it just generates a config file, can you not define it as a function in this single script file?

pkg/config/tlscfg/testdata/README.md
yurishkuro: I was actually referring to the main Makefile. If it's just in this dir then I would simply provide instructions for running the script directly.

pkg/config/tlscfg/testdata/gen-certs.sh
albertteoh: `source` will allow `set -ex` to take affect in the `source`-ed script.

The only reason for keeping it as a separate script was to separate the concern of config file generation from certificate generation, which may be useful for debugging. However, the file generating code is not particularly complex, so it can be moved into a function of the main script.

yurishkuro (approved):
🎉 🎉 🎉

yurishkuro:
one of the commits is not signed (probably the merge). You may want to squash into one:

https://github.com/jaegertracing/jaeger/blob/master/CONTRIBUTING_GUIDELINES.md#missing-sign-offs

PR #2466, opened 2020-09-10
Run the jaeger-agent as a non-root user by default

### LINKED ISSUE #2460: Run the jaeger-agent as a non-root user by default

I am trying to run the jaeger-agent container in a Kubernetes cluster with strict Pod Security Policies set, which requires all containers to run as non-root users.

The Jaeger-agent container runs as root by default (https://github.com/jaegertracing/jaeger/blob/master/cmd/agent/Dockerfile).

Update the Dockerfile to explicitely set a non-root `USER`. There's also some open discussion about this over at the Helm chart repo: https://github.com/jaegertracing/helm-charts/issues/41

I would suggest adding a USER directive to the end of the Dockerfile, setting the running user id to `10001`. I can create a PR for this. 

* Should a different UID be used by default? - I prefer 10001 to 1001 suggested in https://github.com/jaegertracing/helm-charts/issues/41 as the latter UID is often already used by some account on the host machine.

chgl:
- Resolves #2460 

- Added a UID arg, defaulting to 10001, which the build image is running as by default.


cmd/agent/Dockerfile
```diff
@@ -3,8 +3,11 @@ RUN apk add --update --no-cache ca-certificates
...
 FROM scratch
 ARG TARGETARCH=amd64
+ARG USER_UID=10001
 COPY --from=certs /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
 
 EXPOSE 5775/udp 6831/udp 6832/udp 5778
 COPY agent-linux-$TARGETARCH /go/bin/agent-linux
 ENTRYPOINT ["/go/bin/agent-linux"]
+
+USER ${USER_UID}
```
yurishkuro, line 13: Just curious, does this UID have a name? How will it show in `ps`?

yurishkuro (approved):

cmd/agent/Dockerfile
chgl: If there's no name/passwd entry associated with the UID, then the process' user should appear as `unknown` in `ps`. If you run kubectl exec, bash will usually show a prompt like `I have no name!@hostname:`. 

The advantage of using a numeric user vs an actual name is that if a Pod Security Policy in Kubernetes requires a container to be run as root (`runAsNonRoot: true`) then setting any "user" other than 0 here allows the container to be run without having to override the user id in the securityContext.

jpkrohling: For completeness: Kubernetes platforms like OpenShift would typically assign a non-root user dynamically anyway. I expect the same to be true for GKE and AKS.

jpkrohling:
crossdock and unit tests CI jobs restarted.

jpkrohling:
CI looks good. @chgl, could you please rebase this? Once it's done, I'll merge this one.

jpkrohling:
crossdock-otel job restarted

jpkrohling:
Thanks!

PR #2472, opened 2020-09-11
Fix port in cassandra schema creation

### LINKED ISSUE #2467: Support PaaS Cassandra

Connect to a cloud-provider managed Cassandra to reduce the operational overhead on the databases. Unfortunately, for most cloud providers, a non 9042 port is used and SSL is required.

- Cassandra create schema does not support other ports than the default 9042 for the CQL connection.
- 

For operator based installations I already addressed the topic in https://github.com/jaegertracing/jaeger-operator/issues/1179 and we narrowed down some aspects to consider.

To create the Cassandra schema, we have to change the following to support environment variables like `CQLSH_PORT`. 

https://github.com/jaegertracing/jaeger/blob/043d00cc9402378c886aeace116d85e578a44834/plugin/storage/cassandra/schema/docker.sh#L6-L15

To configure SSL certificates, I would suggest to have an option to use your own set of certs or support some "well-known" CAs like Lets Encrypt.

Maybe we can start to discuss the approach. In the meantime, I will work on a PR.

MarianZoll:
what:
- considered CQLSH_PORT in schema creation port
why:
- unable to create schema in Cassandra instances not exposing 9042

Signed-off-by: Marian Zoll <Marian.Zoll@web.de>

Fixes #2467

- Added support for `CQLSH_PORT` in `docker.sh` for schema creations.

### DIFF
--- a/plugin/storage/cassandra/schema/docker.sh
+++ b/plugin/storage/cassandra/schema/docker.sh
@@ -5,6 +5,7 @@
 
 CQLSH=${CQLSH:-"/opt/cassandra/bin/cqlsh"}
 CQLSH_HOST=${CQLSH_HOST:-"cassandra"}
+CQLSH_PORT=${CQLSH_PORT:-"9042"}
 CQLSH_SSL=${CQLSH_SSL:-""}
 CASSANDRA_WAIT_TIMEOUT=${CASSANDRA_WAIT_TIMEOUT:-"60"}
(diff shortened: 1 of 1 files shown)


joe-elliott (approved):
Manually tested this locally and confirmed it worked.

yurishkuro (approved):

Issue #1516, opened 2019-05-05
Use integration tests to generate coverage


yurishkuro:
The `plugin/storage/integration` tests are using `STORAGE` env var to decide which tests to run. It makes sense for tests that require external storage, but badger/grpc-plugin are currently not running at all.

I propose the following changes:
  * always run Badger tests
  * always run grpc-plugin tests, but look for env variable pointing to the plugin binary path (to address [this comment](address https://github.com/jaegertracing/jaeger/pull/1461/files#r280031679))
  * use `-coverpkg` directive to generate test coverage for the actual packages, and merge it with overall coverage (codecov.io allows submitting multiple reports from CI job)

Esh08:
I see that this issue has been added as a good first issue, I'm new to contributing, how may I help? Please let me know, thanks!

yurishkuro:
Help is appreciated. What kind of assistance are you looking for?

yurishkuro:
Perhaps an even better way would be to change the tests from using `integration` build tag to have specific tag for each storage type, so that we don't even run irrelevant tests.

PR #2474, opened 2020-09-12
Enable CI for badger and grpc-plugin

yurishkuro:
Partially address #1516 since the badger and grpc-plugin storage integration tests are not running in CI at all today.

* adds a new CI step to run grpc-plugin and badger integration tests
* minor clean-up of the plugin framework code
* reorder logic in .travis.yml to match the actual order of the matrix steps


yurishkuro:
cc @albertteoh

PR #2475, opened 2020-09-13
Print OpenTelemetry Configuration

### LINKED ISSUE #2253: Print final configuration
Jaeger OTEL components provide a default configuration - e.g. set of receivers, processors and exporters. These components have their own configuration that can be supplied via flags and/or config file.

Users should always know what components are being used and how are they configured. A possible solution is to print the configuration on the standard output. This could be done always at the startup or on-demand via a flag. To protect leaking of credentials it might be better to provide a flag that would print out the configuration just on demand.

joe-elliott:
- Fixes #2253 

- Consolidation of config factory
- Addition of `config.dump-otel` flag
- Moved zipkin host port configuration into the zipkinreceiver `CreateDefaultConfig()`.  This mimics the way the jaegerreceiver works and removes the ZipkinHostPort field from ComponentSettings.

## Notes
- `fmt.Println` is crude but was chosen b/c it would be unaffected by config like a logger and cleanly dumps the config to stdout.
- I am purposefully exiting after printing the config to protect secrets.  Otherwise it might be easy to leave this parameter in and deploy it to production and accidentally dump secrets.

## Todo
- [x]  Tests

@pavolloffay If you have a chance to review this and make sure I'm headed the right direction please do.  I intend to add tests and drop draft status if this change is what is desired.


joe-elliott:
@pavolloffay I think this PR is in a good spot.  Added a basic test to make sure that config factory returned an error if the dump otel flag was passed.  Let me know what you think.

jpkrohling:
crossdoc-otel job restarted.

pavolloffay:
The upstream collector does not have a default config in terms of enabled components, however each component creates a component config with default settings that are unknown to the user. 

So I think this feature might be also useful in the upstream OTel collector. There might be more ways how to approach this either add a flag that prints the full config or print the component config when a component is being started.  I have filed an issue in OTEL https://github.com/open-telemetry/opentelemetry-collector/issues/1791

joe-elliott:
Sounds good, we can leave this for a bit and then close it if otel collector is ok with adding this functionality.

joe-elliott:
Going to close this PR in favor of the one linked by @pavolloffay.  The refactoring in this PR was taken care of in #2495

PR #2477, opened 2020-09-14
Fix failing ServerInUseHostPort test on MacOS

albertteoh:

- Master branch unit tests are failing on `TestServerInUseHostPort`, seemingly only for MacOS (okay for Linux builds).
- IIUC, the reason for the test is to have test coverage of error handling code when listening on GRPC and HTTP ports. To achieve this, the test first listens on those same ports, then attempts to bind to those same ports again via the QueryService `Start()`; and expecting an error to be returned.
- When running this test, no error is returned when binding to `127.0.0.1:8080` while `:8080` is bound, causing the assertion to fail. Interestingly, this behaviour differs between MacOS and Linux (at least on my Centos 8 VM).
- From this, it appears that the Linux OS libraries are stricter than MacOS; the former considering the `0.0.0.0` address to be bindable to the loopback network interface.
- Confirmed with the following simple program which I ran on both OSes:

```
package main

import (
     "fmt"
     "net"
 )

 const (
     hostname0 = ":8080"
     hostname1 = "127.0.0.1:8080"
 )

 func listen(hostname string) {
     fmt.Println("Listening to", hostname)
     if _, err := net.Listen("tcp", hostname); err != nil {
         fmt.Println(err.Error())
     }
 }

 func main() {
     listen(hostname0)
     listen(hostname1)
 }
```

...and here are the results:

**MacOS**
```
$ go run main.go
Listening to :8080
Listening to 127.0.0.1:8080
```

**Centos 8**
```
$ go run main.go
Listening to :8080
Listening to 127.0.0.1:8080
listen tcp 127.0.0.1:8080: bind: address already in use
```

- The fix involves explicitly providing the host IP (loopback in this case) in the host port string.
- Add a stronger assertion on the error.
- Have tried adding a MacOS job in Travis checks to prevent this from happening in future (there've been 2 MacOS specific unit test breakages in the last few weeks already), however, had some difficulties getting the Sarama unit tests to pass, and not sure why they're failing.

### DIFF
--- a/cmd/query/app/server_test.go
+++ b/cmd/query/app/server_test.go
@@ -24,6 +24,7 @@ import (
 	opentracing "github.com/opentracing/opentracing-go"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zaptest/observer"
 
--- a/cmd/query/app/server_test.go
+++ b/cmd/query/app/server_test.go
@@ -83,18 +83,18 @@ func TestServerInUseHostPort(t *testing.T) {
 		httpHostPort string
 		grpcHostPort string
 	}{
-		{"HTTP host port clash on " + conn.Addr().String(), conn.Addr().String(), availableHostPort},
-		{"GRPC host port clash on " + conn.Addr().String(), availableHostPort, conn.Addr().String()},
+		{"HTTP host port clash", conn.Addr().String(), availableHostPort},
+		{"GRPC host port clash", availableHostPort, conn.Addr().String()},
 	}
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
 			server, err := NewServer(
-			        zap.NewNop(), 
-			        &querysvc.QueryService{},
+				zap.NewNop(),
+				&querysvc.QueryService{},
 				&QueryOptions{
-				    HTTPHostPort: tc.httpHostPort, 
-				    GRPCHostPort: tc.grpcHostPort, 
-				    BearerTokenPropagation: true,
+					HTTPHostPort:           tc.httpHostPort,
+					GRPCHostPort:           tc.grpcHostPort,
+					BearerTokenPropagation: true,
 				},
 				opentracing.NoopTracer{},
 			)
--- a/cmd/query/app/server_test.go
+++ b/cmd/query/app/server_test.go
@@ -17,6 +17,7 @@ package app
 import (
 	"context"
 	"net"
+	"os"
 	"sync"
 	"testing"
 	"time"
@@ -26,6 +27,7 @@ import (
 	"github.com/stretchr/testify/mock"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zaptest/observer"
+	"golang.org/x/sys/unix"
 
 	"github.com/jaegertracing/jaeger/cmd/flags"
 	"github.com/jaegertracing/jaeger/cmd/query/app/querysvc"
@@ -74,7 +76,7 @@ func TestServerBadHostPort(t *testing.T) {
 
 func TestServerInUseHostPort(t *testing.T) {
 
-	for _, hostPort := range [2]string{":8080", ":8081"} {
+	for _, hostPort := range [2]string{"127.0.0.1:8080", "127.0.0.1:8081"} {
 		conn, err := net.Listen("tcp", hostPort)
 		assert.NoError(t, err)
 
@@ -84,7 +86,10 @@ func TestServerInUseHostPort(t *testing.T) {
 		assert.NoError(t, err)
 
 		err = server.Start()
-		assert.NotNil(t, err)
+		syscallErr, ok := err.(*net.OpError).Err.(*os.SyscallError)
+		assert.True(t, ok)
+		assert.Equal(t, syscallErr.Err, unix.EADDRINUSE)
+
 		conn.Close()
 		if server.grpcConn != nil {
 			server.grpcConn.Close()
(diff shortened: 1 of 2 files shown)


jpkrohling:
cc @rjs211

cmd/query/app/server_test.go
```diff
@@ -26,6 +27,7 @@ import (
 	"github.com/stretchr/testify/mock"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zaptest/observer"
+	"golang.org/x/sys/unix"
```
yurishkuro, line 30: is it wise to introduce OS dependency without a build tag? How critical is it that we check for specific OS error code?

cmd/query/app/server_test.go
```diff
@@ -74,7 +76,7 @@ func TestServerBadHostPort(t *testing.T) {
 
 func TestServerInUseHostPort(t *testing.T) {
 
-	for _, hostPort := range [2]string{":8080", ":8081"} {
+	for _, hostPort := range [2]string{"127.0.0.1:8080", "127.0.0.1:8081"} {
```
yurishkuro, line 79: a cleaner way to write tests is to listen on port 0 and let the OS assign the unused port. Can we do it here?

albertteoh: I believe the goal of this test is to intentionally have two listeners listening on the same host:port to cause a "Bind address already in use" error, and so there's a need for the port numbers to be deterministic rather than dynamically allocated by the OS.

Please let me know if I've misunderstood your question.

cmd/query/app/server_test.go
albertteoh: Fair question, it's not critical to check the specific OS error code. However, given we are trying to trigger an error that stems from a fairly deep call stack and originating from outside this project, I was also conscious there could be a potentially larger pool of error types returned and so wanted to be as specific as possible to avoid false positives.

Perhaps we can continue to perform type assertions (and assert their success) on the error interfaces until we reach the most specific error, but remove the check on the specific error requiring the unix dependency; so something like the following instead?:

```
err = server.Start()
assert.Error(t, err)
_, ok := err.(*net.OpError).Err.(*os.SyscallError)
assert.True(t, ok)
```

cmd/query/app/server_test.go
yurishkuro: I haven't read the test carefully, but from it's name it's trying to test for port conflict, which does not require listening on a _specific_ port, it just requires trying to bind twice to _the same port_. The first time it can bind to a random port and use it for the second attempt.

cmd/query/app/server_test.go
yurishkuro: My concern is that it just makes the test dependent on some implementation details of the OS, which can change over time, or Go may change its implementation.

Stepping back to what the test is even about - why would we care about port reuse? I think the objective of the test was to check the code path when listening to a port causes an error. It doesn't matter what the error is, just that our code handles it. So the test may do whatever it needs to introduce the error condition. For example, is there any difference in our code behavior (different path) if the port is already take or is completely invalid?

cmd/query/app/server_test.go
albertteoh: That's neat! I didn't think it was possible to retrieve the OS allocated port. Thanks!

albertteoh:
@yurishkuro are there any outstanding concerns that need addressing in this PR?

cmd/query/app/server_test.go
```diff
@@ -73,28 +73,37 @@ func TestServerBadHostPort(t *testing.T) {
...
-		if server.httpConn != nil {
-			server.httpConn.Close()
-		}
-
+	testCases := []struct {
+		name         string
+		httpHostPort string
+		grpcHostPort string
+	}{
+		{"HTTP host port clash on " + conn.Addr().String(), conn.Addr().String(), availableHostPort},
```
yurishkuro, line 86: `"HTTP host port clash on " + conn.Addr().String()` - I think it's better if the test names are stable, this one will include random port number on every run, which is not informative anyway

cmd/query/app/server_test.go
```diff
@@ -73,28 +73,37 @@ func TestServerBadHostPort(t *testing.T) {
...
+		{"GRPC host port clash on " + conn.Addr().String(), availableHostPort, conn.Addr().String()},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			server, err := NewServer(zap.NewNop(), &querysvc.QueryService{},
+				&QueryOptions{HTTPHostPort: tc.httpHostPort, GRPCHostPort: tc.grpcHostPort, BearerTokenPropagation: true},
+				opentracing.NoopTracer{})
```
yurishkuro, line 93: This looks like Java formatting, I think this is more readable:
```suggestion
			server, err := NewServer(
			        zap.NewNop(), 
			        &querysvc.QueryService{},
				&QueryOptions{
				    HTTPHostPort: tc.httpHostPort, 
				    GRPCHostPort: tc.grpcHostPort, 
				    BearerTokenPropagation: true,
				},
				opentracing.NoopTracer{},
			)
```

yurishkuro (approved):
lgtm, aside from minor formatting

cmd/query/app/server_test.go
```diff
@@ -73,28 +73,44 @@ func TestServerBadHostPort(t *testing.T) {
 }
 
 func TestServerInUseHostPort(t *testing.T) {
+	const availableHostPort = "127.0.0.1:0"
+	conn, err := net.Listen("tcp", availableHostPort)
+	defer func() { conn.Close() }()
+	assert.NoError(t, err)
```
yurishkuro, line 79: reversed order.
```suggestion
	require.NoError(t, err)
	defer func() { require.NoError(t, conn.Close()) }()
```

yurishkuro (approved):

PR #2496, opened 2020-09-21
Expose tunning options via expvar

### LINKED ISSUE #1333: Expose the tuning options via expvar

It's not possible to check what's the current value for different tuning options, such as:

```
--collector.num-workers
--collector.queue-size
--discovery.conn-check-timeout
--discovery.min-peers
--memory.max-traces
--processor.*.server-max-packet-size
--processor.*.server-queue-size
--processor.*.workers
--reporter.tchannel.discovery.conn-check-timeout
--reporter.tchannel.discovery.min-peers
--reporter.tchannel.report-timeout
```

During the startup process, store values in expvar variables and make sure /debug/expvar is always mounted on the admin port (right now it is only mounted if `--metrics-backend=expvar` is specified).

dstdfx:
- Resolves #1333

- Mounts expvar handler (/debug/vars) on the admin port regardless of ```--metrics-backend```
- Stores the following collector's options in expvar:
  - collector.num-workers
  - collector.queue-size
- Store the following agent's options in expvar:
  - processor.*.server-max-packet-size
  - processor.*.server-queue-size
  - processor.*.workers
- Store the following storage-related options in expvar: 
  - memory.max-traces
  - downsampling.ratio
  - span-storage.type


jpkrohling:
This test has indeed been flaky...

jpkrohling:
Tests are passing now :-)

cmd/agent/app/flags_test.go
```diff
@@ -51,3 +53,39 @@ func TestBindFlags(t *testing.T) {
...
+	AddFlags(flags)
+	command.PersistentFlags().AddGoFlagSet(flags)
+	v.BindPFlags(command.PersistentFlags())
+
+	err := command.ParseFlags([]string{
+		"--http-server.host-port=:8080",
+		"--processor.jaeger-binary.server-host-port=:1111",
+		"--processor.jaeger-binary.server-max-packet-size=4242",
+		"--processor.jaeger-binary.server-queue-size=42",
+		"--processor.jaeger-binary.workers=42",
```
jpkrohling, line 71: Quite a small thing, but could you change this to be different than the `server-queue-size`? Same as some other test in this PR.

cmd/agent/main.go
```diff
@@ -79,6 +79,9 @@ func main() {
 			if err := agent.Run(); err != nil {
 				return fmt.Errorf("failed to run the agent: %w", err)
 			}
+
+			app.ExposeTuningOptions(builder)
```
jpkrohling, line 83: I'm not sure I'm a big fan of having the command to call this. Sounds like a knowledge it shouldn't have. How about providing this as a private function, which is called by `Start()/Run()`? Or even closer to where they are defined/used, perhaps in the builder themselves. The badger storage plugin does something similar when it's being initialized:

https://github.com/jaegertracing/jaeger/blob/a2fb067f01551b47aef4f58bcc0e925ffeb32f5f/plugin/storage/badger/factory.go#L137

jpkrohling (commented):
I like the base idea, but I think the actual registration with expvar could be closer to where the values are used or defined.

I would also love to see other values, although they might come in a later PR:

```
--memory.max-traces
--span-storage.type (not sure the storate type is actually stored there)
--downsampling.ratio
```

cmd/agent/app/flags_test.go
dstdfx: As I see ```--span-storage.type``` is deprecated in [1.19](https://www.jaegertracing.io/docs/1.19/cli/#jaeger-collector).

jpkrohling: It is since like forever, as we need it in order to setup the storage-related flags, so, people have to use an env var. But my point is that we could expose what's the storage type that is being used by the instance, and I think it is indeed stored in Viper's key for this flag.

jpkrohling (commented):
Looks good to me. One thing that we might want to add is the storage type. What do you think?

dstdfx:
Sure, wilco

dstdfx:
@jpkrohling Is it known issue https://travis-ci.org/github/jaegertracing/jaeger/jobs/729295839?

jpkrohling:
Doesn't sound like it, looks like it's a real problem with the new test:

```
github.com/jaegertracing/jaeger/plugin/storage.TestSetExpvarOptions(0xc000229080)
	/home/travis/gopath/src/github.com/jaegertracing/jaeger/plugin/storage/factory_test.go:364 +0x337
```

dstdfx:
> Doesn't sound like it, looks like it's a real problem with the new test:
> 
> ```
> github.com/jaegertracing/jaeger/plugin/storage.TestSetExpvarOptions(0xc000229080)
> 	/home/travis/gopath/src/github.com/jaegertracing/jaeger/plugin/storage/factory_test.go:364 +0x337
> ```

My bad, didn't notice...

cmd/agent/app/builder.go
```diff
@@ -127,6 +129,21 @@ func (b *Builder) getReporter(primaryProxy CollectorProxy) reporter.Reporter {
 	return reporter.NewMultiReporter(rep...)
 }
 
+func (b *Builder) setExpvarOptions() {
+	for _, p := range b.Processors {
+		prefix := fmt.Sprintf(processorPrefixFmt, p.Model, p.Protocol)
+		if expvar.Get(prefix+suffixServerMaxPacketSize) == nil {
```
yurishkuro, line 135: This looks suspicious - there should be no other place in the code that would set this expvar, and if there were it would be good to panic because the data would be misleading.

I assume the reason for this check is because otherwise unit tests would panic? Which means it's masking the actual problem.

I am inclined to say that instead of writing to expvar directly (which uses global variables and thus is always problematic to test), we might want to use a dedicated metrics factory, which for unit tests can be instantiated to NullFactory, and in production (in main()s) it can be instantiated with expvar backing.

dstdfx: That's right, some of the unit tests are panicing due to expvar rewriting that's why I check expvar variable before creating a new one.

yurishkuro: Would you be interested in following up on my suggestion above and using MetricsFactory abstraction instead of direct expvar access? Our jaeger-lib library provides an implementation of MetricsFactory backed by expvar.

Another option would be to use the same MetricsFactory that we already have, but introduce a convention that certain metrics namespaces, e.g. `internal` would be dispatched to a different implementation (expvar in this case), instead of being piped to the standard metrics exporter.

dstdfx: Looks interesting, sure I'd give it a try but which option is preferable? For me it's still hard to get the whole picture.

yurishkuro: Having a dedicated MetricsFactory for internal parameters potentially requires more wiring, to pass it through all the way from main.  And it's a bit confusing since now we'd have two metric factories everywhere. 

So I would say the last option is most preferable, it relies on MetricsFactory that's already present almost everywhere.

Overall going through MetricsFactory API is preferable over direct expvar access because it allows more flexibility in where the parameters will be exposed in the end, e.g. if in the future we adopt OpenTelemetry SDK that supports z-pages, I think it would make more sense to expose this data via z-pages (can be a nicer UI) than via default expvar handler.

I think the only caveat with using `internal` namespace as a trigger  is that it will not always be the top-level namespace, because some components, such as UDP servers in the agent, are likely passed MetricsFactory that is already namespaced by the caller. But I think it's ok, we could have metrics like `ns1.ns2.internal.ns3.metricName` that would be forked to expvar.

In order to do this you first need to create a PR in jaeger-lib for a new dispatching MetricsFactory, e.g.
```
func NewNamespaceForkingMetricsFactory(
    forkNamespace string,
    forkFactory MetricsFactory,
    defaultFactory MetricsFactory,
) MetricsFactory
```

lmk if you need more details.

dstdfx: So roughly in jaeger-lib we need to implement a separate factory like in [this example](https://github.com/jaegertracing/jaeger-lib/blob/master/metrics/multi/multi.go) and this factory will handle metrics with ```forkNamespace``` by ```forkFactory```?

yurishkuro: Yes! But it wouldn't be "multi" in the same way, i.e. when factory creates a counter, it would be just one counter from the defaultFactory. Basically it should always delegate to the default factory except for Namespace() function:

```go
func (f *Factory) Namespace(scope metrics.NSOptions) metrics.Factory {
    if scope.Name == f.forkNamespace {
        return f.forkFactory.Namespace(scope)
    }
    return &Factory({
        forkNamespace: f.forkNamespace,
        forkFactory: f.forkFactory.Namespace(scope),
        defaultFactory: f.defaultFactory.Namespace(scope),
    })
}
```

cmd/agent/main.go
```diff
@@ -49,9 +51,10 @@ func main() {
 				return err
 			}
 			logger := svc.Logger // shortcut
-			mFactory := svc.MetricsFactory.
-				Namespace(metrics.NSOptions{Name: "jaeger"}).
-				Namespace(metrics.NSOptions{Name: "agent"})
+			baseFactory := svc.MetricsFactory.Namespace(metrics.NSOptions{Name: "jaeger"})
```
yurishkuro, line 54: why isn't "agent" namespace on this line?

cmd/agent/main.go
```diff
@@ -49,9 +51,10 @@ func main() {
...
+			mFactory := fork.New("internal",
+				jexpvar.NewFactory(10).Namespace(metrics.NSOptions{}), // backend for internal opts
```
yurishkuro, line 56: why `.Namespace(metrics.NSOptions{})` ?

cmd/opentelemetry/go.mod
```diff
@@ -15,7 +15,7 @@ require (
 	github.com/spf13/viper v1.7.1
 	github.com/stretchr/testify v1.6.1
 	github.com/uber/jaeger-client-go v2.25.0+incompatible
-	github.com/uber/jaeger-lib v2.2.0+incompatible
+	github.com/uber/jaeger-lib v2.3.1-0.20200925164221-2cac3b34a519+incompatible
```
yurishkuro, line 18: I assume we need a new release?

plugin/storage/factory.go
```diff
@@ -46,6 +46,7 @@ const (
 	badgerStorageType        = "badger"
 	downsamplingRatio        = "downsampling.ratio"
 	downsamplingHashSalt     = "downsampling.hashsalt"
+	spanStorageType          = "span-storage.type"
```
yurishkuro, line 49:
```suggestion
	spanStorageType          = "span-storage-type"
```

and I don't think you need a constant for metric name

plugin/storage/factory.go
```diff
@@ -249,3 +252,11 @@ func (f *Factory) Close() error {
 	}
 	return multierror.Wrap(errs)
 }
+
+func (f *Factory) setExpvarOptions() {
```
yurishkuro, line 256: nit: there's nothing specific to expvar in this function and others like it, consider a different name, like publishParams

plugin/storage/factory.go
yurishkuro: actually, this constant might already be defined elsewhere

cmd/opentelemetry/go.mod
dstdfx: Yeah, that's right

yurishkuro: https://github.com/jaegertracing/jaeger-lib/releases/tag/v2.4.0

yurishkuro: shouldn't this now point to 2.4.0?

yurishkuro (approved):
Looks great! The only question is about otel/go.mod

yurishkuro:
I am going to merge it, please put another PR for otel/go.mod

PR #2504, opened 2020-09-22
Enable coverage in OTEL module and use go-acc for coverage

pavolloffay:
Enable code coverage in OTEL module and use go-acc for coverage. go-acc is also used by OTEL.


yurishkuro:
```
@@             Coverage Diff             @@
##           master    #2504       +/-   ##
Coverage   95.18%   56.63%   -38.56%  
```
@pavolloffay I think this is unacceptable, we might as well give up enforcing coverage numbers. Could you please investigate where the drop is coming from?

Note that we had a convention that some packages could be marked as intentionally not covered, and their coverage were excluded from the totals. The new tool you use may not take that into account and aggregate those packages with 0% coverage. But even then, those packages were typically very small, requiring external resources like Kafka/DB connections (even though, ideally, we should collect coverage for those from integration tests and append to main coverage, which is supported by codecov.io), so I would be surprised if they would account for a 38% drop.

PR #2513, opened 2020-09-25
Disable clock-skew-adjustment by default.

### LINKED ISSUE #1459: Clock Skew Adjuster considered harmful
Edit: duplicate of #961.

The current behavior of clock skew adjustments is surprising and, why well-meant, not beneficial.

Excuse the pun, I'm aware of the existing [issues](https://github.com/jaegertracing/jaeger/issues?q=adjuster).

[The clock skew adjuster](https://github.com/jaegertracing/jaeger/blob/b5e2b65c690c3b4ed55e91f1afe1efb0570dc542/model/adjuster/clockskew.go#L148) currently makes the following assumptions:

1. Spans are from different hosts, with parent and child span having roughly the same duration.
2. Clock Skew is something unknown to people using monitoring
3. It is best to hide this fact and instead of relying on administrators to fix this and offer a best guess workaround as an alternative, it's best to make things look “nice”

Considering 1.:

The algorithm uses the not well documented [`ip` tag](https://github.com/jaegertracing/jaeger/blob/b5e2b65c690c3b4ed55e91f1afe1efb0570dc542/model/adjuster/clockskew.go#L78) of the process to distinguish between hosts and assumes a missing tag [indicates a different host](https://github.com/jaegertracing/jaeger/blob/v1.11.0/model/adjuster/clockskew.go#L134).

People using OpenTracing or OpenCensus usually only know about the service name, so a missing `ip` tag is expected, meaning the algorithm compensates nearly all spans. See

- https://github.com/jaegertracing/jaeger/issues/722
- https://github.com/jaegertracing/jaeger/issues/961
- https://github.com/jaegertracing/jaeger/pull/606

Then it [compares durations](https://github.com/jaegertracing/jaeger/blob/b5e2b65c690c3b4ed55e91f1afe1efb0570dc542/model/adjuster/clockskew.go#L154), which makes only sense one assumes that spans roughly have the same size.

Most child spans would be shorter, so this is rarely the case. If a child span is a few nanoseconds longer I don't care, if it is really longer it should be flagged, but not shifted. It's wrong data and **really** hard to diagnose if it is processed afterwards.

Then the algorithm checks if the end time of the child span is after the end time of the parent. While technically wrong, it is pretty easy to close the child span a couple of nanoseconds too late - probably a non-issue considering tracing, and at most something that should be a warning instead of moving the span around.

Lastly it assumes that span duration somehow correlates to network latency.

If clock skew is an issue, the span context should try to compensate.

Considering 2. and 3. it would IMHO be best to not mess with the data by default and offer _beautification_ as an opt-in workaround, which is different than the currently solutions by turning it on by default and searching for a way to get rid of it.

Also, it might be useful to see clock skew by default, because it may affect other areas too.

I saw very few comments where this skew adjustent is actually wanted. Maybe some of the supporters can chime in?

jpkrohling:

- Fixes #1459

- Changed the default value for the flag `query.max-clock-skew-adjustment` to `0s`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Changes by Version
 -------------------
 
 ### Backend Changes
+* The default value for the flag `query.max-clock-skew-adjustment` has changed to `0s`, meaning that the clock skew adjustment is now disabled by default. See [#1459](https://github.com/jaegertracing/jaeger/issues/1459).
 
 #### Breaking Changes
 * Configurable ES doc count ([#2453](https://github.com/jaegertracing/jaeger/pull/2453), [@albertteoh](https://github.com/albertteoh))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,6 @@ Changes by Version
 -------------------
 
 ### Backend Changes
-* The default value for the flag `query.max-clock-skew-adjustment` has changed to `0s`, meaning that the clock skew adjustment is now disabled by default. See [#1459](https://github.com/jaegertracing/jaeger/issues/1459).
 
 #### Breaking Changes
 * Configurable ES doc count ([#2453](https://github.com/jaegertracing/jaeger/pull/2453), [@albertteoh](https://github.com/albertteoh))
@@ -22,6 +21,8 @@ Changes by Version
     * Operations
     * Dependencies (edges in a dependency graph)
     * Span fetch size for a trace
+
+* The default value for the flag `query.max-clock-skew-adjustment` has changed to `0s`, meaning that the clock skew adjustment is now disabled by default. See [#1459](https://github.com/jaegertracing/jaeger/issues/1459).
 
 #### New Features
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ Changes by Version
 -------------------
 
 ### Backend Changes
+* The default value for the flag `query.max-clock-skew-adjustment` has changed to `0s`, meaning that the clock skew adjustment is now disabled by default. See [#1459](https://github.com/jaegertracing/jaeger/issues/1459).
```
joe-elliott, line 8: Should this be listed as a breaking change as well?

jpkrohling: It does! I swear I thought I was adding it to the *Breaking* changes, not *Backend*...

jpkrohling: Moved.

joe-elliott (approved):

yurishkuro (approved):

PR #2514, opened 2020-09-25
Run coverage for OTEL module too

pavolloffay:
Replaces #2504

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -82,8 +82,8 @@ script:
 
 after_success:
   - if [ "$COVERAGE" == true ]; then mv cover.out coverage.txt ; else echo 'skipping coverage'; fi
-  - if [ "$COVERAGE" == true ]; then bash <(curl -s https://codecov.io/bash) ; else echo 'skipping coverage'; fi
-  - if [ "$COVERAGE" == true ]; then bash cd cmd/opentelemetry/ && <(curl -s https://codecov.io/bash) ; else echo 'skipping coverage'; fi
+  - if [ "$COVERAGE" == true ]; then travis_retry bash <(curl -s https://codecov.io/bash) ; else echo 'skipping coverage'; fi
+  - if [ "$COVERAGE" == true ]; then travis_retry bash cd cmd/opentelemetry/ && <(curl -s https://codecov.io/bash) ; else echo 'skipping coverage'; fi
 
 after_failure:
   - if [ "$CROSSDOCK" == true || "$CROSSDOCK_OTEL" == true ]; then make crossdock-logs ; else echo 'skipping crossdock'; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -83,6 +83,7 @@ script:
 after_success:
   - if [ "$COVERAGE" == true ]; then mv cover.out coverage.txt ; else echo 'skipping coverage'; fi
   - if [ "$COVERAGE" == true ]; then bash <(curl -s https://codecov.io/bash) ; else echo 'skipping coverage'; fi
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -83,6 +83,7 @@ script:
 after_success:
   - if [ "$COVERAGE" == true ]; then mv cover.out coverage.txt ; else echo 'skipping coverage'; fi
   - if [ "$COVERAGE" == true ]; then bash <(curl -s https://codecov.io/bash) ; else echo 'skipping coverage'; fi
+  - if [ "$COVERAGE" == true ]; then bash cd cmd/opentelemetry/ && <(curl -s https://codecov.io/bash) ; else echo 'skipping coverage'; fi
```
yurishkuro, line 86: it would be good to find a way to use retries here, recently codecov.io has been slow/unstable and these submits fail or timeout.

Makefile
```diff
@@ -158,6 +158,13 @@ cover: nocover
 	mv cover-nogen.out cover.out
 	go tool cover -html=cover.out -o cover.html
 
+.PHONY: cover-otel
+cover-otel:
+	@echo pre-compiling OTEL tests
+	@cd ${OTEL_COLLECTOR_DIR} && time go test -i $(shell cd ${OTEL_COLLECTOR_DIR} && go list ./...)
```
yurishkuro, line 164: I think this will give the wrong paths to `go test`, since the paths will be relative to subdir

Makefile
```diff
@@ -158,6 +158,13 @@ cover: nocover
...
+	@cd ${OTEL_COLLECTOR_DIR} && ROOT_PKG=github.com/jaegertracing/jaeger/cmd/opentelemetry ../../scripts/cover.sh $(shell cd ${OTEL_COLLECTOR_DIR} && go list ./...)
```
yurishkuro, line 165: Do we use `.nocover` trick in cmd/opentelemetry anywhere?  If not then it's better to run normal `go test`, which is already capable of aggregating coverage across packages

pavolloffay: no we don't at the moment. I wanted to be aligned with the core module. I can switch to `go test`

.travis.yml
pavolloffay: I have added `travis_retry`. IIRC @jpkrohling added it to other jobs.

pavolloffay:
PR updated.

Makefile
```diff
@@ -158,6 +158,12 @@ cover: nocover
 	mv cover-nogen.out cover.out
 	go tool cover -html=cover.out -o cover.html
 
+.PHONY: cover-otel
+cover-otel:
+	@echo pre-compiling OTEL tests
+	@cd ${OTEL_COLLECTOR_DIR} && time go test -i $(shell cd ${OTEL_COLLECTOR_DIR} && go list ./...)
```
yurishkuro, line 164: I don't think pre-compiling helps much or even needed when running a single `go test`. It made more sense when using cover.sh because that one was testing each module independently.

pavolloffay:
Fixed the retry

yurishkuro:
@pavolloffay could you please rebase off master (not just merge)? Codecov is quite difficult to read when the base revision is not the latest.

pavolloffay:
@yurishkuro done, rebased

yurishkuro:
Still looks like it consistently pulls coverage down to 93%, below the threshold. Did you try to run coverage just for the OTEL pieces? Looks like it's what dragging down the overall level. If that's the case, I think we need to improve it first before merging, otherwise it will just lower the quality bar for everything.

pavolloffay:
rebased after @joe-elliott increased the coverage.

PR #2519, opened 2020-09-29
Add URL option for sampling strategies file

### LINKED ISSUE #2015: Support url for sampling.strategies file
We'd like to allow dev teams to control sampling strategies. This currently is quite involved since jaeger relies on a filesystem-file. In kubernetes this typically means a configMap volume, which requires us to restart all jaeger collector pods when the file is updated. Quite cumbersome.

I'd like Jaeger to support a `sampling.strategies.url` option, where the sampling strategies file was downloaded from a url. The cache timeout could be a separate parameter, or it could use the cache headers from the dowload to determine how often to retrieve updates. 

Imho this would allow much easier changing of the sampling strategies, as we would be able to build a super-simple rest api to provide the json payload.

strategies.json is implemented as a filesystem-file only

See requirement secion

goku321:
- Resolve #2015 

- Added a URL option to support downloading a sampling strategies file from a url.

### DIFF
--- a/plugin/sampling/strategystore/static/options.go
+++ b/plugin/sampling/strategystore/static/options.go
@@ -22,7 +22,7 @@ import (
 )
 
 const (
-	// SamplingStrategiesFile contains the name of CLI opions for config file.
+	// SamplingStrategiesFile contains the name of CLI option for config file.
 	SamplingStrategiesFile           = "sampling.strategies-file"
 	samplingStrategiesReloadInterval = "sampling.strategies-reload-interval"
 )
--- a/plugin/sampling/strategystore/static/strategy_store.go
+++ b/plugin/sampling/strategystore/static/strategy_store.go
@@ -21,7 +21,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
-	"path/filepath"
+	"net/http"
+	"net/url"
 	"sync/atomic"
 	"time"
 
@@ -30,6 +31,9 @@ import (
 	ss "github.com/jaegertracing/jaeger/cmd/collector/app/sampling/strategystore"
 	"github.com/jaegertracing/jaeger/thrift-gen/sampling"
 )
+
+// null represents "null" JSON value.
+const null = "null"
 
 type strategyStore struct {
 	logger *zap.Logger
@@ -45,6 +49,8 @@ type storedStrategies struct {
 	serviceStrategies map[string]*sampling.SamplingStrategyResponse
 }
 
+type strategyLoader func() ([]byte, error)
+
 // NewStrategyStore creates a strategy store that holds static sampling strategies.
 func NewStrategyStore(options Options, logger *zap.Logger) (ss.StrategyStore, error) {
 	ctx, cancelFunc := context.WithCancel(context.Background())
@@ -55,14 +61,15 @@ func NewStrategyStore(options Options, logger *zap.Logger) (ss.StrategyStore, er
 	}
 	h.storedStrategies.Store(defaultStrategies())
 
-	strategies, err := loadStrategies(options.StrategiesFile)
+	loadFn := samplingStrategyLoader(options.StrategiesFile)
+	strategies, err := loadStrategies(loadFn)
 	if err != nil {
 		return nil, err
 	}
 	h.parseStrategies(strategies)
 
 	if options.ReloadInterval > 0 {
-		go h.autoUpdateStrategies(options.ReloadInterval, options.StrategiesFile)
+		go h.autoUpdateStrategies(options.ReloadInterval, loadFn)
 	}
 	return h, nil
 }
@@ -83,35 +90,86 @@ func (h *strategyStore) Close() {
 	h.cancelFunc()
 }
 
-func (h *strategyStore) autoUpdateStrategies(interval time.Duration, filePath string) {
-	lastValue := ""
+func downloadSamplingStrategies(url string) ([]byte, error) {
+	resp, err := http.Get(url)
+	if err != nil {
+		return nil, fmt.Errorf("failed to download sampling strategies: %w", err)
+	}
+
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		if resp.StatusCode == http.StatusServiceUnavailable {
+			return []byte("null"), nil
+		}
+		return nil, fmt.Errorf(
+			"receiving %s while downloading strategies file",
+			resp.Status,
+		)
+	}
+
+	buf := new(bytes.Buffer)
+	if _, err = buf.ReadFrom(resp.Body); err != nil {
+		return nil, fmt.Errorf("failed to read sampling strategies from downloaded JSON: %w", err)
+	}
+	return buf.Bytes(), nil
+}
+
+func isURL(str string) bool {
+	u, err := url.Parse(str)
+	return err == nil && u.Scheme != "" && u.Host != ""
+}
+
+func samplingStrategyLoader(strategiesFile string) strategyLoader {
+	if strategiesFile == "" {
+		return func() ([]byte, error) {
+			// Using null so that it un-marshals to nil pointer.
+			return []byte(null), nil
+		}
+	}
+
+	if isURL(strategiesFile) {
+		return func() ([]byte, error) {
+			return downloadSamplingStrategies(strategiesFile)
+		}
+	}
+
+	return func() ([]byte, error) {
+		currBytes, err := ioutil.ReadFile(strategiesFile)
+		if err != nil {
+			return nil, fmt.Errorf("failed to open strategies file: %w", err)
+		}
+		return currBytes, nil
+	}
+}
+
+func (h *strategyStore) autoUpdateStrategies(interval time.Duration, loader strategyLoader) {
+	lastValue := null
 	ticker := time.NewTicker(interval)
 	defer ticker.Stop()
 	for {
 		select {
 		case <-ticker.C:
-			lastValue = h.reloadSamplingStrategyFile(filePath, lastValue)
+			lastValue = h.reloadSamplingStrategy(loader, lastValue)
 		case <-h.ctx.Done():
 			return
 		}
 	}
 }
 
-func (h *strategyStore) reloadSamplingStrategyFile(filePath string, lastValue string) string {
-	currBytes, err := ioutil.ReadFile(filepath.Clean(filePath))
+func (h *strategyStore) reloadSamplingStrategy(loadFn strategyLoader, lastValue string) string {
+	newValue, err := loadFn()
 	if err != nil {
-		h.logger.Error("failed to load sampling strategies", zap.String("file", filePath), zap.Error(err))
+		h.logger.Error("failed to re-load sampling strategies", zap.Error(err))
 		return lastValue
 	}
-	newValue := string(currBytes)
-	if lastValue == newValue {
+	if lastValue == string(newValue) {
 		return lastValue
 	}
-	if err = h.updateSamplingStrategy(currBytes); err != nil {
-		h.logger.Error("failed to update sampling strategies from file", zap.Error(err))
+	if err := h.updateSamplingStrategy(newValue); err != nil {
+		h.logger.Error("failed to update sampling strategies", zap.Error(err))
 		return lastValue
 	}
-	return newValue
+	return string(newValue)
 }
 
 func (h *strategyStore) updateSamplingStrategy(bytes []byte) error {
@@ -125,24 +183,22 @@ func (h *strategyStore) updateSamplingStrategy(bytes []byte) error {
 }
 
 // TODO good candidate for a global util function
-func loadStrategies(strategiesFile string) (*strategies, error) {
-	if strategiesFile == "" {
-		return nil, nil
-	}
-	data, err := ioutil.ReadFile(strategiesFile) /* nolint #nosec , this comes from an admin, not user */
+func loadStrategies(loadFn strategyLoader) (*strategies, error) {
+	strategyBytes, err := loadFn()
 	if err != nil {
-		return nil, fmt.Errorf("failed to open strategies file: %w", err)
+		return nil, err
 	}
-	var strategies strategies
-	if err := json.Unmarshal(data, &strategies); err != nil {
+
+	var strategies *strategies
+	if err := json.Unmarshal(strategyBytes, &strategies); err != nil {
 		return nil, fmt.Errorf("failed to unmarshal strategies: %w", err)
 	}
-	return &strategies, nil
+	return strategies, nil
 }
 
 func (h *strategyStore) parseStrategies(strategies *strategies) {
 	if strategies == nil {
-		h.logger.Info("No sampling strategies provided, using defaults")
+		h.logger.Info("No sampling strategies provided or URL is unavailable, using defaults")
 		return
 	}
 	newStore := defaultStrategies()
--- a/plugin/sampling/strategystore/static/strategy_store.go
+++ b/plugin/sampling/strategystore/static/strategy_store.go
@@ -23,6 +23,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/url"
+	"path/filepath"
 	"sync/atomic"
 	"time"
 
@@ -32,8 +33,9 @@ import (
 	"github.com/jaegertracing/jaeger/thrift-gen/sampling"
 )
 
-// null represents "null" JSON value.
-const null = "null"
+// null represents "null" JSON value and
+// it un-marshals to nil pointer.
+var nullJSON = []byte("null")
 
 type strategyStore struct {
 	logger *zap.Logger
@@ -60,6 +62,11 @@ func NewStrategyStore(options Options, logger *zap.Logger) (ss.StrategyStore, er
 		cancelFunc: cancelFunc,
 	}
 	h.storedStrategies.Store(defaultStrategies())
+
+	if options.StrategiesFile == "" {
+		h.parseStrategies(nil)
+		return h, nil
+	}
 
 	loadFn := samplingStrategyLoader(options.StrategiesFile)
 	strategies, err := loadStrategies(loadFn)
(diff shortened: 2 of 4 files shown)


go.mod
```diff
@@ -3,83 +3,57 @@ module github.com/jaegertracing/jaeger
 go 1.14
 
 require (
-	github.com/AndreasBriese/bbloom v0.0.0-20190825152654-46b345b51c96 // indirect
```
yurishkuro, line 6: I do not believe this PR requires changes to dependencies. Please exclude the go.mod/go.sum files.

plugin/sampling/strategystore/static/options.go
```diff
@@ -22,9 +22,11 @@ import (
 )
 
 const (
-	// SamplingStrategiesFile contains the name of CLI opions for config file.
+	// SamplingStrategiesFile contains the name of CLI options for config file.
```
yurishkuro, line 25:
```suggestion
	// SamplingStrategiesFile contains the name of CLI option for config file.
```

plugin/sampling/strategystore/static/options.go
```diff
@@ -22,9 +22,11 @@ import (
...
 	SamplingStrategiesFile           = "sampling.strategies-file"
 	samplingStrategiesReloadInterval = "sampling.strategies-reload-interval"
+	// SamplingStrategiesURL contains the name of CLI option for config file URL.
+	SamplingStrategiesURL = "sampling.strategies-url"
```
yurishkuro, line 29: I think the intention of the ticket was to allow `sampling.strategies-file` to contain a URL. A new flag seems redundant.

plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -83,32 +84,73 @@ func (h *strategyStore) Close() {
...
 		}
 	}
 }
 
-func (h *strategyStore) reloadSamplingStrategyFile(filePath string, lastValue string) string {
-	currBytes, err := ioutil.ReadFile(filepath.Clean(filePath))
-	if err != nil {
-		h.logger.Error("failed to load sampling strategies", zap.String("file", filePath), zap.Error(err))
-		return lastValue
+func (h *strategyStore) reloadSamplingStrategy(reloadFrom string, path string, lastValue string) string {
```
yurishkuro, line 109: The caller of this function already does checks for file vs. url. I would pass a loader function here that abstracts the source of the new config, e.g. `loader func () (string, error)`, and only deal with processing the new value.

plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -140,6 +191,28 @@ func loadStrategies(strategiesFile string) (*strategies, error) {
 	return &strategies, nil
 }
 
+func downloadStrategies(strategiesURL string) (*strategies, error) {
+	resp, err := http.Get(strategiesURL)
```
yurishkuro, line 195: this duplicates code from L112

plugin/sampling/strategystore/static/strategy_store.go
goku321: Yes, that's better. Was going to do the same but changed my mind later.

go.mod
goku321: I've synced my fork with the upstream master. Do I still need to exclude go.mod/go.sum files?

goku321:
@yurishkuro Thanks for the suggestions. I've made changes to the pull request

go.mod
yurishkuro: if this PR does not require new dependencies, it should not be updating dependency file. Please do `checkout master <file>` to get the originals.

plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -83,32 +84,76 @@ func (h *strategyStore) Close() {
...
 		case <-h.ctx.Done():
 			return
 		}
 	}
 }
 
-func (h *strategyStore) reloadSamplingStrategyFile(filePath string, lastValue string) string {
-	currBytes, err := ioutil.ReadFile(filepath.Clean(filePath))
+func (h *strategyStore) reloadSamplingStrategy(loader func() (string, error), lastValue string) string {
+	newValue, err := loader()
```
yurishkuro, line 147: I think it'd be much cleaner if the loader is created once in the constructor function

https://github.com/jaegertracing/jaeger/blob/0247828b9b0225151ac24b2156099935f94df6f6/plugin/sampling/strategystore/static/strategy_store.go#L58

plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -83,32 +84,76 @@ func (h *strategyStore) Close() {
...
+		h.logger.Error("failed to load sampling strategies", zap.Error(err))
 		return lastValue
 	}
-	newValue := string(currBytes)
 	if lastValue == newValue {
 		return lastValue
 	}
-	if err = h.updateSamplingStrategy(currBytes); err != nil {
-		h.logger.Error("failed to update sampling strategies from file", zap.Error(err))
+	if err := h.updateSamplingStrategy([]byte(newValue)); err != nil {
```
yurishkuro, line 155: Consider if `([]byte, error)` would be more appropriate signature for the `loader()`, to avoid casting back and forth.

plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -126,9 +171,27 @@ func (h *strategyStore) updateSamplingStrategy(bytes []byte) error {
 
 // TODO good candidate for a global util function
 func loadStrategies(strategiesFile string) (*strategies, error) {
-	if strategiesFile == "" {
-		return nil, nil
+	// Check if it's a URL.
+	if ok := isURL(strategiesFile); ok {
```
yurishkuro, line 175: there is an existing redundancy in the original source, please consolidate it. You should not need to check isURL twice in this source file.

yurishkuro (commented):
top-level question: should there be some local file fallback option if the URL is temporarily unavailable? Because if whichever service provides the strategy for download is down the collector would not be able to start.

goku321:
> top-level question: should there be some local file fallback option if the URL is temporarily unavailable? Because if whichever service provides the strategy for download is down the collector would not be able to start.

How about falling back to default strategies similar to when `options.StrategiesFile` is empty? Also, we can have a retry mechanism for URL option.

yurishkuro:
Default strategy is fine. The difference with StrategiesFile not found is in that case it's ok to fail because it likely indicates a mis-configuration, whereas in case of HTTP server not available then failing is counterproductive.

plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -83,35 +87,86 @@ func (h *strategyStore) Close() {
...
+	return buf.Bytes(), nil
+}
+
+func isURL(str string) bool {
+	u, err := url.Parse(str)
+	return err == nil && u.Scheme != "" && u.Host != ""
+}
+
+func samplingStrategyLoader(strategiesFile string) strategyLoader {
+	return func() ([]byte, error) {
```
yurishkuro, line 120: nit: you only need to test strategiesFile once, and return different functors. Otherwise `strategyLoader` is pointless, you could've just call the function to read stuff
```
if strategiesFile == "" {
    return func() ([]byte, error) {
        return []byte("null"), nil
    }
}
if isURL(strategiesFile) {
    return func() ([]byte, error) {
        return downloadSamplingStrategies(strategiesFile)
    }
}
```

plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -83,35 +87,86 @@ func (h *strategyStore) Close() {
...
+		if strategiesFile == "" {
+			// Using "null" so that it un-marshals to nil pointer.
+			return []byte("null"), nil
```
yurishkuro, line 123: may want to declare this as a constant

plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -83,35 +87,86 @@ func (h *strategyStore) Close() {
...
+			lastValue = h.reloadSamplingStrategy(loader, lastValue)
 		case <-h.ctx.Done():
 			return
 		}
 	}
 }
 
-func (h *strategyStore) reloadSamplingStrategyFile(filePath string, lastValue string) string {
-	currBytes, err := ioutil.ReadFile(filepath.Clean(filePath))
+func (h *strategyStore) reloadSamplingStrategy(loader strategyLoader, lastValue string) string {
```
yurishkuro, line 156:
```suggestion
func (h *strategyStore) reloadSamplingStrategy(loadFn strategyLoader, lastValue string) string {
```

yurishkuro:
please check why the DCO test is failing:
```
Commit sha: 932f6e0, Author: Deepak, Committer: Deepak; Expected "Deepak sah.sslpu@gmail.com", but got "Deepak Sah sah.sslpu@gmail.com".
```
You need to ensure full name in git matches full name in github

plugin/sampling/strategystore/static/strategy_store.go
goku321: Oops, I was being silly. I get this, thanks.

plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -83,35 +90,86 @@ func (h *strategyStore) Close() {
...
+	}
+
+	if isURL(strategiesFile) {
+		return func() ([]byte, error) {
+			return downloadSamplingStrategies(strategiesFile)
+		}
+	}
+
+	return func() ([]byte, error) {
+		currBytes, err := ioutil.ReadFile(strategiesFile)
```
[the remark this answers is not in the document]
yurishkuro: While you're at it, can you fix this line too? I.e. remove `nolint` comment and use `filepath.Clean`
```
./cmd/query/app/static_handler.go:210:	bytes, err := ioutil.ReadFile(uiConfig) /* nolint #nosec , this comes from an admin, not user */
```

plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -83,35 +90,86 @@ func (h *strategyStore) Close() {
...
+	return buf.Bytes(), nil
+}
+
+func isURL(str string) bool {
+	u, err := url.Parse(str)
+	return err == nil && u.Scheme != "" && u.Host != ""
+}
+
+func samplingStrategyLoader(strategiesFile string) strategyLoader {
+	if strategiesFile == "" {
```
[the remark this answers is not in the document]
goku321: Didn't quite understand this one. Can you elaborate it further? 😅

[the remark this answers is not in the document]
yurishkuro: There is a constructor function NewStrategyStore. This struct cannot do anything useful if the strategiesFile is empty, correct? So we could just return an error from there if the param is empty. If that's not sufficient (e.g. because it still needs to return the default strategy), then we can still do that in the constructor and just not execute all the loading/reloading functions if the file name is blank. This way we don't have to keep checking if the file path is empty or not.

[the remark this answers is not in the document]
goku321: I get it. Thanks for the explanation.

plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -83,35 +97,81 @@ func (h *strategyStore) Close() {
...
+		currBytes, err := ioutil.ReadFile(filepath.Clean(strategiesFile))
+		if err != nil {
+			return nil, fmt.Errorf("failed to read strategies file %s: %w", strategiesFile, err)
+		}
+		return currBytes, nil
+	}
+}
+
+func (h *strategyStore) autoUpdateStrategies(interval time.Duration, loader strategyLoader) {
+	lastValue := string(nullJSON)
```
yurishkuro, line 148: Technically, we could have already loaded something before calling this function, so lastValue would be different. But since it's a periodic check anyway, I think it's fine.

plugin/sampling/strategystore/static/strategy_store_test.go
```diff
@@ -293,6 +346,49 @@ func TestAutoUpdateStrategy(t *testing.T) {
...
+	require.NoError(t, err)
+	assert.EqualValues(t, makeResponse(sampling.SamplingStrategyType_PROBABILISTIC, 0.8), *s)
+
+	// verify that reloading in no-op
+	value = store.reloadSamplingStrategy(samplingStrategyLoader(mockServer.URL), string(srcBytes))
+	assert.Equal(t, string(srcBytes), value)
+
+	// update original strategies file with new probability of 0.9
+	newStr = strings.Replace(string(srcBytes), "0.8", "0.9", 1)
+	require.NoError(t, ioutil.WriteFile(srcFile, []byte(newStr), 0644))
```
yurishkuro, line 376: I don't think it's a good idea for the test to override the fixture file. In L316 it is already copied to a temp file that can be overwritten.

Also, if you're testing with a mock HTTP server, why use file in the first place? You can simply replace the content returned by the mock server since you can control it.

plugin/sampling/strategystore/static/strategy_store_test.go
```diff
@@ -293,6 +346,49 @@ func TestAutoUpdateStrategy(t *testing.T) {
 		time.Sleep(1 * time.Millisecond)
 	}
 	assert.EqualValues(t, makeResponse(sampling.SamplingStrategyType_PROBABILISTIC, 0.9), *s)
+
+	// Test auto update strategy with URL option.
```
yurishkuro, line 350: This seems like a pretty separate testing scenario, any reason why it cannot be moved into a new test function with appropriate name?

plugin/sampling/strategystore/static/strategy_store_test.go
```diff
@@ -62,6 +95,26 @@ func TestStrategyStore(t *testing.T) {
 	s, err = store.GetSamplingStrategy(context.Background(), "default")
 	require.NoError(t, err)
 	assert.EqualValues(t, makeResponse(sampling.SamplingStrategyType_PROBABILISTIC, 0.5), *s)
+
+	// Test default strategy when URL is temporarily unavailable.
```
yurishkuro, line 99: could this be in a separate test function?

plugin/sampling/strategystore/static/strategy_store_test.go
```diff
@@ -31,9 +33,40 @@ import (
...
+			w.Write(data)
+		}
+	}
+	return httptest.NewServer(http.HandlerFunc(f))
+}
+
 func TestStrategyStore(t *testing.T) {
 	_, err := NewStrategyStore(Options{StrategiesFile: "fileNotFound.json"}, zap.NewNop())
-	assert.EqualError(t, err, "failed to open strategies file: open fileNotFound.json: no such file or directory")
+	assert.EqualError(t, err, "failed to read strategies file fileNotFound.json: open fileNotFound.json: no such file or directory")
```
yurishkuro, line 69: I recommend using assert.Contains (as in L79) to check only for the text produced by our code, not the parts produced by stdlib functions (since they can change the text between Go versions).

yurishkuro (commented):
Looks great! I left a couple comments around the tests, otherwise LGTM.

plugin/sampling/strategystore/static/strategy_store_test.go
goku321: Yes, it makes sense to move it to a separate test function.

plugin/sampling/strategystore/static/strategy_store_test.go
```diff
@@ -51,22 +72,17 @@ func mockStrategyServer() *httptest.Server {
...
 		}
 	}
 	return httptest.NewServer(http.HandlerFunc(f))
 }
 
-func TestStrategyStore(t *testing.T) {
+func TestStrategyStoreWithFile(t *testing.T) {
 	_, err := NewStrategyStore(Options{StrategiesFile: "fileNotFound.json"}, zap.NewNop())
-	assert.EqualError(t, err, "failed to read strategies file fileNotFound.json: open fileNotFound.json: no such file or directory")
+	assert.Contains(t, err.Error(), "failed to read strategies file fileNotFound.json: open fileNotFound.json: no such file or directory")
```
yurishkuro, line 85:
```suggestion
	assert.Contains(t, err.Error(), "failed to read strategies file fileNotFound.json")
```
only include the string that we generate, not the stdlib portion

plugin/sampling/strategystore/static/strategy_store_test.go
```diff
@@ -346,38 +365,48 @@ func TestAutoUpdateStrategy(t *testing.T) {
...
-	require.NoError(t, ioutil.WriteFile(srcFile, []byte(newStr), 0644))
-	defer func() {
-		// replace original strategies file with old content.
-		require.NoError(t, ioutil.WriteFile(srcFile, originalBytes, 0644), "failed to restore original file content")
-	}()
+	value := store.reloadSamplingStrategy(samplingStrategyLoader(mockServer.URL), string(strategiesJSON))
+	assert.Equal(t, string(strategiesJSON), value)
+
+	// update original strategies with new probability of 0.9
+	strategiesJSON = []byte(`
```
yurishkuro: This is not a good approach to change package variable in-place. Even though it's a var, the general assumption is that package-level vars are constants (in fact I would change its type to string to make a real constant).

An even better approach is to convert it into a function that takes the probability value and returns the full JSON string via Sprintf. Then I would also change makeMockServer to return `(strategy *atomic.String, server *httptest.Server)` (using uber-go/atomic), so that you can manipulate the strategy externally, e.g.:
```go
mockStrategy, server := mockStrategyServer()
...
mockStrategy.Store(strategiesJSON(0.9))
...
```

yurishkuro:
please check the linter failures
```
plugin/sampling/strategystore/static/strategy_store_test.go:130:9: this value of err is never used (SA4006)
plugin/sampling/strategystore/static/strategy_store_test.go:484:11: this value of err is never used (SA4006)
```

yurishkuro (approved):

goku321:
> please check the linter failures
> 
> ```
> plugin/sampling/strategystore/static/strategy_store_test.go:130:9: this value of err is never used (SA4006)
> plugin/sampling/strategystore/static/strategy_store_test.go:484:11: this value of err is never used (SA4006)
> ```

Fixed the linter issues.

PR #2523, opened 2020-09-29
Otel Kafka Exporter Tests

joe-elliott:
- Improves Otel Kafka Exporter Coverage

- Gets Kafka exporter coverage to 100%

While writing these tests I felt strongly that we should remove the otel config merge testing from the individual receivers/exporters and only test it here: https://github.com/jaegertracing/jaeger/blob/master/cmd/opentelemetry/app/defaultconfig/default_config_test.go

However, if desired I can continue down this path!


yurishkuro:
> I felt strongly that we should remove the otel config merge testing from the individual receivers/exporters and only test it here

Don't know the details, but 100% agree that if we can consolidate/generalize testing, we should do it. The only sad thing about Go testing is that it does not aggregate coverage when code in a package is invoked from another package - it can do it but only if tests are run with explicit cross-pollination.

joe-elliott:
The method being tested `CreateDefaultConfig()` loads otel defaults and then overlays Jaeger cli/config file options.  The actual merging of the otel config happens in [default_config.go](https://github.com/jaegertracing/jaeger/blob/master/cmd/opentelemetry/app/defaultconfig/default_config.go#L73).  So it just feels like the otel merging tests should be over in that test file.  Coverage would not be impacted by moving these tests.

cmd/opentelemetry/app/exporter/kafkaexporter/kafka_exporter_test.go
```diff
@@ -51,38 +65,174 @@ func TestDefaultConfig(t *testing.T) {
...
-	assert.Equal(t, "123", kafkaCfg.Authentication.PlainText.Password)
-	assert.Equal(t, "ca.crt", kafkaCfg.Authentication.TLS.CAFile)
-	assert.Equal(t, "key.crt", kafkaCfg.Authentication.TLS.KeyFile)
-	assert.Equal(t, "cert.crt", kafkaCfg.Authentication.TLS.CertFile)
-	assert.Equal(t, true, kafkaCfg.Authentication.TLS.Insecure)
-	assert.Equal(t, "jaeger", kafkaCfg.Authentication.Kerberos.Realm)
-	assert.Equal(t, "/etc/foo", kafkaCfg.Authentication.Kerberos.ConfigPath)
-	assert.Equal(t, "from-jaeger-config", kafkaCfg.Authentication.Kerberos.Username)
+	actualLogs, _ := factory.CreateLogsExporter(context.Background(), component.ExporterCreateParams{}, nil)
+	expectedLogs, _ := factory.Wrapped.CreateLogsExporter(context.Background(), component.ExporterCreateParams{}, nil)
```
albertteoh, line 228: Should there be an `assert.NoError(t, err)`?

That is, could `err != nil` while `actualLogs == expectedLogs` (e.g. both `nil`)?

joe-elliott: No, it's not really testing that.  It's really just confirming that the `ExporterFactory` is just a thin wrapper around the internal `ExporterFactory`.

But you do bring up a good point.  I should be testing both the errs and the return.

joe-elliott:
@pavolloffay I believe this is ready if you'd like to have a look.  Also, are there any docs that specify exactly how the various config options (jaeger cli, jaeger config, otel config) are supposed to override each other?

joe-elliott:
Closing as this is all moving over to a new repo with a new way to handle configuration parsing.

https://github.com/jaegertracing/jaeger-otelcol

PR #2524, opened 2020-09-29
Added .nocover to otel mains

joe-elliott:
- Removing code coverage calculations from otel mains

- Added .nocover to all otel mains


yurishkuro:
I don't think this is necessary. Go does not report coverage for packages that have no test files. This is why we have a CI check that makes sure all packages have _some_ `*_test.go` files, unless the package only contains main or has .nocover marker. You can find some packages with empty_test.go because of that check (usually interface-only packages that don't have testable code).

yurishkuro:
Feels like we should document these practices somewhere, wanna take a stub at that?

joe-elliott:
I added `.nocover` b/c the standard agent/collector/ingester mains all have it.  So I presumed it was being used there to prevent those folders from counting toward code coverage.  I'll admit I didn't dig too deep on it :)

> Feels like we should document these practices somewhere, wanna take a stub at that?

Yup, I'll try to get something together.

yurishkuro (approved):
you're right

Issue #2503, opened 2020-09-22
Migrate lint tools to golangci-lint


pavolloffay:
Migrate `go vet`, `staticcheck`, `gosec` to https://golangci-lint.run/. It is a single tool that can run all static code analysis tools https://medium.com/wesionary-team/introduction-to-linters-in-go-and-know-about-golangci-lint-6486fb2864b3.

It's also used by the OTEL collector.

yurishkuro:
Looks good.

eundoosong:
Hello, I'd like to contribute on this

pavolloffay:
go for it @eundoosong

yurishkuro:
We also have a similar issue #538 to use gometalinter.

Issue #538, opened 2017-11-14
Use  gometalinter


yurishkuro:
https://github.com/alecthomas/gometalinter

also `badtime` linter: https://github.com/m3db/build-tools/tree/master/linters/badtime

yurishkuro:
From a trial run:
```
gometalinter.v1 --deadline=300s ./cmd/... ./crossdock/... ./model/... ./pkg/... ./plugin/... ./storage/... .
cmd/agent/app/builder.go:38:1:warning: agentServiceName is unused (deadcode)
cmd/agent/app/builder.go:38:1:warning: defaultCollectorServiceName is unused (deadcode)
cmd/agent/app/builder.go:59:1:warning: errNoReporters is unused (deadcode)
cmd/agent/app/processors/thrift_processor.go:107::warning: Errors unhandled.,LOW,HIGH (gas)
cmd/agent/app/agent.go:73::warning: Errors unhandled.,LOW,HIGH (gas)
cmd/collector/main.go:116::warning: Errors unhandled.,LOW,HIGH (gas)
cmd/collector/app/zipkin/json.go:210::warning: Errors unhandled.,LOW,HIGH (gas)
cmd/collector/app/zipkin/json.go:215::warning: Errors unhandled.,LOW,HIGH (gas)
cmd/collector/app/zipkin/json.go:220::warning: Errors unhandled.,LOW,HIGH (gas)
cmd/query/app/static_handler.go:60::warning: Errors unhandled.,LOW,HIGH (gas)
cmd/query/app/static_handler.go:108::warning: Errors unhandled.,LOW,HIGH (gas)
storage/spanstore/memory/memory.go:58::warning: Errors unhandled.,LOW,HIGH (gas)
crossdock/services/query.go:72::warning: Errors unhandled.,LOW,HIGH (gas)
cmd/agent/app/httpserver/thrift-0.9.2/ttypes.go:596::warning: cyclomatic complexity 17 of function (*SamplingStrategyResponse).Read() is high (> 10) (gocyclo)
cmd/agent/app/httpserver/thrift-0.9.2/ttypes.go:389::warning: cyclomatic complexity 15 of function (*PerOperationSamplingStrategies).Read() is high (> 10) (gocyclo)
cmd/agent/app/httpserver/thrift-0.9.2/ttypes.go:263::warning: cyclomatic complexity 13 of function (*OperationSamplingStrategy).Read() is high (> 10) (gocyclo)
[...]

PR #2529, opened 2020-10-01
Remove Go report card

yurishkuro:
The way this service works does not allow for investigation of the issues that it raises. Since most of those are just linters anyway, we'd be better off running the actual linters, e.g. #2503, #538.


jpkrohling (approved):

jpkrohling:
The CI job `TESTS=true COVERAGE=true` failed. Restarted.

I think we don't have a `travis_retry` for that, but perhaps we should add? Or is it a specific flaky test that is failing? I forgot to look at the CI logs for the failure :/

PR #2535, opened 2020-10-04
Remove unnecessary ServiceName index seek if tags query is available

burmanm:
- Reduce the amount of queries against the badger backend store if Tags are used as a query parameter. Also, if the Tags is the only query, then we also remove single merge-join phase.

- Tags' key already includes the ServiceName so it makes no sense to add the ServiceName index seek as we can filter it all in a single index seek.


jpkrohling (commented):
Looks sane to me. Just to confirm: this is covered with test cases already, right?

burmanm:
Multiple tests have Tags-queries, yes. The ``TestRandomTraceID`` fetches a query with Tags & ServiceName parameters and as such tests this code path with just the Tags-query being executed.

jpkrohling (approved):

jpkrohling:
Nice to see a PR from you :-)

PR #2536, opened 2020-10-05
[hotrod] Add links to traces

yurishkuro:
Wanted to add this long time ago - a link to the trace from each request using the search URL.

image


examples/hotrod/services/frontend/web_assets/index.html
```diff
@@ -85,7 +85,8 @@ <h4><em>Rides On Demand</em></h4>
       var after = Date.now();
       console.log(data);
       var duration = formatDuration(data.ETA);
-      freshCar.html('HotROD <b>' + data.Driver + '</b> arriving in ' + duration + ' [req: ' + requestID + ', latency: ' + (after-before) + 'ms]');
+      var trace = 'http://localhost:16686/search?limit=20&lookback=1h&service=frontend&tags=%7B%22driver%22%3A%22' + data.Driver + '%22%7D';
```
jpkrohling, line 88: Is it too hard to add a field somewhere with the location of the Jaeger installation? Or to copy the search terms, so that people can just paste into their remote Jaeger instance?

joe-elliott: Yeah, I'm concerned that hotrod is used in a lot of different examples and this link may not always be valid.  I'd say in most cases where people are running hotrod localhost:16686 will be the Jaeger UI, but when it's not this may be confusing.

yurishkuro: added config option `-j http://localhost:16686`

joe-elliott (approved):
Looks good to me.

Issue #1887, opened 2019-10-29
Cannot build binaries


saulshanabrook:
I am trying to follow the CONTRIBUTING.md to build the binaries, in order to contribute them to conda forge https://github.com/conda-forge/staged-recipes/pull/9945

However, when I run `build-binaries-darwin` (I am on a mac), I get:

```bash
$ make build-binaries-darwin
GOOS=darwin /Applications/Xcode.app/Contents/Developer/usr/bin/make build-platform-binaries
CGO_ENABLED=0 installsuffix=cgo go build -o ./cmd/agent/agent-darwin -ldflags "-X github.com/jaegertracing/jaeger/pkg/version.commitSHA=6887ae4b92814c7e54592b5e7081d92a96cdcf64 -X github.com/jaegertracing/jaeger/pkg/version.latestVersion=v1.14.0 -X github.com/jaegertracing/jaeger/pkg/version.date=2019-10-29T16:51:07Z" ./cmd/agent/main.go
esc -pkg mappings -o plugin/storage/es/mappings/gen_assets.go -ignore assets -prefix plugin/storage/es/mappings plugin/storage/es/mappings
CGO_ENABLED=0 installsuffix=cgo go build -o ./cmd/collector/collector-darwin -ldflags "-X github.com/jaegertracing/jaeger/pkg/version.commitSHA=6887ae4b92814c7e54592b5e7081d92a96cdcf64 -X github.com/jaegertracing/jaeger/pkg/version.latestVersion=v1.14.0 -X github.com/jaegertracing/jaeger/pkg/version.date=2019-10-29T16:51:10Z" ./cmd/collector/main.go
CGO_ENABLED=0 installsuffix=cgo go build -tags ui -o ./cmd/query/query-darwin -ldflags "-X github.com/jaegertracing/jaeger/pkg/version.commitSHA=6887ae4b92814c7e54592b5e7081d92a96cdcf64 -X github.com/jaegertracing/jaeger/pkg/version.latestVersion=v1.14.0 -X github.com/jaegertracing/jaeger/pkg/version.date=2019-10-29T16:51:14Z" ./cmd/query/main.go
# github.com/jaegertracing/jaeger/cmd/query/app/ui
cmd/query/app/ui/actual.go:24:19: undefined: assets.FS
make[1]: *** [build-query] Error 2
make: *** [build-binaries-darwin] Error 2
```

Any ideas?

yurishkuro:
did you run this first?
```
[...]

PR #2542, opened 2020-10-08
Adding missing build instruction #1887

listaction:
Adding missing build instruction that allows jaeger binaries to be built.


yurishkuro (approved):
Thanks

yurishkuro:
Please makes sure the commit is signed.

Issue #2493, opened 2020-09-18
Debug images with delve and other base image


jpkrohling:
## Requirement - what kind of business use case are you trying to solve?

We discussed having a different base image for Jaeger, for better troubleshooting, but never moved forward, mostly because the maintenance burden of having two very similar images wasn't worth it. But @bjung82 had a great question, on whether we have a debug image with delve enabled, for remote debugging. 

This sounds like a great idea, and given the recent issues with the agent, sounds like something that would really help users.

## Problem - what in Jaeger blocks you from solving the requirement?

Sometimes, observability is not enough ;-)

## Proposal - what do you suggest to solve the problem or improve the existing situation?

This would require a new Dockerfile for all components, new targets in the Makefile. 

Context: #2465, and would probably also be helpful for #2452

vkartik97:
can I take this up ?

jpkrohling:
It's yours!

jpkrohling:
@aebirim wants to work on this one as well.

yurishkuro:
> whether we have a debug image with delve enabled, for remote debugging.

@jpkrohling is delve needed to be in the image for this?

jpkrohling:
Delve needs to be there, and the Jaeger binaries in the image should not be optimized, so that Delve can extract name and line information.

Issue #2452, opened 2020-09-03
Memory peaks on agent v1.19.2


zigmund:
**Describe the bug**
One group of jaeger agents (3 replicas), deployed on kubernetes. Not so loaded system - 500-700 spans per second.
Used v1.13.1 for year without any issues, upgraded all componenet to 1.19.2 (also replaced Tchannel with gRPC) and agent memory usage become unstable. Before upgrade agent instances used as low as 16 mb with 64 limit, but after upgrade memory usage peaks appeared and agent got oomkilled. I raised limits, but even 512 mb is not enough.

After few hours and tens of oomkills, I downgraded agent to 1.18.1 version and see no issues so far.

**To Reproduce**
1. Deploy agent v1.19.2
2. Send 500-700 spans per second
3. ???
4. Observe memory usage peaks.

**Expected behavior**
Agent memory usage close to lower versions.

**Screenshots**
![image](https://user-images.githubusercontent.com/1713091/92070460-dbcdd600-edcd-11ea-82eb-066d60b948df.png)

**Version (please complete the following information):**
 - OS: Kubernetes nodes based on Ubuntu 20.04 LTS, 5.4.0-39 kernel
 - Jaeger version: 1.19.2
 - Deployment: Kubernetes

**What troubleshooting steps did you try?**
Collected /metrics, will provide if needed.

**Additional context**
-

jpkrohling:
When you downgraded to 1.18.1, did you change any other configuration? Would you mind sharing your agent configuration?

zigmund:
@jpkrohling the only agent parameter used before and after downgrade is `--reporter.grpc.host-port=jaeger-collector:14250`.

Also played for a while after got first OOMkills with parameters, but reverted changes singe they didn't helped much:
```
        - --processor.jaeger-binary.server-queue-size=...
        - --processor.jaeger-binary.workers=...
        - --processor.jaeger-compact.server-queue-size=...
        - --processor.jaeger-compact.workers=...
        - --processor.zipkin-compact.server-queue-size=...
        - --processor.zipkin-compact.workers=...
```

jpkrohling:
I'll take a look at list of commits affecting the agent between 1.18.1 and 1.19.2, perhaps I can spot a suspect there. I'll also try to reproduce the problem here. Do you think it would be reproducible with `minikube` and a simple agent with 64Mi of requests , 256Mi limit? Jaeger's `tracegen` would probably generate enough load ;-)

[...]

PR #2545, opened 2020-10-09
Create debug docker images for jaeger backends

Ashmita152:

Tackles https://github.com/jaegertracing/jaeger/issues/2493

As part of this PR, we are creating separate images for jaeger backend components.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -67,7 +67,7 @@ script:
   - export BRANCH=$(if [ "$TRAVIS_PULL_REQUEST" == "false" ]; then echo $TRAVIS_BRANCH; else echo $TRAVIS_PULL_REQUEST_BRANCH; fi)
   - if [ "$TESTS" == true ]; then make test-ci ; else echo 'skipping tests'; fi
   - if [ "$PROTO_GEN_TEST" == true ]; then make proto && git diff --name-status --exit-code ; else echo 'skipping protoc validation'; fi
-  - if [ "$ALL_IN_ONE" == true ]; then bash ./scripts/travis/build-all-in-one-image.sh ; else echo 'skipping all_in_one'; fi
+  - if [ "$ALL_IN_ONE" == true ]; then make create-baseimg-debugimg && bash ./scripts/travis/build-all-in-one-image.sh ; else echo 'skipping all_in_one'; fi
   - if [ "$CROSSDOCK" == true ]; then travis_retry bash ./scripts/travis/build-crossdock.sh ; else echo 'skipping crossdock'; fi
   - if [ "$CROSSDOCK_OTEL" == true ]; then travis_retry make build-crossdock crossdock-otel ; else echo 'skipping OpenTelemetry crossdock'; fi
   - if [ "$DOCKER" == true ]; then bash ./scripts/travis/build-docker-images.sh ; else echo 'skipping build-docker-images'; fi
--- a/Makefile
+++ b/Makefile
@@ -350,7 +350,7 @@ docker-images-jaeger-backend-debug: SUFFIX = -debug
 docker-images-jaeger-backend docker-images-jaeger-backend-debug: create-baseimg create-debugimg
 	for component in agent collector query ingester ; do \
 		docker build --target $(TARGET) \
-			--tag $(DOCKER_NAMESPACE)/jaeger-$$component:${DOCKER_TAG}$(SUFFIX) \
+			--tag $(DOCKER_NAMESPACE)/jaeger-$$component$(SUFFIX):${DOCKER_TAG} \
 			--build-arg base_image=$(BASE_IMAGE) \
 			--build-arg debug_image=$(DEBUG_IMAGE) \
 			--build-arg TARGETARCH=$(GOARCH) \
--- a/Makefile
+++ b/Makefile
@@ -245,12 +245,12 @@ cmd/query/app/ui/placeholder/gen_assets.go: cmd/query/app/ui/placeholder/public/
 build-all-in-one-linux:
 	GOOS=linux $(MAKE) build-all-in-one
 
-.PHONY: build-all-in-one
-build-all-in-one: build-ui elasticsearch-mappings
-	$(GOBUILD) -tags ui -o ./cmd/all-in-one/all-in-one-$(GOOS)-$(GOARCH) $(BUILD_INFO) ./cmd/all-in-one/main.go
+build-all-in-one-debug build-agent-debug build-query-debug build-collector-debug build-ingester-debug: DISABLE_OPTIMIZATIONS = -gcflags="all=-N -l"
+build-all-in-one-debug build-agent-debug build-query-debug build-collector-debug build-ingester-debug: SUFFIX = -debug
 
-build-agent-debug build-query-debug build-collector-debug build-ingester-debug: DISABLE_OPTIMIZATIONS = -gcflags="all=-N -l"
-build-agent-debug build-query-debug build-collector-debug build-ingester-debug: SUFFIX = -debug
+.PHONY: build-all-in-one build-all-in-one-debug
+build-all-in-one build-all-in-one-debug: build-ui elasticsearch-mappings
+	$(GOBUILD) $(DISABLE_OPTIMIZATIONS) -tags ui -o ./cmd/all-in-one/all-in-one$(SUFFIX)-$(GOOS)-$(GOARCH) $(BUILD_INFO) ./cmd/all-in-one/main.go
 
 .PHONY: build-agent build-agent-debug
 build-agent build-agent-debug:
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,8 @@ JAEGER_IMPORT_PATH=github.com/jaegertracing/jaeger
 STORAGE_PKGS = ./plugin/storage/integration/...
 OTEL_COLLECTOR_DIR = ./cmd/opentelemetry
 
+include docker/Makefile
+
 # all .go files that are not auto-generated and should be auto-formatted and linted.
 ALL_SRC := $(shell find . -name '*.go' \
 				   -not -name 'doc.go' \
@@ -345,9 +347,14 @@ docker-images-jaeger-backend-debug: TARGET = debug
 docker-images-jaeger-backend-debug: SUFFIX = -debug
 
 .PHONY: docker-images-jaeger-backend docker-images-jaeger-backend-debug
-docker-images-jaeger-backend docker-images-jaeger-backend-debug:
+docker-images-jaeger-backend docker-images-jaeger-backend-debug: create-baseimg create-debugimg
 	for component in agent collector query ingester ; do \
-		docker build --target $(TARGET) -t $(DOCKER_NAMESPACE)/jaeger-$$component:${DOCKER_TAG}$(SUFFIX) cmd/$$component --build-arg TARGETARCH=$(GOARCH) ; \
+		docker build --target $(TARGET) \
+			--tag $(DOCKER_NAMESPACE)/jaeger-$$component:${DOCKER_TAG}$(SUFFIX) \
+			--build-arg base_image=$(BASE_IMAGE) \
+			--build-arg debug_image=$(DEBUG_IMAGE) \
+			--build-arg TARGETARCH=$(GOARCH) \
+			cmd/$$component ; \
 		echo "Finished building $$component ==============" ; \
 	done
 	docker build -t $(DOCKER_NAMESPACE)/jaeger-opentelemetry-collector:${DOCKER_TAG} -f ${OTEL_COLLECTOR_DIR}/cmd/collector/Dockerfile cmd/opentelemetry/cmd/collector --build-arg TARGETARCH=$(GOARCH)
--- a/Makefile
+++ b/Makefile
@@ -323,10 +323,13 @@ docker-images-elastic:
 	docker build -t $(DOCKER_NAMESPACE)/jaeger-es-rollover:${DOCKER_TAG} plugin/storage/es -f plugin/storage/es/Dockerfile.rollover
 	@echo "Finished building jaeger-es-indices-clean =============="
 
-.PHONY: docker-images-jaeger-backend
-docker-images-jaeger-backend:
+docker-images-jaeger-backend: TARGET = release
+docker-images-jaeger-backend-debug: TARGET = debug
+
+.PHONY: docker-images-jaeger-backend docker-images-jaeger-backend-debug
+docker-images-jaeger-backend docker-images-jaeger-backend-debug:
 	for component in agent collector query ingester ; do \
-		docker build -t $(DOCKER_NAMESPACE)/jaeger-$$component:${DOCKER_TAG} cmd/$$component --build-arg TARGETARCH=$(GOARCH) ; \
+		docker build --target $(TARGET) -t $(DOCKER_NAMESPACE)/jaeger-$$component:${DOCKER_TAG} cmd/$$component --build-arg TARGETARCH=$(GOARCH) ; \
 		echo "Finished building $$component ==============" ; \
 	done
 	docker build -t $(DOCKER_NAMESPACE)/jaeger-opentelemetry-collector:${DOCKER_TAG} -f ${OTEL_COLLECTOR_DIR}/cmd/collector/Dockerfile cmd/opentelemetry/cmd/collector --build-arg TARGETARCH=$(GOARCH)
@@ -339,7 +342,11 @@ docker-images-tracegen:
 	@echo "Finished building jaeger-tracegen =============="
 
 .PHONY: docker-images-only
-docker-images-only: docker-images-cassandra docker-images-elastic docker-images-jaeger-backend docker-images-tracegen
+docker-images-only: docker-images-cassandra \
+	docker-images-elastic \
+	docker-images-jaeger-backend \
+	docker-images-jaeger-backend-debug \
+	docker-images-tracegen
 
 .PHONY: docker-push
 docker-push:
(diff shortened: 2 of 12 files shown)


yurishkuro (commented):
Thanks for the PR. 

I would strongly prefer having a single Dockerfile for each binary, not copy-pasting content across two files. For example, you are already missing `ARG USER_UID=10001` directive that's in master.

For example, would the following work?

```
# default build
FROM alpine:latest as certs
RUN apk add --update --no-cache ca-certificates
COPY ./scripts/fake-dlv /go/bin/dlv

# debug build
FROM golang:1.15.2-alpine as debug
...

# for debug builds, these vars should be replaced with BASEIMAGE=BUILDIMAGE=debug
ARG BASEIMAGE=scratch
ARG BUILDIMAGE=certs

FROM $BASEIMAGE
COPY --from=$BUILDIMAGE /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
COPY --from=$BUILDIMAGE /go/bin/dlv /go/bin/dlv
...
```

Ashmita152:
Hi @yurishkuro 

Thank you for the review. It definitely makes sense to generate two images out of a single Dockerfile. Let me work on it and update the PR.

Makefile
```diff
@@ -328,13 +328,19 @@ docker-images-jaeger-backend:
 	docker build -t $(DOCKER_NAMESPACE)/jaeger-opentelemetry-agent:${DOCKER_TAG} -f ${OTEL_COLLECTOR_DIR}/cmd/agent/Dockerfile cmd/opentelemetry/cmd/agent --build-arg TARGETARCH=$(GOARCH)
 	docker build -t $(DOCKER_NAMESPACE)/jaeger-opentelemetry-ingester:${DOCKER_TAG} -f ${OTEL_COLLECTOR_DIR}/cmd/ingester/Dockerfile cmd/opentelemetry/cmd/ingester --build-arg TARGETARCH=$(GOARCH)
 
+.PHONY: docker-images-jaeger-backend-debug
```
yurishkuro, line 331: I think this target should just delegate to `docker-images-jaeger-backend` with some extra variables

Ashmita152:
I have another question related to this debug image:

- Do we need any other debugging tool as part of this image other than delve ?

Makefile
```diff
@@ -328,13 +328,19 @@ docker-images-jaeger-backend:
...
+	done
+
 .PHONY: docker-images-tracegen
 docker-images-tracegen:
 	docker build -t $(DOCKER_NAMESPACE)/jaeger-tracegen:${DOCKER_TAG} cmd/tracegen/ --build-arg TARGETARCH=$(GOARCH)
 	@echo "Finished building jaeger-tracegen =============="
 
 .PHONY: docker-images-only
-docker-images-only: docker-images-cassandra docker-images-elastic docker-images-jaeger-backend docker-images-tracegen
+docker-images-only: docker-images-cassandra docker-images-elastic docker-images-jaeger-backend docker-images-jaeger-backend-debug docker-images-tracegen
```
yurishkuro, line 343: The long line is getting difficult to grok. Is there a way to break the dependency targets into multiple lines, something like:
```
a: \
  b \
  c \
  d
```
?

yurishkuro:
> Do we need any other debugging tool as part of this image other than delve ?

I think the biggest ask was not so much the debug tools, but just _any_ tools, because when we build from `scratch` we don't even get a shell for minimal troubleshooting.

jpkrohling:
> I think the biggest ask was not so much the debug tools, but just any tools, because when we build from scratch we don't even get a shell for minimal troubleshooting.

True. For the debug images, I would even base it on something else. Note that the CMD instruction should also be custom for the debug image.

Ashmita152:
Few more questions I have is:

1) let's say a user face an issue in production. Having separate image for debugging means asking user to deploy that debug image in production separately. In case of systems like k8s, we can't have just one pod of different image, that means we need to deploy this debug image in every pod. Do you think this is the right approach. Somehow I am not convinced with this approach. I would better provide a mechanism to attach a sidecar debug container.

2) Is there any particular reason why we are using scratch as the base image instead of let's say alpine (which has minimal shell) ?

jpkrohling:
We talked before about changing the base image (#2116), and to be honest, I forgot to add it to my queue. I guess it's on yours now :-)

That said, for this specific issue here, users would need to deploy a different image anyway, as the process has to be started with `dlv` as the `CMD`.

Still about your first point: while I agree that people should be able to enter the container when they face an issue, I don't think it's unreasonable to ask them to deploy a debug image if problems can't be easily detected (#2452 comes to my mind). They can create a second deployment with this special image, and change the fronting service to direct part of the traffic to it: if there are 10 matching pods, one of them being from the debug deployment, then only 10% of the traffic will reach the debug image.

Ashmita152:
Thank you for answering my questions earlier. This PR is ready for review again. I have updated the PR with:

- Merging two dockerfiles into one and using docker build command's `--target` option (https://docs.docker.com/develop/develop-images/multistage-build/#stop-at-a-specific-build-stage) to build two images.
- Using alpine as the base image for docker images.

I have one followup question:

While trying out delve I found that delve suggest to compile with optimisations disabled for golang binaries (https://github.com/go-delve/delve/blob/master/Documentation/usage/dlv_exec.md) Would you suggest for jaeger components binaries used in debug images ?

Ashmita152:
I noticed @jpkrohling comment in the github issue about disabling optimizations, hence updated the PR accordingly.

Makefile
```diff
@@ -302,7 +305,21 @@ build-binaries-ppc64le:
 	GOOS=linux GOARCH=ppc64le $(MAKE) build-platform-binaries
 
 .PHONY: build-platform-binaries
-build-platform-binaries: build-agent build-collector build-query build-ingester build-all-in-one build-examples build-tracegen build-otel-collector build-otel-agent build-otel-ingester build-otel-all-in-one
+build-platform-binaries: build-agent \
+	build-agent-debug \
```
jpkrohling, line 309: Nice change! Way easier to follow what's going on.

jpkrohling (commented):
It's looking good! I need to try it out myself, but I have a couple of comments:

1. looks like it's missing the all-in-one image?
1. can we somehow have a base image for our components? Or a script that generates the Dockerfile? Most of it seems to be the same plumbing code, and maintaining is probably becoming harder.
1. perhaps as a follow-up PR, but we need to publish the images as well. Take a look at the Travis configuration and look for `DOCKER=true`: that's how/where we are building the images and pushing to the Docker Hub.
1. documentation, which would belong to jaegertracing/documentation

Ashmita152:
Hi @jpkrohling 

Thank you for giving detailed feedback earlier. I have updated the branch with the changes for second point in the review.

> 2. can we somehow have a base image for our components? Or a script that generates the Dockerfile? Most of it seems to be the same plumbing code, and maintaining is probably becoming harder.

 Can I ask for a review again please. I will work on rest three feedbacks once we're convinced with the dockerfiles layout changes since those three will depend on it.

Makefile
```diff
@@ -318,10 +337,19 @@ docker-images-elastic:
...
+docker-images-jaeger-backend: TARGET = release
+docker-images-jaeger-backend-debug: TARGET = debug
+docker-images-jaeger-backend-debug: SUFFIX = -debug
+
+.PHONY: docker-images-jaeger-backend docker-images-jaeger-backend-debug
+docker-images-jaeger-backend docker-images-jaeger-backend-debug: create-baseimg create-debugimg
 	for component in agent collector query ingester ; do \
-		docker build -t $(DOCKER_NAMESPACE)/jaeger-$$component:${DOCKER_TAG} cmd/$$component --build-arg TARGETARCH=$(GOARCH) ; \
+		docker build --target $(TARGET) \
+			--tag $(DOCKER_NAMESPACE)/jaeger-$$component:${DOCKER_TAG}$(SUFFIX) \
```
jpkrohling, line 348: This will yield a tag like this: `jaegertracing/jaeger-agent:latest-debug`. I think I would prefer `jaegertracing/jaeger-agent-debug:latest` instead, but I'm open to hear arguments/experiences with debug versions as well.

cmd/agent/Dockerfile
```diff
@@ -1,13 +1,18 @@
...
+EXPOSE 5775/udp 6831/udp 6832/udp 5778/tcp
 ENTRYPOINT ["/go/bin/agent-linux"]
+USER ${USER_UID}
 
+FROM $debug_image AS debug
+ARG TARGETARCH=amd64
+ARG USER_UID=10001
+COPY agent-debug-linux-$TARGETARCH /go/bin/agent-linux
+EXPOSE 5775/udp 6831/udp 6832/udp 5778/tcp 12345/tcp
+CMD ["/go/bin/dlv", "exec", "/go/bin/agent-linux", "--headless", "--listen=:12345", "--api-version=2", "--log"]
```
jpkrohling, line 17: Shouldn't this be an `ENTRYPOINT` instead? Also, you might want to add `--` as the final argument, so that arguments passed when executing the image are relayed to the underlying binary instead of to dlv:

```
ENTRYPOINT ["/go/bin/dlv", "exec", "/go/bin/collector-linux", "--headless", "--listen=:12345", "--api-version=2", "--log", "--"]
```

Like:

```console
$ docker run -p 14250:14250 -p 12345:12345 -e SPAN_STORAGE_TYPE=memory jaegertracing/jaeger-collector:latest-debug --log-level=debug
API server listening at: [::]:12345
2020-10-13T08:25:42Z info layer=debugger launching process with args: [/go/bin/collector-linux --log-level=debug]
```

I was able to connect to the process in the container, but I wasn't able to set a breakpoint via Visual Studio Code. Were you able to accomplish that? Here's the configuration I'm using:

```json
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Launch remote",
            "type": "go",
            "request": "attach",
            "mode": "remote",
            "remotePath": "/go/bin/collector-linux",
            "port": 12345,
            "host": "127.0.0.1",
            "cwd": "/home/jpkroehling/Projects/src/github.com/jaegertracing/jaeger",
        }
    ]
}
```

jpkrohling (commented):
I'm really excited about this change, and I think we are almost at a working state.

cmd/agent/Dockerfile
Ashmita152: Thank you for the feedback @jpkrohling 

I am not sure but I think since we don't have the source code inside the container that's why we aren't able to set the breakpoint. Let me look into it more.

jpkrohling: Alright, let me know once you figure it out.

Ashmita152: I followed this blog: https://golangforall.com/en/post/go-docker-delve-remote-debug.html to debug the breakpoint issue. I think the issue is due to VSCode's go module detection which isn't working properly. When I tried adding breakpoint in VSCode I got this error:

```
2020-10-13T13:17:04Z debug layer=rpc -> *rpc2.CreateBreakpointOut{"Breakpoint":{"id":0,"name":"","addr":0,"addrs":null,"file":"","line":0,"Cond":"","continue":false,"traceReturn":false,"goroutine":false,"stacktrace":0,"LoadArgs":null,"LoadLocals":null,"hitCount":null,"totalHitCount":0}} error: "could not find file /go/bin/collector-linux/cmd/collector/app/server/http.go"
```

You can notice it's looking for code as per the binary path inside container.

Then I tried with Goland IDE and it worked well.

```
2020-10-13T13:25:58Z info layer=debugger created breakpoint: &api.Breakpoint{ID:21, Name:"", Addr:0xf43f41, Addrs:[]uint64{0xf43f41}, File:"github.com/jaegertracing/jaeger/cmd/collector/app/handler/http_handler.go", Line:59, FunctionName:"github.com/jaegertracing/jaeger/cmd/collector/app/handler.(*APIHandler).RegisterRoutes", Cond:"", Tracepoint:false, TraceReturn:false, Goroutine:false, Stacktrace:0, Variables:[]string(nil), LoadArgs:(*api.LoadConfig)(nil), LoadLocals:(*api.LoadConfig)(nil), HitCount:map[string]uint64{}, TotalHitCount:0x0}
```

These logs came after adding flag `--log-output=debugger,debuglineerr,gdbwire,lldbout,rpc` to the entry-point.

I have updated the branch with the above feedbacks related to CMD to ENTRYPOINT conversion and docker image name change. Thank you.

jpkrohling: I actually got it to work: 

```json
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Launch remote",
            "type": "go",
            "request": "attach",
            "mode": "remote",
            "remotePath": "",
            "port": 12345,
            "host": "127.0.0.1",
            "cwd": "${workspaceRoot}",
        }
    ]
}
```

I think it means you can go ahead :-)

jpkrohling:
To recap, this PR is still missing:

- [ ] apply the same to the all-in-one binary/container image
- [ ] documentation

cmd/query/Dockerfile
```diff
@@ -1,12 +1,14 @@
...
 COPY query-linux-$TARGETARCH /go/bin/query-linux
-
+EXPOSE 16686/tcp
 ENTRYPOINT ["/go/bin/query-linux"]
+
+FROM $debug_image AS debug
+ARG TARGETARCH=amd64
+COPY query-debug-linux-$TARGETARCH /go/bin/query-linux
+EXPOSE 12345/tcp 16686/tcp
+ENTRYPOINT ["/go/bin/dlv", "exec", "/go/bin/query-linux", "--headless", "--listen=:12345", "--api-version=2", "--accept-multiclient", "--log", "--log-output=debugger,debuglineerr,gdbwire,lldbout,rpc", "--"]
```
jpkrohling, line 14: I think we can make it less verbose: people debugging a Jaeger component would be interested mostly (or only) in the Jaeger output.

scripts/travis/build-all-in-one-image.sh
```diff
@@ -40,7 +40,23 @@ upload_to_docker() {
 
 make build-all-in-one GOOS=linux GOARCH=$GOARCH
 repo=jaegertracing/all-in-one
-docker build -f cmd/all-in-one/Dockerfile -t $repo:latest cmd/all-in-one --build-arg TARGETARCH=$GOARCH
+docker build -f cmd/all-in-one/Dockerfile \
+        --target release \
+        --tag $repo:latest cmd/all-in-one \
+        --build-arg base_image=localhost/baseimg:1.0.0-alpine-3.12 \
```
jpkrohling, line 46: Not sure we have everything tied correctly: this image is created via `create-baseimg`, which isn't being called when building the all-in-one:

https://github.com/jaegertracing/jaeger/blob/3ae29da424085fda3dcafbd42f0f0fc5f107e945/.travis.yml#L70

Perhaps we need a Makefile target to build the all-in-one, which builds the base/debug images and calls this script?

Ashmita152: Thank you for catching it. I've updated the PR with a fix. Hopefully it will solve the issue.

Ashmita152:
I am still working on the documentation part.

Ashmita152:
Hello @jpkrohling 

I have raised a [PR](https://github.com/jaegertracing/documentation/pull/452) in documentation repo. I wasn't sure about what to document so most likely I am missing something. Thank you.

jpkrohling:
I'll do a final check next Monday, but looks good to me! If anyone else wants to try this out before that, it would be awesome.

jpkrohling (approved):
LGTM. I tested the collector image before and just tested the all-in-one and it looks good!

Ashmita152:
Resolved the merge conflict due to another PR merge in Makefile.

jpkrohling (approved):

jpkrohling:
Thanks for your contribution!

.travis.yml
```diff
@@ -67,7 +67,7 @@ script:
   - export BRANCH=$(if [ "$TRAVIS_PULL_REQUEST" == "false" ]; then echo $TRAVIS_BRANCH; else echo $TRAVIS_PULL_REQUEST_BRANCH; fi)
   - if [ "$TESTS" == true ]; then make test-ci ; else echo 'skipping tests'; fi
   - if [ "$PROTO_GEN_TEST" == true ]; then make proto && git diff --name-status --exit-code ; else echo 'skipping protoc validation'; fi
-  - if [ "$ALL_IN_ONE" == true ]; then bash ./scripts/travis/build-all-in-one-image.sh ; else echo 'skipping all_in_one'; fi
+  - if [ "$ALL_IN_ONE" == true ]; then make create-baseimg-debugimg && bash ./scripts/travis/build-all-in-one-image.sh ; else echo 'skipping all_in_one'; fi
```
yurishkuro, line 70: Why is the extra step `make create-baseimg-debugimg` needed here but not in DOCKER step in L73?
Also, why not call it from `./scripts/travis/build-all-in-one-image.sh`?

Ashmita152: > Why is the extra step make create-baseimg-debugimg needed here but not in DOCKER step in L73?
My understanding is that script calls: make docker -> make docker-images-only -> make docker-images-jaeger-backend-debug which has dependencies on create-baseimg and create-debugimg.

> Also, why not call it from ./scripts/travis/build-all-in-one-image.sh?
We can do it if you think that is the better place.

jpkrohling: Should probably be done after #2325, as it touches this code as well.

morlay: if need to support multi arch for other components.

base images need to published with multi arch too. 
or just upgrade Dockerfile like https://github.com/morlay/jaeger/blob/master/cmd/all-in-one/Dockerfile

cmd/agent/Dockerfile
```diff
@@ -1,13 +1,18 @@
-FROM alpine:latest as certs
-RUN apk add --update --no-cache ca-certificates
+ARG base_image
+ARG debug_image
 
-FROM scratch
+FROM $base_image AS release
 ARG TARGETARCH=amd64
 ARG USER_UID=10001
```
yurishkuro, line 6: @Ashmita152 was there a reason for TARGETARCH and USER_UID args to be repeated in each target, instead of being defined at the top of the file?

If not, probably another opportunity to clean up / DRY the docker files.

Ashmita152: Sure, I will clean that today.

PR #2547, opened 2020-10-09
Inject version info into index.html

yurishkuro:
Part of https://github.com/jaegertracing/jaeger-ui/issues/365

### DIFF
--- a/cmd/query/app/static_handler.go
+++ b/cmd/query/app/static_handler.go
@@ -30,15 +30,17 @@ import (
 	"go.uber.org/zap"
 
 	"github.com/jaegertracing/jaeger/cmd/query/app/ui"
+	"github.com/jaegertracing/jaeger/pkg/version"
 )
 
 var (
 	favoriteIcon    = "favicon.ico"
 	staticRootFiles = []string{favoriteIcon}
+
+	// The following patterns are searched and replaced in the index.html as a way of customizing the UI.
 	configPattern   = regexp.MustCompile("JAEGER_CONFIG *= *DEFAULT_CONFIG;")
-	basePathPattern = regexp.MustCompile(`<base href="/"`)
-	basePathReplace = `<base href="%s/"`
-	errBadBasePath  = "Invalid base path '%s'. Must start but not end with a slash '/', e.g. '/jaeger/ui'"
+	versionPattern  = regexp.MustCompile("JAEGER_VERSION *= *DEFAULT_VERSION;")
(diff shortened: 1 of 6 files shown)


pkg/version/build.go
```diff
@@ -28,8 +28,8 @@ var (
 // Info holds build information
 type Info struct {
 	GitCommit  string `json:"gitCommit"`
-	GitVersion string `json:"GitVersion"`
-	BuildDate  string `json:"BuildDate"`
+	GitVersion string `json:"gitVersion"`
+	BuildDate  string `json:"buildDate"`
 }
```
yurishkuro, line 33: This could be considered a minor breaking change, I don't know why there was a mismatch in the name casing.

yurishkuro: This JSON output is used in `version` command and `/version` endpoint:
```
$ curl http://localhost:14269/version
{"gitCommit":"","gitVersion":"","buildDate":""}
```
Hard to imaging there would be dependencies on the exact string output of these.

cmd/query/app/static_handler.go
```diff
@@ -113,14 +116,19 @@ func loadIndexBytes(open func(string) (http.File, error), options StaticAssetsHa
...
+	versionString := fmt.Sprintf("JAEGER_VERSION = %s;", string(versionJSON))
+	indexBytes = versionPattern.ReplaceAll(indexBytes, []byte(versionString))
+	// replace base path
 	if options.BasePath == "" {
 		options.BasePath = "/"
 	}
 	if options.BasePath != "/" {
 		if !strings.HasPrefix(options.BasePath, "/") || strings.HasSuffix(options.BasePath, "/") {
-			return nil, fmt.Errorf(errBadBasePath, options.BasePath)
+			return nil, fmt.Errorf("invalid base path '%s'. Must start but not end with a slash '/', e.g. '/jaeger/ui'", options.BasePath)
```
jpkrohling, line 129: How about replacing `'%s'` with `%q` ?

pkg/version/build.go
jpkrohling: I really wouldn't worry about it...

jpkrohling (approved):

joe-elliott (approved):
Looks good to me.

cmd/query/app/static_handler.go
yurishkuro: I think it only makes the logs uglier because JSON logger will have to escape quotes, etc.

PR #2548, opened 2020-10-09
Update static UI assets path in contrib doc

albertteoh:

Following the instruction from the CONTRIBUTING.md file results in a "no such file or directory" error when trying to fetch static UI assets:

```
$ go run ./cmd/all-in-one/main.go --query.static-files jaeger-ui/packages

"Could not create static assets handler","error":"cannot load index.html: cannot open index.html: open jaeger-ui/build/index.html: no such file or directory"
```

The `Makefile` appears to copy the static UI assets into `jaeger-ui/packages/jaeger-ui/build` instead. See: https://github.com/jaegertracing/jaeger/blob/master/Makefile#L234

Changing the command to the correct path as specified in the `build-ui` `Makefile` target results in the all-in-one running successfully:
```
$ go run ./cmd/all-in-one/main.go --query.static-files jaeger-ui/packages/jaeger-ui/build 
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -45,7 +45,7 @@ $ go run -tags ui ./cmd/all-in-one/main.go
 Alternatively, the path to the built UI assets can be provided via `--query.static-files` flag:
 
 ```
-$ go run ./cmd/all-in-one/main.go --query.static-files jaeger-ui/build
+$ go run ./cmd/all-in-one/main.go --query.static-files jaeger-ui/packages/jaeger-ui/build
 ```
 
 ## Project Structure
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,22 +30,14 @@ Then install dependencies and run the tests:
 git submodule update --init --recursive
 make install-tools
 make test
-#if you wish to build platform binaries locally - the step below is needed.
+# if you wish to build platform binaries locally - the step below is needed.
 make build-ui
 ```
 
 ### Running local build with the UI
 
-The `jaeger-ui` submodule contains the source code for the UI assets (requires Node.js 6+). The assets must be compiled first with `make build-ui`, which runs Node.js build and then packages the assets into a Go file that is `.gitignore`-ed. The packaged assets can be enabled by providing a build tag `ui`, e.g.:
-
 ```
-$ go run -tags ui ./cmd/all-in-one/main.go
-```
-
-Alternatively, the path to the built UI assets can be provided via `--query.static-files` flag:
-
-```
-$ go run ./cmd/all-in-one/main.go --query.static-files jaeger-ui/packages/jaeger-ui/build
+$ make run-all-in-one
 ```
 
 ## Project Structure


CONTRIBUTING.md
```diff
@@ -45,7 +45,7 @@ $ go run -tags ui ./cmd/all-in-one/main.go
 Alternatively, the path to the built UI assets can be provided via `--query.static-files` flag:
 
 ```
-$ go run ./cmd/all-in-one/main.go --query.static-files jaeger-ui/build
+$ go run ./cmd/all-in-one/main.go --query.static-files jaeger-ui/packages/jaeger-ui/build
```
jpkrohling, line 48: We actually have a make target now that can be used instead: `make run-all-in-one`

albertteoh: Ah okay, glad you pointed this out. Do you think we should delete this section of the contrib file then (i.e. "Alternatively, ..." + code block)?

Moreover, given we already have the `make build-ui` instruction in the previous section, could we just have a one-liner for this "Running local build with the UI" section of:
```
make run-all-in-one
```

Unless if we feel the first two sentences (i.e. "The jaeger-ui... .gitignored-ed.") are useful information.

jpkrohling: I think so, the whole section here could be replaced with this one make target.

jpkrohling (approved):

CONTRIBUTING.md
```diff
@@ -30,22 +30,14 @@ Then install dependencies and run the tests:
...
 make install-tools
 make test
-#if you wish to build platform binaries locally - the step below is needed.
+# if you wish to build platform binaries locally - the step below is needed.
 make build-ui
 ```
 
 ### Running local build with the UI
 
-The `jaeger-ui` submodule contains the source code for the UI assets (requires Node.js 6+). The assets must be compiled first with `make build-ui`, which runs Node.js build and then packages the assets into a Go file that is `.gitignore`-ed. The packaged assets can be enabled by providing a build tag `ui`, e.g.:
```
yurishkuro, line 39: The make target is great, but why remove the explanation? It's valuable information that now someone needs to dig out by reading source code.

albertteoh: This is a valid point, @yurishkuro; how the UI assets are created and used deserves a more detailed explanation as it's not straightforward (at least to me) to understand from just looking at code and Makefiles.

I think though, that this instruction should belong in the Pre-requisites section as an explanation to what `make build-ui` does. So my proposal is (what do you think?):

### Pre-requisites
Then install dependencies and run the tests:

```
git submodule update --init --recursive
make install-tools
make test
```

The `jaeger-ui` submodule contains the source code for the UI assets (requires Node.js 6+). The assets must be compiled first with `make build-ui`, which runs Node.js build and then packages the assets into a Go file that is `.gitignore`-ed:

```
# if you wish to build platform binaries locally - the step below is needed.
make build-ui
```

### Running local build with the UI

The packaged assets (created by the `make build-ui` step above) can be enabled by providing a build tag `ui`. The following command will start Jaeger all-in-one with this `ui` tag.
```
make run-all-in-one
```

yurishkuro: I prefer UI-related explanations to be together under one section.

PR #2554, opened 2020-10-12
Disable Zipkin server if port/address is not configured

yurishkuro:
In the past Zipkin server would not be started if the port was not configured, but not it's always started.

* run OTEL tests as part of `test-ci` (without coverage)
* change CollectorZipkinHTTPHostPort default to ""
* add logging when Zipkin server is not started
* changed `ports.GetAddressFromCLIOptions()` to return empty string when `port=0 && hostPort=""` (previously it was returning `":"`, which didn't make sense, it's better to return default value for string when no parts of the address are specified)

With this change:

```
$ go run ./cmd/collector help | grep -i zipkin
      --collector.zipkin.host-port string               The host:port (e.g. 127.0.0.1:9411 or :9411) of the collector's Zipkin server (disabled by default)

$ SPAN_STORAGE_TYPE=memory go run ./cmd/collector 2>&1 | grep -i zipkin
{"level":"info","ts":1602520683.32179,"caller":"server/zipkin.go:45","msg":"Not listening for Zipkin HTTP traffic, port not configured"}

$ SPAN_STORAGE_TYPE=memory go run ./cmd/collector --collector.zipkin.host-port 9411 2>&1 | grep -i zipkin
{"level":"info","ts":1602520719.572457,"caller":"server/zipkin.go:49","msg":"Listening for Zipkin HTTP traffic","zipkin host-port":":9411"}
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -391,7 +391,9 @@ install-tools:
 install-ci: install-tools
 
 .PHONY: test-ci
-test-ci: build-examples lint cover
+# TODO (ys) added test-otel to at least ensure tests run in CI,
+#      but this needs to be changed in the lint and cover targets instead
+test-ci: build-examples lint cover test-otel
 
 .PHONY: thrift
 thrift: idl/thrift/jaeger.thrift thrift-image
--- a/cmd/collector/app/builder_flags.go
+++ b/cmd/collector/app/builder_flags.go
@@ -101,7 +101,7 @@ func AddOTELJaegerFlags(flags *flag.FlagSet) {
 
 // AddOTELZipkinFlags adds flag that are exposed by OTEL Zipkin receiver
 func AddOTELZipkinFlags(flags *flag.FlagSet) {
-	flags.String(CollectorZipkinHTTPHostPort, ports.PortToHostPort(0), "The host:port (e.g. 127.0.0.1:9411 or :9411) of the collector's Zipkin server")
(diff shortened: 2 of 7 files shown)


joe-elliott (commented):
This will break at least two things in the otel collector/agent.  I can review after this is merged and provide a PR if that's easiest.

This test will need to be corrected to the new value:
https://github.com/jaegertracing/jaeger/blob/master/cmd/opentelemetry/app/receiver/zipkinreceiver/zipkin_receiver_test.go#L42

This check will need to updated to the new default:
https://github.com/jaegertracing/jaeger/blob/master/cmd/opentelemetry/app/defaultconfig/default_config.go#L167

yurishkuro:
Hold on, how is it that you're saying tests will need to be fixed but they are not failing in the CI? Are we not running them at all in travis?

yurishkuro:
Added a fix to invoke OTEL tests from `test-ci` (larger issue in #2555)

yurishkuro:
Confirmed that OTEL tests are now running and failing with this change (https://travis-ci.org/github/jaegertracing/jaeger/jobs/735111015), so fixed it by changing how GetAddressFromCLIOptions works.

joe-elliott (approved):
nit:  The below line could be simplified to:

```
if zipkin.Endpoint != "" {
```

https://github.com/jaegertracing/jaeger/blob/master/cmd/opentelemetry/app/defaultconfig/default_config.go#L167

Other than that, looks good to me.

PR #2558, opened 2020-10-13
Replaced devbots with issue-labeler

jpkrohling:
The bot that used to add the 'needs-triage' label to new issues stopped working and I can't figure out why. Looks like GitHub has a more official action that does something similar, so, this PR is replacing the 'devbots' with the issue-labeler.

See https://github.com/jaegertracing/jaeger-operator/pull/1251

### DIFF
--- a/.devbots/needs-triage.yml
+++ b/.devbots/needs-triage.yml
@@ -1,2 +1,0 @@
-enabled: true
-label: "needs-triage"
--- a/.github/labeler-needs-triage.yaml
+++ b/.github/labeler-needs-triage.yaml
@@ -1,0 +1,3 @@
+# Add 'needs-triage' label to all new issues
+needs-triage:
+    - '.*'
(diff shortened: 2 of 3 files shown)


objectiser (approved):

yurishkuro:
fwiw I don't find `needs-triage` label particularly useful, we don't seem to pay much attention to it

jpkrohling:
I use it quite a lot to know which issues were reported and that nobody looked at. For the operator, I review them once a day and remove this label, adding other ones, but I noticed that some issues in the main repository are closed while still with `needs-triage`. I need to do a better job in this repository here :-)

yurishkuro:
I used to remove this label when responding, but found it to be too much hustle. I think the semantics of this label is not very well defined either. If we're going after the typical customer support model, then really the labels need to keep changing from "waiting for support" to "waiting for OP feedback". If needs-triage is simply "has anyone looked at it", then perhaps the bot should remove the label automatically once any of the maintainers reply to the ticket.

jpkrohling:
> then perhaps the bot should remove the label automatically once any of the maintainers reply to the ticket.

I think it's actually doable.

PR #2560, opened 2020-10-13
Simplify test coverage collection

### LINKED ISSUE #797: Simplify test coverage scripts

PankhudiB:
Signed-off-by: Pankhudi Bhonsle <psbhonsle08@gmail.com>

- Simplifies test coverage make task
- Resolves #797

- Replace the old script with go test ./... command

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -150,9 +150,6 @@ all-srcs:
 
 .PHONY: cover
 cover: nocover
-	@echo pre-compiling tests
-	@time go test -i $(shell go list ./...)
-	# TODO Switch to single `go test` that already supports multiple packages, but watch out for .nocover dirs.
 	$(GOTEST) -coverprofile cover.out ./...
 	grep -E -v 'model.pb.*.go' cover.out > cover-nogen.out
 	mv cover-nogen.out cover.out


Makefile
```diff
@@ -153,7 +153,7 @@ cover: nocover
 	@echo pre-compiling tests
 	@time go test -i $(shell go list ./...)
 	# TODO Switch to single `go test` that already supports multiple packages, but watch out for .nocover dirs.
-	@./scripts/cover.sh $(shell go list ./...)
+	go test ./...
```
yurishkuro, line 156: this will not produce code coverage results. Try this:

```
$(GOTEST) -coverprofile cover.out ./...
```

yurishkuro:
Resolves #797

Makefile
```diff
@@ -153,7 +153,7 @@ cover: nocover
 	@echo pre-compiling tests
 	@time go test -i $(shell go list ./...)
 	# TODO Switch to single `go test` that already supports multiple packages, but watch out for .nocover dirs.
```
yurishkuro, line 155: please remove the above 3 lines. Pre-compilation no longer makes sense since we invoke a single `go test` command anyway, and TODO is done by this PR.

yurishkuro (commented):
don't know why it happened, but nice to see coverage go up by 0.02% :-)

Makefile
PankhudiB: I think we also no longer need nocover

yurishkuro: we do need it, it's protection against adding packages that don't have test files, which are not included in coverage by default.

yurishkuro (approved):
Thank you!

PankhudiB:
> Thank you!

This was my first ever contribution and i did not feel overwhelmed.  All thanks to you for your help. :)

yurishkuro:
you are welcome, and congrats on the first PR, may it be the first of many.

PR #2571, opened 2020-10-15
Use fossa-contrib/fossa-action instead

smorimoto:
- None

- By using dedicated Action, you can keep your workflow simple and gain some benefits.


.github/workflows/fossa.yml
```diff
@@ -8,20 +8,25 @@ on:
...
+      - name: Use Go ^1.14.x
+        uses: actions/setup-go@v2
+        with:
+          go-version: ^1.14.x
+
+      - name: Add GOPATH to GITHUB_ENV
+        run: echo "GOPATH=$(go env GOPATH)" >>"$GITHUB_ENV"
+
+      - name: Add GOPATH to GITHUB_PATH
+        run: echo "$GOPATH/bin" >>"$GITHUB_PATH"
```
yurishkuro, line 25: ? 

Do these GITHUB_ENV/GITHUB_PATH vars contain file names?

.github/workflows/fossa.yml
```diff
@@ -8,20 +8,25 @@ on:
...
+
+      - name: Run FOSSA scan and upload build data
+        uses: fossa-contrib/fossa-action@v1
```
yurishkuro, line 28: Why isn't this an official action under github.com/fossas ?

.github/workflows/fossa.yml
smorimoto: Is this probably what you are looking for?
https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#environment-files

.github/workflows/fossa.yml
smorimoto: Because this isn't an official action yet.

jpkrohling:
@mergifyio rebase

.github/workflows/fossa.yml
jpkrohling: It's also not clear to me why this is needed. Perhaps leave a comment in the code?

jpkrohling (commented):
The check seems to work, but just need a clarification on the "Add GOPATH" steps.

jpkrohling:
@smorimoto, are you still interested in this one?

smorimoto:
Oh, I'm sorry I missed it.

.github/workflows/fossa.yml
smorimoto: Well, all it does is expose the GOPATH environment variable and add GOPATH to the PATH. Of course, we can do it more typically. Or are you asking why these are necessary? If so, it's necessary because fossa-cli needs them when it analyzes.

yurishkuro:
Thanks!

PR #2573, opened 2020-10-17
Improve `build-ui` Makefile target and fix dependencies

### LINKED ISSUE #2572: Simplify contributing.md
This proposal is motivated by a PR comment originally posted by @yurishkuro in https://github.com/jaegertracing/jaeger/pull/2548#discussion_r502507111.

- Bring back the explanation of what `make build-ui` does under the hood, which contains valuable information for contributors.
- Reduce the number of manual steps required to get started in [contributing.md](https://github.com/jaegertracing/jaeger/blob/master/CONTRIBUTING.md) instructions to at least:
```
make install-tools
make test
```
- More correct make targets:
  - Only run `build-ui` when UI assets are required.
  - Clean up redundant make target dependencies and only assign the `build-ui` dependency to the most specific targets.
  - Add missing `build-ui` dependencies to make targets that depend on UI assets, e.g. `build-all-in-one` should have a dependency on UI assets, but why only `build-all-in-one-linux` takes the `build-ui` dependency?

[contributing.md](https://github.com/jaegertracing/jaeger/blob/master/CONTRIBUTING.md) currently requires these prerequisite steps:

```
git submodule update --init --recursive
make install-tools
make test
# if you wish to build platform binaries locally - the step below is needed.
make build-ui
```

`git submodule update --init --recursive` and `make build-ui` are specific to UI builds. Instead of making these prerequisites, could we make these an explicit dependency within our `Makefile` and only run them when performing UI-related builds?

This has the added benefit of automating these steps where required and, IMHO, is more correct. Consider the following scenarios:
- Only run/build non-UI components; for example: testing agent->collector->kafka and validating with a consumer.
- The jaeger-ui submodule commit is modified in upstream during development and we rebase to pick these changes up.

- Reinstate what `make build-ui` does under the hood (which was removed in #2548) within the Makefile as part of the `build-ui` target, similar to the [`proto` make target](https://github.com/jaegertracing/jaeger/blob/master/Makefile#L454).

- Add `build-ui` dependency to the "lowest common denominator" make targets, namely `build-query`, `build-all-in-one` and `run-all-in-one`.

- `yarn install` and `yarn build` steps take well over a minute to complete even after prior calls, which is particularly undesirable for something like `make run-all-in-one`.
  - Make `build-ui` idempotent by checking for existence of the asset files (`cmd/query/app/ui/*/gen_assets.go`). If there's a need to force UI asset rebuilds, either delete asset files via the `clean` target (this sounds nicer?) or introduce a `build-ui-force` option target.

- As above, move `build-ui` dependency from `build-all-in-one-linux` to the more general `build-all-in-one` target to fix other OS builds.

- Add `git submodule update --init --recursive` as part of the `build-ui` target given their fates are tied to one another (unless if I'm mistaken). This call is idempotent and so is relatively quick to complete after making the initial call. Added benefit of ensuring the correct commit is used in each build/run if it was modified in the upstream master.

- Remove redundant `build-ui` from `docker` target given it is introduced as a dependency in `build-query` (`build-binaries-linux` -> `build-platform-binaries` -> `build-query` -> `build-ui`).

- Simplify the [contributing.md](https://github.com/jaegertracing/jaeger/blob/master/CONTRIBUTING.md) Prerequisites section down to:
```
make install-tools
make test
```

- Maybe there are good reasons to having `git submodule update...` and `make build-ui` as explicit prerequisite steps that I have overlooked?
- What is the reason for requiring `make test` as a prerequisite?

albertteoh:
- Resolves #2572

- Remove `build-ui` step from CONTRIBUTING.md as it's automated by Makefile.
- Add UI asset files in `clean` target to enable rebuilding of UI assets.
- Reinstate the explanation of what `build-ui` does under the hood into the Makefile under its namesake, which was deleted in #2548.
- `build-ui` target dependency should _only_ be used for builds that use `-tags ui` (remove from `docker` and `build-all-in-one-linux` targets).
- Make `build-ui` target idempotent by explicitly checking for the UI assets files' existence. If all expected files exist, do nothing. Otherwise, regenerate UI asset files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -229,13 +229,6 @@ run-all-in-one: build-ui
 .PHONY: build-ui
 build-ui: cmd/query/app/ui/actual/gen_assets.go cmd/query/app/ui/placeholder/gen_assets.go
 	# Do nothing. If you need to force a rebuild of UI assets, run `make clean`.
-	#
-	# The `jaeger-ui` submodule contains the source code for the UI assets (requires Node.js 6+).
-	# The assets must be compiled first with `make build-ui`, which runs Node.js build and then
-	# packages the assets into a Go file that is `.gitignore`-ed.
-	#
-	# The packaged assets can be enabled by providing a build tag `ui`; for example:
-	# $ go run -tags ui ./cmd/all-in-one/main.go
 
 jaeger-ui/packages/jaeger-ui/build/index.html:
 	cd jaeger-ui && yarn install --frozen-lockfile && cd packages/jaeger-ui && yarn build
--- a/Makefile
+++ b/Makefile
@@ -88,8 +88,7 @@ md-to-godoc-gen:
 .PHONY: clean
 clean:
 	rm -rf cover.out .cover/ cover.html lint.log fmt.log \
-		cmd/query/app/ui/actual/gen_assets.go \
-		cmd/query/app/ui/placeholder/gen_assets.go
+		cmd/query/app/ui/actual/gen_assets.go
 
 .PHONY: test
 test: go-gen test-otel
@@ -233,7 +232,7 @@ EXPECT_ASSETS := cmd/query/app/ui/actual/gen_assets.go \
                  cmd/query/app/ui/placeholder/gen_assets.go
 FOUND_ASSETS := $(wildcard $(EXPECT_ASSETS))
 .PHONY: build-ui
-build-ui:
+build-ui: cmd/query/app/ui/actual/gen_assets.go cmd/query/app/ui/placeholder/gen_assets.go
 	# The `jaeger-ui` submodule contains the source code for the UI assets (requires Node.js 6+).
 	# The assets must be compiled first with `make build-ui`, which runs Node.js build and then
 	# packages the assets into a Go file that is `.gitignore`-ed.
@@ -246,16 +245,15 @@ build-ui:
 	#
 	# To force a rebuild of UI assets, run `make clean` which deletes any files matching the
 	# above glob pattern.
-	@echo "Expect UI assets: $(EXPECT_ASSETS)"
-	@echo "Found UI assets: $(FOUND_ASSETS)"
 
-ifeq ($(FOUND_ASSETS), $(EXPECT_ASSETS))
-	@echo "Skipping building UI assets as all expected files were found."
-else
+jaeger-ui/packages/jaeger-ui/build/index.html:
 	cd jaeger-ui && yarn install --frozen-lockfile && cd packages/jaeger-ui && yarn build
+
+cmd/query/app/ui/actual/gen_assets.go: jaeger-ui/packages/jaeger-ui/build/index.html
 	esc -pkg assets -o cmd/query/app/ui/actual/gen_assets.go -prefix jaeger-ui/packages/jaeger-ui/build jaeger-ui/packages/jaeger-ui/build
+
+cmd/query/app/ui/placeholder/gen_assets.go: cmd/query/app/ui/placeholder/public/index.html
 	esc -pkg assets -o cmd/query/app/ui/placeholder/gen_assets.go -prefix cmd/query/app/ui/placeholder/public cmd/query/app/ui/placeholder/public
-endif
 
 .PHONY: build-all-in-one-linux
 build-all-in-one-linux:
--- a/Makefile
+++ b/Makefile
@@ -88,7 +88,7 @@ md-to-godoc-gen:
 .PHONY: clean
 clean:
 	rm -rf cover.out .cover/ cover.html lint.log fmt.log \
-		cmd/query/app/ui/actual/gen_assets.go
+		jaeger-ui/packages/jaeger-ui/build
 
 .PHONY: test
 test: go-gen test-otel
@@ -226,25 +226,16 @@ docker-hotrod:
 run-all-in-one: build-ui
 	go run -tags ui ./cmd/all-in-one --log-level debug
 
-# Be explicit about exactly what assets we want, as asset files could be partially deleted
-# after initial generation; which is the case in Travis CI builds (DOCKER + DEPLOY).
-EXPECT_ASSETS := cmd/query/app/ui/actual/gen_assets.go \
-                 cmd/query/app/ui/placeholder/gen_assets.go
-FOUND_ASSETS := $(wildcard $(EXPECT_ASSETS))
 .PHONY: build-ui
 build-ui: cmd/query/app/ui/actual/gen_assets.go cmd/query/app/ui/placeholder/gen_assets.go
+	# Do nothing. If you need to force a rebuild of UI assets, run `make clean`.
+	#
 	# The `jaeger-ui` submodule contains the source code for the UI assets (requires Node.js 6+).
 	# The assets must be compiled first with `make build-ui`, which runs Node.js build and then
 	# packages the assets into a Go file that is `.gitignore`-ed.
 	#
 	# The packaged assets can be enabled by providing a build tag `ui`; for example:
 	# $ go run -tags ui ./cmd/all-in-one/main.go
-	#
-	# To reduce `build-ui` execution times, no work will be done if the resulting UI assets exist.
-	# These expected assets are cmd/query/app/ui/(actual|placeholder)/gen_assets.go.
-	#
-	# To force a rebuild of UI assets, run `make clean` which deletes any files matching the
-	# above glob pattern.
 
 jaeger-ui/packages/jaeger-ui/build/index.html:
 	cd jaeger-ui && yarn install --frozen-lockfile && cd packages/jaeger-ui && yarn build
--- a/Makefile
+++ b/Makefile
@@ -369,6 +369,8 @@ include crossdock/rules.mk
 # Crossdock tests do not require fully functioning UI, so we skip it to speed up the build.
 .PHONY: build-crossdock-ui-placeholder
 build-crossdock-ui-placeholder:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -222,29 +224,53 @@ docker-hotrod:
...
 	go run -tags ui ./cmd/all-in-one --log-level debug
 
+# Be explicit about exactly what assets we want, as asset files could be partially deleted
+# after initial generation; which is the case in Travis CI builds (DOCKER + DEPLOY).
+EXPECT_ASSETS := cmd/query/app/ui/actual/gen_assets.go \
+                 cmd/query/app/ui/placeholder/gen_assets.go
+FOUND_ASSETS := $(wildcard $(EXPECT_ASSETS))
 .PHONY: build-ui
 build-ui:
+	# The `jaeger-ui` submodule contains the source code for the UI assets (requires Node.js 6+).
```
albertteoh, line 237: Since these comments are indented, they'll be printed. Please let me know if this isn't desirable, and I can unindent to silence them.

Makefile
```diff
@@ -222,29 +224,53 @@ docker-hotrod:
...
+	#
+	# To reduce `build-ui` execution times, no work will be done if the resulting UI assets exist.
+	# These expected assets are cmd/query/app/ui/(actual|placeholder)/gen_assets.go.
+	#
+	# To force a rebuild of UI assets, run `make clean` which deletes any files matching the
+	# above glob pattern.
+	@echo "Expect UI assets: $(EXPECT_ASSETS)"
+	@echo "Found UI assets: $(FOUND_ASSETS)"
+
+ifeq ($(FOUND_ASSETS), $(EXPECT_ASSETS))
```
albertteoh, line 252: I'm open to suggestions of a better approach to checking if `build-ui` was previously run.

Makefile
```diff
@@ -87,7 +87,9 @@ md-to-godoc-gen:
 
 .PHONY: clean
 clean:
-	rm -rf cover.out .cover/ cover.html lint.log fmt.log
+	rm -rf cover.out .cover/ cover.html lint.log fmt.log \
+		cmd/query/app/ui/actual/gen_assets.go \
+		cmd/query/app/ui/placeholder/gen_assets.go
```
yurishkuro, line 92: this file is checked in,  I don't think `clean` should be removing it.

Makefile
yurishkuro: This is what Make was built for - we can to define the files as dependencies of a target. However, the trick is that the files themselves should also be dependency-tracked. I.e. something like this:
```
placeholder/gen_assets.go: placeholder/public/index.html
    esc ...

actual/gen_assets.go: jaeger-ui/packages/jaeger-ui/build/index.html
    esc ...

build-ui: placeholder/gen_assets.go actual/gen_assets.go
    # actually, nothing to do here
```

But, there are still two problems with this:
  * ideally, `jaeger-ui/packages/jaeger-ui/build/index.html` should also be _dependent_ on the actual UI source. This gets pretty complicated to manage.
  * this set-up _always_ requires a real UI build, a situation I would rather avoid. It should only be required when the binary is compiled with `-tags ui`.

albertteoh: Ah yes, good suggestion re: make on file dependencies.

> ideally, jaeger-ui/packages/jaeger-ui/build/index.html should also be dependent on the actual UI source. This gets pretty complicated to manage.

Could we solve this with the following make target spec? I suspect you are suggesting that there's more to consider that I haven't picked up on:
```
jaeger-ui/packages/jaeger-ui/build/index.html:
	cd jaeger-ui && yarn install --frozen-lockfile && cd packages/jaeger-ui && yarn build
```

> this set-up always requires a real UI build, a situation I would rather avoid.

Forgive me for not understanding and my ignorance; what specifically about this set-up requires a real UI build? I thought `yarn build` would result in a real UI build?

> It should only be required when the binary is compiled with -tags ui.

In this PR, I made sure the `build-ui` target is only executed for binaries compiled `-tags ui`, or have I misunderstood something?

yurishkuro: > Could we solve this with the following make target spec?

The main idea of Make is to execute targets when dependencies change. When the target itself is a file AND has no dependencies, then once that file exists the command will never be run:
```
$ cat Makefile
x: file

file:
       echo n > file

$ make x
echo n > file

$ make x
make: Nothing to be done for `x'.
```

albertteoh: We could add the `build` dir to the `clean` target (which doesn't exist when initing the submodule, also note `actual/gen_assets.go` isn't checked in so it should be okay to remove). So we'd have the following:
```
.PHONY: clean
clean:
	rm -rf cover.out .cover/ cover.html lint.log fmt.log \
		cmd/query/app/ui/actual/gen_assets.go \
		jaeger-ui/packages/jaeger-ui/build

.PHONY: build-ui
build-ui: cmd/query/app/ui/actual/gen_assets.go cmd/query/app/ui/placeholder/gen_assets.go
        # Do nothing

jaeger-ui/packages/jaeger-ui/build/index.html:
	cd jaeger-ui && yarn install --frozen-lockfile && cd packages/jaeger-ui && yarn build

cmd/query/app/ui/actual/gen_assets.go: jaeger-ui/packages/jaeger-ui/build/index.html
	esc -pkg assets -o cmd/query/app/ui/actual/gen_assets.go -prefix jaeger-ui/packages/jaeger-ui/build jaeger-ui/packages/jaeger-ui/build

cmd/query/app/ui/placeholder/gen_assets.go: cmd/query/app/ui/placeholder/public/index.html
	esc -pkg assets -o cmd/query/app/ui/placeholder/gen_assets.go -prefix cmd/query/app/ui/placeholder/public cmd/query/app/ui/placeholder/public
```

From a fresh clone, I performed the following test:
```
$ git submodule update --init --recursive

$ make build-ui
cd jaeger-ui && yarn install --frozen-lockfile && cd packages/jaeger-ui && yarn build
esc -pkg assets -o cmd/query/app/ui/actual/gen_assets.go -prefix jaeger-ui/packages/jaeger-ui/build jaeger-ui/packages/jaeger-ui/build
esc -pkg assets -o cmd/query/app/ui/placeholder/gen_assets.go -prefix cmd/query/app/ui/placeholder/public cmd/query/app/ui/placeholder/public

$ make build-ui
# Do nothing

# Trigger a rebuild of UI assets
$ make clean
rm -rf cover.out .cover/ cover.html lint.log fmt.log \
		cmd/query/app/ui/actual/gen_assets.go \
		jaeger-ui/packages/jaeger-ui/build

$ make build-ui
cd jaeger-ui && yarn install --frozen-lockfile && cd packages/jaeger-ui && yarn build
esc -pkg assets -o cmd/query/app/ui/actual/gen_assets.go -prefix jaeger-ui/packages/jaeger-ui/build jaeger-ui/packages/jaeger-ui/build
```

yurishkuro: Sounds good. I would change `# Do nothing` to `# Do nothing. If you need to force a rebuild of UI assets, run make clean.`

I don't think you need to include `cmd/query/app/ui/actual/gen_assets.go` in the `clean` target, because that file will be rebuilt any time index.html is changed (which is being deleted in clean).

Makefile
```diff
@@ -222,29 +223,43 @@ docker-hotrod:
...
-build-ui:
+build-ui: cmd/query/app/ui/actual/gen_assets.go cmd/query/app/ui/placeholder/gen_assets.go
+	# Do nothing. If you need to force a rebuild of UI assets, run `make clean`.
+	#
+	# The `jaeger-ui` submodule contains the source code for the UI assets (requires Node.js 6+).
+	# The assets must be compiled first with `make build-ui`, which runs Node.js build and then
+	# packages the assets into a Go file that is `.gitignore`-ed.
+	#
+	# The packaged assets can be enabled by providing a build tag `ui`; for example:
+	# $ go run -tags ui ./cmd/all-in-one/main.go
```
yurishkuro, line 238: I think CONTRIBUTING file is a better place for this documentation than a makefile, nobody would look here, especially when reading instructions in CONTRIBUTING

yurishkuro (commented):
I compared CI run for this PR (https://travis-ci.org/github/jaegertracing/jaeger/builds/736964904) with recent run on master (https://travis-ci.org/github/jaegertracing/jaeger/builds/735758902). The CROSDOCK steps were a few minutes longer, probably because they are now running yarn to build full UI, a step that we intentionally tried to avoid  (see target `build-crossdock-ui-placeholder`). I think that target needs to be changed to also copy index.html from placeholder dir into jaeger-ui/build

yurishkuro:
btw, please give a different title to this PR, I think changes to makefile are more significant than adding documentation

albertteoh:
> btw, please give a different title to this PR, I think changes to makefile are more significant than adding documentation

Done, please let me know if the new title is appropriate or if you have suggestions to improve it.

Makefile
```diff
@@ -222,29 +223,43 @@ docker-hotrod:
...
+
+jaeger-ui/packages/jaeger-ui/build/index.html:
 	cd jaeger-ui && yarn install --frozen-lockfile && cd packages/jaeger-ui && yarn build
+
+cmd/query/app/ui/actual/gen_assets.go: jaeger-ui/packages/jaeger-ui/build/index.html
```
yurishkuro, line 243: please add to the top of `build-crossdock-ui-placeholder` target:
```
mkdir -p jaeger-ui/packages/jaeger-ui/build/
cp cmd/query/app/ui/placeholder/public/index.html jaeger-ui/packages/jaeger-ui/build/index.html
```
to avoid running yarn for crossdock CI steps

yurishkuro (approved):
Thanks, Albert!

PR #2590, opened 2020-10-27
Clean-up error handling and logging

tzzed:
- Add more idiomatic go in `cmd/collector/main.go` 

- Delete nested code, Add error blank identifiers, and use `Fprint*` to display err.

### DIFF
--- a/cmd/collector/app/collector.go
+++ b/cmd/collector/app/collector.go
@@ -16,6 +16,7 @@ package app
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"net/http"
 	"time"
@@ -91,7 +92,7 @@ func (c *Collector) Start(builderOpts *CollectorOptions) error {
 		SamplingStore: c.strategyStore,
 		Logger:        c.logger,
 	}); err != nil {
-		c.logger.Fatal("could not start gRPC collector", zap.Error(err))
+		return fmt.Errorf("could not start gRPC collector %w", err)
 	} else {
 		c.grpcServer = grpcServer
 	}
@@ -104,7 +105,7 @@ func (c *Collector) Start(builderOpts *CollectorOptions) error {
 		SamplingStore:  c.strategyStore,
 		Logger:         c.logger,
 	}); err != nil {
-		c.logger.Fatal("could not start the HTTP server", zap.Error(err))
+		return fmt.Errorf("could not start the HTTP server %w", err)
 	} else {
 		c.hServer = httpServer
 	}
@@ -118,7 +119,7 @@ func (c *Collector) Start(builderOpts *CollectorOptions) error {
 		AllowedOrigins: builderOpts.CollectorZipkinAllowedOrigins,
 		Logger:         c.logger,
 	}); err != nil {
-		c.logger.Fatal("could not start the Zipkin server", zap.Error(err))
+		return fmt.Errorf("could not start the Zipkin server %w", err)
 	} else {
 		c.zkServer = zkServer
 	}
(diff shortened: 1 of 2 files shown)


cmd/collector/main.go
```diff
@@ -96,15 +96,14 @@ func main() {
 				HealthCheck:    svc.HC(),
 			})
 			collectorOpts := new(app.CollectorOptions).InitFromViper(v)
-			c.Start(collectorOpts)
+			_ = c.Start(collectorOpts)
```
yurishkuro, line 99: if you'd like to clean this up even more, c.Start currently never returns an error, because it logs Fatals. Instead, it should be returning wrapped errors and log.Fatal should only happen here.

cmd/collector/main.go
```diff
@@ -131,7 +130,7 @@ func main() {
 	)
 
 	if err := command.Execute(); err != nil {
-		fmt.Println(err.Error())
+		_, _ = fmt.Fprintln(os.Stderr, err)
```
yurishkuro, line 133: A better alternative is to use stdlib's log.Fatal

cmd/collector/app/collector.go
```diff
@@ -91,7 +102,7 @@ func (c *Collector) Start(builderOpts *CollectorOptions) error {
 		SamplingStore: c.strategyStore,
 		Logger:        c.logger,
 	}); err != nil {
-		c.logger.Fatal("could not start gRPC collector", zap.Error(err))
+		return newCollectorError("could not start gRPC collector", err)
```
yurishkuro, line 105: this seems an overkill, I would simply do `return fmt.Errorf("could not start gRPC collector: %w", err)`, which would properly wrap the error anyway.

cmd/collector/main.go
yurishkuro: I'll unresolve, it will auto-resolve by submitting a change.

cmd/collector/app/collector.go
tzzed: I didn't dare to do it. I think it's much better too.

yurishkuro: There is no need to worry about logger fields here, we just need to wrap the error with explaining messages using Errorf. This is how it's done everywhere else in this project when returning nested errors.

tzzed: More clear thank you !

yurishkuro:
please don't forget to have all commits signed (`git commit -s -amend`)

cmd/collector/app/collector.go
```diff
@@ -143,29 +144,27 @@ func (c *Collector) Close() error {
 	// HTTP server
 	if c.hServer != nil {
 		timeout, cancel := context.WithTimeout(context.Background(), 5*time.Second)
-		err := c.hServer.Shutdown(timeout)
-		if err != nil {
-			c.logger.Error("failed to stop the main HTTP server", zap.Error(err))
+		if err := c.hServer.Shutdown(timeout); err != nil {
+			return fmt.Errorf("failed to stop the main HTTP server %w", err)
```
yurishkuro, line 148: I think logging here is more appropriate, because there is more work left for this function to do, and I am not certain the caller actually logs the returned error (although it should).

cmd/collector/main.go
yurishkuro: can you use log.Fatal here?

cmd/collector/app/collector.go
tzzed: You' re right, it's more safe.

yurishkuro (dismissed):
Thanks

yurishkuro:
you're getting linter failures in CI, please run `make lint`

```
Lint Failures
cmd/collector/app/collector.go:96:9: if block ends with a return statement, so drop this else and outdent its block (move short variable declaration to its own line if necessary)
cmd/collector/app/collector.go:109:9: if block ends with a return statement, so drop this else and outdent its block (move short variable declaration to its own line if necessary)
cmd/collector/app/collector.go:123:9: if block ends with a return statement, so drop this else and outdent its block (move short variable declaration to its own line if necessary)
```

yurishkuro (approved):

Issue #2586, opened 2020-10-25
Flaky test: `TestAddCertsToWatch_remove_ca`


albertteoh:
**Describe the bug**

**To Reproduce**
https://travis-ci.com/github/albertteoh/jaeger/jobs/406895163

```
=== RUN   TestAddCertsToWatch_remove_ca
    cert_watcher_test.go:307: 
        	Error Trace:	cert_watcher_test.go:307
        	Error:      	Should be true
        	Test:       	TestAddCertsToWatch_remove_ca
--- FAIL: TestAddCertsToWatch_remove_ca (0.00s)
```

**Expected behavior**
Passing test.

**What troubleshooting steps did you try?**
Adding the following at the end of the test:
```
fmt.Println(logObserver.All())
```

Prints:
```
[
{{warn 2020-10-25 16:37:28.214382 +1100 AEDT m=+0.001567434  Certificate has been removed, using the last known version undefined } [{certificate 15 0 /var/folders/3p/yms48z2s6v7c8fy2m_1481g00000gn/T/ca.crt611136954 <nil>}]}
{{warn 2020-10-25 16:37:28.214443 +1100 AEDT m=+0.001628482  Certificate has been removed, using the last known version undefined } [{certificate 15 0 /var/folders/3p/yms48z2s6v7c8fy2m_1481g00000gn/T/clientCa.crt633105361 <nil>}]}
]

```

I believe the root cause is a race where the following [`waitUntil`](https://github.com/jaegertracing/jaeger/blob/4062992985bed23b53ef88d6e6b969521de8ad58/pkg/config/tlscfg/cert_watcher_test.go#L303) finishes after the first detected "Certificate has been removed..." log, and before the second `clientCAFile` deletion is detected by the `certWatcher`:
```
waitUntil(func() bool {
	return logObserver.FilterMessage("Certificate has been removed, using the last known version").Len() > 0
}, 100, time.Millisecond*100)
```

Causing the [second assertion](https://github.com/jaegertracing/jaeger/blob/4062992985bed23b53ef88d6e6b969521de8ad58/pkg/config/tlscfg/cert_watcher_test.go#L307) to fail:
```
assert.True(t, logObserver.FilterMessage("Certificate has been removed, using the last known version").FilterField(zap.String("certificate", clientCaFile.Name())).Len() > 0)
```

**Proposed Fix**

[...]

PR #2592, opened 2020-10-27
Fix file descriptor leaks in tlscfg unit tests

chlunde:
There are a few file descriptor leaks in the unit tests, and on an error
condition in certwatcher.

This is an issue when checking for flaky tests with
`go test -test.count ..`. One leak is in an error condition in
certwatcher but it is unlikely to be a runtime issue because it happens
on initialization.

Related #2586


yurishkuro (approved):
lgtm

jpkrohling (approved):

yurishkuro:
Thanks!

PR #2597, opened 2020-10-28
Bump go-plugin to 1.3.0

### LINKED ISSUE #2593: Bump go-plugin to 1.3 accidentally reverted
This change was accidentally reverted (I checked with Pavol). I think the change is as simple as `go get github.com/hashicorp/go-plugin@v1.3.0; go mod tidy; git add go.sum go.mod; cd cmd/opentelemetry/; go mod tidy; git add go.sum`.

One thing I wonder is why the previous change modified `go.sum` in `cmd/opentelemetry/`. I see the replace directive there, but I wonder if *tidy* is the best approach to update `go.sum`, and if there is any point in doing it at all.

**To Reproduce**
1. Check version in go.mod

**Expected behavior**
Find 1.3 since #2261 is merged.

chlunde:
go-plugin was accidentally downgraded in a previous merge.

    go get github.com/hashicorp/go-plugin@v1.3.0; go mod tidy; git add go.sum go.mod
    cd cmd/opentelemetry/; go mod tidy; git add go.sum

Fixes #2593


yurishkuro (dismissed):

chlunde:
rebased

I think `go mod tidy`created a merge conflict because it deleted lines from go.sum. I wonder if it would be better with an "add only" mode for `go mod tidy`.

yurishkuro (approved):
Thanks!

PR #2598, opened 2020-10-28
Remove md-to-godoc

### LINKED ISSUE #2599: Remove md-to-godoc-gen and its dependencies
We have a target in the Makefile that, apparently, isn't being used by anything at the moment: `md-to-godoc-gen`. This depends on a library (`github.com/sectioneight/md-to-godoc`), whose owner seems to have changed the username. As reported in the mailing list, this might become a security risk.

If we do need this library, we should at least change it to use its current location.

yurishkuro:
We have not used it actively, and we don't make much use of godocs.

* Addresses report https://groups.google.com/d/msgid/jaeger-tracing/020db716-6bbc-4aed-a61c-35715f0a5c1fn%40googlegroups.com
* Resolves #2599

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -147,7 +147,7 @@ all-srcs:
 
 .PHONY: cover
 cover: nocover
-	$(GOTEST) -coverprofile cover.out ./...
+	$(GOTEST) -timeout 5m -coverprofile cover.out ./...
 	grep -E -v 'model.pb.*.go' cover.out > cover-nogen.out
 	mv cover-nogen.out cover.out
 	go tool cover -html=cover.out -o cover.html
--- a/Makefile
+++ b/Makefile
@@ -81,12 +81,6 @@ test-and-lint: test fmt lint
 go-gen:
 	@echo skipping go generate ./...
 
-.PHONY: md-to-godoc-gen
-md-to-godoc-gen:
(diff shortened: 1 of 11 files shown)


jpkrohling (dismissed):
LGTM.

jpkrohling:
Did you force-push to trigger the CI?

yurishkuro:
I made a PR from stale master, so force push was after rebase.

jpkrohling (approved):

cmd/agent/app/agent_test.go
```diff
@@ -156,13 +156,14 @@ func TestStartStopRace(t *testing.T) {
 			{
 				Model:    jaegerModel,
 				Protocol: compactProtocol,
+				Workers:  1,
```
yurishkuro, line 159: don't need 10 goroutines in the test

Makefile
```diff
@@ -153,7 +147,7 @@ all-srcs:
 
 .PHONY: cover
 cover: nocover
-	$(GOTEST) -coverprofile cover.out ./...
+	$(GOTEST) -timeout 5m -coverprofile cover.out ./...
```
yurishkuro, line 150: default timeout is 10min, which is the same as Travis's no-output timeout, so Travis kills the job before go test has a chance to timeout and print the stack dump if there is a deadlock.

PR #2602, opened 2020-10-29
Add golangci-lint

daxmc99:
- 
  Fixes https://github.com/jaegertracing/jaeger/issues/2503

- Enable a few linters on golangci-lint to start

Signed-off-by: Dax McDonald <31839142+daxmc99@users.noreply.github.com>

Happy Hacktoberfest 🎃

### DIFF
--- a/.github/workflows/ci-unit-tests.yml
+++ b/.github/workflows/ci-unit-tests.yml
@@ -37,6 +37,5 @@ jobs:
         uses: golangci/golangci-lint-action@v2
         with:
           version: v1.33
-          only-new-issues: true
           args: --verbose
 
--- a/.github/workflows/ci-unit-tests.yml
+++ b/.github/workflows/ci-unit-tests.yml
@@ -28,3 +28,13 @@ jobs:
         file: cover.out
         fail_ci_if_error: true
         verbose: true
+
+  golangci-lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v2
+        with:
+          version: v1.33
+          only-new-issues: true
--- a/.github/workflows/ci-unit-tests.yml
+++ b/.github/workflows/ci-unit-tests.yml
@@ -38,3 +38,5 @@ jobs:
         with:
           version: v1.33
           only-new-issues: true
+          args: --verbose
+
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,69 @@
+# options for analysis running
+run:
+  # default concurrency is a available CPU number
+  # timeout for analysis, e.g. 30s, 5m, default is 1m
+  timeout: 5m
+linters:
+  # please, do not use `enable-all`: it's deprecated and will be removed soon.
+  # inverted configuration with `enable-all` and `disable` is not scalable during updates of golangci-lint
+  disable-all: true
+  enable:
+    - typecheck
+    - dupl
+    - goprintffuncname
+    - govet
+    - gosec
+#    - nolintlint
+#    - rowserrcheck
+    - gofmt
+    - golint
+    - goimports
+#    - misspell
+#    - bodyclose
+#    - unconvert
+#    - ineffassign
+#    - staticcheck
+#    - interfacer
+#    - scopelint
+    - structcheck
+    - deadcode
+#    - depguard
+#    - dogsled
+#    - errcheck
+#    - funlen
+#    - goconst
+#    - gocritic
+#    - gocyclo
+#    - gosimple
+#    - stylecheck
+    - unused
+#    - varcheck
+#    - unparam
+#    - unconvert
+#    - whitespace
+
+issues:
+  include:
+    - EXC0002 # disable excluding of issues about comments from golint
+  # Excluding configuration per-path, per-linter, per-text and per-source
+  exclude-rules:
+    - path: _test\.go
+      linters:
+        - gomnd
+        - dupl
+        - interfacer
+    - path: cmd/opentelemetry
+      linters:
+        - gosec
+
+    - linters:
+        - stylecheck
+      text: "ST1000:"
+    - linters:
+        - gosec
+      text: "G104,G107"
+
+linters-settings:
+  funlen:
+    lines: 80
+    statements: 40
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -14,10 +14,8 @@ linters:
     - gofmt
     - golint
     - goimports
-    - staticcheck
     - structcheck
     - deadcode
-    - errcheck
     - gosimple
     - unused
     - bodyclose
@@ -31,13 +29,22 @@ issues:
         - gomnd
         - dupl
         - interfacer
+        - gosec
+        - bodyclose
+    - path: _nolint\.go
+      linters:
+        - golint
+    - path: /mocks/
+      linters:
+        - goimports
+
 
     - linters:
         - stylecheck
       text: "ST1000:"
     - linters:
         - gosec
-      text: "G104:|G107:"
+      text: "G104:|G107:|G404:"
     - path: cmd/opentelemetry
       linters:
         - gosec
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -11,34 +11,17 @@ linters:
     - typecheck
     - govet
     - gosec
-#    - nolintlint
-#    - rowserrcheck
     - gofmt
     - golint
     - goimports
-#    - misspell
-#    - bodyclose
-#    - unconvert
-#    - ineffassign
     - staticcheck
-#    - interfacer
-#    - scopelint
     - structcheck
     - deadcode
-#    - depguard
-#    - dogsled
-#    - errcheck
-#    - funlen
-#    - goconst
-#    - gocritic
-#    - gocyclo
-#    - gosimple
-#    - stylecheck
+    - errcheck
+    - gosimple
     - unused
-#    - varcheck
-#    - unparam
-#    - unconvert
-#    - whitespace
+    - bodyclose
+    - misspell
 
 issues:
   exclude-use-default: true
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -31,6 +31,7 @@ issues:
         - interfacer
         - gosec
         - bodyclose
+        - golint
     - path: _nolint\.go
       linters:
         - golint
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -9,8 +9,6 @@ linters:
   disable-all: true
   enable:
     - typecheck
-    - dupl
-    - goprintffuncname
     - govet
     - gosec
 #    - nolintlint
@@ -61,9 +59,5 @@ issues:
       text: "ST1000:"
     - linters:
         - gosec
-      text: "G104,G107"
+      text: "G104:G107:"
 
-linters-settings:
-  funlen:
-    lines: 80
-    statements: 40
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -20,7 +20,7 @@ linters:
 #    - bodyclose
 #    - unconvert
 #    - ineffassign
-#    - staticcheck
+    - staticcheck
 #    - interfacer
 #    - scopelint
     - structcheck
@@ -41,23 +41,21 @@ linters:
 #    - whitespace
 
 issues:
-  include:
-    - EXC0002 # disable excluding of issues about comments from golint
-  # Excluding configuration per-path, per-linter, per-text and per-source
+  exclude-use-default: true
   exclude-rules:
     - path: _test\.go
       linters:
         - gomnd
         - dupl
         - interfacer
-    - path: cmd/opentelemetry
-      linters:
-        - gosec
 
     - linters:
         - stylecheck
       text: "ST1000:"
     - linters:
         - gosec
-      text: "G104:G107:"
+      text: "G104:|G107:"
+    - path: cmd/opentelemetry
+      linters:
+        - gosec
 
--- a/cmd/anonymizer/main.go
+++ b/cmd/anonymizer/main.go
@@ -69,6 +69,7 @@ func main() {
 			}
 
 			for _, span := range spans {
+				span := span
 				writer.WriteSpan(&span)
 			}
 			writer.Close()
--- a/crossdock/main.go
+++ b/crossdock/main.go
@@ -111,6 +111,7 @@ func httpHealthCheck(logger *zap.Logger, service, healthURL string) {
 	for i := 0; i < 240; i++ {
 		res, err := http.Get(healthURL)
 		if err == nil && is2xxStatusCode(res.StatusCode) {
+			res.Body.Close()
 			logger.Info("Health check successful", zap.String("service", service))
 			return
 		}
--- a/model/converter/thrift/jaeger/from_domain.go
+++ b/model/converter/thrift/jaeger/from_domain.go
@@ -102,9 +102,8 @@ func (d domainToJaegerTransformer) keyValueToTag(kv *model.KeyValue) *jaeger.Tag
 
 func (d domainToJaegerTransformer) convertKeyValuesToTags(kvs model.KeyValues) []*jaeger.Tag {
 	jaegerTags := make([]*jaeger.Tag, len(kvs))
-	for idx, kv := range kvs {
-		kv := kv
-		jaegerTags[idx] = d.keyValueToTag(&kv)
+	for i := range kvs {
+		jaegerTags[i] = d.keyValueToTag(&kvs[i])
 	}
 	return jaegerTags
 }
--- a/model/converter/thrift/jaeger/from_domain.go
+++ b/model/converter/thrift/jaeger/from_domain.go
@@ -103,6 +103,7 @@ func (d domainToJaegerTransformer) keyValueToTag(kv *model.KeyValue) *jaeger.Tag
 func (d domainToJaegerTransformer) convertKeyValuesToTags(kvs model.KeyValues) []*jaeger.Tag {
 	jaegerTags := make([]*jaeger.Tag, len(kvs))
 	for idx, kv := range kvs {
+		kv := kv
 		jaegerTags[idx] = d.keyValueToTag(&kv)
 	}
 	return jaegerTags
--- a/Makefile
+++ b/Makefile
@@ -178,6 +178,11 @@ lint-staticcheck:
 		>> $(LINT_LOG) || true
 	@[ ! -s "$(LINT_LOG)" ] || (echo "Detected staticcheck failures:" | cat - $(LINT_LOG) && false)
 
+.PHONY: golangci-lint
+golangci-lint:
+	@echo Running golangci-lint
+	golangci-lint -v run
+
 .PHONY: lint
 lint: lint-staticcheck lint-gosec lint-otel
 	$(GOVET) ./...
--- a/cmd/query/app/ui/placeholder.go
+++ b/cmd/query/app/ui/placeholder.go
@@ -17,7 +17,7 @@
 package ui
 
 import (
-	"github.com/jaegertracing/jaeger/cmd/query/app/ui/placeholder"
+	assets "github.com/jaegertracing/jaeger/cmd/query/app/ui/placeholder"
 )
 
 // StaticFiles provides http filesystem with static files for UI
--- a/plugin/storage/badger/spanstore/read_write_test.go
+++ b/plugin/storage/badger/spanstore/read_write_test.go
@@ -139,14 +139,14 @@ func TestIndexSeeks(t *testing.T) {
 		traceOrder := make([]uint64, traces)
 
 		for i := 0; i < traces; i++ {
-			lowId := rand.Uint64()
-			traceOrder[i] = lowId
+			lowID := rand.Uint64()
+			traceOrder[i] = lowID
 			tid = tid.Add(time.Duration(time.Millisecond * time.Duration(i)))
 
 			for j := 0; j < spans; j++ {
 				s := model.Span{
 					TraceID: model.TraceID{
-						Low:  lowId,
+						Low:  lowID,
 						High: 1,
 					},
 					SpanID:        model.SpanID(rand.Uint64()),


yurishkuro:
I would prefer if there is a make target that would allow running all the lint checks locally, not only in GH actions.

jpkrohling:
> I would prefer if there were a make target that would allow running all the lint checks locally, not only in GH actions.

If we have a config in the repo like the `.golangci.yml` from this PR, it would be a matter of only running `golangci-lint` in the base directory, and we could still have the GitHub action for it. The advantage of using an action is that it caches the tool, which would save us a minute or two for each build.

That said, I'd rather have this PR use the official action instead of what we have here. Reference: https://github.com/observatorium/opentelemetry-collector-builder/blob/c725aab7de2c3f782f8273ae4cc96faf21747290/.github/workflows/go.yaml#L27-L39

.github/workflows/ci-golang-lint.yml
```diff
@@ -0,0 +1,19 @@
+name: golangci-lint
+on:
+  push:
+    branches: [branches]
+  pull_request:
+    branches: [branches]
```
yurishkuro, line 6:
```suggestion
    branches: [master]
```

.github/workflows/ci-golang-lint.yml
```diff
@@ -0,0 +1,19 @@
+name: golangci-lint
+on:
+  push:
+    branches: [branches]
```
yurishkuro, line 4:
```suggestion
    branches: [master]
```

.github/workflows/ci-golang-lint.yml
```diff
@@ -0,0 +1,19 @@
...
+  pull_request:
+    branches: [branches]
+
+jobs:
+  name: lint
```
yurishkuro, line 9:
```suggestion
  name: golangci-lint
```

daxmc99:
@yurishkuro thanks for the review! 

Do you think this should be a separate `yml` file? I was planning to nest it under `ci-unit-test.yml`

yurishkuro:
for now let's keep it in ci-unit-tests.yml

Can you actually enable any linters? The job is not checking anything right now.

jpkrohling:
> Can you actually enable any linters?

How did you assess that? The configuration looks sane to me and I would guess that the following linters were used, but I wonder if you are looking at some place with a positive confirmation about the linters that were used for the specific run.

https://github.com/jaegertracing/jaeger/pull/2602/files#diff-6179837f7df53a6f05c522b6b7bb566d484d5465d9894fb04910dd08bb40dcc9R10-R37

yurishkuro:
configuration says `disable-all: true`, and the job completed too quickly

jpkrohling:
It does say `disable-all`, but then it enables linters individually. As opposed to `enable-all` and disable individually. It took about 10s for the linting phase alone, which is what I would expect the job to take.

>   Ran golangci-lint in 9282ms

https://github.com/jaegertracing/jaeger/pull/2602/checks?check_run_id=1478020447

yurishkuro:
`gosec` alone takes 2min:
```sh
$ make lint-gosec
time gosec -quiet -exclude=G104,G107 -exclude-dir=cmd/opentelemetry ./...
      120.03 real       236.28 user       195.93 sys
```

daxmc99:
With the config file, you can use `golangci-lint -v run` to see what the linter picks up and how long it takes.
There are a few open lint issues. We can address those here or in a separate PR. 

<details>
<summary>Output</summary>

```
golangci-lint -v run
INFO [config_reader] Config search paths: [./ /Users/dax/projects/jaeger /Users/dax/projects /Users/dax /Users /]
INFO [config_reader] Used config file .golangci.yml
INFO [lintersdb] Active 10 linters: [deadcode gofmt goimports golint gosec govet staticcheck structcheck typecheck unused]
INFO [loader] Go packages loading at mode 575 (exports_file|files|imports|name|types_sizes|compiled_files|deps) took 1.012619813s
INFO [runner/filename_unadjuster] Pre-built 0 adjustments in 35.883375ms
INFO [linters context/goanalysis] analyzers took 83.376645ms with top 10 stages: buildir: 17.607961ms, goimports: 11.697858ms, gosec: 8.840719ms, gofmt: 7.785945ms, golint: 5.979534ms, SA1012: 1.844425ms, SA1013: 1.646506ms, printf: 1.482018ms, deadcode: 1.420496ms, ctrlflow: 1.231694ms
INFO [linters context/goanalysis] analyzers took 22.52245ms with top 10 stages: buildir: 17.317955ms, U1000: 5.204495ms
INFO [runner/skip dirs] Skipped 1 issues from dir examples/hotrod/services/frontend by pattern (^|/)examples($|/)
INFO [runner/skip dirs] Skipped 1 issues from dir examples/hotrod/services/customer by pattern (^|/)examples($|/)
INFO [runner/skip dirs] Skipped 69 issues from dir examples/hotrod/services/driver by pattern (^|/)examples($|/)
INFO [runner/skip dirs] Skipped 1 issues from dir examples/hotrod/services/route by pattern (^|/)examples($|/)
INFO [runner/max_same_issues] 9/12 issues with text "G404: Use of weak random number generator (math/rand instead of crypto/rand)" were hidden, use --max-same-issues
INFO [runner/max_same_issues] 4/7 issues with text "G306: Expect WriteFile permissions to be 0600 or less" were hidden, use --max-same-issues
INFO [runner/max_same_issues] 4/7 issues with text "File is not `goimports`-ed" were hidden, use --max-same-issues
INFO [runner] Issues before processing: 3983, after processing: 21
INFO [runner] Processors filtering stat (out/in): max_same_issues: 21/38, max_from_linter: 21/21, severity-rules: 21/21, cgo: 3983/3983, filename_unadjuster: 3983/3983, skip_files: 3983/3983, exclude-rules: 39/339, uniq_by_line: 38/39, path_prettifier: 3983/3983, nolint: 39/39, source_code: 21/21, path_shortener: 21/21, sort_results: 21/21, skip_dirs: 3911/3983, autogenerated_exclude: 339/3911, diff: 38/38, max_per_file_from_linter: 38/38, path_prefixer: 21/21, identifier_marker: 339/339, exclude: 339/339
INFO [runner] processing took 38.919486ms with stages: nolint: 16.716437ms, path_prettifier: 7.979696ms, autogenerated_exclude: 6.675571ms, identifier_marker: 3.172654ms, exclude-rules: 1.69026ms, skip_dirs: 1.570182ms, source_code: 442.928µs, cgo: 358.105µs, filename_unadjuster: 255.942µs, max_same_issues: 26.884µs, uniq_by_line: 11.137µs, max_per_file_from_linter: 9.045µs, path_shortener: 4.857µs, max_from_linter: 3.756µs, skip_files: 621ns, exclude: 392ns, diff: 384ns, severity-rules: 264ns, sort_results: 233ns, path_prefixer: 138ns
INFO [runner] linters took 369.097097ms with stages: goanalysis_metalinter: 224.257539ms, unused: 105.765528ms
model/converter/json/from_domain_test.go:142:10: G306: Expect WriteFile permissions to be 0600 or less (gosec)
		err := ioutil.WriteFile(outFile+"-actual.json", buf.Bytes(), 0644)
		       ^
cmd/collector/app/sanitizer/cache/auto_refresh_cache.go:136:40: G404: Use of weak random number generator (math/rand instead of crypto/rand) (gosec)
	return (interval / 2) + time.Duration(rand.Int63n(int64(interval/2)))
	                                      ^
cmd/collector/app/server/grpc_test.go:80:19: G102: Binds to all network interfaces (gosec)
	listener, err := net.Listen("tcp", ":0")
	                 ^
pkg/discovery/grpcresolver/grpc_resolver.go:68:12: G404: Use of weak random number generator (math/rand instead of crypto/rand) (gosec)
	random := rand.New(rand.NewSource(seed))
	          ^
pkg/cassandra/mocks/Iterator.go:18: File is not `goimports`-ed (goimports)
import cassandra "github.com/jaegertracing/jaeger/pkg/cassandra"
import mock "github.com/stretchr/testify/mock"
pkg/cassandra/mocks/Query.go:18: File is not `goimports`-ed (goimports)
import cassandra "github.com/jaegertracing/jaeger/pkg/cassandra"
import mock "github.com/stretchr/testify/mock"
pkg/cassandra/mocks/Session.go:18: File is not `goimports`-ed (goimports)
import cassandra "github.com/jaegertracing/jaeger/pkg/cassandra"
import mock "github.com/stretchr/testify/mock"
pkg/es/wrapper/wrapper_nolint.go:23:30: method Id should be ID (golint)
func (i IndexServiceWrapper) Id(id string) es.IndexService {
                             ^
pkg/es/wrapper/wrapper_nolint.go:28:30: method BodyJson should be BodyJSON (golint)
func (i IndexServiceWrapper) BodyJson(body interface{}) es.IndexService {
                             ^
cmd/ingester/app/processor/decorator/retry.go:54:18: G404: Use of weak random number generator (math/rand instead of crypto/rand) (gosec)
	rand:           rand.New(rand.NewSource(time.Now().UnixNano())),
	                ^
pkg/queue/bounded_queue_test.go:57:3: SA2001: empty critical section (staticcheck)
		startLock.Unlock()
		^
plugin/sampling/strategystore/static/strategy_store_test.go:346:21: G306: Expect WriteFile permissions to be 0600 or less (gosec)
	require.NoError(t, ioutil.WriteFile(dstFile, srcBytes, 0644))
	                   ^
plugin/sampling/strategystore/static/strategy_store_test.go:367:21: G306: Expect WriteFile permissions to be 0600 or less (gosec)
	require.NoError(t, ioutil.WriteFile(dstFile, []byte(newStr), 0644))
	                   ^
storage/spanstore/token_propagation_test.go:25:2: G101: Potential hardcoded credentials (gosec)
	token := "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJhZG1pbiIsIm5hbWUiOiJKb2huIERvZSIsImlhdCI"
	^
cmd/anonymizer/main.go:72:22: G601: Implicit memory aliasing in for loop. (gosec)
				writer.WriteSpan(&span)
				                 ^
model/converter/thrift/jaeger/from_domain.go:106:37: G601: Implicit memory aliasing in for loop. (gosec)
		jaegerTags[idx] = d.keyValueToTag(&kv)
		                                  ^
plugin/storage/badger/spanstore/read_write_test.go:142:4: var `lowId` should be `lowID` (golint)
			lowId := rand.Uint64()
			^
cmd/agent/app/reporter/grpc/reporter_test.go:60:2: SA5001: should check returned error before deferring conn.Close() (staticcheck)
	defer conn.Close()
	^
cmd/agent/app/reporter/grpc/reporter_test.go:98:2: SA5001: should check returned error before deferring conn.Close() (staticcheck)
	defer conn.Close()
	^
cmd/query/app/grpc_handler_test.go:138:12: G102: Binds to all network interfaces (gosec)
	lis, _ := net.Listen("tcp", ":0")
	          ^
cmd/query/app/token_propagation_hander_test.go:31:2: G101: Potential hardcoded credentials (gosec)
	bearerToken := "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJhZG1pbiIsIm5hbWUiOiJKb2huIERvZSIsImlhdCI"
	^
INFO File cache stats: 18 entries of total size 105.4KiB
INFO Memory: 16 samples, avg is 90.0MB, max is 141.1MB
INFO Execution took 1.446295162s
```

</details>

I am specifying `--only-new-issues` in the Github Action config though. I don't believe this will throw any lint issues on the introductory PR. 

> gosec alone takes 2min
This seems like a great reason to use golangci-lint, the go sec errors I am seeing look the same, and golangci-lint runs much faster for me (`1.446295162` seconds)

jpkrohling:
> I am specifying --only-new-issues in the Github Action config though.

I think this is the key :-)

yurishkuro:
I am not sure what only-new-issues means: you still have to scan everything to discover IF there are new issues.

Why can't we run with -v in CI as well? That would at least provide some proof that the linters are running.

.golangci.yml
```diff
@@ -0,0 +1,61 @@
...
+    - unused
+#    - varcheck
+#    - unparam
+#    - unconvert
+#    - whitespace
+
+issues:
+  exclude-use-default: true
+  exclude-rules:
+    - path: _test\.go
```
yurishkuro, line 46: What about the basic golint, is it included? It finds lots of errors in thrift-generated files that we suppress (see greg in the makefile)

.github/workflows/ci-unit-tests.yml
```diff
@@ -28,3 +28,13 @@ jobs:
...
+
+  golangci-lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v2
+        with:
+          version: v1.33
+          only-new-issues: true
```
yurishkuro, line 40: can it be run in verbose mode?

.golangci.yml
daxmc99: Yes, this only excludes `gomnd`,`dupl` and `interfacer`. These typically give false positives in tests (or they will when they are enabled).

.golangci.yml
```diff
@@ -0,0 +1,61 @@
...
+  timeout: 5m
+linters:
+  # please, do not use `enable-all`: it's deprecated and will be removed soon.
+  # inverted configuration with `enable-all` and `disable` is not scalable during updates of golangci-lint
+  disable-all: true
+  enable:
+    - typecheck
+    - govet
+    - gosec
+#    - nolintlint
```
jpkrohling, line 14: Perhaps remove the commented out linters?

daxmc99: Done

.github/workflows/ci-unit-tests.yml
```diff
@@ -28,3 +28,15 @@ jobs:
...
```
yurishkuro, line 40: Could you please explain what `only new issues` means? Does it mean it only scans changed files (if so how does it determine that, against master?), or that it reports errors previously unseen (if so where does it store the state?) ?

yurishkuro:
@jpkrohling my main concern with this is that by adding a whole set of new linters with `new changes only` mode, we're suppressing a lot of existing issues that will make life painful for future contributors, because their PRs may touch files where these issues exist and start breaking, forcing people to fix previously issues and distracting from the main goal of the PRs.

I think I would prefer instead to add linters one by one or in batches along with the corresponding fixes of ALL issues that a given linter flags. That way each PR could be reasonably sized and not push the burden of the actual fixes on the next random person who happens to touch the files.

For example, with the config in this PR:
```
$ golangci-lint -v --max-issues-per-linter 500 --max-same-issues 500 run | grep Error | wc -l
     189
```

jpkrohling:
> because their PRs may touch files where these issues exist and start breaking, forcing people to fix previously issues and distracting from the main goal of the PRs

Good point and I agree with your proposal.

daxmc99:
I'm happy to either reduce the number of enabled linters or fix the found issues. 
To get this merged it might make more sense to enable less linters for now

yurishkuro:
I'd recommend enabling only those linters that currently pass on the whole repository.

.github/workflows/ci-unit-tests.yml
daxmc99: This has been dropped since all linters are currently passing. 
Per  https://golangci-lint.run/usage/configuration/ 
```
 # Show only new issues: if there are unstaged changes or untracked files,
  # only those changes are analyzed, else only changes in HEAD~ are analyzed.
  # It's a super-useful option for integration of golangci-lint into existing
  # large codebase. It's not practical to fix all existing issues at the moment
  # of integration: much better don't allow issues in new code.
  # Default is false.
```
This is meant to allow easier integration into a large codebase by only reporting errors if the added code contains them. It does not report previously seen linter errors (those that are already in master). 
This has been dropped though because all outstanding linter issues have been fixed.

.golangci.yml
daxmc99: Added `golint`.

daxmc99:
After removing a few linters there were a few small issues remaining. These all looked like legitimate issues so I proceeded to fix them

<details>
<summary>Linter errors</summary>

INFO [runner] linters took 142.248311ms with stages: goanalysis_metalinter: 112.589465ms, unused: 7.165119ms 
storage/mocks/ArchiveFactory.go:17: File is not `goimports`-ed (goimports)
import mock "github.com/stretchr/testify/mock"
import spanstore "github.com/jaegertracing/jaeger/storage/spanstore"
import storage "github.com/jaegertracing/jaeger/storage"
storage/mocks/Factory.go:17: File is not `goimports`-ed (goimports)
import dependencystore "github.com/jaegertracing/jaeger/storage/dependencystore"
import metrics "github.com/uber/jaeger-lib/metrics"
import mock "github.com/stretchr/testify/mock"
import spanstore "github.com/jaegertracing/jaeger/storage/spanstore"
import storage "github.com/jaegertracing/jaeger/storage"
import zap "go.uber.org/zap"
cmd/query/app/ui/placeholder.go:20: File is not `goimports`-ed (goimports)
        "github.com/jaegertracing/jaeger/cmd/query/app/ui/placeholder"
plugin/storage/badger/spanstore/read_write_test.go:142:4: var `lowId` should be `lowID` (golint)
                        lowId := rand.Uint64()
                        ^
model/converter/thrift/jaeger/from_domain.go:106:37: G601: Implicit memory aliasing in for loop. (gosec)
                jaegerTags[idx] = d.keyValueToTag(&kv)
                                                  ^
cmd/anonymizer/main.go:72:22: G601: Implicit memory aliasing in for loop. (gosec)
                                writer.WriteSpan(&span)
                                                 ^
crossdock/main.go:112:23: response body must be closed (bodyclose)
                res, err := http.Get(healthURL)
                                    ^
INFO File cache stats: 7 entries of total size 35.1KiB 


</details>

Happy to revert any commits as needed

yurishkuro:
The run in the CI succeeds, but when I run it locally with the current config file I get a few errors. Where is the discrepancy coming from?
```sh
$ golangci-lint -v --max-issues-per-linter 500 --max-same-issues 500 run
```

daxmc99:
Hmm, are you running on this branch? That exact command passes for me with
golangci-lint version `1.33.0`

cmd/anonymizer/main.go
```diff
@@ -69,6 +69,7 @@ func main() {
 			}
 
 			for _, span := range spans {
+				span := span
 				writer.WriteSpan(&span)
```
yurishkuro, line 73: this is weird. I think I know what the linter is complaining about, and a better solution is
```suggestion
			for i := range spans {
				writer.WriteSpan(&spans[i])
```

crossdock/main.go
```diff
@@ -111,6 +111,7 @@ func httpHealthCheck(logger *zap.Logger, service, healthURL string) {
 	for i := 0; i < 240; i++ {
 		res, err := http.Get(healthURL)
 		if err == nil && is2xxStatusCode(res.StatusCode) {
+			res.Body.Close()
```
yurishkuro, line 114: fine here because it's a test code, but strictly speaking we must close when err==nil regardless of status code

model/converter/thrift/jaeger/from_domain.go
```diff
@@ -103,6 +103,7 @@ func (d domainToJaegerTransformer) keyValueToTag(kv *model.KeyValue) *jaeger.Tag
 func (d domainToJaegerTransformer) convertKeyValuesToTags(kvs model.KeyValues) []*jaeger.Tag {
 	jaegerTags := make([]*jaeger.Tag, len(kvs))
 	for idx, kv := range kvs {
+		kv := kv
 		jaegerTags[idx] = d.keyValueToTag(&kv)
```
yurishkuro, line 107:
```suggestion
	for i := range kvs {
		jaegerTags[i] = d.keyValueToTag(&kvs[i])
```

yurishkuro (commented):
you're right, I just ran it on master so it might have tripped on the issues that you fixed.

.golangci.yml
```diff
@@ -0,0 +1,52 @@
...
+linters:
+  # please, do not use `enable-all`: it's deprecated and will be removed soon.
+  # inverted configuration with `enable-all` and `disable` is not scalable during updates of golangci-lint
+  disable-all: true
+  enable:
+    - typecheck
+    - govet
+    - gosec
+    - gofmt
+    - golint
```
yurishkuro, line 15: I find it very suspicious that you're not getting errors from golint given the current exclusions below. Usually `thrift-gen` produces a lot of errors for golint, which is why we excluded it in the makefile

yurishkuro:
Also, what makes it exclude examples?
```
INFO [runner/skip dirs] Skipped 69 issues from dir examples/hotrod/services/driver by pattern (^|/)examples($|/)
INFO [runner/skip dirs] Skipped 1 issues from dir examples/hotrod/services/customer by pattern (^|/)examples($|/)
INFO [runner/skip dirs] Skipped 1 issues from dir examples/hotrod/services/route by pattern (^|/)examples($|/)
INFO [runner/skip dirs] Skipped 1 issues from dir examples/hotrod/services/frontend by pattern (^|/)examples($|/)
```
I think examples code is a fair game for linter (unlike generated code).

crossdock/main.go
daxmc99: Yeah, I wanted to minimize the diff here

cmd/anonymizer/main.go
daxmc99: I am not familiar with the intention of the code but 
`for _, span := range spans {
	// something with &span
```
looks like the developer wants to use copies of the original item, not explicitly manipulating the items in the array. 
If the intention is to use the items directly then I believe this change makes sense.

model/converter/thrift/jaeger/from_domain.go
daxmc99: Same comment as above, just want to confirm that the intention is to use the reference to the object itself and not a copy. 
Although `kvs` is passed in as an arg so it is already being passed by value

cmd/anonymizer/main.go
yurishkuro: writer treats the spans as immutable, so the copying is not necessary, only extra memory allocation

jpkrohling:
Closing due to inactivity.

Issue #2459, opened 2020-09-09
jaeger has a memory leak  when use badger


lichenglin:
**Describe the bug**
I'm doing a stress on badger and found that jaeger has a memory leak .

**To Reproduce**
1. start jaeger with docker
`docker run --rm   --name jaeger-badger -p 16831:6831/udp   -p 16832:6832/udp   -p 15778:5778   -p 15775:5775 -p 26686:16686 -p 24269:14269 -e SPAN_STORAGE_TYPE=badger -e BADGER_EPHEMERAL=false -e BADGER_DIRECTORY_VALUE=/badger/data -e BADGER_DIRECTORY_KEY=/badger/key  jaegertracing/all-in-one:1.19 --badger.span-store-ttl 1m --badger.truncate`

2.run a stress application to send data to jaeger. My application send 7000/s span data to jaeger with udp sender.

3.monitoring the memory of CONTAINER ;

**Expected behavior**
jaeger's memory should be stable at some point not always increase.

**Screenshots**
<img width="982" alt="image" src="https://user-images.githubusercontent.com/8294900/92560762-87cb6180-f2a5-11ea-8696-d49886ca121b.png">
as time goes on ,the memory become higher.

**Version (please complete the following information):**
 - OS: centos7
 - Jaeger version: 1.9
 - Deployment: docker

**What troubleshooting steps did you try?**
I try to adjust some param.
such as `--collector.queue-size`,`--collector.queue-size-memory`
`--badger.span-store-ttl` ,`--badger.consistency`,`--badger.maintenance-interval`

**Additional context**

jpkrohling:
I investigated a possible memory leak some time ago, and documented my findings here: https://issues.redhat.com/browse/TRACING-1071 . Would you mind going over that and checking if your findings match mine?

lichenglin:
> I investigated a possible memory leak some time ago, and documented my findings here: https://issues.redhat.com/browse/TRACING-1071 . Would you mind going over that and checking if your findings match mine?

I have checked.It is the same issue.

<img width="366" alt="image" src="https://user-images.githubusercontent.com/8294900/92671436-d7169e00-f348-11ea-9b86-e459171648be.png">

I use linux command and found the jaeger process use about 10g memory.

But when I use pprof. It shows this.

<img width="781" alt="image" src="https://user-images.githubusercontent.com/8294900/92671495-04fbe280-f349-11ea-92f5-b885ee43f048.png">

[...]

PR #2606, opened 2020-10-29
Use goleak to capture goroutine leak

Ashmita152:

This PR uses goleak (https://github.com/uber-go/goleak) to identify goroutine leak as part of unit tests.

We are adding a `TestMain` function for each test package. Goleak provides a way to exclude top functions, hence we are adding `TolerantVerifyLeakMain` with those exceptions. If this is accepted change, I'll add changes for rest of the test packages too.

### DIFF
--- a/cmd/flags/flags_test.go
+++ b/cmd/flags/flags_test.go
@@ -21,7 +21,12 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"go.uber.org/goleak"
 )
+
+func TestMain(m *testing.M) {
+	goleak.VerifyTestMain(m)
+}
 
 func TestParseJaegerTags(t *testing.T) {
 
--- a/pkg/testutils/testutils.go
+++ b/pkg/testutils/testutils.go
@@ -1,0 +1,46 @@
+// Copyright (c) 2020 The Jaeger Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package testutils
+
+import (
+	"testing"
+
+	"go.uber.org/goleak"
+)
+
+// TolerantVerifyLeakMain verifies go leaks but excludes the go routines that are
+// launched as side effects of some of our dependencies.
+func TolerantVerifyLeakMain(m *testing.M) {
+	goleak.VerifyTestMain(m,
+		// https://github.com/census-instrumentation/opencensus-go/blob/d7677d6af5953e0506ac4c08f349c62b917a443a/stats/view/worker.go#L34
+		goleak.IgnoreTopFunction("go.opencensus.io/stats/view.(*worker).start"),
+		// https://github.com/kubernetes/klog/blob/c85d02d1c76a9ebafa81eb6d35c980734f2c4727/klog.go#L417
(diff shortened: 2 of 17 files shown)


pkg/testutils/testutils.go
```diff
@@ -0,0 +1,46 @@
...
+
+	"go.uber.org/goleak"
+)
+
+// TolerantVerifyLeakMain verifies go leaks but excludes the go routines that are
+// launched as side effects of some of our dependencies.
+func TolerantVerifyLeakMain(m *testing.M) {
+	goleak.VerifyTestMain(m,
+		// https://github.com/census-instrumentation/opencensus-go/blob/d7677d6af5953e0506ac4c08f349c62b917a443a/stats/view/worker.go#L34
+		goleak.IgnoreTopFunction("go.opencensus.io/stats/view.(*worker).start"),
```
yurishkuro, line 28: where do we even use opencensus?

pkg/testutils/testutils.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"testing"
+
+	"go.uber.org/goleak"
+)
+
+// TolerantVerifyLeakMain verifies go leaks but excludes the go routines that are
+// launched as side effects of some of our dependencies.
+func TolerantVerifyLeakMain(m *testing.M) {
```
yurishkuro, line 25: This is unfortunate. Seems to indicate that the tests are not closing the resources properly, or the dependencies are not exposing proper clean-up functions.

pkg/testutils/testutils.go
```diff
@@ -0,0 +1,46 @@
...
+	goleak.VerifyTestMain(m,
+		// https://github.com/census-instrumentation/opencensus-go/blob/d7677d6af5953e0506ac4c08f349c62b917a443a/stats/view/worker.go#L34
+		goleak.IgnoreTopFunction("go.opencensus.io/stats/view.(*worker).start"),
+		// https://github.com/kubernetes/klog/blob/c85d02d1c76a9ebafa81eb6d35c980734f2c4727/klog.go#L417
+		goleak.IgnoreTopFunction("k8s.io/klog/v2.(*loggingT).flushDaemon"),
+		goleak.IgnoreTopFunction("k8s.io/klog.(*loggingT).flushDaemon"),
+		// https://github.com/dgraph-io/badger/blob/v1.5.3/y/watermark.go#L54
+		goleak.IgnoreTopFunction("github.com/dgraph-io/badger/y.(*WaterMark).process"),
```
yurishkuro, line 33: if feels like badger should really have proper clean-ups, maybe we're not invoking them?

cmd/flags/flags_test.go
```diff
@@ -21,8 +21,13 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"go.uber.org/goleak"
 )
 
+func TestMain(m *testing.M) {
```
yurishkuro, line 27: Manually adding these functions is not a scalable approach, next time someone will surely forget to add one.

I would prefer to have a CI check, similar to `make nocover`, that will fail if some packages don't have this function (assuming we want to go with checking these leaks).

yurishkuro (commented):
What is the motivation for using goleak? Does it really help identify issues that are not specific to the tests themselves? E.g. by adding these calls throughout, did you actually run into places where GRs were leaked (not counting external dependencies)?

Ashmita152:
I am very new to golang but I saw goleak is adopted by multiple open-source golang projects like thanos. I just thought it will be good to catch such leaks in unittests. I have no experience with it. 
Possibly the issues with badger we noticed as part of this PR when using goleak with badgerexporter test is maybe related to the behaviour noticed in this https://github.com/jaegertracing/jaeger/issues/2459

yurishkuro:
How did you arrive at the exclusion list, by testing packages individually? I would prefer not to have any exclusions and just not enable goleaks for packages that leak, for now.

Ashmita152:
Right, I added that `TestMain` at each package one by one and when it fails for some package like badgerexporter, it prints the top of the stack which I added to the exclusion list. Sure I will remove that.

PR #2609, opened 2020-10-30
Return buffers to pool on network errors or queue overflow

chlunde:
Use of `sync.Pool` requires a `.Put` for every `.Get`.
On success, the downstream `ThriftProcessor` invoked `.Put` via `.DataRecd`
but in error cases there seems to be a leak.

I think I found a leak when reviewing this code for deadlock-related issues. 

Free memory in error cases.

### DIFF
--- a/cmd/agent/app/servers/tbuffered_server.go
+++ b/cmd/agent/app/servers/tbuffered_server.go
@@ -98,9 +98,11 @@ func (s *TBufferedServer) Serve() {
 				s.metrics.PacketsProcessed.Inc(1)
 				s.updateQueueSize(1)
 			default:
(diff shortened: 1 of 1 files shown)


yurishkuro (approved):
It's not a memory leak, the buffers would be gc-ed, but it doesn't hurt to put them back in error paths.

yurishkuro:
Thanks!

chlunde:
Oh, sorry, I hate that the commit comment I made was incorrect :sob: For some reason I always assumed sync.Pool retained a reference to the objects, allocated as an array or similar..

yurishkuro:
no worries, I changed the commit message during merge

PR #2611, opened 2020-11-03
added multi arch images builds support for make file

morlay:
Signed-off-by: Morlay <morlay.null@gmail.com>

- to support #2292, split from https://github.com/jaegertracing/jaeger/pull/2325

- publishable multi-arch base images
- added `make build-binaries-linux-archs` for all binaries for multi arch
- updates makefile to switch `docker build` to `docker buildx build` for jaeger images
- removes default value of Dockerfile arg `TARGETARCH`

local run（need buildx installed）
```bash
make build-binaries-linux-archs
DOCKER_BASE_NAMESPACE=querycapjaegertracing DOCKER_NAMESPACE=querycapjaegertracing make docker-push

# if not set  DOCKER_BASE_NAMESPACE
# need to 
# - run local registry to push base images 
# - create builder with --driver-opt=network=host (for let buildkit could access `localhost:5000` to pull base images)
docker run -d -p 5000:5000 --name registry registry:2
docker buildx create --use --driver-opt=network=host
```
images https://hub.docker.com/u/querycapjaegertracing

upstream `cassandra` not provide `s390x` image, so now just `amd64 & arm64` first.

**This PR should**

* no break changes of `make docker` and current ci flow. (switch to buildx but still build for amd64 only)
* no Dockfile changes.

**Next PRs will add**

- [ ] `make docker-images-all-in-one`
- [ ] multi arch images pubishing workflow for ci

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,8 +2,6 @@ JAEGER_IMPORT_PATH=github.com/jaegertracing/jaeger
 STORAGE_PKGS = ./plugin/storage/integration/...
 OTEL_COLLECTOR_DIR = ./cmd/opentelemetry
 
-include docker/Makefile
-
 # all .go files that are not auto-generated and should be auto-formatted and linted.
 ALL_SRC := $(shell find . -name '*.go' \
 				   -not -name 'doc.go' \
@@ -68,6 +66,18 @@ COLOR_FAIL=$(shell printf "\033[31mFAIL\033[0m")
 COLORIZE=$(SED) ''/PASS/s//$(COLOR_PASS)/'' | $(SED) ''/FAIL/s//$(COLOR_FAIL)/''
 DOCKER_NAMESPACE?=jaegertracing
 DOCKER_TAG?=latest
+
+# list for multi-arch image publishing
+TARGET_ARCHS ?= amd64 arm64
+
+BUILDX_PUSH ?= false
+DOCKER_BUILD ?= docker build --build-arg=TARGETARCH=$(GOARCH)
+ifeq ($(BUILDX_PUSH),true)
+	DOCKER_BUILD = DOCKER_CLI_EXPERIMENTAL="enabled" docker buildx build --builder=jaeger-builder --push $(foreach arch,${TARGET_ARCHS},--platform=linux/${arch})
+endif
+
+# make sure sub makefiles could use the make variables defined in root makefile.
+include docker/Makefile
 
 MOCKERY=mockery
 
@@ -220,7 +230,7 @@ build-anonymizer:
 .PHONY: docker-hotrod
 docker-hotrod:
 	GOOS=linux $(MAKE) build-examples
-	docker build -t $(DOCKER_NAMESPACE)/example-hotrod:${DOCKER_TAG} ./examples/hotrod --build-arg TARGETARCH=$(GOARCH)
+	$(DOCKER_BUILD) -t $(DOCKER_NAMESPACE)/example-hotrod:${DOCKER_TAG} ./examples/hotrod
 
 .PHONY: run-all-in-one
 run-all-in-one: build-ui
@@ -289,6 +299,12 @@ docker: build-binaries-linux docker-images-only
 build-binaries-linux:
 	GOOS=linux GOARCH=amd64 $(MAKE) build-platform-binaries
 
+.PHONY: build-binaries-linux-archs
+build-binaries-linux-archs:
+	for arch in $(TARGET_ARCHS) ; do \
+		GOOS=linux GOARCH=$${arch} $(MAKE) build-platform-binaries; \
+	done
+
 .PHONY: build-binaries-windows
 build-binaries-windows:
 	GOOS=windows GOARCH=amd64 $(MAKE) build-platform-binaries
@@ -332,13 +348,13 @@ build-all-platforms: build-binaries-linux build-binaries-windows build-binaries-
 
 .PHONY: docker-images-cassandra
 docker-images-cassandra:
-	docker build -t $(DOCKER_NAMESPACE)/jaeger-cassandra-schema:${DOCKER_TAG} plugin/storage/cassandra/
+	$(DOCKER_BUILD) -t $(DOCKER_NAMESPACE)/jaeger-cassandra-schema:${DOCKER_TAG} plugin/storage/cassandra/
 	@echo "Finished building jaeger-cassandra-schema =============="
 
 .PHONY: docker-images-elastic
 docker-images-elastic:
-	docker build -t $(DOCKER_NAMESPACE)/jaeger-es-index-cleaner:${DOCKER_TAG} plugin/storage/es
-	docker build -t $(DOCKER_NAMESPACE)/jaeger-es-rollover:${DOCKER_TAG} plugin/storage/es -f plugin/storage/es/Dockerfile.rollover
+	$(DOCKER_BUILD) -t $(DOCKER_NAMESPACE)/jaeger-es-index-cleaner:${DOCKER_TAG} plugin/storage/es
+	$(DOCKER_BUILD) -t $(DOCKER_NAMESPACE)/jaeger-es-rollover:${DOCKER_TAG} plugin/storage/es -f plugin/storage/es/Dockerfile.rollover
 	@echo "Finished building jaeger-es-indices-clean =============="
 
 docker-images-jaeger-backend: TARGET = release
@@ -348,21 +364,20 @@ docker-images-jaeger-backend-debug: SUFFIX = -debug
 .PHONY: docker-images-jaeger-backend docker-images-jaeger-backend-debug
 docker-images-jaeger-backend docker-images-jaeger-backend-debug: create-baseimg create-debugimg
 	for component in agent collector query ingester ; do \
-		docker build --target $(TARGET) \
+		$(DOCKER_BUILD) --target $(TARGET) \
 			--tag $(DOCKER_NAMESPACE)/jaeger-$$component$(SUFFIX):${DOCKER_TAG} \
 			--build-arg base_image=$(BASE_IMAGE) \
 			--build-arg debug_image=$(DEBUG_IMAGE) \
-			--build-arg TARGETARCH=$(GOARCH) \
 			cmd/$$component ; \
 		echo "Finished building $$component ==============" ; \
 	done
-	docker build -t $(DOCKER_NAMESPACE)/jaeger-opentelemetry-collector:${DOCKER_TAG} -f ${OTEL_COLLECTOR_DIR}/cmd/collector/Dockerfile cmd/opentelemetry/cmd/collector --build-arg TARGETARCH=$(GOARCH)
-	docker build -t $(DOCKER_NAMESPACE)/jaeger-opentelemetry-agent:${DOCKER_TAG} -f ${OTEL_COLLECTOR_DIR}/cmd/agent/Dockerfile cmd/opentelemetry/cmd/agent --build-arg TARGETARCH=$(GOARCH)
-	docker build -t $(DOCKER_NAMESPACE)/jaeger-opentelemetry-ingester:${DOCKER_TAG} -f ${OTEL_COLLECTOR_DIR}/cmd/ingester/Dockerfile cmd/opentelemetry/cmd/ingester --build-arg TARGETARCH=$(GOARCH)
+	$(DOCKER_BUILD) -t $(DOCKER_NAMESPACE)/jaeger-opentelemetry-collector:${DOCKER_TAG} -f ${OTEL_COLLECTOR_DIR}/cmd/collector/Dockerfile cmd/opentelemetry/cmd/collector
+	$(DOCKER_BUILD) -t $(DOCKER_NAMESPACE)/jaeger-opentelemetry-agent:${DOCKER_TAG} -f ${OTEL_COLLECTOR_DIR}/cmd/agent/Dockerfile cmd/opentelemetry/cmd/agent
+	$(DOCKER_BUILD) -t $(DOCKER_NAMESPACE)/jaeger-opentelemetry-ingester:${DOCKER_TAG} -f ${OTEL_COLLECTOR_DIR}/cmd/ingester/Dockerfile cmd/opentelemetry/cmd/ingester
 
 .PHONY: docker-images-tracegen
 docker-images-tracegen:
-	docker build -t $(DOCKER_NAMESPACE)/jaeger-tracegen:${DOCKER_TAG} cmd/tracegen/ --build-arg TARGETARCH=$(GOARCH)
+	$(DOCKER_BUILD) -t $(DOCKER_NAMESPACE)/jaeger-tracegen:${DOCKER_TAG} cmd/tracegen/
 	@echo "Finished building jaeger-tracegen =============="
 
 .PHONY: docker-images-anonymizer
@@ -390,6 +405,10 @@ docker-push:
 		docker push $(DOCKER_NAMESPACE)/jaeger-$$component ; \
 	done
 
+.PHONY: dockerx-push
+dockerx-push: setup-buildx-builder
+	$(MAKE) docker-images-only BUILDX_PUSH=true
+
 .PHONY: build-crossdock-linux
 build-crossdock-linux:
 	GOOS=linux $(GOBUILD) -o ./crossdock/crossdock-linux ./crossdock/main.go
@@ -406,7 +425,7 @@ build-crossdock-ui-placeholder:
 
 .PHONY: build-crossdock
 build-crossdock: build-crossdock-ui-placeholder build-binaries-linux build-crossdock-linux docker-images-cassandra docker-images-jaeger-backend
-	docker build -t $(DOCKER_NAMESPACE)/test-driver:${DOCKER_TAG} crossdock/
+	$(DOCKER_BUILD) -t $(DOCKER_NAMESPACE)/test-driver:${DOCKER_TAG} crossdock/
 	@echo "Finished building test-driver ==============" ; \
 
 .PHONY: build-and-run-crossdock
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -3,18 +3,24 @@ ROOT_IMAGE ?= alpine:3.12
 CERT_IMAGE := alpine:3.12
 GOLANG_IMAGE := golang:1.15-alpine
 
-BASE_IMAGE := localhost/baseimg:$(VERSION)-$(shell echo $(ROOT_IMAGE) | tr : -)
-DEBUG_IMAGE := localhost/debugimg:$(VERSION)-$(shell echo $(GOLANG_IMAGE) | tr : -)
+BASE_IMAGE := $(DOCKER_NAMESPACE)/baseimg:$(VERSION)-$(shell echo $(ROOT_IMAGE) | tr : -)
+DEBUG_IMAGE := $(DOCKER_NAMESPACE)/debugimg:$(VERSION)-$(shell echo $(GOLANG_IMAGE) | tr : -)
 
 create-baseimg-debugimg: create-baseimg create-debugimg
 
+dockerx-baseimg-debugimg: setup-buildx-builder
+	$(MAKE) create-baseimg-debugimg BUILDX_PUSH=true
+
 create-baseimg:
-	docker build -t $(BASE_IMAGE) \
+	$(DOCKER_BUILD) -t $(BASE_IMAGE) \
 		--build-arg root_image=$(ROOT_IMAGE) \
 		--build-arg cert_image=$(CERT_IMAGE) \
 		docker/base
 
 create-debugimg:
-	docker build -t $(DEBUG_IMAGE) \
+	$(DOCKER_BUILD) -t $(DEBUG_IMAGE) \
 		--build-arg golang_image=$(GOLANG_IMAGE) \
 		docker/debug
+
+setup-buildx-builder:
+	bash ./scripts/setup-buildx-builder.sh
--- a/scripts/travis/build-all-in-one-image.sh
+++ b/scripts/travis/build-all-in-one-image.sh
@@ -52,8 +52,8 @@ repo=jaegertracing/all-in-one
 docker build -f cmd/all-in-one/Dockerfile \
         --target release \
         --tag $repo:latest cmd/all-in-one \
-        --build-arg base_image=localhost/baseimg:1.0.0-alpine-3.12 \
-        --build-arg debug_image=localhost/debugimg:1.0.0-golang-1.15-alpine \
+        --build-arg base_image=jaegertracing/baseimg:1.0.0-alpine-3.12 \
+        --build-arg debug_image=jaegertracing/debugimg:1.0.0-golang-1.15-alpine \
         --build-arg TARGETARCH=$GOARCH
 run_integration_test $repo
 upload_to_docker $repo
@@ -63,8 +63,8 @@ repo=jaegertracing/all-in-one-debug
 docker build -f cmd/all-in-one/Dockerfile \
         --target debug \
         --tag $repo:latest cmd/all-in-one \
-        --build-arg base_image=localhost/baseimg:1.0.0-alpine-3.12 \
-        --build-arg debug_image=localhost/debugimg:1.0.0-golang-1.15-alpine \
+        --build-arg base_image=jaegertracing/baseimg:1.0.0-alpine-3.12 \
+        --build-arg debug_image=jaegertracing/debugimg:1.0.0-golang-1.15-alpine \
         --build-arg TARGETARCH=$GOARCH
 run_integration_test $repo
 upload_to_docker $repo
--- a/cmd/agent/Dockerfile
+++ b/cmd/agent/Dockerfile
@@ -2,7 +2,7 @@ ARG base_image
 ARG debug_image
 
 FROM $base_image AS release
-ARG TARGETARCH=amd64
+ARG TARGETARCH
 ARG USER_UID=10001
 COPY agent-linux-$TARGETARCH /go/bin/agent-linux
 EXPOSE 5775/udp 6831/udp 6832/udp 5778/tcp
@@ -10,7 +10,7 @@ ENTRYPOINT ["/go/bin/agent-linux"]
 USER ${USER_UID}
 
 FROM $debug_image AS debug
-ARG TARGETARCH=amd64
+ARG TARGETARCH
 ARG USER_UID=10001
 COPY agent-debug-linux-$TARGETARCH /go/bin/agent-linux
 EXPOSE 5775/udp 6831/udp 6832/udp 5778/tcp 12345/tcp
--- a/cmd/all-in-one/Dockerfile
+++ b/cmd/all-in-one/Dockerfile
@@ -2,7 +2,7 @@ ARG base_image
 ARG debug_image
 
 FROM $base_image AS release
-ARG TARGETARCH=amd64
+ARG TARGETARCH
 
 # Agent zipkin.thrift compact
 EXPOSE 5775/udp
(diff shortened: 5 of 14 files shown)


jpkrohling:
@Ashmita152 would you like to take a look and review this PR?

Ashmita152:
Hi @jpkrohling sure I will try this PR today.

scripts/travis/setup-docker-buildx.sh
```diff
@@ -0,0 +1,61 @@
...
+
+# enabled qemu if needed
+if [[ ! $(docker buildx inspect default | grep Platforms) == *arm64* ]]; then
+  docker run --privileged --rm tonistiigi/binfmt --install all
+fi
+
+# run a local registry for base images push
+# have to do this for multi-arch images
+# multi arch images could not --load
+docker run -d -p 5000:5000 --name registry registry:2
```
Ashmita152, line 58: I am not sure but I'll prefer to push the base image to dockerhub under our organisation instead of starting our own docker-registry in every CI/CD run. That will remove all localhost:5000 too.

Makefile
```diff
@@ -68,6 +68,9 @@ COLOR_FAIL=$(shell printf "\033[31mFAIL\033[0m")
 COLORIZE=$(SED) ''/PASS/s//$(COLOR_PASS)/'' | $(SED) ''/FAIL/s//$(COLOR_FAIL)/''
 DOCKER_NAMESPACE?=jaegertracing
 DOCKER_TAG?=latest
+PLATFORMS?=linux/amd64,linux/arm64
```
Ashmita152, line 71: We are already defining `PLATFORMS` in `docker/Makefile` and we are including that Makefile here in line no. 5, so it should be available already.

Makefile
```diff
@@ -285,6 +288,12 @@ docker: build-binaries-linux docker-images-only
 build-binaries-linux:
 	GOOS=linux GOARCH=amd64 $(MAKE) build-platform-binaries
 
+.PHONY: build-binaries-linux-archs
+build-binaries-linux-archs:
```
Ashmita152, line 292: I see that we have added this new directive. Why didn't we update the existing `build-binaries-linux` directive instead since that is being called as dependencies everywhere ?

Makefile
```diff
@@ -375,7 +384,8 @@ docker-push:
 	if [ $$CONFIRM != "y" ] && [ $$CONFIRM != "Y" ]; then \
 		echo "Exiting." ; exit 1 ; \
 	fi
-	for component in agent cassandra-schema es-index-cleaner es-rollover collector query ingester example-hotrod tracegen; do \
+	DOCKERX_OUTPUT="--push --platform=$(PLATFORMS)" $(MAKE) docker-images-only
```
Ashmita152, line 387: I am very confused about the way` DOCKERX_OUTPUT` is being changed in the Makefile. In the beginning its being set as `DOCKERX_OUTPUT?=--load --platform=linux/amd64` and here we are changing it. What is the idea behind it ? Also I would rename it to something like `DOCKER_BUILDX_PARAMS` because it's more than just `--output` flag.

scripts/travis/setup-docker-buildx.sh
jpkrohling: Agree. I'm actually wondering if we need `docker buildx` at all, given that `buildah` can also generate multi-arch images: 

https://www.youtube.com/watch?v=SYJgkkjqd7s

From what I understand, we don't need qemu or a virtual machine to perform the build, given that we can produce our binaries for the target architectures anyway...

Ashmita152:
One thing I noticed that if we run `docker buildx install` in the beginning, then we don't need to update all lines from `docker build` to `docker buildx build`

```
$ docker buildx install
$ docker build --help | head -4

Usage:	docker buildx build [OPTIONS] PATH | URL | -

Start a build
```

Reference: https://docs.docker.com/buildx/working-with-buildx/#set-buildx-as-the-default-builder

Thought it would be interesting.

scripts/travis/setup-docker-buildx.sh
morlay: @Ashmita152 now in ci, no access to docker.io/jaegertracing for pr.

morlay: @jpkrohling need to research for this.

Without `qemu`, `RUN apk add ssh` may not be executable.

Makefile
morlay: `docker/Makefile` need self scope

`PLATFORMS` could overwrite for all.

Makefile
morlay: Now just for local build.
This pr should now change too lot for CI.

Makefile
morlay: as I said before. buildx will create a builder as backend - [buildkit](https://github.com/moby/buildkit),
which will be running in container.

`docker buildx build` send context to the buildkit and build image layers in container.

`docker buildx build --load` will export the built image layers back to host (for local test in ci)
`docker buildx build --push` will push the built image layers to remote registry like docker.io.

however `docker buildx build --load` only support only one platform `--platform=linux/amd64`

so i have to use those to switch local build or push to remote
`DOCKERX_OUTPUT=--load --platform=linux/amd64` 
`DOCKERX_OUTPUT=--push --platform=linux/amd64,linux/arm64`

morlay:
@Ashmita152 

This feature is for local switch for different projects.

flags of `docker buildx build` could not be used in `docker build`

```
# ensure buildx not as default build
docker buildx uninstall

docker buildx build --help
docker build --help
```

scripts/travis/setup-docker-buildx.sh
jpkrohling: Right, forgot about the `RUN apk add ca` command. Other than that, it would just work, right?

morlay: @jpkrohling Yes. if just COPY file or config static ENV, CMD ... it could work.

[bazel](https://bazel.build) could do this too. like https://github.com/GoogleContainerTools/distroless did.

jpkrohling: So, there could be a builder that runs before this and executes `RUN apk add ca`. The platform-specific build would then just copy those files? Sounds like a great solution to me...

morlay: It could work for base images. But may not for  debug images. 
make devel seems need build platform same sa target platform https://github.com/go-delve/delve/blob/master/_scripts/make.go

morlay:
@jpkrohling @Ashmita152 

is any plan about this?

if you don't want to switch buildx we could refactor Makefile like this.

https://github.com/kiali/kiali/blob/master/make/Makefile.container.mk#L54-L61

For local registry in CI, we could split the base images publishing in another job.

scripts/travis/setup-docker-buildx.sh
jpkrohling: Even if we don't move to scratch images, we should not start a container registry here (which is the original concern from @Ashmita152).

morlay: @jpkrohling this question is.

When we use `docker buildx build`, the backend for docker builds named `buildkit` need to pull images from a container registry.

As a multi-arch image, the tag `localhost:5000/baseimg:1.0.0-alpine-3.12` is ref to a manifest list of platform, not a manifest of image layers.
<details>
<summary>example</summary>
<pre><code>
$ DOCKER_CLI_EXPERIMENTAL=enabled docker manifest inspect morlay/baseimg:1.0.0-alpine-3.12
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
   "manifests": [
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 737,
         "digest": "sha256:6b22c5fd5b30591e4ce460ee722493e6151e5782804caff18bfee6a21685fbc4",
         "platform": {
            "architecture": "amd64",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 737,
         "digest": "sha256:601fe467e2993ecd47ea95b3eb7bb54701cca90da1cfac1af12f59a5b4f76b83",
         "platform": {
            "architecture": "arm64",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 737,
         "digest": "sha256:7c6eb953937fa24a2c4d0a34d70d5b9578cdbfd35e42f5e5afdd9fd5b55f7080",
         "platform": {
            "architecture": "s390x",
            "os": "linux"
         }
      }
   ]
}
</code></pre>
</detials>

it could not be storage in docker local. (means we could not use `docker images` to find this, `docker pull morlay/baseimg:1.0.0-alpine-3.12` only pull the image matched local arch)

**so we need push the baseimg with images for muti arch to some container registry**

as @Ashmita152 said, we could push baseimg to `docker.io/jaegertracing`.

But in CI for PR. there have no access to push to `docker.io/jaegertracing`.

This why I start a container registry here to storage images for muti arch

Ashmita152: Hi @morlay 

> But in CI for PR. there have no access to push to docker.io/jaegertracing.

Can we move the login to docker hub before to provide this access ?

morlay: it could. 
but be careful. 
it may cause a secure issue in PR

jpkrohling: >  it may cause a secure issue in PR

Why would it? We should have those as secrets in Travis and they are suppressed from the output in the CI.

> seems Docker username & password not provided for PR

If you set the appropriate env vars in the Travis job for your fork, it should work. You'd have to override the namespace, so that it won't attempt to push to `docker.io/jaegertracing/all-in-one`, but `quay.io/youraccount/all-in-one`.

morlay: > Why would it? We should have those as secrets in Travis and they are suppressed from the output in the CI.

But i could create some PR and just added some `docker push` to overwrite published images. it not set, the PR may not pass bacause of base image pushing failed.

So I prefer use a local registry
as the said https://github.com/docker/build-push-action#local-registry
 for testing purposes, like PR testings.

morlay: @jpkrohling 

Or could we create another namespace `docker.io/jaegertracing-dev` or `quay.io/jaegertracing-dev`


`docker.io/jaegertracing` only for master branch
`docker.io/jaegertracing-dev` for testing of PRs or other branches.

jpkrohling: You'd still have to override the namespace to `jaegertracing-dev`... So, at least for now, you can override to your Dockerhub/Quay namespace.

morlay: @jpkrohling

in my fork i could override to my Dockerhub/Quay namespace. 
but in PRs of `jaegertracing`, need namespace `jaegertracing-dev` to receive images from PRs.

Without a local registry. base images need to pushing to Dockerhub/Quay in PRs.

morlay:
@jpkrohling @Ashmita152 

I refactor again. 
Now we could keep `docker build` as default. CI flow will not change too much. 
The base images no need to pre publish. so no need local registry any more.

**only change is the final publishing**

```bash
# build all binaries for multi-arch
make build-binaries-linux-archs
```

```bash
# enable BUILDX_PUSH=true to switch buildx and build && push for multi-arch
DOCKER_NAMESPACE="jaegertracing" make dockerx-push
```

```
# for multi namespace could be
DOCKER_NAMESPACE="jaegertracing quay.io/jaegertracing" make dockerx-push
```
Need to updates `-t $(DOCKER_NAMESPACE)/*` to `$(foreach namespace,$(DOCKER_NAMESPACE),--tag=${namespace}/*)`

Makefile
```diff
@@ -379,6 +394,16 @@ docker-push:
 		docker push $(DOCKER_NAMESPACE)/jaeger-$$component ; \
 	done
 
+.PHONY: dockerx-push
+dockerx-push: .ensure-docker-buildx .ensure-buildx-builder
+	@while [ -z "$$CONFIRM" ]; do \
```
jpkrohling, line 399: Would this work in the CI as well? What's the worst that can happen if I call `make dockerx-push` but don't really mean to? I can just cancel it, can't I?

docker/Makefile
```diff
@@ -3,18 +3,48 @@ ROOT_IMAGE ?= alpine:3.12
...
 		docker/base
 
 create-debugimg:
-	docker build -t $(DEBUG_IMAGE) \
+	$(DOCKER_BUILD) -t $(DEBUG_IMAGE) \
 		--build-arg golang_image=$(GOLANG_IMAGE) \
 		docker/debug
+
+# Ensure "docker buildx" is available and enabled. For more details, see: https://github.com/docker/buildx/blob/master/README.md
+.ensure-docker-buildx:
```
jpkrohling, line 26: Why the dot in the name? Also, a beter name would probably be `install-docker-buildx`. To me, `ensure` would convey that it would fail in case the plugin isn't installed.

Makefile
morlay: Just copy from `make docker-push`. 
we could remove it

docker/Makefile
morlay: ensure means install if need.

morlay:
Updated names and removed unnecessary codes.

@jpkrohling How do you think the new workflow? only build multi-arch image in final publishing job.

morlay:
@jpkrohling ?

jpkrohling:
Sorry, was out for a few days. I'll review this one again and will let you know in the couple of days.

jpkrohling:
I haven't forgotten about this, and I'll review it shortly. Sorry for the delay!

jpkrohling (commented):
While I like these changes, I think they bring bigger question for the project itself. I tried to record the questions to discuss during today's bi-weekly call, but it's a holiday in the US, so, we'll probably discuss this in two weeks time:

https://docs.google.com/document/d/1ZuBAwTJvQN7xkWVvEFXj5WU9_JmS5TPiNbxCJSvPqX0#heading=h.reli0pq8oegb

By the way: we don't necessarily need to wait two weeks to discuss this :-)

.travis.yml
```diff
@@ -59,6 +59,7 @@ env:
 install:
   - docker rmi $(docker images -q) || true
   - travis_retry make install-ci
+  - bash ./scripts/travis/upgrade-docker.sh
```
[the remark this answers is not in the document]
Ashmita152: Hi @morlay 

We started moving to Github Actions due to travis credits expiration. As you're already aware of official docker's buildx action: https://github.com/docker/setup-buildx-action I am pretty sure it will make the PR lot simpler. Sorry for the trouble caused by this.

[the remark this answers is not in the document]
morlay: mv to github actions will be simple 🎉
as i setup for jaeger operator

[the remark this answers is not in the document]
morlay: @Ashmita152 I removed codes to setup buildx in travis.

consider setup two release workflows. 

one for base images (could set  `on.push.paths: docker/**/Dockerfile`)

```
make dockerx-baseimg-debugimg
```

another for components.
```
make dockerx-push
```

docker/Makefile
```diff
@@ -3,18 +3,77 @@ ROOT_IMAGE ?= alpine:3.12
...
 		--build-arg cert_image=$(CERT_IMAGE) \
 		docker/base
 
 create-debugimg:
-	docker build -t $(DEBUG_IMAGE) \
+	$(DOCKER_BUILD) -t $(DEBUG_IMAGE) \
 		--build-arg golang_image=$(GOLANG_IMAGE) \
 		docker/debug
+
+# Ensure "docker buildx" is available and enabled. For more details, see: https://github.com/docker/buildx/blob/master/README.md
```
morlay, line 25: copy from https://github.com/kiali/kiali/blob/master/make/Makefile.container.mk#L66-L120 

to make docker buildx could run in ci and local both

.travis.yml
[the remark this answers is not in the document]
Ashmita152: Sure I will do it today.

morlay:
@jpkrohling  now this PR scope changes. only add multi-arch image build for Makefile, and no effects for CIs.

and for questions of google docs. my options:

> Do we want multi arch builds as part of the CI, or only during the release process? 

I think only need to during release process. 

may add a binary check only.  

```
$ file cmd/all-in-one/all-in-one-*
cmd/all-in-one/all-in-one-debug-linux-amd64: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, Go BuildID=ld-A3dPBemVj8KhA6D66/P79Nik2qzz-PUm4_SV6i/faN0475xC2iACbcw4qJd/wESdIHq_6iNuojLMvYuW, not stripped
cmd/all-in-one/all-in-one-debug-linux-arm64: ELF 64-bit LSB executable, ARM aarch64, version 1 (SYSV), statically linked, Go BuildID=4hIwra70fzu8NIMkVc1I/2YkWeaxzHY_-ub2FxFSJ/p9sj0ploAWA8VE5f5hCr/5B9t_659c9_hlp6ZApkn, not stripped
cmd/all-in-one/all-in-one-debug-linux-s390x: ELF 64-bit MSB executable, IBM S/390, version 1 (SYSV), statically linked, Go BuildID=tU-unRCgLcYaz0W36czK/9jJEGBQrsSPDQOYDi3KI/1sQxnw-3aM_BeMLB85uv/2LiZo0AlmoVeOgUkzDbj, not stripped
cmd/all-in-one/all-in-one-linux-amd64:       ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, Go BuildID=jhGgmFXpIlew_u0cg4JK/Z8sX0vyvC0uNXHom8_P9/IdYXLkVSIpgKbYx0WSWu/yMJgn4BfZCdny_rE89e_, not stripped
cmd/all-in-one/all-in-one-linux-arm64:       ELF 64-bit LSB executable, ARM aarch64, version 1 (SYSV), statically linked, Go BuildID=N_2khwD_lRu48H1n2clU/YzXGbzs8QUyIEUuFj7iM/Jo-_zC1HRxIU4QrFqo10/rGpphqYpmiwd2v2r80yi, not stripped
cmd/all-in-one/all-in-one-linux-s390x:       ELF 64-bit MSB executable, IBM S/390, version 1 (SYSV), statically linked, Go BuildID=CEhwmulJWqGBsL8vJNVf/fZGJLDHkSewk42D1pqlr/0Z_ZA814Ylqss3SDn92Y/EZF1DggtOJd-oIu2bM5Z, not stripped
```


> On one hand, our CI build time is already high. On the other hand, not sure we want to release binaries that we are not actually testing (although, I think we are doing that already?)

not neccessary to full test for all archs, amd64 only enough. 

if we could run `GOOS=linux GOARCH=arm64 go build` successfully, we could execute biniary on arm64 env.
so testings for other arch could let open source community to do. 

In fact. with golang, we could be easy to develop arch-less application. 

> If we want to change the release process, should we think about moving to something like goreleaser?

goreleaser just for binaries. nothing need to change for building multi-arch image

docker/Makefile
```diff
@@ -3,18 +3,77 @@ ROOT_IMAGE ?= alpine:3.12
...
+# This does a few things:
+#  1. Makes sure docker is in PATH
+#  2. Downloads and installs buildx if no version of buildx is installed yet
+#  3. Makes sure any installed buildx is a required version or newer
+#  4. Makes sure the user has enabled buildx (either by default or by setting DOCKER_CLI_EXPERIMENTAL env var to 'enabled')
+#  Thus, this target will only ever succeed if a required (or newer) version of 'docker buildx' is available and enabled.
+install-docker-buildx-if-need:
+	@if ! which docker > /dev/null 2>&1; then echo "'docker' is not in your PATH."; exit 1; fi
```
jpkrohling, line 33: This should really be in an external shell script. In fact, I think it was there before, wasn't it?

docker/Makefile
```diff
@@ -3,18 +3,77 @@ ROOT_IMAGE ?= alpine:3.12
...
+	if docker buildx version > /dev/null 2>&1; then \
+	  echo "'docker buildx' is enabled"; \
+	else \
+	  echo "'docker buildx' is not enabled. Set DOCKER_CLI_EXPERIMENTAL=enabled if you want to use it."; \
+	  exit 1; \
+	fi
+
+# Ensure a local builder for multi-arch build. For more details, see: https://github.com/docker/buildx/blob/master/README.md#building-multi-platform-images
+setup-buildx-builder: install-docker-buildx-if-need
+	@if ! docker buildx inspect jaeger-builder > /dev/null 2>&1; then \
```
jpkrohling, line 67: Same as this.

docker/Makefile
```diff
@@ -3,18 +3,77 @@ ROOT_IMAGE ?= alpine:3.12
 CERT_IMAGE := alpine:3.12
 GOLANG_IMAGE := golang:1.15-alpine
 
-BASE_IMAGE := localhost/baseimg:$(VERSION)-$(shell echo $(ROOT_IMAGE) | tr : -)
-DEBUG_IMAGE := localhost/debugimg:$(VERSION)-$(shell echo $(GOLANG_IMAGE) | tr : -)
+BASE_IMAGE := $(DOCKER_NAMESPACE)/baseimg:$(VERSION)-$(shell echo $(ROOT_IMAGE) | tr : -)
```
jpkrohling, line 6: As a good practice, define this var in the beginning of the Makefile, defaulting to the env var:

```Makefile
DOCKER_NAMESPACE?=jaegertracing
```

Same with the other one (`DOCKER_BUILD`).

jpkrohling (commented):
This is looking good. Just need to externalize a few things into scripts. A review by @Ashmita152 would also be nice to have.

Ashmita152:
Looks good overall to me. I have just one question:

- Are `install-docker-buildx-if-need` and `setup-buildx-builder` makefile directives written for installing `docker buildx` locally or for something else ?

morlay:
mainly for local if developer haven't install   buildx and setup a builder named jaeger-builder.

this could be used in CI too.

many good suggestions from kiali maintainers, could check
https://github.com/kiali/kiali/pull/3405

docker/Makefile
morlay: this is a sub Makefile. 
should we need to re-define those twice?

Makefile
```diff
@@ -390,6 +405,10 @@ docker-push:
 		docker push $(DOCKER_NAMESPACE)/jaeger-$$component ; \
 	done
 
+.PHONY: dockerx-push
+dockerx-push: setup-buildx-builder
+	BUILDX_PUSH=true $(MAKE) docker-images-only
```
yurishkuro, line 410: Var before command becomes env var, after command it becomes make var. Do you want `$(MAKE) docker-images-only BUILDX_PUSH=true`?

Makefile
```diff
@@ -69,6 +67,18 @@ COLORIZE=$(SED) ''/PASS/s//$(COLOR_PASS)/'' | $(SED) ''/FAIL/s//$(COLOR_FAIL)/''
 DOCKER_NAMESPACE?=jaegertracing
 DOCKER_TAG?=latest
 
+# list for multi-arch image publishing
+TARGET_ARCHS ?= amd64 arm64
+
+BUILDX_PUSH ?= false
+DOCKER_BUILD ?= docker build --build-arg=TARGETARCH=$(GOARCH)
```
yurishkuro, line 74: TARGETARCH is defined here but not in L76. How does it work, especially if the default value is removed from all Dockerfiles?

Makefile
```diff
@@ -69,6 +67,18 @@ COLORIZE=$(SED) ''/PASS/s//$(COLOR_PASS)/'' | $(SED) ''/FAIL/s//$(COLOR_FAIL)/''
...
+ifeq ($(BUILDX_PUSH),true)
+	DOCKER_BUILD = DOCKER_CLI_EXPERIMENTAL="enabled" docker buildx build --builder=jaeger-builder --push $(foreach arch,${TARGET_ARCHS},--platform=linux/${arch})
+endif
+
+# make sure variables could pass to includes
```
yurishkuro, line 79: what is the meaning of this comment?

Makefile
yurishkuro: Also, wouldn't this line be immediately evaluated with the current value of $(GOARCH) ?

morlay: TARGETARCH is a platform args, which will be auto pass to Dockerfile when we use buildx.
https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope

nope.  when we use `=` instead of `:=` in makefile, `$(GOARCH)` is as a ref value. will be evaluated  when `$(DOCKER_BUILD)` used.

Makefile
morlay: env var could overwrites make var `BUILDX_PUSH` when define with `?=`.
will same as `$(MAKE) docker-images-only BUILDX_PUSH=true`.

i will update this for style.

Makefile
morlay: make sure sub makefiles could use the make variables defined in root makefile.

yurishkuro:
are these warnings expected? https://github.com/jaegertracing/jaeger/pull/2611/checks?check_run_id=1483365080#step:11:375
image

yurishkuro:
This doesn't look good either, pulling all-in-one while trying to build it (cc @Ashmita152 )
image

Makefile
```diff
@@ -390,6 +405,10 @@ docker-push:
 		docker push $(DOCKER_NAMESPACE)/jaeger-$$component ; \
 	done
 
+.PHONY: dockerx-push
+dockerx-push: setup-buildx-builder
```
yurishkuro, line 409: Is this a WIP?  The all-in-one build is only doing amd64, I assume because BUILDX_PUSH=false by default. Wouldn't we want x-platform build to be the default, rather than having a separate target?

morlay: `all-in-one` is not in `make docker-push`, so i haven't add it in `make dockerx-push`

morlay:
@yurishkuro 

> are these warnings expected? https://github.com/jaegertracing/jaeger/pull/2611/checks?check_run_id=1483365080#step:11:375

just because there is no `ARG TARGETARCH` in `docker/base/Dockerfile` && `docker/debug/Dockerfile`.
this will not break anything

scripts/travis/build-all-in-one-image.sh
```diff
@@ -63,8 +63,8 @@ repo=jaegertracing/all-in-one-debug
 docker build -f cmd/all-in-one/Dockerfile \
         --target debug \
         --tag $repo:latest cmd/all-in-one \
-        --build-arg base_image=localhost/baseimg:1.0.0-alpine-3.12 \
-        --build-arg debug_image=localhost/debugimg:1.0.0-golang-1.15-alpine \
```
morlay, line 67: fix https://github.com/jaegertracing/jaeger/pull/2611#issuecomment-736921365

morlay:
no idea why es testing failed.

jpkrohling:
Sharing something I've read this weekend: https://carlosbecker.com/posts/multi-platform-docker-images-goreleaser-gh-actions/
Note how they use `docker manifest` instead of `buildx`: the binaries are still build outside of Docker, via regular Go builds.

morlay:
@jpkrohling 

with `docker manifest` should publish arch suffix images `-arm64` `-amd64` first.

it will be more complex for publishing workflow

after migrated github actions, setup buildx will be very easy

morlay:
waiting for a little long. I have to setup workflow to release most versioned jaeger images

with some hacks https://github.com/querycap/jaeger

GaruGaru:
Any update on this ? 
The changes looks similar to what is proposed on https://github.com/jaegertracing/jaeger/pull/2948

yurishkuro:
Yes, I think we should close this in favor of #2948, which is about to be merged.

PR #2617, opened 2020-11-06
Bump opentelemetry-collector to v0.14.0

Vemmy124:
Adds compatibility with new version of collector

Just refactoring mostly. There is only one open question: should retry queue be enabled by default?
Collector defaults are to enable it, but according to tests it should be disabled


cmd/opentelemetry/app/exporter/elasticsearchexporter/esmodeltranslator/modeltranslator_test.go
```diff
@@ -274,7 +269,18 @@ func addSpan(traces pdata.Traces, name string, traceID pdata.TraceID, spanID pda
 	span.SetName(name)
 	span.SetTraceID(traceID)
 	span.SetSpanID(spanID)
+	span.SetParentSpanID(spanID)
 	span.SetStartTime(pdata.TimestampUnixNano(time.Now().UnixNano()))
 	span.SetEndTime(pdata.TimestampUnixNano(time.Now().UnixNano()))
 	return span
 }
+
+func Byte16ArrayFromString(s string) (result [16]byte) {
```
Vemmy124, line 278: Think I should move these functions not to duplicate code. Would be great if you could help to find their place

jpkrohling: How many callers are there? If there are only two, change them to use something like `pdata.NewTraceID([16]byte{1, 2, 3, 4}` instead of a string. I don't think being a string is part of the test.

jpkrohling (approved):
LGTM

cmd/opentelemetry/app/exporter/badgerexporter/factory.go
```diff
@@ -91,17 +91,17 @@ func (f Factory) CreateDefaultConfig() configmodels.Exporter {
 
 // CreateTraceExporter creates Jaeger Cassandra trace exporter.
 // This function implements OTEL component.ExporterFactory interface.
-func (f Factory) CreateTraceExporter(
+func (f Factory) CreateTracesExporter(
```
yurishkuro, line 94: hehe, what happens when so many people contributing are not native speakers. "trace exporter" was grammatically correct name.

yurishkuro: not a jab at you, @Vemmy124 , but on the upstream change

jpkrohling: I confess it did sound strange to my ears, but I'm not a native speaker myself, so, thought it was a problem with my ears :-)

cmd/opentelemetry/app/exporter/elasticsearchexporter/esmodeltranslator/modeltranslator.go
```diff
@@ -222,21 +222,15 @@ func references(links pdata.SpanLinkSlice, parentSpanID pdata.SpanID, traceID db
...
 func convertSpanID(spanID pdata.SpanID) (dbmodel.SpanID, error) {
-	spanIDInt, err := tracetranslator.BytesToUInt64SpanID(spanID.Bytes())
-	if err != nil {
-		return "", err
-	}
+	spanIDInt := tracetranslator.BytesToUInt64SpanID(spanID.Bytes())
 	if spanIDInt == 0 {
 		return "", errZeroSpanID
 	}
 	return dbmodel.SpanID(fmt.Sprintf("%016x", spanIDInt)), nil
```
yurishkuro, line 229: This probably would be more efficient:
```go
import "encoding/hex"

bytes := spanID.Bytes()
dst := make([]byte, hex.EncodedLen(len(src)))
hex.Encode(dst, src)
return dbmodel.SpanID(dst)
```

cmd/opentelemetry/app/exporter/elasticsearchexporter/esmodeltranslator/modeltranslator.go
```diff
@@ -222,21 +222,15 @@ func references(links pdata.SpanLinkSlice, parentSpanID pdata.SpanID, traceID db
 }
 
 func convertSpanID(spanID pdata.SpanID) (dbmodel.SpanID, error) {
-	spanIDInt, err := tracetranslator.BytesToUInt64SpanID(spanID.Bytes())
-	if err != nil {
-		return "", err
-	}
+	spanIDInt := tracetranslator.BytesToUInt64SpanID(spanID.Bytes())
 	if spanIDInt == 0 {
 		return "", errZeroSpanID
```
yurishkuro, line 227: This feels misplaced to me. This kind of validation should've been done way earlier in the pipeline. The ES backend itself doesn't actually care if IDs are all zeroes, it would still work.

cmd/opentelemetry/app/exporter/elasticsearchexporter/esmodeltranslator/modeltranslator_test.go
```diff
@@ -215,26 +221,16 @@ func TestEmpty(t *testing.T) {
...
+	var zero32Bytes [8]byte
 	tests := []struct {
 		spanID  pdata.SpanID
 		traceID pdata.TraceID
 		err     string
 	}{
-		{
-			traceID: pdata.NewTraceID([]byte("invalid-%")),
-			err:     "TraceID does not have 16 bytes",
-		},
```
yurishkuro, line 229: not sure why these tests are removed - give the current code there would still be some kind of error returned from converters.

yurishkuro (dismissed):
lgtm, but why are some tests removed?

cmd/opentelemetry/app/exporter/badgerexporter/factory.go
Vemmy124: I believe that was bogdandrutu's idea https://github.com/open-telemetry/opentelemetry-collector/issues/1974

cmd/opentelemetry/app/exporter/elasticsearchexporter/esmodeltranslator/modeltranslator_test.go
Vemmy124: But TraceID is now exactly `[16]byte`. Why would the error be there even if only several first bytes are set?

Vemmy124: They are silently truncated according to this PR conversations https://github.com/open-telemetry/opentelemetry-collector/pull/2001

cmd/opentelemetry/app/exporter/badgerexporter/factory.go
Vemmy124: And maybe that's really the problem of being not native speaker, but I agree with him, because the component consumes the structure named pdata.Traces

cmd/opentelemetry/app/exporter/elasticsearchexporter/esmodeltranslator/modeltranslator.go
Vemmy124: You're definitely right, but practically having this error handled here can prevent some time of unnecessary debugging in case of zero IDs and missing validation earlier. I believe this error arises systematically, so anyway there will be collisions by ID and this will only implicitly indicate zero ID error.

Vemmy124:
Seems like a problem with build. Could you take a look please, since you're acquainted with how tests are run?

yurishkuro:
```
Lint Failures
cmd/opentelemetry/app/exporter/badgerexporter/factory.go:92:1: comment on exported method Factory.CreateTracesExporter should be of the form "CreateTracesExporter ..."
cmd/opentelemetry/app/exporter/cassandraexporter/factory.go:74:1: comment on exported method Factory.CreateTracesExporter should be of the form "CreateTracesExporter ..."
cmd/opentelemetry/app/exporter/elasticsearchexporter/factory.go:75:1: comment on exported method Factory.CreateTracesExporter should be of the form "CreateTracesExporter ..."
cmd/opentelemetry/app/exporter/grpcpluginexporter/factory.go:68:1: comment on exported method Factory.CreateTracesExporter should be of the form "CreateTracesExporter ..."
cmd/opentelemetry/app/exporter/jaegerexporter/jaeger_exporter.go:61:1: comment on exported method Factory.CreateTracesExporter should be of the form "CreateTracesExporter ..."
cmd/opentelemetry/app/exporter/kafkaexporter/kafka_exporter.go:100:1: comment on exported method Factory.CreateTracesExporter should be of the form "CreateTracesExporter ..."
cmd/opentelemetry/app/exporter/memoryexporter/factory.go:81:1: comment on exported method Factory.CreateTracesExporter should be of the form "CreateTracesExporter ..."
cmd/opentelemetry/app/processor/resourceprocessor/resource_processor.go:70:1: comment on exported method Factory.CreateTracesProcessor should be of the form "CreateTracesProcessor ..."
cmd/opentelemetry/app/receiver/jaegerreceiver/jaeger_receiver.go:150:1: comment on exported method Factory.CreateTracesReceiver should be of the form "CreateTracesReceiver ..."
cmd/opentelemetry/app/receiver/kafkareceiver/kafka_receiver.go:107:1: comment on exported method Factory.CreateTracesReceiver should be of the form "CreateTracesReceiver ..."
cmd/opentelemetry/app/receiver/zipkinreceiver/zipkin_receiver.go:58:1: comment on exported method Factory.CreateTracesReceiver should be of the form "CreateTracesReceiver ..."
Makefile:198: recipe for target 'go-lint' failed
```

Vemmy124:
I don't understand, how this piece of code passed lint before then. Anyways, let's add comments

Vemmy124:
I wasn't right about the tests correction, fixed now. What is the reason of absence of golangci.yml in the main repo? Wouldn't it be useful for local checks?

Vemmy124:
@yurishkuro @jpkrohling Can we merge now?

jpkrohling:
There's one pending change requested by Yuri: https://github.com/jaegertracing/jaeger/pull/2617#discussion_r518997100

About the linting configuration: I think it would be a welcome change.

cmd/opentelemetry/app/exporter/elasticsearchexporter/esmodeltranslator/modeltranslator.go
Vemmy124: According to benchmark run on my device, this change gave 70ns vs previous 170ns

Vemmy124:
@yurishkuro was right - new method for span ID conversion is a lot faster. Now I believe everything is resolved (well, idk if we reached consensus about spanID validation)

cmd/opentelemetry/app/exporter/badgerexporter/factory.go
yurishkuro: It doesn't matter what exact data type is passed, a component that exports traces (plural) is "trace exporter" (noun used as adjective in singular form).

yurishkuro (approved):

yurishkuro:
Restarted build. I have not seen crossdock test failing too often recently, and here it failed several times, may be related to this change. But could be flaky too, let's see how the build comes back.

Vemmy124:
FWIW in my build on real data some traces are lost somewhere between elasticexporter and elastic cluster. The worst thing is that there are no logs about loss in opentelemetry-collector component until v0.14.0. This might be closely related actually and I will investigate that once we merge this PR. 

Defending my changes in commits, I have to notice, that tests passed with my previous commit, which (at least should) have equivalent semantics with the latest version.

yurishkuro:
Pretty sure the crossdock_otel CI failure is related to this change - it consistently failed 3 times, while passing on master and other PRs.

Vemmy124:
Found the problem. There was no test on the case when only 8 bytes of trace id is generated. Corrected tests to check that as well

yurishkuro (approved):

yurishkuro:
Thanks!

Go phystech.edu :-)

Issue #2618, opened 2020-11-06
Provide multiple datacenters to Cassandra Schema generation scripts


Kerptastic:
## Requirement - what kind of business use case are you trying to solve?
We have a multi-datacenter configuration. Currently, when using the `jaeger-cassandra-schema` container, the keyspace being generated can only be configured using 1) a single datacenter; 2) a hard-coded replication factor of 2. 

Example of current: `{'class': 'NetworkTopologyStrategy', 'dc1': '2' }`

This results in the following error:

```
 panic: first replica is not the primary replica for the token: expected 10.120.42.3 got 10.121.4.7
```

Our Cassandra DBEs propose that the NetworkTopologyStrategy have a minimum replication factor of 3, and that each datacenter be configured during keyspace creation.

Example of requested: `{'class': 'NetworkTopologyStrategy', 'dc1': '3', 'dc2': '3' }`

## Problem - what in Jaeger blocks you from solving the requirement?
We currently use the Jaeger Operator and the CassandraSchema container to prep and manage our Jaeger deployment. Without this change, we need to fork the repo and build our own CassandraSchema container in order to successfully deploy to a multi-dc Cassandra. This would also require us to internally maintain a fork in order to recreate the containers, if need be.

## Proposal - what do you suggest to solve the problem or improve the existing situation?
A suggestion would be to edit the `plugin/storage/cassandra/schema/create|docker.sh` to allow for comma-delimited DATACENTER value to be provided, and created the appropriate `replication_factor` value when multiple datacenters are indeed provided.

I have a PR for this ready, and tested against our Cassandra multi-dc configuration, and it works as expected.

jpkrohling:
Sounds sane to me, but would be good to have an opinion on someone else using Cassandra as well. @joe-elliott, are you using it?

joe-elliott:
We do use Cassandra, but we do not have a multi-dc setup.  The script already supports custom replication factor:

`https://github.com/jaegertracing/jaeger/blob/master/plugin/storage/cassandra/schema/create.sh#L14`

It seems like we just need support for the comma delimited datacenters?  I'd be glad to test the changes on our setup once the PR exists.

Kerptastic:
[...]

PR #2623, opened 2020-11-09
Adding Multi-DC Cassandra Support to Bootstrap Configuration Script(s)

Kerptastic:
- Solves: https://github.com/jaegertracing/jaeger/issues/2618

TL;DR from the issue:
- Multi-DC Cassandra setups are not supported by the jaeger-cassandra bootstrap scripts (i.e., can only provide a single Datacenter value, versus `N` Datacenter values, where `N > 1`)
- Even numbered quorum value is being used for `prod` setups, versus using an odd number

Adding the capability to provide a comma-delimited list of data-centers from the jaeger-operator, through `docker.sh`, and into `create.sh` to allow for multi-DC Cassandra setups.
- Added internal field separator to `create.sh`
- Set default replication factor for prod setups to be 3, and provide minimum odd-number quorum
- Create replication factor string for 1+ Datacenter values when in prod mode
- Updated `echo` statements accordingly

### DIFF
--- a/plugin/storage/cassandra/schema/create.sh
+++ b/plugin/storage/cassandra/schema/create.sh
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+export IFS=","
 
 function usage {
     >&2 echo "Error: $1"
@@ -7,7 +8,7 @@ function usage {
     >&2 echo ""
     >&2 echo "The following parameters can be set via environment:"
     >&2 echo "  MODE               - prod or test. Test keyspace is usable on a single node cluster (no replication)"
-    >&2 echo "  DATACENTER         - datacenter name for network topology used in prod (optional in MODE=test)"
+    >&2 echo "  DATACENTER         - comma seperated list of datacenter names for network topology used in prod (optional in MODE=test)"
     >&2 echo "  TRACE_TTL          - time to live for trace data, in seconds (default: 172800, 2 days)"
     >&2 echo "  DEPENDENCIES_TTL   - time to live for dependencies data, in seconds (default: 0, no TTL)"
     >&2 echo "  KEYSPACE           - keyspace (default: jaeger_v1_{datacenter})"
@@ -30,9 +31,19 @@ if [[ "$MODE" == "" ]]; then
     usage "missing MODE parameter"
 elif [[ "$MODE" == "prod" ]]; then
     if [[ "$DATACENTER" == "" ]]; then usage "missing DATACENTER parameter for prod mode"; fi
+    replication_factor=${REPLICATION_FACTOR:-3}
     datacenter=$DATACENTER
-    replication_factor=${REPLICATION_FACTOR:-2}
-    replication="{'class': 'NetworkTopologyStrategy', '$datacenter': '${replication_factor}' }"
+    topology_config=""
+    count=0
+    for dc in $DATACENTER; do
+      if [[ "$count" -eq "0" ]]; then
+        topology_config="'$dc': '${replication_factor}'"
+      else
+        topology_config="$topology_config, '$dc': '${replication_factor}'"
+      fi
+      count=$((count+1))
+    done
+    replication="{'class': 'NetworkTopologyStrategy', ${topology_config}}"
 elif [[ "$MODE" == "test" ]]; then 
     datacenter=${DATACENTER:-'test'}
     replication_factor=${REPLICATION_FACTOR:-1}
--- a/plugin/storage/cassandra/schema/docker.sh
+++ b/plugin/storage/cassandra/schema/docker.sh
@@ -36,7 +36,7 @@ do
   fi
 done
 
-echo "Generating the schema for the keyspace ${KEYSPACE} and datacenter ${DATACENTER}"
+echo "Generating the schema for the keyspace ${KEYSPACE} and datacenter(s) ${DATACENTER}"
 
 
 if [ -z "$PASSWORD" ]; then


plugin/storage/cassandra/schema/create.sh
```diff
@@ -30,9 +31,19 @@ if [[ "$MODE" == "" ]]; then
     usage "missing MODE parameter"
 elif [[ "$MODE" == "prod" ]]; then
     if [[ "$DATACENTER" == "" ]]; then usage "missing DATACENTER parameter for prod mode"; fi
+    replication_factor=${REPLICATION_FACTOR:-3}
```
joe-elliott, line 34: You should be able to just set the env var `REPLICATION_FACTOR=3` on the container running this script.  I don't think we should change the default value here.

Kerptastic: This is sort of the can of worms I was poking at in the issue, and understand the hesitancy to change to 3. I am not a Cassandra expert by any stretch of imagination, but was informed internally that for quorum, 3 should be the default vs. 2.

Also, if you are running `create.sh` manually or within self baked images, that would work fine, but the issue is that the `docker.sh` (lines 42-46) is where the script is invoked, which is provided by the `jaeger-cassandra-schema` image. From what I understand, folks using the Jaeger Operator Helm Chart(like ourselves), we do not have a way to push in the replication factor via `values.yaml`.

For Example:
```
...
        cassandraCreateSchema:
          enabled: true
          datacenter: us-west2,europe-west4
          mode: "prod"
          image: my-custom-images/jaeger-cassandra-schema-bootstrap:latest
...
```

Or am I overthinking this?

Reference link to `docker.sh`:
https://github.com/jaegertracing/jaeger/blob/44a6d37edeeabb669dc1df61a2099bed09c1bb9a/plugin/storage/cassandra/schema/docker.sh#L45

yurishkuro: There is no "right" value for replication factor, there are pros and cons for different settings. I agree that for backwards compatibility it's best not to change the current default.

Regarding the config schema above, it sounds like it needs a change in the operator.

plugin/storage/cassandra/schema/create.sh
```diff
@@ -30,9 +31,19 @@ if [[ "$MODE" == "" ]]; then
...
+    count=0
+    for dc in $DATACENTER; do
+      if [[ "$count" -eq "0" ]]; then
+        topology_config="'$dc': '${replication_factor}'"
+      else
+        topology_config="$topology_config, '$dc': '${replication_factor}'"
+      fi
+      count=$((count+1))
+    done
+    replication="{'class': 'NetworkTopologyStrategy', ${topology_config}}"
```
joe-elliott, line 46: For more complex multi dc replication strategies does it make sense to just allow the operator to override the entire replication string using an environment variable? 

If we add this then eventually someone will want datacenters with different RF values.

plugin/storage/cassandra/schema/create.sh
Kerptastic: @yurishkuro I agree on both fronts. I think making the config more flexible at the operator level is ultimately the right solution.

plugin/storage/cassandra/schema/create.sh
Kerptastic: @joe-elliott @yurishkuro I agree. I think having the entire string would be simple from an implementation strategy, but possibly a bit harder to debug errors for users who are have little Cassandra experience, or deploying at the operator. I think there are multiple options here:

1. Provide the entire replication config - this would be the easiest to implement, but would be pretty awful to put into a yaml file for the operator config: 
```
{ 'class': 'NetworkTopologyStrategy', 'dc1': '1', 'dc2': '2' ... }
```
2. Provide just the data-center portion, and still use the `mode` setting - this is less offensive to a user of the operator, and pretty simple to validate/implement:
```
'dc1': '1', 'dc2': '2', ...
```
3. Provide a parseable format, such as DC1:RF1,DC2:RF2 - probably the simplest on all fronts, development and using it in yaml, but it would require informing the users specifically what it is:
```
dc1:1,dc2:2,...
```
I would suggest option 3, but I come from the inexperienced with Cassandra camp, and don't know if this would be confusing and hard to understand. I don't fully believe so. Another negative, though, would be losing backwards compatibility with previous versions of the operator, as I would expect a `replicationFactorConfig` value of sorts to be introduced. Obviously there are ways to provide both and bark at the user if both are provided, but I will leave that up for discussion.

Let me know what y'all think!

yurishkuro: I like the idea of just providing a full replacement string for replication config, instead of building complicated piecemeal support.

Kerptastic: @yurishkuro @joe-elliott Alright! Just want to be clear, that preference is for Option 1 above? I can update this PR and start getting to work on the Operator changes. I may need some help on that side, as my golang is pretty elementary, but I'm sure I can get things started.

yurishkuro: Yes, let's do option 1.

Kerptastic: OK - my goal will be to leave the `$DATACENTER` usage as previously provided, and add support for `$REPLICATION_FACTOR_CONFIG`, that when provided, will override the creation of the `replication_factor` with the provided value. Will also PR the Operator. Thanks for the input guys, will circle back over the next few days as I get time to work on this.

yurishkuro:
@Kerptastic I am going to close this, feel free to reopen or create another PR with the new changes.

LuChenjing:
Hi @Kerptastic @yurishkuro ,I met the same issue... May I ask when this PR will been merged?

Kerptastic:
> Hi @Kerptastic @yurishkuro ,I met the same issue... May I ask when this PR will been merged?

Haven't gotten around to finishing this up, and also needed to make a change to the Jaeger operator CRD to take the configuration into account.

LuChenjing:
> > Hi @Kerptastic @yurishkuro ,I met the same issue... May I ask when this PR will been merged?
> 
> Haven't gotten around to finishing this up, and also needed to make a change to the Jaeger operator CRD to take the configuration into account.
Okay，understand, thanks.

PR #2624, opened 2020-11-10
Fix flaky TestReload

### LINKED ISSUE #2622: Flaky test: `TestReload`
**Describe the bug**
https://travis-ci.com/github/jaegertracing/jaeger/jobs/428075939#L2104

```
=== RUN   TestReload
    cert_watcher_test.go:101: 
        	Error Trace:	cert_watcher_test.go:101
        	Error:      	Not equal: 
```

**To Reproduce**
1. Place the following code after [cert_watcher_test.go:L88](https://github.com/jaegertracing/jaeger/blob/master/pkg/config/tlscfg/cert_watcher_test.go#L88)
```
	cert, err = tls.LoadX509KeyPair(filepath.Clean(clientCert), clientKey)
	assert.Equal(t, &cert, watcher.certificate())
```
2. Run test.

**Expected behavior**
The test passes.

Note, the above reproduce step will always fail the test; the intention is to reproduce the symptoms and its purpose is to better illustrate the race condition. That is, that the assertion is made with only the client cert modification, without the private key modification.

 - OS: MacOS
 - Jaeger version: master

Print the logObserver contents each time the cert and key files are modified.

**Additional context**
The root cause is similar to #2586 in that there is a race condition between when the private key file is update is written, to when the derived certificate is updated.

Specifically, in the failing travis job, the certificate is initially created with the server public/private key pair. The certificate is then modified with the client's public/private key pair, with an assertion that the modified certificate should resemble a pre-generated client certificate from the same client public+private key pair.

The race condition here is that the modified cert isn't yet updated with the client's private key, even though the key write completed. The reason why the assertion was made too early is because the `waitUntil` only checks for the existence of the `certFile` log entry when it should also be checking for _both_ the `certFile` and `keyFile`.

**Proposed Fix**
Modify the [`waitUntil` block](https://github.com/jaegertracing/jaeger/blob/master/pkg/config/tlscfg/cert_watcher_test.go#L94) to:
```
	waitUntil(func() bool {
		return logObserver.FilterField(zap.String("certificate", certFile.Name())).Len() > 0 &&
			logObserver.FilterField(zap.String("certificate", keyFile.Name())).Len() > 0
	}, 100, time.Millisecond*200)
```

**Validation**
To prove this fix doesn't just pass due to timing out the `waitUntil` block (and hence slow down test execution time significantly), the test was run locally _without the fix_ which took 0.02s. _With the fix_, the execution time was 0.02-0.04s, which is well below the 20s timeout window.

albertteoh:

- Fixes #2622

Fixes the flaky `TestReload` test by:
- Firstly logging _after_ the modified certificate has finished reloading, as suggested by @chlunde; and use this log entry as our signal for cert file mod completion.
- Adds assertion on `keyFile` log entry, and make both assertions more specific along with diagnostic data on failure to help with debugging if the race condition presents itself again in future.

## Testing
Repeated the following test 3x with no errors.:
```
go test -test.count=5000 -test.run=TestReload$ ./pkg/config/tlscfg/
```

### DIFF
--- a/pkg/config/tlscfg/cert_watcher.go
+++ b/pkg/config/tlscfg/cert_watcher.go
@@ -112,7 +112,12 @@ func (w *certWatcher) watchChangesLoop(rootCAs, clientCAs *x509.CertPool) {
 				w.mu.Unlock()
 				err = e
 			}
-			if err != nil {
+			if err == nil {
+				w.logger.Info("Loaded modified certificate",
+					zap.String("certificate", event.Name),
+					zap.String("event", event.Op.String()))
+
+			} else {
 				w.logger.Error("Failed to load certificate",
 					zap.String("certificate", event.Name),
 					zap.String("event", event.Op.String()),
--- a/pkg/config/tlscfg/cert_watcher_test.go
+++ b/pkg/config/tlscfg/cert_watcher_test.go
@@ -17,6 +17,7 @@ package tlscfg
 import (
 	"crypto/tls"
 	"crypto/x509"
+	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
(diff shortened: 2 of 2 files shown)


yurishkuro (approved):
thanks!

PR #2626, opened 2020-11-11
Bump to latest UI for snapshop builds

yurishkuro:
Usually we don't do that and only bump UI for a release, but is there any harm in staying on top of ui/master for the snapshot builds?


jpkrohling (approved):
+1 to staying up to date with the UI. If anything, we'll catch integration problems earlier.

PR #2632, opened 2020-11-17
[anonymizer] Update Makefile and Dockerfile

Ashmita152:

I realised in v1.21.0 we didn't had the anonymizer binary and docker image due to missing Makefile targets. This PR adds those.

- Update Makefile for anonymizer specific targets.
- Add Dockerfile for anonymizer.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -213,6 +213,10 @@ build-examples:
 build-tracegen:
 	$(GOBUILD) -o ./cmd/tracegen/tracegen-$(GOOS)-$(GOARCH) ./cmd/tracegen/main.go
 
+.PHONY: build-anonymizer
+build-anonymizer:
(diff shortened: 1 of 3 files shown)


yurishkuro:
@jpkrohling do you think it's necessary to publish a Docker image for this?  I feel like since we already distribute binaries, and the use case for the anonymizer is pretty narrow, it may be sufficient to only publish the binary.

jpkrohling:
If we can, sure. It's useful for people who aren't on the target arch that we publish binary for, or people facing bugs on target arches that we don't test (Windows for instance).

yurishkuro:
OK, I created jaeger-anonymizer repo on Docker hub.

yurishkuro (approved):

PR #2634, opened 2020-11-19
Bump opentelemetry to v0.15.0

Vemmy124:
Update of dependency

The only real thing is that any `pdata.Traces` component is non-nullable now, so `IsNil()` and `InitEmpty()` became deprecated.

No changes applied to logic. The config for jaegerreceiver was updated in opentelemetry (initiated and implemented by me actually), but it is orthogonal to current jaeger code.


cmd/opentelemetry/app/exporter/elasticsearchexporter/esmodeltranslator/modeltranslator.go
```diff
@@ -248,7 +248,7 @@ func traceIDToString(high, low uint64) string {
 }
 
 func (c *Translator) process(resource pdata.Resource) *dbmodel.Process {
-	if resource.IsNil() || resource.Attributes().Len() == 0 {
+	if resource.Attributes().Len() == 0 {
```
joe-elliott, line 251: It seems that `IsNil()` was removed and the resource is guaranteed to never be nil?

I find these generate proto wrappers very confusing.  Do you have any insight into why this was removed?

cmd/opentelemetry/app/defaultconfig/default_config.go
```diff
@@ -206,13 +206,15 @@ func createExporters(component ComponentType, storageTypes string, factories com
 
 func enableAgentUDPEndpoints(jaeger *jaegerreceiver.Config) {
 	if jaeger.ThriftCompact == nil {
-		jaeger.ThriftCompact = &confignet.TCPAddr{
+		jaeger.ThriftCompact = &jaegerreceiver.ProtocolUDP{
```
joe-elliott, line 209: Nice.  This will allow us to support max udp packet size in the Jaeger otel collectors.

joe-elliott (dismissed):
This looks good to me.  I do have one question about these proto wrappers.  Any insight you have would be appreciated.

cmd/opentelemetry/app/exporter/elasticsearchexporter/esmodeltranslator/modeltranslator.go
Vemmy124: Well, that went from my issue about panic in marshalling https://github.com/open-telemetry/opentelemetry-collector/issues/1985. Then gradually all proto things became non-nullable. I believe, that using pointers was a mistake from the very beginning

cmd/opentelemetry/app/defaultconfig/default_config.go
Vemmy124: Yeah, it is already configurable in otel receiver config

cmd/opentelemetry/app/exporter/elasticsearchexporter/esmodeltranslator/modeltranslator.go
Vemmy124: Yes, you're right, resource now cannot be nil

cmd/opentelemetry/app/defaultconfig/default_config.go
Vemmy124: Alongwith socket buffer size, number of workers and queue size

Vemmy124:
This is a flaky test I suppose

joe-elliott (approved):
Looks good.

yurishkuro (approved):

PR #2635, opened 2020-11-20
Fix flaky tbuffered server test

### LINKED ISSUE #2631: Flaky test TestTBufferedServer_Metrics
it strikes again https://travis-ci.com/github/yurishkuro/jaeger/jobs/440719322#L679
```
=== RUN   TestTBufferedServer_Metrics
    metricstest.go:49: 
        	Error Trace:	metricstest.go:49
        	            				metricstest.go:37
        	            				tbuffered_server_test.go:156
        	Error:      	Not equal: 
        	            	expected: int(1)
        	            	actual  : int64(0)
        	Test:       	TestTBufferedServer_Metrics
        	Messages:   	expected metric name: thrift.udp.server.read.errors, tags: map[]
--- FAIL: TestTBufferedServer_Metrics (0.01s)
FAIL
coverage: 97.3% of statements
FAIL	github.com/jaegertracing/jaeger/cmd/agent/app/servers	0.051s
```

pkositsyn:

Fixes #2631 

The order for error and second valid packet is changed in `fakeTransport`. Now having second packet dropped guarantees that the error is already processed, which absence was the reason for flaky test

### DIFF
--- a/cmd/agent/app/servers/tbuffered_server_test.go
+++ b/cmd/agent/app/servers/tbuffered_server_test.go
@@ -98,11 +98,13 @@ type fakeTransport struct {
 
 func (t *fakeTransport) Read(p []byte) (n int, err error) {
 	packet := t.packet.Inc()
-	if packet > 2 {
-		if packet > 3 {
-			// return error once when packet==3, otherwise block
-			t.wg.Wait()
-		}
+	if packet == 2 {
+		// return some error packet, followed by valid one
+		return 0, io.ErrNoProgress
+	}
+	if packet > 3 {
+		// don't return error at this moment to avoid flaky test
+		t.wg.Wait()
 		return 0, io.EOF
 	}
 	for i := range p {
@@ -128,9 +130,9 @@ func TestTBufferedServer_Metrics(t *testing.T) {
 	go server.Serve()
 	defer server.Stop()
 
-	// The fakeTransport will allow the server to read exactly two packets and one error.
+	// The fakeTransport will allow the server to read exactly two packets and one error in between.
 	// Since we use the server with queue size == 1, the first packet will be
-	// sent to channel, and the second one dropped.
+	// sent to channel, the error will be processed and the second valid packet dropped.
 
 	packetDropped := false
 	for i := 0; i < 5000; i++ {
--- a/cmd/agent/app/servers/tbuffered_server_test.go
+++ b/cmd/agent/app/servers/tbuffered_server_test.go
@@ -96,6 +96,10 @@ type fakeTransport struct {
 	wg     sync.WaitGroup
 }
 
+// Read simulates three packets received, then blocks until semaphore is released at the end of the test.
+// First packet is returned as normal.
+// Second packet is simulated as error.
+// Third packet is returned as normal, but will be dropped as overflow by the server whose queue size = 1.
 func (t *fakeTransport) Read(p []byte) (n int, err error) {
 	packet := t.packet.Inc()
 	if packet == 2 {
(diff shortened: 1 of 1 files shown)


cmd/agent/app/servers/tbuffered_server_test.go
```diff
@@ -98,11 +98,13 @@ type fakeTransport struct {
...
-			// return error once when packet==3, otherwise block
-			t.wg.Wait()
-		}
+	if packet == 2 {
+		// return some error packet, followed by valid one
+		return 0, io.ErrNoProgress
+	}
+	if packet > 3 {
+		// don't return error at this moment to avoid flaky test
+		t.wg.Wait()
```
yurishkuro, line 107: I don't remember the exact semantics of this test, but from a cursory review, previously the case `packet == 3` would result in EOF with no wait(), while now it falls through to the range loop below. Is that expected?

Vemmy124: Yes, the previous behaviour was: send two packets, the third one will return EOF immediately. And the flake could be if the error wasn't still handled at the moment we expect that and there is no was to watch it. That's why I changed the order for second packet and error - having a dropped second packet guarantees that the error is already handled

Vemmy124:
@jpkrohling Could you review this please?

cmd/agent/app/servers/tbuffered_server_test.go
```diff
@@ -98,11 +98,13 @@ type fakeTransport struct {
...
-		if packet > 3 {
-			// return error once when packet==3, otherwise block
-			t.wg.Wait()
-		}
+	if packet == 2 {
+		// return some error packet, followed by valid one
+		return 0, io.ErrNoProgress
+	}
+	if packet > 3 {
+		// don't return error at this moment to avoid flaky test
```
yurishkuro, line 106:
```suggestion
		// block after 3 packets until the server is shutdown and semaphore released
```

cmd/agent/app/servers/tbuffered_server_test.go
```diff
@@ -98,11 +98,13 @@ type fakeTransport struct {
 
 func (t *fakeTransport) Read(p []byte) (n int, err error) {
```
yurishkuro, line 99: let's add a comment describing behavior
```suggestion
// Read simulates three packets received, then blocks until semaphore is released at the end of the test.
// First packet is returned as normal.
// Second packet is simulated as error.
// Third packet is returned as normal, but will be dropped as overflow by the server whose queue size = 1.
func (t *fakeTransport) Read(p []byte) (n int, err error) {
```

cmd/agent/app/servers/tbuffered_server_test.go
```diff
@@ -128,9 +130,9 @@ func TestTBufferedServer_Metrics(t *testing.T) {
 	go server.Serve()
 	defer server.Stop()
 
-	// The fakeTransport will allow the server to read exactly two packets and one error.
+	// The fakeTransport will allow the server to read exactly two packets and one error in between.
 	// Since we use the server with queue size == 1, the first packet will be
-	// sent to channel, and the second one dropped.
+	// sent to channel, the error will be processed and the second valid packet dropped.
```
yurishkuro, line 135:
```suggestion
	// sent to channel, the error will increment the metric, and the second valid packet dropped.
```

yurishkuro (dismissed):
LGTM, with a few minor corrections to comments to make this more understandable (for me, at least)

yurishkuro (approved):
Thanks!

yurishkuro:
please do `git commit -s -amend` to sign the last commit

pkositsyn:
I tried to do that - but after I commited from github with incorrect name in profile had to merge into my local branch, so now it is third from the top broken commit...

pkositsyn:
Fixed with hard reset + amend

yurishkuro:
Thanks!

jpkrohling:
Sorry for the delay in reviewing, and thanks @yurishkuro for picking it up.

PR #2639, opened 2020-11-22
Cleanup travis config

Ashmita152:

Minor cleanup for travis.yml.

Switch from multiple if-else statements into case statement.

Travis run with this change: https://travis-ci.com/github/Ashmita152/jaeger/builds/203416459

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,42 +9,40 @@ matrix:
   include:
   - go: "1.15.x"
     env:
-    - TESTS=true
-    - COVERAGE=true
+    - TASKS=TESTS,COVERAGE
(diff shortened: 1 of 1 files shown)


yurishkuro:
@Ashmita152 what is the motivation for this change?  If it ain't broke...

If you're interested in cleaning up the CI setup, I think a more promising direction would be to start transitioning to GitHub actions, which we can do piecemeal. The jaegertracing org gets more concurrent GH executors than concurrent Travis executors, so it should be an improvement in how fast CI runs.

Ashmita152:
Thank you @yurishkuro That makes sense to me too. I will close this for now.

PR #2643, opened 2020-11-23
Clean-up Badger factory Close() behavior

### LINKED ISSUE #2642: Badger storage factory panics during Close
**Describe the bug**

The Badger storage factory has a inverted dependency that causes the factory to be closed twice, which leads to 
```
panic: close of closed channel

goroutine 1 [running]:
github.com/jaegertracing/jaeger/plugin/storage/badger.(*Factory).Close(0xc000692000, 0x5388600, 0xc000692000)
	/Users/ysh/dev/jaegertracing/jaeger/plugin/storage/badger/factory.go:175 +0x5b
github.com/jaegertracing/jaeger/plugin/storage.(*Factory).Close(0xc00030e000, 0x0, 0x0)
	/Users/ysh/dev/jaegertracing/jaeger/plugin/storage/factory.go:246 +0x19d
main.main.func1.1()
```

**To Reproduce**

1. Run `$ SPAN_STORAGE_TYPE=badger go run ./cmd/all-in-one`
2. Once it starts, kill process with Ctrl-C

**Expected behavior**
Process should exit cleanly.

 - OS: MacOS
 - Jaeger version: 1.21
 - Deployment: run from source

**Additional context**

The underlying cause is that when Badger Factory creates SpanWriter, it passes itself as a `closer` argument, which results in the following sequence:
1. main -> spanWriter.Close() -> factory.Close()
2. main -> factory.Close() -> panic

**Proposed Solution**

It is not clear why the factory needs to be passed into SpanWriter, it's an inversion of control. SpanWriter.Close() should do the clean-up that it needs to do, but then expect the parent to call Factory.Close() explicitly (which is what main does).

yurishkuro:
Remove inverted dependency of Badger SpanWriter on Badger Storage Factory.

Fixes #2642.

### DIFF
--- a/plugin/storage/badger/spanstore/writer.go
+++ b/plugin/storage/badger/spanstore/writer.go
@@ -189,8 +189,3 @@ func createTraceKV(span *model.Span, encodingType byte, startTime uint64) ([]byt
 
 	return key, bb, err
 }
-
-// Close Implements io.Closer but does nothing.
-func (w *SpanWriter) Close() error {
-	return nil
-}
--- a/plugin/storage/badger/spanstore/writer.go
+++ b/plugin/storage/badger/spanstore/writer.go
@@ -19,7 +19,6 @@ import (
 	"encoding/binary"
 	"encoding/json"
 	"fmt"
-	"io"
 	"time"
 
 	"github.com/dgraph-io/badger"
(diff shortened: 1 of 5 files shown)


plugin/storage/badger/spanstore/writer.go
```diff
@@ -193,7 +190,7 @@ func createTraceKV(span *model.Span, encodingType byte, startTime uint64) ([]byt
 	return key, bb, err
 }
 
-// Close Implements io.Closer and closes the underlying storage
+// Close Implements io.Closer but does nothing.
 func (w *SpanWriter) Close() error {
-	return w.closer.Close()
+	return nil
 }
```
yurishkuro, line 196: I think we could also remove this function altogether (will need a minor clean-up of the tests which expect the writer to implement io.Closer).

objectiser: Agree should be removed.

jpkrohling: Is it only this span writer that doesn't need a closer?

yurishkuro: Memory storage probably doesn't need a closer either. Even in Cassandra writer the state is in the session that is shared with the reader and thus closed by the factory.

jpkrohling: Alright, sounds good to me then.

jpkrohling (approved):

Issue #2645, opened 2020-11-24
Migrate from Travis to GitHub actions


yurishkuro:
We hit the limit of "credits" that Travis provides. Even though the OSS credits are supposed to renew on Nov-23, it's a sign that we need to move away from Travis.

Checklist for migration to GH Actions:
- [x] Unit Test + Coverage
- [x] Protogen Test
- [x] All-In-One
- [x] Crossdock
- [x] Crossdock OpenTelemetry
- [x] Docker Build + Deploy #2740
- [x] ES Integration Tests
- [x] ES OpenTelemetry Integration Tests
- [x] Kafka Integration Tests
- [x] Cassandra Integration Tests
- [x] Memory + Badger Integration Tests
- [x] HotRod Application Build
- [ ] Update build badges in README (may need to add an [expander](https://gist.github.com/pierrejoubert73/902cc94d79424356a8d20be2b382e1ab) section for all workflows, currently only Unit Test and FOSSA badges are shown)
- [ ] Find a way to cancel running workflows on a branch for stale commits. Travis had this as a configuration option, but GH Actions seem to require a special action in each workflow. Without this we'll be running a lot of redundant workflows.

We also need similar changes in the other repositories:

Required before the next Jaeger release:

  * [x] documentation https://github.com/jaegertracing/documentation/issues/461
  * [x] jaeged-ui https://github.com/jaegertracing/jaeger-ui/issues/668

Can be done independently:

  * Split into #2691 (all other repos)

<details>
  <summary>Screenshots from Travis</summary>

![image](https://user-images.githubusercontent.com/3523016/100034096-759f9080-2dc9-11eb-9041-5a40f2a21d25.png)

![image](https://user-images.githubusercontent.com/3523016/100034113-7f28f880-2dc9-11eb-916f-711a753463ae.png)
![image](https://user-images.githubusercontent.com/3523016/100034127-87813380-2dc9-11eb-9dfe-51da83856a8b.png)
  
</details>

cc @jaegertracing/jaeger-maintainers @Ashmita152

Ashmita152:
Sure I can pick this work but I am new to GitHub Actions, so it will take some time to get it working. Maybe we can have parallel builds for Travis and GitHub Actions for sometime and once we are satisfied, we can shutdown travis builds.

yurishkuro:
We do continue using Travis builds, but for some reason they ran out of credits which are not refreshing.

[...]

PR #2649, opened 2020-11-26
Add github action for jaeger integration tests

Ashmita152:

This PR adds GitHub actions for running integration tests for Elasticsearch, Cassandra, Memory, Kafka & Badger.

Part of #2645.

Also it removes `--interactive` and `--tty` flags from `test-compile-es-scripts` makefile directive, which shouldn't be used there and was causing elasticsearch integration test to fail on GH actions with this error (not sure how it is working on travis, maybe different docker version)

```
docker run --rm -it -v /home/runner/work/jaeger/jaeger:/tmp/jaeger python:3-alpine /usr/local/bin/python -m py_compile /tmp/jaeger/plugin/storage/es/esRollover.py
the input device is not a TTY
```


.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,93 @@
...
+    - name: Check out code
+      uses: actions/checkout@v2
+
+    - name: Setup go env
+      uses: actions/setup-go@v2
+      with:
+        go-version: ^1.15
+
+    - name: Install tools
+      run: make install-ci
```
yurishkuro, line 22: do we actually need this step for the integration tests?

Ashmita152: Good point, I looked at travis.yml and it was there in `install` section but I think we mayn't require it. Let me test it once.

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,93 @@
+name: Integration Tests
```
yurishkuro, line 1: I think it might be better to define all our CI jobs in a single file. That way we'll have better control over their orchestration, like using `.needs` to establish dependencies, define common parameters via `env`, like the Go version.

Ashmita152: Sure @yurishkuro that makes sense.

jpkrohling: The down side is that we can only trigger new runs on a per-workflow basis, meaning that all jobs within the workflow will be restarted.

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,93 @@
...
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  es-integration-tests:
```
jpkrohling, line 10: They all look quite similar. If they are indeed the same, just changing a parameter or two, then it might be better to use matrix jobs: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix

.github/workflows/integration-tests.yml
yurishkuro: yeah, if we cannot rerun individual jobs then it's better to have them as separate workflows.

It's annoying that we'd have to repeat a lot of boilerplate. At some point we might want to consider auto-generating the YAML files from smaller pieces. I originally was looking at [this approach](https://gist.github.com/elijahr/12fe5e13b1e1bbcd81eeaa7c06429e19) of using YAML aliases, but that only works in a single file.

jpkrohling: I think we can indeed have multiple workflows in the same file, but it would be a huge YAML no matter how much we Don't-Repeat-Ourselves. Let's see what's the final result, and we can then iterate and refactor.

Ashmita152: Thank you for the review. I am little confused if I should merge unit test and integration tests into one yaml. Using matrix, definitely made the config much cleaner.

Ashmita152: Also, I think our badger integration tests broke due to some recent commit.

yurishkuro:
> I am little confused if I should merge unit test and integration tests into one yaml. Using matrix, definitely made the config much cleaner.

@Ashmita152 we need to be able to restart individual jobs, not the complete CI run. If we can do it with a single workflow file then I would consolidate all CI into one place (just like it was with .travis.yml), otherwise it's better to have independent workflow files.

Ashmita152:
@yurishkuro Looking at the GitHub community threads, it seems like GH actions don't have support restarting individual job. I would suggest maybe we can go ahead with different workflows for now and merge them into one once GitHub actions support it.

yurishkuro:
Let's go with different files. My only other alternative suggestion is to write a shell script that would generate the Yaml files, and execute it as a commit hook. Since the files are only going to be different in one command, the script could be fairly straightforward.

Ashmita152:
Hi @yurishkuro 

It makes sense to have a script generate the workflow files. Let me look at that option.

.github/workflows/cassandra-integration-tests.yml
```diff
@@ -0,0 +1,22 @@
+name: Cassandra Integration Tests
```
yurishkuro, line 1: just a suggestion: call all workflows "Integration Tests", since there's also a distinct job name to uniquely identify each.

Ashmita152:
Hi @yurishkuro 

I thought maybe I can do the script version in another PR since I am not quite familiar with git hooks. To have the CI working quickly, we moved each integration test into separate files. Let me know what you think. Thanks.

.github/workflows/cassandra-integration-tests.yml
```diff
@@ -0,0 +1,22 @@
...
+    branches: [ master ]
+
+jobs:
+  cassandra:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Check out code
+      uses: actions/checkout@v2
+
+    - name: Setup go env
```
yurishkuro, line 16: suggestion: these two steps do not need `name`, just the `uses` keyword

.github/workflows/memory-badger-integration-tests.yml
```diff
@@ -0,0 +1,22 @@
+name: Memory and Badger Integration Tests
```
yurishkuro, line 1: perhaps add this file in the PR that fixes the Badger tests, so that this PR could be merged green.

yurishkuro:
yes, let's start with explicit files, auto-generation can be done later

Ashmita152:
Thank you @yurishkuro updated the PR with the feedback changes.

.github/workflows/cassandra-integration-tests.yml
jpkrohling: I think GitHub actions actually group the workflows by name, not file. If you use the same name in different files, you might end up with a single workflow, no ?

jpkrohling: No, I was wrong: just tested by restarting one of the integration test jobs:

image

yurishkuro (approved):
Thanks!

Issue #2653, opened 2020-11-26
Jaeger collector crashed accidentally due to samara return `nil` error


Betula-L:
**Describe the bug**
Collector panic when print error log since `e.Err` is `nil`.
https://github.com/jaegertracing/jaeger/blob/b8ccf7342ee81a41cc93af6462af7a781530cb38/plugin/storage/kafka/writer.go#L60

It may be caused by sarama bug which return unexpected `nil` to error channel. It's hard to track the root cause of sarama sdk. 

https://github.com/Shopify/sarama/blob/0616f68815691527d3b6ab4d95a2881e09400fe3/async_producer.go#L1072

**Expected behavior**

However, jaeger should not concern about it, just wrap `e.Err`.

Do not panic collector because of kafka sdk bug.

**Version (please complete the following information):**
 - OS: Linux
 - Jaeger version: v1.16.0
 - Deployment: Docker

**Additional context**

panic logs:
```
env:prod
log:panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x10aca5f]
goroutine 42 [running]:
github.com/jaegertracing/jaeger/plugin/storage/kafka.NewSpanWriter.func1(0x161c860, 0xc00079e310, 0xc00010ac00, 0x15f4c80, 0xc000386ce0, 0x15f4c80, 0xc000386d50)
	github.com/jaegertracing/jaeger/plugin/storage/kafka/writer.go:54 +0x12f
created by github.com/jaegertracing/jaeger/plugin/storage/kafka.NewSpanWriter
	github.com/jaegertracing/jaeger/plugin/storage/kafka/writer.go:50 +0x273
```

PR #2654, opened 2020-11-26
fix collector panic due to sarama sdk

Betula-L:
- Fix https://github.com/jaegertracing/jaeger/issues/2653

### DIFF
--- a/plugin/storage/kafka/writer.go
+++ b/plugin/storage/kafka/writer.go
@@ -57,7 +57,9 @@ func NewSpanWriter(
 	}()
 	go func() {
 		for e := range producer.Errors() {
-			logger.Error(e.Err.Error())
+			if e != nil && e.Err != nil {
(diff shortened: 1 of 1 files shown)


plugin/storage/kafka/writer.go
```diff
@@ -57,8 +57,10 @@ func NewSpanWriter(
 	}()
 	go func() {
 		for e := range producer.Errors() {
-			logger.Error(e.Err.Error())
-			writeMetrics.SpansWrittenFailure.Inc(1)
+			if e != nil && e.Err != nil {
+				logger.Error(e.Err.Error())
+				writeMetrics.SpansWrittenFailure.Inc(1)
```
objectiser, line 62: This line should not be within the `if`.

Betula-L: Done

objectiser (approved):
Thanks!

PR #2655, opened 2020-11-26
Move cert loading failure assertion on client public key write

albertteoh:

- Fixes #2654 

- Local testing revealed that `"Failed to load certificate"` is logged every time; however some tests in CI are not logging this failure state, perhaps due to disk I/O variability.
- Given the failure signal is unreliable, this PR removes the failure assertion and only performs an assertion identical to the `waitUntil` condition which should make this test more reliable.

### DIFF
--- a/pkg/config/tlscfg/cert_watcher_test.go
+++ b/pkg/config/tlscfg/cert_watcher_test.go
@@ -97,7 +97,7 @@ func TestReload(t *testing.T) {
 	assert.True(t, logObserver.
 		FilterMessage("Failed to load certificate").
 		FilterField(zap.String("certificate", certFile.Name())).Len() > 0,
-		"Unable to locate 'Failed to load certificate' error in log. All logs: "+fmt.Sprint(logObserver.All()))
+		"Unable to locate 'Failed to load certificate' in log. All logs: "+fmt.Sprint(logObserver.All()))
 
 	// Write the client's private key.
 	keyData, err = ioutil.ReadFile(clientKey)
@@ -115,7 +115,7 @@ func TestReload(t *testing.T) {
 	assert.True(t, logObserver.
 		FilterMessage("Loaded modified certificate").
 		FilterField(zap.String("certificate", keyFile.Name())).Len() > 0,
-		"Unable to locate 'Loaded modified certificate' error in log. All logs: "+fmt.Sprint(logObserver.All()))
+		"Unable to locate 'Loaded modified certificate' in log. All logs: "+fmt.Sprint(logObserver.All()))
 
 	cert, err = tls.LoadX509KeyPair(filepath.Clean(clientCert), clientKey)
 	require.NoError(t, err)
--- a/pkg/config/tlscfg/cert_watcher_test.go
+++ b/pkg/config/tlscfg/cert_watcher_test.go
@@ -99,15 +99,7 @@ func TestReload(t *testing.T) {
 			FilterField(zap.String("certificate", keyFile.Name())).Len() > 0
 	}, 100, time.Millisecond*200)
 
-	// Logged when the cert is modified with the client's public key due to
-	// a mismatch with the existing server private key.
-	assert.True(t, logObserver.
-		FilterMessage("Failed to load certificate").
-		FilterField(zap.String("certificate", certFile.Name())).Len() > 0,
-		"Failed to find wanted logs. All logs: "+fmt.Sprint(logObserver.All()))
-
-	// Logged when the cert is modified with the client's private key,
-	// resulting in both public and private keys matching (from the client).
+	// Logged when the cert is modified with the client's private key.
 	assert.True(t, logObserver.
 		FilterMessage("Loaded modified certificate").
 		FilterField(zap.String("certificate", keyFile.Name())).Len() > 0,
--- a/pkg/config/tlscfg/cert_watcher_test.go
+++ b/pkg/config/tlscfg/cert_watcher_test.go
@@ -97,7 +97,7 @@ func TestReload(t *testing.T) {
 	assert.True(t, logObserver.
 		FilterMessage("Failed to load certificate").
 		FilterField(zap.String("certificate", certFile.Name())).Len() > 0,
-		"Failed to find wanted logs. All logs: "+fmt.Sprint(logObserver.All()))
+		"Unable to locate 'Failed to load certificate' error in log. All logs: "+fmt.Sprint(logObserver.All()))
 
 	// Write the client's private key.
 	keyData, err = ioutil.ReadFile(clientKey)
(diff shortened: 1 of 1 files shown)


objectiser (approved):
According to the comment on https://github.com/jaegertracing/jaeger/blob/1a24f204ba959ee4b2a15e3b4693ffadbe9300bb/pkg/config/tlscfg/cert_watcher_test.go#L97 the "Failed to load certificate" shouldn't be generated anyway - but given the desired state (e.g. "Loaded modified certificate") is asserted, looks ok to remove this check.

pkg/config/tlscfg/cert_watcher_test.go
```diff
@@ -99,15 +99,7 @@ func TestReload(t *testing.T) {
 			FilterField(zap.String("certificate", keyFile.Name())).Len() > 0
 	}, 100, time.Millisecond*200)
 
-	// Logged when the cert is modified with the client's public key due to
-	// a mismatch with the existing server private key.
```
yurishkuro, line 103: It sounds to me that if this doesn't always happen then we have a flaky test conditions that would result in some code branches not executed reliably. The assertion was meant to ensure that they are always executed. If we remove the assertion, it could make the test results more stable, but the code coverage less stable, which is just as bad.

yurishkuro: For the record, the usual reason for flaky test behavior is heavier CPU preemption in the shared CI environments. I see that the wait above is only for 200ms, it's quite easy for the run to hiccup for longer than that, I usually make sure the wait is at least a second. The ideal solution is to make the test conditions deterministic, i.e. if we certain that the condition will happen, then we can safely use longer waits, like 10sec, since it's unlikely the CI run will be CPU starved for that long while at the same time the long waits won't make the test run longer under normal circumstances (especially locally).

objectiser: If the logged error is expected, then the comment I referenced is incorrect. However there is another test that covers the mismatch case that does check for that logged error.

However agree that increasing the wait may have been a good first step.

albertteoh: The wait above is 200ms x 100 iterations = 20sec, which is more than enough time I think. 

@objectiser is right that there is line coverage for the public/private key mismatch case.

The problem as it stands is that we're making two file writes in succession then asserting on both states. If we do, however, want to have a stronger assertion on the behaviour of cert reloads, why don't we assert each state after each file write?

That is, move the failure assertion up, after writing the client's public key to the `certFile`; which should give enough time for the certWatcher to pick up this change in state.

That is:
```
	// Write the client's public key
	certData, err = ioutil.ReadFile(clientCert)
	require.NoError(t, err)
	err = syncWrite(certFile.Name(), certData, 0644)
	require.NoError(t, err)

        // Wait for the cert watcher to log an error due to mismatching private/public keys
	waitUntil(func() bool {
		// Logged when the cert is modified with the client's public key due to
		// a mismatch with the existing server private key.
		return logObserver.FilterMessage("Failed to load certificate").
			FilterField(zap.String("certificate", certFile.Name())).Len() > 0
	}, 100, time.Millisecond*200)

	assert.True(t, logObserver.
		FilterMessage("Failed to load certificate").
		FilterField(zap.String("certificate", certFile.Name())).Len() > 0,
		"Failed to find wanted logs. All logs: "+fmt.Sprint(logObserver.All()))

        // Write the client's private key
	keyData, err = ioutil.ReadFile(clientKey)
	require.NoError(t, err)
	err = syncWrite(keyFile.Name(), keyData, 0644)
	require.NoError(t, err)
        ... // wait for "Loaded modified certificate" as below
```

What do you think @yurishkuro @objectiser?

objectiser: Although the main state this test is interested in is the "loaded modified certificate", so checking for the failure is not strictly necessary, doing it this way may help users examining the test logs that find the error, as they would see the test was anticipating it. So +1 from me.

albertteoh: Thanks for the feedback @objectiser & @yurishkuro, I've moved the failure assertion to after writing the client's public key as suggested.

pkg/config/tlscfg/cert_watcher_test.go
```diff
@@ -82,32 +82,36 @@ func TestReload(t *testing.T) {
...
+	waitUntil(func() bool {
+		// Logged when the cert is reloaded with mismatching client public key and existing server private key.
+		return logObserver.FilterMessage("Failed to load certificate").
+			FilterField(zap.String("certificate", certFile.Name())).Len() > 0
+	}, 100, time.Millisecond*200)
+
+	assert.True(t, logObserver.
+		FilterMessage("Failed to load certificate").
+		FilterField(zap.String("certificate", certFile.Name())).Len() > 0,
+		"Failed to find wanted logs. All logs: "+fmt.Sprint(logObserver.All()))
```
objectiser, line 100: One minor point - could you differentiate this log message from the one at the end of this function - e.g. "Unable to locate 'Failed to load certificate' error in log"?

albertteoh: Yup, good pickup, @objectiser.

objectiser (approved):

objectiser (approved):

pkg/config/tlscfg/cert_watcher_test.go
```diff
@@ -82,36 +82,40 @@ func TestReload(t *testing.T) {
...
 	certData, err = ioutil.ReadFile(clientCert)
 	require.NoError(t, err)
 	err = syncWrite(certFile.Name(), certData, 0644)
 	require.NoError(t, err)
+
+	waitUntil(func() bool {
+		// Logged when the cert is reloaded with mismatching client public key and existing server private key.
+		return logObserver.FilterMessage("Failed to load certificate").
+			FilterField(zap.String("certificate", certFile.Name())).Len() > 0
+	}, 100, time.Millisecond*200)
```
yurishkuro, line 95: nit: instead of 100 * 200ms, I would do 2000 * 10ms - minimize the guaranteed delay, keep total delay the same (absolute # is not particularly important, 20s is probably overkill, but 10s would be fine, enough to outlast any scheduling hiccup).

pkg/config/tlscfg/cert_watcher_test.go
```diff
@@ -82,36 +82,40 @@ func TestReload(t *testing.T) {
...
+
+	assert.True(t, logObserver.
+		FilterMessage("Failed to load certificate").
+		FilterField(zap.String("certificate", certFile.Name())).Len() > 0,
+		"Unable to locate 'Failed to load certificate' in log. All logs: "+fmt.Sprint(logObserver.All()))
```
yurishkuro, line 100: nit: assert uses Printf syntax, you could do
```
"Unable to locate 'Failed to load certificate' in log. All logs: %v", logObserver.All())
```

Issue #2659, opened 2020-11-27
Badger Integration Test Failures


Ashmita152:
**Describe the bug**
We cleaned up the badger factory close() as part of this https://github.com/jaegertracing/jaeger/pull/2643 but we forgot to update the integration tests accordingly since travis is broken.

```
--- FAIL: TestBadgerStorage (46.09s)
    --- FAIL: TestBadgerStorage/GetServices (0.01s)
    --- FAIL: TestBadgerStorage/GetOperations (0.01s)
    --- FAIL: TestBadgerStorage/GetTrace (0.01s)
        --- PASS: TestBadgerStorage/GetTrace/NotFound_error (0.00s)
    --- FAIL: TestBadgerStorage/GetLargeSpans (24.18s)
    --- FAIL: TestBadgerStorage/FindTraces (21.80s)
        --- PASS: TestBadgerStorage/FindTraces/Tags_in_one_spot_-_Tags (0.01s)
        --- PASS: TestBadgerStorage/FindTraces/Tags_in_one_spot_-_Logs (0.01s)
        --- PASS: TestBadgerStorage/FindTraces/Tags_in_one_spot_-_Process (0.01s)
        --- PASS: TestBadgerStorage/FindTraces/Tags_in_different_spots (0.01s)
        --- PASS: TestBadgerStorage/FindTraces/Trace_spans_over_multiple_indices (0.01s)
        --- PASS: TestBadgerStorage/FindTraces/Operation_name (0.01s)
        --- PASS: TestBadgerStorage/FindTraces/Operation_name_+_max_Duration (0.01s)
        --- PASS: TestBadgerStorage/FindTraces/Operation_name_+_Duration_range (0.01s)
        --- PASS: TestBadgerStorage/FindTraces/Duration_range (0.01s)
        --- PASS: TestBadgerStorage/FindTraces/max_Duration (0.01s)
        --- FAIL: TestBadgerStorage/FindTraces/default (21.59s)
        --- PASS: TestBadgerStorage/FindTraces/Tags_+_Operation_name (0.01s)
        --- PASS: TestBadgerStorage/FindTraces/Tags_+_Operation_name_+_max_Duration (0.01s)
        --- PASS: TestBadgerStorage/FindTraces/Tags_+_Operation_name_+_Duration_range (0.01s)
        --- PASS: TestBadgerStorage/FindTraces/Tags_+_Duration_range (0.01s)
[...]

PR #2660, opened 2020-11-27
Fix for failures in badger integration tests

Ashmita152:

Closes https://github.com/jaegertracing/jaeger/issues/2659

Update the integration test based on the cleanup done in badger factory close() function.

### DIFF
--- a/plugin/storage/integration/badgerstore_test.go
+++ b/plugin/storage/integration/badgerstore_test.go
@@ -15,8 +15,6 @@
 package integration
 
 import (
-	"fmt"
-	"io"
 	"os"
 	"testing"
 
@@ -30,22 +28,23 @@ import (
 
 type BadgerIntegrationStorage struct {
 	StorageIntegration
-	logger *zap.Logger
+	logger  *zap.Logger
+	factory *badger.Factory
 }
 
 func (s *BadgerIntegrationStorage) initialize() error {
-	f := badger.NewFactory()
+	s.factory = badger.NewFactory()
 
-	err := f.Initialize(metrics.NullFactory, zap.NewNop())
+	err := s.factory.Initialize(metrics.NullFactory, zap.NewNop())
 	if err != nil {
 		return err
 	}
 
-	sw, err := f.CreateSpanWriter()
+	sw, err := s.factory.CreateSpanWriter()
 	if err != nil {
 		return err
 	}
-	sr, err := f.CreateSpanReader()
+	sr, err := s.factory.CreateSpanReader()
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


plugin/storage/integration/badgerstore_test.go
```diff
@@ -30,22 +28,21 @@ import (
...
-	logger *zap.Logger
+	logger  *zap.Logger
+	factory *badger.Factory
 }
 
 func (s *BadgerIntegrationStorage) initialize() error {
-	f := badger.NewFactory()
-
-	err := f.Initialize(metrics.NullFactory, zap.NewNop())
+	err := s.factory.Initialize(metrics.NullFactory, zap.NewNop())
```
yurishkuro, line 36: This doesn't look quite right, since `initialize()` function is called multiple times, but I am not sure that `factory.Initialize()` is meant to be called more than once. I would suggest moving it into `TestBadgerStorage()` if you want to reuse the factory.

And another issue is that whether `cleanUp()` function is supposed to reset the storage to empty. If so, I would create new factory in `initialize()` every time and close it in `cleanUp()`.

yurishkuro:
Perhaps it's worth adding the GH workflow for badger storage test in this PR, so that we can see the tests actually succeeding.

plugin/storage/integration/badgerstore_test.go
yurishkuro: ^

Ashmita152: @yurishkuro I have updated the PR. May I ask if this is what was intended.

plugin/storage/integration/badgerstore_test.go
```diff
@@ -64,22 +63,7 @@ func (s *BadgerIntegrationStorage) initialize() error {
 	return nil
 }
 
-func (s *BadgerIntegrationStorage) clear() error {
```
yurishkuro, line 67: this function should remain and call factory.Close

Ashmita152: Updated the PR with the feedback.

plugin/storage/integration/badgerstore_test.go
```diff
@@ -65,14 +64,11 @@ func (s *BadgerIntegrationStorage) initialize() error {
 }
 
 func (s *BadgerIntegrationStorage) clear() error {
-	if closer, ok := s.SpanWriter.(io.Closer); ok {
-		err := closer.Close()
-		if err != nil {
-			return err
-		}
-		return nil
+	err := s.factory.Close()
```
yurishkuro, line 67: simply:
```suggestion
	return s.factory.Close()
```

yurishkuro (approved):
lgtm, thanks

Ashmita152:
Thank you, sorry just squashed the commits.

PR #2663, opened 2020-11-28
Add github action for jaeger all-in-one image

Ashmita152:

This PR adds the build job of all-in-one image to github actions. 

1. To make the docker login action work we need to add two environment variables like this 

jaeger

2. To get the branch name we are using github action https://github.com/nelonoel/branch-name which populates env var `BRANCH_NAME`

3. I tried using nodejs official github action (https://github.com/actions/setup-node) instead of calling `scripts/travis/install-ui-deps.sh` but it didn't work and I wasn't able to make it work after several retries. So thought do it in the same way how we do in travis for now.

### DIFF
--- a/.github/workflows/all-in-one-build.yml
+++ b/.github/workflows/all-in-one-build.yml
@@ -1,0 +1,50 @@
+name: Docker
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  all-in-one:
+    runs-on: ubuntu-latest
+    env:
+      BRANCH: ${{ github.head_ref }}
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        submodules: true
+
+    - uses: actions/setup-go@v2
+      with:
+        go-version: ^1.15
+
+    - uses: actions/setup-node@v2-beta
+      with:
+        node-version: '10'
+
+    - uses: docker/login-action@v1
+      id: dockerhub-login
+      with:
+        username: jaegertracingbot
+        password: ${{ secrets.DOCKERHUB_TOKEN }}
+      env:
+        DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
+      if: env.DOCKERHUB_TOKEN != null
+
+    - name: Set DOCKERHUB_LOGIN
+      run: |
+        echo "DOCKERHUB_LOGIN=true" >> $GITHUB_ENV
+      if: steps.dockerhub-login.outcome == "success"
+
+    - name: Export BRANCH variable if undefined
+      run: |
+        echo "BRANCH=${GITHUB_REF##*/}" >> $GITHUB_ENV
+      if: env.BRANCH == null
+
+    - name: Install tools
+      run: make install-ci
+
+    - name: Build, test, and publish all-in-one image
+      run: bash scripts/travis/build-all-in-one-image.sh
--- a/.github/workflows/all-in-one-build.yml
+++ b/.github/workflows/all-in-one-build.yml
@@ -14,6 +14,10 @@ jobs:
       with:
         submodules: true
 
+    - name: Fetch git tags
+      run: |
+        git fetch --prune --unshallow --tags
+
     - uses: actions/setup-go@v2
       with:
         go-version: ^1.15
--- a/.github/workflows/all-in-one-build.yml
+++ b/.github/workflows/all-in-one-build.yml
@@ -9,8 +9,6 @@ on:
 jobs:
   all-in-one:
     runs-on: ubuntu-latest
-    env:
-      BRANCH: ${{ github.head_ref }}
     steps:
     - uses: actions/checkout@v2
       with:
@@ -38,10 +36,15 @@ jobs:
         echo "DOCKERHUB_LOGIN=true" >> $GITHUB_ENV
       if: steps.dockerhub-login.outcome == 'success'
 
-    - name: Export BRANCH variable if undefined
+    - name: Export BRANCH variable for pull_request event
+      run: |
+        echo "BRANCH=${GITHUB_HEAD_REF}" >> $GITHUB_ENV
+      if: github.event_name == 'pull_request'
+
+    - name: Export BRANCH variable for push event
       run: |
         echo "BRANCH=${GITHUB_REF##*/}" >> $GITHUB_ENV
-      if: env.BRANCH == null
+      if: github.event_name == 'push'
 
     - name: Install tools
       run: make install-ci
--- a/.github/workflows/all-in-one-build.yml
+++ b/.github/workflows/all-in-one-build.yml
@@ -25,7 +25,6 @@ jobs:
         node-version: '10'
 
     - uses: docker/login-action@v1
-      id: dockerhub-login
       with:
         username: jaegertracingbot
         password: ${{ secrets.DOCKERHUB_TOKEN }}
@@ -36,7 +35,7 @@ jobs:
     - name: Set DOCKERHUB_LOGIN
       run: |
         echo "DOCKERHUB_LOGIN=true" >> $GITHUB_ENV
-      if: steps.dockerhub-login.outcome == "success"
+      if: env.DOCKERHUB_TOKEN != null
 
     - name: Export BRANCH variable if undefined
       run: |
--- a/.github/workflows/all-in-one-build.yml
+++ b/.github/workflows/all-in-one-build.yml
@@ -1,4 +1,4 @@
-name: Docker
+name: Build all-in-one
 
 on:
   push:
--- a/.github/workflows/all-in-one-build.yml
+++ b/.github/workflows/all-in-one-build.yml
@@ -35,7 +35,7 @@ jobs:
     - name: Export DOCKERHUB_LOGIN variable
       run: |
         echo "DOCKERHUB_LOGIN=true" >> $GITHUB_ENV
-      if: env.DOCKERHUB_TOKEN != null
+      if: success()
 
     - name: Export BRANCH variable if undefined
       run: |
--- a/.github/workflows/all-in-one-build.yml
+++ b/.github/workflows/all-in-one-build.yml
@@ -25,6 +25,7 @@ jobs:
         node-version: '10'
 
     - uses: docker/login-action@v1
+      id: dockerhub-login
       with:
         username: jaegertracingbot
         password: ${{ secrets.DOCKERHUB_TOKEN }}
@@ -35,7 +36,7 @@ jobs:
     - name: Export DOCKERHUB_LOGIN variable
       run: |
         echo "DOCKERHUB_LOGIN=true" >> $GITHUB_ENV
-      if: success()
+      if: steps.dockerhub-login.outcome == 'success'
 
     - name: Export BRANCH variable if undefined
       run: |
--- a/.github/workflows/all-in-one-build.yml
+++ b/.github/workflows/all-in-one-build.yml
@@ -32,7 +32,7 @@ jobs:
         DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
       if: env.DOCKERHUB_TOKEN != null
 
-    - name: Set DOCKERHUB_LOGIN
+    - name: Export DOCKERHUB_LOGIN variable
       run: |
         echo "DOCKERHUB_LOGIN=true" >> $GITHUB_ENV
       if: env.DOCKERHUB_TOKEN != null
--- a/scripts/travis/build-all-in-one-image.sh
+++ b/scripts/travis/build-all-in-one-image.sh
@@ -12,9 +12,9 @@ expected_version="v10"
 version=$(node --version)
 major_version=${version%.*.*}
 if [ "$major_version" = "$expected_version" ] ; then
-  echo "node version is as expected"
+  echo "Node version is as expected: $version"
 else
-  echo "installed version $major_version doesn't match $expected_version"
+  echo "ERROR: installed Node version $version doesn't match expected version $expected_version"
   exit 1
 fi
 
--- a/scripts/travis/build-all-in-one-image.sh
+++ b/scripts/travis/build-all-in-one-image.sh
@@ -8,8 +8,16 @@ BRANCH=${BRANCH:?'missing BRANCH env var'}
 # `GOARCH=<target arch> ./scripts/travis/build-all-in-one-image.sh`.
 GOARCH=${GOARCH:-$(go env GOARCH)}
 
-source ~/.nvm/nvm.sh
-nvm use 10
+expected_version="v10"
+version=$(node --version)
+major_version=${version%.*.*}
+if [ "$major_version" = "$expected_version" ] ; then
+  echo "node version is as expected"
+else
+  echo "installed version $major_version doesn't match $expected_version"
+  exit 1
+fi
+
 make build-ui
 
 set +e
@@ -28,7 +36,7 @@ run_integration_test() {
 
 upload_to_docker() {
   # Only push the docker container to Docker Hub for master branch
-  if [[ ("$BRANCH" == "master" || $BRANCH =~ ^v[0-9]+\.[0-9]+\.[0-9]+$) && "$TRAVIS_SECURE_ENV_VARS" == "true" ]]; then
+  if [[ ("$BRANCH" == "master" || $BRANCH =~ ^v[0-9]+\.[0-9]+\.[0-9]+$) && "DOCKERHUB_LOGIN" == "true" ]]; then
     echo "upload $1 to Docker Hub"
     export REPO=$1
     bash ./scripts/travis/upload-to-docker.sh
@@ -38,6 +46,7 @@ upload_to_docker() {
 }
 
 make build-all-in-one GOOS=linux GOARCH=$GOARCH
+make create-baseimg-debugimg
 repo=jaegertracing/all-in-one
 docker build -f cmd/all-in-one/Dockerfile \
         --target release \
--- a/.github/workflows/es-integration-tests.yml
+++ b/.github/workflows/es-integration-tests.yml
@@ -16,5 +16,8 @@ jobs:
       with:
         go-version: ^1.15
 
+    - name: Install tools
+      run: make install-ci
+
     - name: Run elasticsearch integration tests
       run: bash scripts/travis/es-integration-test.sh
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -16,7 +16,7 @@
     * Title "Release X.Y.Z" 
     * Tag `vX.Y.Z` (note the `v` prefix) and choose appropriate branch
     * Copy the new CHANGELOG.md section into the release notes
-4. The release tag will trigger a build of the docker images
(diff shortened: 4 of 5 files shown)


Ashmita152:
Until we add two environment variables, the build will fail. 

Sample Success Run: https://github.com/Ashmita152/jaeger/pull/8/checks?check_run_id=1466574790

.github/workflows/all-in-one-build.yml
```diff
@@ -0,0 +1,35 @@
...
+    - uses: actions/setup-go@v2
+      with:
+        go-version: ^1.15
+
+    - uses: docker/login-action@v1
+      with:
+        username: ${{ secrets.DOCKERHUB_USERNAME }}
+        password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+    - uses: nelonoel/branch-name@v1.0.1
```
yurishkuro, line 26: I don't like to depend on an external action, and a pretty complicated at that, for something that we do with a single shell line https://github.com/jaegertracing/jaeger/blob/556e3ba94bcf9742ceb6ffc53f806647de3b0e64/.travis.yml#L67

I recommend defining our own small GH action that we can reuse in multiple workflows, as I did in https://github.com/jaegertracing/jaeger-client-go/pull/552. That example also shows how to export an env variable by writing into a special GH file.

Ashmita152: Sure this sounds good.

.github/workflows/all-in-one-build.yml
```diff
@@ -0,0 +1,35 @@
...
+      with:
+        submodules: true
+
+    - uses: actions/setup-go@v2
+      with:
+        go-version: ^1.15
+
+    - uses: docker/login-action@v1
+      with:
+        username: ${{ secrets.DOCKERHUB_USERNAME }}
```
yurishkuro, line 23: Would it make more sense not to encrypt the user name, which happens to be `jaegertracingbot`?

Ashmita152: Yes I do agree.

yurishkuro: I defined the token in secrets
image

scripts/travis/build-all-in-one-image.sh
```diff
@@ -2,7 +2,7 @@
 
 set -exu
 
-BRANCH=${BRANCH:?'missing BRANCH env var'}
+BRANCH=${BRANCH_NAME:?'missing BRANCH env var'}
```
yurishkuro, line 5: can we rename this back to minimize changes?

Ashmita152: Agree, sure

.github/workflows/all-in-one-build.yml
Ashmita152: Thank you.

scripts/travis/build-all-in-one-image.sh
```diff
@@ -28,7 +28,7 @@ run_integration_test() {
 
 upload_to_docker() {
   # Only push the docker container to Docker Hub for master branch
-  if [[ ("$BRANCH" == "master" || $BRANCH =~ ^v[0-9]+\.[0-9]+\.[0-9]+$) && "$TRAVIS_SECURE_ENV_VARS" == "true" ]]; then
```
yurishkuro, line 31: The reason we had this TRAVIS_SECURE_ENV_VARS check was to prevent attempt to deploy snapshot images from pull requests. What would be the equivalent of that?  I assume the DockerHub Token secret won't be available in forks.

PS. It would be good to document this part of the publishing mechanism in RELEASE.md

Ashmita152: I see. sure.

yurishkuro:
> I tried using nodejs official github action (https://github.com/actions/setup-node) instead of calling `scripts/travis/install-ui-deps.sh` but it didn't work

Could you include some details what didn't work?

Not a big deal, I think it's fine to use `scripts/travis/install-ui-deps.sh`, but if you already spent some time investigating it's good to record the findings in case someone wants to try.

BTW, once we're completely off Travis, it would be nice to move scripts/travis/* to just scripts/.

.github/workflows/all-in-one-build.yml
Ashmita152: Not sure if GitHub Actions can see secrets at organisation level.

yurishkuro: Hard to imagine why it wouldn't. The token itself is working, I just tested it
```
$ docker login
Login with your Docker ID to push and pull images from Docker Hub. If you don't have a Docker ID, head over to https://hub.docker.com to create one.
Username: jaegertracingbot
Password:
Login Succeeded
```

Ashmita152: The recent run failed https://github.com/jaegertracing/jaeger/pull/2663/checks?check_run_id=1466696346

yurishkuro: The Settings page says:

> Secrets are not passed to workflows that are triggered by a pull request from a fork.

This is why I was warning about TRAVIS_SECURE_ENV_VARS. The previous setup was intentionally organized in such a way that almost all of the CI job is executed, and only the push to the registry is skipped when it's not from master.

Ashmita152: We might need to define the secret at repo level instead of org level.

Ashmita152: Ah I see.

.github/workflows/all-in-one-build.yml
```diff
@@ -0,0 +1,36 @@
...
+    - name: Export BRANCH variable
+      uses: ./.github/actions/export-branch
+
+    - name: Install tools
+      run: make install-ci
+
+    - name: Install UI dependencies
+      run: bash scripts/travis/install-ui-deps.sh
+
+    - name: Build all-in-one image
```
yurishkuro, line 35:
```suggestion
    - name: Build, test, and publish all-in-one image
```

.github/workflows/all-in-one-build.yml
Ashmita152: Finally able to make it work by doing docker login only if the secret is defined. Not sure if this is okay method.

Ashmita152: I think this will leak secret so definitely not the way.

Ashmita152: I tested in my jaeger fork putting a random string for DOCKERHUB_TOKEN. It should be all fine. I was worried about leaking it in logs but looks like encrypted secrets never gets printed. I tested it. Also their documentation says to use secrets to populate env vars, so it should be all good (https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets#using-encrypted-secrets-in-a-workflow)

docker-registry

Ashmita152:
> Could you include some details about what didn't work?

The issue was mainly around GitHub actions install node at path `/opt/hostedtoolcache/node/10.23.0/x64/bin/node` so when ` source /home/runner/.nvm/nvm.sh` was executing it was failing with VERSION being populated as N/A and hence that script
was failing with 
```
nvm_err 'You need to run "nvm install 10" to install it before using it.'
```
I felt it maybe I am missing something.

yurishkuro:
if we already know that we're running in the correct Node version, we do not need `nvm use`.

scripts/travis/build-all-in-one-image.sh
```diff
@@ -28,7 +28,7 @@ run_integration_test() {
```
yurishkuro, line 28: I think we should replace L11-12 with a check that the correct version of Node is in use and make it the responsibility of the caller. Also, executing `source` command when `set -x` is active creates a mess of the logs.

scripts/travis/build-all-in-one-image.sh
yurishkuro: In the logs the value of BRANCH seems to be `merge`: https://github.com/jaegertracing/jaeger/runs/1467144626?check_suite_focus=true#step:8:1639
```
skip docker upload for PR
+ [[ merge == \m\a\s\t\e\r ]]
+ [[ merge =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]
+ echo 'skip docker upload for PR'
```

yurishkuro (commented):
What is this error about? https://github.com/jaegertracing/jaeger/runs/1467144626?check_suite_focus=true#step:8:1046

> fatal: No names found, cannot describe anything.

.github/actions/export-branch/action.yml
```diff
@@ -0,0 +1,13 @@
+name: 'Export BRANCH variable'
+description: 'Makes BRANCH variable accessible to jobs'
+runs:
+  using: "composite"
+  steps:
+    - name: Export BRANCH variable
+      shell: bash
+      run: |
+        export BRANCH=${GITHUB_REF##*/}
```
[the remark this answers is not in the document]
Ashmita152: > If the branch is directly available, and is typically only used by a single step in the workflow (last step in this case), then why bother with an action instead of just export BRANCH=${GITHUB_REF##*/} inside the run: |?

You're right, in that case we don't need to define an action.

> why do you use ##*/? Wouldn't this work the same: export BRANCH=${GITHUB_REF}?

Generally GITHUB_REF is set as /ref/head/branch-name 
```
$ export GITHUB_REF=refs/head/foobar
$ echo ${GITHUB_REF##*/}
foobar
```

> GITHUB_REF is defined as "The branch or tag ref that triggered the workflow". I wonder if this will confuse any of the scripts.

I will look at older travis runs and see what was the value of BRANCH being evaluated there and make sure that we set the BRANCH variable to the same value.

Ashmita152:
> if we already know that we're running in the correct Node version, we do not need `nvm use`.

I just added this to the GitHub action. Not sure why it was triggering `nvm use`.
```
- uses: actions/setup-node@v1
  with:
    node-version: '10'
```

scripts/travis/build-all-in-one-image.sh
Ashmita152: sure

Ashmita152:
> In the logs the value of BRANCH seems to be merge

Ah sorry I missed it, I should have caught it. I will make sure it's getting populated with the same value as in travis.

Ashmita152:
> What is this error about? https://github.com/jaegertracing/jaeger/runs/1467144626?check_suite_focus=true#step:8:1046
> 
> > fatal: No names found, cannot describe anything.

Let me look into it more.

yurishkuro:
> I just added this to the GitHub action. Not sure why it was triggering `nvm use`.

Because in the current script `nvm use` is unconditional. I am suggesting to replace it with a check for Node version and fail if it does not meet the requirements

yurishkuro:
btw, I just added a workflow for the UI repo based on `setup-node@v2`, it's working fine: https://github.com/jaegertracing/jaeger-ui/pull/664

.github/actions/export-branch/action.yml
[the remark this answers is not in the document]
Ashmita152: Updated the PR removing action and doing things in workflow itself.

1) Setting BRANCH env as ${{ github.head_ref }} at job level which is set only in pull requests (https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#github-context)
2) Later setting BRANCH env if BRANCH is not already set (push to master) from GITHUB_REF.

scripts/travis/build-all-in-one-image.sh
Ashmita152: Updated the PR with check for correct version

Ashmita152:
Updated the PR using setup-node@v2 which works now. Also the BRANCH is populating properly now.

Only issue which is remaining is what if the contributor creates PR from their fork's master branch. We can solve this by:

1) Appending "pr-" infront of the branch name while setting BRANCH env variable in pull request.
2) Updating the current logic of checking whether to push to docker registry with checking whether ${{ github.repository_owner }} variable  equal to `jaegertracing`

scripts/travis/build-all-in-one-image.sh
Ashmita152: This is fixed.

Ashmita152:
> Only issue which is remaining is what if the contributor creates PR from their fork's master branch.

This has been resolved by setting an environment variable DOCKERHUB_LOGIN to true iff when docker login step succeed (which will not happen in case of PR). And docker push will only happen when DOCKERHUB_LOGIN is true.

scripts/travis/build-all-in-one-image.sh
```diff
@@ -8,8 +8,16 @@ BRANCH=${BRANCH:?'missing BRANCH env var'}
...
 
-source ~/.nvm/nvm.sh
-nvm use 10
+expected_version="v10"
+version=$(node --version)
+major_version=${version%.*.*}
+if [ "$major_version" = "$expected_version" ] ; then
+  echo "node version is as expected"
+else
+  echo "installed version $major_version doesn't match $expected_version"
```
yurishkuro, line 17:
```suggestion
  echo "ERROR: installed Node version $version doesn't match expected version $expected_version"
```

.github/workflows/all-in-one-build.yml
```diff
@@ -0,0 +1,50 @@
...
+
+    - name: Export DOCKERHUB_LOGIN variable
+      run: |
+        echo "DOCKERHUB_LOGIN=true" >> $GITHUB_ENV
+      if: steps.dockerhub-login.outcome == 'success'
+
+    - name: Export BRANCH variable if undefined
+      run: |
+        echo "BRANCH=${GITHUB_REF##*/}" >> $GITHUB_ENV
+      if: env.BRANCH == null
```
yurishkuro, line 44: when would this not be null?  It's not a std env var in GH actions: https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables#default-environment-variables

.github/workflows/all-in-one-build.yml
```diff
@@ -0,0 +1,50 @@
+name: Docker
```
yurishkuro, line 1:
```suggestion
name: Build all-in-one
```
due to https://github.com/jaegertracing/jaeger/issues/2645#issuecomment-735256696

yurishkuro: Later on I think we should merge this into the overall `Docker Build + Deploy` workflow. Building the UI is one of the longest parts there, and UI needs to be built for query service as well, so there's really no reason for all-in-one to be separate like that. But we can address it in the PR that does the rest of the builds.

scripts/travis/build-all-in-one-image.sh
```diff
@@ -8,8 +8,16 @@ BRANCH=${BRANCH:?'missing BRANCH env var'}
 # `GOARCH=<target arch> ./scripts/travis/build-all-in-one-image.sh`.
 GOARCH=${GOARCH:-$(go env GOARCH)}
 
-source ~/.nvm/nvm.sh
-nvm use 10
+expected_version="v10"
+version=$(node --version)
+major_version=${version%.*.*}
+if [ "$major_version" = "$expected_version" ] ; then
+  echo "node version is as expected"
```
yurishkuro, line 15:
```suggestion
  echo "Node version is as expected: $version"
```

yurishkuro (commented):
almost there

.github/workflows/all-in-one-build.yml
Ashmita152: First, we are setting BRANCH env variable in the job level

```
    env:
      BRANCH: ${{ github.head_ref }}
```

${{ github.head_ref }} will be only set in case of Pull requests (https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#github-context)

Hence when we push to master (merge PR), env.BRANCH will be null, in which case we are picking from GITHUB_REF.

Ashmita152:
> What is this error about? https://github.com/jaegertracing/jaeger/runs/1467144626?check_suite_focus=true#step:8:1046
> > fatal: No names found, cannot describe anything.

I think this is what is happening, in Makefile

build-all-in-one looks for BUILD_INFO which inturn looks for GIT_CLOSEST_TAG which isn't getting evaluated properly due to fetch-depth set to 1 while cloning the repo by GitHub action.

```
$ git clone https://github.com/Ashmita152/jaeger.git --depth 1
Cloning into 'jaeger'...
remote: Enumerating objects: 1243, done.
remote: Counting objects: 100% (1243/1243), done.
remote: Compressing objects: 100% (1050/1050), done.
remote: Total 1243 (delta 323), reused 485 (delta 129), pack-reused 0
Receiving objects: 100% (1243/1243), 1.10 MiB | 1.10 MiB/s, done.
Resolving deltas: 100% (323/323), done.
$ cd jaeger
$ git describe --abbrev=0 --tags
fatal: No names found, cannot describe anything.
```

yurishkuro:
> build-all-in-one looks for BUILD_INFO which in turn looks for GIT_CLOSEST_TAG

we'll need to fix this because BUILD_INFO is important to include in the binaries.

.github/workflows/all-in-one-build.yml
Ashmita152: Updated the PR to make the steps more self-explanatory.

Ashmita152:
> > build-all-in-one looks for BUILD_INFO which in turn looks for GIT_CLOSEST_TAG
> 
> we'll need to fix this because BUILD_INFO is important to include in the binaries.

This has been resolved.

.github/workflows/all-in-one-build.yml
```diff
@@ -0,0 +1,57 @@
...
+      if: env.DOCKERHUB_TOKEN != null
+
+    - name: Export DOCKERHUB_LOGIN variable
+      run: |
+        echo "DOCKERHUB_LOGIN=true" >> $GITHUB_ENV
+      if: steps.dockerhub-login.outcome == 'success'
+
+    - name: Export BRANCH variable for pull_request event
+      run: |
+        echo "BRANCH=${GITHUB_HEAD_REF}" >> $GITHUB_ENV
```
yurishkuro, line 45: it would be better for debugging to print the derived value
```suggestion
        export BRANCH=${GITHUB_HEAD_REF}
        echo "we are on branch=$BRANCH"
        echo "BRANCH=${BRANCH}" >> $GITHUB_ENV
```

yurishkuro (approved):
Look great! Minor potential future improvements:
  * echo value of BRANCH
  * move docker login steps and BRANCH steps into their own local actions (but only if we need to reuse them elsewhere)

Ashmita152:
Thank you @yurishkuro for the constant help.

After the PR I was expecting one dockerhub image upload after the PR was merged but something looks wrong somewhere in the script.

I can see env variables are populated correctly
```
2020-11-29T20:00:51.1750375Z env:
2020-11-29T20:00:51.1750881Z   GOROOT: /opt/hostedtoolcache/go/1.15.5/x64
2020-11-29T20:00:51.1751426Z   DOCKERHUB_LOGIN: true
2020-11-29T20:00:51.1751881Z   BRANCH: master
```
But somehow it still skipped the upload to dockerhub
```
2020-11-29T20:08:03.2662939Z + upload_to_docker jaegertracing/all-in-one
2020-11-29T20:08:03.2663826Z + [[ master == \m\a\s\t\e\r ]]
2020-11-29T20:08:03.2664409Z + [[ DOCKERHUB_LOGIN == \t\r\u\e ]]
2020-11-29T20:08:03.2665960Z + echo 'skip docker upload for PR'
```

I will look into it more.

PR #2664, opened 2020-11-28
Add instrumentation handler to collector endpoints

### LINKED ISSUE #2264: Expose metric for zipkin spans that could not be parsed

I would like to detect spans that are getting rejected with a 400BAD Request (see https://github.com/istio/istio/issues/24177) with prometheus metrics.

There is currently nothing on `:14269/metrics` which captures such a failure (in this case, a span tag that was not a string).

We build a platform which our product teams use Jaeger to debug, we would like to detect problems before they get raised to us as missing spans in the UI.

Expose a prometheus metric which tracks failed zipkin span reports, so we're able to alert on it.

dimitarvdimitrov:
Resolves #2264 

Currently, we do not expose any HTTP metrics on any of the collector endpoints. This makes it harder to detect when the collector is having issues ingesting bad spans for example.

This change adds a metric that reports the duration of each request and adds HTTP status and path labels to it.

Those metrics do not include the collector's gRPC endpoints.

### DIFF
--- a/cmd/collector/app/server/http_test.go
+++ b/cmd/collector/app/server/http_test.go
@@ -15,14 +15,12 @@
 package server
 
 import (
-	"fmt"
-	"net"
 	"net/http"
+	"net/http/httptest"
 	"testing"
 	"time"
 
 	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
 	"github.com/uber/jaeger-lib/metrics/metricstest"
 	"go.uber.org/zap"
 
@@ -31,7 +29,7 @@ import (
 )
 
 // test wrong port number
-func TestFailToListenHttp(t *testing.T) {
+func TestFailToListenHTTP(t *testing.T) {
 	logger, _ := zap.NewDevelopment()
 	server, err := StartHTTPServer(&HTTPServerParams{
 		HostPort: ":-1",
@@ -41,7 +39,7 @@ func TestFailToListenHttp(t *testing.T) {
 	assert.EqualError(t, err, "listen tcp: address -1: invalid port")
 }
 
-func TestSpanCollectorHttp(t *testing.T) {
+func TestSpanCollectorHTTP(t *testing.T) {
 	logger, _ := zap.NewDevelopment()
 	params := &HTTPServerParams{
 		Handler:        handler.NewJaegerSpanHandler(logger, &mockSpanProcessor{}),
@@ -51,17 +49,12 @@ func TestSpanCollectorHttp(t *testing.T) {
 		Logger:         logger,
 	}
 
-	listener, err := net.Listen("tcp", ":0")
-	require.NoError(t, err)
-	defer listener.Close()
-
-	server := &http.Server{Addr: listener.Addr().String()}
+	server := httptest.NewServer(nil)
 	defer server.Close()
 
-	serveHTTP(server, listener, params)
+	serveHTTP(server.Config, server.Listener, params)
 
-	url := fmt.Sprintf("http://%s", listener.Addr())
-	response, err := http.Post(url, "", nil)
+	response, err := http.Post(server.URL, "", nil)
 	assert.NoError(t, err)
 	assert.NotNil(t, response)
 }
--- a/cmd/collector/app/server/http_test.go
+++ b/cmd/collector/app/server/http_test.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"strings"
 	"testing"
 	"time"
 
@@ -51,17 +52,22 @@ func TestSpanCollectorHttp(t *testing.T) {
 		Logger:         logger,
 	}
 
-	server := &http.Server{Addr: ":12345"}
+	server := &http.Server{Addr: ":0"}
 	defer server.Close()
 
-	listener, err := net.Listen("tcp", ":12345")
+	listener, err := net.Listen("tcp", ":0")
 	require.NoError(t, err)
 	defer listener.Close()
 
 	serveHTTP(server, listener, params)
 
-	url := fmt.Sprintf("http://%s", server.Addr)
+	listenerPort := extractPort(listener.Addr().String())
+	url := fmt.Sprintf("http://%s", listenerPort)
 	response, err := http.Post(url, "", nil)
 	assert.NoError(t, err)
 	assert.NotNil(t, response)
+}
+
+func extractPort(addr string) string {
+	return addr[strings.LastIndexAny(addr, ":"):]
 }
--- a/cmd/collector/app/server/http_test.go
+++ b/cmd/collector/app/server/http_test.go
@@ -1,0 +1,67 @@
+// Copyright (c) 2020 The Jaeger Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package server
+
+import (
+	"fmt"
+	"net"
+	"net/http"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/uber/jaeger-lib/metrics/metricstest"
+	"go.uber.org/zap"
+
+	"github.com/jaegertracing/jaeger/cmd/collector/app/handler"
+	"github.com/jaegertracing/jaeger/pkg/healthcheck"
+)
+
+// test wrong port number
+func TestFailToListenHttp(t *testing.T) {
+	logger, _ := zap.NewDevelopment()
+	server, err := StartHTTPServer(&HTTPServerParams{
+		HostPort: ":-1",
+		Logger:   logger,
+	})
+	assert.Nil(t, server)
+	assert.EqualError(t, err, "listen tcp: address -1: invalid port")
+}
+
+func TestSpanCollectorHttp(t *testing.T) {
+	logger, _ := zap.NewDevelopment()
+	params := &HTTPServerParams{
+		Handler:        handler.NewJaegerSpanHandler(logger, &mockSpanProcessor{}),
+		SamplingStore:  &mockSamplingStore{},
+		MetricsFactory: metricstest.NewFactory(time.Hour),
+		HealthCheck:    healthcheck.New(),
+		Logger:         logger,
+	}
+
+	server := &http.Server{Addr: ":12345"}
+	defer server.Close()
+
+	listener, err := net.Listen("tcp", ":12345")
+	require.NoError(t, err)
+	defer listener.Close()
+
+	serveHTTP(server, listener, params)
+
+	url := fmt.Sprintf("http://%s", server.Addr)
+	response, err := http.Post(url, "", nil)
+	assert.NoError(t, err)
+	assert.NotNil(t, response)
+}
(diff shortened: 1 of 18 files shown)


dimitarvdimitrov:
Should the codecov/patch check be passing? I can't see exactly the reason for it failing, but I'm guessing it's because of the code added in init functions, which aren't tested. Is this a problem?

Also, [`TestReload`](https://github.com/jaegertracing/jaeger/blob/88e9a70814cab98ce6fd389adf7eda13fe9c9416/pkg/config/tlscfg/cert_watcher_test.go#L45) looks like a flaky one: for `274c17c` it passed on [my fork](https://github.com/dimitarvdimitrov/jaeger/runs/1467932228), but failed [on this PR](https://github.com/jaegertracing/jaeger/runs/1467898828),

pkg/instrumentedhandler/metrics.go
```diff
@@ -0,0 +1,63 @@
+// Copyright (c) 2019 The Jaeger Authors.
+// Copyright (c) 2017 Uber Technologies, Inc.
```
yurishkuro, line 2:
```suggestion
// Copyright (c) 2020 The Jaeger Authors.
```

yurishkuro (commented):
Yes, please try to make sure codecov checks are passing, because we've already fallen to a lever barely above the threshold. I am not sure what you mean by init functions, we tend to write all code as testable (i.e. not using things like package-level init).

TestReload is indeed flaky, we have a ticket open for it.

pkg/instrumentedhandler/metrics.go
```diff
@@ -0,0 +1,63 @@
...
+// It will record the HTTP response status, HTTP method, duration and path of the call.
+// The duration will be reported in metrics.Timer and the rest will be labels on that timer.
+func NewMetricsHandler(metricsFactory metrics.Factory) func(http.Handler) http.Handler {
+	return func(h http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			start := time.Now()
+			recorder := &statusRecorder{ResponseWriter: w}
+			h.ServeHTTP(recorder, r)
+
+			requestDuration := metricsFactory.Timer(metrics.TimerOptions{
```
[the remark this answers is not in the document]
dimitarvdimitrov: I only looked at the prometheus implementation of the factory and saw that it caches those, but didn't see the other ones do not. Will add similar caching to that in `collector/app/metrics.go`

pkg/httpmetrics/metrics.go
```diff
@@ -0,0 +1,135 @@
...
+			timer = buildTimer(r.metrics, request)
+			r.timers[cacheKey] = timer
+		}
+		r.lock.Unlock()
+	}
+
+	timer.Record(request.duration)
+}
+
+func (r *requestDurations) cacheKey(request recordedRequest) string {
```
yurishkuro, line 111: This is not necessary. When a struct contains only primitive values, it can be used as a map key directly. Just define a separate struct containing method/path/status (can embed it into recordedRequest).

pkg/httpmetrics/metrics_test.go
```diff
@@ -0,0 +1,50 @@
...
+	})
+
+	mb := metricstest.NewFactory(time.Hour)
+	handler := Wrap(dummyHandlerFunc, mb)
+
+	req, err := http.NewRequest(http.MethodGet, "/subdir/qwerty", nil)
+	assert.NoError(t, err)
+	handler.ServeHTTP(httptest.NewRecorder(), req)
+
+	for i := 0; i < 15; i++ {
```
yurishkuro, line 41: I would recommend s/15/1000. When running in shared CI env, the process can easily be CPU-starved for 15ms

pkg/httpmetrics/metrics.go
```diff
@@ -0,0 +1,135 @@
...
+		return
+	}
+	r.status = status
+	r.wroteHeader = true
+	r.ResponseWriter.WriteHeader(status)
+}
+
+// Wrap returns a handler that wraps the provided one and emits metrics based on the HTTP requests and responses.
+// It will record the HTTP response status, HTTP method, duration and path of the call.
+// The duration will be reported in metrics.Timer and the rest will be labels on that timer.
```
yurishkuro, line 46: May want to add a warning: 
```
//
// Do not use with HTTP endpoints that take parameters from URL path, such as `/user/{user_id}`, 
// because they will result in high cardinality metrics.
```

cmd/collector/app/server/zipkin_test.go
```diff
@@ -0,0 +1,66 @@
...
+func TestSpanCollectorZipkin(t *testing.T) {
+	logger, _ := zap.NewDevelopment()
+	params := &ZipkinServerParams{
+		Handler:        handler.NewZipkinSpanHandler(logger, nil, nil),
+		MetricsFactory: metricstest.NewFactory(time.Hour),
+		HealthCheck:    healthcheck.New(),
+		Logger:         logger,
+	}
+
+	server := &http.Server{Addr: ":12345"}
```
yurishkuro, line 53: it's preferable to run test servers on ephemeral ports, e.g. `&http.Server{Addr: ":0"}`

yurishkuro (commented):
overall lgtm, some minor improvements suggested

pkg/httpmetrics/metrics_test.go
dimitarvdimitrov: > s/15/1000

Sorry, what do you mean by that? Try 15 times, with 1000 ms between each?

Also, don't unit tests run one-by-one?

yurishkuro: sorry, s/15/1000 means replace 15->1000 (regex syntax), i.e. 1000 @ 1ms. 10ms delay is also fine.

cmd/collector/app/server/http_test.go
```diff
@@ -0,0 +1,73 @@
...
+	logger, _ := zap.NewDevelopment()
+	params := &HTTPServerParams{
+		Handler:        handler.NewJaegerSpanHandler(logger, &mockSpanProcessor{}),
+		SamplingStore:  &mockSamplingStore{},
+		MetricsFactory: metricstest.NewFactory(time.Hour),
+		HealthCheck:    healthcheck.New(),
+		Logger:         logger,
+	}
+
+	server := &http.Server{Addr: ":0"}
```
yurishkuro, line 55: Shouldn't this be created _after_ the listener with `Addr: listener.Addr().String()`? Otherwise it's confusing, looks like two different ports are being used.

cmd/collector/app/server/zipkin_test.go
```diff
@@ -0,0 +1,67 @@
...
+	defer server.Close()
+
+	listener, err := net.Listen("tcp", ":0")
+	require.NoError(t, err)
+	defer listener.Close()
+
+	serveZipkin(server, listener, params)
+
+	listenerPort := extractPort(listener.Addr().String())
+	url := fmt.Sprintf("http://%s", listenerPort)
```
yurishkuro, line 63: why do you need to specifically extract the port, instead of just `"http://" + listener.Addr().String()`?

pkg/httpmetrics/metrics.go
```diff
@@ -0,0 +1,121 @@
...
+
+type requestDurations struct {
+	lock    *sync.RWMutex
+	metrics metrics.Factory
+	timers  map[recordedRequestKey]metrics.Timer
+}
+
+func newRequestDurations(metricsFactory metrics.Factory) *requestDurations {
+	return &requestDurations{
+		timers:  map[recordedRequestKey]metrics.Timer{},
```
yurishkuro, line 86:
```suggestion
		timers:  make(map[recordedRequestKey]metrics.Timer),
```

pkg/httpmetrics/metrics.go
```diff
@@ -0,0 +1,121 @@
...
+		metrics: metricsFactory,
+		lock:    &sync.RWMutex{},
```
yurishkuro, line 88: since requestDurations is used via pointer, the lock field does not need to be a pointer

cmd/collector/app/server/zipkin_test.go
dimitarvdimitrov: hm, I thought I had problems with that because it returns an ipv6 address (`[::]:12345`), but it actually works ok

pkg/httpmetrics/metrics.go
dimitarvdimitrov: Is there a reason for this other than consistency/style?

yurishkuro: It reads more intuitive with a make command, the empty brackets are harder to notice.

cmd/collector/app/server/zipkin_test.go
```diff
@@ -0,0 +1,66 @@
...
+	listener, err := net.Listen("tcp", ":0")
+	require.NoError(t, err)
+	defer listener.Close()
+
+	server := &http.Server{Addr: listener.Addr().String()}
+	defer server.Close()
+
+	serveZipkin(server, listener, params)
+
+	url := fmt.Sprintf("http://%s", listener.Addr().String())
```
yurishkuro, line 62: nit: `.String()` is not necessary, since it means Addr() returns a Stringer and %s format will use that automatically.

Alternatively, you could use server.Addr, which is semantically more relevant in this context.

pkg/httpmetrics/metrics.go
```diff
@@ -0,0 +1,121 @@
...
+	lock    sync.RWMutex
+	metrics metrics.Factory
+	timers  map[recordedRequestKey]metrics.Timer
+}
+
+func newRequestDurations(metricsFactory metrics.Factory) *requestDurations {
+	return &requestDurations{
+		timers:  make(map[recordedRequestKey]metrics.Timer),
+		metrics: metricsFactory,
+		lock:    sync.RWMutex{},
```
yurishkuro, line 88: there's no need to init this

jpkrohling (commented):
Looks nice. This doesn't (yet) add metrics to any existing endpoint, right?

dimitarvdimitrov:
> Looks nice. This doesn't (yet) add metrics to any existing endpoint, right?

It does - to collector's `/api/v1/spans`, `/api/v2/spans` and `/api/traces` endpoints

cmd/collector/app/server/http_test.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/stretchr/testify/require"
+	"github.com/uber/jaeger-lib/metrics/metricstest"
+	"go.uber.org/zap"
+
+	"github.com/jaegertracing/jaeger/cmd/collector/app/handler"
+	"github.com/jaegertracing/jaeger/pkg/healthcheck"
+)
+
+// test wrong port number
+func TestFailToListenHttp(t *testing.T) {
```
jpkrohling, line 34:
```suggestion
func TestFailToListenHTTP(t *testing.T) {
```

cmd/collector/app/server/http_test.go
```diff
@@ -0,0 +1,67 @@
...
+	logger, _ := zap.NewDevelopment()
+	server, err := StartHTTPServer(&HTTPServerParams{
+		HostPort: ":-1",
+		Logger:   logger,
+	})
+	assert.Nil(t, server)
+	assert.EqualError(t, err, "listen tcp: address -1: invalid port")
+}
+
+func TestSpanCollectorHttp(t *testing.T) {
```
jpkrohling, line 44:
```suggestion
func TestSpanCollectorHTTP(t *testing.T) {
```

cmd/collector/app/server/http_test.go
```diff
@@ -0,0 +1,67 @@
...
+		HealthCheck:    healthcheck.New(),
+		Logger:         logger,
+	}
+
+	listener, err := net.Listen("tcp", ":0")
+	require.NoError(t, err)
+	defer listener.Close()
+
+	server := &http.Server{Addr: listener.Addr().String()}
+	defer server.Close()
```
jpkrohling, line 59: I might be missing something, but is there a reason to not use `httptest.Server` instead? It should save you from dealing with the listener yourself.

dimitarvdimitrov: really good idea, fixed in https://github.com/jaegertracing/jaeger/pull/2664/commits/00ff54bd38f562163f136300a806d01581bcf1b7

yurishkuro (approved):
Thanks!

PR #2667, opened 2020-11-30
Fix and minor improvements to all-in-one github action

Ashmita152:

This add fix and improvements suggested in https://github.com/jaegertracing/jaeger/pull/2663

### DIFF
--- a/.github/workflows/all-in-one-build.yml
+++ b/.github/workflows/all-in-one-build.yml
@@ -42,7 +42,9 @@ jobs:
 
     - name: Export BRANCH variable for pull_request event
       run: |
-        echo "BRANCH=${GITHUB_HEAD_REF}" >> $GITHUB_ENV
+        export BRANCH=${GITHUB_HEAD_REF}
+        echo "we are on branch=$BRANCH"
+        echo "BRANCH=${BRANCH}" >> $GITHUB_ENV
       if: github.event_name == 'pull_request'
 
     - name: Export BRANCH variable for push event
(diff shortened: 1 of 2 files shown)


yurishkuro (approved):

.github/workflows/all-in-one-build.yml
```diff
@@ -42,7 +42,9 @@ jobs:
 
     - name: Export BRANCH variable for pull_request event
       run: |
-        echo "BRANCH=${GITHUB_HEAD_REF}" >> $GITHUB_ENV
+        export BRANCH=${GITHUB_HEAD_REF}
+        echo "we are on branch=$BRANCH"
+        echo "BRANCH=${BRANCH}" >> $GITHUB_ENV
```
yurishkuro, line 47: The next step could also echo the value

Ashmita152: sure, I will do that while moving this to its own action as you suggested. Since it can be shared with other docker build and push jobs for all jaeger components.

yurishkuro:
the build failed on master https://github.com/jaegertracing/jaeger/runs/1471508769?check_suite_focus=true
```
+ docker login docker.io -u -p
/home/runner/.docker/config.json doesn't exist
```

Ashmita152:
Hi @yurishkuro 

Do you recommend to remove the docker login command from the upload-to-docker.sh script completely.

yurishkuro:
I doubt that we can remove it completely, without credentials we cannot upload the images.

Ashmita152:
> I doubt that we can remove it completely, without credentials we cannot upload the images.

Sorry for my limited knowledge. By unable to upload the images, do you mean other jaeger components images ?

yurishkuro:
nvm, once I saw the PR I see what you meant by removing login - the explicit login in the script should be removed.

PR #2668, opened 2020-11-30
Fix docker login issue with all-in-one build

Ashmita152:

Remove docker login command since that is being done as part of GitHub Actions.

Also remove the extra check for deleting the config.json. I think it was done to prevent any previous docker login session in the same travis runner machine.

### DIFF
--- a/scripts/travis/upload-to-docker.sh
+++ b/scripts/travis/upload-to-docker.sh
@@ -31,15 +31,6 @@ if [[ -n $major ]]; then
   fi
 fi
 
-if [[ -f $HOME/.docker/config.json ]]; then
-  rm -f $HOME/.docker/config.json
-else
-  echo "$HOME/.docker/config.json doesn't exist"
-fi
-
-# Do not enable echo before the `docker login` command to avoid revealing the password.
-set -x
-docker login docker.io -u $DOCKER_USER -p $DOCKER_PASS
 if [[ "${REPO}" == "jaegertracing/jaeger-opentelemetry-collector" || "${REPO}" == "jaegertracing/jaeger-opentelemetry-agent" || "${REPO}" == "jaegertracing/jaeger-opentelemetry-ingester" || "${REPO}" == "jaegertracing/opentelemetry-all-in-one" ]]; then
   # TODO remove once Jaeger OTEL collector is stable
   docker push $REPO:latest
--- a/scripts/travis/upload-to-docker.sh
+++ b/scripts/travis/upload-to-docker.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-set -e
+set -euxf -o pipefail
 
 BRANCH=${BRANCH:?'missing BRANCH env var'}
 IMAGE="${REPO:?'missing REPO env var'}:latest"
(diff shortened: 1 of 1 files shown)


scripts/travis/upload-to-docker.sh
```diff
@@ -31,15 +31,6 @@ if [[ -n $major ]]; then
...
 
-if [[ -f $HOME/.docker/config.json ]]; then
-  rm -f $HOME/.docker/config.json
-else
-  echo "$HOME/.docker/config.json doesn't exist"
-fi
-
-# Do not enable echo before the `docker login` command to avoid revealing the password.
-set -x
-docker login docker.io -u $DOCKER_USER -p $DOCKER_PASS
```
yurishkuro, line 42: +1, but please keep `set -x`. In fact, you can add it to L3 as `set -euxf -o pipefail`

scripts/travis/upload-to-docker.sh
```diff
@@ -31,15 +31,6 @@ if [[ -n $major ]]; then
   fi
```
yurishkuro, line 31: do you know the meaning of substitution in L19?

scripts/travis/upload-to-docker.sh
Ashmita152: sure.

scripts/travis/upload-to-docker.sh
Ashmita152: Not sure, but I will try it and update the comment accordingly.

yurishkuro (approved):

scripts/travis/upload-to-docker.sh
Ashmita152: I think long back BRANCH was evaluated as `/refs/head/branch-name` and hence it was done to remove `/` from branch name since that will result in docker tag:

```
$ a="/refs/head/foobar"
$ echo ${a///}
refsheadfoobar
```

And `/` isn't allowed in docker tag:

```
$ docker tag jaegertracing/jaeger-ingester jaegertracing/jaeger-ingester:testing/testing
Error parsing reference: "jaegertracing/jaeger-ingester:testing/testing" is not a valid repository/tag: invalid reference format
$ docker tag jaegertracing/jaeger-ingester jaegertracing/jaeger-ingester:testing
$
```

Ashmita152:
I think long back BRANCH was evaluated as `/refs/head/branch-name` and hence it was done to remove `/` from branch name since that will result in docker tag:

```
$ a="/refs/head/foobar"
$ echo ${a///}
refsheadfoobar
```

And `/` isn't allowed in docker tag:

```
$ docker tag jaegertracing/jaeger-ingester jaegertracing/jaeger-ingester:testing/testing
Error parsing reference: "jaegertracing/jaeger-ingester:testing/testing" is not a valid repository/tag: invalid reference format
$ docker tag jaegertracing/jaeger-ingester jaegertracing/jaeger-ingester:testing
$
```

yurishkuro:
Do you have a handy doc link on the substitution syntax?

Ashmita152:
@yurishkuro I just played around with bash

Here we are replacing `refs` with `test`
```
$ echo $a
/refs/head/foobar
$ echo ${a/ref/test}
/test/head/foobar
$
``` 

My understanding of why three backslash:

- First backslash signifies that we are going to do string substitution.
- Second blackslash is to escape third backslash.
- Third backslash is the character to replace.
- No character after means replacing with empty string.

yurishkuro:
bummer, build failed again
```
./scripts/travis/upload-to-docker.sh: line 20: TRAVIS_BRANCH: unbound variable
```

Ashmita152:
It was a great idea to add `set -u` It caught it.

PR #2671, opened 2020-11-30
Add github action for elasticsearch otel integration tests

Ashmita152:

Adds integration test for Elasticsearch OpenTelemetry.

Also refactors es-integration-test.sh to accept arg either "default" or "otel" otherwise fails.


.github/workflows/es-otel-integration-tests.yml
```diff
@@ -0,0 +1,23 @@
...
+
+    - uses: actions/setup-go@v2
+      with:
+        go-version: ^1.15
+
+    - name: Install tools
+      run: make install-ci
+
+    - name: Run elasticsearch otel integration tests
+      run: bash scripts/travis/es-integration-test.sh otel
```
yurishkuro, line 23: It may be worth rethinking this test. The script internally runs 3 tests with different S versions
```
run_integration_test "5.6.16"
run_integration_test "6.8.2"
run_integration_test "7.3.0"
```
They can be easily run as different jobs in the same workflow, with the same granularity of restart as today. And afair starting ES is what takes the most time, not the tests themselves, so it would be much more efficient if we ran both default and otel tests sequentially after a given ES is started.

Ashmita152: Good point. Sure will convert to separate jobs.

scripts/travis/es-integration-test.sh
```diff
@@ -1,39 +1,103 @@
 #!/bin/bash
 
-set -e
+set -euxf -o pipefail
 
-GOARCH=$(go env GOARCH)
+usage() {
+  echo $"Usage: $0 (default|otel)"
```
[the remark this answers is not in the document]
Ashmita152: This is me copy pasting from internet :) I think it's not required actually.

.github/workflows/es-otel-integration-tests.yml
Ashmita152: Hi @yurishkuro 

I have updated the PR splitting each into different jobs. Thank you.

.github/workflows/ci-elasticsearch-otel.yml
```diff
@@ -0,0 +1,53 @@
+name: CIT OTEL Elasticsearch
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  elasticsearch-otel-5:
```
yurishkuro, line 10: could we use matrix for this instead of repeating job specs?

Ashmita152: Ah right. Sorry I completely forgot about it.

.github/workflows/ci-elasticsearch-otel.yml
```diff
@@ -0,0 +1,53 @@
...
+
+    - uses: actions/setup-go@v2
+      with:
+        go-version: ^1.15
+
+    - name: Install tools
+      run: make install-ci
+
+    - name: Run elasticsearch otel integration tests
+      run: bash scripts/travis/es-integration-test.sh otel 6.8.2
```
yurishkuro, line 38: can we refactor the script so that default and otel are run one after another? Starting up ES is what takes the most time iirc.

Ashmita152: Do you mean merging `workflows/ci-elasticsearch-otel.yml` and `workflows/ci-elasticsearch-otel.yml` into one ?

yurishkuro: that too, but more importantly not passing default/otel to the script and have it execute both automatically

Ashmita152: I see.

Ashmita152: I am not sure if something remaining in ES from `default`'s integration test run will affect `otel` but I can give it a try.

Ashmita152:
Hi @yurishkuro 

Updated the PR with the feedback. Thank you.

.github/workflows/ci-elasticsearch.yml
```diff
@@ -9,6 +9,9 @@ on:
 jobs:
   elasticsearch:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        version: [5.6.16, 6.8.2, 7.3.0]
```
yurishkuro, line 14: a bit of future proofing - let's define matrix as two values per iteration, a full version (5.6.16) and major-only version (e.g. 5.x), and then define `name` for the job using only the major version (example: https://futurestud.io/tutorials/github-actions-customize-the-job-name). Otherwise the jobs get names like `elasticsearch (5.6.16)`, which must match exactly when configuring the required checks (if we bump the version we'd need to re-configure the repo).

scripts/travis/es-integration-test.sh
```diff
@@ -1,39 +1,86 @@
...
+  echo "Executing token propagation test"
+  run_token_propagation_test
+}
 
-make test-compile-es-scripts
-SPAN_STORAGE_TYPE=elasticsearch ./cmd/query/query-linux-$GOARCH --es.server-urls=http://127.0.0.1:9200 --es.tls=false --es.version=7 --query.bearer-token-propagation=true &
-PID=$(echo $!)
-make token-propagation-integration-test
-kill -9 ${PID}
+main "$@"
```
yurishkuro, line 86: very neat organization of the script!

scripts/travis/es-integration-test.sh
```diff
@@ -1,39 +1,86 @@
...
+  GOOS=linux make build-query
+}
+
+run_integration_test() {
+  local es_version=$1
+  local cid=$(setup_es ${es_version})
+  STORAGE=elasticsearch make storage-integration-test
+  make index-cleaner-integration-test
+  make es-otel-exporter-integration-test
+  teardown_es ${cid}
```
yurishkuro, line 65: Doesn't token propagation test need a running ES? you seem to be stopping it prior to that.

Maybe ES setup/teardown should be in main?

yurishkuro: based on the logs the ES container is definitely being killed before the token test, so how does it pass? does it not need storage?

yurishkuro (commented):
Refactoring looks great!

I'm seeing that error again: `fatal: No names found, cannot describe anything.` https://github.com/jaegertracing/jaeger/pull/2671/checks?check_run_id=1477915552#step:5:1338

scripts/travis/es-integration-test.sh
Ashmita152: Looking at the older version of the script we kill the ES container before token propagation test. I haven't looked into the details of token propagation test to see what it does though. I will check that to make sure.

scripts/travis/es-integration-test.sh
Ashmita152: thank you!

.github/workflows/ci-elasticsearch.yml
Ashmita152: That's a great idea. Is to make sure that we always test against the latest stable version of ES for each major version ?

Ashmita152:
> I'm seeing that error again: fatal: No names found, cannot describe anything.

Nice catch.

.github/workflows/ci-elasticsearch.yml
yurishkuro: Well, I assume at some point we may want to change the exact version being tested, like go from 5.6.16 to 5.7 or so. Right now, if we did that, that would change the job name and leave a required check that could never be satisfied. So it's better to have stable job names.

scripts/travis/es-integration-test.sh
Ashmita152: It looks like token propagation test mocks ES calls and requires only jaeger-query to be available.

yurishkuro (approved):

PR #2657, opened 2020-11-26
Add metrics on the agent that can show connections status between agent with collections

WalkerWang731:
## Which problem is this PR solving?
- Resolves #2656 

## Short description of the changes
- Add `cmd/agent/app/reporter/connect_metrics.go`, used for add new metrics that connections status between agent with collection
- Add `cmd/agent/app/reporter/connect_metrics_test.go`, used for unit test to check func in `cmd/agent/app/reporter/connect_metrics.go`
- Adjust `cmd/agent/app/reporter/grpc/builder.go` func `CreateConnection()`, used for support change metrics as gRPC is changed
- Adjust `cmd/agent/app/reporter/grpc/collector_proxy.go` func `NewCollectorProxy()`, used for `CreateConnection()` newly added parameter requirements
- Adjust `cmd/agent/app/reporter/grpc/builder_test.go`, used for unit test to check func in `cmd/agent/app/reporter/grpc/builder.go` and newly added parameter requirements

## Changelog
Add metrics on the agent that can show connections status between the agent with collections

PR #2675, opened 2020-12-01
Fix the unbound variable error on all-in-one image build

Ashmita152:

When someone creates a PR from fork's master branch, first part of the if condition succeeds and then it looks for DOCKERHUB_LOGIN variable which is only set when we push to master (ie. not for pr) here: https://github.com/jaegertracing/jaeger/blob/master/.github/workflows/ci-all-in-one-build.yml#L38-L41
This is causing unbound variable error. Setting it default to `false` will help in solving the error.

This is currently happening for this PR: https://github.com/jaegertracing/jaeger/pull/2657

### DIFF
--- a/scripts/travis/build-all-in-one-image.sh
+++ b/scripts/travis/build-all-in-one-image.sh
@@ -7,6 +7,7 @@ BRANCH=${BRANCH:?'missing BRANCH env var'}
 # be overrided by passing architecture value to the script:
 # `GOARCH=<target arch> ./scripts/travis/build-all-in-one-image.sh`.
 GOARCH=${GOARCH:-$(go env GOARCH)}
+DOCKERHUB_LOGIN=${DOCKERHUB_LOGIN:-false}
 
 expected_version="v10"
 version=$(node --version)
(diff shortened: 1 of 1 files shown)


scripts/travis/build-all-in-one-image.sh
```diff
@@ -36,7 +36,7 @@ run_integration_test() {
 
 upload_to_docker() {
   # Only push the docker container to Docker Hub for master branch
-  if [[ ("$BRANCH" == "master" || $BRANCH =~ ^v[0-9]+\.[0-9]+\.[0-9]+$) && "$DOCKERHUB_LOGIN" == "true" ]]; then
+  if [[ ("$BRANCH" == "master" || $BRANCH =~ ^v[0-9]+\.[0-9]+\.[0-9]+$) && "${DOCKERHUB_LOGIN:-false}" == "true" ]]; then
```
yurishkuro, line 39: I'd say it's preferable to handle default values at the top of the script, similar to BRANCH and GOARCH, i.e. add
```
DOCKERHUB_LOGIN=${DOCKERHUB_LOGIN:-false}
```

Ashmita152: nice suggestion, updated the PR with the change.

yurishkuro (approved):

PR #2677, opened 2020-12-02
Bump opentelemetry to v0.16.0

pkositsyn:
Signed-off-by: Pavel Kositsyn <kositsyn.pa@phystech.edu>

Updates opentelemetry dependency to new version.

Remove some of deprecated `InitEmpty` calls with `IsNil` checks.


pkositsyn:
Seems like Jaeger config file doesn't have any effect in opentelemetry agent. Is it an expected behaviour? See https://github.com/jaegertracing/jaeger/blob/4d75d3c56b667f3d11377643a2a141a48fe2f749/cmd/opentelemetry/cmd/agent/main.go#L53-L79
viper isn't used in `cfgConfig.DefaultConfigFactory(v)`

pkositsyn:
I see dependencies error for python3.9 in CIT Elasticsearch 5.x job

cmd/opentelemetry/app/defaultconfig/default_config_test.go
```diff
@@ -200,7 +200,10 @@ func TestService(t *testing.T) {
 			test.cfg.Factories = factories
 			createDefaultConfig := test.cfg.DefaultConfigFactory(v)
 
-			cfg, err := createDefaultConfig(viper.New(), factories)
+			// command set flag is parsed into otelViper
+			otelViper := viper.New()
+			command.Flags().StringArray("set", []string{}, "set overrides settings in otelcol")
```
jpkrohling, line 205: Do we use `otelcol` in the other options? I'd prefer to use the full name here ("OpenTelemetry Collector") if we don't use "otelcol" yet in the user interface.

pkositsyn: I'll fix that

joe-elliott:
> Seems like Jaeger config file doesn't have any effect in opentelemetry agent. Is it an expected behaviour? See

It is complex, but I believe it does actually load the config file properly.  Either way we are redoing configuration and I expect this to disappear.

Thank you for bringing up the concern!

pkositsyn:
Yeah, I must have missed something about config load. Can we fix python3.9 for jobs success?

pkositsyn:
Thanks @yurishkuro!

yurishkuro (approved):

PR #2681, opened 2020-12-05
Add gh action for crossdock and crossdock otel ci jobs

Ashmita152:

Since each takes around 10 minutes, keeping them as separate jobs to let run in parallel.

### DIFF
--- a/.github/workflows/ci-crossdock.yml
+++ b/.github/workflows/ci-crossdock.yml
@@ -44,17 +44,8 @@ jobs:
         echo "DOCKERHUB_LOGIN=true" >> $GITHUB_ENV
       if: steps.dockerhub-login.outcome == 'success'
 
-    - name: Export BRANCH variable for pull_request event
-      run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/ci-crossdock.yml
```diff
@@ -0,0 +1,67 @@
...
+
+    - name: Fetch git tags
+      run: |
+        git fetch --prune --unshallow --tags
+
+    - uses: actions/setup-go@v2
+      with:
+        go-version: ^1.15
+
+    - uses: docker/login-action@v1
```
yurishkuro, line 33: we're reusing login and branch logic already, can we move them into reusable actions?

Ashmita152: I don't think GitHub Actions support calling one action from another action so docker login part isn't possible to move yet but I have moved the logic for populating BRANCH env variable into separate action script.

yurishkuro (approved):
🥳

PR #2684, opened 2020-12-10
Add status command

sniperking1234:
Signed-off-by: Chen Zhengwei <chenzhengwei@inspur.com>

- Resolves https://github.com/jaegertracing/jaeger/issues/2361

- Add status command. 
Usage: `<command> status` checks the localhost and the port of the corresponding component.
`<command> status --status.http.host-port` checks for a specific port or URL.

### DIFF
--- a/cmd/status/command_test.go
+++ b/cmd/status/command_test.go
@@ -63,3 +63,10 @@ func TestUnready(t *testing.T) {
 	err := cmd.Execute()
 	assert.Error(t, err)
 }
+
+func TestNoService(t *testing.T) {
+	v := viper.New()
+	cmd := Command(v, 12345)
+	err := cmd.Execute()
+	assert.Error(t, err)
+}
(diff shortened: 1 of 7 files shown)


cmd/status/command_test.go
```diff
@@ -0,0 +1,72 @@
...
+	err := cmd.Execute()
+	assert.NoError(t, err)
+}
+
+func TestOnlyPortConfig(t *testing.T) {
+	ts := httptest.NewServer(http.HandlerFunc(readyHandler))
+	defer ts.Close()
+	v := viper.New()
+	cmd := Command(v, 80)
+	cmd.ParseFlags([]string{"--status.http.host-port=" + strings.TrimPrefix(ts.URL, "http://127.0.0.1")})
```
yurishkuro, line 52: not sure if you're guaranteed to get URL having the exact 127.0.0.1 host (e.g. you could get IPv6 version instead). I'd rather read the port number as the part after the last `:`

yurishkuro (commented):
look great, aside from a small nit in the test!

cmd/status/command_test.go
sniperking1234: Done.

yurishkuro (approved):
Thanks!

PR #2689, opened 2020-12-11
Remove Question issue type and add Discussions

yurishkuro:


joe-elliott (approved):
nit:  "Stack Overflow"

but LGTM!

Issue #123, opened 2017-04-21
REST API for collecting spans


pavolloffay:
REST API will be necessary for collecting span from a web browser.

see https://github.com/uber/jaeger-client-node/issues/109

badiib:
doing this now for thrift format with endpoint:

/api/traces?format=Jaeger.thrift as POST

badiib:
See https://github.com/uber/jaeger/pull/156

yurishkuro:
I don't think this is fully solved. For Browser use case we need JSON endpoint. The current binary thrift endpoint is something we need for blackbox monitoring

badiib:
Sounds good, doing this next:

/api/traces?format=Jaeger.json as POST

black-adder:
curious, why are we doing the "Jaeger" prefix for the format? Are we planning on supporting non-jaeger formats in the long distant future?

badiib:
@black-adder i'd prefer to keep that open as a possibility, yes.

badiib:
assigning to vprithvi as well, as he is doing some work on http span submissions.

pavolloffay:
Done, implemented HTTP endpoint for collecting data in jaeger.thrift format and now we have also zipkin `v1/spans` endpoint.

PR #2707, opened 2020-12-29
Add ability to use JS file for UI configuration (#123 from jaeger-ui)

th3M1ke:
- resolves https://github.com/jaegertracing/jaeger-ui/issues/123

- Added ability to use `javascript` file for UI configuration


cmd/query/app/static_handler.go
```diff
@@ -106,14 +107,10 @@ func loadAndEnrichIndexHTML(open func(string) (http.File, error), options Static
...
+	if configBytes, err := loadUIConfig(options.UIConfigPath); err != nil {
 		return nil, err
-	} else if config != nil {
-		// TODO if we want to support other config formats like YAML, we need to normalize `config` to be
-		// suitable for json.Marshal(). For example, YAML parser may return a map that has keys of type
-		// interface{}, and json.Marshal() is unable to serialize it.
-		bytes, _ := json.Marshal(config)
-		configString = fmt.Sprintf("JAEGER_CONFIG = %v", string(bytes))
+	} else if configBytes != nil {
+		configString = fmt.Sprintf("JAEGER_CONFIG = %v", string(configBytes))
```
yurishkuro, line 113: This doesn't look right. In case of JS file that contains a function there needs to be a part of the template that invokes that function and assigns its _result_ to JAEGER_CONFIG ( or something similar).

cmd/query/app/static_handler.go
```diff
@@ -210,30 +207,45 @@ func loadIndexHTML(open func(string) (http.File, error)) ([]byte, error) {
...
 	var c map[string]interface{}
+	var r []byte
 	var unmarshal func([]byte, interface{}) error
 
 	switch strings.ToLower(ext) {
 	case ".json":
 		unmarshal = json.Unmarshal
+	case ".js":
+		r = bytes.TrimSpace(bytesConfig)
+		if !bytes.HasPrefix(r, []byte("function")) {
```
yurishkuro, line 229: I would suggest to check for string "function UIConfig()" if you want some kind of validation, not the prefix, because there may be comments etc.

yurishkuro (commented):
Good start!

cmd/query/app/static_handler.go
th3M1ke: I would suggest taking a look at this line from the perspective of invoking the JS config file itself. 

My idea is next:
- we add a more strict JS file format validation like `/^function\s[a-zA-Z$_][a-zA-Z0-9$_]+(\s)?\(\)(\s)?{(.*\n)+}$/gi` - in this case we can rely on JS file and assume, that its plain JS function like
```
function LoadCodnfig(){
    console.log('Hello World!');
    alert('Im in Jaeger UI!');

    console.log('asdasd');

    return {
        callMe: () => {
            return Promise.resolve(setTimeout(() => {
                console.log('CALL ME!!!', 1000);
            }))
        },
        asd: 'sdf'
    }
}
```
- next, we add `()` at the moment of injecting config inside `index.html` and we have IIFE inside `index.html`
- in `jaeger-ui` we will add a cache for `getConfig()` like
```
diff --git a/packages/jaeger-ui/src/utils/config/get-config.tsx b/packages/jaeger-ui/src/utils/config/get-config.tsx
index 458ba0c..68dfd47 100644
--- a/packages/jaeger-ui/src/utils/config/get-config.tsx
+++ b/packages/jaeger-ui/src/utils/config/get-config.tsx
@@ -16,15 +16,21 @@ import _get from 'lodash/get';
 
 import processDeprecation from './process-deprecation';
 import defaultConfig, { deprecations } from '../../constants/default-config';
+import { Config } from '../../types/config';
 
 let haveWarnedFactoryFn = false;
 let haveWarnedDeprecations = false;
+let cachedConfig: null | Config = null;
 
 /**
  * Merge the embedded config from the query service (if present) with the
  * default config from `../../constants/default-config`.
  */
 export default function getConfig() {
+  if(cachedConfig) {
+    return cachedConfig;
+  }
+
   const getJaegerUiConfig = window.getJaegerUiConfig;
   if (typeof getJaegerUiConfig !== 'function') {
     if (!haveWarnedFactoryFn) {
@@ -52,6 +58,7 @@ export default function getConfig() {
       rv[key] = { ...defaultConfig[key], ...embedded[key] };
     }
   }
+  cachedConfig = rv;
   return rv;
 }
```

From a usage perspective, we can say, that JS config file is [Function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function) and we invoke it once the page loaded

yurishkuro: Could you show how index.html would be changed to handle JS file? Right now it looks like this, and the middle line is being replaced with the content of JSON file.
```
      function getJaegerUiConfig() {
        const DEFAULT_CONFIG = null;
        const JAEGER_CONFIG = DEFAULT_CONFIG;
        return JAEGER_CONFIG;
      }
```

th3M1ke: For JSON:
```
      function getJaegerUiConfig() {
        const DEFAULT_CONFIG = null;
        const JAEGER_CONFIG = {"a":1, "b":2};
        return JAEGER_CONFIG;
      }
```

For JS:
```
      function getJaegerUiConfig() {
        const DEFAULT_CONFIG = null;
        const JAEGER_CONFIG = function UIConfig() {
          return {"a":1, "b":2};
        }();
        return JAEGER_CONFIG;
      }
```

yurishkuro: I see, you're suggesting to piggy back on the string replacement that currently happens for `const JAEGER_CONFIG = DEFAULT_CONFIG;` line. While this is doable, I wonder if it puts a lot of limitations on what the actual config function can do. If all it does is return a static dictionary, then sure, but we may want to use it as a mechanism for loading plugins, etc. Wouldn't it be better if it was embedded at the top level, not inlined inside an expression?

What if we do this instead:
```javascript
// JAEGER_CONFIG_JS
// the line above may be replaced by user-provided JS file that should define a UIConfig function.

function getJaegerUiConfig() {
        if "UIConfig is not null and is a function" {
             return UIConfig();
        }
        const DEFAULT_CONFIG = null;
        const JAEGER_CONFIG = DEFAULT_CONFIG;
        return JAEGER_CONFIG;
}
```

yurishkuro: regarding this:

> let cachedConfig: null | Config = null;

I think there is a memoize module that we're using, could it be used instead of explicit caching?

th3M1ke: Do I understand you correctly?

JSON:
```
// JAEGER_CONFIG_JS
// the line above may be replaced by user-provided JS file that should define a UIConfig function.

function getJaegerUiConfig() {
        if(typeof window.UIConfig === 'function') {
             return UIConfig();
        }
        const DEFAULT_CONFIG = null;
        const JAEGER_CONFIG = {a:1, b:2};
        return JAEGER_CONFIG;
}
```

JS:
```
function UIConfig() {
   return {a:1, b:2}
}
// the line above may be replaced by user-provided JS file that should define a UIConfig function.

function getJaegerUiConfig() {
        if(typeof window.UIConfig === 'function') {
             return UIConfig();
        }
        const DEFAULT_CONFIG = null;
        const JAEGER_CONFIG = {a:1, b:2};
        return JAEGER_CONFIG;
}
```

yurishkuro: lgtm. any concerns?

th3M1ke: No concerns. Thanks for clarifying! 🤝  WIll update PR shortly

yurishkuro: bonus points if we can get rid of the commend during substitution (e.g. put it all on one line)

th3M1ke:
Have update the PR. @yurishkuro can you take a look?

cmd/query/app/static_handler.go
```diff
@@ -39,6 +40,7 @@ var (
 
 	// The following patterns are searched and replaced in the index.html as a way of customizing the UI.
 	configPattern   = regexp.MustCompile("JAEGER_CONFIG *= *DEFAULT_CONFIG;")
+	configJsPattern = regexp.MustCompile(`(?im)(^\s+)\/\/.*JAEGER_CONFIG_JS.*\n.*`)
```
yurishkuro, line 43:
```suggestion
	configJsPattern = regexp.MustCompile(`(?im)(^\s+)\/\/\s*JAEGER_CONFIG_JS.*\n.*`)
```

yurishkuro: what is the purpose of trailing `.*`? Also, with the `m` mode, does `.*\n` guarantee single-line match?

cmd/query/app/static_handler.go
```diff
@@ -72,6 +74,12 @@ type StaticAssetsHandlerOptions struct {
 	Logger       *zap.Logger
 }
 
+// UIConfigOptions define options for injecting config in index.html in loadAndEnrichIndexHTML
+type UIConfigOptions struct {
```
yurishkuro, line 78: * "options" typically refers to inputs, this is used as output
*  does not need to be public

```suggestion
type loadedConfig struct {
```

cmd/query/app/static_handler.go
```diff
@@ -105,17 +113,11 @@ func loadAndEnrichIndexHTML(open func(string) (http.File, error), options Static
...
+	if configObject, err := loadUIConfig(options.UIConfigPath); err != nil {
 		return nil, err
-	} else if config != nil {
-		// TODO if we want to support other config formats like YAML, we need to normalize `config` to be
-		// suitable for json.Marshal(). For example, YAML parser may return a map that has keys of type
-		// interface{}, and json.Marshal() is unable to serialize it.
-		bytes, _ := json.Marshal(config)
-		configString = fmt.Sprintf("JAEGER_CONFIG = %v", string(bytes))
+	} else if configObject != nil {
+		indexBytes = configObject.regexp.ReplaceAll(indexBytes, append([]byte(`${1}`), configObject.config...))
```
yurishkuro, line 119: I am confused by `${1}` - is it just repeating whitespace? If we match on `// JAEGER...` and replace just that, the whitespace becomes irrelevant.

cmd/query/app/static_handler.go
```diff
@@ -210,30 +212,49 @@ func loadIndexHTML(open func(string) (http.File, error)) ([]byte, error) {
...
 
-	if err := unmarshal(bytes, &c); err != nil {
-		return nil, fmt.Errorf("cannot parse UI config file %v: %w", uiConfig, err)
+	if unmarshal != nil {
+		if err := unmarshal(bytesConfig, &c); err != nil {
+			return nil, fmt.Errorf("cannot parse UI config file %v: %w", uiConfig, err)
+		}
+		// TODO if we want to support other config formats like YAML, we need to normalize `config` to be
+		// suitable for json.Marshal(). For example, YAML parser may return a map that has keys of type
+		// interface{}, and json.Marshal() is unable to serialize it.
```
yurishkuro, line 250: I suggest removing this. We don't have any plans to support anything other than JSON and JS. I would move the unmarshal code directly under `case ".json":` to simplify the logic.

yurishkuro: can also return directly from switch branches instead of maintaining global-like `r`, `re`, `c` variables.

cmd/query/app/static_handler_test.go
```diff
@@ -117,57 +119,67 @@ func TestNewStaticAssetsHandlerErrors(t *testing.T) {
 
 // This test is potentially intermittent
 func TestHotReloadUIConfigTempFile(t *testing.T) {
```
yurishkuro, line 121: Does reload functionality actually depend on the type of file being loaded? It seems it only complicates the test without any real benefits.

cmd/query/app/static_handler_test.go
```diff
@@ -117,57 +119,67 @@ func TestNewStaticAssetsHandlerErrors(t *testing.T) {
...
 func TestLoadUIConfig(t *testing.T) {
 	type testCase struct {
 		configFile    string
-		expected      map[string]interface{}
+		expected      *UIConfigOptions
 		expectedError string
 	}
 
+	jsonRegExpPattern := regexp.MustCompile("JAEGER_CONFIG *= *DEFAULT_CONFIG;")
+	jsRegExpPattern := regexp.MustCompile(`(?im)(^\s+)\/\/.*JAEGER_CONFIG_JS.*\n.*`)
```
yurishkuro, line 181: Could we just use the constants here? Also, do we even need to be comparing which regex is returned? I would rather express the test in terms of what the resulting HTML looks like, not the implementation details of how that HTML is produced.

yurishkuro (commented):
lgtm overall, some clean-up comments

cmd/query/app/static_handler.go
th3M1ke: `.*` - my bad, you are correct `\s+` is best fits.
`m` is for multiline search. I assume that we must replace 2 lines: line with `JAEGER_CONFIG_JS` and following comment on the next line
```
// JAEGER_CONFIG_JS
// the line above may be replaced by user-provided JS file that should define a UIConfig function.
```

cmd/query/app/static_handler.go
th3M1ke: you are correct, it's for white spaces for function UIConfig(). To save origin indent 😶 . Will remove it.

cmd/query/app/static_handler.go
th3M1ke: 👍

cmd/query/app/static_handler_test.go
th3M1ke: Probably no, but to keep consistency I did these changes. Will revert this back

cmd/query/app/static_handler_test.go
th3M1ke: > Could we just use the constants here?

Sure

> Also, do we even need to be comparing which regex is returned? 

This is unit test, I believe, and `LoadUIConfig` is return 
```
type UIConfigOptions struct {
	regexp *regexp.Regexp
	config []byte
}
```
regexp is part of it...


>  I would rather express the test in terms of what the resulting HTML looks like, not the implementation details of how that HTML is produced.

Will update `TestRegisterStaticHandler ` to cover JS file config case

th3M1ke:
Have updated the PR. @yurishkuro can you take a look, please? 🙏

cmd/query/app/static_handler.go
```diff
@@ -210,30 +211,43 @@ func loadIndexHTML(open func(string) (http.File, error)) ([]byte, error) {
...
+		r, _ = json.Marshal(c)
+
+		return &loadedConfig{
+			regexp: configPattern,
+			config: append([]byte("JAEGER_CONFIG = "), append(r, byte(';'))...),
+		}, nil
+	case ".js":
+		r = bytes.TrimSpace(bytesConfig)
+		if !bytes.Contains(r, []byte("function UIConfig(){")) {
+			return nil, fmt.Errorf("wrong JS function format in UI config file format %v", uiConfig)
```
yurishkuro, line 241: more self-explanatory error:
```suggestion
			return nil, fmt.Errorf("UI config file must define function UIConfig(): %v", uiConfig)
```

cmd/query/app/static_handler.go
```diff
@@ -210,30 +211,43 @@ func loadIndexHTML(open func(string) (http.File, error)) ([]byte, error) {
 	return indexBytes, nil
 }
 
-func loadUIConfig(uiConfig string) (map[string]interface{}, error) {
+func loadUIConfig(uiConfig string) (*loadedConfig, error) {
 	if uiConfig == "" {
 		return nil, nil
 	}
 	ext := filepath.Ext(uiConfig)
```
yurishkuro, line 218: for better readability, I would move this next to the `switch` condition:
```go
ext := filepath.Ext(uiConfig)
switch strings.ToLower(ext) {
```

cmd/query/app/static_handler.go
```diff
@@ -210,30 +211,43 @@ func loadIndexHTML(open func(string) (http.File, error)) ([]byte, error) {
...
+		if !bytes.Contains(r, []byte("function UIConfig(){")) {
+			return nil, fmt.Errorf("wrong JS function format in UI config file format %v", uiConfig)
+		}
+
+		return &loadedConfig{
+			regexp: configJsPattern,
+			config: r,
+		}, nil
 	default:
 		return nil, fmt.Errorf("unrecognized UI config file format %v", uiConfig)
```
yurishkuro, line 249:
```suggestion
		return nil, fmt.Errorf("unrecognized UI config file format, expecting .js or .json file: %v", uiConfig)
```

yurishkuro (commented):
lgtm, a couple of minor nits

yurishkuro:
and sorry for delay, did not get the notification of the last change

yurishkuro (approved):
🎉 🎉 🎉 

@th3M1ke this will also require some change in the jaeger-ui, to the html template and the function that retrieves the config. Did you already implement it? It may be worth trying to build query-service with the updated UI as an end-to-end test of JS configs (before we merge this PR, in case of any issues).

th3M1ke:
@yurishkuro I have added a PR with required changes in the `jaeger-ui` repo: https://github.com/jaegertracing/jaeger-ui/pull/677

yurishkuro:
I suggest we bump the UI git submodule to pick up https://github.com/jaegertracing/jaeger-ui/pull/677 (which I just merged). Then once this PR is merged, the latest Docker image for all-in-one can be used to test JS config.

albertteoh:
> I suggest we bump the UI git submodule to pick up [jaegertracing/jaeger-ui#677](https://github.com/jaegertracing/jaeger-ui/pull/677) (which I just merged). Then once this PR is merged, the latest Docker image for all-in-one can be used to test JS config.

Would this require a new release tag of Jaeger-UI?

yurishkuro:
No, submodule links by a commit.

yurishkuro (approved):
Thanks!

PR #2714, opened 2021-01-05
[pkg/queue] Add `StartConsumersWithFactory` function

### LINKED ISSUE #2685: [pkg/queue] Provide factory for consumers

As stated in open-telemetry/opentelemetry-collector#2254 for performance reasons we want to be able to keep state for each consumer of a bounded queue so that we can reuse a certain struct.

It's not currently possible to define a state for each consumer of a bounded queue: the public interface for the bounded queue only allows for providing a function and every consumer runs that function, so you can only keep a global state for all consumers.

https://github.com/jaegertracing/jaeger/blob/1b209472dc6a35f7de90c7a9048cb4e960cc2335/pkg/queue/bounded_queue.go#L61

A possible solution would work like this.

1. Define an interface like
```golang
type Consumer interface {
  Consume(item interface{})
}
```

2. Provide a new way to start a bounded queue with signature
```golang
func (q *BoundedQueue) StartConsumersWithFactory(num int, factory func() Consumer)
```
Reimplement consumer logic internally so that each consumer is built using the factory.

3. Rewrite `StartConsumers` using this factory
```golang
type statelessConsumer struct{
  consumefn func(item interface{}) 
}

func (c *statelessConsumer) Consume(item interface{}){
  c.consumefn(item)
}

func (q *BoundedQueue) StartConsumers(num int, consumer func(item interface{})) { 
  consumer := &statelessConsumer{consumer}
  q.StartConsumersWithFactory(num, func(){
    return consumer 
  })
}
```

I think this would keep backwards-compatible behavior with minimal/no impact to performance but I am not a Go expert so that should be double-checked.

mx-psi:
- Fixes #2685

- Add `StartConsumersWithFactory` function to allow for keeping state per-consumer.
- Rewrite `StartConsumers` with a dummy factory function.
- Add unit test and benchmark for the new function.


pkg/queue/bounded_queue_test.go
```diff
@@ -112,6 +112,80 @@ func TestBoundedQueue(t *testing.T) {
 	assert.False(t, q.Produce("x"), "cannot push to closed queue")
 }
 
+func TestBoundedQueueWithFactory(t *testing.T) {
```
yurishkuro, line 115: is this test essentially a duplicate of the previous one? It's a lot of duplication and complexity. I would rather refactor it into a helper function that can be used with both factory-based and callback-based initialization.

pkg/queue/bounded_queue.go
```diff
@@ -89,6 +95,25 @@ func (q *BoundedQueue) StartConsumers(num int, consumer func(item interface{}))
...
 
+// statelessConsumer wraps a consume function callback
+type statelessConsumer struct {
+	consumefn func(item interface{})
+}
+
+// Consumer consumes an item from a bounded queue
+func (c *statelessConsumer) Consume(item interface{}) {
+	c.consumefn(item)
+}
```
yurishkuro, line 106: I would do this more in line with stdlib/http/HandlerFunc
```suggestion
type ConsumerFunc func(item interface{})

// Consumer consumes an item from a bounded queue
func (c *ConsumerFunc) Consume(item interface{}) {
	c(item)
}
```

pkg/queue/bounded_queue_test.go
```diff
@@ -30,33 +30,14 @@ import (
 	uatomic "go.uber.org/atomic"
 )
 
-// In this test we run a queue with capacity 1 and a single consumer.
```
yurishkuro, line 33: I would keep the comment at the helper function, it describes more the helper than the actual tests.

pkg/queue/bounded_queue_test.go
```diff
@@ -112,6 +93,58 @@ func TestBoundedQueue(t *testing.T) {
...
+		counter.Inc(1)
+	})
+	assert.Equal(t, 1, q.Capacity())
+
+	var startLock sync.Mutex
+
+	startLock.Lock() // block consumers
+	consumerState := newConsumerState(t)
+
+	q.StartConsumers(1, func(item interface{}) {
```
yurishkuro, line 113: There's still a bunch of boiler plate across these tests. What if the helper function looked like this instead:
```go
func helper(t *testing.T, startConsumers func(q *BounderQueue, consumerFn func(item interface{})))
```

Then the tests become
```go
func TestBoundedQueue(t *testing.T) {
    helper(t, func (q *BounderQueue, consumerFn func(item interface{})) {
        q.StartConsumers(1, consumerFn)
    })
}

func TestBoundedQueueWithFactory(t *testing.T) {
    helper(t, func (q *BounderQueue, consumerFn func(item interface{})) {
        q.StartConsumersWithFactory(1, func () Consumer { return ConsumerFunc(consumerFn) })
    })
}
```

pkg/queue/bounded_queue.go
mx-psi: That makes a lot of sense. Should this be exported?

mx-psi: Should be done, marking as solved

mx-psi:
@yurishkuro Thanks for the comments, I think I addressed everything, hopefully this looks better :)

pkg/queue/bounded_queue_test.go
```diff
@@ -33,7 +33,7 @@ import (
 // In this test we run a queue with capacity 1 and a single consumer.
 // We want to test the overflow behavior, so we block the consumer
 // by holding a startLock before submitting items to the queue.
-func TestBoundedQueue(t *testing.T) {
+func helper(t *testing.T, startConsumers func(q *BoundedQueue, consumerFn func(item interface{}))) {
```
yurishkuro, line 36: hehe, I didn't mean the name `helper` literally, was just too lazy to type what you had, but that's fine.

pkg/queue/bounded_queue_test.go
```diff
@@ -112,6 +112,18 @@ func TestBoundedQueue(t *testing.T) {
 	assert.False(t, q.Produce("x"), "cannot push to closed queue")
 }
 
+func TestBoundedQueue(t *testing.T) {
```
yurishkuro, line 115: much cleaner, minimal changes

yurishkuro (approved):
Thanks!

pkg/queue/bounded_queue_test.go
mx-psi: hahaha, I couldn't come up with anything better so I left it like you said 😄

PR #2720, opened 2021-01-08
Copy spans from memory store, fixes #2719

### LINKED ISSUE #2719: Jaeger all-in-one complaints about missing parent spans
**Describe the bug**

I have an issue when jaeger-all-in-one incorrectly complains about missing parent spans:

Here's an example:

```json
{
    "data": [
        {
            "traceID": "ef5dd64a1aa62bae",
            "spans": [
                {
                    "traceID": "ef5dd64a1aa62bae",
                    "spanID": "3e911c0058d13ebb",
                    "flags": 1,
                    "operationName": "trace_me",
                    "references": [
                        {
                            "refType": "CHILD_OF",
                            "traceID": "ef5dd64a1aa62bae",
                            "spanID": "ce8562a5c9781762"
                        }
                    ],
                    "startTime": 1610144303993197,
                    "duration": 30,
                    "tags": [
                        {
                            "key": "internal.span.format",
                            "type": "string",
                            "value": "proto"
                        }
                    ],
                    "logs": [],
                    "processID": "p1",
                    "warnings": [
                        "invalid parent span IDs=ce8562a5c9781762; skipping clock skew adjustment"
                    ]
                },
                {
                    "traceID": "ef5dd64a1aa62bae",
                    "spanID": "ce8562a5c9781762",
                    "flags": 1,
                    "operationName": "top_trace",
                    "references": [],
                    "startTime": 1610144303993035,
                    "duration": 4798745,
                    "tags": [
                        {
                            "key": "sampler.type",
                            "type": "string",
                            "value": "const"
                        },
                        {
                            "key": "sampler.param",
                            "type": "bool",
                            "value": true
                        },
                        {
                            "key": "internal.span.format",
                            "type": "string",
                            "value": "proto"
                        }
                    ],
                    "logs": [],
                    "processID": "p1",
                    "warnings": null
                }
            ],
            "processes": {
                "p1": {
                    "serviceName": "salt",
                    "tags": [
                        {
                            "key": "hostname",
                            "type": "string",
                            "value": "rockpro64"
                        },
                        {
                            "key": "ip",
                            "type": "string",
                            "value": "192.168.1.168"
                        },
                        {
                            "key": "jaeger.version",
                            "type": "string",
                            "value": "Python-4.4.0"
                        }
                    ]
                }
            },
            "warnings": null
        }
    ],
    "total": 0,
    "limit": 0,
    "offset": 0,
    "errors": null
}
```

The error says "invalid parent span IDs=ce8562a5c9781762", but `ce8562a5c9781762` is right there in the JSON.

**To Reproduce**

1. Run jaeger all-in-one
2. Start producing a trace
3. Open the trace before it's full produced
4. Open the trace after it's fully produced
5. Observe lingering warnings about missing parents

**Expected behavior**

No complaints about a missing parent when parent is not in fact missing.

**Screenshots**

The screenshot corresponds to the first JSON snippet in this issue:

image

 - OS: Linux 5.11-rc2
 - Jaeger version: 1.21.1
 - Deployment: Bare metal arm64

**Additional context**

The reason this happens seems to be because `ClockSkew` adjuster runs on real in-memory spans from memory storage engine, as opposed to copies of those spans. So if you happen to request the trace while it's not fully built and the parent is not yet flushed, you'll get a warning for every load. Those warnings are never cleared.

It's possible to get multiple warnings too:

```json
                {
                    "traceID": "eae14db35410faa8",
                    "spanID": "ac4b22f86a3ff0e1",
                    "flags": 1,
                    "operationName": "trace_me",
                    "references": [
                        {
                            "refType": "CHILD_OF",
                            "traceID": "eae14db35410faa8",
                            "spanID": "1da4278e72b02ea1"
                        }
                    ],
                    "startTime": 1610145703000068,
                    "duration": 30,
                    "tags": [
                        {
                            "key": "internal.span.format",
                            "type": "string",
                            "value": "proto"
                        }
                    ],
                    "logs": [],
                    "processID": "p1",
                    "warnings": [
                        "invalid parent span IDs=1da4278e72b02ea1; skipping clock skew adjustment",
                        "invalid parent span IDs=1da4278e72b02ea1; skipping clock skew adjustment"
                    ]
                }
```

bobrik:
Copying allows spans to be freely modified by adjusters and any other code without accidentally altering what is stored in the in-memory store itself. Resolves #2719.

### DIFF
--- a/plugin/storage/memory/memory.go
+++ b/plugin/storage/memory/memory.go
@@ -22,6 +22,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/golang/protobuf/proto"
+
 	"github.com/jaegertracing/jaeger/model"
 	"github.com/jaegertracing/jaeger/model/adjuster"
 	"github.com/jaegertracing/jaeger/pkg/memory/config"
@@ -164,15 +166,19 @@ func (m *Store) GetTrace(ctx context.Context, traceID model.TraceID) (*model.Tra
 	if !ok {
 		return nil, spanstore.ErrTraceNotFound
 	}
-	return m.copyTrace(trace), nil
+	return m.copyTrace(trace)
 }
 
 // Spans may still be added to traces after they are returned to user code, so make copies.
-func (m *Store) copyTrace(trace *model.Trace) *model.Trace {
-	return &model.Trace{
-		Spans:    append([]*model.Span(nil), trace.Spans...),
-		Warnings: append([]string(nil), trace.Warnings...),
+func (m *Store) copyTrace(trace *model.Trace) (*model.Trace, error) {
+	bytes, err := proto.Marshal(trace)
+	if err != nil {
+		return nil, err
 	}
+
+	copied := &model.Trace{}
+	err = proto.Unmarshal(bytes, copied)
+	return copied, err
 }
 
 // GetServices returns a list of all known services
@@ -211,7 +217,12 @@ func (m *Store) FindTraces(ctx context.Context, query *spanstore.TraceQueryParam
 	var retMe []*model.Trace
 	for _, trace := range m.traces {
 		if m.validTrace(trace, query) {
-			retMe = append(retMe, m.copyTrace(trace))
+			copied, err := m.copyTrace(trace)
+			if err != nil {
+				return nil, err
+			}
+
+			retMe = append(retMe, copied)
 		}
 	}
 
--- a/plugin/storage/memory/memory_test.go
+++ b/plugin/storage/memory/memory_test.go
@@ -128,6 +128,14 @@ var childSpan2_1 = &model.Span{
 	StartTime: time.Unix(300, 0),
 }
 
+// This kind of trace cannot be serialized
+var nonSerializableSpan = &model.Span{
+	Process: &model.Process{
+		ServiceName: "naughtyService",
+	},
+	StartTime: time.Date(0, 0, 0, 0, 0, 0, 0, time.UTC),
+}
+
 func withPopulatedMemoryStore(f func(store *Store)) {
 	memStore := NewStore()
 	memStore.WriteSpan(context.Background(), testingSpan)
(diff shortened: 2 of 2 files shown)


plugin/storage/memory/memory.go
```diff
@@ -170,11 +170,37 @@ func (m *Store) GetTrace(ctx context.Context, traceID model.TraceID) (*model.Tra
...
+		OperationName: span.OperationName,
+		References:    span.References,
+		Flags:         span.Flags,
+		StartTime:     span.StartTime,
+		Duration:      span.Duration,
+		Tags:          span.Tags,
+		Logs:          span.Logs,
+		Process:       span.Process,
+		ProcessID:     span.ProcessID,
+		Warnings:      []string(nil),
```
yurishkuro, line 200: why not copy warnings?

plugin/storage/memory/memory.go
```diff
@@ -170,11 +170,37 @@ func (m *Store) GetTrace(ctx context.Context, traceID model.TraceID) (*model.Tra
...
+	spans := make([]*model.Span, len(trace.Spans))
+	for i := range trace.Spans {
+		spans[i] = m.copySpan(trace.Spans[i])
+	}
+	return spans
+}
+
+// Copy static span attributes and discard the ones that can be altered outside of the store.
+func (m *Store) copySpan(span *model.Span) *model.Span {
+	return &model.Span{
```
yurishkuro, line 188: This isn't really a deep copy, we could run into similar issues if say some enrichers add to logs or tags. It would be safer to marshal to proto and back for a fully independent clone

bobrik: Should we do a deep copy of `model.Trace` or just individual `model.Span` members?

plugin/storage/memory/memory.go
bobrik: Good catch, no reason not to copy.

plugin/storage/memory/memory.go
yurishkuro: prob easier to clone Trace, it's also in proto

bobrik: Amended to do a full Trace clone. Some test values needed to be massaged to make `reflect.DeepEqual` happy.

yurishkuro (dismissed):

yurishkuro:
any way to raise the test coverage? It's probably red in the new error handling places.

bobrik:
Any suggestions on how to approach this? I can't figure out how to trick `proto.Marshal` to return an error for `model.Trace`.

yurishkuro:
The following results in an error:
```
	m := &model.Trace{
		Spans: []*model.Span{
			{
				StartTime: time.Date(0, 0, 0, 0, 0, 0, 0, time.UTC),
			},
		},
	}
	_, err := proto.Marshal(m)
```

bobrik:
Added a commit with two more tests to test this scenario, thanks!

yurishkuro (approved):
Thanks!

Issue #2722, opened 2021-01-12
Not support Kafka SASL/PLAIN authentication of SCRAM-SHA-256 or SCRAM-SHA-512 mechanism


WalkerWang731:
**Describe the bug**
In the current version, Kafka authentication only supports `PLAIN` mode, not support `SCRAM-SHA-256` or `SCRAM-SHA-512`. For this issue. I have already added support and commit a new PR later.

**To Reproduce**
1. Set Kafka configuration `server.properties`, like this:
```
listeners=SASL_PLAINTEXT://localhost:9092
security.inter.broker.protocol=SASL_PLAINTEXT
sasl.mechanism.inter.broker.protocol=SCRAM-SHA-512
sasl.enabled.mechanisms=SCRAM-SHA-512
```
For the details can refer: [kafka-documentation](https://kafka.apache.org/documentation/#security_sasl)

Currently, only support mechanisms, like this:
```
listeners=SASL_PLAINTEXT://localhost:9092
security.inter.broker.protocol=SASL_PLAINTEXT
sasl.mechanism.inter.broker.protocol=PLAIN
sasl.enabled.mechanisms=PLAIN
```

**Expected behavior**
Expand to support mechanisms that `SCRAM-SHA-256` or `SCRAM-SHA-512`
Add a new parameter for Kafka, like this:
```
--kafka.producer.plaintext.mechanism string      The plaintext Mechanism for SASL/PLAIN authentication, e.g. 'SCRAM-SHA-256' or 'SCRAM-SHA-512' or 'PLAIN' (default "PLAIN")

--kafka.consumer.plaintext.mechanism string      The plaintext Mechanism for SASL/PLAIN authentication, e.g. 'SCRAM-SHA-256' or 'SCRAM-SHA-512' or 'PLAIN' (default "PLAIN")
```

**Screenshots**
Currently, connection status that using `SCRAM-SHA-256` or `SCRAM-SHA-512`  Kafka server
```
$ jaeger-collector --kafka.producer.brokers "127.0.0.1:9092" --kafka.producer.authentication "plaintext" --kafka.producer.topic jaeger-spans --kafka.producer.plaintext.username user --kafka.producer.plaintext.password password

{"level":"info","ts":1610422257.2999601,"caller":"flags/service.go:116","msg":"Mounting metrics handler on admin server","route":"/metrics"}
[...]

PR #2110, opened 2020-03-02
WIP - Adding scram authentication for kafka

TheDavidKruse:
Co-authored by Zenobius Selvadhason <Zenobius.z.Selvadhason@aexp.com>

Signed-off-by: david kruse <david.kruse@aexp.com>

## Which problem is this PR solving?
- Resolves [#2046](https://github.com/jaegertracing/jaeger/issues/2046) SCRAM authentication between jaeger and kafka

## Short description of the changes
- Added CLI flags to pass in algorithm (SHA), tls config, username, password, etc
- Added package [xdg-go/scram](https://github.com/xdg-go/scram)  package to manage handshake
- Added scram.go to the kafka auth directory

This is a near drop-in from the [sarama scram configuration](https://github.com/Shopify/sarama/tree/master/examples/sasl_scram_client) they listed in their examples.

Issue #2046, opened 2020-01-28
Kafka connection using SCRAM authentication


TheDavidKruse:
## Requirement - what kind of business use case are you trying to solve?

In our current architecture, we secure our connection to our Kafka pipelines using the SCRAM authentication method. We would like to use Go as our Jaeger collector for ease of use in microservices and to replace our current Java implementation.

## Problem - what in Jaeger blocks you from solving the requirement?

The current Go collector doesn't have an implementation to authenticate with Kafka using SCRAM. This forces us into continuing the use with our Java version with higher CPU and memory overhead in our microservice.

## Proposal - what do you suggest to solve the problem or improve the existing situation?

By using [the xdg-go/scram](https://github.com/xdg-go/scram) package we can implement adequate SCRAM authentication between the collector and Kafka. 

I have a pull request able to be made if it's applicable.

## Any open questions to address

Would there be a better SCRAM auth package that could be used?
Would SCRAM with Kafka be useful for anyone else?

yurishkuro:
Maybe this should be solved in Shopify/sarama?  Jaeger simply uses what sarama provides, we don't build additional mechanisms which are not part of the driver.

TheDavidKruse:
That makes sense, let me head over there and see what kind of trouble I can stir up.

Thanks @yurishkuro

TheDavidKruse:
As it turns out, the way we have implemented it comes directly from the way they have in their  (Sarama's) [example](https://github.com/Shopify/sarama/tree/master/examples/sasl_scram_client). 

However, with the example from their repo they do not take into account using Jaeger with Kafka as they have a custom consumer and producer loop. I find that the overlap between what they have implemented in their own way would be easy to implement into Jaeger as it takes care of most of the producer and consumer code and allows for a new authentication method next to TLS and Kerboros.

If it's not worth implementing, is the incorrect implementation location, or is not viewed as something that would benefit Jaeger directly I will close the issue.

zenobiuszeto:
@yurishkuro The PR which we are going to raise will enable the support the SCRAM based SASL fro Jaeger collector for Kafka  which uses `Shopify/sarama`. 

Today we have the support for TLS and kerberos. 

Are you aligned on raising the PR ?

zenobiuszeto:
@yurishkuro I did this case during end of last year

is there a plan to enable SCRAM SASL from Jaeger collector side to kafka ?

[...]

PR #2723, opened 2021-01-12
Add that support Kafka SASL/PLAIN authentication of SCRAM-SHA-256 or SCRAM-SHA-512 mechanism

WalkerWang731:
-  #2722 #2046

- The smallest modification on the current version.
- Different from the previous PR #2110, I just modified `pkg/kafka/auth/plaintext.go`  , `pkg/kafka/auth/options.go` and `pkg/kafka/auth/config.go`, and not add any new file.
- More clearly for option name, using option name is `mechanism` (reference from Kafka parameter name rules )
- About the Prefix, I think to keep using `plaintext` better than `scarm`, because these are both belong to `SASL`, in the future, it can change to `sasl` if needed. Actually, using `plaintext` is ok, it will not be confusing, because it name is `SASL_PLAINTEXT` that from Kafka official security protocol parameter name
- About the parameters limit of `mechanism`,  I followed Kafka name rules (`SCRAM-SHA-256`,  `SCRAM-SHA-512`,  `PLAIN`)
- I reserved the current version `plaintext` mechanism, and add to default. If users do not specify the mechanism, it not affect continued use. if users use the mechanism then just only add `--kafka.ROLE.plaintext.mechanism` parameter then can use it.

## Changelog
Add that support Kafka SASL/PLAIN authentication of SCRAM-SHA-256 or SCRAM-SHA-512 mechanism

### Parameter config of collector (`SPAN_STORAGE_TYPE=kafka`)
```
--kafka.producer.plaintext.mechanism string      The plaintext Mechanism for SASL/PLAIN authentication, e.g. 'SCRAM-SHA-256' or 'SCRAM-SHA-512' or 'PLAIN' (default "PLAIN")
```

### Parameter config of ingester
```
--kafka.consumer.plaintext.mechanism string      The plaintext Mechanism for SASL/PLAIN authentication, e.g. 'SCRAM-SHA-256' or 'SCRAM-SHA-512' or 'PLAIN' (default "PLAIN")
```


WalkerWang731:
Closed this one, because mixed with https://github.com/jaegertracing/jaeger/pull/2657
I recommit a new PR that follows the newest master branch https://github.com/jaegertracing/jaeger/pull/2724

Neustradamus:
The merged PR is:
- https://github.com/jaegertracing/jaeger/pull/2724

PR #2724, opened 2021-01-12
Add support for Kafka SASL/PLAIN authentication via SCRAM-SHA-256 or SCRAM-SHA-512 mechanism

### LINKED ISSUE #2046: Kafka connection using SCRAM authentication

### LINKED ISSUE #2722: Not support Kafka SASL/PLAIN authentication of SCRAM-SHA-256 or SCRAM-SHA-512 mechanism

WalkerWang731:
- Resolves #2722 
- Resolves #2046
- Closes #2110 (replaces it)

- The smallest modification on the current version.
- Different from the previous PR #2110, I just modified `pkg/kafka/auth/plaintext.go`  , `pkg/kafka/auth/options.go` and `pkg/kafka/auth/config.go`, and not add any new file.
- More clearly for option name, using option name is `mechanism` (reference from Kafka parameter name rules )
- About the Prefix, I think to keep using `plaintext` better than `scarm`, because these are both belong to `SASL`, in the future, it can change to `sasl` if needed. Actually, using `plaintext` is ok, it will not be confusing, because it name is `SASL_PLAINTEXT` that from Kafka official security protocol parameter name
- About the parameters limit of `mechanism`,  I followed Kafka name rules (`SCRAM-SHA-256`,  `SCRAM-SHA-512`,  `PLAIN`)
- I reserved the current version `plaintext` mechanism, and add to default. If users do not specify the mechanism, it not affect continued use. if users use the mechanism then just only add `--kafka.ROLE.plaintext.mechanism` parameter then can use it.

## Changelog
Add that support Kafka SASL/PLAIN authentication of SCRAM-SHA-256 or SCRAM-SHA-512 mechanism

### Parameter config of collector (`SPAN_STORAGE_TYPE=kafka`)
```
--kafka.producer.plaintext.mechanism string      The plaintext Mechanism for SASL/PLAIN authentication, e.g. 'SCRAM-SHA-256' or 'SCRAM-SHA-512' or 'PLAIN' (default "PLAIN")
```

### Parameter config of ingester
```
--kafka.consumer.plaintext.mechanism string      The plaintext Mechanism for SASL/PLAIN authentication, e.g. 'SCRAM-SHA-256' or 'SCRAM-SHA-512' or 'PLAIN' (default "PLAIN")
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -67,6 +67,7 @@ require (
 	github.com/uber/jaeger-lib v2.4.0+incompatible
 	github.com/vektra/mockery v0.0.0-20181123154057-e78b021dcbb5
 	github.com/wadey/gocovmerge v0.0.0-20160331181800-b5bfa59ec0ad
+	github.com/xdg/scram v0.0.0-20180814205039-7eeb5667e42c
 	go.mongodb.org/mongo-driver v1.3.2 // indirect
 	go.uber.org/atomic v1.6.0
 	go.uber.org/automaxprocs v1.3.0
--- a/pkg/kafka/auth/plaintext.go
+++ b/pkg/kafka/auth/plaintext.go
@@ -15,17 +15,75 @@
 package auth
 
 import (
+	"crypto/sha256"
+	"crypto/sha512"
+	"fmt"
+	"hash"
+	"strings"
+
 	"github.com/Shopify/sarama"
+	"github.com/xdg/scram"
 )
+
+// XDGSCRAMClient is return a *sarama.SCRAMClient on create SCRAMClientGeneratorFunc when the mechanism is SCRAM-SHA-256 or SCRAM-SHA-512
+type XDGSCRAMClient struct {
+	*scram.Client
+	*scram.ClientConversation
+	scram.HashGeneratorFcn
+}
+
+// Begin prepares the client for the SCRAM exchange
+// with the server with a user name and a password
+func (x *XDGSCRAMClient) Begin(userName, password, authzID string) (err error) {
+	x.Client, err = x.HashGeneratorFcn.NewClient(userName, password, authzID)
+	if err != nil {
+		return err
+	}
+	x.ClientConversation = x.Client.NewConversation()
+	return nil
+}
+
+// Step steps client through the SCRAM exchange. It is
+// called repeatedly until it errors or `Done` returns true.
+func (x *XDGSCRAMClient) Step(challenge string) (response string, err error) {
+	response, err = x.ClientConversation.Step(challenge)
+	return
+}
+
+// Done should return true when the SCRAM conversation
+// is over.
+func (x *XDGSCRAMClient) Done() bool {
+	return x.ClientConversation.Done()
+}
 
 // PlainTextConfig describes the configuration properties needed for SASL/PLAIN with kafka
 type PlainTextConfig struct {
-	UserName string `mapstructure:"username"`
-	Password string `mapstructure:"password" json:"-"`
+	UserName  string `mapstructure:"username"`
+	Password  string `mapstructure:"password" json:"-"`
+	Mechanism string `mapstructure:"mechanism"`
 }
 
-func setPlainTextConfiguration(config *PlainTextConfig, saramaConfig *sarama.Config) {
+func setPlainTextConfiguration(config *PlainTextConfig, saramaConfig *sarama.Config) error {
 	saramaConfig.Net.SASL.Enable = true
 	saramaConfig.Net.SASL.User = config.UserName
 	saramaConfig.Net.SASL.Password = config.Password
+	switch strings.ToUpper(config.Mechanism) {
+	case "SCRAM-SHA-256":
+		saramaConfig.Net.SASL.SCRAMClientGeneratorFunc = func() sarama.SCRAMClient {
+			return &XDGSCRAMClient{HashGeneratorFcn: func() hash.Hash { return sha256.New() }}
+		}
+		saramaConfig.Net.SASL.Mechanism = sarama.SASLTypeSCRAMSHA256
+	case "SCRAM-SHA-512":
+		saramaConfig.Net.SASL.SCRAMClientGeneratorFunc = func() sarama.SCRAMClient {
+			return &XDGSCRAMClient{HashGeneratorFcn: func() hash.Hash { return sha512.New() }}
+		}
+		saramaConfig.Net.SASL.Mechanism = sarama.SASLTypeSCRAMSHA512
+	case "PLAIN":
+		saramaConfig.Net.SASL.Mechanism = sarama.SASLTypePlaintext
+
+	default:
+		return fmt.Errorf("config plaintext.mechanism error: %s, only support 'SCRAM-SHA-256' or 'SCRAM-SHA-512' or 'PLAIN'", config.Mechanism)
+
+	}
+	return nil
 }
--- a/pkg/kafka/auth/plaintext.go
+++ b/pkg/kafka/auth/plaintext.go
@@ -25,8 +25,8 @@ import (
 	"github.com/xdg/scram"
 )
 
-// XDGSCRAMClient is return a *sarama.SCRAMClient on create SCRAMClientGeneratorFunc when the mechanism is SCRAM-SHA-256 or SCRAM-SHA-512
-type XDGSCRAMClient struct {
+// scramClient is the client to use when the auth mechanism is SCRAM
+type scramClient struct {
 	*scram.Client
 	*scram.ClientConversation
 	scram.HashGeneratorFcn
@@ -34,7 +34,7 @@ type XDGSCRAMClient struct {
 
 // Begin prepares the client for the SCRAM exchange
 // with the server with a user name and a password
-func (x *XDGSCRAMClient) Begin(userName, password, authzID string) (err error) {
+func (x *scramClient) Begin(userName, password, authzID string) (err error) {
 	x.Client, err = x.HashGeneratorFcn.NewClient(userName, password, authzID)
 	if err != nil {
 		return err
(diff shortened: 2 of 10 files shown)


go.mod
```diff
@@ -67,6 +67,7 @@ require (
 	github.com/uber/jaeger-lib v2.4.0+incompatible
 	github.com/vektra/mockery v0.0.0-20181123154057-e78b021dcbb5
 	github.com/wadey/gocovmerge v0.0.0-20160331181800-b5bfa59ec0ad
+	github.com/xdg/scram v0.0.0-20180814205039-7eeb5667e42c
```
jpkrohling, line 70: This redirects to `github.com/xdg-go/scram`

pkg/kafka/auth/plaintext.go
```diff
@@ -15,17 +15,75 @@
...
+	"crypto/sha512"
+	"fmt"
+	"hash"
+	"strings"
+
 	"github.com/Shopify/sarama"
+	"github.com/xdg/scram"
 )
 
+// XDGSCRAMClient is return a *sarama.SCRAMClient on create SCRAMClientGeneratorFunc when the mechanism is SCRAM-SHA-256 or SCRAM-SHA-512
```
jpkrohling, line 28: This sounds a bit strange. How about something like this?
```suggestion
// XDGSCRAMClient is the client to use when the auth mechanism is SCRAM
```

But if you don't have clashes anywhere, you can just use `SCRAMClient` instead.

jpkrohling: You might want to do a type assertion as well, like:

```go
var _ sarama.SCRAMClient = (*XDGSCRAMClient)(nil)
```

plugin/storage/kafka/factory_test.go
```diff
@@ -120,6 +120,7 @@ func TestKafkaFactoryDoesNotLogPassword(t *testing.T) {
 				"--kafka.producer.authentication=plaintext",
 				"--kafka.producer.plaintext.username=username",
 				"--kafka.producer.plaintext.password=SECRET",
+				"--kafka.producer.plaintext.mechanism=PLAINT",
```
yurishkuro, line 123:
```suggestion
				"--kafka.producer.plaintext.mechanism=PLAIN",
```
why is this even needed? the mechanism defaults to PLAIN already

pkg/kafka/auth/plaintext.go
yurishkuro: I don't think this type even needs to be public, can just call it `type scramClient struct`...

plugin/storage/kafka/factory_test.go
WalkerWang731: Yes, actually, I forgot to remove it, this parameter is no point. I will remove it on the later commit.

go.mod
WalkerWang731: Yes, actually, this line generate via `make fmt`,
and `github.com/xdg/scram` is dependent on `github.com/Shopify/sarama` (https://github.com/Shopify/sarama/blob/v1.22.1/go.mod#L14)

in jaeger `go.mod`, it has already existed that `github.com/Shopify/sarama`

jpkrohling: My point is that we should use xdg-go instead.

pkg/kafka/auth/plaintext.go
```diff
@@ -15,17 +15,75 @@
...
+// is over.
+func (x *scramClient) Done() bool {
+	return x.ClientConversation.Done()
+}
+
 // PlainTextConfig describes the configuration properties needed for SASL/PLAIN with kafka
 type PlainTextConfig struct {
-	UserName string `mapstructure:"username"`
-	Password string `mapstructure:"password" json:"-"`
+	UserName  string `mapstructure:"username"`
```
jpkrohling, line 61:
```suggestion
	Username  string `mapstructure:"username"`
```

pkg/kafka/auth/plaintext.go
jpkrohling: We still need the type assertion though :-)

go.mod
yurishkuro: The library's README recommends `import "github.com/xdg/scram"`. Since there are no sub-packages, it might work with both imports. I opened a ticket there https://github.com/xdg-go/scram/issues/3.

jpkrohling: Not sure GitHub allows recycling previously used names like npm does (or used to), but if so, this opens a relatively serious security window. So, if we trust code that is under `xdg-go`, we just shouldn't use `xdg`.

pkg/kafka/auth/plaintext.go
WalkerWang731: ok, I will update later.

pkg/kafka/auth/plaintext.go
WalkerWang731: yes, I've already added it. :-)

jpkrohling (approved):
LGTM

yurishkuro (approved):

Neustradamus:
@WalkerWang731 and all others: Thanks a lot for all about SCRAM.

Linked to:
- https://github.com/scram-xmpp/info/issues/1

PR #2729, opened 2021-01-14
[tests] Increase timeout for TestGRPCResolverRoundRobin

yurishkuro:
Resolves #2025.

At first this was just adding extra logging, but per the next comment the logging indicated that it's indeed a timing issue.

Also did some clean-up of the code, like pre-allocating arrays with the right sizes.

### DIFF
--- a/pkg/discovery/grpcresolver/grpc_resolver_test.go
+++ b/pkg/discovery/grpcresolver/grpc_resolver_test.go
@@ -136,10 +136,13 @@ func TestGRPCResolverRoundRobin(t *testing.T) {
 		minPeers    int
 		connections int // expected number of unique connections to servers
 	}{
-		{3, 3}, {5, 5}, {7, 5},
+		{minPeers: 3, connections: 3},
+		{minPeers: 5, connections: 3},
+		// note: test cannot succeed with minPeers < connections because resolver
+		// will never return more than minPeers addresses.
 	}
 	for _, test := range tests {
-		t.Run(fmt.Sprintf("minPeers=%d", test.minPeers), func(t *testing.T) {
+		t.Run(fmt.Sprintf("%+v", test), func(t *testing.T) {
 			res := New(notifier, discoverer, zap.NewNop(), test.minPeers)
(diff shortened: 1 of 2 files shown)


yurishkuro:
ha, interesting, the logging already yielded fruit, the test failed like this:
```
    grpc_resolver_test.go:100: 
        	Error Trace:	grpc_resolver_test.go:100
        	            				grpc_resolver_test.go:156
        	Error:      	Should be true
        	Test:       	TestGRPCResolverRoundRobin/minPeers=3
        	Messages:   	Connection #0 was still not up. Connections so far: map[]
```

Because it's the very first connection that failed, it can only be due to timing, as there are no other conditions. Also, first connections in the subsequent tests are often acquired after 2-3 hundred iterations, almost half of the total time, e.g. this one at 400:
```
=== RUN   TestGRPCResolverRoundRobin/minPeers=7
    grpc_resolver_test.go:95: connected to peer #0 (127.0.0.1:40041) on iteration 440
    grpc_resolver_test.go:95: connected to peer #1 (127.0.0.1:44157) on iteration 1
    grpc_resolver_test.go:95: connected to peer #2 (127.0.0.1:41503) on iteration 0
    grpc_resolver_test.go:95: connected to peer #3 (127.0.0.1:38727) on iteration 1
    grpc_resolver_test.go:95: connected to peer #4 (127.0.0.1:32919) on iteration 0
```
Even though we have a pretty long timeout already (1000 * 10ms = 10s), I am going to increase it even more.

pkg/discovery/grpcresolver/grpc_resolver_test.go
```diff
@@ -84,19 +84,20 @@ func makeSureConnectionsUp(t *testing.T, count int, testc grpctest.TestServiceCl
 	// Make sure connections to all servers are up.
 	for si := 0; si < count; si++ {
 		connected := false
-		for i := 0; i < 1000; i++ {
+		for i := 0; i < 3000; i++ { // 3000 * 10ms = 30s
```
yurishkuro, line 87: this is the main change/fix

PR #2737, opened 2021-01-21
Expose cache ttl for es span writer index+service

necrolyte2:
- Unable to specify a cache duration for the service or index caches
- Our use case requires that we can set these

- Added ability to pass ttl duration for both service and index caches for the elastic spanstore writer
- If not specified it will use the previous defaults of 12h and 48h

### DIFF
--- a/plugin/storage/es/spanstore/writer.go
+++ b/plugin/storage/es/spanstore/writer.go
@@ -65,29 +65,20 @@ type SpanWriterParams struct {
 	TagDotReplacement   string
 	Archive             bool
 	UseReadWriteAliases bool
-	ServiceCacheTTL     string
-	IndexCacheTTL       string
+	ServiceCacheTTL     time.Duration
+	IndexCacheTTL       time.Duration
 }
 
 // NewSpanWriter creates a new SpanWriter for use
 func NewSpanWriter(p SpanWriterParams) *SpanWriter {
-	var err error
-	serviceCacheTTL := serviceCacheTTLDefault
-	if p.ServiceCacheTTL != "" {
-		serviceCacheTTL, err = time.ParseDuration(p.ServiceCacheTTL)
-		if err != nil {
-			p.Logger.Warn("error parsing service cache duration. Using default", zap.Error(err), zap.Duration("default", serviceCacheTTLDefault))
-			serviceCacheTTL = serviceCacheTTLDefault
-		}
+	serviceCacheTTL := p.ServiceCacheTTL
+	if p.ServiceCacheTTL == 0 {
+		serviceCacheTTL = serviceCacheTTLDefault
 	}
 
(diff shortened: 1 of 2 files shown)


plugin/storage/es/spanstore/writer.go
```diff
@@ -53,22 +55,43 @@ type SpanWriter struct {
...
+	MetricsFactory       metrics.Factory
+	IndexPrefix          string
+	IndexDateLayout      string
+	AllTagsAsFields      bool
+	TagKeysAsFields      []string
+	TagDotReplacement    string
+	Archive              bool
+	UseReadWriteAliases  bool
+	ServiceCacheDuration string
+	IndexCacheDuration   string
```
yurishkuro, line 69:
```suggestion
	ServiceCacheDuration time.Duration
	IndexCacheDuration   time.Duration
```

yurishkuro (commented):
* please add test coverage
* how would someone use this functionality? it is not exposed via CLI flags

plugin/storage/es/spanstore/writer.go
necrolyte2: You cannot check if a time.Duration is set or not since it would be 0, but you can check if a string is empty which is why I went the path I did there.

Anybody directly using the writer in their own code can use it and if later somebody wants to plumb in the ability for a flag for the apps to be able to set these could do that in a separate PR

I'm not sure how to test this since everything is in a private property and I cannot check if they are set in a test

yurishkuro: What's wrong with 0?  0 is not a valid value for these parameters, so it works perfectly well as "unset" value.

Tests run in the private package namespace and can test those values.

necrolyte2: 0 is a valid value and would indicate caching forever so you still would not be able to differentiate from cache forever and not being set. Unless you want to assume that nobody would ever want to cache forever(which could be a valid thing, but I would rather be able to know for sure)

I did figure out how to test and just pushed those

yurishkuro: > 0 is a valid value and would indicate caching forever

Technically for the LRU cache that is under under the hood the ttl=0 is a valid param. But in the context of the specific caching scenarios here (index and service names) I do not believe 0 is a valid value, because these caches grow over time. So we can use 0 as "undefined". If people want to cache "forever" they can still do it by providing a very long duration.

necrolyte2: Sounds good. I can make the change

yurishkuro: If you really want to support "undefined" then the other alternative is to use a pointer.

yurishkuro (approved):
Thanks!

PR #2740, opened 2021-01-26
Github action for jaeger packages and docker image

Ashmita152:

- Adds gitbub action to build docker images and packages
- Part of https://github.com/jaegertracing/jaeger/issues/2645

- Release part is still pending. Looking at it, I think we need to use https://github.com/actions/upload-release-asset action but not quite sure how to make it work without changing our current release process.

### DIFF
--- a/.github/workflows/ci-build-binaries.yml
+++ b/.github/workflows/ci-build-binaries.yml
@@ -30,6 +30,3 @@ jobs:
 
     - name: Build binaries
       run: make build-all-platforms
-
-    - name: Package binaries
-      run: bash scripts/travis/package-deploy.sh
--- a/.github/workflows/ci-build-binaries.yml
+++ b/.github/workflows/ci-build-binaries.yml
@@ -31,5 +31,5 @@ jobs:
     - name: Build binaries
       run: make build-all-platforms
 
-    - name: Upload binaries
+    - name: Package binaries
       run: bash scripts/travis/package-deploy.sh
--- a/.github/workflows/ci-build-binaries.yml
+++ b/.github/workflows/ci-build-binaries.yml
@@ -1,0 +1,35 @@
+name: Build binaries
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build-binaries:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        submodules: true
+
+    - name: Fetch git tags
+      run: |
+        git fetch --prune --unshallow --tags
+
+    - uses: actions/setup-go@v2
+      with:
+        go-version: ^1.15
+
+    - name: Export BRANCH variable
+      uses: ./.github/actions/setup-branch
+
+    - name: Install tools
+      run: make install-ci
+
+    - name: Build binaries
+      run: make build-all-platforms
+
+    - name: Upload binaries
+      run: bash scripts/travis/package-deploy.sh
--- a/.github/workflows/ci-build-binaries.yml
+++ b/.github/workflows/ci-build-binaries.yml
@@ -14,10 +14,6 @@ jobs:
       with:
         submodules: true
 
-    - name: Fetch git tags
-      run: |
-        git fetch --prune --unshallow --tags
-
     - uses: actions/setup-go@v2
       with:
         go-version: ^1.15
--- a/.github/workflows/ci-docker-build.yml
+++ b/.github/workflows/ci-docker-build.yml
@@ -1,0 +1,55 @@
+name: Build docker images
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  docker-images:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        submodules: true
+
+    - name: Fetch git tags
+      run: |
+        git fetch --prune --unshallow --tags
+
+    - uses: actions/setup-go@v2
+      with:
+        go-version: ^1.15
+
+    - uses: actions/setup-node@v2-beta
+      with:
+        node-version: '10'
+
+    - uses: docker/login-action@v1
+      id: dockerhub-login
+      with:
+        username: jaegertracingbot
+        password: ${{ secrets.DOCKERHUB_TOKEN }}
+      env:
+        DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
+      if: env.DOCKERHUB_TOKEN != null
+
+    - name: Export DOCKERHUB_LOGIN variable
+      run: |
+        echo "DOCKERHUB_LOGIN=true" >> $GITHUB_ENV
+      if: steps.dockerhub-login.outcome == 'success'
+
+    - name: Export BRANCH variable
+      uses: ./.github/actions/setup-branch
+
+    - name: Install tools
+      run: make install-ci
+
+    - name: Build docker images
+      run: make docker
+      env:
+        DOCKER_NAMESPACE: jaegertracing
+
+    - name: Upload docker images
+      run: bash scripts/travis/upload-all-docker-images.sh
--- a/.github/workflows/ci-docker-build.yml
+++ b/.github/workflows/ci-docker-build.yml
@@ -48,8 +48,6 @@ jobs:
 
     - name: Build docker images
       run: make docker
-      env:
-        DOCKER_NAMESPACE: jaegertracing
 
     - name: Upload docker images
       run: bash scripts/travis/upload-all-docker-images.sh
--- a/.github/workflows/ci-release.yml
+++ b/.github/workflows/ci-release.yml
@@ -46,12 +46,13 @@ jobs:
       run: make install-ci
 
     - name: Build binaries
+      id: build-binaries
       run: make build-all-platforms
 
     - name: Package binaries
+      id: package-binaries
       run: bash scripts/travis/package-deploy.sh
-      env:
-        DEPLOY: true
+      if: steps.build-binaries.outcome == 'success'
 
     - name: Upload binaries
       uses: svenstaro/upload-release-action@v2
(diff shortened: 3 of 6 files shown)


.github/workflows/ci-packages-build.yml
```diff
@@ -0,0 +1,35 @@
+name: Build packages
```
yurishkuro, line 1: let's call this `ci-build-binaries.yml` , `Build binaries`

scripts/travis/upload-all-docker-images.sh
```diff
@@ -1,18 +1,14 @@
 #!/bin/bash
 
-# this script should only be run after build-docker-images.sh
+# this script expects all docker images to be present
```
yurishkuro, line 3:
```suggestion
# this script expects all docker images to be already built, it only uploads them to Docker Hub
```

.github/workflows/ci-packages-build.yml
```diff
@@ -0,0 +1,35 @@
...
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  packages:
```
yurishkuro, line 10:
```suggestion
  build-binaries:
```

.github/workflows/ci-packages-build.yml
```diff
@@ -0,0 +1,35 @@
...
+    - name: Export BRANCH variable
+      uses: ./.github/actions/setup-branch
+
+    - name: Install tools
+      run: make install-ci
+
+    - name: Build binaries
+      run: make build-all-platforms
+
+    - name: Package binaries
```
yurishkuro, line 34:
```suggestion
    - name: Upload binaries
```

yurishkuro (commented):
> Release part is still pending. Looking at it, I think we need to use https://github.com/actions/upload-release-asset action but not quite sure how to make it work without changing our current release process.

well, (a) the current release process is non-functional since Travis refuses to run out jobs, and (b) do you have reasons to suspect that `scripts/travis/package-deploy.sh` would not work?

Ashmita152:
> do you have reasons to suspect that scripts/travis/package-deploy.sh would not work?

Sorry my bad in not explaining my point properly. This PR doesn't add changes equivalent to https://github.com/jaegertracing/jaeger/blob/master/.travis.yml#L93-L104 part of the travis configuration. For that we need to use https://github.com/actions/upload-release-asset But I am confused about how to make it work equivalently because of this `tag: on` travis configuration (https://github.com/jaegertracing/jaeger/blob/master/.travis.yml#L102)

yurishkuro:
> But I am confused about how to make it work equivalently because of this tags: true travis configuration

One option would be to make the whole workflow contingent on the presence of a tag, but then we'd miss any compile errors due to arch differences. So the other alternative is to do what the upload to docker script is doing, i.e. check the tag explicitly and skip the upload of bundles if no tag. Then the workflow will always run.

Ashmita152:
Do you recommend having a new github action like ci-release.yml which will run only when the tag is pushed ?

```
on:
  push:
    tags: [ '*' ]
```

Ashmita152:
Also looking at the example of README of https://github.com/actions/upload-release-asset it seems like it depends on another action https://github.com/actions/create-release (this part we do manually as part of our current release process) 

I was thinking maybe we can probably set `draft: true` in create-release action, thus changing minimal in our current new version release process.

yurishkuro:
We don't want to create a release from actions, because out workflow is (1) create release manually, (2) use CI to upload artifacts into it.

It doesn't look like upload-release-asset depends on create-release, but it does need upload_url for the release (don't know why GH makes it so hard). We could use this one instead https://github.com/bruceadams/get-release, although I would've preferred an official one.  There's an open issue https://github.com/actions/upload-release-asset/issues/34

yurishkuro:
looks like this could be used `${{ github.event.release.upload_url }}`, assuming that the workflow runs at the appropriate time

Ashmita152:
Sure Yuri, Thank you for the feedback. I will try out as you suggested in my forked repo.

.github/workflows/ci-docker-build.yml
```diff
@@ -0,0 +1,55 @@
...
+    - name: Export BRANCH variable
+      uses: ./.github/actions/setup-branch
+
+    - name: Install tools
+      run: make install-ci
+
+    - name: Build docker images
+      run: make docker
+      env:
+        DOCKER_NAMESPACE: jaegertracing
```
yurishkuro, line 52: isn't this the default in the script?

.github/workflows/ci-build-binaries.yml
```diff
@@ -0,0 +1,35 @@
...
+      uses: ./.github/actions/setup-branch
+
+    - name: Install tools
+      run: make install-ci
+
+    - name: Build binaries
+      run: make build-all-platforms
+
+    - name: Package binaries
+      run: bash scripts/travis/package-deploy.sh
```
yurishkuro, line 35: this looks like a noop without DEPLOY=true? Do we need to call it?

.github/workflows/ci-release.yml
```diff
@@ -0,0 +1,73 @@
...
+    - name: Install tools
+      run: make install-ci
+
+    - name: Build binaries
+      run: make build-all-platforms
+
+    - name: Package binaries
+      run: bash scripts/travis/package-deploy.sh
+      env:
+        DEPLOY: true
```
yurishkuro, line 54: I think this is a left-over from Travis configuration. Since you added a standalone workflow for release, I would fix the script to not have this env var, it serves no purpose anymore

Ashmita152:
I tried updating the PR as per your comments.

Since the GitHub Action https://github.com/actions/upload-release-asset doesn't support uploading multiple files in one task by using regex, I had to use https://github.com/marketplace/actions/upload-files-to-a-github-release action instead. 

I tried this PR in my jaeger's fork by creating a release manually (https://github.com/Ashmita152/jaeger/releases/tag/v1.25.0). It triggered the ci-release GitHub Action whose run you can find here: https://github.com/Ashmita152/jaeger/runs/1798106742?check_suite_focus=true

.github/workflows/ci-release.yml
```diff
@@ -0,0 +1,73 @@
...
+      with:
+        file: deploy/*.tar.gz
+        file_glob: true
+        tag: ${{ github.ref }}
+        repo_token: ${{ secrets.GITHUB_TOKEN }}
+
+    - name: Build docker images
+      run: make docker
+      env:
+        DOCKER_NAMESPACE: jaegertracing
```
yurishkuro, line 67: not needed

https://github.com/jaegertracing/jaeger/blob/2ad44d581c71b8cfb66b0c7ef153a0fd5dcdb6db/Makefile#L69

Ashmita152:
Thank you for the feedbacks Yuri. Those were nice catches. I have updated the PR removing those.

Ashmita152:
I have created a new release in my forked repo: https://github.com/Ashmita152/jaeger/releases/tag/v1.35.0 and CI for the release job is: https://github.com/Ashmita152/jaeger/runs/1798332218?check_suite_focus=true

This time the BRANCH is populating properly: 
```
we are on branch=v1.35.0
```

.github/workflows/ci-build-binaries.yml
```diff
@@ -0,0 +1,32 @@
...
+  build-binaries:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        submodules: true
+
+    - name: Fetch git tags
+      run: |
+        git fetch --prune --unshallow --tags
```
yurishkuro, line 19: why do we need this step when building binaries?

yurishkuro: also, when we do need tags, can we achieve the same effect with this setting?
```
    # Number of commits to fetch. 0 indicates all history for all branches and tags.
    # Default: 1
    fetch-depth: ''
```

.github/workflows/ci-release.yml
yurishkuro: let's remove this

.github/workflows/ci-release.yml
```diff
@@ -0,0 +1,71 @@
...
+    - name: Export BRANCH variable
+      uses: ./.github/actions/setup-branch
+
+    - name: Install tools
+      run: make install-ci
+
+    - name: Build binaries
+      run: make build-all-platforms
+
+    - name: Package binaries
```
yurishkuro, line 51: should this step be conditional on success of 'build binaries'?

.github/workflows/ci-release.yml
```diff
@@ -0,0 +1,71 @@
...
+      run: bash scripts/travis/package-deploy.sh
+      env:
+        DEPLOY: true
+
+    - name: Upload binaries
```
yurishkuro, line 56: should this be conditional on success of previous step?

.github/workflows/ci-release.yml
```diff
@@ -0,0 +1,71 @@
...
+      with:
+        file: deploy/*.tar.gz
+        file_glob: true
+        tag: ${{ github.ref }}
+        repo_token: ${{ secrets.GITHUB_TOKEN }}
+
+    - name: Build docker images
+      run: make docker
+
+    - name: Upload docker images
```
yurishkuro, line 67: conditional?

.github/workflows/ci-build-binaries.yml
Ashmita152: I think if we want to pull tags as part of `checkout` action, only option from my understanding is to set `fetch-depth: 0` which will also fetch all the commit history of all branches which will increase build time. I'm unable to find an option for just pulling the tags without commit history.

yurishkuro: maybe it's worth creating a PR for the official action to add a new parameter like `tags:`

yurishkuro (approved):
Thanks!

yurishkuro:
@Ashmita152 seeing some odd `fatal:` logs in the build-binaries:

https://github.com/jaegertracing/jaeger/runs/1803971666?check_suite_focus=true#step:6:9

```
GOOS=linux GOARCH=amd64 make build-platform-binaries
make[1]: Entering directory '/home/runner/work/jaeger/jaeger'
fatal: No names found, cannot describe anything.
```

Ashmita152:
Hi @yurishkuro 

`build-binaries` step is taking ~ 30 minutes right now since it is building for all platforms. Do you recommend dividing it into separate jobs per platform ?

yurishkuro:
yeah, i think it would be neat if we could build them in parallel

PR #2746, opened 2021-01-28
Add changelog entry for Query server TLS support

rjs211:

- Add changelog entry for Query server TLS support

- Adding changelog entry for changes done in #2337

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,19 @@ Changes by Version
 1.22.0 (unreleased)
 -------------------
 
+### Backend Changes
+
+### New Features
+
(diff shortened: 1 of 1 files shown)


rjs211:
@yurishkuro Hello. I added a simple entry.  A few points for further disambiguation could be to specifying the fallback values for the host-ports, something along the lines of:  
```
In case the dedicated ports are used, but either of both flsgs are *not* explicitly set, the default values are used.
```
I havent included the above in this PR as is not a part of #2337 but rather a part of #2387.  Please do let me know how to proceed.

yurishkuro (approved):
That's great! Thanks!

yurishkuro:
Feel free to include that too, the more complete the description of the behavior the better.

Issue #2550, opened 2020-10-11
jaeger-agent --reporter.grpc.tls.enabled=true fails on windows


wscheele:
**Describe the bug**
Due to https://github.com/golang/go/issues/16736, the jaeger-agent is unable to setup a connection to a tls grpc reporter endpoint on windows.

Error message:
```
{"level":"fatal","ts":1602401908.4807255,"caller":"command-line-arguments/main.go:67","msg":"Could not create collector proxy","error":"failed to load TLS config: failed to load CA CertPool: failed to load SystemCertPool: crypto/x509: system root pool is not available on Windows","stacktrace":"main.main.func1\n\tcommand-line-arguments/main.go:67\ngithub.com/spf13/cobra.(*Command).execute\n\tgithub.com/spf13/cobra@v0.0.3/command.go:762\ngithub.com/spf13/cobra.(*Command).ExecuteC\n\tgithub.com/spf13/cobra@v0.0.3/command.go:852\ngithub.com/spf13/cobra.(*Command).Execute\n\tgithub.com/spf13/cobra@v0.0.3/command.go:800\nmain.main\n\tcommand-line-arguments/main.go:102\nruntime.main\n\truntime/proc.go:204"}
```

**To Reproduce**
1. run jaeger-agent.exe with --reporter.grpc.tls.enabled=true

**Expected behavior**
It works

**Version (please complete the following information):**
 - OS: Windows
 - Jaeger version: v1.20.0

**Additional context**
https://github.com/golang/go/issues/16736
https://github.com/sensu/sensu-go/pull/4018

PR #2756, opened 2021-02-01
Use workaround for windows x509.SystemCertPool()

Ashmita152:

Fixes https://github.com/jaegertracing/jaeger/issues/2550

Doing exactly same as https://github.com/sensu/sensu-go/pull/4018


pkg/config/tlscfg/options.go
```diff
@@ -90,6 +90,13 @@ func (p Options) loadCertPool() (*x509.CertPool, error) {
 		if err != nil {
 			return nil, fmt.Errorf("failed to load SystemCertPool: %w", err)
 		}
+		if certPool == nil {
```
yurishkuro, line 93: is this condition specific to Windows? Maybe this whole additional logic should be scoped to windows builds only?

pkg/config/tlscfg/options_unix.go
```diff
@@ -0,0 +1,23 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// +build !windows
+
+package tlscfg
+
+import "crypto/x509"
+
+func appendCerts(rootCAs *x509.CertPool) (*x509.CertPool, error) {
```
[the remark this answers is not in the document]
Ashmita152: I mostly copied the code from sensu-go PR but now reading about it it seems like the goal of `appendCerts` is to load all certificates that are installed in the machine.
On unix systems, SystemCertPool (https://golang.org/pkg/crypto/x509/#SystemCertPool) already does that so we are just defining the function without doing anything. 
On windows, since SystemCertPool doesn't load those installed certificates, we are doing it ourselves.

Ashmita152:
Hi @yurishkuro 

Being very new to golang, I am confused about what changes we need here, would be very helpful if you can guide me.

cmd/opentelemetry/go.sum
```diff
@@ -1024,8 +1024,10 @@ github.com/vektah/gqlparser v1.1.2/go.mod h1:1ycwN7Ij5njmMkPPAOaRFY4rET2Enx7IkVv
 github.com/vektra/mockery v0.0.0-20181123154057-e78b021dcbb5/go.mod h1:ppEjwdhyy7Y31EnHRDm1JkChoC7LXIJ7Ex0VYLWtZtQ=
 github.com/wadey/gocovmerge v0.0.0-20160331181800-b5bfa59ec0ad/go.mod h1:Hy8o65+MXnS6EwGElrSRjUzQDLXreJlzYLlWiHtt8hM=
 github.com/willf/bitset v1.1.3/go.mod h1:RjeCKbqT1RxIR/KWY6phxZiaY1IyutSBfGjNPySAYV4=
+github.com/xdg/scram v0.0.0-20180814205039-7eeb5667e42c h1:u40Z8hqBAAQyv+vATcGgV0YCnDjqSL7/q/JyPhhJSPk=
```
yurishkuro, line 1027: which version of Go are you using? My understanding recent `go mod` stopped adding these entries.

yurishkuro (commented):
So similar to how you split Windows and non-Windows logic in the new files, but you only implemented a narrow function there. I would start with L89
```
certPool, err := systemCertPool()
```
and basically implement this function differently in Win/Linux. Note that this is currently a pointer to a function, so the linux file will be simply
```
var systemCertPool = x509.SystemCertPool // to allow overriding in unit test
```

while in Windows file you can do whatever else is needed

cmd/opentelemetry/go.sum
Ashmita152: I am using latest go1.15.7. Let me look at it.

jpkrohling (approved):
Nice!

pkg/config/tlscfg/options.go
```diff
@@ -86,9 +86,9 @@ func (p *Options) Config(logger *zap.Logger) (*tls.Config, error) {
 
 func (p Options) loadCertPool() (*x509.CertPool, error) {
 	if len(p.CAPath) == 0 { // no truststore given, use SystemCertPool
-		certPool, err := systemCertPool()
+		certPool, err := createCertPool()
```
yurishkuro, line 89: I think more accurate name is loadSystemCertPool

pkg/config/tlscfg/options_unix.go
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) 2021 The Jaeger Authors.
```
yurishkuro, line 1: File name certpool_unix

pkg/config/tlscfg/options.go
```diff
@@ -86,9 +86,9 @@ func (p *Options) Config(logger *zap.Logger) (*tls.Config, error) {
...
 		if err != nil {
-			return nil, fmt.Errorf("failed to load SystemCertPool: %w", err)
+			return nil, fmt.Errorf("failed to create CertPool: %w", err)
```
yurishkuro, line 91: I think the original error message was more accurate

pkg/config/tlscfg/options_unix.go
```diff
@@ -0,0 +1,30 @@
...
+	"crypto/x509"
+	"fmt"
+)
+
+func createCertPool() (*x509.CertPool, error) {
+	certPool, err := systemCertPool()
+	if err != nil {
+		return nil, fmt.Errorf("failed to load SystemCertPool: %w", err)
+	}
+	return certPool, nil
```
yurishkuro, line 29:
```suggestion
	return systemCertPool()
```

pkg/config/tlscfg/options_windows.go
```diff
@@ -0,0 +1,78 @@
...
+			rootCAs.AddCert(c)
+		}
+	}
+	return rootCAs, nil
+}
+
+func createCertPool() (*x509.CertPool, error) {
+	certPool, err := systemCertPool()
+	if err != nil {
+		return nil, fmt.Errorf("failed to load SystemCertPool: %w", err)
```
yurishkuro, line 68:
```suggestion
		return nil, err
```

pkg/config/tlscfg/options_windows.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+	if certPool == nil {
+		certPool = x509.NewCertPool()
+	}
+	certPool, err = appendCerts(certPool)
+	if err != nil {
+		return nil, fmt.Errorf("failed to append SystemCertPool: %w", err)
+	}
+	return certPool, nil
```
yurishkuro, line 77:
```suggestion
	return appendCerts(certPool)
```

yurishkuro (commented):
A bit of cleanup

Ashmita152:
Nice feedbacks Yuri. I can feel the redundant error checking I was doing before. Much cleaner now.

yurishkuro (approved):
thanks!

roederja2:
I'm running version 1.22.0 but I still get failed to load TLS config: failed to load CA CertPool: failed to load SystemCertPool: crypto/x509: system root pool is not available on Window. Specifying --reporter.grpc.tls.ca with an appropriate .pem file fixes it

Ashmita152:
Hi @roederja2 

Thank you for reporting the issue. We will look into it.

roederja2:
Thanks. For completeness these are my arguments:
Doesn't work:
`--reporter.grpc.host-port=dns:///jaeger-collector:443 --reporter.grpc.tls.enabled=true --reporter.grpc.tls.skip-host-verify=true`

works:
`--reporter.grpc.host-port=dns:///jaeger-collector:443 --reporter.grpc.tls.enabled=true --reporter.grpc.tls.skip-host-verify=true --reporter.grpc.tls.ca=C:/Jaeger/ca-certs.pem`

yury-kozlov:
> I can confirm #2550 is still not fixed. I'm running the same arguments as @roederja2.
> Maybe #2550 should be re-opened?

@DuckyC make sure you run agent with parameters: `--reporter.grpc.tls.enabled=true` and `--reporter.grpc.tls.ca=path-to-your-pem` with pem containing required certificate

I had troubles when running agent with additional arguments `reporter.grpc.tls.cert` and  `reporter.grpc.tls.key`, but when I removed them and used only `reporter.grpc.tls.enabled` and `reporter.grpc.tls.ca` everything worked fine.

DuckyC:
Yeah, I ended up having to do that. But the CA is already trusted by the default certificate store in windows.
I had to download the LetsEncrypt CA and point it to that.

PR #2757, opened 2021-02-01
Divide build-binaries job to multiple jobs per platform

Ashmita152:

Right now build-binaries is taking around 30 minutes to complete. This should do this in parallel.

Moving each platform binaries build into separate job.

### DIFF
--- a/.github/workflows/ci-build-binaries.yml
+++ b/.github/workflows/ci-build-binaries.yml
@@ -9,6 +9,22 @@ on:
 jobs:
   build-binaries:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        platform:
+        - name: linux
+          task: build-binaries-linux
+        - name: windows
+          task: build-binaries-windows
+        - name: osx
+          task: build-binaries-darwin
+        - name: system/390
+          task: build-binaries-s390x
+        - name: arm
+          task: build-binaries-arm64
+        - name: powerpc
+          task: build-binaries-ppc64le
+    name: binaries ${{ matrix.platform.name }}
     steps:
     - uses: actions/checkout@v2
       with:
@@ -29,4 +45,4 @@ jobs:
       run: make install-ci
 
     - name: Build binaries
-      run: make build-all-platforms
+      run: make ${{ matrix.platform.task }}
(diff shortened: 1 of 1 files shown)


yurishkuro (dismissed):
Sweet!

 * Build binaries / binaries linux (pull_request) Successful in **9m**

.github/workflows/ci-build-binaries.yml
```diff
@@ -9,6 +9,22 @@ on:
...
+          task: build-binaries-windows
+        - name: osx
+          task: build-binaries-darwin
+        - name: system/390
+          task: build-binaries-s390x
+        - name: arm
+          task: build-binaries-arm64
+        - name: powerpc
+          task: build-binaries-ppc64le
+    name: binaries ${{ matrix.platform.name }}
```
yurishkuro, line 27:
```suggestion
    name: build binaries for ${{ matrix.platform.name }}
```

yurishkuro:
One unfortunate side effect is that it explodes the number of jobs we need to configure as `required` for the branch to pass:
image

.github/workflows/ci-build-binaries.yml
yurishkuro: nvm lets keep as is

Ashmita152:
> One unfortunate side effect is that it explodes the number of jobs we need to configure as required for the branch to pass:
I see. Hopefully we don't hit any limit imposed by github.

Yuri, I have one question: I remember you mentioning that jaegertracing org should get more credits as compared to personal open-source project. But from my experience I haven't seen any difference in the build times for a job in my jaeger's fork vs same build in jaeger.

yurishkuro:
> I remember you mentioning that jaegertracing org should get more credits as compared to personal open-source project.

I am not sure about that. This [page][1] shows 2000 minutes for both personal and orgs. I am not sure if this is enforced yet, because I would've thought we easily exceed 2000 minutes.

It would be nice if GHA were able to automatically cancel running jobs for stale commits, we're burning through those a lot.

[1]: https://docs.github.com/en/github/setting-up-and-managing-billing-and-payments-on-github/about-billing-for-github-actions

Ashmita152:
> It would be nice if GHA were able to automatically cancel running jobs for stale commits, we're burning through those a lot.

I will look into it. Thinking if we can do something like this:

- Create a custom GHA to kill all GHA.
- Add the above custom action as a step if any of the task fails in the jobs which are marked as `required`.

yurishkuro:
I am not sure about the 2nd point. For the first one, there are already GHAs that kill other runs, but when I looked at few of them none were particularly great (may have changed since then)

Issue #2738, opened 2021-01-22
Remove deprecated CLI flags


yurishkuro:
We've accumulated a lot of deprecated flags, it would be good to remove them. Need to verify some of these, but I think all of them are past their expiration date. 

We **should not** clean them in a single PR since they usually require code changes, better to do one PR per flag.

* [x] admin flags #2752
* [x] collector raw ports #2751
* [x] agent tags #2753
* [x] queryPortWarning & queryHOSTPortWarning #2772
* [x] cassandra flags #2789
* [x] es.max-num-spans #2482
* [x] tls flags #2790 

```
$ grep -rin '(deprecated' .
./cmd/flags/admin.go:39:	healthCheckHTTPPortWarning = "(deprecated, will be removed after 2020-03-15 or in release v1.19.0, whichever is later)"
./cmd/flags/admin.go:40:	adminHTTPPortWarning       = "(deprecated, will be removed after 2020-06-30 or in release v1.20.0, whichever is later)"
./cmd/collector/app/builder_flags.go:45:	collectorHTTPPortWarning       = "(deprecated, will be removed after 2020-06-30 or in release v1.20.0, whichever is later)"
./cmd/collector/app/builder_flags.go:46:	collectorGRPCPortWarning       = "(deprecated, will be removed after 2020-06-30 or in release v1.20.0, whichever is later)"
./cmd/collector/app/builder_flags.go:47:	collectorZipkinHTTPPortWarning = "(deprecated, will be removed after 2020-06-30 or in release v1.20.0, whichever is later)"
./cmd/agent/app/reporter/flags.go:51:		flags.String(AgentTagsDeprecated, "", "(deprecated) see --"+agentTags)
./cmd/query/app/flags.go:42:	queryPortWarning        = "(deprecated, will be removed after 2020-08-31 or in release v1.20.0, whichever is later)"
./cmd/query/app/flags.go:43:	queryHOSTPortWarning    = "(deprecated, will be removed after 2020-12-31 or in release v1.21.0, whichever is later)"
./plugin/storage/cassandra/options.go:237:		"(deprecated) Jaeger will automatically detect the version of the dependencies table")
./plugin/storage/cassandra/options.go:241:		"(deprecated) Enable (or disable) host key verification. Use "+nsConfig.namespace+".tls.skip-host-verify instead")
[...]

PR #2772, opened 2021-02-03
Removing deprecated Query Server flags

rjs211:

- resolves part of #2738 

- The deprecated `query.port` and `query.host-port` were removed.
- The redundant test cases were removed
- Even though the behaviour remaiins the same, the code logic is simplified.


rjs211:
The new logic is as follows:

  *  If TLS in enabled in either or both of GRPC or HTTP endpoints, `--query.http-server.host-port`defaults to `:16686` and  `--query.grpc-server.host-port` defaults to `:16685`  
  *  If TLS is disabled in both endpoints (default), both `--query.http-server.host-port` and  `--query.grpc-server.host-port` default to `:16686`  
  
This ensures that if the user doesn't specify the host-ports, the server would still initialize, for both TLS and non-TLS cases, conforming to the existing behaviour.

ports/ports.go
```diff
@@ -38,7 +38,8 @@ const (
 	// CollectorAdminHTTP is the default admin HTTP port (health check, metrics, etc.)
 	CollectorAdminHTTP = 14269
 
-	// QueryGRPC is the default port of GRPC requests for Query trace retrieval
+	// QueryGRPC is the default port of GRPC requests for Query trace retrieval when TLS is enabled.
+	// When TLS is disabled, QueryHTTP is used for both GRPC and HTTP endpoints
```
yurishkuro, line 42:
```suggestion
	// When TLS is disabled and GRPC port not provided, QueryHTTP is used for both GRPC and HTTP endpoints
```

yurishkuro (commented):
Please add an entry to the changelog (and correct the previous one if applicable)

CHANGELOG.md
```diff
@@ -21,12 +23,10 @@ Changes by Version
...
-
+* Add TLS Support for GRPC and HTTP endpoints of the Query server ([#2337](https://github.com/jaegertracing/jaeger/pull/2337), [#2772](https://github.com/jaegertracing/jaeger/pull/2772), [@rjs211](https://github.com/rjs211))
 
-    If TLS is disabled in both endpoints (default), dedicated host-ports are used if either or both of `--query.http-server.host-port` or `--query.grpc-server.host-port` is explicitly set. If only one of the dedicated host-port flags is set, the other flag uses its corresponding default value. If neither of the dedicated host-port flags are explicitly set, the common host-port infered from `--query.host-port` (defaults to `:16686`) is used for both the GRPC and HTTP endpoints.
+  *  If TLS in enabled in either or both of GRPC or HTTP endpoints, `--query.http-server.host-port`defaults to `:16686` and  `--query.grpc-server.host-port` defaults to `:16685`  
+  *  If TLS is disabled in both endpoints (default), both `--query.http-server.host-port` and  `--query.grpc-server.host-port` default to `:16686`
```
yurishkuro, line 29: I am a bit concerned that this will be confusing. Both new flags have default values (different ones), so the traditional behavior of the program when a flag is not specified explicitly is to use the default. However, here we're saying that even though the `help` command will show `:16685` as default, the actual server is going to run on the shared `:16686` port.

Perhaps we could just move this into breaking changes section and always treat the ports as separate, _unless_ the same value is given in the config (which should only work for non-TLS).

@jpkrohling thoughts?

jpkrohling: > Perhaps we could just move this into breaking changes section and always treat the ports as separate, unless the same value is given in the config (which should only work for non-TLS).

If we are breaking, then I'd force people to always use a different port. It makes the code easier to maintain, and I believe this is a good practice anyway...

yurishkuro: I think that would be too "breaking". It's one thing to to change the flags, it's another to completely remove the ability to run on the same port. The same-port CMUX approach seems to work fine for non-TLS, I would rather keep it.

jpkrohling: Sounds good to me.

yurishkuro (approved):
THanks!

PR #2773, opened 2021-02-03
Clean up go.sum to support building with Go 1.16

### LINKED ISSUE #2748: Try building with upcoming Go 1.16
There are some changes to default handling of modules in Go 1.16 (https://tip.golang.org/doc/go1.16), we need to verify that out build still works.

gsagula:
- It allows building Jaeger with upcoming Go version 1.16. 

- Ran `mod tidy` in `./cmd/opentelemetry` in order to build with 1.16. 

Resolves #2748 

Signed-off-by: Gabriel Linden Sagula <glindens@bitfield.org>


yurishkuro:
You're not upgrading any packages. Is it just cleaning up those `h1:` entries from the lock file?

gsagula:
> You're not upgrading any packages. Is it just cleaning up those `h1:` entries from the lock file?

Sorry, wrong description. I had initially upgraded one package but it was not necessary in the end.

yurishkuro (approved):
Thanks!

PR #2782, opened 2021-02-03
Remove downsample flags from the query service

### LINKED ISSUE #2727: Downsampling in query service?
It seems we have the wrong scope of these flags, what meaning do they have for the query service?
```
$ SPAN_STORAGE_TYPE=memory go run ./cmd/query help
...
      --downsampling.hashsalt string               Salt used when hashing trace id for downsampling.
      --downsampling.ratio float                   Ratio of spans passed to storage after downsampling (between 0 and 1), e.g ratio = 0.3 means we are keeping 30% of spans and dropping 70% of spans; ratio = 1.0 disables downsampling. (default 1)
...
```

joe-elliott:
- Fixes #2727 

- Splits the storage factory AddFlags method into two:  `AddFlags` and `AddFlagsPipeline` with the latter meant to be used by Jaeger span pipeline processors (ingesters, distributors).  
- Add appropriate tests


plugin/storage/factory.go
```diff
@@ -201,6 +208,14 @@ func (f *Factory) InitFromViper(v *viper.Viper) {
 }
 
 func (f *Factory) initDownsamplingFromViper(v *viper.Viper) {
+	// if the downsampling flag isn't set then this component used the standard "AddFlags" method
+	// and has no use for downsampling.  the default settings effectively disable downsampling
+	if !v.IsSet(downsamplingRatio) {
```
yurishkuro, line 213: to clarify, `v.IsSet` does not mind if the flag was not even declared? I thought `IsSet` refers to whether the user set it or not.

yurishkuro: perhaps a cleaner alternative is to have a field in the factory that we set to true when addDownsamplingFlags is called

joe-elliott: I believe you're right about `IsSet`.  I don't like adding state here, but I've played with some of these other methods in viper (AllSettings, AllKeys) and they are more complex than they appear on the surface.  Agree the cleanest implementation here will be adding a bool

yurishkuro (approved):

Issue #2498, opened 2020-09-21
Push official images to quay.io, in addition to Docker Hub


aditya-konarde:
# Requirement - what kind of business use case are you trying to solve?

High availability and resiliency of Jaeger in production environments

## Problem - what in Jaeger blocks you from solving the requirement?

Currently, the official images for jaeger are on docker hub. Unless one has a docker paid account, image pulls will be rate-limited and pulls may fail at run-time, causing infrastructure issues.

## Proposal - what do you suggest to solve the problem or improve the existing situation?

$title: Push official images to quay.io, in addition to Docker Hub

We already have an organization where we push the jaeger operator https://quay.io/repository/jaegertracing/jaeger-operator

I propose that we also push all the other images to quay.io, in addition to the docker hub repository here: https://hub.docker.com/u/jaegertracing

## Any open questions to address

None

jpkrohling:
I'm OK with this idea. @yurishkuro, what do you think?

yurishkuro:
_in addition_ sounds fine.

aditya-konarde:
Thanks, I'll start to work on it this week :)

aebirim:
Hi @aditya-konarde, @jpkrohling 

I'm an Outreachy applicant and would like to tackle this issue.

aditya-konarde:
Sounds good. I haven't gotten the time to work on this yet, so please go ahead @aebirim :)

aebirim:
Hi @aditya-konarde will do!

aebirim:
Hi @jpkrohling I've uploaded the 41 images to the quay.io platform. Some of the latest tags didn't work so used the next most recent one. 

I've sent you a message on gitter.

jpkrohling:
@aebirim, ideally, we would push the images to both places during the release process. It's currently called during a Travis build here:

https://github.com/jaegertracing/jaeger/blob/0f88343172c0bb16ad5c1805d18e20d59176ea7e/.travis.yml#L73-L74

Take a look and see if you can follow what it's doing end-to-end, but I think you might need to change only two places:

1. `scripts/travis/upload-all-docker-images.sh`, to call the `upload-to-docker.sh` for both `docker.io` and for `quay.io`
1. `scripts/travis/upload-to-docker.sh`, to accommodate the case where images are pushed to `quay.io`

morlay:
could we push github contrainer registry `ghcr.io/jaegertracing/*` too?

yurishkuro:
[...]

PR #2783, opened 2021-02-04
Add quay support

Ashmita152:

I am still trying to figure out ways to test it properly before merging. Looking for feedback to validate the approach.

Solves https://github.com/jaegertracing/jaeger/issues/2498

- All docker images are now built with two tags one for dockerhub and another for quay.io.
- All docker images are now uploaded to both dockerhub and quay.io.
- All github actions now try to login to quay.io registry.


yurishkuro:
Thanks @Ashmita152 .

To my taste, there's too much repetition. What if we encapsulate the docker vs quay only within the scripts/travis/upload-to-docker.sh script? I don't think anything above it needs to know about where the images are being uploaded. And ideally we would even move the logging in steps into the same script, as there's no reason to do it before the actual upload decision is made. I assume we can pass the login/password via env and it won't be leaked to the logs (we should verify that).

Ashmita152:
Hi @yurishkuro 

Thank you for the feedback. I see your point. Let me think about it.

Ashmita152:
Hi @yurishkuro 

I have updated the PR with your suggestion. I hope this time, I took the expected approach. 

Two things we need to do before merging:
1) create jaegertracingbot user in quay.io and allow it to be able to upload images to jaegertracing quay org.
2) get its token and set QUAY_TOKEN envvar in github.

Regards.

yurishkuro:
did you try pushing the login steps into `scripts/travis/upload-to-docker.sh`? Before GHA we used to do login manually, and if it is done in that script, we do not need to keep adding the login steps in all the GHA workflows.

yurishkuro:
@Ashmita152 see my test PR #2797. This is the GH run: https://github.com/jaegertracing/jaeger/runs/1846319012?check_suite_focus=true

The env var is never printed in the logs when logging into docker. We can remove all the login steps from the workflows.

Ashmita152:
Sure Yuri, sounds good. I will update it.

Ashmita152:
Hi @yurishkuro 

I just realised that the DOCKERHUB_TOKEN is exposed in the above GitHub Action run you shared.

Screenshot 2021-02-07 at 9 41 48 PM

I agree with you that we need to cleanup this login. I am trying to find a way to move dockerhub login and quay login as a separate GitHub Action if possible.

Ashmita152:
Oh I misunderstood your point. You are right, it will never show up in logs.

Ashmita152:
Hi Yuri,

I decided to keep the dockerhub+quay login part as a separate github action mainly because of two reasons:
- This way those tokens are only visible to this github action script, which I felt more safer since it won't change often.
- Our upload-to-docker.sh isn't very neat so I didn't want to add login part there.

Let me know in case you feel otherwise.

scripts/travis/build-all-in-one-image.sh
```diff
@@ -36,13 +37,13 @@ run_integration_test() {
 }
 
 upload_to_docker() {
-  # Only push the docker container to Docker Hub for master/release branch and when dockerhub login is done
-  if [[ ("$BRANCH" == "master" || $BRANCH =~ ^v[0-9]+\.[0-9]+\.[0-9]+$) && "$DOCKERHUB_LOGIN" == "true" ]]; then
-    echo "upload $1 to Docker Hub"
+  # Only push the docker image to dockerhub/quay.io for master/release branch and when dockerhub or quay.io login is done
+  if [[ ("$BRANCH" == "master" || $BRANCH =~ ^v[0-9]+\.[0-9]+\.[0-9]+$) && ("$DOCKERHUB_LOGIN" == "true" || "$QUAY_LOGIN" == "true") ]]; then
```
yurishkuro, line 41: I still think we need to push to upload script the login-related checks. If in the future we need to add one more registry, I only want to have to change the upload script. The var checks are not essential here, and we didn't have them before GHA.

scripts/travis/upload-to-docker.sh
```diff
@@ -22,23 +25,44 @@ echo "REPO=$REPO, BRANCH=$BRANCH, TAG=$TAG, IMAGE=$IMAGE"
 # add major, major.minor and major.minor.patch tags
 if [[ -n ${major:-} ]]; then
   docker tag $IMAGE $REPO:${major}
+  docker tag $IMAGE ${QUAY_PREFIX}/$REPO:${major}
   if [[ -n ${minor:-} ]]; then
     docker tag $IMAGE $REPO:${major}.${minor}
+    docker tag $IMAGE ${QUAY_PREFIX}/$REPO:${major}.${minor}
```
yurishkuro, line 31: This doesn't look quite right - this tagging will apply to Docker as well? Maybe we can simply prefix the REPO at the beginning once if quay login is in effect, and run the same set of commands otherwise. Let's separate the logic of the script from the destination registry.

scripts/travis/build-all-in-one-image.sh
Ashmita152: If I remember correctly, those DOCKERHUB_LOGIN var check wasn't there in Travis days because we used to set DOCKERHUB_TOKEN from Travis UI. After moving to GHA, we set them Github org secrets which aren't accessible in case of PRs that's why we added it. My understanding is if we remove those checks and if someone raises a PR from their fork's master branch, it will try to push images to dockerhub and fail.

scripts/travis/upload-to-docker.sh
Ashmita152: If I get it correctly, we should make the script upload-to-docker.sh to accept a parameter with acceptable values as "dockerhub" or "quay" and should remove all if dockerhub/quay conditions and make the script agnostic of the upstream registry ?

yurishkuro: this is what I am suggesting:

1. In the workflows, we call `upload-to-docker.sh` and pass secrets as env vars (instead of having a dedicated login action)
2. `upload-to-docker.sh` looks like this:

```sh
function do_upload() {
    registry = $1
    prefix = $2
    token_var = $3

    marker=.$registry.login
    if [ ! -f $marker && -v $token_var ]; then
        printenv $token_var | docker login ...
        touch marker
    fi

    if [ ! -f $marker ]; then
        echo "skipping upload to $registry, not logged in:
        return
    fi

    # current content of upload-to-docker.sh before this PR
}

do_upload "docker.io" "" "DOCKERHUB_TOKEN"
do_upload "quay.io" "quay prefix" "QUAY_TOKEN"
```

This way, the _only_ thing that individual actions need to do about logging in is pass the token as env. The login logic is encapsulated into the upload script, the fact that it does docker and quay is also encapsulated, and the login is efficient by using a cache "marker" file

Ashmita152: Sounds great Yuri.

jpkrohling:
A couple of comments:

- the username should be externalized, as Quay supports robot accounts tied to the organization, which aren't real accounts. The username would then be something like `jaegertracing+github_workflows`
- you can assume that all repositories are prefixed, like: `docker.io`, `quay.io`, ...

jpkrohling:
I just added the following as "repository secrets"

image

The DOCKERHUB_TOKEN and JAEGERTRACINGBOT_GITHUB_TOKEN were already set as organization secrets.

scripts/travis/hotrod-integration-test.sh
```diff
@@ -18,14 +18,19 @@ fi
...
 fi
 
-bash ./scripts/travis/upload-to-docker.sh
+DOCKERHUB_USERNAME=${DOCKERHUB_USERNAME:-}
+DOCKERHUB_TOKEN=${DOCKERHUB_TOKEN:-}
+QUAY_USERNAME=${QUAY_USERNAME:-}
+QUAY_TOKEN=${QUAY_TOKEN:-}
+
+bash scripts/travis/upload-to-registry.sh "docker.io" $REPO $DOCKERHUB_USERNAME $DOCKERHUB_TOKEN
+bash scripts/travis/upload-to-registry.sh "quay.io" $REPO $QUAY_USERNAME $QUAY_TOKEN
```
yurishkuro, line 36:
```suggestion
bash scripts/travis/upload-to-registry.sh
```

This is the only statement we need here. The enumeration of repositories should be hidden INSIDE `upload-to-registry.sh`.

scripts/travis/upload-to-registry.sh
```diff
@@ -0,0 +1,109 @@
...
+
+try_login() {
+  local registry=$1
+  local user=$2
+  local token=$3
+  local marker=.${registry}.login
+
+  if [ ! -f ${marker} ]; then
+    printenv ${token}  | docker login ${registry} --username ${user} --password-stdin
+    if [ $? -eq 0 ]; then
```
yurishkuro, line 61: is this check necessary? You have `set -e -pipefail`, so the script should simply fail on error

scripts/travis/upload-all-docker-images.sh
```diff
@@ -5,13 +5,12 @@
 set -euxf -o pipefail
 
 BRANCH=${BRANCH:?'missing BRANCH env var'}
-DOCKERHUB_LOGIN=${DOCKERHUB_LOGIN:-false}
 
-# Only push images to Docker Hub for master branch or for release tags vM.N.P and when dockerhub login is done
-if [[ ("$BRANCH" == "master" || $BRANCH =~ ^v[0-9]+\.[0-9]+\.[0-9]+$) && "$DOCKERHUB_LOGIN" == "true" ]]; then
-  echo "upload to Docker Hub, BRANCH=$BRANCH"
+# Only push images to dockerhub/quay.io for master branch or for release tags vM.N.P
+if [[ "$BRANCH" == "master" || $BRANCH =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
```
yurishkuro, line 10: seems like every script is doing this check, so could be another candidate to move into upload-to-registry

scripts/travis/hotrod-integration-test.sh
yurishkuro: also, passing secrets like this is not safe, they will be printed. But there's no need to pass them, upload-to-registry already has access to env.

scripts/travis/upload-to-registry.sh
```diff
@@ -0,0 +1,113 @@
...
+      fi
+    fi
+  fi
+}
+
+try_login() {
+  local registry=$1
+  local user=$2
+  local token=$3
+  local marker=.${registry}.login
```
yurishkuro, line 57: this is better taken as `$4` parameter to remove cross-function dependency on the marker file name

scripts/travis/upload-to-registry.sh
```diff
@@ -0,0 +1,113 @@
...
+      *)
+        docker push ${registry}/${image}
+        ;;
+    esac
+  fi
+
+  # upload snapshot images
+  if [ ! -f ${marker} ]; then
+    echo "skipping upload to ${registry}, not logged in!"
+  else
```
yurishkuro, line 91: this is redundant, I would just merge the `else` parts of these two if statements

yurishkuro (commented):
Looks great! Almost all redundancy is eliminated, should be much easier to maintain going forward.

scripts/travis/upload-to-registry.sh
```diff
@@ -0,0 +1,113 @@
...
+    echo "pushing snapshot image ${snapshot_image}"
+    docker tag ${image} ${registry}/${snapshot_image}
+    docker push ${registry}/${snapshot_image}
+  fi
+}
+
+main() {
+  check_args "$@"
+
+  read -r tag major minor patch <<< "$(compute_image_tag ${BRANCH})"
```
yurishkuro, line 102: I've never seen triple `<<<` - what does it mean?

scripts/travis/upload-to-registry.sh
```diff
@@ -0,0 +1,113 @@
...
+  label_release_tag $1 $2 ${major} ${minor} ${patch}
+  upload_images $1 $2 $3 $4
```
yurishkuro, line 104: what are numbered params refer to here? can we use symbolic names?

scripts/travis/upload-to-registry.sh
```diff
@@ -0,0 +1,113 @@
...
+}
+
+DOCKERHUB_USERNAME=${DOCKERHUB_USERNAME:-}
+DOCKERHUB_TOKEN=${DOCKERHUB_TOKEN:-}
+QUAY_USERNAME=${QUAY_USERNAME:-}
+QUAY_TOKEN=${QUAY_TOKEN:-}
+
+main "docker.io" $1 ${DOCKERHUB_USERNAME} ${DOCKERHUB_TOKEN}
+main "quay.io" $1 ${QUAY_USERNAME} ${QUAY_TOKEN}
```
yurishkuro, line 113: I think you need
```suggestion
main "docker.io" $1 ${DOCKERHUB_USERNAME} DOCKERHUB_TOKEN
main "quay.io" $1 ${QUAY_USERNAME} QUAY_TOKEN
```

Ashmita152: I didn't get this part Yuri. Wont passing them without $ will cause token set as the string "DOCKERHUB_TOKEN" instead of evaluating its value.

scripts/travis/upload-to-registry.sh
Ashmita152: I mostly copy pasted from internet. Seems like is called "Here Strings" from bash manpage.

```
   Here Strings
       A variant of here documents, the format is:

              <<<word

       The word is expanded and supplied to the command on its standard input.
```

scripts/travis/upload-to-registry.sh
yurishkuro: If you pass them as `${QUAY_TOKEN}` then the values will be printed to the log. The idea here is to pass this by-reference, not by-value, i.e. we pass only the string that contains the variable name `"QUAY_TOKEN"`, and then we use that with `printenv` to access the actual value or `-v` to test for value. This way the value is never passed directly and will never be seen in the logs.

scripts/travis/upload-to-registry.sh
```diff
@@ -0,0 +1,112 @@
...
+  fi
+}
+
+main() {
+  local registry=$1
+  local image=$2
+  local user=$3
+  local token=$4
+
+  check_args "$@"
```
yurishkuro, line 99: this call should be outside of the function, since you're checking params for the whole script, which only needs to happen once.

scripts/travis/upload-to-registry.sh
```diff
@@ -0,0 +1,112 @@
...
+
+  read -r tag major minor patch <<< "$(compute_image_tag ${BRANCH})"
+  label_release_tag ${registry} ${image} ${major} ${minor} ${patch}
+  upload_images ${registry} ${image} ${user} ${token}
+}
+
+DOCKERHUB_USERNAME=${DOCKERHUB_USERNAME:-}
+DOCKERHUB_TOKEN=${DOCKERHUB_TOKEN:-}
+QUAY_USERNAME=${QUAY_USERNAME:-}
+QUAY_TOKEN=${QUAY_TOKEN:-}
```
yurishkuro, line 109:
```suggestion
DOCKERHUB_USERNAME=${DOCKERHUB_USERNAME:-"jaegertracingbot"}
DOCKERHUB_TOKEN=${DOCKERHUB_TOKEN:-}
QUAY_USERNAME=${QUAY_USERNAME:-"jaegertracingbot+github_workflows"}
QUAY_TOKEN=${QUAY_TOKEN:-}
```

* names are not passed from action so need defaults
* please move these statements to the top of the script, it makes it explicit what expectations on env it has

scripts/travis/upload-to-registry.sh
```diff
@@ -0,0 +1,112 @@
...
+  fi
+}
+
+try_login() {
+  local registry=$1
+  local user=$2
+  local token=$3
+  local marker=$4
+
+  if [ ! -f ${marker} ]; then
```
yurishkuro, line 59:
```suggestion
  if [ ! -f ${marker} && -v ${token} ]; then
```

(assuming that `&&` syntax is correct). This makes the script tolerant when one of the registry tokens is not passed.

scripts/travis/upload-to-registry.sh
```diff
@@ -0,0 +1,112 @@
...
+upload_images() {
+  local registry=$1
+  local image=$2
+  local user=$3
+  local token=$4
+  local marker=.${registry}.login
+
+  try_login ${registry} ${user} ${token} ${marker}
+
+  if [ ! -f ${marker} ]; then
```
yurishkuro, line 74:
```suggestion
  if [ ! -f ${marker} && -v ${token} ]; then
```

yurishkuro (commented):
It would be good if you try running the upload script manually (you can build one of the images locally and try to push it to your personal account at quay/dockerhub) and include a paste of the execution logs.

Ashmita152:
Thank you for the feedbacks. Sure Yuri, I will test this PR from my fork properly both from merge to master + release perspective and share with the links.

scripts/travis/upload-to-registry.sh
```diff
@@ -0,0 +1,112 @@
...
+upload_images() {
+  local registry=$1
+  local image=$2
+  local user=$3
+  local token=$4
+  local marker=.${registry}.login
+
+  try_login ${registry} ${user} ${token} ${marker}
+
+  if [ ! -f ${marker} && -v ${token} ]; then
```
yurishkuro, line 79: my bad, don't need `-v` here

yurishkuro (dismissed):

jpkrohling:
Nice!

yurishkuro:
I think this is almost ready to merge (just one tiny fix), but it would be nice to see a test run logs. Of course, we could merge and see if the publishing succeeds and fix forward.

Ashmita152:
Right, I am still testing with my fork. I will share the result once I am done testing.

scripts/travis/upload-to-registry.sh
```diff
@@ -0,0 +1,112 @@
...
+upload_images() {
+  local registry=$1
+  local image=$2
+  local user=$3
+  local token=$4
+  local marker=.${registry}.login
+
+  try_login ${registry} ${user} ${token} ${marker}
+
+  if [ ! -f ${marker} ] && [ -z ${token} ]; then
```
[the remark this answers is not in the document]
yurishkuro: another way to refactor this is to do something like this:
```
if [ try_login ${registry} ${user} ${token} ] ; then
    # do upload
else
    # echo no upload
fi
```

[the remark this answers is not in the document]
yurishkuro: This way the marker can be hidden inside try_login

Ashmita152:
Hi @yurishkuro 

I can confirm that this PR works as expected. I tested by merging my branch in my fork.

I wasn't able to make the quay's robot user work for pushing to quay image and keep getting this error while pushing from robot account.

```
unauthorized: access to the requested resource is not authorized
``` 

Then I used the main quay user account. 

I found one bug and fixed it here: https://github.com/jaegertracing/jaeger/pull/2783/commits/3b2ea2e2a742501d0ea1624533661d4bfcc91ff6

Otherwise for quay.io, 
```
docker push ${registry}/${image}
```
This will fail because image will be missing because we was never labelling the latest tag for quay.

All the GH actions completed successfully: https://github.com/Ashmita152/jaeger/runs/1873169008
All Quay Images: https://quay.io/organization/ashmita1 (quay by default creates private images so you won't be able to see many images)
All Dockerhub Images: https://hub.docker.com/u/ashmita1

Regards.

scripts/travis/upload-to-registry.sh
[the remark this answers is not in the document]
Ashmita152: Trying it.

yurishkuro (approved):

yurishkuro:
Given the positive test evidence, I think it's time to merge, and see if it works with real credentials / repos.

scripts/travis/upload-to-registry.sh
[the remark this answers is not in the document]
Ashmita152: I am little confused about this change. If we use this way, let's say we don't specify token for one of the registries, won't it will still try to upload to that registry since we no longer checking for marker file and deciding based on the return status of try_login status.

yurishkuro:
@jpkrohling any thoughts?
```
printenv QUAY_TOKEN | docker login quay.io --username jaegertracingbot+github_workflows --password-stdin
Error response from daemon: Get https://quay.io/v2/: unauthorized: Could not find robot with specified username
```

scripts/travis/upload-to-registry.sh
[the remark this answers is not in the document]
yurishkuro: In this specific location, if the marker file does not exist, it means we did not log in, so no point in trying to upload. That's why the extra `&& -v $token` condition was wrong.

Otherwise, if you're asking about the `if [ try_login ]` version, then the function should return true only if it successfully logged in, i.e. it needs to be modified slightly to return false when token is not defined.

I wouldn't worry about it, lets leave it as it is now.

jpkrohling:
Oops, it should be `jaegertracing+github_workflows` (no `bot`) I updated the env var QUAY_USERNAME with the correct username.

scripts/travis/upload-to-registry.sh
```diff
@@ -0,0 +1,113 @@
+#!/bin/bash
+
+set -euxf -o pipefail
+
+DOCKERHUB_USERNAME=${DOCKERHUB_USERNAME:-"jaegertracingbot"}
+DOCKERHUB_TOKEN=${DOCKERHUB_TOKEN:-}
+QUAY_USERNAME=${QUAY_USERNAME:-"jaegertracingbot+github_workflows"}
```
jpkrohling, line 7: Looks like this isn't being take from the env var set as secret in the project. The right username for this is `jaegertracing+github_workflows`

PR #2789, opened 2021-02-05
Remove deprecated cassandra flags

albertteoh:

- Relates to #2738

- Removes deprecated cassandra flags

### DIFF
--- a/pkg/cassandra/config/config.go
+++ b/pkg/cassandra/config/config.go
@@ -44,7 +44,6 @@ type Configuration struct {
 	Port                 int            `yaml:"port" mapstructure:"port"`
 	Authenticator        Authenticator  `yaml:"authenticator" mapstructure:",squash"`
 	DisableAutoDiscovery bool           `yaml:"disable_auto_discovery" mapstructure:"-"`
-	EnableDependenciesV2 bool           `yaml:"enable_dependencies_v2" mapstructure:"-"`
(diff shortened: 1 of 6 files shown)


pkg/cassandra/config/config.go
```diff
@@ -44,7 +44,6 @@ type Configuration struct {
 	Port                 int            `yaml:"port" mapstructure:"port"`
 	Authenticator        Authenticator  `yaml:"authenticator" mapstructure:",squash"`
 	DisableAutoDiscovery bool           `yaml:"disable_auto_discovery" mapstructure:"-"`
-	EnableDependenciesV2 bool           `yaml:"enable_dependencies_v2" mapstructure:"-"`
```
yurishkuro, line 47: I expect there to be some code that actually reads this field, not seeing anything like that removed in the PR.

yurishkuro (commented):
please add changelog entry

pkg/cassandra/config/config.go
albertteoh: Did a recursive case-insensitive search in my jaeger root project dir and couldn't find `EnableDependenciesV2` or `enable_dependencies_v2`.

Maybe something wrong with my search, can you see any references?

yurishkuro: found it - the logic for that flag was already removed in #1364

yurishkuro (approved):

PR #2798, opened 2021-02-07
TLS support for HTTP Endpoint of Collector server

rjs211:
- solves part of #2249 

- independent TLS flags are exposed for gRPC and HTTP endpoints, enabling the user to provide different set of key, cert, CA-Cert , etc for each communication channal.

- provides the option of enabling TLS/mTLS in none, either one or both of HTTP and gRPC endpoints.

## Todo
- CHANGELOG.md entry


cmd/collector/app/collector.go
```diff
@@ -45,10 +45,11 @@ type Collector struct {
...
 	// state, read only
-	hServer    *http.Server
-	zkServer   *http.Server
-	grpcServer *grpc.Server
-	tlsCloser  io.Closer
+	hServer       *http.Server
+	zkServer      *http.Server
+	grpcServer    *grpc.Server
+	TLSGRPCCloser io.Closer
+	TLSHTTPCloser io.Closer
```
yurishkuro, line 52:
```suggestion
	tlsGRPCCertWatcherCloser io.Closer
	tlsHTTPCertWatcherCloser io.Closer
```

cmd/collector/app/builder_flags.go
```diff
@@ -40,12 +40,18 @@ const (
 	collectorZipkinAllowedHeaders = "collector.zipkin.allowed-headers"
 )
 
-var tlsFlagsConfig = tlscfg.ServerFlagsConfig{
+var tlsGRPCFlagsConfig = tlscfg.ServerFlagsConfig{
 	Prefix:       "collector.grpc",
 	ShowEnabled:  true,
```
yurishkuro, line 45: Perhaps we should remove this option, it is never false
```
$ grx 'ShowEnabled: ' .
./cmd/collector/app/builder_flags.go:45:	ShowEnabled:  true,
./cmd/agent/app/reporter/grpc/flags.go:36:	ShowEnabled:    true,
./cmd/query/app/flags.go:52:	ShowEnabled:  true,
./cmd/query/app/flags.go:58:	ShowEnabled:  true,
./plugin/storage/cassandra/options.go:250:		ShowEnabled:    true,
./plugin/storage/es/options.go:139:		ShowEnabled:    true,
./pkg/config/tlscfg/flags_test.go:51:				ShowEnabled:    true,
./pkg/config/tlscfg/flags_test.go:98:				ShowEnabled:  true,
./pkg/kafka/auth/config.go:94:		ShowEnabled:    true,
./pkg/kafka/auth/options.go:113:		ShowEnabled:    true,
```

yurishkuro (commented):
Overall lgtm. The unit tests are failing to compile though.

cmd/collector/app/builder_flags.go
rjs211: Do you want me to do this as a part of the same PR?

yurishkuro: up to you

yurishkuro (commented):
Code lgtm, but coverage `will decrease by 0.10%`. Can you add tests to exercise the new paths?

cmd/collector/app/server/http_test.go
```diff
@@ -58,3 +85,174 @@ func TestSpanCollectorHTTP(t *testing.T) {
...
+				HostPort:       fmt.Sprintf(":%d", ports.CollectorHTTP),
+				Handler:        handler.NewJaegerSpanHandler(logger, &mockSpanProcessor{}),
+				SamplingStore:  &mockSamplingStore{},
+				MetricsFactory: metricstest.NewFactory(time.Hour),
+				HealthCheck:    healthcheck.New(),
+				Logger:         logger,
+				TLSConfig:      test.TLS,
+			}
+			server, err := StartHTTPServer(params)
+			defer server.Close()
```
yurishkuro, line 216: linter is unhappy about missing error check

rjs211: I fixed and pushed it in my fork. for some reason, it is not reflected in the PR yet...

https://github.com/rjs211/jaeger/blob/a07970e548104d644238e17e9416b6521d4b2d2e/cmd/collector/app/server/http_test.go#L214-L221

yurishkuro:
any other tests could be added to increase the PR coverage above 95%?

cmd/collector/app/collector.go
```diff
@@ -165,8 +168,12 @@ func (c *Collector) Close() error {
...
 	}
 
-	if err := c.tlsCloser.Close(); err != nil {
-		c.logger.Error("failed to close TLS certificate watcher", zap.Error(err))
+	if err := c.tlsGRPCCertWatcherCloser.Close(); err != nil {
+		c.logger.Error("failed to close gRPC TLS certificate watcher", zap.Error(err))
+	}
+
+	if err := c.tlsHTTPCertWatcherCloser.Close(); err != nil {
+		c.logger.Error("failed to close HTTP TLS certificate watcher", zap.Error(err))
```
rjs211, line 176: @yurishkuro  According to [codecov report](https://codecov.io/gh/jaegertracing/jaeger/pull/2798/diff
) , only these errors are not covered. The rest are covered. I am not sure how to simulate this error.  Please let me know if you have a way to do so.

yurishkuro: I suggest replacing these with 
```
// watchers actually never return errors from Close
_ = c.tlsGRPCCertWatcherCloser.Close()
_ = c.tlsHTTPCertWatcherCloser.Close()
```

cmd/collector/app/server/http_test.go
```diff
@@ -58,3 +85,176 @@ func TestSpanCollectorHTTP(t *testing.T) {
...
+			},
+			expectError:       false,
+			expectServerFail:  false,
+			expectClientError: true,
+		},
+		{
+			name: "should pass with TLS client with cert to trusted TLS server requiring cert",
+			TLS: tlscfg.Options{
+				Enabled:      true,
+				CertPath:     testCertKeyLocation + "/example-server-cert.pem",
```
rjs211, line 167: This is an example of mTLS right? @davidcmitchell  @yurishkuro

yurishkuro: Yes. Once this PR lands, the client mTLS will be _technically_ feasible, but most Jaeger SDKs do not actually support attaching client certificates to requests (Java SDK might).

Also, mTLS seems like a lot of management overhead. If TLS from client is really necessary, I would consider using plain TLS in combination with plain auth in the headers, which I think Jaeger clients already support (but Jaeger backed ignores it afaik).

yurishkuro:
codeov is stuck again, but the report shows +0.04%

https://codecov.io/github/jaegertracing/jaeger/commit/03107bed1254f3fc1b5fa2c6ae6d833856412c64

rjs211:
@yurishkuro Actually im yet to make CHANGELOG.md entry. Is it possible to revert the commit? I dont want to waste another PR for that..

yurishkuro:
No need, as this is not a breaking change but a new feature it will be added as part of the release process. Also, we have an abundant number of available PR numbers, so it's ok to waste them 😛

cmd/collector/app/server/http_test.go
davidcmitchell: I would push for mTLS since it's frequently a security requirement for interservice communication. Otherwise, should we create an issue to implement plain auth alongside TLS (so that it's not ignored by the jaeger backend)?

yurishkuro: Feel free to open a discussion on that. The challenge with supporting plain auth is how you manage the allowed credentials - do we need to invent some special config file listing all user names & passwords? Do we need to integrate with some credentials store like LDAP?

Issue #2771, opened 2021-02-02
Upgrade google.golang.org/grpc to 1.35.0


yurishkuro:
The dependabot PR #2765 ran into compatibility issue, needs manual upgrade

yurishkuro:
We may also want to include github.com/golang/protobuf -> 1.4.3 which also failed in #2774 due to deprecation. It will require regenerating proto types, which probably requires upgrading https://github.com/jaegertracing/docker-protobuf

albertteoh:
Had a look at this. The following test fails upon upgrading google.golang.org/grpc to 1.35.0:
```
--- FAIL: TestTraceSpanIDMarshalProto (0.00s)
    --- FAIL: TestTraceSpanIDMarshalProto/protobuf (0.00s)
panic: invalid Go type model.TraceID for field jaeger.api_v2.SpanRef.trace_id [recovered]
	panic: invalid Go type model.TraceID for field jaeger.api_v2.SpanRef.trace_id
```

I _think_ the panic is due to lack of support for marshalling structs like [`model.TraceID`](https://github.com/jaegertracing/jaeger/blob/master/model/ids.go#L35) to bytes. That is, the type switch determines our [target type is `bytes`](https://github.com/protocolbuffers/protobuf-go/blob/master/internal/impl/convert.go#L132), but the source type of `struct` is not supported.

Tried adding `|| t.Kind() == reflect.Struct || t.Kind() == reflect.Uint64` ([TraceID](https://github.com/jaegertracing/jaeger/blob/master/model/ids.go#L35) and [SpanID](https://github.com/jaegertracing/jaeger/blob/master/model/ids.go#L41) respectively) to get past this point.
I finally reached a point where we [require a coder](https://github.com/protocolbuffers/protobuf-go/blob/master/internal/impl/codec_tables.go#L423) that knows how to un/marshal the source type (`struct` or `uint64`), but only `string` and `[]byte` are supported. I can see a `uint64` codec, so that should be solvable, but for `struct`, it's almost as if we need to recursively call `Marshal` until we reach primitive types.

Anyone have any ideas on how to solve this?

jpkrohling:
> Anyone have any ideas on how to solve this?

@pavolloffay and I had similar problems in the past when trying to upgrade to 1.30, and I reckon Pavol did have a solution, but I can't remember exactly what it was. A change here would also imply a change to the protobuf image that we use to generate the IDL code. See:

[...]

PR #2807, opened 2021-02-10
Upgraded OTel to 0.20.0

joe-elliott:
- Upgrades OTel to the latest release

- Copied `generateAndRegisterManualResolver` from the "google.golang.org/grpc/resolver/manual" package.  This appears to be a test function we are using to init the resolver?  Unsure why this was used.
- Removed all reference to the `queued_retry` processor as it no longer exists.
- The `Status` object can no longer be nil.  Adjusted code accordingly.


cmd/agent/app/reporter/grpc/builder.go
```diff
@@ -80,7 +82,7 @@ func (b *ConnBuilder) CreateConnection(logger *zap.Logger, mFactory metrics.Fact
 			return nil, errors.New("at least one collector hostPort address is required when resolver is not available")
 		}
 		if len(b.CollectorHostPorts) > 1 {
-			r, _ := manual.GenerateAndRegisterManualResolver()
+			r := generateAndRegisterManualResolver()
```
yurishkuro, line 85: is it preemptive change needed to upgrade grpc later?

yurishkuro (approved):

cmd/agent/app/reporter/grpc/builder.go
joe-elliott: It's due to the otel grpc dependency upgrading here:
https://github.com/jaegertracing/jaeger/pull/2807/files#diff-c941bfc88e74f7f559df1ab09fdfeb7c228f0dfce84ef863555812222a943e25R1469

and since the otel components use this functionality it needs to be up to date with 1.35.0.

yurishkuro: sorry, that link doesn't point me to anywhere. I just meant that there's no grpc upgrade in go.mod in this PR.

joe-elliott: the grpc upgrade is actually due to the otel upgrade.  was trying to link to the go.sum changes under `./cmd/opentelemetry`, but it does not appear to work.

0.20.0 otel collector requires grpc 1.35.0.  since the components under `./cmd/opentelemetry` reference Jaeger core, it also has to be up to date with 1.35.0 even though it is not explicitly using it in its own go.mod.

yurishkuro: I see. Have you tried bumping grpc to the same ver in the main go.mod?  I just tried it and it seems to work, and I think it's better to keep the two in sync

```
$ go get google.golang.org/grpc@v1.35.0
go: downloading google.golang.org/grpc v1.35.0
go: downloading google.golang.org/genproto v0.0.0-20200526211855-cb27e3aa2013

$ make build-collector
(no errors)
```

jpkrohling: > Have you tried bumping grpc to the same ver in the main go.mod? 

Isn't this what dependabot did in another PR?

joe-elliott: google.golang.org/grpc is currently at 1.29.1.

attempting an upgrade now to 1.35.0, but getting some failed tests.  i will see if i can't iron these out today.

jpkrohling: Might be worth checking this: https://github.com/jaegertracing/jaeger/issues/2771

yurishkuro:
@joe-elliott should we close this?

joe-elliott:
Yup, I think this no longer makes sense.

PR #2820, opened 2021-02-20
2814 - Throw an error when --es.use-ilm is set without --es.use-aliases

### LINKED ISSUE #2814: Error when `--es.use-ilm` is set without `--es.use-aliases`

`--es.use-ilm` must always be used in conjunction with `--es.use-aliases` to ensure ES writers and readers refer to the single index mapping. Otherwise, ILM will not work as it does not have a rollover alias to work on.

As such, we should provide a useful error to users who mistakenly set `--es.use-ilm` without enabling `--es.use-aliases`.

This is a follow-up task to [this PR review comment](https://github.com/jaegertracing/jaeger/pull/2796/files#r571716159).

The proposal is to add validation at the top of each ES [reader](https://github.com/jaegertracing/jaeger/blob/master/plugin/storage/es/factory.go#L141) and [writer](https://github.com/jaegertracing/jaeger/blob/master/plugin/storage/es/factory.go#L162) constructor with the following logic:
```
if cfg.GetUseILM() && !cfg.GetUseReadWriteAliases() {
	return nil, fmt.Errorf("the error message")
}
```

bhiravabhatla:
Signed-off-by: santosh <bsantosh@thoughtworks.com>

 Resolves #2814 

- Adds a check to make sure that --es.use-ilm is always used in conjunction with--es.use-aliases

 Add validation at the top of each ES reader and writer constructor with the following logic:

if cfg.GetUseILM() && !cfg.GetUseReadWriteAliases() {
	return nil, fmt.Errorf("the error message")
}

### DIFF
--- a/plugin/storage/es/factory.go
+++ b/plugin/storage/es/factory.go
@@ -138,6 +138,9 @@ func createSpanReader(
 	cfg config.ClientBuilder,
 	archive bool,
 ) (spanstore.Reader, error) {
+	if cfg.GetUseILM() && !cfg.GetUseReadWriteAliases() {
+		return nil, fmt.Errorf("--es.use-ilm must always be used in conjunction with --es.use-aliases to ensure ES writers and readers refer to the single index mapping")
+	}
 	return esSpanStore.NewSpanReader(esSpanStore.SpanReaderParams{
 		Client:              client,
 		Logger:              logger,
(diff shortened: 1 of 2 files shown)


yurishkuro (approved):

yurishkuro:
Thanks!

Issue #2749, opened 2021-01-29
Replace the use of "esc" with standard "embed" package


yurishkuro:
We're currently using `github.com/mjibson/esc` to embed certain files (like UI assets) into binaries.

Upcoming Go 1.16 supports this with a new standard package [embed](https://tip.golang.org/pkg/embed/) and `//go:embed` directive.

* [x] investigate what it would take to switch to `embed` #2835
* [x] embed HotROD assets #2835
* [x] embed Elasticsearch mappings #2822
* [x] embed UI assets (NB: we should investigate the impact on binary size compared to the current `esc` embedding. Maybe consider compressing the assets)
  * [x] placeholder index.html #2846
  * [x] actual UI #3399

See also: #2748

yurishkuro:
A blog post that might be useful (I did not read it yet): https://lakefs.io/working-with-embed-in-go/

lonewolf3739:
@yurishkuro I would like to work on this.

yurishkuro:
+1

PR #2822, opened 2021-02-22
2813 - Move all mapping related code to mappings package.

### LINKED ISSUE #2813: Refactor Elasticsearch mapping code to implement cleaner design
From @yurishkuro  comment - https://github.com/jaegertracing/jaeger/pull/2796#discussion_r571714232

* Create a struct SpanServiceMappingsBuilder and move all functions like GetSpanServiceMappings, FixMapping, LoadMapping to take that struct as receiver instead of passing a dozen parameters to them. 
* This will also allow to omit assigning the es.TemplateBuilder field (it should default to es.TextTemplateBuilder{} automatically), keeping the code cleaner yet still providing testability.
* Use that struct directly to pass values to tmpl.Execute(), instead of declaring an inner struct.
* Move all the mapping-related business into mappings package, instead of being in the factory

Needs a cleaner design. Follow up to https://github.com/jaegertracing/jaeger/pull/2796 PR

bhiravabhatla:
Signed-off-by: santosh <bsantosh@thoughtworks.com>

Resolves #2813 

-  Move all mapping related code to mappings package.
- Implement Feedback on refactoring mappings code

- Create a struct SpanServiceMappingsBuilder and move all functions like GetSpanServiceMappings, FixMapping, LoadMapping to take that struct as receiver instead of passing a dozen parameters to them.
- Move all the mapping-related business into mappings package, instead of being in the factory
- Rename ESPrefix to IndexPrefix to be more readable


bhiravabhatla:
@albertteoh /@yurishkuro. Could not implement below points - 

- This will also allow to omit assigning the es.TemplateBuilder field (it should default to es.TextTemplateBuilder{} automatically), keeping the code cleaner yet still providing testability.
- Use that struct directly to pass values to tmpl.Execute(), instead of declaring an inner struct.

As I needed to pass the template builder interface for mocking in tests.

bhiravabhatla:
Would look at failing integration tests.

plugin/storage/es/mappings/mapping.go
```diff
@@ -0,0 +1,96 @@
...
+	templateBuilder es.TemplateBuilder
+	shards          int64
+	replicas        int64
+	esVersion       uint
+	esPrefix        string
+	useILM          bool
+}
+
+// NewBuilder constructs and returns an initialized MappingBuilder
+func NewBuilder(tb es.TemplateBuilder, shards, replicas int64, esVersion uint, esPrefix string, useILM bool) *MappingBuilder {
```
yurishkuro, line 34: I suggest simply declaring all the fields public so that the caller can initialize them directly. This method does not do anything useful, but has a hard to decipher signature at the call sites

plugin/storage/es/mappings/mapping.go
```diff
@@ -0,0 +1,96 @@
...
+	tmpl, err := mb.templateBuilder.Parse(loadMapping(mapping))
+	if err != nil {
+		return "", err
+	}
+	writer := new(bytes.Buffer)
+
+	if mb.esPrefix != "" {
+		mb.esPrefix += "-"
+	}
+	values := struct {
```
yurishkuro, line 84: why not use `MappingBuilder` itself as the container of values?

plugin/storage/es/mappings/mapping.go
bhiravabhatla: Agreed. I thought of initializing mapping as well in the constructor - dint pan out. I ll just make fields public as suggested.

plugin/storage/es/mappings/mapping.go
bhiravabhatla: True. Will do.

bhiravabhatla:
Wierd, integration tests passed locally on elasticsearch v7.3.0.

plugin/storage/es/mappings/mapping.go
bhiravabhatla: Done

plugin/storage/es/mappings/mapping.go
bhiravabhatla: Done

bhiravabhatla:
> Wierd, integration tests passed locally on elasticsearch v7.3.0.

Ok, I am able reproduce locally as well. Not entirely sure whats causing this though. Will spend some time on this

bhiravabhatla:
I am seeing this in my elasticsearch logs during the test runs.

`
"stacktrace": ["org.elasticsearch.transport.RemoteTransportException: [ca859ca2c689][127.0.0.1:9300][indices:data/read/search[phase/query]]",
"Caused by: java.lang.IllegalArgumentException: Fielddata is disabled on text fields by default. Set fielddata=true on [serviceName] in order to load fielddata in memory by uninverting the inverted index. Note that this can however use significant memory. Alternatively use a keyword field instead."
`
@albertteoh / @yurishkuro  - Any idea on this.

albertteoh:
> Fielddata is disabled on text fields by default. Set fielddata=true on [serviceName] ...

I saw this once when I had index mapping issues; I think it means your `serviceName` field was set to type `text` which is ES's dynamic mapping sets for strings if there is no existing template. If it's type `text`, it means individual words in the string cannot be indexed, I think.

I would suggest querying ES on both your span and service mappings to confirm this is indeed the case (type is `text` instead of `keyword`):
```
GET jaeger-span-write/_mapping
GET jaeger-service-write/_mapping
```

And check your templates to ensure `serviceName` is present and correct (should set type to `keyword`) in both the mappings:
```
GET _template/jaeger-span
GET _template/jaeger-service
```

And from there, it's just a matter of pinning down the root cause from what's changed in this PR.

bhiravabhatla:
> > Fielddata is disabled on text fields by default. Set fielddata=true on [serviceName] ...
> 
> I saw this once when I had index mapping issues; I think it means your `serviceName` field was set to type `text` which is what ES sets for strings if there is no existing template (dynamic mapping). If it's type `text`, it means individual words in the string cannot be indexed, I think.
> 
> I would suggest querying ES on both your span and service mappings to confirm this is indeed the case (type is `text` instead of `keyword`):
> 
> ```
> GET jaeger-span-write/_mapping
> GET jaeger-service-write/_mapping
> ```
> 
> And check your templates to ensure `serviceName` is present and correct (should set type to `keyword`) in both the mappings:
> 
> ```
> GET _template/jaeger-span
> GET _template/jaeger-service
> ```
> 
> And from there, it's just a matter of pinning down the root cause from what's changed in this PR.

Thank you @albertteoh  - that helped. Index template pattern was the issue, index prefix was getting suffixed with "-" twice. Added a suffix check to avoid it. Tests passed locally.

plugin/storage/es/mappings/gen_assets.go
```diff
@@ -213,7 +213,7 @@ var _escData = map[string]*_escFile{
 		name:    ".nocover",
```
yurishkuro, line 213: consider using the new `embed` package (see #2749)

plugin/storage/es/mappings/jaeger-dependencies.json
```diff
@@ -1,8 +1,8 @@
 {
   "template": "*jaeger-dependencies-*",
   "settings":{
-    "index.number_of_shards": {{ .NumberOfShards }},
```
yurishkuro, line 4: Why do we need to change the mappings in this PR?  Isn't it simpler to change the field names in the class?

yurishkuro: having said that, ESIndex->IndexPrefix is a welcome change

plugin/storage/es/mappings/mapping_test.go
```diff
@@ -0,0 +1,316 @@
...
+		shards      int64
+		replicas    int64
+		esVersion   uint
+		indexPrefix string
+		useILM      bool
+	}
+	tests := []struct {
+		name                       string
+		args                       args
+		mockNewTextTemplateBuilder func() es.TemplateBuilder
```
yurishkuro, line 165: I would just use `MappingBuilder` struct directly, there's no benefit in the indirections

plugin/storage/es/mappings/mapping_test.go
```diff
@@ -0,0 +1,316 @@
...
+	ta.On("Execute", mock.Anything, mock.Anything).Return(errors.New("template load error"))
+	tb.On("Parse", mock.Anything).Return(&ta, nil)
+
+	mappingBuilder := MappingBuilder{
+		TemplateBuilder: &tb,
+		Shards:          5,
+		Replicas:        5,
+		EsVersion:       7,
+		IndexPrefix:     "",
+		UseILM:          false,
```
albertteoh, line 312: these could be removed since the test isn't asserting on the resulting mapping string

plugin/storage/es/mappings/mapping_test.go
```diff
@@ -0,0 +1,316 @@
...
+			assert.Equal(t, got, want)
+		})
+	}
+}
+
+func TestMappingBuilder_loadMapping(t *testing.T) {
+	tests := []struct {
+		name        string
+		indexPrefix string
+		useILM      bool
```
albertteoh, line 73: these are never set or used so they can be removed

plugin/storage/es/mappings/mapping.go
```diff
@@ -0,0 +1,81 @@
...
+
+func (mb *MappingBuilder) fixMapping(mapping string) (string, error) {
+
+	tmpl, err := mb.TemplateBuilder.Parse(loadMapping(mapping))
+	if err != nil {
+		return "", err
+	}
+	writer := new(bytes.Buffer)
+
+	if mb.IndexPrefix != "" && !strings.HasSuffix(mb.IndexPrefix, "-") {
```
albertteoh, line 73: is there a test to cover the case where the `IndexPrefix` has an existing `-` suffix?

plugin/storage/es/mappings/mapping.go
```diff
@@ -0,0 +1,81 @@
...
+type MappingBuilder struct {
+	TemplateBuilder es.TemplateBuilder
+	Shards          int64
+	Replicas        int64
+	EsVersion       uint
+	IndexPrefix     string
+	UseILM          bool
+}
+
+// GetMapping returns the render mapping based on elasticsearch version
```
albertteoh, line 34: `s/render/rendered/`

plugin/storage/es/mappings/mapping_test.go
```diff
@@ -0,0 +1,316 @@
...
+				Replicas:        3,
+				EsVersion:       tt.esVersion,
+				IndexPrefix:     "",
+				UseILM:          false,
+			}
+			got, err := mb.GetMapping(tt.mapping)
+			require.NoError(t, err)
+			want := ""
+			if tt.esVersion == 7 {
+				want, err = mb.fixMapping("/" + tt.mapping + "-7.json")
```
albertteoh, line 58: Although these tests achieve the goal of line coverage, they appear execute the same logic between the `want` and `got`. This means that if there's a bug in, say, `fixMapping()`, we wouldn't catch it.

I think we ought to have a static file in a fixtures directory of `mappings` that represents the `want`ed resulting mapping, then assert against the contents of this file.

plugin/storage/es/mappings/jaeger-dependencies.json
bhiravabhatla: Yep,  I thought of doing the same but, as I was changing the files anyway for indexPrefix, so renamed other variables as well. Do you want me to revert it?

yurishkuro: Nah, let's keep it. If we have integration tests that exercise all these mappings, then it's a safe change

plugin/storage/es/mappings/mapping_test.go
bhiravabhatla: Yep, agreed. Would add fixtures as suggested.

plugin/storage/es/mappings/mapping.go
bhiravabhatla: Nope. Will add one

plugin/storage/es/mappings/mapping_test.go
```diff
@@ -0,0 +1,316 @@
...
+				require.NoError(t, err)
```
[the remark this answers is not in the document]
bhiravabhatla: @albertteoh  dint get this - you want me to move `require.NoError(t, err)` after else?

plugin/storage/es/mappings/mapping.go
bhiravabhatla: Fixed in next commit

plugin/storage/es/mappings/mapping_test.go
bhiravabhatla: Added in next commit

plugin/storage/es/mappings/mapping.go
bhiravabhatla: Added as part of GetMapping unit test

plugin/storage/es/mappings/mapping_test.go
bhiravabhatla: Removed in next commit

plugin/storage/es/mappings/mapping_test.go
bhiravabhatla: Yep, Removed in next commit

plugin/storage/es/mappings/gen_assets.go
bhiravabhatla: Used embed instead of esc. Would remove corresponding make target dependency as well.

plugin/storage/es/mappings/mapping_test.go
bhiravabhatla: Dint get this.

yurishkuro: rather than defining a struct for tests and copying values into it, just use MappingBuilder struct directly when defining test cases.

bhiravabhatla: Not part of this commit. Will work on it now.

plugin/storage/es/mappings/mapping.go
```diff
@@ -0,0 +1,86 @@
...
+	"bytes"
+	"embed"
+	"strings"
+
+	"github.com/jaegertracing/jaeger/pkg/es"
+)
+
+//go:embed *.json
+// MAPPINGS contains embeded index templates.
+var MAPPINGS embed.FS
```
yurishkuro, line 27: does this need to be public?
```suggestion
var mappingsFS embed.FS
```

Issue #2781, opened 2021-02-03
Remove temporary usage of a Thrift fork


jpkrohling:
The PR #2780 introduced the usage of a personal fork for Apache Thrift Go libraries to address a memory problem with the agent. Once a new version of Thrift is released that includes the fix, we need to revert that and use the official library again.

jpkrohling:
Looks like Thrift 0.14 is now released: https://github.com/apache/thrift/releases/tag/v0.14.0

Any takers?

albertteoh:
I can take a look at this.

albertteoh:
@jpkrohling is there a way to confirm that this new thrift tag fixes issues #2638 and #2452? I also tried the reproducer in https://github.com/jaegertracing/jaeger/issues/2638#issuecomment-742592854 but I see the following output:
```
$ go run main.go
error while processing: *jaeger.Batch error reading struct: *jaeger.Span error reading struct: EOF
memory usage before: 0 MiB
memory usage after: 0 MiB
finished
```
I was expecting some non-zero memory usage numbers and no error; but maybe I'm wrong?

@Mario-Hofstaetter, @zigmund: Were you able to confirm that changes in #2780 fixed the memory leak issue?

zigmund:
@albertteoh is there docker image containing https://github.com/jaegertracing/jaeger/pull/2780?
Or I should use `jaegertracing/jaeger-agent:latest`? Docker hub says it was pushed last time 2 days ago.

jpkrohling:
`jaegertracing/jaeger-agent:latest` should indeed contain the fix.

jpkrohling:
> @jpkrohling is there a way to confirm that this new thrift tag fixes issues #2638 and #2452? I also tried the reproducer in [#2638 (comment)](https://github.com/jaegertracing/jaeger/issues/2638#issuecomment-742592854) but I see the following output:
> 
> ```
> $ go run main.go
> error while processing: *jaeger.Batch error reading struct: *jaeger.Span error reading struct: EOF
> memory usage before: 0 MiB
> memory usage after: 0 MiB
> finished
> ```
> 
> I was expecting some non-zero memory usage numbers and no error; but maybe I'm wrong?
> 

[...]

PR #2828, opened 2021-02-23
Remove existing OpenTelemetry Collector code

jpkrohling:
The current OpenTelemetry code isn't representative anymore of where we want to go for Jaeger v2. In order to reduce confusion to users, we discussed during the last weekly meeting about removing this code. Exporters that were created for this are still available in earlier tags and can be recovered from there when needed.


pavolloffay:
Are these components going to be migrated to  https://github.com/jaegertracing/jaeger-otelcol? 

The main reason for having the code here was for the ease of development. Many times changes in the core Jaeger components were required.

joe-elliott:
There were a number of places that these "AddOtelFlags" methods were added:

https://github.com/jaegertracing/jaeger/blob/master/cmd/ingester/app/flags.go#L92

Should we clean those up as well?

jpkrohling:
From what I understand, we'll first adapt this repository here to have multiple go modules, and then move the contents of jaeger-otelcol to this repo here. The goal is to avoid the circular dependency between otelcol and Jaeger, while still being able to have the binaries being built as part of this single repository.

@joe-elliott is working on the reorganization, and @kevinearls will work on moving things here. Some of the components removed by this PR will get a new life after this reorganization.

jpkrohling:
For context, the decision to remove this here right now also has to do with #2781, as we can't bump Thrift to 0.14 in the otelcol modules as it depends on a Zipkin Thrift repository that is still on 0.13. Unfortunately, Thrift 0.14 is incompatible with 0.13, as it adds `context.Context` to some methods...

jpkrohling:
> Should we clean those up as well?

Done!

pavolloffay:
> For context, the decision to remove this here right now also has to do with #2781, as we can't bump Thrift to 0.14 in the otelcol modules as it depends on a Zipkin Thrift repository that is still on 0.13. Unfortunately, Thrift 0.14 is incompatible with 0.13, as it adds context.Context to some methods...

The same problem might pop up after in OTEL collector as well. Better would be to bump zipkin to thrift 0.14 as well.

jpkrohling:
> The same problem might pop up after in OTEL collector as well.

Indeed, it will affect that as well.

joe-elliott (approved):
Looks good to me

joe-elliott (changes requested):
Apologies, found a few more non-blocking otel references after approving:

https://github.com/jaegertracing/jaeger/blob/master/.gitignore#L37
https://github.com/jaegertracing/jaeger/blob/master/Makefile#L3
https://github.com/jaegertracing/jaeger/blob/master/Makefile#L177
https://github.com/jaegertracing/jaeger/blob/master/scripts/upload-all-docker-images.sh#L33
https://github.com/jaegertracing/jaeger/blob/master/scripts/upload-to-registry.sh#L84

scripts/build-all-in-one-image.sh
```diff
@@ -67,9 +67,3 @@ docker build -f cmd/all-in-one/Dockerfile \
         --build-arg TARGETARCH=$GOARCH
 run_integration_test $repo
 upload_to_docker $repo
-
-make build-otel-all-in-one GOOS=linux GOARCH=$GOARCH
-repo=jaegertracing/opentelemetry-all-in-one
-docker build -f cmd/opentelemetry/cmd/all-in-one/Dockerfile -t $repo:latest cmd/opentelemetry/cmd/all-in-one --build-arg TARGETARCH=$GOARCH
-run_integration_test $repo
-upload_to_docker $repo
```
yurishkuro, line 75: there's more stuff in scripts:
```
$ grx opentelemetry scripts
scripts/upload-all-docker-images.sh:33:	opentelemetry-collector
scripts/upload-all-docker-images.sh:34:	opentelemetry-agent
scripts/upload-all-docker-images.sh:35:	opentelemetry-ingester
scripts/build-all-in-one-image.sh:72:repo=jaegertracing/opentelemetry-all-in-one
scripts/build-all-in-one-image.sh:73:docker build -f cmd/opentelemetry/cmd/all-in-one/Dockerfile -t $repo:latest cmd/opentelemetry/cmd/all-in-one --build-arg TARGETARCH=$GOARCH
scripts/upload-to-registry.sh:84:      "jaegertracing/jaeger-opentelemetry-collector" | "jaegertracing/jaeger-opentelemetry-agent" | "jaegertracing/jaeger-opentelemetry-ingester" | "jaegertracing/opentelemetry-all-in-one" )
```

jpkrohling: Done

cmd/agent/app/reporter/grpc/flags.go
```diff
@@ -41,11 +41,6 @@ var tlsFlagsConfig = tlscfg.ClientFlagsConfig{
 func AddFlags(flags *flag.FlagSet) {
 	flags.Uint(retry, defaultMaxRetry, "Sets the maximum number of retries for a call")
 	flags.Int(discoveryMinPeers, 3, "Max number of collectors to which the agent will try to connect at any given time")
-	AddOTELFlags(flags)
-}
-
-// AddOTELFlags adds flags that are exposed by OTEL collector
-func AddOTELFlags(flags *flag.FlagSet) {
 	flags.String(collectorHostPort, "", "Comma-separated string representing host:port of a static list of collectors to connect to directly")
```
albertteoh, line 44: I wonder why these were considered specific to OTEL in the first place

albertteoh (approved):
lgtm

cmd/agent/app/reporter/grpc/flags.go
pavolloffay: Not all flags were added to OTEL distros in other words some flags made sense only for legacy Jaeger.

joe-elliott (approved):

jpkrohling:
I'm merging with two approvals, any left overs can be removed in a follow-up PR.

PR #2835, opened 2021-02-25
Upgrade to Go 1.16; use embed package in hotrod

yurishkuro:
* Upgrade go.mod and GitHub Actions to Go 1.16
* Fixes part of #2749 for HotRod assets


examples/hotrod/services/frontend/web_assets/assets.go
```diff
@@ -0,0 +1,7 @@
+package webassets
```
yurishkuro, line 1: didn't find an easy way to remove `web_assets/` dir name from the files when embedding

albertteoh (approved):
Nice! Unfortunate about the `web_assets` dir name. 

Also, not sure if you wanted to include the `cmd/opentelemetry/go.sum` diff in this PR.

yurishkuro:
> Unfortunate about the web_assets dir name.

What do you mean?

> Also, not sure if you wanted to include the cmd/opentelemetry/go.sum diff in this PR.

Yes, because it refused to build otherwise.

albertteoh:
> What do you mean?

I meant not being to find an easy way to remove the `web_assets` dir name due to the underscore in the name.

yurishkuro:
no, that's not what I meant. It doesn't bother me that the dir name has underscore, I kept it to minimize changes. If we want we can rename the dir to match the package name.

What I meant was that I had to create a .go file in that dir in order to embed the files. My initial attempt was to put the embed directive in the service code, like this:
```
//go:embed web_assets/*
```
However, that makes them available at `/web_assets/...` URL, instead of root `/`. There is a std function `http.StripPrefix`, we kinda need the reverse one, to preserve root URL and prepend `/web_assets` when accessing files from the embedded file system. There's no std function like that, and the current approach seemed easier.

We will likely run into similar issue with the main UI assets, because they are in `cmd/query/app/ui/actual/public` but the embedding is in `cmd/query/app/ui/actual`. So we may need to tweak the build to copy them differently.

I was surprised that there doesn't seem to be a way to ask embed.FS to represent subdirectory as another FS, that would've solved the issue. But I have not looked too deeply.

PR #2846, opened 2021-02-25
Support embedding assets from a subdir

yurishkuro:
- add util function for altering `http.FileSystem` behavior
- refactor HotROD embedding to read from a subdir directly
- replace `esc` with embedding for the main UI placeholder


yurishkuro:
@albertteoh fix for that nested dir

pkg/httpfs/prefixed.go
```diff
@@ -0,0 +1,45 @@
...
+
+package httpfs
+
+import (
+	"net/http"
+)
+
+// AddPrefixFS returns a FileSystem that adds a path prefix to all files
+// before delegating to the underlying fs.
+func AddPrefixFS(prefix string, fs http.FileSystem) http.FileSystem {
```
albertteoh, line 23: nit: `AddPrefixFS` implies the func adds a prefix to the given filesystem, when in fact, it's creating a new struct. As such, I suggest renaming to `NewPrefixedFS`.

pkg/httpfs/prefixed.go
```diff
@@ -0,0 +1,45 @@
...
+	return &addPrefixFS{
+		prefix: prefix,
+		fs:     fs,
+	}
+}
+
+type addPrefixFS struct {
```
albertteoh, line 30: nit: related to above, `addPrefixFS` has a verb prefix which is better suited to a func name; how about `prefixedFS`?

pkg/httpfs/prefixed_test.go
```diff
@@ -0,0 +1,46 @@
...
+	"net/http"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+//go:embed test_assets/*
+var assetFS embed.FS
+
+func TestAddPrefixFS_Root(t *testing.T) {
```
albertteoh, line 28: could these be implemented as table-driven tests to reduce duplication?

pkg/httpfs/prefixed.go
```diff
@@ -0,0 +1,45 @@
...
+func (fs *addPrefixFS) Open(name string) (http.File, error) {
+	prefixedName := fs.prefix + name
+	if name == "/" {
+		// Return the dir itself when asked for the root.
+		// This is what http.FS() also does to allow redirects
+		// from `/`` to `/index.html`.
+		prefixedName = fs.prefix
+	}
+	f, err := fs.fs.Open(prefixedName)
+	return f, err
```
albertteoh, line 44: `return fs.fs.Open(prefixedName)`

pkg/httpfs/prefixed.go
yurishkuro: Perhaps, on the other hand, std lib often uses naming pattern without New. And the business function is exactly what you said - adding prefix to a file system, returning a struct is besides the point. I think New is typically used when a public struct is returned, not in case of decorators.

Having said that, the FS prefix is rather unnecessary: httpfx.AddPrefix()

albertteoh (approved):
nice! Looking forward to removing use of `esc` altogether; I noticed it's been a common question/problem from folks trying to run/build the backend components because they haven't installed it or `$GOPATH` isn't in their `$PATH`.

yurishkuro:
Indeed. And the gen_assets kept being regenerated with only difference being the timestamp, pretty annoying.

We need to merge #2822 before the last esc piece can be removed.

PR #2850, opened 2021-02-26
WIP - Replace esc with "embed" for packaging UI assets

### LINKED ISSUE #3229: Replace esc with "embed" for packaging UI assets
To complete the issue #2749, there's one PR that needs some attention: #2850. At its current state, it adds the contents of the Jaeger UI build to a Go asset in an uncompressed format. This causes an increase of approximately 10MB to the binary size of the Query/all-in-one.

Ideally, we'd find a way to store the contents of the UI in a compressed format in the binary and serve the compressed asset directly to the client, without having to decompress/recompress.

yurishkuro:
Part of #2749, closes #3229

  * [x] infra changes to use `go:embed` in the build to pack UI assets
  * [x] use `embed.FS` as a source for HTTP server for static assets
  * [ ] an adapter that implements `http.FileSystem ` that can read gzip-compressed files from `embed.FS` (similar to https://github.com/vearutop/statigz, but we want to avoid 3rd party dependency)


yurishkuro:
As I was afraid, the binary size goes up a lot without doing some explicit compression:
```
# with esc
-rwxr-xr-x      43M Feb 25 23:15 ./cmd/all-in-one/all-in-one-darwin-amd64*
# with embed
-rwxr-xr-x      56M Feb 25 23:21 ./cmd/all-in-one/all-in-one-darwin-amd64*
```

yurishkuro:
If all files are gzipped, then the size is comparable with esc (even a tiny bit smaller)
```
-rwxr-xr-x      42M Feb 25 23:40 ./cmd/all-in-one/all-in-one-darwin-amd64*
```

yurishkuro:
Found a brand new library by @vearutop that serves gzipped files https://github.com/vearutop/statigz. Not sure if we want to use it. It might conflict with our usage of `gorilla.CompressHandler`, although we could change it so that it doesn't apply to static assets.

yurishkuro:
@albertteoh fyi

vearutop:
Combining `statigz` with `gorilla.CompressHandler` would work, but in a sub optimal way.

The problem is that `gorilla.CompressHandler` removes `Accept-Encoding` header before passing request to next handler.
https://github.com/gorilla/handlers/blob/v1.5.1/compress.go#L119
```go
r.Header.Del(acceptEncoding)
```
In such case `statigz` would think user agent is not capable to receive compressed response and will decompress it on the fly, then this decompressed response will be compressed again by `gorilla.CompressHandler`.

albertteoh (approved):

jpkrohling (dismissed):

jpkrohling:
What's the status of this PR?

yurishkuro:
It needs to be done, but I didn't have time. Currently it blows up the size of the binaries unless we introduce a new lib dependency.

jpkrohling:
> It needs to be done, but I didn't have time. Currently it blows up the size of the binaries unless we introduce a new lib dependency.

Do you think an Outreachy candidate would be able to tackle this if a good description is provided in an issue?

yurishkuro:
yes, it's a fairly straightforward piece.

yurishkuro:
incorporated into #3399

PR #2857, opened 2021-03-01
WIP - Upgrade GRPC/Proto to Latest

### LINKED ISSUE #2771: Upgrade google.golang.org/grpc to 1.35.0

joe-elliott:
- Upgrades GPRC/Proto to the latest
- Fixes #2771 

- Changes to use the latest proto builder from OTel 
  - we can build a jaeger image with latest and use it if desired
- Copy in removed method `generateAndRegisterManualResolver`
  - this method looks wild to me.  Using a time value for the scheme?  Possibly something to clean up in a future PR
- Regenerated proto with the latest builder
- Changed a few packages to use gogoproto over google proto
- Added `/proto-gen/api_v2/codec.go`.  This forces grpc to use gogoproto over google proto.
  - I hate the location of this, but unsure if there's a better place that will be always be imported in a package that uses grpc.


Makefile
```diff
@@ -60,7 +60,7 @@ SWAGGER_IMAGE=quay.io/goswagger/swagger:$(SWAGGER_VER)
 SWAGGER=docker run --rm -it -u ${shell id -u} -v "${PWD}:/go/src/" -w /go/src/ $(SWAGGER_IMAGE)
 SWAGGER_GEN_DIR=swagger-gen
 
-JAEGER_DOCKER_PROTOBUF=jaegertracing/protobuf:0.2.0
+JAEGER_DOCKER_PROTOBUF=otel/build-protobuf:0.2.1
```
yurishkuro, line 63: Could we make the corresponding upgrade in our `protobuf` image?

Alternatively, are there significant differences between them? Maybe it would make sense to deprecate ours altogether.

model/codec.go
```diff
@@ -0,0 +1,48 @@
...
+
+// Marshal implements encoding.Codec
+func (c *gogoCodec) Marshal(v interface{}) ([]byte, error) {
+	return proto.Marshal(v.(proto.Message))
+}
+
+// Unmarshal implements encoding.Codec
+func (c *gogoCodec) Unmarshal(data []byte, v interface{}) error {
+	return proto.Unmarshal(data, v.(proto.Message))
+}
```
yurishkuro, line 48: I assume this is already explicitly covered by the existing tests for marshalling.

yurishkuro (commented):
Is `model/model/model.pb.go` file needed?

Makefile
jpkrohling: Why did we stop using the upstream image? The last time I've checked, the image was well-maintained.

https://github.com/TheThingsIndustries/docker-protobuf

yurishkuro: We never used upstream image. It doesn't include gogo iirc.

joe-elliott:
> Is model/model/model.pb.go file needed?
Nope.  Has been removed.

model/codec.go
joe-elliott: Yes, all grpc/proto marshalling/unmarshalling will now go through this file.

Makefile
joe-elliott: Trying to use the upstream image (`thethingsindustries/protoc:3.1`) fails with 

> protoc-gen-swagger/options/annotations.proto: File not found.
> query.proto:25:1: Import "protoc-gen-swagger/options/annotations.proto" was not found or had errors.

We can upgrade our image if we want.  I'm fine with either option, but I figured it was worth discussing.  Here is the diff to upgrade `jaegertracing/protobuf`:

https://gist.github.com/joe-elliott/249686fa8329d7700b1044bff5a34aee

This was largely taken from the otel protobuf image.

proto-gen/api_v2/codec.go
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) 2021 The Jaeger Authors.
```
yurishkuro, line 1: maybe putting this into `/model` is better, because no other APIs can be used without model, but not vice versa.

Makefile
yurishkuro: Let's upgrade ours, can you create a PR?

proto-gen/api_v2/codec.go
joe-elliott: I also prefer model, but this is where the cracks in this approach show.  At this point what I have pushed works, but it is based on the ordering of the `init()` in two different packages.  The init method in `/proto-gen/api_v2/codec.go` has to occur after this init:

https://github.com/grpc/grpc-go/blob/master/encoding/proto/proto.go#L33

The only package in Jaeger that eventually includes `google.golang.org/grpc/encoding/proto` is `api_v2` so putting codec.go in this package guarantees that we are called second and gogoproto is the registered coded for proto with grpc.   See [this code](https://github.com/grpc/grpc-go/blob/master/encoding/encoding.go#L113) to see the registration process.

What this means is that future, unrelated changes to Jaeger could magically break this by accidentally calling the `grpc/encoding/proto` init after our init.  The good news is that CI would catch this, the bad news is that this is bad.

I don't _hate_ this change.  It allows us to stay up to date with OTel and grpc, but there are obvious drawbacks.  I support either moving forward (with additional comments in codec.go to explain the above issue) or being patient with gogoproto for a few months to see if things resolve.

joe-elliott:
github is not updating, but currently all tests are passing aside from "CIT Memory And Badger".  It is failing with 

```    integration.go:184: plugin error: rpc error: code = Unavailable desc = connection error: desc = "transport: error while dialing: dial unix /tmp/plugin047540702: connect: connection refused"```

I will look into this when I get a chance.

File: proto-gen/api_v2/codec.go (hunk 3 above)
Comment (yurishkuro) [reply proto-gen/api_v2/codec.go:1]: I am also fine with this, but I suggest literally pasting your explanation into the comments.

Also, if we move codec.go into model/, can we not add an _ import into it to ensure that `grpc/encoding/proto` is called first?

**Comment by yurishkuro** [comment]
> currently all tests are passing aside from "CIT Memory And Badger"

It would be exercising gRPC storage plugin infra. A bit concerning that it fails, since wire format / handshakes should not have changed with gRPC upgrade.

**Comment by yurishkuro** [comment]
I wonder if this has something to do with grpc-plugin module itself and which version of gRPC it usees. There isn't good logging from the test, unfortunately, my guess is that either start-up fails of the initial handshake fails
```
    integration.go:89: Waiting for storage backend to update documents, iteration 1 out of 30
    integration.go:184: grpc stream error: rpc error: code = Unavailable desc = transport is closing
```

**Comment by joe-elliott** [comment]
Ok, I hit what may be a real blocker that requires either serious changes in the hashicorp go plugin or gogoproto.  The memstore plugin is exiting with the below error.  [Others are seeing similar issues.](https://github.com/containerd/ttrpc/issues/62)

```
panic: protobuf tag not enough fields in Empty.state:
```

Which comes from this line in the gogo proto unmarshalling code:

https://github.com/gogo/protobuf/blob/b03c65ea87cdc3521ede29f62fe3ce239267c1bc/proto/table_unmarshal.go#L341

This is coming from hashicorps use of the Empty type here:

https://github.com/hashicorp/go-plugin/blob/0c19a1387af146eaa3335dad0a021741344ac3e9/internal/plugin/grpc_stdio.pb.go#L9

The empty type in now a wrapper around a google.golang.org/protobuf type here:

https://github.com/protocolbuffers/protobuf-go/blob/master/types/known/emptypb/empty.pb.go#L52

Note that the `empty` type has 3 metadata fields: `state`, `sizeCache` and `unknownFields`.  [These fields were formerly](https://github.com/golang/protobuf/commit/b9f5089fb9d4f7e6d1126161d93549e300c70f50#diff-b9c985a1ebe662c4712fd9979a296fbf966136acd072c74f597e1ff48711a56e) ??, `XXX_sizecache` and `XXX_unrecognized`.

So I can actually get tests to pass by ignoring these fields, but I have to directly edit gogo proto vendored code to do it.  Tests pass by extending this line to include the new fields:

https://github.com/gogo/protobuf/blob/master/proto/table_unmarshal.go#L327

```
if f.Name == "XXX_NoUnkeyedLiteral" || f.Name == "XXX_sizecache" || f.Name == "state" || f.Name == "sizeCache" || f.Name == "unknownFields" {
    continue
}
```

I don't know if I can take this any farther.  The above change technically works, but doesn't feel like a valuable PR to gogoproto.  I don't have any real, deep understanding of what this change does in the broader context of proto parsing.

pavolloffay:
done in #3056 

thanks @joe-elliott and @albertteoh for the prior work

PR #2861, opened 2021-03-04
Use official Apache Thrift instead of Jaeger's fork

### LINKED ISSUE #2781: Remove temporary usage of a Thrift fork

jpkrohling:
Closes #2781


jpkrohling:
@yurishkuro do you know what this is? https://github.com/jaegertracing/jaeger/tree/master/pkg/clientcfg/clientcfghttp/thrift-0.9.2

albertteoh (dismissed):

albertteoh:
Looks like more changes are required to resolve the lint checks?

pkg/clientcfg/clientcfghttp/handler_test.go
```diff
@@ -91,7 +90,7 @@ func testHTTPHandler(t *testing.T, basePath string) {
 				err = resp.Body.Close()
 				require.NoError(t, err)
 				if endpoint == "/" {
-					objResp := &tSampling092.SamplingStrategyResponse{}
+					objResp := &sampling.SamplingStrategyResponse{}
```
yurishkuro, line 93: What is the motivation for this change? We use v092 for backwards compatibility with some old clients that used to expect numeric enums in the JSON.

jpkrohling: That explains it. The 0.9.2 client code base is failing to compile because of the Thrift incompatibility issues, and I was wondering why this was using that old client.

Not sure how to move forward here: we can't use 0.9.2, as it won't compile with the Thrift we have now as dependency (0.14.1).

yurishkuro: What exactly is the dependency? We have a fork of 0.9.2 embedded in the repo, it wasn't supposed to interact with the main version of Thrift.

jpkrohling: The context changes, like here: 

https://github.com/jaegertracing/jaeger/blob/e4cacd0a2f4b362b9a0a03efef4311c45aa5899a/pkg/clientcfg/clientcfghttp/thrift-0.9.2/ttypes.go#L76

In the latest Thrift, this function requires a context. Should I just manually change those methods then to comply with the new signature?

yurishkuro: Ah, I see, so this package is just our old generated types, not the full thrift/lib. Then I'd say we drop this support, it's been several years since the clients were upgraded. It was introduced back before agent was even oss, in #61. 

Would be good to get confirmation from @vprithvi that it's not being used. There used to be a metric counting requests to this old endpoint https://github.com/jaegertracing/jaeger/pull/61/files#diff-2f12a4e2fcded1e774b911087f15091aead73a569c7d67caaa8be009d521dca2R63, but I am not seeing it in the recent code.

yurishkuro (approved):

yurishkuro:
Let's add a "breaking change" notice to CHANGELOG

PR #2862, opened 2021-03-05
Enable logging in ES client

### LINKED ISSUE #2821: Log ES queries

Lack of visibility into Elasticsearch queries executed by readers.

See: https://github.com/jaegertracing/jaeger/issues/2816#issuecomment-782147138

Add a package-level helper function that can be used when logging service, operation and span queries:
```
func logQuery(logger *zap.Logger, query *elastic.SearchSource) {
	// Ignore errors since we're only debug logging and should not prevent queries from being executed.
	qMap, _ := query.Source()
	qJson, _ := json.Marshal(qMap)
	logger.Debug("Querying elasticsearch", zap.String("query", string(qJson)))
}
```

Logging queries requires the ability to build the query beforehand and, ideally, reuse this same query object to execute against Elasticsearch. 

Unfortunately, for service and operation queries, existing code uses [SearchService](https://github.com/jaegertracing/jaeger/blob/5062366157fff65a052dc6ac6779e7621054248d/plugin/storage/es/spanstore/service_operation.go#L113) which does not provide a means to set a query source for execution, leading to duplicating the query object, unlike span searches which use `MultiSearch()` that allows [setting the query source](https://github.com/jaegertracing/jaeger/blob/5062366157fff65a052dc6ac6779e7621054248d/plugin/storage/es/spanstore/reader.go#L350) in its individual search requests.

albertteoh:

- Fixes #2821

- Enables logging options in ES client.

## Testing

Manually tested with the following combination of flags on jaeger-collector:
- `ES_LOG_LEVEL=debug` only: logs ES queries as expected
- `ES_LOG_LEVEL=debug LOG_LEVEL=debug`: logs both ES queries and usual collector debug logs as expected
- `ES_LOG_LEVEL=debug LOG_LEVEL=error`: logs ES queries as expected

Manually tested with the following combination of flags on jaeger-query:
- `ES_LOG_LEVEL=debug` only: only logs primary ES client queries.
- `ES_LOG_LEVEL=debug ES_ARCHIVE_LOG_LEVEL=debug`: logs both primary ES client as well as archive ES client queries.

**Example logs: jaeger-query** 
With `ES_LOG_LEVEL=debug`:
```
# Request
{"level":"debug","ts":1614933783.168494,"caller":"zapgrpc/zapgrpc.go:94","msg":"POST /jaeger-service-2021-03-05%2Cjaeger-service-2021-03-04%2Cjaeger-service-2021-03-03%2Cjaeger-service-2021-03-02/_search?ignore_unavailable=true&rest_total_hits_as_int=true HTTP/1.1\r\nHost: 127.0.0.1:9200\r\nUser-Agent: elastic/6.2.35 (darwin-amd64)\r\nTransfer-Encoding: chunked\r\nAccept: application/json\r\nContent-Type: application/json\r\nAccept-Encoding: gzip\r\n\r\n5d\r\n{\"aggregations\":{\"distinct_services\":{\"terms\":{\"field\":\"serviceName\",\"size\":9999}}},\"size\":0}\r\n0\r\n\r\n\n"}

# Response
{"level":"debug","ts":1614933783.1721869,"caller":"zapgrpc/zapgrpc.go:94","msg":"HTTP/1.1 200 OK\r\nContent-Type: application/json; charset=UTF-8\r\n\r\n{\"took\":2,\"timed_out\":false,\"_shards\":{\"total\":5,\"successful\":5,\"skipped\":0,\"failed\":0},\"hits\":{\"total\":9,\"max_score\":null,\"hits\":[]},\"aggregations\":{\"distinct_services\":{\"doc_count_error_upper_bound\":0,\"sum_other_doc_count\":0,\"buckets\":[{\"key\":\"jaeger-query\",\"doc_count\":9}]}}}\n"}
```

With `ES_LOG_LEVEL=info`:
```
# Request + response
{"level":"info","ts":1614933693.67656,"caller":"zapgrpc/zapgrpc.go:94","msg":"POST http://127.0.0.1:9200/jaeger-service-2021-03-05%2Cjaeger-service-2021-03-04%2Cjaeger-service-2021-03-03%2Cjaeger-service-2021-03-02/_search?ignore_unavailable=true&rest_total_hits_as_int=true [status:200, request:0.003s]"}
```

**Example logs: jaeger-collector**
With `ES_LOG_LEVEL=debug`:
```
# Request
{"level":"debug","ts":1614933923.268276,"caller":"zapgrpc/zapgrpc.go:94","msg":"POST /_bulk HTTP/1.1\r\nHost: 127.0.0.1:9200\r\nUser-Agent: elastic/6.2.35 (darwin-amd64)\r\nContent-Length: 3985\r\nAccept: application/json\r\nContent-Type: application/x-ndjson\r\nAccept-Encoding: gzip\r\n\r\n{\"index\":{\"_index\":\"jaeger-service-2021-03-05\",\"_id\":\"5d2c6ee5873dd4a6\"}}\n{\"serviceName\":\"jaeger-query\",\"operationName\":\"/api/services\"}\n{\"index\":{\"_index\":\"jaeger-span-2021-03-05\"}}\n{\"traceID\":\"667fd072e817ec39\",\"spanID\":\"667fd072e817ec39\",\"flags\":1,\"operationName\":\"/api/services\",\"references\":[],\"startTime\":1614933922387422,\"startTimeMillis\":1614933922387,\"duration\":3363,\"tags\":[{\"key\":\"sampler.type\",\"type\":\"string\",\"value\":\"const\"},...


# Response
{"level":"debug","ts":1614933923.308722,"caller":"zapgrpc/zapgrpc.go:94","msg":"HTTP/1.1 200 OK\r\nContent-Type: application/json; charset=UTF-8\r\n\r\n{\"took\":38,\"errors\":false,\"items\":[{\"index\":{\"_index\":\"jaeger-service-2021-03-05\",\"_type\":\"_doc\",\"_id\":\"5d2c6ee5873dd4a6\",\"_version\":2,\"result\":\"updated\",\"_shards\":{\"total\":2,\"successful\":2,\"failed\":0}...
```

With `ES_LOG_LEVEL=info`:
```
# Request + response
{"level":"info","ts":1614934150.339668,"caller":"zapgrpc/zapgrpc.go:94","msg":"POST http://127.0.0.1:9200/_bulk [status:200, request:0.011s]"}
```


albertteoh:
cc @dekimsey

pkg/es/config/config.go
```diff
@@ -362,6 +363,18 @@ func (c *Configuration) getConfigOptions(logger *zap.Logger) ([]elastic.ClientOp
...
 	options = append(options, elastic.SetHttpClient(httpClient))
 	options = append(options, elastic.SetBasicAuth(c.Username, c.Password))
+
+	// Elastic client requires a "Printf"-able logger.
+	l := zapgrpc.NewLogger(logger)
+	switch {
+	case logger.Core().Enabled(zap.DebugLevel):
+		l = zapgrpc.NewLogger(logger, zapgrpc.WithDebug())
+		options = append(options, elastic.SetTraceLog(l))
+	case logger.Core().Enabled(zap.InfoLevel):
```
yurishkuro, line 373: I am not sure this is a good idea. Our default log level is info, so we're going to be adding a log line for every span written in the collector.

yurishkuro (dismissed):

pkg/es/config/config.go
albertteoh: Okay, fair point about the collector. The motivation for this PR was to log ES queries so I think just showing the tracelog when `debug` logging is enabled is enough, and the rest, including info level in jaeger, should default to `ErrorLevel` logging in ES client.

It also brings up the question of whether the debug logs for collector would be too much as it would log the entire span's content which can be very large. The only log line the collector writes when `log-level=debug` is:
```
{"level":"debug","ts":1615007007.183977,"caller":"app/span_processor.go:149","msg":"Span written to the storage by the collector","trace-id":"4a9afb787bbb4f3f","span-id":"4a9afb787bbb4f3f"}
```

which I can see a use case for when debugging ingestion. With ES queries in the logging mix, it may be hard to see the above logs without grepping. As such, maybe we need to isolate this feature to jaeger-query and not jaeger-collector?

yurishkuro: You're probably better positioned to answer this since you're running this setup in prod.

Another option instead of piggy-backing on the main log level is to add just an `--es.` flag for log level, then people can go nuts if they want to, all the way to full payload.

albertteoh: Actually, we don't use jaeger-collector with ES as the backing store; although I do like your suggestion of having a separate flag (I propose `--es.log-level`), which I'll aim to do in the coming day or so.

albertteoh: @yurishkuro I've added the `--es.log-level` (and `--es-archive.log-level`) flag and tested it together with `--log-level` in different combinations of values. Both the `--es` and non-`--es` flags are decoupled, working independently of one another.

yurishkuro (approved):

PR #2865, opened 2021-03-05
Fix es-archive namespace default values

### LINKED ISSUE #2853: Fix archive default values reported for ES and Cassandra
**Describe the bug**
Noticed when @mehta-ankit_gitlab raised the following question in Gitter:
> Question regarding Archive option for Jaeger using Elasticsearch backend.
--es-archive.max-doc-count is set to 0 and --es-archive.max-span-age is set to 0s as default.
Does someone know, if these defaults mean age of the span (inside the archive index) is ignored when querying it.

When checking the `--help` docs, the default values are missing or incorrect for the `*-archive*` flags.

I confirmed the defaults should be the same as the non-archive equivalents as these are the values used for archive flags internally if unset; i.e. this bug only impacts `--help` documentation.

This also impacts the jaeger documentation: https://www.jaegertracing.io/docs/latest/cli/

**To Reproduce**
1. `docker run -e SPAN_STORAGE_TYPE=elasticsearch jaegertracing/jaeger-query --help` .
2. Same problem when changing storage type to `cassandra`.
3. Same problem with `jaeger-collector` as well.

**Expected behavior**
Got: 
```
      --es-archive.max-doc-count int                       The maximum document count to return from an Elasticsearch query. This will also apply to aggregations.
      --es-archive.max-span-age duration                   The maximum lookback for spans in Elasticsearch (default 0s)
      --es-archive.num-replicas int                        The number of replicas per index in Elasticsearch
      --es-archive.num-shards int                          The number of shards per index in Elasticsearch
```

Want:
```
      --es-archive.max-doc-count int                       The maximum document count to return from an Elasticsearch query. This will also apply to aggregations. (default 10000)
      --es-archive.max-span-age duration                   The maximum lookback for spans in Elasticsearch (default 72h0m0s)
      --es-archive.num-replicas int                        The number of replicas per index in Elasticsearch (default 1)
      --es-archive.num-shards int                          The number of shards per index in Elasticsearch (default 5)
```

 - OS: Linux & MacOS
 - Jaeger version: latest
 - Deployment: Dev

**Additional Context**
I propose the fix should be to share the primary (e.g. non-archive) default configs with all other additional namespaces like `archive`.

albertteoh:

- Fixes #2853

- Sets the default config of "other" namespaces to the primary's default config.
- As discussed in #2853, the primary default of `72h` is misleading to users because an archive fetch should have no "lookback" limit, fetching any traceID that is in archive storage. 

  This PR also removes the `--es-archive.max-span-age` flag to avoid confusing users.

## Testing
- Confirmed `--es-archive.max-span-age` no longer appears in `--help` for both query and collector.
- Confirmed archive defaults are correctly based on primary config defaults for both query and collector. E.g.:
```
      --es-archive.max-doc-count int                       The maximum document count to return from an Elasticsearch query. This will also apply to aggregations. (default 10000)
 ```

### CODE REVIEW

**Review by yurishkuro — DISMISSED** [review]
Might warrant a "breaking change" entry in the changelog

**Comment by albertteoh** [comment]
> Might warrant a "breaking change" entry in the changelog

👍 Breaking change entry added.

**Comment by yurishkuro** [comment]
the ES integration tests don't seem to like this change

File: plugin/storage/es/options.go [hunk 1]
```diff
@@ -83,35 +83,41 @@ type namespaceConfig struct {
...
-			},
-			namespace: primaryNamespace,
+			Configuration: defaultConfig,
+			namespace:     primaryNamespace,
 		},
 		others: make(map[string]*namespaceConfig, len(otherNamespaces)),
 	}
 
+	// Other namespaces need to be explicitly enabled.
+	defaultConfig.Enabled = false
```
albertteoh, line 115: This should fix both the failing ES integration tests and Crossdock builds.

The cause was due to archive mode being enabled by default and attempting to either ping a non-existent ES for a version (failing ES integration test) or attempting to communicate with ES on the wrong (default) IP address (failing Crossdock builds).

The original behaviour was to have archive mode disabled.

PR #2870, opened 2021-03-09
Inject trace context to grpc metadata

### LINKED ISSUE #2869: Support trace propagation for gRPC plugin

Support trace propagation for gRPC plugin

Currently, the trace context is not propagated while using gRPC plugin system. Thus, for example, traces from query API called by UI app and those from actual gRPC plugin are separated. But, they are expected to work together to provide more performance insight for the additional storage.

So far, there have been already many use-cases for gRPC storage plugin, including Logz.io, InfluxDB and Couchbase as mentioned in the issue #638.

In our production cluster, we are using [AlibabaCloud LogService](https://www.alibabacloud.com/product/log-service) as our storage backend which is heavily inspired by https://github.com/aliyun/aliyun-log-jaeger. And it works as a gRPC plugin in our case.

Inject trace context manually into grpc metadata since `DialOption` is not supported by go-plugin module now as mentioned in this issue https://github.com/hashicorp/go-plugin/issues/166 and this project seems not in active dev.

It would work similar to `upgradeContextWithBearerToken(Context) -> Context`.

A PR will be open to resolve this issue.

lujiajing1126:
Signed-off-by: Megrez Lu <lujiajing1126@gmail.com>

Resolve #2869 

- Update `go-plugin` to `1.4.2`
- Add interceptors to support tracing propagation


plugin/storage/grpc/shared/archive.go
```diff
@@ -44,7 +44,7 @@ type archiveWriter struct {
 
 // GetTrace takes a traceID and returns a Trace associated with that traceID from Archive Storage
 func (r *archiveReader) GetTrace(ctx context.Context, traceID model.TraceID) (*model.Trace, error) {
-	stream, err := r.client.GetArchiveTrace(upgradeContextWithBearerToken(ctx), &storage_v1.GetTraceRequest{
+	stream, err := r.client.GetArchiveTrace(upgradeWithTraceContext(upgradeContextWithBearerToken(ctx)), &storage_v1.GetTraceRequest{
```
jpkrohling, line 47: I think it might be time to split the statement here, creating the context in one and using it in another.

plugin/storage/grpc/shared/grpc_client.go
```diff
@@ -52,11 +54,45 @@ type grpcClient struct {
...
 	if hasToken {
-		requestMetadata := metadata.New(map[string]string{
-			spanstore.BearerTokenKey: bearerToken,
-		})
-		return metadata.NewOutgoingContext(ctx, requestMetadata)
+		md, ok := metadata.FromOutgoingContext(ctx)
+		if !ok {
+			md = metadata.New(nil)
+		} else {
+			md = md.Copy()
```
jpkrohling, line 61: Why do you need a copy?

plugin/storage/grpc/shared/metadata.go
```diff
@@ -0,0 +1,51 @@
+// Copyright (c) 2019 The Jaeger Authors.
```
jpkrohling, line 1: oops?

plugin/storage/grpc/shared/grpc_client.go
```diff
@@ -85,7 +121,7 @@ func (c *grpcClient) ArchiveSpanWriter() spanstore.Writer {
 
 // GetTrace takes a traceID and returns a Trace associated with that traceID
 func (c *grpcClient) GetTrace(ctx context.Context, traceID model.TraceID) (*model.Trace, error) {
-	stream, err := c.readerClient.GetTrace(upgradeContextWithBearerToken(ctx), &storage_v1.GetTraceRequest{
+	stream, err := c.readerClient.GetTrace(upgradeWithTraceContext(upgradeContextWithBearerToken(ctx)), &storage_v1.GetTraceRequest{
```
jpkrohling, line 124: Given the number of places calling the two upgrade functions, it might be a good idea to combine both in another function, and use only this other function.

plugin/storage/grpc/shared/archive.go
lujiajing1126: I've used a functional type with a composition function to refactor this part. I suppose it is fine now.

plugin/storage/grpc/shared/grpc_client.go
lujiajing1126: Removed. I've checked it is unnecessary.

plugin/storage/grpc/shared/metadata.go
lujiajing1126: Changed to 2021

plugin/storage/grpc/shared/grpc_client.go
lujiajing1126: Refactored in the same way

lujiajing1126:
@jpkrohling Any idea why unit test failed? I don't think it is related to my latest modification

plugin/storage/grpc/shared/grpc_client.go
```diff
@@ -46,20 +51,63 @@ type grpcClient struct {
...
+// upgradeWithTraceContext injects the current active span given into the outgoing gRPC metadata,
+// and returns the context being decorated.
+// If no global tracer is registered or no active span is found in the context,
+// returns the original context.
+func upgradeWithTraceContext(ctx context.Context) context.Context {
+	if !opentracing.IsGlobalTracerRegistered() {
+		return ctx
+	}
+
+	if span := opentracing.SpanFromContext(ctx); span != nil {
```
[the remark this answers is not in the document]
lujiajing1126: Yes. I agree with you that a child span should be created. But this is acutally a compromise here that I make use of the active span from the context.

The difficulty stems from the fact that `go-plugin` does not support customized `DialOption` which I have mentioned in the issue #2869.

The server-side code can use https://github.com/opentracing-contrib/go-grpc

lujiajing1126:
@jpkrohling @jpkrohling I've updated to the latest `go-plugin(v1.4.2)` to enable client-side gRPC tracing. Could you kindly have a review again?

plugin/storage/grpc/config/config.go
```diff
@@ -16,11 +16,14 @@ package config
 
 import (
 	"fmt"
+	"github.com/opentracing/opentracing-go"
```
yurishkuro, line 19: please move to the next group (I think `make fmt` will do this automatically)

yurishkuro (commented):
* I see you're only changing grpc-client. Are there similar changes needed for the server?
* since there are no tests for tracing instrumentation of the plugin, could you try a manual test and post a screenshot of the trace?

plugin/storage/grpc/config/config.go
lujiajing1126: Fixed

lujiajing1126:
> * I see you're only changing grpc-client. Are there similar changes needed for the server?

The plugin developer can enable tracing while starting the plugin server.

```go
grpc.ServeWithGRPCServer(
				store.NewAliyunSLSStore(&appConfig, logger),
				func(options []googleGRPC.ServerOption) *googleGRPC.Server {
					return plugin.DefaultGRPCServer([]googleGRPC.ServerOption{
						googleGRPC.UnaryInterceptor(otgrpc.OpenTracingServerInterceptor(tracer)),
						googleGRPC.StreamInterceptor(otgrpc.OpenTracingStreamServerInterceptor(tracer)),
					})
				},
			)
```

Do you need to make it a default option?

> * since there are no tests for tracing instrumentation of the plugin, could you try a manual test and post a screenshot of the trace?

image

yurishkuro:
I don't know if there's a place to make this a default option, at minimum it would make sense to mention this in the plugin/README, and perhaps add to examples/memstore-plugin/main.go ?

lujiajing1126:
> I don't know if there's a place to make this a default option, at minimum it would make sense to mention this in the plugin/README, and perhaps add to examples/memstore-plugin/main.go ?

I've added a `tracing` section in the document as well as code in the memstore example.

jpkrohling (approved):

jpkrohling:
Should 1.23 wait for this one?

yurishkuro (approved):

Issue #2760, opened 2021-02-02
Flaky test TestKafkaStorage/GetTrace


jpkrohling:
Seen here: https://github.com/jaegertracing/jaeger/pull/2759/checks?check_run_id=1813633773. The mentioned PR does not touch Kafka at all.

```
    integration.go:188: 
        	Error Trace:	integration.go:188
        	Error:      	Should be true
        	Test:       	TestKafkaStorage/GetTrace
    trace_compare.go:57: 
        	Error Trace:	trace_compare.go:57
        	            				integration.go:189
        	Error:      	Expected value not to be nil.
        	Test:       	TestKafkaStorage/GetTrace
--- FAIL: TestKafkaStorage (4.15s)
    --- FAIL: TestKafkaStorage/GetTrace (3.02s)
```

LostLaser:
I'd like to give this one a shot

yurishkuro:
+1

LostLaser:
If anyone is following this, I have been able to get a consistent reproduction of this behavior by running the described test in quick succession. I haven't found a specific root cause that would explain the behavior as of yet but I hope to soon.

yurishkuro:
nice!

LostLaser:
As mentioned a few days ago, I was able to reproduce the same error message mentioned in this issue by running this test case in quick succession given these conditions:
1. The docker container that hosts the Kafka/Zookeeper instance persists between runs (variance from the test suite that stops the docker process after each run)
2. The test case is ran, and then ran again at most approximately 15 seconds after the initial run. Subsequent runs within the 15 second time frame of the previous will produce errors.

Based on the attached Wireshark logs I have narrowed this issue down to a few possibilities:
1. The connection after each run is not being properly terminated. 
2. The Kafka container used to test is misconfigured in a way that restricts connections.
3. There is a bug in the shopify/sarama package that is used to interface with Kafka with a potential root cause being that there is no assertion the Kafka group has been joined before sending messages.

The most important thing to note about the Wireshark logs, other than the fact that there are no fetch calls for the failing case, is the order in which the join group response arrives when comparing the successful run and the failing run. In some instances the join group response would come ~2-3 seconds after the test had already failed.

### Wireshark logs
Failing run output:
![kafka-failure](https://user-images.githubusercontent.com/29441668/107727144-40139080-6cb0-11eb-87d8-332921fbee0f.png)
Successful run output:
[...]

PR #2873, opened 2021-03-10
Increasing test timeout for TestKafkaStorage

LostLaser:
Signed-off-by: Jacob Goldsworthy <jtgolds6@gmail.com>

- Addresses: #2760

- Increase timeout of storage integration tests to allow for longer storage settle time

### DIFF
--- a/plugin/storage/integration/integration.go
+++ b/plugin/storage/integration/integration.go
@@ -36,7 +36,7 @@ import (
 )
 
 const (
-	iterations = 40
+	iterations = 100
 )
 
 // StorageIntegration holds components for storage integration test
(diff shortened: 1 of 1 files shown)


plugin/storage/integration/integration.go
```diff
@@ -36,7 +36,7 @@ import (
 )
 
 const (
-	iterations = 30
+	iterations = 40
```
yurishkuro, line 39: go to 100

albertteoh (dismissed):
lgtm

plugin/storage/integration/integration.go
LostLaser: Acked writes didn't seem to help unfortunately. I attempted to switch the acks to "local" as well as "all" to no avail. I'm guessing the ack never actually comes into effect because there is probably a timeout on the message transfer that is greater than the test time we have set. I ran kafkacat to see if the spans ever made it to the cluster and it seemed like the messages would always make it after some time.

yurishkuro (approved):

yurishkuro:
thanks @LostLaser

PR #2874, opened 2021-03-11
[es] Add remote read clusters option for cross-cluster querying

### LINKED ISSUE #1883: Elasticsearch Cross Cluster Query

dgrizzanti:
- Enables elasticsearch [cross-cluster querying](https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-cross-cluster-search.html)

Resolves #1883
Updates to PR #1892

The cross cluster query api functions by appending a pre-configured remote cluster to an index e.g. cluster_one:index_name. This change simply applies a prefix for each remote cluster and for each index. You will end up with `total_indicies = indicies * remote_clusters`. Also added a special case local which will not append a prefix and query the local cluster.

This works with both standard and rollover indices since it mostly comes down to a simple string manipulation.

### DIFF
--- a/pkg/es/config/config.go
+++ b/pkg/es/config/config.go
@@ -45,6 +45,7 @@ import (
 // Configuration describes the configuration properties needed to connect to an ElasticSearch cluster
 type Configuration struct {
 	Servers               []string       `mapstructure:"server_urls"`
+	RemoteReadClusters    []string       `mapstructure:"remote_read_clusters"`
 	Username              string         `mapstructure:"username"`
 	Password              string         `mapstructure:"password" json:"-"`
 	TokenFilePath         string         `mapstructure:"token_file"`
@@ -89,6 +90,7 @@ type TagsAsFields struct {
 // ClientBuilder creates new es.Client
 type ClientBuilder interface {
 	NewClient(logger *zap.Logger, metricsFactory metrics.Factory) (es.Client, error)
+	GetRemoteReadClusters() []string
 	GetNumShards() int64
 	GetNumReplicas() int64
 	GetMaxSpanAge() time.Duration
@@ -193,6 +195,9 @@ func (c *Configuration) NewClient(logger *zap.Logger, metricsFactory metrics.Fac
 
 // ApplyDefaults copies settings from source unless its own value is non-zero.
 func (c *Configuration) ApplyDefaults(source *Configuration) {
+	if len(c.RemoteReadClusters) == 0 {
+		c.RemoteReadClusters = source.RemoteReadClusters
+	}
 	if c.Username == "" {
 		c.Username = source.Username
 	}
@@ -244,6 +249,11 @@ func (c *Configuration) ApplyDefaults(source *Configuration) {
 	if c.LogLevel == "" {
 		c.LogLevel = source.LogLevel
 	}
+}
+
+// GetRemoteReadClusters returns list of remote read clusters
+func (c *Configuration) GetRemoteReadClusters() []string {
+	return c.RemoteReadClusters
 }
 
 // GetNumShards returns number of shards from Configuration
--- a/plugin/storage/es/options_test.go
+++ b/plugin/storage/es/options_test.go
@@ -31,6 +31,7 @@ func TestOptions(t *testing.T) {
 	assert.Empty(t, primary.Username)
 	assert.Empty(t, primary.Password)
 	assert.NotEmpty(t, primary.Servers)
+	assert.Empty(t, primary.RemoteReadClusters)
 	assert.Equal(t, int64(5), primary.NumShards)
 	assert.Equal(t, int64(1), primary.NumReplicas)
 	assert.Equal(t, 72*time.Hour, primary.MaxSpanAge)
@@ -58,6 +59,7 @@ func TestOptionsWithFlags(t *testing.T) {
 		"--es.num-replicas=10",
 		"--es.index-date-separator=",
 		// a couple overrides
+		"--es.remote-read-clusters=cluster_one,cluster_two",
 		"--es.aux.server-urls=3.3.3.3, 4.4.4.4",
 		"--es.aux.max-span-age=24h",
 		"--es.aux.num-replicas=10",
@@ -77,6 +79,7 @@ func TestOptionsWithFlags(t *testing.T) {
 	assert.Equal(t, "hello", primary.Username)
 	assert.Equal(t, "/foo/bar", primary.TokenFilePath)
 	assert.Equal(t, []string{"1.1.1.1", "2.2.2.2"}, primary.Servers)
+	assert.Equal(t, []string{"cluster_one", "cluster_two"}, primary.RemoteReadClusters)
 	assert.Equal(t, 48*time.Hour, primary.MaxSpanAge)
 	assert.True(t, primary.Sniffer)
 	assert.True(t, primary.SnifferTLSEnabled)
@@ -101,6 +104,19 @@ func TestOptionsWithFlags(t *testing.T) {
 	assert.Equal(t, "test,tags", aux.Tags.Include)
 	assert.Equal(t, "2006.01.02", aux.IndexDateLayout)
 	assert.True(t, primary.UseILM)
+}
+
+func TestEmptyRemoteReadClusters(t *testing.T) {
+	opts := NewOptions("es", "es.aux")
+	v, command := config.Viperize(opts.AddFlags)
+	err := command.ParseFlags([]string{
+		"--es.remote-read-clusters=",
+	})
+	require.NoError(t, err)
+	opts.InitFromViper(v)
+
+	primary := opts.GetPrimary()
+	assert.Equal(t, []string{}, primary.RemoteReadClusters)
 }
 
 func TestMaxSpanAgeSetErrorInArchiveMode(t *testing.T) {
--- a/plugin/storage/es/spanstore/reader.go
+++ b/plugin/storage/es/spanstore/reader.go
@@ -116,6 +116,7 @@ type SpanReaderParams struct {
 	TagDotReplacement   string
 	Archive             bool
 	UseReadWriteAliases bool
+	RemoteReadClusters  []string
 }
 
 // NewSpanReader returns a new SpanReader with a metrics.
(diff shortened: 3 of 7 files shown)


plugin/storage/es/spanstore/reader_test.go
```diff
@@ -149,35 +149,40 @@ func TestSpanReaderIndices(t *testing.T) {
...
 		{params: SpanReaderParams{Client: client, Logger: logger, MetricsFactory: metricsFactory,
 			IndexPrefix: "foo:", Archive: true},
-			index: "foo:" + indexPrefixSeparator + spanIndex + archiveIndexSuffix},
+			indices: []string{"foo:" + indexPrefixSeparator + spanIndex + archiveIndexSuffix}},
 		{params: SpanReaderParams{Client: client, Logger: logger, MetricsFactory: metricsFactory,
 			IndexPrefix: "foo:", Archive: true, UseReadWriteAliases: true},
-			index: "foo:" + indexPrefixSeparator + spanIndex + archiveReadIndexSuffix},
+			indices: []string{"foo:" + indexPrefixSeparator + spanIndex + archiveReadIndexSuffix}},
+		{params: SpanReaderParams{Client: client, Logger: logger, MetricsFactory: metricsFactory,
+			IndexPrefix: "", Archive: false, RemoteReadClusters: []string{"cluster_one", "cluster_two"}},
```
albertteoh, line 177: It'd be good to add two more test cases for `RemoteReadClusters: []string{"cluster_one", "cluster_two"}}` together with because the `addRemoteReadClusters` calls could be removed for archive and alias mode without test failure:
- `UseReadWriteAliases: true`
- `Archive: true`

plugin/storage/es/spanstore/reader_test.go
```diff
@@ -149,35 +149,40 @@ func TestSpanReaderIndices(t *testing.T) {
...
+			indices: []string{spanIndex + archiveIndexSuffix}},
 		{params: SpanReaderParams{Client: client, Logger: logger, MetricsFactory: metricsFactory,
 			IndexPrefix: "foo:", Archive: true},
-			index: "foo:" + indexPrefixSeparator + spanIndex + archiveIndexSuffix},
+			indices: []string{"foo:" + indexPrefixSeparator + spanIndex + archiveIndexSuffix}},
 		{params: SpanReaderParams{Client: client, Logger: logger, MetricsFactory: metricsFactory,
 			IndexPrefix: "foo:", Archive: true, UseReadWriteAliases: true},
-			index: "foo:" + indexPrefixSeparator + spanIndex + archiveReadIndexSuffix},
+			indices: []string{"foo:" + indexPrefixSeparator + spanIndex + archiveReadIndexSuffix}},
+		{params: SpanReaderParams{Client: client, Logger: logger, MetricsFactory: metricsFactory,
```
albertteoh, line 176: Please refer to @yurishkuro's [comment](https://github.com/jaegertracing/jaeger/pull/1892/files#r344831587) about making the test cases multiline.

plugin/storage/es/spanstore/reader.go
```diff
@@ -139,24 +140,43 @@ type timeRangeIndexFn func(indexName string, indexDateLayout string, startTime t
...
-		return func(indices string, indexDateLayout string, startTime time.Time, endTime time.Time) []string {
+		return addRemoteReadClusters(func(indices string, indexDateLayout string, startTime time.Time, endTime time.Time) []string {
 			return []string{indices + "read"}
+		}, remoteReadClusters)
+	}
+	return addRemoteReadClusters(timeRangeIndices, remoteReadClusters)
+}
+
+// Elasticsearch cross cluster query api example: GET /twitter,cluster_one:twitter,cluster_two:twitter/_search
+// Add a remote cluster prefix for each cluster and for each index and add it to the list of original indices
```
albertteoh, line 164: nit: Please end comment sentences with a full-stop. 

Same as [above](https://github.com/jaegertracing/jaeger/pull/2874/files#diff-de5aae4da3cb03d3a73919aca1f6f38181b906dfd205bb0def4bf7b9d6042992R254) though I understand this convention is not followed in the adjacent functions below it.

dgrizzanti:
@albertteoh thanks for the comments and suggestions. I think I covered all of them if you can take a look again. Thanks!

plugin/storage/integration/integration.go
```diff
@@ -36,7 +36,7 @@ import (
 )
 
 const (
-	iterations = 30
+	iterations = 100
```
albertteoh, line 39: Were these changes brought in from #2873?

albertteoh:
> @albertteoh thanks for the comments and suggestions. I think I covered all of them if you can take a look again. Thanks!

[One more](https://github.com/jaegertracing/jaeger/pull/2874/files#r592908331) :) Also, it seems another commit was pulled into this PR: https://github.com/jaegertracing/jaeger/pull/2874/commits/510523707eec305870eafa17845459aaf7bd7a04.

plugin/storage/integration/integration.go
dgrizzanti: I did rebase so may be they got pulled in by accident? I flip it back so it doesn't show up in this PR.

plugin/storage/es/spanstore/reader.go
```diff
@@ -139,24 +140,44 @@ type timeRangeIndexFn func(indexName string, indexDateLayout string, startTime t
...
+// Elasticsearch cross cluster api example GET /twitter,cluster_one:twitter,cluster_two:twitter/_search.
+func addRemoteReadClusters(fn timeRangeIndexFn, remoteReadClusters []string) timeRangeIndexFn {
+	return func(indexPrefix string, indexDateLayout string, startTime time.Time, endTime time.Time) []string {
+		jaegerIndices := fn(indexPrefix, indexDateLayout, startTime, endTime)
+		if len(remoteReadClusters) == 0 {
+			return jaegerIndices
 		}
+
+		for _, jaegerIndex := range jaegerIndices {
+			for _, remoteCluster := range remoteReadClusters {
```
albertteoh, line 173: Related to the earlier comment, if `remoteReadClusters` defaults to `[]string{""}`, the returned indices count would be doubled with the new indices looking like `:<index-name>`. Is that expected?

plugin/storage/integration/integration.go
albertteoh: Yup, I had more luck with `git merge` and did recall having similar problems when using `git rebase`. Thanks 🙏

pkg/es/config/config.go
```diff
@@ -193,6 +195,9 @@ func (c *Configuration) NewClient(logger *zap.Logger, metricsFactory metrics.Fac
 
 // ApplyDefaults copies settings from source unless its own value is non-zero.
 func (c *Configuration) ApplyDefaults(source *Configuration) {
+	if len(c.RemoteReadClusters) == 0 {
+		c.RemoteReadClusters = source.RemoteReadClusters
```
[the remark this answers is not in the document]
dgrizzanti: Thanks @albertteoh - I didn't realize this nuance with `strings.Split()` but after looking at the docs, this makes sense. I didn't see a great (easy) way around this since the cli takes in strings, so I updated `options.go` to check the length and only split when necessary (otherwise default to `[]string{}`). I added 2 tests cases around options as well to cover this.

plugin/storage/es/spanstore/reader.go
dgrizzanti: This should be avoided now with the changes to `plugin/storage/es/options.go`

plugin/storage/es/options_test.go
```diff
@@ -103,6 +106,19 @@ func TestOptionsWithFlags(t *testing.T) {
...
+	opts := NewOptions("es", "es.aux")
+	v, command := config.Viperize(opts.AddFlags)
+	err := command.ParseFlags([]string{
+		"--es.remote-read-clusters=",
+	})
+	require.NoError(t, err)
+	opts.InitFromViper(v)
+
+	primary := opts.GetPrimary()
+	assert.Equal(t, []string{}, primary.RemoteReadClusters)
```
albertteoh, line 119: I think the following would be more expressive/readable: `assert.Empty(t, primary.RemoteReadClusters)`

pkg/es/config/config.go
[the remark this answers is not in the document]
albertteoh: Thanks for adding the test cases to cover the empty string. I think there's also the following case: `--es.remote-read-clusters=cluster_one,,cluster_two`, where the second item is empty but I think we can consider this a user error.

albertteoh (approved):
lgtm, thanks @dgrizzanti.

yurishkuro:
Thank you!

PR #2886, opened 2021-03-18
Setup CI for v2

### LINKED ISSUE #2883: Setup CI for v2

We need to set up a custom CI infra for the v2, as there are things from v1 that we might not need, do differently, or that don't exist yet.

- [ ] Set the current CI to ignore PRs that include only changes to v2
- [ ] Create a new Makefile inside the v2 with the targets we need for v2
- [ ] Create a new v2-specific GitHub workflows

jpkrohling:
- Moved v1-specific workflows to .github/workflows/v1/
- Makefile for v1 should ignore v2 files (linting, header checks, ...)
- Add the manifests for the agent
- Add builder configuration to build the agent
- Add goreleaser to perform a release when new v2 tags are created

Fixes #2883


v2/storage/memory/config.go
```diff
@@ -1,10 +1,10 @@
-// Copyright (c) 2021 The Jaeger Authors.
```
albertteoh, line 1: Why remove the copyright year?

jpkrohling: My understanding from other communities is that it isn't really required. And not having it makes it less cumbersome when the year changes.

v2/Makefile
```diff
@@ -0,0 +1,54 @@
+VERSION ?= 2.0.0-apha2
```
Ashmita152, line 1: Is it a typo in alpha.

v2/.goreleaser.yaml
```diff
@@ -0,0 +1,145 @@
...
+    binary: jaeger-agent
+    flags: 
+      - -trimpath
+    gcflags:
+      - all=-N -l
+    env:
+      - CGO_ENABLED=0
+    goos:
+      - linux
+      - darwin
```
Ashmita152, line 24: Do we need to support windows in v2 as well as we do in v1 ?

v2/Makefile
jpkrohling: Great catch!

v2/.goreleaser.yaml
jpkrohling: Yes. I temporarily removed as I hit an error and forgot to put it back. Thanks for spotting it!

v2/.goreleaser.yaml
```diff
@@ -0,0 +1,145 @@
...
+  ## 1. regular, production optimized
+  ## 2. debug, with no optimization or inlining
+  - &agent-build
+    id: jaeger-agent
+    dir: ./_build/agent
+    binary: jaeger-agent
+    flags: 
+      - -trimpath
+    gcflags:
+      - all=-N -l
```
Ashmita152, line 19: Does this means we are turning off compiler optimizations in normal image build too ?

jpkrohling: Oops!

PR #2900, opened 2021-03-25
change variable name replacing golang's keyword "error"

wangchenglong01:
- Code specification：change variable name replacing golang's keyword "error"

- "error" is golang's keyword. It's better not to make it as a variable.

### DIFF
--- a/cmd/query/main.go
+++ b/cmd/query/main.go
@@ -145,8 +145,8 @@ func main() {
 		app.AddFlags,
 	)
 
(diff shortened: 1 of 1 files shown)


yurishkuro (commented):
may want to do it in another place too
```
./cmd/anonymizer/main.go:92:	if error := command.Execute(); error != nil {
./cmd/query/main.go:148:	if error := command.Execute(); error != nil {
```

PR #2906, opened 2021-03-26
Add jaeger-query HTTP handler diagnostic logging

albertteoh:

- Jaeger-query was swallowing json marshalling errors, making it difficult to root-cause issues during development (while adding a new `/metrics` API endpoint).

- Handle `json.Marshal` and `ResponseWriter.Write` errors and log them. Log success at debug level as well.


yurishkuro (commented):
The error paths can me tested

albertteoh:
> The error paths can me tested

Yup, sloppiness on my part. I had a go at writing tests for it, please let me know if there's a more idiomatic approach.

cmd/query/app/http_handler.go
```diff
@@ -454,7 +454,15 @@ func (aH *APIHandler) writeJSON(w http.ResponseWriter, r *http.Request, response
...
+	if err != nil {
+		aH.handleError(w, fmt.Errorf("failed marshalling HTTP response to JSON: %w", err), http.StatusInternalServerError)
+		return
+	}
 	w.Header().Set("Content-Type", "application/json")
-	w.Write(resp)
+	if b, err := w.Write(resp); err != nil {
+		aH.handleError(w, fmt.Errorf("failed writing HTTP response: %w", err), http.StatusInternalServerError)
+	} else {
+		aH.logger.Debug("Successfully wrote HTTP response", zap.Int("bytes", b))
```
yurishkuro, line 466: I'm not a fan of this debug logging, is it necessary? It should be observable from the client side already.

cmd/query/app/http_handler_test.go
```diff
@@ -185,16 +186,59 @@ func TestLogOnServerError(t *testing.T) {
...
+func (h *httpResponseErrWriter) Write([]byte) (int, error) {
+	return 0, fmt.Errorf("failed to write")
+}
+func (h *httpResponseErrWriter) WriteHeader(statusCode int) {}
+func (h *httpResponseErrWriter) Header() http.Header {
+	return http.Header{}
+}
+
+// ErrHandlerFunc injects the httpResponseErrWriter into the HTTP handler.
+type ErrHandlerFunc func(http.ResponseWriter, *http.Request)
```
yurishkuro, line 202: This type isn't necessary, you can inline in L262 using http.HandlerFunc

yurishkuro (approved):

albertteoh:
Thanks for the review!

PR #2909, opened 2021-03-28
Add deprecation notes in documentation

### LINKED ISSUE #2113: Tracking and removing deprecated CLI flags

albertteoh:

- Fixes #2113

- Sets in stone the agreed upon proposed protocol for handling CLI flag deprecation and removal.
- README.md updated for the benefit of users who want to understand the stability guarantees.
- CONTRIBUTING.md updated to provide instructions to contributors introducing and removing deprecated CLI flags.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -192,7 +192,7 @@ After the merge make sure referenced issues were closed.
 * When reading the flag into config, log a warning with the same warning.
 * Take care of deprecated flags in `initFromViper` functions, do not pass them to business functions.
 
-## Removing Deprecated CLI Flags
+### Removing Deprecated CLI Flags
 * Ensure all references to the flag's variables have been removed in code.
 * Ensure a "Breaking Changes" entry is added in the [CHANGELOG](./CHANGELOG.md) indicating which CLI flag
 is being removed and which CLI flag should be used in favor of this removed flag.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -173,12 +173,12 @@ Before merging a PR make sure:
 Merge the PR by using "Squash and merge" option on Github. Avoid creating merge commits.
 After the merge make sure referenced issues were closed.
 
-## Deprecated CLI Flags
+## Deprecating CLI Flags
 
-* If a flag is deprecated in release N, it can be removed in release N+2.
-* When adding a (deprecated) prefix to the flags, indicate via a deprecation message that the flag could be removed after 3 months. For example:
+* If a flag is deprecated in release N, it can be removed in release N+2 or three months later, whichever is later.
+* When adding a (deprecated) prefix to the flags, indicate via a deprecation message that the flag could be removed in the future. For example:
   ```
-  deprecated, will be removed after 2020-03-15 or in release v1.19.0, whichever is later
+  (deprecated, will be removed after 2020-03-15 or in release v1.19.0, whichever is later)
   ```
 * At the top of the file where the flag name is defined, add a constant and a comment, e.g.
   ```
@@ -189,7 +189,7 @@ After the merge make sure referenced issues were closed.
   ```
   flagSet.Int(healthCheckHTTPPort, 0, healthCheckHTTPPortWarning+" see --"+adminHTTPHostPort)
   ```
-* When reading the flag into config, log a warning with the same warning.
+* When parsing a deprecated flag into config, log a warning with the same deprecation message
 * Take care of deprecated flags in `initFromViper` functions, do not pass them to business functions.
 
 ### Removing Deprecated CLI Flags
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -172,3 +172,32 @@ Before merging a PR make sure:
 
 Merge the PR by using "Squash and merge" option on Github. Avoid creating merge commits.
 After the merge make sure referenced issues were closed.
+
+## Deprecated CLI Flags
+
+* If a flag is deprecated in release N, it can be removed in release N+2.
+* When adding a (deprecated) prefix to the flags, indicate via a deprecation message that the flag could be removed after 3 months. For example:
+  ```
+  deprecated, will be removed after 2020-03-15 or in release v1.19.0, whichever is later
+  ```
+* At the top of the file where the flag name is defined, add a constant and a comment, e.g.
+  ```
+  // TODO deprecated flag to be removed
+  healthCheckHTTPPortWarning = "(deprecated, will be removed after 2020-03-15 or in release v1.19.0, whichever is later)"
+  ```
+* Use that constant as the prefix to the help text, e.g.
+  ```
+  flagSet.Int(healthCheckHTTPPort, 0, healthCheckHTTPPortWarning+" see --"+adminHTTPHostPort)
+  ```
+* When reading the flag into config, log a warning with the same warning.
+* Take care of deprecated flags in `initFromViper` functions, do not pass them to business functions.
+
+## Removing Deprecated CLI Flags
+* Ensure all references to the flag's variables have been removed in code.
+* Ensure a "Breaking Changes" entry is added in the [CHANGELOG](./CHANGELOG.md) indicating which CLI flag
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -172,3 +172,32 @@ Before merging a PR make sure:
 
 Merge the PR by using "Squash and merge" option on Github. Avoid creating merge commits.
 After the merge make sure referenced issues were closed.
+
+## Deprecated CLI Flags
```
albertteoh, line 176: This section is mostly a copy of the description from #2113.

README.md
```diff
@@ -203,3 +203,27 @@ If you would like to add your organization to the list, please comment on our
...
+
+In short, for a deprecated CLI flag, you should expect to see the following message in the `--help` documentation:
+```
+(deprecated, will be removed after yyyy-mm-dd or in release vX.Y.Z, whichever is later)
+```
+
+A grace period of at least **3 months** or **two minor version bumps** from the first release
+containing the deprecation notice (whichever is later) will be provided before the deprecated CLI flag _can_ be deleted.
+
+For example, if hypothetically, v1.28.0 is released on 01/06/2021 containing a deprecation notice for a CLI flag.
```
yurishkuro, line 227:
```suggestion
For example, if hypothetically, v1.28.0 is released on 01-Jun-2021 containing a deprecation notice for a CLI flag.
```
Use universally readable dates

README.md
```diff
@@ -203,3 +203,27 @@ If you would like to add your organization to the list, please comment on our
 [slack]: https://cloud-native.slack.com/archives/CGG7NFUJ3
 [slack-join]: https://slack.cncf.io
 [slack-img]: https://img.shields.io/badge/slack-join%20chat%20%E2%86%92-brightgreen?logo=slack
```
yurishkuro, line 205: Please keep references at the end of the doc

README.md
```diff
@@ -203,3 +203,27 @@ If you would like to add your organization to the list, please comment on our
...
+
+## Releases
```
yurishkuro, line 207: This doesn't feel like the right place, needs to be much higher in the doc. Also, rather than being under "releases" heading, perhaps it should be phrased more in terms of compatibility guarantees.

Lastly, as a general observation, we don't have a clear separation of responsibilities between readme and the docs. We should think about it.

yurishkuro: > Would it make sense to move this before: "Related Repositories"?

+1

> What are the use cases for README vs. docs?

Don't know, it varies widely between projects. For example, [k8s readme][1] is very short, doesn't even have such sections as license & contributing (probably expecting people to look into respective files).

> It seems the docs cater for operational use cases

Not necessarily, the Intro & Features pages in the docs largely repeat what's in the README. I think it would be good to move more stuff into docs and keep the README more development-focused, to minimize duplication of content. Another useful template is [Envoy][2], it's README is slightly longer than k8s's, but also avoids listing features and such.

But that doesn't actually answer the question where to put compatibility guarantees, I can still see both places as candidates.

[1]: https://github.com/kubernetes/kubernetes
[2]: https://github.com/envoyproxy/envoy

albertteoh: > But that doesn't actually answer the question where to put compatibility guarantees, I can still see both places as candidates.

I've added this topic for further discussion in the upcoming Jaeger community call to see if the group has a particular preference.

CONTRIBUTING.md
yurishkuro, line 176:
```suggestion
## Deprecating CLI Flags
```

CONTRIBUTING.md
```diff
@@ -172,3 +172,32 @@ Before merging a PR make sure:
...
+
+* If a flag is deprecated in release N, it can be removed in release N+2.
```
yurishkuro, line 178:
```suggestion
* If a flag is deprecated in release N, it can be removed in release N+2 or three months later, whichever is later.
```

CONTRIBUTING.md
```diff
@@ -172,3 +172,32 @@ Before merging a PR make sure:
...
+* When adding a (deprecated) prefix to the flags, indicate via a deprecation message that the flag could be removed after 3 months. For example:
```
yurishkuro, line 179:
```suggestion
* When adding a (deprecated) prefix to the flags, indicate via a deprecation message that the flag could be removed in the future. For example:
```

CONTRIBUTING.md
```diff
@@ -172,3 +172,32 @@ Before merging a PR make sure:
...
+  ```
+  deprecated, will be removed after 2020-03-15 or in release v1.19.0, whichever is later
```
yurishkuro, line 181:
```suggestion
  (deprecated, will be removed after 2020-03-15 or in release v1.19.0, whichever is later)
```

CONTRIBUTING.md
```diff
@@ -172,3 +172,32 @@ Before merging a PR make sure:
...
+* At the top of the file where the flag name is defined, add a constant and a comment, e.g.
+  ```
+  // TODO deprecated flag to be removed
+  healthCheckHTTPPortWarning = "(deprecated, will be removed after 2020-03-15 or in release v1.19.0, whichever is later)"
+  ```
+* Use that constant as the prefix to the help text, e.g.
+  ```
+  flagSet.Int(healthCheckHTTPPort, 0, healthCheckHTTPPortWarning+" see --"+adminHTTPHostPort)
+  ```
+* When reading the flag into config, log a warning with the same warning.
```
yurishkuro, line 192:
```suggestion
* When parsing a deprecated flag into config, log a warning with the same deprecation message
```

README.md
```diff
@@ -92,6 +92,23 @@ using Zipkin libraries, you do not have to rewrite all that code. Jaeger provide
 by accepting spans in Zipkin formats (Thrift or JSON v1/v2) over HTTP. Switching from Zipkin backend is just a matter
 of routing the traffic from Zipkin libraries to the Jaeger backend.
 
+## Version Compatibility Guarantees
+
+Occasionally, CLI flags can be deprecated due to, for example, usability improvements or new functionality.
+In such situations, developers introducing the deprecation are required to follow [these guidelines](./CONTRIBUTING.md#deprecated-cli-flags).
```
yurishkuro, line 98:
```suggestion
In such situations, developers introducing the deprecation are required to follow [these guidelines](./CONTRIBUTING.md#deprecating-cli-flags).
```

README.md
```diff
@@ -92,6 +92,23 @@ using Zipkin libraries, you do not have to rewrite all that code. Jaeger provide
...
+
+In short, for a deprecated CLI flag, you should expect to see the following message in the `--help` documentation:
+```
+(deprecated, will be removed after yyyy-mm-dd or in release vX.Y.Z, whichever is later)
+```
+
+A grace period of at least **3 months** or **two minor version bumps** from the first release
+containing the deprecation notice (whichever is later) will be provided before the deprecated CLI flag _can_ be deleted.
```
yurishkuro, line 106:
```suggestion
A grace period of at least **3 months** or **two minor version bumps** (whichever is later) from the first release
containing the deprecation notice will be provided before the deprecated CLI flag _can_ be deleted.
```

yurishkuro (approved):

PR #2911, opened 2021-03-29
Variable 'url' collides with imported package "net/url" name

wangchenglong01:
- Variable 'url' collides with imported package "net/url" name

- Variable 'url' collides with imported package "net/url" name


yurishkuro (commented):
It is not illegal in Go to mask symbols from enclosing scope. Let's not make superficial changes.

wangchenglong01:
> It is not illegal in Go to mask symbols from enclosing scope. Let's not make superficial changes.

ok

PR #2916, opened 2021-04-02
Check error conditions in crossdock tests

wangchenglong01:
- add err check

- add err check


wangchenglong01:
@joe-elliott
Hi, Do the added code need to add tests?

wangchenglong01:
@joe-elliott 
codecov/patch shows that the code was not covered by tests
image

But the code is covered by the path crossdock/services/agent_test TestGetSamplingRate, crossdock/services/query_test TestGetTraces methods.

How do I solve the problem?

albertteoh:
> But the code is covered by the path crossdock/services/agent_test TestGetSamplingRate, crossdock/services/query_test TestGetTraces methods.

If you try running a debugger over the test, you should find that the new lines of code are not covered by existing tests because they don't trigger an error on `ioutil.ReadAll`.

To cover this case, you can do something like the following:
```
// lyingHandler responds saying there's something in the response body but sends nothing in the body.
type lyingHandler struct{}

func (h *lyingHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
	w.Header().Set("Content-Length", "1")
}

func TestGetSamplingRateReadAllErr(t *testing.T) {
	handler := &lyingHandler{}
	server := httptest.NewServer(handler)
	defer server.Close()

	agent := NewAgentService(server.URL, zap.NewNop())
	_, err := agent.GetSamplingRate("svc", "op")
	assert.EqualError(t, err, "unexpected EOF")
}
```

wangchenglong01:
@albertteoh
I got it. Thanks very much, and I have added tests.

albertteoh (approved):
LGTM, thanks.

crossdock/services/agent_test.go
```diff
@@ -104,4 +104,13 @@ func TestGetSamplingRate(t *testing.T) {
 
 	_, err = agent.GetSamplingRate("bad_svc", "op")
 	assert.Error(t, err)
+
+	// Test ioutil.ReadAll error
```
jpkrohling, line 108: Wouldn't it be better to have this on its own test func?

crossdock/services/query_test.go
```diff
@@ -61,4 +61,13 @@ func TestGetTraces(t *testing.T) {
 
 	_, err = query.GetTraces("bad_svc", "op", map[string]string{"key": "value"})
 	assert.Error(t, err)
+
+	// Test ioutil.ReadAll error
```
jpkrohling, line 65: Same here: perhaps extract this into its own test func.

jpkrohling (approved):
LGTM, even though I don't see the practical effect of this change.

crossdock/services/query_test.go
wangchenglong01: good idea.
And I have changed them following @albertteoh 's example.

PR #2918, opened 2021-04-02
fix redundant type conversion

jinglina:
remove redundant type conversion for string


yurishkuro (commented):
Please sign commits

jinglina:
I just try to sign, but failed ,SAD

PR #2935, opened 2021-04-13
[es] Add index rollover mode that can choose day and hour

### LINKED ISSUE #2933: ES index rotation mode can choice with day and hour

As our Jaeger span count increasing, more and more large ES index size, the large size index will be caused more performance issues for the ES side. To avoid this issue, an hourly rotate index is the most convenient option when in systems with large amounts of data. 
I think can provide a parameter for the user choice, the default is "day" mode (like that current) user also can change to "hour" mode
let me create a PR to finish it

WalkerWang731:
- Resolves #2933

- Add a parameter `index-rollover-frequency`, default is day mode, will not affect existing usage
- Minimize changes to ensure compatibility with previous code 

### Parameter config of `collector` `ingester` `query`  (SPAN_STORAGE_TYPE=elasticsearch)
```
--es-archive.index-rollover-frequency string                     Rotates Jaeger indices over the given period. For example "day" creates "jaeger-span-yyyy-HH-dd" every day after UTC 12AM. Valid options: [hour, day]. Jaeger also support Elasticsearch ILM to manage indices, reference(https://www.jaegertracing.io/docs/deployment/#elasticsearch-ilm-support) (default "day")
--es.index-rollover-frequency string                             Rotates Jaeger indices over the given period. For example "day" creates "jaeger-span-yyyy-HH-dd" every day after UTC 12AM. Valid options: [hour, day]. Jaeger also support Elasticsearch ILM to manage indices, reference(https://www.jaegertracing.io/docs/deployment/#elasticsearch-ilm-support) (default "day")
```

### DIFF
--- a/pkg/es/config/config.go
+++ b/pkg/es/config/config.go
@@ -63,7 +63,7 @@ type Configuration struct {
 	BulkFlushInterval     time.Duration  `mapstructure:"-"`
 	IndexPrefix           string         `mapstructure:"index_prefix"`
 	IndexDateLayout       string         `mapstructure:"index_date_layout"`
-	IndexRotate           string         `mapstructure:"-"`
+	IndexRollover         string         `mapstructure:"-"`
 	Tags                  TagsAsFields   `mapstructure:"tags_as_fields"`
 	Enabled               bool           `mapstructure:"-"`
 	TLS                   tlscfg.Options `mapstructure:"tls"`
--- a/pkg/es/config/config.go
+++ b/pkg/es/config/config.go
@@ -44,34 +44,34 @@ import (
 
 // Configuration describes the configuration properties needed to connect to an ElasticSearch cluster
 type Configuration struct {
-	Servers               []string       `mapstructure:"server_urls"`
-	RemoteReadClusters    []string       `mapstructure:"remote_read_clusters"`
-	Username              string         `mapstructure:"username"`
-	Password              string         `mapstructure:"password" json:"-"`
-	TokenFilePath         string         `mapstructure:"token_file"`
-	AllowTokenFromContext bool           `mapstructure:"-"`
-	Sniffer               bool           `mapstructure:"sniffer"` // https://github.com/olivere/elastic/wiki/Sniffing
-	SnifferTLSEnabled     bool           `mapstructure:"sniffer_tls_enabled"`
-	MaxDocCount           int            `mapstructure:"-"`                     // Defines maximum number of results to fetch from storage per query
-	MaxSpanAge            time.Duration  `yaml:"max_span_age" mapstructure:"-"` // configures the maximum lookback on span reads
-	NumShards             int64          `yaml:"shards" mapstructure:"num_shards"`
-	NumReplicas           int64          `yaml:"replicas" mapstructure:"num_replicas"`
-	Timeout               time.Duration  `validate:"min=500" mapstructure:"-"`
-	BulkSize              int            `mapstructure:"-"`
-	BulkWorkers           int            `mapstructure:"-"`
-	BulkActions           int            `mapstructure:"-"`
-	BulkFlushInterval     time.Duration  `mapstructure:"-"`
-	IndexPrefix           string         `mapstructure:"index_prefix"`
-	IndexDateLayout       string         `mapstructure:"index_date_layout"`
-	IndexRollover         string         `mapstructure:"-"`
-	Tags                  TagsAsFields   `mapstructure:"tags_as_fields"`
-	Enabled               bool           `mapstructure:"-"`
-	TLS                   tlscfg.Options `mapstructure:"tls"`
-	UseReadWriteAliases   bool           `mapstructure:"use_aliases"`
-	CreateIndexTemplates  bool           `mapstructure:"create_mappings"`
-	UseILM                bool           `mapstructure:"use_ilm"`
-	Version               uint           `mapstructure:"version"`
-	LogLevel              string         `mapstructure:"log_level"`
+	Servers                []string       `mapstructure:"server_urls"`
+	RemoteReadClusters     []string       `mapstructure:"remote_read_clusters"`
+	Username               string         `mapstructure:"username"`
+	Password               string         `mapstructure:"password" json:"-"`
+	TokenFilePath          string         `mapstructure:"token_file"`
+	AllowTokenFromContext  bool           `mapstructure:"-"`
+	Sniffer                bool           `mapstructure:"sniffer"` // https://github.com/olivere/elastic/wiki/Sniffing
+	SnifferTLSEnabled      bool           `mapstructure:"sniffer_tls_enabled"`
+	MaxDocCount            int            `mapstructure:"-"`                     // Defines maximum number of results to fetch from storage per query
+	MaxSpanAge             time.Duration  `yaml:"max_span_age" mapstructure:"-"` // configures the maximum lookback on span reads
+	NumShards              int64          `yaml:"shards" mapstructure:"num_shards"`
+	NumReplicas            int64          `yaml:"replicas" mapstructure:"num_replicas"`
+	Timeout                time.Duration  `validate:"min=500" mapstructure:"-"`
+	BulkSize               int            `mapstructure:"-"`
+	BulkWorkers            int            `mapstructure:"-"`
+	BulkActions            int            `mapstructure:"-"`
+	BulkFlushInterval      time.Duration  `mapstructure:"-"`
+	IndexPrefix            string         `mapstructure:"index_prefix"`
+	IndexDateLayout        string         `mapstructure:"index_date_layout"`
+	IndexRolloverFrequency string         `mapstructure:"-"`
+	Tags                   TagsAsFields   `mapstructure:"tags_as_fields"`
+	Enabled                bool           `mapstructure:"-"`
+	TLS                    tlscfg.Options `mapstructure:"tls"`
+	UseReadWriteAliases    bool           `mapstructure:"use_aliases"`
+	CreateIndexTemplates   bool           `mapstructure:"create_mappings"`
+	UseILM                 bool           `mapstructure:"use_ilm"`
+	Version                uint           `mapstructure:"version"`
+	LogLevel               string         `mapstructure:"log_level"`
 }
 
 // TagsAsFields holds configuration for tag schema.
--- a/pkg/es/config/config.go
+++ b/pkg/es/config/config.go
@@ -63,6 +63,7 @@ type Configuration struct {
 	BulkFlushInterval     time.Duration  `mapstructure:"-"`
 	IndexPrefix           string         `mapstructure:"index_prefix"`
 	IndexDateLayout       string         `mapstructure:"index_date_layout"`
+	IndexRotate           string         `mapstructure:"-"`
 	Tags                  TagsAsFields   `mapstructure:"tags_as_fields"`
 	Enabled               bool           `mapstructure:"-"`
 	TLS                   tlscfg.Options `mapstructure:"tls"`
--- a/plugin/storage/es/options.go
+++ b/plugin/storage/es/options.go
@@ -46,7 +46,7 @@ const (
 	suffixTimeout             = ".timeout"
 	suffixIndexPrefix         = ".index-prefix"
 	suffixIndexDateSeparator  = ".index-date-separator"
-	suffixIndexRotate         = ".index-rotate"
+	suffixIndexRollover       = ".index-rollover"
 	suffixTagsAsFields        = ".tags-as-fields"
 	suffixTagsAsFieldsAll     = suffixTagsAsFields + ".all"
 	suffixTagsAsFieldsInclude = suffixTagsAsFields + ".include"
@@ -67,7 +67,7 @@ const (
 	// default separator for Elasticsearch index date layout.
 	defaultIndexDateSeparator = "-"
 
-	defaultIndexRotate = "day"
+	defaultIndexRollover = "day"
 )
 
 // TODO this should be moved next to config.Configuration struct (maybe ./flags package)
@@ -210,9 +210,10 @@ func addFlags(flagSet *flag.FlagSet, nsConfig *namespaceConfig) {
 		defaultIndexDateSeparator,
 		"Optional date separator of Jaeger indices. For example \".\" creates \"jaeger-span-2020.11.20\".")
 	flagSet.String(
-		nsConfig.namespace+suffixIndexRotate,
-		defaultIndexRotate,
-		"Optional rotation opportunity of Jaeger indices. For example \"day\" creates \"jaeger-span-yyyy-HH-dd\" every day after UTC 12AM. Valid opportunity: [hour, day]")
+		nsConfig.namespace+suffixIndexRollover,
+		defaultIndexRollover,
+		"Rotates Jaeger indices over the given period. For example \"day\" creates \"jaeger-span-yyyy-HH-dd\" every day after UTC 12AM. Valid options: [hour, day]. "+
+			"Jaeger also support Elasticsearch ILM to manage indices, reference(https://www.jaegertracing.io/docs/deployment/#elasticsearch-ilm-support)")
 	flagSet.Bool(
 		nsConfig.namespace+suffixTagsAsFieldsAll,
 		nsConfig.Tags.AllAsFields,
@@ -302,6 +303,7 @@ func initFromViper(cfg *namespaceConfig, v *viper.Viper) {
 	cfg.BulkFlushInterval = v.GetDuration(cfg.namespace + suffixBulkFlushInterval)
 	cfg.Timeout = v.GetDuration(cfg.namespace + suffixTimeout)
 	cfg.IndexPrefix = v.GetString(cfg.namespace + suffixIndexPrefix)
+	cfg.IndexDateLayout = initDateLayout(strings.ToLower(v.GetString(cfg.namespace+suffixIndexRollover)), v.GetString(cfg.namespace+suffixIndexDateSeparator))
 	cfg.Tags.AllAsFields = v.GetBool(cfg.namespace + suffixTagsAsFieldsAll)
 	cfg.Tags.Include = v.GetString(cfg.namespace + suffixTagsAsFieldsInclude)
 	cfg.Tags.File = v.GetString(cfg.namespace + suffixTagsFile)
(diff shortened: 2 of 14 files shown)


albertteoh:
Sorry for the slow response, I can take a look at this tomorrow.

WalkerWang731:
> Sorry for the slow response, I can take a look at this tomorrow.

Actually, just only `jaeger-span`  index need rotation hourly, `jaeger-service` seems do not need rotation, because in our environments, `jaeger-span` can reach more than 100TB, but `jaeger-service`  just only 500MB around.(maybe just occur our scenario)
I have not thought of good that how to just rotate `jaeger-span` and look for the most graceful solution
so just commit this version code, if we want separate `jaeger-span` and `jaeger-service` rotation logic, we have to change since the factory function.

albertteoh:
Thanks @WalkerWang731 , have you considered the [esRollover functionality available in Jaeger](https://www.jaegertracing.io/docs/1.22/deployment/#elasticsearch-rollover)?

There is also support for [Elasticsearch's ILM functionality](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-lifecycle-management.html) that is a bit more sophisticated, providing automated rollover without the need for a cron job to execute the rollover as is required with `esRollover`: https://www.jaegertracing.io/docs/1.22/deployment/#elasticsearch-ilm-support.

I believe both approaches support rollover by time units (days, hours, minutes), number of documents or by size.

WalkerWang731:
> Thanks @WalkerWang731 , have you considered the [esRollover functionality available in Jaeger](https://www.jaegertracing.io/docs/1.22/deployment/#elasticsearch-rollover)?
>
> There is also support for [Elasticsearch's ILM functionality](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-lifecycle-management.html) that is a bit more sophisticated, providing automated rollover without the need for a cron job to execute the rollover as is required with `esRollover`: https://www.jaegertracing.io/docs/1.22/deployment/#elasticsearch-ilm-support.
>
> I believe both approaches support rollover by time units (days, hours, minutes), number of documents or by size.

Thanks @albertteoh , For the [Elasticsearch's ILM functionality](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-lifecycle-management.html) and [esRollover functionality available in Jaeger](https://www.jaegertracing.io/docs/1.22/deployment/#elasticsearch-rollover) they looks good. 
Currently, these 3 methods each have advantages and disadvantages:

| method                                                       | advantages                                    | disadvantages                                                |
| ------------------------------------------------------------ | --------------------------------------------- | ------------------------------------------------------------ |
| [esRollover functionality available in Jaeger](https://www.jaegertracing.io/docs/1.22/deployment/#elasticsearch-rollover) | support `age` `docs` `size`                   | need deploy cron job                                         |
| [Elasticsearch's ILM functionality](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-lifecycle-management.html) | support `age` `docs` `size` and automatically | just only support 7 or later                                 |
| This commit                                                  | don't need other operation                    | just support `age` of `day` and `hour`, the service index also will be rotated |

So, although the solution of this commit is not the most perfect, but the operation of the rotation is also simplified as much as possible to provide users with a simpler way. And maybe we can add the other two ways in the help text. 

Actually, this commit part will not have any impact on the code structure, which is a minor change,

But I have to say I'm really not sure whether need to merge, need your review and assessment, I just want to give a simple way to rotate the index.

Thanks, above are my suggestions, for your information.

albertteoh:
How do you plan to delete old indices? It seems having too many indices can impact shard performance.

WalkerWang731:
> How do you plan to delete old indices? It seems having too many indices can impact shard performance.

Yes, more shards will impact ES performance, delete old indices can use `Elasticsearch's ILM` and have not limited the version. For example, we use ES 6 version. Because just only set `Cold phase` and `Delete phase`, so it don't impact jaeger write and query

pkg/es/config/config.go
```diff
@@ -63,6 +63,7 @@ type Configuration struct {
 	BulkFlushInterval     time.Duration  `mapstructure:"-"`
 	IndexPrefix           string         `mapstructure:"index_prefix"`
 	IndexDateLayout       string         `mapstructure:"index_date_layout"`
+	IndexRotate           string         `mapstructure:"-"`
```
albertteoh, line 66: ES uses the term `Rollover`, I think we should stay consistent with terminology.

plugin/storage/es/spanstore/reader.go
```diff
@@ -200,8 +200,12 @@ func timeRangeIndices(indexName, indexDateLayout string, startTime time.Time, en
 	firstIndex := indexWithDate(indexName, indexDateLayout, startTime)
 	currentIndex := indexWithDate(indexName, indexDateLayout, endTime)
 	for currentIndex != firstIndex {
-		indices = append(indices, currentIndex)
-		endTime = endTime.Add(-24 * time.Hour)
+		if len(indices) == 0 {
```
albertteoh, line 203: Could be simplified:
```
if len(indices) == 0 || indices[len(indices)-1] != currentIndex {
}
```

albertteoh (commented):
@bhiravabhatla, please remind us why Jaeger's ILM support was limited to ES 7 and does not supported for [ES 6's ILM feature](https://www.elastic.co/guide/en/elasticsearch/reference/6.8/_setting_up_a_new_policy.html)?

I would prefer users adopt a single index management solution for rollover and cleanup (either via scripts+cronjob or ILM) rather than require setting a flag + ILM policy for cleanup.

However, I appreciate using esRollover/Cleaner scripts needs a bit more effort on setup and maintenance, and ILM is limited to version 7. 

If we determine that ILM isn't feasible in ES 6, this proposal is okay with me so long as we document usage and the recommended alternatives clearly both here and in the official documentation.

bhiravabhatla:
> @bhiravabhatla, please remind us why Jaeger's ILM support was limited to ES 7 and does not supported for [ES 6's ILM feature](https://www.elastic.co/guide/en/elasticsearch/reference/6.8/_setting_up_a_new_policy.html)?
> 
> I would prefer users adopt a single index management solution for rollover and cleanup (either via scripts+cronjob or ILM) rather than require setting a flag + ILM policy for cleanup.
> 
> However, I appreciate using esRollover/Cleaner scripts needs a bit more effort on setup and maintenance, and ILM is limited to version 7.
> 
> If we determine that ILM isn't feasible in ES 6, this proposal is okay with me so long as we document usage and the recommended alternatives clearly both here and in the official documentation.
@albertteoh 
As I understand, ILM was not OSS in 6.x - correct me if I am wrong, was one of the reasons - https://github.com/jaegertracing/jaeger/pull/2935#pullrequestreview-639465402. 
Not entirely sure about this.

Another reason is is_write_index is only supported for ES Versions > 6.4. We use is_write_index for ILM implementation.

WalkerWang731:
Hi, @albertteoh
 I have adjusted it, please review continue
 - adjust terminology from rotate to rollover, include option parameters
 - add reference into `.index-rollover` help
 - simplified code

plugin/storage/es/options.go
```diff
@@ -205,7 +208,12 @@ func addFlags(flagSet *flag.FlagSet, nsConfig *namespaceConfig) {
 	flagSet.String(
 		nsConfig.namespace+suffixIndexDateSeparator,
 		defaultIndexDateSeparator,
-		"Optional date separator of Jaeger indices. For example \".\" creates \"jaeger-span-2020.11.20 \".")
+		"Optional date separator of Jaeger indices. For example \".\" creates \"jaeger-span-2020.11.20\".")
+	flagSet.String(
+		nsConfig.namespace+suffixIndexRollover,
+		defaultIndexRollover,
+		"Rotates Jaeger indices over the given period. For example \"day\" creates \"jaeger-span-yyyy-HH-dd\" every day after UTC 12AM. Valid options: [hour, day]. "+
```
albertteoh, line 215: `Rollover Jaeger indices for the given period.`

`s/jaeger-span-yyyy-HH-dd/jaeger-span-yyyy-MM-dd/`

plugin/storage/es/options.go
```diff
@@ -205,7 +208,12 @@ func addFlags(flagSet *flag.FlagSet, nsConfig *namespaceConfig) {
...
+			"Jaeger also support Elasticsearch ILM to manage indices, reference(https://www.jaegertracing.io/docs/deployment/#elasticsearch-ilm-support)")
```
albertteoh, line 216: `Jaeger additionally supports manual and automated (via ILM) index management. Reference:  https://www.jaegertracing.io/docs/deployment/#elasticsearch-rollover.`

plugin/storage/es/options.go
```diff
@@ -295,7 +303,7 @@ func initFromViper(cfg *namespaceConfig, v *viper.Viper) {
 	cfg.BulkFlushInterval = v.GetDuration(cfg.namespace + suffixBulkFlushInterval)
 	cfg.Timeout = v.GetDuration(cfg.namespace + suffixTimeout)
 	cfg.IndexPrefix = v.GetString(cfg.namespace + suffixIndexPrefix)
-	cfg.IndexDateLayout = initDateLayout(v.GetString(cfg.namespace + suffixIndexDateSeparator))
+	cfg.IndexDateLayout = initDateLayout(strings.ToLower(v.GetString(cfg.namespace+suffixIndexRollover)), v.GetString(cfg.namespace+suffixIndexDateSeparator))
```
albertteoh, line 306: I think it's a bit more readable to separate these into two separate variables as it starts to get a bit unwieldy with the additional string getter and manipulation functions:
```
rolloverBy := strings.ToLower(v.GetString(cfg.namespace + suffixIndexRollover))
separator := v.GetString(cfg.namespace + suffixIndexDateSeparator)
cfg.IndexDateLayout = initDateLayout(rolloverBy, separator)
```

plugin/storage/es/options.go
```diff
@@ -343,6 +351,12 @@ func stripWhiteSpace(str string) string {
 	return strings.Replace(str, " ", "", -1)
 }
 
-func initDateLayout(separator string) string {
-	return fmt.Sprintf("2006%s01%s02", separator, separator)
+func initDateLayout(RolloverBy, separator string) string {
```
albertteoh, line 354: `s/RolloverBy/rolloverBy/`

plugin/storage/es/options.go
```diff
@@ -343,6 +351,12 @@ func stripWhiteSpace(str string) string {
...
+	seps := []interface{}{separator, separator}
+	indexLayout := "2006%s01%s02"
```
albertteoh, line 356: nit: I would add a comment just to make it a bit clearer:
```
// Default "day" index layout.
indexLayout := "2006%s01%s02"
```

albertteoh:
Thanks @WalkerWang731, just a few minor comments.

By the way, it's good practice to create PRs off a branch in your fork, instead of `master`, and keeping your fork's `master` clean and in sync with remote `master`. I recall having some problems in the past when submitting PRs directly from master, but can't recall what they were. :P

Anybody understand what causes the following build errors?
```
+ docker login docker.io --username jaegertracingbot --password-stdin
```

WalkerWang731:
> Thanks @WalkerWang731, just a few minor comments.
> 
> By the way, it's good practice to create PRs off a branch in your fork, instead of `master`, and keeping your fork's `master` clean and in sync with remote `master`. I recall having some problems in the past when submitting PRs directly from master, but can't recall what they were. :P
> 
> Anybody understand what causes the following build errors?
> 
> ```
> + docker login docker.io --username jaegertracingbot --password-stdin
> Error: Cannot perform an interactive login from a non TTY device
> Error: Process completed with exit code 1.
> ```

Hi @albertteoh, 
Thanks for your suggestions, I have updated follow your suggestions.
About `Error: Cannot perform an interactive login from a non TTY device` issue, maybe need @Ashmita152 help to take a look? I'm not sure too

albertteoh (approved):
lgtm, thanks @WalkerWang731.

pkg/es/config/config.go
```diff
@@ -63,6 +63,7 @@ type Configuration struct {
 	BulkFlushInterval     time.Duration  `mapstructure:"-"`
 	IndexPrefix           string         `mapstructure:"index_prefix"`
 	IndexDateLayout       string         `mapstructure:"index_date_layout"`
+	IndexRollover         string         `mapstructure:"-"`
```
yurishkuro, line 66: This is a poor name, what does "index rollover" mean? From the flag description it seems to mean "frequency", so let's name it IndexRolloverFrequency (and all other vars/flags respectively)

plugin/storage/es/options.go
```diff
@@ -205,7 +208,12 @@ func addFlags(flagSet *flag.FlagSet, nsConfig *namespaceConfig) {
 	flagSet.String(
 		nsConfig.namespace+suffixIndexDateSeparator,
 		defaultIndexDateSeparator,
-		"Optional date separator of Jaeger indices. For example \".\" creates \"jaeger-span-2020.11.20 \".")
+		"Optional date separator of Jaeger indices. For example \".\" creates \"jaeger-span-2020.11.20\".")
+	flagSet.String(
+		nsConfig.namespace+suffixIndexRollover,
+		defaultIndexRollover,
+		"Rotates Jaeger indices over the given period. For example \"day\" creates \"jaeger-span-yyyy-MM-dd\" every day after UTC 12AM. Valid options: [hour, day]. "+
+			"Jaeger additionally supports manual and automated (via ILM) index management. Reference: https://www.jaegertracing.io/docs/deployment/#elasticsearch-rollover.")
```
yurishkuro, line 216: > Jaeger additionally supports manual and automated (via ILM) index management.

This is confusing. Doesn't this day/hour enhancement classifies as "manual"? If not, what other "manual" rollover are we referring to?

plugin/storage/es/spanstore/reader.go
```diff
@@ -200,8 +200,10 @@ func timeRangeIndices(indexName, indexDateLayout string, startTime time.Time, en
 	firstIndex := indexWithDate(indexName, indexDateLayout, startTime)
 	currentIndex := indexWithDate(indexName, indexDateLayout, endTime)
 	for currentIndex != firstIndex {
-		indices = append(indices, currentIndex)
-		endTime = endTime.Add(-24 * time.Hour)
+		if len(indices) == 0 || indices[len(indices)-1] != currentIndex {
+			indices = append(indices, currentIndex)
+		}
+		endTime = endTime.Add(-1 * time.Hour)
```
yurishkuro, line 206: So this will do 24 iterations on every invocation, even though the default is daily rotation? Could we make this more efficient?

plugin/storage/es/options.go
```diff
@@ -205,7 +208,12 @@ func addFlags(flagSet *flag.FlagSet, nsConfig *namespaceConfig) {
 	flagSet.String(
 		nsConfig.namespace+suffixIndexDateSeparator,
 		defaultIndexDateSeparator,
-		"Optional date separator of Jaeger indices. For example \".\" creates \"jaeger-span-2020.11.20 \".")
+		"Optional date separator of Jaeger indices. For example \".\" creates \"jaeger-span-2020.11.20\".")
+	flagSet.String(
+		nsConfig.namespace+suffixIndexRollover,
+		defaultIndexRollover,
+		"Rotates Jaeger indices over the given period. For example \"day\" creates \"jaeger-span-yyyy-MM-dd\" every day after UTC 12AM. Valid options: [hour, day]. "+
```
yurishkuro, line 215: I suggest adding more detailed explanation to the website (for next release) and linking there.

plugin/storage/es/options.go
```diff
@@ -343,6 +354,15 @@ func stripWhiteSpace(str string) string {
...
+	seps := []interface{}{separator, separator}
+
+	// Default "day" index layout.
+	indexLayout := "2006%s01%s02"
+
+	if rolloverBy == "hour" {
+		indexLayout = "2006%s01%s02%s15"
+		seps = append(seps, separator)
+	}
+	return fmt.Sprintf(indexLayout, seps...)
```
yurishkuro, line 367:
```suggestion
func initDateLayout(rolloverFreq, sep string) string {
    // default to daily format
    indexLayout := "2006" + sep + "01" + sep + "02"
    if rolloverFreq == "hour" {
        indexLayout = indexLayout + sep + "15"
    }
    return indexLayout
}
```

plugin/storage/es/options.go
albertteoh: +1

plugin/storage/es/options.go
albertteoh: "manual" here refers to the esRollover/Cleaner python scripts that are used to manually rollover/delete an index (automated via cronjobs).

This day/hour enhancement is a simple and convenient way to rollover an index by duration **only** and lacks the ability to:
- Rollover on other dimensions such as by size or number of ES documents.
- Clean older indices.

Both the esRollover/Cleaner python scripts and ILM (supported in Jaeger) provide the above, which is why I suggested the term "index management" here because I consider these as complete index management solutions and want users to be aware of these more comprehensive but "higher-effort" alternatives.

Agree this is confusing, perhaps we could rephrase to: "This does not delete old indices. For details on complete index management solutions supported by Jaeger, refer to: https://www.jaegertracing.io/docs/deployment/#elasticsearch-rollover".

Open to suggestions.

plugin/storage/es/spanstore/reader.go
WalkerWang731: ok let me take a look

albertteoh:
> I also want to limit jaeger-service indices to daily rollover

My two cents... I think it's better to maintain index format consistency between spans and services to avoid surprises and more work to manage these indices. I would also think that service index sizes would be fairly small and so not have as much an impact on ES performance?

Either that or introduce `--index-rollover-span-frequency` and `--index-rollover-service-frequency`.

WalkerWang731:
> > I also want to limit jaeger-service indices to daily rollover
> 
> My two cents... I think it's better to maintain index format consistency between spans and services to avoid surprises and more work to manage these indices. I would also think that service index sizes would be fairly small and so not have as much an impact on ES performance?
> 
> Either that or introduce `--index-rollover-span-frequency` and `--index-rollover-service-frequency`.

Hi @albertteoh 
From our experience, one node less than 600 shards is safe, not impact ES performance, 600 shards are the total number, include replica shards, and also depends on how long will keep it. 

The formula: (SHARD_COUNT * INDEX_COUNT * (REPLICA_COUNT+ 1) ) * KEEP_DAY / NODE_COUNT
- KEEP_DAY is the number of indices that actually exist during that time, daily is one day * 1, hourly is one day * 24
- If the `jaeger-service` index also is hourly rollover then will expand shard count.
- Actually, just keep shard count the total as same previous. For example, the previous index shards count is 600(daily), so the current index shards count is 25(hourly, 600 / 24) will be ok.

I also agree that keep the index format consistency between spans and services, but provide like as `--index-rollover-span-frequency` and `--index-rollover-service-frequency`  will be complex, we must change the factory func and more. I always thought this was a small commit and didn't want to impact too much code.

So, according to the above, limiting `jaeger-service` shards just for the more optimization. We also keep the current code, and temporarily without regard for the `jaeger-service` shards, how about? 

Do you have any better suggestions, please let me know.

Thank you!

albertteoh:
Hi @WalkerWang731, sorry I don't completely follow; are you saying it should be okay to leave this PR as is, or is it quite critical to ensure the service index must always be on a daily rollover?

Please also make sure:
- all your commits are signed.
- test coverage is adequate to pass the build.

WalkerWang731:
> Hi @WalkerWang731, sorry I don't completely follow; are you saying it should be okay to leave this PR as is, or is it quite critical to ensure the service index must always be on a daily rollover?
> 
Hi @albertteoh 
Ah sorry, I'm not to say clearly, I think it's okay for the current PR. service index rollover of daily is not a must and urgent

> Please also make sure:
> 
> * all your commits are signed.
> * test coverage is adequate to pass the build.

Yeah, I think it's should be ok now, unit test has passed, except docker relevant issues. I'm not sure whether need @Ashmita152 help to take a look?
```
+ '[' '!' -f .docker.io.login ']'
+ '[' -v DOCKERHUB_TOKEN ']'
+ printenv DOCKERHUB_TOKEN
+ docker login docker.io --username jaegertracingbot --password-stdin
```

WalkerWang731:
close this PR because [can't fix the signoff issue ](https://github.com/jaegertracing/jaeger/pull/2935/checks?check_run_id=2445917988) and let me recreate a new PR.

PR #2948, opened 2021-04-21
Add s390x support on multiarch docker images

kun-lu20:
This change is to add s390x support on multiarch Jaeger docker images

Signed-off-by: Kun-Lu <kun.lu@ibm.com>

- Problem described in #2292 

- modifies CI workflows (`ci-all-in-one-build.yml`, `ci-docker-build.yml`, `ci-release.yml`) to build and publish multi-arch docker images on `docker.io` and `quay.io`.
- adds `s390x` support on multi-arch docker images of `baseimg`, `all-in-one` and all the Jaeger components, except debug images and `Jaeger-Cassandra-Schema` whose base image doesn’t have `s390x` support yet.
- uses local registry service to hold: 
    1. multi-arch `baseimg`,
    2. multi-arch `all-in-one` image for local integration test.
- updates `Makefile` and `scripts/build-all-in-one-image.sh`, adds `scripts/build-upload-docker-images.sh` to use `docker buildx` to build and publish multi-arch docker images.
- removes `scripts/upload-all-docker-images.sh` since its uploading function has been implemented in `scripts/build-upload-docker-images.sh`. 
- removes default value of `ARG TARGETARCH` in Dockerfiles.
- all the building, integration testing, tag computation and publishing procedures are the same as existing ones (including all the multi-arch docker images and docker images which currently don’t support `s390x`).


Ashmita152 (commented):
Thank you for working on it.

scripts/build-all-in-one-image.sh
```diff
@@ -45,17 +45,108 @@ upload_to_docker() {
...
+  QUAY_TOKEN=${QUAY_TOKEN:-}
+
+  ###############Compute the tag
+  BASE_BUILD_IMAGE=$1
+
+  ## if we are on a release tag, let's extract the version number
+  ## the other possible value, currently, is 'master' (or another branch name)
+  if [[ $BRANCH == v* ]]; then
+      COMPONENT_VERSION=$(echo ${BRANCH} | grep -Po "([\d\.]+)")
+      MAJOR_MINOR=$(echo ${COMPONENT_VERSION} | awk -F. '{print $1"."$2}')
```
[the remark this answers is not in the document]
kun-lu20: The `COMPONENT_VERSION` variable in `L71` is actually for patch components, like `1.22.0`. I've changed the name of this variable to `MAJOR_MINOR_PATCH` to avoid ambiguity.

scripts/build-all-in-one-image.sh
```diff
@@ -45,17 +45,108 @@ upload_to_docker() {
...
+      MAJOR1=$(echo ${COMPONENT_VERSION} | awk -F. '{print $1}')
```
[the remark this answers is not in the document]
kun-lu20: Yes, I've updated the code here.

scripts/build-all-in-one-image.sh
```diff
@@ -45,17 +45,108 @@ upload_to_docker() {
...
+
+  if [ "${MAJOR1}x" != "x" ]; then
+      MAJOR1_IMAGE="${BASE_BUILD_IMAGE}:${MAJOR1}"
+      IMAGE_TAGS="${IMAGE_TAGS} --tag docker.io/${MAJOR1_IMAGE}"
+  fi
+
+  IMAGE_TAGS="${IMAGE_TAGS} --tag docker.io/${SNAPSHOT_TAG}"
+
+  ## for quay.io
+  IMAGE_TAGS="${IMAGE_TAGS} --tag quay.io/${BASE_BUILD_IMAGE} --tag quay.io/${BUILD_IMAGE}"
```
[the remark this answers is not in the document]
kun-lu20: Yes, we can. I've updated the script to combine them.

Makefile
```diff
@@ -72,6 +72,17 @@ MOCKERY=mockery
 
 .DEFAULT_GOAL := test-and-lint
 
+BASE_IMAGE_MULTIARCH := localhost:5000/baseimg:$(VERSION)-$(shell echo $(ROOT_IMAGE) | tr : -)
+PLATFORMS=linux/amd64,linux/s390x
+INGESTER_TAG=$(subst JAGERCOMP,ingester,$(IMAGE_TAGS))
```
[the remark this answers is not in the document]
kun-lu20: Yes, I think we need to, since each jaeger component has its unique name, hence the unique tag value.

`JAGERCOMP` is not a variable but a const string. We use it in the tag computation procedure in `scripts/build-upload-docker-images.sh` to get the value of variable `IMAGE_TAGS`. Then in `Makefile`  `L77-L84`, we replace it with the name of each jaeger component in variable `IMAGE_TAGS` to get each actual tag value.

cmd/all-in-one/Dockerfile
```diff
@@ -28,6 +28,8 @@ EXPOSE 16686
 COPY all-in-one-linux-$TARGETARCH /go/bin/all-in-one-linux
 COPY sampling_strategies.json /etc/jaeger/
 
+RUN chmod +x /go/bin/all-in-one-linux
```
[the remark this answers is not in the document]
kun-lu20: Yes, you are right, the binary is already executable. I've removed this line from Dockerfile.

.github/workflows/ci-all-in-one-build.yml
```diff
@@ -6,9 +6,14 @@ on:
...
 
-jobs:
+jobs:   
   all-in-one:
     runs-on: ubuntu-latest
+    services:
+      registry:
+        image: registry:2
+        ports:
+          - 5000:5000
```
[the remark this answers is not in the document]
kun-lu20: Yes, we still need to publish the ports, otherwise we couldn't access the local registry service (localhost:5000) to hold multi-arch `baseimg`, and push and pull multi-arch `all-in-one` image while running local integration test.

kun-lu20:
@Ashmita152 Thank you very much for your review comments! I've responded to them and made corresponding code changes. Please take a look when you are available.

scripts/build-all-in-one-image.sh
```diff
@@ -45,17 +45,96 @@ upload_to_docker() {
   fi
 }
 
-make build-all-in-one GOOS=linux GOARCH=$GOARCH
+upload_multiarch_to_docker() {
```
yurishkuro, line 48: Why is there so much code just for all-in-one? I assume it would be the same for other Docker images, so let's keep a single copy.

.github/workflows/ci-all-in-one-build.yml
```diff
@@ -6,9 +6,14 @@ on:
   pull_request:
     branches: [ master ]
 
-jobs:
+jobs:
```
yurishkuro, line 9: remove whitespace

.github/workflows/ci-docker-build.yml
```diff
@@ -32,11 +39,14 @@ jobs:
     - name: Install tools
       run: make install-ci
 
-    - name: Build docker images
-      run: make docker
```
yurishkuro, line 36: what happens to `make docker` target now?

Makefile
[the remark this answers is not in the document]
yurishkuro: I am not a fan of these cross-dependences. If the string that contains `JAGERCOMP` is defined in the script, why can't the script resolve it as needed, without pushing the substitution into Makefile?

Also, Jaeger is spelled with 'e'.

Makefile
```diff
@@ -348,6 +361,73 @@ docker-images-only: docker-images-cassandra \
 	docker-images-tracegen \
 	docker-images-anonymizer
 
+.PHONY: multiarch-docker
```
yurishkuro, line 364: Couldn't we just move all this code into the script? Why do they need to be in the Makefile? I would restrict Makefile to building the binaries only, and then have all docker packaging etc. be done in the script.

Makefile
```diff
@@ -348,6 +361,73 @@ docker-images-only: docker-images-cassandra \
...
+.PHONY: docker-images-anonymizer-multiarch
+docker-images-anonymizer-multiarch:
+	docker buildx build --output "$(PUSHTAG)" \
+    	--progress=plain \
+    	--platform=$(PLATFORMS) \
+    	$(ANONYMIZER_TAG) \
+		cmd/anonymizer/
+	@echo "Finished building multiarch jaeger-anonymizer =============="
+
 .PHONY: docker-push
```
yurishkuro, line 431: we should delete this target, it's not used from GH Actions, and it probably incorrect given these changes.

Makefile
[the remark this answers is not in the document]
kun-lu20: Got it. I've corrected the spelling and moved the related code to the script.

scripts/build-all-in-one-image.sh
kun-lu20: Yes, you're right. I've extracted the duplicate code from `scripts/build-all-in-one-image.sh` and `scripts/build-upload-docker-images.sh` and put them in newly created scripts.

.github/workflows/ci-all-in-one-build.yml
kun-lu20: Whitespace removed.

.github/workflows/ci-docker-build.yml
kun-lu20: This target is unused now, I've removed it from Makefile.

Makefile
kun-lu20: Yes, we could. I've moved all code related to building and uploading multi-arch docker images from Makefile to the scripts.

Makefile
kun-lu20: OK, I've deleted this target.

kun-lu20:
Hi @yurishkuro , thank you so much for your review! I've made changes as per your comments, please take a look when you are available. Thanks again!

Ashmita152 (commented):
It's looking great. One minor feedback: Do we need to append `for-multiarch-image` in the script names.

scripts/build-upload-docker-images.sh
```diff
@@ -0,0 +1,93 @@
+#!/bin/bash
+
+set -euxf -o pipefail
+
+build_upload_multiarch_images(){
+  for component in agent collector query ingester
+  do
+		docker buildx build --output "${PUSHTAG}" \
```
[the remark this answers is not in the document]
kun-lu20: Thanks for reminding me, I've fixed the indentation issues.

scripts/build-upload-docker-images.sh
```diff
@@ -0,0 +1,93 @@
...
+    	--progress=plain --target release \
+    	--build-arg base_image="localhost:5000/baseimg:1.0.0-alpine-3.12" \
+			--build-arg debug_image="golang:1.15-alpine" \
```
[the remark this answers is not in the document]
kun-lu20: Yes, you're right, we've noticed this. But building `debugimg:1.0.0-golang-1.15-alpine` needs **`go-delve`** which does not have **`s390x`** support yet. So **at this moment** we only make **release version** of docker images (like `all-in-one`, `jaeger-agent`, `jaeger-tracegen`, etc.) **multi-arch** and using `debug_image="golang:1.15-alpine"` in `build-arg`. In fact, the argument `debug_image` is not used when building release version of images.

The **debug version** of images (like `all-in-one-debug` or `jaeger-agent-debug` ) still use `debug_image=localhost/debugimg:1.0.0-golang-1.15-alpine` and are **`amd64` specific**.

scripts/build-upload-docker-images.sh
```diff
@@ -0,0 +1,93 @@
...
+			cmd/${component}
+		echo "Finished building ${component} =============="
+	done
+
+  docker buildx build --output "${PUSHTAG}" \
+		--progress=plain \
+		--platform=${PLATFORMS} \
+    $(echo ${IMAGE_TAGS} | sed "s/JAEGERCOMP/es-index-cleaner/g") \
+		plugin/storage/es
+	docker buildx build --output "${PUSHTAG}" \
```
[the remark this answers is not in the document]
kun-lu20: Yes, we can, I've updated the code to implement this.

scripts/compute-tag-for-multiarch-image.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+
+set -exu
+
+###############Compute the tag
+BASE_BUILD_IMAGE=$1
+
+## if we are on a release tag, let's extract the version number
+## the other possible value, currently, is 'master' (or another branch name)
+if [[ $BRANCH == v* ]]; then
```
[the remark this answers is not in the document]
kun-lu20: Yes, you're right, I've updated the code to use strict regex match.

kun-lu20:
> It's looking great. One minor feedback: Do we need to append `for-multiarch-image` in the script names.

@Ashmita152 Thank you very much! I've removed `for-multiarch-image` from script names and made code changes as per your review comments. Please take a look when you are available.

kun-lu20:
@yurishkuro Could you please review this again? Thank you very much!

yurishkuro:
@kun-lu20 could you please describe what kind of testing you did to make sure this all works? Ideally you would run this in your fork with a tag and override the target repos in the registries to your private ones, so that the end result are the images pushed to registries that can be validated.

kun-lu20:
@yurishkuro Thanks for your response! Yes, each time before I committed the code changes, I would create a testing branch in our fork with the name `vM.N.P`, and replaced the registry namespace and credentials with my private ones, so that the code changes and GA workflows could be tested in our fork. 
The built images were pushed onto repos on `docker.io` and `quay.io` registries under my private namespace `kunlu20` (publicly available) and were primitively validated as per the instructions on https://www.jaegertracing.io/docs/1.22/getting-started/ and https://www.jaegertracing.io/docs/1.22/deployment/.

scripts/build-upload-docker-images.sh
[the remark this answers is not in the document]
Ashmita152: > In fact, the argument debug_image is not used when building release version of images.

If this argument is not used, is it mandatory to pass it ? It just creates confusion, isn't it ?

scripts/build-upload-docker-images.sh
```diff
@@ -0,0 +1,87 @@
...
+make build-binaries-s390x
+
+PLATFORMS="linux/amd64,linux/s390x"
+bash scripts/build-multiarch-baseimg.sh
+
+IMAGE_TAGS=$(bash scripts/compute-tags.sh "jaegertracing/jaeger-JAEGERCOMP")
+
+# Only push multi-arch images to dockerhub/quay.io for master branch or for release tags vM.N.P
+if [[ "$BRANCH" == "master" || $BRANCH =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+  echo "build multiarch images and upload to dockerhub/quay.io, BRANCH=$BRANCH"
```
Ashmita152, line 52: I can see that in some places within this script you are using space while at other places you are using tab for indentation.

scripts/compute-tags.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+
+set -exu
+
+###############Compute the tag
```
Ashmita152, line 5: can we change thing to just single #

scripts/compute-tags.sh
```diff
@@ -0,0 +1,39 @@
...
+fi
+
+if [ "${MAJOR}x" != "x" ]; then
+    MAJOR_IMAGE="${BASE_BUILD_IMAGE}:${MAJOR}"
+    IMAGE_TAGS="${IMAGE_TAGS} --tag docker.io/${MAJOR_IMAGE} --tag quay.io/${MAJOR_IMAGE}"
+fi
+
+IMAGE_TAGS="${IMAGE_TAGS} --tag docker.io/${SNAPSHOT_TAG} --tag quay.io/${SNAPSHOT_TAG}"
+
+################################
```
Ashmita152, line 37: would suggest to remove it.

scripts/build-upload-docker-images.sh
```diff
@@ -0,0 +1,87 @@
...
+	done
+
+	for component in es-index-cleaner es-rollover tracegen anonymizer
+	do
+		if [[ "${component}" == "es-rollover" ]]; then
+			docker_file_arg="--file plugin/storage/es/Dockerfile.rollover"
+		else
+			docker_file_arg=""
+		fi
+		if [[ "${component}" =~ ^es-.* ]]; then
```
Ashmita152, line 26: Won't this regex match for es-rollover too ? Is that intended ?

Ashmita152: This multiple if-else condition is making things confusing. Can we make it cleaner with either switch-case or some other way ?

yurishkuro: a simpler approach is to define a function with args like docker_file_arg and then call the function for each component with necessary args

kun-lu20: @Ashmita152  Yes, this regex match for both `es-index-cleaner` and `es-rollover` , since the value of their `dir_arg` is the same (`plugin/storage/es`).

scripts/compute-tags.sh
kun-lu20: OK, removed

scripts/compute-tags.sh
kun-lu20: Yes, changes done.

scripts/build-upload-docker-images.sh
kun-lu20: I've updated this script to use tab for indentation. Thanks.

scripts/build-upload-docker-images.sh
kun-lu20: I've separated out a function docker_buildx_build() in this script and call it for each component with necessary args. Thanks.

kun-lu20:
Hi @yurishkuro @Ashmita152 , I've made code changes as per your comments and tested them in our fork. The built images were pushed onto repos on `docker.io` and `quay.io` registries under my private namespace `kunlu20` . Could you please review this again? Thanks!

scripts/build-upload-docker-images.sh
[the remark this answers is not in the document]
kun-lu20: Yes, since the debug images also use the same Dockerfile, we need to **keep** this arg and **pass** it. Although it would create confusion at this moment, once `go-delve` is supported by `s390x`, we can use this arg for building multi-arch debug images. Otherwise we need to separate each of these Dockerfiles into two Dockerfiles, one for `release` target, the other for `debug` target.

scripts/build-multiarch-baseimg.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
kun-lu20: Yes, we can. I've removed this script and incorporated the code for `multi-arch` base image creation into the Makefile. Since the `multi-arch` images already include the `amd64` version, I kept only **one** target for creating the `multi-arch` base image in the Makefile and made relevant changes in other scripts.

kun-lu20:
Hi @Ashmita152 , thank you so much for your feedback! I checked the code and removed or incorporated the unnecessary duplicate parts, and I also added multi-arch support for images of two more components. The following code changes have been made:

1. Remove the duplicate code from `scripts/upload-to-registry.sh`, modify relevant code in other scripts.
2. Search for duplicate code in other scripts (like `scripts/build-all-in-one-image.sh`) and adjust them.
3. Add multi-arch image support for component `jaegertracing/example-hotrod` and `jaegertracing/test-driver`.
4. Move multi-arch `baseimg` creation code to `docker/Makefile`, keep only one target for creating the multi-arch base image in the Makefile and made relevant changes in other scripts.

Please take a look when you are available, Thanks again!

kun-lu20:
Hi @yurishkuro @Ashmita152 , could you please review this again? Thanks!

kun-lu20:
@yurishkuro Could you please have a look? Thank you very much!

jpkrohling:
@Ashmita152, @kun-lu20, this PR seems to have grown in scope, making it harder to fully grasp what's going on. Would you be open to refactoring the current build we have in smaller pieces, making room for the final goal, which is to have multi-arch builds?

kun-lu20:
@jpkrohling Thanks for your response! Sure, we can divide current code changes into smaller pieces and go step by step. I suggest firstly we could add s390x support on multi-arch docker images for `baseimg` and `all-in-one`, then we could work on other Jaeger components. Please let me know what you think, thanks again! @jpkrohling @Ashmita152

yurishkuro:
Actually my concern was not so much with covering all images, but with duplication of the code for building them, which now gets a lot more complicated due to buildx. If we put aside the Python based images for ES tools, all other images should be built in nearly identical way. The one exception is the images that require UI assets, but they could still be covered by a single make target. So what I would like to see is refactoring all the different scripts to converge to common shared scripts for building and pushing the images, and then extending that single point with multi-arch support.

kun-lu20:
@yurishkuro Yes, that makes sense. We could try to incorporate the different scripts for building and pushing the images in one place and add multi-arch support on it.

der-eismann:
It would be awesome to add support for `arm64` while you're at it, should only be a few more chars to support Apple M1 and AWS m6g instances as well.

kun-lu20:
@der-eismann Yes, we will try to add multi-arch support for docker images on Intel+s390x first, since we can validate results on these platforms. Adding other platforms after that would not be difficult.

kun-lu20:
Hi @yurishkuro , I created the following initial plan for refactoring the whole image building and uploading process and made a simple prototype to verify its feasibility. Could you please take a look? Thanks.

We could create a common shared script `build-upload-a-docker-image.sh`. This shared script will use  `docker buildx build`  for all the docker images. It would have the following arguments:
1. `component_name`
2. `base_image_debug_image` ('N' or corresponding tags of base image and debug image) (`all-in-one` and `Jaeger-backend` components would use this arg)
3. `extra_image_tags` ('N' or additional image tags for this image) (`crossdock` component would use this arg)
4. `directory_path`
5. `Dockerfile_path`
6. `target` ('N', 'release', 'debug') (`all-in-one` and `Jaeger-backend` components would use this arg)
7. `platform` ('linux/amd64,linux/s390x' or 'linux/amd64')
8. `local_test_only` ('Y' for pushing the images only to local registry) (`all-in-one` and `hotrod` component would use this arg)
9. `push_for_master_only` (push images to remote registry for Master branch only ('Y') or also for release tags vM.N.P ('N')) (`crossdock` component would use this arg)

The logic inside this new script will include:
1. According to current branch name, determine if we need to push the built multi-arch docker image to remote registries.
2. Calculate the whole image tag set.
3. Use the above arguments to construct the complete `docker buildx build` command statement, execute the command to build/upload the docker image.

The image building and uploading process is now implemented in this **single** script. Other scripts such as `build-all-in-one-image.sh` or `build-upload-docker-images.sh` only need to use the Makefile targets to generate binaries and invoke this script with corresponding arguments for multi-arch/single-arch images building and releasing. Thus the code redundancy would be **eliminated**. 

I think the 9 arguments in the above shared script would cover all the requirements of various `Jaeger` components. However, having too many arguments might complicate the invoke of the script. We thought that using configuration files might be helpful to reduce the number of arguments. Could you please give us some suggestions on this? Thank you very much!

yurishkuro:
@kun-lu20 You could env vars for passing parameters or getops:

https://github.com/jaegertracing/jaeger/blob/c5642b708be9e2577cc1c494889ae97946ccc78a/pkg/config/tlscfg/testdata/gen-certs.sh#L15-L24

You could also use defaults so that you only need to pass less-frequent values. 

You can also consider NOT creating a single mega-script, but separating into build / upload (assuming this would make the list of args smaller, otherwise keep one script).

You may want to keep crossdock separate and not introduce numerous if/else branches in the script just for crossdock. It looks like 2 of the parameters are only needed for crossdock, so you can eliminate them by keeping it separate.

kun-lu20:
Hi @yurishkuro , thanks for your suggestions! 

I believe we could use `env vars`, `getops` and default values to simplify the argument passing. 

Since both the image building and uploading process could be done in one `docker buildx build` command, we'll still keep them in one script. 

Keeping `crossdock` separate would reduce the number of arguments and if/else branches in the shared script and would not induce much redundancy, so it's definitely worth trying it.

docker/Makefile
```diff
@@ -3,18 +3,21 @@ ROOT_IMAGE ?= alpine:3.13
 CERT_IMAGE := alpine:3.13
 GOLANG_IMAGE := golang:1.15-alpine
 
-BASE_IMAGE := localhost/baseimg:$(VERSION)-$(shell echo $(ROOT_IMAGE) | tr : -)
-DEBUG_IMAGE := localhost/debugimg:$(VERSION)-$(shell echo $(GOLANG_IMAGE) | tr : -)
+BASE_IMAGE := localhost:5000/baseimg:$(VERSION)-$(shell echo $(ROOT_IMAGE) | tr : -)
+DEBUG_IMAGE := localhost:5000/debugimg:$(VERSION)-$(shell echo $(GOLANG_IMAGE) | tr : -)
```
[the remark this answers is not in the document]
kun-lu20: Yes, you're right, I'll change the names of base image and debug image to stable local names. Thanks!

scripts/build-crossdock.sh
```diff
@@ -10,12 +10,18 @@ make build-and-run-crossdock
...
 if [[ "$BRANCH" == "master" ]]; then
   echo 'upload images to dockerhub/quay.io'
+  make build-crossdock-binary GOOS=linux GOARCH=amd64
+  make build-crossdock-binary GOOS=linux GOARCH=s390x
+  PLATFORMS="linux/amd64,linux/s390x"
+  REPO=jaegertracing/test-driver
+  IMAGE_TAGS=$(bash scripts/compute-tags.sh $REPO)
+  IMAGE_TAGS="${IMAGE_TAGS} --tag docker.io/${REPO}:${COMMIT} --tag quay.io/${REPO}:${COMMIT}"
+  bash scripts/docker-login.sh
+  docker buildx build --push \
```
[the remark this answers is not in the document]
kun-lu20: You're right, we do not need to make crossdock images multi-arch. I'll revert this script to use `docker build`.

scripts/compute-tags.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+
+set -exu
+
+# Compute the tag
```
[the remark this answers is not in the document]
kun-lu20: Updated as per the suggestion

scripts/compute-tags.sh
```diff
@@ -0,0 +1,37 @@
...
+BASE_BUILD_IMAGE=$1
```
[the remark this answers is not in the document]
kun-lu20: Updated as per the suggestion

docker/Makefile
```diff
@@ -3,18 +3,21 @@ ROOT_IMAGE ?= alpine:3.13
 CERT_IMAGE := alpine:3.13
```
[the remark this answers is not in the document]
kun-lu20: Updated as per the suggestion

scripts/build-crossdock.sh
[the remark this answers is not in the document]
kun-lu20: Hi @yurishkuro , since we've removed script `upload-to-registry.sh` for eliminating redundancy, I think here we could set `platforms` to `linux/amd64` and keep using `docker buildx` to upload images for code simplicity. I've updated script `build-crossdock.sh`. Please let me know what you think, thanks!

docker/Makefile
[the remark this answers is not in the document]
kun-lu20: Hi @yurishkuro , I've changed the names of `baseimg` and `debugimg` to stable local names. I still set the `debug_image` arg to `golang:1.15-alpine` for building multi-arch images in `build-all-in-one-image.sh` and `build-upload-docker-images.sh`, since the built `debugimg` doe not support `s390x` yet (due to lack of `go-delve` support on s390x). For building amd64 images, the `debug_image` arg is set to the stable local name.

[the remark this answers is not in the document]
kun-lu20: Hi @yurishkuro , I've made some changes to remove cross-script dependencies of `debug_image` image names. Please take a look when you are available. Thanks!

kun-lu20:
Hi @jpkrohling @yurishkuro , could you please have a look at the recent code changes? I am looking forward to hearing the comments from you. Thank you very much!

Makefile
```diff
@@ -269,9 +269,6 @@ build-collector build-collector-debug:
 build-ingester build-ingester-debug:
 	$(GOBUILD) $(DISABLE_OPTIMIZATIONS) -o ./cmd/ingester/ingester$(SUFFIX)-$(GOOS)-$(GOARCH) $(BUILD_INFO) ./cmd/ingester/main.go
 
-.PHONY: docker
```
jpkrohling, line 272: I still find the docker and docker-push targets useful. I had to use them during the last release, as I had to manually build/publish the images.

docker/debug/Dockerfile
```diff
@@ -1,11 +1,16 @@
...
-    go get github.com/go-delve/delve/cmd/dlv && \
-    cd /go/src/github.com/go-delve/delve && \
-    make install
+RUN apk add --update --no-cache ca-certificates make git
+RUN if [[ "$TARGETARCH" != "s390x" ]] ; then \
+        go get github.com/go-delve/delve/cmd/dlv && \
+        cd /go/src/github.com/go-delve/delve && \
+        make install; \
+    else \
+        touch /go/bin/dlv; \
```
jpkrohling, line 12: Wait, what? Can you add a comment here describing what's going on? Is dlv not supported at all for s390x? If so, we shouldn't be publishing this image for this arch.

scripts/build-crossdock.sh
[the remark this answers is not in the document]
jpkrohling: What's the penalty in terms of clock time? buildx is slower than regular builds, isn't it?

scripts/build-upload-a-docker-image.sh
```diff
@@ -0,0 +1,70 @@
...
+	c)
+		component_name=${OPTARG}
+		;;
+	b)
+		base_debug_img_arg="--build-arg base_image=localhost:5000/baseimg_alpine:latest --build-arg debug_image=localhost:5000/debugimg_alpine:latest "
+		;;
+	d)
+		dir_arg=${OPTARG}
+		;;
+    f)
```
jpkrohling, line 23: The indentation seems inconsistent between the flags.

scripts/build-upload-docker-images.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
```
jpkrohling, line 1: This script would then replace the `make docker docker-push` targets, right?

jpkrohling (commented):
Looks OK to me, had only a couple of comments. But in general, I think we could think a bit about what the ideal world would be in terms of building both the binaries and the images. I feel like there are quite a few things that can be improved.

scripts/build-upload-docker-images.sh
kun-lu20: Yes, it will replace these targets to build and upload multi-arch images.

Makefile
kun-lu20: I see, now you can use `scripts/build-upload-docker-images.sh` to build/publish multi-arch images.

docker/debug/Dockerfile
kun-lu20: `go-delve` does not have s390x support yet. Since Dockerfile of `release` version images also requires the `debug_image` arg to be populated, we did this in `docker/debug/Dockerfile` to make a different version of `debugimg` for s390x and eliminate cross-script dependencies of `debug_image` arg values.

BTW, `debug` version images such as `all-in-one-debug` and `jaeger-agent-debug` are not currently built for s390x (we use `--platform="linux/amd64"` arg in `docker buildx` command for these images).

scripts/build-crossdock.sh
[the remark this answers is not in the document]
kun-lu20: Yeah, `docker buildx` might be slower than regular build, but it builds the images concurrently and would bring the benefits of easy maintenance of multi-arch images.

scripts/build-upload-a-docker-image.sh
kun-lu20: issue solved. Thanks!

docker/debug/Dockerfile
jpkrohling: We need a comment here then, possibly instructing our future selves which conditions should exist for us to remove this. For instance, "Once go-delve adds support for s390x (see issue #...) remove this entire conditional"

scripts/build-crossdock.sh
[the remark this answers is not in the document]
jpkrohling: How much slower? If we talking about a couple of seconds slower in total, it's fine. If it adds 5 minutes to the build, it's not acceptable.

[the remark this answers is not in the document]
kun-lu20: I made a test from my end. The workflow `ci-all-in-one-build`, `ci-docker-build` and `ci-hodrod` will take 10m 47s, 9m 21s and 2m 2s when using regular build for image building respectively. When using `docker buildx` to build multi-arch images, these workflows will take 10m 14s, 12m 46s and 2m 19s. 

The workflow `ci-docker-build` might be slower when building multi-arch images since it has 13 different images to build. Maybe we could reduce the frequency of this workflow or divide it into smaller parallel tasks to mitigate the time penalty.

docker/debug/Dockerfile
kun-lu20: Comment added. Thanks!

scripts/build-crossdock.sh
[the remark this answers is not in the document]
jpkrohling: I think the penalty isn't high enough right now to block this PR. We should certainly consider rethinking our CI, to avoid building all the things all the time if we don't need them for the tests. We can then have a nightly build to publish the snapshot images.

jpkrohling (approved):
LGTM. I'll give a couple of days for @yurishkuro and @Ashmita152 to review it again, in case they are interested. If I don't hear complaints from them, I'll merge this on Wed.

scripts/build-crossdock.sh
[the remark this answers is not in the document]
kun-lu20: Sure, I've opened an issue #3113 for tracking. Thanks!

yurishkuro (approved):

jpkrohling:
@kun-lu20, could you please rebase this PR? Looks like I can't rebase it myself. Once it's done, I'll merge this.

kun-lu20:
Thanks @yurishkuro @jpkrohling !

scripts/build-upload-docker-images.sh
albertteoh: @kun-lu20 @Ashmita152 on my macbook I tried running `sudo bash scripts/build-upload-docker-images.sh` (after `brew install buildkit`) and getting the error:
```
error: failed to solve: rpc error: code = Unknown desc = failed to do request: Head http://localhost:5000/v2/baseimg_alpine/blobs/sha256:595b0fe564bb9444ebfe78288079a01ee6d7f666544028d5e96ba610f909ee43: dial tcp 127.0.0.1:5000: connect: connection refused
```

I tried removing the `--push` from the `create-baseimg` target just to explore options to resolve this then added `darwin/amd64` to the platform list but seeing this error:
```
 => ERROR [darwin/amd64 internal] load metadata for docker.io/library/alpine:3.13                                                                                                                                  8.2s
 => [linux/s390x internal] load metadata for docker.io/library/alpine:3.13                                                                                                                                         8.1s
 => [linux/arm64 internal] load metadata for docker.io/library/alpine:3.13                                                                                                                                         8.1s
 => [linux/amd64 internal] load metadata for docker.io/library/alpine:3.13                                                                                                                                         8.1s
 => [auth] library/alpine:pull token for registry-1.docker.io                                                                                                                                                      0.0s
------
 > [darwin/amd64 internal] load metadata for docker.io/library/alpine:3.13:
------
error: failed to solve: rpc error: code = Unknown desc = failed to solve with frontend dockerfile.v0: failed to create LLB definition: no match for platform in manifest sha256:f51ff2d96627690d62fee79e6eecd9fa87429a38142b5df8a3bfbb26061df7fc: not found
make: *** [create-baseimg] Error 1
```

Any tips on how to resolve or troubleshoot this?

kun-lu20: I think the cause of the first error is that `docker buildx` could not connect to the local registry container running on port 5000. 
I haven't tried it on macOS, but from the error log, it looks like `alpine:3.13` doesn't support platform `darwin/amd64` yet.

albertteoh: Thanks @kun-lu20! I managed to get past the error with:
```
brew install buildkit

# start the local registry
docker run -d -p 5000:5000 --name registry registry:2

# create the buildx container and ensure it can reach the registry container on the host network
docker buildx create --driver-opt=network=host --use
```

then I was able to successfully run:
```
make docker-images-only
```

You can ignore my attempt to add `darwin/amd64`, I forgot that Docker Desktop for mac runs its own linux VM.

kun-lu20: No problem! I think now you could run `scripts/build-upload-docker-images.sh` successfully as well.

PR #2966, opened 2021-04-28
YAAAAS - Yet Another Attempt At Adaptive Sampling

### LINKED ISSUE #365: Adaptive Sampling

joe-elliott:
- Fixes #365 

This PR builds on #2818 and as such @Ashmita152 is named as a coauthor. Currently it only wires config for adaptive sampling to the existing Cassandra storage implementations. 

## Additional changes
- Adjusts hotrod jaeger tracer creation to allow configuration through env vars 
- Makes changes to ./docker-compose/jaeger-docker-compose.yaml to test adaptive sampling with hotrod.


jpkrohling:
Do you want an initial review, or should I wait for the remaining items to be done?

joe-elliott:
Let me move this a little farther forward before a first review. I have everything wired up but I'd like to see it work at least a little before we start reviews.

jpkrohling:
I'll mark this as a draft then.

wgb1990:
@joe-elliott

Sorry, I asked a very simple question. In the process of reading the source code, I found that the Store.InsertThroughput method was not called anywhere

// Store writes and retrieves sampling data to and from storage.
type Store interface {
// InsertThroughput inserts aggregated throughput for operations into storage.
InsertThroughput(throughput []*model.Throughput) error
......
}

yurishkuro:
I realized there are a few files missing from what was available internally at Uber. I will work with @vprithvi to get them committed (they were all cleaned up for oss, but for some reason not moved).

yurishkuro:
The missing files have been added in #3065

cmd/collector/main.go
```diff
@@ -79,21 +85,33 @@ func main() {
...
 				logger.Fatal("Failed to create sampling strategy store", zap.Error(err))
 			}
-
 			c := app.New(&app.CollectorParams{
 				ServiceName:    serviceName,
 				Logger:         logger,
 				MetricsFactory: metricsFactory,
 				SpanWriter:     spanWriter,
 				StrategyStore:  strategyStore,
+				Aggregator:     aggregator,
```
yurishkuro, line 114: Note: we intentionally kept collector unaware of "aggregators", it only needed to know about generic "span processors"

joe-elliott: I had it this way (and do not mind moving back) but preferred this way for two reasons:

1) collector can own the lifecycle of the Aggregator and call stop when it is closed
2) otherwise the main function has to do the translation from Aggregator -> ProcessSpan which just felt like clutter in main

joe-elliott:
Allright, I'm going to drop draft status on this PR as I think it's ready for review. codecov is not passing yet, but I wanted to see how the review shook out before expanding the tests. There is one final "todo" that I'm not sure how to handle, but I'm still mulling over it.

```
$ curl http://localhost:5778/sampling?service="frontend"
{"strategyType":"PROBABILISTIC","operationSampling":{"defaultSamplingProbability":0.5,"defaultLowerBoundTracesPerSecond":0.016666666666666666,"perOperationStrategies":[{"operation":"HTTP GET /config","probabilisticSampling":{"samplingRate":0.008823601197058824}},{"operation":"HTTP GET /dispatch","probabilisticSampling":{"samplingRate":0.008333401130555557}},{"operation":"HTTP GET /","probabilisticSampling":{"samplingRate":0.15000122035}}],"defaultUpperBoundTracesPerSecond":0}}
```

```
cqlsh:jaeger_v1_dc1> select * from operation_throughput;

 bucket | ts                                   | throughput
--------+--------------------------------------+----------------------------------------------------------------------------------------------------------------------------------------
      8 | 2e95481a-cdf5-11eb-84c2-0242ac1d0005 | frontend,HTTP GET /,2,1.000000\nfrontend,HTTP GET /config,34,"0.500000,1.000000"\nfrontend,HTTP GET /dispatch,36,"1.000000,0.500000"\n
      2 | f1d4c95a-ce0d-11eb-91db-0242ac1d0005 |                                                                                               frontend,HTTP GET /dispatch,1,0.008333\n
      2 | 761bdf37-cdf5-11eb-84ca-0242ac1d0005 |                                                         frontend,HTTP GET /config,5,0.500000\nfrontend,HTTP GET /dispatch,4,0.500000\n
      7 | 52587dfd-cdf5-11eb-84c6-0242ac1d0005 |                                                                                                                                       
      6 | 99df500d-cdf5-11eb-84ce-0242ac1d0005 |                                                                                                                                       
      9 | ce11878d-ce0d-11eb-91d7-0242ac1d0005 |                         frontend,HTTP GET /config,0,0.008824\nfrontend,HTTP GET /dispatch,0,0.008333\nfrontend,HTTP GET /,0,0.150001\n
```

yurishkuro:
Are we saving throughput data as some sort of CSV format? Maybe worth converting to JSON before we commit, because JSON can be further extended in the future, but CSV would make any extension difficult.

joe-elliott:
> Are we saving throughput data as some sort of CSV format? Maybe worth converting to JSON before we commit, because JSON can be further extended in the future, but CSV would make any extension difficult.

Yes, that code is here:
https://github.com/jaegertracing/jaeger/blame/master/plugin/storage/cassandra/samplingstore/storage.go#L138

There are two sets of methods to translate between strings and probabilities/throughput. I don't know enough about the original implementation to comment on why csv was chosen. json would be heavier weight to parse/store if there were a large number of operations to store, but I can look into it.

examples/hotrod/pkg/tracing/init.go
```diff
@@ -30,14 +30,18 @@ import (
 
 // Init creates a new instance of Jaeger tracer.
 func Init(serviceName string, metricsFactory metrics.Factory, logger log.Factory) opentracing.Tracer {
-	cfg, err := config.FromEnv()
-	if err != nil {
-		logger.Bg().Fatal("cannot parse Jaeger env vars", zap.Error(err))
+	cfg := &config.Configuration{
```
jpkrohling, line 33: It might become clearer later, but why is this change needed?

plugin/sampling/leaderelection/leader_election.go
```diff
@@ -31,6 +31,8 @@ const (
 // ElectionParticipant partakes in leader election to become leader.
 type ElectionParticipant interface {
 	IsLeader() bool
+	Start() error
+	Close() error
```
jpkrohling, line 35: Can you include `io.Closer` instead?

cmd/collector/main.go
jpkrohling: If ProcessSpan was an interface, implementations complying with io.Closer could be closed by the spanProcessor. Without that, I think I also prefer the current code from this PR.

plugin/sampling/strategystore/adaptive/options.go
```diff
@@ -111,6 +113,10 @@ type Options struct {
 	// FollowerLeaseRefreshInterval is the duration to sleep if this processor is a follower
 	// (ie. failed to gain the leader lock).
 	FollowerLeaseRefreshInterval time.Duration
+
+	// OverrideHostname overrides the hostname returned from os.Hostname for use in the adaptive sampling
```
jpkrohling, line 117: Why is this needed?

plugin/storage/cassandra/factory.go
```diff
@@ -147,6 +152,19 @@ func (f *Factory) CreateArchiveSpanWriter() (spanstore.Writer, error) {
 	return cSpanStore.NewSpanWriter(f.archiveSession, f.Options.SpanStoreWriteCacheTTL, f.archiveMetricsFactory, f.logger, options...), nil
 }
 
+// CreateLockAndSamplingStore creates a distributedlock.Lock and samplingstore.Store for use with adaptive sampling
+func (f *Factory) CreateLockAndSamplingStore() (distributedlock.Lock, samplingstore.Store, error) {
+	hostname, err := os.Hostname() // todo(jpe): wire up --sampling.override-hostname in ./plugin/sampling/strategystore/adaptive/options.go?
```
jpkrohling, line 157: Now I see why you need the override :-) I think JGroups used to add a short random string to the hostname in order to allow multiple instances on the same hostname.

plugin/storage/cassandra/schema/v003.cql.tmpl
```diff
@@ -201,4 +201,29 @@ CREATE TABLE IF NOT EXISTS ${keyspace}.dependencies_v2 (
         'max_threshold': '32',
         'class': 'org.apache.cassandra.db.compaction.SizeTieredCompactionStrategy'
     }
-    AND default_time_to_live = ${dependencies_ttl};
\ No newline at end of file
+    AND default_time_to_live = ${dependencies_ttl};
+
+-- adaptive sampling tables
```
jpkrohling, line 206: Doesn't it deserve a new file?

plugin/storage/factory_test.go
```diff
@@ -392,29 +401,69 @@ func TestPublishOpts(t *testing.T) {
...
+
+	// test succeeds (supported or not)
+	f := Factory{
+		factories: map[string]storage.Factory{
+			"supports":         supports,
+			"does-not-support": doesNotSupport,
+		},
+	}
+
+	_, _, err := f.CreateLockAndSamplingStore()
```
jpkrohling, line 421: Can you assert that it returns `supports`?

plugin/storage/factory_test.go
```diff
@@ -392,29 +401,69 @@ func TestPublishOpts(t *testing.T) {
...
+	assert.NoError(t, err)
+
+	delete(f.factories, "supports")
+	_, _, err = f.CreateLockAndSamplingStore()
```
jpkrohling, line 425: Can you assert that the store is nil?

storage/factory.go
```diff
@@ -53,6 +59,9 @@ var (
 
 	// ErrArchiveStorageNotSupported can be returned by the ArchiveFactory when the archive storage is not supported by the backend.
 	ErrArchiveStorageNotSupported = errors.New("archive storage not supported")
+
+	// ErrLockAndSamplingStoreNotSupported can be returned by the StorageFactory when the sampling storage is not supported by the backend.
+	ErrLockAndSamplingStoreNotSupported = errors.New("archive storage not supported")
```
jpkrohling, line 64:
```suggestion
	ErrLockAndSamplingStoreNotSupported = errors.New("lock/sampling storage not supported")
```

jpkrohling (commented):
Looks good! I still have to run the end-to-end test, though. I'll try to run it on Monday and will approve the PR then if all goes well.

plugin/storage/cassandra/factory.go
joe-elliott: I prefer this approach and will use it. I was having trouble wiring up the `sampling.override-hostname` param to the two places that needed it.

examples/hotrod/pkg/tracing/init.go
joe-elliott: This makes hotrod respect env vars which I used in `/docker-compose/jaeger-docker-compose.yml` to setup hotrod as a local test of adaptive sampling.

Without this change hotrod is hardcoded at `const`/`1`.

jpkrohling: Right, I see. Now, it is const/1, but overridden from the env var if values are provided there. Sounds good!

plugin/sampling/strategystore/adaptive/strategy_store.go
```diff
@@ -15,27 +15,22 @@
...
-		hostname, err = os.Hostname()
-		if err != nil {
-			return nil, err
-		}
-		logger.Info("Adaptive sampling hostname retrieved from os.Hostname()", zap.String("hostname", hostname))
+	hostname, err := hostname.AsIdentifier()
+	if err != nil {
+		return nil, err
 	}
+	logger.Info("Retrieved unique hostname from the OS for use in adaptive sampling", zap.String("hostname", hostname))
```
jpkrohling, line 33: This might now be confusing, as "hostname" isn't really the hostname. How about renaming to `participantName`? The message could then be `Using %q as the participant name for this instance`.

plugin/storage/cassandra/factory.go
```diff
@@ -154,10 +154,11 @@ func (f *Factory) CreateArchiveSpanWriter() (spanstore.Writer, error) {
 
 // CreateLockAndSamplingStore creates a distributedlock.Lock and samplingstore.Store for use with adaptive sampling
 func (f *Factory) CreateLockAndSamplingStore() (distributedlock.Lock, samplingstore.Store, error) {
-	hostname, err := os.Hostname() // todo(jpe): wire up --sampling.override-hostname in ./plugin/sampling/strategystore/adaptive/options.go?
+	hostname, err := hostname.AsIdentifier()
 	if err != nil {
 		return nil, nil, err
 	}
+	f.logger.Info("Retrieved unique hostname from the OS for use in the distributed lock", zap.String("hostname", hostname))
```
jpkrohling, line 161: Same here.

jpkrohling (approved):
LGTM, just a couple of minor comments about log messages.

jpkrohling:
Linting is failing with:

`[/home/runner/work/jaeger/jaeger/pkg/hostname/hostname.go:18] - G404 (CWE-338): Use of weak random number generator (math/rand instead of crypto/rand) (Confidence: MEDIUM, Severity: HIGH)`

jpkrohling:
@joe-elliott, I see in a test here a reference to the flag `--sampling.initial-sampling-probability`, but I can't seem to be able to use it with this PR:

```console
$ git log -1
commit b72f4704b4dac191d147fa34b92f23dbf5804280 (HEAD -> adaptive-sampling)
Author: Joe Elliott <number101010@gmail.com>
Date:   Fri Jun 25 16:21:59 2021 -0400

    added hostname
    
    Signed-off-by: Joe Elliott <number101010@gmail.com>
$ SPAN_STORAGE_TYPE=cassandra go run -tags ui ./cmd/collector/ --log-level debug --sampling.initial-sampling-probability=.5
Error: unknown flag: --sampling.initial-sampling-probability
$ SPAN_STORAGE_TYPE=cassandra go run -tags ui ./cmd/collector/ --help | grep sampling.initial-sampling-probability
$
```

Would you mind writing a couple of lines describing how I can test this?

cmd/collector/app/sampling/strategystore/factory.go
```diff
@@ -27,8 +30,11 @@ import (
 // plugin.Configurable
 type Factory interface {
 	// Initialize performs internal initialization of the factory.
-	Initialize(metricsFactory metrics.Factory, logger *zap.Logger) error
+	Initialize(metricsFactory metrics.Factory, logger *zap.Logger, lock distributedlock.Lock, store samplingstore.Store) error
```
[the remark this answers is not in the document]
joe-elliott: Works. Moving to last.

cmd/collector/app/sampling/strategystore/interface.go
```diff
@@ -25,3 +25,15 @@ type StrategyStore interface {
...
+// Aggregator defines an interface used to aggregate operation throughput.
+type Aggregator interface {
+	// RecordThroughput records throughput for an operation for aggregation.
+	RecordThroughput(service, operation, samplerType string, probability float64)
+
+	// Start starts aggregating operation throughput.
+	Start()
+
+	// Stop stops the aggregator from aggregating throughput.
+	Stop()
```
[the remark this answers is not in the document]
joe-elliott: Agree.

plugin/storage/cassandra/schema/v003.cql.tmpl
joe-elliott: Perhaps I could add more comments to explain why it's creating these tables? If we create a second file we will need changes to `jaeger-cassandra-schema` to optionally apply it. Perhaps changes to the operator as well?

I think the cost of the extra empty tables is low as long as we add some comments so people know why they exist?

plugin/storage/factory_test.go
joe-elliott: Added logic to test that the appropriate mocks were returned.

joe-elliott:
> Would you mind writing a couple of lines describing how I can test this?
In the case of the above lines you simply need to define the sampling type (default is static):

```
SAMPLING_TYPE=adaptive SPAN_STORAGE_TYPE=cassandra go run -tags ui ./cmd/collector/ --log-level debug --sampling.initial-sampling-probability=.5
```

Also please review `/docker-compose/jaeger-docker-compose.yml`. I made some small changes here to run a full jaeger stack with hotrod and cassandra setup to use adaptive sampling.

docker-compose/jaeger-docker-compose.yml
```diff
@@ -1,11 +1,32 @@
 version: '2'
 
 services:
+    hotrod:
+      image: jaegertracing/example-hotrod:latest
```
yurishkuro, line 5: if this docker-compose/jaeger-docker-compose.yml is used in any crossdock tests then including hotrod might create conflicts in the ports (but might be ok if host names are used for addressing)

joe-elliott:
@jpkrohling 

> Are the participant names expected to be different here?

The way I had written it, yes, but I recognize that might be confusing. Improved the hostname package to always return the same value when called in the same process.

docker-compose/jaeger-docker-compose.yml
joe-elliott: `docker-compose/jaeger-docker-compose.yml` does not seem to be referenced by anything else in the repo. I figured it was a dev/debugging tool so it seemed ok to make updates.

yurishkuro: We used to import it (https://github.com/jaegertracing/jaeger-client-java/pull/276/files) but then removed in favor of only running all-in-one (https://github.com/jaegertracing/jaeger-client-java/pull/654/files)

jpkrohling (approved):

jpkrohling:
I think this is good to go! Let's get user feedback and iterate :-)

joe-elliott:
Let's go!

model/span.go
```diff
@@ -159,3 +195,15 @@ func (f Flags) IsFirehoseEnabled() bool {
 func (f Flags) checkFlags(bit Flags) bool {
 	return f&bit == bit
 }
+
+func getParam(samplerParamTag KeyValue) (float64, error) {
```
yurishkuro, line 199:
```suggestion
func samplerParamToFloat(samplerParamTag KeyValue) (float64, error) {
```

cmd/collector/app/sampling/strategystore/factory.go
```diff
@@ -27,8 +30,11 @@ import (
...
 	// Initialize performs internal initialization of the factory.
-	Initialize(metricsFactory metrics.Factory, logger *zap.Logger) error
+	Initialize(metricsFactory metrics.Factory, lock distributedlock.Lock, store samplingstore.Store, logger *zap.Logger) error
 
 	// CreateStrategyStore initializes the StrategyStore and returns it.
-	CreateStrategyStore() (StrategyStore, error)
+	CreateStrategyStore() (StrategyStore, Aggregator, error)
+
+	// RequiresLockAndSamplingStore indicates whether this strategy store requires a Lock and SamplingStore
+	RequiresLockAndSamplingStore() (bool, error)
```
yurishkuro, line 39: I'm not clear on the purpose of this. First, how can a factory for sampling store not require a sampling store? Second, who calls this and what changes depending on the output? 

Let's add better docs.

cmd/collector/main.go
```diff
@@ -80,21 +86,33 @@ func main() {
 				logger.Fatal("Failed to create span writer", zap.Error(err))
 			}
 
+			requireLockAndSamplingStore, err := strategyStoreFactory.RequiresLockAndSamplingStore()
+			if err != nil {
+				logger.Fatal("Failed to determine if lock and sampling store is required.", zap.Error(err))
+			}
+			var lock distributedlock.Lock
+			var samplingStore samplingstore.Store
+			if requireLockAndSamplingStore {
```
yurishkuro, line 95: What is the purpose of this check? I see that all factories implement the Create method (but many return nils), isn't that the same signal as the extra bool?

storage/factory.go
```diff
@@ -45,6 +47,10 @@ type Factory interface {
 
 	// CreateDependencyReader creates a dependencystore.Reader.
 	CreateDependencyReader() (dependencystore.Reader, error)
+
+	// CreateLockAndSamplingStore creates a distributedlock.Lock and samplingstore.Store.
+	//  These are used for adaptive sampling.
```
yurishkuro, line 52: Sampling store could be static though, can it not? So not always for adaptive.

cmd/collector/main.go
joe-elliott: Yeah, I was torn on this. I added it b/c I liked the explicitness of only attempting to create a sampling store/lock if it was necessary. Without this adaptive sampling will log something like:

"lock and sampling store nil. this is probably b/c your backend does not support adaptive sampling"

However, after all the factory/interface/implementation changes I question its value. Let me take a pass at removing it.

storage/factory.go
joe-elliott: The `SamplingStore` is explicitly for adaptive sampling. It is for storing and retrieving the various datapoints necessary to calculate the current sampling rates:

https://github.com/jaegertracing/jaeger/blob/master/storage/samplingstore/interface.go#L25

cmd/collector/main.go
yurishkuro: I think another aspect here is that originally the storage factory was not meant to be this all encompassing central entry point, i.e. in theory you should be able to configure Cassandra-based Span storage and another backend for Dependencies or Sampling. This is why the env var is specifically SPAN_STORAGE_TYPE, but we never really made clear separation.

joe-elliott: I have removed the "requireLockAndSamplingStore" functionality and feel like things look cleaner. It will be easy to revert if we want to go back.

> I think another aspect here is that originally the storage factory was not meant to be this all encompassing central entry point, i.e. in theory you should be able to configure Cassandra-based Span storage and another backend for Dependencies or Sampling. This is why the env var is specifically SPAN_STORAGE_TYPE, but we never really made clear separation.

This would require some work in `plugin/storage/factory.go` but may not be that bad. I think most of the code would be isolated here.

docker-compose/jaeger-docker-compose.yml
yurishkuro: Let's add a header to this file explaining why it exists, where it can be / should not be used.

plugin/sampling/strategystore/adaptive/strategy_store.go
```diff
@@ -0,0 +1,46 @@
...
+	"go.uber.org/zap"
+
+	"github.com/jaegertracing/jaeger/pkg/distributedlock"
+	"github.com/jaegertracing/jaeger/pkg/hostname"
+	"github.com/jaegertracing/jaeger/plugin/sampling/leaderelection"
+	"github.com/jaegertracing/jaeger/storage/samplingstore"
+)
+
+// NewStrategyStore creates a strategy store that holds adaptive sampling strategies.
+func NewStrategyStore(options Options, metricsFactory metrics.Factory, logger *zap.Logger, lock distributedlock.Lock, store samplingstore.Store) (*Processor, error) {
```
yurishkuro, line 28: I am now confused by what drives what. NewStrategyStore returns Processor? Is Processor API surface bigger than StrategyStore? If so, should this be renamed to reflect that?

plugin/sampling/strategystore/factory.go
```diff
@@ -23,30 +23,32 @@ import (
...
 )
 
-const (
-	staticStrategyStoreType = "static"
-)
+// Kind is a datatype holding the type of strategy store.
+type Kind string
 
-var allSamplingTypes = []string{staticStrategyStoreType} // TODO support adaptive
+var allSamplingTypes = []Kind{"static", "adaptive"}
```
yurishkuro, line 36: maybe define these as constants of Kind type? The literals are used in more than one place.

plugin/storage/badger/factory.go
```diff
@@ -164,6 +167,11 @@ func (f *Factory) CreateDependencyReader() (dependencystore.Reader, error) {
 	return depStore.NewDependencyStore(sr), nil
 }
 
+// CreateLockAndSamplingStore implements storage.ArchiveFactory. It is not supported by this storage plugin.
+func (f *Factory) CreateLockAndSamplingStore() (distributedlock.Lock, samplingstore.Store, error) {
```
yurishkuro, line 171: Another approach we could take that would minimize changes in this diff like this no-op addition to all factories is to just create a new optional interface SamplingStoreFactory and typecast-check for it where needed. I don't like all this noise being added to factories that don't need it.

plugin/sampling/strategystore/adaptive/strategy_store.go
joe-elliott: `Processor` implements the `strategystore.StrategyStore` interface which is used by the serving code to respond when a user hits `/sampling?service="test"`.  This is used for both adaptive and static sampling (and all future sampling served off this endpoint).

```
type StrategyStore interface {
	GetSamplingStrategy(ctx context.Context, serviceName string) (*sampling.SamplingStrategyResponse, error)
}
```

`samplingstore.Store` is an interface that abstracts away a backend (cassandra, elastic, ...) for use by the adaptive sampling code to record data for adaptive sampling.

```
type Store interface {
	InsertThroughput(throughput []*model.Throughput) error
	InsertProbabilitiesAndQPS(hostname string, probabilities model.ServiceOperationProbabilities, qps model.ServiceOperationQPS) error
	GetThroughput(start, end time.Time) ([]*model.Throughput, error)
	GetProbabilitiesAndQPS(start, end time.Time) (map[string][]model.ServiceOperationData, error)
	GetLatestProbabilities() (model.ServiceOperationProbabilities, error)
}
```

plugin/storage/badger/factory.go
joe-elliott: agreed. done.

jpkrohling (dismissed):

yurishkuro (approved):

joe-elliott:
woohoo!

Issue #2967, opened 2021-04-29
ILM policy is not configurable


jrRibeiro:
## Requirement - what kind of business use case are you trying to solve?

Following https://github.com/jaegertracing/jaeger/pull/2796 (great work, btw), we now have the ability to use ILM to manage indices. In _multi-tenancy_ scenarios, when running several Jaeger instances using the same Elastic storage, we can use different indices for each instance by defining its own prefix. However, different indices might need different ILM policies (some indices need different rollover and retention configurations than others).

## Problem - what in Jaeger blocks you from solving the requirement?

Right now, an ILM policy is hard-coded (named `jaeger-ilm-policy`) and it is not possible to change.

## Proposal - what do you suggest to solve the problem or improve the existing situation?

Make the policy's name configurable using an env var (to use with USE_ILM). This way, we can set the ILM policy to use and change the mappings of the index to use different policies per index.

I can open a new PR if you agree with my proposal. :slightly_smiling_face:

yurishkuro:
+1

PR #2971, opened 2021-04-30
Allow the ILM policy name to be configurable

jrRibeiro:

- Resolves https://github.com/jaegertracing/jaeger/issues/2967

- Add an option `--ilm-policy-name` to `esmapping-generator`, to set the policy name in the mappings file;
- Set the mapping's option in the rollover script (using an env var - `ES_ILM_POLICY_NAME`).
- The policy name defaults to `jaeger-ilm-policy` (both in the mapping generator and in the `init` action of the rollover script), so no breaking changes are added.

### DIFF
--- a/cmd/esmapping-generator/app/flags.go
+++ b/cmd/esmapping-generator/app/flags.go
@@ -20,21 +20,23 @@ import (
 
 // Options represent configurable parameters for jaeger-esmapping-generator
 type Options struct {
-	Mapping     string
-	EsVersion   uint
-	Shards      int64
-	Replicas    int64
-	IndexPrefix string
-	UseILM      string // using string as util is being used in python and using bool leads to type issues.
+	Mapping       string
+	EsVersion     uint
+	Shards        int64
+	Replicas      int64
+	IndexPrefix   string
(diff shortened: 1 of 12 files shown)


yurishkuro (dismissed):

jrRibeiro:
> @jrRibeiro I am not sure our CI covers this part of ES maintenance, did you do some manual testing? If so could you post some test results / evidence?

Yes I did some manual testing but it's better to actually test the change as part of the integration tests of the index rollover.

I just added a new integration test (and refactored the tests a bit) to test both the default ILM policy name (to make sure that not setting the env var still works) and also a new test to set the env var and test if the ILM policy associated with the created indices matches the one we define.

yurishkuro (dismissed):
LGTM 

@albertteoh can you take a look at the integration test changes?

albertteoh (dismissed):
LGTM too.

albertteoh (approved):

albertteoh:
Thanks @jrRibeiro!

jrRibeiro:
> Thanks @jrRibeiro!

Thank you @yurishkuro and @albertteoh!

Issue #2954, opened 2021-04-24
"Monitor" tab for service health metrics


albertteoh:
### Jaeger-Query

Owners: @albertteoh 

- [x] #2946: Add metrics query API spec
- [x] #2977: Add Metrics Reader interface
- [x] #2983, #2988, #3004: Add M3 reader implementation
- [x] #3049: Add factory and flags
- [x] #3055: Add TLS support
- [x] #3091: Add GRPC handler
- [x] #3095: Add HTTP handler
- [x] #3060, #3061: Update query service
- [x] #3079: Hookup metrics query to "main"
- [x] #3171: [Explore options to make monitor tab work for all-in-one (or create issue)](https://github.com/jaegertracing/jaeger/issues/3107)

### Jaeger-UI

Owners: @th3M1ke 

- [x] Approve UX & UI design (this issue)
- [x] https://github.com/jaegertracing/jaeger-ui/pull/815: Create Monitoring Tab for Jaeger UI

### Documentation

Owners: @albertteoh 

- [x] https://github.com/jaegertracing/documentation/pull/539: Add usage documentation for metrics query API (noting that this is "experimental")
- [x] https://github.com/jaegertracing/documentation/issues/553: Add usage documentation for Monitor tab UI

## Requirement - what kind of business use case are you trying to solve?
The main proposal is documented in: https://github.com/jaegertracing/jaeger/issues/2736.

The motivation is to help identify interesting traces (high qps, slow or erroneous) without knowing the service or operations up-front. 

Use cases include:
- Post deployment sanity checks across the org, or on known dependent services in the request chain.
- Monitoring and root-causing when alerted of an issue.
- Better onboarding experience for new users of Jaeger UI.
- Long-term trend analysis of QPS, errors and latencies.
- Capacity planning.

## Proposal - what do you suggest to solve the problem or improve the existing situation?

Add a new "Monitoring" tab situated after "Compare" containing service-level request rates, error rates, latency and impact (= `latency * request rate` to avoid "false positives" from low QPS endpoints with high latencies). 

The data will be sourced from [jaeger-query's new metrics endpoints](https://github.com/jaegertracing/jaeger/issues/2736).

As the jaeger-query metrics endpoints require opt-in to be enabled, the Monitor tab will have a sensible empty state, perhaps a link to documentation on how to enable metrics querying capabilities.

### Workflow
[...]

PR #2977, opened 2021-05-04
Add metrics reader interface and gen proto

albertteoh:

- Relates to #2954

- Adds the Metrics Reader interface bridging the gRPC and HTTP API handlers with the underlying metrics storage implementation.
- Adds the generated proto API and data model code used by the reader interface for returning the proto modelled metrics data as a response.


proto-gen/api_v2/service.pb.go
```diff
@@ -0,0 +1,2062 @@
+// Code generated by protoc-gen-gogo. DO NOT EDIT.
```
yurishkuro, line 1: `service.pb.go` is an odd name

storage/metricsstore/interface.go
```diff
@@ -0,0 +1,83 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package metricsstore
+
+import (
+	"context"
+	"time"
+
+	"github.com/jaegertracing/jaeger/proto-gen/api_v2"
```
yurishkuro, line 21: perhaps we should move the gen-types under api_v2/metrics?

albertteoh: +1

proto-gen/api_v2/service.pb.go
albertteoh: yeah, I'd originally tried `query.proto` but `protoc` complains about filenames shadowing one another with the existing `query.proto`, which is why I chose `service.proto` but agree it could be better named.

I've gone with `metricsquery.proto`. Let me know if you prefer another name.

yurishkuro (approved):

Issue #2923, opened 2021-04-06
Use of `startTimeMillis` for findTraces API , when es-aliases are used


sivatarunp:
## Requirement - what kind of business use case are you trying to solve?

Improve the query performance in Jaeger, for Elasticsearch storage, when --es.use-aliases is set to true

## Problem - what in Jaeger blocks you from solving the requirement?
      
Currently, when we use aliases for Elasticsearch storage, the findTraces API  queries all indices present under jaeger-span-read alias (irrespective of time range we give in UI). Due to this, when the data set is huge, significant amount of time is being used for querying unnecessary shards which are not in the given time range. 

## Proposal - what do you suggest to solve the problem or improve the existing situation?
    
The findTraces API use the `startTime` field for querying, which is a `long` field. Elasticsearch in built has a feature, to skip shards before querying, when the query is a range query on `date` field. 
    
https://discuss.elastic.co/t/timeline-query-on-timestamped-indices/129328/2 is the related discussion for the same.
    
Hence modifying the findTraces API , to use `startTimeMillis` field(which is already present in the data we store) which is a `date` type field, can help in skipping unnecessary shards hence improving the query performance

albertteoh:
@sivatarunp that's an interesting find; did you get a chance to test this out to see how much of an improvement this change would make to the FindTraces query?

sivatarunp:
@albertteoh . We have couple of things here.
1. Changing the timefield to startTimeMillis.
2. Adding a time range query with startTimeMillis field in the msearch query used to fetch individual trace data, in the findTraces page. 
We observed a huge difference, in query times. It's more predominantly seen, when we have the hot and warm nodes setup. Same query got timed out without the above changes

albertteoh:
@sivatarunp are you able to quantify the query time improvement by providing some numbers from your tests?

sivatarunp:
@albertteoh Here is a panel I could build from grafana community dashboards. The current query took more than 4m and even timed out, where as with above changes results were under 1 min, for the same data set[](url)
<img width="1425" alt="Screenshot 2021-04-19 at 10 24 25 AM" src="https://user-images.githubusercontent.com/41666755/115183677-e25d5380-a0f9-11eb-967b-b8482b77995f.png">

albertteoh:
[...]

PR #2978, opened 2021-05-04
Changed Range Query to use startTimeMillis date field instead of startTime field.

Sreevani871:
-  Solving https://github.com/jaegertracing/jaeger/issues/2923

If `es.use-aliases` cofiguration field is set to true, jaeger-span-read alias is used in Elasticsearch which is applying to all indexes. Using time range query Elasticsearch will skip the unnecessary shards while searching. At present in the code, the `startTime` field is used in range query but it is stored as `type: long` in Elasticsearch mapping. it's not skipping the unnecessary shards since Elasticsearch timeRange queries works with only `type: date` field as discussed in the issue: https://github.com/jaegertracing/jaeger/issues/2923. 
Using startTimeMillisField irrespective of `es.use-aliases: true` will improve the performance of the ES query.
-  Changed TimeRange query to use startTimeMillis field(type:date) instead of startTime(type:long).
-  Added time range term query to the ES query used in the multiRead method to avail the performance benefit.


plugin/storage/es/spanstore/reader.go
```diff
@@ -363,15 +364,18 @@ func (s *SpanReader) multiRead(ctx context.Context, traceIDs []model.TraceID, st
 		searchRequests := make([]*elastic.SearchRequest, len(traceIDs))
 		for i, traceID := range traceIDs {
 			query := buildTraceByIDQuery(traceID)
+			boolQuery := elastic.NewBoolQuery().Must(startTimeRangeQuery)
+			query = boolQuery.Must(query) // Adding time range query on startTimeMillis field.
```
yurishkuro, line 368: clean-up
```suggestion
			traceQuery := buildTraceByIDQuery(traceID)
			query := elastic.NewBoolQuery()
			    .Must(startTimeRangeQuery)
			    .Must(traceQuery)
```

plugin/storage/es/spanstore/reader.go
```diff
@@ -352,7 +353,7 @@ func (s *SpanReader) multiRead(ctx context.Context, traceIDs []model.TraceID, st
 	// i.e starts in one and ends in another.
 	indices := s.timeRangeIndices(s.spanIndexPrefix, s.indexDateLayout, startTime.Add(-time.Hour), endTime.Add(time.Hour))
 	nextTime := model.TimeAsEpochMicroseconds(startTime.Add(-time.Hour))
-
+	startTimeRangeQuery := s.buildStartTimeQuery(startTime.Add(-time.Hour), endTime.Add(time.Hour))
```
yurishkuro, line 356: Why +/- an hour and not some other interval?  Also, who determines startTime/endTime arguments to this function? The external API usually has just the traceID argument.

plugin/storage/es/spanstore/reader.go
```diff
@@ -363,15 +364,18 @@ func (s *SpanReader) multiRead(ctx context.Context, traceIDs []model.TraceID, st
...
-			s := s.sourceFn(query, nextTime)
-
-			searchRequests[i] = elastic.NewSearchRequest().
+			src := s.sourceFn(query, nextTime)
+			searchRequest := elastic.NewSearchRequest().
 				IgnoreUnavailable(true).
-				Source(s)
+				Source(src)
+
+			searchRequests[i] = searchRequest
```
yurishkuro, line 378: searchRequest: seems like unnecessary var. What was wrong with the original code?

plugin/storage/es/spanstore/reader.go
```diff
@@ -617,7 +621,10 @@ func (s *SpanReader) buildDurationQuery(durationMin time.Duration, durationMax t
 func (s *SpanReader) buildStartTimeQuery(startTimeMin time.Time, startTimeMax time.Time) elastic.Query {
 	minStartTimeMicros := model.TimeAsEpochMicroseconds(startTimeMin)
 	maxStartTimeMicros := model.TimeAsEpochMicroseconds(startTimeMax)
-	return elastic.NewRangeQuery(startTimeField).Gte(minStartTimeMicros).Lte(maxStartTimeMicros)
+	// startTimeMillisField is date field in ES mapping.
+	// Using date field in range queries helps to skip search on unnecessary shards at Elasticsearch side.
+	// https://discuss.elastic.co/t/timeline-query-on-timestamped-indices/129328/2
+	return elastic.NewRangeQuery(startTimeMillisField).Gte(minStartTimeMicros / 1000).Lte(maxStartTimeMicros / 1000)
```
yurishkuro, line 627: @albertteoh this is out of my domain knowledge of ES, PTAL

plugin/storage/es/spanstore/reader.go
Sreevani871: ` who determines startTime/endTime arguments to this function? The external API usually has just the traceID argument. 
`
Here multiRead operation is used by FindTraces API call and GetTrace API call, where FindTraces API calls include start and end time information as part of the request. So this information can be used to have range query as part of `_msearch ES query` while reading spans data of traces so that ES will skip unnecessary shards which are out of time range while processing the request at ES side.

In case of  GetTrace API, startTime is always maxSpanAge and endTime as current time passed as arguments to multiRead method call. Here startTimeRangeQuery is always static for GetTrace API calls, though it will have performance benefit if ES has data older than maxSpanAge.

`Why +/- an hour and not some other interval?`
In the existing code if use-aliases flag is not being set to true, while getting timeRangeIndices  +/- hour interval is added to start and end time to fetch the traces straddle two indexes are retrieved. So using the 1h tradeoff interval here also.

plugin/storage/es/spanstore/reader.go
Sreevani871: yes, internally we used it for logging so added the same here. Will remove this.

plugin/storage/es/spanstore/reader.go
albertteoh: IIUC this query searches for spans that reside in the `startTimeRangeQuery`. Can we handle traces that have spans outside of the +/- hour search window? I think it's technically possible to have traces that go on for many hours or even days, though unlikely.

albertteoh (commented):
@Sreevani871 please create a new PR against a branch once you've addressed the review comments. Creating the PR from master results in the CI build failures we're seeing in this PR.

Please also check if the unit tests are passing.

plugin/storage/es/spanstore/reader.go
Sreevani871: yes @albertteoh. If aliases are not used, timeRangeIndices are calculated based on start and end time in the existing code. So here also if the spans go on for many hours/days may not present in timeRangeIndices list. Correct me If am wrong here.

`Which API calls are affected by these changes?
`Here only FindTraces API calls will be affected, Since the response returned from FindTraces API is used to generate the list/summary page view in UI, even if spans having a longer duration greater than +/- hour are not part of FindTraces API response, only some of the parameters in the summary page UI such as span count and span count grouped by service are being affected. This tradeoff can be considered to improve query performance since we already have this tradeoff in the case of timeRangeIndices. What do you think?

Sreevani871:
> @Sreevani871 please create a new PR against a branch once you've addressed the review comments. Creating the PR from master results in the CI build failures we're seeing in this PR.
> 
> Please also check if the unit tests are passing.

Okay will do that

plugin/storage/es/spanstore/reader.go
albertteoh: What happens when the user clicks on a trace from the summary page view in the UI? Does the UI make another request to jaeger-query to fetch all spans for that trace (GetTrace API) or is everything loaded into memory from the FindTraces API call? 

From some local testing, I observed the latter behaviour; my GetTrace handler breakpoint wasn't hit when clicking on a trace from the summary page. This was perhaps introduced in the UI as a performance optimisation. However, if I refresh the specific trace, the UI does call the GetTrace API.

Sreevani871: Yes, FindTraces API call loads all spans for resulted TraceIds for a given timeRange in FindTraceIds sub call. So When the user clicks on a particular trace from the summary page, the UI is loading the spans of a particular trace received as part of the FindTraceAPI call instead of making an extra GetTrace API call. But on refresh or directly landing on GetTrace page(https://<IP:PORT>/trace/<TRACE_ID>) UI will make GetTrace API call to the backend. 

With the changes proposed in place, GetTrace API call queries for spans of a trace b/w maxSpanAge-1h to now+1h. Here we will get all spans of a trace.

albertteoh: Thanks for confirming @Sreevani871, this means that if a user clicks on a trace from the summary page, some spans could be missing leading to a malformed trace, is that correct? If so, I think this scenario would be undesirable.

From what we've discovered, this might be why a start-end time window isn't applied in the span-fetch query here because we need to search through all indices as we don't know the trace time boundary up front.

Sreevani871: @albertteoh As I mentioned earlier incase of es.use-aliases:false, in the code timeRangeIndices list is calculated based on startTime and EndTime https://github.com/jaegertracing/jaeger/blob/master/plugin/storage/es/spanstore/reader.go#L198-L209.
According to above code for say:  
```
startTime: <Timestamp1> => Date: 2021-04-20
endTime: <Timestamp2> => Date: 2021-04-22
Indices list ==> [jaeger-span-2021-04-20*, jaeger-span-2021-04-21*, jaeger-span-2021-04-22*]
```
Here while fetching spans for traces in FindTraces API, we are only searching in 3 indices, not searching all indices. Here also if some spans are present in jaeger-span-2021-04-19* and jaeger-span-2021-04-21* indices, we will see the malformed view of trace only when we click trace from summary view.

Since the above behaviour is already existed in the code in case of es.use-aliases:false, can the same tradeoff be used in case of aliases as well to improve the query performance. Here also we can have +/- 1d instead of +/- 1h. 

Why because, in the case of timeRangeIndices search is restricted to only some set of indices it will be very fast, but incase of jaeger-span-read alias it is pointing to all indices in ES and searching across all indices is very slow.

What do you say?

albertteoh: > As I mentioned earlier incase of es.use-aliases:false, in the code timeRangeIndices list is calculated based on startTime and EndTime

@Sreevani871 thanks for the clear explanation, that makes sense to me now, and sorry for not picking up on what you mentioned earlier.

> Since the above behaviour is already existed in the code in case of es.use-aliases:false, can the same tradeoff be used in case of aliases as well to improve the query performance.

I agree, I think the query performance improvement is worth the trade-off of missing spans for the rare case of a long-running trace, which as you rightly pointed out, is already the case for the non-alias mode.

> Here also we can have +/- 1d instead of +/- 1h.

Yes, I think +/- 1hr is not enough to at least be at parity with existing behaviour when es.use-aliases:false for the more common daily index rollovers.

Sreevani871: > Yes, I think +/- 1hr is not enough to at least be at parity with existing behaviour when es.use-aliases:false for the more common daily index rollovers.

Agree. Will change to +/- 1d.

Sreevani871:
Creating a new PR hence closing this.

PR #2979, opened 2021-05-05
Verify FindTraces() received a query

esnible:
Fixes a problem where FindTraces misbehaves if sent no query.


yurishkuro (approved):
nice! thanks

PR #2980, opened 2021-05-08
Changed Range Query to use startTimeMillis date field instead of startTime field

Sreevani871:
-  Solving https://github.com/jaegertracing/jaeger/issues/2923

If `es.use-aliases` cofiguration field is set to true, jaeger-span-read alias is used in Elasticsearch which is applying to all indexes. Using time range query Elasticsearch will skip the unnecessary shards while searching. At present in the code, the `startTime` field is used in range query but it is stored as `type: long` in Elasticsearch mapping. it's not skipping the unnecessary shards since Elasticsearch timeRange queries works with only `type: date` field as discussed in the issue: https://github.com/jaegertracing/jaeger/issues/2923. 
Using startTimeMillisField irrespective of `es.use-aliases: true` will improve the performance of the ES query.
-  Changed TimeRange query to use startTimeMillis field(type:date) instead of startTime(type:long).
-  Added time range term query to the ES query used in the multiRead method to avail the performance benefit.

Closed PR: https://github.com/jaegertracing/jaeger/pull/2978


plugin/storage/es/spanstore/reader.go
```diff
@@ -362,13 +368,22 @@ func (s *SpanReader) multiRead(ctx context.Context, traceIDs []model.TraceID, st
...
+			var query *elastic.BoolQuery
+			if s.useReadWriteAliases {
+				query = elastic.NewBoolQuery().
+					Must(startTimeRangeQuery).
+					Must(traceQuery)
+			} else {
+				query = elastic.NewBoolQuery().
+					Must(traceQuery)
+			}
+
```
albertteoh, line 381: We can simplify to:
```
                        query := elastic.NewBoolQuery().
				Must(traceQuery)
			if s.useReadWriteAliases {
				startTimeRangeQuery := s.buildStartTimeQuery(startTime.Add(-time.Hour*24), endTime.Add(time.Hour*24))
				query = query.Must(startTimeRangeQuery)
			}
```

albertteoh (approved):

bhiravabhatla:
This actually brought our average search time on Jaeger UI by 1/10. For search all the traces for a particular service/operation in last 2 hours - it used to take 30 secs on average [ We use ES backend and aliases ]. It is down to 3 seconds now. Awesome work!

PR #2988, opened 2021-05-12
Implement Prometheus metrics reader constructor

albertteoh:

- Relates to #2954.

- Adds the Prometheus metrics reader constructor implementation and tests.


plugin/metrics/prometheus/metricsstore/reader.go
```diff
@@ -16,22 +16,45 @@ package metricsstore
...
 // NewMetricsReader returns a new MetricsReader, assigning the first reachable host:port from the provided list.
 // This host:port forms part of the URL to call when making queries to the underlying metrics store.
 func NewMetricsReader(logger *zap.Logger, hostPorts []string, connTimeout time.Duration) (*MetricsReader, error) {
-	// TODO: Implement me
-	return &MetricsReader{}, nil
+	if len(hostPorts) < 1 {
+		return nil, fmt.Errorf("no prometheus query host:port provided")
+	}
+	errs := make([]error, 0)
+	for _, hostPort := range hostPorts {
```
yurishkuro, line 43: This is kind of an odd behavior, why do we need to loop through hosts looking for a reachable one? Why wouldn't they all be reachable? Or to put another way, what's the point of providing unreachable host:port pairs as input?

If these addresses are somehow "unstable", then they might as well become unreachable _after_ the reader is created.

albertteoh: Thanks for pointing this out @yurishkuro.

The initial motivation was to maintain config consistency with ES and Cassandra span stores in how they both supply multiple server URLs, presumably for redundancy but yes, good call that the selected address could be unreachable after the reader is created.

This constructor behaviour was based on following ES's lead in how it initiates a new client by health checking each server: https://github.com/olivere/elastic/blob/v6.2.35/client.go#L1153.

However, the piece I'm missing which you correctly picked up on, is maintaining this pool of connections and finding the first healthy connection on each request, for example: https://github.com/olivere/elastic/blob/v6.2.35/client.go#L1301

The trade-off here is that it's a fair bit more complicated.

What direction do you suggest we take? Have a simple single host:port configuration? Or support a pool of connections that we select on each request?

yurishkuro: Does Prom only expose HTTP API, or is there gRPC version? gRPC would've taken care of the connection pooling.

albertteoh: To the best of my knowledge, Prometheus only exposes an [HTTP API](https://prometheus.io/docs/prometheus/latest/querying/api/). 

I found some notes in Prometheus docs referring to [using gRPC for remote storage reads/writes in the future](https://prometheus.io/docs/prometheus/latest/storage/#overview), but that's not related to the query API.

albertteoh: Another alternative is to use the Prometheus go client; but it means including another dependency.

Worth noting is this client takes a single URL: https://github.com/prometheus/client_golang/blob/master/api/client.go#L78

Given we're only using in a single Prometheus endpoint, I feel the Prometheus client offers more than we need and may be a bit of an overkill.
 
At the same time, we can adopt this client's approach of just a single connection URL, which I assume would be the more idiomatic given it's an official Prometheus client.

yurishkuro: aren't we already using prom client for metrics?

albertteoh: Yes, you're right. However, the [README](https://github.com/prometheus/client_golang#client-for-the-prometheus-http-api) says that the client used to query the HTTP API is still in alpha:

> The api/prometheus directory contains the client for the Prometheus HTTP API. It allows you to write Go applications that query time series data from a Prometheus server. It is still in alpha stage.

albertteoh: If we want to proceed with the Prom client library, I think it would be best we modify our [metrics data model](https://github.com/jaegertracing/jaeger/tree/master/model/proto/metrics) to match prometheus' to avoid mapping between OpenTelemetry and Prometheus' data models. I guess this might mean using [OpenMetrics](https://github.com/OpenObservability/OpenMetrics/blob/main/proto/openmetrics_data_model.proto), unless if there's a good reason to stick with OpenTelemetry's.

Otherwise, should we maintain the simple http client as we have now, configured with a single Prom address?

yurishkuro: I don't have a strong opinion. Which of the two formats more mature at this point? I thought OTLP was still changing its proto.

yurishkuro: We just discussed this in the Jaeger call. General agreement was that OpenMetrics would be a better choice:
  * OTLP is still changing, and we already took a copy of the data model which might be different now
  * OTLP is supposed to be fully compatible with OpenMetrics by the end of the year
  * OpenMetrics is the de-factor Prometheus format used by many different backends already
  * OpenMetrics is stable
  * OTEL will support OTLP<->OpenMetrics conversion in the future so we would still be able to implement support for OTLP-native backends

albertteoh: Okay, thanks for raising this in the Jaeger call. Those reasons make sense so I'll make the switch over to OpenMetrics.

It also sounds like we're okay with using the Prom client for HTTP queries (even though it's in alpha), but please correct me if my assumption is wrong.

yurishkuro: > It also sounds like we're okay with using the Prom client for HTTP queries

sgtm

albertteoh:
PR updated to use prom client and OpenMetrics.

Did an initial e2e test with a POC version and the prom client looks okay so far. Still a bit of work involved to convert from the prom client's data model to OpenMetrics, and couldn't find a library that does this conversion, which was a little surprising. It wasn't too difficult in any case.

plugin/metrics/prometheus/metricsstore/reader.go
```diff
@@ -18,38 +18,53 @@ import (
...
+	}
+	logger.Info("Prometheus reader initialized", zap.String("addr", hostPort))
+	return mr, nil
 }
 
 // GetLatencies gets the latency metrics for the given set of latency query parameters.
-func (m *MetricsReader) GetLatencies(ctx context.Context, params *metricsstore.LatenciesQueryParameters) ([]metrics.Metric, error) {
+func (m *MetricsReader) GetLatencies(ctx context.Context, params *metricsstore.LatenciesQueryParameters) (mf metrics.MetricFamily, err error) {
 	// TODO: Implement me
-	return nil, nil
```
yurishkuro, line 40: nit: we try to stay away from named return values (imo one of the language features where Go really dropped the ball on readability)

yurishkuro (dismissed):

model/proto/metrics/openmetrics.proto
```diff
@@ -0,0 +1,242 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// The OpenMetrics protobuf schema which defines the protobuf wire format.
+// Ensure to interpret "required" as semantically required for a valid message.
+// All string fields MUST be UTF-8 encoded strings.
+// Based on: https://github.com/OpenObservability/OpenMetrics/blob/v1.0.0/proto/openmetrics_data_model.proto
```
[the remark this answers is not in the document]
yurishkuro: describing additions in readme is a bit removed. If there is a concrete & single block of text that's been added compared to official file, I would mark it with comments in the file directly. When looking at the source file, it's convenient to have it be self-descriptive, e.g.
```
// This file is a copy of {path to source}, with the following additions 
//  * ...
//  * ...

// -- BEGIN ADDITIONS
...
// -- END ADDITIONS
```

plugin/metrics/prometheus/metricsstore/reader.go
```diff
@@ -16,40 +16,68 @@ package metricsstore
...
+	return metrics.MetricFamily{}, nil
 }
 
 // GetCallRates gets the call rate metrics for the given set of call rate query parameters.
-func (m *MetricsReader) GetCallRates(ctx context.Context, params *metricsstore.CallRateQueryParameters) ([]metrics.Metric, error) {
+func (m *MetricsReader) GetCallRates(ctx context.Context, params *metricsstore.CallRateQueryParameters) (metrics.MetricFamily, error) {
 	// TODO: Implement me
-	return nil, nil
+	return metrics.MetricFamily{}, nil
```
yurishkuro, line 74: Any reason to return MetricFamily by value instead of a pointer?

yurishkuro (dismissed):

plugin/metrics/prometheus/metricsstore/reader.go
albertteoh: No strong reason, it's just a continuation from the API definition which I defined as not nullable.

Perhaps it's better to pass by pointer for performance reasons?

albertteoh: and the reason for defining `MetricFamily` as not nullable is to act as a forcing-function to return an empty result for both cases where there is no data in the result as well as an error, which I think is more meaningful (as it also contains meta-data), rather than `nil`.

plugin/metrics/prometheus/metricsstore/reader_test.go
```diff
@@ -25,17 +25,28 @@ import (
...
-	reader, err := NewMetricsReader(logger, nil, time.Second)
+	reader, err := NewMetricsReader(logger, "localhost:1234", defaultTimeout)
 	assert.NoError(t, err)
 	assert.NotNil(t, reader)
 }
 
+func TestNewMetricsReaderInvalidAddress(t *testing.T) {
+	logger := zap.NewNop()
+	reader, err := NewMetricsReader(logger, "\n", defaultTimeout)
+	const wantErrMsg = `parse "http://\n": net/url: invalid control character in URL`
```
yurishkuro, line 40: it's better not to depend on exact error messages from stdlib, but wrap the `error.Wrap` the error with our own message and assert only on our own string, e.g.:
```
require.Error(t, err)
assert.Contains(t, expectedMsg, err.Error())
```

yurishkuro (approved):

plugin/metrics/prometheus/metricsstore/reader_test.go
albertteoh: yup, I'll address this in the next PR, thanks!

Issue #3019, opened 2021-05-20
Support specifying cipher suites in tls connection


jannyHou:
## Requirement - what kind of business use case are you trying to solve?

Hello team, in PR https://github.com/jaegertracing/jaeger/pull/2798 the tls connection is enabled in collector, it would be good to support configuring `cipher` for it.

It would be similar to the fluentd tls settings in https://docs.fluentd.org/configuration/transport-section#tls-setting, you can config the `ciphers` field.

## Problem - what in Jaeger blocks you from solving the requirement?

Currently people could not configure the `ciphers` field.

## Proposal - what do you suggest to solve the problem or improve the existing situation?

Same as the requirement

## Any open questions to address

n/a

clock21am:
@yurishkuro can I take this?

yurishkuro:
It's yours

yurishkuro:
I wasn't too familiar with cipher suites, but now that I read about them, I am questioning why we need this.

1) The documentation says that they are only supported up to TLS 1.2, phased out in TLS 1.3. The default behavior is to use whichever ciphers the server supports.

2) I don't see an established format of describing the cipher family. The Go struct uses an array of uint16's, Fluentd doesn't describe the format but gives a weird example `"ALL:!aNULL:!eNULL:!SSLv2"`, other sources show strings like `"ECDHE-ECDSA-AES256-GCM-SHA384"`

Ashmita152:
>  why we need this

In my understanding, some organisations fulfil certain compliances (especially fintech regulated by operating country's law) that have strict requirements that all SSL communication should happen only within certain ciphers. In those cases, this option will be quite useful to specify as a config param.

clock21am:
make sense @Ashmita152

jpkrohling:
@yurishkuro, the server provides a list of ciphers to clients, and cipher negotiation happens. The key is to strike a balance between modern features and old clients. For instance, there might be older Java clients supporting TLS 1.1 only. Your example of Fluentd is basically filtering out classes of ciphers, while the last example is an example of a concrete cipher that would be part of a class. They should both be compatible.

Allowing the configuration of ciphers would allow admins to adjust the balance I mentioned before: if they are sure all their applications are capable of using modern ciphers, they can remove older, less safe ones. This way, a man-in-the-middle would have a harder time performing a downgrade attack.

jpkrohling:
>They should both be compatible.

[...]

PR #3026, opened 2021-05-22
Add: support specifying cipher suites in tls connection #3019

clock21am:
- Add  support specifying cipher suites in tls connection #3019

- Add cipher parameter in tls config

### DIFF
--- a/pkg/config/tlscfg/options.go
+++ b/pkg/config/tlscfg/options.go
@@ -21,6 +21,7 @@ import (
 	"io"
 	"io/ioutil"
 	"path/filepath"
+	"strconv"
 
 	"go.uber.org/zap"
 )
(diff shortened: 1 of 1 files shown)


clock21am:
@yurishkuro Let me know what do you think?

pkg/config/tlscfg/options.go
```diff
@@ -18,6 +18,7 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"fmt"
+	"github.com/coreos/etcd/pkg/tlsutil"
```
yurishkuro, line 21: I don't think we want an external dependency for such minor functionality. Also, I am not seeing this package in https://github.com/etcd-io/etcd/tree/main/pkg

clock21am: @yurishkuro removed this external dependency and made it a sample. Hope this makes sense.

clock21am: as it was mentioned in this document  https://docs.fluentd.org/configuration/transport-section#tls-setting do you want to provide some default cipher in the repo itself?

clock21am: and one more doubt why naming ciphers if want to pass just string ? are we expecting multiple ciphers ?

yurishkuro:
I have questions on the original ticket.

jpkrohling:
What's the status of this PR?

clock21am:
@jpkrohling there were questions on the original tickets have to look but haven’t got the chance can take a look in the evening though

yurishkuro:
Already addressed by #3564

PR #3030, opened 2021-05-25
Perform log.fatal if TLS flags are used when tls.enabled=false #2893

### LINKED ISSUE #2893: Error/warn if TLS flags are used when tls.enabled=false

Motivated by [this comment in Slack](https://cloud-native.slack.com/archives/CGG7NFUJ3/p1616518570051700):
> we tried the CLI flag --es.tls.skip-host-verify and the enviornment variable ES_TLS_SKIP_HOST_VERIFY
> Neither one seems to have an effect
> ...
> Found the issue, you also need to set --es.tls.enabled/ES_TLS_ENABLED 🤦 

The user was, rightfully, confused when the `--es.tls.skip-host-verify` was set but wasn't working.

Log a warning or error if any `--*.tls.*` flag is set when `--*.tls.enabled == false`, maybe somewhere [here](https://github.com/jaegertracing/jaeger/blob/master/pkg/config/tlscfg/flags.go#L73) and [here](https://github.com/jaegertracing/jaeger/blob/master/pkg/config/tlscfg/flags.go#L89).

Should we log a warning or error (and prevent startup of service)? I would lean towards the latter to be more explicit, though I think it would be considered "breaking" behaviour. The former warning could be easily missed.

clock21am:
Signed-off-by: Rajdeep Kaur <rajdeep51994@gmail.com>

-  Resolves #2893 Terminate the program if tls.enabled=false

-  Added log.fatal if tls.enabled=false for client as well as server


pkg/config/tlscfg/flags.go
```diff
@@ -74,6 +75,8 @@ func (c ClientFlagsConfig) InitFromViper(v *viper.Viper) Options {
 	var p Options
 	if c.ShowEnabled {
 		p.Enabled = v.GetBool(c.Prefix + tlsEnabled)
+	} else {
+		log.Fatalf("ES_TLS_ENABLED is set to false unable to talk to server(s)")
```
yurishkuro, line 79: We try not to use std log package, as it makes the code not-testable. The preferred way is to return an error and make sure it is logged upstream where the logger is already available / initialized.

clock21am: but don't we want to crash instead of logging or we can log it from the logger and then crash it?

yurishkuro: the top-level logger (e.g. somewhere in main) can log fatal, but all other packages should be returning errors

clock21am: okay

clock21am:
@yurishkuro let me know what you think now

cmd/all-in-one/main.go
```diff
@@ -119,8 +119,14 @@ by default uses only in-memory database.`,
 			aOpts := new(agentApp.Builder).InitFromViper(v)
 			repOpts := new(agentRep.Options).InitFromViper(v, logger)
 			grpcBuilder := agentGrpcRep.NewConnBuilder().InitFromViper(v)
-			cOpts := new(collectorApp.CollectorOptions).InitFromViper(v)
-			qOpts := new(queryApp.QueryOptions).InitFromViper(v, logger)
+			cOpts, err := new(collectorApp.CollectorOptions).InitFromViper(v)
+			if err != nil {
+				logger.Fatal("Incorrect value of flag tls.enabled", zap.Error(err))
```
yurishkuro, line 124: the message should be "Failed to initialize collector", because we do not know the reason why it failed (it will be contained in the `err` which will be printed separately

cmd/collector/app/builder_flags.go
```diff
@@ -102,8 +103,15 @@ func (cOpts *CollectorOptions) InitFromViper(v *viper.Viper) *CollectorOptions {
 	cOpts.DynQueueSizeMemory = v.GetUint(collectorDynQueueSizeMemory) * 1024 * 1024 // we receive in MiB and store in bytes
 	cOpts.NumWorkers = v.GetInt(collectorNumWorkers)
 	cOpts.QueueSize = v.GetInt(collectorQueueSize)
-	cOpts.TLSGRPC = tlsGRPCFlagsConfig.InitFromViper(v)
-	cOpts.TLSHTTP = tlsHTTPFlagsConfig.InitFromViper(v)
-
-	return cOpts
+	TLSGRPC, err := tlsGRPCFlagsConfig.InitFromViper(v)
+	if err != nil {
+		return cOpts, errors.New("ES_TLS_ENABLED is set to false unable to talk to server(s)")
```
yurishkuro, line 108: same issue - the message is "Failed to initialize TLS", we don't know why at this point. And you need to capture the original error (https://blog.golang.org/go1.13-errors):
```
fmt.Errorf("Failed to initialize TLS: %w", err)
```

cmd/collector/app/builder_flags.go
```diff
@@ -102,8 +103,15 @@ func (cOpts *CollectorOptions) InitFromViper(v *viper.Viper) *CollectorOptions {
 	cOpts.DynQueueSizeMemory = v.GetUint(collectorDynQueueSizeMemory) * 1024 * 1024 // we receive in MiB and store in bytes
 	cOpts.NumWorkers = v.GetInt(collectorNumWorkers)
 	cOpts.QueueSize = v.GetInt(collectorQueueSize)
-	cOpts.TLSGRPC = tlsGRPCFlagsConfig.InitFromViper(v)
-	cOpts.TLSHTTP = tlsHTTPFlagsConfig.InitFromViper(v)
-
-	return cOpts
+	TLSGRPC, err := tlsGRPCFlagsConfig.InitFromViper(v)
```
yurishkuro, line 106: I would recommend this syntax:
```
if tls, err := tlsGRPCFlagsConfig.InitFromViper(v); err == nil {
    cOpts.TLSGRPC = tls
} else {
    return ...
}
```

cmd/query/app/flags_test.go
```diff
@@ -41,7 +41,7 @@ func TestQueryBuilderFlags(t *testing.T) {
 		"--query.additional-headers=whatever:thing",
 		"--query.max-clock-skew-adjustment=10s",
 	})
-	qOpts := new(QueryOptions).InitFromViper(v, zap.NewNop())
+	qOpts, _ := new(QueryOptions).InitFromViper(v, zap.NewNop())
```
yurishkuro, line 44:
```suggestion
	qOpts, err := new(QueryOptions).InitFromViper(v, zap.NewNop())
	require.NoError(t, err)
```

pkg/config/tlscfg/flags.go
```diff
@@ -70,10 +70,12 @@ func (c ServerFlagsConfig) AddFlags(flags *flag.FlagSet) {
 }
 
 // InitFromViper creates tls.Config populated with values retrieved from Viper.
-func (c ClientFlagsConfig) InitFromViper(v *viper.Viper) Options {
+func (c ClientFlagsConfig) InitFromViper(v *viper.Viper) (Options, error) {
 	var p Options
 	if c.ShowEnabled {
 		p.Enabled = v.GetBool(c.Prefix + tlsEnabled)
+	} else {
+		return p, errors.New("ES_TLS_ENABLED is set to false unable to talk to server(s)")
```
yurishkuro, line 78: I think the point of the issue was to raise an error if enabled=false but **other flags are present**. There's nothing wrong with enabled=false by itself. Also, this is not specific to ES, if you want to refer to the actual property it would be `c.Prefix + tlsEnabled`

cmd/all-in-one/main.go
clock21am: yeah i was thinking the same thanks :)

pkg/config/tlscfg/flags.go
clock21am: noted

clock21am:
@yurishkuro let me know what you think now

cmd/agent/main.go
```diff
@@ -60,7 +61,10 @@ func main() {
 				baseFactory)
 
 			rOpts := new(reporter.Options).InitFromViper(v, logger)
-			grpcBuilder := grpc.NewConnBuilder().InitFromViper(v)
+			grpcBuilder, err := grpc.NewConnBuilder().InitFromViper(v)
+			if err != nil {
+				log.Fatal("Failed  to initialize grpc connection: %w", err)
```
yurishkuro, line 66:
```suggestion
				logger.Fatal("Failed  to initialize grpc connection", zap.Error(err))
```

cmd/agent/app/reporter/grpc/flags.go
```diff
@@ -46,13 +46,17 @@ func AddFlags(flags *flag.FlagSet) {
...
 	hostPorts := v.GetString(collectorHostPort)
 	if hostPorts != "" {
 		b.CollectorHostPorts = strings.Split(hostPorts, ",")
 	}
 	b.MaxRetry = uint(v.GetInt(retry))
-	b.TLS = tlsFlagsConfig.InitFromViper(v)
 	b.DiscoveryMinPeers = v.GetInt(discoveryMinPeers)
-	return b
+	if TLS, err := tlsFlagsConfig.InitFromViper(v); err != nil {
+		return b, err
```
yurishkuro, line 57: * not sure if this will pass the linter, if not you can swap the condition and do a return in the `else` part.
* it looks odd when local var is upper case, use `tls` or `cfg`

cmd/collector/app/builder_flags_test.go
```diff
@@ -15,6 +15,7 @@
 package app
 
 import (
+	"github.com/stretchr/testify/require"
```
yurishkuro, line 18: did you run `make lint`?

pkg/config/tlscfg/flags.go
```diff
@@ -70,31 +70,35 @@ func (c ServerFlagsConfig) AddFlags(flags *flag.FlagSet) {
...
 	p.KeyPath = v.GetString(c.Prefix + tlsKey)
 	if c.ShowServerName {
 		p.ServerName = v.GetString(c.Prefix + tlsServerName)
 	}
 	p.SkipHostVerify = v.GetBool(c.Prefix + tlsSkipHostVerify)
-	return p
+	if c.ShowEnabled {
+		p.Enabled = v.GetBool(c.Prefix + tlsEnabled)
+	} else {
+		return p, fmt.Errorf("%s has been set to False", c.Prefix+tlsEnabled)
```
yurishkuro, line 85: * `c.ShowEnabled == false` doesn't mean that tls was disabled, it means that the CLI option does not need to be present
* even if TLS is disabled (`p.Enable == false`), doesn't mean that this is an error.  The error is when it is false but any other flags are non-empty: `(!p.Enabled || !c.ShowEnabled) && anyOtherFieldIsNotEmpty`

clock21am:
@yurishkuro ran lint and tests let me know what you think now

plugin/metrics/prometheus/options.go
```diff
@@ -68,11 +68,17 @@ func (opt *Options) AddFlags(flagSet *flag.FlagSet) {
 }
 
 // InitFromViper initializes the options struct with values from Viper.
-func (opt *Options) InitFromViper(v *viper.Viper) {
+func (opt *Options) InitFromViper(v *viper.Viper) error {
 	cfg := &opt.Primary
 	cfg.ServerURL = stripWhiteSpace(v.GetString(cfg.namespace + suffixServerURL))
 	cfg.ConnectTimeout = v.GetDuration(cfg.namespace + suffixConnectTimeout)
-	cfg.TLS = cfg.getTLSFlagsConfig().InitFromViper(v)
+	if tls, err := cfg.getTLSFlagsConfig().InitFromViper(v); err == nil {
```
albertteoh, line 75: The failing unit test job is due to [linter errors](https://github.com/jaegertracing/jaeger/pull/3030/checks?check_run_id=2871465877#step:5:59), the majority of which are:

> if block ends with a return statement, so drop this else and outdent its block (move short variable declaration to its own line if necessary)

Whether if the linter is right about this being a good style or not [is debatable](https://github.com/golang/lint/issues/17#issuecomment-344050288). :)

One option is:
```
	tls, err := cfg.getTLSFlagsConfig().InitFromViper(v)
	if err != nil {
		return err
	}
	cfg.TLS = tls
	return nil
```

To check for linter errors, run `make lint`.

clock21am: @albertteoh don't think see this is the output of make lint and it does point to any of my changes 

` make lint
time staticcheck ./... \
                | grep -v \
                        -e model/model.pb.go \
                        -e proto-gen \
                        -e _test.pb.go \
                        -e thrift-gen/ \
                        -e swagger-gen/ \
                >> .lint.log || true

real    0m5.547s
user    0m5.165s
sys     0m7.180s
time gosec -quiet -exclude=G104,G107 ./...
Results:


[/Users/rkaur/github/jaeger/plugin/sampling/strategystore/adaptive/processor.go:226] - G404 (CWE-338): Use of weak random number generator (math/rand instead of crypto/rand) (Confidence: MEDIUM, Severity: HIGH)
    225: func addJitter(jitterAmount time.Duration) {
  > 226:        delay := (jitterAmount / 2) + time.Duration(rand.Int63n(int64(jitterAmount/2)))
    227:        time.Sleep(delay)



[/Users/rkaur/github/jaeger/pkg/discovery/grpcresolver/grpc_resolver.go:68] - G404 (CWE-338): Use of weak random number generator (math/rand instead of crypto/rand) (Confidence: MEDIUM, Severity: HIGH)
    67:         seed := time.Now().UnixNano()
  > 68:         random := rand.New(rand.NewSource(seed))
    69:         r := &Resolver{



[/Users/rkaur/github/jaeger/crossdock/services/tracehandler.go:308] - G404 (CWE-338): Use of weak random number generator (math/rand instead of crypto/rand) (Confidence: MEDIUM, Severity: HIGH)
    307:        // A random 8-byte hex
  > 308:        return fmt.Sprintf("%x", rand.Int63())
    309: }



[/Users/rkaur/github/jaeger/cmd/ingester/app/processor/decorator/retry.go:54] - G404 (CWE-338): Use of weak random number generator (math/rand instead of crypto/rand) (Confidence: MEDIUM, Severity: HIGH)
    53:         propagateError: false,
  > 54:         rand:           rand.New(rand.NewSource(time.Now().UnixNano())),
    55: }



[/Users/rkaur/github/jaeger/cmd/collector/app/sanitizer/cache/auto_refresh_cache.go:136] - G404 (CWE-338): Use of weak random number generator (math/rand instead of crypto/rand) (Confidence: MEDIUM, Severity: HIGH)
    135: func getRandomSleepTime(interval time.Duration) time.Duration {
  > 136:        return (interval / 2) + time.Duration(rand.Int63n(int64(interval/2)))
    137: }



[/Users/rkaur/github/jaeger/cmd/anonymizer/app/uiconv/reader.go:40] - G304 (CWE-22): Potential file inclusion via variable (Confidence: HIGH, Severity: MEDIUM)
    39: func NewReader(capturedFile string, logger *zap.Logger) (*Reader, error) {
  > 40:         cf, err := os.OpenFile(capturedFile, os.O_RDONLY, os.ModePerm)
    41:         if err != nil {



[/Users/rkaur/github/jaeger/cmd/anonymizer/app/uiconv/extractor.go:38] - G304 (CWE-22): Potential file inclusion via variable (Confidence: HIGH, Severity: MEDIUM)
    37: func NewExtractor(uiFile string, traceID string, reader *Reader, logger *zap.Logger) (*Extractor, error) {
  > 38:         f, err := os.OpenFile(uiFile, os.O_CREATE|os.O_WRONLY, os.ModePerm)
    39:         if err != nil {



[/Users/rkaur/github/jaeger/model/converter/thrift/jaeger/from_domain.go:106] - G601 (CWE-118): Implicit memory aliasing in for loop. (Confidence: MEDIUM, Severity: MEDIUM)
    105:        for idx, kv := range kvs {
  > 106:                jaegerTags[idx] = d.keyValueToTag(&kv)
    107:        }



[/Users/rkaur/github/jaeger/cmd/anonymizer/main.go:72] - G601 (CWE-118): Implicit memory aliasing in for loop. (Confidence: MEDIUM, Severity: MEDIUM)
    71:                         for _, span := range spans {
  > 72:                                 writer.WriteSpan(&span)
    73:                         }



Summary:
  Gosec  : dev
  Files  : 441
  Lines  : 73829
  Nosec  : 10
  Issues : 9

      164.77 real       166.34 user       224.53 sys
make: *** [lint-gosec] Error 1
`

clock21am: @albertteoh ooh it is `make go-lint`

albertteoh: `make lint` should run `make go-lint`, but it looks like it's `gosec -quiet -exclude=G104,G107 ./...` that's failing for you before it gets to call `make go-lint`. This is odd, because `gosec` passes in CI (and for me as well).

Did you run `make install-tools` to install `gosec`? `which gosec` should point to `$GOPATH/bin/gosec`.

clock21am: installed explicitly gosec but didn't run `make install-tools`

clock21am: it should show a warning now should not show failing

clock21am: ooh less than 95 got it

albertteoh (commented):
To help with identifying test coverage gaps locally, I know of a few options (folks reading this, please feel free to add other suggestions):
- your IDE should support executing tests with coverage.
- run `make cover` from the project root. This checks coverage for the entire project so it takes some time.
- I personally like to use a command that I install in .zshrc that is like `make cover` but runs from a path relative to the current dir, so it's much quicker. Happy to share that if you prefer this option.

pkg/config/tlscfg/flags.go
```diff
@@ -70,31 +71,61 @@ func (c ServerFlagsConfig) AddFlags(flags *flag.FlagSet) {
 }
 
 // InitFromViper creates tls.Config populated with values retrieved from Viper.
-func (c ClientFlagsConfig) InitFromViper(v *viper.Viper) Options {
+func (c ClientFlagsConfig) InitFromViper(v *viper.Viper) (Options, error) {
 	var p Options
+	var checkAnyFieldIsEmpty bool
 	if c.ShowEnabled {
```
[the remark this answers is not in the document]
yurishkuro: They weren't always true, some settings did not apply to some use cases. This may have changed since then.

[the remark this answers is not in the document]
albertteoh: @clock21am The `Show*` flags have been removed in #3106 which should simplify efforts for you in this PR. Please merge these changes in from master.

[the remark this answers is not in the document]
albertteoh: and thanks for confirming, Yuri.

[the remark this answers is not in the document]
clock21am: Thanks for the review @albertteoh @yurishkuro i will rebase my branch and fix it

pkg/kafka/auth/config.go
```diff
@@ -95,7 +95,7 @@ func (config *AuthenticationConfig) InitFromViper(configPrefix string, v *viper.
 		ShowServerName: true,
 	}
 
-	config.TLS = tlsClientConfig.InitFromViper(v)
+	config.TLS, _ = tlsClientConfig.InitFromViper(v)
```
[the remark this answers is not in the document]
clock21am: because it has been set to true later https://github.com/jaegertracing/jaeger/blob/master/pkg/kafka/auth/config.go#L96

pkg/config/tlscfg/flags.go
```diff
@@ -60,23 +61,31 @@ func (c ServerFlagsConfig) AddFlags(flags *flag.FlagSet) {
...
 	p.Enabled = v.GetBool(c.Prefix + tlsEnabled)
 	p.CAPath = v.GetString(c.Prefix + tlsCA)
 	p.CertPath = v.GetString(c.Prefix + tlsCert)
 	p.KeyPath = v.GetString(c.Prefix + tlsKey)
 	p.ServerName = v.GetString(c.Prefix + tlsServerName)
 	p.SkipHostVerify = v.GetBool(c.Prefix + tlsSkipHostVerify)
-	return p
+	areTLSFlagsUsed := v.IsSet(c.Prefix+tlsCA) || v.IsSet(c.Prefix+tlsCert) || v.IsSet(c.Prefix+tlsKey) || v.IsSet(c.Prefix+tlsServerName)
+	if (!p.Enabled) && (p.SkipHostVerify && areTLSFlagsUsed) {
+		return p, fmt.Errorf("%s has been disable", c.Prefix+tlsEnabled)
```
yurishkuro, line 74:
```suggestion
		return p, fmt.Errorf("TLS parameters do not apply when %s=false", c.Prefix+tlsEnabled)
```

cmd/query/app/flags_test.go
```diff
@@ -162,11 +166,84 @@ func TestQueryOptionsPortAllocationFromFlags(t *testing.T) {
...
+	for _, test := range flagPortCases {
+		t.Run(test.name, func(t *testing.T) {
+			v, command := config.Viperize(AddFlags)
+			command.ParseFlags(test.flagsArray)
+			v.Set("query.http.tls.enabled", "false")
+			v.Set("query.http.tls.cert", "abc")
+			v.Set("query.http.tls.ca", "def")
+			v.Set("query.http.tls.key", "xyz")
+			_, err := new(QueryOptions).InitFromViper(v, zap.NewNop())
+			require.Error(t, err, "query.http.tls.enabled has been disable")
```
yurishkuro, line 246: This does not test for the type of error, only that it is not nil (the last argument is only used when logging a failure).

If you want to test for the actual error, I would suggest this:
```
require.Error(t, err)
assert.Contains(t, err.Error(), "some relevant substring of the message")
```

jpkrohling:
What's the status of this PR?

clock21am:
Hey

> What's the status of this PR?

hey will start working on it again..somehow lost it will update it today.

yurishkuro (approved):

yurishkuro:
@clock21am thank you!

PR #3038, opened 2021-05-27
Report unit test result on pull request

### LINKED ISSUE #2859: Publish test report
One feature I always missed from Jenkins was the ability to easily view the test report for a specific run, and apparently, there's a good candidate for this feature being proposed for the otelcol repository:

https://github.com/open-telemetry/opentelemetry-collector/pull/2531
https://github.com/EnricoMi/publish-unit-test-result-action

We should try it out in this repository as well.

ddmendes:
From request jaegertracing#2859
Uses gotest.tools/gotestsum to convert go test report to a compatible
junit xml report.
Uses github action EnricoMi/publish-unit-test-result-action to report
on pull request unit tests results.

Signed-off-by: Davi Diorio Mendes <d.mendes@sidi.org.br>

- Resolves #2859 

- Convert unit test report to junit xml report style
- Publish this report with EnricoMi/publish-unit-test-result-action

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,6 +35,7 @@ GOOS ?= $(shell go env GOOS)
 GOARCH ?= $(shell go env GOARCH)
 GOBUILD=CGO_ENABLED=0 installsuffix=cgo go build -trimpath
 GOTEST=go test -v $(RACE)
+GOTESTREPORT=gotestsum --junitfile test-report.xml -- -v $(RACE)
 GOLINT=golint
 GOVET=go vet
 GOFMT=gofmt
@@ -145,7 +146,7 @@ all-srcs:
 
 .PHONY: cover
 cover: nocover
-	$(GOTEST) -timeout 5m -coverprofile cover.out ./...
+	$(GOTESTREPORT) -timeout 5m -coverprofile cover.out ./...
 	grep -E -v 'model.pb.*.go' cover.out > cover-nogen.out
 	mv cover-nogen.out cover.out
 	go tool cover -html=cover.out -o cover.html
@@ -406,6 +407,7 @@ install-tools:
 	go install github.com/mjibson/esc
 	go install github.com/securego/gosec/cmd/gosec
 	go install honnef.co/go/tools/cmd/staticcheck
+	go install gotest.tools/gotestsum
 
 .PHONY: install-ci
 install-ci: install-tools
--- a/go.mod
+++ b/go.mod
@@ -11,7 +11,6 @@ require (
 	github.com/crossdock/crossdock-go v0.0.0-20160816171116-049aabb0122b
 	github.com/dgraph-io/badger v1.6.2
 	github.com/dgryski/go-farm v0.0.0-20200201041132-a6ae2369ad13 // indirect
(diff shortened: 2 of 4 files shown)


yurishkuro:
Doesn't seem like it can publish the test results on PRs:
```
2021-05-27 00:38:26 +0000 - publish - WARNING - This action is running on a pull_request event for a fork repository. It cannot do anything useful like creating check runs or pull request comments.
```
cc @jpkrohling

go.mod
```diff
@@ -70,6 +68,7 @@ require (
 	google.golang.org/grpc v1.29.1
 	gopkg.in/ini.v1 v1.52.0 // indirect
 	gopkg.in/yaml.v2 v2.4.0
+	gotest.tools/gotestsum v1.6.4 // indirect
```
yurishkuro, line 71: In order for this dep to remain here, you need to import it in `tools.go`

Makefile
```diff
@@ -35,6 +35,7 @@ GOOS ?= $(shell go env GOOS)
 GOARCH ?= $(shell go env GOARCH)
 GOBUILD=CGO_ENABLED=0 installsuffix=cgo go build -trimpath
 GOTEST=go test -v $(RACE)
+GOTESTREPORT=gotestsum --junitfile test-report.xml -- -v $(RACE)
```
yurishkuro, line 38: perhaps instead of adding new var we can simply modify GOTEST. One nice thing about gotestsum is that it colors the output, which we had to do manually
```
GOTESTREPORT=gotestsum --format testname --junitfile test-report.xml -- -v $(RACE)
```

ddmendes: I tried it first but it seems several targets uses GOTEST without running install-ci so almost all the PR checks failed on that attempt.

ddmendes:
> Doesn't seem like it can publish the test results on PRs:
> 
> ```
> 2021-05-27 00:38:26 +0000 - publish - WARNING - This action is running on a pull_request event for a fork repository. It cannot do anything useful like creating check runs or pull request comments.
> ```
> 
> cc @jpkrohling

Ow, I missed this part of the documentation https://github.com/EnricoMi/publish-unit-test-result-action#support-fork-repositories-and-dependabot-branches

Will work on it as well.

Makefile
yurishkuro: those targets should be easy to fix though. Or we could make GOTEST point to a shell script that will ensure the module is installed before executing

jpkrohling:
> Will work on it as well.

Did I read right that the solution is to not run this target on pull requests from forks?

ddmendes:
> > Will work on it as well.
> 
> Did I read right that the solution is to not run this target on pull requests from forks? 

Not out of the box. They say it should work if the test-report.xml is uploaded to main repo using actions/upload_artifact and on a workflow for unit-test completed we may download it and publish. Will try it once I have a free time again.

jpkrohling:
@ddmendes, do you have any updates about this one? Not sure there's an easy solution to this :-/

ddmendes:
Hi, @jpkrohling. I didn't try it yet but I would like to. Sorry by the long time. Need to organize myself to fit this.

PR #3043, opened 2021-05-30
Run the cassandra schema job as a non-root user by default

chgl:

By default, the cassandra schema job runs as a root user. This doesn't work on clusters with restrictive security policies. Although I did create https://github.com/jaegertracing/jaeger-operator/pull/1167 in the past, I think it's more convenient to have this change directly within the schema job and have a more secure default.

Adds a new user `nonroot` to the schema image and sets this as the default user.


pavolloffay (approved):

plugin/storage/cassandra/Dockerfile
```diff
@@ -3,4 +3,9 @@ FROM cassandra:3.11
 COPY schema/* /cassandra-schema/
 
 ENV CQLSH_HOST=cassandra
+
+RUN groupadd -g 65532 nonroot && \
```
pavolloffay, line 7: instead of using a number I would prefer using a name and then reuse that name in the `USER` statement

chgl: Thanks for the feedback! This is mostly due to laziness: when using a numeric value in the user directive, kubernetes/pod security policies are able to determine that the pod/container is running as a non-root user. When using a name, kubernetes can't automatically determine that the user id is != 0, so one would have to explicitly set the securityContext's runAsUser to 65532.

Here, the group name and username are set as 'nonroot' and both group and user IDs are set to an arbitrary (inspired by the distroless projects 'nonroot' user) large number.

It's basically the same as I did in https://github.com/jaegertracing/jaeger/pull/2466 , where a similar remark was made in the comments.

pavolloffay: thanks for the explanation this is grey area for me :). Can I ask why `65532`? Does the number have any exact meaning (I found that other projects are using it as well)?

chgl: Sure! The number `65532` is just muscle-memory from using Google's distroless container images regularly, which defines a user called `nonroot` with UID `65532`, see https://github.com/GoogleContainerTools/distroless/issues/443. It seems appropriate since it's greater than `10000`, so unlikely to collide with any existing UID on the host-system (see also http://www.linfo.org/uid.html via https://github.com/zegl/kube-score/issues/204).

If Kubernetes gets support for user-namespaces eventually, this will probably all become irrelevant, but at this time almost all Kubernetes manifest linting tools will complain about running containers as root.

yurishkuro: how is this different from what we do in ./cmd/agent/Dockerfile ?
```
ARG USER_UID=10001
USER ${USER_UID}
```

chgl: @yurishkuro not too different, I noticed when I just set the `USER <some-id>`, cqlsh would warn about trying to write some log file to /.cassandra. When creating a dedicated user with a home directory, it would instead write to `/home/nonroot/.cassandra`. I know that these logs are effectively useless, but it seemed like a cleaner solution.

In case of the agent, since it doesn't ever try to access the file system, setting just some random UID without an actual user behind it works just as well. For consistency I could use the same UID (10001) here as well and/or allow specifying it as a build-time argument (although I doubt it will ever need to be changed). Thoughts?

pavolloffay (approved):

plugin/storage/cassandra/Dockerfile
jpkrohling: I think the current code is fine.

jpkrohling (approved):

Issue #169, opened 2017-05-18
[Experimental] Provide GraphQL query service


saminzadeh:
Instead of an arbitrary "RESTful" HTTP API service, I think a better solution would be to implement the query service over [GraphQL](http://graphql.org/).

Here is my sample implementation of what the service would like:

https://launchpad.graphql.com/wxn5zk8mz
HTTP Endpoint: https://wxn5zk8mz.lp.gql.zone/graphql

**Some of the benefits:**
- Clients can request what they want, resulting in smaller payloads and less unused data. Really beneficial in the list view of traces, we don't need to guess what a client needs as designs/needs may change.
- Documentation for free (via introspection) and strongly typed API backed by a spec: https://facebook.github.io/graphql/
- A descriptive data model to be more expressive with complex queries. (more robust the URL encoded params)
- GraphQL queries can be issues via HTTP, TChannel, or any communication protocol.
- Simple computed fields and object type relationships

This will help solve the following issues as well:
https://github.com/uber/jaeger/issues/158
https://github.com/uber/jaeger/issues/123

yurishkuro:
@saminzadeh any thoughts about https://rejoiner.io/ ?

saminzadeh:
> @saminzadeh any thoughts about https://rejoiner.io/ ?

@yurishkuro I'm aware but have had not time to research it yet. Looks very promising and worth investigating

rubenvp8510:
@yurishkuro  Wondering if is still plans to do this, I would like to contribute and work on this one.

yurishkuro:
We don't have active plans to move in this direction. I think we need better justification of the costs/benefits before we can undertake this work. Also, we now support gRPC API for querying, so I would like to deprecate the old REST API, but we have not decided what should replace it.

cc @tiffon @everett980 - do you have any thoughts on this?

rubenvp8510:
Well, I was thinking this could be used by the UI, not sure what is the status of the gRPC  in the browsers.

yurishkuro:
@rubenvp8510 sure, but we need to get some very useful new capabilities that we cannot get otherwise, if we decide to undertake this significant renovation.

jpkrohling:
> but we need to get some very useful new capabilities that we cannot get otherwise

Being able to retrieve only a limited set of fields (only operation names, for instance) is already a new useful capability, IMO.

pavolloffay:
We (Red Hat) would like to move this forward and introduce GraphQL API. 

[...]

PR #3051, opened 2021-06-03
POC: Add GraphQL API for query service

pavolloffay:
Updates https://github.com/jaegertracing/jaeger/issues/169

Please do not nit-pick this is a POC with rather ugly code. I just want to demonstrate how GraphQL API might work with the query service.

#### Goals
* Documented and stable query API
* Return OTLP JSON proto model - https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/collector/trace/v1/trace_service.proto#L41 
  * Use "standard"/well defined model
  * The returned data can be directly reingested 
 
##### https://github.com/99designs/gqlgen/

* GraphQL SDL first approach 
* Tool `gqlgen` generates handlers (and optionally model) from schema file

Issues:
* In order to be compatible with OTLP trace JSON `JSONPb` marshaller has to be used. GraphQL framework allows to use  custom marshallers, however this works only well for scalar types. If the marshaller is used for objects then selecting properties/fields in the query does not work (e.g. `"query": "query {trace(traceId: \"todo\") { resourceSpans{instrumentationLibrarySpans{spans{name}}} }}"}` to select the span name). There is a manual and a quite ugly workaround (I will add a pointer to the code), also note that GraphQL query has to specify at least one object propery. 
* The GraphQL response wraps the returned JSON into the following object. Hence the response cannot be directly processed and ingested as OTPL JSON traces.

```json
{
  "data": {
    "<queryName>/getTrace": {
        "resourceSpans": {}
    }
}
```

##### Other notes

Example curl request
```
curl -ivX POST -H 'Content-Type: application/json' -d '{"query": "query {trace(traceId: \"<traceId>\") { resourceSpans{instrumentationLibrarySpans{spans{name}}} }}"}' http://localhost:8080/query
```

Example GraphQL queries in the GraphQL editor `http://localhost:8080/`
```graphql
query getTrace{ trace(traceId: "foo") { resourceSpans{instrumentationLibrarySpans{ instrumentationLibrary{name} spans{traceId}}}  } }

query findTraces {
  traces(service: "payments", startMicros: 10, endMicros:150) {
    name
    traceId
  }
}
```


cmd/query/app/graphql/graph/schema.resolvers.go
```diff
@@ -0,0 +1,91 @@
...
+		TraceId: []byte{0, 1, 2, 3, 4},
+		Name:    "hello",
+	}
+
+	// Remove fields that hasn't been requested
+	// this is a workaround because:
+	// * a custom marshaller is used for OTLP Span
+	//   to ensure compatibility with OTLP JSON (e.g. standard JSON marshalling cannot be used for protos)
+	// * a custom marshaller works well only with scalar types
+	if !names["resourceSpans.instrumentationLibrarySpans.spans.traceId"] {
```
pavolloffay, line 51: Here is the workaround to make selecting object properties/fields work. This is only needed when custom object marshaller is used. The custom marshaller `JSONPb` for `ResourceSpans` has to be used to be compatible with OTLP trace JSON.

yurishkuro:
btw, do we know how https://github.com/dgraph-io/dgraph is going graphql with Go & gRPC?

pavolloffay:
by looking at dependencies it might be using https://github.com/graph-gophers/graphql-go

pavolloffay:
Closing in favour of https://github.com/jaegertracing/jaeger/pull/3086

PR #3055, opened 2021-06-04
Add TLS support for Prometheus Reader

albertteoh:

- Relates to #2954.

- Adds TLS encryption to Prometheus server.

## Testing
- As [documented](https://prometheus.io/docs/guides/tls-encryption/), Prometheus does not natively support TLS encryption for connections to its HTTP API, however, it does suggest an alternative means of enforcing TLS encryption at the proxy layer with something like nginx.
- This PR was tested using nginx as a reverse proxy to Prometheus, as suggested above.

```
# Queries succeed: Directly connect to Prometheus, bypassing nginx
$ METRICS_STORAGE_TYPE=prometheus ./cmd/query/query-darwin-amd64 \
  --prometheus.server-url http://localhost:9090

# Queries succeed: Enable TLS and provide public key
$ METRICS_STORAGE_TYPE=prometheus ./cmd/query/query-darwin-amd64 \
  --prometheus.server-url https://example.com/prometheus \
  --prometheus.tls.ca=example.com.crt \
  --prometheus.tls.enabled

# Queries fail: TLS not enabled
$ METRICS_STORAGE_TYPE=prometheus ./cmd/query/query-darwin-amd64 \
  --prometheus.server-url https://example.com/prometheus  \
  --prometheus.tls.ca=example.com.crt 
...failed executing metrics query: Post "https://example.com/prometheus/api/v1/query_range": x509: certificate signed by unknown authority...

# Queries fail: Missing public key
$ METRICS_STORAGE_TYPE=prometheus ./cmd/query/query-darwin-amd64 \
  --prometheus.server-url https://example.com/prometheus  \
  --prometheus.tls.enabled
...failed executing metrics query: Post "https://example.com/prometheus/api/v1/query_range": x509: certificate signed by unknown authority...
```

### DIFF
--- a/pkg/prometheus/config/config.go
+++ b/pkg/prometheus/config/config.go
@@ -22,7 +22,7 @@ import (
 
 // Configuration describes the options to customize the storage behavior.
 type Configuration struct {
-	ServerURL      string         `validate:"nonzero" mapstructure:"server"`
-	ConnectTimeout time.Duration  `validate:"nonzero" mapstructure:"timeout"`
-	TLS            tlscfg.Options `mapstructure:"tls"`
+	ServerURL      string
+	ConnectTimeout time.Duration
+	TLS            tlscfg.Options
 }
--- a/pkg/prometheus/config/config.go
+++ b/pkg/prometheus/config/config.go
@@ -14,10 +14,15 @@
 
 package config
 
-import "time"
+import (
+	"time"
+
+	"github.com/jaegertracing/jaeger/pkg/config/tlscfg"
+)
 
 // Configuration describes the options to customize the storage behavior.
 type Configuration struct {
-	HostPort       string        `validate:"nonzero" mapstructure:"server"`
-	ConnectTimeout time.Duration `validate:"nonzero" mapstructure:"timeout"`
+	ServerURL      string         `validate:"nonzero" mapstructure:"server"`
+	ConnectTimeout time.Duration  `validate:"nonzero" mapstructure:"timeout"`
+	TLS            tlscfg.Options `mapstructure:"tls"`
 }
--- a/plugin/metrics/prometheus/metricsstore/reader.go
+++ b/plugin/metrics/prometheus/metricsstore/reader.go
@@ -16,6 +16,7 @@ package metricsstore
 
 import (
 	"context"
+	"crypto/tls"
 	"fmt"
 	"net"
 	"net/http"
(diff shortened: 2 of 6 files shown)


plugin/metrics/prometheus/options.go
```diff
@@ -21,14 +21,15 @@ import (
 
 	"github.com/spf13/viper"
 
+	"github.com/jaegertracing/jaeger/pkg/config/tlscfg"
 	"github.com/jaegertracing/jaeger/pkg/prometheus/config"
 )
 
 const (
-	suffixHostPort       = ".host-port"
+	suffixServerURL      = ".server-url"
```
albertteoh, line 29: Motivation for renaming:

- To use TLS with Prometheus requires the protocol scheme `https`, as with how ES's server URLs are defined. Hence, I felt it made more sense that users are required to provide the appropriate protocol scheme of either `http` or `https`.
- Consistency with ES options `server-urls`, and made more sense than `host-port` if requiring the protocol scheme.

pkg/prometheus/config/config.go
```diff
@@ -14,10 +14,15 @@
...
+	"time"
+
+	"github.com/jaegertracing/jaeger/pkg/config/tlscfg"
+)
 
 // Configuration describes the options to customize the storage behavior.
 type Configuration struct {
-	HostPort       string        `validate:"nonzero" mapstructure:"server"`
-	ConnectTimeout time.Duration `validate:"nonzero" mapstructure:"timeout"`
+	ServerURL      string         `validate:"nonzero" mapstructure:"server"`
```
yurishkuro, line 25: > `validate:"nonzero" mapstructure:"server"`

Where are these being used?  I thought they were introduced when @pavolloffay tried to use OTEL's config mechanism to populate Jaeger configs, but we stopped doing that.

yurishkuro: If not used, I would remove the annotations

plugin/metrics/prometheus/metricsstore/reader.go
```diff
@@ -247,3 +241,25 @@ func logErrorToSpan(span opentracing.Span, err error) {
 	ottag.Error.Set(span, true)
 	span.LogFields(otlog.Error(err))
 }
+
+func getHTTPRoundTripper(c *config.Configuration, logger *zap.Logger) (http.RoundTripper, error) {
+	if !c.TLS.Enabled {
+		return nil, nil
```
yurishkuro, line 247: that means you never want to support non-TLS connections?

albertteoh: It means to use the `DefaultRoundTripper` supplied by Prometheus client (the client will do this on our behalf), but you make a good point that we lose our `ConnectTimeout` configuration with this approach. I've written a test to fail this code and fixed the implementation.

pkg/prometheus/config/config.go
albertteoh: The annotations are based on existing example config structs, and I didn't know what they were used for as well tbh; removed.

yurishkuro (approved):

albertteoh:
Thanks for the review @yurishkuro!

PR #3056, opened 2021-06-04
Upgrade gRPC to 1.38.x

### LINKED ISSUE #2172: grpc: Server.Serve failed to create ServerTransport

### LINKED ISSUE #2771: Upgrade google.golang.org/grpc to 1.35.0

pavolloffay:
Builds on top of #2857 

Resolves #2771
Resolves #2172


proto-gen/api_v2/codec.go
```diff
@@ -0,0 +1,45 @@
...
+
+// Name implements encoding.Codec
+func (c *gogoCodec) Name() string {
+	return "proto"
+}
+
+// Marshal implements encoding.Codec
+func (c *gogoCodec) Marshal(v interface{}) ([]byte, error) {
+	t := reflect.TypeOf(v)
+	if strings.Contains(t.String(), "Empty") {
```
pavolloffay, line 32: This does the trick to pass the grpc storage test. This is obviously not good. if we go forward we should whitelist jaeger types to use gogo and use standard impl by default. 

cc) @joe-elliott

proto-gen/api_v2/codec.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"github.com/gogo/protobuf/proto"
+	"google.golang.org/grpc/encoding"
+	gproto "google.golang.org/protobuf/proto"
+	"reflect"
+	"strings"
+)
+
+func init() {
+	encoding.RegisterCodec(newCodec())
```
pavolloffay, line 12: @bogdandrutu or @tigrannajaryan could you please loop in here? This PR updates to the latest gRPC and protobuf. 

Jaeger uses gogo proto as OTELcol and it uses the same features - custom type for IDs and nonullable fields. To make this with the latest dependencies this codes registration is needed.  I could not find the custom codec registration in OTELcol. Do you know why it is not needed?

proto-gen/api_v2/codec.go
joe-elliott: Yeah, that's pretty rough. I understand why that works, but we're really holding things together with shoestrings and duct tape here. At the least it needs a comment to explain why it exists.

The good news is that Jaeger tests are robust enough we can feel confident in these changes, but I think someone is going to have to take on a major project to get us off of gogoproto for Jaeger's long term health.

pavolloffay: The idea behind this is to use gogo marshaller only for Jaeger types.

pavolloffay: Added comment and condition to use gogo only for jaeger types - based on the import path.

pavolloffay:
This is ready for review/merge

proto-gen/codec.go
```diff
@@ -0,0 +1,54 @@
...
+
+	gogoproto "github.com/gogo/protobuf/proto"
+	"google.golang.org/grpc/encoding"
+	"google.golang.org/protobuf/proto"
+)
+
+const jaegerProtoGenPkgPath = "github.com/jaegertracing/jaeger/proto-gen"
+
+func init() {
+	encoding.RegisterCodec(newCodec())
```
jpkrohling, line 15: Is this an idempotent call?

jpkrohling (commented):
It would be good to have a review from someone with more experience with gRPC/Gogo, as I don't really understand how `proto-gen/codec.go` fixes our problems.

In any case, is this something we'd want for the long-term, or do you think we still need a better solution in the future?

proto-gen/codec.go
pavolloffay: yes

cmd/agent/app/reporter/grpc/builder.go
```diff
@@ -123,3 +125,12 @@ func (b *ConnBuilder) CreateConnection(logger *zap.Logger, mFactory metrics.Fact
 
 	return conn, nil
 }
+
+// generateAndRegisterManualResolver was removed from grpc.
```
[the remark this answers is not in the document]
pavolloffay: It seems they leave the work for the caller.

pavolloffay:
> In any case, is this something we'd want for the long-term, or do you think we still need a better solution in the future?

correct, this is rather a short-term solution. Since the gogo is compatible only with go proto V1 and it will not be made compatible with v2 we will have to migrate. It seems that solutions like https://github.com/planetscale/vtprotobuf (based on https://github.com/gogo/protobuf/issues/691#issuecomment-854136033) will be way forward. It is a plugin for proto compiler that generates serialization methods, it does not modify generated model classes.
 This has a couple of implications for us:
* no support for nonnullable fields
* no support for custom types e.g. (TraceID)

pavolloffay:
@joe-elliott / @yurishkuro could you please review?

go.mod
```diff
@@ -26,10 +26,10 @@ require (
 	github.com/gogo/googleapis v1.4.1
 	github.com/gogo/protobuf v1.3.2
 	github.com/golang/mock v1.4.3 // indirect
-	github.com/golang/protobuf v1.3.5
+	github.com/golang/protobuf v1.5.2
 	github.com/gorilla/handlers v1.5.1
-	github.com/gorilla/mux v1.8.0
-	github.com/grpc-ecosystem/go-grpc-middleware v1.3.0
+	github.com/gorilla/mux v1.7.4
```
yurishkuro, line 31: is this supposed to go backwards?

proto-gen/codec_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+	"github.com/golang/protobuf/proto"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/protobuf/types/known/emptypb"
+
+	"github.com/jaegertracing/jaeger/model"
+)
+
+func TestCodecMarshallAndUnmarshall_jaeger_type(t *testing.T) {
```
yurishkuro, line 14: are you sure tests are being executed in this proto-gen package? I thought we were excluding them (you have a fmt issue above so pretty sure linter is not running here). I think it's better not to mix generated code with manual code.

yurishkuro (commented):
Can you try running a ser/deser benchmark before this upgrade and after?

proto-gen/codec_test.go
pavolloffay: To what package should we move this then? I can think of `model` only.  

> I think it's better not to mix generated code with manual code.

We already do mix packages e.g. `model` package.

yurishkuro: in the model we do it out of necessity, since gen-ed code doubles as domain model and the custom types ser/deset methods need to be there. But we usually don't write manual code in ***-gen packages.

You can create a new package like pkg/gogocodec

pavolloffay: The best would be get rid of this, I cannot find custom codec in OTELcol, so perhaps it's doable to do it without it.

pavolloffay: The custom codec is needed only if a custom type (e.g. `TaceID) is used directly in the request (probably also response) object - https://github.com/jaegertracing/jaeger-idl/blob/master/proto/api_v2/query.proto#L38 via https://github.com/jaegertracing/jaeger-idl/blob/master/proto/api_v2/query.proto#L128. If the `TraceID` was wrapped into another message the serialization would work without the custom codec.

proto-gen/api_v2/codec.go
tigrannajaryan: >Do you know why it is not needed?

I am not very familiar with this specific functionality of gogo, I don't know the answer to this.

go.mod
pavolloffay: not sure why it went backwards, maybe a mistake during rebase.

pavolloffay:
Benchmark results look good to me. See the results below. The benchmark code is in this PR in the `model` package.

This PR:
```
GOMAXPROCS=1 go test -bench=. -test.benchtime=10s -benchmem -cpuprofile profile_cpu.out ./model                                                                                             
goos: linux
goarch: amd64
pkg: github.com/jaegertracing/jaeger/model
cpu: Intel(R) Core(TM) i7-9850H CPU @ 2.60GHz
BenchmarkKeyValueIsLessAndEquals 	313754542	        38.84 ns/op	       0 B/op	       0 allocs/op
BenchmarkKeyValuesSort           	83939259	       154.1 ns/op	      24 B/op	       1 allocs/op
BenchmarkKeyValuesWrapping       	1000000000	         0.2424 ns/op	       0 B/op	       0 allocs/op
BenchmarkSpanHash                	  552618	     21807 ns/op	    2695 B/op	     112 allocs/op
BenchmarkBatchSerialization/marshal         	15824212	       749.9 ns/op	     480 B/op	       7 allocs/op
BenchmarkBatchSerialization/unmarshal       	 7336722	      1737 ns/op	    1592 B/op	      25 allocs/op
PASS
ok  	github.com/jaegertracing/jaeger/model	68.924s
```

Master with `github.com/gogo/protobuf/proto`
```
GOMAXPROCS=1 go test -bench=. -test.benchtime=10s -benchmem -cpuprofile profile_cpu.out ./model                                                                                               
go: downloading github.com/golang/protobuf v1.3.5
goos: linux
goarch: amd64
pkg: github.com/jaegertracing/jaeger/model
cpu: Intel(R) Core(TM) i7-9850H CPU @ 2.60GHz
BenchmarkKeyValueIsLessAndEquals 	306750854	        39.29 ns/op	       0 B/op	       0 allocs/op
BenchmarkKeyValuesSort           	84167293	       151.4 ns/op	      24 B/op	       1 allocs/op
BenchmarkKeyValuesWrapping       	1000000000	         0.2390 ns/op	       0 B/op	       0 allocs/op
BenchmarkSpanHash                	  544084	     21888 ns/op	    2695 B/op	     112 allocs/op
BenchmarkBatchSerialization/marshal         	11489486	      1047 ns/op	     624 B/op	      10 allocs/op
BenchmarkBatchSerialization/unmarshal       	 7310154	      1633 ns/op	    1592 B/op	      25 allocs/op
PASS
ok  	github.com/jaegertracing/jaeger/model	68.099s
```

Master with `ithub.com/golang/protobuf/proto`
```
GOMAXPROCS=1 go test -bench=. -test.benchtime=10s -benchmem -cpuprofile profile_cpu.out ./model                                                                                                 
goos: linux
goarch: amd64
pkg: github.com/jaegertracing/jaeger/model
cpu: Intel(R) Core(TM) i7-9850H CPU @ 2.60GHz
BenchmarkKeyValueIsLessAndEquals 	306549897	        42.74 ns/op	       0 B/op	       0 allocs/op
BenchmarkKeyValuesSort           	77749569	       158.2 ns/op	      24 B/op	       1 allocs/op
BenchmarkKeyValuesWrapping       	1000000000	         0.2391 ns/op	       0 B/op	       0 allocs/op
BenchmarkSpanHash                	  553023	     21951 ns/op	    2695 B/op	     112 allocs/op
BenchmarkBatchSerialization/marshal         	11297779	      1064 ns/op	     624 B/op	      10 allocs/op
BenchmarkBatchSerialization/unmarshal       	 7264723	      1635 ns/op	    1592 B/op	      25 allocs/op
PASS
ok  	github.com/jaegertracing/jaeger/model	68.930s
```

pavolloffay:
> Do I read it right that golang/protobuf and gogo/protobuf are on the same ballpark? If this benchmark is correct, is it even worth using gogo at all in our case?

We do not use gogo only for perf but also for using custom types (e.g. `TraceID`) and minimize allocations by not using nullable fields - this is not reflected in serialization benchmark. Also even when using golang/proto serialization the gogo generated code is most likely used. To properly compare gogo and stdlib the benchmark would have to generate model twice. For purpose of this PR ignore the golang/protobuf result.

yurishkuro (approved):

plugin/storage/integration/grpc_test.go
```diff
@@ -44,6 +44,8 @@ func (s *GRPCStorageIntegrationTestSuite) initialize() error {
 	err := command.ParseFlags([]string{
 		"--grpc-storage-plugin.binary",
 		s.pluginBinaryPath,
+		"--grpc-storage-plugin.log-level",
```
joe-elliott, line 47: I think I added this when I was trying to determine why the plugin wasn't working. It's not necessary but I don't think it hurts anything either.

cmd/agent/app/reporter/grpc/builder.go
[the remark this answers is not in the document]
joe-elliott: Perhaps we should move to `WithResolver`?

https://github.com/grpc/grpc-go/pull/3960

Not much info about why it was removed here.

PR #3060, opened 2021-06-06
Refactor QueryService tests

albertteoh:

- Relates to #2954.

- Refactor QueryService tests for easier maintenance and reduce required changes when introducing metrics reader to query service:
  - Return a `testMocks` struct rather than return multiple mock objects.
  - Only use what's required, instead of multiple blank identifiers for unwanted mocks.
  - Introduce functional options on a single test bootstrapping function instead of mostly duplicated functions for specific combinations of parameters.

### DIFF
--- a/cmd/query/app/querysvc/query_service_test.go
+++ b/cmd/query/app/querysvc/query_service_test.go
@@ -59,118 +59,119 @@ var (
 	}
 )
 
-type testMocks struct {
-	spanReader *spanstoremocks.Reader
-	depsReader *depsmocks.Reader
+type testQueryService struct {
+	queryService *QueryService
+	spanReader   *spanstoremocks.Reader
+	depsReader   *depsmocks.Reader
 
 	archiveSpanReader *spanstoremocks.Reader
 	archiveSpanWriter *spanstoremocks.Writer
 }
 
-type QueryServiceOptionApplier func(*testMocks, *QueryServiceOptions)
+type testOption func(*testQueryService, *QueryServiceOptions)
 
-func withArchiveSpanReader() QueryServiceOptionApplier {
-	return func(mocks *testMocks, options *QueryServiceOptions) {
+func withArchiveSpanReader() testOption {
+	return func(mocks *testQueryService, options *QueryServiceOptions) {
 		r := &spanstoremocks.Reader{}
-		options.ArchiveSpanReader = r
 		mocks.archiveSpanReader = r
+		options.ArchiveSpanReader = r
 	}
 }
 
-func withArchiveSpanWriter() QueryServiceOptionApplier {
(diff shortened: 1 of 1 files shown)


cmd/query/app/querysvc/query_service_test.go
```diff
@@ -59,42 +59,63 @@ var (
...
-		ArchiveSpanWriter: archiveWriteStorage,
+type testMocks struct {
+	spanReader *spanstoremocks.Reader
+	depsReader *depsmocks.Reader
+
+	archiveSpanReader *spanstoremocks.Reader
+	archiveSpanWriter *spanstoremocks.Writer
+}
+
+type QueryServiceOptionApplier func(*testMocks, *QueryServiceOptions)
```
yurishkuro, line 70: usually these types are named simply XyzOption

cmd/query/app/querysvc/query_service_test.go
```diff
@@ -59,42 +59,63 @@ var (
...
 			return trace, errAdjustment
-		}),
+		})
 	}
-	qs := NewQueryService(readStorage, dependencyStorage, options)
-	return qs
 }
 
-func initializeTestService() (*QueryService, *spanstoremocks.Reader, *depsmocks.Reader) {
+func initializeTestService(optionAppliers ...QueryServiceOptionApplier) (*QueryService, *testMocks) {
```
yurishkuro, line 96: since you're already introducing a new struct testMocks, you might instead return a single `testQueryService` struct which contains both the `*QS` and the mocks at top level

yurishkuro (approved):

cmd/query/app/querysvc/query_service_test.go
albertteoh: glad you mentioned this as this was also what I wanted to do initially, just thought there may have been a preference to separate the "thing under test" from the mocks.

albertteoh:
Thanks Yuri!

cmd/query/app/querysvc/query_service_test.go
albertteoh: btw, I'm planning to do the same thing for [`QueryServiceOptions`](https://github.com/jaegertracing/jaeger/blob/c5642b708be9e2577cc1c494889ae97946ccc78a/cmd/query/app/querysvc/query_service.go#L41) in the next PR, as I think MetricsQuery reader belongs in this struct. 

Can you think of a good name for the `Option`? `QueryServiceOptionsOption` seems a little too stutter-y and I'm a little short on other ideas on names. :P

cmd/query/app/querysvc/query_service_test.go
yurishkuro: tqsOption - these are private types, no need to be especially creative

PR #3061, opened 2021-06-06
Add metrics query capability to query service

albertteoh:

- Relates to #2954.

- Add metrics query capabilities to QueryService.
- As it's an opt-in reader, it has been added to `QueryServiceOptions` instead of being a mandatory reader like the span or dependencies reader.


cmd/query/app/querysvc/query_service.go
```diff
@@ -51,6 +54,9 @@ type QueryService struct {
 	options          QueryServiceOptions
 }
 
+// QSOption is the functional option for configuring QueryServiceOptions.
+type QSOption func(qOpts *QueryServiceOptions)
```
yurishkuro, line 58: Why do we need this as public?

cmd/query/app/querysvc/query_service.go
```diff
@@ -125,6 +131,32 @@ func (qs QueryService) GetDependencies(ctx context.Context, endTs time.Time, loo
 	return qs.dependencyReader.GetDependencies(ctx, endTs, lookback)
 }
 
+// MetricsQueryEnabled returns whether if metric query capabilities are enabled, opted-in by setting the
+// METRICS_STORAGE_TYPE environment variable.
+func (qs QueryService) MetricsQueryEnabled() bool {
```
yurishkuro, line 136: Why is this needed as public? For tests you can check qs.options directly

cmd/query/app/querysvc/query_service.go
```diff
@@ -125,6 +131,32 @@ func (qs QueryService) GetDependencies(ctx context.Context, endTs time.Time, loo
...
+	return qs.options.MetricsReader != nil
+}
+
+// GetLatencies is the queryService implementation of metricsstore.Reader.
+func (qs QueryService) GetLatencies(ctx context.Context, params *metricsstore.LatenciesQueryParameters) (*metrics.MetricFamily, error) {
+	return qs.options.MetricsReader.GetLatencies(ctx, params)
```
yurishkuro, line 142: I think you will want to test for nil

cmd/query/app/querysvc/query_service.go
albertteoh: The original motivation was to use it here: https://github.com/jaegertracing/jaeger/blob/c5642b708be9e2577cc1c494889ae97946ccc78a/cmd/query/app/flags.go#L125

to avoid modifying the caller's signature if metrics querying is opted-out, but I think it's not a particularly large impact to simply add it as a parameter and pass in `nil` if opted-out, so I've removed this option for now.

cmd/query/app/querysvc/query_service.go
albertteoh: To allow the HTTP and gRPC handlers to return early if the metrics query endpoints are called but not enabled like so:

HTTP Handler:
```
if !aH.queryService.MetricsQueryEnabled() {
	aH.handleError(w, errMetricsQueryDisabled, http.StatusNotImplemented)
	return
}
```

GRPC Handler:
```
if !g.queryService.MetricsQueryEnabled() {
	return nil, status.Error(codes.Unimplemented, "metrics querying is currently disabled")
}
```

An alternative could be to remove this endpoint and have the `QueryService.GetLatencies`, etc. functions check for `nil` `MetricsReader` then return a `MetricsReaderDisabled` error type and use [`errors.Is`](https://golang.org/pkg/errors/#Is) to handle the "not implemented" case. 

What do you think?

cmd/query/app/querysvc/query_service.go
albertteoh: As above, these nil checks are performed by the handlers so at this point, `MetricsReader` should never be nil. However, if we take the proposed alternative of introducing a `MetricsQueryDisabled` error, then we could perform the `nil` check here.

albertteoh:
@yurishkuro, are there any further questions/concerns that need addressing in this PR?

cmd/query/app/querysvc/query_service.go
albertteoh: @yurishkuro  I've moved metrics-related endpoints into a new `MetricsQueryService` and also added nil checks from each endpoint.

yurishkuro (approved):

albertteoh:
Thanks Yuri! I'll look into your suggestion for the next PR.

PR #3078, opened 2021-06-09
Use withResolver API in gRPC reporter

pavolloffay:
Related to https://github.com/jaegertracing/jaeger/pull/3056#discussion_r647636261

See https://github.com/grpc/grpc-go/pull/3960
See https://github.com/grpc/grpc-go/pull/3700


cmd/agent/app/reporter/grpc/builder.go
```diff
@@ -82,7 +82,8 @@ func (b *ConnBuilder) CreateConnection(logger *zap.Logger, mFactory metrics.Fact
 			return nil, errors.New("at least one collector hostPort address is required when resolver is not available")
 		}
 		if len(b.CollectorHostPorts) > 1 {
-			r, _ := generateAndRegisterManualResolver()
+			r := manual.NewBuilderWithScheme(strconv.FormatInt(time.Now().UnixNano(), 36))
```
pavolloffay, line 85: Can somebody explain why are we using a random scheme and what it does?

pavolloffay:
@joe-elliott could you please review?

cmd/agent/app/reporter/grpc/builder.go
joe-elliott: https://github.com/grpc/grpc-go/blob/master/dialoptions.go#L609

> will be matched against the current scheme...

Is the random scheme to prevent it from ever being matched? Can this code just be removed?

pavolloffay: the test suite failed when I didn't call `WithResolvers`.

joe-elliott: ok, not 100% sure but i've done some code diving:

https://github.com/grpc/grpc-go/blob/b1f7648a9fc72ce76cbcd42d8e2c60d9d9bed9fc/clientconn.go#L251

when the dialer wants to choose its resolver it looks for one matching the scheme of the target.  On this line we're just using the resolver scheme when creating the target:

```
dialTarget = r.Scheme() + ":///round_robin"
```

This basically guarantees that our target scheme lines up with the resolver scheme and the right resolver is chosen. I believe it's a random number so it never conflicts with a "real resolver":

https://github.com/grpc/grpc-go/blob/master/internal/resolver/passthrough/passthrough.go#L25
https://github.com/grpc/grpc-go/blob/master/internal/resolver/dns/dns_resolver.go#L158

We could leave it as is or we could change it to something like `jaeger_manual` if we're feeling fancy.

joe-elliott (approved):
lgtm

one comment on the scheme, but it can be merged as is.

cmd/agent/app/reporter/grpc/builder.go
pavolloffay: Let's feel fancy in this case,  I like the predictability :)

pavolloffay:
@joe-elliott could you please re-approve? I have changed one thing to make it fancier xD.

jpkrohling:
CI failed with:

> Running go fmt on ALL_SRC ...
> ./scripts/updateLicenses.sh >> .fmt.log
> ./scripts/import-order-cleanup.sh stdout > .import.log
> Go fmt, license check, or import ordering failures, run 'make fmt'

joe-elliott (approved):

yurishkuro (approved):

PR #3079, opened 2021-06-10
Hook up MetricsQueryService to main funcs

albertteoh:

- Relates to #2954.

- Instantiates the MetricsQueryService from both the `all-in-one` and `query` executables to be passed on to the HTTP and gRPC handlers. 
- `MetricsQueryService` exposes an `Enabled()` endpoint to determine if  the MetricsQuery feature is enabled (`METRICS_STORAGE_TYPE` env var is set), allowing HTTP and gRPC handlers to respond with a sensible error message/code.


cmd/query/app/server.go
```diff
@@ -111,11 +111,15 @@ func createGRPCServer(querySvc *querysvc.QueryService, options *QueryOptions, lo
 	server := grpc.NewServer(grpcOpts...)
 
 	handler := NewGRPCHandler(querySvc, logger, tracer)
+
+	// TODO: Register MetricsQueryService
```
albertteoh, line 115: Will be addressed in a follow-up PR implementing MetricsQuery support for the gRPC handler.

cmd/query/app/server.go
```diff
@@ -111,11 +111,15 @@ func createGRPCServer(querySvc *querysvc.QueryService, options *QueryOptions, lo
...
 	api_v2.RegisterQueryServiceServer(server, handler)
+
 	return server, nil
 }
 
-func createHTTPServer(querySvc *querysvc.QueryService, queryOpts *QueryOptions, tracer opentracing.Tracer, logger *zap.Logger) (*http.Server, error) {
+func createHTTPServer(querySvc *querysvc.QueryService, metricsQuerySvc *querysvc.MetricsQueryService, queryOpts *QueryOptions, tracer opentracing.Tracer, logger *zap.Logger) (*http.Server, error) {
+	// TODO: Add HandlerOptions.MetricsQueryService
```
albertteoh, line 122: Will be addressed in a follow-up PR implementing MetricsQuery support for the HTTP handler.

cmd/query/app/querysvc/metrics_query_service.go
```diff
@@ -28,44 +28,47 @@ type MetricsQueryService struct {
...
 
 // NewMetricsQueryService returns a new MetricsQueryService.
-// A nil reader will result in a nil MetricsQueryService being returned.
 func NewMetricsQueryService(reader metricsstore.Reader) *MetricsQueryService {
 	return &MetricsQueryService{
 		metricsReader: reader,
 	}
 }
 
+func (mqs MetricsQueryService) Enabled() bool {
```
albertteoh, line 40: Hopefully I understood your suggestion correctly, @yurishkuro.

yurishkuro: no, what I meant was adding this:
```
type nullMetricsQueryService struct {}
func (nullMetricsQueryService) GetLatencies(ctx context.Context, params *metricsstore.LatenciesQueryParameters) (*metrics.MetricFamily, error) { return nil, errNotEnabled }
...
```
and using it in RPC handlers. Basically, solve the problem with polymorphism, not with a bunch of nil checks all over the place.

albertteoh: that's pretty cool :) although, unless if I'm mistaken, I see two minor issues:
- The handlers will need to parse inputs, etc. then call `GetLatencies`, before it knows that Metrics Querying is disabled. I think it would be better to return early; but not sure it's possible with this approach.
- Once calling `GetLatencies`, the handler needs to differentiate the `errNotEnabled` from an "exceptional" error case so it can return an appropriate error code rather than the generic `StatusInternalServerError` code, which is important as the UI needs to handle both cases differently. 
  We could follow a pattern like [this](https://github.com/jaegertracing/jaeger/blob/master/cmd/query/app/http_handler.go#L415) but there's no guarantee the error could be wrapped up, so I imagine we'd need use something like `errors.Is`.

what do you think?

yurishkuro: and yes, you'd want to use `error.Is`, it's the new Go way, what we have with `if err == spanstore.ErrTraceNotFound {` predates that.

albertteoh: > what's the downside of parsing inputs?

not a large downside, and the emphasis wasn't so much on the input parsing but on when the handler knows if the feature is enabled or disabled. I feel the intent is clearer when reading the handler code instead of needing to follow it to the point where the error is handled:
```
metricsHandler() {
    if notEnabled {
       return notImplementedErrorCode
    }
    res, err := service.call()
    ...
}
```

In any case, it's a small matter I think, and happy to go ahead with your suggestion.

cmd/query/app/querysvc/metrics_query_service.go
```diff
@@ -16,56 +16,90 @@ package querysvc
...
 // GetMinStepDuration is the queryService implementation of metricsstore.Reader.
-func (mqs MetricsQueryService) GetMinStepDuration(ctx context.Context, params *metricsstore.MinStepDurationQueryParameters) (time.Duration, error) {
-	if mqs.metricsReader == nil {
-		return 0, errNilReader
-	}
+func (mqs enabledMetricsQueryService) GetMinStepDuration(ctx context.Context, params *metricsstore.MinStepDurationQueryParameters) (time.Duration, error) {
 	return mqs.metricsReader.GetMinStepDuration(ctx, params)
 }
+
+// GetLatencies is the queryService implementation of metricsstore.Reader.
```
yurishkuro, line 87: private types do not require comments

cmd/query/app/querysvc/metrics_query_service.go
```diff
@@ -16,56 +16,90 @@ package querysvc
...
+		GetCallRates(ctx context.Context, params *metricsstore.CallRateQueryParameters) (*metrics.MetricFamily, error)
+		GetErrorRates(ctx context.Context, params *metricsstore.ErrorRateQueryParameters) (*metrics.MetricFamily, error)
+		GetMinStepDuration(ctx context.Context, params *metricsstore.MinStepDurationQueryParameters) (time.Duration, error)
+	}
+
+	// errMetricsQueryDisabled is the error returned by disabledMetricsQueryService.
+	errMetricsQueryDisabled struct{}
+
+	// enabledMetricsQueryService represents an "enabled" MetricsQueryService implementation.
+	enabledMetricsQueryService struct {
```
yurishkuro, line 38: I think just `metricsQueryService` would be more natural name.

cmd/all-in-one/main.go
```diff
@@ -272,3 +296,31 @@ func initTracer(metricsFactory metrics.Factory, logger *zap.Logger) io.Closer {
...
+// metricsQueryEnabled returns whether if metrics querying capabilities are enabled.
+// To avoid introducing a breaking change, the Metrics Querying feature must be
+// explicitly opted-in by setting the METRICS_STORAGE_TYPE env var.
+// An unset METRICS_STORAGE_TYPE will disable this feature.
+func metricsQueryEnabled(fc metricsPlugin.FactoryConfig) bool {
+	return fc.MetricsStorageType != ""
+}
+
+func createMetricsReaderFactory(fc metricsPlugin.FactoryConfig) (*metricsPlugin.Factory, error) {
+	if !metricsQueryEnabled(fc) {
```
yurishkuro, line 309: are these checks still necessary? wouldn't you always get _something_ back (perhaps disabled service)?

albertteoh: Yeah, it's still necessary otherwise: `Cannot initialize metrics store factory: unknown metrics type "". Valid types are [prometheus]`

Do you think it's worth doing something similar with the factories? i.e. have `disabledMetricsReaderFactory` that simply returns a `nil` Reader? That way, these checks won't be necessary.

yurishkuro: I think you can extend the same approach - consider default metrics store type to be "none" and return a "disabled reader" in that case. Basically push this all the way down (you won't need "disabled service" in this case).

cmd/query/app/querysvc/metrics_query_service.go
```diff
@@ -16,22 +16,18 @@ package querysvc
...
-	"errors"
 	"time"
 
 	"github.com/jaegertracing/jaeger/proto-gen/api_v2/metrics"
 	"github.com/jaegertracing/jaeger/storage/metricsstore"
 )
 
-// MetricsQueryService contains the underlying reader required for querying the metrics store.
+// MetricsQueryService provides a means of querying R.E.D metrics from an underlying metrics store.
 type MetricsQueryService struct {
```
yurishkuro, line 26: What is `MetricsQueryService` supposed to add on top of `metricsstore.Reader`?

Note that we used to have HTTP handler use SpanReader directly, but then refactored it into QueryService because we needed to share it with GRPC Handler **and** the query service was an abstraction on top of two span readers: primary and archive. In your case there doesn't seem to be any additional abstraction needed, you just proxy all calls into the reader directly. Not a huge issue to leave as is, but just curious if there's any thinking that a layer of abstraction would be needed in the future.

plugin/metrics/factory.go
```diff
@@ -22,12 +22,16 @@ import (
...
 	"github.com/jaegertracing/jaeger/plugin"
+	"github.com/jaegertracing/jaeger/plugin/metrics/disabled"
 	"github.com/jaegertracing/jaeger/plugin/metrics/prometheus"
 	"github.com/jaegertracing/jaeger/storage"
 	"github.com/jaegertracing/jaeger/storage/metricsstore"
 )
 
 const (
+	// DisabledStorageType is the storage type used when METRICS_STORAGE_TYPE is unset.
+	DisabledStorageType = ""
```
yurishkuro, line 33: doesn't seem to be used outside of this pkg.

plugin/metrics/disabled/metricsstore/reader_test.go
```diff
@@ -0,0 +1,74 @@
+// Copyright (c) 2021 The Jaeger Authors.
```
yurishkuro, line 1: > plugin/metrics/disabled/metricsstore/

a bit of overkill with nesting of packages, I would've kept it next to Factory under metrics/disabled

yurishkuro (approved):
small nits, lgtm otherwise

cmd/query/app/querysvc/metrics_query_service.go
albertteoh: Indeed `MetricsQueryService` adds little on top of `metricsstore.Reader`. It has exactly the same function signature as the Reader and yes, it simply forwards the call. I also don't see a need for multiple metrics readers at runtime, for now at least.

It was added to follow the same design as span/dependency querying to avoid confusing/surprising those following the code and seeing a `QueryService` being used by handlers to query spans/dependencies, then a `Reader` used to query metrics; when they may expect the handlers to refer to "QueryService" abstractions for querying spans, dependencies **and** metrics.

One possibility is to avoid confusion, yet minimizing unnecessary code, is to make `MetricsQueryService` an interface, embedding `metricsstore.Reader`, meaning we can remove the `MetricsQueryService` struct and its proxy functions altogether and simply give the handlers a `metricsstore.Reader` instance. That way, handlers still refer to the "MetricsQueryService" alias for metrics, but instead using a Reader implementation. Something like:
```
type MetricsQueryService interface {
	metricsstore.Reader
}
```

yurishkuro (approved):

Issue #3062, opened 2021-06-06
Expose `jaeger_version` info metric with `pkg/version/build.go` as labels


mindw:
## Requirement - what kind of business use case are you trying to solve?
As a metrics consumer I would like to tell what version of jaeger components are deployed. There are very common use cases for this and it is both a common and encouraged practice.
For example, the golang version is exposed be default:
```
go_info{job="jaeger-agent", version="go1.16"}
```
## Problem - what in Jaeger blocks you from solving the requirement?
* only the jaeger-query version can be found via the "About Jaeger"  menu in the GUI.
* Impacts certain observability patterns https://www.robustperception.io/exposing-the-software-version-to-prometheus

## Proposal - what do you suggest to solve the problem or improve the existing situation?
Add jaeger_<component>_info metric for each component using the pre-existing content in `pkg/version/`.

PR #3087, opened 2021-06-10
Add build_info metrics for jaeger components

Ashmita152:

Resolves https://github.com/jaegertracing/jaeger/issues/3062

Following this [blog](https://www.robustperception.io/exposing-the-software-version-to-prometheus), exposing build_info metrics for every jaeger component.

Example for [Jaeger Agent metrics endpoint](http://localhost:14271/metrics):

Screenshot 2021-06-11 at 2 00 41 AM

( I have a local git tag as v1.32.0 hence version is showing that)


albertteoh (approved):

jpkrohling:
I like this idea, but we went in a different direction with #1333. The main reason is this, seen in a comment from @yurishkuro:

> Good idea. the only issue I foresee is that some metrics systems (in particular, Uber's M3) flush all metrics periodically, which results in the single data point in the time series db, thus if you're looking at the last hour of metrics and that point was earlier, you won't get any data.
>
> Perhaps instead of metrics (which are meant for time series, even if gauges), these values are better exposed via expvar.

Ashmita152:
Thank you @albertteoh @jpkrohling for the review. I will look at the PR associated with that issue and update it accordingly.

Ashmita152:
On reading more about expvar, I have a question about it:

It looks like expvar is golang related endpoint rather than prometheus related endpoint. That means if we expose it via expvar, it is exposed at /debug/vars which doesn't seems to be consumable by prometheus (since it's not exposing in prometheus format)
As per [the](https://www.robustperception.io/exposing-the-software-version-to-prometheus) blog linked to the issue, they are trying to add version label to other timeseries using some kind of join. This will not be possible if we expose it as expvar.

Looking at other tools like prometheus, alertmanager, victoriametrics, they are exposing this build information as metrics rather than expvar. 

I tried exposing the gauge as expvar which came something like this at http://localhost:14271/debug/vars 

expvar

I am not sure what will be the right way.

joe-elliott:
Responding to @jpkrohling/@yurishkuro 's comment, we use `_info` metrics regurlarly that are scraped by prometheus to build dashboards that include versioning information. I don't know the details of M3 but in Prometheus this series is always available to be joined against as long as the application is publishing it.

Ashmita152:
Hi @jpkrohling @yurishkuro 

May I ask for your review for this PR please.

Regards.

jpkrohling:
Should we convert the ones from #1333 as well, in name of consistency?

Ashmita152:
I don't have any strong opinion but I think command line flags should be kept as expvar unless they are changing dynamically during the runtime.

jpkrohling:
At least one value might be changing: the dyn-queue-size value.

PR #3090, opened 2021-06-13
Use object type instead of nested type for logs field in index mapping for jaeger-span

### LINKED ISSUE #3085: [Jaeger Collector - ES Backend ]Use object type instead of nested type for logs field in index mapping for jaeger-span

We use jaeger with Elasticsearch backend. We see that in jaeger-span index template - we use nested fields (Ref - https://www.elastic.co/guide/en/elasticsearch/reference/current/nested.html)  for logs (in index mapping - https://github.com/jaegertracing/jaeger/blob/7dedc4637eeb80e0605a82720e0c23feabe623b0/plugin/storage/es/mappings/jaeger-span-7.json#L72).

This is  actually causing an explosion fixedbit sets memory usage. More about Nested fields and why they should be avoided unless required explicitly of querying [here](https://www.elastic.co/guide/en/elasticsearch/reference/current/nested.html#nested).

If we change the type of logs field to object and store the logs as flattened key value pairs - It saving lot of heap memory. We have done some analysis on the same results below:

Current state of one our ES Nodes ( where we store Jaeger spans)

Overal Heap Used - 

image

Heap used for FixedBit Sets

image

Out of 22GB of used heap space, 7G is used for fixedBit sets storage

We did an exercise with a 50 GB jaeger-span index. We updated the index template ( Changed the log field mapping to object type) and reindexed the span  index to another index. We could see a huge drop in fixedBit Sets Memory usage.

image

This index was taking up around 250MB in memory to store fixedBit cache - after the change its down to 19 MB. As Nested fields are stored as separate documents - the documents count also reduced drastically - Disk storage remained constant as expected

Nested fields are stored as seperate documents and references (offset details etc) are stored in memory. This leads to increase in memory. We have cases where we add more 50 logs for a span record. All these logs are then getting indexed as separate documents and contribute to fieldBit set cache. 

Can we change the mapping for logs field to object instead nested. Thoughts?

CC - @albertteoh

bhiravabhatla:
- Resolves #3085 

- Change index mapping for logs field to object type from nested field.

### DIFF
--- a/plugin/storage/es/mappings/fixtures/jaeger-span-7.json
+++ b/plugin/storage/es/mappings/fixtures/jaeger-span-7.json
@@ -65,29 +65,46 @@
         "type":"integer"
       },
       "logs":{
-        "type":"nested",
         "dynamic":false,
-        "properties":{
-          "timestamp":{
-            "type":"long"
-          },
-          "fields":{
-            "type":"nested",
+        "type": "object",
+        "enabled":true,
+        "properties": {
+          "fields": {
             "dynamic":false,
-            "properties":{
-              "key":{
-                "type":"keyword",
-                "ignore_above":256
+            "type": "object",
+            "enabled":true,
+            "properties": {
+              "type": {
+                "type": "text",
+                "fields": {
+                  "keyword": {
+                    "ignore_above": 256,
+                    "type": "keyword"
+                  }
+                }
               },
-              "value":{
-                "type":"keyword",
-                "ignore_above":256
+              "value": {
+                "type":"text",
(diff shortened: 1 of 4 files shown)


jpkrohling:
The tests seem to be failing due to the changes from this PR.

bhiravabhatla:
@jpkrohling  - Yep - there is an on going discussion on the issue thread. Need direction on the same - until its resolved, cant go ahead with this PR. https://github.com/jaegertracing/jaeger/issues/3085

jpkrohling:
What's the status of this PR?

PR #3091, opened 2021-06-15
Add MetricsQueryService  grcp handler

albertteoh:

- Relates to #2954.

- Adds the gRPC handler.

### DIFF
--- a/cmd/query/app/grpc_handler.go
+++ b/cmd/query/app/grpc_handler.go
@@ -17,6 +17,7 @@ package app
 import (
 	"context"
 	"errors"
+	"time"
 
 	"github.com/opentracing/opentracing-go"
 	"go.uber.org/zap"
@@ -51,19 +52,7 @@ type GRPCHandler struct {
 	metricsQueryService querysvc.MetricsQueryService
 	logger              *zap.Logger
 	tracer              opentracing.Tracer
-	clock               clock
-}
-
-// NewGRPCHandler returns a GRPCHandler
-func NewGRPCHandler(queryService *querysvc.QueryService, metricsQueryService querysvc.MetricsQueryService, logger *zap.Logger, tracer opentracing.Tracer, clock clock) *GRPCHandler {
-	gH := &GRPCHandler{
-		queryService:        queryService,
-		metricsQueryService: metricsQueryService,
-		logger:              logger,
-		tracer:              tracer,
-		clock:               clock,
-	}
-	return gH
+	nowFn               func() time.Time
 }
 
 // GetTrace is the gRPC handler to fetch traces based on trace-id.
@@ -260,10 +249,10 @@ func (g *GRPCHandler) handleErr(msg string, err error) error {
 	g.logger.Error(msg, zap.Error(err))
 
 	// Avoid wrapping "expected" errors with an "Internal Server" error.
-	switch {
-	case errors.Is(err, disabled.ErrDisabled):
+	if errors.Is(err, disabled.ErrDisabled) {
 		return errGRPCMetricsQueryDisabled
-	case errors.Is(err, errMissingServiceNames), errors.Is(err, errMissingQuantile):
+	}
+	if _, ok := status.FromError(err); ok {
 		return err
 	}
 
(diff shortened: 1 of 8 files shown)


cmd/query/app/grpc_handler.go
```diff
@@ -35,21 +39,30 @@ const (
...
+	queryService        *querysvc.QueryService
+	metricsQueryService querysvc.MetricsQueryService
+	logger              *zap.Logger
+	tracer              opentracing.Tracer
+	clock               clock
 }
 
 // NewGRPCHandler returns a GRPCHandler
-func NewGRPCHandler(queryService *querysvc.QueryService, logger *zap.Logger, tracer opentracing.Tracer) *GRPCHandler {
+func NewGRPCHandler(queryService *querysvc.QueryService, metricsQueryService querysvc.MetricsQueryService, logger *zap.Logger, tracer opentracing.Tracer, clock clock) *GRPCHandler {
```
albertteoh, line 58: is it worth adopting the functional options pattern here given the list of params is growing?

yurishkuro: Lately I've been trying to stay away from functional options and just use an options/params struct, it's less work to implement and has better discoverability.

yurishkuro: In this concrete example, we don't even need any additional struct, just remove this function and initialize GRPCHandler directly

cmd/query/app/grpc_handler.go
```diff
@@ -177,3 +190,123 @@ func (g *GRPCHandler) GetDependencies(ctx context.Context, r *api_v2.GetDependen
 
 	return &api_v2.GetDependenciesResponse{Dependencies: dependencies}, nil
 }
+
+// GetLatencies is the gRPC handler to fetch latency metrics.
+func (g *GRPCHandler) GetLatencies(ctx context.Context, r *metrics.GetLatenciesRequest) (*metrics.GetMetricsResponse, error) {
```
yurishkuro, line 195: recently we've seen some tickets showing that apparently with some client libs it's possible to send a request where r==nil, which crashes the server.

cmd/query/app/grpc_handler.go
```diff
@@ -177,3 +190,123 @@ func (g *GRPCHandler) GetDependencies(ctx context.Context, r *api_v2.GetDependen
...
+	if err == nil {
+		return nil
+	}
+	g.logger.Error(msg, zap.Error(err))
+
+	// Avoid wrapping "expected" errors with an "Internal Server" error.
+	switch {
+	case errors.Is(err, disabled.ErrDisabled):
+		return errGRPCMetricsQueryDisabled
+	case errors.Is(err, errMissingServiceNames), errors.Is(err, errMissingQuantile):
```
yurishkuro, line 266: rather than enumerating, could we check if it's already an instance of status.Error?

cmd/query/app/grpc_handler.go
yurishkuro: I see you already have this check in g.newBaseQueryParameters, but should call this first then

cmd/query/app/util.go
```diff
@@ -15,9 +15,20 @@
 package app
 
 import (
+	"time"
+
 	"github.com/jaegertracing/jaeger/storage/spanstore"
 )
 
+// clock is a wrapper around the time package to allow mocking of the current time.
```
yurishkuro, line 23: You can avoid this type by using the function signature directly `func() time.Time`, and then pass `time.Now` as main impl

yurishkuro (approved):
lgtm, sans minor nits

cmd/query/app/grpc_handler.go
albertteoh: yeah, at least the grpc clients generated in this project fail to marshal a nil message over the wire; but I trust in what you've seen from other client libs, so I've added a nil check for the request.

Note, the original nil check in `g.newBaseQueryParameters` was just for the `MetricsQueryBaseRequest`, which is a common attribute among all the request objects, but it doesn't check the parent request for nil.

cmd/query/app/grpc_handler.go
```diff
@@ -177,3 +180,135 @@ func (g *GRPCHandler) GetDependencies(ctx context.Context, r *api_v2.GetDependen
...
+	}
+	var baseRequest *metrics.MetricsQueryBaseRequest
+	switch v := r.(type) {
+	case *metrics.GetLatenciesRequest:
+		baseRequest = v.BaseRequest
+	case *metrics.GetCallRatesRequest:
+		baseRequest = v.BaseRequest
+	case *metrics.GetErrorRatesRequest:
+		baseRequest = v.BaseRequest
+	}
```
albertteoh, line 276: is there a better way?

The reason for taking this approach is to avoid the same nil check in each of the endpoint handlers (`GetLatencies`, etc...) so they can just pass the parent request object in, and have a single place to perform the nil check.
 
The trade-off is given each of these request objects belong to different types, but I need to access the common `BaseRequest` out of each request, a type switch was necessary.

yurishkuro (approved):

albertteoh:
Thanks Yuri!

PR #52, opened 2017-02-27
Introduce query API handler and query parser

badiib:

PR #3095, opened 2021-06-16
Add HTTP handler for metrics querying

albertteoh:

- Relates to #2954.

- Adds the HTTP handlers for metrics querying.


cmd/query/app/handler_archive_test.go
```diff
@@ -36,7 +36,7 @@ func TestGetArchivedTrace_NotFound(t *testing.T) {
 	for _, tc := range []spanstore.Reader{nil, mockReader} {
 		archiveReader := tc // capture loop var
 		t.Run(fmt.Sprint(archiveReader), func(t *testing.T) {
-			withTestServer(t, func(ts *testServer) {
```
albertteoh, line 39: `t` was never used.

cmd/query/app/http_handler.go
```diff
@@ -115,6 +127,10 @@ func (aH *APIHandler) RegisterRoutes(router *mux.Router) {
 	// TODO - remove this when UI catches up
 	aH.handleFunc(router, aH.getOperationsLegacy, "/services/{%s}/operations", serviceParam).Methods(http.MethodGet)
 	aH.handleFunc(router, aH.dependencies, "/dependencies").Methods(http.MethodGet)
+	aH.handleFunc(router, aH.latencies, "/metrics/latencies/{%s}", servicesParam).Methods(http.MethodGet)
```
yurishkuro, line 130: Are you particularly set on passing service in the URL?  We tried to move away from that (see L126)

cmd/query/app/json_marshaler.go
```diff
@@ -0,0 +1,74 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+
+	"github.com/gogo/protobuf/jsonpb"
+	"github.com/gogo/protobuf/proto"
+)
+
+type jsonMarshaler interface {
```
yurishkuro, line 26: nit: no need to change that now, but you could've declared this just as a function and returned functions directly from newXXX methods, without any intermediate structs.

cmd/query/app/http_handler.go
```diff
@@ -295,6 +299,155 @@ func (aH *APIHandler) dependencies(w http.ResponseWriter, r *http.Request) {
...
+	ok := true
+	if formVal := r.FormValue(paramName); formVal != "" {
+		var err error
+		parsedValue, err = strconv.ParseBool(formVal)
+		ok = !aH.handleParseError(w, err, paramName)
+	}
+	return parsedValue, ok
+}
+
+func (aH *APIHandler) parseTimeParam(w http.ResponseWriter, r *http.Request, paramName string, defaultTime *time.Time) (t time.Time, ok bool) {
```
yurishkuro, line 399: would these functions be better placed into queryParser?

cmd/query/app/json_marshaler.go
albertteoh: ah yes, similar to what we did with `time.Now`

cmd/query/app/http_handler.go
albertteoh: Nope, I'm not particularly attached to passing service as a URL path segment. Just thought that having services in the URL path is an affordance to suggest it is a mandatory parameter.

What was the reason behind moving away from it? I only found this comment, but couldn't find the discussion on why there was final agreement on moving in this direction: https://github.com/jaegertracing/jaeger/pull/52#discussion_r103296286

yurishkuro: > What was the reason behind moving away from it?

1. part of my holy war on high cardinality REST URLs (when params are encoded in the URL we can't use URL as the span name)
2. when the service name is part of the URL people sometimes tend to think it can go there unencoded, which breaks when the service name contains a slash

in the end though - not very strong reasons

cmd/query/app/http_handler.go
```diff
@@ -115,6 +126,10 @@ func (aH *APIHandler) RegisterRoutes(router *mux.Router) {
 	// TODO - remove this when UI catches up
 	aH.handleFunc(router, aH.getOperationsLegacy, "/services/{%s}/operations", serviceParam).Methods(http.MethodGet)
 	aH.handleFunc(router, aH.dependencies, "/dependencies").Methods(http.MethodGet)
+	aH.handleFunc(router, aH.latencies, "/metrics/latencies").Methods(http.MethodGet)
```
albertteoh, line 129: removed the services from the URL path.

cmd/query/app/http_handler.go
```diff
@@ -295,6 +299,60 @@ func (aH *APIHandler) dependencies(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	structuredRes := structuredResponse{
+		Data: minStep.Milliseconds(),
+	}
+	aH.writeJSON(w, r, &structuredRes)
+}
+
+func (aH *APIHandler) metrics(w http.ResponseWriter, r *http.Request, getMetrics func(context.Context, metricsstore.BaseQueryParameters) (*metrics.MetricFamily, error)) {
+	requestParams, err := aH.queryParser.parseMetricsQueryParams(r)
```
albertteoh, line 345: moved param parsing logic into `queryParser`.

cmd/query/app/query_parser.go
```diff
@@ -76,15 +93,15 @@ type traceQueryParameters struct {
 //     key := strValue
 //     keyValue := strValue ':' strValue
 //     tags :== 'tags=' jsonMap
-func (p *queryParser) parse(r *http.Request) (*traceQueryParameters, error) {
+func (p *queryParser) parseTraceQueryParams(r *http.Request) (*traceQueryParameters, error) {
 	service := r.FormValue(serviceParam)
 	operation := r.FormValue(operationParam)
 
-	startTime, err := p.parseTime(startTimeParam, r)
+	startTime, err := p.parseTime(r, startTimeParam, time.Microsecond)
```
albertteoh, line 100: the `search` endpoint defines times in microseconds since epoch (why?), but other endpoints such as `dependencies` and `metrics` use milliseconds since epoch, since those are the time units JS (i.e. Jaeger UI) use e.g. `Date.now()`, so we need to pass the units to the `parseTime` function to support both cases.

cmd/query/app/query_parser.go
```diff
@@ -143,31 +160,125 @@ func (p *queryParser) parse(r *http.Request) (*traceQueryParameters, error) {
...
+	bqp.EndTime = &endTs
+	bqp.Lookback = &lookback
+	bqp.Step = &step
+	bqp.RatePer = &ratePer
+	return bqp, err
+}
+
+// parseTime parses the time parameter of an HTTP request that is represented the number of "units" since epoch.
+// If the time parameter is empty, the current time will be returned.
+func (p *queryParser) parseTime(r *http.Request, paramName string, units time.Duration) (time.Time, error) {
```
albertteoh, line 203: as above, `units` are required here to support both micros and millis since epoch.

cmd/query/app/query_parser.go
```diff
@@ -143,31 +160,125 @@ func (p *queryParser) parse(r *http.Request) (*traceQueryParameters, error) {
...
+func (p *queryParser) parseDuration(r *http.Request, paramName string, units *time.Duration, defaultDuration time.Duration) (d time.Duration, err error) {
+	d = defaultDuration
+	formValue := r.FormValue(paramName)
+	switch {
+	case formValue == "":
+		return d, nil
+
+	// If no units are supplied, assume parsing of duration strings like 5ms.
+	case units == nil:
+		if d, err = time.ParseDuration(formValue); err == nil {
```
albertteoh, line 231: the `search` endpoint defines durations as duration strings, presumably to support human-friendly clients like curl, etc. so this maintains support for such duration parameter values.

albertteoh (commented):
I've tidied a few things up:
- use encapsulation for duration parsing
- move query param parsing logic for each endpoint into `queryParser`
- fix up comments
- unexport unnecessarily exported var
- wrap errors

PTAL.

cmd/query/app/query_parser.go
yurishkuro: We should add this to comments in the code. The search operates on span latencies, which are expressed as microseconds in the data model, so it makes sense to support high accuracy. The search UI itself does not insist on exact units because it supports string like 1ms. We had a debate over whether units should be handled by the UI instead of the backend service, but here we are, since Go makes parsing 1ms very easy.

The dependencies API does not operate on the latency space, instead its timestamps are just time range selections, and the typical backend granularity of those is on the order of 15min or more. So microseconds aren't really that useful in this domain, although I certainly would've preferred having a consistent time representation.

Metrics is a new domain, you need to decide which representation makes more sense. I think it's closer to dependencies, where you're asking about wall clock time, not the latency domain.

albertteoh: > The search operates on span latencies, which are expressed as microseconds in the data model

Out of curiosity, why micros and not posix nanos which I think also fit into an int64? Were there some limitations in storage or clients to handle posix nanos?

> Metrics is a new domain, you need to decide which representation makes more sense. I think it's closer to dependencies, where you're asking about wall clock time, not the latency domain.

I think milliseconds are most appropriate (I've documented these reasons in comments within `queryParser`):
- The main client in mind is Jaeger UI. Being a react.js app, its built-in time precision is milliseconds so it simplifies usage for Jaeger UI.
- The min step size is `1ms` for the Prometheus-compliant storage backends I've tested.
- It also follows that data point timestamps are reported with a millisecond precision, at least from my tests so far.

cmd/query/app/query_parser.go
```diff
@@ -46,22 +49,62 @@ const (
...
+	if err != nil {
+		return 0, err
+	}
+	return time.Duration(i) * (dup.units), nil
 }
 
-// parse takes a request and constructs a model of parameters
+// parseTraceQueryParams takes a request and constructs a model of parameters.
+//
+// Why start/end parameters are expressed in microseconds:
```
albertteoh, line 99: @yurishkuro I've added your comments here for posterity. Please advise if they would be better placed else where.

cmd/query/app/query_parser.go
```diff
@@ -143,36 +187,154 @@ func (p *queryParser) parse(r *http.Request) (*traceQueryParameters, error) {
...
+		return dqp, err
+	}
+
+	dqp.lookback, err = parseDuration(r, lookbackParam, durationUnitsParser{units: time.Millisecond}, defaultDependencyLookbackDuration)
+	return dqp, err
+}
+
+// parseMetricsQueryParams takes a request and constructs a model of metrics query parameters.
+//
+// Why the API is designed using an end time (endTs) and lookback:
```
albertteoh, line 207: Also added reasons for various parameter decisions here. Again, if you can think of a better place to put these, I can move these.

cmd/query/app/http_handler_test.go
```diff
@@ -514,23 +517,26 @@ func TestGetServicesSuccess(t *testing.T) {
...
+	defer ts.server.Close()
 	expectedOperations := []spanstore.Operation{{Name: ""}, {Name: "get", SpanKind: "server"}}
-	readMock.On(
+	ts.spanReader.On(
 		"GetOperations",
 		mock.AnythingOfType("*context.valueCtx"),
-		spanstore.OperationQueryParameters{ServiceName: "abc/trifle"},
+		spanstore.OperationQueryParameters{
+			ServiceName: "abc/trifle",
+			SpanKind:    "server",
```
albertteoh, line 538: here's the assertion to catch my earlier breaking API change.

cmd/query/app/query_parser.go
yurishkuro: > Out of curiosity, why micros and not posix nanos which I think also fit into an int64? Were there some limitations in storage or clients to handle posix nanos?

legacy from zipkin origins, which still only supports micros, whereas our domain model uses Go's time.Time which has nanos. But our Thrift API uses micros.

cmd/query/app/query_parser.go
```diff
@@ -76,15 +121,16 @@ type traceQueryParameters struct {
 //     key := strValue
 //     keyValue := strValue ':' strValue
 //     tags :== 'tags=' jsonMap
-func (p *queryParser) parse(r *http.Request) (*traceQueryParameters, error) {
+func (p *queryParser) parseTraceQueryParams(r *http.Request) (*traceQueryParameters, error) {
+	dp := durationStringParser{}
```
yurishkuro, line 125: declared too far from usage

cmd/query/app/query_parser.go
```diff
@@ -46,22 +49,64 @@ const (
...
+		spanstore.TraceQueryParameters
+		traceIDs []model.TraceID
+	}
+
+	dependenciesQueryParameters struct {
+		endTs    time.Time
+		lookback time.Duration
+	}
+
+	durationParser interface {
```
yurishkuro, line 73: embrace functional programming - this is just a function, and the two parsers can be easily implemented to return functions, without using any structs

cmd/query/app/query_parser.go
```diff
@@ -143,36 +189,169 @@ func (p *queryParser) parse(r *http.Request) (*traceQueryParameters, error) {
...
+		return b, newParseError(err, paramName)
+	}
+	return b, nil
+}
+
+// parseSpanKindParam parses the input span kinds to filter for in the metrics query.
+// Valid input span kinds are the string representations from the OpenTelemetry model/proto/metrics/otelspankind.proto.
+// For example:
+// - "SPAN_KIND_SERVER"
+// - "SPAN_KIND_CLIENT"
```
yurishkuro, line 329: is this consistent with the strings we used in the Operations API?  I thought we used strings client/server, as defined in OpenTracing.

cmd/query/app/json_marshaler.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+
+	"github.com/gogo/protobuf/jsonpb"
+	"github.com/gogo/protobuf/proto"
+)
+
+type jsonMarshaler = func(writer io.Writer, response interface{}) error
```
albertteoh, line 26: I decided to follow your advice and embrace functional programming here as well; it looks much more concise :)

cmd/query/app/query_parser.go
albertteoh: Good pickup, I couldn't find formal docs in comments but found evidence where "server" and "client" are used in the [operations API unit tests](https://github.com/jaegertracing/jaeger/blob/master/cmd/query/app/http_handler_test.go#L581-L582). 

Added a mapping from jaeger/opentracing span kinds to OTEL span kinds.

cmd/query/app/query_parser.go
```diff
@@ -46,22 +49,68 @@ const (
...
+		spanstore.TraceQueryParameters
+		traceIDs []model.TraceID
+	}
+
+	dependenciesQueryParameters struct {
+		endTs    time.Time
+		lookback time.Duration
+	}
+
+	durationParser = func(s string) (time.Duration, error)
```
albertteoh, line 82: converted the duration parsers to functions. I did initially try to adopt functional programming but was thrown by the fact that the unit duration parser needed to maintain state (the duration units), but hadn't thought that we can get around this via closures. 

Thanks for pointing this out, @yurishkuro, it looks a lot neater, and I've learnt something new! :)

cmd/query/app/query_parser_test.go
```diff
@@ -171,3 +172,153 @@ func TestParseTraceQuery(t *testing.T) {
...
+}
+
+func TestParseRepeatedSpanKinds(t *testing.T) {
+	q := "x?service=foo&spanKind=unspecified&spanKind=internal&spanKind=server&spanKind=client&spanKind=producer&spanKind=consumer"
+	request, err := http.NewRequest(http.MethodGet, q, nil)
+	require.NoError(t, err)
+	parser := &queryParser{
+		timeNow: func() time.Time {
+			return time.Now()
+		},
```
yurishkuro, line 243:
```suggestion
		timeNow: time.Now
```

cmd/query/app/query_parser.go
```diff
@@ -143,36 +193,184 @@ func (p *queryParser) parse(r *http.Request) (*traceQueryParameters, error) {
...
+		return defaultSpanKinds, nil
+	}
+	otelSpanKinds, err := mapSpanKinds(jaegerSpanKinds)
+	if err != nil {
+		return defaultSpanKinds, newParseError(err, paramName)
+	}
+	return otelSpanKinds, nil
+}
+
+func mapSpanKinds(spanKinds []string) ([]string, error) {
```
yurishkuro, line 360:
```suggestion
func mapSpanKindsToOpenTelemetry(spanKinds []string) ([]string, error) {
```

cmd/query/app/query_parser.go
```diff
@@ -143,36 +193,184 @@ func (p *queryParser) parse(r *http.Request) (*traceQueryParameters, error) {
...
+	otelSpanKinds := make([]string, len(spanKinds))
+	var ok bool
+	for i, spanKind := range spanKinds {
+		if otelSpanKinds[i], ok = jaegerToOtelSpanKind[spanKind]; !ok {
+			return otelSpanKinds, fmt.Errorf("unsupported span kind: '%s'", spanKind)
```
yurishkuro, line 365: nit (more narrow var scope):
```suggestion
	for i, spanKind := range spanKinds {
		if v, ok := jaegerToOtelSpanKind[spanKind]; ok {
		    otelSpanKinds[i] = v
		} else {
			return otelSpanKinds, fmt.Errorf("unsupported span kind: '%s'", spanKind)
```

yurishkuro (approved):

albertteoh:
Thanks for the review, Yuri!

Issue #2987, opened 2021-05-11
Jaeger is OOMKilled when use badger as storage


Sallyan:
**version**: all-in-one:1.18.1
**issue**: 
when i use badge as storage, Jaeger requests much more memory than in-memory storage and it keeps OOMKiller. 
At the beginning Jaeger requests just hundred MB memory when it uses in-memory storage. But it requests over 5Gi memory when changed to Badge. 
**Jaeger yaml:**
```
apiVersion: jaegertracing.io/v1
kind: Jaeger
metadata:
  name: tracing-jaeger
  namespace: kyma-system
spec:
  agent:
    config: {}
    options: {}
    resources: {}
  allInOne:
    config: {}
    image: eu.gcr.io/kyma-project/external/jaegertracing/all-in-one:1.18.1
    options:
      log-level: info
    resources: {}
  annotations:
    sidecar.istio.io/inject: "true"
    sidecar.istio.io/rewriteAppHTTPProbers: "true"
  collector:
    config: {}
    options: {}
    resources: {}
  ingester:
    config: {}
    options: {}
    resources: {}
  ingress:
    enabled: false
    openshift: {}
    options: {}
    resources: {}
    security: none
  query:
    options: {}
    resources: {}
  resources:
    limits:
      cpu: 500m
      memory: 8Gi
    requests:
      cpu: 200m
      memory: 6Gi
  sampling:
    options: {}
  storage:
    cassandraCreateSchema: {}
    dependencies:
      resources: {}
      schedule: 55 23 * * *
    elasticsearch:
      nodeCount: 3
      redundancyPolicy: SingleRedundancy
      resources:
        limits:
          memory: 16Gi
        requests:
          cpu: "1"
          memory: 16Gi
      storage: {}
    esIndexCleaner:
      numberOfDays: 7
      resources: {}
      schedule: 55 23 * * *
    esRollover:
      resources: {}
      schedule: 0 0 * * *
    options:
      badger:
        directory-key: /badger/key
        directory-value: /badger/data
        ephemeral: false
        span-store-ttl: 24h
        truncate: true
      cassandra:
        keyspace: jaeger_v1_datacenter3
        servers: cassandra.default.svc
      es:
        server-urls: http://elasticsearch-client.default.svc:9200
      memory:
        max-traces: 10000
    type: badger
  strategy: allinone
  ui:
    options:
      dependencies:
        menuEnabled: true
      menu:
      - items:
        - label: Documentation
[...]

PR #3096, opened 2021-06-17
Badger: v1 to v3 upgrade

Ashmita152:

Relates to https://github.com/jaegertracing/jaeger/issues/2459 https://github.com/jaegertracing/jaeger/issues/2987

- Badger v3 removes Truncate  and TableLoadingMode options https://github.com/dgraph-io/badger/pull/1555
- Badger v3 makes MemoryMap as the default loading option https://github.com/dgraph-io/badger/pull/733

## Benchmarks

Badger is well known for high memory usage. We are using very version of badger which makes it almost unusable. I did some benchmark between badger-v1 and badger-v3 using all-in-one image and tracegen. 

#### Badger v1 benchmarks
Started all-in-one with badger-v1 backend from master:
```
❯ SPAN_STORAGE_TYPE=badger ~/all-in-one-badger-v1 --badger.ephemeral=false --badger.directory-value=/tmp/badger/v1/data --badger.directory-key=/tmp/badger/v1/key  --badger.span-store-ttl=1m --badger.consistency --badger.maintenance-interval=1m
```
Started tracegen:
```
❯ go run ./cmd/tracegen -pause 50us -duration 1m
```
**Result**:

From htop:
```
CPU: 107%
Residential Memory: 3840MB
```
```
❯ du -sh /tmp/badger/v1
741M    /tmp/badger/v1
```

#### Badger v3 benchmarks
Started all-in-one with badger-v3 backend from PR branch:
```
❯ SPAN_STORAGE_TYPE=badger ~/all-in-one-badger-v3 --badger.ephemeral=false --badger.directory-value=/tmp/badger/v3/data --badger.directory-key=/tmp/badger/v3/key  --badger.span-store-ttl=1m --badger.consistency --badger.maintenance-interval=1m
```
Started tracegen:
```
❯ go run ./cmd/tracegen -pause 50us -duration 1m
```
**Result**:

From htop:
```
CPU: 6%
Residential Memory: 677MB
```
```
❯ du -sh /tmp/badger/v3
48M    /tmp/badger/v3
```

I see good performance gain in terms of all three resources: cpu, memory and disk.

Although this is a breaking change but I think we should move in this direction. We know badger makes breaking changes in every major version release.


albertteoh:
If we want to avoid a breaking change, we could deprecate support for badger v1 (no longer maintain it, along with clear warning logs) and create another span storage type badger_v3. Then eventually decommission badger v1 after a few releases.

pavolloffay:
How do people migrate data from version to version in badger? Is there any way to migrate? Would it require writing a custom tool?

plugin/storage/badger/options.go
```diff
@@ -133,11 +132,6 @@ func addFlags(flagSet *flag.FlagSet, nsConfig NamespaceConfig) {
 		nsConfig.MetricsUpdateInterval,
 		"How often the badger metrics are collected by Jaeger. Format is time.Duration (https://golang.org/pkg/time/#Duration)",
 	)
-	flagSet.Bool(
-		nsConfig.namespace+suffixTruncate,
-		nsConfig.Truncate,
-		"If write-ahead-log should be truncated on restart. this will cause data loss.",
```
yurishkuro, line 139: Let's Follow our deprecation policy, keep this flag and log a warning when it is used

yurishkuro (commented):
LGTM. I don't think we should keep v1 around.

Please edit the change log and describe the breaking changes, such as renaming of metrics.

Ashmita152:
Thank you everyone for the review.  I will work on fixing the build failures and other feedbacks.

> How do people migrate data from version to version in badger? Is there any way to migrate? Would it require writing a custom tool?
Reading through the dgraph github issues, it seems like they suggest to do backup and restore which seems compatible across different versions.

Ashmita152:
This PR is ready for review again. Thank you.

plugin/storage/badger/factory.go
```diff
@@ -117,7 +115,9 @@ func (f *Factory) Initialize(metricsFactory metrics.Factory, logger *zap.Logger)
 		opts.ValueDir = f.Options.Primary.ValueDirectory
 
 		// These options make no sense with ephemeral data
-		opts.Truncate = f.Options.Primary.Truncate
+		if f.Options.Primary.Truncate {
```
yurishkuro, line 118: Since this flag has no effect anymore, we should limit its existence to `options.go`. There's no need to translate it from CLI flags to a struct, we can just warn right there.

Ashmita152: I have a question:
 In options.go there is no logger object. Hence to add a warning log, we need to add a logger which will need to change `Configurable` interface's `InitFromViper` signature to have a logger object and hence will change in every type that is implementing that interface. Do you think this is the right way to achieve it, I wanted to check if I am thinking in right direction.

yurishkuro: yes, it's fine, if you look at all InitFromViper there are some that take a logger for this reason:
```
./cmd/flags/admin.go:77:func (s *AdminServer) initFromViper(v *viper.Viper, logger *zap.Logger) {
./cmd/agent/app/reporter/flags.go:55:func (b *Options) InitFromViper(v *viper.Viper, logger *zap.Logger) *Options {
./cmd/query/app/flags.go:103:func (qOpts *QueryOptions) InitFromViper(v *viper.Viper, logger *zap.Logger) *QueryOptions {
```

Ashmita152: Thank you so much @yurishkuro

Ashmita152:
This PR is ready for review again.

Ashmita152:
This PR is ready for review again.

plugin/storage/badger/options.go
```diff
@@ -44,6 +45,9 @@ type NamespaceConfig struct {
 	ReadOnly              bool          `mapstructure:"read_only"`
 }
 
+// TODO deprecated flag to be removed
+const truncateWarning = "(deprecated since v1.24.0, will be removed in release v1.26.0)"
```
albertteoh, line 49: nit: it may be worth including the time dimension too (3 months notice) as in the example from [CONTRIBUTING.md](https://github.com/jaegertracing/jaeger/blob/master/CONTRIBUTING.md#deprecating-cli-flags), which would be more pertinent if we decide to have more frequent releases; or did we decide to only use release versions to determine the deprecation notice period?

plugin/storage/badger/options.go
```diff
@@ -160,6 +164,9 @@ func initFromViper(cfg *NamespaceConfig, v *viper.Viper) {
 	cfg.MetricsUpdateInterval = v.GetDuration(cfg.namespace + suffixMetricsInterval)
 	cfg.Truncate = v.GetBool(cfg.namespace + suffixTruncate)
 	cfg.ReadOnly = v.GetBool(cfg.namespace + suffixReadOnly)
+	if cfg.Truncate {
```
albertteoh, line 167: nit: using `v.IsSet(cfg.namespace + suffixTruncate)` would notify users of the deprecation if they were explicitly doing `--badger.truncate=false`, even though it's redundant, but who knows. :)

albertteoh (commented):
a couple of nits, otherwise LGTM.

plugin/storage/badger/options.go
```diff
@@ -136,7 +140,7 @@ func addFlags(flagSet *flag.FlagSet, nsConfig NamespaceConfig) {
 	flagSet.Bool(
 		nsConfig.namespace+suffixTruncate,
 		nsConfig.Truncate,
```
yurishkuro, line 138: Shouldn't this field be removed from config struct ?

yurishkuro (approved):

jpkrohling:
> Reading through the badger github issues, it seems like they suggest to do backup and restore which seems compatible across different versions.

Could you add a link to this in the changelog? So that people know where the migration instructions are located.

Ashmita152:
I tried the steps and thought to be include as part of documentation: https://github.com/jaegertracing/documentation/pull/507

PR #3111, opened 2021-06-28
Switch to user nobody for esCleaner and esRollover

majidazimi:
`esCleaner.py` and `esRollover.py` run under user `root` which causes problem when setting pod/container security context (`runAsNonRoot: true`). Since base image is alpine and it already has user `nobody`, it would be good to switch the user to facilitate this.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -65,7 +65,7 @@ require (
 	golang.org/x/net v0.0.0-20210614182718-04defd469f4e
 	golang.org/x/sys v0.0.0-20210615035016-665e8c7367d1
 	google.golang.org/grpc v1.38.0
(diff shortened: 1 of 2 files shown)


majidazimi:
I believe there are some problems in the pipeline configuration. I see such messages:
```
+ '[' '!' -f .docker.io.login ']'
+ '[' -v DOCKERHUB_TOKEN ']'
+ printenv DOCKERHUB_TOKEN
+ docker login docker.io --username jaegertracingbot --password-stdin
```

albertteoh:
@majidazimi please add your changes to a feature branch from your fork, avoid making changes to your `master`.

PR #3112, opened 2021-06-28
Switch to user nobody for es cleaner and rollover

majidazimi:
`esCleaner.py` and `esRollover.py` run under user root which causes problem when setting pod/container security context (runAsNonRoot: true). Since base image is alpine and it already has user nobody, it would be good to switch the user to facilitate this.

### DIFF
--- a/plugin/storage/es/Dockerfile
+++ b/plugin/storage/es/Dockerfile
@@ -6,4 +6,5 @@ RUN pip install urllib3==1.24.3
 RUN pip install elasticsearch elasticsearch-curator
 COPY esCleaner.py /es-index-cleaner/
 
+USER nobody
 ENTRYPOINT ["python3", "/es-index-cleaner/esCleaner.py"]
--- a/plugin/storage/es/Dockerfile.rollover
+++ b/plugin/storage/es/Dockerfile.rollover
@@ -8,4 +8,5 @@ COPY ./mappings/* /mappings/
 COPY esRollover.py /es-rollover/
 COPY esmapping-generator /usr/bin/
 
+USER nobody
 ENTRYPOINT ["python3", "/es-rollover/esRollover.py"]


plugin/storage/es/Dockerfile
```diff
@@ -6,4 +6,5 @@ RUN pip install urllib3==1.24.3
 RUN pip install elasticsearch elasticsearch-curator
 COPY esCleaner.py /es-index-cleaner/
 
+USER nobody
```
albertteoh, line 9: (disclaimer: I'm no expert in k8s)
Is it a good practice to use the username `nobody` instead of the numeric uid `65534`?

I see a few references which seem to suggest using the numeric representation [[1]](https://github.com/helm/helm/issues/4818) [[2]](https://git.blindage.org/Kubernetes/hetzner-dns/commit/606260d49483b5adcdef1a71767f8df13b2b5d0f) [[3]](https://github.com/kubernetes/kubernetes/issues/40958).

majidazimi: makes sense. I switched to explicit user id.

albertteoh: Thanks! Did you manage to test that `USER 65534` works with `runAsNonRoot: true` for the esCleaner and esRollover? I'm not sure we have integration tests covering k8s.

albertteoh: I also suggest adding a similar comment to [this PR](https://git.blindage.org/Kubernetes/hetzner-dns/commit/606260d49483b5adcdef1a71767f8df13b2b5d0f) explaining why we need the `nobody` user and why the numeric uid is used, for posterity.

jpkrohling: Another reference: https://github.com/open-telemetry/opentelemetry-collector/pull/2838

jpkrohling:
Could you please rebase this?

PR #3139, opened 2021-07-14
Fix base path in grpc gateway for api_v3

pavolloffay:


cmd/query/app/apiv3/grpc_gateway.go
```diff
@@ -34,7 +34,11 @@ func RegisterGRPCGateway(ctx context.Context, logger *zap.Logger, r *mux.Router,
 	grpcGatewayMux := runtime.NewServeMux(
 		runtime.WithMarshalerOption(runtime.MIMEWildcard, jsonpb),
 	)
-	r.PathPrefix("/v3/").Handler(http.StripPrefix(basePath, grpcGatewayMux))
```
pavolloffay, line 37: this was causing the issue. The default base path is passed as `/`. The `StripPrefix` is noop only when the base path is empty and the old tests were testing only empty base path.

jpkrohling (approved):

yurishkuro (approved):

PR #3169, opened 2021-07-28
Set lookback in ES rollover to distant past

### LINKED ISSUE #3152: Traces in a rolled-over index are unable to be hotlinked via /trace/{id}
**Describe the bug**

- Search a trace in your 'history' that should be in an index that is in a rollover'd index. 
- You are able to find this via the search and click on it. This gives a working view of your trace at `/trace/{your-id}`
- Open a new tab
- Insert the link
- You get a 404 when directly going to the /`/trace/{your-id}` page

When using a trace that is 'fresh' i.e. has not been in a rollover index, the `/trace/{your-id}` works.

**Expected behavior**
Directly 'hotlinking' to any `/trace/{your-id}` should work. Regardless if this lives in a roll-over'd index.

 - Jaeger version: 1.21

So many things, but your own tracing and debug lvls are empty.

**Additional context**
I'm aware of other issues before, I have searched them. The key difference here is:
- We can find the traces via search. It's merely based on the direct link to `/trace/{your-id}`

I think, but I got no evidence for this, that the direct page on `/trace/{your-id}` does not account for the flag `es.use-aliases`

pavolloffay:
Resolves #3152

@albertteoh could you please review?

### DIFF
--- a/plugin/storage/es/spanstore/reader.go
+++ b/plugin/storage/es/spanstore/reader.go
@@ -129,10 +129,17 @@ type SpanReaderParams struct {
 
 // NewSpanReader returns a new SpanReader with a metrics.
 func NewSpanReader(p SpanReaderParams) *SpanReader {
+	maxSpanAge := p.MaxSpanAge
+	// If rollover is enabled set lookback far into the past
+	// In rollover only read alias is used to query the data so looking far into the past should
+	// not affect performance.
+	if p.UseReadWriteAliases {
+		maxSpanAge = time.Hour * 24 * 365 * 100
+	}
 	return &SpanReader{
 		client:                        p.Client,
 		logger:                        p.Logger,
-		maxSpanAge:                    p.MaxSpanAge,
+		maxSpanAge:                    maxSpanAge,
 		serviceOperationStorage:       NewServiceOperationStorage(p.Client, p.Logger, 0), // the decorator takes care of metrics
 		spanIndexPrefix:               indexNames(p.IndexPrefix, spanIndex),
 		serviceIndexPrefix:            indexNames(p.IndexPrefix, serviceIndex),
--- a/plugin/storage/es/spanstore/reader.go
+++ b/plugin/storage/es/spanstore/reader.go
@@ -60,6 +60,8 @@ const (
 	tagValueField          = "value"
 
 	defaultNumTraces = 100
+
+	rolloverMaxSpanAge = time.Hour * 24 * 365 * 100
 )
 
 var (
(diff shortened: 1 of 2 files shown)


plugin/storage/es/spanstore/reader.go
```diff
@@ -129,10 +129,17 @@ type SpanReaderParams struct {
 
 // NewSpanReader returns a new SpanReader with a metrics.
 func NewSpanReader(p SpanReaderParams) *SpanReader {
+	maxSpanAge := p.MaxSpanAge
+	// If rollover is enabled set lookback far into the past
+	// In rollover only read alias is used to query the data so looking far into the past should
+	// not affect performance.
```
albertteoh, line 135: I suggest we reword this to describe why `maxSpanAge` is set to a large number. A suggestion:
```
Setting the maxSpanAge to a large duration will ensure all spans in the "read" alias are accessible by queries (query window = now - maxSpanAge).
When read/write aliases are enabled, required for index rollovers, only the "read" alias is queried and therefore should not affect performance.
```

plugin/storage/es/spanstore/reader.go
```diff
@@ -129,10 +129,17 @@ type SpanReaderParams struct {
...
+	if p.UseReadWriteAliases {
+		maxSpanAge = time.Hour * 24 * 365 * 100
```
albertteoh, line 137: I think we should extend `TestNewSpanReader`, maybe converting it to a table-driven test, to prevent a regression on this functionality.

pavolloffay: +1

pavolloffay:
@albertteoh PR updated

plugin/storage/es/spanstore/reader_test.go
```diff
@@ -132,14 +131,33 @@ func withArchiveSpanReader(readAlias bool, fn func(r *spanReaderTest)) {
...
+				MaxSpanAge:          time.Hour * 72,
+				UseReadWriteAliases: true,
+			},
+			maxSpanAge: time.Hour * 24 * 365 * 100,
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			reader := NewSpanReader(test.params)
+			require.NotNil(t, reader)
```
albertteoh, line 158: i think you're missing the assertion on `test.maxSpanAge` here (I would rename it `test.wantMaxSpanAge` to make it clear it's the expected value)?

pavolloffay: doh, fixed

albertteoh (approved):

PR #3171, opened 2021-08-01
Add ATM dev environment docker-compose and API doc

### LINKED ISSUE #3107: Metrics Query Development Environment

As part of #2954, I'd like to ensure there is a decent development environment that users and contributors alike can easily spin up/down, something like jaeger all-in-one.

The metrics querying capabilities of jaeger-query are made possible through the availability of OTEL-generated span metrics data, so a number of components need to be spun up in order to have "instant" graphs seen in the new Monitor tab, namely:

- Something to generate traces, ideally for multiple services and operations at varying rates and properties to make the graphs more interesting to look at.
- OTEL collector to receive these spans then compute and expose these metrics.
- A Prometheus server to scrape these metrics and expose endpoints for querying the metrics data.
- Jaeger all-in-one providing the ability to query and view the metrics and drill-down into the related traces.

It wouldn't be feasible (AFAIK) to have these components combined into a single container like jaeger all-in-one.
I propose to introduce a docker-compose file defining the aforementioned components, much like the existing [jaeger-docker-compose.yml](https://github.com/jaegertracing/jaeger/blob/master/docker-compose/jaeger-docker-compose.yml), and perhaps in the same directory with a Makefile target to easily bring up the dev environment.

The docker-compose file would look something like this, with some modifications like using official rather than personal images: https://github.com/albertteoh/atm-dev/blob/main/docker-compose.yml

Note that for generating traces, I provide two options in the above sample [docker-compose.yml file](https://github.com/albertteoh/atm-dev/blob/main/docker-compose.yml):
- [synthetic-load-generator](https://github.com/Omnition/synthetic-load-generator) by Splunk, which is now an archived repo. The reason why I have a personal image is due to a config-related bug in the upstream repo which I fixed in my local fork.
- HotROD

The benefit of synthetic-load-generator is that it generates fairly complex and configurable traces at varying rates of calls, errors and durations, without human intervention; unlike HotROD. This is desirable, particularly when viewing metrics as there is a continuous stream of metrics updates over time.

The problem is the repo is archived (and so no longer maintained) and contains a bug (which I've fixed locally) and so would involve pushing an alternative image managed by us.

HotROD, on the other hand, is familiar to users, is maintained by the community, and publishes official tagged images, but requires user intervention to generate traces and metrics.

albertteoh:

Resolves #3107

- Adds a docker-compose file with instructions on how to bring up a development/demo environment for the metrics query feature.
- Additionally documents the metrics query API and available query parameters with example curl commands.

### DIFF
--- a/docker-compose/monitor/README.md
+++ b/docker-compose/monitor/README.md
@@ -1,0 +1,173 @@
+# atm-dev
+Development environment for ATM
+
+# Getting Started
+
+## Bring up/down the dev environment
+```
+docker compose up
+docker compose down
+```
+
+## Example 1
+Fetch call rates for both the driver and frontend services, grouped by operation, from now,
+looking back 1 second with a sliding rate-calculation window of 1m and step size of 1 millisecond
+
+```bash
+curl http://localhost:16686/api/metrics/calls\?service\=driver\&service\=frontend\&groupByOperation\=true\&endTs\="$(date +%s)"000\&lookback\=1000\&step\=100\&ratePer\=60000 | jq .
+```
+
+
+## Example 2
+Fetch P95 latencies for both the driver and frontend services from now,
+looking back 1 second with a sliding rate-calculation window of 1m and step size of 1 millisecond, where the span kind is either "server" or "client".
+
+```bash
+curl http://localhost:16686/api/metrics/latencies\?service\=driver\&service\=frontend\&quantile\=0.95\&endTs\="$(date +%s)"000\&lookback\=1000\&step\=100\&ratePer\=60000\&spanKind\=server\&spanKind\=client | jq .
+```
+
+## Example 3
+Fetch error rates for both driver and frontend services using default parameters.
+```bash
+curl http://localhost:16686/api/metrics/errors\?service\=driver\&service\=frontend | jq .
+```
+
+## Example 4
+Fetch the minimum step size supported by the underlying metrics store.
+```bash
+curl http://localhost:16686/api/metrics/minstep | jq .
+```
+
+# HTTP API
+
+## Query Metrics
+
+`/api/metrics/{metric_type}?{query}`
+
+Where (Backus-Naur form):
+```
+metric_type = 'latencies' | 'calls' | 'errors'
+
+query = services , [ '&' optionalParams ]
+
+optionalParams = param | param '&' optionalParams
+
+param =  groupByOperation | quantile | endTs | lookback | step | ratePer | spanKinds
+
+services = service | service '&' services
+service = 'service=' strValue
+  - The list of services to include in the metrics selection filter, which are logically 'OR'ed.
+  - Mandatory.
+
+quantile = 'quantile=' floatValue
+  - The quantile to compute the latency 'P' value. Valid range (0,1].
+  - Mandatory for 'latencies' type.
+
+groupByOperation = 'groupByOperation=' boolValue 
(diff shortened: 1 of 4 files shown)


docker-compose/monitor/README.md
```diff
@@ -0,0 +1,173 @@
+# atm-dev
+Development environment for ATM
```
yurishkuro, line 2: what is ATM? Please spell out.

albertteoh: Thanks for picking up on this, I've elaborated a little more on what ATM stands for, the architecture of the development/demo environment and explanation of each component with relevant links, and an illustration of how the components interact with one another.

docker-compose/monitor/README.md
```diff
@@ -0,0 +1,183 @@
...
+curl http://localhost:16686/api/metrics/errors\?service\=driver\&service\=frontend | jq .
+```
+
+## Example 4
+Fetch the minimum step size supported by the underlying metrics store.
+```bash
+curl http://localhost:16686/api/metrics/minstep | jq .
+```
+
+# HTTP API
```
albertteoh, line 51: I'm planning to add this section into https://www.jaegertracing.io/docs/latest/apis/, would that be an appropriate place or is it sufficient to leave this API doc here?

docker-compose/monitor/docker-compose.yml
```diff
@@ -0,0 +1,39 @@
+version: "3.5"
+services:
+  jaeger:
+    networks:
+      - backend
+    image: jaegertracing/all-in-one:latest
```
jpkrohling, line 6: Is this file being tested via CI? If not, it might be better to use a fixed version.

docker-compose/monitor/docker-compose.yml
```diff
@@ -0,0 +1,39 @@
...
+    ports:
+      - "14250:14250"
+      - "14268:14268"
+      - "6831:6831/udp"
+      - "16686:16686"
+      - "16685:16685"
+  otel_collector:
+    networks:
+      - backend
+    image: otel/opentelemetry-collector-contrib:latest
```
jpkrohling, line 19: Same here.

docker-compose/monitor/docker-compose.yml
```diff
@@ -0,0 +1,39 @@
...
+    volumes:
+      - "./otel-collector-config.yml:/etc/otelcol/otel-collector-config.yml"
+    command: --config /etc/otelcol/otel-collector-config.yml
```
jpkrohling, line 22: Not a compose expert, but shouldn't this depend on the jaeger one being up?

docker-compose/monitor/docker-compose.yml
albertteoh: > Is this file being tested via CI?

Nope.

> it might be better to use a fixed version.

I'd like to consider one hypothetical scenario, keeping in mind that the ATM/Monitor feature is still under active development (the UI isn't ready yet).

We set the jaeger all-in-one tag to v1.24 here. If someone makes a change now and merges to `master` before v1.25 is released, if I understand correctly, will users need to wait until v1.26 to obtain/use the features just merged?

docker-compose/monitor/docker-compose.yml
albertteoh: yup, thanks! 😊

jpkrohling (approved):

albertteoh:
Thanks for the review Yuri/Juca!

PR #3190, opened 2021-08-06
Add dependency management files for Python scripts

jpkrohling:
When building those scripts for Red Hat's build of Jaeger, I faced some dependency problems and had to use a setup similar to what this PR contains. The changes in this PR might be useful to more folks, especially those looking for reprodubicle builds.

﻿Signed-off-by: Juraci Paixão Kröhling <juraci@kroehling.de>


plugin/storage/es/setup.cfg
```diff
@@ -0,0 +1,3 @@
+[metadata]
+name = jaeger-es
+version = 1.25.0
```
yurishkuro, line 3: How important is it to have an accurate version number here? Feels like a serious overhead to the process to update this very minor corner.

plugin/storage/es/requirements.txt
```diff
@@ -0,0 +1,56 @@
+#
```
yurishkuro, line 1: Where is this file used in the build?

plugin/storage/es/setup.cfg
jpkrohling: I believe this is relevant to consumers of this package, if this package was to be published at some central repository, which isn't happening for these scripts.

plugin/storage/es/requirements.txt
jpkrohling: Right, sorry, this should have been part of the Dockefiles. PR updated.

plugin/storage/es/setup.cfg
pavolloffay: Our python scripts are not published anywhere. Can we omit this file?

jpkrohling: I'm not a Python expert, but apparently, some tools do expect Python projects to contain at least a setup.cfg and the requirements.txt files. Perhaps we can have 0.0.1 as the version?

pavolloffay:
Could you please explain why do we need `requirements-build.txt`, `requirements.txt` and `requirements-build.in`? Most of the tutorials I found use only `requirements.txt`

jpkrohling:
I added a few words about them to the readme file.

plugin/storage/es/setup.cfg
pavolloffay: the question is which tool requires setup.cfg in jaeger. Is it https://github.com/jazzband/pip-tools?

pavolloffay:
That helps thanks. If my understanding is right we need all those 3 files for https://github.com/jazzband/pip-tools. Then the higher-level question is why do we need https://github.com/jazzband/pip-tools? The simplest python project does not require this 3rd party build tool.

plugin/storage/es/Dockerfile
```diff
@@ -1,11 +1,11 @@
 FROM python:3-alpine3.11
 
-# Temporary fix for https://github.com/jaegertracing/jaeger/issues/1494
-RUN pip install urllib3==1.24.3
-# Temporary fix for https://github.com/jaegertracing/jaeger/issues/3179
-RUN pip install elasticsearch==7.13.0
+COPY requirements-build.txt requirements-build.txt
+COPY requirements.txt requirements.txt
```
yurishkuro, line 4: if we're always copying both files in prod output there is no reason to separate build from main requirements, since they are both required for prod.

yurishkuro (commented):
.cfg file doesn't bother me as long as we don't need to track the version.

plugin/storage/es/Dockerfile
jpkrohling: Correct, but downstream (like our Dockerfiles) might run on hermetic environments so that we need to ensure that all build requirements are installed before we install the runtime dependencies. For instance, the "cython" requirement may be resolved to be installed after "elasticsearch-curator", as the latter doesn't list the former in its list of (runtime) dependencies.

We use "Cachito" internally, and this is a link to the explanation on their side: https://github.com/release-engineering/cachito/blob/master/docs/pip.md#build-dependencies

plugin/storage/es/setup.cfg
jpkrohling: In my specific case, it's Cachito, but I assume this might be true for other similar tools: https://github.com/release-engineering/cachito/blob/master/docs/pip.md#setuppy-setupcfg

jpkrohling:
>Then the higher-level question is why do we need https://github.com/jazzband/pip-tools?

Developers looking to update a dependency would need this `pip-tools` package in order to recompile the list of dependencies based (`.txt` files) based on the input files (`.in`). The `.txt` files should not be manually changed.

jpkrohling:
>If my understanding is right we need all those 3 files

Just a small correction: there are 4 files, being two input files, and two generated files.

jpkrohling:
I'm closing this, as the Python scripts are not being replaced by Golang implementations.

PR #3193, opened 2021-08-06
Add SendGetBodyAs flag on elasticsearch storage

### LINKED ISSUE #3191: Unable to find traces with Elasticsearch deployed behind Google Cloud load balancer
**Describe the bug**
When using a Google Cloud HTTP load balancer in front of our Elastic Search cluster, Jaeger Query component is unable to search for traces.

**To Reproduce**
1. Deploy an elasticsearch cluster in Google Cloud
2. Deploy a load balancer in front of the cluster with an HTTPS route
3. Configure jaeger-query to use the elasticsearch cluster
4. Search for traces

**Expected behavior**
I should be able to search for traces

**Screenshots**
After intercepting HTTP traffic between jaeger-query and elasticsearch, this is the query blocked by the load balancer with a 400 error:
```
GET https://elasticsearch.xxxx/_msearch?rest_total_hits_as_int=true HTTP/1.1
Host: elasticsearch.xxxx
User-Agent: elastic/6.2.37 (windows-amd64)
Accept: application/json
Content-Type: application/json
Accept-Encoding: gzip

{"ignore_unavailable":true,"index":"jaeger-jaeger-span-2021-08-06"}
{"query":{"bool":{"must":{"term":{"traceID":"ca1eb6d7b60a624a97acd85c4ccf9ad1"}}}},"search_after":[1628247456972000],"size":10000,"sort":[{"startTime":{"order":"asc"}}],"terminate_after":10000}
{"ignore_unavailable":true,"index":"jaeger-jaeger-span-2021-08-06"}
```

**Version:**
 - Tested on Linux and Windows
 - Jaeger version: 1.25
 - Kubernetes, Docker and .exe

**Additional context**
Google Cloud Load balancer does not support body in HTTP GET queries and blocks it. The ElasticSearch client should be able to make queries with POST instead of GET (maybe if a flag is activated). Meanwhile, we're using TCP load balancer but it's not a best practice.
https://cloud.google.com/load-balancing/docs/https#illegal_request_and_response_handling

NatMarchand:
Signed-off-by: Nathanael Marchand <nathanael.marchand@outlook.com>

Resolves #3191 

This PR configures the property SendGetBodyAs on the elastic search client to allow sending HTTP requests with a body with another verb than GET (which might cause networking issue on some WAF or load balancers).
In order to do that, I've created an option -es.send-get-body-as to specify the verb to use.

I open this PR early to gather feedback as Go is not my native development language ;)

### DIFF
--- a/plugin/storage/es/options.go
+++ b/plugin/storage/es/options.go
@@ -59,6 +59,7 @@ const (
 	suffixVersion                        = ".version"
 	suffixMaxDocCount                    = ".max-doc-count"
 	suffixLogLevel                       = ".log-level"
+	suffixSendGetBodyAs                  = ".send-get-body-as"
 	// default number of documents to return from a query (elasticsearch allowed limit)
 	// see search.max_buckets and index.max_result_window
 	defaultMaxDocCount        = 10_000
@@ -68,6 +69,7 @@ const (
 	defaultIndexDateSeparator = "-"
 
 	defaultIndexRolloverFrequency = "day"
+	defaultSendGetBodyAs          = ""
 )
 
 // TODO this should be moved next to config.Configuration struct (maybe ./flags package)
@@ -110,6 +112,7 @@ func NewOptions(primaryNamespace string, otherNamespaces ...string) *Options {
 		RemoteReadClusters:   []string{},
 		MaxDocCount:          defaultMaxDocCount,
 		LogLevel:             "error",
+		SendGetBodyAs:        defaultSendGetBodyAs,
 	}
 	options := &Options{
 		Primary: namespaceConfig{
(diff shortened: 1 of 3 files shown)


plugin/storage/es/options.go
```diff
@@ -265,6 +268,10 @@ func addFlags(flagSet *flag.FlagSet, nsConfig *namespaceConfig) {
 		nsConfig.namespace+suffixLogLevel,
 		nsConfig.LogLevel,
 		"The Elasticsearch client log-level. Valid levels: [debug, info, error]")
+	flagSet.String(
+		nsConfig.namespace+suffixSendGetBodyAs,
+		nsConfig.SendGetBodyAs,
+		"The Elasticsearch client SendGetBodyAs value.")
```
yurishkuro, line 274: I would put a bit more explanation in the description

yurishkuro (commented):
looks reasonable. Please run `make fmt` to fix formatting issues.

plugin/storage/es/options.go
NatMarchand: sure, will do

NatMarchand:
I've formatted code, improved the description such as suggested. I'm not sure if I have to add tests before removing Draft ?

yurishkuro (commented):
usually we have some minimum test to ensure that the value is parsed properly into the struct

yurishkuro (approved):

yurishkuro:
Thanks!

NatMarchand:
Thank you for your quick feedback and approval ! Glad to have contributed :)

Issue #3044, opened 2021-05-31
Official OpenSearch storage option


wiardvanrij:
## Requirement - what kind of business use case are you trying to solve?

At the moment Jaeger supports Cassandra and Elasticsearch (ES).

With recent changes in ES license, AWS created a fork (OpenSearch).
For long-term commitment and architectural decisions I would like to know Jaeger's thoughts/vision/commitment towards OpenSearch as storage. 

I.e.: Should we migrate from our current AWS ES 'PaaS' towards a dedicated 'ES' setup, rather than upgrading towards OpenSearch

## Problem - what in Jaeger blocks you from solving the requirement?

The problem I could foresee is that we hit a compatibility issue on OpenSearch side, where Jaeger only supports ES. It would require a change in Jaeger, which might not be feasible as, at the moment, Jaeger does not support OpenSearch officially. 

Now one could argue that "But the specs are the same" - this is still the case for _now_. Things change. 
For example Elasticsearch could implement new features which in return Jaeger starts using, while OpenSearch goes a different direction. At this point we lose compatibility. 

## Any open questions to address

I would like to point out this change: https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1005
The sole reason is to: 
> This PR needs to ensure we are sending data to licensed ES.

Now this does not impact Jaeger at all, but it does make a statement that ES has no intentions to create an ecosystem where multiple projects, implementing the same API, are compatible. 
Therefore I strongly believe OpenSearch should be officially acknowledged as separated storage option since we cannot say "ES compatible engine". - With defining the separation, it still leaves the question open if this should get official support as Jaeger compatible storage.

pavolloffay:
Great topic! :) 

Let me recap the current state: Jaeger at the moment supports ES 5, 6 and 7. And therefore should be compatible with the latest OpenSearch version which if I am not mistaken it is based on ES7.

To me, it makes sense to add official OpenSearch support and start testing against official images. Though, I am not sure what the situation will be in the long run. Officially supporting both adds maintenance cost. I am personally curios if there will be native go client for OpenSearch. Jaeger currently uses https://github.com/olivere/elastic and we were planning to migrate to https://github.com/elastic/go-elasticsearch.

@jkowall can probably loop in as well and comment on OpenSearch support (at least from logz.io perspective).

jkowall:
[...]

PR #3198, opened 2021-08-10
Automatically detect OpenSearch version

pavolloffay:
Updates #3044 

* Automatically set ES version to 7 if OpenSearch 1x is used
* Add OpenSearch to itests

### DIFF
--- a/plugin/storage/integration/elasticsearch_test.go
+++ b/plugin/storage/integration/elasticsearch_test.go
@@ -21,6 +21,7 @@ import (
 	"net/http"
 	"os"
 	"strconv"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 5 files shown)


pavolloffay:
@albertteoh would you like to review?

yurishkuro (dismissed):

jpkrohling:
Are the test failures expected?

plugin/storage/integration/elasticsearch_test.go
```diff
@@ -68,6 +69,12 @@ func (s *ESStorageIntegration) getVersion() (uint, error) {
 	if err != nil {
 		return 0, err
 	}
+	// OpenSearch is based on ES 7.x
+	if strings.Contains(pingResult.TagLine, "OpenSearch") {
```
pavolloffay, line 73: this is unfortunate we should refactor the test to use the pkg/es/config

pavolloffay:
> Are the test failures expected?

no it needed a bit more love. It should be fixed now and it needs re-approval.

pavolloffay:
still failed, I will have a look

pavolloffay:
alright so the rollover script needed the "opensearch" logic as well.

albertteoh (approved):

PR #3211, opened 2021-08-16
Add CLI Option for gRPC Max Receive Message Size

### LINKED ISSUE #3210: gRPC Collector Does Not Support Configurable Max Message Size
**Describe the bug**
The Jaeger gRPC Collector Service has a default max receive message size of 4mb and does not support overriding this max size even though gRPC allows for it.

Source from `cmd/collector/app/server/server.go`:
```go
var defaultServerOptions = serverOptions{
	maxReceiveMessageSize: defaultServerMaxReceiveMessageSize,
	maxSendMessageSize:    defaultServerMaxSendMessageSize,
	connectionTimeout:     120 * time.Second,
	writeBufferSize:       defaultWriteBufSize,
	readBufferSize:        defaultReadBufSize,
}
```

Where `defaultServerMaxReceiveMessageSize` is defined as:
```go
const (
	defaultServerMaxReceiveMessageSize = 1024 * 1024 * 4
```

There is a method available to set the max receive message size also but it is not called and has no configuration options associated with it:
```go
// MaxRecvMsgSize returns a ServerOption to set the max message size in bytes the server can receive.
// If this is not set, gRPC uses the default 4MB.
func MaxRecvMsgSize(m int) ServerOption {
	return newFuncServerOption(func(o *serverOptions) {
		o.maxReceiveMessageSize = m
	})
}
```

**To Reproduce**
See my initial question on this in the Discussion section: [gRPC Collector Configurable Max Message Size #3200](https://github.com/jaegertracing/jaeger/discussions/3200)

**Expected behavior**
Some users of Jaeger (such as myself) may want the ability to log larger Spans and since gRPC is flexible and allows a larger message size limit, we should expose a setting to allow overriding the default max message size.

For example, a new CLI option for the Collector:
```
collector.grpc-max-receive-message-length
```

To change this behavior, we basically just need to add a new Collector CLI option as mentioned and then add the following code to `cmd/collector/app/server/grpc.go -> StartGRPCServer`:
```go
// StartGRPCServer based on the given parameters
func StartGRPCServer(params *GRPCServerParams) (*grpc.Server, error) {
	var server *grpc.Server
	var grpcOpts []grpc.ServerOption

        // New - set max receive message size via new params property populated via CLI option
	grpcOpts = append(grpcOpts, grpc.MaxRecvMsgSize(params.MaxReceiveMessageLength))

	if params.TLSConfig.Enabled {
		// user requested a server with TLS, setup creds
		tlsCfg, err := params.TLSConfig.Config(params.Logger)
		if err != nil {
			return nil, err
		}
```

We'd also have to update the `GRPCServerParams` type and some other code as well but that's the most important part of it.

 - OS: Linux
 - Jaeger version: Jaeger v1.21.0
 - Deployment: Jaeger deployed via Kubernetes but traced application is running on bare metal (Linux via WSL)

See my initial question on this in the Discussion section: [gRPC Collector Configurable Max Message Size #3200](https://github.com/jaegertracing/jaeger/discussions/3200)

js8080:
Resolves #3210 

Added a new CLI option for the gRPC Collector to override the default
max receive message size which is normally 4MB since some users of
Jaeger may want to record Spans which exceed the default max size.

Signed-off-by: Justin Stauffer <justin.stauffer@gmail.com>

### DIFF
--- a/cmd/collector/app/builder_flags.go
+++ b/cmd/collector/app/builder_flags.go
@@ -35,7 +35,7 @@ const (
 	collectorZipkinAllowedHeaders        = "collector.zipkin.allowed-headers"
 	collectorZipkinAllowedOrigins        = "collector.zipkin.allowed-origins"
 	collectorZipkinHTTPHostPort          = "collector.zipkin.host-port"
-	collectorGRPCMaxReceiveMessageLength = "collector.grpc.max-receive-message-length"
+	collectorGRPCMaxReceiveMessageLength = "collector.grpc-server.max-message-size"
 )
 
 var tlsGRPCFlagsConfig = tlscfg.ServerFlagsConfig{
--- a/cmd/collector/app/builder_flags.go
+++ b/cmd/collector/app/builder_flags.go
@@ -35,7 +35,7 @@ const (
 	collectorZipkinAllowedHeaders        = "collector.zipkin.allowed-headers"
 	collectorZipkinAllowedOrigins        = "collector.zipkin.allowed-origins"
 	collectorZipkinHTTPHostPort          = "collector.zipkin.host-port"
-	collectorGRPCMaxReceiveMessageLength = "collector.grpc-max-receive-message-length"
+	collectorGRPCMaxReceiveMessageLength = "collector.grpc.max-receive-message-length"
 )
 
 var tlsGRPCFlagsConfig = tlscfg.ServerFlagsConfig{
--- a/cmd/collector/app/builder_flags.go
+++ b/cmd/collector/app/builder_flags.go
@@ -26,15 +26,16 @@ import (
 )
 
 const (
-	collectorDynQueueSizeMemory   = "collector.queue-size-memory"
-	collectorGRPCHostPort         = "collector.grpc-server.host-port"
-	collectorHTTPHostPort         = "collector.http-server.host-port"
-	collectorNumWorkers           = "collector.num-workers"
-	collectorQueueSize            = "collector.queue-size"
-	collectorTags                 = "collector.tags"
-	collectorZipkinAllowedHeaders = "collector.zipkin.allowed-headers"
-	collectorZipkinAllowedOrigins = "collector.zipkin.allowed-origins"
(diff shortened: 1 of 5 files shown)


cmd/collector/app/builder_flags.go
```diff
@@ -26,15 +26,16 @@ import (
...
+	collectorDynQueueSizeMemory          = "collector.queue-size-memory"
+	collectorGRPCHostPort                = "collector.grpc-server.host-port"
+	collectorHTTPHostPort                = "collector.http-server.host-port"
+	collectorNumWorkers                  = "collector.num-workers"
+	collectorQueueSize                   = "collector.queue-size"
+	collectorTags                        = "collector.tags"
+	collectorZipkinAllowedHeaders        = "collector.zipkin.allowed-headers"
+	collectorZipkinAllowedOrigins        = "collector.zipkin.allowed-origins"
+	collectorZipkinHTTPHostPort          = "collector.zipkin.host-port"
+	collectorGRPCMaxReceiveMessageLength = "collector.grpc.max-receive-message-length"
```
yurishkuro, line 38: given that we already have a grouping `collector.grpc-server`, I think a better name would be
```
collector.grpc-server.max-message-size
```

Ashmita152:
You would need to raise the PR from another branch of your fork (other than master) to fix these CI failures.

js8080:
> You would need to raise the PR from another branch of your fork (other than master) to fix these CI failures.

Oh, shoot... I didn't realize that. Should I just delete this whole PR and then re-do it?

cmd/collector/app/builder_flags.go
js8080: I will do that.

js8080:
> > You would need to raise the PR from another branch of your fork (other than master) to fix these CI failures.
> 
> Oh, shoot... I didn't realize that. Should I just delete this whole PR and then re-do it?

I made a new PR in https://github.com/jaegertracing/jaeger/pull/3214 from a branch on my fork instead of from master. Closing this PR.

Issue #3209, opened 2021-08-16
Refactor CI jobs to use goreleaser


Ashmita152:
Our current CI is too complicated to understand.

## Requirement - what kind of business use case are you trying to solve?
I am creating this issue to start discussion around CI jobs refactoring. From my perspective, I can see following problems in our current CI implementation:

1) We do most of the stuffs ourselves in shell scripts which build tools like [goreleaser](https://github.com/goreleaser/goreleaser) can do for us. With buildx coming into picture, it has become even more complicated.
2) We have one huge Makefile which keeps on growing.
3) Lack of documentation around build scripts.

## Proposal - what do you suggest to solve the problem or improve the existing situation?
I would like to propose following solutions for the above problems:
1) Let [goreleaser](https://github.com/goreleaser/goreleaser) do the heavy lifting for all our binary build, docker build and docker push related tasks. @jpkrohling already tried it for v2 in https://github.com/jaegertracing/jaeger/pull/2886 and it looked much cleaner.
2) Divide root Makefile into multiple Makefiles. 
    - Add each jaeger component specific directives to each `cmd/<app>/Makefile` and include them in the root Makefile.
    - With goreleaser into play, lot of directives should get removed.
3) Document the flow of build jobs.

Let me know your thoughts.

yurishkuro:
I don't have a strong opinion about goreleaser. What makes it better? From the description it seems a sizeable portion of it is dealing with pushing GH release & homebrew.

I am not sure what splitting the Makefile gives us. The amount of content per binary is actually minimal in it. I would rather deal with a single 600-lines Makefile (which isn't that large) than hunt around for the same information in multiple dirs.

I think the last round of refactoring for buildx actually consolidated a lot of repeated steps into smaller number of scripts, which is great. The build process is still, indeed, pretty complicated.

Ashmita152:
The trigger point for opening this issue is https://github.com/jaegertracing/jaeger/pull/3204 It clearly shows our CI is quite rigid. It works perfectly well on daily basis PRs but when someone need to change something in those shell scripts, it requires quite an effort and requires complete knowledge since we have too much coupling among those scripts.  Amount of conversation in https://github.com/jaegertracing/jaeger/pull/2611 and https://github.com/jaegertracing/jaeger/pull/2948 reflect that pretty well.

Regarding goreleaser, I can see a few benefit:
[...]

PR #3213, opened 2021-08-17
Add docker buildx make target

pavolloffay:
Updates #3209 

This PR adds make target to configure docker buildx command. The idea is to allow contributors to run the build locally the same way how it runs on CI.

### DIFF
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -21,3 +21,13 @@ create-debugimg:
 		--build-arg golang_image=$(GOLANG_IMAGE) \
 		--platform=$(PLATFORMS) \
 		docker/debug
+
+.PHONY: prepare-docker-buildx
+prepare-docker-buildx:
+	 docker buildx create --use --name=jaeger-build --buildkitd-flags="--allow-insecure-entitlement security.insecure --allow-insecure-entitlement network.host" --driver-opt="network=host"
+	 docker run --rm -d -p 5000:5000 --name registry registry:2
+
+.PHONY: clean-docker-buildx
+clean-docker-buildx:
+	docker buildx rm jaeger-build
+	docker rm -f registry
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -9,14 +9,14 @@ PLATFORMS := linux/amd64,linux/s390x,linux/ppc64le,linux/arm64
 
 create-baseimg-debugimg: create-baseimg create-debugimg
 
-create-baseimg:
+create-baseimg: prepare-docker-buildx
 	docker buildx build -t $(BASE_IMAGE) --push \
 		--build-arg root_image=$(ROOT_IMAGE) \
 		--build-arg cert_image=$(CERT_IMAGE) \
 		--platform=$(PLATFORMS) \
 		docker/base
 
-create-debugimg:
+create-debugimg: prepare-docker-buildx
 	docker buildx build -t $(DEBUG_IMAGE) --push \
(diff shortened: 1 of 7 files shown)


docker/Makefile
```diff
@@ -21,3 +21,13 @@ create-debugimg:
 		--build-arg golang_image=$(GOLANG_IMAGE) \
 		--platform=$(PLATFORMS) \
 		docker/debug
+
+.PHONY: prepare-docker-buildx
+prepare-docker-buildx:
```
albertteoh, line 26: This is a nice idea. It would be even nicer if it was a seamless experience for local dev. For example, if I run the following locally:
```
make docker-images-jaeger-backend
```

I still get the error, because the local registry isn't started as part of this make target:
```
error: failed to solve: rpc error: code = Unknown desc = failed to do request: Head "http://localhost:5000/v2/baseimg_alpine/blobs/sha256:48167268581ca7f86ce8f794a24d144b0509556cf4c584eec9e95afebe88bcc6": dial tcp [::1]:5000: connect: connection refused
```

Given the only time we need the registry is when we do a `push`, would it make sense to add this target as a dependency to both `create-baseimg` and `create-debugimg`? i.e.

```
create-baseimg: prepare-docker-buildx
...
create-debugimg: prepare-docker-buildx
...
```

That way, developers won't need to know to run `make prepare-docker-buildx` for their docker builds to work.

If the above is feasible, it also means we could do away with the `- run: make prepare-docker-buildx` within our github actions `steps`, given `create-baseimg` and/or `create-debugimg` _should_ be run.

The drawback I see is that `make prepare-docker-buildx` isn't idempotent and I don't have any good suggestions right now to solve for it, except for making sure that `make clean-docker-buildx` is always run at the end.

The other alternative is we document this extra manual step of `make prepare-docker-buildx` clearly somewhere.

pavolloffay: btw. are you on macos? Were you able to run it?

albertteoh: Yes, I'm on macos and successfully ran this make target together with other docker buildx.

pavolloffay: I didn't mind requiring devs to execute `make prepare-docker-buildx` explicitly. It's quite easy to understand what needs to be run from the scripts.

I have made it idempotent in the last commit and removed prepare step from the ci jobs.


For instance IIRC `install-tools` target is not invoked automatically before a target that uses tools.

pavolloffay:
@albertteoh could you please review?

docker/Makefile
albertteoh: > It's quite easy to understand what needs to be run from the scripts.

yes, I agree, but knowing to where to find the scripts (I presume you mean the scripts run by github actions) seems less intuitive than looking in the Makefile since that's where the make targets are defined. For example, they may just want to build custom docker images for testing purposes and run `make docker-images-jaeger-backend`, and not have any desire to replicate CI locally.

albertteoh (approved):
works like a charm, and nice to see less GHA config as well.

btw, looks like you need to rebase with master, meaning it would invalidate my approval. I'll re-approve after you rebase, if I'm still awake :)

docker/Makefile
pavolloffay: When I am trying to build an unknown project I usually first go to CI config to see what is required to run. If there is a clear "prepare" step I can follow it and understand what needs to be done. 

If the "prepare" for OOTB then it's good as well until it does not interfere with tools and services running on the host. In this case it might be the docker registry for instance.

Issue #3080, opened 2021-06-10
Schema support for cassandra 4


nothinux:
Hi, i'm trying to initilize schema in cassandra 4, but i got `'Unknown property 'dclocal_read_repair_chance'`.
after i check changelog for cassandra 4, `dclocal_read_repair_chance` has removed [link](https://github.com/apache/cassandra/blob/6c30261bc8d696f599a1e9ebea8a5bffbd0ac9ed/NEWS.txt#L255).

So temporarily I removed `AND dclocal_read_repair_chance=0.0` in plugin/storage/cassandra/schema/v003.cql.tmpl, and it works well.

i know cassandra 4 still RC, i'm interested to contribute, would sending a pull request v004.cql.tmpl similar to v003.cql.tmpl but without the `AND dclocal_read_repair_chance=0.0` solve the problem? or is there something else to do?

Thank you

albertteoh:
You may need to provide a migration script from v003 -> v004 as well like https://github.com/jaegertracing/jaeger/blob/master/plugin/storage/cassandra/schema/migration/V002toV003.sh.

PR #3225, opened 2021-08-23
Add cassandra v4 support

Ashmita152:
Resolves https://github.com/jaegertracing/jaeger/issues/3080

We don't need to add v003tov004.sh because the change (removal of dclocal_read_repair_chance config) is in table metadata which gets handled seemlessly within cassandra upgrade (tested locally) when users upgrade their cassandra cluster from v3.11 to v4.0. So there won't be any scenario, where existing user will need migrate their data.

### DIFF
--- a/.github/workflows/ci-cassandra.yml
+++ b/.github/workflows/ci-cassandra.yml
@@ -15,11 +15,11 @@ jobs:
         - distribution: cassandra
           major: 3.x
           image: 3.11
-          schema: v001 v002 v003
+          schema: v003
         - distribution: cassandra
           major: 4.x
           image: 4.0
-          schema: v004 v004 v004
+          schema: v004
     name: ${{ matrix.version.distribution }} ${{ matrix.version.major }}
     steps:
     - uses: actions/checkout@v2.3.4
@@ -31,18 +31,8 @@ jobs:
     - name: Run cassandra integration tests
       run: bash scripts/cassandra-integration-test.sh ${{ matrix.version.schema }}
     services:
-      cassandra1:
+      cassandra:
         image: ${{ matrix.version.distribution }}:${{ matrix.version.image }}
         ports:
           - 9042:9042
           - 9160:9160
-      cassandra2:
-        image: ${{ matrix.version.distribution }}:${{ matrix.version.image }}
-        ports:
-          - 9043:9042
-          - 9161:9160
-      cassandra3:
-        image: ${{ matrix.version.distribution }}:${{ matrix.version.image }}
-        ports:
-          - 9044:9042
-          - 9162:9160
--- a/.github/workflows/ci-cassandra.yml
+++ b/.github/workflows/ci-cassandra.yml
@@ -9,6 +9,16 @@ on:
 jobs:
   cassandra:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        version:
+        - image: cassandra
+          tag: 3.11
+          schema: v001 v002 v003
+        - image: cassandra
+          tag: 4.0
+          schema: v004 v004 v004
+    name: ${{ matrix.version.image }} ${{ matrix.version.tag }}
     steps:
     - uses: actions/checkout@v2.3.4
 
@@ -17,4 +27,20 @@ jobs:
         go-version: ^1.17
 
     - name: Run cassandra integration tests
-      run: bash scripts/cassandra-integration-test.sh
+      run: bash scripts/cassandra-integration-test.sh ${{ matrix.version.schema }}
+    services:
+      cassandra1:
+        image: ${{ matrix.version.image }}:${{ matrix.version.tag }}
+        ports:
+          - 9042:9042
+          - 9160:9160
+      cassandra2:
+        image: ${{ matrix.version.image }}:${{ matrix.version.tag }}
+        ports:
+          - 9043:9042
+          - 9161:9160
+      cassandra3:
+        image: ${{ matrix.version.image }}:${{ matrix.version.tag }}
+        ports:
+          - 9044:9042
+          - 9162:9160
--- a/.github/workflows/ci-cassandra.yml
+++ b/.github/workflows/ci-cassandra.yml
@@ -12,13 +12,15 @@ jobs:
     strategy:
       matrix:
         version:
-        - image: cassandra
-          tag: 3.11
+        - distribution: cassandra
+          major: 3.x
+          image: 3.11
           schema: v001 v002 v003
-        - image: cassandra
-          tag: 4.0
+        - distribution: cassandra
+          major: 4.x
+          image: 4.0
           schema: v004 v004 v004
-    name: ${{ matrix.version.image }} ${{ matrix.version.tag }}
+    name: ${{ matrix.version.distribution }} ${{ matrix.version.major }}
     steps:
     - uses: actions/checkout@v2.3.4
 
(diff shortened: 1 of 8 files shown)


.github/workflows/ci-cassandra.yml
```diff
@@ -30,17 +32,17 @@ jobs:
       run: bash scripts/cassandra-integration-test.sh ${{ matrix.version.schema }}
```
yurishkuro, line 32: `${{ matrix.version.schema }}` is not used in the script as far as I can tell. Instead it does this explicitly:
```
docker run --network integration_test -e CQLSH_HOST=cassandra -e TEMPLATE=/cassandra-schema/v001.cql.tmpl jaeger-cassandra-schema-integration-test
docker run --network integration_test -e CQLSH_HOST=cassandra2 -e TEMPLATE=/cassandra-schema/v002.cql.tmpl jaeger-cassandra-schema-integration-test
```

.github/workflows/ci-cassandra.yml
```diff
@@ -9,6 +9,18 @@ on:
...
+      matrix:
+        version:
+        - distribution: cassandra
+          major: 3.x
+          image: 3.11
+          schema: v001 v002 v003
+        - distribution: cassandra
+          major: 4.x
+          image: 4.0
+          schema: v004 v004 v004
```
yurishkuro, line 22: the repetition doesn't make sense

.github/workflows/ci-cassandra.yml
Ashmita152: I noticed that v003.cql.tmpl was never being tested as part of that script. Hence I am passing those 3 templates as args and referring them as $1, $2, $3 in the script.

.github/workflows/ci-cassandra.yml
Ashmita152: Since we v001, v002, v003 for cassandra 3.11, I had to repeat it. We anyhow need 2 containers atleast since https://github.com/jaegertracing/jaeger/blob/master/plugin/storage/integration/cassandra_test.go#L69-L71 is connecting to cassandra1 and https://github.com/jaegertracing/jaeger/blob/master/plugin/storage/integration/cassandra_test.go#L88-L91 is connecting to cassandra2  so thought to keep all 3.

scripts/cassandra-integration-test.sh
```diff
@@ -2,29 +2,13 @@
...
-docker run -d --name cassandra2 --network integration_test -p 9043:9042 -p 9161:9160 cassandra:3.9
-
 # Build the schema container and run it rather than using the existing container in Docker Hub since that
 # requires this current build to succeed before this test can use it; chicken and egg problem.
 docker build -t jaeger-cassandra-schema-integration-test plugin/storage/cassandra/
-docker run --network integration_test -e CQLSH_HOST=cassandra -e TEMPLATE=/cassandra-schema/v001.cql.tmpl jaeger-cassandra-schema-integration-test
-docker run --network integration_test -e CQLSH_HOST=cassandra2 -e TEMPLATE=/cassandra-schema/v002.cql.tmpl jaeger-cassandra-schema-integration-test
+docker run -e CQLSH_HOST=localhost -e CQLSH_PORT=9042 -e TEMPLATE=/cassandra-schema/$1.cql.tmpl --network=host jaeger-cassandra-schema-integration-test
+docker run -e CQLSH_HOST=localhost -e CQLSH_PORT=9043 -e TEMPLATE=/cassandra-schema/$2.cql.tmpl --network=host jaeger-cassandra-schema-integration-test
+docker run -e CQLSH_HOST=localhost -e CQLSH_PORT=9044 -e TEMPLATE=/cassandra-schema/$3.cql.tmpl --network=host jaeger-cassandra-schema-integration-test
```
yurishkuro, line 10: where would this container at 9044 be used?  The integration test only seems to use the first two

https://github.com/jaegertracing/jaeger/blob/ea1ac8f85342f6efebce42c0424fa7681fbd9a27/plugin/storage/integration/cassandra_test.go#L90

Ashmita152: This is added to validate that v003.cql.tmpl applies properly. I thought it was missing hence thought to add.

yurishkuro: I think we should revisit how we test, are we going to test old schema versions forever?

Ashmita152: In my opinion, unless we deprecate any schema, we should test it.

yurishkuro: Well, the old schemas *are* deprecated. It's been almost a year since v3 was implemented, we don't need to support people running on old schemas forever.

Ashmita152: Do you think we should just test for v003.cql.tmpl and v004.cql.tmpl in that case.

yurishkuro: At this point, yes.

scripts/cassandra-integration-test.sh
```diff
@@ -2,29 +2,11 @@
 
 set -ex
 
-# Clean up before starting.
```
yurishkuro, line 5: what is the motivation for removing setup & cleanup? They are especially useful when running tests locally

Ashmita152: I thought to use GA's [service containers](https://docs.github.com/en/actions/guides/about-service-containers) for managing the containers (like we do in kafka integration test) so we can do parallel tests and don't need to run containers on two ports and duplicate things in integration test (as was done before). I feel this is lot cleaner approach but think otherwise, I can revert back to earlier way.

yurishkuro (dismissed):

plugin/storage/cassandra/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM cassandra:3.11
+FROM cassandra:4.0
```
Ashmita152, line 1: This is done because cassandra v4.0 image contains cqlsh 6.0.0 which is backward compatible with cassandra v3.x (https://pypi.org/project/cqlsh/)

plugin/storage/cassandra/schema/create.sh
```diff
@@ -21,9 +21,30 @@ function usage {
...
+# Extract cassandra version
+#
+# $ cqlsh -e "show version"
+# [cqlsh 5.0.1 | Cassandra 3.11.11 | CQL spec 3.4.4 | Native protocol v4]
+#
+cas_version=$(cqlsh -e "show version" \
+    | awk -F "|" '{print $2}' \
+    | awk -F " " '{print $2}' \
+    | awk -F "." '{print $1}' \
+)
```
Ashmita152, line 33: Tested with cassandra v3.11 and v4.0 containers:

```
root@fa16f4b2b767:/# cqlsh -e "show version"
[cqlsh 6.0.0 | Cassandra 4.0.0 | CQL spec 3.4.5 | Native protocol v5]
root@fa16f4b2b767:/# cas_version=$(cqlsh -e "show version" \
    | awk -F "|" '{print $2}' \
    | awk -F " " '{print $2}' \
    | awk -F "." '{print $1}' \
)
root@fa16f4b2b767:/# echo $cas_version
4
```

```
root@c120e02677e6:/# cqlsh -e "show version"
[cqlsh 5.0.1 | Cassandra 3.11.11 | CQL spec 3.4.4 | Native protocol v4]
root@c120e02677e6:/# cas_version=$(cqlsh -e "show version" \
>     | awk -F "|" '{print $2}' \
>     | awk -F " " '{print $2}' \
>     | awk -F "." '{print $1}' \
> )
root@c120e02677e6:/# echo $cas_version
3
```

yurishkuro (approved):

Issue #2996, opened 2021-05-15
SIGSEGV on empty grpc findTraces query


FauxFaux:
When processing a request for `findTraces` with no parameters, the server experiences a segmentation fault/segfault.

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1232df8]

goroutine 29502 [running]:
github.com/jaegertracing/jaeger/cmd/query/app.(*GRPCHandler).FindTraces(0xc006f19060, 0xc006c303c0, 0x1e2bb20, 0xc006c710c0, 0xc006f19060, 0x20)
	github.com/jaegertracing/jaeger/cmd/query/app/grpc_handler.go:88 +0x98
github.com/jaegertracing/jaeger/proto-gen/api_v2._QueryService_FindTraces_Handler(0x14833a0, 0xc006f19060, 0x1e292a0, 0xc006c76000, 0x26d5680, 0xc000274000)
	github.com/jaegertracing/jaeger/proto-gen/api_v2/query.pb.go:994 +0x10b
google.golang.org/grpc.(*Server).processStreamingRPC(0xc00023c340, 0x1e2f7e0, 0xc0004a8300, 0xc000274000, 0xc006f226f0, 0x2687aa0, 0x0, 0x0, 0x0)
	google.golang.org/grpc@v1.29.1/server.go:1329 +0xcd8
google.golang.org/grpc.(*Server).handleStream(0xc00023c340, 0x1e2f7e0, 0xc0004a8300, 0xc000274000, 0x0)
	google.golang.org/grpc@v1.29.1/server.go:1409 +0xc5c
google.golang.org/grpc.(*Server).serveStreams.func1.1(0xc0004c5630, 0xc00023c340, 0x1e2f7e0, 0xc0004a8300, 0xc000274000)
	google.golang.org/grpc@v1.29.1/server.go:746 +0xa5
created by google.golang.org/grpc.(*Server).serveStreams.func1
	google.golang.org/grpc@v1.29.1/server.go:744 +0xa5
```

**To Reproduce**

Using the grpc-js generated client from `ts-proto`, `QueryServiceClient`: `qsc.findTraces({})`, i.e. make a request with none of the required parameters populated.

**Expected behavior**
An error to be returned, and the server to not be down / restart.

**Version:**
 - OS: Ubuntu
 - Jaeger version: jaegertracing/all-in-one:1.22
[...]

PR #3232, opened 2021-08-25
Fix sigsegv 2996

akuzni2:
- 
#2996 

Added validation checks to the gRPC server handler methods that implement the QueryServiceClient interface. 

Couple of notes:
- I created a gRPC Go client to attempt to test when the request parameters are `nil` - however the client looks like it panics before it reaches the server.  Is there a way I can write a unit test for this functionality? 
- `GetServices` service method: the request parameter `r *api_v2.GetServicesRequest` not used so I didn't add any validation for it.

Questions for the following service methods

`GetOperations`
    - Not sure about behavior for scenarios of GetOperations when Service or Span are "". Possible that these are valid arguments?

### DIFF
--- a/cmd/query/app/grpc_handler_test.go
+++ b/cmd/query/app/grpc_handler_test.go
@@ -264,6 +264,23 @@ func assertGRPCError(t *testing.T, err error, code codes.Code, msg string) {
 	assert.Equal(t, code, s.Code())
 	assert.Contains(t, s.Message(), msg)
 }
+func TestGetTraceEmptyTraceIDFailure_GRPC(t *testing.T) {
+	withServerAndClient(t, func(server *grpcServer, client *grpcClient) {
+
+		server.spanReader.On("GetTrace", mock.AnythingOfType("*context.valueCtx"), mock.AnythingOfType("model.TraceID")).
+			Return(mockTrace, nil).Once()
+
+		res, err := client.GetTrace(context.Background(), &api_v2.GetTraceRequest{
+			TraceID: model.TraceID{},
+		})
+
+		assert.NoError(t, err)
+
+		spanResChunk, err := res.Recv()
+		assert.Error(t, err)
+		assert.Nil(t, spanResChunk)
+	})
+}
 
 func TestGetTraceDBFailureGRPC(t *testing.T) {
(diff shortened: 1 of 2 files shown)


cmd/query/app/grpc_handler_test.go
```diff
@@ -264,6 +264,23 @@ func assertGRPCError(t *testing.T, err error, code codes.Code, msg string) {
...
+			Return(mockTrace, nil).Once()
+
+		res, err := client.GetTrace(context.Background(), &api_v2.GetTraceRequest{
+			TraceID: model.TraceID{},
+		})
+
+		assert.NoError(t, err)
+
+		spanResChunk, err := res.Recv()
+		assert.Error(t, err)
```
jpkrohling, line 280:
```suggestion
		assert.Equal(t, errUninitializedTraceID, err)
```

cmd/query/app/grpc_handler_test.go
```diff
@@ -332,6 +349,17 @@ func TestArchiveTraceNotFoundGRPC(t *testing.T) {
 	})
 }
 
+func TestArchiveTraceEmptyTraceFailureGRPC(t *testing.T) {
+	withServerAndClient(t, func(server *grpcServer, client *grpcClient) {
+
+		_, err := client.ArchiveTrace(context.Background(), &api_v2.ArchiveTraceRequest{
+			TraceID: model.TraceID{},
+		})
+		assert.Error(t, err)
```
jpkrohling, line 358: Same here, you can assert the concrete error with Equal.

jpkrohling (approved):
Looks good, just two very minor improvements.

jpkrohling:
@akuzni2, do you think you can improve the test coverage of this PR?

cmd/query/app/grpc_handler_test.go
akuzni2: Good call - fixed

cmd/query/app/grpc_handler_test.go
akuzni2: Updated

akuzni2:
> @akuzni2, do you think you can improve the test coverage of this PR?

@jpkrohling - updated tests to meet code cov. Added tests to pass `nil` into the `GRPCHandler`

PR #3236, opened 2021-08-28
Enable gzip compression for collector grpc endpoint.

### LINKED ISSUE #3235: Enable gzip compression in collector GRPC endpoint

I would like to optimize network traffic used by collector. Right now out internal deployment receives about 50mb/s of data. 

We use internal C++ client implementation and grpc transport protocol. After enabling grpc compression in client config I received `grpc: Decompressor is not installed for grpc-encoding "gzip"` error from collector. 

Seems like gzip compressor is not enabled by default.

Enable gzip compressor that comes with grpc-go library.

slon:

Resolves #3235

Import gzip compressor in cmd/collector/app/handler. That way gzip compressor is linked into every binary that exposes collector GRPC endpoint.

### DIFF
--- a/cmd/collector/app/handler/grpc_handler.go
+++ b/cmd/collector/app/handler/grpc_handler.go
@@ -19,6 +19,7 @@ import (
 
 	"go.uber.org/zap"
 	"google.golang.org/grpc/codes"
+	_ "google.golang.org/grpc/encoding/gzip"
 	"google.golang.org/grpc/status"
 
 	"github.com/jaegertracing/jaeger/cmd/collector/app/processor"
--- a/cmd/collector/app/handler/grpc_handler_test.go
+++ b/cmd/collector/app/handler/grpc_handler_test.go
@@ -110,6 +110,23 @@ func TestPostSpans(t *testing.T) {
 	}
 }
 
+func TestGRPCCompressionEnabled(t *testing.T) {
+	processor := &mockSpanProcessor{}
+	server, addr := initializeGRPCTestServer(t, func(s *grpc.Server) {
+		handler := NewGRPCHandler(zap.NewNop(), processor)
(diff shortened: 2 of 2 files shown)


yurishkuro (approved):

cmd/collector/app/handler/grpc_handler.go
```diff
@@ -21,6 +21,9 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 
+	// enable gzip compression support
+	_ "google.golang.org/grpc/encoding/gzip"
```
yurishkuro, line 25: this won't work, run `make fmt`. The comment isn't necessary, ".../encoding/gzip" hints at that already

slon: This comment is required by golint.

```
cmd/collector/app/handler/grpc_handler.go:22:2: a blank import should be only in a main or test package, or have a comment justifying it
```

cmd/collector/app/handler/grpc_handler.go
```diff
@@ -19,6 +19,7 @@ import (
 
 	"go.uber.org/zap"
 	"google.golang.org/grpc/codes"
+	_ "google.golang.org/grpc/encoding/gzip" // enable gzip compression support
```
albertteoh, line 22: I think the linter error is from the [old `golint` implementation](https://github.com/golang/lint) which has been deprecated in favour of `go vet` and `staticcheck`. I tested PR this with `golangci-lint` and does not gripe about this line having a blank import.

If we want to avoid including this comment, we could try waiting for #3237 to be merged (if it's approved) and rebase?

slon: Thank you!

I'll wait for the #3237, and then update this PR to the original version.

yurishkuro (approved):

PR #3237, opened 2021-08-29
Replace old linters with golangci-lint

albertteoh:
- #2503 and relates to #2602

- Replaces our existing lint checks with [golangci-lint](https://golangci-lint.run/).
- Goal is to achieve parity (wrt linters) with the existing checks and minimal code changes. If we wish to include more linters, it's a straightforward config change.


.github/workflows/ci-unit-tests.yml
```diff
@@ -22,9 +22,15 @@ jobs:
     - name: Run unit tests
       run: make test-ci
 
+    - name: golangci-lint
```
albertteoh, line 25: I ended up placing the lint step within the unit-tests, which is similar behaviour to the Makefile. However, I feel it would be nicer as a separate job in order to enable concurrency, but am [getting errors relating to undefined types](https://github.com/jaegertracing/jaeger/runs/3454770737?check_suite_focus=true). Oddly, testing on my fork doesn't have this issue.

I believe it's the same config as is with #2602 and does not appear to have encountered this problem. Maybe @daxmc99 has some ideas about this?

albertteoh: It might be worth noting that (_I think_) those "undefined" errors could be due to not having run `go build` on jaeger, which explains why moving this step as part of the unit tests job works.

Makefile
```diff
@@ -161,39 +155,9 @@ fmt:
...
-		>> $(LINT_LOG) || true
-	@[ ! -s "$(LINT_LOG)" ] || (echo "Detected staticcheck failures:" | cat - $(LINT_LOG) && false)
-
 .PHONY: lint
-lint: lint-staticcheck lint-gosec
-	$(GOVET) ./...
-	$(MAKE) go-lint
-	@echo Running go fmt on ALL_SRC ...
-	@$(GOFMT) -e -s -l $(ALL_SRC) > $(FMT_LOG)
-	./scripts/updateLicenses.sh >> $(FMT_LOG)
```
yurishkuro, line 187: This must remain

Makefile
```diff
@@ -161,39 +155,9 @@ fmt:
...
-	./scripts/import-order-cleanup.sh stdout > $(IMPORT_LOG)
```
yurishkuro, line 188: I'm not sure goimport supports the grouping/ordering we want

Makefile
```diff
@@ -400,16 +364,13 @@ changelog:
...
 	go install github.com/mjibson/esc
-	go install github.com/securego/gosec/cmd/gosec
-	go install honnef.co/go/tools/cmd/staticcheck
 
 .PHONY: install-ci
 install-ci: install-tools
 
 .PHONY: test-ci
-test-ci: build-examples lint cover
+test-ci: build-examples cover
```
yurishkuro, line 373: I would consider killing this target and invoking directly in GA

pavolloffay: I like a minimal CI configuration. I want to be able to easily replicate locally what CI is doing - e.g. run the same make target.

Makefile
pavolloffay: it supports grouping but it might create multiple groups.

Makefile
albertteoh: Nice catch, I've added this back into the `lint` target. Tested by removing the license from `grpc_handler.go`:
```
$ make lint
golangci-lint run
./scripts/updateLicenses.sh > .fmt.log
License check failures, run 'make fmt'
cmd/collector/app/handler/grpc_handler.go
make: *** [lint] Error 1
```

Makefile
albertteoh: Actually, it's a good point, our script will enforce correct ordering of groups:

e.g.
```
import (
	"context"

	"github.com/jaegertracing/jaeger/cmd/collector/app/processor"

	"go.uber.org/zap"
	"google.golang.org/grpc/codes"
	"google.golang.org/grpc/status"
)
```
will give error:
```
cmd/collector/app/handler/grpc_handler.go imports out of order
```

but `goimports` gives no error. It does, however, enforce grouping of stdlib, 3rd party and local, just not ordering. I'll bring back our import checker script.

Makefile
albertteoh: Yup, I agree that it's a better developer experience if the CI scripts can be traced down to the make target then "replayed" locally, which a GA step doesn't afford.

As such, I've removed the GA step to run golangci-lint and kept the original setup where the `test-ci` target will run the `lint` target. Fewer changes, and cleaner.

Makefile
```diff
@@ -400,10 +369,8 @@ changelog:
 .PHONY: install-tools
 install-tools:
 	go install github.com/wadey/gocovmerge
-	go install golang.org/x/lint/golint
 	go install github.com/mjibson/esc
-	go install github.com/securego/gosec/cmd/gosec
-	go install honnef.co/go/tools/cmd/staticcheck
+	go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.42.0
```
albertteoh, line 373: Add this to the tools so CI is able to run golangci-lint, and ensure devs also have the right tools installed to test linting.

.golangci.yml
```diff
@@ -0,0 +1,35 @@
...
+  gosec:
+    # To specify a set of rules to explicitly exclude.
+    # Available rules: https://github.com/securego/gosec#available-rules
+    excludes:
+      - G104
+      - G107
+      - G404
+      - G601
+  gosimple:
+    go: "1.17"
```
albertteoh, line 19: Bumped the go version since we've moved onto 1.17 already.

albertteoh:
Please let me know if there are any issues that still need addressing.

Makefile
```diff
@@ -161,39 +157,12 @@ fmt:
...
 	./scripts/import-order-cleanup.sh stdout > $(IMPORT_LOG)
-	@[ ! -s "$(FMT_LOG)" -a ! -s "$(IMPORT_LOG)" ] || (echo "Go fmt, license check, or import ordering failures, run 'make fmt'" | cat - $(FMT_LOG) && false)
-
-.PHONY: go-lint
-go-lint:
-	@cat /dev/null > $(LINT_LOG)
-	@echo Running go lint...
-	@$(GOLINT) $(ALL_PKGS) | grep -v _nolint.go >> $(LINT_LOG) || true;
-	@[ ! -s "$(LINT_LOG)" ] || (echo "Lint Failures" | cat - $(LINT_LOG) && false)
+	@[ ! -s "$(FMT_LOG)" -a ! -s "$(IMPORT_LOG)" ] || (echo "License check or import ordering failures, run 'make fmt'" | cat - $(FMT_LOG) && false)
```
yurishkuro, line 165: This does not print import_log

yurishkuro (dismissed):

Makefile
albertteoh: Address in [8b3a050](https://github.com/jaegertracing/jaeger/pull/3237/commits/8b3a050ff2123238276adb124fb7717f62e7c0ac)

.golangci.yml
```diff
@@ -0,0 +1,36 @@
...
+
+linters-settings:
+  gosec:
+    # To specify a set of rules to explicitly exclude.
+    # Available rules: https://github.com/securego/gosec#available-rules
+    excludes:
+      - G104
+      - G107
+      - G404
+      - G601
```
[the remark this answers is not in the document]
yurishkuro: let's fix them in another PR

yurishkuro (approved):

albertteoh:
thanks for the review Yuri/Pavol! 🙏🏼

Issue #3179, opened 2021-08-04
Migrate away from curator


pavolloffay:
Library `elasticsearch-python` that is used by curator added a check to verify that ES is elastic.co product https://github.com/elastic/elasticsearch-py/releases/tag/v7.14.0.

So now our index cleaner and rollover scritps fail to run with old ES (5.x and even old -oss versions). Here is failed build https://github.com/jaegertracing/jaeger/pull/3178/checks?check_run_id=3240658056#step:6:835

I propose to migrate the python scripts to golang and as a quick fix we can pin the `elasticserch-python` to an older version without the check.

jpkrohling:
Yes, please!

pavolloffay:
done

PR #3258, opened 2021-09-10
Use es-rollover golang implementation

rubenvp8510:

Updates #3179


Makefile
```diff
@@ -188,6 +184,10 @@ build-esmapping-generator-linux:
 build-es-index-cleaner:
 	$(GOBUILD) -o ./cmd/es-index-cleaner/es-index-cleaner-$(GOOS)-$(GOARCH) ./cmd/es-index-cleaner/main.go
 
+.PHONY: build-es-rollover
+build-es-rollover:
+	$(GOBUILD) -o ./cmd/es-rollover/es-rollover-$(GOOS)-$(GOARCH) ./cmd/es-rollover/main.go
```
yurishkuro, line 189: please add .gitignore in that folder and exclude these new binaries

Makefile
```diff
@@ -109,10 +109,6 @@ grpc-plugin-storage-integration-test:
 	(cd examples/memstore-plugin/ && go build .)
 	STORAGE=grpc-plugin $(MAKE) storage-integration-test
 
-.PHONY: test-compile-es-scripts
-test-compile-es-scripts:
```
yurishkuro, line 113: this is called from 
```
./scripts/es-integration-test.sh:97:  make test-compile-es-scripts
```

yurishkuro:
are the CLI flags in the new image compatible with the old python script?

pavolloffay:
The CI didn't pass

cmd/es-rollover/app/index_options_test.go
```diff
@@ -61,7 +61,7 @@ func TestRolloverIndices(t *testing.T) {
 			expected: []expectedValues{
 				{
 					prefix:               "mytenant-jaeger-span-archive",
-					templateName:         "mytenant-jaeger-span",
```
pavolloffay, line 64: Why is this changing? Shouldn't the template name keep the prefix name? 

The new golang implementation has to be backwards compatible.

pavolloffay (approved):

pavolloffay:
> are the CLI flags in the new image compatible with the old python script?

Yes this is the goal

cmd/es-rollover/app/index_options_test.go
rubenvp8510: I shouldn't 

https://github.com/jaegertracing/jaeger/pull/3258/files#diff-59567de3853ac61d1a7460823020ee5c2fe7680353df87011f4cd1df1b132f45L83

Look as this calls, the template has no prefix.  It was my mistake to include the prefix in the first PR for this.

rubenvp8510: wait. I see one mistake , don't merge this please.

rubenvp8510: Fixed, you're right, the template name should be prefixed, the mapping name used to build the template should not.

pavolloffay (approved):

Issue #3203, opened 2021-08-12
Duplicate tag keys due to collector level tags


pbdger:
**Describe the bug**
By using the --collctor.tags parameter, duplicate key entries can be created by the agent.

**To Reproduce**

*Set tags parameter --collector.tags=collector=setViaAgentCli
*Start the Agent
*Create a tracer in Your code with tag collector=setInTracerCreation

**Expected behavior**
tags are defined as key:value pairs (https://opentracing.io/docs/overview/tags-logs-baggage/#tags).
Therefore, the expectation is that when the trace is displayed at the process, there is only one tag key agent.

**Screenshots**
![image](https://user-images.githubusercontent.com/81152194/129173030-dd1c9124-23de-4043-a1fb-b8bf910bcce2.png)

**Version:**

*OS: MacOs
*Jaeger version: 1.25
*Deployment: Docker

**Additional context**
The fix would also need to define if the tag from the --collector.tags parameter always overwrites a tag from the tracer or if the tags from the tracer is lead.

pavolloffay:
The attributes/tags in jaeger are not map - it is a list so duplicate entires are supported by design.

yurishkuro:
@pavolloffay I think it would be a valid feature to have a CLI flag to control what to do with dups added by collector/agent: keep, drop, replace.

pavolloffay:
I don't have any objections for it.

yurishkuro:
agent is no more, closing

PR #3273, opened 2021-09-18
Add cli flag to handle duplicate tag keys

Ashmita152:

Resolves https://github.com/jaegertracing/jaeger/issues/3203

- As recommended by Albert, this PR adds a cli flag --collector.use-on-tag-key-conflict with three possible values: "span", "collector" and "both". Using "both" as the default behaviour to maintain backward compatibility.


cmd/collector/app/options.go
```diff
@@ -31,24 +31,27 @@ const (
 	DefaultQueueSize = 2000
 	// DefaultGRPCMaxReceiveMessageLength is the default max receivable message size for the gRPC Collector
 	DefaultGRPCMaxReceiveMessageLength = 4 * 1024 * 1024
+	// DefaultHandleDuplicateTagKey is the default behaviour of collector on receiving duplicate tag key
+	DefaultHandleDuplicateTagKey = "keep"
```
albertteoh, line 35: I understand we want to be backwards compatible with original behaviour, though I'd like to understand, is the original behaviour considered correct or a bug?

cmd/collector/app/span_processor.go
```diff
@@ -180,23 +182,55 @@ func (sp *spanProcessor) processItemFromQueue(item *queueItem) {
...
-	}
-	dedupKey := make(map[string]struct{})
-	for _, tag := range span.Process.Tags {
-		if value, ok := sp.collectorTags[tag.Key]; ok && value == tag.AsString() {
-			sp.logger.Debug("ignore collector process tags", zap.String("key", tag.Key), zap.String("value", value))
-			dedupKey[tag.Key] = struct{}{}
+func (sp *spanProcessor) deDupTags(span *model.Span) {
+	deDupKey := make(map[string]int, len(span.Process.Tags))
+	deDupKeyVal := make(map[string]int, len(span.Process.Tags))
+	for idx, tag := range span.Process.Tags {
```
albertteoh, line 188: `idx` isn't used after setting them in the respective maps. I suggest changing the maps to be `map[string]bool`, and set them to `true`.

Then you could do `if !dedupKey[k] {`

albertteoh:
The [build is failing](https://github.com/jaegertracing/jaeger/pull/3273/checks?check_run_id=3638919562#step:9:184), looks like the first error is:
```
+ make create-baseimg-debugimg
docker buildx inspect jaeger-build > /dev/null || docker buildx create --use --name=jaeger-build --buildkitd-flags="--allow-insecure-entitlement security.insecure --allow-insecure-entitlement network.host" --driver-opt="network=host"
jaeger-build
docker inspect registry > /dev/null || docker run --rm -d -p 5000:5000 --name registry registry:2
error: no builder "jaeger-build" found
Error: No such object: registry
Unable to find image 'registry:2' locally
```

cmd/collector/app/span_processor.go
Ashmita152: `idx` is being used in `replace` case for deleting the tag from array.

cmd/collector/app/options.go
Ashmita152: I would let maintainers to decide the default value. In my opinion, we should keep the current behaviour as default.

Ashmita152:
Updated the PR with feedbacks.

cmd/collector/app/span_processor.go
```diff
@@ -180,23 +182,55 @@ func (sp *spanProcessor) processItemFromQueue(item *queueItem) {
...
+			if _, ok := deDupKeyVal[k]; !ok {
+				span.Process.Tags = append(span.Process.Tags, model.String(k, v))
+			}
 		}
+	case "collector":
+		// replace with collector tags if it has the same key in spans
+		for k, v := range sp.collectorTags {
+			if _, ok := deDupKeyVal[k]; !ok {
+				if _, ok = deDupKey[k]; ok {
+					idx := deDupKey[k]
```
albertteoh, line 218: We could flatten and simplify to something like:
```
			if _, ok := deDupKeyVal[k]; ok {
                                 continue
                         }
			if idx, ok := deDupKey[k]; ok {
                                 // you now have idx set
                         }
```

cmd/collector/app/span_processor.go
```diff
@@ -180,23 +182,55 @@ func (sp *spanProcessor) processItemFromQueue(item *queueItem) {
...
+					copy(span.Process.Tags[idx:], span.Process.Tags[idx+1:])
+					span.Process.Tags = append(span.Process.Tags[:len(span.Process.Tags)-1], model.String(k, v))
```
albertteoh, line 220: If I understand correctly, this is an O(n) time operation, together with a loop over the collector tags, _worst case_.

I think it's okay as it is, especially when _I think_ the modus operandi is not to set the collector tags and this case would only be hit if there are conflicts and the tag values are not identical.

Having said that, I was wondering if there is there a way we can simplify this?

I threw something together (which probably isn't perfect) to see if it helps us come up with ideas on how to simplify a bit. In particular, my goals were:
- Reduce the additional memory/complexity needed for maintaining additional maps
  - I still needed one additional map to store a copy of the collector tags.
- _Try_ to dedup the tags in a single pass over the process tags
  - I still needed one the additional loop over the collector tags at the end.
- Improve readability
  - Would like to get your thoughts on this.

Let me know what you think:
```
func (sp *spanProcessor) dedupTags(span *model.Span) {
	// Copy the collector tags to keep track of which of these need to be appended.
	cTags := make(map[string]string, len(sp.collectorTags))
	for k, v := range sp.collectorTags {
		cTags[k] = v
	}
	var procTags []model.KeyValue
	for _, sTag := range span.Process.Tags {
		appendSpanTag := true
		// If there's a tag key conflict.
		if cTagVal, ok := sp.collectorTags[sTag.Key]; ok {
			switch sp.useOnTagKeyConflict {
			case "both":
				// If both key and value are conflicting, there's no need to add the collector's tag.
				if cTagVal == sTag.AsString() {
					delete(cTags, sTag.Key)
				}
			case "collector":
				appendSpanTag = false
			case "span":
				// If we choose the span's tag, then remove this conflicting tag from the collector tags.
				delete(cTags, sTag.Key)
			}
		}
		if appendSpanTag {
			procTags = append(procTags, sTag)
		}
	}
	// Append the remaining collector tags.
	for k, v := range cTags {
		procTags = append(procTags, model.String(k, v))
	}
	span.Process.Tags = procTags
}
```

cmd/collector/app/span_processor_test.go
```diff
@@ -383,6 +383,142 @@ func TestSpanProcessorWithCollectorTags(t *testing.T) {
...
 
+func TestSpanProcessorWithCollectorTagsWithUseOnTagKeyConflictAsBoth(t *testing.T) {
+	testCollectorTags := map[string]string{
+		"extra": "tag",
+		"env":   "prod",
+		"node":  "172.22.18.161",
+	}
+
+	w := &fakeSpanWriter{}
+	p := NewSpanProcessor(w, nil, Options.CollectorTags(testCollectorTags), Options.UseOnTagKeyConflict("both")).(*spanProcessor)
```
albertteoh, line 394: Given `"both"` is the default, I suggest adding (maybe with a table driven test), a test where the `UseOnTagKeyConflict` option isn't provided, which we should expect to have the same result as `"both"` being explicitly set.

cmd/collector/app/span_processor_test.go
```diff
@@ -383,6 +383,142 @@ func TestSpanProcessorWithCollectorTags(t *testing.T) {
...
+		"env":   "prod",
+		"node":  "172.22.18.161",
+	}
+
+	w := &fakeSpanWriter{}
+	p := NewSpanProcessor(w, nil, Options.CollectorTags(testCollectorTags), Options.UseOnTagKeyConflict("span")).(*spanProcessor)
+
+	defer assert.NoError(t, p.Close())
+	span := &model.Span{
+		Process: model.NewProcess("unit-test-service", []model.KeyValue{
```
albertteoh, line 446: I think it would be nice to add an additional tag for each test's span that has a tag that doesn't exist in the `testCollectorTags`, which will catch cases in our logic that somehow assume overlapping keys or keys that only exist in the collector tags, i.e.:
```
{
	Key:  "only-in",
	VStr: "this-span",
},
```

cmd/collector/app/span_processor.go
Ashmita152: I am not good in data-structures but this is definitely lot better and simpler to understand. I decided to use it.

cmd/collector/app/span_processor_test.go
Ashmita152: There is a test in the master for that: https://github.com/jaegertracing/jaeger/blob/master/cmd/collector/app/span_processor_test.go#L338

albertteoh (approved):
Looks pretty good to me, just a couple minor suggestions.

Please also rebase with master.

cmd/collector/app/builder_flags_test.go
```diff
@@ -58,7 +59,18 @@ func TestCollectorOptionsWithFlags_CheckMaxReceiveMessageLength(t *testing.T) {
...
+func TestCollectorOptionsWithFlags_UseOnTagKeyConflictAsInvalid(t *testing.T) {
+	c := &CollectorOptions{}
+	v, command := config.Viperize(AddFlags)
+	command.ParseFlags([]string{
+		"--collector.use-on-tag-key-conflict=invalid",
+	})
+	c.InitFromViper(v, zap.NewNop())
+
+	assert.Equal(t, "both", c.CollectorUseOnTagKeyConflict)
+}
```
albertteoh, line 76: The test is failing due to my `logger.Fatal` suggestion. This should resolve the failing test.
```suggestion
func TestCollectorOptionsWithFlags_UseOnTagKeyConflictAsInvalid(t *testing.T) {
	defer func() {
		if r := recover(); r != nil {
			assert.Contains(t, r, "Invalid configuration. Allowed values: both, collector, span")
		}
	}()
	c := &CollectorOptions{}
	v, command := config.Viperize(AddFlags)
	command.ParseFlags([]string{
		"--collector.use-on-tag-key-conflict=invalid",
	})
	c.InitFromViper(v, zap.NewNop().WithOptions(zap.OnFatal(zapcore.WriteThenPanic)))
}
```

Ashmita152:
Thank you so much @albertteoh I didn't know we can unit test for log.Fatal too. I agree without we should just fail on invalid parameter.

cmd/collector/app/span_processor.go
```diff
@@ -180,23 +182,46 @@ func (sp *spanProcessor) processItemFromQueue(item *queueItem) {
 	sp.metrics.InQueueLatency.Record(time.Since(item.queuedTime))
 }
 
-func (sp *spanProcessor) addCollectorTags(span *model.Span) {
-	if len(sp.collectorTags) == 0 {
-		return
+func (sp *spanProcessor) dedupTags(span *model.Span) {
```
yurishkuro, line 185: just brainstorming how to make this func do as few allocations as possible:
```
init:
   cTagsBits := map[string]uint64 {
      "tag1": 0x1,
      "tag2": 0x2,
      "tag3": 0x4, 
      etc
   }
   
dedupTags(spanTags)
   var skipmask uint64
   for i, tag in spanTags { // efficiency tip: don't actually use `tag` var as it requires copying, use spanTags[i]
      if tag.key in cTagsBits {
         // duplicate found
         if dedupeMode == keepBoth {
             continue
         }
         // replace it in-place based on dedupe-mode requested
         spanTags[i] = ...
         // and label it to be skipped
         skipmask = skipmask | cTagsBits[tag.key]
      }
   }
   for _, key in cTagsBits {
        if mask & cTagsBits[key] == 0 {
            spanTags = append(spanTags, ...) // append to the same slice
        }
   }
```
this limits the number of collector tags to 64, which I think is more than reasonable

cmd/collector/app/span_processor.go
```diff
@@ -180,23 +182,46 @@ func (sp *spanProcessor) processItemFromQueue(item *queueItem) {
...
+		procTags = append(procTags, model.String(k, v))
+	}
+	span.Process.Tags = procTags
+}
+
+func (sp *spanProcessor) addCollectorTags(span *model.Span) {
+	if len(sp.collectorTags) == 0 {
+		return
+	}
+	sp.dedupTags(span)
```
yurishkuro, line 224: this behavior should be identical to the agent's tags, so we should have a util class somewhere that stores the tags and adds them to span

cmd/collector/app/builder_flags.go
```diff
@@ -85,14 +89,15 @@ func AddFlags(flags *flag.FlagSet) {
 	flags.String(collectorZipkinAllowedHeaders, "content-type", "Comma separated list of allowed headers for the Zipkin collector service, default content-type")
 	flags.String(collectorZipkinAllowedOrigins, "*", "Comma separated list of allowed origins for the Zipkin collector service, default accepts all")
 	flags.String(collectorZipkinHTTPHostPort, "", "The host:port (e.g. 127.0.0.1:9411 or :9411) of the collector's Zipkin server (disabled by default)")
+	flags.String(collectorUseOnTagKeyConflict, "both", "Specifies how the collector will handle duplicate tag keys. (values: span, collector, both)")
```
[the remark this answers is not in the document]
albertteoh: Yeah, I agree.

cmd/collector/app/span_processor.go
albertteoh: This is an elegant solution, assuming a 64 tag limit is acceptable.

PR #3286, opened 2021-09-27
Fixing otel configuration in docker compose

Ashmita152:

- Fixing otel configuration in docker-compose

- As per this https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/5428 they have modified the tls configuration key.


albertteoh (approved):
thanks!

albertteoh:
> What version of the collector are we using?

In what context? Did you mean which otel collector (or jaeger collector) version is used in the docker-compose? If so, it's the "latest" image tags.

PR #3299, opened 2021-09-30
Close #3270: Prevent rollover lookback from passing the Unix epoch

### LINKED ISSUE #3270: Existing traces can't be opened from the trace id lookup UI (404 error received) - ES backend
**Describe the bug**
Trace id's cannot be opened through the trace lookup UI (the search bar at the top) the API 404 page is shown.
However I can find open the same trace from search UI. 

**To Reproduce**
1. set up Jaeger with Elasticsearch and indexes rollover (either via utilities of ILM)  and put some traces
3. search for a trace and open it
4. hit refresh (F5)
5. 404 Not Found displayed
6. same 404 behavior in case of opening the trace via lookup 

**Expected behavior**
Existing trace should be opened 

**Screenshots**
image
image
image

 - OS: Linux
 - Jaeger version: 1.26
 - Deployment: Kubernetes 

Verified the indexes and policies setup, Jaeger configuration, verified the presence of the trace in question

**Additional context**
Span and service indexes created with name jaeger-span-000001 and jaeger-service-000001 with aliases jaeger-span-read, jaeger-span-write and  jaeger-service-read, jaeger-service-write. The issue is reproducible right away after fresh setup.
Both collector and UI are started with arguments:
``` 
"--es.use-aliases=true",
"--es.num-replicas=0",
"--es.num-shards=1",
"--es.create-index-templates=false",
"--es.log-level=error",
"--es.use-ilm=true",
"--es.max-span-age=168h0m0s",
```
I've take a look into the code, and I think the issue is in `func (s *SpanReader) multiRead(ctx context.Context, traceIDs []model.TraceID, startTime, endTime time.Time) ([]*model.Trace, error)` which is called from api/trace/{id}. In case of using aliases it should make a query to the read alias `jaeger-span-read` in the case, and not trying to construct the multi index search call which will fail due to different index naming not the expected date formatted but - 000001, 000002, etc. However I'm not a golang developer and I may miss something.

In case if any other information will be required I'll try to provide it asap.

ctreatma:

Resolves #3270

Version 1.26 introduced an automatic configuration for the query lookback
when using ElasticSearch with aliases enabled.  When aliases are enabled,
the ES plugin will look back 100 years.  This pre-dates the Unix epoch, and
while such dates can be modeled as negative timestamps, the model defined
in `jaeger/model/time.go` only supports unsigned timestamps.  As a result,
the 100-year lookback ends up overflowing the time model, resulting in a
distant-future lookback date, rather than a distant-past lookback date.

While the time model could be updated to support negative timestamps, it
seems unlikely that any Jaeger users would reasonably need to search for
spans from the 1920s.  This reduces the automatic lookback to 50 years to
remove the overflow issue while still providing an extremely long search
window that should serve even the most ambitious searches of historical
trace data.

### DIFF
--- a/plugin/storage/es/spanstore/reader.go
+++ b/plugin/storage/es/spanstore/reader.go
@@ -61,7 +61,7 @@ const (
 
 	defaultNumTraces = 100
 
-	rolloverMaxSpanAge = time.Hour * 24 * 365 * 100
+	rolloverMaxSpanAge = time.Hour * 24 * 365 * 50
 )
 
 var (
--- a/plugin/storage/es/spanstore/reader_test.go
+++ b/plugin/storage/es/spanstore/reader_test.go
@@ -149,7 +149,7 @@ func TestNewSpanReader(t *testing.T) {
 				MaxSpanAge:          time.Hour * 72,
 				UseReadWriteAliases: true,
 			},
-			maxSpanAge: time.Hour * 24 * 365 * 100,
+			maxSpanAge: time.Hour * 24 * 365 * 50,
 		},
 	}
 	for _, test := range tests {


pavolloffay:
@ctreatma great fix :) 

Could you please sign the commits (DCO check)? https://github.com/jaegertracing/jaeger/blob/master/CONTRIBUTING_GUIDELINES.md#certificate-of-origin---sign-your-work

pavolloffay (approved):

albertteoh (approved):

pavolloffay (approved):

oliversalzburg:
What's the workaround until this is released? Downgrade to 1.25?

albertteoh:
> What's the workaround until this is released? Downgrade to 1.25?

@oliversalzburg in case you missed the notification, this fix is now available in release [1.27](https://github.com/jaegertracing/jaeger/releases/tag/v1.27.0).

PR #3302, opened 2021-10-04
Changed sampling type env var and updated collector help text

joe-elliott:
As suggested here: https://github.com/jaegertracing/documentation/pull/526

This PR renamed `SAMPLING_TYPE` env var to `SAMPLING_CONFIG_TYPE`. Additionally it extends the collector help text to include information about this new env var.

New text:
```
$ go run ./cmd/collector env
2021/10/04 10:00:45 maxprocs: Leaving GOMAXPROCS=8: CPU quota undefined

their names to upper case and replacing punctuation with underscores. For example:

SPAN_STORAGE_TYPE string         The type of backend [cassandra, opensearch, elasticsearch, memory, kafka, badger, grpc-plugin] used for trace storage. Multiple backends can be specified as comma-separated list, e.g. "cassandra,elasticsearch" (currently only for writing spans). Note that "kafka" is only valid in jaeger-collector; it is not a replacement for a proper storage backend, and only used as a buffer for spans when Jaeger is deployed in the collector+ingester configuration.  (default "cassandra")

DEPENDENCY_STORAGE_TYPE string   The type of backend used for service dependencies storage. (default "${SPAN_STORAGE_TYPE}")

SAMPLING_CONFIG_TYPE string      The method [file, adaptive] used for determining the sampling rates served to clients configured with remote sampling enabled. "file" uses a periodically reloaded file and "adaptive" dynamically adjusts sampling rates based on current traffic.  (default "file")
```

### DIFF
--- a/plugin/sampling/strategystore/factory.go
+++ b/plugin/sampling/strategystore/factory.go
@@ -34,10 +34,10 @@ type Kind string
 
 const (
 	samplingTypeAdaptive = "adaptive"
-	samplingTypeStatic   = "file"
+	samplingTypeFile     = "file"
 )
 
-var AllSamplingTypes = []string{samplingTypeStatic, samplingTypeAdaptive}
+var AllSamplingTypes = []string{samplingTypeFile, samplingTypeAdaptive}
 
 // Factory implements strategystore.Factory interface as a meta-factory for strategy storage components.
 type Factory struct {
@@ -65,7 +65,7 @@ func NewFactory(config FactoryConfig) (*Factory, error) {
 
 func (f *Factory) getFactoryOfType(factoryType Kind) (strategystore.Factory, error) {
 	switch factoryType {
-	case samplingTypeStatic:
+	case samplingTypeFile:
 		return static.NewFactory(), nil
 	case samplingTypeAdaptive:
 		return adaptive.NewFactory(), nil
--- a/plugin/sampling/strategystore/factory.go
+++ b/plugin/sampling/strategystore/factory.go
@@ -34,10 +34,10 @@ type Kind string
 
 const (
 	samplingTypeAdaptive = "adaptive"
-	samplingTypeStatic   = "static"
+	samplingTypeStatic   = "file"
 )
 
-var allSamplingTypes = []Kind{samplingTypeStatic, samplingTypeAdaptive}
+var AllSamplingTypes = []string{samplingTypeStatic, samplingTypeAdaptive}
 
 // Factory implements strategystore.Factory interface as a meta-factory for strategy storage components.
 type Factory struct {
@@ -70,7 +70,7 @@ func (f *Factory) getFactoryOfType(factoryType Kind) (strategystore.Factory, err
 	case samplingTypeAdaptive:
 		return adaptive.NewFactory(), nil
 	default:
-		return nil, fmt.Errorf("unknown sampling strategy store type %s. Valid types are %v", factoryType, allSamplingTypes)
+		return nil, fmt.Errorf("unknown sampling strategy store type %s. Valid types are %v", factoryType, AllSamplingTypes)
 	}
 }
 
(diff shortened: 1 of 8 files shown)


plugin/sampling/strategystore/factory.go
```diff
@@ -34,10 +34,10 @@ type Kind string
 
 const (
 	samplingTypeAdaptive = "adaptive"
-	samplingTypeStatic   = "static"
+	samplingTypeStatic   = "file"
```
yurishkuro, line 37: Based on https://github.com/jaegertracing/jaeger/pull/2966/files#diff-3ee6a5bb916b37c246c98086c1195bb67cbbc19e0166c8c136472b918267c884L31, the value `static` existed before, so we need to support it per backwards compatibility policy (can be removed after 3 months). So I would suggest naming this const samplingTypeFile, and then handling the value "static" in the switch statement in the factory as legacy (ideally with a logged deprecation warning).

joe-elliott:
Updated to accept the `SAMPLING_TYPE` env var as a deprecated method of setting the sampling type and `static` as a deprecated value.

Preferred is `SAMPLING_CONFIG_TYPE` and `file`.

```
$ SAMPLING_TYPE=static go run ./cmd/collector
2021/10/04 15:58:17 maxprocs: Leaving GOMAXPROCS=8: CPU quota undefined
WARNING: Using deprecated 'SAMPLING_TYPE' env var. Please switch to 'SAMPLING_CONFIG_TYPE'.
WARNING: Using deprecated 'static' value for SAMPLING_CONFIG_TYPE. Please switch to 'file'.
{"level":"info","ts":1633377497.5333173,"caller":"flags/service.go:117","msg":"Mounting metrics handler on admin server","route":"/metrics"}
...
```

plugin/sampling/strategystore/factory_config.go
```diff
@@ -16,32 +16,60 @@ package strategystore
...
 //   * `adaptive` - built-in
-func FactoryConfigFromEnv() (*FactoryConfig, error) {
-	strategyStoreType := os.Getenv(SamplingTypeEnvVar)
-	if strategyStoreType == "" {
-		strategyStoreType = samplingTypeStatic
+func FactoryConfigFromEnv(log io.Writer) (*FactoryConfig, error) {
+	strategyStoreType := getStrategyStoreTypeFromEnv(log)
+	if strategyStoreType != samplingTypeAdaptive &&
+		strategyStoreType != samplingTypeFile &&
+		strategyStoreType != deprecatedSamplingTypeStatic {
```
yurishkuro, line 45: it might be even cleaner if you check for and override `static` in `getStrategyStoreTypeFromEnv`, because there was no release where `SAMPLING_CONFIG_TYPE=static` was valid.

plugin/sampling/strategystore/factory_config.go
```diff
@@ -16,32 +16,60 @@ package strategystore
...
+	strategyStoreType := os.Getenv(SamplingTypeEnvVar)
+	if strategyStoreType != "" {
+		return strategyStoreType
+	}
+
+	// accept the old env var but warn
+	strategyStoreType = os.Getenv(deprecatedSamplingTypeEnvVar)
+	if strategyStoreType != "" {
+		fmt.Fprintf(log, "WARNING: Using deprecated '%s' env var. Please switch to '%s'.\n", deprecatedSamplingTypeEnvVar, SamplingTypeEnvVar)
+		return strategyStoreType
```
yurishkuro, line 70:
```suggestion
	        if strategyStoreType == deprecatedSamplingTypeStatic {
		        fmt.Fprintf(log, "WARNING: Using deprecated '%s' value for %s. Please switch to '%s'.\n", strategyStoreType, SamplingTypeEnvVar, samplingTypeFile)
		        strategyStoreType = samplingTypeFile
	        }
		return strategyStoreType
```

yurishkuro (approved):

yurishkuro (approved):

PR #3319, opened 2021-10-11
cmd/es-rollover/app/init: fix args order in strings.Contains

quasilyte:
If we do `strings.Contains("resource_already_exists_exception", errorMap["type"].(string))`,
we're basically doing `"resource_already_exists_exception" == errorMap["type"].(string)`.
I would assume that we intended to test whether `errorMap["type"]` contains
`"resource_already_exists_exception"` as a substring, so the reverse `strings.Contains` order
is more applicable here.

- 

-


pavolloffay:
It should pass DCO in https://github.com/jaegertracing/jaeger/pull/3324

Issue #1976, opened 2019-12-13
Move Helm chart to new jaegertracing/helm-charts


scottrigby:
## Requirement - what kind of business use case are you trying to solve?

Keeping Jaeger Helm charts accessible for end users, and also improving maintainability of the chart(s) moving forward.

## Problem - what in Jaeger blocks you from solving the requirement?

The `stable` and `incubator` Helm chart repositories, currently listed on the [Helm Hub](https://hub.helm.sh/), have begun a 1 year deprecation timeline. After 6 months they will be de-listed from the Hub. See helm/charts > [Status of the Project](https://github.com/helm/charts#status-of-the-project) and [Deprecation Timeline](https://github.com/helm/charts#deprecation-timeline).

The impact for end users is in 6 months the `incubator/jaeger` and `stable/jaeger-operator` will no longer be listed on the Helm Hub, as they are now: https://hub.helm.sh/charts?q=jaeger. Additionally, since [Helm 3 was released](https://helm.sh/blog/helm-3-released/), new versions of helm do not even include the `stable` repo by default. This means after 6 months, if the Jaeger charts are not moved to a new home and listed on the Hub, new users will not have a direct way to discover or install them.

## Proposal - what do you suggest to solve the problem or improve the existing situation?

Move Jaeger Helm charts to prepare for `stable` and `incubator` charts repo deprecation timeline.

1. Create a jaegertracing Helm charts gitpub project, such as https://github.com/jaegertracing/helm-charts, to be used as a [Helm repository](https://helm.sh/docs/topics/chart_repository/) for Jaeger-related charts.
2. Move `incubator/jaeger` and `stable/jaeger-operator` from https://github.com/helm/charts to a new GitHub project. Note that retaining git history from the beginning is much easier than trying to splice it in later (see https://github.com/goharbor/harbor-helm/issues/48)
3. Include disclaimer README similar to https://github.com/elastic/helm-charts#elastic-stack-kubernetes-helm-chart. Also note these charts are community maintained, and add a basic statement about how to get involved (because it will be your own GitHub repo, you no longer need to rely on the helm/charts `OWNERS` file, but can add collaborators directly to the repo, if per-chart access is desired use GitHub native [CODEOWNERS](https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners), etc). This all helps keep the burden low for the Jaeger team, and and shows support for Jaeger Helm Charts without adding additional liability
[...]

PR #3327, opened 2021-10-17
Update Jaeger chart link

isbang:
- #1976

- I think it is more reasonable to link to Jaeger chart repository not k8s chart repository


isbang:
Ok. I will re-create this PR. Thank you!

Issue #3305, opened 2021-10-05
Storage backends for adaptive sampling


yurishkuro:
Since v1.27 adaptive sampling is supported in the backend, but it only works with Cassandra as the backing store. We need to implement it for other types of stores, e.g.
  * memory-only (for all-in-one)
  * Badger
  * Elasticsearch / OpenSearch

lonewolf3739:
I wanted to try out this feature but realised not supported for different backends. I can take a stab at this if nobody is already working on it.

meneketehe:
Hi, does anyone working on this? I would like to work on Elasticsearch storage support.

meneketehe:
I have some questions before I start implementing the feature.
1. What is the purpose of the `bucket` column in the `operation_throughput` and `sampling_probabilities` tables in the Cassandra storage backend? Is it solely for performance, or are there other considerations I'm missing?
2. Do I need to use index-per-day pattern? Do I need to support rollover and index-cleaner for adaptive sampling?

Here is my idea to store the document, feedbacks are welcome!

jaeger-throughputs
Is it better if we encode the `service`, `operation`, `count`, and `probabilities` field into a single string? Since we only query the `timestamp` field
```
// mapping
{
  "mappings": {
    "properties": {
      "timestamp": {
        "type": "long"
      },
      "service": {
        "type": "keyword",
        "index": false
      },
      "operation": {
        "type": "keyword",
        "index": false
      },
      "count": {
        "type": "long",
        "index": false
      },
      "probabilities": {
        "type": "keyword",
        "index": false
      }
    }
  }
}
```
```
// example
{
  "timestamp": 1485467191639875,
  "service": "svc",
  "operation": "op",
  "count": 40,
  "probabilities": ["0.1", "0.5"]
}
```


jaeger-probabilities-and-qps
```
// mapping
{
  "mappings": {
    "properties": {
      "timestamp": {
        "type": "long"
      },
      "hostname": {
        "type": "keyword",
        "index": false
      },
      "probabilities": {
        "type": "object",
        "dynamic": false,
        "properties": {
          "operations": {
            "type": "object",
            "dynamic": false,
            "properties": {
              "operation": {
                "type": "keyword",
                "index": false
              },
              "probability": {
                "type": "keyword",
                "index": false
              },
              "qps": {
                "type": "long",
                "index": false
              }
            }
          },
          "service": {
            "type": "keyword",
            "index": false
          }
        }
      }
    }
  }
}
```
```
// example
[...]

PR #3335, opened 2021-10-21
Add in-memory storage support for adaptive sampling

srikanthccv:
Signed-off-by: Srikanth Chekuri <srikanth.chekuri92@gmail.com>

- Part of #3305


lonewolf3739:
Thanks @yurishkuro for the review. I took the default unbounded number of traces as an inspiration but given the processor implementation it makes sense to limit the number of entries to N. Regarding probabilities, I think we can go with the simplest approach first.

PR #3341, opened 2021-10-27
Add auth token propagation for metrics reader

### LINKED ISSUE #3340: Authorization header does not propagate to `/api/metrics/*` endpoints
**Describe the bug**
Jaeger query doesn't propagate auth token on the `/api/metrics/*` endpoints.

**To Reproduce**
Steps to reproduce the behaviour:
1. Start jaeger-query with QUERY_BEARER_TOKEN_PROPAGATION=true
2. make a GET request to `/api/metrics/latencies?service=jaeger-query&endTs=1635219714184&...` with an authorization token
3. inspect jaeger-query request headers

**Expected behaviour**
The `/api/metrics/*` endpoints should propagate the auth token the same way it does on other endpoints (/operations, /traces, etc.)

- OS: [macOS]
- Jaeger version: [1.27]
- Deployment: [Docker]

albertteoh:

- Resolves #3340 

- Adds propagation of authorization token from the inbound request to the outbound request to a metrics server.
- This is achieve by wrapping the `http.Transport` into a `tokenAuthTransport` struct that implements the `http.RoundTripper`'s `RoundTrip` function, which injects the Authorization bearer token header into the outbound request.
- Follows the existing pattern used by the ES reader: https://github.com/jaegertracing/jaeger/blob/master/pkg/es/config/config.go#L532


pkg/bearertoken/transport.go
```diff
@@ -0,0 +1,77 @@
...
+		return nil, errors.New("no http.RoundTripper provided")
+	}
+	token := tr.defaultToken
+	if tr.allowOverrideFromCtx {
+		headerToken, _ := GetBearerToken(r.Context())
+		if headerToken != "" {
+			token = headerToken
+		}
+	}
+	r.Header.Set("Authorization", "Bearer "+token)
```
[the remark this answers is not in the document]
albertteoh: Yeah, I considered it but was _technically_ a breaking change; I don't know enough about Auth headers to determine if this is a bug or correct behaviour, though sounds like it's a bug given your suggestion.

pkg/bearertoken/context.go
```diff
@@ -12,30 +12,32 @@
...
 
-type contextKey string
+type contextKeyType string
 
-// BearerTokenKey is the string literal used internally in the implementation of this context.
-const BearerTokenKey = "bearer.token"
-const bearerToken = contextKey(BearerTokenKey)
+const (
+	// Key is the key name for the bearer token context value.
+	Key = "bearer.token"
```
albertteoh, line 23: `Key` is used outside of this package [here](https://github.com/jaegertracing/jaeger/blob/6685913c38a7168bc32c6a8cfb5a37546029f390/plugin/storage/grpc/shared/grpc_client.go#L76).

pkg/bearertoken/http.go
```diff
@@ -12,18 +12,20 @@
...
 	"go.uber.org/zap"
-
-	"github.com/jaegertracing/jaeger/storage/spanstore"
 )
 
-func bearerTokenPropagationHandler(logger *zap.Logger, h http.Handler) http.Handler {
+// PropagationHandler returns a http.Handler containing the logic to extract
+// the Authorization token from the http.Request and inserts it into the http.Request
+// context for easier access to the request token via GetBearerToken for bearer token
+// propagation use cases.
```
yurishkuro, line 27: -

```suggestion
// PropagationHandler returns a http.Handler containing the logic to extract
// the Bearer token from the Authorization header of the http.Request and insert it into request.Context 
// for propagation. The token can be accessed via GetBearerToken.
```

pkg/bearertoken/context.go
albertteoh: Addressed in 696cb5f.

yurishkuro (approved):

albertteoh:
Thanks for the review!

PR #3345, opened 2021-10-28
fix sending on closed channel

### LINKED ISSUE #3342: BoundedQueue.Produce may panic after BoundedQueue.Stop is called
**Describe the bug**
BoundedQueue.Produce may panic because of sending on closed channel after BoundedQueue.Stop is called.

When q.stopped.Store(1)  is called in method BoundedQueue.Stop,  some producers may run in code below concurrently.

https://github.com/jaegertracing/jaeger/blob/c5642b708be9e2577cc1c494889ae97946ccc78a/pkg/queue/bounded_queue.go#L124-L131

That means BounedQueue.Stop method cannot stop all producer from sending data to channel.

If code  in BoundedQueue.Stop
https://github.com/jaegertracing/jaeger/blob/c5642b708be9e2577cc1c494889ae97946ccc78a/pkg/queue/bounded_queue.go#L151
happens before code in BoundedQueue.Produce
https://github.com/jaegertracing/jaeger/blob/c5642b708be9e2577cc1c494889ae97946ccc78a/pkg/queue/bounded_queue.go#L133

BoundedQueue.Produce will panic, because there are no synchronization mechanisms to guarantee the execution order between channel closing and enqueuing.

**To Reproduce**
1. To make the described scenario happen every time,  sleep for a while before sending data to channel in BoundedQueue.Produce.

> 	if q.Size() >= q.Capacity() {
		// note that all items will be dropped if the capacity is 0
		q.onDroppedItem(item)
		return false
	}
	**time.Sleep(time.Second)**
	q.size.Add(1)

2. add a unit test, stop the queue immediately after producing.
func TestBoundedQueueClose(t *testing.T) {
	q := NewBoundedQueue(1000, func(item interface{}) {
	})
	q.StartConsumersWithFactory(1, func() Consumer {
		return ConsumerFunc(func(item interface{}) {})
	})
	started := make(chan interface{})
	finished := make(chan interface{})
	go func() {
		started <- struct{}{}
		a := q.Produce("test")
		t.Log(a)
		finished <- struct{}{}
	}()
	<-started
	q.Stop()
	<-finished
}
3. unit test fails and prints the error stack.

>  panic: send on closed channel

goroutine 6 [running]:
github.com/jaegertracing/jaeger/pkg/queue.(*BoundedQueue).Produce(0x14000012460, {0x100afc4a0, 0x100b326b8})
	/Users/tt/opensource/jaeger/pkg/queue/bounded_queue.go:132 +0xc4
github.com/jaegertracing/jaeger/pkg/queue.TestBoundedQueueClose.func3(0x140000703c0, 0x14000012460, 0x14000116ea0, 0x14000070420)
	/Users/tt/opensource/jaeger/pkg/queue/bounded_queue_test.go:372 +0x6c
created by github.com/jaegertracing/jaeger/pkg/queue.TestBoundedQueueClose
	/Users/tt/opensource/jaeger/pkg/queue/bounded_queue_test.go:370 +0x204

**Expected behavior**
adding synchronization mechanisms between  closing and enqueuing operations.

 - OS:  Linux
 - Jaeger version:  Latest

Sniper91:
Resolves #3342 
- add  mutex between closing and enqueuing channel operations.


yurishkuro:
I don't really follow this change, the logic is made even more complicated than it is. I don't think it's worth solving, this queue won't be used in jaeger-v2.

PR #3350, opened 2021-10-28
Issue #3349 - enable `-shuffle=on` in tests

### LINKED ISSUE #3349: Enable test shuffle for more robust test suite

Go 1.17 introduced the `-shuffle` flag to avoid deterministic test execution order. The goal of using `-shuffle` is to make sure that the test suite is stable and each test is well isolated from the other. 

I suggest we introduce `-shuffle` in the `test` target

rbroggi:
- Resolves #3349 

- Enabling `shuffling` in tests


rbroggi:
it seems that this enhancement is already spotting some tests-interdependencies... What do you think?

yurishkuro:
Generally I am not aware of inter-test dependencies, we always tried to keep tests independent. I see that TestAggregateThroughput is failing, which is a relatively recent test, some side-effect dependency might have sneaked in & we should fix it.

plugin/sampling/strategystore/adaptive/processor_test.go
```diff
@@ -79,8 +79,15 @@ var (
...
 func TestAggregateThroughput(t *testing.T) {
+	testThroughput := []*model.Throughput{
+		{Service: "svcA", Operation: "GET", Count: 4, Probabilities: map[string]struct{}{"0.1": {}}},
+		{Service: "svcA", Operation: "GET", Count: 4, Probabilities: map[string]struct{}{"0.2": {}}},
+		{Service: "svcA", Operation: "PUT", Count: 5, Probabilities: map[string]struct{}{"0.1": {}}},
+		{Service: "svcB", Operation: "GET", Count: 3, Probabilities: map[string]struct{}{"0.1": {}}},
+	}
+
 	p := &Processor{}
-	aggregatedThroughput := p.aggregateThroughput(testThroughputs)
```
rbroggi, line 83: It's very likely that this global variable is the source of the flakiness as it's not `read-only` as it should therefore impacting also this case.

rbroggi:
> Generally I am not aware of inter-test dependencies, we always tried to keep tests independent. I see that TestAggregateThroughput is failing, which is a relatively recent test, some side-effect dependency might have sneaked in & we should fix it.

Found the source of the bug: global variable :relieved:

My approach in the test was to change all the global variables of the file to methods which return the same value, this way we cannot share state across tests (even in future developments)

rbroggi:
Update on this one:

`TestHotReloadUIConfigTempFile` fails now. I think `shuffle` is already paying some dividends spoting flakiness. 
It will probably take a while for me to troubleshoot those issues as I'm not yet familiar with the code. 

Another intriguing thing is that the `all-in-one` tests fail but only for the `debug` image, the `release` one works fine :thinking:

plugin/sampling/strategystore/adaptive/processor_test.go
yurishkuro: I am a bit concerned by this approach. The input to `p.aggregateThroughput` is _supposed to be treated as read-only_ by that function. If it's not, that might be the bug / source of the test dependencies, not the fault of the test. The `testThroughputs` is not a global variable, it's logically a constant. By using a function instead of the same value we simply obscure the bug elsewhere.

rbroggi: Let me see if I understand your point:

since method `p.aggregateThroughput` is *supposed* to treat its inputs as `read-only`, you think that my approach could hide a potential bug in the `p.aggregateThroughput` that could potentially mutate the input? 

If this is what you mean I think my change is still valid because it's enforcing independence **between** tests. And what we could have is a specific test that tests for `p.aggregateThroughput` immutability. Currently the thing that is actually mutating the content of the global variable is a test on its body (see in master `TestRealisticRunCalculationLoop`  in `processor_test:445`)

rbroggi: I have created the following test case and to your point it indeed fail (which means that the method is indeed mutating the arguments):

```go
func TestAggregateThroughputInputsImmutability(t *testing.T) {
	p := &Processor{}
	in := testThroughputs()
	_ = p.aggregateThroughput(in)
	assert.Equal(t, in, testThroughputs())
}

```

rbroggi: And if indeed it's supposed not to mutate the arguments the issue is exactly here:

`plugin/sampling/strategystore/adaptive/processor.go:313` 

Pointer semantics :yum:

plugin/sampling/strategystore/adaptive/processor_test.go
```diff
@@ -69,18 +71,29 @@ var (
...
-	testCalculator = calculationstrategy.CalculateFunc(func(targetQPS, qps, oldProbability float64) float64 {
+func testCalculator() calculationstrategy.ProbabilityCalculator {
+	return calculationstrategy.CalculateFunc(func(targetQPS, qps, oldProbability float64) float64 {
 		factor := targetQPS / qps
 		return oldProbability * factor
 	})
-)
+}
+
+func TestAggregateThroughputInputsImmutability(t *testing.T) {
```
rbroggi, line 87: if you run this test method on `master` branch it will fail

plugin/sampling/strategystore/adaptive/processor.go
```diff
@@ -310,12 +310,26 @@ func (p *Processor) aggregateThroughput(throughputs []*model.Throughput) service
 			t.Count += throughput.Count
 			t.Probabilities = merge(t.Probabilities, throughput.Probabilities)
 		} else {
-			aggregatedThroughput[service][operation] = throughput
```
rbroggi, line 313: This is probably a bug (unintended pointer copy makes so that the input argument `throughputs` is changed during this method execution)

rbroggi: bug hunted thanks to `shuffle` :yum:

plugin/sampling/strategystore/adaptive/processor_test.go
yurishkuro: +1

Given that the CI still fails due to some other issues, do you want to merge this PR without the `-shuffle=on` in the Makefile?

rbroggi: I think it will profitable to keep the interesting commenting history therefore:

https://github.com/jaegertracing/jaeger/pull/3360

PR #3355, opened 2021-10-29
Remove dependencies on unused tools, install tools explicitly instead of via go.mod

rbroggi:

golint has been deprecated and is no longer used in this repository. Eliminating the useless import in the tools.go.


tools.go
```diff
@@ -25,5 +25,4 @@ import (
 	_ "github.com/securego/gosec/cmd/gosec"
 	_ "github.com/vektra/mockery/cmd/mockery"
 	_ "github.com/wadey/gocovmerge"
-	_ "golang.org/x/lint/golint"
```
yurishkuro, line 28: actually, we need to replace it with import of `github.com/golangci/golangci-lint`, because it is being explicitly installed by version in `install-tools` Makefile target instead of using go.mod as we do with other tools.

pkg/es/wrapper/wrapper_nolint.go
```diff
@@ -16,9 +16,6 @@ package eswrapper
 
 import "github.com/jaegertracing/jaeger/pkg/es"
 
-// Some of the functions of elastic.BulkIndexRequest violate golint rules,
-// e.g. Id() should be ID() and BodyJson() should be BodyJSON().
-
```
yurishkuro, line 21: this feels like the relevant comment that explains the need for this file.

tools.go
rbroggi: what about simply removing this file? It seems build tag 'tool' is nowhere called... 

github.com/golangci/golangci-lint is not a valid package (it's a binary)

yurishkuro: This file is needed so that the respective packages are added to go.mod and when we do `go install` we use the versions pinned in go.mod, not the random/latest ones.

I am not sure what you mean that github.com/golangci/golangci-lint is not a valid package - it's the source code of the tool , we currently install it as 
```
go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.42.0
```
but once we have it in go.mod with the version we should just install it with `go install github.com/golangci/golangci-lint/cmd/golangci-lint`

tools.go
```diff
@@ -19,11 +19,8 @@ package jaeger
 
 // https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module
 import (
-	_ "honnef.co/go/tools/cmd/staticcheck"
```
rbroggi, line 22: not used in our current configuration of `golangci-lint` or in any other form

tools.go
```diff
@@ -19,11 +19,8 @@ package jaeger
...
-
+	_ "github.com/golangci/golangci-lint/cmd/golangci-lint"
 	_ "github.com/mjibson/esc"
-	_ "github.com/securego/gosec/cmd/gosec"
```
rbroggi, line 25: indirect dependency of `golangci-lint` will be naturally included in `go.mod` as indirect dependency.

Makefile
```diff
@@ -409,12 +410,8 @@ generate-zipkin-swagger: init-submodules
 	$(SWAGGER) generate server -f ./idl/swagger/zipkin2-api.yaml -t $(SWAGGER_GEN_DIR) -O PostSpans --exclude-main
 	rm $(SWAGGER_GEN_DIR)/restapi/operations/post_spans_urlbuilder.go $(SWAGGER_GEN_DIR)/restapi/server.go $(SWAGGER_GEN_DIR)/restapi/configure_zipkin.go $(SWAGGER_GEN_DIR)/models/trace.go $(SWAGGER_GEN_DIR)/models/list_of_traces.go $(SWAGGER_GEN_DIR)/models/dependency_link.go
 
-.PHONY: install-mockery
```
rbroggi, line 412: I took the liberty to merge all the tools installation into a single make target. makes sense?

tools.go
rbroggi: Ok, I have tried to replicate your suggestion to the best of my understanding. Please notice that:

1. having `golangci-lint` in our `go.mod` is bringing *a lot* of indirect dependencies as well (see the `go.mod` file). This is a consequence of the nature of `golangci-lint` - it is a linter aggregator and therefore it declares all the supported linters as its dependencies. In Jaeger case this might be a big price to pay given that we are only using 3 out of the [very long list of supported linters](https://golangci-lint.run/usage/linters/).
2. Since we do not use `gosec` directly but only indirectly through `golangci-lint` I have removed the explicit dependency from the `tools.go` file. 
3. I have removed altogether the import of `honnef.co/go/tools/cmd/staticcheck` as we are not using it directly or indirectly. (correct me if I'm wrong)
4. I took the liberty to merge `install-mockery` into `install-tools` target because it felt tidier IMO. Please let me know if you prefer to keep them separate (and if there is a rational for that I would be interested to know as well).

yurishkuro: Hm, I think it was a mistake that staticcheck was removed https://github.com/jaegertracing/jaeger/pull/3237/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L169 -- cc @albertteoh

Makefile
yurishkuro: +1

pkg/es/wrapper/wrapper_nolint.go
yurishkuro: please restore the commeng

yurishkuro:
> having golangci-lint in our go.mod is bringing a lot of indirect dependencies

I tend to agree, this went into the wrong direction, we don't want to be tracking so many dependencies.  I think we should go back to installing golang-ci via explicit version and not list it in the tools.go

Question, however: when we install golang-ci directly, not via go.mod, (a) does it auto-install all other required tools, and (b) does it manage their versions or always gets the latest?

tools.go
albertteoh: The `staticcheck` linter is included by default: https://golangci-lint.run/usage/linters/#enabled-by-default-linters

Confirmed with:
```
$ make lint
...
INFO [lintersdb] Active 11 linters: [deadcode gofmt gosec gosimple govet ineffassign staticcheck structcheck typecheck unused varcheck] 
```

albertteoh:
> (a) does it auto-install all other required tools, and 

Yes, that's my understanding. To test it, I removed by `$GOPATH/pkg/mod` dir and `$GOPATH/bin/*` and ran:

```
$  make lint                   
golangci-lint -v run
make: golangci-lint: No such file or directory
make: *** [lint] Error 1

$ make install-tools
...

$ make lint
golangci-lint -v run
...
INFO Execution took 10.59982082s                  
./scripts/updateLicenses.sh > .fmt.log
./scripts/import-order-cleanup.sh stdout > .import.log
```

> (b) does it manage their versions or always gets the latest?

My understanding is that it manages their versions:

```
$ make install-tools
...
go: downloading github.com/polyfloyd/go-errorlint v0.0.0-20210510181950-ab96adb96fea
...
```

is consistent with golangci-lint v1.42.0's go-errorlint dependency: https://github.com/golangci/golangci-lint/blob/v1.42.0/go.mod#L61.

Switching to golangci-lint's master branch, we see this [dependency version has been upgraded](https://github.com/golangci/golangci-lint/blob/master/go.mod):
```
	github.com/polyfloyd/go-errorlint v0.0.0-20210722154253-910bb7978349
```

tools.go
rbroggi: I wanted to complement with some info:

As a matter of fact in the [official golangci-lint documentation](https://golangci-lint.run/usage/install/) the `go install github.com/golangci/golangci-lint/cmd/golangci-lint@<version>` is the advised best approach. 

Some things I think we should tackle nevertheless:

1. Having the tools versioning in `tools.go` or in `make install-tools` sounds good to me, but I think that it would be simpler/clearer if we pick one instead of using a hybrid approach (some tools versioned in `go.mod` and others explicitly specified in the `install-tools`). This is why earlier I suggested deleting the `tools.go`. In the context of this PR I suggest we only do a cleanup and follow up on actions 3 and 4 below.
2. Even if you choose to keep the `tools.go` we do not need to declare `gosec` and `staticcheck` as those are indirect dependencies of `golangci-lint` and will be taken care of by `install-tools` target.
3. (to be done in separate PR) `https://github.com/wadey/gocovmerge` last commit was in 2016 - I would consider either:
    1. vendoring it
    2. copying it to this repository
    3. since the script is very simple, create our own version and maintain it in the scope of this repo 
4. (to be done in separate PR) considering to upgrade `mockery` to v2.

pkg/es/wrapper/wrapper_nolint.go
rbroggi: I did it already

tools.go
yurishkuro: I thought we don't need gocovmerge anymore since we no longer test pkg by pkg

rbroggi: I think you are absolutely right :) 

I have updated the PR.

pkg/es/wrapper/wrapper_nolint.go
yurishkuro: it still shows as deleted in the latest version of PR

rbroggi: my bad. Updated now

yurishkuro (approved):

PR #3356, opened 2021-10-29
Favor `ubuntu-20.04` over `ubuntu-latest`

rbroggi:

`latest` tag in github actions should be avoided for the same reasons it is avoided in container images


yurishkuro:
Pinning concrete release number just creates maintenance issues going forward. I don't see much risk in depending on latest from CI.

PR #3379, opened 2021-11-06
Improve release notes script

albertteoh:

- When generating release notes, the most relevant commit messages for the release are those that are before the previous release tag.
- `make changelog` nicely outputs these relevant changes, however, the release manager either needs to scroll to the top where the relevant commits are then copy-paste them.
- `make changelog` depends on a personal docker image and would be a burden to the owning engineer to keep publishing improvements/updates.
- In most cases, `dependabot` version bumps are not relevant to a changelog.

This PR proposes the following improvements:

- Uses argparse to manage arguments, providing sensible defaults, which has the following benefits:
  - More robust parameter handling.
  - Better help documentation.
- Given python3 is in mainstream use (and python 2 discontinued), I've switched to using f-strings for readability and reduced stutter.
- Add a convenience option to `--exclude-dependabot`.
- If `--commit-num` is not supplied, the script will find the number of commits from HEAD of `master` to the previous tag and list out those commits.

### Examples

Using the default parameters:
```
$ python ./scripts/release-notes.py
There are 29 new commits since v1.27.0
https://api.github.com/repos/jaegertracing/jaeger/commits?per_page=29
Retrieved 29 commits
* Prepare release 1.28.0 ([@albertteoh](https://github.com/albertteoh) in [#3378](https://github.com/jaegertracing/jaeger/pull/3378))
* Do not throw error on empty indices in Elasticsach rollover lookback ([@jkandasa](https://github.com/jkandasa) in [#3369](https://github.com/jaegertracing/jaeger/pull/3369))
...
* Bump github.com/dgraph-io/badger/v3 from 3.2103.1 to 3.2103.2 ([@dependabot[bot]](https://github.com/apps/dependabot) in [#3310](https://github.com/jaegertracing/jaeger/pull/3310))
* List Yarn as pre-requisite for run-all-in-one ([@rsvoboda](https://github.com/rsvoboda) in [#3307](https://github.com/jaegertracing/jaeger/pull/3307))
```

All optional params provided:
```
$ python ./scripts/release-notes.py --token-file my-github-token-file --num-commits 3 --exclude-dependabot --repo jaeger-ui    
https://api.github.com/repos/jaegertracing/jaeger-ui/commits?per_page=3
Retrieved 3 commits
* Prepare release v1.18.0 ([@albertteoh](https://github.com/albertteoh) in [#846](https://github.com/jaegertracing/jaeger-ui/pull/846))
* fix deep dependency graph not show in uiEmbed=v0 mode ([@leroy-chen](https://github.com/leroy-chen) in [#768](https://github.com/jaegertracing/jaeger-ui/pull/768))

(Skipped 1 dependabot commit)
```

### DIFF
--- a/scripts/release-notes.py
+++ b/scripts/release-notes.py
@@ -95,13 +95,11 @@ def main(token, repo, num_commits, exclude_dependabot):
 
 
 if __name__ == "__main__":
-    if "OAUTH_TOKEN" not in os.environ:
-        eprint("Please set OAUTH_TOKEN environment variable: " +
-               "https://github.com/settings/tokens/new?description=GitHub%20Changelog%20Generator%20token")
-        sys.exit(1)
-
     parser = argparse.ArgumentParser(description='List changes based on git log for release notes.')
 
+    parser.add_argument('--token', type=str,
+                        help='The personal github token to access the github API. ' +
+                             '(default: uses OATH_TOKEN environment variable)')
     parser.add_argument('--repo', type=str, default='jaeger',
                         help='The repository name to fetch commit logs from. (default: jaeger)')
     parser.add_argument('--exclude-dependabot', action='store_true',
@@ -111,4 +109,13 @@ if __name__ == "__main__":
                              '(default: number of commits before the previous tag)')
 
     args = parser.parse_args()
-    main(os.environ["OAUTH_TOKEN"], args.repo, args.num_commits, args.exclude_dependabot)
+
+    token = args.token
+    if not token:
+        if "OAUTH_TOKEN" not in os.environ:
+            eprint("Please provide a --token or set OAUTH_TOKEN environment variable to a generated token: " +
+                   "https://github.com/settings/tokens/new?description=GitHub%20Changelog%20Generator%20token")
+            sys.exit(1)
+        token = os.environ["OAUTH_TOKEN"]
+
+    main(token, args.repo, args.num_commits, args.exclude_dependabot)
--- a/scripts/release-notes.py
+++ b/scripts/release-notes.py
@@ -91,7 +91,7 @@ def main(token, repo, num_commits, exclude_dependabot):
         print(f'* {msg} ([@{author}]({author_url}) in [#{pull_id}]({pull_url}))')
 
     if skipped_dependabot:
-        print(f"\n(Skipped {skipped_dependabot} dependabot commits)")
+        print(f"\n(Skipped {skipped_dependabot} dependabot commit{'' if skipped_dependabot == 1 else 's'})")
 
 
 if __name__ == "__main__":
--- a/scripts/release-notes.py
+++ b/scripts/release-notes.py
@@ -10,8 +10,9 @@
 
 import argparse
 import json
+import os.path
 import urllib.parse
-import os
+from os.path import expanduser
 import sys
 from urllib.request import (
     urlopen,
@@ -23,7 +24,7 @@ def eprint(*args, **kwargs):
     print(*args, file=sys.stderr, **kwargs)
 
 
-def num_commits_before_prev_tag(token, base_url):
+def num_commits_since_prev_tag(token, base_url):
     tags_url = f"{base_url}/tags"
 
     req = Request(tags_url)
@@ -36,7 +37,7 @@ def num_commits_before_prev_tag(token, base_url):
     compare_results = json.loads(urlopen(req).read())
     num_commits = compare_results['behind_by']
 
-    print(f"master is behind {prev_release_tag} by {num_commits} commits")
+    print(f"There are {num_commits} new commits since {prev_release_tag}")
     return num_commits
 
 
@@ -48,7 +49,10 @@ def main(token, repo, num_commits, exclude_dependabot):
 
     # If num_commits isn't set, get the number of commits made since the previous release tag.
     if not num_commits:
-        num_commits = num_commits_before_prev_tag(token, base_url)
+        num_commits = num_commits_since_prev_tag(token, base_url)
+
+    if not num_commits:
+        return
 
     # Load commits
     data = urllib.parse.urlencode({'per_page': num_commits})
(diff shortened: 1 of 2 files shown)


scripts/release-notes.py
```diff
@@ -1,67 +1,121 @@
...
+    req.add_header("Authorization", f"token {token}")
+    tags = json.loads(urlopen(req).read())
+    prev_release_tag = tags[0]['name']
+    compare_url = f"{base_url}/compare/master...{prev_release_tag}"
+    req = Request(compare_url)
+    req.add_header("Authorization", f"token {token}")
+    compare_results = json.loads(urlopen(req).read())
+    num_commits = compare_results['behind_by']
+
+    print(f"master is behind {prev_release_tag} by {num_commits} commits")
```
yurishkuro, line 39: "behind" sounds like you need a pull. I'd rather say "N new commits since prevtag "

scripts/release-notes.py
```diff
@@ -1,67 +1,121 @@
...
-    msg = msg.replace('(#{pull})'.format(pull=pull_no), '').strip()
-    print('* {msg} ([@{author}]({author_url}) in [#{pull}]({pull_url}))'.format(
-        msg=msg,
-        author=author,
-        author_url=author_url,
-        pull=pull_no,
-        pull_url=pull_url,
-    ))
+
+def num_commits_before_prev_tag(token, base_url):
```
yurishkuro, line 26:
```suggestion
def num_commits_since_prev_tag(token, base_url):
```

scripts/release-notes.py
```diff
@@ -1,67 +1,121 @@
...
+    parser.add_argument('--exclude-dependabot', action='store_true',
+                        help='Excludes dependabot commits. (default: false)')
+    parser.add_argument('--num-commits', type=int,
+                        help='Print this number of commits from git log. ' +
+                             '(default: number of commits before the previous tag)')
+
+    args = parser.parse_args()
+
+    token = args.token
+    if not token:
```
yurishkuro, line 114: NB: it's generally not a good practice to accept secrets as cmd line args, because other users on the box can see them in ps. Better options are either read it from stdin or from a file (which can have default name like ~/.github_token) which can be set only once.

yurishkuro (commented):
Do you think then that we can drop recommendation to use docker image and only use this script?

albertteoh:
> Do you think then that we can drop recommendation to use docker image and only use this script?

That's what I prefer (and have made the change in the PR). Happy to use the docker image if that's the general preference of maintainers. The benefit of a docker image is that it provides all the correct version dependencies required for the python script, which I think was particularly useful during the sunsetting period of python 2.

However, I believe python 3 should be in the mainstream now and, _I think_, the libs used in this script come with the default python 3 installation.

scripts/release-notes.py
```diff
@@ -1,67 +1,133 @@
...
+    parser.add_argument('--exclude-dependabot', action='store_true',
+                        help='Excludes dependabot commits. (default: false)')
+    parser.add_argument('--num-commits', type=int,
+                        help='Print this number of commits from git log. ' +
+                             '(default: number of commits before the previous tag)')
+
+    args = parser.parse_args()
+    generate_token_url = "https://github.com/settings/tokens/new?description=GitHub%20Changelog%20Generator%20token"
+    generate_err_msg = (f"Please generate a token from this URL: {generate_token_url} and "
+                        f"place it in the token-file. The token only needs default permissions.")
```
yurishkuro, line 118: > The token only needs default permissions.

Not sure what "default permissions" mean - I would instead explicitly recommend `chmod 0600`

yurishkuro:
```suggestion
                        f"place it in the token-file. Protect the file so only you can read it: chmod 0600 <file>.")
```

scripts/release-notes.py
albertteoh: Good point, addressed in f78e3d1. 

Note: I've removed support for the `OAUTH_TOKEN` env var (a breaking change) because:
- It simplifies the script. Avoids handling different ways of supplying the token.
- Prevents users from inlining the token env var in the python command, which defeats the purpose of avoiding exposing the token over `ps`.

yurishkuro: +1

scripts/release-notes.py
albertteoh: Ah, I wanted to respect what was stated in the original [Makefile comment](https://github.com/jaegertracing/jaeger/blob/master/Makefile#L365). I thought it was referring to the privileges (scopes) given to the generated token in terms of API access. You can see them when clicking on this link: https://github.com/settings/tokens/new?description=GitHub%20Changelog%20Generator%20token

Perhaps it could be worded a little better.

yurishkuro (approved):

scripts/release-notes.py
```diff
@@ -1,67 +1,133 @@
...
-    ))
+
+def num_commits_since_prev_tag(token, base_url):
+    tags_url = f"{base_url}/tags"
+
+    req = Request(tags_url)
+    req.add_header("Authorization", f"token {token}")
+    tags = json.loads(urlopen(req).read())
+    prev_release_tag = tags[0]['name']
+    compare_url = f"{base_url}/compare/master...{prev_release_tag}"
```
jpkrohling, line 34: Perhaps externalize the branch name? We might (should!) eventually replace it with `main`.

jpkrohling (approved):

scripts/release-notes.py
albertteoh: To clarify, did you mean placing it in a variable within the script?

jpkrohling: yes, but feel free to merge it without for now

albertteoh: Thanks @jpkrohling, I've addressed this in 1e8111a.

jpkrohling (approved):

PR #3383, opened 2021-11-10
Add remote gRPC option for storage plugin

### LINKED ISSUE #3377: Add the ability to use the gRPC storage plugin with a remote endpoint

Ability to have the gRPC storage API be implemented on a remote endpoint (not only locally over a socket).

**I'm happy to submit a PR if the community is open to this.**

Right now the [gRPC storage plugin](https://github.com/jaegertracing/jaeger/tree/master/plugin/storage/grpc) can only be used via the go-hashicorp plugin (over a socket). I'd like to expose the storage API on another machine, so having a way for the gRPC plugin to use a regular gRPC client instead would be much easier and also more efficient.

In the short term I've solved this with a [gRPC plugin](https://github.com/timescale/promscale/tree/master/cmd/jaeger-query-proxy) that simply proxies the request to a gRPC client with a remote endpoint so the communication is 

jaeger --(go-hashicorp-plugin)-->jaeger-query-proxy--(gRPC)--> grpc Endpoint implementing the storage API. 

But the jaeger-query-proxy implements the same exact API as the grpc Endpoint implementing the storage API. so it would be much more efficient to just have

jaeger --(gRPC)--> grpc Endpoint implementing the storage API. 

I think there would need to be some additional configuration options added to the grpc plugin ([here](https://github.com/jaegertracing/jaeger/blob/master/plugin/storage/grpc/config/config.go#L32)):
- 	TLS                bool          `yaml:"tls"`
-	CaFile             string        `yaml:"cafile,omitempty"`
-	ServerAddr         string        `yaml:"grpc-server"`
-	ServerHostOverride string        `yaml:"grpc-server-host-override,omitempty"`
-	ConnectTimeout     time.Duration `yaml:"connection-timeout,omitempty"`

And then in the [plugin Build function](https://github.com/jaegertracing/jaeger/blob/master/plugin/storage/grpc/config/config.go#L50) we'd need to initialize a regular gRPC client if using a remote connection (instead of plugin.NewClient from the hashicorp plugin).

Of course the existing hashicorp way would still work and not be effected.

I don't think this would be a big change and again I'd be happy to submit a PR. 

I think in the choice between the plugin or the remote client would be done depending on whether the PluginBinary configuration option is an empty string or not.

I'm hoping that the hashicorp client and the gRPC client would provide an equivalent interface but I'd have to see during implementation.

cevian:
Resolves #3377 

Adds the option to host a gRPC storage API on a remote endpoint
using regular gRPC. Previously the plugin system only supported
local socket connections through the go-hashicorp plugin system.


cevian:
I couldn't find how the existing plugin was tested end-to-end, so couldn't write better tests. If more tests are needed, please point me in the right direction.

cevian:
Also wasn't sure about the CLI option naming.

plugin/storage/grpc/config/config.go
```diff
@@ -15,24 +15,32 @@
...
 type Configuration struct {
-	PluginBinary            string `yaml:"binary" mapstructure:"binary"`
-	PluginConfigurationFile string `yaml:"configuration-file" mapstructure:"configuration_file"`
-	PluginLogLevel          string `yaml:"log-level" mapstructure:"log_level"`
+	PluginBinary             string        `yaml:"binary" mapstructure:"binary"`
+	PluginConfigurationFile  string        `yaml:"configuration-file" mapstructure:"configuration_file"`
+	PluginLogLevel           string        `yaml:"log-level" mapstructure:"log_level"`
+	RemoteServerAddr         string        `yaml:"server" mapstructure:"server"`
+	RemoteTLS                bool          `yaml:"tls" mapstructure:"tls"`
+	RemoteCAFile             string        `yaml:"cafile" mapstructure:"cafile"`
```
yurishkuro, line 41: We have a shared config/options struct for configuring TLS, see other storage configs for examples

yurishkuro (commented):
Can we adapt grpc-plugin storage integration test to test via remote grpc as well?

cevian:
@yurishkuro for the integration tests I will need to either:
1) modify examples/memstore-plugin to be able to accept a gRPC address cli flag to listen on (and then either start the plugin or the grpc server depending on the flag)  
or
2) create a new examples/memstore-remote to only act as a gRPC server (not the plugin)

Option 1 is less code changes but the examples become a bit more muddled. Do you have any preference if we want a separate example for this?

cevian:
Hmm I like that idea. Let me try it out.

plugin/storage/integration/grpc_test.go
```diff
@@ -87,6 +145,31 @@ func (s *GRPCStorageIntegrationTestSuite) cleanUp() error {
 	return s.initialize()
 }
 
+type memoryStorePlugin struct {
```
cevian, line 148: This is copied from `examples/memstore-plugin/main.go` and I'd like to avoid duplicate code but I don't know where to put this so it can be shared. Ideas most welcome.

plugin/storage/grpc/config/config.go
```diff
@@ -43,11 +51,62 @@ type ClientPluginServices struct {
...
+		opts = append(opts, grpc.WithTransportCredentials(creds))
+	} else {
+		opts = append(opts, grpc.WithInsecure())
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), c.RemoteConnectTimeout)
+	defer cancel()
+	conn, err := grpc.DialContext(ctx, c.RemoteServerAddr, opts...)
+	if err != nil {
+		return nil, fmt.Errorf("error connecting to Promscale GRPC server: %w", err)
```
yurishkuro, line 96:
```suggestion
		return nil, fmt.Errorf("error connecting to remote storage: %w", err)
```

plugin/storage/grpc/options.go
```diff
@@ -16,17 +16,23 @@ package grpc
...
-	pluginBinary            = "grpc-storage-plugin.binary"
-	pluginConfigurationFile = "grpc-storage-plugin.configuration-file"
-	pluginLogLevel          = "grpc-storage-plugin.log-level"
-	defaultPluginLogLevel   = "warn"
+	pluginBinary             = "grpc-storage-plugin.binary"
+	pluginConfigurationFile  = "grpc-storage-plugin.configuration-file"
+	pluginLogLevel           = "grpc-storage-plugin.log-level"
+	remotePrefix             = "grpc-storage"
+	pluginServer             = remotePrefix + ".server"
+	pluginConnectionTimeout  = remotePrefix + ".connection-timeout"
```
yurishkuro, line 33:
```suggestion
	remoteServer             = remotePrefix + ".server"
	remoteConnectionTimeout  = remotePrefix + ".connection-timeout"
```

plugin/storage/grpc/config/config.go
```diff
@@ -43,11 +51,62 @@ type ClientPluginServices struct {
...
+	return nil
+}
+
+func (c *Configuration) buildRemote(logger *zap.Logger) (*ClientPluginServices, error) {
+	opts := []grpc.DialOption{
+		grpc.WithUnaryInterceptor(otgrpc.OpenTracingClientInterceptor(opentracing.GlobalTracer())),
+		grpc.WithStreamInterceptor(otgrpc.OpenTracingStreamClientInterceptor(opentracing.GlobalTracer())),
+		grpc.WithBlock(),
+	}
+	var err error
```
yurishkuro, line 80: do you need this? I only see one assignment in L94

plugin/storage/grpc/options.go
```diff
@@ -35,16 +41,32 @@ type Options struct {
...
+
 // AddFlags adds flags for Options
 func (opt *Options) AddFlags(flagSet *flag.FlagSet) {
+	var tlsFlagsConfig = tlsFlagsConfig()
+	tlsFlagsConfig.AddFlags(flagSet)
+
 	flagSet.String(pluginBinary, "", "The location of the plugin binary")
 	flagSet.String(pluginConfigurationFile, "", "A path pointing to the plugin's configuration file, made available to the plugin with the --config arg")
 	flagSet.String(pluginLogLevel, defaultPluginLogLevel, "Set the log level of the plugin's logger")
+	flagSet.String(remoteServer, "", "The server address for the remote gRPC server")
```
yurishkuro, line 58: "server" twice
```suggestion
	flagSet.String(remoteServer, "", "The remote storage gRPC server address")
```

plugin/storage/grpc/options.go
```diff
@@ -35,16 +41,32 @@ type Options struct {
...
+	flagSet.Duration(remoteConnectionTimeout, defaultConnectionTimeout, "The connection timeout for connecting to the remote server")
```
yurishkuro, line 59:
```suggestion
	flagSet.Duration(remoteConnectionTimeout, defaultConnectionTimeout, "The remote storage gRPC server connection timeout")
```

plugin/storage/grpc/options.go
```diff
@@ -35,16 +41,32 @@ type Options struct {
...
+func tlsFlagsConfig() tlscfg.ClientFlagsConfig {
+	return tlscfg.ClientFlagsConfig{
+		Prefix: remotePrefix,
+	}
+}
+
 // AddFlags adds flags for Options
 func (opt *Options) AddFlags(flagSet *flag.FlagSet) {
+	var tlsFlagsConfig = tlsFlagsConfig()
+	tlsFlagsConfig.AddFlags(flagSet)
```
yurishkuro, line 53:
```suggestion
	tlsFlagsConfig().AddFlags(flagSet)
```

plugin/storage/grpc/options.go
```diff
@@ -35,16 +41,32 @@ type Options struct {
...
 }
 
 // InitFromViper initializes Options with properties from viper
 func (opt *Options) InitFromViper(v *viper.Viper) {
+	var tlsFlagsConfig = tlsFlagsConfig()
 	opt.Configuration.PluginBinary = v.GetString(pluginBinary)
 	opt.Configuration.PluginConfigurationFile = v.GetString(pluginConfigurationFile)
 	opt.Configuration.PluginLogLevel = v.GetString(pluginLogLevel)
+	opt.Configuration.RemoteServerAddr = v.GetString(remoteServer)
+	opt.Configuration.RemoteTLS = tlsFlagsConfig.InitFromViper(v)
```
yurishkuro, line 70:
```suggestion
	opt.Configuration.RemoteTLS = tlsFlagsConfig().InitFromViper(v)
```

plugin/storage/grpc/shared/plugin.go
```diff
@@ -36,8 +36,8 @@ type StorageGRPCPlugin struct {
 	ArchiveImpl ArchiveStoragePlugin
 }
 
-// GRPCServer implements plugin.GRPCPlugin. It is used by go-plugin to create a grpc plugin server.
-func (p *StorageGRPCPlugin) GRPCServer(broker *plugin.GRPCBroker, s *grpc.Server) error {
+// RegisterServer registers the plugin with the server
+func (p *StorageGRPCPlugin) RegisterServer(s *grpc.Server) error {
```
yurishkuro, line 40:
```suggestion
func (p *StorageGRPCPlugin) RegisterHandlers(s *grpc.Server) error {
```

plugin/storage/integration/grpc_test.go
```diff
@@ -18,45 +18,102 @@
...
+	}
+
+	s.server = googleGRPC.NewServer()
+	queryPlugin := shared.StorageGRPCPlugin{
+		Impl:        memStorePlugin,
+		ArchiveImpl: memStorePlugin,
+	}
+
+	err := queryPlugin.RegisterServer(s.server)
+	if err != nil {
```
yurishkuro, line 76:
```suggestion
	if err := queryPlugin.RegisterServer(s.server); err != nil {
```

plugin/storage/integration/grpc_test.go
```diff
@@ -18,45 +18,102 @@
...
+
+	listener, err := net.Listen("tcp", "localhost:2001")
+	if err != nil {
+		return err
+	}
+	s.wg.Add(1)
+	go func() {
+		defer s.wg.Done()
+		err = s.server.Serve(listener)
+		if err != nil {
```
yurishkuro, line 88:
```suggestion
		if err := s.server.Serve(listener); err != nil {
```

plugin/storage/integration/grpc_test.go
yurishkuro: plugin/storage/grpc/memory/

plugin/storage/integration/grpc_test.go
```diff
@@ -97,9 +179,35 @@ func TestGRPCStorage(t *testing.T) {
 	if configPath == "" {
 		t.Log("PLUGIN_CONFIG_PATH env var not set")
 	}
+
+	flags := []string{
+		"--grpc-storage-plugin.binary", binaryPath,
+		"--grpc-storage-plugin.log-level", "debug",
+	}
+	if configPath != "" {
```
yurishkuro, line 187: this check seems unnecessary, will it fail if we append `--grpc-storage-plugin.configuration-file` with empty string value?

cevian:
Thanks for the detailed review @yurishkuro. I think I addressed all the issues

plugin/storage/integration/grpc_test.go
```diff
@@ -87,6 +140,31 @@ func (s *GRPCStorageIntegrationTestSuite) cleanUp() error {
 	return s.initialize()
 }
 
+type memoryStorePlugin struct {
```
yurishkuro, line 143: do you need this type in the test, or can you use one from plugin/storage/grpc/memory/plugin.go ?

plugin/storage/grpc/factory.go
```diff
@@ -124,3 +124,8 @@ func (f *Factory) CreateArchiveSpanWriter() (spanstore.Writer, error) {
 	}
 	return f.archiveStore.ArchiveSpanWriter(), nil
 }
+
+// Close closes the resources held by the factory
```
yurishkuro, line 128: because we have this:
```
./plugin/storage/factory.go:278:			if closer, ok := factory.(io.Closer); ok {
```
I suggest adding this after struct declaration:
```
var _ io.Closer = (*Factory)(nil)
```

plugin/storage/grpc/config/config.go
```diff
@@ -43,11 +51,61 @@ type ClientPluginServices struct {
...
+	} else {
+		return c.buildRemote(logger)
+	}
+}
+
+func (c *Configuration) Close() error {
+	if c.PluginBinary == "" {
+		return c.RemoteTLS.Close()
+	}
+	return nil
```
yurishkuro, line 71:
```suggestion
	return c.RemoteTLS.Close()
```
the check is not necessary, TLSConfig.Close() is safe to call

yurishkuro (commented):
Please add some tests to plugin/storage/grpc/memory

I recommend running `make test lint` locally before submitting

yurishkuro (approved):
Thanks!

Issue #3380, opened 2021-11-08
Upgrade HotROD example to use OpenTelemetry SDK


yurishkuro:
Per #3362

* [x] replace Jaeger SDK with OTEL SDK using an OT Bridge 
* [x] replace OT instrumentation with OpenTelemetry
* [x] Remove references to OpenTracing API
  * [x] from HotROD code
  * [x] from HotROD tests `examples/hotrod/pkg/tracing/rpcmetrics/`
  * [x] from README #4790
* [x] Remove Jaeger exporter, leave only OTLP, and update README and docker-compose

rbroggi:
Will try to work on that during the next weekend.

rbroggi:
Before jumping right into the code I have been studying the opentelemetry golang API. Given that the notion  of logs is not complete in opentelemetry yet but is something used in Jaeger, I wanted to make sure we have a strategy to migrate what is currently 'logged' in HotROD to what in opentelemetry is called an 'Event'. Is this a good understanding? The only other option in open telemetry is to use Attributes (key-value) which are what in jaeger-client we call Tags. 

in the [article](https://medium.com/opentracing/take-opentracing-for-a-hotrod-ride-f6e3141f7941) there is also a section that describes the difference between tags vs Logs.

yurishkuro:
Let me clarify - at this time our intention is not to upgrade everything to OTEL API, but keep OT instrumentation and use OT-to-OTEL bridge to replace Jaeger SDK with OTEL SDK. Therefore your question is not relevant because that decision will be made by the bridge.

rbroggi:
> Let me clarify - at this time our intention is not to upgrade everything to OTEL API, but keep OT instrumentation and use OT-to-OTEL bridge to replace Jaeger SDK with OTEL SDK. Therefore your question is not relevant because that decision will be made by the bridge.

Copy that! I'm glad I asked the question! :smile:

yurishkuro:
getting close to completing this

yurishkuro:
The only remaining references to opentracing are in documentation files, still need to clean them up:
```
./README.md:71:### Native support for OpenTracing
./README.md:73:Jaeger backend, Web UI, and instrumentation libraries have been designed from the ground up to support the [OpenTracing standard](https://opentracing.io/specification/).
./README.md:74:  * Represent traces as directed acyclic graphs (not just trees) via [span references](https://github.com/opentracing/specification/blob/master/specification.md#references-between-spans)
[...]

PR #3390, opened 2021-11-14
Use OpenTelemetry SDK in HotROD 

rbroggi:
* replacing the jaeger-client-go ot tracer implementation with the ot-otel bridge implementation

- Issue #3380

- First iteration in replacing the jaeger client ot tracer implementation with the otel bridge ot implementation


examples/hotrod/pkg/tracing/init.go
```diff
@@ -16,56 +16,39 @@
...
-// Init creates a new instance of Jaeger tracer.
-func Init(serviceName string, metricsFactory metrics.Factory, logger log.Factory) opentracing.Tracer {
-	cfg := &config.Configuration{
-		Sampler: &config.SamplerConfig{},
-	}
-	cfg.ServiceName = serviceName
-	cfg.Sampler.Type = "const"
-	cfg.Sampler.Param = 1
+// Init creates a new bridge opentrace tracer that sends spans to Jaeger. The
+// bridge tracer is bind to an open telemetry tracer
```
yurishkuro, line 32:
```suggestion
// Init initializes OpenTelemetry SDK and uses OTel-OpenTracing Bridge
// to return an OpenTracing-compatible tracer.
```

yurishkuro (commented):
jaeger-ui submodule update should not be in this PR, please make sure to rebase off master

rbroggi:
Some comments:

With this change something is not really going as planned as the DAG I get is:

image

instead of the one in the [article](https://medium.com/opentracing/take-opentracing-for-a-hotrod-ride-f6e3141f7941) 

Another change that will have to be documented is that we also need to open the grpc port while running the all-in-one container for the go binary to reach jaeger:

```bash
docker run -d -p6831:6831/udp -p16686:16686 -p14268:14268 jaegertracing/all-in-one:latest
```

Will there be a way to update the medium article (it was very useful to me, it would be a shame to loose it)

rbroggi:
The frontend tracers are not connecting to the rest:

image

rbroggi:
> The frontend tracers are not connecting to the rest:
> 
> image

Dear @yurishkuro, would you have any advice on how I could troubleshoot this problem efficiently?

By investigating the code I fail to understand why this change had an impact only in the `frontend` component of the HOTRod. Well I'm not even sure whether the problem lies 100% on the frontend, the only real info I have is that there is a discontinuity between the trace from frontend and the others.

yurishkuro:
I suspect this has something to do with the propagator of the context on the wire. What happens when the code calls tracer.Inject() and the tracer is the one from the bridge?

yurishkuro:
This was finished in #4187

PR #3395, opened 2021-11-16
feat: expose rackID option in ingester samara client 

shyimo:
Signed-off-by: shaimoria <shai.moria@zooz.com>

It is common to have a Kafka cluster spanning multiple datacenters. For example, a common deployment is within an AWS region in which each availability zone is treated as a datacenter. In order to reduce cross region/rack network cost, we can tell the consumer which rack he's located on. this way the consumer will fetch only from partitions that are located on he's region/rack 

- expose `kafka.consumer.rack-id` cli argument in order to ingester consumers to fetch from closest replica


cmd/ingester/app/flags.go
```diff
@@ -111,6 +113,10 @@ func AddFlags(flagSet *flag.FlagSet) {
 		KafkaConsumerConfigPrefix+SuffixEncoding,
 		DefaultEncoding,
 		fmt.Sprintf(`The encoding of spans ("%s") consumed from kafka`, strings.Join(kafka.AllEncodings, "\", \"")))
+	flagSet.String(
+		KafkaConsumerConfigPrefix+SuffixRackID,
+		"",
+		"Configure the 'rack' in which the consumer resides to enable - must be supported by kafka server")
```
yurishkuro, line 119: Incomplete sentence

yurishkuro (commented):
Please explain the use case in the PR description

cmd/ingester/app/flags.go
shyimo: Hi @yurishkuro.
i update the description

shyimo:
@yurishkuro add explanation for the use case

yurishkuro (approved):

PR #3399, opened 2021-11-19
Add adapter that implements http.FileSystem for reading gzip files

### LINKED ISSUE #2749: Replace the use of "esc" with standard "embed" package

Ashmita152:
Co-authored-by: Yuri Shkuro <github@ysh.us>

- Resolves #2749

- Build all-in-one from this branch which came around 51M while from master it comes around 52M.


pkg/embed/gzip.go
```diff
@@ -0,0 +1,116 @@
+// Copyright (c) 2021 The Jaeger Authors.
```
yurishkuro, line 1: @Ashmita152 was this implementation copied from somewhere? If so, we need to add attribution.

yurishkuro (commented):
@Ashmita152 this is great, thanks for putting this PR together. I rebased it off my original PR and made a few refactoring changes.

pkg/gzipfs/gzip_test.go
```diff
@@ -0,0 +1,127 @@
...
+			path:            "testdata/foobar.gz",
+			expectedMode:    0444,
+			expectedName:    "foobar.gz",
+			expectedSize:    38,
+			expectedContent: "hello world",
+			expectedModTime: time.Date(1, 1, 1, 0, 0, 0, 0 /* nanos */, time.UTC),
+		},
+		{
+			name:            "compressed file without gz extension",
+			path:            "testdata/foobaz",
```
yurishkuro, line 63: Q: why do we need a separate file for this? Could we not have the same test against `testdata/foobar` Path?

pkg/gzipfs/gzip_test.go
```diff
@@ -0,0 +1,127 @@
...
+
+			stat, err := f.Stat()
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			name := stat.Name()
+			if name != c.expectedName {
+				t.Fatalf("name is wrong, expected %s, got %s", c.expectedName, name)
+			}
```
yurishkuro, line 88: nit: this is unnecessary verbose, we could achieve the same result with
```suggestion
			assert.Equals(t, c.expectedName, stat.Name())
```

pkg/gzipfs/gzip_test.go
yurishkuro: nvm

Ashmita152:
Hi Yuri,

Thank you so much for reviewing and cleaning up the code. I tried my best to improve the code coverage but unable to increase it further.

pkg/gzipfs/gzip.go
```diff
@@ -0,0 +1,118 @@
...
+		return f, err
+	}
+
+	gr, err := gzip.NewReader(f)
+	if err != nil {
+		return f, err
+	}
+	defer gr.Close()
+
+	c, err := io.ReadAll(gr)
```
yurishkuro, line 109: One way to cause an error in ReadAll is to have a .gz file that has the correct gz header (which will be read & validated by gzip.NewReader), but the rest of the content is truncated such that further reads would fail.

yurishkuro:
@Ashmita152 I added the remaining tests (error from ReadAll was a bit tricky to induce, I added a comment).

yurishkuro (approved):

Ashmita152:
Thank you Yuri, for adding the remaining tests and increasing the coverage to 100%.

PR #3412, opened 2021-11-27
Proposal: Introduce proxy

siavashs:
Jaeger proxy is a new component which can use multiple jaeger query
instances as upstreams and merges the result of queries.
This component can be useful in island architecture setups where applications
communicate across islands but jaeger service is provided locally per island.

The implementation is based on the query component and also reuses the ui.
The communication between proxy and query upstreams is based on GRPC.
All queries are performed in parallel and results are merged and returned.
The proxy also performs adjustments on traces.

The proxy component expects the proper tags to be applied during collection.
This is important to identify the source of a span (e.g. `cluster=east`).

As an extra feature it provides an option to map existing tags to new ones:
```
source tag:
hostname: app-abcd1234

cli flag:
--proxy.tags.map hostname:pod

result:
hostname: app-abcd1234
pod: app-abcd1234
```
This is useful when integrating Jaeger datasource with other data sources
which might be using different labels (`pod` instead of `hostname`, etc.)
e.g. Grafana Loki.

This is an initial implementation as a proposal so some work still remains:
- Fixing applicable tests which were copied from query component
- Adding tests specific to proxy component
- Handle GRPC client options

Signed-off-by: Siavash Safi <siavash.safi@gmail.com>


yurishkuro (changes requested):
We could use a federating proxy. However, it seems you just made a full copy of `cmd/query`. There's absolutely no need to do that. The existing cmd/query already abstracts the actual "query service" with `cmd/query/app/querysvc/query_service`. We can simply introduce a new CLI flag that enables federation mode on the existing query service, which would instantiate a different implementation of `querysvc`. The rest of the functionality such as UI embedding, http/grpc servers, etc. can remain intact.

pavolloffay:
We were internally talking about a similar use-case. It would be great to start an issue with a design proposal before we move to the implementation.

Anyways, it's great to see this @siavashs.

yeya24:
This is super useful for our setup as well and I mentioned this in https://github.com/kiali/kiali/issues/4635. Looking forward to it.

yurishkuro:
Please reopen with recommended changes

Issue #3411, opened 2021-11-25
OpenSearch integration test often fails in CI


yurishkuro:
I've noticed that the integration with OpenSearch spuriously fails quite often ([latest example](https://github.com/jaegertracing/jaeger/runs/4327831762?check_suite_focus=true)). It also cancels all other ES tests so it's hard to say if the failure is specific to OpenSearch.
```
=== RUN   TestElasticsearchStorage
    elasticsearch_test.go:193: elastic search is not ready
```

PR #3416, opened 2021-11-29
Wait for elasticsearch/opensearch to be ready before running the integration tests

Ashmita152:
Signed-off-by: Ashmita Bohara <ashmita.bohara@logz.io>

- Resolves https://github.com/jaegertracing/jaeger/issues/3411

- We are waiting for elasticsearch/opensearch to be ready before running the integration tests
- Bumping opensearch container version to 1.2.0.

### DIFF
--- a/scripts/es-integration-test.sh
+++ b/scripts/es-integration-test.sh
@@ -68,10 +68,14 @@ wait_for_it() {
     ''%{http_code}''
   )
   local counter=0
-  while [[ "$(curl ${params[@]} ${url})" != "200" && ${counter} < 30 ]]; do
+  while [[ "$(curl ${params[@]} ${url})" != "200" && ${counter} -le 30 ]]; do
     sleep 2
     counter=$((counter+1))
     echo "waiting for ${url} to be up..."
+    if [ ${counter} -eq 30 ]; then
+      echo "ERROR: elasticsearch/opensearch is down"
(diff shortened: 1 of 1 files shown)


pavolloffay (approved):

scripts/es-integration-test.sh
```diff
@@ -58,6 +58,23 @@ setup_query() {
...
+    --write-out
+    ''%{http_code}''
+  )
+  local counter=0
+  while [[ "$(curl ${params[@]} ${url})" != "200" && ${counter} < 30 ]]; do
+    sleep 2
+    counter=$((counter+1))
+    echo "waiting for ${url} to be up..."
+  done
+}
```
yurishkuro, line 76: I would recommend failing the whole script if 200 was never achieved, this way we'd get a much earlier and clearer failure signal.

yurishkuro: i.e. something like this:
```
while counter < 30 {
    status = $(curl)
    if status == 200 then return 0
}
return 1
```

yurishkuro (changes requested):
per comments

Ashmita152:
Hi Yuri, sorry for the huge delay on my side.

yurishkuro (approved):
👏 👏 👏

PR #3422, opened 2021-12-02
Add MaxConnectionAge[Grace] to collector's gRPC server

### LINKED ISSUE #1678: distirbution of traces/span  amongst collector
Are collector  load balanced ?

We have our jaegertracing setup working  with back end configured as  elastic search. Currently we have  two collector replica set up .  There are 5-10 services which sends traces to the collector  ( the number of services , keep changing ) .  I see  collectors are not evenly loaded   with traffic.  One collector reaches to the max queue usage  where as  other collector is hardly  using 20-30% capacity .  This causes  the drop from the  collector which is loaded to the capacity .    
Can we load balance the traffic  (spans)  amongst the both collector ?  I am not sure if there is any config  and i am missing it.

jpkrohling:
Closes #1678 by adding two new flags, allowing operators to control the max-connection-age and max-connection-age-grace durations for the collector's gRPC server. Setting the max-connection-age to a few seconds/minutes will cause clients to rediscover collector nodes periodically, by forcing the subchannels to go away after it reaches a certain age. The grace option allows in-flight connections to complete before being forcefully closed.

﻿Signed-off-by: Juraci Paixão Kröhling <juraci@kroehling.de>

### DIFF
--- a/cmd/collector/app/builder_flags.go
+++ b/cmd/collector/app/builder_flags.go
@@ -17,6 +17,7 @@ package app
 
 import (
 	"flag"
+	"time"
 
 	"github.com/spf13/viper"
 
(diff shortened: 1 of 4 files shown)


yurishkuro (approved):

pavolloffay:
This is great. Are there any recommended values for Kubernetes environment @jpkrohling ?

jpkrohling:
This is highly dependent on how elastic the deployment is: the more frequent the changes are, the lower the max-connection-age should be.

PR #3447, opened 2021-12-14
Let build be customized by env vars for OSX

### LINKED ISSUE #3445: Build on a Mac
**Describe the bug**

Jaeger container builds do not work on a Mac (using Docker desktop).  Or, if they do, the technique is not documented.

`make create-baseimg` yields

```
error: failed to solve: failed to do request: Head "http://localhost:5000/v2/baseimg_alpine/blobs/sha256:be307f383ecc62b27a29b599c3fc9d3129693a798e7fcce614f09174cfe2d354": dial tcp 127.0.0.1:5000: connect: connection refused
```

**To Reproduce**

See above.

 - OS: Mac OSX
 - Jaeger version: Main/Master, 1.29.0 I think
 - Deployment: none
 - Docker Desktop: 4.2.0

**Additional context**

On the Mac, anything about the host should be `docker.for.mac.localhost` instead of `localhost`!  I couldn't figure out how to make that happen, though.

esnible:

Resolves #3445

With this change, I can build on a Mac.  The biggest problem is that the Mac doesn't let the containers see the real _localhost_, but instead puts the Mac's localhost at _host.docker.internal_.

First, I must tell Docker to let host.docker.internal be insecure:

```bash
cat <<EOF | > /tmp/buildkitd.toml
[registry."host.docker.internal:5000"]
  http = true
EOF

docker buildx create --use --config /tmp/buildkitd.toml
```

I then build using

make build-binaries-linux
PLATFORMS=linux/amd64 DOCKER_REPO=host.docker.internal:5000 make docker-images-jaeger-backend

(Note that I am confident about every change except for the `--load`.  I am not sure what that is needed!  Without it, the build complained that "No output specified for docker-container driver. Build result will only remain in the build cache. To push result image into registry use --push or to load image into docker use --load ")

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -309,7 +309,6 @@ docker-images-jaeger-backend docker-images-jaeger-backend-debug: create-baseimg 
 			--build-arg base_image=$(BASE_IMAGE) \
 			--build-arg debug_image=$(DEBUG_IMAGE) \
 			--build-arg TARGETARCH=$(GOARCH) \
-			--load \
 			cmd/$$component ; \
 		echo "Finished building $$component ==============" ; \
 	done
--- a/Makefile
+++ b/Makefile
@@ -309,6 +309,7 @@ docker-images-jaeger-backend docker-images-jaeger-backend-debug: create-baseimg 
 			--build-arg base_image=$(BASE_IMAGE) \
 			--build-arg debug_image=$(DEBUG_IMAGE) \
 			--build-arg TARGETARCH=$(GOARCH) \
+			--load \
 			cmd/$$component ; \
 		echo "Finished building $$component ==============" ; \
 	done
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -3,9 +3,10 @@ ROOT_IMAGE ?= alpine:3.14
 CERT_IMAGE := $(ROOT_IMAGE)
 GOLANG_IMAGE := golang:1.17-alpine
 
-BASE_IMAGE := localhost:5000/baseimg_alpine:latest
-DEBUG_IMAGE := localhost:5000/debugimg_alpine:latest
-PLATFORMS := linux/amd64,linux/s390x,linux/ppc64le,linux/arm64
+DOCKER_REPO ?= localhost:5000
+BASE_IMAGE ?= $(DOCKER_REPO)/baseimg_alpine:latest
+DEBUG_IMAGE ?= $(DOCKER_REPO)/debugimg_alpine:latest
+PLATFORMS ?= linux/amd64,linux/s390x,linux/ppc64le,linux/arm64
 
 create-baseimg-debugimg: create-baseimg create-debugimg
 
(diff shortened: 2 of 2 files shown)


docker/Makefile
```diff
@@ -3,9 +3,10 @@ ROOT_IMAGE ?= alpine:3.14
 CERT_IMAGE := $(ROOT_IMAGE)
 GOLANG_IMAGE := golang:1.17-alpine
 
-BASE_IMAGE := localhost:5000/baseimg_alpine:latest
-DEBUG_IMAGE := localhost:5000/debugimg_alpine:latest
-PLATFORMS := linux/amd64,linux/s390x,linux/ppc64le,linux/arm64
+DOCKER_REPO ?= localhost:5000
```
yurishkuro, line 6: "REPO" is misleading, I would call this DOCKER_REGISTRY

Makefile
```diff
@@ -309,6 +309,7 @@ docker-images-jaeger-backend docker-images-jaeger-backend-debug: create-baseimg
 			--build-arg base_image=$(BASE_IMAGE) \
 			--build-arg debug_image=$(DEBUG_IMAGE) \
 			--build-arg TARGETARCH=$(GOARCH) \
+			--load \
```
yurishkuro, line 312: according to the [docs][], this is a shorthand for `--output=type=docker`. I would use the latter, it makes the command self-explanatory.

[docs]: https://docs.docker.com/engine/reference/commandline/buildx_build/

yurishkuro: also, I don't see `--load` it my `docker help build`, but I do see `--output` flag, so it's probably better for compatibility too.

yurishkuro: Actually both `--load` and `--output=type=docker` fail for my version of Docker. When you say "it complains", does it fail or just print a warning? The docs say that by default the build results are "exported to `docker images`", which sounds like what we want because we then publish them explicitly.

esnible: I will make the changes requested here.
I have tried with both `docker version` client and server 20.10.10 and .11.

The warning I saw was 

```
WARN[0000] No output specified for docker-container driver. Build result will only remain in the build cache. To push result image into registry use --push or to load image into docker use --load 
```

This warning did not trouble me.  However, after a successful build, `docker images` didn't show any new images!  On the advice of the warning I added `--load` and repeated the _make_, after which I saw four newly built images.

I will dig into this tomorrow, perhaps I have misconfigured something.

esnible:
Thanks for your patience everyone.

I removed ~/.docker, reinstalled Docker Desktop for Mac, and rebooted my laptop.

After those environment changes everything works without `--load`.
After those environment changes I find I don't need `DOCKER_REGISTRY=host.docker.internal:5000` any more.  Something in my old config needed it; on the fresh Docker Desktop install it works fine without it.  Thus, I can't test overriding DOCKER_REGISTRY.

Thus, the only value of this PR is the ability to speed builds by doing `PLATFORMS=linux/amd64 make docker-images-jaeger-backend`.  Would it make sense to
- close this PR
- revise this PR just for `PLATFORMS`
- keep `DOCKER_REGISTRY` exposed for users whose configuration requires it
?

yurishkuro (approved):

PR #3485, opened 2022-01-19
Log the remote IP:port for query API users

esnible:
Experimental code to log the remote IP address for the query API.

### DIFF
--- a/cmd/query/app/server.go
+++ b/cmd/query/app/server.go
@@ -29,6 +29,7 @@ import (
 	"go.uber.org/zap/zapcore"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/stats"
 
 	"github.com/jaegertracing/jaeger/cmd/query/app/apiv3"
 	"github.com/jaegertracing/jaeger/cmd/query/app/querysvc"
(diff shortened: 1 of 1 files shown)


yurishkuro (commented):
in GDPR/ePD an IP address is considered sensitive user data and logging it should be avoided.

esnible:
@yurishkuro I agree that logging this data is undesirable.

What I noticed is that after https://github.com/jaegertracing/jaeger/pull/2827 I see messages in the jaeger-query log that look like 

```
{
  "level": "warn",
  "ts": 1642500323.338672,
  "caller": "channelz/logging.go:62",
  "msg": "[core]grpc: Server.Serve failed to create ServerTransport: connection error: desc = \"transport: http2Server.HandleStreams received bogus greeting from client: \\\"\\\\x16\\\\x03\\\\x01\\\\x02\\\\x00\\\\x01\\\\x00\\\\x01\\\\xfc\\\\x03\\\\x03;\\\\xcd\\\\xcc\\\\xe0K\\\\xdbt\\\\x8aO\\\\xd9C2\\\\xb4\\\"\"",
  "system": "grpc",
  "grpc_log": true
}
```

The truth is I really only want to get the remote IP in the case where I am about to generate a warning.

The two ways I can think of to determine the cause of the warning is to run `wireshark` on live data or to log the IPs.  The error message itself comes from _google.golang.org/grpc/server.go_:890 and _grpc/internal/transport/http2_server.go_:302 and I couldn't figure out any other way to get a message out of those libraries.

PR #3506, opened 2022-02-01
fix(integration-tests): allow multiple polling attempts

johanneswuerbach:

Slow backends (in my case a grpc plugin) require multiple attempts until `FindTraces` returns a result. While `waitForCondition` correctly waits for the result, `tracesMatch` has [already affected the test result](https://github.com/jaegertracing/jaeger/blob/7872d1b07439c3f2d316065b1fd53e885b26a66f/plugin/storage/integration/integration_test.go#L336) and marked the test as failed, if > 1 try are required.

Similar to other tests, the check inside `waitForCondition` is only returning a boolean result to break the loop, but the actual comparison is moved outside the loop.


yurishkuro (approved):
thanks!

Issue #3514, opened 2022-02-04
Jaeger binaries were compiled with an older Go version which has security vulnerabilities


apm-opentt:
**Describe the bug**
Our company's internal security scan tool has found security vulnerabilities in the latest Jaeger binaries.  May I request Jaeger team to uplift the Go compiler version on your build machines to the latest when building the next Jaeger release?  
Here is the vulnerabilities detected:
https://nvd.nist.gov/vuln/detail/CVE-2021-41772

**To Reproduce**
1.  Use internal scanning tool to scan Go binaries
2.
3.

**Expected behavior**
No security vulnerabilities.

**Screenshots**

**Version (please complete the following information):**
 - OS: [Linux]
 - Jaeger version: [1.30.0]
 - Deployment: [Kubernetes]

**What troubleshooting steps did you try?**

**Additional context**

pavolloffay:
Jaeger 1.31. is going out soon https://github.com/jaegertracing/jaeger/blob/main/RELEASE.md#release-managers. We will make sure it is compiled against the newest golang version.

yurishkuro:
v1.31.0 release in progress using Go 1.17.6

yurishkuro:
```
$ docker run --rm --name jaeger jaegertracing/all-in-one-debug:1.31

$ docker exec -it e0561c06f0e9 sh
/go # go version
go version go1.17.6 linux/amd64
```

PR #3516, opened 2022-02-04
Bump Go compiler in CI to 1.17.6

yurishkuro:
Related to #3514

### DIFF
--- a/.github/workflows/ci-all-in-one-build.yml
+++ b/.github/workflows/ci-all-in-one-build.yml
@@ -21,7 +21,7 @@ jobs:
 
     - uses: actions/setup-go@v2
       with:
-        go-version: ^1.17
+        go-version: ^1.17.6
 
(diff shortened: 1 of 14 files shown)


.github/workflows/ci-all-in-one-build.yml
```diff
@@ -21,7 +21,7 @@ jobs:
 
     - uses: actions/setup-go@v2
       with:
-        go-version: ^1.17
+        go-version: ^1.17.6
```
pavolloffay, line 24: What version does the 1.17 pick up? I would assume it is the latest one?

yurishkuro: It probably does, but why risk it. 1.18 is supposed to come out in Feb iiuc, so we'll switch to that and won't have the patch no frozen.

tianruyun: good

Issue #3523, opened 2022-02-09
query dependency from es return empty. Does 'IndexDateLayout' missed when init es.dependencystore ?


galric:
**Describe the bug**
system architecture and deep dependency graph return empty result while data exists in ES. 
does the function 'createDependencyReader' in 'https://github.com/jaegertracing/jaeger/blob/main/plugin/storage/es/factory.go' forget to init IndexDateLayout ?

current:

```
	reader := esDepStore.NewDependencyStore(esDepStore.DependencyStoreParams{
		Client:              client,
		Logger:              logger,
		IndexPrefix:         cfg.GetIndexPrefix(),
		MaxDocCount:         cfg.GetMaxDocCount(),
		UseReadWriteAliases: cfg.GetUseReadWriteAliases(),
	})

```

may should be:

```
	reader := esDepStore.NewDependencyStore(esDepStore.DependencyStoreParams{
		Client:              client,
		Logger:              logger,
		IndexPrefix:         cfg.GetIndexPrefix(),
		MaxDocCount:         cfg.GetMaxDocCount(),
		UseReadWriteAliases: cfg.GetUseReadWriteAliases(),
		IndexDateLayout:     cfg.GetIndexDateLayoutDependencies(),
	})

```

add this

```

IndexDateLayout:     cfg.GetIndexDateLayoutDependencies(),

```

to make sure 
```
func indexWithDate(indexNamePrefix, indexDateLayout string, date time.Time) string {
	return indexNamePrefix + date.UTC().Format(indexDateLayout)
}
```
from https://github.com/jaegertracing/jaeger/blob/main/plugin/storage/es/dependencystore/storage.go 
not alway return a empty string 

which effect DependencyStore.GetDependencies can only search 'jaeger-dependencies-' not 'jaeger-dependencies-2022-02-08'

galric:
i test the diff in my local environment, and it seems works

mindw:
@galric got hit by this also - with the same conclusion.

jkowall:
Seems to be fixed per https://github.com/jaegertracing/jaeger/pull/3524

PR #3524, opened 2022-02-12
fix indexDateLayout for elasticsearch dependencies #3523

ilyamor:


NimrodAvni78 (approved):

frzifus (commented):
thanks @ilyamor, could you please [sign off your work](https://github.com/jaegertracing/jaeger/blob/main/CONTRIBUTING_GUIDELINES.md#certificate-of-origin---sign-your-work)

plugin/storage/es/factory.go
```diff
@@ -218,6 +218,7 @@ func createDependencyReader(
 		Client:              client,
 		Logger:              logger,
 		IndexPrefix:         cfg.GetIndexPrefix(),
+		IndexDateLayout:     cfg.GetIndexDateLayoutDependencies(),
```
albertteoh, line 221: Thanks for fixing this, I wonder if we can update the unit tests to prevent this from happening again. Maybe via a type assertion of the returned `dependencystore.Reader`?

ilyamor: not sure I find it useful, looks like a dummy test

albertteoh (approved):

PR #3564, opened 2022-03-05
Add support for ciphersuites in tls config

Ashmita152:
Signed-off-by: Ashmita <ashmita.bohara@logz.io>

Part of  https://github.com/jaegertracing/jaeger/issues/3019

In golang's tls package, there are two kinds of ciphers: secure and insecure, we need to support both hence we are creating a map of all ciphers (mapping cipherName to cipherID) 

I will raise min tls version support as separate pr.

### DIFF
--- a/pkg/config/tlscfg/ciphersuites.go
+++ b/pkg/config/tlscfg/ciphersuites.go
@@ -1,0 +1,64 @@
+// Copyright (c) 2022 The Jaeger Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tlscfg
+
+import (
+	"crypto/tls"
+	"fmt"
+	"strings"
+)
+
+var (
+	// https://golang.org/pkg/crypto/tls/#pkg-constants
+	ciphers         = map[string]uint16{}
+	insecureCiphers = map[string]uint16{}
+)
+
+func init() {
+	for _, suite := range tls.CipherSuites() {
+		ciphers[suite.Name] = suite.ID
+	}
+	for _, suite := range tls.InsecureCipherSuites() {
+		insecureCiphers[suite.Name] = suite.ID
+	}
+}
+
+func allCiphers() map[string]uint16 {
+	acceptedCiphers := make(map[string]uint16, len(ciphers))
+	for k, v := range ciphers {
+		acceptedCiphers[k] = v
+	}
+	for k, v := range insecureCiphers {
+		acceptedCiphers[k] = v
+	}
+	return acceptedCiphers
+}
+
+// TLSCipherSuites returns a list of cipher suite IDs from the cipher suite names passed.
+func TLSCipherSuites(cipherNames string) ([]uint16, error) {
+	if cipherNames == "" {
+		return nil, nil
+	}
+	ciphersIntSlice := make([]uint16, 0)
+	possibleCiphers := allCiphers()
+	for _, cipher := range strings.Split(cipherNames, ",") {
+		intValue, ok := possibleCiphers[cipher]
+		if !ok {
+			return nil, fmt.Errorf("cipher suite %s not supported or doesn't exist", cipher)
+		}
+		ciphersIntSlice = append(ciphersIntSlice, intValue)
+	}
+	return ciphersIntSlice, nil
+}
--- a/pkg/config/tlscfg/ciphersuites.go
+++ b/pkg/config/tlscfg/ciphersuites.go
@@ -17,48 +17,29 @@ package tlscfg
 import (
 	"crypto/tls"
 	"fmt"
-	"strings"
 )
-
-var (
-	// https://golang.org/pkg/crypto/tls/#pkg-constants
-	ciphers         = map[string]uint16{}
-	insecureCiphers = map[string]uint16{}
-)
-
-func init() {
-	for _, suite := range tls.CipherSuites() {
-		ciphers[suite.Name] = suite.ID
-	}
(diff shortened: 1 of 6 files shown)


pkg/config/tlscfg/ciphersuites.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// TLSCipherSuites returns a list of cipher suite IDs from the cipher suite names passed.
+func TLSCipherSuites(cipherNames string) ([]uint16, error) {
+	if cipherNames == "" {
+		return nil, nil
+	}
+	ciphersIntSlice := make([]uint16, 0)
+	possibleCiphers := allCiphers()
+	for _, cipher := range strings.Split(cipherNames, ",") {
```
yurishkuro, line 56: Typically we avoid doing string splits all over the place. This is business function and should take a proper array of strings. Splitting should happen closer to the actual configuration code, ie when reading the combined string from viper (it may already support list-like arguments)

pkg/config/tlscfg/ciphersuites.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+	return acceptedCiphers
+}
+
+// TLSCipherSuites returns a list of cipher suite IDs from the cipher suite names passed.
+func TLSCipherSuites(cipherNames string) ([]uint16, error) {
+	if cipherNames == "" {
+		return nil, nil
+	}
+	ciphersIntSlice := make([]uint16, 0)
```
yurishkuro, line 54:
```suggestion
	var ciphersIDs []uint16
```

yurishkuro: This way when no values are passed the func would return nil

pkg/config/tlscfg/options.go
```diff
@@ -41,17 +42,24 @@ var systemCertPool = x509.SystemCertPool // to allow overriding in unit test
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get cipher suite ids from cipher suite names: %w", err)
+	}
+
 	// #nosec G402
 	tlsCfg := &tls.Config{
 		RootCAs:            certPool,
 		ServerName:         p.ServerName,
 		InsecureSkipVerify: p.SkipHostVerify,
+		CipherSuites:       cipherSuiteIds,
```
yurishkuro, line 60: Default value must be nil (see comment in the function)

pkg/config/tlscfg/ciphersuites.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	"crypto/tls"
+	"fmt"
+	"strings"
+)
+
+var (
+	// https://golang.org/pkg/crypto/tls/#pkg-constants
+	ciphers         = map[string]uint16{}
+	insecureCiphers = map[string]uint16{}
```
yurishkuro, line 26: Do we need these? The ticket was not specific which ciphers people needed, I would rather not support those with known security issues. If we do support them I would maybe introduce another cli flag for them, like insecure ciphers.

pkg/config/tlscfg/ciphersuites.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+func init() {
```
yurishkuro, line 29: There's no need for pkg init, this code is not perf sensitive. You can do this directly in allCiphers()

pkg/config/tlscfg/ciphersuites.go
```diff
@@ -0,0 +1,64 @@
...
+	for _, suite := range tls.CipherSuites() {
+		ciphers[suite.Name] = suite.ID
+	}
+	for _, suite := range tls.InsecureCipherSuites() {
+		insecureCiphers[suite.Name] = suite.ID
+	}
+}
+
+func allCiphers() map[string]uint16 {
+	acceptedCiphers := make(map[string]uint16, len(ciphers))
```
yurishkuro, line 39:
```suggestion
	acceptedCiphers := make(map[string]uint16)
```

pkg/config/tlscfg/ciphersuites.go
```diff
@@ -0,0 +1,64 @@
...
+		acceptedCiphers[k] = v
+	}
+	for k, v := range insecureCiphers {
+		acceptedCiphers[k] = v
+	}
+	return acceptedCiphers
+}
+
+// TLSCipherSuites returns a list of cipher suite IDs from the cipher suite names passed.
+func TLSCipherSuites(cipherNames string) ([]uint16, error) {
```
yurishkuro, line 50:
```suggestion
func CipherSuiteNamesToIDs(cipherNames []string) ([]uint16, error) {
```

TLS prefix is redundant in this pkg

Ashmita152:
👋 Hello Yuri

Thank you for the feedbacks. I tried to resolve them. It is ready for review again.

pkg/config/tlscfg/ciphersuites.go
```diff
@@ -0,0 +1,45 @@
...
+	acceptedCiphers := make(map[string]uint16)
+	for _, suite := range tls.CipherSuites() {
+		acceptedCiphers[suite.Name] = suite.ID
+	}
+	return acceptedCiphers
+}
+
+// CipherSuiteNamesToIDs returns a list of cipher suite IDs from the cipher suite names passed.
+func CipherSuiteNamesToIDs(cipherNames []string) ([]uint16, error) {
+	if len(cipherNames) == 0 {
```
yurishkuro, line 32: This check is not necessary, same will be achieved by the loop immediately exiting

pkg/config/tlscfg/flags.go
```diff
@@ -57,6 +58,7 @@ func (c ServerFlagsConfig) AddFlags(flags *flag.FlagSet) {
 	flags.String(c.Prefix+tlsCert, "", "Path to a TLS Certificate file, used to identify this server to clients")
 	flags.String(c.Prefix+tlsKey, "", "Path to a TLS Private Key file, used to identify this server to clients")
 	flags.String(c.Prefix+tlsClientCA, "", "Path to a TLS CA (Certification Authority) file used to verify certificates presented by clients (if unset, all clients are permitted)")
+	flags.String(c.Prefix+tlsCipherSuites, "", "Space-separated list of cipher suites for the server, values are from tls package constants (https://golang.org/pkg/crypto/tls/#pkg-constants).")
```
yurishkuro, line 61: I don't think we have any flags that are space separated, we always use commas, because commas do not need require quotes on command line.

yurishkuro (approved):
thanks!

Issue #3566, opened 2022-03-06
Support specifying minimum TLS version in TLS connection


sirzooro:
## Requirement - what kind of business use case are you trying to solve?

There are standards which require use of some minimum TLS version (e.g. TLS 1.2+) and selected cipher list. Issue https://github.com/jaegertracing/jaeger/issues/3019 has been closed few hours ago, so cipher list can be configured now. However minimum TLS version still cannot be configured.

## Problem - what in Jaeger blocks you from solving the requirement?

There is no option to specify minimum TLS version.

## Proposal - what do you suggest to solve the problem or improve the existing situation?

Add option to specify minimum TLS version.

PR #3567, opened 2022-03-06
Add minimum and maximum tls version support in tls config

Ashmita152:

- Closes https://github.com/jaegertracing/jaeger/issues/3566

- Add support for minimum tls version

### DIFF
--- a/pkg/config/tlscfg/ciphersuites.go
+++ b/pkg/config/tlscfg/ciphersuites.go
@@ -21,10 +21,10 @@ import (
 
 // https://pkg.go.dev/crypto/tls#pkg-constants
 var versions = map[string]uint16{
-	"VersionTLS10": tls.VersionTLS10,
-	"VersionTLS11": tls.VersionTLS11,
-	"VersionTLS12": tls.VersionTLS12,
-	"VersionTLS13": tls.VersionTLS13,
+	"1.0": tls.VersionTLS10,
+	"1.1": tls.VersionTLS11,
+	"1.2": tls.VersionTLS12,
+	"1.3": tls.VersionTLS13,
 }
 
 func allCiphers() map[string]uint16 {
@@ -51,16 +51,8 @@ func CipherSuiteNamesToIDs(cipherNames []string) ([]uint16, error) {
 
 // VersionNameToID returns the version ID from version name
 func VersionNameToID(versionName string) (uint16, error) {
-	if len(versionName) == 0 {
-		return DefaultVersion(), nil
-	}
 	if version, ok := versions[versionName]; ok {
 		return version, nil
 	}
 	return 0, fmt.Errorf("unknown tls version %q", versionName)
-}
-
-func DefaultVersion() uint16 {
-	// TLS1.2 is considered the safest default
-	return tls.VersionTLS12
 }
--- a/pkg/config/tlscfg/ciphersuites.go
+++ b/pkg/config/tlscfg/ciphersuites.go
@@ -19,6 +19,14 @@ import (
 	"fmt"
 )
 
+// https://pkg.go.dev/crypto/tls#pkg-constants
+var versions = map[string]uint16{
+	"VersionTLS10": tls.VersionTLS10,
+	"VersionTLS11": tls.VersionTLS11,
+	"VersionTLS12": tls.VersionTLS12,
+	"VersionTLS13": tls.VersionTLS13,
+}
+
 func allCiphers() map[string]uint16 {
 	acceptedCiphers := make(map[string]uint16)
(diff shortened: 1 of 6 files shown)


pkg/config/tlscfg/ciphersuites.go
```diff
@@ -19,6 +19,14 @@ import (
 	"fmt"
 )
 
+// https://pkg.go.dev/crypto/tls#pkg-constants
+var versions = map[string]uint16{
+	"VersionTLS10": tls.VersionTLS10,
+	"VersionTLS11": tls.VersionTLS11,
+	"VersionTLS12": tls.VersionTLS12,
+	"VersionTLS13": tls.VersionTLS13,
```
yurishkuro, line 27: we didn't have much choice with cipher names, but here we have full control over names, and "VersionTLS13" seems like unnecessary complicated naming. We can simply use strings like "1.1", ..., "1.3".

pkg/config/tlscfg/ciphersuites.go
```diff
@@ -40,3 +48,19 @@ func CipherSuiteNamesToIDs(cipherNames []string) ([]uint16, error) {
...
+		return DefaultVersion(), nil
+	}
+	if version, ok := versions[versionName]; ok {
+		return version, nil
+	}
+	return 0, fmt.Errorf("unknown tls version %q", versionName)
+}
+
+func DefaultVersion() uint16 {
+	// TLS1.2 is considered the safest default
```
yurishkuro, line 64: tls.Config says default is 1.0. I don't think we need to invent our own default - if the user did not set the value then we don't pass anything to tls.Config

pkg/config/tlscfg/flags.go
```diff
@@ -60,6 +61,7 @@ func (c ServerFlagsConfig) AddFlags(flags *flag.FlagSet) {
 	flags.String(c.Prefix+tlsKey, "", "Path to a TLS Private Key file, used to identify this server to clients")
 	flags.String(c.Prefix+tlsClientCA, "", "Path to a TLS CA (Certification Authority) file used to verify certificates presented by clients (if unset, all clients are permitted)")
 	flags.String(c.Prefix+tlsCipherSuites, "", "Comma-separated list of cipher suites for the server, values are from tls package constants (https://golang.org/pkg/crypto/tls/#pkg-constants).")
+	flags.String(c.Prefix+tlsMinVersion, "", "Minimum TLS version supported (Possible values: VersionTLS10, VersionTLS11, VersionTLS12, VersionTLS13)")
```
yurishkuro, line 64: Go supports both min and max version, why not implement support for both?

pkg/config/tlscfg/ciphersuites.go
```diff
@@ -40,3 +48,19 @@ func CipherSuiteNamesToIDs(cipherNames []string) ([]uint16, error) {
 	}
 	return ciphersIDs, nil
 }
+
+// VersionNameToID returns the version ID from version name
+func VersionNameToID(versionName string) (uint16, error) {
+	if len(versionName) == 0 {
+		return DefaultVersion(), nil
```
yurishkuro, line 55: default is better handled in the caller, this is purely a conversion function

pkg/config/tlscfg/options.go
```diff
@@ -52,12 +53,18 @@ func (p *Options) Config(logger *zap.Logger) (*tls.Config, error) {
 		return nil, fmt.Errorf("failed to get cipher suite ids from cipher suite names: %w", err)
 	}
 
+	minVersionId, err := VersionNameToID(p.MinVersion)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get minimum tls version: %w", err)
+	}
+
```
yurishkuro, line 60:
```suggestion
	var minVersionId uint16
	var err error
	if p.MinVersion != "" {
	    minVersionId, err = VersionNameToID(p.MinVersion)
	    if err != nil {
		    return nil, fmt.Errorf("failed to get minimum tls version: %w", err)
	    }
	}
```
i.e. do not change the default in Go struct if user did not provide any value

pkg/config/tlscfg/options.go
```diff
@@ -42,6 +44,9 @@ var systemCertPool = x509.SystemCertPool // to allow overriding in unit test
 
 // Config loads TLS certificates and returns a TLS Config.
 func (p *Options) Config(logger *zap.Logger) (*tls.Config, error) {
+	var err error
```
yurishkuro, line 47: redundant since it's redefined in L50

yurishkuro (approved):

yurishkuro (approved):

yurishkuro:
👏

Issue #3563, opened 2022-03-04
Favicon returns 500


forestsword:
**Describe the bug**

Looks like after the change https://github.com/jaegertracing/jaeger/pull/3399 the favicon can no longer be served. It's now an embedded gzipped file. All the versions since 1.28 return a 500, _seeker can't seek_ error.

It can be fixed by simply removing `favoriteIcon` from `StaticRootFiles` https://github.com/jaegertracing/jaeger/blob/e36f9fec8eb4a3b9a808aa76d248094fd7f657be/cmd/query/app/static_handler.go#L41

**To Reproduce**
1. Run the all-in-one after 1.28 (also affects query)
2. Open the UI in a browser

**Expected behavior**
Return the favicon with a 200 result.

**Screenshots**

![2022-03-04_13-35](https://user-images.githubusercontent.com/31450008/156764671-2994c4db-370e-4b86-9ae1-754e9984c911.png)

**Version (please complete the following information):**

All versions after 1.28

yurishkuro:
cc @Ashmita152

yurishkuro:
curiously enough, if I run all-in-one from source, favicon is returned normally, but when I run the docker command it returns 500.

PR #3569, opened 2022-03-07
Fix favicon returning 500 inside container

Ashmita152:

- Resolves https://github.com/jaegertracing/jaeger/issues/3563

- net/http depends on mime-type to set the Content-Type for a http response (https://github.com/golang/go/blob/master/src/net/http/fs.go#L173-L176)
- mime-type is calculated using https://pkg.go.dev/mime#TypeByExtension (https://github.com/golang/go/blob/master/src/net/http/fs.go#L235)
- In alpine, there is no default installation of mime-type library hence net/http was setting application/text for favicon.ico which was leading to 500. Adding mailcap package will install /etc/mime.types (https://pkgs.alpinelinux.org/contents?branch=edge&name=mailcap&arch=x86_64&repo=main) and fix the issue by setting the right content-type.

cc @yurishkuro

### DIFF
--- a/cmd/all-in-one/all_in_one_test.go
+++ b/cmd/all-in-one/all_in_one_test.go
@@ -62,6 +62,10 @@ func TestAllInOne(t *testing.T) {
 	if err := healthCheck(); err != nil {
 		t.Fatal(err)
 	}
+	// Check if the favicon icon is available
+	if err := faviconCheck(); err != nil {
+		t.Fatal(err)
+	}
 	createTrace(t)
 	getAPITrace(t)
 	getSamplingStrategy(t)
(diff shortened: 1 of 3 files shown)


docker/base/Dockerfile
```diff
@@ -2,7 +2,7 @@ ARG cert_image
 ARG root_image
 
 FROM $cert_image AS cert
-RUN apk add --update --no-cache ca-certificates
+RUN apk add --update --no-cache ca-certificates mailcap
 
 FROM $root_image
 COPY --from=cert /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
```
yurishkuro, line 8: don't we need to add a COPY command here to copy the mime.types files?

yurishkuro:
maybe we can add a test to after UI image is built to ensure that curl for favicon returns ok?

yurishkuro:
btw, nice find!

docker/base/Dockerfile
Ashmita152: ohh right. I missed it.

cmd/all-in-one/all_in_one_test.go
```diff
@@ -138,6 +142,18 @@ func healthCheck() error {
 	return fmt.Errorf("query service is not ready")
 }
 
+func faviconCheck() error {
```
yurishkuro, line 145: I don't think this one is testing the docker image

Ashmita152: Hi Yuri,
My understanding may be wrong but looking at this code https://github.com/jaegertracing/jaeger/blob/main/scripts/build-all-in-one-image.sh#L23-L32 it seems we first bring up the newly built all-in-one image and then run the integration test from this file against the container.

yurishkuro (approved):

PR #3573, opened 2022-03-09
Add SAMPLING_STORAGE_TYPE environment variable

joe-elliott:
- Adds support for a new environment variable `SAMPLING_STORAGE_TYPE`. This allows the user to independently configure
their `SPAN_STORAGE_TYPE` and `SAMPLING_STORAGE_TYPE`. e.g. the user could choose to store their spans in elasticsearch while putting their adaptive sampling data in cassandra. This also paves the way for simpler to operate sampling storage types like redis or memcached.

- Add support for `SAMPLING_STORAGE_TYPE` to plugin/storage/factory_config.go
- Add support to plugin/storage/factory.go to return the correct factory or fall back to previous behavior if unspecified.
- Add relevant testing.

### DIFF
--- a/plugin/storage/factory_config_test.go
+++ b/plugin/storage/factory_config_test.go
@@ -26,6 +26,7 @@ import (
 func clearEnv() {
 	os.Setenv(SpanStorageTypeEnvVar, "")
 	os.Setenv(DependencyStorageTypeEnvVar, "")
+	os.Setenv(SamplingStorageTypeEnvVar, "")
 }
 
 func TestFactoryConfigFromEnv(t *testing.T) {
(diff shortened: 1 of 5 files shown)


yurishkuro:
Just an observation - this will be subject to the existing limitation that a given kind of storage (e.g. cassandra) can only be configured once, i.e. it's not possible to have span and sampling storage use different instances of cassandra (or even different namespaces).

I don't think it should block this PR, but it may be worth thinking how that can be improved. For example, we have the mechanism for configuring different "named" storage, namely "archive", via flags like `--es.archive.***`. We could extend that mechanism to storage types, e.g. `--es.dependencies.***` or `--es.sampling.***`.

Of course, if the new solution prevents sharing the same instance as we can today that would be a regression.

yurishkuro (approved):

plugin/storage/factory_config_test.go
```diff
@@ -37,15 +38,18 @@ func TestFactoryConfigFromEnv(t *testing.T) {
 	assert.Equal(t, cassandraStorageType, f.SpanWriterTypes[0])
 	assert.Equal(t, cassandraStorageType, f.SpanReaderType)
 	assert.Equal(t, cassandraStorageType, f.DependenciesStorageType)
+	assert.Equal(t, "", f.SamplingStorageType)
```
albertteoh, line 41: nit: could use `assert.Empty(t, f.SamplingStorageType)`

albertteoh (approved):
LGTM. Does this mean we need to add support for `SAMPLING_STORAGE_TYPE` in jaeger-operator similar to https://github.com/jaegertracing/jaeger-operator/issues/1700?

joe-elliott:
> LGTM. Does this mean we need to add support for SAMPLING_STORAGE_TYPE in jaeger-operator similar to https://github.com/jaegertracing/jaeger-operator/issues/1700?

Oof, I don't know much about the operator. I had intended to PR the docs, but can't say I feel confident in my ability to make a well thought out operator change.

> this will be subject to the existing limitation that a given kind of storage (e.g. cassandra) can only be configured once, i.e. it's not possible to have span and sampling storage use different instances of cassandra (or even different namespaces).

This is a good point. I will include a note about this in my docs PR.

yurishkuro (approved):

PR #3610, opened 2022-04-01
Ilm disable template

rbizos:
clone of https://github.com/jaegertracing/jaeger/pull/3599 from another source branch to fix CI

## Problem

With ILM, index patter overridden without lifecycle policy set by collector if not explicitly disabled with --es.create-index-templates=false
Proposal


albertteoh:
Thanks!

rbizos:
My pleasure, thanks for the review

Issue #3578, opened 2022-03-10
Jaeger ingester panics on convertProcess function


juan-ramirez-sp:
**Describe the bug**
The Jaeger ingester is panicking when taking data from kafka and attempting to write it into elasticsearch.

I believe the following code is the reason for the panic. 

https://github.com/jaegertracing/jaeger/blob/4f55a70524b29ef87178b4deffbd5a99be5d017a/plugin/storage/es/spanstore/dbmodel/from_domain.go#L123

There seems to be no safeguards for a null process or null tags?

This is the following stack trace 

```
{"level":"info","ts":1646779105.5014815,"caller":"consumer/processor_factory.go:65","msg":"Creating new processors","partition":31}
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0xd94dc6]

goroutine 84516 [running]:
github.com/jaegertracing/jaeger/plugin/storage/es/spanstore/dbmodel.FromDomain.convertProcess(...)
	github.com/jaegertracing/jaeger/plugin/storage/es/spanstore/dbmodel/from_domain.go:123
github.com/jaegertracing/jaeger/plugin/storage/es/spanstore/dbmodel.FromDomain.convertSpanEmbedProcess({0x0, 0xc0003c48a0, {0x1499618, 0xf23872}}, 0xc00c544690)
	github.com/jaegertracing/jaeger/plugin/storage/es/spanstore/dbmodel/from_domain.go:64 +0x126
github.com/jaegertracing/jaeger/plugin/storage/es/spanstore/dbmodel.FromDomain.FromDomainEmbedProcess(...)
	github.com/jaegertracing/jaeger/plugin/storage/es/spanstore/dbmodel/from_domain.go:43
github.com/jaegertracing/jaeger/plugin/storage/es/spanstore.(*SpanWriter).WriteSpan(0xc0005f8cc0, {0x0, 0x0}, 0xc00c544690)
	github.com/jaegertracing/jaeger/plugin/storage/es/spanstore/writer.go:152 +0x7a
github.com/jaegertracing/jaeger/cmd/ingester/app/processor.KafkaSpanProcessor.Process({{0x14ad480, 0x1dee2a8}, {0x14ad340, 0xc0005f8cc0}, {0x0, 0x0}}, {0x14af760, 0xc0111950e0})
[...]

PR #3631, opened 2022-04-15
Sanitize spans with null process or empty service name

yurishkuro:
Sanitize spans with empty service name or null process. This should protect the pipeline from malformed spans submitted externally, as reported in #3578. It **will not** protect against malformed spans written directly to Kafka - we could do that too, but it changes the expectations of the deployment that Kafka is a public API into ingester, which would be a new guarantee and will require ingester to run preprocessing logic similar to collector.

### DIFF
--- a/cmd/collector/app/sanitizer/sanitizer_test.go
+++ b/cmd/collector/app/sanitizer/sanitizer_test.go
@@ -22,6 +22,10 @@ import (
 	"github.com/jaegertracing/jaeger/model"
 )
 
+func TestNewStandardSanitizers(t *testing.T) {
+	NewStandardSanitizers()
+}
+
 func TestChainedSanitizer(t *testing.T) {
 	var s1 SanitizeSpan = func(span *model.Span) *model.Span {
 		span.Process = &model.Process{ServiceName: "s1"}
--- a/cmd/collector/app/sanitizer/zipkin/span_sanitizer_test.go
+++ b/cmd/collector/app/sanitizer/zipkin/span_sanitizer_test.go
@@ -28,6 +28,10 @@ var (
 	positiveDuration = int64(1)
 )
 
+func TestNewStandardSanitizers(t *testing.T) {
+	NewStandardSanitizers()
+}
+
 func TestChainedSanitizer(t *testing.T) {
 	sanitizer := NewChainedSanitizer(NewSpanDurationSanitizer())
 


yurishkuro:
@albertteoh can you review, please?

cmd/collector/app/sanitizer/empty_service_name_sanitizer.go
```diff
@@ -0,0 +1,41 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package sanitizer
+
+import (
+	"github.com/jaegertracing/jaeger/model"
+)
+
+var (
```
albertteoh, line 21: Why not `const`?

cmd/collector/app/span_processor.go
```diff
@@ -96,13 +96,18 @@ func newSpanProcessor(spanWriter spanstore.Writer, additional []ProcessSpan, opt
 	}
 	boundedQueue := queue.NewBoundedQueue(options.queueSize, droppedItemHandler)
 
+	sanitizers := append([]sanitizer.SanitizeSpan{}, sanitizer.StandardSanitizers...)
```
albertteoh, line 99: `sanitizer.StandardSanitizers` is a globally accessible slice that can be mutated. What do you think about unexporting `sanitizer.StandardSanitizers` and adding a `sanitizer.GetStandardSanitizers()` func?

Then (I don't think we need to append to an empty slice?):
```
sanitizers := sanitizer.GetStandardSanitizers()
```

albertteoh (commented):
Looks good overall, just a question around the standard sanitizers variable.

albertteoh:
@yurishkuro, apologies for the late review, I've been away for a couple of weeks.

scripts/es-integration-test.sh
```diff
@@ -68,7 +69,7 @@ wait_for_it() {
     ''%{http_code}''
   )
   local counter=0
-  while [[ "$(curl ${params[@]} ${url})" != "200" && ${counter} -le 30 ]]; do
+  while [[ "$(curl ${params[@]} ${url})" != "200" && ${counter} -le 60 ]]; do
```
yurishkuro, line 72: 1min (2*30) seems insufficient sometimes, we can wait longer if it takes OS longer to start

albertteoh (dismissed):
LGTM. Seems some lines might need to be covered by tests though.

PR #3633, opened 2022-04-17
Pin golandci-lint to hash pending v1.45.3 release

yurishkuro:
This re-enables 3 out of 4 default linters that were disabled due to incompatibility with Go 1.18.

### DIFF
--- a/cmd/query/app/apiv3/otlp_translator.go
+++ b/cmd/query/app/apiv3/otlp_translator.go
@@ -25,6 +25,7 @@ import (
 
 	"go.opentelemetry.io/collector/model/pdata"
 	semconv "go.opentelemetry.io/collector/model/semconv/v1.5.0"
+	"go.opentelemetry.io/collector/pdata/ptrace"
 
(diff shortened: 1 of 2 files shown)


cmd/query/app/apiv3/otlp_translator.go
```diff
@@ -376,7 +377,7 @@ func uint64ToTraceID(high, low uint64) []byte {
 // See: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#status
 func statusCodeFromHTTP(httpStatusCode int) pdata.StatusCode {
 	if httpStatusCode >= 100 && httpStatusCode < 399 {
-		return pdata.StatusCodeUnset
+		return ptrace.StatusCodeUnset
 	}
-	return pdata.StatusCodeError
+	return ptrace.StatusCodeError
```
yurishkuro, line 382: re-enabling staticheck caught this deprecation notice

PR #3639, opened 2022-04-24
fix: close grpc client when in close

vuuihc:
Signed-off-by: vuuihc <zshyou@gmail.com>

the grpc Client Close is not called when exit

Close the grpc Client when exit


yurishkuro (commented):
+1

can we add a test to validate that?

vuuihc:
I missed that it's Close in RunAndThen, I'll close this PR.

PR #3640, opened 2022-04-24
feature: add streamingSpanWriterPlugin

### LINKED ISSUE #3636: Add a client stream WriteSpan method in grpc storage plugin
Hi, thanks for this outstanding project. I have a performance improvement suggestion for the grpc storage plugin. 

When saving traces with [jaeger-clickhouse storage plugin](https://github.com/jaegertracing/jaeger-clickhouse), I found that the all-in-one process's  CPU utilization is relatively high, while the write throughput is not large enough. Through performance analysis, I found that grpc communication occupies a relatively large proportion of CPU, and this can be optimized.

In the grpc service definition of jaeger's storage grpc plugin, the `WriteSpan` method is an unary rpc method

Write to storage with stream rpc. Specifically, we may add a client stream method `WriteSpanStreamingly`

``` proto
service SpanWriterPlugin {
    // spanstore/Writer
    rpc WriteSpan(WriteSpanRequest) returns (WriteSpanResponse);
    rpc WriteSpanStreamingly(stream WriteSpanRequest) returns (WriteSpanResponse); // add this method
    rpc Close(CloseWriterRequest) returns (CloseWriterResponse);
}
```
In my test, using client stream RPC compared to using unary RPC, under the same throughput pressure, the CPU utilization decreased by 35%; under the same hardware and software parameter configuration,  the maximum throughput increased by 76%. Below is the detailed data.

#### environment
```
CPU=10 * (Intel(R) Xeon(R) Platinum 8255C CPU @ 2.50GHz)
MEM=40GB
COLLECTOR_QUEUE_SIZE=4096
COLLECTOR_NUM_WORKERS=102
"the same write pressure": call greeter grpc service with ghz, concurrency=100 and num=5000000, 11 spans per request.
```
#### data summary
|       |under the same write pressure CPU usage      |  Maximum throughput under the same hardware and software conditions (spans/sec) |
| ------|----- | ----------- |
| WriteSpan |    186%  | 25974   |
|WriteSpanStreamingly | 120% | 45775     |
| Diff  | ↓35%  |   ↑76% |

#### pprof  flame graph
unary
unary

stream
stream

If there is no problem, I am willing to open a PR to complete it.

vuuihc:
Signed-off-by: vuuihc <zshyou@gmail.com>

Resolves #3636 

add streamingSpanWriterPlugin to improve the write performance of grpc plugin.


plugin/storage/grpc/shared/streaming_writer_grpc_client.go
```diff
@@ -0,0 +1,72 @@
...
+	"google.golang.org/grpc/status"
+
+	"github.com/jaegertracing/jaeger/model"
+	"github.com/jaegertracing/jaeger/proto-gen/storage_v1"
+	"github.com/jaegertracing/jaeger/storage/spanstore"
+)
+
+type streamingWriterGRPCClient struct {
+	*grpcClient
+	streamingWriterClient storage_v1.StreamingSpanWriterPluginClient
```
yurishkuro, line 32:
```suggestion
	client storage_v1.StreamingSpanWriterPluginClient
```

plugin/storage/grpc/shared/streaming_writer_grpc_client.go
```diff
@@ -0,0 +1,72 @@
...
+	stream                storage_v1.StreamingSpanWriterPlugin_WriteSpanStreamClient
+}
+
+func NewStreamingWriterGPRCClient(c *grpc.ClientConn) *streamingWriterGRPCClient {
+	return &streamingWriterGRPCClient{
+		grpcClient:            NewGRPCClient(c),
```
yurishkuro, line 38: what is this for?

plugin/storage/grpc/config/config.go
```diff
@@ -122,6 +122,9 @@ func (c *Configuration) buildPlugin(logger *zap.Logger) (*ClientPluginServices,
 		HandshakeConfig: shared.Handshake,
 		VersionedPlugins: map[int]plugin.PluginSet{
 			1: shared.PluginMap,
+			2: map[string]plugin.Plugin{
```
yurishkuro, line 125: It's been a while since I looked into this, could you elaborate what `VersionedPlugins` does?

yurishkuro (commented):
Thanks for the PR.

I feel like I am missing something. It's not clear to me why we need a separate "streaming client" writer, could we not do it from the existing client? 

But more importantly, I think the overall approach we had taken when we extended the plugin contract with "archive" storage was to add Capabilities field that tells the client if the plugin implementation supports additional interfaces. So I am surprised that you don't have any changed to the Factory files.

Lastly, I would like to have integration tests to exercise the new functionality.

plugin/storage/grpc/config/config.go
vuuihc: VersionedPlugins is used in go-plugins to negotiate plugin protoversion between servers and clients. Go-plugin will find the latest version supported by both the server and the client, and start the plugin server. 

In this scenario, it is more reasonable to negotiate with Capabilities, I'll refactor the code.

plugin/storage/grpc/shared/streaming_writer_grpc_client.go
vuuihc: @yurishkuro Thanks for the feedback.

> I feel like I am missing something. It's not clear to me why we need a separate "streaming client" writer, could we not do it from the existing client?

We can do it in the existing client, I want to write the code more elegantly, if there is any suggestion please do not hesitate to teach in the next review.

> But more importantly, I think the overall approach we had taken when we extended the plugin contract with "archive" storage was to add Capabilities field that tells the client if the plugin implementation supports additional interfaces. So I am surprised that you don't have any changed to the Factory files.

ok, I'll change to using Capabilities to negotiate.

> Lastly, I would like to have integration tests to exercise the new functionality.

ok, I'll add it

plugin/storage/grpc/factory.go
```diff
@@ -38,9 +38,10 @@ type Factory struct {
 
 	builder config.PluginBuilder
 
-	store        shared.StoragePlugin
-	archiveStore shared.ArchiveStoragePlugin
-	capabilities shared.PluginCapabilities
+	store               shared.StoragePlugin
+	archiveStore        shared.ArchiveStoragePlugin
+	streamingSpanWriter shared.StreamingSpanWriterPlugin
+	capabilities        shared.PluginCapabilities
```
yurishkuro, line 44: I just realized something - we probably didn't need to introduce Capabilities in the code at all if we used gRPC reflection instead. But that ship has sailed.

plugin/storage/grpc/shared/streaming_writer.go
```diff
@@ -0,0 +1,59 @@
...
+
+// streamingSpanWriter wraps storage_v1.StreamingSpanWriterPluginClient into spanstore.Writer
+type streamingSpanWriter struct {
+	client storage_v1.StreamingSpanWriterPluginClient
+	stream storage_v1.StreamingSpanWriterPlugin_WriteSpanStreamClient
+}
+
+// WriteSpan write span into stream
+func (s *streamingSpanWriter) WriteSpan(ctx context.Context, span *model.Span) error {
+	if s.stream == nil {
```
yurishkuro, line 38: Is it possible that this method will be called concurrently? A general SpanWriter is expected to be concurrent because Collector uses the same storage from multiple workers. I don't think there's any synchronization point from top-level SpanWriter to this class.

examples/memstore-plugin/main.go
```diff
@@ -72,8 +72,9 @@ func main() {
 
 	memStorePlugin := grpcMemory.NewStoragePlugin(memory.NewStore(), memory.NewStore())
 	grpc.ServeWithGRPCServer(&shared.PluginServices{
-		Store:        memStorePlugin,
-		ArchiveStore: memStorePlugin,
+		Store:               memStorePlugin,
+		ArchiveStore:        memStorePlugin,
+		StreamingSpanWriter: memStorePlugin,
```
yurishkuro, line 77: iirc we're using this memstore-plugin for integration tests of the storage plugin framework, just like any other storage backend. However, in this implementation it looks like the streaming writes is the only write path that will be exercised, because you unconditionally switch to streaming writer if it's supported.

One way to address this is to run the tests another time with a different CLI flag such that in the first run `StreamingSpanWriter` field above will be left nil, and only populated if a special CLI flag is set.

Note that we already run gRPC test >1 times (`plugin/storage/integration/grpc_test.go`), to test both the hashicorp plugin and the regular remote storage.

yurishkuro (commented):
lgtm overall, I kicked off the tests

examples/memstore-plugin/main.go
vuuihc: Got it, I set streamingWriter to be off by default and add a plugin flag `grpc-storage-plugin.write-type==streaming` to turn it on. Then add a new test method in `grpc_test.go.`

plugin/storage/grpc/shared/streaming_writer.go
vuuihc: Yes, `stream.Send` can't be called concurrently too, let me modify it

plugin/storage/grpc/factory.go
vuuihc: yeah

vuuihc:
@yurishkuro Hi yuri, all the issues you commented have been resolved.

plugin/storage/grpc/options.go
```diff
@@ -29,10 +29,12 @@ const (
 	pluginBinary             = "grpc-storage-plugin.binary"
 	pluginConfigurationFile  = "grpc-storage-plugin.configuration-file"
 	pluginLogLevel           = "grpc-storage-plugin.log-level"
+	pluginWriterType         = "grpc-storage-plugin.writer-type"
```
yurishkuro, line 32: I think we had a mis-communication. I don't know if this flag is needed, because if the plugin implements streaming API I don't see why it shouldn't be used automatically, can you think of any reasons?  Even if there are reasons, the plugin author is in control of the Capabilities response and could implement their own configuration to disable streaming if there users want that.

When I was talking about a cmd line option I was only referring to the mock plugin we use in integration tests, where automatic upgrade to streaming API meant the regular API would never be tested.

Granted, this version achieves the same more generically, I am not too opposed to it, but it makes the streaming feature a bit more clunky to use imo. Perhaps we could design it so that the streaming upgrade still happens automatically, but the option allows to turn it off. For example, we could make the default value for this flag to be `auto` instead of `unary`.

plugin/storage/grpc/options.go
```diff
@@ -55,6 +57,7 @@ func (opt *Options) AddFlags(flagSet *flag.FlagSet) {
 	flagSet.String(pluginBinary, "", "The location of the plugin binary")
 	flagSet.String(pluginConfigurationFile, "", "A path pointing to the plugin's configuration file, made available to the plugin with the --config arg")
 	flagSet.String(pluginLogLevel, defaultPluginLogLevel, "Set the log level of the plugin's logger")
+	flagSet.String(pluginWriterType, defaultWriterType, "The plugin span writer's grpc rpc type")
```
yurishkuro, line 60:
```suggestion
	flagSet.String(pluginWriterType, defaultWriterType, "When the plugin supports streaming API for writes, this option controls which API will be used (streaming, unary, auto), with 'auto' causing an automatic upgrade to the  streaming version")
```

plugin/storage/integration/grpc_test.go
```diff
@@ -164,6 +164,31 @@ func TestGRPCStorage(t *testing.T) {
...
+	if configPath == "" {
+		t.Log("PLUGIN_CONFIG_PATH env var not set")
+	}
+
+	flags := []string{
+		"--grpc-storage-plugin.binary", binaryPath,
+		"--grpc-storage-plugin.log-level", "debug",
+		"--grpc-storage-plugin.writer-type", "streaming",
+		"--grpc-storage-plugin.configuration-file", configPath,
+	}
```
yurishkuro, line 183: this is largely repeating the code from the previous function; let's move into a helper `getPluginFlags`

yurishkuro: also, the actual values seem identical to the previous test, did you want to include the new option here? It can be appended after calling getPluginFlags()

plugin/storage/grpc/options.go
vuuihc: I also think this flag may not be necessary.
Now the plugin only accepts the `-config` flag, so I'm not sure how to use a command-line option to instruct the plugin whether to fill in the streamWriter field, maybe add a field to the memory plugin's config for the integration test as my latest commit, or do you think there is a more appropriate way?

plugin/storage/integration/grpc_test.go
vuuihc: done

plugin/storage/grpc/options.go
vuuihc: I removed this flag for simplicity.

pkg/memory/config/config.go
```diff
@@ -16,5 +16,6 @@ package config
 
 // Configuration describes the options to customize the storage behavior
 type Configuration struct {
-	MaxTraces int `yaml:"max-traces" mapstructure:"max_traces"`
+	MaxTraces             int  `yaml:"max-traces" mapstructure:"max_traces"`
+	EnableStreamingWriter bool `yaml:"enable-streaming-writer" mapstructure:"enable_streaming_writer"`
```
yurishkuro, line 20: I am not sure this is a good place for this flag. Memory storage by itself has no need for this flag, it only makes sense in the context of the memstore-plugin.

vuuihc: Got it, I deleted it here and kept it in the memory plugin

plugin/storage/grpc/shared/streaming_writer.go
```diff
@@ -0,0 +1,103 @@
...
+// WriteSpan write span into stream
+func (s *streamingSpanWriter) WriteSpan(ctx context.Context, span *model.Span) error {
+	stream, err := s.getStream(ctx)
+	if err != nil {
+		return fmt.Errorf("plugin getStream error: %w", err)
+	}
+	if err := stream.Send(&storage_v1.WriteSpanRequest{Span: span}); err != nil {
+		return fmt.Errorf("plugin Send error: %w", err)
+	}
+	s.putStream(stream)
```
yurishkuro, line 62: If I understand your code correctly, you pass a stream object through the channel over and over. That seems like a very roundabout way of synchronizing access to it (vs. just using a mutex). Also, it looks like it might open multiple streams (up to a 1000), e.g. when one is in use, the read from the channel will fall back to L87. Did you really want that?

vuuihc: Yes, this is because stream.Send in gRPC is not concurrency safe (refer to: https://github.com/grpc/grpc-go/blob/master/stream.go#L123), but it seems that we have no way to bind streams and the upper layer goroutine worker, so a stream pool is implemented in this way.
Theoretically, the size of the stream pool will not exceed `collectorNumWorkers` because get and put are executed synchronously.
Btw, in my test, compared to unary writeSpan
1. Sharing a stream through mutex, although the CPU utilization will decrease, the throughput will decrease 
2. Sharing streams through the stream pool, the throughput will increase while the CPU utilization decreases.

examples/memstore-plugin/main.go
```diff
@@ -71,10 +72,14 @@ func main() {
 	opentracing.SetGlobalTracer(tracer)
 
 	memStorePlugin := grpcMemory.NewStoragePlugin(memory.NewStore(), memory.NewStore())
-	grpc.ServeWithGRPCServer(&shared.PluginServices{
+	service := &shared.PluginServices{
 		Store:        memStorePlugin,
 		ArchiveStore: memStorePlugin,
-	}, func(options []googleGRPC.ServerOption) *googleGRPC.Server {
+	}
+	if v.GetBool("enable_streaming_writer") {
```
yurishkuro, line 79: wouldn't it be simpler to declare this flag next to L41 instead of going through the configuration file?

vuuihc: If using the same level of flag as `config`, it looks like we also need to modify `plugin/storage/grpc/config/config.go` (add code to parse this flag, then pass the flag in `buildPlugin`). This is similar to the previous `grpc-storage-plugin.writer-type` flag, but transfers the use of the flag to the plugin author.
If I understand correctly, we want to automatically upgrade the rpc method (don't want to explicitly add a flag to the user), but want to test both rpc methods in the integration test.
Now it seems that there are several options, which one do you prefer?
Option 1: Add a flag similar to `grpc-storage-plugin.writer-type` and let the plugin (plugin server) handle this flag.
Option 2: Add a flag like `grpc-storage-plugin.writer-type`, the default is auto. Handle this flag in jaeger (plugin client).
Option 3: Read the flag from config in the integration test, which is the current code.
Or do you have any other suggestions?

yurishkuro: Option 4 - add `flag.BoolPtr` in this file (`examples/memstore-plugin/main.go`) (instead of reading it through config file / viper) and use it in L79.

vuuihc: If I understand correctly, then we also need to parse and transparently transmit this flag to the plugin in [`plugin/storage/grpc/config/config.go:119` ](https://github.com/jaegertracing/jaeger/blob/main/plugin/storage/grpc/config/config.go#L119), because **the flag passed in [`plugin/storage/integration/grpc_test.go:152`](https://github.com/jaegertracing/jaeger/blob/main/plugin/storage/integration/grpc_test.go#L152) is for jaeger, not for then plugin**. Do you think this is okay?

yurishkuro: Oh, I see what you mean. Ok let's leave it with the config file then.

yurishkuro:
I will try to do a final review tomorrow

yurishkuro (approved):

PR #3654, opened 2022-05-04
Resolves #3403 : Adding FilterProcessTags to filter out which Span.Process.Tags are sent from the agent

### LINKED ISSUE #3403: Using ES_TAGS_AS_FIELDS_INCLUDE writes a giant blob of tags as well as the fields
**Describe the bug**
I'm not entirely sure from the documentation if this is intended, or a bug...

I'm trying to use `ES_TAGS_AS_FIELDS_INCLUDE` to decide which tags get written as fields to Elasticsearch.  This works quite nicely, however as well as writing my desired fields I also get a blob called `tags` which contains the raw data.

The purpose of me using `ES_TAGS_AS_FIELDS_INCLUDE` is to reduce the ingestion data for stuff we don't care about, so it's counter productive for me to also write a big blob such as this.

**To Reproduce**
1. Configure a collector with `ES_TAGS_AS_FIELDS_INCLUDE` as a subset of available tags
2. Observe what gets written to elasticsearch

**Expected behavior**
The `tags` field to not be written, or have a documented means of not writing it.  I appreciate i could drop it on the ES side, but at that point it's already been transferred (data egress costs) and also means i'm maintaining two lots of config for fields, in different places.

**Screenshots**
Screenshot 2021-11-22 at 09 25 18

 - Jaeger version: 1.27.0
 - Deployment: Kubernetes

Reading documentation.

**Additional context**

SamrathPalSingh:
Adding flag to filter Span.Process.Tags on the agent side. This flag can be used to specify which Span.Process.Tags get sent from the Jaeger agent. This flag is created as a way to do what ES_TAGS_AS_FIELDS_INCLUDE does but on the agent side so that only the certain tags are sent. It reduces the network cost of sending data which is not required on the collector side.

The current changes are for Span.Process.Tags, but it can be extended to Span.Tags as well.

### DIFF
--- a/cmd/agent/app/reporter/flags.go
+++ b/cmd/agent/app/reporter/flags.go
@@ -32,9 +32,6 @@ const (
 	GRPC Type = "grpc"
 
 	agentTags = "agent.tags"
-
-	//tags to be included for sending to collector
-	filterProcessTags = "filterProcessTags"
 )
 
 // Type defines type of reporter.
@@ -42,18 +39,13 @@ type Type string
 
 // Options holds generic reporter configuration.
 type Options struct {
-	ReporterType          Type
-	AgentTags             map[string]string
-	FilterProcessTags string
+	ReporterType Type
+	AgentTags    map[string]string
 }
 
 // AddFlags adds flags for Options.
 func AddFlags(flags *flag.FlagSet) {
 	flags.String(reporterType, string(GRPC), fmt.Sprintf("Reporter type to use e.g. %s", string(GRPC)))
-
-	//Comma seperated values of tags to be sent to collector
-	flags.String(filterProcessTags, string(""), fmt.Sprintf("Tags to be included in the blob sent to collector"))
-
 	if !setupcontext.IsAllInOne() {
 		flags.String(agentTags, "", "One or more tags to be added to the Process tags of all spans passing through this agent. Ex: key1=value1,key2=${envVar:defaultValue}")
 	}
@@ -62,9 +54,6 @@ func AddFlags(flags *flag.FlagSet) {
 // InitFromViper initializes Options with properties retrieved from Viper.
 func (b *Options) InitFromViper(v *viper.Viper, logger *zap.Logger) *Options {
 	b.ReporterType = Type(v.GetString(reporterType))
-
-	b.FilterProcessTags = v.GetString(filterProcessTags)
-
 	if !setupcontext.IsAllInOne() {
 		if len(v.GetString(agentTags)) > 0 {
 			b.AgentTags = flags.ParseJaegerTags(v.GetString(agentTags))
--- a/cmd/agent/app/reporter/flags.go
+++ b/cmd/agent/app/reporter/flags.go
@@ -32,6 +32,9 @@ const (
 	GRPC Type = "grpc"
 
 	agentTags = "agent.tags"
+
+	//tags to be included for sending to collector
+	filterProcessTags = "filterProcessTags"
 )
 
 // Type defines type of reporter.
@@ -39,13 +42,18 @@ type Type string
 
 // Options holds generic reporter configuration.
 type Options struct {
-	ReporterType Type
-	AgentTags    map[string]string
+	ReporterType          Type
+	AgentTags             map[string]string
+	FilterProcessTags string
 }
 
 // AddFlags adds flags for Options.
 func AddFlags(flags *flag.FlagSet) {
 	flags.String(reporterType, string(GRPC), fmt.Sprintf("Reporter type to use e.g. %s", string(GRPC)))
+
+	//Comma seperated values of tags to be sent to collector
+	flags.String(filterProcessTags, string(""), fmt.Sprintf("Tags to be included in the blob sent to collector"))
+
 	if !setupcontext.IsAllInOne() {
 		flags.String(agentTags, "", "One or more tags to be added to the Process tags of all spans passing through this agent. Ex: key1=value1,key2=${envVar:defaultValue}")
 	}
@@ -54,6 +62,9 @@ func AddFlags(flags *flag.FlagSet) {
 // InitFromViper initializes Options with properties retrieved from Viper.
 func (b *Options) InitFromViper(v *viper.Viper, logger *zap.Logger) *Options {
 	b.ReporterType = Type(v.GetString(reporterType))
+
+	b.FilterProcessTags = v.GetString(filterProcessTags)
+
 	if !setupcontext.IsAllInOne() {
 		if len(v.GetString(agentTags)) > 0 {
 			b.AgentTags = flags.ParseJaegerTags(v.GetString(agentTags))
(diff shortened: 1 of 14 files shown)


cmd/agent/app/reporter/grpc/reporter.go
```diff
@@ -101,3 +107,37 @@ func makeModelKeyValue(agentTags map[string]string) []model.KeyValue {
 
 	return tags
 }
+
+func makeSliceOfTags(includedTags string) []string {
+	if includedTags != "" {
+		tags := strings.Split(strings.Trim(includedTags, " "), ",")
```
yurishkuro, line 113: parsing of config parameters should be done close to config / flags code, not in the business type like Reporter

yurishkuro:
This does not resolve #3403 (I edited description), this is completely independent functionality.

cmd/agent/app/reporter/grpc/reporter.go
```diff
@@ -32,19 +33,21 @@ import (
...
 // Reporter reports data to collector over gRPC.
 type Reporter struct {
-	collector api_v2.CollectorServiceClient
-	agentTags []model.KeyValue
-	logger    *zap.Logger
-	sanitizer zipkin2.Sanitizer
+	collector                  api_v2.CollectorServiceClient
+	agentTags                  []model.KeyValue
+	logger                     *zap.Logger
+	filterProcessTagsSlice []string
```
yurishkuro, line 39: the Reporter should not care what kind of clean-up is performed on the spans. It can accept a span sanitizer, similar to zipkin sanitizer.

cmd/agent/app/reporter/grpc/reporter.go
```diff
@@ -101,3 +107,37 @@ func makeModelKeyValue(agentTags map[string]string) []model.KeyValue {
...
+	}
+}
+
+//remove the tags not present in filterProcessTags flag
+func pruneUnwantedTags(filterProcessTagsSlice []string, spans []*model.Span, process *model.Process) []*model.Span {
+	for _, span := range spans {
+		if span.Process != nil && span.Process.Tags != nil {
+			finalTags := make([]model.KeyValue, 0)
+			for _, tag := range span.Process.Tags {
+				if contains(filterProcessTagsSlice, tag.Key) {
```
yurishkuro, line 126: the sanitizer should convert (during initialization) the array of tag names into a `map[string]struct{}` for faster lookup

cmd/agent/app/reporter/flags.go
```diff
@@ -32,20 +32,28 @@ const (
 	GRPC Type = "grpc"
 
 	agentTags = "agent.tags"
+
+	//tags to be included for sending to collector
+	filterProcessTags = "filterProcessTags"
```
yurishkuro, line 37: "filter" is ambiguous. This could be called `agent.drop-process-tags` (later paired with `agent.drop-span-tags`).

yurishkuro (commented):
What is the use case that this is solving?

SamrathPalSingh:
> What is the use case that this is solving?

Basically, in issue #3403 the reporter is suggesting a way to stop the blob on the agent side instead of on the collector. With this implementation I intend to only send the tags which are required to be stored on the collector side. That way the blob still get's stored on the ES side but only with the Span.Process.Tags which we want.

Am I missing something here ?  The reporter is talking about Span.Process.Tags, right ?

SamrathPalSingh:
@yurishkuro got it. I have made changes to drop Tags if if --es.tags-as-fields.include or --es.tags-as-fields.all is used. Does this make sense. I have reverted the rest of the changes on agent's side.

plugin/storage/es/spanstore/dbmodel/from_domain.go
```diff
@@ -45,17 +45,32 @@ func (fd FromDomain) FromDomainEmbedProcess(span *model.Span) *Span {
...
+			OperationName:   span.OperationName,
+			StartTime:       model.TimeAsEpochMicroseconds(span.StartTime),
+			StartTimeMillis: model.TimeAsEpochMicroseconds(span.StartTime) / 1000,
+			Duration:        model.DurationAsMicroseconds(span.Duration),
+			Tags:            tags,
+			Tag:             tagsMap,
+			Logs:            fd.convertLogs(span.Logs),
+		}
+	} else {
+		return Span{
```
yurishkuro, line 62: this is a poor choice to copy & paste code that differs in only one line.

yurishkuro (commented):
This is not right either. Dropping the full map makes sense when `fd.allTagsAsFields` is true. But then only selected tags are lifted up to fields, dropping the map is equivalent to data loss. The correct behavior would be to reduce the map by removing the keys that are lifted into fields, but leave all others in the map.

Also, this new behavior needs to be opt-in, not automatic.

SamrathPalSingh:
Adding es.tags-as-fields.remove flag. If set to true, it will remove the the tags which are not present in es.tags-as-fields.include. By default, only the tags which are lifted into the fields will be removed, the rest will stay. 

@yurishkuro I've also changed the duplicate code part.

yurishkuro:
Please describe in words what it is that you're trying to achieve. With data examples if that helps.

SamrathPalSingh:
In issue #3403 the reporter is looking for a way to drop the tags which are not present in --es.tags-as-fields.include flag. The idea is to reduce the amount of data stored. This adds functionality to do exactly that. By adding the --es.tags-as-fields.remove flag we can drop the unwanted tags if required. 

Let's take the reporters example. To quote directly from the issue:

_...The purpose of me using ES_TAGS_AS_FIELDS_INCLUDE is to reduce the ingestion data for stuff we don't care about, so it's counter productive for me to also write a big blob such as this....

Expected behavior:
The tags field to not be written when using ES_TAGS_AS_FIELDS_INCLUDE, or have a documented means of not writing it...._

If ES_TAGS_AS_FIELDS_INCLUDE is used, the collector will create fields of the selected tags and store the rest of the tags as a blob of raw data (see the screen shot attached in issue #3403). In the screenshot, you will see the fields (tag.http@protocol, tag.http@status_code etc ) extracted from the raw tags. The last entry in the same screenshot is called 'tags' and it stores the blob of raw data containing the tags which were not present in ES_TAGS_AS_FIELDS_INCLUDE. In the reporter's use case it is counter productive to store 'tags' as he has already extracted the data he needs in the form of fields. So, the last entry is a waste of storage. He basically wants to have a way of dropping this blob of raw tags.

Hope this explains what I'm trying to do here.

yurishkuro:
If a span has 5 tags, but the include flag lists only one, you are going to completely drop the rest 4. I don't believe this was the behavior asked for in the ticket. The ticket was about data duplication, not about dropping the blob part altogether.

SamrathPalSingh:
Data duplication is already handled in the current version of the code. A tag can be in either 'tags' (raw data) or in a field, it cannot be in both. I have asked the Reporter to clarify this in the issue.

SamrathPalSingh:
He has replied back saying this is exactly what is needs

SamrathPalSingh:
These changes can be made to OPTEL directly.

PR #3659, opened 2022-05-06
Add a Context that flows through span processors to SpanWriters

esnible:
This PR replaces https://github.com/jaegertracing/jaeger/pull/3605 and is an alternative to https://github.com/jaegertracing/jaeger/pull/3660

This PR adds a Context to `ProcessSpans()`.  This allows key/value pairs to flow through the processor down to the SpanWriter.

I hope to build on this when I introduce tenancy.

I have chosen to implement sending the tenant through the processors this way because
- We already use a context on the read path, e.g. _/proto-gen/storage_v1/storage.pb.go_ which has `GetTrace(ctx context.Context, in *GetTraceRequest, opts ...grpc.CallOption) (SpanReaderPlugin_GetTraceClient, error)`; it is symmetric to add a context to the write path.
- Contexts are traditionally used when data needs to flow through to an inner API; that is what we are doing.
- This is very flexible; only processors and writers that care need to look for the keys they honor
- Currently only the saveSpan() processor uses the context, and it only uses it to call the spanWriter.  However, in the future, other processors might take advantage of the context settings.
 
cc @pavolloffay

### DIFF
--- a/cmd/collector/app/processor/interface.go
+++ b/cmd/collector/app/processor/interface.go
@@ -15,6 +15,7 @@
 package processor
 
 import (
+	"context"
 	"errors"
 	"io"
 
@@ -28,6 +29,7 @@ var ErrBusy = errors.New("server busy")
 type SpansOptions struct {
 	SpanFormat       SpanFormat
 	InboundTransport InboundTransport
+	Context          context.Context
 }
 
 // SpanProcessor handles model spans
--- a/cmd/collector/app/root_span_handler_test.go
+++ b/cmd/collector/app/root_span_handler_test.go
@@ -15,6 +15,7 @@
 package app
 
 import (
+	"context"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -43,7 +44,7 @@ func TestHandleRootSpan(t *testing.T) {
 
 	// Testing non-root span
 	span := &model.Span{References: []model.SpanRef{{SpanID: model.NewSpanID(1), RefType: model.ChildOf}}}
-	processor(span)
+	processor(span, context.TODO())
 	assert.Equal(t, 0, aggregator.callCount)
 
 	// Testing span with service name but no operation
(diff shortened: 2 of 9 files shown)


cmd/collector/app/processor/interface.go
```diff
@@ -28,6 +29,7 @@ var ErrBusy = errors.New("server busy")
 type SpansOptions struct {
 	SpanFormat       SpanFormat
 	InboundTransport InboundTransport
+	Context          context.Context
```
yurishkuro, line 32: have a concern about this one. First, it is not idiomatic and explicitly goes against recommendation to avoid storing context in structs. Second, the reason we have this struct is because we have a queue sitting between the RPC endpoint accepting spans and the processing pipeline. Which means the context coming from RPC has a different life-cycle than the processing life-cycle. For example, by the time the item is processed from the queue the RPC context may have been cancelled already, which won't affect passing the tenancy, but might affect sending DB requests if the same context is piped all the way through the writers.

I think it would be better to capture the tenancy explicitly at this boundary and let the queue consumer create its own context (and re-save the tenancy there if necessary)

cmd/collector/app/root_span_handler_test.go
```diff
@@ -43,27 +44,27 @@ func TestHandleRootSpan(t *testing.T) {
 
 	// Testing non-root span
 	span := &model.Span{References: []model.SpanRef{{SpanID: model.NewSpanID(1), RefType: model.ChildOf}}}
-	processor(span)
+	processor(span, context.TODO())
```
yurishkuro, line 47: I think you want context.Background() in the tests - TODO() implies you want to change it in the future

cmd/collector/app/span_processor.go
```diff
@@ -156,17 +157,21 @@ func (sp *spanProcessor) saveSpan(span *model.Span) {
...
 
-func (sp *spanProcessor) countSpan(span *model.Span) {
+func (sp *spanProcessor) countSpan(span *model.Span, ctx context.Context) {
 	sp.bytesProcessed.Add(uint64(span.Size()))
 	sp.spansProcessed.Inc()
 }
 
 func (sp *spanProcessor) ProcessSpans(mSpans []*model.Span, options processor.SpansOptions) ([]bool, error) {
-	sp.preProcessSpans(mSpans)
+	ctx := options.Context
```
yurishkuro, line 166: this is what I meant in the previous comment - I think this should always be context.Background() (with some enrichment) due to different life-cycles.

PR #3660, opened 2022-05-06
Flow context to SpanWriter

esnible:
This PR replaces https://github.com/jaegertracing/jaeger/pull/3605 and is an alternative to https://github.com/jaegertracing/jaeger/pull/3659

This PR allows key/value pairs to reach the SpanWriter, but doesn't make them available to the processors.

I hope to build on this when I introduce tenancy.

I have chosen to implement sending the tenant through the processors this way because
- This PR is very small
- Contexts are traditionally used when data needs to flow through to an inner API; that is what we are doing.

cc @pavolloffay

### DIFF
--- a/cmd/collector/app/span_processor.go
+++ b/cmd/collector/app/span_processor.go
@@ -45,7 +45,7 @@ type spanProcessor struct {
 	preProcessSpans    ProcessSpans
 	filterSpan         FilterSpan             // filter is called before the sanitizer but after preProcessSpans
 	sanitizer          sanitizer.SanitizeSpan // sanitizer is called before processSpan
-	processSpan        ProcessSpan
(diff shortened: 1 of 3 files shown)


cmd/collector/app/span_processor.go
```diff
@@ -164,9 +170,15 @@ func (sp *spanProcessor) countSpan(span *model.Span) {
 func (sp *spanProcessor) ProcessSpans(mSpans []*model.Span, options processor.SpansOptions) ([]bool, error) {
 	sp.preProcessSpans(mSpans)
 	sp.metrics.BatchSize.Update(int64(len(mSpans)))
+
+	ctx := options.Context
```
yurishkuro, line 174: https://github.com/jaegertracing/jaeger/pull/3659#discussion_r867153563

PR #3661, opened 2022-05-06
Flow tenant through processors as a string

esnible:
This PR replaces https://github.com/jaegertracing/jaeger/pull/3659 which passed the tenant ID string as part of a context.  It is intended as the first part of a true replacement for the proof-of-concept https://github.com/jaegertracing/jaeger/pull/3605

cc @pavolloffay


pavolloffay:
@esnible is the intention here to make the writers tenant aware or have a writer per tenant and route to the appropriate writer in the span processor?

esnible:
@pavolloffay The intention is to have the writers be tenant-aware.

@yurishkuro said "I see benefits to having per-tenant queues for better isolation" but later said "My suggestion would be to start with the simplest form: everything identical and shared, but tenancy exists at the data level."  This is an attempt to get the tenant down to the SpanWriter with as little churn as possible as a first PR.

storage/spanstore/interface.go
```diff
@@ -86,3 +86,11 @@ type Operation struct {
 	Name     string
 	SpanKind string
 }
+
+// tenantKeyType is a custom type for the key "tenant", following context.Context convention
```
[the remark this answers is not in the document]
yurishkuro: also, these should probably be defined at storage/ level, not storage/spanstore

storage/tenant.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+// GetTenant retrieves a tenant associated with a Context
+func GetTenant(ctx context.Context) string {
+	tenant := ctx.Value(tenantKey)
+	if tenant == nil {
+		return ""
+	}
+
+	return tenant.(string)
```
yurishkuro, line 39: perhaps we could make it safer
```go
if s, ok := tenant.(string) {
    return s
}
return ""
```

storage/tenant.go
```diff
@@ -0,0 +1,40 @@
...
+// tenantKeyType is a custom type for the key "tenant", following context.Context convention
+type tenantKeyType string
+
+const (
+	// tenantKey holds tenancy for spans
+	tenantKey = tenantKeyType("tenant")
+)
+
+// WithTenant creates a Context with a tenant association
+func WithTenant(ctx context.Context, tenant string) context.Context {
```
yurishkuro, line 28: let's add basic tests here to keep the coverage (and remove empty_test.go)

cmd/collector/app/span_processor_test.go
```diff
@@ -566,3 +577,28 @@ func TestAdditionalProcessors(t *testing.T) {
...
+		Tenant: dummyTenant,
+	})
+	assert.NoError(t, err)
+	assert.Equal(t, []bool{true}, res)
+	assert.NoError(t, p.Close())
+
+	// Verify that the dummy tenant from SpansOptions context made it to writer
+	assert.Equal(t, true, w.tenants[dummyTenant])
+	// Verify no other tenantKey context values made it to writer
+	assert.True(t, reflect.DeepEqual(w.tenants, map[string]bool{dummyTenant: true}))
```
yurishkuro, line 603: I think you can simply use assert.Equal(), it should compare the maps correctly iirc.

yurishkuro (dismissed):

yurishkuro (approved):

PR #3681, opened 2022-05-16
Fix Prometheus factory signature

### LINKED ISSUE #3677: PROMETHEUS_SERVER_URL broken since 1.34
**Describe the bug**
After upgrading to 1.34 from 1.33 the `PROMETHEUS_SERVER_URL` value is not read anymore.

Using v1.33

```json
{"level":"info","ts":1652451118.0158124,"caller":"metricsstore/reader.go:94",
 "msg":"Prometheus reader initialized","addr":"http://172.16.128.175:9090"} 
```

And v1.34, which shows the default value.

```json
{"level":"info","ts":1652451438.1780195,"caller":"metricsstore/reader.go:94",
 "msg":"Prometheus reader initialized","addr":"http://localhost:9090"}
```

**To Reproduce**
1. Run 1.33 with monitor tab
2. Upgrade to 1.34
3. Notice the 500's error in the monitor tab

**Expected behavior**

**Screenshots**

 - OS: Linux
 - Jaeger version: 1.34
 - Deployment: Docker

I was using `PROMETHEUS_SERVERURL` env variable and switched to a `config-file`, same issue. Here is the config file.

```
log-level: info

admin:
  http:
    host-port: "0.0.0.0:29937"

query:
  base-path: "/jaeger"
  grpc-server:
    host-port: "0.0.0.0:24507"
  http-server:
    host-port: "0.0.0.0:25050"

es:
  server-urls: "http://opensearch-tracing.service.consul:9200"

prometheus:
  server-url: "http://172.16.128.175:9090"
```

**Additional context**

Asked for help on Slack: https://cloud-native.slack.com/archives/CGG7NFUJ3/p1652443407988039

yurishkuro:
Resolves #3677 
  * adds `plugin.Configurable' validation to all factories to avoid re-introducing the bug
  * remove error return value from Prometheus factory and log it as fatal inside the function, similar to all other factories

cc @albertteoh

### DIFF
--- a/pkg/testutils/logger.go
+++ b/pkg/testutils/logger.go
@@ -30,7 +30,7 @@ import (
 // NewLogger creates a new zap.Logger backed by a zaptest.Buffer, which is also returned.
 func NewLogger() (*zap.Logger, *Buffer) {
 	core, buf := newRecordingCore()
-	logger := zap.New(core)
+	logger := zap.New(core, zap.OnFatal(zapcore.WriteThenPanic))
 	return logger, buf
 }
 
--- a/plugin/metrics/disabled/factory.go
+++ b/plugin/metrics/disabled/factory.go
@@ -20,8 +20,11 @@ import (
 	"github.com/spf13/viper"
 	"go.uber.org/zap"
 
+	"github.com/jaegertracing/jaeger/plugin"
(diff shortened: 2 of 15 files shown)


albertteoh (approved):
LGTM, thanks.

PR #3688, opened 2022-05-17
Allow Jaeger's GRPC handler to flow the tenant from an HTTP header

esnible:

If Jaeger is started with for example `--collector.tenant-header=x-tenant`, flow the value of the x-tenant header to the tenant parameter for span processing.

If Jaeger is not started with `--collector.tenant-header`, this is essentially a no-op.

cc @pavolloffay


cmd/collector/app/handler/grpc_handler.go
```diff
@@ -60,3 +69,23 @@ func (g *GRPCHandler) PostSpans(ctx context.Context, r *api_v2.PostSpansRequest)
...
+
+// WithTenant creates a Context with a tenant association
+func (g *GRPCHandler) getTenant(ctx context.Context) (string, error) {
+	// If the collector isn't looking for a tenant, no tenant is needed
+	if g.tenantHeader == "" {
+		return "", nil
+	}
+	md, ok := metadata.FromIncomingContext(ctx)
+	if !ok {
+		return "", status.Errorf(codes.Internal, "missing metadata")
```
yurishkuro, line 81: is it an error to be missing metadata?

cmd/collector/app/builder_flags.go
```diff
@@ -40,6 +40,7 @@ const (
 	collectorGRPCMaxReceiveMessageLength = "collector.grpc-server.max-message-size"
 	collectorMaxConnectionAge            = "collector.grpc-server.max-connection-age"
 	collectorMaxConnectionAgeGrace       = "collector.grpc-server.max-connection-age-grace"
+	collectorTenantHeader                = "collector.tenant-header"
```
yurishkuro, line 43: gRPC is not the only way to submit data to collector. Is the intention to require the same header name across transports?

cmd/collector/app/handler/grpc_handler.go
yurishkuro: fair enough, but then I would not treat it as `codes.Internal`, but similar to other ones and I would alter the error message, e.g. "missing metadata, tenant header is required"

cmd/collector/app/builder_flags.go
pavolloffay: Do we need to specify tenants at the deployment time? I wanted to avoid restarting Jaeger (and changing deployments) when a new tenant is added.

yurishkuro: The configuration can be extended to accept tenants in other forms, like watching a file or by integrating with another source. This would be a good opportunity to start introducing a control plane where tenants can be provided externally via api call, such that the integration with the source can be decoupled from Jaeger code.

pkg/config/tenancy/flags.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/spf13/viper"
+)
+
+const (
+	tenancyEnabled = "multi_tenancy.enabled"
+	tenancyHeader  = "multi_tenancy.header"
+	validTenants   = "multi_tenancy.tenants"
+)
+
+// TenancyFlagsConfig describes which CLI flags for TLS client should be generated.
```
yurishkuro, line 31: the type does not seem necessary, we could just have global functions. Or, at least, it should be the same type as Options, that's the pattern we used, i.e. the config/options type exposes functions to initialize itself via CLI.

pkg/config/tenancy/tenancy.go
```diff
@@ -0,0 +1,77 @@
...
+
+func (tenantUnaware) Valid(candidate string) bool {
+	return true
+}
+
+type tenantList struct {
+	tenants map[string]bool
+}
+
+func (tl tenantList) Valid(candidate string) bool {
```
yurishkuro, line 55: are you sure you want by-value receiver here? There's going to be a heap allocation anyway when the struct is converted into Guard interface.

cmd/collector/app/handler/grpc_handler.go
```diff
@@ -20,28 +20,42 @@ import (
...
 func (g *GRPCHandler) PostSpans(ctx context.Context, r *api_v2.PostSpansRequest) (*api_v2.PostSpansResponse, error) {
+	var tenant string
+	if g.tenancyConfig.Enabled {
+		var err error
+		tenant, err = g.validateTenant(ctx)
+		if err != nil {
+			g.logger.Error("rejecting spans (tenancy)", zap.Error(err))
+			return nil, err
+		}
+	}
```
yurishkuro, line 57: suggest moving this to validateTenant
```suggestion
	tenant, err := g.validateTenant(ctx)
        if err != nil {
	        g.logger.Error("failed to validate tenant, rejecting spans", zap.Error(err))
	        return nil, err
        }
```

pkg/config/tenancy/tenancy.go
```diff
@@ -0,0 +1,77 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tenancy
+
+// TenancyConfig holds the settings for multi-tenant Jaeger
+type TenancyConfig struct {
+	Enabled bool
+	Header  string
+	Valid   Guard
```
yurishkuro, line 21: this looks weird in the usage: `config.Valid.Valid()`. How about:
```
type TenancyConfig struct {
    ...
    guard Guard
}

func (c *TenancyConfig) Valid(v string) bool {
    return c.guard.Valid()
}
```

especially since the current Valid variable is force-assigned in the constructor & cannot be user-supplied.

cmd/collector/app/builder_flags.go
pavolloffay: I would propose tackling the tenant restrictions in a separate PR.

More sophisticated deployments should do tenant validation in the auth layer (proxy). Which makes me think do we even need Jaeger to do this validation? A valid requirement for predefined tenants might be if the storage implementation needs to do some initialization per tenant. In the case of ES it should not be required, but maybe other storages like C* need to create a new DB that cannot be done in-flight of the first store call.

yurishkuro: The validation I'm most concerned about is not whether tenant is valid, but whether it is sent at all. Without that we open a hole that could allow other tenants to read your data.

I agree that there needs to be a stronger authentication somewhere other than a header, which is trivial to spoof.

@jkowall any experience to share here?

esnible: I will relax the handling of valid tenants so that if `multi_tenancy.tenants` is undefined any tenant will pass.  This will allow the spanwriter to reject spans.  In the future we can provide a way for the writer to reject earlier.

Storage might have rules about what a valid tenant can be that we can't express on a list.  For example, some storage might reject tenants that include non-ASCII characters.  Or some storage might want to hot-enable/disable a tenant.

yurishkuro (commented):
lgtm

cmd/collector/app/handler/grpc_handler_test.go
```diff
@@ -152,3 +168,167 @@ func TestPostSpansWithError(t *testing.T) {
...
+			batch: model.Batch{Process: &model.Process{ServiceName: "batch-process"}, Spans: []*model.Span{{OperationName: "test-op", Process: &model.Process{ServiceName: "bar"}}}},
+
+			// NewGRPCHandler rejects spans with multiple values for tenant header
+			mustFail:        true,
+			expected:        nil,
+			expectedTenants: nil,
+		},
+	}
+	for _, test := range tests {
+		_, err := client.PostSpans(test.ctx, &api_v2.PostSpansRequest{
```
pavolloffay, line 257: the individual test cases should be executed separately via `t.Run`

cmd/collector/app/handler/grpc_handler_test.go
```diff
@@ -152,3 +168,167 @@ func TestPostSpansWithError(t *testing.T) {
...
+	}
+
+	processor := &mockSpanProcessor{}
+	handler := NewGRPCHandler(zap.NewNop(), processor, tenancy.NewTenancyConfig(&tenancy.Options{
+		Enabled: true,
+		Header:  tenantHeader,
+		Tenants: validTenants,
+	}))
+	for _, test := range tests {
+		tenant, err := handler.validateTenant(test.ctx)
```
pavolloffay, line 326: the same here, test cases should be run via `t.Run`

pkg/config/tenancy/flags.go
```diff
@@ -0,0 +1,50 @@
...
+const (
+	tenancyEnabled = "multi_tenancy.enabled"
+	tenancyHeader  = "multi_tenancy.header"
+	validTenants   = "multi_tenancy.tenants"
+)
+
+// AddFlags adds flags for tenancy to the FlagSet.
+func AddFlags(flags *flag.FlagSet) {
+	flags.Bool(tenancyEnabled, false, "Enable tenancy header when receiving or querying")
+	flags.String(tenancyHeader, "x-tenant", "HTTP header carrying tenant")
```
pavolloffay, line 33: As a first version, we could have a single flag `tenant-header` which would enable tenancy if it is defined. 

Internally we will most likely not use the predefined list of tenants (e.g. in the observatorium project) and if the storage layer does not need to have access to the tenant names I find this configuration redundant.

cmd/collector/app/handler/grpc_handler_test.go
esnible: I followed the format of `TestPostSpans()`.  No tests in _cmd/collector/app/handler_ use `t.Run()`.

esnible:
I struggled with re-basing after https://github.com/jaegertracing/jaeger/pull/3707 .  Instead I copied the code changes and tests and manually applied them to _main_.

cmd/collector/app/handler/grpc_handler.go
```diff
@@ -45,13 +48,20 @@ func NewGRPCHandler(logger *zap.Logger, spanProcessor processor.SpanProcessor) *
...
 func (g *GRPCHandler) PostSpans(ctx context.Context, r *api_v2.PostSpansRequest) (*api_v2.PostSpansResponse, error) {
+	tenant, err := g.validateTenant(ctx)
+	if err != nil {
+		g.logger.Error("rejecting spans (tenancy)", zap.Error(err))
+		return nil, err
+	}
+
 	batch := &r.Batch
-	err := g.batchConsumer.consume(batch)
+	err = g.batchConsumer.consume(batch, tenant)
```
yurishkuro, line 64: I would suggest changing the signature to `consume(context, batch)` and handling tenancy inside. This way it can be reused in OTLP receiver.

yurishkuro (approved):

PR #3700, opened 2022-05-25
Flow tenant from query

esnible:
This PR ensures the client-supplied tenant is validated by the HTTP/gRPC layer, and passed through the Context to the storage layer.

This builds upon https://github.com/jaegertracing/jaeger/pull/3688 and should will be rebased after that PR merges

Please review this for direction only until after 3688 is merged.  Some stuff in the current PR duplicates 3688.

The basic idea is to merely guard the spanstore.Reader and the dependencystore.Reader.

cc @pavolloffay

### DIFF
--- a/cmd/all-in-one/main.go
+++ b/cmd/all-in-one/main.go
@@ -268,7 +268,6 @@ func startQuery(
 	metricsQueryService querysvc.MetricsQueryService,
 	baseFactory metrics.Factory,
 ) *queryApp.Server {
-	fmt.Printf("@@@ ecs REACHED startQuery, tenancy is %v\n", qOpts.Tenancy.Enabled)
 	if qOpts.Tenancy.Enabled {
 		spanReader = tenancy.NewGuardedSpanReader(spanReader, &qOpts.Tenancy)
 		depReader = tenancy.NewGuardedDependencyReader(depReader, &qOpts.Tenancy)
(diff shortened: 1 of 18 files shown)


cmd/all-in-one/main.go
```diff
@@ -267,6 +268,11 @@ func startQuery(
 	metricsQueryService querysvc.MetricsQueryService,
 	baseFactory metrics.Factory,
 ) *queryApp.Server {
+	if qOpts.Tenancy.Enabled {
+		spanReader = tenancy.NewGuardedSpanReader(spanReader, &qOpts.Tenancy)
+		depReader = tenancy.NewGuardedDependencyReader(depReader, &qOpts.Tenancy)
```
yurishkuro, line 273: nit: tenancy guard is a cross-cutting concern, similar to metrics. We usually do not apply metrics decorators in the main functions, but do it centrally in the storage factory facade. I think it would be better to do similar thing with the tenancy guard decorator, to ensure it's always present, instead of copy-pasting the same across many main()s

yurishkuro:
This is difficult to review given many previous changes. I would recommend the "stacked diffs" approach here, i.e. instead of this PR targeting `main` for merge, target the branch of the base PR.

cmd/all-in-one/main.go
esnible: @yurishkuro The PR for collector had an explicit guard check in PostSpans().  This PR wrapped the spanReader, similar to how Metrics are wrapped in several _main.go_ with `spanReader = spanstoreMetrics.NewReadMetricsDecorator(spanReader, metricsFactory)`.

Do you prefer:
- explicit guard in each method?
- explicit wrapper in each _main.go_ that constructs a spanReader
- Add tenancy to `storage.FactoryConfig`

yurishkuro: my preference is to enable the guard at the factory level

PR #3705, opened 2022-05-26
Add flag to enable/disable dependencies on rollover

rubenvp8510:

-  Solves https://github.com/jaegertracing/jaeger/issues/3706

- This PR adds a flag to enable/disable the dependencies on roll-over, if the dependencies is disabled it won't rollover dependencies index.

### DIFF
--- a/cmd/es-rollover/app/flags.go
+++ b/cmd/es-rollover/app/flags.go
@@ -21,25 +21,27 @@ import (
 )
 
 const (
-	indexPrefix   = "index-prefix"
-	archive       = "archive"
-	username      = "es.username"
-	password      = "es.password"
-	useILM        = "es.use-ilm"
-	ilmPolicyName = "es.ilm-policy-name"
-	timeout       = "timeout"
+	indexPrefix      = "index-prefix"
+	archive          = "archive"
+	username         = "es.username"
+	password         = "es.password"
+	useILM           = "es.use-ilm"
+	ilmPolicyName    = "es.ilm-policy-name"
+	timeout          = "timeout"
+	skipDependencies = "skip-dependencies"
 )
 
 // Config holds the global configurations for the es rollover, common to all actions
 type Config struct {
-	IndexPrefix   string
-	Archive       bool
-	Username      string
-	Password      string
-	TLSEnabled    bool
-	ILMPolicyName string
-	UseILM        bool
-	Timeout       int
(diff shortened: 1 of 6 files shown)


yurishkuro:
Could you please fill out "Which problem is this PR solving" section? Or is there a ticket that describes that?

cmd/es-rollover/app/flags.go
```diff
@@ -51,6 +53,8 @@ func AddFlags(flags *flag.FlagSet) {
 	flags.Bool(useILM, false, "Use ILM to manage jaeger indices")
 	flags.String(ilmPolicyName, "jaeger-ilm-policy", "The name of the ILM policy to use if ILM is active")
 	flags.Int(timeout, 120, "Number of seconds to wait for master node response")
+	flags.Bool(dependenciesEnabled, false, "Enable dependency index rollover")
```
yurishkuro, line 56: this doesn't seem like a backwards compatible change. Maybe you want to change the flag to "skip-dependencies", where default value false would retain the existing behavior.

yurishkuro: Note: this only applies to the CLI flag, in the rest of the code it's fine to use `dependenciesEnabled` since the default value does not play a role there when the param is passed explicitly.

rubenvp8510: Fair point, I'll do the change

rubenvp8510: On the other hand, having this feature enable is indeed breaking some old jaeger installations.

frzifus (approved):

cmd/es-rollover/app/flags.go
```diff
@@ -51,6 +53,8 @@ func AddFlags(flags *flag.FlagSet) {
 	flags.Bool(useILM, false, "Use ILM to manage jaeger indices")
 	flags.String(ilmPolicyName, "jaeger-ilm-policy", "The name of the ILM policy to use if ILM is active")
 	flags.Int(timeout, 120, "Number of seconds to wait for master node response")
+	flags.Bool(skipDependencies, false, "Enable dependency index rollover")
```
yurishkuro, line 56:
```suggestion
	flags.Bool(skipDependencies, false, "Disable rollover for dependencies index")
```

yurishkuro (approved):

PR #3735, opened 2022-06-06
Tenant handling for queries using interceptor

esnible:

This replaces the draft https://github.com/jaegertracing/jaeger/pull/3700

This PR flows the tenant header through the query APIs.
This PR blocks requests with invalid tenant or missing tenant header from reaching the query APIs.

@pavolloffay @yurishkuro I had planned to only deliver tenant header checking in this PR.  In interactive testing I discovered that enabling tenancy was causing self-spans to be rejected.  I added a simple implementation for that.  It is possible we should break that implementation out (and deliver it separately first.)

I was uncertain how to handle the CLI options for self-span reporting.  I only test with Jaeger reporting against itself.  Probably I should introduce new CLI options `-reporter.tenant_header` and `-reporter.tenant` so that Jaeger can send spans to a different Jaeger.  Do you concur?  Or should we use the `--multi_tenancy.header` and a special value like `_self` for the tenant?

### DIFF
--- a/cmd/agent/app/proxy_builders.go
+++ b/cmd/agent/app/proxy_builders.go
@@ -16,11 +16,16 @@ package app
 
 import (
 	"github.com/jaegertracing/jaeger/cmd/agent/app/reporter/grpc"
+	"google.golang.org/grpc/metadata"
 )
 
 // GRPCCollectorProxyBuilder creates CollectorProxyBuilder for GRPC reporter
 func GRPCCollectorProxyBuilder(builder *grpc.ConnBuilder) CollectorProxyBuilder {
 	return func(opts ProxyBuilderOptions) (proxy CollectorProxy, err error) {
-		return grpc.NewCollectorProxy(builder, opts.AgentTags, opts.Metrics, opts.Logger)
+		md := metadata.New(map[string]string{})
+		if opts.TenancyHeader != "" {
+			md = metadata.New(map[string]string{opts.TenancyHeader: opts.Tenant})
+		}
+		return grpc.NewCollectorProxy(builder, opts.AgentTags, opts.Metrics, md, opts.Logger)
 	}
 }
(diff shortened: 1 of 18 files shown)


yurishkuro:
> In interactive testing I discovered that enabling tenancy was causing self-spans to be rejected.

So I think there are two issues here to unpack.

1) Are the self-spans being reported without a tenant in the first place, and hence being rejected? I think fixing this requires changes in the SDKs to allow configuring the tenant, which with Jaeger SDKs being retired is somewhat problematic (although the change is relatively small and only to the Jaeger SDK, so we could do it if necessary). Another alternative is to switch Jaeger to use OTEL SDKs internally, which needs to happen sometime in any case, but it's an independent task.

2) If the self-spans were sent with a tenant, would they still be rejected because the tenant is not one of those explicitly configured? We could approach this in two ways, either (a) always configure the SDKs to send the default tenant `jaeger` and have that tenant automatically enabled in the backend (in addition to what may be user-configured), or (b) expect the user to always explicitly configure the tenant to be used by the SDK and add that tenant into allow-list. I would prefer to go with (a), and perhaps later we could add options to disable that, if someone asks.

esnible:
@yurishkuro The self-spans are being reported without a tenant.
- We could send along the tenant `jaeger` or `_self`
- We could send a user-configurable self tenant such as "us-east.prod.jaeger"
- We could send along the tenant making the query (as it is query that triggers these spans).

I didn't notice this when I added tenancy checking to gRPC PostSpans because there are no tests where we query and then verify the server posted some spans about the query it performed.

The self spans and a `context.Context` flow through thrift processor, agent processor, client metrics reporter, and gRPC reporter.  Adding tenant metadata at any of these levels allows the query spans to be stored.

yurishkuro:
Since you mention the agent, it receives spans over UDP without any metadata. To be able to include tenant there we will need to extend the span data model itself.

PR #3742, opened 2022-06-08
Optionally supply tenancy when reporting spans

esnible:

With https://github.com/jaegertracing/jaeger/pull/3688 we introduced `--multi_tenancy.enabled=true` which enables the collector to reject spans with missing or invalid tenant header.

This PR restores correct span reporting when tenancy is enabled.  Without this change, when a multi-tenant Jaeger gets a (currently untenanted) query, it reports spans related to handling the query without a tenant.  Those (usually) self spans are rejected.

Some of the work in this PR was previously done in the draft PR for query tenancy, https://github.com/jaegertracing/jaeger/pull/3735 .  That PR was getting big so the reporting stuff will be removed and handled in this PR.

### DIFF
--- a/cmd/agent/app/proxy_builders.go
+++ b/cmd/agent/app/proxy_builders.go
@@ -15,8 +15,6 @@
 package app
 
 import (
-	"fmt"
-
 	"google.golang.org/grpc/metadata"
 
 	"github.com/jaegertracing/jaeger/cmd/agent/app/reporter/grpc"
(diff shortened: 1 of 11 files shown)


cmd/agent/app/proxy_builders.go
```diff
@@ -15,12 +15,18 @@
...
 	"github.com/jaegertracing/jaeger/cmd/agent/app/reporter/grpc"
 )
 
 // GRPCCollectorProxyBuilder creates CollectorProxyBuilder for GRPC reporter
 func GRPCCollectorProxyBuilder(builder *grpc.ConnBuilder) CollectorProxyBuilder {
 	return func(opts ProxyBuilderOptions) (proxy CollectorProxy, err error) {
-		return grpc.NewCollectorProxy(builder, opts.AgentTags, opts.Metrics, opts.Logger)
+		md := metadata.New(map[string]string{})
+		if builder.CollectorTenancyHeader != "" {
+			md = metadata.New(map[string]string{builder.CollectorTenancyHeader: builder.CollectorTenant})
```
yurishkuro, line 28: With this change every span received by the Agent will be tagged with a single tenant when forwarded to Collector. While this is a viable deployment model in some situations, I don't think it's ultimately the right one, especially when the Agent runs as a host agent receiving data from potentially multiple co-located applications (which could be different tenants).

For example, this makes it impossible to test multi-tenancy with all-in-one if the other tenants are also going through the agent.

My preference would be for the agent to extract the tenant directly from the spans it receives, and the current solution could be an opt-in extension to either override the tenant or inject it when empty (for cases when agent is used as a sidecar).

esnible:
Closing in favor of https://github.com/jaegertracing/jaeger/pull/3750

PR #3750, opened 2022-06-09
Optionally supply tenancy when reporting spans v2

esnible:
Replacement for https://github.com/jaegertracing/jaeger/pull/3742

This PR attempts to address https://github.com/jaegertracing/jaeger/pull/3742#discussion_r892803416

I am creating this as a draft, for direction, to make sure I understand the requirement before doing all of the necessary work.

### DIFF
--- a/cmd/agent/app/builder.go
+++ b/cmd/agent/app/builder.go
@@ -244,7 +244,6 @@ func CreateCollectorProxy(
 	opts ProxyBuilderOptions,
 	builders map[reporter.Type]CollectorProxyBuilder,
 ) (CollectorProxy, error) {
-	fmt.Printf("@@@ ecs REACHED CreateCollectorProxy, opts=%#v\n", opts)
 	builder, ok := builders[opts.ReporterType]
 	if !ok {
 		return nil, fmt.Errorf("unknown reporter type %s", string(opts.ReporterType))
--- a/cmd/agent/app/builder.go
+++ b/cmd/agent/app/builder.go
@@ -244,6 +244,7 @@ func CreateCollectorProxy(
 	opts ProxyBuilderOptions,
 	builders map[reporter.Type]CollectorProxyBuilder,
 ) (CollectorProxy, error) {
+	fmt.Printf("@@@ ecs REACHED CreateCollectorProxy, opts=%#v\n", opts)
 	builder, ok := builders[opts.ReporterType]
 	if !ok {
 		return nil, fmt.Errorf("unknown reporter type %s", string(opts.ReporterType))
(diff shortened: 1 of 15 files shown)


yurishkuro:
Could you describe what solution specifically you're implementing?  There seems to be a mix of things, but the main issue with agent being multi-tenant does not seem to be addressed.

esnible:
I'd like to operate a cloud-based service that consumes OpenTelemetry traces and provides a Jaeger UI and gRPC query API.

I have a reverse proxy that implements single-sign-on and adds a header, `x-tenant`, when talking to the OTel collector and to Jaeger query.

My goal is to allow different groups of users to interact with the same Jaeger and see different data.

I did a proof-of-concept at https://github.com/jaegertracing/jaeger/pull/3605 with a multi-tenant memory storage and partial implementation of multi-tenant collecting and query.  I am now trying to complete that work correctly, in small PRs.

After https://github.com/jaegertracing/jaeger/pull/3688 the tenant from the PostSpans header is passed along to storage.  While working on a similar PR to pass the header from the query API to the spanstore I noticed Jaeger's logs reporting that if multitenant collection was enabled the opentracing middleware spans for query were being rejected because they lacked tenancy.

I believe the self-traces from Jaeger's HTTP server "belong" to the service operator, not the service user.  This is why I created https://github.com/jaegertracing/jaeger/pull/3742 , to supply a service-operator tenant for those spans.

You correctly pointed out that I wasn't thinking about reporting in general.  I only thought of correcting the span dropping introduced by my tenant checking on PostSpans.  It does make sense to let the developer choose if reporting is done based on the active tenant or an "internal" tenant.

I've spent a day looking at the code and creating draft PRs.  I tried a hack where I encode the tenant as a tag.

A clean solution will require either a multitenant `opentracing.Tracer`, or Jaeger to report queries using a different mechanism.  I lack community experience and don't know which design is preferred.

yurishkuro:
Basically, in order to support self-reported spans, we have these options with additional work in each:
  * continue using Jaeger Go SDK internally sending spans over UDP:
    * Thrift data format must be extended to capture tenant
    * Jaeger Go SDK must be extended to populate the tenant in Thrift packets
  * continue using Jaeger Go SDK internally but switch to sending spans over HTTP or gRPC:
    * Jaeger Go SDK must be extended to populate the tenant header in HTTP/gRPC requests
  * switch to use OTEL SDK internally
    * I would also switch to use gRPC sending directly to collector
    * (not sure if already done) OTEL SDK must support sending x-tenant header 

Could we just ignore the problem of self-reported spans for now and focus more on the production-quality deployments by enabling collector and query service?

PR #3764, opened 2022-06-20
Replace 'operation' query param by  'operationName'  in `/api/traces` API

pjgg:
Currently `api/traces` returns an structure with several fields as `operationName`

for example:

```
{
               "traceID":"c54574036f1d750cb58f84a4275e9f8e",
               "spanID":"70a59e0d6f079f65",
               "operationName":"/trace/ping",
               .
               .
               .
}

```

`OperationName` represents the same concept as `operation` trace API query param. For exmaple, `/api/traces?service=vertx-web-client&operation=ping`

This PR tries to unify both domains in a single domain. My proposal is to use `operationName` as a first choice, doesn't matter if we are talking about query params or return pojos. `operation` query param is still supported in order to be backward compatible.

### DIFF
--- a/cmd/query/app/query_parser.go
+++ b/cmd/query/app/query_parser.go
@@ -33,17 +33,18 @@ import (
 const (
 	defaultQueryLimit = 100
 
-	operationParam   = "operation"
-	tagParam         = "tag"
-	tagsParam        = "tags"
-	startTimeParam   = "start"
-	limitParam       = "limit"
-	minDurationParam = "minDuration"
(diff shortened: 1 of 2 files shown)


yurishkuro:
The REST API is intentionally designed to use short words in the URL, like `service` instead of `serviceName`. This change would make it more verbose and internally inconsistent, so we can't accept this PR. In the future I would recommend opening a ticket to discuss the change you propose, before spending time working on the code.

PR #3782, opened 2022-06-26
Add flag to enable span size metrics reporting

### LINKED ISSUE #2126: `jaeger_collector_spans_bytes` is always 0
Gain information about per span sizes through Jaeger

Currently `jaeger_collector_spans_bytes` always reports 0.

After upgrading to 1.17 I added `jaeger_collector_spans_bytes` up on a graph to get more information about per span sizing.  Unfortunately the gauge appears to be pinned at 0.

image

I'm unable to at the moment, but when I get some time I will dig deeper on this.

ymtdzzz:
Resolves #2126 

- add `--collector.enable-span-size-metrics` that enables processed span metrics even if `--collector.queue-size-memory` is disabled

## Details
- `--collector.enable-span-size-metrics` is enabled but `--collector.queue-size-memory` is disabled

<details>

```yaml
apiVersion: jaegertracing.io/v1
kind: Jaeger
metadata:
  name: jaeger
spec:
  strategy: AllInOne
  imagePullPolicy: Never
  allInOne:
    image: jaegeraio:wip
    options:
      collector:
        enable-span-size-metrics: true
```

image

</details>

- `--collector.enable-processed-span-metrics` is disabled but `--collector.queue-size-memory` is enabled (no change)

<details>

```yaml
apiVersion: jaegertracing.io/v1
kind: Jaeger
metadata:
  name: jaeger
spec:
  strategy: AllInOne
  imagePullPolicy: Never
  allInOne:
    image: jaegeraio:wip
    options:
      collector:
        queue-size-memory: 10
```

image

</details>

### DIFF
--- a/cmd/collector/app/flags/flags.go
+++ b/cmd/collector/app/flags/flags.go
@@ -30,11 +30,11 @@ import (
 )
 
 const (
-	flagDynQueueSizeMemory          = "collector.queue-size-memory"
-	flagNumWorkers                  = "collector.num-workers"
-	flagQueueSize                   = "collector.queue-size"
-	flagCollectorTags               = "collector.tags"
-	flagProcessedSpanMetricsEnabled = "collector.enable-processed-span-metrics"
+	flagDynQueueSizeMemory     = "collector.queue-size-memory"
+	flagNumWorkers             = "collector.num-workers"
+	flagQueueSize              = "collector.queue-size"
+	flagCollectorTags          = "collector.tags"
+	flagSpanSizeMetricsEnabled = "collector.enable-span-size-metrics"
 
 	flagSuffixHostPort = "host-port"
 
@@ -125,8 +125,8 @@ type CollectorOptions struct {
 	}
 	// CollectorTags is the string representing collector tags to append to each and every span
 	CollectorTags map[string]string
-	// ProcessedSpanMetricsEnabled determines whether to enable processed span metrics
-	ProcessedSpanMetricsEnabled bool
+	// SpanSizeMetricsEnabled determines whether to enable metrics based on processed span size
+	SpanSizeMetricsEnabled bool
 }
 
 type serverFlagsConfig struct {
@@ -166,7 +166,7 @@ func AddFlags(flags *flag.FlagSet) {
 	flags.Int(flagQueueSize, DefaultQueueSize, "The queue size of the collector")
 	flags.Uint(flagDynQueueSizeMemory, 0, "(experimental) The max memory size in MiB to use for the dynamic queue.")
 	flags.String(flagCollectorTags, "", "One or more tags to be added to the Process tags of all spans passing through this collector. Ex: key1=value1,key2=${envVar:defaultValue}")
-	flags.Bool(flagProcessedSpanMetricsEnabled, false, "Enables processed span metrics.")
+	flags.Bool(flagSpanSizeMetricsEnabled, false, "Enables metrics based on processed span size, which are more expensive to calculate.")
 
 	addHTTPFlags(flags, httpServerFlagsCfg, ports.PortToHostPort(ports.CollectorHTTP))
 	addGRPCFlags(flags, grpcServerFlagsCfg, ports.PortToHostPort(ports.CollectorGRPC))
@@ -243,7 +243,7 @@ func (cOpts *CollectorOptions) InitFromViper(v *viper.Viper, logger *zap.Logger)
 	cOpts.NumWorkers = v.GetInt(flagNumWorkers)
 	cOpts.QueueSize = v.GetInt(flagQueueSize)
 	cOpts.DynQueueSizeMemory = v.GetUint(flagDynQueueSizeMemory) * 1024 * 1024 // we receive in MiB and store in bytes
-	cOpts.ProcessedSpanMetricsEnabled = v.GetBool(flagProcessedSpanMetricsEnabled)
+	cOpts.SpanSizeMetricsEnabled = v.GetBool(flagSpanSizeMetricsEnabled)
 
 	if err := cOpts.HTTP.initFromViper(v, logger, httpServerFlagsCfg); err != nil {
 		return cOpts, fmt.Errorf("failed to parse HTTP server options: %w", err)
--- a/cmd/collector/app/flags/flags.go
+++ b/cmd/collector/app/flags/flags.go
@@ -30,10 +30,11 @@ import (
 )
 
 const (
-	flagDynQueueSizeMemory = "collector.queue-size-memory"
-	flagNumWorkers         = "collector.num-workers"
-	flagQueueSize          = "collector.queue-size"
-	flagCollectorTags      = "collector.tags"
+	flagDynQueueSizeMemory          = "collector.queue-size-memory"
+	flagNumWorkers                  = "collector.num-workers"
+	flagQueueSize                   = "collector.queue-size"
+	flagCollectorTags               = "collector.tags"
+	flagProcessedSpanMetricsEnabled = "collector.enable-processed-span-metrics"
 
 	flagSuffixHostPort = "host-port"
 
@@ -124,6 +125,8 @@ type CollectorOptions struct {
 	}
 	// CollectorTags is the string representing collector tags to append to each and every span
 	CollectorTags map[string]string
+	// ProcessedSpanMetricsEnabled determines whether to enable processed span metrics
+	ProcessedSpanMetricsEnabled bool
 }
 
 type serverFlagsConfig struct {
@@ -163,6 +166,7 @@ func AddFlags(flags *flag.FlagSet) {
 	flags.Int(flagQueueSize, DefaultQueueSize, "The queue size of the collector")
 	flags.Uint(flagDynQueueSizeMemory, 0, "(experimental) The max memory size in MiB to use for the dynamic queue.")
 	flags.String(flagCollectorTags, "", "One or more tags to be added to the Process tags of all spans passing through this collector. Ex: key1=value1,key2=${envVar:defaultValue}")
+	flags.Bool(flagProcessedSpanMetricsEnabled, false, "Enables processed span metrics.")
 
 	addHTTPFlags(flags, httpServerFlagsCfg, ports.PortToHostPort(ports.CollectorHTTP))
 	addGRPCFlags(flags, grpcServerFlagsCfg, ports.PortToHostPort(ports.CollectorGRPC))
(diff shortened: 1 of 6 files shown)


cmd/collector/app/span_processor.go
```diff
@@ -123,10 +123,12 @@ func newSpanProcessor(spanWriter spanstore.Writer, additional []ProcessSpan, opt
...
 	processSpanFuncs := []ProcessSpan{options.preSave, sp.saveSpan}
 	if options.dynQueueSizeMemory > 0 {
-		// add to processSpanFuncs
 		options.logger.Info("Dynamically adjusting the queue size at runtime.",
 			zap.Uint("memory-mib", options.dynQueueSizeMemory/1024/1024),
 			zap.Uint("queue-size-warmup", options.dynQueueSizeWarmup))
+	}
+	if options.dynQueueSizeMemory > 0 || options.processedSpanMetricsEnabled {
+		// add to processSpanFuncs
 		processSpanFuncs = append(processSpanFuncs, sp.countSpan)
```
ymtdzzz, line 132: As mentioned in https://github.com/jaegertracing/jaeger/issues/2126#issuecomment-600663551 , `sp.spansProcessed.Inc()` could be independent but I didn't change this for simplicity.

cmd/collector/app/span_processor.go
```diff
@@ -123,10 +123,12 @@ func newSpanProcessor(spanWriter spanstore.Writer, additional []ProcessSpan, opt
 
 	processSpanFuncs := []ProcessSpan{options.preSave, sp.saveSpan}
```
yurishkuro, line 124: Couldn't we just do this? I don't think this requires any new cmd line options
```suggestion
	processSpanFuncs := []ProcessSpan{options.preSave, sp.saveSpan, sp.countSpan}
```

ymtdzzz: Thank you for your suggestion!
From the discussions in https://github.com/jaegertracing/jaeger/issues/2126 , my understanding is that this metrics should be an optional due to performance concerns ( https://github.com/jaegertracing/jaeger/issues/2126#issuecomment-600038153 ).
If there is no problem in always enabling it, the code will be as suggested.

yurishkuro: oh, i see. Then I guess it's ok to have a flag, but I would still make the counting only dependent on that flag, not on the dynqueue

ymtdzzz: Hmm, `bytesProcessed` is used by both metrics and dynqueue, so the counting is needed if either flag is true.

https://github.com/jaegertracing/jaeger/blob/9b4bbf4f75bac3333f1099e26cc7cf3dd79773bf/cmd/collector/app/span_processor.go#L269

How about to count in either case and use `--collector.enable-processed-span-metrics` flag to determine whether to update the metrics? This makes the role of the flag straightforward.

```go
func (sp *spanProcessor) updateGauges() {
        if sp.processedSpanMetricsEnabled {
                sp.metrics.SpansBytes.Update(int64(sp.bytesProcessed.Load()))
        }
```

yurishkuro: The bumping of the metric itself is very cheap, there's no need to protect it with a flag. 

There's certainly a number of places touching `bytesProcessed`. I am not particularly liking this design - couldn't this int64 variable overflow with time? But that's a problem separate from this PR.

The main concern from the ticket was the overhead of `span.Size()` calls.

ymtdzzz: This performance concern is existing issue of dynqueue and it doesn't come up unless the flag is true, so why not discuss about performance and design on a separate ticket?

I think the following issues should be handled by separate PRs.
- issue that metrics cannot be collected unless dynqueue is enabled (resolved by this PR)
- performance tuning the counting
- improving design not to depend too much on bytesProcessed and to avoid risk of overflow

The latter two seem to require more discussion before code changing.

cmd/collector/app/flags/flags.go
```diff
@@ -30,10 +30,11 @@ import (
...
 const (
-	flagDynQueueSizeMemory = "collector.queue-size-memory"
-	flagNumWorkers         = "collector.num-workers"
-	flagQueueSize          = "collector.queue-size"
-	flagCollectorTags      = "collector.tags"
+	flagDynQueueSizeMemory          = "collector.queue-size-memory"
+	flagNumWorkers                  = "collector.num-workers"
+	flagQueueSize                   = "collector.queue-size"
+	flagCollectorTags               = "collector.tags"
+	flagProcessedSpanMetricsEnabled = "collector.enable-processed-span-metrics"
```
yurishkuro, line 37:
```suggestion
	flagProcessedSpanMetricsEnabled = "collector.enable-span-size-metrics"
```

cmd/collector/app/flags/flags.go
```diff
@@ -163,6 +166,7 @@ func AddFlags(flags *flag.FlagSet) {
 	flags.Int(flagQueueSize, DefaultQueueSize, "The queue size of the collector")
 	flags.Uint(flagDynQueueSizeMemory, 0, "(experimental) The max memory size in MiB to use for the dynamic queue.")
 	flags.String(flagCollectorTags, "", "One or more tags to be added to the Process tags of all spans passing through this collector. Ex: key1=value1,key2=${envVar:defaultValue}")
+	flags.Bool(flagProcessedSpanMetricsEnabled, false, "Enables processed span metrics.")
```
yurishkuro, line 169:
```suggestion
	flags.Bool(flagProcessedSpanMetricsEnabled, false, "Enables metrics based on processed span size, which are more expensive to calculate.")
```

cmd/collector/app/flags/flags.go
```diff
@@ -124,6 +125,8 @@ type CollectorOptions struct {
 	}
 	// CollectorTags is the string representing collector tags to append to each and every span
 	CollectorTags map[string]string
+	// ProcessedSpanMetricsEnabled determines whether to enable processed span metrics
+	ProcessedSpanMetricsEnabled bool
```
yurishkuro, line 129:
```suggestion
	SpanSizeMetricsEnabled bool
```

cmd/collector/app/flags/flags.go
```diff
@@ -30,10 +30,11 @@ import (
...
 const (
-	flagDynQueueSizeMemory = "collector.queue-size-memory"
-	flagNumWorkers         = "collector.num-workers"
-	flagQueueSize          = "collector.queue-size"
-	flagCollectorTags      = "collector.tags"
+	flagDynQueueSizeMemory          = "collector.queue-size-memory"
+	flagNumWorkers                  = "collector.num-workers"
+	flagQueueSize                   = "collector.queue-size"
+	flagCollectorTags               = "collector.tags"
+	flagProcessedSpanMetricsEnabled = "collector.enable-span-size-metrics"
```
yurishkuro, line 37:
```suggestion
	flag SpanSizeMetricsEnabled = "collector.enable-span-size-metrics"
```
let's keep consistent naming

cmd/collector/app/options.go
```diff
@@ -26,21 +26,22 @@ import (
...
+	spanFilter                  FilterSpan
+	numWorkers                  int
+	blockingSubmit              bool
+	queueSize                   int
+	dynQueueSizeWarmup          uint
+	dynQueueSizeMemory          uint
+	reportBusy                  bool
+	extraFormatTypes            []processor.SpanFormat
+	collectorTags               map[string]string
+	processedSpanMetricsEnabled bool
```
yurishkuro, line 44: please keep naming consistent

cmd/collector/app/options.go
```diff
@@ -156,6 +157,13 @@ func (options) CollectorTags(extraTags map[string]string) Option {
 	}
 }
 
+// ProcessedSpanMetricsEnabled creates an Option that initializes the processedSpanMetrics boolean
+func (options) ProcessedSpanMetricsEnabled(processedSpanMetrics bool) Option {
```
yurishkuro, line 161: please keep naming consistent

cmd/collector/app/flags/flags.go
ymtdzzz: oh ok, I will check it all again

cmd/collector/app/span_processor_test.go
```diff
@@ -400,27 +400,74 @@ func TestSpanProcessorWithCollectorTags(t *testing.T) {
...
-		_, g := mb.Snapshot()
-		if b := g["spans.bytes"]; b > 0 {
-			assert.Equal(t, p.bytesProcessed.Load(), uint64(g["spans.bytes"]))
-			return
-		}
-		time.Sleep(time.Millisecond)
+			if tt.expectedUpdateGauge {
+				assert.Fail(t, "gauge hasn't been updated within a reasonable amount of time")
+			}
+			assert.NoError(t, p.Close())
```
yurishkuro, line 468: this won't be executed if you return in L460. I suggest using defer to close p.

cmd/collector/app/span_processor_test.go
```diff
@@ -400,27 +400,74 @@ func TestSpanProcessorWithCollectorTags(t *testing.T) {
...
-	mb := metricstest.NewFactory(time.Hour)
-	m := mb.Namespace(metrics.NSOptions{})
+	tests := []struct {
+		name                  string
+		enableDynQueueSizeMem bool
+		enableSpanMetrics     bool
+		expectedUpdateGauge   bool
+	}{
+		{
+			name:                  "enable-dyn-queue-size-enable-metrics",
```
yurishkuro, line 410:
```suggestion
			name:                  "enable dyn-queue-size, enable metrics",
```
nit: remove dashes for better readability, separate with commas

cmd/collector/app/span_processor_test.go
```diff
@@ -400,27 +400,74 @@ func TestSpanProcessorWithCollectorTags(t *testing.T) {
...
+			} else {
+				opts = append(opts, Options.DynQueueSizeMemory(0))
+			}
+			p := NewSpanProcessor(w, nil, opts...).(*spanProcessor)
+			p.background(10*time.Millisecond, p.updateGauges)
+
+			p.processSpan(&model.Span{}, "")
+			assert.NotEqual(t, uint64(0), p.bytesProcessed)
+
+			for i := 0; i < 15; i++ {
```
yurishkuro, line 453:
```suggestion
			for i := 0; i < 10000; i++ {
```
doesn't affect test speed in most cases, but makes it more resilient in case of VM freezes

yurishkuro (approved):
thanks!

PR #3786, opened 2022-06-28
Introduce ParentReference adjuster

bobrik:
As demonstrated in https://github.com/jaegertracing/jaeger-ui/issues/966, jaeger-ui expects the first reference to be of `CHILD_OF` type to properly render the tree. This new adjuster achieves exactly that.

### DIFF
--- a/model/adjuster/parent_reference.go
+++ b/model/adjuster/parent_reference.go
@@ -1,0 +1,52 @@
+// Copyright (c) 2022 The Jaeger Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package adjuster
+
+import (
+	"github.com/jaegertracing/jaeger/model"
+)
+
+// ParentReference returns an Adjuster that puts CHILD_OF references first.
+// This is necessary to match jaeger-ui expectations:
+// * https://github.com/jaegertracing/jaeger-ui/issues/966
+func ParentReference() Adjuster {
+	return Func(func(trace *model.Trace) (*model.Trace, error) {
+		for _, span := range trace.Spans {
+			if len(span.References) == 0 {
+				continue
+			}
+
+			for i := range span.References {
+				if i == 0 {
+					continue
+				}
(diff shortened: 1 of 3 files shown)


model/adjuster/sort_references.go
```diff
@@ -0,0 +1,38 @@
+// Copyright (c) 2019 The Jaeger Authors.
+// Copyright (c) 2017 Uber Technologies, Inc.
```
yurishkuro, line 2:
```suggestion
// Copyright (c) 2022 The Jaeger Authors.
```

model/adjuster/sort_references.go
```diff
@@ -0,0 +1,38 @@
...
+// SortReferences returns an Adjuster that puts CHILD_OF references first.
+// This is necessary to match jaeger-ui expectations:
+// * https://github.com/jaegertracing/jaeger-ui/issues/966
+func SortReferences() Adjuster {
+	return Func(func(trace *model.Trace) (*model.Trace, error) {
+		for _, span := range trace.Spans {
+			sort.SliceStable(span.References, func(i, j int) bool {
+				iType := span.References[i].GetRefType()
+				jType := span.References[j].GetRefType()
+				return iType == model.SpanRefType_CHILD_OF && jType == model.SpanRefType_FOLLOWS_FROM
```
yurishkuro, line 33: I don't think this is a sufficient condition. The logic to pick the correct parent should be:
  * first ref with refType=CHILD_OF and the same trace ID
  * otherwise, first ref that has the same trace ID
  * otherwise no change

Also, I would avoid the actual sorting, and only do the swap into the first position.

model/adjuster/sort_references.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"sort"
+
+	"github.com/jaegertracing/jaeger/model"
+)
+
+// SortReferences returns an Adjuster that puts CHILD_OF references first.
+// This is necessary to match jaeger-ui expectations:
+// * https://github.com/jaegertracing/jaeger-ui/issues/966
+func SortReferences() Adjuster {
```
yurishkuro, line 27:
```suggestion
func ParentReference() Adjuster {
```
I don't think we need to sort, only swap the parent into the first position.

model/adjuster/sort_references.go
bobrik: Makes sense, updated.

model/adjuster/parent_reference.go
```diff
@@ -0,0 +1,52 @@
...
+			if len(span.References) == 0 {
+				continue
+			}
+
+			for i := range span.References {
+				if i == 0 {
+					continue
+				}
+
+				if span.References[0].RefType == model.SpanRefType_CHILD_OF && span.References[0].TraceID == span.TraceID {
```
yurishkuro, line 36: this could be outside the loop

yurishkuro (approved):

PR #3787, opened 2022-06-29
Ignore the problem of self-reported spans when multi-tenant enabled

esnible:

If the experimental tenancy feature is enabled, every self-reported span generates two **error** level log entries with full stack traces.

This PR causes Jaeger to log missing tenancy at **debug** level.

The goal is to mitigate the problem of self-reports while we focus on adding tenancy to the query APIs.

See https://github.com/jaegertracing/jaeger/pull/3750#issuecomment-1151819300

### DIFF
--- a/cmd/agent/app/reporter/grpc/reporter.go
+++ b/cmd/agent/app/reporter/grpc/reporter.go
@@ -20,6 +20,8 @@ import (
 
 	"go.uber.org/zap"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 
 	zipkin2 "github.com/jaegertracing/jaeger/cmd/collector/app/sanitizer/zipkin"
 	"github.com/jaegertracing/jaeger/model"
@@ -71,7 +73,12 @@ func (r *Reporter) send(ctx context.Context, spans []*model.Span, process *model
 	req := &api_v2.PostSpansRequest{Batch: batch}
 	_, err := r.collector.PostSpans(ctx, req)
 	if err != nil {
-		r.logger.Error("Could not send spans over gRPC", zap.Error(err))
+		stat, ok := status.FromError(err)
+		if ok && stat.Code() == codes.PermissionDenied && stat.Message() == "missing tenant header" {
+			r.logger.Debug("Could not send spans over gRPC", zap.Error(err))
+		} else {
+			r.logger.Error("Could not send spans over gRPC", zap.Error(err))
+		}
 	}
 	return err
 }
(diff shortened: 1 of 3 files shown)


yurishkuro (dismissed):

cmd/agent/app/reporter/grpc/reporter.go
```diff
@@ -71,7 +73,12 @@ func (r *Reporter) send(ctx context.Context, spans []*model.Span, process *model
 	req := &api_v2.PostSpansRequest{Batch: batch}
 	_, err := r.collector.PostSpans(ctx, req)
 	if err != nil {
-		r.logger.Error("Could not send spans over gRPC", zap.Error(err))
+		stat, ok := status.FromError(err)
+		if ok && stat.Code() == codes.PermissionDenied && stat.Message() == "missing tenant header" {
+			r.logger.Debug("Could not send spans over gRPC", zap.Error(err))
```
pavolloffay, line 78: It's worth distinguishing the error messages a bit more, not just by the error level.

cmd/agent/app/reporter/grpc/reporter_test.go
```diff
@@ -178,3 +181,49 @@ func TestReporter_MakeModelKeyValue(t *testing.T) {
...
+func TestReporter_MultitenantEmitBatch(t *testing.T) {
+	handler := &mockMultitenantSpanHandler{}
+	s, addr := initializeGRPCTestServer(t, func(s *grpc.Server) {
+		api_v2.RegisterCollectorServiceServer(s, handler)
+	})
+	defer s.Stop()
+	conn, err := grpc.Dial(addr.String(), grpc.WithTransportCredentials(insecure.NewCredentials()))
+	//nolint:staticcheck // don't care about errors
+	defer conn.Close()
+	require.NoError(t, err)
```
yurishkuro, line 212:
```suggestion
	conn, err := grpc.Dial(addr.String(), grpc.WithTransportCredentials(insecure.NewCredentials()))
	require.NoError(t, err)
	defer func() { require.NoError(t, conn.Close()) }()
```

cmd/agent/app/reporter/grpc/reporter_test.go
```diff
@@ -178,3 +181,49 @@ func TestReporter_MakeModelKeyValue(t *testing.T) {
...
+	rep := NewReporter(conn, nil, zap.NewNop())
+
+	tm := time.Unix(158, 0)
```
yurishkuro, line 215: why not `time.Now()`?

pavolloffay (approved):

PR #3791, opened 2022-07-01
Tenancy for queries

esnible:

This PR replaces https://github.com/jaegertracing/jaeger/pull/3735

- This PR flows the tenant header through the query APIs.
- This PR blocks requests with invalid tenant or missing tenant header from reaching the query APIs.
- This PR does **NOT** do authorization.  It is assumed authorization is handled by a reverse proxy or other non-Jaeger component at this time.
- This PR requires actual distinguishing of the tenant to be done by the storage layer.  The purpose of this PR is to be sure the storage layer receives a tenant that is valid when Jaeger is started with `--multi_tenant.enabled=true`.

cc @pavolloffay


cmd/query/app/apiv3/grpc_gateway.go
```diff
@@ -25,16 +25,32 @@ import (
...
+	"github.com/jaegertracing/jaeger/pkg/config/tenancy"
 	"github.com/jaegertracing/jaeger/pkg/config/tlscfg"
 	"github.com/jaegertracing/jaeger/proto-gen/api_v3"
 )
 
 // RegisterGRPCGateway registers api_v3 endpoints into provided mux.
 func RegisterGRPCGateway(ctx context.Context, logger *zap.Logger, r *mux.Router, basePath string, grpcEndpoint string, grpcTLS tlscfg.Options) error {
+	return RegisterGRPCGatewayWithTenancy(ctx, logger, r, basePath, grpcEndpoint, grpcTLS,
+		tenancy.Options{
+			Enabled: false,
```
pavolloffay, line 37: why is this hardcoded? It deserves a comment

pavolloffay: making this API more configurable seems better. The function has also a lot of parameters, maybe having a params struct would simplify it as well.

pkg/config/tenancy/context.go
```diff
@@ -0,0 +1,81 @@
...
+)
+
+var emptyMD = metadata.New(map[string]string{})
+
+func tenantFromMetadata(md metadata.MD, tenancyHeader string) (string, error) {
+	tenants := md.Get(tenancyHeader)
+	if len(tenants) < 1 {
+		return "", status.Errorf(codes.PermissionDenied, "missing tenant header")
+	} else if len(tenants) > 1 {
+		return "", status.Errorf(codes.PermissionDenied, "extra tenant header")
```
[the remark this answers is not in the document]
esnible: What values are you thinking?  Typically PermissionDenied messages are cryptic, as they often get sent to the client, who is only supposed to know enough to fix the problem, but not enough to guess how to get around the security.

pkg/config/tenancy/context.go
```diff
@@ -0,0 +1,81 @@
...
+
+// MetadataAnnotator returns a function suitable for propagating tenancy
+// via github.com/grpc-ecosystem/grpc-gateway/runtime.NewServeMux
+func (tc *TenancyConfig) MetadataAnnotator() func(context.Context, *http.Request) metadata.MD {
+	return func(ctx context.Context, req *http.Request) metadata.MD {
+		tenant := req.Header.Get(tc.Header)
+		if tenant == "" {
+			return emptyMD
+		}
+		return metadata.New(map[string]string{
```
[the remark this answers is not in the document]
esnible: The /v3 HTTP API is implemented by calling the gRPC implementation.  I believe you did this work.  MetadataAnnotator() is a new middleware that will be called once per v3 query.

I believe github.com/grpc-ecosystem/grpc-gateway, or possibly Go's GRPC itself, takes care of merging all gRPC metadata together, so there shouldn't be a concern about other values.

If you are concerned about a `map` allocation per query these could be cached by the middleware.

pkg/config/tenancy/context.go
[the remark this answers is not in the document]
pavolloffay: We could put there the list of tenants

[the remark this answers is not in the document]
esnible: That's a security no-no.  We can't say 'You identified as tenant "dummy", to use this site you must represent yourself as "ibm", "redhat", or "google".'  If we give any information as part of PermissionDenied we will fail any kind of security audit or pen test.

pavolloffay (approved):

pkg/config/tenancy/interceptor.go
```diff
@@ -0,0 +1,73 @@
...
+type tenantedServerStream struct {
+	grpc.ServerStream
+	context context.Context
+}
+
+func (tss *tenantedServerStream) Context() context.Context {
+	return tss.context
+}
+
+// NewGuardingStreamInterceptor blocks handling of streams whose tenancy header isn't doesn't meet tenancy requirements
```
yurishkuro, line 38:
```suggestion
// NewGuardingStreamInterceptor blocks handling of streams whose tenancy header doesn't meet tenancy requirements.
```

pkg/config/tenancy/tenancy.go
```diff
@@ -35,9 +35,14 @@ type Options struct {
 
 // NewTenancyConfig creates a tenancy configuration for tenancy Options
 func NewTenancyConfig(options *Options) *TenancyConfig {
```
yurishkuro, line 37: this file would be better named `config.go`

pkg/config/tenancy/context.go
```diff
@@ -0,0 +1,81 @@
...
+	})
+}
+
+// MetadataAnnotator returns a function suitable for propagating tenancy
+// via github.com/grpc-ecosystem/grpc-gateway/runtime.NewServeMux
+func (tc *TenancyConfig) MetadataAnnotator() func(context.Context, *http.Request) metadata.MD {
+	return func(ctx context.Context, req *http.Request) metadata.MD {
+		tenant := req.Header.Get(tc.Header)
+		if tenant == "" {
+			return emptyMD
```
yurishkuro, line 75: this feels dangerous, isn't it possible to have some other interceptor etc. in the chain that will try to modify metadata, such that the singleton won't be appropriate?

is it possible to simply return nil?

pkg/config/tenancy/interceptor.go
```diff
@@ -0,0 +1,73 @@
+// Copyright (c) 2022 The Jaeger Authors.
```
yurishkuro, line 1: it's weird that we have interceptors and other functionality within package `config/*`. If config is only one of tenancy-related functions, then I would suggest:
```
pkg/
    tenancy/
        config.go (for TenancyConfig and validators)
        context.go (instead of being under storage/)
        grpc.go (for grpc-related pieces)
        http.go (for http-related pieces)
```

cmd/query/app/apiv3/grpc_gateway.go
```diff
@@ -25,16 +25,24 @@ import (
...
-func RegisterGRPCGateway(ctx context.Context, logger *zap.Logger, r *mux.Router, basePath string, grpcEndpoint string, grpcTLS tlscfg.Options) error {
+func RegisterGRPCGateway(ctx context.Context, logger *zap.Logger, r *mux.Router, basePath string, grpcEndpoint string, grpcTLS tlscfg.Options, tenancyOptions tenancy.Options) error {
 	jsonpb := &runtime.JSONPb{}
-	grpcGatewayMux := runtime.NewServeMux(
+
+	muxOpts := []runtime.ServeMuxOption{
 		runtime.WithMarshalerOption(runtime.MIMEWildcard, jsonpb),
-	)
+	}
+	tc := tenancy.NewTenancyConfig(&tenancyOptions)
```
yurishkuro, line 40: why do we transform options to config here instead of doing once elsewhere and the reusing config across different handlers?

pkg/config/tenancy/context.go
esnible: Can't pass `nil` because grpc-gateway's `runtime.WithMetadata()` expects a function that takes an `MD`, not a `*MD`.  With the current implementation of grpc-gateway returning a singleton isn't a problem, but there is no guarantee of that, and I will pass along a new MD.  This is only called if tenancy is enabled, and it will be a rare case that tenancy is enabled and clients aren't sending tenancy information.

storage/empty_test.go
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) 2018 The Jaeger Authors.
```
yurishkuro, line 1: Did you mean to also delete `storage/tenant.go`?

cmd/query/app/server.go
```diff
@@ -160,7 +161,8 @@ func createHTTPServer(querySvc *querysvc.QueryService, metricsQuerySvc querysvc.
 	}
 
 	ctx, closeGRPCGateway := context.WithCancel(context.Background())
-	if err := apiv3.RegisterGRPCGateway(ctx, logger, r, queryOpts.BasePath, queryOpts.GRPCHostPort, queryOpts.TLSGRPC); err != nil {
+	tc := tenancy.NewTenancyConfig(&queryOpts.Tenancy)
```
yurishkuro, line 164: what about the API handler above, I assume it should also be wrapped in ExtractTenant

storage/empty_test.go
yurishkuro: +1

cmd/query/app/server.go
esnible: Thanks for your review.  *This comment was extremely helpful and identified several flaws.*

I have fixed the problems, but please don't merge yet! ;-)

The problems were not detected for two reasons.  I was testing with GetTrace, a streaming RPC, and ignoring tenancy for GetService, a unary RPC.  This PR needs unary RPC tests.  I have added one but need to increase coverage.
The other problem is that the tests set up the server using `newGRPCServer()` in cmd/query/app/grpc_handler_test.go, but the actual Jaeger uses `createGRPCServer()` which is nearly the same in implementation but was not tested for tenancy and didn't work.

I want to do at least one more commit of additional tests.

cc @pavolloffay

esnible:
@pavolloffay I have added extensive tests.  Please re-review.

cmd/query/app/server.go
pavolloffay: I am not familiar with `newGRPCServer` nor `createGRPCServer` but if there is a way to unify these APIs we should consider doing it (e.g. in a follow up PR).

cmd/query/app/grpc_handler_test.go
```diff
@@ -219,7 +228,7 @@ func initializeTestServerGRPCWithOptions(t *testing.T, options ...testOption) *g
 	logger := zap.NewNop()
 	tracer := opentracing.NoopTracer{}
 
-	server, addr := newGRPCServer(t, q, tqs.metricsQueryService, logger, tracer)
+	server, addr := newGRPCServer(t, q, tqs.metricsQueryService, logger, tracer, &tenancy.TenancyConfig{})
```
pavolloffay, line 231: Shouldn't be `tenancyConfig` passed instead of creating a new one?

pavolloffay (approved):

cmd/query/app/grpc_handler_test.go
esnible: Now fixed, but of course that dismissed your review....

pavolloffay (approved):

pavolloffay (dismissed):

pkg/tenancy/http.go
```diff
@@ -0,0 +1,66 @@
...
+	"net/http"
+
+	"go.uber.org/zap"
+	"google.golang.org/grpc/metadata"
+)
+
+// PropagationHandler returns a http.Handler containing the logic to extract
+// the tenancy header of the http.Request and insert the tenant into request.Context
+// for propagation. The token can be accessed via tenancy.GetTenant().
+func ExtractTenantHTTPHandler(tc *TenancyConfig, h http.Handler, logger *zap.Logger) http.Handler {
```
yurishkuro, line 28: the logger is not used, let's remove it

cmd/query/app/http_handler.go
```diff
@@ -139,9 +144,14 @@ func (aH *APIHandler) handleFunc(
 	args ...interface{},
 ) *mux.Route {
 	route = aH.route(route, args...)
+	var handler http.Handler
+	handler = http.HandlerFunc(f)
+	if aH.tenancyConfig.Enabled {
+		handler = tenancy.ExtractTenantHTTPHandler(aH.tenancyConfig, handler, zap.NewNop())
```
yurishkuro, line 150:
```suggestion
		handler = tenancy.ExtractTenantHTTPHandler(aH.tenancyConfig, handler, ah.logger)
```

cmd/query/app/handler_options.go
```diff
@@ -61,6 +62,13 @@ func (handlerOptions) QueryLookbackDuration(queryLookbackDuration time.Duration)
 	}
 }
 
+// Tracer creates a HandlerOption that initializes tenancy
+func (handlerOptions) Tenancy(options *tenancy.Options) HandlerOption {
```
yurishkuro, line 66: why pass options instead of tenancy.Config?

cmd/query/app/server.go
yurishkuro: > `tc := tenancy.NewTenancyConfig(&queryOpts.Tenancy)`

there are way too many places doing this conversion - we should only create tenancy config once (in the future it could be "live" object that self-updates).

pkg/tenancy/grpc.go
```diff
@@ -0,0 +1,114 @@
...
+
+	return tenants[0], nil
+}
+
+// NewGuardingUnaryInterceptor blocks handling of RPCs whose tenancy header doesn't meet tenancy requirements.
+// It also ensures the tenant is directly in the context, rather than context metadata.
+func NewGuardingUnaryInterceptor(tc *TenancyConfig) grpc.UnaryServerInterceptor {
+	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+		// Handle case where tenant is directly in the context
+		tenant := GetTenant(ctx)
```
yurishkuro, line 89: code below is identical to the previous function, please move to a single helper function like
```
tenant, err := GetAndValidateTenant(ctx)
if err != nil { return err }
if tenant == "" {
  return handler()
} else {
  return wrappedHandler()
}
```

cmd/query/app/server.go
esnible: @yurishkuro. I want to get this right.

I was using TenancyOptions for something human-readable, for Viper, and TenancyConfig for something machine readable, optimized for fast checking, following the pattern of `tlscfg.Options` and `tls.Config`.  Would it make sense to change tenency.TenancyConfig to `tenancy.TenancyManager`?

Currently the PR converts tenancy.Options to TenancyConfig in 4  non-test places.  Twice in `createGRPCServer()` -- I will fix that, once in `createHTTPServer()`, and once in `BuildHandlers()` (already merged; for trace ingestion).  Currently TenancyConfig is lightweight, and I didn't want to change the signatures to the create*Server() functions.  If we think of Tenancy is a "service" then I should create it once.  The next commit will address this.

yurishkuro: `TenancyConfig` is different from `tls.Config` in that the former is an object with behavior (an internal "service"), but the latter is just data passed elsewhere. I think from the naming it would be clearer if it was called `tenancy.TenancyManager`.

> Currently TenancyConfig is lightweight

It's only lightweight in the current implementation. Because it's a service, it can become arbitrarily complicated in the future, e.g. it could start refreshing the list of tenants from some storage, etc. By instantiating it in many places we introduce technical debt from the start that will need to be cleaned up later, so I'd rather fix it now and create this service once.

cmd/collector/app/span_handler_builder.go
```diff
@@ -76,7 +76,7 @@ func (b *SpanHandlerBuilder) BuildHandlers(spanProcessor processor.SpanProcessor
 			zs.NewChainedSanitizer(zs.NewStandardSanitizers()...),
 		),
 		handler.NewJaegerSpanHandler(b.Logger, spanProcessor),
-		handler.NewGRPCHandler(b.Logger, spanProcessor, tenancy.NewTenancyConfig(&b.CollectorOpts.GRPC.Tenancy)),
+		handler.NewGRPCHandler(b.Logger, spanProcessor, tenancy.NewTenancyManager(&b.CollectorOpts.GRPC.Tenancy)),
```
yurishkuro, line 79: we should treat TM similar to Logger - pass it to SpanHandlerBuilder at construction.

another peculiar thing: why is it `b.CollectorOpts.GRPC.Tenancy`? I.e. why tenancy options are within GRPC options?

cmd/query/app/http_handler.go
```diff
@@ -84,6 +85,7 @@ type APIHandler struct {
 	queryService        *querysvc.QueryService
 	metricsQueryService querysvc.MetricsQueryService
 	queryParser         queryParser
+	tenancyConfig       *tenancy.TenancyManager
```
yurishkuro, line 88:
```suggestion
	tenancyMgr       *tenancy.TenancyManager
```

cmd/query/app/http_handler.go
```diff
@@ -112,6 +114,9 @@ func NewAPIHandler(queryService *querysvc.QueryService, options ...HandlerOption
 	if aH.tracer == nil {
 		aH.tracer = opentracing.NoopTracer{}
 	}
+	if aH.tenancyConfig == nil {
```
yurishkuro, line 117: realistically, will we ever be in a situation where this is nil?  `cmd/query/app/server.go` does pass this option.

cmd/query/app/server.go
```diff
@@ -119,6 +120,12 @@ func createGRPCServer(querySvc *querysvc.QueryService, metricsQuerySvc querysvc.
 
 		grpcOpts = append(grpcOpts, grpc.Creds(creds))
 	}
+	if options.Tenancy.Enabled {
```
yurishkuro, line 123: you probably want `tm.Enabled` here

pkg/tenancy/config.go
```diff
@@ -33,16 +33,21 @@ type Options struct {
 	Tenants []string
 }
 
-// NewTenancyConfig creates a tenancy configuration for tenancy Options
-func NewTenancyConfig(options *Options) *TenancyConfig {
-	return &TenancyConfig{
+// NewTenancyManager creates a tenancy configuration for tenancy Options
```
yurishkuro, line 36:
```suggestion
// NewTenancyManager creates TenancyManager from tenancy Options
```

pkg/tenancy/grpc_test.go
```diff
@@ -0,0 +1,169 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+)
+
+func TestStreamingTenancyInterceptor(t *testing.T) {
+	tests := []struct {
```
yurishkuro, line 28: Is there a difference between these test cases and those in the next test?  Could we, perhaps, pull them out into a shared struct or function, and only differ in the business logic?

cmd/collector/app/handler/grpc_handler.go
```diff
@@ -58,12 +58,12 @@ type batchConsumer struct {
...
 	spanOptions   processor.SpansOptions
-	tenancyConfig tenancy.TenancyConfig
+	tenancyMgr    *tenancy.TenancyManager
 }
 
-func newBatchConsumer(logger *zap.Logger, spanProcessor processor.SpanProcessor, transport processor.InboundTransport, spanFormat processor.SpanFormat, tenancyConfig *tenancy.TenancyConfig) batchConsumer {
-	if tenancyConfig == nil {
-		tenancyConfig = &tenancy.TenancyConfig{}
+func newBatchConsumer(logger *zap.Logger, spanProcessor processor.SpanProcessor, transport processor.InboundTransport, spanFormat processor.SpanFormat, tenancyMgr *tenancy.TenancyManager) batchConsumer {
+	if tenancyMgr == nil {
```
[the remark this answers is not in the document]
esnible: This line only became part of the PR because of the rename from `TenancyConfig` to `TenancyManager`.  It was already merged this way.  _oltp_receiver.go_:146 actually does pass in nil.  My thought at that time was to default to an empty handler.  If you would prefer it to be explicit I will also add an `error` return and fail if `nil` is passed in.  Thoughts?

cmd/collector/app/span_handler_builder.go
esnible: This line only became part of the PR because of the rename from `TenancyConfig` to `TenancyManager`.  The collector doesn't use the interceptors, which were introduced for query.  (I tend to use interceptors as a last resort, as many devs struggle with them.)  Now that I introduced them for query it makes sense to use them for collection as well, and I will refactor this code to use them.

cmd/query/app/http_handler.go
esnible: I am happy to remove the defaulting; I will add an error return at the same time.

cmd/collector/app/handler/grpc_handler.go
```diff
@@ -104,7 +104,7 @@ func (c *batchConsumer) consume(ctx context.Context, batch *model.Batch) error {
 }
 
 func (c *batchConsumer) validateTenant(ctx context.Context) (string, error) {
```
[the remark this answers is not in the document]
esnible: This line only became part of the PR because of the rename from `TenancyConfig` to `TenancyManager`. The collector doesn't use the interceptors, which were introduced for query.  This PR is large; I propose backporting the new query interceptor to the collector should be in a follow-up.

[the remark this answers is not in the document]
yurishkuro: +1

cmd/collector/app/collector.go
```diff
@@ -94,6 +95,7 @@ func (c *Collector) Start(options *flags.CollectorOptions) error {
 		CollectorOpts:  options,
 		Logger:         c.logger,
 		MetricsFactory: c.metricsFactory,
+		TenancyMgr:     tenancy.NewTenancyManager(&options.GRPC.Tenancy),
```
yurishkuro, line 98: when collector is instantiated from all-in-one, will this cause multiple tenancy managers to exist?

yurishkuro (commented):
I think this PR is mostly ok to merge, but I would still prefer it is cleaned up to only instantiate tenancy manager in main's. I see at least two places where it's being created adhoc in the components that should instead be receiving it as a dependency.

esnible:
@yurishkuro Thank you for your patience.  I was timid.  With the latest commit, only the _main.go_ and test cases create TenancyManager.  Thus there will now only be one of them even in the all-in-one case.

yurishkuro (approved):
🥳

Issue #3804, opened 2022-07-08
[Feature]: Visualize uninstrumented services in the dependency diagrams


yurishkuro:
### Requirement

Visualize services in the dependency diagram even when they are not instrumented, but known from the caller side.

### Problem

When the trace leaf nodes that represent outbound calls to uninstrumented services, those services are not shown in the dependency diagram (e.g. see how Zipkin shows them in #3803).

### Proposal

Jaeger can infer that there is an existing callee service when the caller service logs a span with tag `span.kind=client` without the corresponding `span.kind=server` span.

There are several places in the code base where this will need to be accounted for:
  * in the in-memory storage used by all-in-one (the easiest to start with)
  * in the Flink/Spark jobs for production usage

Aside from changing the graph logic, another alternative is to have a trace enrichment which will add artificial `server` spans to the trace. Then the graph building logic would not need to change at all, and the inferred nodes could also be shown in the single-trace views.

### Open questions

Deciding what to call the missing callee services can be tricky. We will need to implement a heuristic that derives the name from some of the tags of the client span:
  * based on OpenTracing [semantic conventions](https://github.com/opentracing/specification/blob/master/semantic_conventions.md)
    * `peer.service`
    * `peer.address`
    * `peer.ip?` + `peer.port`
  * based on similar OTEL semantic conventions

There was a discussion in OTEL once about labeling the type of downstream service (e.g. an SQL db, etc), which could also be taken into account when naming the derived services.

paule96:
Hi @yurishkuro,

I like your idea to just add it to the trace enrichment. Can you help me where to find the code of this? So I can try my luck to implement it? :)
From a first look into the code I didn't find the right place.

yurishkuro:
@paule96 the dependencies calculation done by all-in-one happens inside `memory` storage:

https://github.com/jaegertracing/jaeger/blob/711488f94f25d4a5e7900d0ba450ecb5bfdfda7b/plugin/storage/memory/memory.go#L63

It may not be very straightforward to combine this with the enricher idea because dependencies logic does not query for traces, it just accesses them directly.

pranoyk:
@yurishkuro I am planning to look into this issue. Will share my findings here and also create a WIP PR. 
Also, do let me know if I should be aware of anything to make changes for this issue

nidhey27:
Hi @yurishkuro,
I noticed that this issue is still open. Is it currently being worked on? If not, I would like to take this up and contribute a fix.
[...]

PR #3805, opened 2022-07-10
[WIP] Visualize uninstrumented services in the dependency diagrams

paule96:

This PR resolves the feature #3804 for the memory storage.

- Add some development setup changes to make it easy for new developers on this project to just check out hit F5 and be up and running
   - The developer now needs just to clone the repository in container volumne with the help of the `remote containers development` extension of vscode.
   - To start debugging he just needs to hit F5
   - It's not required for the developer to install any tools on his PC -> only docker is required
   - It also configures the recommended settings from the contribution guide
- add the peer detection in the memory storage provider

### DIFF
--- a/plugin/storage/memory/memory.go
+++ b/plugin/storage/memory/memory.go
@@ -71,6 +71,20 @@ func (m *Store) GetDependencies(ctx context.Context, endTs time.Time, lookback t
 		trace, _ := m.deduper.Adjust(orig)
 		if m.traceIsBetweenStartAndEnd(startTs, endTs, trace) {
 			for _, s := range trace.Spans {
+				// this finds unmanged peers of the service. So this are all external dependcies
+				peer := m.findPeer(s)
+				if peer != nil {
+					depKey := s.Process.ServiceName + "&&&" + peer.VStr
+					if _, ok := deps[depKey]; !ok {
+						deps[depKey] = &model.DependencyLink{
+							Parent:    s.Process.ServiceName,
+							Child:     peer.VStr,
+							CallCount: 1,
+						}
+					} else {
+						deps[depKey].CallCount++
+					}
+				}
 				parentSpan := m.findSpan(trace, s.ParentSpanID())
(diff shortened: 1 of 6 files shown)


plugin/storage/memory/memory.go
```diff
@@ -71,6 +71,20 @@ func (m *Store) GetDependencies(ctx context.Context, endTs time.Time, lookback t
 		trace, _ := m.deduper.Adjust(orig)
 		if m.traceIsBetweenStartAndEnd(startTs, endTs, trace) {
 			for _, s := range trace.Spans {
+				// this finds unmanged peers of the service. So this are all external dependcies
+				peer := m.findPeer(s)
```
yurishkuro, line 75: This will not address the issue in the ticket. While peer.service is a recommended tag, it's often impossible for the instrumentation to know.

paule96: How you mean? What is the case this tag doesn't catch? And shouldn't this be more an issue for the collector? Shouldn't the observability layer trust the open telemetry spec?

paule96: Or did you referring too the other informations that the open telemetry gives for peers?
Something like

- peer.port
- peer.ipv6
- peer.ipv4
- peer.hostname
- peer.address

So we should combine all these information for the key?

yurishkuro: There are two issues.
  1. You're force-adding a dependency based on the presence of `peer.service` tag regardless of whether the current node is the leaf in the call graph or an inner node that would produce a dependency anyway from its child span.
  2. None of the tags you listed are actually _required_ to be present in the span. In a typical HTTP client instrumentation, the instrumentation layer does not know the name of the downstream service, only its URL, so `peer.service` tag is likely to be missing, while other `peer.*` tags might be present. We would need some heuristic function that would decide how to name the downstream service in this case.

yurishkuro:
closing stale PR

PR #3813, opened 2022-07-13
Add capability to build docker all-in-one

albertteoh:
- As a developer, I want to quickly test changes to SPM locally, which depends on the `jaegertracing/all-in-one:latest` image.
- Building this image is currently not supported in the current Makefile.

- Add `all-in-one` component to the list of components to build docker images for.
- Note: I don't believe this impacts existing CI builds, which utilise the `scripts/build-all-in-one-image.sh` for all-in-one, and `scripts/build-upload-docker-images.sh` for other components. These scripts, in turn, execute their own docker commands (not sure why, instead of using Makefile targets).

## Testing

- Used to locally test changes from https://github.com/jaegertracing/jaeger-ui/pull/971.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -308,21 +308,19 @@ docker-images-jaeger-backend-debug: SUFFIX = -debug
 .PHONY: docker-images-jaeger-backend docker-images-jaeger-backend-debug
 docker-images-jaeger-backend docker-images-jaeger-backend-debug: create-baseimg create-debugimg
 	for component in "jaeger-agent" "jaeger-collector" "jaeger-query" "jaeger-ingester" "all-in-one" ; do \
-		regex="(jaeger-)(.*)"; \
-		component_prefix=""; \
+		regex="jaeger-(.*)"; \
 		component_suffix=$$component; \
 		if [[ $$component =~ $$regex ]]; then \
-			component_prefix="$${BASH_REMATCH[1]}"; \
-			component_suffix="$${BASH_REMATCH[2]}"; \
+			component_suffix="$${BASH_REMATCH[1]}"; \
 		fi; \
 		docker buildx build --target $(TARGET) \
-			--tag $(DOCKER_NAMESPACE)/$$component_prefix$$component_suffix$(SUFFIX):${DOCKER_TAG} \
+			--tag $(DOCKER_NAMESPACE)/$$component$(SUFFIX):${DOCKER_TAG} \
 			--build-arg base_image=$(BASE_IMAGE) \
 			--build-arg debug_image=$(DEBUG_IMAGE) \
 			--build-arg TARGETARCH=$(GOARCH) \
 			--load \
-			cmd/$$component_suffix ; \
-		echo "Finished building $$component_suffix ==============" ; \
+			cmd/$$component_suffix; \
+		echo "Finished building $$component ==============" ; \
 	done; \
 
 .PHONY: docker-images-tracegen
--- a/Makefile
+++ b/Makefile
@@ -306,15 +306,26 @@ docker-images-jaeger-backend-debug: SUFFIX = -debug
 
 .PHONY: docker-images-jaeger-backend docker-images-jaeger-backend-debug
 docker-images-jaeger-backend docker-images-jaeger-backend-debug: create-baseimg create-debugimg
-	for component in agent collector query ingester ; do \
-		docker build --target $(TARGET) \
-			--tag $(DOCKER_NAMESPACE)/jaeger-$$component$(SUFFIX):${DOCKER_TAG} \
+	# Save the current Internal Field Separator (IFS) for later recovery, then set it to a comma-separator.
+	# The reason for using a comma-separator is to support the empty prefix required for all-in-one.
+	OLDIFS=$IFS; \
+	IFS=","; \
+	for pair in "jaeger,agent" "jaeger,collector" "jaeger,query" "jaeger,ingester" ",all-in-one" ; do \
+		set -- $$pair; \
+		prefix=$$1; \
+		component=$$2; \
+		docker buildx build --target $(TARGET) \
+			--tag $(DOCKER_NAMESPACE)/$$prefix$$component$(SUFFIX):${DOCKER_TAG} \
 			--build-arg base_image=$(BASE_IMAGE) \
 			--build-arg debug_image=$(DEBUG_IMAGE) \
 			--build-arg TARGETARCH=$(GOARCH) \
+			--load \
 			cmd/$$component ; \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -306,15 +306,26 @@ docker-images-jaeger-backend-debug: SUFFIX = -debug
...
+	for pair in "jaeger,agent" "jaeger,collector" "jaeger,query" "jaeger,ingester" ",all-in-one" ; do \
+		set -- $$pair; \
+		prefix=$$1; \
+		component=$$2; \
+		docker buildx build --target $(TARGET) \
+			--tag $(DOCKER_NAMESPACE)/$$prefix$$component$(SUFFIX):${DOCKER_TAG} \
 			--build-arg base_image=$(BASE_IMAGE) \
 			--build-arg debug_image=$(DEBUG_IMAGE) \
 			--build-arg TARGETARCH=$(GOARCH) \
+			--load \
```
albertteoh, line 322: Necessary for registering the images into docker so they are visible via `docker images`.

Makefile
```diff
@@ -306,15 +306,26 @@ docker-images-jaeger-backend-debug: SUFFIX = -debug
...
 .PHONY: docker-images-jaeger-backend docker-images-jaeger-backend-debug
 docker-images-jaeger-backend docker-images-jaeger-backend-debug: create-baseimg create-debugimg
-	for component in agent collector query ingester ; do \
-		docker build --target $(TARGET) \
-			--tag $(DOCKER_NAMESPACE)/jaeger-$$component$(SUFFIX):${DOCKER_TAG} \
+	# Save the current Internal Field Separator (IFS) for later recovery, then set it to a comma-separator.
+	# The reason for using a comma-separator is to support the empty prefix required for all-in-one.
+	OLDIFS=$IFS; \
+	IFS=","; \
+	for pair in "jaeger,agent" "jaeger,collector" "jaeger,query" "jaeger,ingester" ",all-in-one" ; do \
```
albertteoh, line 313: Open to other suggestions on a better approach.

yurishkuro: perhaps it's simpler to use strings like `jaeger-agent` and then just regex-out the `jaeger-` prefix when referring to the working dir `cmd/$$component`

albertteoh: Good suggestion, done in: [353cc39](https://github.com/jaegertracing/jaeger/pull/3813/commits/353cc392292a3284dce8662edddeb747155866a4)

Makefile
```diff
@@ -306,15 +307,23 @@ docker-images-jaeger-backend-debug: SUFFIX = -debug
...
+	for component in "jaeger-agent" "jaeger-collector" "jaeger-query" "jaeger-ingester" "all-in-one" ; do \
+		regex="(jaeger-)(.*)"; \
+		component_prefix=""; \
+		component_suffix=$$component; \
+		if [[ $$component =~ $$regex ]]; then \
+			component_prefix="$${BASH_REMATCH[1]}"; \
+			component_suffix="$${BASH_REMATCH[2]}"; \
+		fi; \
+		docker buildx build --target $(TARGET) \
+			--tag $(DOCKER_NAMESPACE)/$$component_prefix$$component_suffix$(SUFFIX):${DOCKER_TAG} \
```
yurishkuro, line 319:
```suggestion
			--tag $(DOCKER_NAMESPACE)/$$component$(SUFFIX):${DOCKER_TAG} \
```

Makefile
```diff
@@ -306,15 +307,23 @@ docker-images-jaeger-backend-debug: SUFFIX = -debug
...
 			--build-arg base_image=$(BASE_IMAGE) \
 			--build-arg debug_image=$(DEBUG_IMAGE) \
 			--build-arg TARGETARCH=$(GOARCH) \
-			cmd/$$component ; \
-		echo "Finished building $$component ==============" ; \
-	done
+			--load \
+			cmd/$$component_suffix ; \
+		echo "Finished building $$component_suffix ==============" ; \
```
yurishkuro, line 325:
```suggestion
		echo "Finished building $$component ==============" ; \
```

Makefile
```diff
@@ -306,15 +307,23 @@ docker-images-jaeger-backend-debug: SUFFIX = -debug
 
 .PHONY: docker-images-jaeger-backend docker-images-jaeger-backend-debug
 docker-images-jaeger-backend docker-images-jaeger-backend-debug: create-baseimg create-debugimg
-	for component in agent collector query ingester ; do \
-		docker build --target $(TARGET) \
-			--tag $(DOCKER_NAMESPACE)/jaeger-$$component$(SUFFIX):${DOCKER_TAG} \
+	for component in "jaeger-agent" "jaeger-collector" "jaeger-query" "jaeger-ingester" "all-in-one" ; do \
+		regex="(jaeger-)(.*)"; \
+		component_prefix=""; \
```
yurishkuro, line 312: this var is not needed

albertteoh: Thanks, that's simpler: [ae6b0e1](https://github.com/jaegertracing/jaeger/pull/3813/commits/ae6b0e1f79744c70838326b023e897ebe7df8cc1)

Makefile
```diff
@@ -306,15 +307,21 @@ docker-images-jaeger-backend-debug: SUFFIX = -debug
...
+			--tag $(DOCKER_NAMESPACE)/$$component$(SUFFIX):${DOCKER_TAG} \
 			--build-arg base_image=$(BASE_IMAGE) \
 			--build-arg debug_image=$(DEBUG_IMAGE) \
 			--build-arg TARGETARCH=$(GOARCH) \
-			cmd/$$component ; \
+			--load \
+			cmd/$$component_suffix; \
 		echo "Finished building $$component ==============" ; \
-	done
+	done; \
```
yurishkuro, line 324: why does this have `\` continuation?

Makefile
```diff
@@ -1,4 +1,5 @@
-JAEGER_IMPORT_PATH=github.com/jaegertracing/jaeger
+SHELL := /bin/bash
```
yurishkuro, line 1: is this needed?

Makefile
albertteoh: It was a remnant of the previous implementation where the IFS needed to be reset. Removed in [c649fa6](https://github.com/jaegertracing/jaeger/pull/3813/commits/c649fa6732d0660ef020bc081019ae41931e4986)

yurishkuro (approved):

Makefile
albertteoh: Yes, because regex matches via the `=~` operator requires `[[ ... ]]`, which is only supported in `bash`, and not `sh`. Without it, the following error will result: `/bin/sh: 4: [[: not found`.

Issue #3696, opened 2022-05-23
Provide a .zip of the Windows binaries 


jamers99:
## Requirement - what kind of business use case are you trying to solve?
Provide a way to download the .zip of the binaries at https://www.jaegertracing.io/download/#binaries

## Problem - what in Jaeger blocks you from solving the requirement?
Many Windows folks do not have the tools to unzip a .tar.gz file

## Proposal - what do you suggest to solve the problem or improve the existing situation?
Publish the binaries in .zip format as well, perhaps it should even be the default download for Windows.

## Any open questions to address
None

adhithyasrinivasan:
I will take this up

PR #3817, opened 2022-07-17
Added windows zip file generation

adhithyasrinivasan:
- 
- Fixing this issue #3696

- Added a function to generate zip file instead of tar file for windows machines.

Signed-off-by: adhithyasrinivasan [adhithyasrinivasan@gmail.com)


scripts/package-deploy.sh
```diff
@@ -46,6 +46,24 @@ function package {
     rm -rf $PACKAGE_STAGING_DIR
 }
 
+# package pulls built files for the platform ($1). If you pass in a file
+# extension ($2) it will be used on the binaries
+function package_zip {
```
yurishkuro, line 51: I prefer not to duplicate the logic from `package` function. I suggest passing a parameter that indicates the desired compression.

adhithyasrinivasan: Done

yurishkuro (approved):
Thanks

PR #3819, opened 2022-07-20
[ingester/fix] Apply sanitizers to avoid panic on span.process=nil

### LINKED ISSUE #3578: Jaeger ingester panics on convertProcess function

locmai:
- Fix #3578

- Using the process.GetTags() and process.GetServiceName() instead of directly reference to those fields to avoid panic due to nil pointer dereference.


plugin/storage/es/spanstore/dbmodel/from_domain_test.go
```diff
@@ -108,6 +108,32 @@ func TestTagMap(t *testing.T) {
...
+	converter := NewFromDomain(false, nil, ":")
+	dbSpanWithNilTags := converter.FromDomainEmbedProcess(&spanWithNilProcessTags)
+	dbSpanWithNilProcess := converter.FromDomainEmbedProcess(&spanWithNilProcess)
+
+	assert.Equal(t, 3, len(dbSpanWithNilTags.Tags))
+	assert.Equal(t, 3, len(dbSpanWithNilProcess.Tags))
+	assert.Equal(t, 0, len(dbSpanWithNilTags.Process.Tags))
+	assert.Equal(t, 0, len(dbSpanWithNilProcess.Process.Tags))
+
+	tagsMap := map[string]interface{}(nil)
```
locmai, line 130: not sure if this was expected but this is what I got from the `convertKeyValuesString` func if the Tags was nil.

plugin/storage/es/spanstore/dbmodel/from_domain.go
```diff
@@ -120,9 +120,9 @@ func (fd FromDomain) convertLogs(logs []model.Log) []Log {
 }
 
 func (fd FromDomain) convertProcess(process *model.Process) Process {
-	tags, tagsMap := fd.convertKeyValuesString(process.Tags)
+	tags, tagsMap := fd.convertKeyValuesString(process.GetTags())
 	return Process{
-		ServiceName: process.ServiceName,
+		ServiceName: process.GetServiceName(),
```
yurishkuro, line 125: So this is exactly the type of change that I did not want to make. It avoids a panic, but does it silently, creating invalid data in the process (service name is not allowed to be empty, it could break many other places, including the UI).

I've added a sanitizer in #3631. We can try invoking it early in the Kafka consumer path.

locmai: Thanks for taking a look at this and making things clearer. Let me take a look at the sanitizer and try what you suggested.

It make sense that the service name must not be empty, so I might revert the change to process.GetServiceName() here. For the process.GetTags, I think it still nicely handles the `nil` cases those led to the panics.

yurishkuro: > I think it still nicely handles the nil cases those led to the panics.

Yes, but again, it does that accidentally, it so happens that in this context the behavior is appropriate. But fundamentally you're trying to work around malformed data that should've never gotten to this point in the pipeline. *I don't mind this specific change*, but at least with the current code we knew that there is a data issue (even though panicing on it is not ideal reaction).

plugin/storage/es/spanstore/dbmodel/from_domain_test.go
yurishkuro, line 130:
```suggestion
	nilMap := map[string]interface{}(nil)
```

plugin/storage/es/spanstore/dbmodel/from_domain_test.go
```diff
@@ -108,6 +108,32 @@ func TestTagMap(t *testing.T) {
...
+	}
+
+	spanWithNilProcessTags := model.Span{Tags: tags, Process: &model.Process{Tags: nil}}
+	spanWithNilProcess := model.Span{Tags: tags, Process: nil}
+
+	converter := NewFromDomain(false, nil, ":")
+	dbSpanWithNilTags := converter.FromDomainEmbedProcess(&spanWithNilProcessTags)
+	dbSpanWithNilProcess := converter.FromDomainEmbedProcess(&spanWithNilProcess)
+
+	assert.Equal(t, 3, len(dbSpanWithNilTags.Tags))
```
yurishkuro, line 125:
```suggestion
	assert.Len(t, 3, dbSpanWithNilTags.Tags)
```

plugin/storage/es/spanstore/dbmodel/from_domain.go
trangmaiq: Hi @yurishkuro , thank you so much for your review. As you mentioned, I understand it's important to detect data issue in our system, so should we return error here and let caller know that we do not accept the nil process for example? 

I think panic should be considered if we do want application immediately crash. Otherwise, we can consider error.

Thank you for your explanation and effort.

yurishkuro: If we apply the sanitizer earlier in the pipeline then this code does need to change as it will never have to deal with nil process.

plugin/storage/es/spanstore/dbmodel/from_domain_test.go
```diff
@@ -108,6 +108,26 @@ func TestTagMap(t *testing.T) {
...
 }
 
+func TestNilProcess(t *testing.T) {
+	tags := []model.KeyValue{
+		model.String("foo", "foo"),
+		model.Bool("a", true),
+		model.Int64("b.b", 1),
+	}
+
+	spanWithNilTags := model.Span{Tags: tags, Process: &model.Process{Tags: nil}}
```
yurishkuro, line 118: This test would pass without any code changes in the PR. It's the nil Process that's causing panic in production, not the nil tags (nil tags are still iterable).

yurishkuro (approved):

PR #3827, opened 2022-07-21
Tenancy for memory storage

esnible:

This PR adds tenancy support to the memory storage.

### DIFF
--- a/plugin/storage/memory/memory.go
+++ b/plugin/storage/memory/memory.go
@@ -27,18 +27,25 @@ import (
 	"github.com/jaegertracing/jaeger/model"
 	"github.com/jaegertracing/jaeger/model/adjuster"
 	"github.com/jaegertracing/jaeger/pkg/memory/config"
+	"github.com/jaegertracing/jaeger/pkg/tenancy"
 	"github.com/jaegertracing/jaeger/storage/spanstore"
 )
 
 // Store is an in-memory store of traces
 type Store struct {
 	sync.RWMutex
+	config    config.Configuration
+	perTenant map[string]*Tenant
+}
+
+// Tenant is an in-memory store of traces for a single tenant
+type Tenant struct {
+	sync.RWMutex
 	ids        []*model.TraceID
 	traces     map[model.TraceID]*model.Trace
 	services   map[string]struct{}
 	operations map[string]map[spanstore.Operation]struct{}
 	deduper    adjuster.Adjuster
-	config     config.Configuration
 	index      int
 }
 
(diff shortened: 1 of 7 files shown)


plugin/storage/memory/memory.go
```diff
@@ -27,18 +27,25 @@ import (
...
 	"github.com/jaegertracing/jaeger/model/adjuster"
 	"github.com/jaegertracing/jaeger/pkg/memory/config"
+	"github.com/jaegertracing/jaeger/pkg/tenancy"
 	"github.com/jaegertracing/jaeger/storage/spanstore"
 )
 
 // Store is an in-memory store of traces
 type Store struct {
+	sync.RWMutex
+	config    config.Configuration
```
yurishkuro, line 37: I would recommend keeping config at the Tenant struct level (copy the same one for now), because it seems very reasonable to want per-tenant quota settings rather than one global setting. Or maybe not?

plugin/storage/memory/memory.go
```diff
@@ -50,17 +57,40 @@ func NewStore() *Store {
...
 		traces:     map[model.TraceID]*model.Trace{},
 		services:   map[string]struct{}{},
 		operations: map[string]map[spanstore.Operation]struct{}{},
 		deduper:    adjuster.SpanIDDeduper(),
-		config:     configuration,
 	}
 }
 
+// GetTenant returns the per-tenant storage.  Note that tenantID has already been checked for by the collector or query
+func (st *Store) GetTenant(tenantID string) *Tenant {
```
yurishkuro, line 76: does this need to be public?

plugin/storage/memory/memory.go
yurishkuro: Sorry, what I meant was this:
```go
type Store struct {
  // Each tenant gets a copy of default config.
  // In the future this can be extended to contain per-tenant configuration.
  defaultConfig config.Config
}
type Tenant struct {
  config config.Config
}

func (s *Store) newTenant() *Tenant {
  return &Tenant{
    config: s.defaultConfig
  }
}
```

plugin/storage/memory/memory.go
```diff
@@ -59,8 +74,32 @@ func WithConfiguration(configuration config.Configuration) *Store {
 	}
 }
 
+// getTenant returns the per-tenant storage.  Note that tenantID has already been checked for by the collector or query
+func (st *Store) getTenant(tenantID string) *Tenant {
+	tenant, ok := st.perTenant[tenantID]
+	if !ok {
+		// We do the lookup twice to skip locking on retrieval of existing tenant
```
yurishkuro, line 81: Can't really do this because you're doing unprotected read above that could cause a concurrency issue. Test with `-race` should catch this. If you're worried about locking, you can either use RWLock or use a sync.Map which is specifically designed for frequent reads / rare writes.

plugin/storage/memory/memory.go
```diff
@@ -50,17 +60,42 @@ func NewStore() *Store {
...
 	return &Store{
-		ids:        make([]*model.TraceID, configuration.MaxTraces),
+		defaultConfig: configuration,
+		perTenant:     make(map[string]*Tenant),
+	}
+}
+
+func (st *Store) newTenant() *Tenant {
+	return &Tenant{
+		ids:        make([]*model.TraceID, st.defaultConfig.MaxTraces),
```
yurishkuro, line 70: nit: can decouple this func from the knowledge of defaultConfig:
```
// caller
newTenant(st.defaultConfig)

func (st *Store) newTenant(cfg config.Config) *Tenant {
  // ...
}
```

plugin/storage/memory/memory.go
```diff
@@ -290,7 +330,7 @@ func (m *Store) validSpan(span *model.Span, query *spanstore.TraceQueryParameter
 	return true
 }
 
-func (m *Store) flattenTags(span *model.Span) model.KeyValues {
+func flattenTags(span *model.Span) model.KeyValues {
 	retMe := span.Tags
 	retMe = append(retMe, span.Process.Tags...)
```
yurishkuro, line 335: this looks like existing bug - it could mutate `span.Tags`

plugin/storage/memory/memory_test.go
```diff
@@ -53,6 +54,32 @@ var testingSpan = &model.Span{
 	StartTime: time.Unix(300, 0).UTC(),
 }
 
+var traceID2 = model.NewTraceID(2, 3)
+
+var testingSpan2 = &model.Span{
```
yurishkuro, line 59: nit: turn `testingSpan` into a function that takes a few args distinguishing the traces (like trace id)

yurishkuro (approved):

PR #3834, opened 2022-08-01
Normalize multi-tenancy flags prefix

yurishkuro:
Changing `--multi_tenancy...` to `multi-tenancy...` CLI flag prefix.
The usual pattern for CLI flags in Jaeger is to use `-` between words, so that they can be typed without using the SHIFT key.


yurishkuro:
cc @esnible

albertteoh (approved):

Issue #3851, opened 2022-08-04
[Bug]: Release v1.37.0 is missing windows build in assets


atri-dastidar-snkeos:
### What happened?

[The latest release v1.37.0](https://github.com/jaegertracing/jaeger/releases/tag/v1.37.0) is missing windows build in assets.

### Steps to reproduce

N/A

### Expected behavior

Expected to see a windows build (jaeger-1.37.0-windows-amd64.tar.gz) as with previous releases.

atri-dastidar-snkeos:
@Ashmita152 @yurishkuro thanks for the fix. Is [v1.37.0](https://github.com/jaegertracing/jaeger/releases/tag/v1.37.0) going to be re-released with the windows binaries?

PR #3854, opened 2022-08-05
Fix upload of zip file in release github action

Ashmita152:
Resolves https://github.com/jaegertracing/jaeger/issues/3851

Earlier we used to have tar.gz for windows release. https://github.com/jaegertracing/jaeger/pull/3817 changed the windows release from tar.gz to zip format. It broke jaeger windows users hence we need to restore earlier format too. This PR also restores the backward compatibility i.e. now we have both tar.gz and zip for windows release.

I tested the file_glob in a test repo with test release:
- https://github.com/Ashmita152/test-upload/blob/master/.github/workflows/ci-release.yml
- https://github.com/Ashmita152/test-upload/releases/tag/v0.4

### DIFF
--- a/scripts/package-deploy.sh
+++ b/scripts/package-deploy.sh
@@ -28,19 +28,17 @@ function stage-platform-files {
     stage-file ./examples/hotrod/hotrod-$PLATFORM $PACKAGE_STAGING_DIR/example-hotrod$FILE_EXTENSION
 }
 
-# package pulls built files for the platform ($1). If you pass in a file
-# extension ($2) it will be used on the binaries
+# package pulls built files for the platform ($2) and compresses it using the compression ($1).
+# If you pass in a file extension ($3) it will be look for binaries with that extension.
 function package {
-    
     local COMPRESSION=$1
     local PLATFORM=$2
     local FILE_EXTENSION=$3
-    # script start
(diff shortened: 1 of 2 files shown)


yurishkuro (approved):

scripts/package-deploy.sh
```diff
@@ -28,19 +28,17 @@ function stage-platform-files {
     stage-file ./examples/hotrod/hotrod-$PLATFORM $PACKAGE_STAGING_DIR/example-hotrod$FILE_EXTENSION
 }
 
-# package pulls built files for the platform ($1). If you pass in a file
-# extension ($2) it will be used on the binaries
+# package pulls built files for the platform ($2) and compresses it using the compression ($1).
+# If you pass in a file extension ($3) it will be look for binaries with that extension.
```
albertteoh, line 32: nit: My understanding of `stage-platform-files` is that it copies binary files over into the `$PACKAGE_STAGING_DIR` and adds the `$FILE_EXTENSION` to that copied filename, and AFAICT, doesn't look for binaries with `$FILE_EXTENSION`. Perhaps we should update this sentence.

albertteoh (commented):
Looks like I just missed the boat 😄

Issue #3877, opened 2022-08-17
[Bug]: Reproducible UI builds


TripleDogDare:
### What happened?

UI builds are not reproducible.

There's at least two issues that I've discovered so far.
1. gzipped files are inserting file timestamps. This can be resolved in the Makefile fairly easily.
2. `asset-manifest.json` does not maintain a consistent sort order. I'll probably need to add something to `./scripts` to resolve this.

### Steps to reproduce

1. `make rebuild-ui`
2. `find ./cmd/query/app/ui/actual/ -type f -exec sha256sum {} \; | tee ui.gz.sha256sum`
3. `find ./cmd/query/app/ui/actual -type f -name '*.gz' -delete`
4. `make-rebuild-ui`
5. `sha256sum -c ui.gz.sha256sum`

### Expected behavior

sha256sum check should return OK

### Additional context

Other discussion at #3876

yurishkuro:
I think this is solved, if not please list additional items required.

PR #3880, opened 2022-08-17
Fixes reproducibility for binaries

TripleDogDare:
This does not help with reproducibility checks across build systems. But
should help ensure that multiple builds on the same build system are
reproducible.

Resolves Issue: #3877

Adds reproducibility checks to the make file and CI.
Fixes timestamps embedded in gzipped UI files.
Fixes reproducibility for asset-manifest.json before compression/embedding.

### DIFF
--- a/.github/workflows/ci-build-binaries.yml
+++ b/.github/workflows/ci-build-binaries.yml
@@ -49,15 +49,3 @@ jobs:
 
     - name: Build binaries
       run: make ${{ matrix.platform.task }}
-
-    - name: Collect checksums
-      run: make collect-sums
-
-    - name: Clean
-      run: make clean
-
-    - name: Build binaries
-      run: make ${{ matrix.platform.task }}
-
-    - name: Validate reproducibility
-      run: make test-checksums
--- a/.github/workflows/ci-build-binaries.yml
+++ b/.github/workflows/ci-build-binaries.yml
@@ -49,3 +49,15 @@ jobs:
 
     - name: Build binaries
       run: make ${{ matrix.platform.task }}
+
+    - name: Collect checksums
+      run: make collect-sums
+
+    - name: Clean
+      run: make clean
+
+    - name: Build binaries
+      run: make ${{ matrix.platform.task }}
+
+    - name: Validate reproducibility
+      run: make test-checksums
--- a/Makefile
+++ b/Makefile
@@ -241,7 +241,7 @@ build-collector build-collector-debug:
 
 .PHONY: build-ingester build-ingester-debug
 build-ingester build-ingester-debug:
-	$(GOBUILD) $(DISABLE_OPTIMIZATIONS) -o ./cmd/ingester/ingester$(SUFFIX)-$(GOOS)-$(GOARCH) ./cmd/ingester/main.go
+	$(GOBUILD) $(DISABLE_OPTIMIZATIONS) -o ./cmd/ingester/ingester$(SUFFIX)-$(GOOS)-$(GOARCH) $(BUILD_INFO) ./cmd/ingester/main.go
 
 .PHONY: build-remote-storage build-remote-storage-debug
 build-remote-storage build-remote-storage-debug:
@@ -603,19 +603,13 @@ certs:
 certs-dryrun:
 	cd pkg/config/tlscfg/testdata && ./gen-certs.sh -d
 
-.PHONY: collect-sums
-collect-sums:
-	find cmd -type f -executable -exec sha256sum {} \; | sort -k2 > sha256sum.combined.txt
-
-.PHONY: test-checksums
-test-checksums:
-	sha256sum --strict --check ./sha256sum.combined.txt
-
 .PHONY: repro-check
 repro-check:
+	# Check local reproducibility of generated executables.
 	$(MAKE) clean
 	$(MAKE) build-all-platforms
-	$(MAKE) collect-sums
+	# Generate checksum for all executables under ./cmd
+	find cmd -type f -executable -exec sha256sum {} \; | sort -k2 > sha256sum.combined.txt
 	$(MAKE) clean
 	$(MAKE) build-all-platforms
-	$(MAKE) test-checksums
+	sha256sum --strict --check ./sha256sum.combined.txt
--- a/Makefile
+++ b/Makefile
@@ -33,8 +33,9 @@ else
 endif
 GOOS ?= $(shell go env GOOS)
 GOARCH ?= $(shell go env GOARCH)
-GOBUILD=CGO_ENABLED=0 installsuffix=cgo go build -trimpath
-GOTEST=go test -v $(RACE)
+GOCACHE=$(abspath .gocache)
+GOBUILD=GOCACHE=$(GOCACHE) CGO_ENABLED=0 installsuffix=cgo go build -trimpath
+GOTEST=GOCACHE=$(GOCACHE) go test -v $(RACE)
 GOFMT=gofmt
 GOFUMPT=gofumpt
 FMT_LOG=.fmt.log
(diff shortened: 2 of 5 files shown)


.github/workflows/ci-build-binaries.yml
```diff
@@ -49,3 +49,15 @@ jobs:
...
       run: make ${{ matrix.platform.task }}
+
+    - name: Collect checksums
+      run: make collect-sums
+
+    - name: Clean
+      run: make clean
+
+    - name: Build binaries
+      run: make ${{ matrix.platform.task }}
```
yurishkuro, line 60: I wouldn't want to do this, this is already a very long CI job, it's not worth making it twice as long just to verify repro.

TripleDogDare: Actually less because I'm not cleaning the go build cache, which is something I should check on as well. So it helped with something! But I agree the operational cost is something to consider. I do think it makes sense to ship checksums with releases and include it in the release tagging process.

yurishkuro: Yes, adding checksums to the release process would be useful.

TripleDogDare: Do you prefer additional commits or squashed commits?

yurishkuro: doesn't matter, we squash on merge, just make sure all commits are signed.

Makefile
```diff
@@ -351,6 +372,7 @@ docker-images-only: docker-images-cassandra \
 .PHONY: build-crossdock-binary
 build-crossdock-binary:
 	$(GOBUILD) -o ./crossdock/crossdock-$(GOOS)-$(GOARCH) ./crossdock/main.go
+	sha256sum crossdock/crossdock-$(GOOS)-$(GOARCH) > ./crossdock/crossdock-$(GOOS)-$(GOARCH).sha256sum.txt
```
yurishkuro, line 375: Couldn't we do it in scripts/package-deploy.sh instead of repeating command for every binary?  checksum files are not useful by themselves outside of packaging a release

Makefile
```diff
@@ -203,14 +213,19 @@ cmd/query/app/ui/actual/index.html.gz: jaeger-ui/packages/jaeger-ui/build/index.
...
 
 jaeger-ui/packages/jaeger-ui/build/index.html:
 	$(MAKE) rebuild-ui
 
 .PHONY: rebuild-ui
 rebuild-ui:
 	cd jaeger-ui && yarn install --frozen-lockfile && cd packages/jaeger-ui && yarn build
+	# Asset manifest must be in sorted order for binary reproducibility.
+	mv jaeger-ui/packages/jaeger-ui/build/asset-manifest.json .asset-manifest.json
+	cat .asset-manifest.json | python -c 'import json, sys; o = json.loads(sys.stdin.read()); print(json.dumps(o, indent=2, sort_keys=True))' > jaeger-ui/packages/jaeger-ui/build/asset-manifest.json
```
yurishkuro, line 228: My preference would be to do this in the UI project using Node.js, so that UI build is fully responsible for producing stable/repeatable artifacts.

TripleDogDare: I made a pass at this and... well I'm kind of at a loss for why it doesn't work.
https://github.com/TripleDogDare/jaeger-ui/pull/1/files

Honestly could probably just delete the file as I don't think it's being used at all.

Makefile
TripleDogDare: It would require a different repro-check script and would omit anything that isn't currently explicitly packaged for release.

Makefile
TripleDogDare: Here's a pass at resolving this in the UI build.
https://github.com/jaegertracing/jaeger-ui/pull/981

Makefile
TripleDogDare: I removed all these checksums and changed how this works. I'm actually happier with how this turned out.

Makefile
```diff
@@ -237,7 +241,7 @@ build-collector build-collector-debug:
 
 .PHONY: build-ingester build-ingester-debug
 build-ingester build-ingester-debug:
-	$(GOBUILD) $(DISABLE_OPTIMIZATIONS) -o ./cmd/ingester/ingester$(SUFFIX)-$(GOOS)-$(GOARCH) $(BUILD_INFO) ./cmd/ingester/main.go
+	$(GOBUILD) $(DISABLE_OPTIMIZATIONS) -o ./cmd/ingester/ingester$(SUFFIX)-$(GOOS)-$(GOARCH) ./cmd/ingester/main.go
```
yurishkuro, line 244: why did you drop `$(BUILD_INFO)` here?

Makefile
```diff
@@ -598,3 +602,20 @@ certs:
...
 certs-dryrun:
 	cd pkg/config/tlscfg/testdata && ./gen-certs.sh -d
+
+.PHONY: collect-sums
+collect-sums:
+	find cmd -type f -executable -exec sha256sum {} \; | sort -k2 > sha256sum.combined.txt
+
+.PHONY: test-checksums
+test-checksums:
+	sha256sum --strict --check ./sha256sum.combined.txt
```
yurishkuro, line 612: where is `sha256sum` from?

Makefile
```diff
@@ -598,3 +602,20 @@ certs:
 .PHONY: certs-dryrun
 certs-dryrun:
 	cd pkg/config/tlscfg/testdata && ./gen-certs.sh -d
+
```
yurishkuro, line 605: are these targets necessary?

scripts/package-deploy.sh
```diff
@@ -71,3 +75,4 @@ package zip windows-amd64 .exe
 package tar linux-s390x
 package tar linux-arm64
 package tar linux-ppc64le
+find deploy \( ! -name '*sha256sum.txt' \) -type f -exec sha256sum {} \; | sed -r 's/(\w+\s+).*(jaeger)/\1\2/' | sort -k2 | tee ./deploy/jaeger-$VERSION.sha256sum.txt
```
yurishkuro, line 78: please add a comment explaining this line

scripts/package-deploy.sh
```diff
@@ -1,15 +1,14 @@
...
-
 # stage-file copies the file $1 with the specified path $2
-# if no file exists it will silently continue
+# if no file exists it will abort
 function stage-file {
-    if [ -f $1 ]; then
-        echo "Copying $1 to $2"
-        cp $1 $2
-    else
+    if [ ! -f $1 ]
```
yurishkuro, line 5: would it be better to simply do `set -euxf -o pipefail` at the top, as we do in most our scripts?

Makefile
TripleDogDare: I believe it is part of GNU core-utils.

Makefile
TripleDogDare: Ah good catch. It definitely should be there and deleting it was probably an accident.

Makefile
TripleDogDare: I think having repro-check is useful to give everything a whirl without thinking too hard about how to do that. That said, the collect and test sums targets could be inlined there.

scripts/package-deploy.sh
TripleDogDare: Yeah I'd agree with that.

TripleDogDare: If you include `-x` then the echo for copying doesn't really make sense and the whole function can just be removed.

Makefile
```diff
@@ -598,3 +602,14 @@ certs:
...
+.PHONY: repro-check
+repro-check:
+	# Check local reproducibility of generated executables.
+	$(MAKE) clean
+	$(MAKE) build-all-platforms
+	# Generate checksum for all executables under ./cmd
+	find cmd -type f -executable -exec sha256sum {} \; | sort -k2 > sha256sum.combined.txt
+	$(MAKE) clean
+	$(MAKE) build-all-platforms
+	sha256sum --strict --check ./sha256sum.combined.txt
```
yurishkuro, line 615: can we not use `shasum -a 256`? It should work on both Mac and Linux without additional dependencies

TripleDogDare: `shasum` is included by `perl` package which is usually shipped with full Debian based distributions. I can't say that it will always be available in Linux as it's not part of coreutils. Particularly if this is run in a reduced size container which will likely have `perl-base` instead of `perl`.

TripleDogDare: This would mostly just affect maintainers so it's a limited set of users.
Options:
1) Use `shasum` and require it to be installed by any edge-case Linux users. Presumably CI will function properly on [ubuntu-latest](https://github.com/jaegertracing/jaeger/blob/main/.github/workflows/ci-release.yml#L10) image, but it might not.
  - We could also leave it as-is in `package-deploy.sh` and only apply the chosen change to `Makefile`.
2) Require macOS users to do `brew install coreutils`.
3) Try to detect which commands exist and switch between the two.

yurishkuro: shasum works fine in our CI, I just tested: https://github.com/jaegertracing/jaeger/runs/8061728418?check_suite_focus=true#step:5:27

yurishkuro (dismissed):

TripleDogDare:
Got a random test failure.
```
2022-08-31T20:45:56.0221232Z --- FAIL: TestGRPCResolverRoundRobin (0.05s)
2022-08-31T20:45:56.0221979Z     --- PASS: TestGRPCResolverRoundRobin/{minPeers:3_connections:3} (0.02s)
2022-08-31T20:45:56.0222782Z     --- FAIL: TestGRPCResolverRoundRobin/{minPeers:5_connections:3} (0.02s)
```
Clearly not related. Rebased on main to see if that clears it.

PR #3887, opened 2022-08-26
fix: streaming span writer is not working in grpc based remote storage plugin

arajkumar:
- streaming span writer is not working in grpc remote storage plugin
- errors other than io.EOF from grpc framework causes crash with streaming span writer


arajkumar:
cc @yurishkuro

yurishkuro (approved):

PR #3892, opened 2022-08-31
fix: jaeger-agent sampling endpoint returns backwards incompatible JSON

### LINKED ISSUE #3891: [Bug]: Jaeger Agent's http /sampling endpoint returns JSON which breaks older jaeger clients
### What happened?

Jaeger Agent's sampling endpoint's `perOperationStrategies` field returns `null` instead of `[]`, which breaks older Java jaeger-clients.

### Steps to reproduce

As of 1.23, sampling endpoint returns the following when using grpc as transport between agent and collector.
```
$ curl "localhost:5778/sampling?service=none" | jq
 
  "probabilisticSampling": {
    "samplingRate": 0
  },
  "operationSampling": {
    "defaultSamplingProbability": 0.001,
    "defaultLowerBoundTracesPerSecond": 1e-05,
    "perOperationStrategies": null
  }
}
```

### Expected behavior

As of version 1.17, sampling endpoint returns the following when using tchannel as transport between agent and collector

```
$ curl "localhost:5778/sampling?service=none" | jq
 
  "probabilisticSampling": {
    "samplingRate": 0
  },
  "operationSampling": {
    "defaultSamplingProbability": 0.001,
    "defaultLowerBoundTracesPerSecond": 1e-05,
    "perOperationStrategies": []
  }
}
```

This is because the grpc/proto response from collector uses nil as the default value for slice, and json.Marshal then encodes nil as `null`. It is possible that tchannel/thrift didn't follow the same conventions and used empty arrays instead.

From https://pkg.go.dev/encoding/json#Marshal, 

> Array and slice values encode as JSON arrays, except that []byte encodes as a base64-encoded string, and a nil slice encodes as the null JSON value.

vprithvi:
- Sampling endpoint's `perOperationStrategies` field contains `[]` as a default value instead of nil. This was the behavior as of jaeger-agent v1.17 when using tchannel/thrift to communicate with collector.
- Returning an empty slice for `perOperationStrategies` instead of the default nil slice allows
  json.Marshal to marshal it to `[]` instead of `null`
- Resolves #3891

### DIFF
--- a/model/converter/thrift/jaeger/sampling_from_domain.go
+++ b/model/converter/thrift/jaeger/sampling_from_domain.go
@@ -67,14 +67,12 @@ func convertPerOperationFromDomain(s *api_v2.PerOperationSamplingStrategies) *sa
 		DefaultLowerBoundTracesPerSecond: s.GetDefaultLowerBoundTracesPerSecond(),
 		DefaultUpperBoundTracesPerSecond: &s.DefaultUpperBoundTracesPerSecond,
 	}
-	if s.GetPerOperationStrategies() != nil {
-		r.PerOperationStrategies = make([]*sampling.OperationSamplingStrategy, len(s.GetPerOperationStrategies()))
-		for i, k := range s.PerOperationStrategies {
-			r.PerOperationStrategies[i] = convertOperationFromDomain(k)
-		}
-	} else {
-		// Initialize to an empty array so that using json.Marshal results in [] instead of null.
-		r.PerOperationStrategies = make([]*sampling.OperationSamplingStrategy, 0)
+
+	perOp := s.GetPerOperationStrategies()
+	// Default to empty array so that json.Marshal returns [] instead of null (Issue #3891).
+	r.PerOperationStrategies = make([]*sampling.OperationSamplingStrategy, len(perOp))
+	for i, k := range perOp {
+		r.PerOperationStrategies[i] = convertOperationFromDomain(k)
 	}
 	return r
 }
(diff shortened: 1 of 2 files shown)


model/converter/thrift/jaeger/sampling_from_domain.go
```diff
@@ -72,6 +72,9 @@ func convertPerOperationFromDomain(s *api_v2.PerOperationSamplingStrategies) *sa
 		for i, k := range s.PerOperationStrategies {
 			r.PerOperationStrategies[i] = convertOperationFromDomain(k)
 		}
+	} else {
+		// Initialize to an empty array so that using json.Marshal results in [] instead of null.
+		r.PerOperationStrategies = make([]*sampling.OperationSamplingStrategy, 0)
 	}
```
yurishkuro, line 78: I think you can simply remove the if statement and only keep its content
```
        perOp := s.GetPerOperationStrategies()
        // Want an empty array by default so that using json.Marshal results in [] instead of null.
        r.PerOperationStrategies = make([]*sampling.OperationSamplingStrategy, len(perOp))
        for i, k := range perOp {
            r.PerOperationStrategies[i] = convertOperationFromDomain(k)
        }
```

vprithvi: D'oh! ty!

yurishkuro (approved):

abliqo (approved):

vprithvi:
@yurishkuro ty! The builds seem to be failing when publishing to docker with the following, how to proceed?

```
+ docker login -u jaegertracingbot docker.io --password-stdin
```

yurishkuro:
ah, it's because you created PR from your main branch, it will continue failing. Needs to be on a named branch.

vprithvi:
> ah, it's because you created PR from your main branch, it will continue failing. Needs to be on a named branch.

I see, I created https://github.com/jaegertracing/jaeger/pull/3897 with the same contents;

PR #3898, opened 2022-09-05
Use all span kinds as default for metrics

ricoberger:
Hi everyone, if this change doesn't make sense or if I missed something to configure the span kind feel free to close the PR.

Background: The span kind for some services could have another value then "SPAN_KIND_SERVER", e.g. the spans from Istio are using "SPAN_KIND_CLIENT". Because of the default span kind and the fact that the span kind is not configureable in the frontend (the query parameter is never set in the API calls) it was not possible to use the service performance monitoring with some services.

Instead of just using "SPAN_KIND_SERVER" as the default span kind for the metrics, the default value now includes all span kinds.

### DIFF
--- a/cmd/query/app/default_params.go
+++ b/cmd/query/app/default_params.go
@@ -28,5 +28,12 @@ var (
 	defaultMetricsQueryLookbackDuration = time.Hour
 	defaultMetricsQueryStepDuration     = 5 * time.Second
 	defaultMetricsQueryRateDuration     = 10 * time.Minute
-	defaultMetricsSpanKinds             = []string{metrics.SpanKind_SPAN_KIND_SERVER.String()}
+	defaultMetricsSpanKinds             = []string{
+		metrics.SpanKind_SPAN_KIND_UNSPECIFIED.String(),
+		metrics.SpanKind_SPAN_KIND_INTERNAL.String(),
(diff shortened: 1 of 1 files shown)


ankitnayan:
@ricoberger that would create a lot of noise IMHO in the metrics expected in SPM. Say, 1 server call has downstream 1000 calls to other dbs, queues and other services. With the suggested approach, all the spans will be used to calculate metrics for that 1 server call.

albertteoh:
Closing this PR for now. Feel free to re-open if you wish to discuss further.

ricoberger:
Hi, sorry I completely forgot about this. It's definitely ok to close the PR we tested this further during the last weeks and implemented a solution in our Jaeger frontend, where the user can select the span kind.

albertteoh:
> we tested this further during the last weeks and implemented a solution in our Jaeger frontend, where the user can select the span kind.

Nice one @ricoberger! Would you be open to contributing what you've done to the https://github.com/jaegertracing/jaeger-ui/ project? I think it's a valid use case to query on non-server span kinds and would be valuable to others in the community.

nifrasinnovent:
@albertteoh  `a dropdown in Jaeger UI's Monitor tab to select the span kind, and perhaps cache that selection per service`
 This feature is applicable, if not is there any workaround to accomplish this for now?

PR #3919, opened 2022-09-20
Change tags_opname_trace to have only one parent

### LINKED ISSUE #3918: [Feature]: Add more realistic scenario to find trace integration test
### Requirement

This is a follow up of https://github.com/jaegertracing/jaeger/issues/3912 

As a developer working on automatically running the TestExternalGRPCStorage tests against a custom storage that uses the [OTEL translator libs](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/7b5eb5280c15a5cc195d68757ba4c24c2c7f8eba/pkg/translator/jaeger) and the gRPC plugin.

I want a more realistic scenario for the `TestExternalGRPCStorage/FindTraces/Tags_+_Operation_name` so that I can have more confidence on the results of the integration test.

### Problem

We are using the [OTEL translator pkg](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/7b5eb5280c15a5cc195d68757ba4c24c2c7f8eba/pkg/translator) to ingest the spans, since that's the datamodel with started with:

#### Spans in fixture have different trace_id:

This test sends a span (a) from a trace and has 3 references to 2 other spans (b and c):

- a CHILD_OF b
- a CHILD_OF of c
- a FOLLOWS_FROM c

https://github.com/jaegertracing/jaeger/blob/cac21f82950545b613f45d3984d96e1498bfd15d/plugin/storage/integration/fixtures/traces/tags_opname_trace.json#L4-L23

The problem is with the child relationship and the fact that each span has a different `traceId`. This breaks the parent relationship because when the translator tries to get the parent span it cannot find a match for the parent. ParentSpanID method makes the following assertion `ref.TraceID == s.TraceID && ref.RefType == ChildOf` and since each span belongs to a different trace no match is found:

https://github.com/jaegertracing/jaeger/blob/cac21f82950545b613f45d3984d96e1498bfd15d/model/span.go#L99-L109

#### OTEL doesn't have a concept of refType

The OTEL model doesn't define a span reference type (in the case of OTEL it would be link). When translating from otel to jaeger the parent span is assign the `CHILD_OF` refType, and all other links are assigned `FOLLOWS_FROM`:

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/7b5eb5280c15a5cc195d68757ba4c24c2c7f8eba/pkg/translator/jaeger/traces_to_jaegerproto.go#L260-L281

This explicitly says:

>  Since Jaeger RefType is not captured in internal data,  use SpanRefType_FOLLOWS_FROM by default. 
> SpanRefType_CHILD_OF supposed to be set only from parentSpanID.

This is a problem for the test because it's only setting one `CHILD_OF` when the test expects 2 from the result.

The expected based on the fixture is:

```
                "references": [
                    {
                        "trace_id": "AAAAAAAAAAAAAAAAAAAAEg==",
                        "span_id": "AAAAAAAAAP8="
                    },
                    {
                        "trace_id": "AAAAAAAAAAAAAAAAAAAAEg==",
                        "span_id": "AAAAAAAAAAI="
                    },
                    {
                        "trace_id": "AAAAAAAAAAAAAAAAAAAAAQ==",
                        "span_id": "AAAAAAAAAAI=",
                        "ref_type": 1
                    }
                ],
```

The first 2 are `ref_type=0` which is `CHILD_OF`.

The actual result we are getting because the translator sets only one `CHILD_OF` is:

```
                "references": [
                    {
                        "trace_id": "AAAAAAAAAAAAAAAAAAAAEg==",
                        "span_id": "AAAAAAAAAP8="
                    },
                    {
                        "trace_id": "AAAAAAAAAAAAAAAAAAAAEg==",
                        "span_id": "AAAAAAAAAAI=",
                        "ref_type": 1
                    },
                    {
                        "trace_id": "AAAAAAAAAAAAAAAAAAAAAQ==",
                        "span_id": "AAAAAAAAAAI=",
                        "ref_type": 1
                    }
                ],
```

As you can see, since the translator takes the first `CHILD_OF` for the span parent id all the other relationships are FOLLOWS_FROM.

### Proposal

Change the relationships that are being set in the fixture. Currently the following is being set:

- span_a from trace_a CHILD_OF span_b from trace_b
- span_a from trace_a CHILD_OF span_c from trace_c
- span_a from trace_a FOLLOWS_FROM span_c from trace_c

We could change it to:

- span_a from trace_a CHILD_OF span_b from trace_a
- span_a from trace_a FOLLOWS_FROM span_b from trace_a
- span_a from trace_a FOLLOWS_FROM span_c from trace_c

```
diff --git a/plugin/storage/integration/fixtures/traces/tags_opname_trace.json b/plugin/storage/integration/fixtures/traces/tags_opname_trace.json
index 664c19f9..86abf00a 100644
--- a/plugin/storage/integration/fixtures/traces/tags_opname_trace.json
+++ b/plugin/storage/integration/fixtures/traces/tags_opname_trace.json
@@ -7,13 +7,13 @@
       "references": [
         {
           "refType": "CHILD_OF",
-          "traceId": "AAAAAAAAAAAAAAAAAAAA/w==",
+          "traceId": "AAAAAAAAAAAAAAAAAAAAEg==",
           "spanId": "AAAAAAAAAP8="
         },
         {
-          "refType": "CHILD_OF",
-          "traceId": "AAAAAAAAAAAAAAAAAAAAAQ==",
-          "spanId": "AAAAAAAAAAI="
+          "refType": "FOLLOWS_FROM",
+          "traceId": "AAAAAAAAAAAAAAAAAAAAEg==",
+          "spanId": "AAAAAAAAAP8="
         },
         {
           "refType": "FOLLOWS_FROM",
```

I'll go ahead and open a PR with these changes. If you feel this is not the correct approach feel free to close both the issue and PR, or let me know and I'll close them.

Thanks.

### Open questions

_No response_

alejandrodnm:
Resolves #3918

Set the fixture `tags_opname_trace.json` used in the integration test

TestExternalGRPCStorage/FindTraces/Tags_+_Operation_name

To have the following references:

- A CHILD_OF to a different span that shares the same traceId.
- A FOLLOWS_FROM to the same span used in the `CHILD_OF` relationship.
- A FOLLOWS_FROM to a span belonging to a different trace.

### DIFF
--- a/plugin/storage/integration/fixtures/traces/tags_opname_trace.json
+++ b/plugin/storage/integration/fixtures/traces/tags_opname_trace.json
@@ -7,13 +7,13 @@
       "references": [
         {
           "refType": "CHILD_OF",
-          "traceId": "AAAAAAAAAAAAAAAAAAAA/w==",
+          "traceId": "AAAAAAAAAAAAAAAAAAAAEg==",
           "spanId": "AAAAAAAAAP8="
         },
         {
-          "refType": "CHILD_OF",
-          "traceId": "AAAAAAAAAAAAAAAAAAAAAQ==",
-          "spanId": "AAAAAAAAAAI="
+          "refType": "FOLLOWS_FROM",
+          "traceId": "AAAAAAAAAAAAAAAAAAAAEg==",
+          "spanId": "AAAAAAAAAP8="
         },
         {
           "refType": "FOLLOWS_FROM",


plugin/storage/integration/fixtures/traces/tags_opname_trace.json
```diff
@@ -7,13 +7,13 @@
...
           "refType": "CHILD_OF",
-          "traceId": "AAAAAAAAAAAAAAAAAAAA/w==",
+          "traceId": "AAAAAAAAAAAAAAAAAAAAEg==",
           "spanId": "AAAAAAAAAP8="
         },
         {
-          "refType": "CHILD_OF",
-          "traceId": "AAAAAAAAAAAAAAAAAAAAAQ==",
-          "spanId": "AAAAAAAAAAI="
+          "refType": "FOLLOWS_FROM",
```
yurishkuro, line 14: Wouldn't this reduce the effectiveness of the test? Jaeger allows multiple child-of refs so that users can model fork-join relationships

yurishkuro: Your reference to OpenMetrics threw me off a bit, until I clicked on the link that referred to OpenTracing.

Yes, `opentracing.ref_type` is what should be used, it's why it was added to the transformation rules.

alejandrodnm: OMG, my heads a mess right now with the names of things.

Thanks :)

PR #3922, opened 2022-09-23
Fix empty operationNames used in storage integration tests

arajkumar:
Few fixtures in storage integration tests has empty operationName which doesn't soundcorrect.

Signed-off-by: Arunprasad Rajkumar <ar.arunprasad@gmail.com>

### DIFF
--- a/plugin/storage/integration/fixtures/traces/default.json
+++ b/plugin/storage/integration/fixtures/traces/default.json
@@ -3,7 +3,7 @@
     {
       "traceId": "AAAAAAAAAAAAAAAAAAAAEQ==",
       "spanId": "AAAAAAAAAAM=",
-      "operationName": "",
+      "operationName": "placeholder",
       "references": [],
       "startTime": "2017-01-26T16:46:31.639875Z",
       "duration": "100000ns",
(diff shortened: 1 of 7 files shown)


arajkumar:
@yurishkuro I'm not sure whether `span.operationName` can be empty, though semantically it doesn't add any value. We([timescale/promscale](https://github.com/timescale/promscale)) encountered the problem when running these tests against our jaeger storage implementation.

yurishkuro:
What kind of issue did you run into? Empty operation name seems like a reasonable edge case.

yurishkuro:
I don't think Jaeger requires this constraint, but it makes sense. Do OpenTelemetry SDK enforce span_name being non-empty?

arajkumar:
I had gone through the [otel sdk](https://github.com/open-telemetry/opentelemetry-go/blob/d7bfe6675fe79b0c01cae3d4c5f06330b2fece73/sdk/trace/tracer.go#L37), it doesn't seem to enforce non empty span name.

yurishkuro:
Right, that's sort of what I thought, so why do you want that restraint in the database?  An empty string in operation name doesn't really break things. The alternative is to have a place somewhere in the pipeline that will fix the name to something like "unknown" (e.g. Jaeger does it that for service name).

arajkumar:
Yup, we have a[ similar workaround](https://github.com/timescale/promscale/pull/1627/commits/3b54ddf6188eee75dd9a6e49754e692fa8f1f594), just wanted to know this can be fixed in the test origin itself. Thanks @yurishkuro for your thoughts. I will close this issue.

PR #3944, opened 2022-10-04
Expose storage integration helpers as go pkg

arajkumar:
gRPC storage implementations which lives out of tree has to copy integration test helpers and fixtures.

This commit exposes storage integration test helpers as golang packages so that out of tree grpc storage implementations can import and run tests from their context.

Prior to this change out of tree implementations copies these files into their repo[1]. This is not good in terms of keeping fixtures upto date.

This commit does the following,
- Rename test helper files(integration_test.go => integration.go, trace_compare_test.go => test_compare.go) to expose those as go packages.
- Use `go embed` instead of `os.ReadFile` package to read test fixtures, this helps to reuse fixtures
- Add a field in StorageIntegration struct to skip tests

[1] https://github.com/jaegertracing/jaeger-clickhouse/tree/main/integration

Signed-off-by: Arunprasad Rajkumar ar.arunprasad@gmail.com


yurishkuro:
perhaps we should add `.nocover` to this package, because it's going to be dragging the coverage down, since it has a bunch of code that is only executed in case of test failures, so to have it covered we'd need to simulate those failures with some mock test suite (ideally we'd do exactly that, but it's some extra work).

The `.nocover` should contain FIXME, e.g.
```
FIXME - branches for storage test failures are not covered
```

yurishkuro (approved):

yurishkuro:
my bad, `.nocover` only takes effect in dirs that have `.go` files without tests.

arajkumar:
Thanks @yurishkuro

PR #3947, opened 2022-10-05
Separate index teamplates for primary and archive

naveedyahyazadeh:
- 

-

### DIFF
--- a/plugin/storage/es/mappings/jaeger-span-archive.json
+++ b/plugin/storage/es/mappings/jaeger-span-archive.json
@@ -1,11 +1,10 @@
 {
   "template": [
-    "*jaeger-span-*",
-    "-jaeger-span-archive"
+    "jaeger-span-archive"
   ],
   "settings":{
-    "index.number_of_shards": 3,
-    "index.number_of_replicas": 3,
+    "index.number_of_shards": {{ .Shards }},
+    "index.number_of_replicas": {{ .Replicas }},
     "index.mapping.nested_fields.limit":50,
     "index.requests.cache.enable":true,
     "index.mapper.dynamic":false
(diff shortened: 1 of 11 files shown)


plugin/storage/es/mappings/jaeger-span.json
```diff
@@ -1,5 +1,8 @@
 {
-  "template": "*jaeger-span-*",
+  "template": [
+    "*jaeger-span-*",
+    "-jaeger-span-archive"
```
yurishkuro, line 4: 1. why does this not have `*`s at both ends?
2. if this template already covers archive, why do you need `jaeger-span-archive.json`?

plugin/storage/es/mappings/jaeger-span-archive.json
```diff
@@ -0,0 +1,167 @@
+{
```
yurishkuro, line 1: couldn't we just reuse `jaeger-span.json`?

yurishkuro (commented):
do you think you could add some integration tests under plugin/storage/integration to validate the archive functionality?

plugin/storage/es/mappings/jaeger-span.json
yurishkuro: What I meant was: why can't we parameterize this file and reuse it for both, instead of creating a nearly identical duplicate? E.g.
```
  "template": [
    "{{ .Template1 }}",
    "{{ .Template2 }}"
```

naveedyahyazadeh: > 

Is it possible to run the integration tests on local? I don't seem to be able to get past the health check stage of `elasticsearch_test.go`

yurishkuro: I think you should be able to run the test with `bash scripts/es-integration-test.sh elasticsearch 7.14.0`

PR #3956, opened 2022-10-08
Generate and publish sbom on release

### LINKED ISSUE #3943: [Feature]: Implementations for 100% Security Compliance for CLOmonitor
### Requirement

To receive awards for compliance with CNCF best practices, Jaeger needs to implement SBOMs that will clear [this check](https://clomonitor.io/docs/topics/checks/#software-bill-of-materials-sbom). 

### Problem

Software Bill of Materials (SBOM) - is a list of all the dependencies that went into your current build and deliverables. It contains more information about your project than an usual lockfile, also compiling individual dependencies license, security and origin information into one neat packet.
SBOMS are increasingly becoming a requirement to be provided with deliverables. For example the recent [NSA, CISA, ODNI Software Supply Chain Guidance for Developers](https://www.nsa.gov/Press-Room/News-Highlights/Article/Article/3146465/nsa-cisa-odni-release-software-supply-chain-guidance-for-developers/) recommend that producers of open source projects provide a SBOM for every project as a mitigation and assurance of good development practice.
SBOMS can also be checked by users adopting software, creating visibility into what constituent dependencies were used to build a given package, creating more transparency and visibility into security issues, mitigations and footprint. This information can be used by them to assess the security, as well as provide feedback and contributions back to your project.
Adopting SBOMs is easy. There are two standards out today, [CycloneDX](https://cyclonedx.org/) and  [SPDX](https://github.com/spdx/tools). CLOMonitor by CNCF can automatically check your project for the presence of these files and give you guidance on how to get started creating one.

### Proposal

_No response_

### Open questions

Two additional security compliance measures not detailed here but needed for 100% completion are: 
1. Token permissions
2. Updating the docs for signing releases

mmorel-35:

- Fixes #3943

- Generate a sbom as a bom.json file and publish it on release

### DIFF
--- a/.github/workflows/ci-release.yml
+++ b/.github/workflows/ci-release.yml
@@ -41,10 +41,16 @@ jobs:
       run: bash scripts/package-deploy.sh
       if: steps.build-binaries.outcome == 'success'
 
+    - name: Generate SBOM
+      uses: CycloneDX/gh-gomod-generate-sbom@v1
+      with:
+        version: v1
+        args: mod -licenses -json -output bom.json
+
     - name: Upload binaries
       uses: svenstaro/upload-release-action@2.3.0
       with:
-        file: '{deploy/*.tar.gz,deploy/*.zip,deploy/*.sha256sum.txt}'
+        file: '{bom.json,deploy/*.tar.gz,deploy/*.zip,deploy/*.sha256sum.txt}'
         file_glob: true
         tag: ${{ github.ref }}
         repo_token: ${{ secrets.GITHUB_TOKEN }}


jkowall:
Thanks for doing this! I can add the link to Readme after.

yurishkuro:
This will only publish SBOM for Go code. We also have UI code, and the content of the base Docker images.
* Should we publish more than one SBOM, at minimum binaries-sbom (Go + UI) and images-sbom (including image layers)?
* Should we be using syft-based https://github.com/anchore/sbom-action instead of CycloneDX? syft seems to cover both source code and binaries.

jkowall:
> This will only publish SBOM for Go code. We also have UI code, and the content of the base Docker images.
> 
> * Should we publish more than one SBOM, at minimum binaries-sbom (Go + UI) and images-sbom (including image layers)?
> * Should we be using syft-based https://github.com/anchore/sbom-action instead of CycloneDX? syft seems to cover both source code and binaries.

That's why I just used Syft on the base all-in-one image, which should get us where we need to be. This seems like a one-time thing, since we don't add a huge amount of code and libraries to the project at this point. Probably over-engineering it to build it into the pipeline. Most folks who really need to track libraries in projects will use a service like Snyk, Whitesource, etc they will not just rely on a file published in a project. I considered this work as a box checking exercise.

yurishkuro:
> Probably over-engineering it to build it into the pipeline.

I don't think so - we have dependabot enabled and update dependencies _constantly_, so the one-off SBOM will be out of date usually by the next release. If we can't automate it, I'd rather not publish SBOM at all than publish a misleading one.

E.g. this year, 92 dependency bumps (not counting manual ones):
```
$ gl main@{2022-01-01}..HEAD | grep dependabot | wc -l
      92
```

yurishkuro:
Done in #3987

PR #3966, opened 2022-10-12
Add support for OpenSearch 2.x

gaurav-05:
- 

-


.github/workflows/ci-opensearch.yml
```diff
@@ -16,6 +16,9 @@ jobs:
         - major: 1.x
           image: 1.0.0
           distribution: opensearch
+        - major: 2.x
```
albertteoh, line 19: I think we might need to add a test case for 2.x in: https://github.com/jaegertracing/jaeger/blob/main/pkg/es/client/cluster_client_test.go

gaurav-05: Added, please review.

go.mod
```diff
@@ -29,7 +29,7 @@ require (
 	github.com/grpc-ecosystem/grpc-opentracing v0.0.0-20180507213350-8e809c8a8645
 	github.com/hashicorp/go-hclog v1.3.1
 	github.com/hashicorp/go-plugin v1.4.5
-	github.com/kr/pretty v0.3.0
+	github.com/kr/pretty v0.3.1
```
yurishkuro, line 32: Please exclude go.mod changes from this PR, it's better to commit dependency upgrades separately.

gaurav-05: I updated my branch, seems that was an update issue as those changes were not from my PR. You may please check now.

yurishkuro (approved):
thank you

PR #3983, opened 2022-10-18
Bump sarama to 1.33.0

pavolloffay:

Sarama 1.32.0 should not be used https://github.com/Shopify/sarama/blob/main/go.mod#L42 

1.33.0 seems to be the latest sarama version that we can use. 1.34.0 does not work due to dependency on the deprecated sarama-cluster.


albertteoh (approved):
There are some merge conflicts.

pavolloffay:
PR rebased

PR #3996, opened 2022-10-24
Add instructions to send traces in SPM Dev Env

albertteoh:
- Follow up to: https://github.com/jaegertracing/jaeger/pull/3986#discussion_r999526750

- Remove unused jaeger all-in-one exposed ports.
- Add instructions explaining how to traces from a custom application with examples and screenshots.


docker-compose/monitor/README.md
```diff
@@ -50,27 +94,27 @@ curl "http://localhost:16686/api/metrics/calls?service=driver&service=frontend&g
...
 
-## Example 4
+### Example 4
 Fetch the minimum step size supported by the underlying metrics store.
 ```bash
 curl "http://localhost:16686/api/metrics/minstep" | jq .
 ```
 
-# HTTP API
+# HTTP API Specification
```
albertteoh, line 117: I considered defining this as a swagger spec, but given its only "user" is Jaeger UI, I don't think it's worth the effort.

docker-compose/monitor/jaeger_example.py
```diff
@@ -0,0 +1,30 @@
...
+
+resource = Resource(attributes={
+    "service.name": "my_service"
+})
+
+trace.set_tracer_provider(TracerProvider(resource=resource))
+
+jaeger_exporter = JaegerExporter(
+    collector_endpoint="http://localhost:14278/api/traces",
+)
```
yurishkuro, line 19: I would suggest using OTLP exporter instead and exposing OTLP port in the OTEL Collector.

albertteoh: For my learning, why the preference for OTLP over Jaeger formatted spans?

Done in [ab3dd6b](https://github.com/jaegertracing/jaeger/pull/3996/commits/ab3dd6b6f5f42a6347ce7659f5c2da7b9e63e9f2).

yurishkuro: Because jaeger SDKs are deprecated, the recommended path is native OTEL

docker-compose/monitor/docker-compose.yml
```diff
@@ -11,11 +11,7 @@ services:
       - METRICS_STORAGE_TYPE=prometheus
       - PROMETHEUS_SERVER_URL=http://prometheus:9090
     ports:
-      - "14250:14250"
-      - "14268:14268"
-      - "6831:6831/udp"
       - "16686:16686"
-      - "16685:16685"
```
yurishkuro, line 18: isn't one of these ports still needed to receive data from OTEL collector?

albertteoh: No, all ports are accessible between containers. Only the ports that need to be exposed to the "outside world" should be listed under `ports`.

yurishkuro (approved):

PR #3998, opened 2022-10-24
Feat: Dedicated Opensearch Storage Backend Support

derek-ho:
Signed-off-by: Derek Ho <dxho@amazon.com>

- 

-


yurishkuro:
We already support opensearch. How is this different?

yurishkuro:
no activity, closing

Issue #3960, opened 2022-10-10
[Security]: Sign Releases


jkowall:
### Requirement

Please create a GPG key and store the private key in a secret. This will be used to modify the build shell script to create signed releases. 

http://ekaia.org/blog/2009/05/10/creating-new-gpgkey/ > To create the key

Please provide the name of the github secret for the private key.

### Problem

Need to create and store GPG key to allow for verification of the builds being from the Jaeger project

### Proposal

Modify the script here to add the signing process in https://github.com/jaegertracing/jaeger/blob/2e8c6b24492df7232b1f2814cfaa8cdf004c3ec4/scripts/package-deploy.sh#L36

per: https://wiki.debian.org/Creating%20signed%20GitHub%20releases

yurishkuro:
I suggest someone first creates a PR that demonstrates this process working (using a personal private key set in the fork).

For instance, the article suggests using `gpg --armor --detach-sign`, which (a) picks the key from the local key ring, and (b) asks for a passphrase to decode it from the ring. It's not going to work in a shell script like that exactly.

Open question: we already have a key pair known to maintainers (used to encrypt/decrypt other project secrets), should we use that one or create a new dedicated key for release signing?

yurishkuro:
cc @jaegertracing/jaeger-maintainers

yurishkuro:
This looks like an example of dealing with gpg keys in a script: https://github.com/operator-framework/operator-sdk/pull/4034/files#diff-cf3a54e7ea95d0803bdc9605ebf170aa569cfef0fc83d1f1a78cde09215ce133

PR #4033, opened 2022-11-09
Add the signing of releases to build process

jkowall:
Signed-off-by: jkowall <jkowall@kowall.net>

Resolves https://github.com/jaegertracing/jaeger/issues/3960 

This adds in signing to the build process, the following steps must be taken to configure 2 new secrets in the repo by another team member.

1. Generate a GPG key : gpg --gen-key
   Answer the prompts accordingly for the email specified (you can set a password as well)

2. pgp --list-keys
  Find the GUID for the key you want to use ex: 79FAF16A8C42B3593143DD7C8DE95E7E9BC9CA3C 

3. Get the contents of the key: gpg --export-secret-keys -a 8DE95E7E9BC9CA3C
4. Copy and paste (or pipe to a file and use that). This should be added as a secret called GPG_PRIVATE_KEY
 
5. The password you created should be added as a secret called GPG_PASSWORD

Please let me know if you have questions. We can either re-run the last build or wait for the next release to test it.

### DIFF
--- a/.github/workflows/ci-release.yml
+++ b/.github/workflows/ci-release.yml
@@ -43,6 +43,12 @@ jobs:
     - name: Install tools
       run: make install-ci
 
+    - name: Configure GPG Key
+      uses: crazy-max/ghaction-import-gpg@v1
+      env:
+        GPG_PRIVATE_KEY: ${{ secrets.GPG_PRIVATE_KEY }}
+        PASSPHRASE: ${{ secrets.GPG_PASSWORD }}
+
     - name: Build binaries
       id: build-binaries
       run: make build-all-platforms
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-release.yml
```diff
@@ -43,6 +43,12 @@ jobs:
     - name: Install tools
       run: make install-ci
 
+    - name: Configure GPG Key
+      uses: crazy-max/ghaction-import-gpg@v1
+      env:
+        GPG_PRIVATE_KEY: ${{ secrets.GPG_PRIVATE_KEY }}
+        PASSPHRASE: ${{ secrets.GPG_PASSWORD }}
```
yurishkuro, line 50:
```suggestion
        PASSPHRASE: ${{ secrets. GPG_PASSPHRASE }}
```

jkowall:
One missing step if you want to add the key into the keyserver:

gpg --keyserver pgp.mit.edu --send-key 79FAF16A8C42B3593143DD7C8DE95E7E9BC9CA3C

yurishkuro (approved):

PR #4050, opened 2022-11-16
Fix httpmetrics deadlock cause by panic

### LINKED ISSUE #2944: Prometheus 'Label value is not valid UTF-8' cause API timeout
**Describe the bug**
Prometheus 'Label value is not valid UTF-8' cause API request timeout.

**To Reproduce**
1. docker run -d -e COLLECTOR_ZIPKIN_HOST_PORT=:9411   -p 5775:5775/udp   -p 6831:6831/udp   -p 6832:6832/udp   -p 5778:5778   -p 16686:16686   -p 14268:14268   -p 14250:14250   -p 9411:9411   jaegertracing/all-in-one:1.22
2. curl http://localhost:9411/%C0%AE%C0%AE
3. Then all the other requests to http://localhost:9411/ will be timeout.

log:
```
2021/04/18 06:51:16 http: panic serving 172.17.0.1:33336: label value "/\xc0\xae\xc0\xae" is not valid UTF-8
goroutine 195 [running]:
net/http.(*conn).serve.func1(0xc000489f40)
	net/http/server.go:1801 +0x147
panic(0x1392360, 0xc0002c3630)
	runtime/panic.go:975 +0x47a
github.com/prometheus/client_golang/prometheus.(*HistogramVec).WithLabelValues(0xc0004fc500, 0xc000575770, 0x3, 0x3, 0xc000575650, 0xc000575770)
	github.com/prometheus/client_golang@v1.5.1/prometheus/histogram.go:489 +0xda
github.com/uber/jaeger-lib/metrics/prometheus.(*Factory).Timer(0xc0000edc20, 0x15b70ed, 0x15, 0xc000575620, 0x15bbfcd, 0x19, 0x0, 0x0, 0x0, 0x1512b80, ...)
	github.com/uber/jaeger-lib@v2.4.0+incompatible/metrics/prometheus/factory.go:181 +0x2ce
github.com/uber/jaeger-lib/metrics/fork.(*Factory).Timer(0xc0003fe870, 0x15b70ed, 0x15, 0xc000575620, 0x15bbfcd, 0x19, 0x0, 0x0, 0x0, 0xc0006799f5, ...)
	github.com/uber/jaeger-lib@v2.4.0+incompatible/metrics/fork/fork.go:48 +0x65
github.com/jaegertracing/jaeger/pkg/httpmetrics.buildTimer(...)
	github.com/jaegertracing/jaeger/pkg/httpmetrics/metrics.go:111
github.com/jaegertracing/jaeger/pkg/httpmetrics.(*requestDurations).record(0xc00056c090, 0xc0000bbc80, 0x3, 0xc00057a858, 0x5, 0xc00057a940, 0x3, 0x10578)
	github.com/jaegertracing/jaeger/pkg/httpmetrics/metrics.go:101 +0x465
github.com/jaegertracing/jaeger/pkg/httpmetrics.Wrap.func1(0x1e1c2c0, 0xc000158380, 0xc00014ea00)
	github.com/jaegertracing/jaeger/pkg/httpmetrics/metrics.go:63 +0x1fe
net/http.HandlerFunc.ServeHTTP(0xc00050de60, 0x1e1c2c0, 0xc000158380, 0xc00014ea00)
	net/http/server.go:2042 +0x44
github.com/rs/cors.(*Cors).Handler.func1(0x1e1c2c0, 0xc000158380, 0xc00014ea00)
	github.com/rs/cors@v1.7.0/cors.go:219 +0x1b9
net/http.HandlerFunc.ServeHTTP(0xc00050de80, 0x1e1c2c0, 0xc000158380, 0xc00014ea00)
	net/http/server.go:2042 +0x44
net/http.serverHandler.ServeHTTP(0xc0000d0380, 0x1e1c2c0, 0xc000158380, 0xc00014ea00)
	net/http/server.go:2843 +0xa3
net/http.(*conn).serve(0xc000489f40, 0x1e225c0, 0xc00025a400)
	net/http/server.go:1925 +0x8ad
created by net/http.(*Server).Serve
	net/http/server.go:2969 +0x36c
```

**Expected behavior**
 Other requests not timeout.

**Screenshots**
null

 - OS: MacOS & Linux
 - Jaeger version: 1.22
 - Deployment: Docker

The error is reported by Prometheus, so I add `-e METRICS_BACKEND=expvar` to disable Prometheus.

**Additional context**
null

ianzhang1988:
- fix #2944 
- `buildTimer` would panic, cause lock not being release.

- put unlock in `defer` statement. so lock can release after panic.
- put `httpmetrics.Wrap` inside `recoveryHandler`. just seems to be right.

### DIFF
--- a/cmd/collector/app/server/http.go
+++ b/cmd/collector/app/server/http.go
@@ -89,7 +89,7 @@ func serveHTTP(server *http.Server, listener net.Listener, params *HTTPServerPar
 	cfgHandler.RegisterRoutes(r)
 
 	recoveryHandler := recoveryhandler.NewRecoveryHandler(params.Logger, true)
-	server.Handler = httpmetrics.Wrap(recoveryHandler(r), params.MetricsFactory)
+	server.Handler = recoveryHandler(httpmetrics.Wrap(r, params.MetricsFactory))
 	go func() {
 		var err error
 		if params.TLSConfig.Enabled {
--- a/pkg/httpmetrics/metrics.go
+++ b/pkg/httpmetrics/metrics.go
@@ -91,17 +91,25 @@ func newRequestDurations(metricsFactory metrics.Factory) *requestDurations {
 func (r *requestDurations) record(request recordedRequest) {
 	cacheKey := request.key
 
-	r.lock.RLock()
-	timer, ok := r.timers[cacheKey]
-	r.lock.RUnlock()
-	if !ok {
-		r.lock.Lock()
+	var timer metrics.Timer
+	var ok bool
+
+	func() {
+		r.lock.RLock()
+		defer r.lock.RUnlock()
 		timer, ok = r.timers[cacheKey]
-		if !ok {
-			timer = buildTimer(r.metrics, cacheKey)
-			r.timers[cacheKey] = timer
-		}
-		r.lock.Unlock()
+	}()
+
+	if !ok {
+		func() {
+			r.lock.Lock()
+			defer r.lock.Unlock()
+			timer, ok = r.timers[cacheKey]
+			if !ok {
+				timer = buildTimer(r.metrics, cacheKey)
+				r.timers[cacheKey] = timer
+			}
+		}()
 	}
 
 	timer.Record(request.duration)


yurishkuro (commented):
Thanks for the PR, but I don't think this is a good solution. Generating metrics should never affect the execution of requests. Your solution would avoid crashing the process, but it still fails the incoming request.

I think the best solution would be to sanitize the input better and ensure that we don't pass illegal characters to Prometheus SDK.

ianzhang1988:
Yes, this maybe not a perfect fit for the issue. 

Although at least it's a safer way of using lock. And allow other correct requests being processed.

yurishkuro:
It may be safer, but it results in invalid metrics being reported, especially if requests fail for other reasons than the bad characters.

ianzhang1988:
I think metrics would be droped after panic from promethues, not sure though.

I still think this could be usefull. I have being using jaeger collecting opentelemetry spans. And every few hours my service got a ileagle url ( from Network scanning which is not part of my services ). Then collector would stop responding http 202 to opentelemetry client causing timeout and drop spans. And collector memory useage increase over time.  Stay in this state untill OOM kill. Take me days to find this cause.

I would argure that this may save others some time locating this problem in a similar network environment.

yurishkuro:
It may be fixing one specific issue, but it's a dirty fix as it introduces other issues. I have a cleaner fix in #4051.

PR #4056, opened 2022-11-18
Fix cassandra schema scripts to be able to run from a remote node

aoterolorenzo:
Signed-off-by: Alberto Otero Lorenzo <lorenzoa@vmware.com>

The `create.sh` file from the cassandra-schema migration scripts for Jaeger, executes an 'a priori' inoffensive
command `cqlsh -e "show version"`. This command shouldn't need of authentication, and it doesn't... But only when the target is the local cassandra node (`$CQLSH_HOST` points to `localhost`). Otherwise, the CLI tool will report an error asking for auth when running the scrips over a remote node:

```
Generating the schema for the keyspace jaeger_v1_dc1 and datacenter dc1
Connection error: ('Unable to connect to any servers', {'10.101.195.23:9042': AuthenticationFailed('Remote end requires authentication')})
Using template file /cassandra-schema/v004.cql.tmpl with parameters:
    mode = test
    datacenter = dc1
    keyspace = jaeger_v1_dc1
    replication = {'class': 'SimpleStrategy', 'replication_factor': '1'}
    trace_ttl = 172800
    dependencies_ttl = 0
```

An if is added to the logic for `create.sh`'s command `cqlsh -e "show version"` to be executed with authentication arguments in case a password is set. This doesn't check if the node where the CQL is executed is remote or not, but instead it provides auth in case it is provided, local or remote, covering all use cases.


plugin/storage/cassandra/schema/create.sh
```diff
@@ -26,11 +26,19 @@ dependencies_ttl=${DEPENDENCIES_TTL:-0}
...
-    | awk -F "." '{print $1}' \
-)
+if [ -z "$PASSWORD" ]; then
+  cas_version=$(cqlsh -e "show version" \
+      | awk -F "|" '{print $2}' \
+      | awk -F " " '{print $2}' \
+      | awk -F "." '{print $1}' \
+  )
+else
+  cas_version=$(cqlsh -u ${USER} -p ${PASSWORD} -e "show version" \
```
yurishkuro, line 36: Is there another way to pass credentials? Passing via cmd line arguments is not secure, they can be observed via ps.

aoterolorenzo: I'm afraid not. Seems only `$CQLSH_HOST` and `$CQLSH_PORT` can be provided as env vars, and the authentication can only be performed through CLI args or plain config file ([see ref](https://docs.datastax.com/eol/en/security/6.0/security/secCqlshCredentials.html)).

In fact, there's already calls of this kind on the parent file [`docker.sh`](https://github.com/jaegertracing/jaeger/blob/main/plugin/storage/cassandra/schema/docker.sh#L21-L24):
```
if [ -z "$PASSWORD" ]; then
  ${CQLSH} ${CQLSH_SSL} ${CQLSH_HOST} ${CQLSH_PORT} -e "describe keyspaces"
else
  ${CQLSH} ${CQLSH_SSL} ${CQLSH_HOST} ${CQLSH_PORT} -u ${USER} -p ${PASSWORD} -e "describe keyspaces"
```

```
if [ -z "$PASSWORD" ]; then
  MODE="${MODE}" DATACENTER="${DATACENTER}" KEYSPACE="${KEYSPACE}" /cassandra-schema/create.sh "${TEMPLATE}" | ${CQLSH} ${CQLSH_SSL} ${CQLSH_HOST} ${CQLSH_PORT}
else
  MODE="${MODE}" DATACENTER="${DATACENTER}" KEYSPACE="${KEYSPACE}" /cassandra-schema/create.sh "${TEMPLATE}" | ${CQLSH} ${CQLSH_SSL} ${CQLSH_HOST} ${CQLSH_PORT} -u ${USER} -p ${PASSWORD}
fi
```

Is simply that `cqlsh -e "show version"` is so simple and straight forward that you don't think about a need of authentication to use it, so it's not until you run it from a remote machine when you realize of this (which indeed makes complete sense, to need auth for executing ANY command or query over a remote end).

yurishkuro: Also, the version check is only necessary if template is not provided.

aoterolorenzo: I think for the moment this would keep the current script behavior solving the problem and keeping consistency with the rest of the code, so I would accept it as a fix. Of course a proper refactor would be welcome if you guys find the time 🙂

aoterolorenzo:
Hi there! Any update on this?

yurishkuro:
I have alternative fix

PR #4083, opened 2022-12-02
Restore NewGRPCHandler method

### LINKED ISSUE #4082: [Bug]: `cmd/query/app.NewGRPCHandler` removed without alternatives
### Requirement

Ability to get a GRPCHandler which can be used with a custom server

### Problem

In #3091, it looks like `app.NewGRPCHandler` was [removed](https://github.com/jaegertracing/jaeger/pull/3091/files#diff-a6e90941a25731f98ece903b59eaed1d2b921007f34eb26ff66d21a17567a9edL45-L51) accidentally - as this is a exported function, it breaks semver.

### Proposal

Add `app.NewGRPCHandler` back.

### Open questions

_No response_

vprithvi:
Signed-off-by: Prithvi Raj <vprithvi@users.noreply.github.com>

- Resolves #4082

- Add a NewGRPCHandler method. This method does not have the same signature as the previous method, so it is still breaking semver, but it provides a straightforward fix for users who which to customize the server.


cmd/query/app/grpc_handler.go
```diff
@@ -57,6 +57,39 @@ type GRPCHandler struct {
...
+	options GRPCHandlerOptions,
+) *GRPCHandler {
+	if options.Logger == nil {
+		options.Logger = zap.NewNop()
+	}
+
+	if options.Tracer == nil {
+		options.Tracer = opentracing.NoopTracer{}
+	}
+
```
yurishkuro, line 79: please add a test that passes empty options struct and validates that all 3 fields are not-nil in the result

cmd/query/app/grpc_handler_test.go
```diff
@@ -155,15 +155,13 @@ func newGRPCServer(t *testing.T, q *querysvc.QueryService, mq querysvc.MetricsQu
 		)
 	}
 	grpcServer := grpc.NewServer(grpcOpts...)
-	grpcHandler := &GRPCHandler{
-		queryService:        q,
-		metricsQueryService: mq,
-		logger:              logger,
-		tracer:              tracer,
-		nowFn: func() time.Time {
+	grpcHandler := NewGRPCHandler(q, mq, GRPCHandlerOptions{
```
yurishkuro, line 158: +1 to actually use the func to prevent it from being deleted

yurishkuro (approved):

PR #4097, opened 2022-12-08
Fix ci-release.yml perms

joe-elliott:
Release CI is currently failing with:

image

Currently testing this change in my fork

### DIFF
--- a/.github/workflows/ci-release.yml
+++ b/.github/workflows/ci-release.yml
@@ -9,7 +9,7 @@ on:
 # and https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions
 permissions:
   deployments: write
-  contents: read
+  contents: read,write
 
 jobs:
   publish-release:


yurishkuro (approved):

.github/workflows/ci-release.yml
```diff
@@ -9,7 +9,7 @@ on:
 # and https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions
 permissions:
   deployments: write
-  contents: read
+  contents: read,write
```
yurishkuro, line 12: oh-oh
```
The workflow is not valid. .github/workflows/ci-release.yml (Line: 12, Col: 13): Unexpected value 'read,write'
```

yurishkuro: maybe this needs an array syntax `["read","write"]`?

yurishkuro: yeah,  would expect write to include read

yurishkuro: but also curious what "content" this is writing (GH documentation is so bad for these, they just list them without explanation)

yurishkuro: probably this: `PATCH /repos/{owner}/{repo}/releases/assets/{asset_id} (write)`

Issue #3414, opened 2021-11-29
Product/File version in Windows executables


pocki:
## Requirement - what kind of business use case are you trying to solve?
Actual the windows executables (jaeger-agent.exe, jaeger-collector.exe, ...) don't have a version set. It is hard to identify the current used version and check if the actual used version is outdated.

![image](https://user-images.githubusercontent.com/2448303/143830707-153c3f71-a84d-42db-a0db-5b95b2525351.png)

## Problem - what in Jaeger blocks you from solving the requirement?
Checking the actual used version and identity an outdated version.
We have some applications running native on Windows (Server) and therefore the agent is also running native on Windows.

## Proposal - what do you suggest to solve the problem or improve the existing situation?
Set the Product and/or File Version of the windows executables on compile or release.

yurishkuro:
Jaeger binaries support "version" subcommand, is that not enough?

pocki:
Thanks for the information. It's a little bit hidden.
Normally the services are always running and you just want to check the version. 
On Windows users are more likely to check it the properties page instead of extra typing in a command (also when the parameter for version nearly every app uses different)

pocki:
Sorry, I am not familiar with Go and Go compiler.

smolgeat:
Hi can I take this?

yurishkuro:
@smolgeat sure

xmarcoied:
Hi @yurishkuro. 
>  I don't know if Go compiler supports setting these properties on the Windows binary, but I don't have any objections to including them.

I don't think there's a way to straightforward doing this configurations with Go Compile. but I am thinking to use a 3rd party service/repo to embed such info through a `go:generate`. What do you think about this approach? I can draft a PR for one.

yurishkuro:
That might be ok, I would recommend checking what other projects do (e.g. does Hugo have Windows binaries? Do they tag them like that?)

keremgocen:
looking for a beginner-friendly issue if this is available

yurishkuro:
it is

ananyak19:
Hey can I work on this?

yurishkuro:
Yes

ananyak19:
Hi @pocki I am a first time contributor, can you please guide on how to get started here and where I need to make the changes in order to solve the issue?

pocki:
As you can see above, I am also not familiar with Go and the Go Compiler. 
[...]

PR #4108, opened 2022-12-13
Set Product/File version in Windows executables

ananyak19:
- https://github.com/jaegertracing/jaeger/issues/3414

-


albertteoh:
Can you please provide more details on the motivation behind this PR:
- What specific problem is this trying to solve.
- What is the final outcome of this change, e.g. what the windows explorer properties would look like for the jaeger exe.
- How do you plan to manage version updates for each new release?
- How can/should we incorporate this into our release process?

yurishkuro:
Yeah, this PR is not usable as is. Judging by the file names, this may have something to do with some of these tools:
  * https://pkg.go.dev/github.com/hallazzang/syso
  * https://github.com/tc-hib/go-winres
  * https://github.com/akavel/rsrc

But the files are supposed to be generated at build time, not checked in. What we would need instead:
  * a Makefile target to generate those files
  * if the tool is a Go program, it needs to be installed like the other build-time tools we install
  * It needs to use proper Jaeger icon
  * the config JSON needs to be filled out as appropriate for Jaeger project. 
  * I am not clear if this will automatically use git tags for versions, or if we need to generate a different config file during build time where those values are populated (instead of `0.0.0.0`)

yurishkuro:
closing as there was no follow-up, feel free to reopen if comments are addressed

PR #4141, opened 2023-01-09
Support tenant header propagation in query service

pavolloffay:
Support tenant header propagation in the query HTTP APIs.

Our use case is to properly propagate tenant name (set in a configurable header) through the query service to the Tempo grpc plugin (right only the authorization header is propagated which is a bit of hack https://github.com/grafana/tempo/blob/main/cmd/tempo-query/tempo/plugin.go#L329-LL330)

### DIFF
--- a/cmd/query/app/flags.go
+++ b/cmd/query/app/flags.go
@@ -44,6 +44,7 @@ const (
 	queryStaticFiles        = "query.static-files"
 	queryUIConfig           = "query.ui-config"
 	queryTokenPropagation   = "query.bearer-token-propagation"
+	queryTenantPropagation  = "query.tenant-propagation"
 	queryAdditionalHeaders  = "query.additional-headers"
 	queryMaxClockSkewAdjust = "query.max-clock-skew-adjustment"
 )
@@ -72,6 +73,8 @@ type QueryOptions struct {
 	UIConfig string
 	// BearerTokenPropagation activate/deactivate bearer token propagation to storage
 	BearerTokenPropagation bool
+	// TenantHeaderPropagation activate/deactivate propagation of tenant header
+	TenantHeaderPropagation string
 	// TLSGRPC configures secure transport (Consumer to Query service GRPC API)
 	TLSGRPC tlscfg.Options
 	// TLSHTTP configures secure transport (Consumer to Query service HTTP API)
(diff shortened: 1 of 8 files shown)


cmd/query/app/flags.go
```diff
@@ -93,6 +96,7 @@ func AddFlags(flagSet *flag.FlagSet) {
 	flagSet.String(queryStaticFiles, "", "The directory path override for the static assets for the UI")
 	flagSet.String(queryUIConfig, "", "The path to the UI configuration file in JSON format")
 	flagSet.Bool(queryTokenPropagation, false, "Allow propagation of bearer token to be used by storage plugins")
+	flagSet.String(queryTenantPropagation, "", "Enables propagation of a tenant header")
```
yurishkuro, line 99: unclear description - is this a boolean or the name of the header?

cmd/query/app/flags.go
```diff
@@ -44,6 +44,7 @@ const (
 	queryStaticFiles        = "query.static-files"
 	queryUIConfig           = "query.ui-config"
 	queryTokenPropagation   = "query.bearer-token-propagation"
+	queryTenantPropagation  = "query.tenant-propagation"
```
yurishkuro, line 47: Please see https://github.com/jaegertracing/jaeger/pull/3688/files, there is already flag defined for the tenancy, can we not reuse that?

joe-elliott:
Personal opinion is that this change should not be tied to tenancy. It would be easier to add a flag that takes a list of headers and then propagate anything that matches.

yurishkuro:
> It would be more straightforward to add a flag that takes a list of headers and then propagate anything that matches.

+1, the more we can generalize the better. Could the same be done with bearer token though? Or would it need some kind of mapping when headers are transitioned from HTTP to gRPC?

pkg/bearertoken/context.go
```diff
@@ -14,11 +14,16 @@
...
+import (
+	"context"
+)
 
 type contextKeyType int
 
-const contextKey = contextKeyType(iota)
+const (
+	bearerTokenContextKey = contextKeyType(iota)
+	tenantHeaderContextKey
```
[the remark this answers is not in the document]
pavolloffay: It's ` pkg/tenancy` . I totally forgot about https://github.com/jaegertracing/jaeger/pull/3688. I will try to adjust the PR to use the `tenancy` package. It seems that only the ingestion part is using it for now.

pavolloffay:
superseded by #4151

PR #4151, opened 2023-01-12
Support tenant header propagation in query service and grpc-plugin

pavolloffay:

Replaces #4141

### DIFF
--- a/cmd/query/main.go
+++ b/cmd/query/main.go
@@ -163,6 +163,8 @@ func main() {
 		storageFactory.AddFlags,
 		app.AddFlags,
 		metricsReaderFactory.AddFlags,
+		// add tenancy flags here to avoid panic caused by double registration in all-in-one
+		tenancy.AddFlags,
 	)
 
 	if err := command.Execute(); err != nil {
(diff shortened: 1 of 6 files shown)


cmd/query/main.go
```diff
@@ -163,6 +163,8 @@ func main() {
 		storageFactory.AddFlags,
 		app.AddFlags,
 		metricsReaderFactory.AddFlags,
+		// add tenancy flags here to avoid panic caused by double registration in all-in-one
```
yurishkuro, line 166: what about all-in-one/main, does it need changes?

cmd/query/main.go
```diff
@@ -163,6 +163,8 @@ func main() {
...
+		tenancy.AddFlags,
```
yurishkuro, line 167: So is just adding flags sufficient to make query propagate tenant?  I was expecting to see some usage of these options.

pavolloffay: Just the flags were not exposed. I think @esnible mainly tested the all-in-one which exposed flags correctly. 

Query tenancy handler initialization:
* https://github.com/jaegertracing/jaeger/blob/b92aa3edc5b682ef0afa89463b069630f3c273cc/cmd/query/app/http_handler.go#L148
* https://github.com/jaegertracing/jaeger/blob/c1bb2946e670129314bf88d9730d8ed7566766d4/cmd/query/app/server.go#L123

yurishkuro (approved):

PR #4159, opened 2023-01-13
Watch directories for certificate hot-reload

### LINKED ISSUE #3968: [Bug]: Jaeger is not reloading the renewed certificates from filesystem even during connection setup
### What happened?

We are using Jaeger for tracing purpose. The Jaeger collector makes connection with search engine to store the trace data in TLS mode. The certificates get renewed at a periodic interval. But the renewed certificates are not reloaded by Jaeger as in the collector log it shows the below line only. 

Certificate has been removed, using the last known version ...

This line comes from https://github.com/jaegertracing/jaeger/blob/main/pkg/config/tlscfg/cert_watcher.go#L92

So Jaeger is able to recognize when the certificate gets renewed. But It is not reloading the new certificates as the below line is not printed in the Jaeger collector logs.

Loading modified certificate ...

https://github.com/jaegertracing/jaeger/blob/main/pkg/config/tlscfg/cert_watcher.go#L97

So we would like to know the expected behaviour.

### Steps to reproduce

1. Install Jaeger Collector and search engine with TLS
2. Check the renewed certificate and logs at the periodic interval
3. In the log it shows "Certificate has been removed, using the last known version ..."

### Expected behavior

Jaeger should reload the renewed certificates from filesystem even during connection setup

The below line should be printed in the Jaeger collector logs

Loading modified certificate ...

### Relevant log output

_No response_

### Screenshot

_No response_

### Additional context

_No response_

### Jaeger backend version

v1.36.0

### SDK

_No response_

### Pipeline

_No response_

### Stogage backend

_No response_

### Operating system

Linux

### Deployment model

Docker, Kubernetes

### Deployment configs

_No response_

tsaarni:
This PR fixes the certificate hot-reload on Kubernetes.

Resolves #3968

This PR moves the inotify watch from certificate files to monitoring the parent directory instead, as also recommended [here](https://github.com/fsnotify/fsnotify/blob/c6f5cfa163edb0f1bb78be3a77053ee14c48a3ce/backend_inotify.go#L246-L254) by the [fsnotify](https://github.com/fsnotify/fsnotify) library.

Without this change, certificate hot-reload only works if the content of certificate or key files is overwritten. That approach is rarely used for update since it is not atomic and it risks files being read in the middle of the update. Therefore more often the files are replaced by a rename operation, by swapping the old file with a new one. It allows the file to be completely written and closed before exposing it to the application. However, in this update scheme, the old file gets deleted and the current inotify watch on file level is automatically terminated. No reload happens and warning message like following was printed into log

```json
{"msg":"Certificate has been removed, using the last known version","certificate":"/certs/server-ca.pem"}
```

Example use case:

When Kubelet writes certificate and key files on Kubernetes Secret volume, the target files are symbolic links to files in a different directory - a hidden subdirectory in a same level as the files themselves. During update, that directory is swapped with a new one, while the symbolic links remains the same. This guarantees atomic swap for both certificate and key files at once.  It also means that any rename event received at the parent directory level might indicate that the files were replaced, even if name of the renamed file was not any of the files being monitored. Therefore this PR proposes checking the hashes of the files to detect changes.

With this change, following update schemes will work:
* Replace file content in-place (old behaviour: e.g. `cat newfile > cert.pem`)
* Update the files by making any rename operation in the parent directory (Kubernetes secret volume update scheme: `mv ..data_tmp ..data`)
* Update the files by renaming the target files (e.g. `mv cert.pem.tmp cert.pem`)

Signed-off-by: Tero Saarni <tero.saarni@est.tech>

### DIFF
--- a/pkg/config/tlscfg/cert_watcher.go
+++ b/pkg/config/tlscfg/cert_watcher.go
@@ -34,11 +34,11 @@ import (
 // The certificate and key can be obtained via certWatcher.certificate.
 // The consumers of this API should use GetCertificate or GetClientCertificate from tls.Config to supply the certificate to the config.
 type certWatcher struct {
+	mu           sync.RWMutex
 	opts         Options
+	logger       *zap.Logger
 	watcher      *fsnotify.Watcher
 	cert         *tls.Certificate
-	logger       *zap.Logger
-	mu           *sync.RWMutex
 	caHash       string
 	clientCAHash string
 	certHash     string
@@ -57,19 +57,20 @@ func newCertWatcher(opts Options, logger *zap.Logger) (*certWatcher, error) {
 		}
 		cert = &c
 	}
+
 	watcher, err := fsnotify.NewWatcher()
 	if err != nil {
 		return nil, err
 	}
 
 	w := &certWatcher{
-		cert:    cert,
 		opts:    opts,
-		watcher: watcher,
 		logger:  logger,
-		mu:      &sync.RWMutex{},
+		cert:    cert,
+		watcher: watcher,
 	}
-	if err := w.addWatches(watcher, opts); err != nil {
+
+	if err := w.setupWatchedPaths(); err != nil {
 		watcher.Close()
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


tsaarni:
Hi @pavolloffay, I noticed that you did the certificate reload implementation few years back. It would be greatly appreciated if you could take a look at this PR! Thank you :pray:

tsaarni:
I've corrected a formatting mistake. @yurishkuro could you please re-trigger the tests. Thank you!

tsaarni:
Coverage results were accidentally decreased, hopefully fixed now with the latest fix. Not sure about [CIT OpenSearch failure](https://github.com/jaegertracing/jaeger/actions/runs/3917247108/jobs/6697303016#step:9:554) though. It looks to me there might have been similar errors before with that same job, maybe a fluke?

tsaarni:
Yes apparently CIT OpenSearch failure was a fluke. However coverage was still under target %. I think one test was missing even before: replace CA certs with bad ones during update. I added that now. Besides that one, the missed lines are error branches that might be difficult to produce without stubbing the library calls.

yurishkuro:
Yeah OS tests have been failing quite often recently. I'll rerun them till they pass. 

Just curious, why was it necessary to regenerate test certificates?

tsaarni:
> Yeah OS tests have been failing quite often recently. I'll rerun them till they pass.

Ok, thank you!

> Just curious, why was it necessary to regenerate test certificates?

I wanted to have second server certificate for new test case `TestReload_kubernetes_secret_update()` and therefore I modified `gen-certs.sh` to generate that (`example-server-2-key.pem` and `example-server-2-cert.pem`). When running that script, it regenerated the old test certs too. It is not completely necessary, it can be avoided as well.

BTW as a shameless plug, I've created [a tool to generate test certs]( https://github.com/tsaarni/certyaml) with [Go API](https://github.com/tsaarni/certyaml#go-api)  to avoid storing test certs and keys in the first place. But since it is just maintained by me alone, I assume Jaeger does not want that as a new dependency :-)

yurishkuro:
I merged some fixes for opensearch integration test, please rebase, hopefully it should clear the failure

tsaarni:
@yurishkuro I've now rebased, as well as removed the test data changes that were not necessary and squashed commits.

yurishkuro:
@tsaarni I added a couple commits to DRY-up some of the code, but was not able to push changed to your fork. Do you want to cherry-pick the two extra commits from #4165? Otherwise I could merge that new PR directly.

tsaarni:
@yurishkuro It is OK to me if you push #4165 instead of this PR. Thank you!

tsaarni:
> @yurishkuro It is OK to me if you push #4165 instead of this PR. Thank you!

I ended up doing cherry-pick here as well but either way is OK!

yurishkuro (approved):
Thanks!

PR #4161, opened 2023-01-15
Refactor scripts/es-integration-test.sh

yurishkuro:
- OpenSearch integration tests very frequently fail
- This change adds retries, but does not fundamentally fix anything else, except for making sure that we can get the container logs in case of a future failure

- Setup storage once before executing tests (even though token test does not require a real ES, it could be just a unit test)
- Remove `-rm` from storage Docker commands so that the container is still available to do `docker logs` after the timeout
- Change polling from 5s to 10s, for a total of 600sec
- If storage is not reachable in 600sec then kill the container and retry two more times
- Use trap to tear down storage (useful for local tests as otherwise container may remain if the tests fail)


yurishkuro:
@albertteoh fancy to review this too?

plugin/storage/integration/token_propagation_test.go
```diff
@@ -88,23 +88,25 @@ func TestBearTokenPropagation(t *testing.T) {
 
 	// Run elastic search mocked server in background..
 	// is not a full server, just mocked the necessary stuff for this test.
-	srv := &http.Server{Addr: ":9200"}
+	srv := &http.Server{Addr: ":19200"}
```
yurishkuro, line 91: no reason to run test server on the official ES port (the integration script also changed to run query-service against this fake service)

Issue #3766, opened 2022-06-21
Remove dependency on jaeger-lib


yurishkuro:
There are several issues with the dependency on jaeger-lib:
  * it is shared with jaeger-client-go, which is now deprecated, so the library does not need to be standalone
  * library does not use go modules, migrating it to `go mod` is non-trivial because it's already v2
  * because it cannot be migrated to go modules, it cannot be upgraded to later versions of go-kit. The older 0.11 version of go-kit has dependencies with known vulnerabilities
  * there's not much need to depend on go-kit in the first place, we're only using it for its expvar package, which is relatively trivial to implement (especially without the histograms)

There are two directions we can take:
  1. migrate to OpenTelemetry instead. This is preferable, but will involve non-trivial amount of work. Good news is that work can likely be done piecemeal.
  2. Keep using jaeger-lib API for metrics and pull it inside this repo as `internal`.
    * unfortunately, this does not fully remove dependency on jaeger-lib because we're still using jaeger-client-go (cf. #3381, #3380)
  3. Keep using jaeger-lib API for metrics but implement it with OpenTelemetry SDK. This seems like the least resistance path (need to prototype it to test viability).

(Update 2023-01-29) Current plan:
  * [x] Clone jaeger-lib into Jaeger repo (done in #3767, #3772)
  * [x] Upgrade HotROD to use OTEL SDK (via OT bridge), which removes the need for uber/jaeger-lib (done in #4187)
  * [x] Upgrade tracegen to use OTEL SDK (done in #4189)
  * [x] Upgrade collector/query/all-in-one to use OTEL SDK (via OT bridge)
  * [x] Fix crossdock (maybe we don't need SDK testing there anymore)

Issue #3381, opened 2021-11-08
Upgrade internal use of tracing to OpenTelemetry


yurishkuro:
This is aimed as an intern project for the LFX Mentorship. Please refer to https://github.com/jaegertracing/jaeger/issues/4453. [Apply here](https://mentorship.lfx.linuxfoundation.org/project/b8009398-1252-4f63-82fe-363846ccc11d).

Historically, the Jaeger backend used the OpenTracing API, with Jaeger's own Go SDK `jaeger-client-go`, for instrumenting its own internals for distributed tracing. Since Jaeger's SDKs have been deprecated (#3362), we want to upgrade the Jaeger backend to use the OpenTelemetry tracing API and SDK directly.

## Expected Outcome:
  - Replace the use of OpenTracing API with OpenTelemetry
  - Remove `jaeger-client-go` and `jaeger-lib` as dependencies
  - Remove `opentracing-go` and `opentracing-contrib/*` as dependencies
  - Switch to standard OpenTelemetry instrumentation libraries where available (e.g. for HTTP, gRPC)
  - Rethink/rework `crossdock` integration tests to test end-to-end flow with OpenTelemetry data
  - Publish a blog post on medium.com/jaegertracing documenting the experience

## Proposed approach

I don't think using OTEL-OT Bridge makes sense for this project, since it will create an extra work for adding the bridge, but will not make the actual migration any easier. Instead, I would do this in stages:
  * Right now various components of the backend take `opentracing.Tracer` as argument. As a first step, we create a wrapper object that will contain that same tracer, and pass the wrapper object to components. This will be one-time large refactoring but without any functional changes.
  * Once the first step is done, we can add OTEL Tracer to the same wrapper object.
  * Then start changing components one by one by switching them to use OTEL instrumentation.
  * When all of them have been migrated, we can remove the OpenTracing part.

Some related issues:
* [ ] https://github.com/jaegertracing/jaeger/issues/4534

yurishkuro:
Filed as a project for LFX Mentorship https://github.com/cncf/mentoring/pull/947

swastik959:
@yurishkuro hi I am new to this org and want to work on this issue I have worked on some kubernetes sigs and has a knowledge of golang

swastik959:
@yurishkuro  can you guide me through its process of getting selected

yurishkuro:
[...]

PR #4187, opened 2023-01-28
Use OpenTelemetry SDK in HotROD 🚗

### LINKED ISSUE #3380: Upgrade HotROD example to use OpenTelemetry SDK

yurishkuro:
Based on earlier PR #3390 by @rbroggi. 

- Resolves #3380

- Switch from jaeger-client-go SDK to OTel SDK paired with ot-otel bridge
- Add cli flag to select which Otel Exporter to use (Jaeger, OTLP or stdout)
- Add favicon 🚗
- Added OTEL version of rpcmetrics from jaeger-client-go
- Remove dependency on github.com/uber/jaeger-lib in HotROD (addresses one of outstanding tasks in #3766)


examples/hotrod/pkg/tracing/init.go
```diff
@@ -16,56 +16,69 @@
...
 
-	_, err := cfg.FromEnv()
-	if err != nil {
-		logger.Bg().Fatal("cannot parse Jaeger env vars", zap.Error(err))
-	}
+// Init initializes OpenTelemetry SDK and uses OTel-OpenTracing Bridge
+// to return an OpenTracing-compatible tracer.
+func Init(serviceName string, exporterType string, logger log.Factory) opentracing.Tracer {
+	once.Do(func() {
+		otel.SetTextMapPropagator(propagation.TraceContext{})
```
yurishkuro, line 45: @rbroggi your PR was mostly spot-on, this was the only thing missing, so traces did not continue across services.

albertteoh (approved):
LGTM; do you think this should be the approach (i.e. using the OTEL -> OT shim) to take for #3381?

examples/hotrod/pkg/tracing/init.go
```diff
@@ -16,56 +16,69 @@
...
-}
-
-type jaegerLoggerAdapter struct {
-	logger log.Logger
-}
 
-func (l jaegerLoggerAdapter) Error(msg string) {
-	l.logger.Error(msg)
+	tp := sdktrace.NewTracerProvider(
+		sdktrace.WithBatcher(exp, sdktrace.WithMaxExportBatchSize(1)),
```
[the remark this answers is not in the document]
yurishkuro: or, this should be removed, I used it when troubleshooting.

yurishkuro:
@albertteoh my intention was to make this change work with the original [HotROD blog post](https://medium.com/opentracing/take-opentracing-for-a-hotrod-ride-f6e3141f7941) (probably time to update it to current state). But it has a section that demonstrates generation of RPC metrics by using span observer in the tracer. This functionality is now broken, but I suspect it's possible to reproduce it via an additional span processor passed to the OTEL SDK

yurishkuro:
> do you think this should be the approach (i.e. using the OTEL -> OT shim) to take for https://github.com/jaegertracing/jaeger/issues/3381?

I don't have a strong opinion there, but it's probably better to just do a full upgrade to OTEL. I kinda wanted to try out the OT-OTEL shim in the Hotrod first, since I haven't seen any pubs describing a successful use of the shim on existing code. Not that I know it works (maybe write a blog post about it),  it may be worth doing a follow-up PR and move to OTEL directly. It will involve using different instrumentation libraries though.

examples/hotrod/pkg/tracing/init.go
rbroggi: Thx for reaching out. I have been quite in the shadows lately. Hope to come back to some contributions soon.

PR #4195, opened 2023-01-30
ci(lint): bump golangci-lint to v1.50.1

mmorel-35:
- bump golangci-lint to v1.50.1

- fixes contextlint
- disable gosec g114

### DIFF
--- a/cmd/collector/app/zipkin/http_handler.go
+++ b/cmd/collector/app/zipkin/http_handler.go
@@ -92,7 +92,7 @@ func (aH *APIHandler) saveSpans(w http.ResponseWriter, r *http.Request) {
 	var tSpans []*zipkincore.Span
 	switch contentType {
 	case "application/x-thrift":
-		tSpans, err = zipkin.DeserializeThrift(bodyBytes)
+		tSpans, err = zipkin.DeserializeThrift(r.Context(), bodyBytes)
 	case "application/json":
 		tSpans, err = zipkindeser.DeserializeJSON(bodyBytes)
 	default:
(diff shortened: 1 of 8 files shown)


cmd/collector/app/zipkin/http_handler.go
```diff
@@ -92,7 +92,7 @@ func (aH *APIHandler) saveSpans(w http.ResponseWriter, r *http.Request) {
 	var tSpans []*zipkincore.Span
 	switch contentType {
 	case "application/x-thrift":
-		tSpans, err = zipkin.DeserializeThrift(bodyBytes)
+		tSpans, err = zipkin.DeserializeThrift(r.Context(), bodyBytes)
```
yurishkuro, line 95: what's the point of passing context here? DeserializeThrift is not an IO-bound function.

mmorel-35: First there was 
> Function `DeserializeThrift` should pass the context parameter (contextcheck)

If I just use
> zipkin.DeserializeThrift(context.Background(), bodyBytes)
I get
> Non-inherited new context, use function like `context.WithXXX` or `r.Context` instead (contextcheck)

yurishkuro: I see the problem: https://github.com/jaegertracing/jaeger/blob/74d6a1250ee5c9db6aa776510927e36fe5869681/model/converter/thrift/zipkin/deserialize.go#L40-L41

That's why it's flagging this specific function vs. other functions. It's a false positive because all serialization happens in-memory, without IO, but the way Thrift APIs are defined it looks like there is an IO and requires a context.

yurishkuro (approved):

Issue #3842, opened 2022-08-02
[Feature]: Publish description and readme to Docker Hub


yurishkuro:
### Requirement

As a user downloading Jaeger images from Docker Hub, I want to be able to read a description and basic usage in the Docker Hub repositories.

### Problem

Right now most of Jaeger's Docker Hub repositories contain neither description nor the readme.

### Proposal

Use Docker Hub API to push the README files from the respective cmd/*** dirs when publishing Docker images.

Example: https://github.com/peter-evans/dockerhub-description/blob/main/src/dockerhub-helper.ts

YashasviChaurasia:
@yurishkuro i would like to work on this issue.

yurishkuro:
Great, go for it. Don't worry about the descriptions themselves, we first want to set up the process, and then we can improve the readmes.

YashasviChaurasia:
@yurishkuro 
### I have Tested the Workflow and have tested it on my personal repository to upload Readme files on two different docker-hub repos namely "yashasvichaurasia/testing" and "yashasvichaurasia/testlife"

Following are some screenshots:

<img width="1440" alt="Screenshot 2022-08-07 at 1 34 01 AM" src="https://user-images.githubusercontent.com/46622381/183264598-8bc091b0-4e17-4df1-baad-b1c37ba0db44.png">
<img width="1440" alt="Screenshot 2022-08-07 at 1 35 31 AM" src="https://user-images.githubusercontent.com/46622381/183264603-3af84a01-a86b-491c-8bcf-7fc62a285cab.png">
<img width="1440" alt="Screenshot 2022-08-07 at 1 36 39 AM" src="https://user-images.githubusercontent.com/46622381/183264608-4e92a69f-0a4b-41d7-967b-5416cf3a6c2d.png">
<img width="1440" alt="Screenshot 2022-08-07 at 1 42 58 AM" src="https://user-images.githubusercontent.com/46622381/183264723-2d11e9a7-68d6-4f85-92e4-14c19fe81633.png">

YashasviChaurasia:
I have written the jobs in '.yml' in the following template where I have assumed the admin username and password to be in the following repository secrets as we would require admin privileges while accessing docker hub while using this workflow:

```
    - name: Docker Hub Description
      uses: peter-evans/dockerhub-description@v3
      with:
        username: ${{ secrets.DOCKERHUB_USERNAME }}
        password: ${{ secrets.DOCKERHUB_PASSWORD }}
[...]

PR #4224, opened 2023-02-08
[Feature] : Update Docker Hub Description

SaarthakMaini:
This is a commit to integrate the pushing of the README.md file to the DockerHub Description to provide more information to the user

Reference Issue : #3842

-  Added a shell script to push the README.md file to the DockerHub repository to show the information about the image
-  Incorporate this script into the CI process.

### DIFF
--- a/scripts/update-dockerhub-description.sh
+++ b/scripts/update-dockerhub-description.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+image_name="all-in-one"
+readme_path="./README.md"
+token="$DOCKERHUB_TOKEN"
+repository="jaegertracing/$image_name"
+url="https://hub.docker.com/v2/repositories/$repository/"
+readme_content=$(cat $readme_path)
+payload="{\"full_description\":\"$readme_content\"}"
+
+response=$(curl -s -X PATCH \
+  -H "Authorization: JWT $token" \
+  -H "Content-Type: application/json" \
+  -d "$payload" \
+  $url)
+
+status_code=$(echo "$response" | jq -r .status_code)
+
+if [ $status_code -eq 200 ]; then
+  echo "Successfully updated description for $repository"
+else
+  echo "Failed to update description for $repository with status code $status_code"
+fi
--- a/.github/workflows/ci-docker-build.yml
+++ b/.github/workflows/ci-docker-build.yml
@@ -50,3 +50,8 @@ jobs:
       env:
         DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
         QUAY_TOKEN: ${{ secrets.QUAY_TOKEN }}
+
+    - name: Update Docker Hub Description
+      run: bash scripts/update-dockerhub-description.sh
+      env:
+        DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}


scripts/update-dockerhub-description.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+image_name="all-in-one"
```
yurishkuro, line 3: Note: there are two CI workflows that build Docker images, `ci-docker-build.yml` and `ci-all-in-one-build.yml`. You're only uploading README for one image

scripts/update-dockerhub-description.sh
```diff
@@ -0,0 +1,23 @@
...
+readme_path="./README.md"
```
yurishkuro, line 4: This is the root README? I think it's better to first check if there is a more narrow `./cmd/{binary}/README` file

scripts/update-dockerhub-description.sh
```diff
@@ -0,0 +1,23 @@
...
+token="$DOCKERHUB_TOKEN"
+repository="jaegertracing/$image_name"
+url="https://hub.docker.com/v2/repositories/$repository/"
```
yurishkuro, line 7: We upload images to both Docker Hub and quay.io. Does the latter also have an API to upload README?

scripts/update-dockerhub-description.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
```
yurishkuro, line 2: we usually do
```
set -euxf -o pipefail
```
to assist in debugging and fail fast.

scripts/update-dockerhub-description.sh
```diff
@@ -0,0 +1,23 @@
...
+image_name="all-in-one"
+readme_path="./README.md"
+token="$DOCKERHUB_TOKEN"
+repository="jaegertracing/$image_name"
+url="https://hub.docker.com/v2/repositories/$repository/"
+readme_content=$(cat $readme_path)
+payload="{\"full_description\":\"$readme_content\"}"
```
yurishkuro, line 9: L9 might cause a large output because of `set -x`. Instead of reading content in, you can create a temp file.

Also, inserting content of a file inside a JSON `"..."` string is very likely to result in malformed JSON.

scripts/update-dockerhub-description.sh
```diff
@@ -0,0 +1,23 @@
...
+
+response=$(curl -s -X PATCH \
+  -H "Authorization: JWT $token" \
```
yurishkuro, line 12: with `set -x` this will cause the token to be printed to the logs, we cannot allow that.

scripts/update-dockerhub-description.sh
```diff
@@ -0,0 +1,23 @@
...
+  -H "Content-Type: application/json" \
+  -d "$payload" \
+  $url)
+
+status_code=$(echo "$response" | jq -r .status_code)
+
+if [ $status_code -eq 200 ]; then
+  echo "Successfully updated description for $repository"
+else
+  echo "Failed to update description for $repository with status code $status_code"
```
yurishkuro, line 22: may be useful to log the full response, in case there is an error message there

yurishkuro:
Tip: in the PR description, change
  * Reference Issue : #3842
to
  * Resolves #3842
This allows the issue to auto-close on PR merge.

SaarthakMaini:
Thank you for the code review! 

I'll make the necessary changes and update the PR soon

SaarthakMaini:
I have made the necessary changes and the code can be reviewed now!

Please feel free to suggest changes and I would modify the code accordingly.

Thank You!

scripts/update-dockerhub-description.sh
```diff
@@ -0,0 +1,66 @@
+#!/bin/bash
+
+set -euxf -o pipefail
+
+if [ "${WORKFLOW}" == "ci-docker-build.yml" ]; then
+  binary="docker"
```
yurishkuro, line 6: there is no such binary / image in Jaeger named `jaegertracing/docker`. We publish 2-3 dozen different images, you can see them all here: https://hub.docker.com/u/jaegertracing

yurishkuro (commented):
My recommendation:
  * change your script to accept the exact name of the Docker image, and the exact path to the README file as arguments
  * some images today do not have dedicated README files, so log a warning instead (with path) so that we can then incrementally go and add them. I don't think it makes sense to use the top-level project README for those.
  * note that both workflows eventually call the same script `scripts/build-upload-a-docker-image.sh`. This script gets the exact name of the image and the directory where its Dockerfile is located, so you can simply call your new script from there instead of calling it directly from the workflow (substitute Dockerfile path for README path, e.g. `$(dirname $DOCKERFILE)/README.md`)

scripts/update-dockerhub-description.sh
```diff
@@ -0,0 +1,57 @@
...
+dockerhub_repository="jaegertracing/$image_name"
+dockerhub_url="https://hub.docker.com/v2/repositories/$dockerhub_repository/"
+
+quay_repository="quay.io/jaegertracing/$image_name"
+quay_url="https://quay.io/api/v1/repository/$quay_repository/description"
+
+if [ ! -f "$readme_path" ]; then
+  echo "Warning: the dedicated README file for Docker image $image_name was not found at path $readme_path"
+  echo "It is recommended to have a dedicated README file for each Docker image"
+  exit 1
```
yurishkuro, line 17: this may kill parent script, since we use `set -ex`

scripts/update-dockerhub-description.sh
```diff
@@ -0,0 +1,57 @@
...
+
+tempfile=$(mktemp)
+cat $readme_path > $tempfile
+
+set +x
+
+dockerhub_response=$(curl -s -X PATCH \
+  -H "Authorization: JWT $DOCKERHUB_TOKEN" \
+  -H "Content-Type: multipart/form-data" \
+  -F "full_description=@$tempfile" \
```
yurishkuro, line 28: is the content supposed to be encoded somehow?

scripts/build-upload-a-docker-image.sh
```diff
@@ -68,4 +68,6 @@ docker buildx build --output "${PUSHTAG}" \
 	${IMAGE_TAGS} \
 	${dir_arg}
 
+bash update-dockerhub-description.sh ${IMAGE_TAGS} ${dir_arg}
```
yurishkuro, line 71:
```suggestion
bash update-dockerhub-description.sh "${IMAGE_TAGS}" ${dir_arg}
```
IMAGE_TAGS may contain all kinds of things, do you actually need them all, or do you only need namespace/component?

yurishkuro (commented):
Please show how you tested this. I recommend testing one one image that you upload to a personal docker hub repo

scripts/update-dockerhub-description.sh
SaarthakMaini: I agree!

So what I'm thinking of doing is that I can remove "exit 1" and instead of that 

`echo "Continue script execution..."`
`readme_path=../$(realpath $(dirname $0)/README.md)`

I think 

`readme_path = ../README.md` would also work fine

Is there any better way of approaching this problem?

Thank You!

scripts/build-upload-a-docker-image.sh
SaarthakMaini: I have updated it and the final command in build-upload-a-docker-image.sh looks like:

`bash update-dockerhub-description.sh ${component_name} ${dir_arg}/README.md`

`namespace="jaegertracing"` is already covered in the update-dockerhub-description.sh

The first argument taken is the image_name and the second argument is the readme_path

scripts/update-dockerhub-description.sh
yurishkuro: what if you just call `exit 0`? The caller will treat it as successful exit.

SaarthakMaini:
I tried testing the script. 

It's giving me the following error:

`Failed to update Docker Hub README for saarthakmainiofficial/testfile with status code null
Full response: {"message":"unauthorized","errinfo":{}}
`

I'm doing the following for authorization:
`
dockerhub_response=$(curl -s -X PATCH \
  -H "Authorization: JWT <Hard Coded Docker API Token>" \
  -H "Content-Type: application/json" \
  -F "full_description=@$tempfile" \
  $dockerhub_url)`
  
  I'm trying to look into it and solve the issue.
  
  I would like to ask suggestions on what I can do about it in the meanwhile
  
  Thank You!

yurishkuro:
> It's giving me the following error:

I would recommend (a) doing `docker login` and (b) publishing under your own DockerHub namespace as a test.

PR #4231, opened 2023-02-10
Support secure OTLP exporter config for hotrod

graphaelli:

hotrod currently forces all connections to insecure in https://github.com/jaegertracing/jaeger/blob/fedeb4cab75399e4672b77efe6a067a7bd148ddf/examples/hotrod/pkg/tracing/init.go#L85

preventing SSL connections from succeeding.  For example:

```
$ OTEL_EXPORTER_OTLP_ENDPOINT=https://hotrod.apm.us-east4.gcp.elastic-cloud.com:443 ./hotrod all -x otlp

2023/02/10 13:00:04 traces export: failed to send to http://hotrod.apm.us-east4.gcp.elastic-cloud.com:443/v1/traces: 400 Bad Request
```

- use secure connections when env var `OTEL_EXPORTER_OTLP_ENDPOINT` uses `https` scheme or `OTEL_EXPORTER_OTLP_INSECURE=false`

### DIFF
--- a/examples/hotrod/pkg/tracing/init.go
+++ b/examples/hotrod/pkg/tracing/init.go
@@ -18,6 +18,8 @@ package tracing
 import (
 	"context"
 	"fmt"
+	"os"
+	"strings"
 	"sync"
 
 	"github.com/opentracing/opentracing-go"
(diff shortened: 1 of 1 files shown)


yurishkuro:
except that now you broke the insecure mode, which is much more prevalent in demo setups
```shell
go run ./examples/hotrod all -x otlp
```

graphaelli:
It doesn't break insecure mode, it changes the default on something user must opt into anyway.  The original behavior is restored with:

`OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4318 go run ./examples/hotrod all -x otlp`

or 

`OTEL_EXPORTER_OTLP_INSECURE=true go run ./examples/hotrod all -x otlp`

The goal is to make the following work, which I suspect is the most likely way folks engage with `hotrod` (that is, via docker images).

```
docker run -it --rm -v $PWD/ca-certificates.crt:/etc/ssl/certs/ca-certificates.crt -e OTEL_EXPORTER_OTLP_ENDPOINT=my-https-endpoint jaegertracing/example-hotrod:latest all -x otlp 
```

That said, if that's too big of an impact I'm fine to withdraw this.

graphaelli:
I suppose we could just look for an `https://` prefix no the env var and not add the option in that case?

yurishkuro (commented):
LGTM. Please make sure all commits are signed (or squash into one, usually easier to fix DCO this way)

yurishkuro (approved):

Issue #3924, opened 2022-09-26
[Feature]: Elasticsearch config details can taken by an env variable through file inside pod filesystem. 


crombus:
### Requirement

Elasticsearch config details can taken by an env variable through file inside pod filesystem. For now we have to pass through configmap.

### Problem

Integrating with vault which injector to inject passwords inside the podfilesystem.

### Proposal

Create env variable for elastic in jaeger which takes password from file. Similarly to keycloak.

### Open questions

Any other way we can do so, Please suggest for the time being.

yurishkuro:
Would it also require watching the file for password changes?

crombus:
yes, for rotating passwords.

haanhvu:
I'll work on this

haanhvu:
@yurishkuro Could you point me to which files I need to change to implement this? For now I can see the ES config is here: https://github.com/jaegertracing/jaeger/blob/main/pkg/es/config/config.go

yurishkuro:
There are several parts to this (best done in different PRs):

1. Refactor file system watchers into one module. Right now two modules in Jaeger support file reloads: `cmd/query/app/static_handler.go` and `pkg/config/tlscfg/cert_watcher.go`. They both rely on 3rd party `fsnotify` module, however the latter is much more involved and can deal with symlinks and whole directories being replaced, which was reported several time as a problem for the original version of `cert_watcher`. So I would like to move this to a shared functionality (into `pkg/fswatcher` module, which currently doesn't do much), and refactor the other two places to use it.
2. Once that shared watcher is available, use it to add a new option to ES storage to read the password from a file (as well as token-file, which is already supported).
3. Lastly, we need to actually be able to support changing password/token on the fly. IIRC all auth settings are done at the start when ES client is initialized, so this requires a bit of research what it would take if those arguments need to be modified - is the code structured such that ES client can be easily swapped?

yurishkuro:
To provide more details about the proposal.

Step 1 is to move the file & directory watching logic from `tlscfg` to `pkg/fswatcher` completely, and expose a much simpler API to the other packages, not the low-level API of the current fswatcher. The API I am thinking of is the following:
```go
type FSWatcher interface {
  WatchFiles(paths []string, onChange func())
  Close()
}
```

The implementation can do this:
  * load file contents and store their checksums
  * register watchers with fsnotifier for both the files and their directoties (as currently done in `tlscfg`)
[...]

PR #4232, opened 2023-02-12
Migrate the use of fsnotify to fswatcher in cert_watcher.go

haanhvu:
Signed-off-by: haanhvu <haanh6594@gmail.com>

First step for #3924 as discussed [here](https://github.com/jaegertracing/jaeger/issues/3924#issuecomment-1424754016)

### DIFF
--- a/pkg/fswatcher/fs_watcher.go
+++ b/pkg/fswatcher/fs_watcher.go
@@ -20,6 +20,7 @@ import "github.com/fsnotify/fsnotify"
 // Primarily used for mocking the fsnotify lib.
 type Watcher interface {
 	Add(name string) error
+	Close() error
 	Events() chan fsnotify.Event
 	Errors() chan error
 }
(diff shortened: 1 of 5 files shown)


haanhvu:
@yurishkuro You said [here](https://github.com/jaegertracing/jaeger/issues/3924#issuecomment-1424754016) that we need to make both `static_handler.go` and `cert_watcher.go` use `fswatcher`. But I already see `static_handler.go` use it. So I just make `cert_watcher.go` use it in this PR.

Please help review!

yurishkuro (approved):
👍

pkg/fswatcher/fs_watcher.go
```diff
@@ -34,6 +35,11 @@ func (f *fsnotifyWatcherWrapper) Add(name string) error {
 	return f.fsnotifyWatcher.Add(name)
 }
 
+// Close closes the watcher.
+func (f *fsnotifyWatcherWrapper) Close() error {
```
yurishkuro, line 39: please add a test for this func

haanhvu: Ok I'll do it.

Is there any way to check Codecov failure locally? I ran `make test` and as I can remember it didn't report it.

yurishkuro: Not to my knowledge. You can only compare package level coverage before and after the change with go test -cover.

yurishkuro (approved):

yurishkuro:
🎉   Thanks!

PR #4235, opened 2023-02-13
Add option to read password from file in ES storage

haanhvu:
Signed-off-by: haanhvu <haanh6594@gmail.com>

Second step for #3924 as discussed [here](https://github.com/jaegertracing/jaeger/issues/3924#issuecomment-1424754016)

### DIFF
--- a/pkg/es/config/config.go
+++ b/pkg/es/config/config.go
@@ -29,6 +29,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/fsnotify/fsnotify"
 	"github.com/olivere/elastic"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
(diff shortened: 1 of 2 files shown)


haanhvu:
@yurishkuro This is the second step as you described [here](https://github.com/jaegertracing/jaeger/issues/3924#issuecomment-1424754016). There may still be errors, but could you check if this is the right implementation approach?

I'll add tests once we have the right approach. After finishing the second step, I'll finish the third step also in this PR. (I think these two steps should go together.)

pkg/es/config/config.go
```diff
@@ -541,7 +556,40 @@ func GetHTTPRoundTripper(c *Configuration, logger *zap.Logger) (http.RoundTrippe
 	return transport, nil
 }
 
-func loadToken(path string) (string, error) {
+func watchPasswordFromFile(path string, passwordChan chan<- string) error {
```
yurishkuro, line 559: @haanhvu so this is exactly the path that I did not go through. Please see the new comment on the ticket: https://github.com/jaegertracing/jaeger/issues/3924#issuecomment-1428567250

haanhvu:
Close in favor of #4342

PR #4239, opened 2023-02-17
Allow SPM to pass bearer token to Prometheus

### LINKED ISSUE #4219: [Feature]: spm bearer token required for prometheus
### Requirement

I am using a global prometheus for spm where I need to present a bearer token to get access.
In kubernetes the token is available at /var/run/secrets/kubernetes.io/serviceaccount/token
Currently no parameter is available to define the token neither as value nor as file.

### Problem

spm gets a 401 when connected to prometheus

### Proposal

create a parameter like prometheus.bearer-token

### Open questions

_No response_

ChillOrb:
Resolves  #4219

- Added functionality for users to pass bearer token from file path or use pass in request context to access SPM


plugin/metrics/prometheus/metricsstore/reader.go
```diff
@@ -264,8 +267,33 @@ func getHTTPRoundTripper(c *config.Configuration, logger *zap.Logger) (rt http.R
...
+	}
+	if token != "" || c.AllowTokenFromContext {
+		transport = bearertoken.RoundTripper{
+			Transport:       httpTransport,
+			OverrideFromCtx: c.AllowTokenFromContext,
+			StaticToken:     token,
+		}
+	}
+
+	return transport, nil
```
albertteoh, line 290: I think we can simplify this to:
```
return bearertoken.RoundTripper{
		Transport:       httpTransport,
		OverrideFromCtx: c.AllowTokenFromContext,
		StaticToken:     token,
	}, nil
```

Also, the above suggestion should fix a bug where it's possible for `transport` to not be set if `token == ""`, which means we lose our `httpTransport` configuration, including TLS configuration (it would be nice if there's a way to test TLS enablement).

ChillOrb:
Closing PR , since more work is required before we can resolve the issue as discussed [here](https://github.com/jaegertracing/jaeger/issues/4219#issuecomment-1435278004)

PR #4251, opened 2023-02-22
#4163 | Speed-up CI by using published UI artifacts 

### LINKED ISSUE #4163: Speed-up CI by using published UI artifacts
### Requirement

As a developer, I want Jaeger CI to run faster when submitting PRs.

### Problem

The query-service and all-in-one binaries & docker images require rebuilding all UI assets, which takes quite a while, and these workflows run on every CI.

### Proposal

Since the jaeger-ui submodule is almost always pinned to a published UI release, we can build the UI artifacts once upon UI release (implemented in https://github.com/jaegertracing/jaeger-ui/commit/15017d5570a508e1ef9368618e2cd8f0abacd16a, available since https://github.com/jaegertracing/jaeger-ui/releases/tag/v1.27.4) and just download them as part of the backend CI. 

**Implementation**

Currently the `rebuild-ui` Makefile target performs a full UI build: https://github.com/jaegertracing/jaeger/blob/65ad4fccf8e451184878cce79bb615edc4c94f17/Makefile#L216-L217

Instead, we can try to do this first:
* `cd jaeger-ui`
* determine of the pinned commit has a published release tag, e.g. `v1.27.4`
* check if the corresponding UI release has the `assets.tar.gz` file uploaded (e.g. https://github.com/jaegertracing/jaeger-ui/releases/download/v1.27.4/assets.tar.gz)
* if yes, download the file and unzip it into `jaeger-ui/packages/jaeger-ui/build/`

If we can't find the assets archive in the release, or the pinned commit is not a release tag, then execute the current logic.

### Open questions

_No response_

shubbham1219:

Resolves #4163

- Attempt to download UI assets from GitHub release, fall back to regular manual build

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -214,8 +214,8 @@ jaeger-ui/packages/jaeger-ui/build/index.html:
 
 .PHONY: rebuild-ui
 rebuild-ui:
-	cd jaeger-ui && yarn install --frozen-lockfile && cd packages/jaeger-ui && yarn build
-
+	./scripts/rebuild-ui.sh
+	
 .PHONY: build-all-in-one-linux
 build-all-in-one-linux:
 	GOOS=linux $(MAKE) build-all-in-one
--- a/Makefile
+++ b/Makefile
@@ -214,7 +214,7 @@ jaeger-ui/packages/jaeger-ui/build/index.html:
 
 .PHONY: rebuild-ui
 rebuild-ui:
-	chmod u+x scripts/rebuild-ui.sh && ./scripts/rebuild-ui.sh
+	bash ./scripts/rebuild-ui.sh
 	
 .PHONY: build-all-in-one-linux
 build-all-in-one-linux:
--- a/Makefile
+++ b/Makefile
@@ -214,7 +214,7 @@ jaeger-ui/packages/jaeger-ui/build/index.html:
 
 .PHONY: rebuild-ui
 rebuild-ui:
-	./scripts/rebuild-ui.sh
+	chmod u+x scripts/rebuild-ui.sh && ./scripts/rebuild-ui.sh
 	
 .PHONY: build-all-in-one-linux
 build-all-in-one-linux:
--- a/scripts/rebuild-ui.sh
+++ b/scripts/rebuild-ui.sh
@@ -9,13 +9,10 @@ if [[ "$tag" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]];  then
     temp_file=$(mktemp)
     trap "rm -f ${temp_file}" EXIT
     release_url="https://github.com/jaegertracing/jaeger-ui/releases/download/${tag}/assets.tar.gz"
-
-     if curl -O "$release_url"; then
-        curl --location --output "$temp_file" "$release_url"
-        mkdir -p packages/jaeger-ui/build/
-        tar -zxvf "$temp_file" packages/jaeger-ui/build/
-        exit 0
-    fi
+    curl --location --output "$temp_file" "$release_url"
+    mkdir -p packages/jaeger-ui/build/
+    tar -zxvf "$temp_file" packages/jaeger-ui/build/
+    exit 0
 fi
 
 # do a regular full build
--- a/scripts/rebuild-ui.sh
+++ b/scripts/rebuild-ui.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+
+set -euxf -o pipefail
+
+cd jaeger-ui
+
+tag=$(git describe --exact-match --tags)
+if [[ "$tag" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]];  then
+    temp_file=$(mktemp)
+    trap "rm -f ${temp_file}" EXIT
+    release_url="https://github.com/jaegertracing/jaeger-ui/releases/download/${tag}/assets.tar.gz"
+
+     if curl -O "$release_url"; then
+        curl --location --output "$temp_file" "$release_url"
+        mkdir -p packages/jaeger-ui/build/
+        tar -zxvf "$temp_file" packages/jaeger-ui/build/
+        exit 0
+    fi
+fi
+
+# do a regular full build
+yarn install --frozen-lockfile && cd packages/jaeger-ui && yarn build
+
+
+
+
+
--- a/scripts/rebuild-ui.sh
+++ b/scripts/rebuild-ui.sh
@@ -9,10 +9,11 @@ if [[ "$tag" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]];  then
     temp_file=$(mktemp)
     trap "rm -f ${temp_file}" EXIT
     release_url="https://github.com/jaegertracing/jaeger-ui/releases/download/${tag}/assets.tar.gz"
-    curl --location --output "$temp_file" "$release_url"
-    mkdir -p packages/jaeger-ui/build/
-    tar -zxvf "$temp_file" packages/jaeger-ui/build/
-    exit 0
+    if curl --silent --fail --location --output "$temp_file" "$release_url"; then
+        mkdir -p packages/jaeger-ui/build/
+        tar -zxvf "$temp_file" packages/jaeger-ui/build/
+        exit 0
+    fi
 fi
 
 # do a regular full build


Makefile
```diff
@@ -214,7 +214,31 @@ jaeger-ui/packages/jaeger-ui/build/index.html:
 
 .PHONY: rebuild-ui
 rebuild-ui:
-	cd jaeger-ui && yarn install --frozen-lockfile && cd packages/jaeger-ui && yarn build
+	cd jaeger-ui
+	
+	if git describe --exact-match --tags $(git rev-parse HEAD) >/dev/null 2>&1; then
```
yurishkuro, line 219: Makefiles don't work with such multi-line scripts, unless you escape them. It may be easier to move all logic into `scripts/rebuild-ui.sh`

Makefile
```diff
@@ -214,7 +214,31 @@ jaeger-ui/packages/jaeger-ui/build/index.html:
...
+		# Get the release version from the tag
+		release_version=$(git describe --exact-match --tags $(git rev-parse HEAD) | cut -d "v" -f 2 | cut -d "-" -f 1)
+
+		# Check if the corresponding UI release has the assets.tar.gz file uploaded
+		release_url="https://github.com/jaegertracing/jaeger-ui/releases/download/v${release_version}/assets.tar.gz"
+
+		if curl --output /dev/null --silent --head --fail "$release_url"; then
+			# Download the file and unzip it into packages/build/
+			
+			curl -L -o assets.tar.gz "$release_url"
```
yurishkuro, line 229: I would prefer we use a temp file 
```
tmpfile=$(mktemp)
trap rm tmpfile EXIT
```

yurishkuro (commented):
Please explain how the change was tested

shubbham1219:
@yurishkuro  I have made some commits on this. I am not exactly sure how to get the version of the head commit. The command $(git describe --exact-match --tags $(git rev-parse HEAD))  gives me the Hash associated with the latest commit but not the version. Can you look into the commit once?

yurishkuro:
This seems to work:
```shell
# when on release tag
$ cd jaeger-ui
$ git describe --exact-match --tags
v1.27.3

# when on a commit without release tag
$ git checkout HEAD^1
$ git describe --exact-match --tags
fatal: no tag exactly matches '9b1ee69fa04482fa6e31e186f0ec790f21f28322'
# above also returns with an error code
```

scripts/rebuild-ui.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+
+cd ../jaeger-ui
+
+if git describe --exact-match --tags $(git rev-parse HEAD) >/dev/null 2>&1; then
```
yurishkuro, line 5: don't think you need `$(git rev-parse HEAD)`, by default it gives you current commit

scripts/rebuild-ui.sh
```diff
@@ -0,0 +1,37 @@
...
+    
+    # Get the release version from the tag  
+    release_version = $(git describe --exact-match --tags $(git rev-parse HEAD))
```
yurishkuro, line 8: no reason to run the same git command twice, read it into a var once

scripts/rebuild-ui.sh
```diff
@@ -0,0 +1,37 @@
...
+cd ../jaeger-ui
+
+if git describe --exact-match --tags $(git rev-parse HEAD) >/dev/null 2>&1; then
+    
+    # Get the release version from the tag  
+    release_version = $(git describe --exact-match --tags $(git rev-parse HEAD)) 
+
+
+    # Check if the corresponding UI release has the assets.tar.gz file uploaded
+    release_url="https://github.com/jaegertracing/jaeger-ui/releases/download/${release_version}/assets.tar.gz"
```
yurishkuro, line 12: I wouldn't recommend using the output from git directly like that. Instead, match it against the release tag patter (vN.M.m) and only then use in the URL

scripts/rebuild-ui.sh
```diff
@@ -0,0 +1,37 @@
...
+
+    temp_file=$(mktemp)
+
+    # Download the file to the temporary file and extract it into packages/build/
+    if curl --output /dev/null --silent --head --fail "$release_url"; then
+        curl --location --output "$temp_file" "$release_url"
+        tar -zxvf "$temp_file" -C packages/jaeger_ui/build/
+        rm -f "$temp_file"
```
yurishkuro, line 20: it's safer to use a trap right after you defined the var

scripts/rebuild-ui.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+
```
yurishkuro, line 2:
```suggestion

set -euxf -o pipefail

```

scripts/rebuild-ui.sh
```diff
@@ -0,0 +1,37 @@
...
+
+    # Check if the corresponding UI release has the assets.tar.gz file uploaded
+    release_url="https://github.com/jaegertracing/jaeger-ui/releases/download/${release_version}/assets.tar.gz"
+
+    temp_file=$(mktemp)
+
+    # Download the file to the temporary file and extract it into packages/build/
+    if curl --output /dev/null --silent --head --fail "$release_url"; then
+        curl --location --output "$temp_file" "$release_url"
```
yurishkuro, line 18: Why do you need to check for `--head` first instead of trying to download right away? I assume it would still return an error exit code if the URL is invalid.

scripts/rebuild-ui.sh
```diff
@@ -0,0 +1,37 @@
...
+        tar -zxvf "$temp_file" -C packages/jaeger_ui/build/
+        rm -f "$temp_file"
+    
+    else
```
yurishkuro, line 22: rather than having two duplicate `else`, just call `exit 0` here. Something like:
```
tag=$(git describe ...)
if tag matches pattern; then
    temp_file=$(mktemp)
    trap "rm -f ${temp_file}" EXIT
    if curl ... ; then
        untar
        exit 0
    fi
fi

# do a regular full build
yarn install --frozen-lockfile && cd packages/jaeger-ui && yarn build
```

shubbham1219:
I have added more commits. 

1)Currently the $(git describe --exact-match --tags) command returns v1.27.3. Then it matches it with release tag patter(vN.M.m) and then tries the download it right away. As the assets file does not exist for this version, it fails with an exit code 0 and stops the execution. Do I need to make a change such that if it fails we perform a full build?

2) I tried to hardcode the tag variable with "v1.27.4" and then running the make run-all-in-one command. It was able to download the assets file and unzip it but later the build failed with the error 
{"level":"panic","ts":1677953320.4700572,"caller":"app/static_handler.go:58","msg":"Could not create static assets handler","error":"cannot load index.html: cannot open index.html: open actual/index.html.gz: file does not exist"

Is this only because the current backend version is not compatible with the assets file that was downloaded or are there more changes required?

yurishkuro:
you can bump your local Jaeger repo to v1.27.4 of the UI:
```shell
$ pwd
.../dev/jaegertracing/jaeger
$ cd jaeger-ui
$ git checkout v1.27.4
```

to restore it back
```shell
$ pwd
.../dev/jaegertracing/jaeger
$ git submodule update
```

yurishkuro:
The error seems to suggest that your build did not place the assets in the right place where they could be picked up by the UI build.

Makefile
```diff
@@ -214,7 +214,9 @@ jaeger-ui/packages/jaeger-ui/build/index.html:
 
 .PHONY: rebuild-ui
 rebuild-ui:
-	cd jaeger-ui && yarn install --frozen-lockfile && cd packages/jaeger-ui && yarn build
+	cd scripts && ./rebuild-ui.sh
```
yurishkuro, line 217:
```suggestion
	./scripts/rebuild-ui.sh
```

shubbham1219: Yeah -C flag was creating packages/jaeger-ui/build/ again. Have fixed this with the latest commit

shubbham1219:
The build succeeds after we bump the UI version to v1.27.4. But fails for the version v1.27.3 as it tries to download right away and does not find the assets file and it exits after that.

Makefile
```diff
@@ -214,8 +214,8 @@ jaeger-ui/packages/jaeger-ui/build/index.html:
 
 .PHONY: rebuild-ui
 rebuild-ui:
-	cd jaeger-ui && yarn install --frozen-lockfile && cd packages/jaeger-ui && yarn build
-
+	chmod u+x scripts/rebuild-ui.sh && ./scripts/rebuild-ui.sh
+
```
yurishkuro, line 218: just call it with bash
```suggestion
	bash ./scripts/rebuild-ui.sh

```

scripts/rebuild-ui.sh
```diff
@@ -0,0 +1,27 @@
...
+
+cd jaeger-ui
+
+tag=$(git describe --exact-match --tags)
+if [[ "$tag" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]];  then
+    temp_file=$(mktemp)
+    trap "rm -f ${temp_file}" EXIT
+    release_url="https://github.com/jaegertracing/jaeger-ui/releases/download/${tag}/assets.tar.gz"
+
+     if curl -O "$release_url"; then
```
yurishkuro, line 13: when I ran this, I got `assets.tar.gz` in the local dir. Why do you need to call curl twice?

shubbham1219: I guess it was part of some earlier commit wherein I was first checking if the file exists instead of downloading right away. I have fixed this.

scripts/rebuild-ui.sh
```diff
@@ -0,0 +1,24 @@
...
+
+tag=$(git describe --exact-match --tags)
+if [[ "$tag" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]];  then
+    temp_file=$(mktemp)
+    trap "rm -f ${temp_file}" EXIT
+    release_url="https://github.com/jaegertracing/jaeger-ui/releases/download/${tag}/assets.tar.gz"
+    curl --location --output "$temp_file" "$release_url"
+    mkdir -p packages/jaeger-ui/build/
+    tar -zxvf "$temp_file" packages/jaeger-ui/build/
+    exit 0
```
yurishkuro, line 15: if the archive is not found, we want to fall back to explicit build
```suggestion
    if curl --silent --fail --location --output "$temp_file" "$release_url"; then
        mkdir -p packages/jaeger-ui/build/
        tar -zxvf "$temp_file" packages/jaeger-ui/build/
        exit 0
    fi
```

yurishkuro (approved):

yurishkuro:
Thank you!

bobrik:
This broke our internal build where we have some patches for jaeger-ui applied. Perhaps we should check for a clean git tree before downloading the jaeger-ui archive from GitHub.

bobrik:
I made #4553 to rectify this.

PR #4260, opened 2023-02-26
Make fswatcher higher-level for cert_watcher and static_handler

haanhvu:
Redo first step #3924 as discussed [here](https://github.com/jaegertracing/jaeger/issues/3924#issuecomment-1424754016) and [here](https://github.com/jaegertracing/jaeger/issues/3924#issuecomment-1428567250)


haanhvu:
@yurishkuro I redo first step as you suggested [here](https://github.com/jaegertracing/jaeger/pull/4242#discussion_r1111278866).

Could you review if the new implementation and test approaches make sense?

Even if they make sense, I have a failed test in `TestWatchFilesChangeAndRemove`: `Unable to locate 'Content changed' in log. All logs: [{{info 2023-02-26 23:38:15.087952009 +0700 +07 m=+0.001966662  Content changed undefined } []} {{warn 2023-02-26 23:38:15.098235482 +0700 +07 m=+0.012250144  ./test.dochas been removed. undefined } []}]`

From the error message (expexted log: `Content changed`, result log: `Content changed undefined`) it seems like I misused the `zapcore` and `zaptest` packages somehow?

pkg/fswatcher/fswatcher.go
```diff
@@ -0,0 +1,129 @@
...
+	fsnotifyWatcher *fsnotify.Watcher
+}
+
+// WatchFiles adds files' directories to the watcher, store each file's hashed content,
+// and watch for their changes and errors. If a file's hashed content changes, invoke onChange().
+func (f *fsWatcherWrapper) WatchFiles(paths []string, onChange func(), log *zap.Logger) error {
+	fileHashContentMap := make(map[string]string)
+	uniqueDirs := make(map[string]bool)
+
+	go func() error {
```
yurishkuro, line 46: You have a race condition here. First execute the initialization logic from L71, and only then start a gorouting for watching.

pkg/fswatcher/fswatcher.go
```diff
@@ -0,0 +1,129 @@
...
+		for {
+			select {
+			case event := <-f.fsnotifyWatcher.Events:
```
yurishkuro, line 49: What happens if WatchFiles() is called multiple times (as would happen from cert_watcher)? Since you're starting multiple goroutines, only one of them might get an event, but the event may be for a different group of files.

There are different ways to fix this, perhaps the simplest is to remove WatchFiles() interface altogether and only have a constructor function, which:
  * takes the file names and an onChange
  * initializes hashes
  * starts a goroutine for watching parent directories
This means the application might end up with more than one active fsnotifier & channels, but I think it will make the code much simpler than if we were trying to reuse a single fsnotifier but then figure out which of the file groups' onChange needs to be called.

pkg/fswatcher/fswatcher.go
```diff
@@ -0,0 +1,129 @@
...
+					event.Op&fsnotify.Write == fsnotify.Write ||
+					event.Op&fsnotify.Rename == fsnotify.Rename {
+					ok, newHash := isModified(event.Name, fileHashContentMap[event.Name])
+					if ok {
+						fileHashContentMap[event.Name] = newHash
+						onChange()
+					}
+				}
+				if event.Op&fsnotify.Remove == fsnotify.Remove {
+					if log == nil {
```
yurishkuro, line 60: * if logger is declared as argument, we can assume it won't be nil.
* our standard name is `logger`, not `log`

pkg/fswatcher/fswatcher.go
```diff
@@ -0,0 +1,129 @@
...
+	uniqueDirs := make(map[string]bool)
+
+	go func() error {
+		for {
+			select {
+			case event := <-f.fsnotifyWatcher.Events:
+				if event.Op&fsnotify.Create == fsnotify.Create ||
+					event.Op&fsnotify.Write == fsnotify.Write ||
+					event.Op&fsnotify.Rename == fsnotify.Rename {
+					ok, newHash := isModified(event.Name, fileHashContentMap[event.Name])
```
yurishkuro, line 53: I am not sure you want to rely on the `event.Name` here, since you're watching parent directories, not the actual files. Just loop through all files and check each of them for change.

pkg/fswatcher/fswatcher.go
```diff
@@ -0,0 +1,129 @@
...
+					if ok {
+						fileHashContentMap[event.Name] = newHash
```
yurishkuro, line 55: you need to check all files in a group and potentially update their new hashes

pkg/fswatcher/fswatcher.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+// fsWatcherWrapper wraps the fsnotify.Watcher and implements FSWatcher.
+type fsWatcherWrapper struct {
+	fsnotifyWatcher *fsnotify.Watcher
+}
+
+// WatchFiles adds files' directories to the watcher, store each file's hashed content,
+// and watch for their changes and errors. If a file's hashed content changes, invoke onChange().
+func (f *fsWatcherWrapper) WatchFiles(paths []string, onChange func(), log *zap.Logger) error {
```
yurishkuro, line 42: please preserve the long comment from `(w *certWatcher) watchChangesLoop`

pkg/fswatcher/fswatcher_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+func assertLogs(t *testing.T, f func() bool, errorMsg string, logObserver *observer.ObservedLogs) {
+	assert.Eventuallyf(t, f,
+		10*time.Second, 10*time.Millisecond,
+		errorMsg,
+		delayedFormat{
+			fn: func() interface{} { return logObserver.All() },
+		},
+	)
+}
```
yurishkuro, line 105: Just an observation: `pkg/config/tlscfg/cert_watcher_test.go` is 489 lines. Are you sure your tests are as comprehensive as they are there?

One potential way to check that is to include a refactoring of `cert_watcher.go` in this PR to use the new helper class, but keep all the unit tests in `cert_watcher_test.go` to see if they still work. Good news is that `cert_watcher_test.go` does not depend on fsnotifier mocks, so the tests should be at a high-enough level of abstraction .

yurishkuro (approved):

haanhvu:
@yurishkuro I've done what you suggested in comments.

After refactoring cert_watcher, there's 1 failure at [test case 4 in `TestAddCertsToWatch_err`](https://github.com/jaegertracing/jaeger/pull/4260/files#diff-d7950062b5351c2ff4ce121d82c6458ee19573ff987f85c7c3b827dddd8a2588L373).

Here's the full error:
```
--- FAIL: TestAddCertsToWatch_err (0.00s)
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x62693d]

goroutine 6 [running]:
testing.tRunner.func1.2({0x661580, 0x884b80})
	/usr/local/go/src/testing/testing.go:1389 +0x24e
testing.tRunner.func1()
	/usr/local/go/src/testing/testing.go:1392 +0x39f
panic({0x661580, 0x884b80})
	/usr/local/go/src/runtime/panic.go:838 +0x207
github.com/jaegertracing/jaeger/pkg/config/tlscfg.(*certWatcher).watchCertPair(0x0)
	/home/admin1/go/src/github.com/jaegertracing/jaeger/pkg/config/tlscfg/cert_watcher.go:98 +0x3d
github.com/jaegertracing/jaeger/pkg/config/tlscfg.TestAddCertsToWatch_err(0x0?)
	/home/admin1/go/src/github.com/jaegertracing/jaeger/pkg/config/tlscfg/cert_watcher_test.go:387 +0x246
testing.tRunner(0xc0001076c0, 0x6d2240)
	/usr/local/go/src/testing/testing.go:1439 +0x102
created by testing.(*T).Run
	/usr/local/go/src/testing/testing.go:1486 +0x35f
```

I have looked at different aspects but couldn't see why there's this error, and only in this test case.

Could you help take a look? Thanks!

pkg/fswatcher/fswatcher.go
```diff
@@ -0,0 +1,129 @@
...
+
+// NewFSWatcher creates a new fsWatcherWrapper, wrapping the fsnotify.Watcher.
+func NewFSWatcher() (FSWatcher, error) {
+	w, err := fsnotify.NewWatcher()
+	return &fsWatcherWrapper{fsnotifyWatcher: w}, err
+}
+
+// isModified returns true if the file has been modified since the last check.
+func isModified(file string, previousHash string) (bool, string) {
+	if file == "" {
```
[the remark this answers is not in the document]
haanhvu: I copied it from the current cert_watcher: https://github.com/jaegertracing/jaeger/blob/82d7f9b6f2da42c2fb483ac8bdd0385a0bec9d2f/pkg/config/tlscfg/cert_watcher.go#L97

pkg/fswatcher/fswatcher.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+// Close closes the watcher.
+func (f *FSWatcher) Close() error {
+	return f.watcher.Close()
+}
+
+// isModified returns true if the file has been modified since the last check.
+func (f *FSWatcher) isModified(filepath string, previousHash string) (bool, string) {
+	if filepath == "" {
```
yurishkuro, line 119: empty paths are already discarded in L71, no need to check for empty

pkg/fswatcher/fswatcher.go
```diff
@@ -0,0 +1,144 @@
...
+		return false, ""
+	}
+	hash, err := hashFile(filepath)
+	if err != nil {
+		f.logger.Warn("File has been removed, using the last known version", zap.String("file", filepath))
+		return false, ""
```
yurishkuro, line 125: I think file not being readable qualifies as modified, why return false?

pkg/fswatcher/fswatcher.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+// isModified returns true if the file has been modified since the last check.
+func (f *FSWatcher) isModified(filepath string, previousHash string) (bool, string) {
+	if filepath == "" {
+		return false, ""
+	}
+	hash, err := hashFile(filepath)
+	if err != nil {
+		f.logger.Warn("File has been removed, using the last known version", zap.String("file", filepath))
```
yurishkuro, line 124: log line doesn't make sense, it should be `"unable to read the file", zap.Error(err), ...`

pkg/config/tlscfg/options.go
```diff
@@ -114,7 +114,9 @@ func (p *Options) Config(logger *zap.Logger) (*tls.Config, error) {
 		}
 	}
 
-	go p.certWatcher.watchChangesLoop(tlsCfg.RootCAs, tlsCfg.ClientCAs)
+	p.certWatcher.watchCertPair()
+	p.certWatcher.watchCert(p.certWatcher.opts.CAPath, tlsCfg.RootCAs)
+	p.certWatcher.watchCert(p.certWatcher.opts.ClientCAPath, tlsCfg.ClientCAs)
```
yurishkuro, line 119: what's the reason for calling these three methods externally, instead of doing it all in certWatcher() constructor?

yurishkuro: There can also be errors in those methods, which are being ignored here. I don't see a good reason for making options.go aware of all these internal details.

pkg/fswatcher/fswatcher.go
haanhvu: Yeah I copied it from the current cert_watcher. But you're right in the new implementation it should return true instead

pkg/config/tlscfg/options.go
haanhvu: > what's the reason for calling these three methods externally, instead of doing it all in certWatcher() constructor?

These three methods (and the current `watchChangesLoop`) need `RootCAs` and `ClientCAs` from `tlsCfg`. For now, `tlsCfg` and its fields are initialized in [`func (p *Options) Config`](https://github.com/jaegertracing/jaeger/pull/4260/files/3ff7ee086c435d30f03dcd9f726e21b74fc37b17#diff-0001cdb928045f195b69e9e8546205bdbc96779bb25510f927ae8d6badd3781bL46).

haanhvu: I'll make tlsCfg a parameter in certWatcher constructor so that we can call these three all there

pkg/config/tlscfg/cert_watcher_test.go
```diff
@@ -373,16 +386,23 @@ func TestAddCertsToWatch_err(t *testing.T) {
 			opts: Options{
 				CAPath:       caCert,
 				ClientCAPath: caCert,
-				CertPath:     serverCert,
 				KeyPath:      "doesnotexists",
```
haanhvu, line 389: I edited this test case because when both CertPath and KeyPath are non-empty, [LoadX509KeyPair is called and returns an error and a nil certWatcher](https://github.com/jaegertracing/jaeger/blob/1d3bc1dc6334ce6b09546d9ddfadc5765820a5e2/pkg/config/tlscfg/cert_watcher.go#L56-L59). That was why I got a [nil pointer dereference when called watchCertPair outside certWatcher constructor](https://github.com/jaegertracing/jaeger/pull/4260#issuecomment-1453968676).

I think there's no point testing LoadX509KeyPair and getting a nil certWatcher here. So delete CertPath in this case to let watchCertPair tested.

haanhvu (commented):
@yurishkuro I resolved all your comments. The tests all passed too. Could you review?

pkg/config/tlscfg/cert_watcher.go
```diff
@@ -195,46 +98,35 @@ func (w *certWatcher) attemptReload(rootCAs, clientCAs *x509.CertPool) {
...
+			zap.Error(err),
+		)
+		w.Close()
+	}
 }
 
-func (w *certWatcher) reloadIfModified(certPath string, certHash *string, certPool *x509.CertPool) {
-	if mod, newHash := w.isModified(certPath, *certHash); mod {
+func (w *certWatcher) watchCert(certPath string, certPool *x509.CertPool) {
+	onCertChange := func() {
```
yurishkuro, line 117: I would move this to a top-level function, to reduce unnecessary indentation

pkg/config/tlscfg/cert_watcher.go
```diff
@@ -195,46 +98,35 @@ func (w *certWatcher) attemptReload(rootCAs, clientCAs *x509.CertPool) {
...
-	}
-	hash, err := hashFile(file)
-	if err != nil {
-		w.logger.Warn("Certificate has been removed, using the last known version", zap.String("certificate", file))
-		return false, ""
+	certWatcher, err := fswatcher.NewFSWatcher([]string{certPath}, onCertChange, w.logger)
+	if err == nil {
+		w.watchers = append(w.watchers, *certWatcher)
+	} else {
+		w.logger.Error("Cannot set up watcher for certificate", zap.String("certificate", certPath), zap.Error(err))
```
yurishkuro, line 129: this code path is executed during construction, it should propagate the error back to the caller

pkg/config/tlscfg/cert_watcher_test.go
yurishkuro: this just tells me that nil pointer error is introduced in this change, not that the test needs changing. I made comment elsewhere that you're not handling / propagating errors properly.

haanhvu:
@yurishkuro I resolved all your comments. The tests all passed too. Could you review? Thanks!

pkg/config/tlscfg/cert_watcher_test.go
```diff
@@ -204,7 +199,7 @@ func TestReload_err_watch(t *testing.T) {
 	opts := Options{
 		CAPath: "doesnotexists",
 	}
-	watcher, err := newCertWatcher(opts, zap.NewNop())
+	watcher, err := newCertWatcher(opts, nil, nil, nil)
```
yurishkuro, line 202: It's safer to pass a logger in case future implementation tries to use it before returning the error

pkg/fswatcher/fswatcher.go
```diff
@@ -0,0 +1,148 @@
...
+
+	go fsw.watch(fileHashContentMap, onChange, logger)
+
+	return fsw, nil
+}
+
+// Watch watches for Events and Errors of files.
+// Each time an Event happen, all the files are checked for content change.
+// If a file's content changes, its hashed content is updated and onChange is invoked.
+func (f *FSWatcher) watch(fileHashContentMap map[string]string, onChange func(), logger *zap.Logger) {
```
yurishkuro, line 95: Nit: the way the code is structured now, it doesn't make sense to pass these arguments to the watch() function, they should be fields on f 

Nit 2: seems like w would be more appropriate receiver name

pkg/fswatcher/fswatcher_test.go
```diff
@@ -0,0 +1,102 @@
...
+	"go.uber.org/zap/zaptest/observer"
+)
+
+func TestFSWatcherAddFiles(t *testing.T) {
+	w, err := NewFSWatcher([]string{"foo"}, nil, nil)
+	assert.Error(t, err)
+	assert.IsType(t, &FSWatcher{}, w)
+
+	w, err = NewFSWatcher([]string{"../../cmd/query/app/fixture/ui-config.json"}, nil, nil)
+	assert.NoError(t, err)
```
yurishkuro, line 36: Test is leaking goroutine, need to close w

pkg/fswatcher/fswatcher_test.go
```diff
@@ -0,0 +1,102 @@
...
+	w, err = NewFSWatcher([]string{"../../cmd/query/app/fixture/static/asset.txt", "../../cmd/query/app/fixture/ui-config.json"}, nil, nil)
+	assert.NoError(t, err)
+	assert.IsType(t, &FSWatcher{}, w)
+
+	err = w.Close()
+	assert.NoError(t, err)
+}
+
+func TestFSWatcherWatchFileChangeAndRemove(t *testing.T) {
+	testFile, err := os.Create("test.doc")
```
yurishkuro, line 52: Is this file creates in the source code dir? It should use temp dir, like cert watcher test does

pkg/fswatcher/fswatcher_test.go
```diff
@@ -0,0 +1,102 @@
...
+	require.NoError(t, err)
+
+	_, err = testFile.WriteString("test content")
+	require.NoError(t, err)
+
+	zcore, logObserver := observer.New(zapcore.InfoLevel)
+	logger := zap.New(zcore)
+
+	onChange := func() {
+		logger.Info("Content changed")
```
yurishkuro, line 62: Nit: better test signal would be to read the file content and include it in the log, that way you can be sure the write log line is matched against

pkg/fswatcher/fswatcher_test.go
```diff
@@ -0,0 +1,102 @@
...
+	require.IsType(t, &FSWatcher{}, w)
+
+	testFile.WriteString("test content changed")
+	assertLogs(t,
+		func() bool {
+			return logObserver.FilterMessage("Content changed").Len() > 0
+		},
+		"Unable to locate 'Content changed' in log. All logs: %v", logObserver)
+
+	os.Remove(testFile.Name())
```
yurishkuro, line 75: I think this test is still too simplistic. The main complications cert watcher had to deal with was movements of parent directories, symlinking, etc. That's why I would like the complexity of cert watcher tests to be reproduced here. It would still be simpler since we don't need to test different certs separately (the benefit on abstracting that into file groups)

yurishkuro (commented):
Almost there, I think the main code is mostly in good shape, some comments around tests.

yurishkuro:
@tsaarni fyi

pkg/fswatcher/fswatcher.go
```diff
@@ -0,0 +1,148 @@
...
+		case event, ok := <-f.watcher.Events:
+			if !ok {
+				return
+			}
+			logger.Info("Received event", zap.String("event", event.String()))
+			for file, hash := range fileHashContentMap {
+				changed, newHash := f.isModified(file, hash)
+				if changed {
+					fileHashContentMap[file] = newHash
+					onChange()
```
tsaarni, line 107: Not saying it is necessarily a big problem, but I think this logic will trigger some extra/spurious reloads. The `onChange()` callback seems to be called for each changed file one-by-one. Consider the callee that handles reload of certificate and private key. Since it has no "big picture" of the change, will it load them twice? 

In the previous implementation, the hash calculation was "closer" to the logic that was knowledgeable about what the files are and cert+key was handled together as one ([link](https://github.com/jaegertracing/jaeger/blob/579b24bab0fa8db11425a49289bfadedb9a4b796/pkg/config/tlscfg/cert_watcher.go#L177-L179)).

haanhvu:
@yurishkuro @tsaarni I resolved all the comments except for adding new tests.

For now, I'm having problem with reading temp file in [TestFSWatcherWatchFileChangeAndRemove](https://github.com/jaegertracing/jaeger/pull/4260/files#diff-6f2c8d7485d1a72e8e2a0a1be20b72ab50173b4353214c5b40ce4aa9ad32732fR74). The test error says [it can't read temp dir or file](https://github.com/jaegertracing/jaeger/actions/runs/4382534002/jobs/7671691280#step:7:3265).

Could you help take a look to see where's the problem?

pkg/fswatcher/fswatcher.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	uniqueDirs := make(map[string]bool)
+
+	for _, p := range filepaths {
+		if p == "" {
+			continue
+		}
+		if h, err := hashFile(p); err == nil {
+			w.fileHashContentMap[p] = h
```
yurishkuro, line 77: We leak fsnotifier here. I suggest moving L70-88 to a function and if it returns an error calling watcher.Close()

pkg/fswatcher/fswatcher_test.go
```diff
@@ -0,0 +1,119 @@
...
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+	"go.uber.org/zap/zaptest/observer"
+)
+
+func TestFSWatcherAddFiles(t *testing.T) {
+	//Add an unreadable file
+	w, err := NewFSWatcher([]string{"foo"}, nil, nil)
+	assert.Error(t, err)
+	assert.IsType(t, &FSWatcher{}, w)
```
yurishkuro, line 34: Shouldn't this be nil instead?

pkg/fswatcher/fswatcher_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+	//Add a readable file
+	w, err = NewFSWatcher([]string{"../../cmd/query/app/fixture/ui-config.json"}, nil, nil)
+	assert.NoError(t, err)
+	assert.IsType(t, &FSWatcher{}, w)
+	err = w.Close()
+	assert.NoError(t, err)
```
yurishkuro, line 41: Can combine 
```suggestion
	assert.NoError(t, w.Close())
```

pkg/config/tlscfg/cert_watcher.go
```diff
@@ -15,41 +15,34 @@
...
-	"os"
-	"path"
 	"path/filepath"
 	"sync"
 
-	"github.com/fsnotify/fsnotify"
 	"go.uber.org/zap"
 
 	"github.com/jaegertracing/jaeger/pkg/fswatcher"
+	"github.com/jaegertracing/jaeger/pkg/multierror"
```
yurishkuro, line 28: main branch was changed to use stdlib's `errors.Join()`

pkg/config/tlscfg/cert_watcher.go
```diff
@@ -60,27 +53,33 @@ func newCertWatcher(opts Options, logger *zap.Logger) (*certWatcher, error) {
...
 
 func (w *certWatcher) Close() error {
-	return w.watcher.Close()
+	var errs []error
+	for _, w := range w.watchers {
+		if err := w.Close(); err != nil {
+			errs = append(errs, err)
+		}
+	}
+	return multierror.Wrap(errs)
```
yurishkuro, line 82:
```suggestion
	return errors.Join(errs...)
```

pkg/fswatcher/fswatcher.go
```diff
@@ -0,0 +1,166 @@
...
+		} else {
+			w.Close()
+			return err
+		}
+		dir := path.Dir(p)
+		if _, ok := uniqueDirs[dir]; !ok {
+			if err := w.watcher.Add(dir); err != nil {
+				w.Close()
```
yurishkuro, line 95:
```suggestion
```
move this to before L71 instead

pkg/fswatcher/fswatcher.go
```diff
@@ -0,0 +1,166 @@
...
+	uniqueDirs := make(map[string]bool)
+
+	for _, p := range filepaths {
+		if p == "" {
+			continue
+		}
+		if h, err := hashFile(p); err == nil {
+			w.fileHashContentMap[p] = h
+		} else {
+			w.Close()
```
yurishkuro, line 89: do this in the caller, that was the reason I suggested to factor it into a function

yurishkuro (approved):
lgtm!

@tsaarni do you have any other concerns?

haanhvu:
> lgtm!
> 
> @tsaarni do you have any other concerns?

@yurishkuro lgty? What about the race detector in the unit test? I'm investigating it today (if you could, then please investigate too). For now it seems like there's a race detector in FSWatcher.watch() when it watches multiple files.

yurishkuro:
right... so the main code is ok because the shared map `fileHashContentMap` is only accessed by one goroutine. However, in the tests you are also reading this map for validation, and it's being detected (correctly) as a race condition. I suggest adding a mutex around accesses to this map (this is how it was previously done in cert_watcher).

haanhvu:
> right... so the main code is ok because the shared map `fileHashContentMap` is only accessed by one goroutine. However, in the tests you are also reading this map for validation, and it's being detected (correctly) as a race condition. I suggest adding a mutex around accesses to this map (this is how it was previously done in cert_watcher).

@yurishkuro Yeah I just found out and thought adding read/write mutex was the solution too. But actually `fileHashContentMap` is not exposed outside of FSWatcher. In reality I'm not sure we'll ever need to read it. So do you think it's worth to add a mutex for it? In the test I just read it to check if FSWatcher updated `fileHashContentMap` successfully. If this check is not necessary we could remove.

yurishkuro:
Usually the good practice for unit tests is to test the external behavior of the component, not internal implementation details. The purpose of the hash map is to suppress no-op changes. This can be tested by writing the same file content into a new temp file and moving that file into the watched file name - this should trigged an event, but the checksum should match and the onChange should not be called. This way you test the expected behavior without depending on internal details. Of course, there should be another test where replacing a file with different content does invoke the callback.

There may be slight technical difficulty ensuring that the callback is *not* called, because of async behavior of the watcher - you won't know how long to wait to be certain that onChange was not called. For this, I would perhaps build another internal callback that's invoked after the loop regardless of whether onChange was called, e.g.
```go
changed := false 
// loop through files
if changed {
  onChange()
}
internalAfterEvents(changed) 
```

`internalAfterEvents` can be a function pointer initialized to no-op in the constructor, and the unit test can override it. This way the onChange call or not call can be validated.

Having said that, adding a mutex might be a smaller change at this point. This is not a high throughput code, so the mutex's perf overhead is not a concern.

tsaarni (dismissed):
Looks good to me! No further comments on my side.

cmd/query/app/static_handler.go
```diff
@@ -63,18 +62,17 @@ func RegisterStaticHandler(r *mux.Router, logger *zap.Logger, qOpts *QueryOption
...
 // StaticAssetsHandler handles static assets
 type StaticAssetsHandler struct {
-	options    StaticAssetsHandlerOptions
-	indexHTML  atomic.Value // stores []byte
-	assetsFS   http.FileSystem
-	newWatcher func() (fswatcher.Watcher, error)
+	options   StaticAssetsHandlerOptions
+	indexHTML atomic.Value // stores []byte
+	assetsFS  http.FileSystem
+	watcher   fswatcher.FSWatcher
```
yurishkuro, line 68:
```suggestion
	watcher   *fswatcher.FSWatcher
```

cmd/query/app/static_handler.go
```diff
@@ -93,23 +91,24 @@ func NewStaticAssetsHandler(staticAssetsRoot string, options StaticAssetsHandler
...
-		assetsFS:   assetsFS,
-		newWatcher: options.NewWatcher,
+		options:  options,
+		assetsFS: assetsFS,
+	}
+
+	watcher, err := fswatcher.NewFSWatcher([]string{options.UIConfigPath}, h.reloadUIConfig, h.options.Logger)
+	if err != nil {
+		return nil, err
+	} else {
```
yurishkuro, line 107: else clause is redundant, can assign outside of if{}

pkg/config/tlscfg/cert_watcher.go
```diff
@@ -36,20 +33,16 @@ import (
...
-	watcher      fswatcher.Watcher
-	cert         *tls.Certificate
-	caHash       string
-	clientCAHash string
-	certHash     string
-	keyHash      string
+	mu       sync.RWMutex
+	opts     Options
+	logger   *zap.Logger
+	watchers []fswatcher.FSWatcher
```
yurishkuro, line 39:
```suggestion
	watchers []*fswatcher.FSWatcher
```

pkg/fswatcher/fswatcher.go
```diff
@@ -0,0 +1,169 @@
...
+			if !ok {
+				return
+			}
+			w.logger.Info("Received event", zap.String("event", event.String()))
+			var changed bool
+			for file, hash := range w.fileHashContentMap {
+				fileChanged, newHash := w.isModified(file, hash)
+				if fileChanged {
+					changed = fileChanged
+					w.mu.Lock()
```
yurishkuro, line 123: you are accessing map in L119, the lock needs to be taken before that.

cmd/query/app/static_handler.go
```diff
@@ -93,23 +91,23 @@ func NewStaticAssetsHandler(staticAssetsRoot string, options StaticAssetsHandler
...
-		options:    options,
-		assetsFS:   assetsFS,
-		newWatcher: options.NewWatcher,
+		options:  options,
+		assetsFS: assetsFS,
 	}
 
+	watcher, err := fswatcher.NewFSWatcher([]string{options.UIConfigPath}, h.reloadUIConfig, h.options.Logger)
+	if err != nil {
+		return nil, err
```
yurishkuro, line 106: we can add a test for this by passing invalid file name

pkg/config/tlscfg/cert_watcher.go
```diff
@@ -60,27 +53,33 @@ func newCertWatcher(opts Options, logger *zap.Logger) (*certWatcher, error) {
...
-		watcher: watcher,
+		opts:   opts,
+		logger: logger,
+		cert:   cert,
 	}
 
-	if err := w.setupWatchedPaths(); err != nil {
-		watcher.Close()
+	if err := w.watchCertPair(); err != nil {
+		return nil, err
```
yurishkuro, line 63: can test this with invalid file names

pkg/config/tlscfg/cert_watcher.go
```diff
@@ -60,27 +53,33 @@ func newCertWatcher(opts Options, logger *zap.Logger) (*certWatcher, error) {
...
+	}
+	if err := w.watchCert(w.opts.CAPath, rootCAs); err != nil {
 		return nil, err
 	}
+	if err := w.watchCert(w.opts.ClientCAPath, clientCAs); err != nil {
+		return nil, err
```
yurishkuro, line 69: can test this with invalid file names

pkg/config/tlscfg/cert_watcher_test.go
```diff
@@ -340,83 +326,6 @@ func createTimestampDir(t *testing.T, dir string, ca, cert, key string) {
 	require.NoError(t, err)
 }
 
-func TestAddCertsToWatch_err(t *testing.T) {
```
yurishkuro, line 343: why delete this test? I think it could've helped the coverage I commented on earlier.

pkg/fswatcher/fswatcher.go
```diff
@@ -0,0 +1,169 @@
...
+
+	return w, nil
+}
+
+func (w *FSWatcher) setupWatchedPaths(filepaths []string) error {
+	uniqueDirs := make(map[string]bool)
+
+	for _, p := range filepaths {
+		if p == "" {
+			continue
```
yurishkuro, line 87: this one can be easily tested

yurishkuro (commented):
Code lgtm, I would like to bring the test coverage up a bit, many places look like low hanging fruit to test.

pkg/config/tlscfg/cert_watcher_test.go
haanhvu: I thought invalid file names were already tested in `fswatcher`, so there's no need to test it again in `static_handler` and `cert_watcher`?

yurishkuro: there are different levels of error handling, the test here would exercise error code paths that only exist in cert_watcher

cmd/query/app/static_handler.go
haanhvu: If UI config file name is invalid, then the error has already been returned in L96, through loadAndEnrichIndexHTML?

pkg/config/tlscfg/cert_watcher.go
```diff
@@ -60,27 +53,33 @@ func newCertWatcher(opts Options, logger *zap.Logger) (*certWatcher, error) {
...
 	return w, nil
 }
 
 func (w *certWatcher) Close() error {
-	return w.watcher.Close()
+	var errs []error
+	for _, w := range w.watchers {
+		if err := w.Close(); err != nil {
+			errs = append(errs, err)
+		}
```
yurishkuro, line 80: you can actually fool the code coverage here by always appending error, because Join ignores nils
```suggestion
		errs = append(errs, w.Close())
```

yurishkuro (approved):

pkg/config/tlscfg/options.go
```diff
@@ -114,7 +114,6 @@ func (p *Options) Config(logger *zap.Logger) (*tls.Config, error) {
 		}
 	}
 
-	go p.certWatcher.watchChangesLoop(tlsCfg.RootCAs, tlsCfg.ClientCAs)
 	return tlsCfg, nil
 }
```
yurishkuro, line 119: GH doesn't allow to comment on L142, but it's causing a race condition in the CI.
I think we could use the same mutex in certWatcher to serialize access to the cert pools.
```
WARNING: DATA RACE
Read at 0x00c0000105e8 by goroutine 22:
  crypto/x509.(*CertPool).addCertFunc()
      /opt/hostedtoolcache/go/1.20.2/x64/src/crypto/x509/cert_pool.go:189 +0x6b8
  crypto/x509.(*CertPool).AppendCertsFromPEM()
      /opt/hostedtoolcache/go/1.20.2/x64/src/crypto/x509/cert_pool.go:227 +0x4f1
  github.com/jaegertracing/jaeger/pkg/config/tlscfg.addCertToPool()
      /home/runner/work/jaeger/jaeger/pkg/config/tlscfg/options.go:142 +0x1d7
 ```

yurishkuro: I added a mutex, but I think there is a bigger issue with race conditions, which is outside the scope of this PR. Namely, the race condition showed that certPools themselves are not safe for concurrent access. The mutex I added prevents competing writes, but it's not clear to me what that means to the actual RPC layers that utilize this TLS configuration. It's possible that they only read the certPool at the beginning, or perhaps at rare re-connection events. But if they do read pools from time to time, this is a clear race condition with use trying to update those pools. I don't really see how this can be solved in Go standard tls package.

yurishkuro (approved):

pkg/config/tlscfg/options.go
yurishkuro: @tsaarni I wonder if you ran into this race condition

tsaarni: @yurishkuro I have not bumped into it, but we have refrained from rotating CA certificates at runtime. 

I agree that it does not seem to be possible to solve the race. 

When it comes to the impact, CA certificate change on disk while processing TLS handshake could cause connection failure, maybe even crash? `tls.Config.RootCAs` gets used at every reconnect, during processing of TLS handshake messages ([here](https://github.com/golang/go/blob/20e9b7f1b53d49fd66e0344b1d0d42d3cf5e47b6/src/crypto/tls/handshake_client.go#L862)). Similarly for servers, `tls.Config.ClientCAs` gets used when processing the TLS handshake ([here](https://github.com/golang/go/blob/20e9b7f1b53d49fd66e0344b1d0d42d3cf5e47b6/src/crypto/tls/handshake_server.go#L812)). The pointer to `CertPool` gets used when verifying the peer certificate during call `certs[0].Verify(opts)` (defined [here](https://github.com/golang/go/blob/20e9b7f1b53d49fd66e0344b1d0d42d3cf5e47b6/src/crypto/x509/verify.go#L751)).

Another observation: `certPool.AppendCertsFromPEM()` only appends new CA certificates. The old ones are never removed even if they would not be present on disk anymore. There isn't support in Go to remove certificates from an existing `CertPool`. 

In practise, it is quite common that only client and server credentials are rotated at runtime but modifying trust anchors require restart. It seems to me that the design of go crypto is (mostly) aligned with this.

BTW for TLS servers, there is callback `tls.Config.GetConfigForClient` that in theory could be used to return `tls.Config` with a different (atomic) copy of `tls.Config.ClientCAs`. There is no similar function for TLS clients.

yurishkuro: I booked #4316

Issue #4268, opened 2023-03-01
[Bug]: es.index-prefix not consistent in es jobs


OpenGuidou:
### What happened?

To be able to configure the index-prefix with ES, the flag is es.index-prefix for the collector and query components.
For the jobs (es-index-cleaner, es-rollover and es-mapping-generator), it's using directly index-prefix without the prefix.

If we're using a common configuration for all components, then the jobs are not configured correctly with the indexprefix.
Example with the jaeger helm chart:

https://github.com/jaegertracing/helm-charts/blob/main/charts/jaeger/templates/_helpers.tpl#L371

### Steps to reproduce

Deploy the jaeger helm chart with the elasticsearch storage enabled.
The jobs won't be able to find the indexes.

### Expected behavior

It should be aligned and use the es. prefix as well

### Jaeger backend version

Latest

### Stogage backend

Elasticsearch

### Operating system

any

### Deployment model

Any

yurishkuro:
If the helm chart is using incorrect flags, we should fix the chart.

OpenGuidou:
Then why use the prefix for the username and password ?
https://github.com/jaegertracing/jaeger/blob/main/cmd/es-index-cleaner/app/flags.go#L29

Was it not to align the flags across all usages ?

yurishkuro:
It's fine to align all usages, but the way we do it is by supporting both new and old flags for some time, see https://github.com/jaegertracing/jaeger#version-compatibility-guarantees

OpenGuidou:
Fair enough, thanks
I'll update de PR based on that

rbaret:
I ran into this bug this week. I had to go through the es-index-cleaner code to understand why my indices where not cleaned as per requested (No indices to delete was the only indication in the logs of the container although it was listing all indices). No indication of any sort regarding the indexPrefix being incorrect...

I had to understand how env vars were used in the es-index-cleaner to notice it was not working with the ES_INDEX_PREFIX envvar name passed by default but was expecting INDEX_PREFIX instead...

If you don't want to solve this, at least update the documentation to mention that INDEX_PREFIX in esIndexCleaner.extraEnv should be set when using storage.elasticsearch.indexPrefix ...

yurishkuro:
we will address this in jaeger-v2, where we're redesigning configuration completely.

PR #4269, opened 2023-03-01
Align ES index-prefix parameter for jobs

OpenGuidou:
Proposal to fix https://github.com/jaegertracing/jaeger/issues/4268
As it's a non-backward compatible change, let me know how you want to handle this.
I'm open to changes.

- Align ES index-prefix flag for jobs to what's defined for other jaeger components. This allows to have a single configuration for all items.

### DIFF
--- a/cmd/es-index-cleaner/app/flags.go
+++ b/cmd/es-index-cleaner/app/flags.go
@@ -18,16 +18,19 @@ import (
 	"flag"
 
 	"github.com/spf13/viper"
+	"go.uber.org/zap"
 )
 
 const (
-	indexPrefix        = "index-prefix"
-	archive            = "archive"
-	rollover           = "rollover"
-	timeout            = "timeout"
-	indexDateSeparator = "index-date-separator"
-	username           = "es.username"
-	password           = "es.password"
+	deprecatedIndexPrefix        = "index-prefix"
+	deprecatedIndexPrefixWarning = "(deprecated, will be removed after 2023-06-05 or in release v1.44.0, whichever is later)"
+	archive                      = "archive"
+	rollover                     = "rollover"
+	timeout                      = "timeout"
+	indexDateSeparator           = "index-date-separator"
+	username                     = "es.username"
+	password                     = "es.password"
+	indexPrefix                  = "es.index-prefix"
 )
 
 // Config holds configuration for index cleaner binary.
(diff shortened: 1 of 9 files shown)


cmd/es-index-cleaner/app/flags.go
```diff
@@ -44,18 +46,23 @@ type Config struct {
...
 	flags.String(password, "", "The password required by storage")
+	flags.String(indexPrefix, "", "Index prefix")
 }
 
 // InitFromViper initializes config from viper.Viper.
 func (c *Config) InitFromViper(v *viper.Viper) {
 	c.IndexPrefix = v.GetString(indexPrefix)
+	if c.IndexPrefix == "" {
+		// try with deprecated flag
+		c.IndexPrefix = v.GetString(deprecatedIndexPrefix)
```
yurishkuro, line 64: this should log a warning about using deprecated flag

yurishkuro (commented):
I really don't think this PR is useful. It clearly shows that the naming of the flags is all over the place and inconsistent. Fixing one single flag doesn't make the situation any better, it only introduces churn and can break people in 3m who don't notice the deprecation. I don't see any benefits in breaking people for a cosmetic change.

OpenGuidou:
> 

We can take the opportunity to review the cli flags in those commands, there are not so many.
In my opinion it's still better to fix it at the root and risk breaking people not noticing the warning than pushing the inconsistency downwards to the rest of the ecosystem integration, like with the helm chart.

cmd/es-index-cleaner/app/flags.go
OpenGuidou: solved

yurishkuro:
> In my opinion it's still better to fix it at the root and risk breaking people not noticing the warning than pushing the inconsistency downwards to the rest of the ecosystem integration, like with the helm chart.

I disagree. One of the worst sins for a project to do is break users. Your better have strong reasons if you do that, and here I don't see a good reason. It would be nice if we, the maintainers, were more diligent when this inconsistency was introduced the first time, but right now is has already been pushed downwards and adopted, few years ago.

The _complexity_ is also arguable. All these scripts are narrowly scoped to ES maintenance, I don't think it's especially difficult to grasp 10 or so flags, whether they are named consistently or not.

yurishkuro:
will be fixed differently in jaeger-v2

Issue #4172, opened 2023-01-21
Change "docket images" CI/CD task to build for fewer architectures


yurishkuro:
### Requirement

As a developer I like CI/CD runs to take as little time as possible.

### Problem

"Build docker images" workflow takes 30min now, because it builds binaries and images for all possible architectues.

### Proposal

Consider reducing the number of architectures built for PRs, and only keep the full set for the main branch / release workflow.

Personally I would limit the PR builds to linux/amd86 architecture.

shubbham1215:
Hi @yurishkuro, Could you give a brief about what changes would be a part of this ticket? I am looking to contribute to an issue for my academic requirement. I have already picked https://github.com/jaegertracing/jaeger/issues/4163 and will be working on it this week but the instructor has also mentioned picking something that would require a good amount of time to finish. Will this be a good issue? Or if there is some other issue you can assign me that as well.

yurishkuro:
@shubbham1215 the change would be mostly refactoring shell scripts / Makefile to pass fewer arch params to the docker build process. Around a week of work, including ramp-up.

> good amount of time to finish

what amount of time do you want to allocate?

shubbham1215:
Okay. I think this issue is fine as it would require me much more than a week to finish it. Could you assign me this as well? Thanks.

Eileen-Yu:
Hi @yurishkuro Here is an initialized PR for this issue:
https://github.com/jaegertracing/jaeger/pull/4286

Since I'm pretty new here, would you help with review and let me know if anything you'd like to share?

PR #4286, opened 2023-03-09
Refactor: build linux artifacts only for PR

Eileen-Yu:
- Resolves https://github.com/jaegertracing/jaeger/issues/4172

- scripts/build-upload-docker-images.sh: Set platforms variable dynamically based on ~~env `PLATFORMS`~~ the first argument
- .github/workflows/ci-docker-build.yml: Add a new job for PR only and pass `pr-only` as the first argument

### DIFF
--- a/.github/workflows/ci-docker-build.yml
+++ b/.github/workflows/ci-docker-build.yml
@@ -44,7 +44,12 @@ jobs:
 
     - uses: docker/setup-qemu-action@e81a89b1732b9c48d79cd809d8d81d79c4647a18
 
+    - name: Build only linux/amd64 docker image for Pull Request
+      if: github.ref_name != 'main'
+      run: bash scripts/build-upload-docker-images.sh pr-only
+
     - name: Build and upload all docker images
+      if: github.ref_name == 'main'
       run: bash scripts/build-upload-docker-images.sh
       env:
         DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
--- a/scripts/build-upload-docker-images.sh
+++ b/scripts/build-upload-docker-images.sh
@@ -2,16 +2,22 @@
 
 set -euxf -o pipefail
 
+mode=${1-main}
+
 make create-baseimg-debugimg
 
-# build multi-arch binaries
 make build-binaries-linux
-make build-binaries-s390x
-make build-binaries-ppc64le
-make build-binaries-arm64
 
-# build multi-arch docker images
-platforms="linux/amd64,linux/s390x,linux/ppc64le,linux/arm64"
+if [ "$mode" = "pr-only" ]; then
+  # build artifacts for linux/amd64 only for pull requests
+  platforms="linux/amd64"
+else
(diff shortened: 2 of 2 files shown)


.github/workflows/ci-docker-build.yml
```diff
@@ -44,7 +44,16 @@ jobs:
 
     - uses: docker/setup-qemu-action@e81a89b1732b9c48d79cd809d8d81d79c4647a18
 
+    - name: Build and upload only linux/amd64 docker image for Pull Request
+      if: github.ref_name != 'main'
+      run: bash scripts/build-upload-docker-images.sh
+      env:
+        PLATFORMS: "linux/amd64"
+        DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
+        QUAY_TOKEN: ${{ secrets.QUAY_TOKEN }}
```
yurishkuro, line 53: Since we do not upload images for PRs, I think it's not necessary to pass tokens

.github/workflows/ci-docker-build.yml
```diff
@@ -44,7 +44,16 @@ jobs:
 
     - uses: docker/setup-qemu-action@e81a89b1732b9c48d79cd809d8d81d79c4647a18
 
+    - name: Build and upload only linux/amd64 docker image for Pull Request
```
yurishkuro, line 47:
```suggestion
    - name: Build only linux/amd64 docker image for Pull Request
```

scripts/build-upload-docker-images.sh
```diff
@@ -2,16 +2,19 @@
 
 set -euxf -o pipefail
 
+# build multi-arch docker images by default
+platforms="${PLATFORMS:-linux/amd64,linux/s390x,linux/ppc64le,linux/arm64}"
```
yurishkuro, line 6: You're not really using the value of env var effectively, so instead of trying to pass full platform name and then doing ugly check in L13 you can pass an argument to the script to indicate pr-only mode and used that to decide which commands to execute and which platforms to build.

scripts/build-upload-docker-images.sh
```diff
@@ -2,16 +2,22 @@
 
 set -euxf -o pipefail
 
+mode=${1-main}
```
Eileen-Yu, line 5: I passed the `pr-only` arg to the script when pull-request, and set the default arg as `main` (to avoid unbound variable error).

Not sure if it is good, or there is better way to deal with this?

scripts/build-upload-docker-images.sh
Eileen-Yu: Thx for reviewing! @yurishkuro 
Your suggestion sounds  great.
Not sure if understood, but I tried passing the `pr-only` arg to the script.
Would that be good?

.github/workflows/ci-docker-build.yml
Eileen-Yu: Sounds good👍

yurishkuro:
Comparing runtime, this change is 24m -> 12m!

yurishkuro (approved):
🎉

yurishkuro:
Thank you @Eileen-Yu

PR #4295, opened 2023-03-12
Allow SPM to pass bearer token to Prometheus #4247

### LINKED ISSUE #4219: [Feature]: spm bearer token required for prometheus

ChillOrb:
- Resolves #4219  

- Follow up to #4247


ChillOrb:
@albertteoh I tried to revert submodule [change in /jaegar-ui ](https://github.com/jaegertracing/jaeger/pull/4247#pullrequestreview-1332340632) but it does not seem to me working. 
I tried to create[ new branch ](https://github.com/ChillOrb/jaeger/tree/dev2) and create a PR after adding changes but as soon as i make `git pull` i see submodule changes in `git status` . After reading [about submodules](https://git-scm.com/book/en/v2/Git-Tools-Submodules) 

`By default, the git pull command recursively fetches submodules changes, as we can see in the output of the first command above. However, it does not update the submodules. This is shown by the output of the git status command, which shows the submodule is “modified”, and has “new commits”. What’s more, the brackets showing the new commits point left (<), indicating that these commits are recorded in MainProject but are not present in the local DbConnector checkout. To finalize the update, you need to run git submodule update:`

How should i proceed with this? 

Maybe i can pull in all submodules update using `git submodule update --remote` ?

ChillOrb:
> > as i make `git pull`
> 
> @ChillOrb, just for my understanding, why is there a need to do a `git pull`?
> 
> Technically, you should be able to create a new PR from your [previous PR](https://github.com/jaegertracing/jaeger/pull/4247) with:
> 
> ```
> # From your jaeger fork root directory on "main" branch.
> $ git checkout -b dev
> $ wget https://github.com/jaegertracing/jaeger/pull/4247.diff -O /tmp/4247.diff
> $ git apply /tmp/4247.diff
> $ git add .
> $ git commit -sm "My commit message"
> $ git push origin dev
> # Then create PR from dev branch.
> ```
> 
> Then, after making changes for an existing PR:
> 
> ```
> $ git add .
> $ git commit -sm "My change"
> $ git push origin dev
> ```
> 
>  > How should i proceed with this?
> 
> I would try running these commands from your project root then push the changes (more generally, I would run this if my submodules go "out of sync"):
> 
> ```
> $ git submodule deinit -f .
> $ git submodule update --init
> ```

@albertteoh , issue was with my main branch and therefore propagated to dev . I forced reset my main from head of jaegar:main , synced dev with main and selectively added files from dev2 to my dev. Sorry for all the confusion. Took me sometime to get this sorted.

Really appreciate you helping me out on this. 

Please review my PR again

plugin/metrics/prometheus/metricsstore/reader_test.go
```diff
@@ -357,6 +358,49 @@ func TestGetRoundTripper(t *testing.T) {
...
 
+func TestGetRoundTripperToken(t *testing.T) {
+	tokenFilePath := "test/test.txt"
+	wantBearer := "token from file"
+	logger := zap.NewNop()
+	// Create temp file with token if tokenFilePath is provided
+	dir, file := filepath.Split(tokenFilePath)
+	err := os.MkdirAll(dir, 0o750)
+	require.NoError(t, err)
+	err = os.WriteFile(filepath.Join(dir, file), []byte(wantBearer), 0o660)
```
albertteoh, line 369: Instead of `filepath.Split(tokenFilePath)`, could we use `filepath.Dir(tokenFilePath)` to just get the `dir` so that we can recursively create the directory structure for the test file, then replace `filepath.Join(dir, file)` with just `tokenFilePath` (`file` isn't really needed)?

plugin/metrics/prometheus/metricsstore/reader.go
```diff
@@ -264,8 +265,26 @@ func getHTTPRoundTripper(c *config.Configuration, logger *zap.Logger) (rt http.R
 		TLSHandshakeTimeout: 10 * time.Second,
 		TLSClientConfig:     ctlsConfig,
 	}
+
+	token := ""
+	if c.TokenFilePath != "" {
+		tokenFromFile, err := loadToken(c.TokenFilePath)
+		if err != nil {
+			return nil, err
```
albertteoh, line 273: We'll need to add a couple of tests to cover the error cases here and on L287.

ChillOrb: @albertteoh To cover these i was thinking of adding back table test case [here](https://github.com/ChillOrb/jaeger/blob/2c56f164f8a40f4a9d8d901753131454bc5abe73/plugin/metrics/prometheus/metricsstore/reader_test.go#L361) and [here](https://github.com/ChillOrb/jaeger/blob/2c56f164f8a40f4a9d8d901753131454bc5abe73/plugin/metrics/prometheus/metricsstore/reader_test.go#L505) 

I was thinking of using `wantErr bool` field and to mock an error maybe i can mock the behavior where user provides incorrect file path ( path does not exist error). This would however add some clumsy `if else` blocks where i check `if wantErr{ require.Err()}else{require.NoErr()}`.

Does this sound ok?

albertteoh: I think a better approach is to add a seperate simple test to cover the error case. This should cover both lines:
```
func TestGetRoundTripperTokenError(t *testing.T) {
	tokenFilePath := "this file does not exist"

	_, err := getHTTPRoundTripper(&config.Configuration{
		TokenFilePath: tokenFilePath,
	}, nil)

	require.Error(t, err)

	pathErr := &os.PathError{}
	assert.ErrorAs(t, err, &pathErr)
	assert.EqualError(t, pathErr, fmt.Sprintf("open %s: no such file or directory", tokenFilePath))
}
```

plugin/metrics/prometheus/metricsstore/reader_test.go
```diff
@@ -357,6 +358,49 @@ func TestGetRoundTripper(t *testing.T) {
 	}
 }
 
+func TestGetRoundTripperToken(t *testing.T) {
+	tokenFilePath := "test/test.txt"
+	wantBearer := "token from file"
+	logger := zap.NewNop()
```
albertteoh, line 364: nit: I just realised that `logger` is only used if TLS is enabled, which is not the case for this test. So we can remove this and just pass `nil` into `getHTTPRoundTripper`, to simplify the test a little.

ChillOrb: Done

plugin/metrics/prometheus/metricsstore/reader.go
ChillOrb: @albertteoh thanks, i was overthinking this

plugin/metrics/prometheus/metricsstore/reader.go
```diff
@@ -264,8 +265,26 @@ func getHTTPRoundTripper(c *config.Configuration, logger *zap.Logger) (rt http.R
...
+		}
+		token = tokenFromFile
+	}
 	return bearertoken.RoundTripper{
 		Transport:       httpTransport,
 		OverrideFromCtx: true,
```
albertteoh, line 279: For the benefit of anyone else looking over this PR, as mentioned earlier, I'll create a separate PR to address removing this line, and improve the `TestGetRoundTripperTLSConfig` test.

albertteoh (dismissed):
This looks good to me, thanks, @ChillOrb! I'll give other maintainers a chance to look over this PR before merging.

plugin/metrics/prometheus/metricsstore/reader_test.go
```diff
@@ -357,6 +358,62 @@ func TestGetRoundTripper(t *testing.T) {
...
+	require.NoError(t, err)
+	err = os.WriteFile(filepath.Join(dir, file), []byte(wantBearer), 0o660)
+	require.NoError(t, err)
+	defer func() {
+		err = os.RemoveAll(tokenFilePath)
+		require.NoError(t, err)
+	}()
+
+	rt, err := getHTTPRoundTripper(&config.Configuration{
+		ServerURL:      "https://localhost:1234",
```
yurishkuro, line 376: why not call this *after* the test server is created and its ServerURL is known?

plugin/metrics/prometheus/metricsstore/reader_test.go
```diff
@@ -357,6 +358,62 @@ func TestGetRoundTripper(t *testing.T) {
...
+
+	_, err := getHTTPRoundTripper(&config.Configuration{
+		TokenFilePath: tokenFilePath,
+	}, nil)
+
+	require.Error(t, err)
+
+	pathErr := &os.PathError{}
+	assert.ErrorAs(t, err, &pathErr)
+	assert.EqualError(t, pathErr, fmt.Sprintf("open %s: no such file or directory", tokenFilePath))
```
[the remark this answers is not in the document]
ChillOrb: I agree , how does _[this](https://github.com/ChillOrb/jaeger/blob/714f86397dfdea11361e5cc066a0e1577481a8bf/plugin/metrics/prometheus/metricsstore/reader.go#L288)_ look?

plugin/metrics/prometheus/metricsstore/reader_test.go
```diff
@@ -457,3 +514,21 @@ func assertMetrics(t *testing.T, gotMetrics *metrics.MetricFamily, wantLabels ma
 	actualVal := mps[0].Value.(*metrics.MetricPoint_GaugeValue).GaugeValue.Value.(*metrics.GaugeValue_DoubleValue).DoubleValue
 	assert.Equal(t, float64(9223372036854), actualVal)
 }
+
+func TestLoadToken(t *testing.T) {
```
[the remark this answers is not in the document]
ChillOrb: I agree , removed .. thanks

plugin/metrics/prometheus/metricsstore/reader_test.go
ChillOrb: Yep , since url is not required i have simplified and removed it

plugin/metrics/prometheus/metricsstore/reader_test.go
```diff
@@ -357,6 +358,62 @@ func TestGetRoundTripper(t *testing.T) {
 	}
 }
 
+func TestGetRoundTripperToken(t *testing.T) {
+	tokenFilePath := "test/test.txt"
+	wantBearer := "token from file"
+	// Create temp file with token if tokenFilePath is provided
+	dir, file := filepath.Split(tokenFilePath)
+	err := os.MkdirAll(dir, 0o750)
```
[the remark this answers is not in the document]
ChillOrb: I feel having a temp file is better than creating a directory for this test case . _[This](https://github.com/ChillOrb/jaeger/blob/d858ca2f51a3bdaaff76a61feab186a397f60048/plugin/metrics/prometheus/metricsstore/reader_test.go#L364)_ the change i have made to the function... Please share your views, thanks

plugin/metrics/prometheus/metricsstore/reader.go
```diff
@@ -264,8 +265,26 @@ func getHTTPRoundTripper(c *config.Configuration, logger *zap.Logger) (rt http.R
 		TLSHandshakeTimeout: 10 * time.Second,
 		TLSClientConfig:     ctlsConfig,
 	}
+
+	token := ""
+	if c.TokenFilePath != "" {
+		tokenFromFile, err := loadToken(c.TokenFilePath)
```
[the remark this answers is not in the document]
ChillOrb: I was not aware of this case , thanks for pointing this out. As per my understanding if Jaeger is deployed on K8 cluster then token rotation is configurable(@hourly by default) and thus ought to be watched. Is that correct? 

Also, what if Jaeger is hosted on some other orchestration platform? 

Need some help deciding on next steps here..

@yurishkuro @albertteoh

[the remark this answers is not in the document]
yurishkuro: I would recommend just creating an issue for it to see if there's actually a demand for auto-reload of Prom token. Implementation-wise it should be using the fswatcher introduced in #4260

plugin/metrics/prometheus/metricsstore/reader.go
```diff
@@ -264,8 +265,28 @@ func getHTTPRoundTripper(c *config.Configuration, logger *zap.Logger) (rt http.R
...
+		StaticToken:     token,
 	}, nil
 }
+
+func loadToken(path string) (string, error) {
+	b, err := os.ReadFile(filepath.Clean(path))
+	if err != nil {
+		// wrapping the error message
+		err = fmt.Errorf("failed to get token from file: %w", err)
+		return "", err
```
yurishkuro, line 289: unnecessary verbosity
```suggestion
		return "", fmt.Errorf("failed to get token from file: %w", err)
```

yurishkuro (approved):

plugin/metrics/prometheus/metricsstore/reader_test.go
```diff
@@ -357,6 +356,53 @@ func TestGetRoundTripper(t *testing.T) {
...
+	defer func() { assert.NoError(t, os.Remove(file.Name())) }()
+
+	_, err = file.Write([]byte(wantBearer))
+	require.NoError(t, err)
+	require.NoError(t, file.Close())
+
+	rt, err := getHTTPRoundTripper(&config.Configuration{
+		ConnectTimeout: time.Second,
+		TokenFilePath:  file.Name(),
+	}, nil)
```
yurishkuro, line 373: @ChillOrb I made a few tweaks
* simplify temp file creation
* make sure to call file.Close() after writing content to flush the write
* change ConnectTimeout to 1sec

yurishkuro: and use const for wantBearer

yurishkuro:
Thanks!

PR #4298, opened 2023-03-13
Prepare release v1.43.0

pavolloffay:
TODOs

- [x] Update UI once https://github.com/jaegertracing/jaeger-ui/pull/1264 is merged

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,7 +36,7 @@ next release
 
 ### UI Changes
 
-* UI pinned to version [1.28.0](https://github.com/jaegertracing/jaeger-ui/blob/main/CHANGELOG.md#v1280-2023-03-15).
+* UI pinned to version [1.28.1](https://github.com/jaegertracing/jaeger-ui/blob/main/CHANGELOG.md#v1281-2023-03-15).
 
 1.42.0 (2023-02-05)
 -------------------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,7 +39,7 @@ next release
 
 ### UI Changes
 
-* UI pinned to version [1.28.0](https://github.com/jaegertracing/jaeger-ui/blob/main/CHANGELOG.md#v1280-2023-03-13).
+* UI pinned to version [1.28.0](https://github.com/jaegertracing/jaeger-ui/blob/main/CHANGELOG.md#v1280-2023-03-15).
 
 1.42.0 (2023-02-05)
 -------------------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,22 +20,19 @@ next release
 #### New Features
 
 * Support secure OTLP exporter config for hotrod ([@graphaelli](https://github.com/graphaelli) in [#4231](https://github.com/jaegertracing/jaeger/pull/4231))
+* Jaeger YugabyteDB(YCQL) support ([@HarshDaryani896](https://github.com/HarshDaryani896) in [#4220](https://github.com/jaegertracing/jaeger/pull/4220))
 
 #### Bug fixes, Minor Improvements
-* 
-* [chore]: Replace pkg/multierror with standard errors.Join ([@ClementRepo](https://github.com/ClementRepo) in [#4293](https://github.com/jaegertracing/jaeger/pull/4293))
-* [chore] remove pkg/multicloser ([@yurishkuro](https://github.com/yurishkuro) in [#4291](https://github.com/jaegertracing/jaeger/pull/4291))
-* Refactor: build linux artifacts only for PR ([@Eileen-Yu](https://github.com/Eileen-Yu) in [#4286](https://github.com/jaegertracing/jaeger/pull/4286))
-* #4163 | Speed-up CI by using published UI artifacts ([@shubbham1219](https://github.com/shubbham1219) in [#4251](https://github.com/jaegertracing/jaeger/pull/4251))
+
+* Replace pkg/multierror with standard errors.Join ([@ClementRepo](https://github.com/ClementRepo) in [#4293](https://github.com/jaegertracing/jaeger/pull/4293))
+* Remove pkg/multicloser ([@yurishkuro](https://github.com/yurishkuro) in [#4291](https://github.com/jaegertracing/jaeger/pull/4291))
+* Refactor build linux artifacts only for PR ([@Eileen-Yu](https://github.com/Eileen-Yu) in [#4286](https://github.com/jaegertracing/jaeger/pull/4286))
+* Speed-up CI by using published UI artifacts ([@shubbham1219](https://github.com/shubbham1219) in [#4251](https://github.com/jaegertracing/jaeger/pull/4251))
 * Update Go version to 1.20 ([@SaarthakMaini](https://github.com/SaarthakMaini) in [#4206](https://github.com/jaegertracing/jaeger/pull/4206))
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,34 @@ next release
...
+1.43.0 (2023-03-13)
+-------------------
+### Backend Changes
+
+#### New Features
+
+* Support secure OTLP exporter config for hotrod ([@graphaelli](https://github.com/graphaelli) in [#4231](https://github.com/jaegertracing/jaeger/pull/4231))
+
+#### Bug fixes, Minor Improvements
+*
```
albertteoh, line 25: empty bullet point

CHANGELOG.md
```diff
@@ -13,6 +13,34 @@ next release
...
+### Backend Changes
+
+#### New Features
+
+* Support secure OTLP exporter config for hotrod ([@graphaelli](https://github.com/graphaelli) in [#4231](https://github.com/jaegertracing/jaeger/pull/4231))
+
+#### Bug fixes, Minor Improvements
+* 
+* [chore]: Replace pkg/multierror with standard errors.Join ([@ClementRepo](https://github.com/ClementRepo) in [#4293](https://github.com/jaegertracing/jaeger/pull/4293))
+* [chore] remove pkg/multicloser ([@yurishkuro](https://github.com/yurishkuro) in [#4291](https://github.com/jaegertracing/jaeger/pull/4291))
```
albertteoh, line 27: What do you think about removing the `[chore]` prefixes and Upper case the sentences (same below):
```
* Replace ...
* Remove ...
```

CHANGELOG.md
```diff
@@ -13,6 +13,34 @@ next release
...
+* Refactor: build linux artifacts only for PR ([@Eileen-Yu](https://github.com/Eileen-Yu) in [#4286](https://github.com/jaegertracing/jaeger/pull/4286))
+* #4163 | Speed-up CI by using published UI artifacts ([@shubbham1219](https://github.com/shubbham1219) in [#4251](https://github.com/jaegertracing/jaeger/pull/4251))
```
albertteoh, line 29:
```suggestion
* Speed-up CI by using published UI artifacts ([@shubbham1219](https://github.com/shubbham1219) in [#4251](https://github.com/jaegertracing/jaeger/pull/4251))
```

CHANGELOG.md
```diff
@@ -13,6 +13,34 @@ next release
...
+* Improve contributing guidelines ([@yurishkuro](https://github.com/yurishkuro) in [82d7f9b](https://github.com/jaegertracing/jaeger/commit/82d7f9b6f2da42c2fb483ac8bdd0385a0bec9d2f))
+* [chore]: Use http.MethodGet instead of "GET" ([@my-git9](https://github.com/my-git9) in [#4248](https://github.com/jaegertracing/jaeger/pull/4248))
+* Add internal action to setup Node.js version appropriate for Jaeger UI ([@yurishkuro](https://github.com/yurishkuro) in [#4243](https://github.com/jaegertracing/jaeger/pull/4243))
+* Updating all-in-one path ([@bigfleet](https://github.com/bigfleet) in [#4234](https://github.com/jaegertracing/jaeger/pull/4234))
+* Migrate the use of fsnotify to fswatcher in cert_watcher.go ([@haanhvu](https://github.com/haanhvu) in [#4232](https://github.com/jaegertracing/jaeger/pull/4232))
+* Restore baggage support in HotROD 🚗 ([@yurishkuro](https://github.com/yurishkuro) in [#4225](https://github.com/jaegertracing/jaeger/pull/4225))
+* Jaeger YugabyteDB(YCQL) support ([@HarshDaryani896](https://github.com/HarshDaryani896) in [#4220](https://github.com/jaegertracing/jaeger/pull/4220))
```
albertteoh, line 37: This sounds like a new feature to me, what does everyone think?

CHANGELOG.md
pavolloffay: The changelog is auto-generated. I can can make manual changes to make it nicer.

pavolloffay:
@albertteoh thanks for the review. PR updated

CHANGELOG.md
```diff
@@ -13,6 +13,31 @@ next release
 
 ### UI Changes
 
+1.43.0 (2023-03-13)
```
albertteoh, line 16: Update to today's date?

albertteoh (dismissed):
LGTM

CHANGELOG.md
pavolloffay: fixed

albertteoh (approved):

pavolloffay:
The UI test `Check_if_the_favicon_icon_is_available` is failing and the UI does not load if I run it locally `go run -tags ui ./cmd/all-in-one/main.go`

yurishkuro:
@mszabo-wikia I remember icon file moved in the UI recently

pavolloffay:
I think this is the root cause https://github.com/jaegertracing/jaeger-ui/issues/1275

yurishkuro:
@pavolloffay do you want to just downgrade UI to 1.27.4 and get the release out? It might take time to figure out how the UI build changed.

yurishkuro:
> `Check_if_the_favicon_icon_is_available `

I'm glad we had this test :-). It won't work going forward, but the situation with the blank screen in the UI makes me think we should alter it to look for some other specific element in the UI home page. Ideally it would be something like the Search button, which would only be generated if the correct React app is loaded, but that won't happen via curl.

yurishkuro:
I added a bit of logging to static handler, things look ok from the server side
```
serving /static/index-b9997246.js
serving /static/index-3c2d679b.css
```
and in Chrome too:
image

I think it's really something about the UI build output, the app is incomplete or we're using a wrong mode.

@mszabo-wikia any thoughts?

mszabo-wikia:
Looks like there's this error:
```
Uncaught TypeError: Unknown theme type: undefined, name: undefined
    withSuffix http://localhost:16686/static/index-b9997246.js:69
```

mszabo-wikia:
Although https://github.com/ant-design/ant-design/issues/19002#issuecomment-965958565 is in Chinese, it seems the observed error matches & from some quick local testing the UI loads properly after applying that update to the build config. So this looks to be an antd 3.9.0 -> 3.26 update regression.

mszabo-wikia:
Filed https://github.com/jaegertracing/jaeger-ui/pull/1276. [ee7ce64f6784e5e00718a58680a43098f83a3c94 ](https://github.com/jaegertracing/jaeger-ui/commit/cda1746aa10dad4105cff06c7b279dec367f8b00) is the first commit that breaks without the additional config, so antd 3.26 indeed seems to be at fault.

yurishkuro (dismissed):
I expect favicon test would still fail, need to remove it

cmd/all-in-one/all_in_one_test.go
```diff
@@ -138,9 +138,9 @@ func healthCheck(t *testing.T) {
 	)
 }
 
-func faviconCheck(t *testing.T) {
+func jaegerLogoCheck(t *testing.T) {
 	t.Log("Checking favicon...")
-	resp, err := http.Get(queryURL + "/favicon.ico")
+	resp, err := http.Get(queryURL + "/static/jaeger-logo-ab11f618.svg")
```
yurishkuro, line 143: I wonder if the hash in the filename may change over time and break this test. If it's based on the file checksum then it should be stable. We can improve the test later, e.g. request home page first, and look for logo URL in HTML.

yurishkuro (approved):

cmd/all-in-one/all_in_one_test.go
pavolloffay: the hash should not change if the logo does not change

PR #4302, opened 2023-03-15
Log access to static assets; remove favicon test

yurishkuro:
Only merge after UI upgrade to 1.28

### DIFF
--- a/cmd/query/app/flags.go
+++ b/cmd/query/app/flags.go
@@ -69,7 +69,7 @@ type QueryOptions struct {
 	BasePath string
 	// StaticAssets is the path for the static assets for the UI (https://github.com/uber/jaeger-ui)
 	StaticAssets string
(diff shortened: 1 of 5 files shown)


cmd/query/app/flags.go
```diff
@@ -68,6 +69,8 @@ type QueryOptions struct {
 	BasePath string
 	// StaticAssets is the path for the static assets for the UI (https://github.com/uber/jaeger-ui)
 	StaticAssets string
+	// LogStaticAssetsAccess tells static handler to log access to static access, useful in debugging
```
albertteoh, line 72:
```suggestion
	// LogStaticAssetsAccess tells static handler to log access to static assets, useful in debugging
```

albertteoh (approved):

albertteoh (approved):

PR #4310, opened 2023-03-17
fix Backend prometheus replaced by Const

dsy3502:


pkg/metrics/builder.go
```diff
@@ -68,7 +68,7 @@ func (b *Builder) InitFromViper(v *viper.Viper) *Builder {
 // If the metrics backend supports HTTP endpoint for scraping, it is stored in the builder and
 // can be later added by RegisterHandler function.
 func (b *Builder) CreateMetricsFactory(namespace string) (metrics.Factory, error) {
-	if b.Backend == "prometheus" {
+	if b.Backend == defaultMetricsBackend {
```
yurishkuro, line 71: This makes the code less readable. The acceptable change would be to add a new constant:
```
backendPrometheus = "prometheus"
defaultMetricsBackend = backendPrometheus
...
if b.Backend == backendPrometheus
```

yurishkuro:
feel free to reopen once comments are addressed

PR #4293, opened 2023-03-11
[chore]: Replace pkg/multierror with standard errors.Join

ClementRepo:
## Which problem is this PR solving?
Resolves #4292

PR #4206, opened 2023-02-04
Update Go version to 1.20

SaarthakMaini:
## Which problem is this PR solving?
- Resolves #4202 

## Short description of the changes
All the files mentioned in issue #4202 have been updated with Go version 1.20 replacing version 1.19

Signed-off-by: Saarthak Maini <saarthakmainiofficial@gmail.com>

Issue #4202, opened 2023-02-02
[Feature]: Migrate to go 1.20 


scalalang2:
### Requirement

[Go 1.20 officially released today](https://tip.golang.org/doc/go1.20)
There's improvements on the garbage collector. It informs that its overall performace improves by up to 2% and reduces memory overheads. I guess that migrating to go 1.20 will provides improved performance, bug fixes, security enhancements and etc.

### Problem

There's no problem on Jaeger.
This proposal only include a simple suggestion.

Thanks.

### Proposal

Migrate to go runtime 1.20.

yurishkuro:
Files to change:
```
$ grep -r 1.19 .
./go.mod:3:go 1.19
./docker/Makefile:4:GOLANG_IMAGE := golang:1.19-alpine
./.golangci.yml:2:  go: '1.19'
./.golangci.yml:30:    go: "1.19"
./.github/workflows/fossa.yml:28:          go-version: 1.19.x
./.github/workflows/ci-release.yml:34:        go-version: 1.19.x
./.github/workflows/ci-unit-tests.yml:33:        go-version: 1.19.x
./.github/workflows/ci-docker-build.yml:34:        go-version: 1.19.x
./.github/workflows/ci-build-binaries.yml:51:        go-version: 1.19.x
./.github/workflows/ci-crossdock.yml:34:        go-version: 1.19.x
./.github/workflows/ci-elasticsearch.yml:46:        go-version: 1.19.x
./.github/workflows/ci-grpc-badger.yml:27:        go-version: 1.19.x
./.github/workflows/ci-hotrod.yml:33:        go-version: 1.19.x
./.github/workflows/ci-protogen-tests.yml:29:        go-version: 1.19.x
./.github/workflows/ci-all-in-one-build.yml:33:        go-version: 1.19.x
./.github/workflows/ci-cassandra.yml:39:        go-version: 1.19.x
./.github/workflows/ci-opensearch.yml:43:        go-version: 1.19.x
./.github/workflows/ci-kafka.yml:27:        go-version: 1.19.x
```

SaarthakMaini:
PR for the issue has already been made and would be merged during the next Release Cycle.

scalalang2:
👍

mx-psi:
[...]

PR #4328, opened 2023-03-22
Add back support for Go 1.19 when using Jaeger as a library

mx-psi:

- Relates to #4202

This partially reverts #4206 and fully reverts #4291 and #4293.

On the OpenTelemetry Collector project we aim to provide support for all officially supported Go versions. Per [Go's support policy][1], Go 1.19 will remain a supported Go version until Go 1.21 is released.

Since we take a dependency on Jaeger as a library, we would prefer Jaeger to remain compatible with Go 1.19 as a library to be able to keep it updated.

This patch does not revert some changes on #4206; Jaeger binaries would continue to be built with Go 1.20 and benefit from all performance and security improvements that the 1.20 release cycle offers. It does impose the requirement that new code is compatible with Go 1.19 (which sounds okay to me since there was no specific need for Go 1.20 mentioned on the related issues, the only obvious downside is delaying the removal of some helper code).

[1]: https://go.dev/doc/devel/release#policy

Signed-off-by: Pablo Baeyens <pablo.baeyens@datadoghq.com>

### DIFF
--- a/.github/workflows/ci-unit-tests-1.19.yml
+++ b/.github/workflows/ci-unit-tests-1.19.yml
@@ -1,0 +1,48 @@
+name: Unit Tests
+
+on:
+  push:
+    branches: [main]
+
+  pull_request:
+    branches: [main]
+
+# See https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions
+permissions:  # added using https://github.com/step-security/secure-workflows
+  contents: read
+
+env:
+  # Using upload token helps against rate limiting errors.
+  # Cannot define it as secret as we need it accessible from forks.
+  # See https://github.com/codecov/codecov-action/issues/837
+  CODECOV_TOKEN: f457b710-93af-4191-8678-bcf51281f98c
+
+jobs:
+  unit-tests-1.19:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Harden Runner
+      uses: step-security/harden-runner@1f99358870fe1c846a3ccba386cc2b2246836776
+      with:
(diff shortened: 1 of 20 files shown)


mx-psi:
@yurishkuro I would love to know if this is something you and the Jaeger project would be willing to accept before I continue working on it. Would you mind chiming in?

yurishkuro:
I'd be fine if this was only the go.mod change, but rolling back so many changes doesn't make sense to me. And frankly neither does the OTEL collector policy - it's not a library where last two versions policy makes sense, collector is a binary, you should be able to upgrade the runtime as you see fit.

mx-psi:
@yurishkuro 

> I'd be fine if this was only the go.mod change, but rolling back so many changes doesn't make sense to me

To be clear, the rest of the changes are just "don't use `errors.Join` and use the previous thing", but I understand that this is bigger than expected, I will close the PR.

> neither does the OTEL collector policy - it's not a library where last two versions policy makes sense, collector is a binary, you should be able to upgrade the runtime as you see fit.

The Collector is consumed as a library for a wide variety of use cases. This includes downstream distros, OpenTelemetry Collector builder users, vendor-specific observability agents like the Datadog Agent or the Grafana Agent or [tooling](https://github.com/CtrlSpice/otel-desktop-viewer). It's even used as a library by [Jaeger itself](https://github.com/jaegertracing/jaeger/blob/6ab3f0184ca2cd8900079998a38b1690bf1c03d2/go.mod#L50-L56) :)

yurishkuro:
I opened #4329 to decide if we want to institute similar policy going forward.

Issue #3949, opened 2022-10-06
[Feature]: Support accepting OTLP via Kafka


yurishkuro:
### Requirement

As a user of OpenTelemetry Collector, I would like to send OTLP-encoded data to Jaeger.

### Problem

Jaeger ingester today only understands 3 formats, `json` (Jaeger Protobuf as JSON), `protobuf` (Jaeger Protobuf), and `zipkin-thrift`.

### Proposal

Add two more recognized encodings, `otlp-json` and `otlp-proto`.

It should be fairly simple to implement since we already import OTLP->Jaeger decoders when receiving OTLP via RPC.

NonsoAmadi10:
Hello yurishkuro, Please can I be assigned to this. I want to try my hands on this issue

NonsoAmadi10:
Hello Yurishkuro, can I be assigned to this please? I want this to be my first open source contribution

NonsoAmadi10:
Also what guide can I follow as to how I can make my Pull Request?

yurishkuro:
> Also what guide can I follow as to how I can make my Pull Request?

Are you asking in general how to do it on GitHub [[1]], or specifically in Jaeger project [[2]]?

> I would be requiring a marshal and an unmashaller for both encodings, Is this alright?

I think you only need to unmarshall https://github.com/jaegertracing/jaeger/blob/3d67ac86f53866332568b0f542265b00d86407fc/cmd/ingester/app/builder/builder.go#L35

But when introducing a new encoding you probably don't want it to be available as output encoding of the collector, only the input encoding in the ingester. We probably don't support such asymmetry cleanly at the moment. The simplest solution is to fail collector at startup if otlp is selected at output encoding. But if you can implement bidirectional marshalling, thenwe could go with full symmetric support (OTEL contrib does have translation of OTLP to and from Jaeger proto).

[1]: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request
[2]: https://github.com/jaegertracing/jaeger/blob/main/CONTRIBUTING.md

NonsoAmadi10:
Wow, alright. I would keep this in mind while implementing. Thank you Yuri!! I'd be raising my first PR soon for you to look and give feedback

yurishkuro:
@Vishal-Chdhry give it a try

SaarthakMaini:
Hello,

Has this issue been resolved ?

[...]

PR #4333, opened 2023-03-23
Feature: Added Support accepting OTLP via Kafka

metapox:
- Adds two more recognized encodings, otlp-json and otlp-proto.
- https://github.com/jaegertracing/jaeger/issues/3949

-


cmd/ingester/app/processor/span_processor.go
```diff
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"log"
```
yurishkuro, line 21: we don't use log package, only zap

metapox: I see

Tell me about it.
Because span_processor does not hold the logger object of zap,
We need to propagate zap objects from parallel_processor to span_processor.
The interface changes, is this acceptable?

Or should I create a new logger in span_processor as follows?
https://github.com/jaegertracing/jaeger/blob/main/cmd/flags/service.go#L93-L96

yurishkuro: `*zap.Logger` is only created in `main()` and passed through to all components that need it. It's ok to change the component constructor's interface to accept a logger.

yurishkuro (commented):
the unit tests here are not simulating a multi-message batch, we need to include that as a test scenario

plugin/storage/kafka/writer_test.go
```diff
@@ -47,7 +47,7 @@ var (
 				RefType: model.ChildOf,
 			},
 		},
-		Flags:     model.Flags(1),
+		Flags:     model.Flags(0),
```
yurishkuro, line 50: not clear why this change is needed

yurishkuro:
closing due to inactivity

Issue #1051, opened 2018-09-05
Jaeger Query can OOM when retrieving large traces


vprithvi:
## Requirement - what kind of business use case are you trying to solve?
- Retrieve large traces
- Be resilient against bad instrumentation using same traceID for all traces

## Problem - what in Jaeger blocks you from solving the requirement?
Jaeger Query OOMs on retrieval of large traces on Cassandra. 
If someone is crafty, they can easily create a trace with millions of spans, and attempt to retrieve it to systematically bring down all jaeger-query instances. 

## Any open questions to address
- Does this affect ES as well?

pavolloffay:
Related issue to ES, although its related to overloading ES not query https://github.com/jaegertracing/jaeger/issues/960

jpkrohling:
We probably want to add this scenario to our test cases

cc @jkandasa

yurishkuro:
We may consider adding LIMIT to Cassandra queries by trace ID.

pavolloffay:
Limit on the number of spans in a trace?

yurishkuro:
Well there are two separate things. We're seeing some traces in production with many millions of spans so we want to implement a limit on ingestion. But separately, there can be a limit on reading side to avoid loading too much even if that many spans were saved.

annanay25:
The olivere/elastic library supports sending a range query using `From(0)` and `Size(100)` to fetch a predefined number of documents - for ex: https://github.com/olivere/elastic/blob/release-branch.v6/example_test.go#L172

Would it make sense to pass this range as a param in MultiSearch here - https://github.com/jaegertracing/jaeger/blob/master/plugin/storage/es/spanstore/reader.go#L256 ?

pavolloffay:
Perhaps yes, I am not 100% percent sure how the current search works. Would you like to submit a patch?

annanay25:
Sure, I'll give it a shot soon.

annanay25:
@pavolloffay could you review this? https://github.com/annanay25/jaeger/commit/2d167b86a14f1a88a6e7ed7cee84399f8403ba88
Reduced the default fetch limit from 10,000 to 1,000. And added a `FetchNext` function to implement scroll.

yurishkuro:
[...]

PR #4336, opened 2023-03-23
BUG: Remove TerminateAfter from Elasticsearch/Opensearch query resulting in incomplete span count/list

### LINKED ISSUE #4330: [Bug]: Elastic/Opensearch backend does not return all trace spans
### What happened?

The Jaeger query is not resolving all spans with Elasticsearch/Opensearch Backend.
This happens as the `_msearch` request uses the `terminate_after` option in the body, which result in a wrong `hits.total` value. Due to the wrong `hits.total` value the search loop stops before getting all spans. 

### Steps to reproduce

1. Setup Jaeger with Elasticsearch or Openearch Backend
2. Ingest trace with more then 40k spans and leave the `--es.max-doc-count` untouched
3. Try to get the trace via api or UI

### Expected behavior

I would expect jaeger to resolve all spans from the trace.

### Relevant log output

_No response_

### Screenshot

_No response_

### Additional context

`teminate_after` is set to max doc count.
https://github.com/jaegertracing/jaeger/blob/main/plugin/storage/es/spanstore/reader.go#L208

The total hits is compared to the length of all saved trace span hits.
https://github.com/jaegertracing/jaeger/blob/main/plugin/storage/es/spanstore/reader.go#L432

Example of how `teminate_after` behaves with hits total:
```
GET staging-jaeger-span-read/_search
{

  "terminate_after": 5
}

{
  "took": 24,
  "timed_out": false,
  "terminated_early": true,
  "_shards": {
    "total": 51,
    "successful": 51,
    "skipped": 0,
    "failed": 0
  },
  "hits": {
    "total": {
      "value": 230,
      "relation": "eq"
    },
    "max_score": 1,
    "hits":  [...]
  }
}
```
```
GET staging-jaeger-span-read/_search

{
  "took": 33,
  "timed_out": false,
  "_shards": {
    "total": 51,
    "successful": 51,
    "skipped": 0,
    "failed": 0
  },
  "hits": {
    "total": {
      "value": 10000,
      "relation": "gte"
    },
    "max_score": 1,
    "hits": [...]
  }
}
```

Possible solutions I can think of:
1. Dont set terminate_after or set it to 0 - Could lead to and increased query time per msearch request
2. Get total hits per trace with `_search` and `track_total_hits` once before and match against the correct total hits value - additional time spend due to extra request.

### Jaeger backend version

_No response_

### SDK

_No response_

### Pipeline

_No response_

### Stogage backend

Opensearch 2.6.0

### Operating system

Linux

### Deployment model

Kubernetes

### Deployment configs

_No response_

Jakob3xD:
Closes #4330 

- Remove TerminateAfter from Elasticsearch/Opensearch query resulting in incomplete span count/list.

### DIFF
--- a/plugin/storage/es/spanstore/reader.go
+++ b/plugin/storage/es/spanstore/reader.go
@@ -204,8 +204,7 @@ func getSourceFn(archive bool, maxDocCount int) sourceFn {
 	return func(query elastic.Query, nextTime uint64) *elastic.SearchSource {
 		s := elastic.NewSearchSource().
 			Query(query).
-			Size(maxDocCount).
-			TerminateAfter(maxDocCount)
+			Size(maxDocCount)
 		if !archive {
 			s.Sort("startTime", true).
 				SearchAfter(nextTime)
--- a/plugin/storage/es/spanstore/reader_test.go
+++ b/plugin/storage/es/spanstore/reader_test.go
@@ -1296,3 +1296,29 @@ func TestBuildTraceByIDQuery(t *testing.T) {
 		assert.Equal(t, test.query, q)
 	}
 }
+
+func TestTerminateAfterNotSet(t *testing.T) {
+	srcFn := getSourceFn(false, 99)
+	searchSource := srcFn(elastic.NewMatchAllQuery(), 1)
+	sp, err := searchSource.Source()
+	require.NoError(t, err)
+
+	searchParams, ok := sp.(map[string]interface{})
+	require.True(t, ok)
+
+	termAfter, ok := searchParams["terminate_after"]
+	require.False(t, ok)
+	assert.Nil(t, termAfter)
+
+	query, ok := searchParams["query"]
+	require.True(t, ok)
+
+	queryMap, ok := query.(map[string]interface{})
+	require.True(t, ok)
+	_, ok = queryMap["match_all"]
+	require.True(t, ok)
+
+	size, ok := searchParams["size"]
+	require.True(t, ok)
+	assert.Equal(t, 99, size)
+}


plugin/storage/es/spanstore/reader.go
albertteoh, line 1: We should write a test for this change; a suggestion:

```
type fakeQuery struct {
	fnSource func() (interface{}, error)
}

func (f fakeQuery) Source() (interface{}, error) {
	return "fake query source", nil
}

func TestTerminateAfterNotSet(t *testing.T) {
	srcFn := getSourceFn(false, 99)
	searchSource := srcFn(fakeQuery{}, 1)
	sp, err := searchSource.Source()
	require.NoError(t, err)

	searchParams, ok := sp.(map[string]interface{})
	require.True(t, ok)

	termAfter, ok := searchParams["terminate_after"]
	require.False(t, ok)
	assert.Nil(t, termAfter)

	query, ok := searchParams["query"]
	require.True(t, ok)

	queryStr, ok := query.(string)
	require.True(t, ok)
	assert.Equal(t, "fake query source", queryStr)

	size, ok := searchParams["size"]
	require.True(t, ok)
	assert.Equal(t, 99, size)
}
```

albertteoh (commented):
It seems quite reasonable to me to remove the terminate after. IIUC, its primary use case is to protect shards from using a large amount of memory while also improving query times. 
However, I feel query correctness should take precedence, especially if shard capacity can be configured.

I've just added a suggestion relating to testing.

plugin/storage/es/spanstore/reader.go
Jakob3xD: Thanks for your reply.
I am looking at the current tests and I am not 100% sure where to add your suggestion. Can you guide me on where to add it or add it your self?

Beside that I don't fully understand the current tests due to the complexity of types, strucs and functions of jaeger, but I have a suggestion for the tests. We have a `GetLargeSpans` test which creates a trace with 10008 spans by calling the loadParseAndWriteLargeTrace function. Increasing the number to something over 50000 would trigger my issue with `terminate_after`, but the tests fails on my workstation as it times out. So instead of increasing the spans we could decrease the defaultMaxDocCount in the elasticsearch_test.go to a value below 2000 and also trigger the issue. Maybe we can overwrite the maxDoc only of the `GetLargeSpans` test some how?

From where I got those numbers? The test create an index with 5 shards so with maxDocs of 10k your have a max of 5*10000 docs to search. If you have 10008 docs you need a value below 10008 / 5 to trigger an error caused by `terminate_after`.

albertteoh: > Can you guide me on where to add it or add it your self?

I was thinking to add this into plugin/storage/es/spanstore/reader_test.go. This just checks the query configuration set by `getSourceFn`. The test should fail without your changes in this PR.

> We have a GetLargeSpans test

Yup, I think adding an integration test is a good idea, so long as it is reliable (not flaky).

> Maybe we can overwrite the maxDoc only of the GetLargeSpans test some how?

We can introduce a `maxDocCount` function parameter to `initSpanstore`, which is then passed in from the test functions.

Jakob3xD: > The test should fail without your changes in this PR.

Yes, it fails without it.



> As you can see, the kafka integration tests are currently failing

It seems to me that other PR also fail on the kafka tests.

> We can introduce a maxDocCount function parameter to initSpanstore, which is then passed in from the test functions.

I think I understand most of the elastic integration tests now and I don't see an easy way of doing this.
Golang test calls `TestElasticsearchStorage` calling -> `testElasticsearchStorage` calling -> `s.initializeES` calling -> `initSpanstore`. `initSpanstore` defines the `SpanReader` which is later on used for all the tests. The actual tests are generic so overwriting the `SpanReader` inside the test is not an option.

Jakob3xD: The failing FOSS and Kafka test are not related to the MR. Fossa can be retried and Kafka seems to be a general issue.

CHANGELOG.md
```diff
@@ -47,6 +47,7 @@ next release
 * Updating all-in-one path ([@bigfleet](https://github.com/bigfleet) in [#4234](https://github.com/jaegertracing/jaeger/pull/4234))
 * Migrate the use of fsnotify to fswatcher in cert_watcher.go ([@haanhvu](https://github.com/haanhvu) in [#4232](https://github.com/jaegertracing/jaeger/pull/4232))
 * Restore baggage support in HotROD 🚗 ([@yurishkuro](https://github.com/yurishkuro) in [#4225](https://github.com/jaegertracing/jaeger/pull/4225))
+* Remove TerminateAfter for elastic queries ([@jakob3xd](https://github.com/Jakob3xD)) in [#4336](https://github.com/jaegertracing/jaeger/pull/4336)
```
albertteoh, line 50: Thanks for thinking about adding this, but it's not necessary. We have scripts that will generate the changelog on each release. Please remove this entry.

albertteoh (approved):

albertteoh:
For posterity, `TerminateAfter` was first introduced in [this PR](https://github.com/jaegertracing/jaeger/pull/1202/files#diff-c2d43ea87d6708a8018f31e96af64225b523611237b8ecfc24f11a34e38ae8ffR272) to address [jaeger-query OOMing](https://github.com/jaegertracing/jaeger/issues/1051). 

However, based on [this SO answer](https://stackoverflow.com/questions/40423696/terminate-after-in-elasticsearch) and the [ES docs](https://www.elastic.co/guide/en/elasticsearch/client/java-api/current/java-search-terminate-after.html#java-search-terminate-after), I don't see how `TerminateAfter` will help reduce memory pressure in jaeger-query although, iiuc, it can help reduce memory pressure in individual ES shards.

This bug fix is technically a breaking change, so I think we should report it as such in release notes.

yurishkuro:
I think you meant this PR #1283 (#1202 was closed unmerged)

yurishkuro:
This [SO explanations](https://stackoverflow.com/a/40428335/2673284) also confirms that TerminateAfter is the wrong thing to do.

PR #4337, opened 2023-03-24
Add GO param to Makefile

yurishkuro:
`go` command was hardcoded, making it difficult to run with `gotip`. Now can do `make test GO=gotip`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 SHELL := /bin/bash
 JAEGER_IMPORT_PATH = github.com/jaegertracing/jaeger
 STORAGE_PKGS = ./plugin/storage/integration/...
+GO = go
 
 include docker/Makefile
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,6 +1,7 @@
 SHELL := /bin/bash
 JAEGER_IMPORT_PATH = github.com/jaegertracing/jaeger
 STORAGE_PKGS = ./plugin/storage/integration/...
+GO = go
```
albertteoh, line 4: What do you think about using `?=`? It means users can `export GO=gotip` just once then just `make test` each time during development, instead of needing to alway run `make test GO=gotip`.

It also offers the option of `make test GO=gotip` if users don't wish to "permanently" set their `GO` env var.

yurishkuro: I personally don't like such auto-import of env vars, who knows what else is in there...

PR #4342, opened 2023-03-26
Support reloading ES client's password from file

haanhvu:
Resolves the final 2nd and 3rd steps for #3924 as specified [here](https://github.com/jaegertracing/jaeger/issues/3924#issuecomment-1424754016)


haanhvu:
@yurishkuro Could you pls review if this implementation approach seems to make sense? If yes, I'll continue to fix current broken tests and add related tests.

plugin/storage/es/factory.go
```diff
@@ -236,8 +266,26 @@ var _ io.Closer = (*Factory)(nil)
...
-	return f.Options.GetPrimary().TLS.Close()
+	errs = append(errs, f.Options.GetPrimary().TLS.Close())
+	return errors.Join(errs...)
+}
+
+func (f *Factory) onPasswordChange(config config.ClientBuilder, esClient es.Client) {
+	b, err := os.ReadFile(filepath.Clean(config.GetPasswordFilePath()))
+	if err == nil {
+		newPassword := strings.TrimRight(string(b), "\r\n")
+		config.SetPassword(newPassword)
```
yurishkuro, line 284: why do we need to set pwd on the config?

plugin/storage/es/factory.go
```diff
@@ -236,8 +266,26 @@ var _ io.Closer = (*Factory)(nil)
...
+		esClient.SetBasicAuth(config.GetUsername(), newPassword)
```
yurishkuro, line 285: is `esClient.SetBasicAuth` method reentrant? That is, would the take effect after the client is already created? It would be good to add a unit test that validates it (assuming we can mock the ES server with a test server, which iirc we do somewhere by using a fake http.RoundTripper)

plugin/storage/es/factory.go
```diff
@@ -94,12 +100,36 @@ func (f *Factory) Initialize(metricsFactory metrics.Factory, logger *zap.Logger)
 		return fmt.Errorf("failed to create primary Elasticsearch client: %w", err)
 	}
 	f.primaryClient = primaryClient
+
+	f.watchers = make([]*fswatcher.FSWatcher, 2)
+	onPrimaryClientPasswordChange := func() { f.onPasswordChange(f.primaryConfig, f.primaryClient) }
```
yurishkuro, line 105:
```suggestion
	onPrimaryClientPasswordChange := func() { 
		f.onPasswordChange(f.primaryConfig, f.primaryClient)
	}
```

plugin/storage/es/factory.go
```diff
@@ -94,12 +100,36 @@ func (f *Factory) Initialize(metricsFactory metrics.Factory, logger *zap.Logger)
...
+			err)
+	}
+	f.watchers = append(f.watchers, watcher)
+
 	if f.archiveConfig.IsStorageEnabled() {
 		f.archiveClient, err = f.archiveConfig.NewClient(logger, metricsFactory)
 		if err != nil {
 			return fmt.Errorf("failed to create archive Elasticsearch client: %w", err)
 		}
+		onArchiveClientPasswordChange := func() { f.onPasswordChange(f.archiveConfig, f.archiveClient) }
```
yurishkuro, line 121: duplicated code, can we DRY it ?

plugin/storage/es/factory.go
haanhvu: Now I think we don't.

Let's see if I understand correctly:
- Users can set password in `config.Password` and/or `config.PasswordFilePath`
- The password from `config.PasswordFilePath` will override `config.Password`
- Later if `config.PasswordFilePath` is empty or deleted, we'll use `config.Password` instead

Is that correct?

plugin/storage/es/factory.go
haanhvu: From my study of the codebase of `github.com/olivere/elastic`, yes.

You can see the implementation of `esClient.SetBasicAuth` [here](https://github.com/jaegertracing/jaeger/pull/4342/files#diff-178ba3fec75295e1f7b9539382116134f29d80c7e52305cf8a87b4ffc8fdeb80R84-R87).

Basically what it does is what `github.com/olivere/elastic` does to set username and password on ES client: [get a ClientOptionFunc from calling SetBasicAuth](https://github.com/olivere/elastic/blob/29ee98974cf1984dfecf53ef772d721fb97cb0b9/client.go#L466) then [apply that ClientOptionFunc on ES client](https://github.com/olivere/elastic/blob/release-branch.v7/client.go#L347).

I'll add a test for this.

plugin/storage/es/factory.go
yurishkuro: Only-once is a feasible implementation of the PasswordFilePath flag, but the point of watching the file for updates is to detect if the password is changed and begin using the new value. Same with token file.

haanhvu: Yeah I understand `PasswordFilePath` needs to be watched. I already did it in the implementation.

What I'm discussing about is the (maybe rare) case when users set password in both `PasswordFilePath` and `Password`. In this case, the password from `PasswordFilePath` should override `Password`, right?

And later, in case `PasswordFilePath` is removed or empty-content, what would we do?

Initially, in the code that you reviewed, I called `SetPassword` to make sure `Password` and `PasswordFilePath` always had the same non-empty password. And in case `PasswordFilePath` is removed or empty-content, we'd use the latest non-empty password stored in `Password`. But now I'm not sure if it's a good idea.

yurishkuro: We should not allow setting both string pwd and file name, it's a configuration error that can halt the startup.

If the watched file becomes empty, our default behavior in all other components that watch files is to use the previously read value, and log a message informing the operator about that.

haanhvu:
I'm changing the approach from resetting the password to swapping the whole ES client. That might be easier to test and wouldn't introduce subtle errors...

haanhvu:
@yurishkuro Could you help me a little here?

My approach is creating [a channel of ES options](https://github.com/jaegertracing/jaeger/pull/4342/files#diff-de5aae4da3cb03d3a73919aca1f6f38181b906dfd205bb0def4bf7b9d6042992R449) and [a channel of ES client](https://github.com/jaegertracing/jaeger/pull/4342/files#diff-de5aae4da3cb03d3a73919aca1f6f38181b906dfd205bb0def4bf7b9d6042992R101). Everytime password from file changes, a new array of options is channeled to create a new client. That client is [channeled to the factory](https://github.com/jaegertracing/jaeger/pull/4342/files#diff-bd7140c4b097bbb7e2462f2b512528d3f45af2317fe008211c293553bff71c56R94-R95).

Some tests are broken but basically the one problem here is the client channel can receive the client, but cannot assign client in the factory. The test result says [assert.NotNil(t, <-f.primaryClientChan)](https://github.com/jaegertracing/jaeger/pull/4342/files#diff-fb005680ccab1461b654fe14dcb0a53eb862af4f9bfc15b279731e287083b5daR99) passes but [assert.NotNil(t, f.primaryClient)](https://github.com/jaegertracing/jaeger/pull/4342/files#diff-fb005680ccab1461b654fe14dcb0a53eb862af4f9bfc15b279731e287083b5daR100) fails. When I debugged the problem seems to be in [assigning client in factory](https://github.com/jaegertracing/jaeger/pull/4342/files#diff-bd7140c4b097bbb7e2462f2b512528d3f45af2317fe008211c293553bff71c56R100-R102).

Could you take a look to see what I did wrong here?

yurishkuro:
I think channels are overkill for this. There is already a callback mechanism in fswatcher that can be used to propagate all the changes. Critically though, your changes so far don't actually affect how the storage works, because `factory.CreateSpanReader()` is called only once, which will instantiate the client only once: https://github.com/jaegertracing/jaeger/blob/6cbc11eb17e0badd94d75faff746934323772d36/plugin/storage/es/factory.go#L195-L196

If recreating the client is the only option (which may be fine), then this is how  I would go about it:
  * don't pass a client to NewSpanReader, but pass a function that returns a client
  * the function will retrieve the client from `f.primaryClient` which should be turned into atomic.Value
  * pass a callback function to fswatcher such that when that call back is invoked the client is re-created and replaced in `f.primaryClient`

plugin/storage/es/factory.go
```diff
@@ -236,8 +256,57 @@ var _ io.Closer = (*Factory)(nil)
...
 	if cfg := f.Options.Get(archiveNamespace); cfg != nil {
 		cfg.TLS.Close()
 	}
-	return f.Options.GetPrimary().TLS.Close()
+	errs = append(errs, f.Options.GetPrimary().TLS.Close())
+	return errors.Join(errs...)
+}
+
+// initializeClient initializes either primary or archive ES client
+func (f *Factory) initializeClient(config config.ClientBuilder, options []elastic.ClientOptionFunc,
```
yurishkuro, line 271: assuming this function is necessary, its interface should be simpler: create and return a client or error. There is no need to pass atomic holder, for example.

pkg/es/config/config.go
```diff
@@ -94,7 +95,8 @@ type TagsAsFields struct {
 
 // ClientBuilder creates new es.Client
 type ClientBuilder interface {
-	NewClient(logger *zap.Logger, metricsFactory metrics.Factory) (es.Client, error)
+	NewClient(logger *zap.Logger, options []elastic.ClientOptionFunc, metricsFactory metrics.Factory) (es.Client, error)
+	GetConfigOptions(logger *zap.Logger) ([]elastic.ClientOptionFunc, error)
```
yurishkuro, line 99: First, I just put a PR #4396 to get rid of this interface since it's an unnecessary complication. Once it merges please rebase.

Second, I don't think you need to be exposing ES options in this way. Because all other functions now just use the Configuration struct, when the password file is changed and callback is called, you can do this:
```
cfgClone = *f.primaryConfig // create a copy of the config
cfgClone.Password = // read password from file
newClient := f.newClientFn(&cfgClone, ...)
f.client.Store(newClient)
```

haanhvu: > Second, I don't think you need to be exposing ES options in this way. Because all other functions now just use the Configuration struct, when the password file is changed and callback is called, you can do this:

I could have done that with the old interface too. The main reason I exposed and stored options was to store the unchangeable options. Then when password changes, we just need to append the new authentication option to the list of options. If you don't think that's necessary, I'll recreate the client from scratch as you said.

plugin/storage/es/factory.go
```diff
@@ -96,45 +101,48 @@ func (f *Factory) Initialize(metricsFactory metrics.Factory, logger *zap.Logger)
...
+		f.atomicArchiveClient.Store(&archiveClient)
 	}
+
 	return nil
 }
 
 // CreateSpanReader implements storage.Factory
 func (f *Factory) CreateSpanReader() (spanstore.Reader, error) {
-	return createSpanReader(f.metricsFactory, f.logger, f.primaryClient, f.primaryConfig, false)
+	return createSpanReader(f.metricsFactory, f.logger, *f.atomicPrimaryClient.Load(), f.primaryConfig, false)
```
yurishkuro, line 119: You cannot dereference client here, if you do it will not be swappable once the reader is created. You need to pass the whole atomic var.

This indicates that you don't have a unit test which validates this new capability of updating password via file.

plugin/storage/es/factory.go
```diff
@@ -148,7 +156,7 @@ func createSpanReader(
 		return nil, fmt.Errorf("--es.use-ilm must always be used in conjunction with --es.use-aliases to ensure ES writers and readers refer to the single index mapping")
 	}
 	return esSpanStore.NewSpanReader(esSpanStore.SpanReaderParams{
-		Client:                        client,
+		Client:                        func() es.Client { return client },
```
yurishkuro, line 159: So you have a function here but it always returns the same client.

plugin/storage/es/factory.go
```diff
@@ -96,13 +101,29 @@ func (f *Factory) Initialize(metricsFactory metrics.Factory, logger *zap.Logger)
 	if err != nil {
 		return fmt.Errorf("failed to create primary Elasticsearch client: %w", err)
 	}
-	f.primaryClient = primaryClient
+	f.primaryClient.Store(&primaryClient)
+
+	f.watchers = make([]*fswatcher.FSWatcher, 2)
```
yurishkuro, line 106: Don't preallocate, you don't know if the second watcher will be added. Just use append

plugin/storage/es/factory.go
```diff
@@ -96,45 +101,48 @@ func (f *Factory) Initialize(metricsFactory metrics.Factory, logger *zap.Logger)
 	if err != nil {
 		return fmt.Errorf("failed to create primary Elasticsearch client: %w", err)
 	}
-	f.primaryClient = primaryClient
+	f.atomicPrimaryClient.Store(&primaryClient)
```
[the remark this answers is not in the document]
haanhvu: Let me see. This is the first time I work with atomic. The doc says `These functions require great care to be used correctly`. And I just saw a bunch of lock copy lint failures...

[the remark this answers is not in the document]
yurishkuro: You're right, atomic.Pointer only accepts pointers to things https://go.dev/play/p/omLHKDd5P9o

plugin/storage/es/factory.go
```diff
@@ -96,13 +101,29 @@ func (f *Factory) Initialize(metricsFactory metrics.Factory, logger *zap.Logger)
 	if err != nil {
 		return fmt.Errorf("failed to create primary Elasticsearch client: %w", err)
 	}
-	f.primaryClient = primaryClient
+	f.primaryClient.Store(&primaryClient)
+
+	f.watchers = make([]*fswatcher.FSWatcher, 0)
```
yurishkuro, line 106: don't need this statement, append will work just fine with nil slice

plugin/storage/es/factory.go
```diff
@@ -239,8 +260,31 @@ var _ io.Closer = (*Factory)(nil)
...
 	if cfg := f.Options.Get(archiveNamespace); cfg != nil {
-		cfg.TLS.Close()
+		errs = append(errs, cfg.TLS.Close())
+	}
+	errs = append(errs, f.Options.GetPrimary().TLS.Close())
+	return errors.Join(errs...)
+}
+
+func (f *Factory) onPrimaryPasswordChange() {
+	primaryClient, err := f.newClientFn(f.primaryConfig, f.logger, f.metricsFactory)
```
yurishkuro, line 275: You are using the old config here, not the new password. This needs a unit test

plugin/storage/es/factory.go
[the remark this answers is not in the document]
haanhvu: Ah, I misunderstood your question. I thought you meant should we use atomic.Pointer or atomic.Value^^ Anyway there's a lint failure regarding passing atomic.Pointer into function: https://github.com/jaegertracing/jaeger/pull/4342/commits/47a7cad6eae97d9900493c85b2062cd4fcaaad17#diff-bd7140c4b097bbb7e2462f2b512528d3f45af2317fe008211c293553bff71c56L111

I would like to resolve this (and your remaining reviews) before adding tests.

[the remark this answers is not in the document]
haanhvu: I fixed the lint failures. I'll push it soon.

plugin/storage/es/factory.go
haanhvu: The current tests and linters are fine now so I'm adding new tests now.

But could you help explain why password is not updated here? I thought when we recall f.newClientFn(f.primaryConfig, ...) then we'd recall [getConfigOptions](https://github.com/jaegertracing/jaeger/blob/main/pkg/es/config/config.go#L100) and get the latest password from password file path in config?

yurishkuro: This is why when you develop new functionality you often start with a unit test that verifies what you're trying to implement. In this case we want to observe that a new client is created once pwd file is changed - if you had that test you would've caught several bugs I already pointed out.

haanhvu: Yeah sorry I'll add that test now^^ I have exams until Thursday so maybe I'll push the test after that.

yurishkuro: no need to apologize, your contributions are greatly appreciated.

plugin/storage/es/factory_test.go
```diff
@@ -238,4 +237,48 @@ func TestInitFromOptions(t *testing.T) {
...
+	primaryClient, err := f.newClientFn(f.primaryConfig.Load(), nil, nil)
+	require.NoError(t, err)
+
+	var expectedPrimaryClient atomic.Pointer[es.Client]
+	expectedPrimaryClient.Store(&primaryClient)
+	assert.Equal(t, expectedPrimaryClient.Load(), f.primaryClient.Load())
+
+	_, err = passwordFile.WriteString("baz")
+	require.NoError(t, err)
+	f.onPrimaryPasswordChange()
```
haanhvu, line 282: I don't know why but I have to explicitly call f.onPrimaryPasswordChange() here to make the password change. Don't know if it's a problem of the watcher in the implementation or a problem in my test.

haanhvu (commented):
@yurishkuro I added a test for password change in primary config.

Also, in the implementation, there was a race detected so I changed config type to atomic.

Please help review! The test passed but there're probably some problems.

plugin/storage/es/factory_test.go
```diff
@@ -238,4 +237,48 @@ func TestInitFromOptions(t *testing.T) {
...
+
+	passwordFile.WriteString("bar")
+
+	c := escfg.Configuration{
+		Username:         "foo",
+		PasswordFilePath: passwordFile.Name(),
+	}
+
+	f := NewFactory()
+	f.newClientFn = func(c *escfg.Configuration, logger *zap.Logger, metricsFactory metrics.Factory) (es.Client, error) {
```
yurishkuro, line 255: why do you need this function? It looks like it's repeating the same code as in the factory.

You need to verify that the es.Client is updated once the password is modified. The easiest way is to do `factory.getPrimaryClient` and check its authentication settings, assuming they are accessible. If not accessible, the other option is to create a fake server and invoke a method on the client that would cause it to make an HTTP call. We don't care if the call succeeds or not, but the fake server will be able to inspect the auth headers in that call, which is the ultimate validation.

plugin/storage/es/factory_test.go
yurishkuro: Also, because watcher is running in a separate goroutine, you cannot immediately validate if the client was updated, usually we use assert.Eventually (grep for examples in this repo)

pkg/es/config/config.go
```diff
@@ -412,7 +424,7 @@ func GetHTTPRoundTripper(c *Configuration, logger *zap.Logger) (http.RoundTrippe
 	return transport, nil
 }
 
-func loadToken(path string) (string, error) {
+func LoadFileContent(path string) (string, error) {
```
yurishkuro, line 427: TODO this should probably remain being called LoadToken, since it's not a plain file read

PR #4351, opened 2023-03-29
Add option to configure max received message size for jaeger-query

### LINKED ISSUE #4350: [Bug]: HTTP Error: grpc stream error: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (7833328 vs. 4194304)
### What happened?

As a user with large traces, I want to configure the grpc plugin max receive message length so that jaeger can display large traces (e.g. with more than 10000 spans).

### Steps to reproduce

1. Send a large trace response from a grpc storage plugin
2. The server returns the error

### Expected behavior

The server should allow increasing the maximum message size in parity to collector

### Relevant log output

```shell
{
  "level": "error",
  "ts": 1680074638.9349482,
  "caller": "app/http_handler.go:495",
  "msg": "HTTP handler, Internal Server Error",
  "error": "grpc stream error: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (7833328 vs. 4194304)",
  "stacktrace": "github.com/jaegertracing/jaeger/cmd/query/app.(*APIHandler).handleError\n\tgithub.com/jaegertracing/jaeger/cmd/query/app/http_handler.go:495\ngithub.com/jaegertracing/jaeger/cmd/query/app.(*APIHandler).getTrace\n\tgithub.com/jaegertracing/jaeger/cmd/query/app/http_handler.go:437\nnet/http.HandlerFunc.ServeHTTP\n\tnet/http/server.go:2110\ngithub.com/opentracing-contrib/go-stdlib/nethttp.MiddlewareFunc.func5\n\tgithub.com/opentracing-contrib/go-stdlib@v1.0.0/nethttp/server.go:154\nnet/http.HandlerFunc.ServeHTTP\n\tnet/http/server.go:2110\nnet/http.HandlerFunc.ServeHTTP\n\tnet/http/server.go:2110\ngithub.com/gorilla/mux.(*Router).ServeHTTP\n\tgithub.com/gorilla/mux@v1.8.0/mux.go:210\ngithub.com/jaegertracing/jaeger/cmd/query/app.additionalHeadersHandler.func1\n\tgithub.com/jaegertracing/jaeger/cmd/query/app/additional_headers_handler.go:28\nnet/http.HandlerFunc.ServeHTTP\n\tnet/http/server.go:2110\ngithub.com/gorilla/handlers.CompressHandlerLevel.func1\n\tgithub.com/gorilla/handlers@v1.5.1/compress.go:100\nnet/http.HandlerFunc.ServeHTTP\n\tnet/http/server.go:2110\ngithub.com/gorilla/handlers.recoveryHandler.ServeHTTP\n\tgithub.com/gorilla/handlers@v1.5.1/recovery.go:78\nnet/http.serverHandler.ServeHTTP\n\tnet/http/server.go:2951\nnet/http.(*conn).serve\n\tnet/http/server.go:1992"
}
```

### Screenshot

image

### Additional context

_No response_

### Jaeger backend version

v1.42

### SDK

custom storage plugin

### Pipeline

In a nutshell, a custom storage plugin that mutates the trace.

See https://github.com/kubewharf/kelemetry/blob/14752b56543603fe3c84bc817cb437ede51fe1d6/docs/DEPLOY.md for full pipeline, but it is mostly irrelevant.

### Stogage backend

custom storage backend

### Operating system

Linux

### Deployment model

Kubernetes

### Deployment configs

```shell
N/A
```

SOF3:
Currently this option only exists for collector but not for query.

This option is required when a storage returns a very large grpc message (>4MiB).

- Resolves #4350

- Add a `--grpc-storage.max-recv-msg-size` option.

### DIFF
--- a/plugin/storage/grpc/options.go
+++ b/plugin/storage/grpc/options.go
@@ -33,6 +33,7 @@ const (
 	remotePrefix             = "grpc-storage"
 	remoteServer             = remotePrefix + ".server"
 	remoteConnectionTimeout  = remotePrefix + ".connection-timeout"
+	grpcMaxRecvMsgSize       = remotePrefix + ".max-recv-msg-size"
 	defaultPluginLogLevel    = "warn"
 	defaultConnectionTimeout = time.Duration(5 * time.Second)
 )
@@ -58,6 +59,7 @@ func (opt *Options) AddFlags(flagSet *flag.FlagSet) {
 	flagSet.String(pluginLogLevel, defaultPluginLogLevel, "Set the log level of the plugin's logger")
 	flagSet.String(remoteServer, "", "The remote storage gRPC server address as host:port")
 	flagSet.Duration(remoteConnectionTimeout, defaultConnectionTimeout, "The remote storage gRPC server connection timeout")
+	flagSet.Int(grpcMaxRecvMsgSize, 4<<20, "The maximum receivable message size to receive from the remote storage gRPC server")
 }
 
 // InitFromViper initializes Options with properties from viper
@@ -73,5 +75,6 @@ func (opt *Options) InitFromViper(v *viper.Viper) error {
 	}
 	opt.Configuration.RemoteConnectTimeout = v.GetDuration(remoteConnectionTimeout)
 	opt.Configuration.TenancyOpts = tenancy.InitFromViper(v)
+	opt.Configuration.MaxRecvMsgSize = v.GetInt(grpcMaxRecvMsgSize)
 	return nil
 }
--- a/plugin/storage/grpc/config/config.go
+++ b/plugin/storage/grpc/config/config.go
@@ -46,6 +46,7 @@ type Configuration struct {
 	RemoteTLS               tlscfg.Options
 	RemoteConnectTimeout    time.Duration `yaml:"connection-timeout" mapstructure:"connection-timeout"`
 	TenancyOpts             tenancy.Options
+	MaxRecvMsgSize          int `yaml:"max-receive-message-size" mapstructure:"max-receive-message-size"`
 
 	pluginHealthCheck     *time.Ticker
 	pluginHealthCheckDone chan bool
@@ -84,6 +85,7 @@ func (c *Configuration) Close() error {
 
 func (c *Configuration) buildRemote(logger *zap.Logger) (*ClientPluginServices, error) {
 	opts := []grpc.DialOption{
+		grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(c.MaxRecvMsgSize)),
 		grpc.WithUnaryInterceptor(otgrpc.OpenTracingClientInterceptor(opentracing.GlobalTracer())),
 		grpc.WithStreamInterceptor(otgrpc.OpenTracingStreamClientInterceptor(opentracing.GlobalTracer())),
 		grpc.WithBlock(),


ChillOrb:
@SOF3 What if someone wants to set it as -1 (no limit) [ref](https://grpc.github.io/grpc/core/group__grpc__arg__keys.html#ga813f94f9ac3174571dd712c96cdbbdc1)?

SOF3:
@ChillOrb the argument syntax is copied from the equivalent option in collector `--collector.grpc-server.max-message-size`.

ChillOrb:
Looking at this a bit more, I found that currently grpc-go only supports `MaxRecvMsgSize` and not something like `MaxRecvMsgLength`(which is a valid option [link](https://grpc.github.io/grpc/core/group__grpc__arg__keys.html#ga813f94f9ac3174571dd712c96cdbbdc1)) so for [link](https://pkg.go.dev/google.golang.org/grpc#MaxRecvMsgSize) and condition for > 0 makes sense. 

As for the following variable name`flagSuffixGRPCMaxReceiveMessageLength` which is accepted from `"max-message-size"` [link](https://github.com/jaegertracing/jaeger/blob/bc8e1d16524dcbdac80683ebc007e22cac8406c7/cmd/collector/app/flags/flags.go#L45)  should instead be called `flagSuffixGRPCMaxReceiveMessageSize` since `MaxMessageLength` has a different meaning in grpc terms and can be set as -1 and thus should not have >0 condition. 

@yurishkuro What do you think? 

> @SOF3 What if someone wants to set it as -1 (no limit) [ref](https://grpc.github.io/grpc/core/group__grpc__arg__keys.html#ga813f94f9ac3174571dd712c96cdbbdc1)?

yurishkuro:
Go API calls is "size", which is what we call it too. "length" is from some other documentation.

yurishkuro:
any follow-up?

cmd/query/app/server.go
```diff
@@ -110,6 +110,9 @@ func (s Server) HealthCheckStatus() chan healthcheck.Status {
 func createGRPCServer(querySvc *querysvc.QueryService, metricsQuerySvc querysvc.MetricsQueryService, options *QueryOptions, tm *tenancy.Manager, logger *zap.Logger, tracer opentracing.Tracer) (*grpc.Server, error) {
 	var grpcOpts []grpc.ServerOption
 
+	if options.GRPCMaxReceiveMessageSize > 0 {
+		grpcOpts = append(grpcOpts, grpc.MaxRecvMsgSize(options.GRPCMaxReceiveMessageSize))
```
yurishkuro, line 114: I don't think this is the right place. This is a [setting for gRPC server][1] that runs as part of jaeger-query. That server allows people to download traces, but it does not _receive_ any large payloads.

The original ticket mentioned gRPC plugin, i.e. where the path is `client <- jaeger-query/gRPC <- grpc-storage-plugin <- remote storage`, and it's in the last hop that throws the exception because the returned message is too large.

[1]: https://pkg.go.dev/google.golang.org/grpc#MaxRecvMsgSize

SOF3: In the original issue, the last hop uses the HTTP/JSON frontend, so there is no gRPC involved there.

The path was

```mermaid
client -->|HTTP/JSON| "jaeger-query" -->|gRPC| "storage-plugin" --> "blackbox out of question"
```

yurishkuro: In this workflow the proposed change will have no effect whatsoever, because the server whose setting is being changed is not even in the picture. It would be if the client was calling via gRPC, but even then it would have no effect on plugin-blackbox interaction, which is where the error was raised.

SOF3: @yurishkuro No, the error was raised on the layer between jaeger-query and storage-plugin. We are not even using any grpc behind the storage plugin and blackbox, and we aren't even exposing the grpc query port. I have verified that this patch has fixed the issue, at least for the case I reproduced.

SOF3: Well, the error came from query-plugin interaction not plugin-blackbox interaction. But you're right, the changed code affects client-query instead of query-plugin. I wonder how this somehow fixed the issue in my test...

SOF3:
Rewrote the commit, changing the grpc storage instad of the query command.

plugin/storage/grpc/options.go
```diff
@@ -73,5 +75,6 @@ func (opt *Options) InitFromViper(v *viper.Viper) error {
 	}
 	opt.Configuration.RemoteConnectTimeout = v.GetDuration(remoteConnectionTimeout)
 	opt.Configuration.TenancyOpts = tenancy.InitFromViper(v)
+	opt.Configuration.MaxRecvMsgSize = v.GetInt(grpcMaxRecvMsgSize)
```
yurishkuro, line 78: please add to tests

plugin/storage/grpc/options.go
```diff
@@ -33,6 +33,7 @@ const (
 	remotePrefix             = "grpc-storage"
 	remoteServer             = remotePrefix + ".server"
 	remoteConnectionTimeout  = remotePrefix + ".connection-timeout"
+	grpcMaxRecvMsgSize       = remotePrefix + ".max-recv-msg-size"
```
yurishkuro, line 36:
```suggestion
	grpcMaxRecvMsgSize       = remotePrefix + ".max-message-size"
```
to be consistent with `cmd/collector/app/flags/flags.go:45`

SOF3: Would this be ambiguous that it also affects the sent message size? (although I don't see any chance a storage plugin client would send a large request anyway).

yurishkuro: why not apply the same setting both ways?

SOF3: Well basically I just didn't see the motivation for configuring the send size.

yurishkuro: the motivation is using the same param and not introducing a new name that does not match existing flags

yurishkuro:
old PR, not solving the issue

PR #4371, opened 2023-04-06
Fix typo in readme of hotrod example

ruthetum:
- Invalid docker run script about hotrod example

- Fixes a minor typo in readme of hotrod example


yurishkuro (approved):
Thx

PR #4382, opened 2023-04-13
Allow follows-from reference as a parent span id

### LINKED ISSUE #4380: [Bug]: non-root spans with FollowsFrom reference incorrectly represent as trace in collector metrics
### What happened?

We noticed that non-root spans with **FollowsFrom** reference are showing as traces in collector metrics, e.g. `traces.received`, `traces.rejected` metrics.

The reason is that in collector metrics, `ParentSpanID()` is used to determine if a span is a root span, e.g. below
https://github.com/jaegertracing/jaeger/blob/3d67ac86f53866332568b0f542265b00d86407fc/cmd/collector/app/metrics.go#L233-L237

And the current implementation of `ParentSpanID()` only considers **ChildOf** relationship. So if a span has only **FollowsFrom** reference, the `ParentSpanID()` doesn't consider that and returns 0.
https://github.com/jaegertracing/jaeger/blob/506adaf2bf4f8364aa21d3f3001d1a6dcc6e2e97/model/span.go#L99-L109

Not sure any particular reason that **FollowsFrom**  is not considered in the current implementation of `ParentSpanID()`.

The proposal is to either 1) fix in collector metric to replace the `ParentSpanID()`, 2) fix in `ParentSpanID()`.
Prefer 2) because we use this method a lot to determine a span is a root span internally, but changing its behavior might break other places in jaeger OSS.

### Steps to reproduce

Create spans with **FollowsFrom** relationship, and check the collector trace metrics.

### Expected behavior

non-root spans with FollowsFrom reference should not show as trace in collector metrics.

### Relevant log output

_No response_

### Screenshot

_No response_

### Additional context

_No response_

### Jaeger backend version

_No response_

### SDK

_No response_

### Pipeline

_No response_

### Stogage backend

_No response_

### Operating system

_No response_

### Deployment model

_No response_

### Deployment configs

_No response_

kubarydz:
Resolves #4380

handle FOLLOWS_FROM as a parent span ID


model/span.go
```diff
@@ -101,7 +101,7 @@ func (s *Span) NormalizeTimestamps() {
 func (s *Span) ParentSpanID() SpanID {
 	for i := range s.References {
 		ref := &s.References[i]
-		if ref.TraceID == s.TraceID && ref.RefType == ChildOf {
+		if ref.TraceID == s.TraceID && (ref.RefType == ChildOf || ref.RefType == FollowsFrom) {
```
yurishkuro, line 104: if no unit tests break I think it's reasonable. But we always treated ChildOf as higher priority, so a better fix would be (a) search for ChildOf ref, (b) if not found, search for FollowsFrom() (this can be optimized to be done in one loop)

kubarydz: that makes sense, I will make adjustments

kubarydz:
right, sorry about that, should be ok now

yurishkuro:
how would this change affect clock skew?  https://github.com/jaegertracing/jaeger/issues/448

kubarydz:
Well they would not be considered root spans, which they are now. Tbh now I'm not sure anymore if a span with follows-from reference but no child-of reference should be considered a root span.

yurishkuro:
Conceptually the parentSpanID field represents an ancestor span, that _happened-before_ the current span and is the _cause_ of the current span. In that regard there is no strong difference between child-of or follows-from reference types, they actually tell more about the parent than about the child. I thought about this when OpenTelemetry model was designed and I was ok with only having a parent-span-id in OTEL, without explicit references like in OpenTracing. However, there is a difference between them for the clock alignment purpose. Namely, and naively, that the child-of span is supposed to finish before the parent is finished, which means we have both lower and upper boundary of how far the child span's timestamps can be moved. The same restriction does not exist with follows-from span because it can legitimately finished after the parent. But the reason I say "naively" is because even with child-of span there is actually no guarantee that it finishes before the parent. All we know is that child-of span _sends a response to the parent_ before parent finishes, whereas finishing the child span may happen any time  after response is sent (sometimes long after, due to async behavior). Which means that even for child-of spans the clock alignment can realistically only be done on their startTime.

Long story short, this just highlights the fact that our clock alignment algorithm is pretty simplistic at the moment. But even that simple algorithm could be broken if we treat the two ref types identically (namely, as child-of spans are treated), which is what I think will happen with this change because unfortunately the algorithm uses parentSpanID field instead of inspecting the references explicitly. Hence my concern with this change.

kubarydz:
Thanks for the exhaustive answer.
It is now my understanding that follows-from can be used to determine parentSpanID everywhere except in the clock skew adjuster. If that's ok I would propose additional change in building sub graphs:
https://github.com/jaegertracing/jaeger/blob/5d6a2e88f372806c29847a46b3735e2385e70d55/model/adjuster/clockskew.go#L123-L126
specifically here, since adjuster just wants to know if it should adjust the clock. So we could just ask the span if it should be adjusted (`isChild()`,  `shouldBeClockAdjusted()`, `isClockskewAdjustable()` or something similar). Then the logic of clockskew adjuster would not be dependent on how we determine if span is root span.
I might still be missing something, but it seems to me like clockskew adjuster has a different idea of what a root span is, which creates a problem.

yurishkuro (approved):

yurishkuro:
I had a second look at the current logic in clockSkew adjuster was incorrect anyway, so let's leave it

PR #4412, opened 2023-04-25
enable gocritic linter

mmorel-35:
- enable gocritic linter
- disabled  dupArg, exitAfterDefer and appendAssign rules of gocritic until they are addressed. 

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -3,16 +3,12 @@ issues:
   exclude-rules:
     # Exclude some linters from running on tests files.
     - path: _test\.go
-      linters:
-        - gosec
-    - linters:
-        - gocritic
+      linters: [gosec]
+    - linters: [gocritic]
       text: "dupArg"
-    - linters:
-        - gocritic
+    - linters: [gocritic]
       text: "exitAfterDefer"
-    - linters:
-        - gocritic
+    - linters: [gocritic]
       text: "appendAssign"
   max-issues-per-linter: 0
   max-same-issues: 0
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,12 +1,35 @@
-run:
-  go: '1.20'
-  timeout: 10m
-  skip-dirs:
-    - mocks
-    - thrift-0.9.2
-    - .*-gen
-  skip-files:
-    - ".*.pb.go$"
+issues:
+  # Excluding configuration per-path, per-linter, per-text and per-source
+  exclude-rules:
+    # Exclude some linters from running on tests files.
+    - path: _test\.go
+      linters:
+        - gosec
+    - linters:
+        - gocritic
+      text: "dupArg"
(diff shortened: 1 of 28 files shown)


pkg/metrics/metrics.go
```diff
@@ -93,41 +94,42 @@ func Init(m interface{}, factory Factory, globalTags map[string]string) error {
...
+			default:
 				return fmt.Errorf(
 					"Field [%s]: Buckets should only be defined for Timer and Histogram metric types",
 					field.Name)
 			}
 		}
 		help := field.Tag.Get("help")
 		var obj interface{}
-		if field.Type.AssignableTo(counterPtrType) {
+		switch {
```
albertteoh, line 105: I tend to prefer the `switch` statement instead of if/else if chains because it's a bit cleaner and easier to read, but others maintainers may disagree.

albertteoh (approved):
Some nice improvements, thanks!

pkg/metrics/metrics.go
mmorel-35: As I started to code with Java I used to be one of those. But different language, different best practices and switch case is one of those differences. Gocritic is widely used which means that the golang community follow it’s rules. If the majority of the developers of this project don’t want to conform to this rule it can be excuded in the golangci-lint config file. 
By the way, I’m not aware of any golang linter forbidding switch cases let me know if you do ;)

.golangci.yml
```diff
@@ -1,12 +1,35 @@
...
-    - .*-gen
-  skip-files:
-    - ".*.pb.go$"
+issues:
+  # Excluding configuration per-path, per-linter, per-text and per-source
+  exclude-rules:
+    # Exclude some linters from running on tests files.
+    - path: _test\.go
+      linters:
+        - gosec
```
yurishkuro, line 7: could this be written as `linters: [gosec]`? It's much easier to read

cmd/agent/app/testutils/in_memory_reporter.go
```diff
@@ -58,12 +58,12 @@ func (i *InMemoryReporter) EmitBatch(_ context.Context, batch *jaeger.Batch) (er
 func (i *InMemoryReporter) ZipkinSpans() []*zipkincore.Span {
 	i.mutex.Lock()
 	defer i.mutex.Unlock()
-	return i.zSpans[:]
+	return i.zSpans
```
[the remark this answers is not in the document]
mmorel-35: Please see this rule https://go-critic.com/overview.html#unslice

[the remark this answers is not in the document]
yurishkuro: confirmed https://go.dev/play/p/ZlBx7iZwkW6

yurishkuro (approved):

PR #4414, opened 2023-04-26
feat: implement gRPC storage plugin for metrics

### LINKED ISSUE #4413: [Feature]: gRPC storage plugin for metrics
### Requirement

As a Jaeger operator, I want to enable the Jaeger UI "Monitoring" tab for my SREs, DevOps, Engineers.

### Problem

[Service Performance Monitoring](https://www.jaegertracing.io/docs/1.44/spm/) is a feature of Jaeger that is often overlooked.

My team can't use it because our metrics backend is InfluxDB. (InfluxDB is also our tracing backend.) The only Jaeger metrics plugin requires Prometheus.

### Proposal

I offer to write a new `metricsstore` plugin, which backends to a remote (not sidecar) third-party gRPC service.

At the same time, I offer to write a third-party remote gRPC backend for the proposed plugin. This backend would support InfluxDB, and would live here:
https://github.com/influxdata/influxdb-observability/tree/main/jaeger-influxdb

### Open questions

[Several questions have been raised and discussed.](https://github.com/jaegertracing/jaeger/discussions/4369)

Where should the new plugin live? I propose that `plugin/storage/grpc/shared/interface.go` change like this (I hope the rest is obvious):

```diff
  type StoragePlugin interface {
    SpanReader()       spanstore.Reader
    SpanWriter()       spanstore.Writer
    DependencyReader() dependencystore.Reader
+   MetricsReader()    metricsstore.Reader
  }
...
  type Capabilities struct {
    ArchiveSpanReader   bool
    ArchiveSpanWriter   bool
    StreamingSpanWriter bool
+   MetricsReader       bool
  }
```

jacobmarble:
Resolves #4413

- define the metrics gRPC API in protos
- refactor: move two subpackages of plugin/storage/grpc to pkg/grpc
- implement the gRPC metrics plugin

A POC is implemented here: https://github.com/influxdata/influxdb-observability/pull/239

## TODO
- [ ] initial review by maintainers
- [ ] add unit tests (new _test.go files with TODO comments)

### DIFF
--- a/plugin/metrics/factory.go
+++ b/plugin/metrics/factory.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/jaegertracing/jaeger/plugin"
 	"github.com/jaegertracing/jaeger/plugin/metrics/disabled"
+	"github.com/jaegertracing/jaeger/plugin/metrics/grpc"
 	"github.com/jaegertracing/jaeger/plugin/metrics/prometheus"
 	"github.com/jaegertracing/jaeger/storage"
 	"github.com/jaegertracing/jaeger/storage/metricsstore"
@@ -33,6 +34,7 @@ const (
 	disabledStorageType = ""
 
 	prometheusStorageType = "prometheus"
+	grpcStorageType       = "grpc-plugin"
 )
 
 // AllStorageTypes defines all available storage backends.
@@ -67,6 +69,8 @@ func (f *Factory) getFactoryOfType(factoryType string) (storage.MetricsFactory, 
 	switch factoryType {
 	case prometheusStorageType:
 		return prometheus.NewFactory(), nil
+	case grpcStorageType:
+		return grpc.NewFactory(), nil
 	case disabledStorageType:
 		return disabled.NewFactory(), nil
 	}
@@ -76,7 +80,9 @@ func (f *Factory) getFactoryOfType(factoryType string) (storage.MetricsFactory, 
 // Initialize implements storage.MetricsFactory.
 func (f *Factory) Initialize(logger *zap.Logger) error {
 	for _, factory := range f.factories {
-		factory.Initialize(logger)
+		if err := factory.Initialize(logger); err != nil {
+			return err
+		}
 	}
 	return nil
 }
--- a/plugin/metrics/grpc/options_test.go
+++ b/plugin/metrics/grpc/options_test.go
@@ -1,0 +1,17 @@
+// Copyright (c) 2023 The Jaeger Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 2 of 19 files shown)


pkg/grpc/shared/metrics.go
jacobmarble, line 1: TODO write tests

pkg/grpc/grpc.go
```diff
@@ -0,0 +1,47 @@
+// Copyright (c) 2019 The Jaeger Authors.
```
yurishkuro, line 1: why did you need to move `grpc` from storage? `pkg` is our general-purpose modules, this does not really fit here because it's coupled with business-specific storage APIs

jacobmarble: I anticipated that you would request these sub-packages of `plugin/storage/grpc` be moved to `pkg/grpc` because both `plugin/storage/grpc` and `plugin/metrics/grpc` depend on them. I'll move them back.

jacobmarble: Rather than making the PR more complicated with an additional file-move PR, I've removed the commit that moved `plugin/storage/grpc/config` and `plugin/storage/grpc/shared` to `pkg/grpc`.

plugin/metrics/factory.go
```diff
@@ -33,6 +34,7 @@ const (
 	disabledStorageType = ""
 
 	prometheusStorageType = "prometheus"
+	grpcStorageType       = "grpc-plugin"
```
yurishkuro, line 37: So we have `--metrics-storage-type=grpc-plugin` but then factory flags under `--grpc-metrics.*`? I can't think if a better name, but at least we should keep these in sync, so let's go with `grpc-metrics`

plugin/metrics/factory.go
```diff
@@ -67,6 +69,8 @@ func (f *Factory) getFactoryOfType(factoryType string) (storage.MetricsFactory,
 	switch factoryType {
 	case prometheusStorageType:
 		return prometheus.NewFactory(), nil
+	case grpcStorageType:
+		return grpc.NewFactory(), nil
```
yurishkuro, line 73: could we add a test?

plugin/metrics/grpc/options_test.go
```diff
@@ -0,0 +1,17 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package grpc
+
+// TODO write tests
```
yurishkuro, line 17: add tests?

yurishkuro (commented):
It's a bit hard to grok from the changes, but what does this mean in practice that a MetricsReader service is added to the main storage plugin? Since it's controlled by capabilities API, it means existing storages don't have to implement it, which is good. But I am also interested in the reverse - what if we I am using one of standard backends for span storage and want grpc-remote implementation for metrics store? Is that possible?

PR #4444, opened 2023-05-07
Update README.md - fixing docker compose command

tqi-raurora:
The correct command is "docker compose up" and not "docker-compose up"
Reference: https://docs.docker.com/engine/reference/commandline/compose/


yurishkuro:
thanks. `docker-compose` is also correct, but for older versions of docker.

PR #4449, opened 2023-05-09
[Jaeger-Agent] Use RawConn.Control to get fd instead of Fd()

### LINKED ISSUE #4448: [Bug]: Jaeger agent gets stuck when closing with SocketBufferSize set
#### Problem
With `SocketBufferSize` set in config, the jaeger-agent gets stuck when closing.
https://github.com/jaegertracing/jaeger/blob/5c3c20dcd309afca924174bece0ad16671105279/cmd/agent/app/builder.go#L90

#### Root cause
This is because `Close()`of `net.UDPConn` will be blocked if `Fd()` is used to get the file descriptor.

The place to set the `SocketBufferSize` in jaeger-agent is below :
https://github.com/jaegertracing/jaeger/blob/5c3c20dcd309afca924174bece0ad16671105279/cmd/agent/app/servers/thriftudp/socket_buffer.go#L25-L32

And the implementation of `Fd()`:
```
func (f *File) Fd() uintptr {
	if f == nil {
		return ^(uintptr(0))
	}

	// If we put the file descriptor into nonblocking mode,
	// then set it to blocking mode before we return it,
	// because historically we have always returned a descriptor
	// opened in blocking mode. The File will continue to work,
	// but any blocking operation will tie up a thread.
	if f.nonblock {
		f.pfd.SetBlocking()
	}

	return uintptr(f.pfd.Sysfd)
}
```
**The exact same problem has been discussed here: https://github.com/golang/go/issues/29277**

Will create PR for fix.

### Steps to reproduce

1. Set `SocketBufferSize` with any valid number
2. Run agent
3. Stop agent

This issue can also reproduce in unit test:
https://github.com/jaegertracing/jaeger/commit/e25216c40b64cde6d4edd91bf0e2f54d07c64a21

### Expected behavior

Jaeger agent can stop successfully.

### Relevant log output

_No response_

### Screenshot

_No response_

### Additional context

_No response_

### Jaeger backend version

latest

### SDK

without sdk

### Pipeline

_No response_

### Stogage backend

_No response_

### Operating system

Linux, Mac

### Deployment model

_No response_

### Deployment configs

_No response_

ChenX1993:
Resolves #4448

Jaeger agent gets stuck when closing with SocketBufferSize set. This is because `Close()` of `net.UDPConn` will be blocked if `Fd()` is used to get the file descriptor.
Use `RawConn.Control` instead to get fd to set the socket buffer.

Same issue was discussed here: https://github.com/golang/go/issues/29277
The fix refers to here: https://github.com/brucespang/go-tcpinfo/pull/3

### DIFF
--- a/cmd/agent/app/agent_test.go
+++ b/cmd/agent/app/agent_test.go
@@ -184,6 +184,36 @@ func TestStartStopRace(t *testing.T) {
 	t.Fatal("Expecting server exit log")
 }
 
+func TestStartStopWithSocketBufferSet(t *testing.T) {
+	resetDefaultPrometheusRegistry()
+	cfg := Builder{
+		Processors: []ProcessorConfiguration{
+			{
+				Model:    jaegerModel,
+				Protocol: compactProtocol,
+				Workers:  1,
+				Server: ServerConfiguration{
+					HostPort:         "127.0.0.1:0",
+					SocketBufferSize: 10,
+				},
+			},
+		},
+	}
+	mBldr := &metricsbuilder.Builder{HTTPRoute: "/metrics", Backend: "prometheus"}
+	metricsFactory, err := mBldr.CreateMetricsFactory("jaeger")
+	mFactory := fork.New("internal", metrics.NullFactory, metricsFactory)
+	require.NoError(t, err)
+	agent, err := cfg.CreateAgent(fakeCollectorProxy{}, zap.NewNop(), mFactory)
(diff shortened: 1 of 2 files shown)


yurishkuro (commented):
please make sure commits are signed

ChenX1993:
> please make sure commits are signed

Signed

yurishkuro (approved):

PR #4452, opened 2023-05-13
Add support for the spanmetrics connector

### LINKED ISSUE #4345: [Feature/SPM]: Support spanmetrics connector
### Requirement

As a jaeger operator, I want to use the newly introduced [spanmetrics connector](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/connector/spanmetricsconnector) for the following reasons:
- Introduces a number of [improvements](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/connector/spanmetricsconnector#span-to-metrics-processor-to-span-to-metrics-connector) to the spanmetrics processor.
- The [spanmetrics processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/spanmetricsprocessor) is [deprecated](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/19751) and so future enhancements will no longer be added to the processor.
- Avoid being in a situation where the spanmetrics processor is decommissioned preventing me from upgrading my OTEL collector version.

### Problem

The known breaking issues include:
- Renaming the `calls_total` metric to `calls` [1], while also allowing for a namespace prefix [2], so a metric name could resemble: `my_namespace_calls`.
  - \[1\]: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/18677
  - \[2\]: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/18697
- Renaming the `latency` metric to `duration` [3], while also allowing for a namespace prefix [4], so a metric name could resemble: `my_namespace_duration_bucket`.
  - \[3\]: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/19371
  - \[4\]:  https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/18697

### Proposal

Make the metric names configurable.

Perhaps introduce an `spm` parameter namespace where metric names can be configured. e.g. `--spm.calls-metric-name` and `--spm.latency-metric-name`.

This would also require an update to the example provided in `docker-compose/monitor`.

Other suggestions welcome.

### Open questions

_No response_

albertteoh:
Resolves #4345

- Adds support for spanmetrics connector via config parameters, while maintaining backwards compatibility with spanmetrics processor.
- Some quality of life changes added to help contributors with making changes to the SPM feature.

## Testing

- Add test cases for connector case, and rely on existing tests to assert backwards compatibility.
- Manual testing via docker-compose to ensure both the service-level and operation-level metrics are visible in the Monitor tab.


docker-compose/monitor/Makefile
```diff
@@ -0,0 +1,29 @@
+.PHONY: build
+build: clean-jaeger
+	cd ../../ && \
+	make build-all-in-one && \
+	make docker-images-jaeger-backend && \
+	docker tag jaegertracing/all-in-one:latest jaegertracing/all-in-one:dev
+
+.PHONY: run
```
albertteoh, line 8: Intentionally named this `run` instead of `run-connector` to imply it's the default make target to use to bring up the system locally.

docker-compose/monitor/README.md
```diff
@@ -83,6 +85,40 @@ Then navigate to the Monitor tab at http://localhost:16686/monitor to view the R
 
 ![My Service RED Metrics](images/my_service_metrics.png)
 
+## Migrating to Span Metrics Connector
```
albertteoh, line 88: I intend to add something similar to this to the Jaeger documentation after this is approved.

docker-compose/monitor/docker-compose-processor.yml
albertteoh, line 1: This is the same as the original `docker-compose.yml` file, _bar_ an additional `LOG_LEVEL=debug` env var set to help with debugging.

I wanted to have separate "run" configs so it's easy to test backwards compatibility now and for future changes.

plugin/metrics/prometheus/factory.go
```diff
@@ -48,7 +48,7 @@ func (f *Factory) AddFlags(flagSet *flag.FlagSet) {
 // InitFromViper implements plugin.Configurable.
 func (f *Factory) InitFromViper(v *viper.Viper, logger *zap.Logger) {
 	if err := f.options.InitFromViper(v); err != nil {
-		logger.Fatal("Failed to initialize metrics storage factory", zap.Error(err))
+		logger.Panic("Failed to initialize metrics storage factory", zap.Error(err))
```
albertteoh, line 51: Added for testability. `logger.Fatal` does an `os.Exit` which can't be tested.

plugin/metrics/prometheus/metricsstore/dbmodel/to_domain.go
```diff
@@ -23,37 +23,54 @@ import (
...
 
 // toDomainLabels converts Prometheus' representation of metric labels to Jaeger's.
-func toDomainLabels(promLabels model.Metric) []*metrics.Label {
+func (d Translator) toDomainLabels(promLabels model.Metric) []*metrics.Label {
 	labels := make([]*metrics.Label, len(promLabels))
 	j := 0
 	for k, v := range promLabels {
-		labels[j] = &metrics.Label{Name: string(k), Value: string(v)}
+		labelName := string(k)
+		if newLabel, ok := d.labelMap[labelName]; ok {
```
albertteoh, line 70: This is needed because the spanmetrics connector renamed `operation` to `span_name`. I made it a method of a `Translator` struct because it was a little tedious (and awkward) to pass the map through multiple levels of functions.

plugin/metrics/prometheus/metricsstore/reader.go
```diff
@@ -89,16 +96,23 @@ func NewMetricsReader(logger *zap.Logger, cfg config.Configuration) (*MetricsRea
...
+		operationLabel:    cfg.OperationLabel,
 	}
+
 	logger.Info("Prometheus reader initialized", zap.String("addr", cfg.ServerURL))
 	return mr, nil
 }
 
 // GetLatencies gets the latency metrics for the given set of latency query parameters.
-func (m *MetricsReader) GetLatencies(ctx context.Context, requestParams *metricsstore.LatenciesQueryParameters) (*metrics.MetricFamily, error) {
+func (m MetricsReader) GetLatencies(ctx context.Context, requestParams *metricsstore.LatenciesQueryParameters) (*metrics.MetricFamily, error) {
```
albertteoh, line 115: MetricsReader's state doesn't change from external packages (and probably not at all) during its lifetime; so I thought to make it read-only.

plugin/metrics/prometheus/metricsstore/reader.go
```diff
@@ -179,29 +224,28 @@ func (m *MetricsReader) executeQuery(ctx context.Context, p metricsQueryParams)
...
 		return &metrics.MetricFamily{}, fmt.Errorf("failed executing metrics query: %w", err)
 	}
 	if len(warnings) > 0 {
-		m.logger.Warn("Warnings detected on Prometheus query", zap.Any("warnings", warnings))
+		m.logger.Warn("Warnings detected on Prometheus query", zap.Any("warnings", warnings), zap.String("query", promQuery), zap.Any("range", queryRange))
 	}
 
-	m.logger.Debug("Prometheus query results", zap.String("results", mv.String()))
-	return dbmodel.ToDomainMetricsFamily(
+	m.logger.Debug("Prometheus query results", zap.String("results", mv.String()), zap.String("query", promQuery), zap.Any("range", queryRange))
```
albertteoh, line 236: It was hard to correlate the request with the response, so I just decided to move the query string in the response, which also helps reduce noise in logs.

plugin/metrics/prometheus/options.go
```diff
@@ -63,9 +81,29 @@ func NewOptions(primaryNamespace string) *Options {
 // AddFlags from this storage to the CLI.
 func (opt *Options) AddFlags(flagSet *flag.FlagSet) {
 	nsConfig := &opt.Primary
-	flagSet.String(nsConfig.namespace+suffixServerURL, defaultServerURL, "The Prometheus server's URL, must include the protocol scheme e.g. http://localhost:9090")
-	flagSet.Duration(nsConfig.namespace+suffixConnectTimeout, defaultConnectTimeout, "The period to wait for a connection to Prometheus when executing queries.")
-	flagSet.String(nsConfig.namespace+suffixTokenFilePath, defaultTokenFilePath, "The path to a file containing the bearer token which will be included when executing queries against the Prometheus API.")
+	flagSet.String(nsConfig.namespace+suffixServerURL, defaultServerURL,
```
albertteoh, line 84: I considered adding a single "OTEL component type" (value one of: "processor" or "connector") flag which hardcodes all the breaking changes to the component type, instead of all these new flags, but there were some drawbacks that made me avoid this path:

- Couples Jaeger too closely to OTEL collector.
- If the spanmetrics connector decides to change a metric name, it would involve a code change in Jaeger. Having separate config options means users can just update their config in this scenario.

Let me know if you disagree.

plugin/metrics/prometheus/options.go
```diff
@@ -75,6 +113,19 @@ func (opt *Options) InitFromViper(v *viper.Viper) error {
 	cfg.ServerURL = stripWhiteSpace(v.GetString(cfg.namespace + suffixServerURL))
 	cfg.ConnectTimeout = v.GetDuration(cfg.namespace + suffixConnectTimeout)
 	cfg.TokenFilePath = v.GetString(cfg.namespace + suffixTokenFilePath)
+	cfg.MetricNamespace = v.GetString(cfg.namespace + suffixMetricNamespace)
+	cfg.CallsMetricName = v.GetString(cfg.namespace + suffixCallsMetricName)
+	cfg.LatencyMetricName = v.GetString(cfg.namespace + suffixLatencyMetricName)
+	cfg.LatencyUnit = v.GetString(cfg.namespace + suffixLatencyUnit)
+	cfg.OperationLabel = v.GetString(cfg.namespace + suffixOperationLabel)
+
+	if v.IsSet(cfg.namespace + suffixLatencyUnit) {
```
albertteoh, line 122: I'm not sure if this is the best place to perform config validation; let me know if it should be done somewhere else.

docker-compose/monitor/README.md
```diff
@@ -26,11 +26,16 @@ The following diagram illustrates the relationship between these components:
 
 # Getting Started
```
yurishkuro, line 27: Using or developing?

docker-compose/monitor/README.md
yurishkuro: I would instead change the docs to only describe connector and point to readme for migration.

docker-compose/monitor/README.md
```diff
@@ -83,6 +85,40 @@ Then navigate to the Monitor tab at http://localhost:16686/monitor to view the R
 
 ![My Service RED Metrics](images/my_service_metrics.png)
 
+## Migrating to Span Metrics Connector 
+
+### Background
+
+A new [Connector](https://pkg.go.dev/go.opentelemetry.io/collector/connector#section-readme) component was introduced
```
yurishkuro, line 92:
```suggestion
A new [Connector](https://pkg.go.dev/go.opentelemetry.io/collector/connector#section-readme) API was introduced
```

docker-compose/monitor/README.md
```diff
@@ -83,6 +85,40 @@ Then navigate to the Monitor tab at http://localhost:16686/monitor to view the R
...
+
+Assuming the OpenTelemetry Collector is running with the [Span Metrics Connector][spanmetricsconnector] correctly
+configured, the following configuration should be applied to jaeger-query or jaeger-all-in-one:
+
+As command line parameters:
+```shell
+--prometheus.query.namespace=span_metrics
+--prometheus.query.duration-metric-name=duration
+--prometheus.query.duration-unit=ms
+--prometheus.query.span-name-label=span_name
```
yurishkuro, line 111: I'm skeptical about introducing this level of flexibility, is it really that necessary? It adds both code complexity and mental complexity for the users. To support connector we could just expose a single flag and deal with the renaming internally.

docker-compose/monitor/docker-compose-connector.yml
```diff
@@ -3,21 +3,26 @@ services:
   jaeger:
     networks:
       - backend
-    image: jaegertracing/all-in-one:latest
+    image: jaegertracing/all-in-one:dev
```
yurishkuro, line 6: Maybe parameterize to allow running without rebuilding.

albertteoh: Okay, I've parameterised this, which is set in the Makefile in: [ba0a3d7](https://github.com/jaegertracing/jaeger/pull/4452/commits/ba0a3d759c91ad1ef6e63da110e509c42eaad9ef). 

Let me know if that addresses your comment and, if not, can you please clarify what you mean?

docker-compose/monitor/README.md
albertteoh: Yeah, good point, I ignored the user who just wants to demo it. I've updated the Makefile and README instructions in: [ba0a3d7](https://github.com/jaegertracing/jaeger/pull/4452/commits/ba0a3d759c91ad1ef6e63da110e509c42eaad9ef).

docker-compose/monitor/README.md
albertteoh: Renamed in: [d5752b2](https://github.com/jaegertracing/jaeger/pull/4452/commits/d5752b267c9ba92ddf60657a8973e7ff8ce66801)

docker-compose/monitor/README.md
yurishkuro: 1) I as thinking just a Boolean, default to false for backwards compat, in 2 releases default to true, in next two releases remove altogether.

2) is the dependency on the naming directly in spm or in prom storage impl?

albertteoh: > I as thinking just a Boolean, default to false for backwards compat, in 2 releases default to true, in next two releases remove altogether.

Sounds good. 👍🏼 

> is the dependency on the naming directly in spm or in prom storage impl?

The name changes are in the metrics stored in Prometheys themselves. Example:
- Before (spanmetrics processor): `latency_bucket{service_name="foo", operation="GetCustomer", ...}`
- After (spanmetrics connector): `my_namespace_duration_milliseconds_bucket{service_name="foo", span_name="GetCustomer", ...}`

So main changes I've made are in the Prometheus reader in Jaeger, in how it puts together the Prometheus query.

albertteoh: My initial thinking was that having this boolean parameter under the prometheus namespace doesn't seem quite right because it relates to OTEL and the data contract it has with Jaeger via the SPM feature; which is why I felt having it under an `spm` namespace made more sense.

But given it's going to be removed after 2 releases, maybe it's not worth the extra effort to create a new namespace?

yurishkuro: The reason I asked was that it seems even as Boolean it is scoped to prom implementation, not the top level spm.

albertteoh: Sorry, do you mean you prefer a boolean under the `prometheus` scope? I don't have a strong opinion either way, especially given it will be temporary.

By the way, if we go down this path of a boolean parameter, there are some new parameters that are still necessary to:
- Configure the namespace.
- Configure the unit used for latency histograms.

Those parameters are also configurable in the spanmetrics connector and influence the metric's name/labels. So those will need to remain.

yurishkuro: > do you mean you prefer a boolean under the prometheus scope?

yes

albertteoh: I've added a `prometheus.query.support-spanmetrics-connector` flag in: [e52b74b](https://github.com/jaegertracing/jaeger/pull/4452/commits/e52b74b63478c3a09d00dc98ba003b95972f2b67)

Let me know what you think.

docker-compose/monitor/Makefile
```diff
@@ -0,0 +1,50 @@
+.PHONY: build
+build: clean-jaeger
+	cd ../../ && \
+	make build-all-in-one && \
+	make docker-images-jaeger-backend && \
+	docker tag jaegertracing/all-in-one:latest jaegertracing/all-in-one:dev
```
yurishkuro, line 6: what's the reason for aliasing latest tag to dev tag?

docker-compose/monitor/Makefile
```diff
@@ -0,0 +1,50 @@
...
+run-dev: export JAEGER_IMAGE_TAG = dev
+run-dev: _run-connector
+
+# _run-connector is the base target to bring up the system required for SPM using the new OTEL spanmetrics connector.
+.PHONY: _run-connector
+_run-connector: export OTEL_IMAGE_TAG = latest
+_run-connector: export OTEL_CONFIG_SRC = ./otel-collector-config-connector.yml
+_run-connector: export PROMETHEUS_QUERY_SUPPORT_SPANMETRICS_CONNECTOR = true
+_run-connector: export PROMETHEUS_QUERY_NAMESPACE = span_metrics
+_run-connector: export PROMETHEUS_QUERY_DURATION_UNIT = ms
```
yurishkuro, line 25: when running collector with connector, does it require these args?  If not, can we not make these the default values and only require overriding them when different?

docker-compose/monitor/Makefile
```diff
@@ -0,0 +1,50 @@
...
+_run-connector:
+	docker compose -f docker-compose.yml up
+
+# The run-dev-processor make target is setup to allow devs to run the older spanmetrics processor setup, for example,
```
yurishkuro, line 29:
```suggestion
# run the older spanmetrics processor setup, for example,
```

docker-compose/monitor/README.md
```diff
@@ -42,12 +44,37 @@ docker compose down
...
+
+## Backwards compatibility testing with spanmetrics processor
+
+```bash
+make run-dev-processor
 ```
 
+For each "run" make target, you should expect to see the following in the Monitor tab after a few minutes:
+
+![Monitor Screenshot](https://github.com/jaegertracing/jaeger/assets/26584478/f0fd5d26-c486-4de6-892b-ac48c3e8e425)
```
yurishkuro, line 76: how is this link determined? Can we put it under `images/` instead?

plugin/metrics/prometheus/options.go
```diff
@@ -63,9 +76,24 @@ func NewOptions(primaryNamespace string) *Options {
...
+		"The period to wait for a connection to Prometheus when executing queries.")
+	flagSet.String(nsConfig.namespace+suffixTokenFilePath, defaultTokenFilePath,
+		"The path to a file containing the bearer token which will be included when executing queries against the Prometheus API.")
+
+	flagSet.Bool(nsConfig.namespace+suffixSupportSpanmetricsConnector, defaultSupportSpanmetricsConnector,
+		`Whether if queries should be adjusted for OpenTelemetry Collector's spanmetrics connector.'`)
+	flagSet.String(nsConfig.namespace+suffixMetricNamespace, defaultMetricNamespace,
+		`The metric namespace that is prefixed to the metric name. A '.' separator will be added between `+
+			`the namespace and the metric name.`)
+	flagSet.String(nsConfig.namespace+suffixLatencyUnit, defaultLatencyUnit,
```
yurishkuro, line 91: why not flagSet.Duration?

docker-compose/monitor/Makefile
albertteoh: Ah, I initially aliased the tag because I noticed `docker rmi -f ...:latest` wasn't cleaning things up properly. I was still getting an older version.

However, I tested it again and seems to be working fine now. I _suspect_ because I'd switched over to using `docker system prune -f`.

I like the explicitness of a `latest` (for users that just want to demo SPM) vs `dev` (for users who are developers of SPM) version and this explicitness is mirrored in make targets, which seems nice as well (`make run` vs `make run-dev`).

Thoughts?

docker-compose/monitor/Makefile
albertteoh: Nope, it doesn't. I just put it there for illustrative purposes. Removed in: [17f42b7](https://github.com/jaegertracing/jaeger/pull/4452/commits/17f42b70209bb85f8c78627295a4838e94d8adb0).

docker-compose/monitor/Makefile
albertteoh: Updated in: [f030268](https://github.com/jaegertracing/jaeger/pull/4452/commits/f030268105324ef71b6e89201e55c5959d732848).

docker-compose/monitor/README.md
albertteoh: Good idea, updated in [f030268](https://github.com/jaegertracing/jaeger/pull/4452/commits/f030268105324ef71b6e89201e55c5959d732848).

plugin/metrics/prometheus/options.go
albertteoh: I like the thinking behind the idea, but it means users need to configure this value to `1ms` or `1s`, which seems a little awkward to me. `ms` or `s` results in a parse error.

yurishkuro: are ok, fair enough

albertteoh:
@yurishkuro please let me know if there's anything else that needs addressing in this PR.

yurishkuro (approved):

hdong69:
hi @albertteoh @yurishkuro. Are the changes included in all the images of Jaeger componets (agent, colector, query,....) version 1.45 ? If not, is any addional configuration needed to add ? Thanks

albertteoh:
> Are the changes included in all the images of Jaeger componets (agent, colector, query,....) version 1.45 ?

@hdong69 this change should be available from the next release (version 1.46.0), and is only relevant to jaeger all-in-one and jaeger-query components.

>  is any addional configuration needed to add ?

Please find instructions here: https://github.com/jaegertracing/jaeger/tree/main/docker-compose/monitor#migrating

hdong69:
@albertteoh thanks for your feedback. so I understand that following the instruction (adding some variables in jaeger-query,...) could make the current version of jaeger query (jaegertracing/jaeger-query:1.45) works ? please correct me if i am wrong. thanks

albertteoh:
> so I understand that following the instruction (adding some variables in jaeger-query,...) could make the current version of jaeger query (jaegertracing/jaeger-query:1.45) works (curently, p95 doesnot appear in my Jaeger UI ) ?

Apologies, I wasn't clear. The instructions above only apply from version 1.46.

jaeger-query 1.45 will only work (for Latencies) with OTEL collector versions 0.70.0 and below. It was broken because of a breaking change to the prometheusexporter in OTEL collector.

I'll create a PR that updates the v1.46 documentation to include these details.

hdong69:
> > so I understand that following the instruction (adding some variables in jaeger-query,...) could make the current version of jaeger query (jaegertracing/jaeger-query:1.45) works (curently, p95 doesnot appear in my Jaeger UI ) ?
> 
> Apologies, I wasn't clear. The instructions above only apply from version 1.46.
> 
> jaeger-query 1.45 will only work (for Latencies) with OTEL collector versions 0.70.0 and below. It was broken because of a breaking change to the prometheusexporter in OTEL collector.
> 
> I'll create a PR that updates the v1.46 documentation to include these details.

okay i got it. Many thanks

tmc:
Can we get a 1.46 release out?

PR #4470, opened 2023-05-26
Bump mongo driver to latest

pavolloffay:


albertteoh (commented):
Curious, what's the motivation for the version bump?

yurishkuro:
oh boy...   this is coming as a dependency of `go-openapi`. @pavolloffay I think it was you who introduced `go-openapi`, how much do we need it in this repo?

yurishkuro (approved):

PR #4494, opened 2023-05-31
Make OTLP receiver enabled by default

yurishkuro:
This simplifies compatibility with OTEL/OTLP by not requiring to set the option to enable OTLP receiver. The option is still supported and allows OTLP receiver to be disabled when needed.


albertteoh (approved):

lynieya:
dont ask me, its update all posted

PR #4496, opened 2023-06-02
[ci]: skipping debug builds when not making a release

psk001:
Adjusted the build scripts to skip building debug versions unless the GH actions runs for a release by including a SKIP_DEBUG variable in command.


psk001:
@yurishkuro please review the PR and suggest changes.

scripts/build-all-in-one-image.sh
```diff
@@ -46,10 +48,14 @@ run_integration_test localhost:5000/$repo
...
+if [ "$SKIP_DEBUG" = true ] ; then
+  make build-all-in-one-debug GOOS=linux GOARCH=$GOARCH
+  repo=${repo}-debug
+  #build all-in-one-debug image locally for integration test
+  bash scripts/build-upload-a-docker-image.sh -l -b -c all-in-one-debug -d cmd/all-in-one -t debug
+  run_integration_test localhost:5000/$repo
+  #build all-in-one-debug image and upload to dockerhub/quay.io
+  bash scripts/build-upload-a-docker-image.sh -b -c all-in-one-debug -d cmd/all-in-one -t debug
+else 
+  echo "Skipping debug commands"
```
yurishkuro, line 60: looks like this should be in the `then` part

.github/workflows/ci-all-in-one-build.yml
```diff
@@ -48,3 +48,4 @@ jobs:
       env:
         DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
         QUAY_TOKEN: ${{ secrets.QUAY_TOKEN }}
+        SKIP_DEBUG: ${{inputs.SKIP_DEBUG}}
```
yurishkuro, line 51: We don't run actions manually. The main release is done by `.github/workflows/ci-release.yml`, so there it should be enabling debug builds, and here they should be explicitly disabled.

yurishkuro (commented):
Another place where debug images are built is scripts/build-upload-docker-images.sh

Note that that script already accepts `pr-only` argument, so we can adjust it directly to exclude debug logic for pr-only, without any changes to the workflows. And for consistency we should implement a similar flag for `scripts/build-all-in-one-image.sh`, instead of the new SKIP_DEBUG variable.

psk001:
Got them all, working on it

yurishkuro (approved):

Issue #4459, opened 2023-05-23
[Feature]: Support CORS headers on OTEL HTTP endpoint


yurishkuro:
### Requirement

Allow exporting traces from browser directly to Jaeger. 

### Problem

Access from the browser usually requires addressing the CORS issue.

https://stackoverflow.com/questions/76266837/open-telemetry-cors-issue-exporting-trace-data-to-jaeger

### Proposal

Support CORS headers similar to how they are supported on Zipkin endpoint (see  #2056). If OTLP HTTP Receiver already supports these, we need to allow configuring them.

yurishkuro:
If you don't know outright how to implement it, you can help by doing research:
  * does OTLP receiver support this? If so, how/where it can be configured.

samyakjain10:
Hi @sid-008, was just wondering if you are still working on this issue

sid-008:
No I am not, I don't really know how to outright implement it, you can take a shot at it

shubmjagtap:
Hi @yurishkuro  can i work on this ? @sbdtu5498 opened PR but its been 14 days . So can i work on this ?

shubmjagtap:
yes thanks

sbdtu5498:
@yurishkuro didn't get time to catch up on the work again. I have closed the pr for the time being.

PR #4510, opened 2023-06-06
enabling CORS for collector HTTP endpoints

sbdtu5498:

- Resolving #4459 

@yurishkuro  I have hard-coded the CORS functionality in the Collector HTTP server to make cross-origin requests on port 14268. I think this would solve the issue that has been discussed in the Stack Overflow link provided in the issue. This is still in draft mode as I wasn't aware of the stuff such as how much control we want over the CORS handling. So here we can implement CORS via handlers or directly via the HTTP server (that I have done currently).

Implementing CORS in the handler: You can add the CORS headers directly within each handler function in the handler files (http_thrift_handler.go, otlp_receiver.go, thrift_handler.go). This approach allows us to have fine-grained control over CORS headers for each specific handler. Also, controlling via CLI arguments is possible. It is also easily doable and can be done after a bit more discussion.

Implementing CORS in the HTTP server: We can implement a wrapper around the HTTP server in the http.go to add the CORS headers. This approach would apply the CORS headers to all requests handled by the HTTP server, regardless of the specific handler. This is what I have done and it enables CORS on port 1428 (I have tested it locally via cURL request).

Can you please provide me with the details of the features that you want to see in this feature such as CLI arguments for fine-tuning CORS functionalities, etc? Thanks!

### DIFF
--- a/cmd/collector/app/server/http.go
+++ b/cmd/collector/app/server/http.go
@@ -99,6 +99,23 @@ func serveHTTP(server *http.Server, listener net.Listener, params *HTTPServerPar
 
 	recoveryHandler := recoveryhandler.NewRecoveryHandler(params.Logger, true)
 	server.Handler = httpmetrics.Wrap(recoveryHandler(r), params.MetricsFactory, params.Logger)
+	
+	// CORS handler
+	corsHandler := func(h http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			// Required CORS headers
+			w.Header().Set("Access-Control-Allow-Origin", "*")
+			w.Header().Set("Access-Control-Allow-Methods", "GET, POST, OPTIONS")
+			w.Header().Set("Access-Control-Allow-Headers", "Content-Type")
+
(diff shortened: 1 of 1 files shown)


sbdtu5498:
> Implementing CORS in the handler: You can add the CORS headers directly within each handler function in the handler files (http_thrift_handler.go, otlp_receiver.go, thrift_handler.go). This approach allows us to have fine-grained control over CORS headers for each specific handler. Also, controlling via CLI arguments is possible. It is also easily doable and can be done after a bit more discussion 

@pmuls99 I have mentioned it in my PR. It is just a first draft.

yurishkuro:
This PR adds CORS headers on the collector's Jaeger-native HTTP port, but that port does not accept any data in JSON format, only binary Thrift, so it's not going to address the original ticket. The recommended path for people would be to use OTLP JSON exporter from the browser, meaning that we need CORS on OTLP receiver, which is already [supporter](https://github.com/open-telemetry/opentelemetry-collector/blob/main/receiver/otlpreceiver/config.md#confighttp-httpserversettings), but our configuration does not allow to specify it.

PR #4512, opened 2023-06-06
fix: change es version check to support elasticsearch 8+

kjschnei001:
- Allows for writing of spans to a v8 elasticsearch index.
- I am utilizing the library only for writing spans to an elasticsearch instance, so I can't confirm that these changes are sufficient for elasticsearch 8+ support in jaeger, but I do know that the `_type` field is incompatible in elasticsearch v8, and removing it allows me to successfully write to an index in elasticsearch v8. 
- Even if it doesn't fully satisfy the requirements for v8 support, it does feel like a very safe change to make, as it doesn't modify the logic for v7 and lower versions of elasticsearch.

- This PR modifies the version check for elasticsearch to apply v7 logic to any version equal to or greater than v7.


yurishkuro:
The change needs to be demonstrated with CI running successfully against v8. I added it to the matrix, but I think there is some infra issue where a started DB is not recognized by our script, needs investigation.

yurishkuro:
I think this is already fixed, we have CI running for esv8. Closing.

PR #4515, opened 2023-06-07
[cassandra] Add Configuration.Close() to ensure TLS cert watcher is closed

### LINKED ISSUE #4514: [Bug]: Leaky Go Routine in Jaeger Fswatcher.go
### What happened?

There is a leaky go routine in Jaeger fswatcher.go library. In fswatcher.go, the new go routine is called whenever New Function is created but that go routine was never terminated.

### Steps to reproduce

This only when the [new Cassandra Cluster](https://github.com/jaegertracing/jaeger/blob/main/pkg/cassandra/config/config.go#L109) is created directly with no option to close the[ fswatcher.go](https://github.com/jaegertracing/jaeger/blob/main/pkg/fswatcher/fswatcher.go#L77) goroutine.

### Expected behavior

Cassandra Cluster should have a public Close method that allows user to Close the leaky goroutine.

### Relevant log output

```shell
`[Goroutine 122 in state select, with github.com/jaegertracing/jaeger/pkg/fswatcher.(*FSWatcher).watch on top of the stack:
goroutine 122 [select]:
github.com/jaegertracing/jaeger/pkg/fswatcher.(*FSWatcher).watch(0xc0005d9d40)
        external/com_github_jaegertracing_jaeger/pkg/fswatcher/fswatcher.go:112 +0xc5
created by github.com/jaegertracing/jaeger/pkg/fswatcher.New
        external/com_github_jaegertracing_jaeger/pkg/fswatcher/fswatcher.go:77 +0x125`
```

### Screenshot

_No response_

### Additional context

_No response_

### Jaeger backend version

v1.46.0

### SDK

_No response_

### Pipeline

_No response_

### Stogage backend

_No response_

### Operating system

_No response_

### Deployment model

_No response_

### Deployment configs

_No response_

kennyaz:
Resolves #4514 

Short description of the changes
Implement a Close Function in Cassandra to allow us to close the leaky-go routine.


yurishkuro (approved):
Thanks!

kennyaz:
The first push failed because of the test case in TestProxyBuilder in build_test.go file is calling fswatcher.go Close function twice. Added protection so that if the channel is already closed, it will not try to attempt to close the channel again.

pkg/fswatcher/fswatcher.go
```diff
@@ -133,12 +139,21 @@ func (w *FSWatcher) watch() {
...
 }
 
-// Close closes the watcher.
+// Close closes the watcher and stops the background go routine of FSWatcher.
 func (w *FSWatcher) Close() error {
+	if !w.isClosed {
+		close(w.stop)
+		w.wg.Wait()
+	}
+	w.isClosed = true
```
yurishkuro, line 155: unprotected bool can be a source of race condition.

kennyaz: I see, let me add the mutex lock to prevent the race conditions.

yurishkuro (approved):

yurishkuro:
please make sure all commits are signed (easiest is to squash into one)

kennyaz:
I squashed and signed all my commits. Can you approve it again?

pkg/fswatcher/fswatcher.go
```diff
@@ -133,12 +139,20 @@ func (w *FSWatcher) watch() {
...
+			w.wg.Done()
+			return
 		}
 	}
 }
 
-// Close closes the watcher.
+// Close closes the watcher and stops the background go routine of FSWatcher.
 func (w *FSWatcher) Close() error {
+	if w.isClosed.CompareAndSwap(false, true) {
```
yurishkuro, line 151: I was looking closer at this and I am not convinced this is all needed. When we call `w.watcher.Close()`, it closes the `Events` and `Errors` channels, so our goroutine will exit naturally. Yes, there is a race condition if you're checking for its exit immediately, but that cannot be completely avoided, even with these changes the goroutine may go to sleep after L143 lowers the semaphore.

If you observed that the goroutine is still running for a long time after close, there may be something else going on.

We can still use the waitgroup introduced here, but without the channel/bool:
```go
func (w *FSWatcher) Close() error {
	// watcher.Close() takes care of multiple calls by returning "already closed" error
	if err := w.watcher.Close(); err != nil {
		return err
	}
	w.wg.Wait()
	return nil
}
```

kennyaz: I see. let me send out a new commit with these changes.

yurishkuro: Can you elaborate how you determined that the GR was still running? Can it be reproduced with a unit test?

yurishkuro: +1

pkg/cassandra/config/config.go
```diff
@@ -163,6 +163,10 @@ func (c *Configuration) NewCluster(logger *zap.Logger) (*gocql.ClusterConfig, er
 	return cluster, nil
 }
 
+func (c *Configuration) Close() error {
```
yurishkuro, line 166: Since this function did not exist before, nothing is calling it, so just adding it here is not going to address the issue in the ticket.

kennyaz: So I looked at the code base, the leaky go routine only happens when new Cassandra Cluster is created directly with no option to close. So, the leaky-go routine is happening on our side, not on your end. I am thinking about changing a ticket from bug to feature request so that users who created the new Cassandra Cluster will have the option to close the fswatcher leaky go routine.

yurishkuro (approved):

PR #4518, opened 2023-06-08
[hotrod] Reduce span exporter's batch timeout to let the spans be exported sooner

GLVSKiriti:
In HotROD appbatching is not required it should export traces asap so BatchTimeOut value can be lowered from 5sec solves the issue
Resolves jaegertracing/jaeger-ui#1180

### DIFF
--- a/examples/hotrod/pkg/tracing/init.go
+++ b/examples/hotrod/pkg/tracing/init.go
@@ -21,6 +21,7 @@ import (
 	"os"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/opentracing/opentracing-go"
 	"go.opentelemetry.io/otel"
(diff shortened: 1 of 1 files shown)


yurishkuro (approved):

examples/hotrod/pkg/tracing/init.go
```diff
@@ -62,7 +63,7 @@ func Init(serviceName string, exporterType string, metricsFactory metrics.Factor
 	rpcmetricsObserver := rpcmetrics.NewObserver(metricsFactory, rpcmetrics.DefaultNameNormalizer)
 
 	tp := sdktrace.NewTracerProvider(
-		sdktrace.WithBatcher(exp),
+		sdktrace.WithBatcher(exp, sdktrace.WithBatchTimeout(1000*time.Millisecond)),
```
albertteoh, line 66: nit: could express this as `time.Second`.

Issue #2744, opened 2021-01-27
Kerberos authentication for KDCs like Active Directory that do not support FAST negotiation


kaz-zahed:
## Requirement - what kind of business use case are you trying to solve?

Kerberos authentication for KDCs like Active Directory that do not support FAST negotiation.

## Problem - what in Jaeger blocks you from solving the requirement?

Jaeger does not expose the configuration property that allows disabling FAST negotiation for Kerberos authentication.

## Proposal - what do you suggest to solve the problem or improve the existing situation?

Shopify/sarama Kafka library exposed this configuration property with release v1.27.0 with this [pull request](https://github.com/Shopify/sarama/pull/1466). Jaeger's Kafka Kerberos authentication configuration can be updated to allow configuring this property.

pmuls99:
Hey @kaz-zahed, @yurishkuro seems like the PR raised for this issue is closed. Could I take this up and raise a PR?

yurishkuro:
Feel free to try

pmuls99:
Hey @yurishkuro , I just raised a PR (#4520)addressing the issue. Could you please take a look and give me pointers on improvement!

PR #4520, opened 2023-06-09
Add *.kerberos.disable-fast-negotiation option to Kafka consumer

pmuls99:
- Solves #2744 

- Added a variable in KerborosConfig struct and configured it 
- Added flags and the configuration for the same

### DIFF
--- a/pkg/kafka/auth/options.go
+++ b/pkg/kafka/auth/options.go
@@ -87,8 +87,7 @@ func addKerberosFlags(configPrefix string, flagSet *flag.FlagSet) {
 	flagSet.Bool(
 		configPrefix+kerberosPrefix+suffixKerberosDisablePAFXFAST,
 		defaultKerberosDisablePAFXFast,
-		"Disable FAST negotiation when not supported by KDC's like Active Directory")
-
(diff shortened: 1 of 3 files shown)


pkg/kafka/auth/options.go
```diff
@@ -82,6 +84,11 @@ func addKerberosFlags(configPrefix string, flagSet *flag.FlagSet) {
 		configPrefix+kerberosPrefix+suffixKerberosKeyTab,
 		defaultKerberosKeyTab,
 		"Path to keytab file. i.e /etc/security/kafka.keytab")
+	flagSet.Bool(
+		configPrefix+kerberosPrefix+suffixKerberosDisablePAFXFAST,
+		defaultKerberosDisablePAFXFast,
+		"Disable FAST negotiation when not supported by KDC's like Active Directory")
```
yurishkuro, line 90: Q: why are the variables in the code have additional "PA FX" prefix? Is there semantic difference between PF-FX-FAST and just FAST? (it would be useful to include a link to the spec in the help text)

pmuls99: So the additional PA FX refers to the cookie used. I considered naming the variable like that because they were done the same way in the sarama library. There is not a logical difference between both so we can consider naming it FAST. Whats your say?

yurishkuro (approved):

PR #4533, opened 2023-06-16
feat: Added OTEL `TracerProvider` instead of `opentracing.Tracer` for hotROD example app service (Redis)

afzal442:

Related to #3380 
Part of #3381 

- This adds `OTEL tracerProvider` and removes `OT tracer` for Redis service hotROD app

Before:
r2

After:
r1

### DIFF
--- a/examples/hotrod/pkg/tracing/init.go
+++ b/examples/hotrod/pkg/tracing/init.go
@@ -44,27 +44,23 @@ import (
 
 var once sync.Once
 
-// InitOTEL initializes OpenTelemetry SDK
-// to return an Otel tracer.
+// InitOTEL initializes OpenTelemetry SDK.
 func InitOTEL(serviceName string, exporterType string, metricsFactory metrics.Factory, logger log.Factory) trace.Tracer {
 	_, oteltp := initBOTH(serviceName, exporterType, metricsFactory, logger)
-	otel.SetTracerProvider(oteltp)
 
-	logger.Bg().Debug("Added OTEL tracer", zap.String("service-name", serviceName))
+	logger.Bg().Debug("Created OTEL tracer", zap.String("service-name", serviceName))
 	return oteltp.Tracer(serviceName)
 }
 
-// InitOP initializes OTel-OpenTracing Bridge
-// to return an OpenTracing-compatible tracer.
+// Init returns OTel-OpenTracing Bridge.
 func Init(serviceName string, exporterType string, metricsFactory metrics.Factory, logger log.Factory) opentracing.Tracer {
-	optracer, _ := initBOTH(serviceName, exporterType, metricsFactory, logger)
+	otTracer, _ := initBOTH(serviceName, exporterType, metricsFactory, logger)
 
-	logger.Bg().Debug("created OTEL->OT bridge", zap.String("service-name", serviceName))
-	return optracer
+	logger.Bg().Debug("Created OTEL->OT bridge", zap.String("service-name", serviceName))
+	return otTracer
 }
 
-// Init initializes OpenTelemetry SDK and uses OTel-OpenTracing Bridge
-// to return an OpenTracing-compatible tracer and Otel tracer.
+// initBOTH initializes OpenTelemetry SDK and uses OTel-OpenTracing Bridge
 func initBOTH(serviceName string, exporterType string, metricsFactory metrics.Factory, logger log.Factory) (opentracing.Tracer, trace.TracerProvider) {
 	once.Do(func() {
 		otel.SetTextMapPropagator(
--- a/examples/hotrod/pkg/tracing/init.go
+++ b/examples/hotrod/pkg/tracing/init.go
@@ -46,18 +46,18 @@ var once sync.Once
 
 // InitOTEL initializes OpenTelemetry SDK
 // to return an Otel tracer.
-func InitOTEL(serviceName string, exporterType string, metricsFactory metrics.Factory, logger log.Factory) trace.TracerProvider {
-	_, oteltp := InitBOTH(serviceName, exporterType, metricsFactory, logger)
+func InitOTEL(serviceName string, exporterType string, metricsFactory metrics.Factory, logger log.Factory) trace.Tracer {
+	_, oteltp := initBOTH(serviceName, exporterType, metricsFactory, logger)
 	otel.SetTracerProvider(oteltp)
 
 	logger.Bg().Debug("Added OTEL tracer", zap.String("service-name", serviceName))
-	return oteltp
+	return oteltp.Tracer(serviceName)
 }
 
 // InitOP initializes OTel-OpenTracing Bridge
 // to return an OpenTracing-compatible tracer.
-func InitOP(serviceName string, exporterType string, metricsFactory metrics.Factory, logger log.Factory) opentracing.Tracer {
-	optracer, _ := InitBOTH(serviceName, exporterType, metricsFactory, logger)
+func Init(serviceName string, exporterType string, metricsFactory metrics.Factory, logger log.Factory) opentracing.Tracer {
+	optracer, _ := initBOTH(serviceName, exporterType, metricsFactory, logger)
 
 	logger.Bg().Debug("created OTEL->OT bridge", zap.String("service-name", serviceName))
 	return optracer
@@ -65,7 +65,7 @@ func InitOP(serviceName string, exporterType string, metricsFactory metrics.Fact
 
 // Init initializes OpenTelemetry SDK and uses OTel-OpenTracing Bridge
 // to return an OpenTracing-compatible tracer and Otel tracer.
-func InitBOTH(serviceName string, exporterType string, metricsFactory metrics.Factory, logger log.Factory) (opentracing.Tracer, trace.TracerProvider) {
+func initBOTH(serviceName string, exporterType string, metricsFactory metrics.Factory, logger log.Factory) (opentracing.Tracer, trace.TracerProvider) {
 	once.Do(func() {
 		otel.SetTextMapPropagator(
 			propagation.NewCompositeTextMapPropagator(
@@ -90,7 +90,7 @@ func InitBOTH(serviceName string, exporterType string, metricsFactory metrics.Fa
 			semconv.ServiceNameKey.String(serviceName),
 		)),
 	)
-	otTracer, _ := otbridge.NewTracerPair(tp.Tracer(""))
+	otTracer, _ := otbridge.NewTracerPair(tp.Tracer(serviceName))
 	return otTracer, tp
 }
 
--- a/examples/hotrod/pkg/tracing/init.go
+++ b/examples/hotrod/pkg/tracing/init.go
@@ -34,6 +34,7 @@ import (
 	"go.opentelemetry.io/otel/sdk/resource"
 	sdktrace "go.opentelemetry.io/otel/sdk/trace"
 	semconv "go.opentelemetry.io/otel/semconv/v1.7.0"
+	"go.opentelemetry.io/otel/trace"
 	"go.uber.org/zap"
 
 	"github.com/jaegertracing/jaeger/examples/hotrod/pkg/log"
@@ -43,9 +44,28 @@ import (
 
 var once sync.Once
 
-// Init initializes OpenTelemetry SDK and uses OTel-OpenTracing Bridge
+// InitOTEL initializes OpenTelemetry SDK
+// to return an Otel tracer.
+func InitOTEL(serviceName string, exporterType string, metricsFactory metrics.Factory, logger log.Factory) trace.TracerProvider {
+	_, oteltp := InitBOTH(serviceName, exporterType, metricsFactory, logger)
+	otel.SetTracerProvider(oteltp)
+
+	logger.Bg().Debug("Added OTEL tracer", zap.String("service-name", serviceName))
+	return oteltp
+}
+
+// InitOP initializes OTel-OpenTracing Bridge
 // to return an OpenTracing-compatible tracer.
-func Init(serviceName string, exporterType string, metricsFactory metrics.Factory, logger log.Factory) opentracing.Tracer {
+func InitOP(serviceName string, exporterType string, metricsFactory metrics.Factory, logger log.Factory) opentracing.Tracer {
+	optracer, _ := InitBOTH(serviceName, exporterType, metricsFactory, logger)
+
+	logger.Bg().Debug("created OTEL->OT bridge", zap.String("service-name", serviceName))
+	return optracer
+}
+
+// Init initializes OpenTelemetry SDK and uses OTel-OpenTracing Bridge
+// to return an OpenTracing-compatible tracer and Otel tracer.
+func InitBOTH(serviceName string, exporterType string, metricsFactory metrics.Factory, logger log.Factory) (opentracing.Tracer, trace.TracerProvider) {
 	once.Do(func() {
 		otel.SetTextMapPropagator(
 			propagation.NewCompositeTextMapPropagator(
@@ -71,8 +91,7 @@ func Init(serviceName string, exporterType string, metricsFactory metrics.Factor
 		)),
 	)
 	otTracer, _ := otbridge.NewTracerPair(tp.Tracer(""))
-	logger.Bg().Debug("created OTEL->OT bridge", zap.String("service-name", serviceName))
-	return otTracer
+	return otTracer, tp
 }
 
 // withSecure instructs the client to use HTTPS scheme, instead of hotrod's desired default HTTP
--- a/examples/hotrod/services/driver/redis.go
+++ b/examples/hotrod/services/driver/redis.go
@@ -22,8 +22,8 @@ import (
 	"math/rand"
 	"sync"
 
-	"github.com/opentracing/opentracing-go"
-	"github.com/opentracing/opentracing-go/ext"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
 	"go.uber.org/zap"
 
 	"github.com/jaegertracing/jaeger/examples/hotrod/pkg/delay"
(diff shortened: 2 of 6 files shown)


afzal442:
Hi there, currently, it is out of trace with parent context. I am not sure why. on it.

examples/hotrod/pkg/tracing/init.go
```diff
@@ -43,6 +44,40 @@ import (
...
+			semconv.SchemaURL,
+			semconv.ServiceNameKey.String(serviceName),
+		)),
+	)
+	otel.SetTracerProvider(tp)
+	// _, otelTracerProvider := otbridge.NewTracerPair(tp.Tracer(""))
+	tracer := otel.Tracer("my otel tracer")
+	// logger.Bg().Debug("created OTEL->OT bridge", zap.String("service-name", serviceName))
+	return tracer
+}
```
yurishkuro, line 79: all of this is already done in the Init function below. You can simply change that function to return two objects:
```
func Init() (opentracing.Tracer, trace.TracerProvider) {}
```

yurishkuro: Or at least refactor in such a way that we don't duplicate the code, e.g. 
```
func InitOTEL() trace.TracerProvider {
  _, otel := initBoth()
  return otel
}
// rename Init() to initBoth()
func initBoth() (opentracing.Tracer, trace.TracerProvider) {}
```

examples/hotrod/services/driver/redis.go
```diff
@@ -35,27 +35,38 @@ import (
...
-	tracer opentracing.Tracer // simulate redis as a separate process
+	tracer trace.Tracer // simulate redis as a separate process
 	logger log.Factory
 	errorSimulator
 }
 
 func newRedis(otelExporter string, metricsFactory metrics.Factory, logger log.Factory) *Redis {
 	return &Redis{
-		tracer: tracing.Init("redis", otelExporter, metricsFactory, logger),
+		tracer: tracing.InitOTEL("redis", otelExporter, metricsFactory, logger),
```
yurishkuro, line 45: InitOTEL should be returning TracerProvider. Then you get the tracer from it by passing the layer/instrumentation name.

examples/hotrod/services/driver/redis.go
```diff
@@ -35,27 +35,38 @@ import (
...
 	}
 }
 
 // FindDriverIDs finds IDs of drivers who are near the location.
 func (r *Redis) FindDriverIDs(ctx context.Context, location string) []string {
-	if span := opentracing.SpanFromContext(ctx); span != nil {
+
+	// tracer := otel.Tracer(serviceName)
+	ctx, span := r.tracer.Start(ctx, "FindDriverIDs")
+	if span != nil {
```
yurishkuro, line 55: why would it be nil here? Start always returns a span

examples/hotrod/services/driver/redis.go
```diff
@@ -35,27 +35,38 @@ import (
...
+		_, span := r.tracer.Start(ctx, "FindDriverIDs", trace.WithSpanKind(trace.SpanKindServer) /*...set child of option*/)
```
yurishkuro, line 56: why do you start a 2nd span?

examples/hotrod/services/driver/redis.go
```diff
@@ -35,27 +35,38 @@ import (
...
+
+		span.SetAttributes(attribute.Key("param.location").String(location))
+
+		span.SetName("FindDriverIDs")
```
yurishkuro, line 60: name was already passed to Start() function

examples/hotrod/services/driver/redis.go
yurishkuro: according to L66 this used to be Client-type span.

examples/hotrod/services/driver/redis.go
```diff
@@ -70,18 +81,29 @@ func (r *Redis) FindDriverIDs(ctx context.Context, location string) []string {
...
 		defer span.Finish()
 		ctx = opentracing.ContextWithSpan(ctx, span)
-	}
+	} */
 	// simulate RPC delay
 	delay.Sleep(config.RedisGetDelay, config.RedisGetDelayStdDev)
 	if err := r.checkError(); err != nil {
-		if span := opentracing.SpanFromContext(ctx); span != nil {
-			ext.Error.Set(span, true)
+		if _, span := r.tracer.Start(ctx, "GetDriver"); span != nil {
```
yurishkuro, line 105: you already have the span created in L84

examples/hotrod/pkg/tracing/init.go
afzal442: Ah thanks, that'll be nice.

examples/hotrod/services/driver/redis.go
```diff
@@ -35,27 +35,34 @@ import (
...
-		tracer: tracing.Init("redis", otelExporter, metricsFactory, logger),
+		tracer: tracing.InitOTEL("redis", otelExporter, metricsFactory, logger),
 		logger: logger,
 	}
 }
 
 // FindDriverIDs finds IDs of drivers who are near the location.
 func (r *Redis) FindDriverIDs(ctx context.Context, location string) []string {
-	if span := opentracing.SpanFromContext(ctx); span != nil {
+	tracer := r.tracer.Tracer("")
```
yurishkuro, line 52: This should be only done once in the constructor.
```suggestion
	tracer := r.tracer.Tracer("redis-manual")
```

examples/hotrod/services/driver/redis.go
```diff
@@ -35,27 +35,34 @@ import (
...
+	if _, span := tracer.Start(ctx, "FindDriverIDs", trace.WithSpanKind(trace.SpanKindClient) /*...set child of option*/); span != nil {
```
yurishkuro, line 53: Don't need if, it can never be nil

examples/hotrod/services/driver/redis.go
```diff
@@ -35,27 +35,34 @@ import (
...
+		span.SetAttributes(attribute.Key("param.location").String(location))
+
+		defer span.End()
+		ctx = trace.ContextWithSpan(ctx, span)
```
yurishkuro, line 57: Doesn't Start already return modified context?

examples/hotrod/pkg/tracing/init.go
```diff
@@ -43,9 +44,28 @@ import (
...
+	otel.SetTracerProvider(oteltp)
+
+	logger.Bg().Debug("Added OTEL tracer", zap.String("service-name", serviceName))
+	return oteltp
+}
+
+// InitOP initializes OTel-OpenTracing Bridge
 // to return an OpenTracing-compatible tracer.
-func Init(serviceName string, exporterType string, metricsFactory metrics.Factory, logger log.Factory) opentracing.Tracer {
+func InitOP(serviceName string, exporterType string, metricsFactory metrics.Factory, logger log.Factory) opentracing.Tracer {
```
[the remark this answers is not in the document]
afzal442: That means we will keep as was `func Init(...)` for now becaz other function call gets affected.

examples/hotrod/services/driver/redis.go
afzal442: Ah, I see, was just ignoring the fact. Thanks

examples/hotrod/services/driver/redis.go
afzal442: Yes, it does, but I thought like L57 `ctx` gets modified with the parent context. but not sure.
https://github.com/jaegertracing/jaeger/blob/bfed6e71ceb633296aab05df10a8fec995c7ff8c/examples/hotrod/services/driver/redis.go#L57

examples/hotrod/services/driver/redis.go
afzal442: `func InitOTEL(...)` should do this job. Right?

yurishkuro: No. InitOTEL returns TracerProvider. Different layers in a service can use provider to obtain differently named Tracers, e.g. if your server endpoint is HTTP using Gorilla framework, the instrumentation may ask for a tracer called "gorilla", whereas a gRPC instrumentation may ask for tracer called "gRPC".

examples/hotrod/pkg/tracing/init.go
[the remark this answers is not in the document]
yurishkuro: yes

examples/hotrod/services/driver/redis.go
yurishkuro: This is a difference between OT and OTEL APIs. In OTEL you don't need to do that as it's already done when you call tracer.Start

examples/hotrod/services/driver/redis.go
```diff
@@ -35,27 +35,34 @@ import (
...
 		logger: logger,
 	}
 }
 
 // FindDriverIDs finds IDs of drivers who are near the location.
 func (r *Redis) FindDriverIDs(ctx context.Context, location string) []string {
-	if span := opentracing.SpanFromContext(ctx); span != nil {
+	tracer := r.tracer.Tracer("")
+	if _, span := tracer.Start(ctx, "FindDriverIDs", trace.WithSpanKind(trace.SpanKindClient) /*...set child of option*/); span != nil {
+		span.SetAttributes(attribute.Key("param.location").String(location))
```
yurishkuro, line 54: let's change this a bit:
```suggestion
		span.SetAttributes(attribute.Key("param.driver.location").String(location))
```

yurishkuro:
Just tested this PR as is - it's a good sign that Redis spans appear to work just fine with the rest of the trace. Should make this easier when we migrate the main Jaeger code base.

examples/hotrod/services/driver/redis.go
```diff
@@ -35,27 +35,32 @@ import (
...
 
 // FindDriverIDs finds IDs of drivers who are near the location.
 func (r *Redis) FindDriverIDs(ctx context.Context, location string) []string {
-	if span := opentracing.SpanFromContext(ctx); span != nil {
+	_, span := r.tracer.Start(ctx, "FindDriverIDs", trace.WithSpanKind(trace.SpanKindClient))
+	span.SetAttributes(attribute.Key("param.driver.location").String(location))
+
+	defer span.End()
+
+	/* if span := opentracing.SpanFromContext(ctx); span != nil {
```
albertteoh, line 57: Can the commented-out opentracing instrumentation be removed?

examples/hotrod/pkg/tracing/init.go
```diff
@@ -70,9 +90,8 @@ func Init(serviceName string, exporterType string, metricsFactory metrics.Factor
 			semconv.ServiceNameKey.String(serviceName),
 		)),
 	)
-	otTracer, _ := otbridge.NewTracerPair(tp.Tracer(""))
-	logger.Bg().Debug("created OTEL->OT bridge", zap.String("service-name", serviceName))
-	return otTracer
+	otTracer, _ := otbridge.NewTracerPair(tp.Tracer(serviceName))
```
albertteoh, line 93: I think now we've established that you can mix OT and OTEL tracers for a single trace. 

Given the above, I wonfer if we should remove the otebridge dependency altogether.

examples/hotrod/services/driver/redis.go
afzal442: I think it can be taken as reference to anyone who wants re-instrument their application this way.

examples/hotrod/pkg/tracing/init.go
yurishkuro: @albertteoh we cannot remove the bridge right now, because only `redis.go` was changed to use OTEL instrumentation. My recommendation is to merge this PR when clean, then apply the same principles to all other places where OT instrumentation is still used. Only then we can remove the bridge dependency.

yurishkuro: And yes, it was great to see that OT and OTEL instrumentations were able to interoperate when using the bridge.

examples/hotrod/services/driver/redis.go
yurishkuro: I agree w/ @albertteoh - it's never a good idea to keep commented out code.

examples/hotrod/services/driver/redis.go
```diff
@@ -35,27 +35,25 @@ import (
...
 // FindDriverIDs finds IDs of drivers who are near the location.
 func (r *Redis) FindDriverIDs(ctx context.Context, location string) []string {
-	if span := opentracing.SpanFromContext(ctx); span != nil {
-		span := r.tracer.StartSpan("FindDriverIDs", opentracing.ChildOf(span.Context()))
-		span.SetTag("param.location", location)
-		ext.SpanKindRPCClient.Set(span)
-		defer span.Finish()
-		ctx = opentracing.ContextWithSpan(ctx, span)
-	}
+	_, span := r.tracer.Start(ctx, "FindDriverIDs", trace.WithSpanKind(trace.SpanKindClient))
```
yurishkuro, line 52:
```suggestion
	ctx, span := r.tracer.Start(ctx, "FindDriverIDs", trace.WithSpanKind(trace.SpanKindClient))
```
You are using `ctx` in L65, so need to override it

examples/hotrod/services/driver/redis.go
```diff
@@ -70,23 +68,21 @@ func (r *Redis) FindDriverIDs(ctx context.Context, location string) []string {
...
+	_, span := r.tracer.Start(ctx, "GetDriver", trace.WithSpanKind(trace.SpanKindClient))
+	span.SetAttributes(attribute.Key("param.driverID").String(driverID))
+
+	defer span.End()
+
 	// simulate RPC delay
 	delay.Sleep(config.RedisGetDelay, config.RedisGetDelayStdDev)
 	if err := r.checkError(); err != nil {
-		if span := opentracing.SpanFromContext(ctx); span != nil {
-			ext.Error.Set(span, true)
```
yurishkuro, line 84: this should have an OTEL equivalent, like setting span status to error.

examples/hotrod/services/driver/redis.go
```diff
@@ -70,23 +68,21 @@ func (r *Redis) FindDriverIDs(ctx context.Context, location string) []string {
...
-		span := r.tracer.StartSpan("GetDriver", opentracing.ChildOf(span.Context()))
-		span.SetTag("param.driverID", driverID)
-		ext.SpanKindRPCClient.Set(span)
-		defer span.Finish()
-		ctx = opentracing.ContextWithSpan(ctx, span)
-	}
+	_, span := r.tracer.Start(ctx, "GetDriver", trace.WithSpanKind(trace.SpanKindClient))
+	span.SetAttributes(attribute.Key("param.driverID").String(driverID))
+
+	defer span.End()
```
yurishkuro, line 74:
```suggestion
	ctx, span := r.tracer.Start(ctx, "GetDriver", trace.WithSpanKind(trace.SpanKindClient))
	span.SetAttributes(attribute.Key("param.driverID").String(driverID))
	defer span.End()
```

examples/hotrod/pkg/tracing/init.go
```diff
@@ -43,9 +44,28 @@ import (
...
+	otel.SetTracerProvider(oteltp)
+
+	logger.Bg().Debug("Added OTEL tracer", zap.String("service-name", serviceName))
+	return oteltp.Tracer(serviceName)
+}
+
+// InitOP initializes OTel-OpenTracing Bridge
 // to return an OpenTracing-compatible tracer.
 func Init(serviceName string, exporterType string, metricsFactory metrics.Factory, logger log.Factory) opentracing.Tracer {
+	optracer, _ := initBOTH(serviceName, exporterType, metricsFactory, logger)
```
yurishkuro, line 60:
```suggestion
	otTracer, _ := initBOTH(serviceName, exporterType, metricsFactory, logger)
```

examples/hotrod/pkg/tracing/init.go
```diff
@@ -43,9 +44,28 @@ import (
 
 var once sync.Once
 
-// Init initializes OpenTelemetry SDK and uses OTel-OpenTracing Bridge
+// InitOTEL initializes OpenTelemetry SDK
+// to return an Otel tracer.
```
yurishkuro, line 48:
```suggestion
// InitOTEL initializes OpenTelemetry SDK.
```
No need to explain in the comment what the function returns, it's obvious from the signature.

examples/hotrod/pkg/tracing/init.go
```diff
@@ -43,9 +44,28 @@ import (
...
+func InitOTEL(serviceName string, exporterType string, metricsFactory metrics.Factory, logger log.Factory) trace.Tracer {
+	_, oteltp := initBOTH(serviceName, exporterType, metricsFactory, logger)
+	otel.SetTracerProvider(oteltp)
```
yurishkuro, line 51: If we can get away with it, I would prefer not to set this global. Right now you certainly don't need it, so let's see how other OTEL instrumentation is done in the later PRs.

examples/hotrod/pkg/tracing/init.go
```diff
@@ -43,9 +44,28 @@ import (
...
+
+	logger.Bg().Debug("Added OTEL tracer", zap.String("service-name", serviceName))
```
yurishkuro, line 53:
```suggestion
	logger.Bg().Debug("Created OTEL tracer", zap.String("service-name", serviceName))
```

examples/hotrod/pkg/tracing/init.go
```diff
@@ -43,9 +44,28 @@ import (
...
+	return oteltp.Tracer(serviceName)
+}
+
+// InitOP initializes OTel-OpenTracing Bridge
 // to return an OpenTracing-compatible tracer.
 func Init(serviceName string, exporterType string, metricsFactory metrics.Factory, logger log.Factory) opentracing.Tracer {
+	optracer, _ := initBOTH(serviceName, exporterType, metricsFactory, logger)
+
+	logger.Bg().Debug("created OTEL->OT bridge", zap.String("service-name", serviceName))
```
yurishkuro, line 62:
```suggestion
	logger.Bg().Debug("Created OTEL->OT bridge", zap.String("service-name", serviceName))
```
Usually the log statements start with a capital letter, to read like full sentences. But the error messages are traditionally not capitalized.

examples/hotrod/pkg/tracing/init.go
afzal442: Thanks for that.

yurishkuro:
I briefly checked the blog [post][1], and I don't think there is an issue with baggage specifically for the `redis` service, but it may come up in the future changes. So this PR is ok to merge.

[1]: https://medium.com/opentracing/take-opentracing-for-a-hotrod-ride-f6e3141f7941

yurishkuro:
What would be nice is if we had an integration test for HotROD that was validating all the points being made in the blog post.

PR #4548, opened 2023-06-22
[hotROD] Upgrade hotrod to support otel tracer

afzal442:
* Part of #3380
* This PR tries to instrument the hotROD app with otel downgrading opentracing. Related https://github.com/jaegertracing/jaeger/pull/4533#discussion_r1234363250

- Upgrades hotROD application to support `otel tracer`


examples/hotrod/pkg/tracing/init.go
```diff
@@ -45,11 +45,11 @@ import (
 var once sync.Once
 
 // InitOTEL initializes OpenTelemetry SDK.
-func InitOTEL(serviceName string, exporterType string, metricsFactory metrics.Factory, logger log.Factory) trace.Tracer {
+func InitOTEL(serviceName string, exporterType string, metricsFactory metrics.Factory, logger log.Factory) trace.TracerProvider {
 	_, oteltp := initBOTH(serviceName, exporterType, metricsFactory, logger)
 
 	logger.Bg().Debug("Created OTEL tracer", zap.String("service-name", serviceName))
-	return oteltp.Tracer(serviceName)
+	return oteltp
```
yurishkuro, line 52: I knew you'd have to do that, but left it for you to discover :-)

examples/hotrod/services/driver/server.go
```diff
@@ -32,7 +32,7 @@ import (
 // Server implements jaeger-demo-frontend service
 type Server struct {
 	hostPort string
-	tracer   opentracing.Tracer
+	tracer   *trace.TracerProvider
```
yurishkuro, line 35: Is `trace.TracerProvider` a struct or an interface? You don't need a pointer if it's an interface.

examples/hotrod/services/driver/server.go
```diff
@@ -42,14 +42,14 @@ var _ DriverServiceServer = (*Server)(nil)
 
 // NewServer creates a new driver.Server
 func NewServer(hostPort string, otelExporter string, metricsFactory metrics.Factory, logger log.Factory) *Server {
-	tracer := tracing.Init("driver", otelExporter, metricsFactory, logger)
+	tracer := tracing.InitOTEL("driver", otelExporter, metricsFactory, logger)
```
yurishkuro, line 45:
```suggestion
	tracerProvider := tracing.InitOTEL("driver", otelExporter, metricsFactory, logger)
```

afzal442:
Hi @yurishkuro, while I try to update the hotROD app to support otel tracer, logs under gRPC interceptor don't appear. Another thing is I wondered if I should make it otel instrumented as per previous https://github.com/jaegertracing/jaeger/pull/4533#discussion_r1234363250.

yurishkuro:
> logs under gRPC interceptor don't appear like before

that's a problem. Can you investigate? Logs are being written to span via OT span.Log() API. You can check what OTEL Bridge does when it implements this method - I assume it should be translating them into OTEL Span Events. If that works fine, the next step to check is what jaeger2otel transformer is doing in otel-collector-contrib, maybe it's missing the logic to take Span Events and convert them back into Jaeger's span logs.

A good way to test this is to write a unit test.

examples/hotrod/services/driver/server.go
afzal442: Sorry by mistake, I left it when I was testing around this.

yurishkuro:
Please link the description back to the main tracking issue.

examples/hotrod/services/driver/server.go
```diff
@@ -32,7 +32,7 @@ import (
 // Server implements jaeger-demo-frontend service
 type Server struct {
 	hostPort string
-	tracer   opentracing.Tracer
+	tracer   trace.TracerProvider
```
yurishkuro, line 35: Is this being used later?

afzal442: Interesting, I didn't find anywhere.

yurishkuro: let's remove if not used

examples/hotrod/services/route/client.go
```diff
@@ -30,19 +29,26 @@ import (
...
-	tracer   opentracing.Tracer
 	logger   log.Factory
 	client   *tracing.HTTPClient
 	hostPort string
 }
 
+// Transport wraps a RoundTripper. If a request is being traced with
+// Tracer, Transport will inject the current span into the headers,
+// and set HTTP related tags on the span.
+type Transport struct {
```
yurishkuro, line 40: This doesn't seem to add any new functionality compared to stdlib, what's the point? Previous code used custom roundtripper

examples/hotrod/pkg/tracing/http.go
```diff
@@ -21,17 +21,44 @@ import (
...
 
 // HTTPClient wraps an http.Client with tracing instrumentation.
 type HTTPClient struct {
-	Tracer opentracing.Tracer
+	Tracer trace.TracerProvider
 	Client *http.Client
 }
 
+// Tracer holds tracing details for one HTTP request.
+type Tracer struct {
```
yurishkuro, line 43: Isn't there std http instrumentation in OTEL similar to grpc interceptors?

yurishkuro: https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp

examples/hotrod/pkg/tracing/http.go
```diff
@@ -57,6 +60,19 @@ func (c *HTTPClient) GetJSON(ctx context.Context, endpoint string, url string, o
...
 		return errors.New(string(body))
 	}
+
 	decoder := json.NewDecoder(res.Body)
 	return decoder.Decode(out)
 }
+
+// GetJson executes HTTP GET against specified url and tried to parse
+// the response into out object.
+func (c *HTTPClient) GetJson(ctx context.Context, url string, out interface{}) error {
```
yurishkuro, line 70: why new method instead of modifying the original one?

examples/hotrod/services/route/client.go
```diff
@@ -30,20 +30,18 @@ import (
...
 // NewClient creates a new route.Client
-func NewClient(tracer opentracing.Tracer, logger log.Factory, hostPort string) *Client {
+func NewClient(tracer trace.TracerProvider, logger log.Factory, hostPort string) *Client {
 	return &Client{
-		tracer: tracer,
 		logger: logger,
 		client: &tracing.HTTPClient{
-			Client: &http.Client{Transport: &nethttp.Transport{}},
-			Tracer: tracer,
+			Client:         &http.Client{Transport: &otelhttp.Transport{}},
```
yurishkuro, line 43: It doesn't seem like there are situations where you would NOT pass this value. Therefore, I would recommend creating a constructor `tracing.NewHTTPClient(tracerProvider)` to minimize the call at the call sites

examples/hotrod/pkg/tracing/http.go
```diff
@@ -24,12 +24,15 @@ import (
...
 	"github.com/opentracing/opentracing-go"
+	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
+	"go.opentelemetry.io/otel/trace"
 )
 
 // HTTPClient wraps an http.Client with tracing instrumentation.
 type HTTPClient struct {
-	Tracer opentracing.Tracer
-	Client *http.Client
+	Tracer         opentracing.Tracer
```
yurishkuro, line 33: is this ever passed anymore?

examples/hotrod/services/driver/server_test.go
```diff
@@ -0,0 +1,90 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"go.opentelemetry.io/otel/trace"
+	"go.uber.org/zap"
+
+	"github.com/jaegertracing/jaeger/examples/hotrod/pkg/log"
+)
+
+func TestFindNearest(t *testing.T) {
```
yurishkuro, line 28: what are we testing here? Better not have tests than have those that don't test much.

examples/hotrod/services/frontend/server.go
```diff
@@ -72,16 +73,17 @@ func NewServer(options ConfigOptions, tracer opentracing.Tracer, logger log.Fact
...
 	mux := s.createServeMux()
 	s.logger.Bg().Info("Starting", zap.String("address", "http://"+path.Join(s.hostPort, s.basepath)))
-	return http.ListenAndServe(s.hostPort, mux)
+	return http.ListenAndServe(s.hostPort, otelhttp.NewHandler(mux, s.basepath,
+		otelhttp.WithTracerProvider(s.tracer)))
 }
 
 func (s *Server) createServeMux() http.Handler {
-	mux := tracing.NewServeMux(true, s.tracer, s.logger)
+	// mux := tracing.NewServeMux(true, s.tracer, s.logger)
```
yurishkuro, line 81: same line

examples/hotrod/services/frontend/server.go
```diff
@@ -72,16 +73,17 @@ func NewServer(options ConfigOptions, tracer opentracing.Tracer, logger log.Fact
...
 	p := path.Join("/", s.basepath)
-	mux.Handle(p, http.StripPrefix(p, http.FileServer(s.assetFS)))
-	mux.Handle(path.Join(p, "/dispatch"), http.HandlerFunc(s.dispatch))
-	mux.Handle(path.Join(p, "/config"), http.HandlerFunc(s.config))
-	return mux
+	s.mux.Handle(p, otelhttp.WithRouteTag(p, http.StripPrefix(p, http.FileServer(s.assetFS))))
```
yurishkuro, line 83: s.mux is never initialized. Why does it need to be stored?

yurishkuro: This function is asked to create & return a handler, which happens to be implemented as Mux. Why does it need to be stored in `Server` instead of created locally here, as was done before?

examples/hotrod/pkg/tracing/http.go
```diff
@@ -18,45 +18,33 @@ package tracing
...
-	req, err := http.NewRequest(http.MethodGet, url, nil)
-	if err != nil {
-		return err
+func NewHTTPClient(tp trace.TracerProvider) HTTPClient {
+	return HTTPClient{
+		TracerProvider: tp,
+		Client:         &http.Client{Transport: &otelhttp.Transport{}},
 	}
-	req = req.WithContext(ctx)
-	req, ht := nethttp.TraceRequest(c.Tracer, req, nethttp.OperationName("HTTP GET: "+endpoint))
```
afzal442, line 43: Hi Yuri! I am not able to figure out how we can transform it to support otel tracer request.

yurishkuro (commented):
Q: why are we accumulating such a large PR? Can we not fix each service independently?

afzal442:
I see, I thought it could be done all at a time. But creating separate PR for each svc would be great though.

yurishkuro:
as long as we validate that each PR does not break the functionality (like baggage propagation).

examples/hotrod/pkg/tracing/mux.go
```diff
@@ -41,36 +40,16 @@ func NewServeMux(copyBaggage bool, tracer opentracing.Tracer, logger log.Factory
...
-		handler,
-		nethttp.OperationNameFunc(func(r *http.Request) string {
-			return "HTTP " + r.Method + " " + pattern
-		}),
-		// Jaeger SDK was able to accept `jaeger-baggage` header even for requests without am active trace.
-		// OTEL Bridge does not support that, so we use Baggage propagator to manually extract the baggage
-		// into Context (in otelBaggageExtractor handler below), and once the Bridge creates a Span,
-		// we use this SpanObserver to copy OTEL baggage from Context into the Span.
-		nethttp.MWSpanObserver(func(span opentracing.Span, r *http.Request) {
-			if !tm.copyBaggage {
```
afzal442, line 63: @yurishkuro I did workaround [here ](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/instrumentation/net/http/otelhttp/v0.42.0/instrumentation/net/http/otelhttp/handler.go#L64). I think otel sdk doesn't support for `jaeger-baggage` header. Anyway to handle such? 🤔

yurishkuro: Did you read about this part in my blog post? This isn't about jaeger-baggage header, but about the bridge not allowing baggage-only, without a valid span (which doesn't come from the front end).

afzal442: I see. I read that a few days ago. TBH, but made my assumptions on it. thanks
image

afzal442: underneath, I found that. Thanks

examples/hotrod/pkg/tracing/mux.go
```diff
@@ -40,37 +40,28 @@ func NewServeMux(copyBaggage bool, tracer opentracing.Tracer, logger log.Factory
 // TracedServeMux is a wrapper around http.ServeMux that instruments handlers for tracing.
 type TracedServeMux struct {
 	mux         *http.ServeMux
-	copyBaggage bool
-	tracer      opentracing.Tracer
+	tracer      trace.TracerProvider
 	logger      log.Factory
+	req         http.Request
```
yurishkuro, line 45: `TracedServeMux` is a type of HTTP Handler which deals with many requests, it cannot have a member variable like this.

examples/hotrod/pkg/tracing/mux.go
```diff
@@ -40,37 +40,28 @@ func NewServeMux(copyBaggage bool, tracer opentracing.Tracer, logger log.Factory
...
-			for _, m := range bag.Members() {
-				if b := span.BaggageItem(m.Key()); b == "" {
-					span.SetBaggageItem(m.Key(), m.Value())
-				}
-			}
-		}),
-	)
+	// otel.SetTextMapPropagator(propagation.NewCompositeTextMapPropagator(propagation.TraceContext{}, propagation.Baggage{}))
+
+	ctx := tm.req.Context()
```
yurishkuro, line 54: `Handle` function is called at initialization time, not at request handing time. You can put this logic into an HTTP handler, similar to what `otelBaggageExtractor` or `otelhttp.NewHandler` functions are doing. You just need to make sure the ordering is correct, because `middleware` from above was running _after_ the span was created.

It's also not clear to me if you even need this at all. The original reason I had to add `middleware` function was because the baggage was extracted from request via OTEL Propagator and placed into `Context`, but the rest of the code was using OpenTracing instrumentation and was expecting to get the baggage from the `opentracing.Span`, so `middleware` was doing a copy from Context to Span. Since we now are fully migrating to OTEL instrumentation, I think just `otelBaggageExtractor` will be sufficient (as long as we change the code that reads the baggage to read it from Context, not from Span).

We may also not even need `otelBaggageExtractor` eventually, since it was working around the limitation of the Bridge where it was not parsing baggage if the request had no tracing context (as happens between UI and `frontend`). However, with OTEL instrumentation baggage and span creation are decoupled, so I assume baggage will still be parsed correctly by the propagators into Context.

examples/hotrod/services/driver/client.go
```diff
@@ -30,25 +30,23 @@ import (
...
 // NewClient creates a new driver.Client
-func NewClient(tracer opentracing.Tracer, logger log.Factory, hostPort string) *Client {
+func NewClient(tracerProvider trace.TracerProvider, logger log.Factory, hostPort string) *Client {
 	conn, err := grpc.Dial(hostPort, grpc.WithTransportCredentials(insecure.NewCredentials()),
 		grpc.WithUnaryInterceptor(
-			otgrpc.OpenTracingClientInterceptor(tracer)),
+			otelgrpc.UnaryClientInterceptor(otelgrpc.WithTracerProvider(tracerProvider))),
 		grpc.WithStreamInterceptor(
-			otgrpc.OpenTracingStreamClientInterceptor(tracer)))
+			otelgrpc.StreamClientInterceptor(otelgrpc.WithTracerProvider(tracerProvider))))
```
yurishkuro, line 43: Are we still planning to split this PR into more pieces? I prefer to have PRs that deal with one problem at a time, and we have the baggage issue that might be more complicated across several services, so it's best to merge all the OTEL upgrades before tackling it.

examples/hotrod/pkg/tracing/mux.go
afzal442: Thanks Yuri. It was an great explanation. I am able to extract context from you now. 😄 
> Handle function is called at initialization time, not at request handling time. 

That's why we needed to call this `func (tm *TracedServeMux) Handle(...)` while starting the server.
> It's also not clear to me if you even need this at all.

That means we don't need to create this for baggage propagator https://github.com/jaegertracing/jaeger/blob/e427e196ec409ce577ae2acb1fc4e640f580f68f/examples/hotrod/pkg/tracing/mux.go#L54-L60 since we have OTEL instrumentation does this job and maybe not eventually `otelBaggageExtractor`.

examples/hotrod/services/driver/client.go
afzal442: I see. Let me grep it over and give you update here.

examples/hotrod/cmd/frontend.go
```diff
@@ -45,7 +45,7 @@ var frontendCmd = &cobra.Command{
 		logger := log.NewFactory(zapLogger)
 		server := frontend.NewServer(
 			options,
-			tracing.Init("frontend", otelExporter, metricsFactory, logger),
+			tracing.InitOTEL("frontend", otelExporter, metricsFactory, logger),
```
afzal442, line 48: @yurishkuro in order to separate out a PR, we'll have to create a duplicate instance of tracer here to be initiated by NewServer constructor.

afzal442: Unfortunately that way didn't work out for me. We are loosing span from context.

examples/hotrod/services/driver/client.go
afzal442: Sorry Yuri, I am not able to split this PR. We are getting stuck while considering `NewServer` creation.

examples/hotrod/pkg/tracing/http.go
```diff
@@ -18,45 +18,33 @@ package tracing
...
-	defer ht.Finish()
+}
 
-	res, err := c.Client.Do(req)
+// GetJSON executes HTTP GET against specified url and tried to parse
+// the response into out object.
+func (c *HTTPClient) GetJSON(ctx context.Context, url string, out interface{}) error {
+	resp, err := otelhttp.Get(ctx, url)
```
yurishkuro, line 43: this means your are not using the HTTPClient, nor the TracerProvider. `otelhttp.Get` is a bad design, let's not use it. You don't really need to change this GetJSON method, as long as in instantiate the Client as you do above with otel's Transport

examples/hotrod/pkg/tracing/http.go
```diff
@@ -18,45 +18,33 @@ package tracing
...
-// GetJSON executes HTTP GET against specified url and tried to parse
-// the response into out object.
-func (c *HTTPClient) GetJSON(ctx context.Context, endpoint string, url string, out interface{}) error {
-	req, err := http.NewRequest(http.MethodGet, url, nil)
-	if err != nil {
-		return err
+func NewHTTPClient(tp trace.TracerProvider) HTTPClient {
+	return HTTPClient{
+		TracerProvider: tp,
+		Client:         &http.Client{Transport: &otelhttp.Transport{}},
```
yurishkuro, line 36: you need to use NewTransport, not create the empty struct manually, and pass the Tracer as options.

examples/hotrod/pkg/tracing/mux.go
```diff
@@ -18,74 +18,39 @@ package tracing
...
-				if b := span.BaggageItem(m.Key()); b == "" {
-					span.SetBaggageItem(m.Key(), m.Value())
-				}
-			}
-		}),
-	)
-	tm.mux.Handle(pattern, otelBaggageExtractor(middleware))
+	middleware := otelhttp.NewHandler(handler, pattern,
+		otelhttp.WithTracerProvider(tm.tracer))
+	tm.mux.Handle(pattern, middleware)
```
yurishkuro, line 50: this looks right

examples/hotrod/services/customer/client.go
```diff
@@ -30,21 +28,16 @@ import (
 
 // Client is a remote client that implements customer.Interface
 type Client struct {
-	tracer   opentracing.Tracer
 	logger   log.Factory
-	client   *tracing.HTTPClient
+	client   tracing.HTTPClient
```
yurishkuro, line 32: what's the motivation for removing the pointer?

examples/hotrod/services/customer/server.go
```diff
@@ -57,8 +58,8 @@ func (s *Server) Run() error {
 }
 
 func (s *Server) createServeMux() http.Handler {
-	mux := tracing.NewServeMux(false, s.tracer, s.logger)
-	mux.Handle("/customer", http.HandlerFunc(s.customer))
+	mux := tracing.NewServeMux(s.tracer, s.logger)
+	mux.Handle("/customer", otelhttp.WithRouteTag("/customer", http.HandlerFunc(s.customer)))
```
yurishkuro, line 62: why do we need `otelhttp.WithRouteTag`? Why not do it inside TracedServerMux?

examples/hotrod/services/driver/client.go
yurishkuro: that's probably because you're introducing too many changes in the library functions, which I don't think are necessary.

examples/hotrod/services/frontend/best_eta.go
```diff
@@ -76,9 +77,8 @@ func (eta *bestETA) Get(ctx context.Context, customerID string) (*Response, erro
 	}
 	eta.logger.For(ctx).Info("Found customer", zap.Any("customer", customer))
 
-	if span := opentracing.SpanFromContext(ctx); span != nil {
-		span.SetBaggageItem("customer", customer.Name)
-	}
+	span := trace.SpanFromContext(ctx)
+	span.AddEvent("customer", trace.WithAttributes(attribute.Key(customer.Name).String(customer.Location)))
```
yurishkuro, line 81: this wasn't there before, why are you adding an event?

examples/hotrod/services/frontend/server.go
```diff
@@ -76,7 +76,7 @@ func (s *Server) Run() error {
 }
 
 func (s *Server) createServeMux() http.Handler {
-	mux := tracing.NewServeMux(true, s.tracer, s.logger)
+	mux := tracing.NewServeMux(s.tracer, s.logger)
```
yurishkuro, line 79: you could've kept the boolean parameter to minimize the changes and PR each server separately. Then at the end we can do a round of clean-up. As it stands now, you have a PR with a dozen different things going on.

examples/hotrod/services/route/client.go
```diff
@@ -58,7 +51,7 @@ func (c *Client) FindRoute(ctx context.Context, pickup, dropoff string) (*Route,
 	v.Set("dropoff", dropoff)
 	url := "http://" + c.hostPort + "/route?" + v.Encode()
 	var route Route
-	if err := c.client.GetJSON(ctx, "/route", url, &route); err != nil {
+	if err := c.client.GetJSON(ctx, url, &route); err != nil {
```
yurishkuro, line 54: another unnecessary change

examples/hotrod/pkg/tracing/http.go
```diff
@@ -22,26 +22,30 @@ import (
...
+		TracerProvider: tp,
+		Client:         &http.Client{Transport: otelhttp.NewTransport(http.DefaultTransport, otelhttp.WithTracerProvider(tp))},
+	}
 }
 
 // GetJSON executes HTTP GET against specified url and tried to parse
 // the response into out object.
-func (c *HTTPClient) GetJSON(ctx context.Context, endpoint string, url string, out interface{}) error {
-	req, err := http.NewRequest(http.MethodGet, url, nil)
+func (c *HTTPClient) GetJSON(ctx context.Context, url string, out interface{}) error {
```
yurishkuro, line 44: Please keep the `endpoint string` parameter. It will minimize the changes in this PR, which create unnecessary noise.

examples/hotrod/pkg/tracing/mux.go
```diff
@@ -41,51 +39,22 @@ func NewServeMux(copyBaggage bool, tracer opentracing.Tracer, logger log.Factory
...
-					span.SetBaggageItem(m.Key(), m.Value())
-				}
-			}
-		}),
-	)
-	tm.mux.Handle(pattern, otelBaggageExtractor(middleware))
+	handler = otelhttp.WithRouteTag(pattern, handler)
+
+	middleware := otelhttp.NewHandler(handler, pattern,
+		otelhttp.WithTracerProvider(tm.tracer))
```
yurishkuro, line 53:
```suggestion
	middleware := otelhttp.NewHandler(
		otelhttp.WithRouteTag(pattern, handler),
		pattern,
		otelhttp.WithTracerProvider(tm.tracer))
```

examples/hotrod/services/frontend/best_eta.go
```diff
@@ -76,10 +76,6 @@ func (eta *bestETA) Get(ctx context.Context, customerID string) (*Response, erro
 	}
 	eta.logger.For(ctx).Info("Found customer", zap.Any("customer", customer))
 
-	if span := opentracing.SpanFromContext(ctx); span != nil {
-		span.SetBaggageItem("customer", customer.Name)
-	}
```
yurishkuro, line 81: there needs to be an equivalent of this in OTEL baggage syntax

examples/hotrod/pkg/tracing/http.go
```diff
@@ -22,26 +22,30 @@ import (
...
-	Tracer opentracing.Tracer
-	Client *http.Client
+	TracerProvider trace.TracerProvider
+	Client         *http.Client
+}
+
+func NewHTTPClient(tp trace.TracerProvider) *HTTPClient {
+	return &HTTPClient{
+		TracerProvider: tp,
+		Client:         &http.Client{Transport: otelhttp.NewTransport(http.DefaultTransport, otelhttp.WithTracerProvider(tp))},
```
yurishkuro, line 38: avoid long lines. Without formatting it's difficult to match which params to where, one has to count parentheses
```suggestion
		Client:         &http.Client{
			Transport: otelhttp.NewTransport(
				http.DefaultTransport, 
				otelhttp.WithTracerProvider(tp),
			),
		},
```

afzal442: Ah sorry! next time.

examples/hotrod/services/frontend/best_eta.go
```diff
@@ -76,9 +78,10 @@ func (eta *bestETA) Get(ctx context.Context, customerID string) (*Response, erro
...
+	span := trace.SpanFromContext(ctx)
+	bag := baggage.FromContext(ctx)
+	m := bag.Member(customer.Name)
+	span.SetAttributes(attribute.Key(m.Key()).String(m.Value()))
```
yurishkuro, line 84: you're doing it backwards. The original code was setting _baggage item_, making it available downstream. You're _reading_ baggage (that nobody sets) and storing it as a span attribute (which is a form of logging but has no runtime impact on the downstream).

afzal442: I see. I really stumbled upon that.

examples/hotrod/services/frontend/best_eta.go
```diff
@@ -76,9 +77,11 @@ func (eta *bestETA) Get(ctx context.Context, customerID string) (*Response, erro
...
 	eta.logger.For(ctx).Info("Found customer", zap.Any("customer", customer))
 
-	if span := opentracing.SpanFromContext(ctx); span != nil {
-		span.SetBaggageItem("customer", customer.Name)
-	}
+	/* bag := baggage.FromContext(ctx)
+	m := bag.Member("customer") */
+	m, _ := baggage.NewMember("customer", customer.Name)
+	bag, _ := baggage.New(m)
+	ctx = baggage.ContextWithBaggage(ctx, bag)
```
yurishkuro, line 84: +1. Please remove commented code

afzalbin64: @yurishkuro I am not able to make sure if I get it correct. LMK how I can test this out.

yurishkuro: you should be able to see metrics by customer in `/debug/vars`

afzalbin64: Yes! but when I try to inspect `m.Key()` or `m.Value()`, there is null value being appeared. Even it is not for `session` baggage ctx as well. But `session.sec` works out.

yurishkuro: you're ignoring the errors. If you log them, you will see this:
```
2023-07-12T14:58:36.008-0400	ERROR	frontend/best_eta.go:83	invalid value: "Trom Chocolatier"	{"service": "frontend", "trace_id": "a2cba41c7586b75f76e80cee7d122818", "span_id": "f6a2a37ee8b39539"}
2023-07-12T14:58:36.008-0400	ERROR	frontend/best_eta.go:88	invalid baggage list-member	{"service": "frontend", "trace_id": "a2cba41c7586b75f76e80cee7d122818", "span_id": "f6a2a37ee8b39539"}
```

examples/hotrod/services/frontend/best_eta.go
```diff
@@ -76,9 +77,10 @@ func (eta *bestETA) Get(ctx context.Context, customerID string) (*Response, erro
 	}
 	eta.logger.For(ctx).Info("Found customer", zap.Any("customer", customer))
 
-	if span := opentracing.SpanFromContext(ctx); span != nil {
-		span.SetBaggageItem("customer", customer.Name)
-	}
+	m, _ := baggage.NewMember("customer", customer.Name)
+	bag := baggage.FromContext(ctx)
+	bag, _ = bag.SetMember(m)
```
yurishkuro, line 82:
```suggestion
	m, err := baggage.NewMember("customer", customer.Name)
	if err != nil {
		eta.logger.For(ctx).Error("cannot create baggage member", zap.Error(err))
	}
	bag := baggage.FromContext(ctx)
	bag, err = bag.SetMember(m)
	if err != nil {
		eta.logger.For(ctx).Error("cannot set baggage member", zap.Error(err))
	}
```

afzal442: Cool! ignoring `err` costs me here.

afzal442: @yurishkuro after deep investigation, it came out working with the response in the HTTP request. e.g. ```Loading customer      {"service": "customer", "component": "mysql", "trace_id": "63718f14ff8d04607e60d95da54784ae", "span_id": "b60a6202006008bc", "customer_id": "567"}``` here, when I try to use any of the items included say "span_id" or customer.ID or customerID or "mysql", I am getting baggage items out of it. Make sure the service should be "customer".

afzal442: Most insane. when I try to add logs into it, it's still not working. `INFO    customer/database.go:86 Getting customer      {"service": "customer", "component": "mysql", "trace_id": "49af5501b6c5c75acfb1e69d19f1e481", "span_id": "e9e53f530cb6aacc", "customer_name": "Amazing Coffee Roasters"}`

afzal442: after making minor changes, I got sth like below
```
"route.calc.by.customer.sec": {"Amazing_Coffee_Roasters": 0.498, "Japanese_Desserts": 0.486},
"route.calc.by.session.sec": {"1581": 0.498, "7406": 0.486}
```

yurishkuro: Yes, that's the bug it OTEL - users are not supposed to mangle the values like this in order to put them into baggage.

yurishkuro (approved):
🎉 🎉 🎉

PR #4549, opened 2023-06-23
[WIP] Enabling CORS settings on OTLP Reciever HTTP endpoint

### LINKED ISSUE #4459: [Feature]: Support CORS headers on OTEL HTTP endpoint

pmuls99:
-  Resolves #4459 

- Allowed headers and Allowed origins as a slice of strings as input
- Added the configuration option in the OTLP reciever file


pmuls99:
@yurishkuro , I still have some changes to make to the PR. If I am on the right track to solve the issue, I would make those changes with the next commit.

cmd/collector/app/flags/flags.go
```diff
@@ -202,6 +207,8 @@ func addHTTPFlags(flags *flag.FlagSet, cfg serverFlagsConfig, defaultHostPort st
 	flags.Duration(cfg.prefix+"."+flagSuffixHTTPIdleTimeout, 0, "See https://pkg.go.dev/net/http#Server")
 	flags.Duration(cfg.prefix+"."+flagSuffixHTTPReadTimeout, 0, "See https://pkg.go.dev/net/http#Server")
 	flags.Duration(cfg.prefix+"."+flagSuffixHTTPReadHeaderTimeout, 2*time.Second, "See https://pkg.go.dev/net/http#Server")
+	flags.String(cfg.prefix+"."+flagSuffixHTTPAllowedHeaders, "content-type", "Allowed headers for the OTLP HTTP port , default content-type")
+	flags.String(cfg.prefix+"."+flagSuffixHTTPAllowedOrigins, "*", "Allowed origins for the OTLP HTTP port , default accepts all")
```
yurishkuro, line 211: This could add these flags at the top level, e.g. `--collector.allowed-headers`, which is not what we want.

I would rather do a little bit of refactoring and create a new package `pkg/config/corscfg`, similar to `tlscfg`, which will define a shared structure & flags for these parameters. We can then use the new package in both Zipkin endpoint and OTLP endpoint.

We can also reuse OTEL's `confighttp.CORSSettings` as the underlying structure for these params.

cmd/collector/app/flags/flags.go
```diff
@@ -153,6 +156,8 @@ type HTTPOptions struct {
 	ReadHeaderTimeout time.Duration
 	// IdleTimeout sets the respective parameter of http.Server
 	IdleTimeout time.Duration
+	// Allowed Origins
```
yurishkuro, line 159: unnecessary comment

cmd/collector/app/flags/flags.go
```diff
@@ -153,6 +156,8 @@ type HTTPOptions struct {
...
+	CORSSettings *confighttp.CORSSettings
```
yurishkuro, line 160:
```suggestion
	CORSSettings confighttp.CORSSettings
```

cmd/collector/app/flags/flags.go
pmuls99: Flags at the top level as in while calling the flag or internally? because addHTTPFlags is a common function for both collector http port and the otlp server http port and is called with respective server flags. Correct me if I am wrong.

Yes refactoring in a similar way like tlscfg seems like a nice decision will get on the work with that!

pmuls99:
@yurishkuro , worked on what mistake there was in the code. Could you please review the commit and tell if there is anything to improve/change ?

pkg/config/corscfg/flags.go
```diff
@@ -0,0 +1,43 @@
...
+type CorsFlagsConfig struct {
+	Prefix string
+}
+
+func (c CorsFlagsConfig) AddFlags(flags *flag.FlagSet) {
+	flags.String(c.Prefix+corsAllowedHeaders, "content-type", "Allowed headers for the HTTP port , default content-type")
+	flags.String(c.Prefix+corsAllowedOrigins, "*", "Allowed origins for the HTTP port , default accepts all")
+}
+
+func (c CorsFlagsConfig) InitFromViper(v *viper.Viper) (Settings, error) {
```
yurishkuro, line 38: why return an error if it can never happen?

yurishkuro (commented):
directionally lgtm

pmuls99:
@yurishkuro , have made the suggested changes. Also we need to update the docs since the flags for zipkin server have changed !

cmd/collector/app/server/zipkin.go
```diff
@@ -39,8 +39,7 @@ type ZipkinServerParams struct {
 	TLSConfig      tlscfg.Options
 	HostPort       string
 	Handler        handler.ZipkinSpansHandler
-	AllowedOrigins string
-	AllowedHeaders string
+	CORSSettings   corscfg.Options
```
yurishkuro, line 42: to keep the existing naming conventions:
```suggestion
	CORSConfig   corscfg.Options
```

pkg/config/corscfg/flags.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+func (c Flags) AddFlags(flags *flag.FlagSet) {
+	flags.String(c.Prefix+corsAllowedHeaders, "content-type", "Allowed headers for the HTTP port , default content-type")
+	flags.String(c.Prefix+corsAllowedOrigins, "*", "Allowed origins for the HTTP port , default accepts all")
+}
+
+func (c Flags) InitFromViper(v *viper.Viper) Options {
+	var p Options
+	p.AllowedHeaders = v.GetStringSlice(c.Prefix + corsAllowedHeaders)
```
yurishkuro, line 40: I see that previously we were stripping whitespace
```
strings.Split(strings.ReplaceAll(params.AllowedOrigins, " ", ""), ",")
```
Is it going to be done automatically by `viper.GetStringSlice`? If not it's not a backwards compatible change.

yurishkuro: can you add tests to this package? it's small and has no external deps, no reason not to write tests, and in particular validate ^^^

cmd/collector/app/collector.go
```diff
@@ -146,8 +146,7 @@ func (c *Collector) Start(options *flags.CollectorOptions) error {
 		Handler:        c.spanHandlers.ZipkinSpansHandler,
 		TLSConfig:      options.Zipkin.TLS,
 		HealthCheck:    c.hCheck,
-		AllowedHeaders: options.Zipkin.AllowedHeaders,
-		AllowedOrigins: options.Zipkin.AllowedOrigins,
+		CORSSettings:   options.Zipkin.CORSSettings,
```
yurishkuro, line 149:
```suggestion
		CORSConfig:   options.Zipkin.CORSSettings,
```

pkg/config/corscfg/flags.go
pmuls99: I just was checking with you if i was going directionally right ! Will write the tests and validate !

pmuls99: I was just checking with you if I was going in the right direction , since I am new to the community. I will write the tests and validate soon, sorry for the delay!

PR #4550, opened 2023-06-25
OTEL span logging in HotRod example

albertteoh:
## Description

This is a throw-away draft to illustrate the changes necessary to get logging working for the HotRod example, after switching the `driver.Server` to use the OTEL tracer.

The key to the problem is using the `otelgrpc.UnaryServerInterceptor` during the switch which, AFAIK, when extracting the span context, inserts an OTEL span into the context.

This means that the `logger.For(ctx)` lookup for the span will fail to find an OT span:

```go
if otSpan := ot.SpanFromContext(ctx); otSpan != nil {
```

and return the default stdout logger, which explains why the logs are visible in the terminal but not in spans when run locally.

This PR adds an `otelSpanLogger`, which is almost identical to the OT `spanLogger`, with a different encoding (incomplete) implementation. When extracted from the context, it will successfully log to the span as an event and surface as a log entry in Jaeger UI:

Screenshot 2023-06-25 at 9 53 20 pm

Note, just the bare minimum and least-effort changes were introduced, and so should not be simply copy-pasted as a solution, but rather, be carefully considered before incorporating. Little care was taken for code correctness, style, etc.

The changes from this PR are made on top of https://github.com/jaegertracing/jaeger/pull/4548, so there may be some unrelated changes present.


examples/hotrod/pkg/log/factory.go
```diff
@@ -53,7 +53,14 @@ func (b Factory) For(ctx context.Context) Logger {
 				zap.String("span_id", otelSpan.SpanContext().SpanID().String()),
 			}
 		}
-
+		return logger
+	}
+	if otelSpan := trace.SpanFromContext(ctx); otelSpan != nil {
```
albertteoh, line 58: A hack to fallback to trying to fetch the OTEL span if trying to fetch the OT span fails.

afzal442: Thanks @albertteoh to hacking around this. Supposedly, I thought we needed this. You made it. Awesome! 💯

yurishkuro (commented):
I think this reveals one gotcha we will need to work around everywhere - when plain OTEL is used upper in the call chain, the OT instrumentation in lower layers will not find the span (OTEL->OT), while the reverse is apparently working (OT->OTEL). This is something to point out in the blog, and also in the bridge README.

examples/hotrod/pkg/log/factory.go
```diff
@@ -53,7 +53,14 @@ func (b Factory) For(ctx context.Context) Logger {
 				zap.String("span_id", otelSpan.SpanContext().SpanID().String()),
 			}
 		}
-
+		return logger
+	}
```
yurishkuro, line 57: I don't recall where L50-55 came from, they seem to do something similar.

examples/hotrod/pkg/log/spanlogger.go
```diff
@@ -16,15 +16,244 @@
...
 	"go.uber.org/zap/zapcore"
 )
 
+type otelSpanLogger struct {
+	logger     *zap.Logger
+	span       trace.Span
+	spanFields []zapcore.Field
+}
+
+func (sl otelSpanLogger) Debug(msg string, fields ...zapcore.Field) {
```
yurishkuro, line 38: These methods do not directly depend on underlying span so we could keep a single spanLogger type and only differentiate lower in the implementation which span it is dealing with.

examples/hotrod/pkg/log/spanlogger.go
```diff
@@ -16,15 +16,244 @@
...
+	"go.opentelemetry.io/otel/trace"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
 )
 
+type otelSpanLogger struct {
+	logger     *zap.Logger
+	span       trace.Span
+	spanFields []zapcore.Field
+}
```
yurishkuro, line 36: Another thing that's worth looking into is this: when the Context contains an OT span generated by OTEL Bridge, is there a way to access the real OTEL Span that it wraps? Because if so, rather than maintaining two implementations here we can simply upgrade this spanLogger to only work with OTEL API.

yurishkuro: Namely, if they don't have it already, I recommend creating a PR for the bridge to implement an Unwrap method:
```go
func Unwrap(opentracing.Span) (trace.Span, error)
```

https://github.com/open-telemetry/opentelemetry-go/blob/dc187e74af528ce081811c4eae17a7930be1232b/bridge/opentracing/bridge.go#L89

yurishkuro: Doesn't look like they have a public method for that https://pkg.go.dev/go.opentelemetry.io/otel/bridge/opentracing

albertteoh: It looks like we can keep a single spanLogger because when writing the span to the context, its tracer is obtained and, because it's a bridge tracer, it's able to inject the span in the "OTEL-way" with the correct context key and OTEL span interface. See: https://github.com/opentracing/opentracing-go/blob/master/gocontext.go#L14.

I've opened another PR that takes advantage of this fact: https://github.com/jaegertracing/jaeger/pull/4556.

albertteoh:
Superseded by https://github.com/jaegertracing/jaeger/pull/4556.

PR #4553, opened 2023-06-26
Rebuild jaeger-ui if the tree does not match any tag exactly

bobrik:
The previous behavior resulted in the prebuild upstream package being downloaded, even if the local tree has any changes, whether they are commited or not. This broke our internal build that has some patches applied. See:

* https://github.com/jaegertracing/jaeger/pull/4251#issuecomment-1608393236

Let's use a more strict approach and only use the prebuild package if the tag is exact and does not contain any extra changes.

### DIFF
--- a/scripts/rebuild-ui.sh
+++ b/scripts/rebuild-ui.sh
@@ -4,13 +4,13 @@ set -euxf -o pipefail
 
 cd jaeger-ui
 
-LAST_TAG=$(git describe --abbrev=0 --tags 2>/dev/null)
-BRANCH_HASH=$(git rev-parse HEAD)
-LAST_TAG_HASH=$(git rev-parse $LAST_TAG)
+LAST_TAG=$(git describe --tags --dirty 2>/dev/null)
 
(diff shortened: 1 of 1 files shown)


yurishkuro:
@bobrik could you elaborate a bit how your internal build is setup? Do you have a fork of both Jaeger and Jaeger-UI.

bobrik:
Yes, we have patches for both `jaeger` and `jaeger-ui` internally, so we build our own binaries for deployment.

yurishkuro:
I wonder if it makes more sense for you to have an option that can be passed to `make` that will skip the tag check & download and always go with full rebuild.

bobrik:
The only scenario where I can see it being useful is you want to rebuild with a changed environment (like node version), everything else _should_ be covered by this change already. For this a bit more finessing might be required (I remember needing to nuke `node_modules` to make it build for me) and if you are in that territory, you probably can run `yarn build` directly.

I can add `make rebuild-ui` that would pass an env variable to `./scripts/rebuild-ui.sh` to skip the download, but I'm not sure if it's something anyone would actually use.

yurishkuro (approved):

PR #4556, opened 2023-07-01
Fix OTEL logging in HotRod example

albertteoh:
- A cleaned up version of https://github.com/jaegertracing/jaeger/pull/4550.
- It uses a single spanLogger for both OpenTracing and OpenTelemetry spans.
  - The logging works for both types spans because of [this clever line of code](https://github.com/opentracing/opentracing-go/blob/master/gocontext.go#L14) where `tracerWithHook` is an OTEL bridge tracer, allowing span to be stored in the `ctx` the "OTEL-way", making the span discoverable as a legitimate OTEL span such that `ok == true`:

```go
if span, ok := ctx.Value(currentSpanKey).(Span); ok { ... }
```

## Testing

Ran locally to confirm logs appearing for both OpenTracing spans:

Screenshot 2023-07-01 at 8 12 16 pm

and OpenTelemetry spans:
Screenshot 2023-07-01 at 8 12 08 pm


examples/hotrod/pkg/log/spanlogger.go
```diff
@@ -16,39 +16,40 @@
...
 
 type spanLogger struct {
 	logger     *zap.Logger
-	span       opentracing.Span
+	span       trace.Span
 	spanFields []zapcore.Field
 }
 
 func (sl spanLogger) Debug(msg string, fields ...zapcore.Field) {
-	sl.logToSpan("Debug", msg, fields...)
```
albertteoh, line 35: The log level is not used.

examples/hotrod/pkg/log/spanlogger.go
```diff
@@ -57,99 +58,164 @@ func (sl spanLogger) With(fields ...zapcore.Field) Logger {
...
+// otTagToOTelAttr converts given key-value into attribute.KeyValue.
+// Note that some conversions are not obvious:
+// - int -> int64
+// - uint -> string
+// - int32 -> int64
+// - uint32 -> int64
+// - uint64 -> string
+// - float32 -> float64
+//
+// See: https://github.com/open-telemetry/opentelemetry-go/blob/main/bridge/opentracing/bridge.go#L538
```
albertteoh, line 186: As the comment states, the mappings below are taken from the OTEL bridge to maintain consistency.

examples/hotrod/pkg/log/spanlogger.go
yurishkuro: Maybe we can add it as attribute?

examples/hotrod/pkg/log/spanlogger.go
```diff
@@ -57,99 +58,164 @@ func (sl spanLogger) With(fields ...zapcore.Field) Logger {
 	return spanLogger{logger: sl.logger.With(fields...), span: sl.span, spanFields: sl.spanFields}
 }
 
-func (sl spanLogger) logToSpan(level string, msg string, fields ...zapcore.Field) {
-	// TODO rather than always converting the fields, we could wrap them into a lazy logger
-	fa := fieldAdapter(make([]log.Field, 0, 2+len(fields)))
-	fa = append(fa, log.String("event", msg))
-	fa = append(fa, log.String("level", level))
```
yurishkuro, line 64: Level was used!

examples/hotrod/pkg/log/spanlogger.go
```diff
@@ -57,99 +58,164 @@ func (sl spanLogger) With(fields ...zapcore.Field) Logger {
...
+func (sl spanLogger) logToSpan(msg string, fields ...zapcore.Field) {
+	sl.span.AddEvent(
+		msg,
+		trace.WithAttributes(otLogFieldsToOTelAttrs(fields)...),
+	)
+}
+
+func otLogFieldsToOTelAttrs(fields []zapcore.Field) []attribute.KeyValue {
```
yurishkuro, line 68:
```suggestion
func logFieldsToOTelAttrs(fields []zapcore.Field) []attribute.KeyValue {
```

examples/hotrod/pkg/log/spanlogger.go
```diff
@@ -57,99 +58,164 @@ func (sl spanLogger) With(fields ...zapcore.Field) Logger {
...
+}
+
+func (e *bridgeFieldEncoder) AddObject(key string, marshaler zapcore.ObjectMarshaler) error {
+	e.pairs = append(e.pairs, otTagToOTelAttr(key, marshaler))
+	return nil
 }
 
-type fieldAdapter []log.Field
+func (e *bridgeFieldEncoder) AddBinary(key string, value []byte) {
+	e.pairs = append(e.pairs, otTagToOTelAttr(key, value))
```
yurishkuro, line 91: Since we know the type here, why do we call into method that erases the type and has to match on it at runtime? We could call the right otel attribute constructor right here.

albertteoh: Just out of convenience, particularly for the "unsupported" types which fall back to the `default` case; although I like the fact that there's less indirection, so I've made direct calls in: [b1d21a1](https://github.com/jaegertracing/jaeger/pull/4556/commits/b1d21a13dee7d7083ce6873a0f0c1bd9d4557bba).

examples/hotrod/pkg/log/spanlogger.go
albertteoh: Ha! Yes, you're right, my bad for this oversight and thanks for the callout. Added "level" as an attribute to the logs in: [a31b067](https://github.com/jaegertracing/jaeger/pull/4556/commits/a31b0679180f7e62c1576ccbb2cfe7903d0d0da5)

Note that the MySQL spans are missing the log level:

Screenshot 2023-07-02 at 7 20 44 am

That's because `LogFields` is invoked directly from the active span here: https://github.com/jaegertracing/jaeger/blob/343fcc6ddd3bdb94dec864bddc2ff8794d6666a8/examples/hotrod/pkg/tracing/mutex.go#L50.

Is this something we should fix in this PR?

examples/hotrod/pkg/log/spanlogger.go
albertteoh: Yup, added this in: a31b067

examples/hotrod/pkg/log/spanlogger.go
albertteoh: Renamed in: [b1d21a1](https://github.com/jaegertracing/jaeger/pull/4556/commits/b1d21a13dee7d7083ce6873a0f0c1bd9d4557bba)

examples/hotrod/pkg/log/spanlogger.go
yurishkuro: looks fine as is. Level is not meaningful in the span logs, the only reason to transfer it when converting real logs is to keep data fidelity.

yurishkuro (approved):

PR #4574, opened 2023-07-07
Replace Jaeger SDK with OTEL SDK + OT Bridge

afzal442:
* Part of #3381 
* This PR adds OTEL Provider to jtracer pkg

- Upgrades jaeger tracing to support `otel tracer`


pkg/jtracer/jtracer.go
```diff
@@ -14,16 +14,57 @@
...
-	OT opentracing.Tracer
+	OT   opentracing.Tracer
+	OTEL trace.TracerProvider
 }
 
 func OT(t opentracing.Tracer) JTracer {
 	return JTracer{OT: t}
 }
 
+func OTEL(tp trace.TracerProvider) JTracer {
```
albertteoh, line 39: I prefer a single `New(t opentracing.Tracer, tp trace.TracerProvider) JTracer` constructor to keep things simple; you can pass in noop variants of each tracer if you like, in tests. That means we can remove the `NoOp()` and `NoOpOtel()` funcs.

In fact, if we're exporting the `OT` `OTEL` fields, I don't think we even need a constructor func which will make things even simpler. I only see a need for it if we unexport the `JTracer` fields.

Thoughts?

yurishkuro: I would keep whichever ctors are currently convenient, and clean-up after the migration is completed. We'd have to do the final clean-up anyway, since we'd want to eliminate any OT references.

pkg/jtracer/jtracer.go
```diff
@@ -14,16 +14,57 @@
...
+		return nil, fmt.Errorf("failed to create exporter: %w", err)
+	}
+
+	tracerProvider := sdktrace.NewTracerProvider(
+		sdktrace.WithBatcher(exporter),
+		sdktrace.WithSampler(sdktrace.AlwaysSample()),
+	)
+
+	otel.SetTracerProvider(tracerProvider)
+	otel.SetTextMapPropagator(propagation.NewCompositeTextMapPropagator(propagation.TraceContext{}, propagation.Baggage{}))
```
yurishkuro, line 67:
```suggestion
	otel.SetTextMapPropagator(
		propagation.NewCompositeTextMapPropagator(
			propagation.TraceContext{}, 
			propagation.Baggage{},
		)
	)
```

pkg/jtracer/jtracer.go
```diff
@@ -14,16 +14,57 @@
...
+		context.Background(),
+		otlptracehttp.NewClient(opts...),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create exporter: %w", err)
+	}
+
+	tracerProvider := sdktrace.NewTracerProvider(
+		sdktrace.WithBatcher(exporter),
+		sdktrace.WithSampler(sdktrace.AlwaysSample()),
```
yurishkuro, line 63: two considerations here
  * we should allow the exact settings to be customizable, e.g. by relying on OTEL's env vars
  * we should also have some sensible defaults when no env vars are given

plugin/metrics/prometheus/metricsstore/reader.go
```diff
@@ -287,17 +289,20 @@ func promqlDurationString(d *time.Duration) string {
 	return string(b)
 }
 
-func startSpanForQuery(ctx context.Context, metricName, query string) (opentracing.Span, context.Context) {
-	span, ctx := opentracing.StartSpanFromContext(ctx, metricName)
-	ottag.DBStatement.Set(span, query)
-	ottag.DBType.Set(span, "prometheus")
-	ottag.Component.Set(span, "promql")
+func startSpanForQuery(ctx context.Context, metricName, query string) (trace.Span, context.Context) {
+	tp, _ := jtracer.NewOTELProvider()
```
yurishkuro, line 293: provider is a (expensive) singleton, you cannot instantiate it for every span

plugin/metrics/prometheus/metricsstore/reader.go
```diff
@@ -287,17 +289,20 @@ func promqlDurationString(d *time.Duration) string {
 	return string(b)
 }
 
-func startSpanForQuery(ctx context.Context, metricName, query string) (opentracing.Span, context.Context) {
-	span, ctx := opentracing.StartSpanFromContext(ctx, metricName)
-	ottag.DBStatement.Set(span, query)
-	ottag.DBType.Set(span, "prometheus")
-	ottag.Component.Set(span, "promql")
+func startSpanForQuery(ctx context.Context, metricName, query string) (trace.Span, context.Context) {
```
yurishkuro, line 292: follow the OTEL syntax/convention here:
```suggestion
func startSpanForQuery(ctx context.Context, metricName, query string) (context.Context, trace.Span) {
```

plugin/metrics/prometheus/metricsstore/reader.go
```diff
@@ -287,17 +289,20 @@ func promqlDurationString(d *time.Duration) string {
...
+	tp, _ := jtracer.NewOTELProvider()
+	ctx, span := tp.Tracer("").Start(ctx, metricName)
```
yurishkuro, line 294: tracer must be named

pkg/jtracer/jtracer.go
yurishkuro: having said that, I would recommend only having `New() JTracer` and let it create both as needed, instead of forcing main's here and there create those manually (and inconsistently)

albertteoh: > I would recommend only having New() JTracer and let it create both as needed

That sounds reasonable. For tests, would we still need to do: `JTracer{OT: opentracing.NoopTracer{}}` (for example)?

yurishkuro: we should be aiming to minimize changes and churn. Tests should be able to call NoOp(), but why do they need to distinguish between NoOpOtel and OT? A single function can create JTracer with both sub-tracers initialized to noop.

pkg/jtracer/jtracer.go
```diff
@@ -14,16 +14,57 @@
...
 
+func OTEL(tp trace.TracerProvider) JTracer {
+	return JTracer{OTEL: tp}
+}
+
 func NoOp() JTracer {
 	return JTracer{OT: opentracing.NoopTracer{}}
 }
+
+func NoOpOtel() JTracer {
```
yurishkuro, line 47: So we certainly want to follow the principle of doing small PRs that perform one focused change. But this PR introduces changes like this function that is not actually used. Why do we need it? How do we know having this function is a good thing if we're not using it?

afzal442: I see. I made it in case we need it for future. For now, I don't see its use. I am going to revert it.

pkg/jtracer/jtracer.go
afzal442: Hi, I could create `New() JTracer` constructor on top of Jtracer struct. But there is a lot of files being addressed by func Jtracer.OT().

albertteoh: The question in my mind is, will there be a situation where we need both otel and ot tracers? If so, then we'll need the New(). If not, I think it's okay to keep OT() and OTEL() constructors.

albertteoh: Also worth thinking about testability. Do we want to inject a mock tracer provider to assert the spans produced? Or am I overthinking things? If so, we may need to consider dependency injection. 

Thoughts?

yurishkuro: In order to make incremental changes, we will need a period where the OT and OTEL instrumentations coexist, meaning we'll need to use the OTEL bridge which can satisfy both OT and OTEL interfaces. So it's still a single New function that returns connected tracers.

As for testing, New can only be called in main (which we don't test), but everything else will be getting JTracer as a dependency, which can be manipulated for testing (usually by installing an exporter that accumulates the data and allows to inspect it).

pkg/jtracer/jtracer.go
```diff
@@ -14,16 +14,65 @@
...
 
 type JTracer struct {
-	OT opentracing.Tracer
+	OT   opentracing.Tracer
+	OTEL trace.TracerProvider
 }
 
-func OT(t opentracing.Tracer) JTracer {
-	return JTracer{OT: t}
+var globalTracerProvider trace.TracerProvider
```
yurishkuro, line 37: No globals please, it's always a code smell

pkg/jtracer/jtracer.go
```diff
@@ -14,16 +14,65 @@
...
+
+func New(t opentracing.Tracer, tp trace.TracerProvider) JTracer {
```
yurishkuro, line 39: My suggestion was no-args New(). It can create all the necessary pieces, similar to tracing.Init in hotrod

pkg/jtracer/jtracer.go
```diff
@@ -14,16 +14,65 @@
...
+	globalTracerProvider = sdktrace.NewTracerProvider(
+		sdktrace.WithSpanProcessor(bsp),
+		sdktrace.WithResource(resource.NewWithAttributes(
+			semconv.SchemaURL,
+			semconv.ServiceNameKey.String("otlp"),
+		)),
+	)
+
+	otel.SetTracerProvider(globalTracerProvider)
+	otel.SetTextMapPropagator(
```
yurishkuro, line 69: This part is a necessary evil, but I prefer to draw the line at it as far as globals. Let's not rely on global tracer provider. If you do, it's hard to unit test components that rely on that global.

pkg/jtracer/jtracer.go
```diff
@@ -14,16 +14,83 @@
...
+	OT   opentracing.Tracer
+	OTEL trace.TracerProvider
 }
 
-func OT(t opentracing.Tracer) JTracer {
-	return JTracer{OT: t}
+var once sync.Once
+
+func New() JTracer {
+	return JTracer{OT: initOT(), OTEL: initOTEL()}
```
yurishkuro, line 42: You're creating the same thing twice here

pkg/jtracer/jtracer.go
```diff
@@ -14,16 +14,83 @@
...
+
+	once.Do(func() {
+		otel.SetTextMapPropagator(
+			propagation.NewCompositeTextMapPropagator(
+				propagation.TraceContext{},
+				propagation.Baggage{},
+			))
+	})
+
+	otTracer, _ := otbridge.NewTracerPair(tracerProvider.Tracer(""))
```
yurishkuro, line 93: What is the 2nd value you're ignoring here? This is prod code, not sample like hotrod, we try not to ignore these.

Also, how are you going to shutdown them? Jaeger used to return Closer, OTEL might have close as a method, but your main in not calling close.

plugin/metrics/prometheus/metricsstore/reader.go
```diff
@@ -287,17 +288,20 @@ func promqlDurationString(d *time.Duration) string {
...
 }
 
-func startSpanForQuery(ctx context.Context, metricName, query string) (opentracing.Span, context.Context) {
-	span, ctx := opentracing.StartSpanFromContext(ctx, metricName)
-	ottag.DBStatement.Set(span, query)
-	ottag.DBType.Set(span, "prometheus")
-	ottag.Component.Set(span, "promql")
-	return span, ctx
+func startSpanForQuery(ctx context.Context, metricName, query string) (context.Context, trace.Span) {
+	tp := jtracer.New().OTEL
```
yurishkuro, line 292: We never init tracers in lower level components, only in main.

afzal442: Sorry @yurishkuro , any way to initialise it. I mean `package main` ?

yurishkuro: you already added code to initialize it in main. You just need to pass it down to components, similar to other dependencies like metrics factory or logger.

plugin/metrics/prometheus/metricsstore/reader.go
```diff
@@ -101,6 +103,7 @@ func NewMetricsReader(logger *zap.Logger, cfg config.Configuration) (*MetricsRea
 		callsMetricName:   buildFullCallsMetricName(cfg),
 		latencyMetricName: buildFullLatencyMetricName(cfg),
 		operationLabel:    operationLabel,
+		tracer:            jtracer.New(),
```
afzal442, line 106: @yurishkuro Should I be able to pass it through func NewMetricsReader(...tracer trace.Trace) ?

yurishkuro: yes

yurishkuro: you can pick another component to validate the jtracer changes, there are components where JTracer is already passed in.

afzalbin64: Sorry! I couldn't get that. Can I use `GetLatencies(...)` method to initiate that?

yurishkuro: I mean use try to change other components than metricsstore.

yurishkuro: For example, you're passing JTracer here: `cmd/query/app/grpc_handler.go`, but later extracting OT Tracer from it - why not change that path to use OTEL TracerProvider instead?

pkg/jtracer/jtracer.go
```diff
@@ -14,16 +14,77 @@
...
+
+// initBoth initializes OpenTelemetry SDK and uses OTel-OpenTracing Bridge
+func initBoth() (opentracing.Tracer, trace.TracerProvider) {
+	opts := []otlptracehttp.Option{otlptracehttp.WithInsecure()}
+	traceExporter, err := otlptrace.New(
+		context.Background(),
+		otlptracehttp.NewClient(opts...),
+	)
+	if err != nil {
+		log.Fatal("failed to create exporter: %w", err)
```
yurishkuro, line 58: the default behavior on errors should be to wrap and return them, not log them. Some caller upstream may decide to log (and is more likely to have access to the logger).

pkg/jtracer/jtracer.go
```diff
@@ -14,16 +14,77 @@
...
+				propagation.TraceContext{},
+				propagation.Baggage{},
+			))
+	})
+
+	// Use the bridgeTracer as your OpenTracing tracer(otTrace).
+	otTracer, wrapperTracerProvider := otbridge.NewTracerPair(tracerProvider.Tracer(""))
+
+	// Shutdown the tracerProvider to clean up resources
+	err = tracerProvider.Shutdown(context.Background())
```
yurishkuro, line 84: you can't shut it down right after creating :-) Shutdown should be called from main.

pkg/jtracer/jtracer.go
```diff
@@ -14,16 +14,77 @@
...
+		otlptracehttp.NewClient(opts...),
+	)
+	if err != nil {
+		log.Fatal("failed to create exporter: %w", err)
+	}
+
+	// Register the trace exporter with a TracerProvider, using a batch
+	// span processor to aggregate spans before export.
+	bsp := sdktrace.NewBatchSpanProcessor(traceExporter)
+	tracerProvider := sdktrace.NewTracerProvider(
```
yurishkuro, line 64: what about sampling, is it defaulted to 100%?

pkg/jtracer/jtracer.go
```diff
@@ -14,16 +14,77 @@
...
-	return JTracer{OT: opentracing.NoopTracer{}}
+	return JTracer{OT: opentracing.NoopTracer{}, OTEL: trace.NewNoopTracerProvider()}
+}
+
+// initBoth initializes OpenTelemetry SDK and uses OTel-OpenTracing Bridge
+func initBoth() (opentracing.Tracer, trace.TracerProvider) {
+	opts := []otlptracehttp.Option{otlptracehttp.WithInsecure()}
+	traceExporter, err := otlptrace.New(
+		context.Background(),
+		otlptracehttp.NewClient(opts...),
```
yurishkuro, line 55: it's better to use grpc, this is Jaeger's default protocol

cmd/query/app/http_handler_test.go
```diff
@@ -303,10 +305,14 @@ func TestGetTrace(t *testing.T) {
...
-			jTracer := jtracer.OT(jaegerTracer)
-			defer jaegerCloser.Close()
+			exporter := tracetest.NewInMemoryExporter()
+			tracerProvider := sdktrace.NewTracerProvider(
+				sdktrace.WithSyncer(exporter),
+				sdktrace.WithSampler(sdktrace.AlwaysSample()),
+			)
+			// Use the bridgeTracer as OpenTracing tracer(otTrace).
+			otTracer, _ := otbridge.NewTracerPair(tracerProvider.Tracer(""))
+			jTracer := jtracer.JTracer{OT: otTracer}
```
yurishkuro, line 315:
```suggestion
			jTracer := jtracer.JTracer{OT: otTracer, OTEL: tracerProvider}
```

pkg/jtracer/jtracer.go
afzal442: I suspect I may go wrong here.

pkg/jtracer/jtracer.go
```diff
@@ -14,16 +14,109 @@
...
+// newExporter returns a console exporter.
+func newExporter(ctx context.Context) (sdktrace.SpanExporter, error) {
+	ctx, cancel := context.WithTimeout(ctx, time.Second)
+	defer cancel()
+	server := grpc.NewServer()
+
+	lis, _ := net.Listen("tcp", ":0")
+	go func() {
+		err := server.Serve(lis)
+		log.Fatal("failed to create gRPC server connection: %w", err)
```
afzal442, line 104: @yurishkuro how would I log this err? you told something like wrapping this into upstream logger.

yurishkuro: why are you starting a server from `newExporter`?

yurishkuro: https://github.com/jaegertracing/jaeger/blob/e40a3e5a0ea1ab2a4634348bc3dd2093775651af/cmd/tracegen/main.go#L102

afzal442: That's my bad. I would consider this otlp-grpc https://github.com/jaegertracing/jaeger/blob/e40a3e5a0ea1ab2a4634348bc3dd2093775651af/cmd/tracegen/main.go#L115-L117
Right?

yurishkuro: yes

cmd/query/main.go
```diff
@@ -76,28 +73,15 @@ func main() {
...
-			)
 			if err != nil {
 				logger.Fatal("Failed to initialize tracer", zap.Error(err))
 			}
-			defer closer.Close()
-			opentracing.SetGlobalTracer(tracer)
-			jtracer := jtracer.OT(tracer)
+
+			jtracer := jtracer.New()
+			defer jtracer.Close(context.Background())
```
yurishkuro, line 84: I think it's better to call Close at the end of `svc.RunAndThen` at L130

pkg/jtracer/jtracer.go
```diff
@@ -14,16 +14,103 @@
...
+var once sync.Once
+
+func New() JTracer {
+	jt := JTracer{}
+	zc := zap.NewDevelopmentConfig()
+	zc.Level = zap.NewAtomicLevelAt(zapcore.Level(-8)) // level used by OTEL's Debug()
+	logger, err := zc.Build()
+	if err != nil {
+		panic(err)
+	}
```
yurishkuro, line 52: if you need the logger, you should take is as argument to New(). Observability singletons are never initialized in utility functions, only in main.

pkg/jtracer/jtracer.go
```diff
@@ -14,16 +14,103 @@
...
+
+	return otTracer, tracerProvider
+}
+
+// newExporter returns a console exporter.
+func newExporter() (sdktrace.SpanExporter, error) {
+	client := otlptracegrpc.NewClient(
+		otlptracegrpc.WithInsecure(),
+	)
+	// Set up a otlp-trace exporter
```
yurishkuro, line 101:
```suggestion
```

pkg/jtracer/jtracer.go
```diff
@@ -14,16 +14,103 @@
...
+			))
+	})
+
+	// Use the bridgeTracer as your OpenTracing tracer(otTrace).
+	otTracer, _ := otbridge.NewTracerPair(tracerProvider.Tracer(""))
+
+	return otTracer, tracerProvider
+}
+
+// newExporter returns a console exporter.
```
yurishkuro, line 96:
```suggestion
```

pkg/jtracer/jtracer.go
```diff
@@ -14,16 +14,103 @@
...
+
+	return traceExporter, nil
+}
+
+// Shutdown the tracerProvider to clean up resources
+func (jt JTracer) Close(ctx context.Context) {
+	err := jt.OTEL.Shutdown(ctx)
+	if err != nil {
+		log.Fatalf("Error shutting down tracer provider: %v", err)
+	}
```
yurishkuro, line 115:
```suggestion
func (jt JTracer) Close(ctx context.Context) error {
	return jt.OTEL.Shutdown(ctx)
```

pkg/jtracer/jtracer.go
yurishkuro: why the 2nd value ignored?

pkg/jtracer/jtracer.go
```diff
@@ -14,16 +14,103 @@
...
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	semconv "go.opentelemetry.io/otel/semconv/v1.4.0"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+)
 
 type JTracer struct {
-	OT opentracing.Tracer
+	OT   opentracing.Tracer
+	OTEL *sdktrace.TracerProvider
```
yurishkuro, line 39: Isn't there a corresponding interface in `otel/`? Why do we depend on the struct?

cmd/all-in-one/main.go
```diff
@@ -274,8 +274,8 @@ func startQuery(
 ) *queryApp.Server {
 	spanReader = storageMetrics.NewReadMetricsDecorator(spanReader, baseFactory.Namespace(metrics.NSOptions{Name: "query"}))
 	qs := querysvc.NewQueryService(spanReader, depReader, *queryOpts)
-	jtracer := jtracer.OT(opentracing.GlobalTracer())
-	server, err := queryApp.NewServer(svc.Logger, qs, metricsQueryService, qOpts, tm, jtracer)
+	tracer := jtracer.JTracer{OT: opentracing.GlobalTracer()}
```
yurishkuro, line 277: changes here should be similar to cmd/query/main

cmd/query/app/handler_options.go
```diff
@@ -64,7 +64,8 @@ func (handlerOptions) QueryLookbackDuration(queryLookbackDuration time.Duration)
 // Tracer creates a HandlerOption that initializes OpenTracing tracer
 func (handlerOptions) Tracer(tracer jtracer.JTracer) HandlerOption {
 	return func(apiHandler *APIHandler) {
-		apiHandler.tracer = tracer
+		apiHandler.tracer.OT = tracer.OT
+		apiHandler.tracer.OTEL = tracer.OTEL
```
yurishkuro, line 68: why does apiHandler need both? If it passes tracer further downstream, it should pass JTracer. If it uses it directly for instrumentation, then it should only take one of these.

pkg/jtracer/jtracer.go
afzal442: I wondered how I can close this if we use `trace.TracerProvider` interface.

yurishkuro: There are several ways:
  1. you can return Closer, similar to how Jaeger SDK's creation returns `(opentracing.Tracer, io.Closer, error)`
  2. you can store the `provider.Shutdown` function as a private member of JTracer and invoke it from this Close method

afzal442: Thanks @yurishkuro for your masterpiece. 😄

cmd/all-in-one/main.go
```diff
@@ -287,6 +288,10 @@ func startQuery(
 	if err := server.Start(); err != nil {
 		svc.Logger.Fatal("Could not start jaeger-query service", zap.Error(err))
 	}
+	if err = tracer.Close(context.Background()); err != nil {
```
yurishkuro, line 291: so when do you think this will be executed?

cmd/query/main.go
```diff
@@ -133,6 +132,9 @@ func main() {
 					logger.Error("Failed to close storage factory", zap.Error(err))
 				}
 			})
+			if err = jtracer.Close(context.Background()); err != nil {
```
yurishkuro, line 135: move this _inside_ `RunAndThen` callback

pkg/jtracer/jtracer.go
```diff
@@ -52,20 +69,9 @@ func New() JTracer {
...
-	return JTracer{OT: opentracing.NoopTracer{}, OTEL: &sdktrace.TracerProvider{}}
-}
-
-// initBoth initializes OpenTelemetry SDK and uses OTel-OpenTracing Bridge
-func (jt JTracer) initBoth() (opentracing.Tracer, *sdktrace.TracerProvider) {
-	traceExporter, err := newExporter()
+	traceExporter, err := otelExporter()
 	if err != nil {
-		jt.log.Sugar().Fatalf("failed to create exporter", zap.Any("error", err))
+		logger.Sugar().Fatalf("failed to create exporter", zap.Any("error", err))
```
yurishkuro, line 74: return errors, don't log them

pkg/jtracer/jtracer.go
```diff
@@ -88,17 +94,19 @@ func (jt JTracer) initBoth() (opentracing.Tracer, *sdktrace.TracerProvider) {
 	})
 
 	// Use the bridgeTracer as your OpenTracing tracer(otTrace).
-	otTracer, _ := otbridge.NewTracerPair(tracerProvider.Tracer(""))
+	otTracer, wrapperTracerProvider := otbridge.NewTracerPair(tracerProvider.Tracer(""))
```
yurishkuro, line 97: does `wrapperTracerProvider` have a Shutdown method too?

pkg/jtracer/jtracer.go
```diff
@@ -88,17 +94,19 @@ func (jt JTracer) initBoth() (opentracing.Tracer, *sdktrace.TracerProvider) {
...
 
-	return otTracer, tracerProvider
+	closer := func() error {
+		return tracerProvider.Shutdown(context.Background())
+	}
+
+	return otTracer, wrapperTracerProvider, logger, closer
```
yurishkuro, line 103: why do you return logger here? logger is a dependency of this whole module

cmd/all-in-one/main.go
afzal442: Sorry! I think it was supposed to be here `svc.RunAndThen(func() {`

cmd/all-in-one/main.go
```diff
@@ -194,7 +195,7 @@ by default uses only in-memory database.`,
 			agent := startAgent(cp, aOpts, logger, metricsFactory)
 
 			// query
-			querySrv := startQuery(
+			querySrv, tracer := startQuery(
```
yurishkuro, line 198: why do you create a tracer in startQuery when there is initTracer function called in L107?

cmd/query/main.go
```diff
@@ -148,6 +123,9 @@ func main() {
 				if err := storageFactory.Close(); err != nil {
 					logger.Error("Failed to close storage factory", zap.Error(err))
 				}
+				if err = jtracer.Close(context.Background()); err != nil {
+					svc.Logger.Fatal("Error shutting down tracer provider", zap.Error(err))
```
yurishkuro, line 127:
```suggestion
					logger.Fatal("Error shutting down tracer provider", zap.Error(err))
```

pkg/jtracer/jtracer.go
```diff
@@ -14,16 +14,97 @@
...
+	OTEL   trace.TracerProvider
+	closer func() error
+	err    error
 }
 
-func OT(t opentracing.Tracer) JTracer {
-	return JTracer{OT: t}
+var once sync.Once
+
+func New() JTracer {
```
yurishkuro, line 43:
```suggestion
func New() (*JTracer, error) {
```

cmd/all-in-one/main.go
afzal442: @yurishkuro Can we rmv the initTracer() in favor of jtracer.New() ? Sorry if I misunderstood here.

yurishkuro: I'd say we should replace it, not remove. It makes no sense why a method `startQuery` would return a tracer object, it breaks all single purpose principles.

cmd/all-in-one/main.go
```diff
@@ -271,11 +275,11 @@ func startQuery(
...
 	baseFactory metrics.Factory,
 	tm *tenancy.Manager,
-) *queryApp.Server {
+) (*queryApp.Server, jtracer.JTracer) {
 	spanReader = storageMetrics.NewReadMetricsDecorator(spanReader, baseFactory.Namespace(metrics.NSOptions{Name: "query"}))
 	qs := querysvc.NewQueryService(spanReader, depReader, *queryOpts)
-	jtracer := jtracer.OT(opentracing.GlobalTracer())
-	server, err := queryApp.NewServer(svc.Logger, qs, metricsQueryService, qOpts, tm, jtracer)
+	tracer := jtracer.New()
+	server, err := queryApp.NewServer(svc.Logger, qs, metricsQueryService, qOpts, tm, tracer)
```
afzal442, line 282: So here server is getting initialised during callback. As we have initiated a otel tracer here, can we replace it initTracer()? If so, how to close this?

pkg/jtracer/jtracer.go
afzal442: I didn't find anywhere. Sorry if any pointers.

cmd/all-in-one/main.go
yurishkuro: You create tracer outside of this function and pass it in as parameter.

afzal442: sorry, I didn't get this. I mean I created `tracer` as NoOpTracer

yurishkuro: You are creating tracer in L103. Just pass it as argument to `startQuery()`, don't create a new one here.

cmd/all-in-one/main.go
```diff
@@ -287,32 +286,16 @@ func startQuery(
...
-	)
+func initTracer() error {
+	tracer, err := jtracer.New()
 	if err != nil {
-		logger.Fatal("Failed to initialize tracer", zap.Error(err))
+		log.Fatal("Failed to initialize tracer", zap.Error(err))
 	}
-	opentracing.SetGlobalTracer(tracer)
-	return closer
+	return tracer.Close(context.Background())
```
yurishkuro, line 298: what do you expect to happen after this line?

afzal442: I think tracer should shutdown after the init. Or I should be returning tracer obj and get it closed inside `svc.RunAndThen(func() {`

yurishkuro: how do you expect to use the tracer later if you shut it down on init?

pkg/jtracer/jtracer.go
```diff
@@ -14,16 +14,95 @@
...
+	closer func() error
 }
 
-func OT(t opentracing.Tracer) JTracer {
-	return JTracer{OT: t}
+var once sync.Once
+
+func New() (*JTracer, error) {
+	jt := JTracer{}
+	opentracingTracer, otelTracerProvider, closed, error := jt.initBoth(context.Background())
```
albertteoh, line 44: There are a lot of returns from this method, and also it doesn't need to have a receiver. 

I have a suggestion for breaking this up a little to make it a little easier to read:

1. create a func (e.g. `initOTEL`) that returns just the otel tracer provider. Remove `initBoth`.
2. build the opentracing tracer separately from within `New`
3. build the closer separately from within `New`

Also, the `otlp` serviceName we saw in the failing integration test was due to this config below:
```go
semconv.ServiceNameKey.String("otlp")
```

It seems the resource attributes (that contain the service name) need to be determined when the tracer provider is created, so we may need to pass in the service name into `New`. This implies a tracer provider per service.

Try to avoid using `error` as a variable name; it's the type name for the error interface, so it can be pretty confusing if overloaded as a var name.

We don't need `jt` and `initOTEL` does not need to have the `JTracer` as a receiver.

So putting all that together:
```go
func New(serviceName string) (*JTracer, error) {
	ctx := context.Background()
	tracerProvider, err := initOTEL(ctx, serviceName)
	if err != nil {
		return nil, err
	}
	// Use the bridgeTracer as your OpenTracing tracer(otTrace).
	otelTracer := tracerProvider.Tracer("github.com/jaegertracing/jaeger/pkg/jtracer")
	otTracer, wrappedOTELTracer := otbridge.NewTracerPair(otelTracer)

	closer := func() error {
		return tracerProvider.Shutdown(ctx)
	}

	return &JTracer{
		OT:     otTracer,
		OTEL:   wrappedOTELTracer,
		closer: closer,
	}, nil
}
```

pkg/jtracer/jtracer.go
```diff
@@ -14,16 +14,95 @@
...
+
+	return &JTracer{
+		OT:     opentracingTracer,
+		OTEL:   otelTracerProvider,
+		closer: closed,
+	}, error
```
albertteoh, line 50: As these codcov warnings suggest, it would be nice to add some tests for this file.

afzal442: Yes, TODO!

cmd/all-in-one/main.go
```diff
@@ -270,12 +269,11 @@ func startQuery(
 	depReader dependencystore.Reader,
 	metricsQueryService querysvc.MetricsQueryService,
 	baseFactory metrics.Factory,
-	tm *tenancy.Manager,
+	tm *tenancy.Manager, jt jtracer.JTracer,
```
yurishkuro, line 272:
```suggestion
	tm *tenancy.Manager, 
	jt jtracer.JTracer,
```

cmd/all-in-one/main.go
```diff
@@ -287,32 +285,16 @@ func startQuery(
...
-	}
-	traceCfg, err := traceCfg.FromEnv()
-	if err != nil {
-		logger.Fatal("Failed to read tracer configuration", zap.Error(err))
-	}
-	tracer, closer, err := traceCfg.NewTracer(
-		jaegerClientConfig.Metrics(jlibadapter.NewAdapter(svc.MetricsFactory)),
-		jaegerClientConfig.Logger(jaegerClientZapLog.NewLogger(logger)),
-	)
+func initTracer() jtracer.JTracer {
```
yurishkuro, line 292: this function is not necessary, you can do the same directly in main, and log the error using logger, not the `log` package

pkg/jtracer/jtracer.go
yurishkuro: small nit: wrappedOTELTracer -> wrappedTracerProvider

yurishkuro: another nit: the closer function should take context as argument provided by the caller

plugin/metrics/prometheus/metricsstore/reader.go
```diff
@@ -26,14 +26,15 @@ import (
 	"time"
 	"unicode"
 
-	"github.com/opentracing/opentracing-go"
```
yurishkuro, line 29: can we undo changes to this file? I think we decided not to touch it in this PR

afzal442: Yes, sure I am.

albertteoh:
There are two (integration) test failure errors:

1. Got an empty result from a trace query instead of a single trace.
  ```
=== RUN   TestAllInOne/getAPITrace
    all_in_one_test.go:105: 
        	Error Trace:	/home/runner/work/jaeger/jaeger/cmd/all-in-one/all_in_one_test.go:105
        	Error:      	"[]" should have 1 item(s), but has 0
        	Test:       	TestAllInOne/getAPITrace
  ```
2. Got a service name "otlp" instead of "jaeger-query" from the trace query.
```
=== RUN   TestAllInOne/getServicesAPIV3
    all_in_one_test.go:160: 
        	Error Trace:	/home/runner/work/jaeger/jaeger/cmd/all-in-one/all_in_one_test.go:160
        	Error:      	Not equal: 
        	            	expected: []string{"jaeger-query"}
        	            	actual  : []string{"otlp"}
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,3 +1,3 @@
        	            	 ([]string) (len=1) {
        	            	- (string) (len=12) "jaeger-query"
        	            	+ (string) (len=4) "otlp"
        	            	 }
        	Test:       	TestAllInOne/getServicesAPIV3
```

I'll explain each one in turn.

## 1. Got an empty result from a trace query instead of a single trace.

There are two important invocations in the integration test:
1. [Create the trace](https://github.com/jaegertracing/jaeger/blob/d1cf716fcc59c0b64c48cc7d2a5de6a8c424eab5/cmd/all-in-one/all_in_one_test.go#L64C1-L64C1): more importantly, the trace is [created with the `jaeger-debug-id=debug`](https://github.com/jaegertracing/jaeger/blob/d1cf716fcc59c0b64c48cc7d2a5de6a8c424eab5/cmd/all-in-one/all_in_one_test.go#L73).
    1. This is [little known feature](https://github.com/jaegertracing/documentation/issues/349) of jaeger-client-go that supports the [`jaeger-debug-id`](https://github.com/jaegertracing/jaeger-client-go/blob/master/constants.go#L34) header to allow easier identification of a specific set of traces for debugging purposes. It's achieved by inserting the `jaeger-debug-id` tag key into the span that's emitted to the collector (or agent) such that you can then search for the span with the tag filter `jaeger-debug-id=debug` for example.
2. [Get the trace created above](https://github.com/jaegertracing/jaeger/blob/d1cf716fcc59c0b64c48cc7d2a5de6a8c424eab5/cmd/all-in-one/all_in_one_test.go#L65): a request to the jaeger-query HTTP API is made with the tag filter request parameter of `tag=jaeger-debug-id:debug` ([ref](https://github.com/jaegertracing/jaeger/blob/d1cf716fcc59c0b64c48cc7d2a5de6a8c424eab5/cmd/all-in-one/all_in_one_test.go#L49)). 
    1. This leads to an empty result. Why? It's because we've switched to using [JTracer](https://github.com/jaegertracing/jaeger/pull/4574/files#diff-0888ca3408f76362e5e951617b16f40ac09e6ff66e400f48097903f2289f4d8bR103) instead of jaeger-client-go's tracer in this PR for jaeger all-in-one.
    2. Why does using JTracer lead to an empty query result when filtering on `tag=jaeger-debug-id:debug`? It's because the OTEL SDK does not appear to support this functionality, as @yurishkuro calls out in [this opentelemetry-specification issue](https://github.com/open-telemetry/opentelemetry-specification/issues/231):
        > In addition, the debug flag can be set by the user even before the trace is created, by including a special header `jaeger-debug-id: anything`. When Jaeger sees this header in the incoming request, it's equivalent to setting `sampling.priority=1` and `jaeger-debug-id=$value` tags on the span. Storing the header value as a correlation ID allows finding the trace later. E.g. I can send a curl request with `jaeger-debug-id: yuri-test-1`.
    3. Since the `jaeger-debug-id:debug` tag is not inserted into the span by jaeger all-in-one during the integration test, that span cannot be found with the `tag=jaeger-debug-id:debug` filter resulting in a failed test.

We _could_, as a workaround, omit the `tag=jaeger-debug-id:debug` filter from the API request but I think it will weaken the test expectation because the act of hitting the `/api/traces` endpoint results in emitting a span, and so may be unclear where the query result's span came from. Maybe other folks can think of a better workaround?

## 2. Got a service name "otlp" instead of "jaeger-query" from the trace query.

Caused by the following config when creating the tracer provider:
```go
semconv.ServiceNameKey.String("otlp")
```

It seems the resource attributes (that contain the service name) need to be determined when the tracer provider is created, so we may need to pass in the service name into New. This implies a tracer provider per service.

I think this has since been "fixed" in [a2eb44a](https://github.com/jaegertracing/jaeger/pull/4574/commits/a2eb44a6ba9ac8a4c6fed3ea1dbdd62fdb4a4186), but the name just needs updating from `query-service` to `jaeger-query`.

cmd/all-in-one/main.go
```diff
@@ -103,7 +100,10 @@ by default uses only in-memory database.`,
 				svc.MetricsFactory.Namespace(metrics.NSOptions{Name: "jaeger"}))
 			version.NewInfoMetrics(metricsFactory)
 
-			tracerCloser := initTracer(svc)
+			tracer, err := jtracer.New("query-service")
```
albertteoh, line 103: This is the cause of [the failing test](https://github.com/jaegertracing/jaeger/actions/runs/5563637373/jobs/10162715011?pr=4574#step:11:1029).

```suggestion
			tracer, err := jtracer.New("jaeger-query")
```

yurishkuro:
> the trace is created with the `jaeger-debug-id=debug`

I suspect this is done in all-in-one in order to have something to search by later on. I don't know of an equivalent with OTEL SDK, i.e. something that the _caller_ provides and that ends up in the trace. With Jaeger SDK, `jaeger-debug-id` was one option, but also any other baggage would also work because iirc the SDK would store that baggage as span tags (although I may be wrong, it may only happen when SetBaggage API is called by the app).

Going back to the main objective - the integration test needs some identifier to retrieve the trace. The most ideal solution would be for OTEL SDK to return trace/span ID in the response. There is already a [W3C spec](https://github.com/w3c/trace-context/blob/551a5b0855171281e98b4c2a814bf9e1f837cd53/spec/21-http_response_header_format.md?plain=1) for that. The header format is identical to `traceparent` for a given span when sent to a downstream service, meaning we could use the existing W3C Propagator and just write a small middleware to inject it into response headers. Then the integration test can parse the header (also using W3C Trace Context Propagator) and use trace-id for querying the backend.

yurishkuro:
I added the traceresponse bit to HotROD, we can copy it for the main repo: #4584

cmd/all-in-one/main.go
afzal442: Ah, sorry! I missed that.

cmd/all-in-one/all_in_one_test.go
```diff
@@ -46,12 +47,13 @@ const (
 	agentURL      = "http://" + agentHostPort
 
 	getServicesURL         = queryURL + "/api/services"
-	getTraceURL            = queryURL + "/api/traces?service=jaeger-query&tag=jaeger-debug-id:debug"
 	getSamplingStrategyURL = agentURL + "/sampling?service=whatever"
 
 	getServicesAPIV3URL = queryURL + "/api/v3/services"
 )
 
+var getTraceURL = queryURL + "/api/traces/"
```
yurishkuro, line 55: why move this out of the const list?

cmd/query/app/http_handler_test.go
```diff
@@ -303,10 +305,14 @@ func TestGetTrace(t *testing.T) {
...
-			jaegerTracer, jaegerCloser := jaeger.NewTracer("test", jaeger.NewConstSampler(true), reporter)
-			jTracer := jtracer.OT(jaegerTracer)
-			defer jaegerCloser.Close()
+			exporter := tracetest.NewInMemoryExporter()
+			tracerProvider := sdktrace.NewTracerProvider(
+				sdktrace.WithSyncer(exporter),
+				sdktrace.WithSampler(sdktrace.AlwaysSample()),
+			)
+			// Use the bridgeTracer as OpenTracing tracer(otTrace).
+			otTracer, _ := otbridge.NewTracerPair(tracerProvider.Tracer(""))
```
yurishkuro, line 314: the 2nd value is wrapper provider. Does it have a Close/Shutdown method? If it does, you need to defer-call it

pkg/jtracer/jtracer.go
```diff
@@ -14,16 +14,98 @@
...
+
+	"github.com/opentracing/opentracing-go"
+	"go.opentelemetry.io/otel"
+	otbridge "go.opentelemetry.io/otel/bridge/opentracing"
+	"go.opentelemetry.io/otel/exporters/otlp/otlptrace"
+	"go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc"
+	"go.opentelemetry.io/otel/propagation"
+	"go.opentelemetry.io/otel/sdk/resource"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	semconv "go.opentelemetry.io/otel/semconv/v1.4.0"
```
yurishkuro, line 30: why 1.4.0? There are much higher versions

yurishkuro:
integration tests now pass!

afzal442:
Variable declared by const doesn't allow to change it further.

afzal442:
Yes tracerresponse obj in the header made this magic. :)

cmd/all-in-one/main.go
```diff
@@ -271,11 +273,11 @@ func startQuery(
 	metricsQueryService querysvc.MetricsQueryService,
 	baseFactory metrics.Factory,
 	tm *tenancy.Manager,
+	jt *jtracer.JTracer,
 ) *queryApp.Server {
 	spanReader = storageMetrics.NewReadMetricsDecorator(spanReader, baseFactory.Namespace(metrics.NSOptions{Name: "query"}))
 	qs := querysvc.NewQueryService(spanReader, depReader, *queryOpts)
-	jtracer := jtracer.OT(opentracing.GlobalTracer())
-	server, err := queryApp.NewServer(svc.Logger, qs, metricsQueryService, qOpts, tm, jtracer)
+	server, err := queryApp.NewServer(svc.Logger, qs, metricsQueryService, qOpts, tm, *jt)
```
yurishkuro, line 280:
```suggestion
	server, err := queryApp.NewServer(svc.Logger, qs, metricsQueryService, qOpts, tm, jt)
```

cmd/query/app/handler_options.go
```diff
@@ -64,7 +64,7 @@ func (handlerOptions) QueryLookbackDuration(queryLookbackDuration time.Duration)
 // Tracer creates a HandlerOption that initializes OpenTracing tracer
 func (handlerOptions) Tracer(tracer jtracer.JTracer) HandlerOption {
 	return func(apiHandler *APIHandler) {
-		apiHandler.tracer = tracer
+		apiHandler.tracer.OTEL = tracer.OTEL
```
yurishkuro, line 67:
```suggestion
		apiHandler.tracer = tracer
```

cmd/query/app/http_handler_test.go
```diff
@@ -303,10 +305,14 @@ func TestGetTrace(t *testing.T) {
...
-			jaegerTracer, jaegerCloser := jaeger.NewTracer("test", jaeger.NewConstSampler(true), reporter)
-			jTracer := jtracer.OT(jaegerTracer)
-			defer jaegerCloser.Close()
+			exporter := tracetest.NewInMemoryExporter()
+			tracerProvider := sdktrace.NewTracerProvider(
+				sdktrace.WithSyncer(exporter),
+				sdktrace.WithSampler(sdktrace.AlwaysSample()),
+			)
+			// Use the bridgeTracer as OpenTracing tracer(otTrace).
+			otTracer, wrappedTracerProvider := otbridge.NewTracerPair(tracerProvider.Tracer(""))
```
yurishkuro, line 314: `defer wrappedTracerProvider.Shutdown` (or tracerProvider.Shutdown if wrapped doesn't support that)

cmd/query/main.go
```diff
@@ -75,29 +72,10 @@ func main() {
...
-					Type:  "const",
-					Param: 1.0,
-				},
-				RPCMetrics: true,
-			}
-			traceCfg, err = traceCfg.FromEnv()
+			jtracer, err := jtracer.New("jaeger-query")
 			if err != nil {
-				logger.Fatal("Failed to read tracer configuration", zap.Error(err))
+				logger.Fatal("Failed to create exporter:", zap.Error(err))
```
yurishkuro, line 77:
```suggestion
				logger.Fatal("Failed to create tracer:", zap.Error(err))
```

cmd/query/main.go
```diff
@@ -128,7 +106,7 @@ func main() {
 				dependencyReader,
 				*queryServiceOptions)
 			tm := tenancy.NewManager(&queryOpts.Tenancy)
-			server, err := app.NewServer(svc.Logger, queryService, metricsQueryService, queryOpts, tm, jtracer)
+			server, err := app.NewServer(svc.Logger, queryService, metricsQueryService, queryOpts, tm, *jtracer)
```
yurishkuro, line 109:
```suggestion
			server, err := app.NewServer(svc.Logger, queryService, metricsQueryService, queryOpts, tm, jtracer)
```

pkg/jtracer/jtracer.go
```diff
@@ -14,16 +14,98 @@
...
+
+func New(serviceName string) (*JTracer, error) {
+	ctx := context.Background()
+	tracerProvider, err := initOTEL(ctx, serviceName)
+	if err != nil {
+		return nil, err
+	}
+	// Use the bridgeTracer as your OpenTracing tracer(otTrace).
+	otelTracer := tracerProvider.Tracer("github.com/jaegertracing/jaeger/pkg/jtracer")
+	otTracer, wrappedTracerProvider := otbridge.NewTracerPair(otelTracer)
```
yurishkuro, line 50: does this provide Shutdown function? If it does, does it also invoke tracerProvider.Shutdown?

pkg/jtracer/jtracer.go
```diff
@@ -14,16 +14,98 @@
...
+// initOTEL initializes OTEL Tracer
+func initOTEL(ctx context.Context, svc string) (*sdktrace.TracerProvider, error) {
+	traceExporter, err := otelExporter(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	// Register the trace exporter with a TracerProvider, using a batch
+	// span processor to aggregate spans before export.
+	bsp := sdktrace.NewBatchSpanProcessor(traceExporter, sdktrace.WithBlocking())
```
yurishkuro, line 76: why do we want `sdktrace.WithBlocking()` here?

pkg/jtracer/jtracer_test.go
```diff
@@ -14,18 +14,4 @@
...
-	"github.com/jaegertracing/jaeger/pkg/jtracer"
-)
-
-func TestOT(t *testing.T) {
-	mockTracer := opentracing.NoopTracer{}
-	jtracer := jtracer.OT(opentracing.NoopTracer{})
-
-	assert.Equal(t, mockTracer, jtracer.OT)
-}
+// TODO
```
yurishkuro, line 17: remove this file, otherwise it will cause 0% coverage to be reported.

pkg/jtracer/testutils/.nocover
```diff
@@ -0,0 +1 @@
+requires connection to OTLP receiver
```
yurishkuro, line 1:
```suggestion
FIXME
```
when we write an explanation in this file it usually means that unit testing is not feasible without external dependencies. In this case this is testable, we just skipped it.

pkg/jtracer/jtracer_test.go
afzal442: when I did this, I was getting linting error.

pkg/jtracer/jtracer.go
afzal442: I don't see. Sorry!

pkg/jtracer/jtracer_test.go
yurishkuro: that's because you didn't have .nocover file

yurishkuro: It's an intentional safety check.

afzal442: when test file is not there, linting causes an error. `No jtracer_test.go`

yurishkuro: did you commit the deletion? The list of files for linting may be read from git

afzal442: That's correct. I did not.

cmd/all-in-one/main.go
```diff
@@ -103,7 +100,10 @@ by default uses only in-memory database.`,
 				svc.MetricsFactory.Namespace(metrics.NSOptions{Name: "jaeger"}))
 			version.NewInfoMetrics(metricsFactory)
 
-			tracerCloser := initTracer(svc)
+			tracer, err := jtracer.New("jaeger-query")
```
yurishkuro, line 103:
```suggestion
			tracer, err := jtracer.New("jaeger-all-in-one")
```

pkg/jtracer/testutils/.nocover
```diff
@@ -0,0 +1 @@
+FIXME
```
yurishkuro, line 1: it's supposed to be `pkg/jtracer/.nocover`

cmd/query/app/grpc_handler.go
```diff
@@ -73,8 +73,8 @@ func NewGRPCHandler(queryService *querysvc.QueryService,
 		options.Logger = zap.NewNop()
 	}
 
-	if options.Tracer.OT == nil {
-		options.Tracer = jtracer.NoOp()
+	if options.Tracer.OTEL == nil {
+		options.Tracer = *jtracer.NoOp()
```
yurishkuro, line 77: there should be no dereferencing of the pointer, JTracer struct is always passed / stored as a pointer. And the check should be `if options.Tracer == nil {`

cmd/query/app/grpc_handler_test.go
```diff
@@ -927,7 +927,7 @@ func initializeTenantedTestServerGRPCWithOptions(t *testing.T, tm *tenancy.Manag
 	logger := zap.NewNop()
 	tracer := jtracer.NoOp()
 
-	server, addr := newGRPCServer(t, q, tqs.metricsQueryService, logger, tracer, tm)
+	server, addr := newGRPCServer(t, q, tqs.metricsQueryService, logger, *tracer, tm)
```
yurishkuro, line 930:
```suggestion
	server, addr := newGRPCServer(t, q, tqs.metricsQueryService, logger, tracer, tm)
```

cmd/query/app/http_handler.go
```diff
@@ -111,8 +112,8 @@ func NewAPIHandler(queryService *querysvc.QueryService, tm *tenancy.Manager, opt
 	if aH.logger == nil {
 		aH.logger = zap.NewNop()
 	}
-	if aH.tracer.OT == nil {
-		aH.tracer = jtracer.NoOp()
+	if aH.tracer.OTEL == nil {
+		aH.tracer = *jtracer.NoOp()
```
yurishkuro, line 116:
```suggestion
	if aH.tracer == nil {
		aH.tracer = jtracer.NoOp()
```

cmd/query/app/http_handler_test.go
```diff
@@ -303,12 +305,17 @@ func TestGetTrace(t *testing.T) {
...
-			jTracer := jtracer.OT(jaegerTracer)
-			defer jaegerCloser.Close()
-
-			ts := initializeTestServer(HandlerOptions.Tracer(jTracer))
+			exporter := tracetest.NewInMemoryExporter()
+			tracerProvider := sdktrace.NewTracerProvider(
+				sdktrace.WithSyncer(exporter),
+				sdktrace.WithSampler(sdktrace.AlwaysSample()),
+			)
+			// Use the bridgeTracer as OpenTracing tracer(otTrace).
```
yurishkuro, line 313:
```suggestion
```
no new information in the comment

cmd/query/app/server.go
```diff
@@ -132,7 +132,7 @@ func createGRPCServer(querySvc *querysvc.QueryService, metricsQuerySvc querysvc.
 
 	handler := NewGRPCHandler(querySvc, metricsQuerySvc, GRPCHandlerOptions{
 		Logger: logger,
-		Tracer: tracer,
+		Tracer: *tracer,
```
yurishkuro, line 135:
```suggestion
		Tracer: tracer,
```

cmd/query/app/grpc_handler.go
afzal442: I will note this hopefully.

cmd/query/app/grpc_handler.go
```diff
@@ -85,7 +85,7 @@ func NewGRPCHandler(queryService *querysvc.QueryService,
 		queryService:        queryService,
 		metricsQueryService: metricsQueryService,
 		logger:              options.Logger,
-		tracer:              options.Tracer,
+		tracer:              *options.Tracer,
```
yurishkuro, line 88: there should be no pointer dereferencing, please make sure the type is always `*JTracer`

pkg/jtracer/empty_test.go
```diff
@@ -14,18 +14,4 @@
...
-	"github.com/jaegertracing/jaeger/pkg/jtracer"
-)
-
-func TestOT(t *testing.T) {
-	mockTracer := opentracing.NoopTracer{}
-	jtracer := jtracer.OT(opentracing.NoopTracer{})
-
-	assert.Equal(t, mockTracer, jtracer.OT)
-}
+// TODO
```
yurishkuro, line 17: if you keep this file, you cannot pass the code coverage check in CI
```
codecov/patch — 29.72% of diff hit (target 95.00%)
```

afzal442: Thanks lot for that call.

cmd/all-in-one/main.go
```diff
@@ -271,48 +273,24 @@ func startQuery(
...
+	jt *jtracer.JTracer,
 ) *queryApp.Server {
 	spanReader = storageMetrics.NewReadMetricsDecorator(spanReader, baseFactory.Namespace(metrics.NSOptions{Name: "query"}))
 	qs := querysvc.NewQueryService(spanReader, depReader, *queryOpts)
-	jtracer := jtracer.OT(opentracing.GlobalTracer())
-	server, err := queryApp.NewServer(svc.Logger, qs, metricsQueryService, qOpts, tm, jtracer)
+	server, err := queryApp.NewServer(svc.Logger, qs, metricsQueryService, qOpts, tm, jt)
 	if err != nil {
-		svc.Logger.Fatal("Could not start jaeger-query service", zap.Error(err))
+		svc.Logger.Fatal("Could not start jaeger-all-in-one service", zap.Error(err))
```
yurishkuro, line 282:
```suggestion
		svc.Logger.Fatal("Could not create jaeger-query service", zap.Error(err))
```

cmd/all-in-one/main.go
```diff
@@ -271,48 +273,24 @@ func startQuery(
...
 	}
 	go func() {
 		for s := range server.HealthCheckStatus() {
 			svc.SetHealthCheckStatus(s)
 		}
 	}()
 	if err := server.Start(); err != nil {
-		svc.Logger.Fatal("Could not start jaeger-query service", zap.Error(err))
+		svc.Logger.Fatal("Could not start jaeger-all-in-one service", zap.Error(err))
```
yurishkuro, line 290:
```suggestion
		svc.Logger.Fatal("Could not start jaeger-query service", zap.Error(err))
```

cmd/all-in-one/all_in_one_test.go
```diff
@@ -70,10 +72,13 @@ func TestAllInOne(t *testing.T) {
 func createTrace(t *testing.T) {
 	req, err := http.NewRequest(http.MethodGet, getServicesURL, nil)
 	require.NoError(t, err)
-	req.Header.Add("jaeger-debug-id", "debug")
 
 	resp, err := httpClient.Do(req)
 	require.NoError(t, err)
+	traceResponse := resp.Header.Get("traceresponse")
+	traceID := strings.Split(traceResponse, "-")[1]
```
albertteoh, line 79: nit: I'd do this in two steps so we can have an explicit assertion on the expected number of parts in the traceresponse:

```go
parts := strings.Split(traceResponse, "-")
require.Len(t, parts, 4) // [version] [trace-id] [child-id] [trace-flags]
traceID := parts[1]
```

cmd/all-in-one/all_in_one_test.go
albertteoh, line 55: Why is this defined here instead of in the `const` section above?

cmd/all-in-one/all_in_one_test.go
afzal442: Nice observation. Thanks @albertteoh

cmd/all-in-one/all_in_one_test.go
afzal442: We can't modify it if it is declared as `const`.

albertteoh: Ah yes, you're right, I missed that.

cmd/all-in-one/main.go
```diff
@@ -103,7 +100,10 @@ by default uses only in-memory database.`,
 				svc.MetricsFactory.Namespace(metrics.NSOptions{Name: "jaeger"}))
 			version.NewInfoMetrics(metricsFactory)
 
-			tracerCloser := initTracer(svc)
+			tracer, err := jtracer.New("jaeger-all-in-one")
+			if err != nil {
+				logger.Fatal("Failed to initialize tracer", zap.Error(err))
```
albertteoh, line 105: Hmmm... I wonder why we don't annotate and return the error here instead of logging it.

cmd/query/app/http_handler_test.go
```diff
@@ -303,12 +305,16 @@ func TestGetTrace(t *testing.T) {
...
-			jTracer := jtracer.OT(jaegerTracer)
-			defer jaegerCloser.Close()
-
-			ts := initializeTestServer(HandlerOptions.Tracer(jTracer))
+			exporter := tracetest.NewInMemoryExporter()
+			tracerProvider := sdktrace.NewTracerProvider(
+				sdktrace.WithSyncer(exporter),
+				sdktrace.WithSampler(sdktrace.AlwaysSample()),
+			)
+			otTracer, wrappedTracerProvider := otbridge.NewTracerPair(tracerProvider.Tracer(""))
```
albertteoh, line 313: Why don't we give the tracer a name here?

pkg/jtracer/jtracer.go
```diff
@@ -14,16 +14,98 @@
...
+	}, nil
+}
+
+func NoOp() *JTracer {
+	return &JTracer{OT: opentracing.NoopTracer{}, OTEL: trace.NewNoopTracerProvider()}
 }
 
-func OT(t opentracing.Tracer) JTracer {
-	return JTracer{OT: t}
+// initOTEL initializes OTEL Tracer
```
albertteoh, line 67:
```suggestion
// initOTEL initializes OTEL Tracer.
```

pkg/jtracer/.nocover
```diff
@@ -0,0 +1 @@
+FIXME
```
albertteoh, line 1: I thought an empty file is sufficient?

albertteoh:
Is it a good time to remove the [WIP] from the PR title?

cmd/all-in-one/main.go
```diff
@@ -271,48 +273,24 @@ func startQuery(
...
-	traceCfg, err := traceCfg.FromEnv()
-	if err != nil {
-		logger.Fatal("Failed to read tracer configuration", zap.Error(err))
-	}
-	tracer, closer, err := traceCfg.NewTracer(
-		jaegerClientConfig.Metrics(jlibadapter.NewAdapter(svc.MetricsFactory)),
-		jaegerClientConfig.Logger(jaegerClientZapLog.NewLogger(logger)),
-	)
-	if err != nil {
-		logger.Fatal("Failed to initialize tracer", zap.Error(err))
```
afzal442, line 312: I think this is what it is. @albertteoh  we won't return here as well.

albertteoh: Yeah, just questioning why; I just can't see a reason why the error shouldn't returned, but maybe there's a good reason.

cmd/all-in-one/main.go
yurishkuro: I think mostly because if we return an error, it is intercepted at the top level of main like this:
```
		log.Fatal(err)
	}
```
Here the zapp logger is not available so the stdlib `log` is used, which is formatted differently. E.g.:
```shell
$ go run ./cmd/all-in-one --admin.http.tls.enabled=true --admin.http.tls.key=invalid
. . .
2023/07/20 11:43:08 cannot initialize admin server: failed to watch key pair invalid and : open invalid: no such file or directory
exit status 1
```
With zapp logs you get line number and other nice things.

pkg/jtracer/.nocover
yurishkuro: it may work, but the content of these files is printed when you run `make nocover`

cmd/all-in-one/main.go
afzal442: That's cool thing I noticed here. 🏷️

cmd/query/app/http_handler_test.go
afzal442: @yurishkuro LMK if we need to give it any name here.

yurishkuro: I don't think so, this is a test and you're not going to validate the output of OT Bridge.

Are you sure you even need OpenTracing tracer for this test? What if keep it as nil? Might just work.

afzal442: That worked. Thanks

yurishkuro (approved):

PR #4578, opened 2023-07-11
[tracegen] Enable BlockOnQueueFull in OTel SDK to avoid dropped spans

haanhvu:
[tracegen] Enable BlockOnQueueFull in OTel SDK to make spans not dropped in spikes

Call `WithBlocking()` from OTel SDK to enable `BlockOnQueueFull=true` to make spans not dropped when the queue is full.

Tested locally:
- Before the change: Spans in tracegen were dropped when generated to jaeger-collector, even with just 1m duration and 1 worker.
- With the change: Spans in tracegen were not dropped even with 1h duration and 10 workers.

Screenshot 2023-07-12 005149

### DIFF
--- a/cmd/tracegen/main.go
+++ b/cmd/tracegen/main.go
@@ -81,8 +81,7 @@ func createTracers(cfg *tracegen.Config, logger *zap.Logger) ([]trace.Tracer, fu
 		logger.Sugar().Infof("using %s trace exporter for service %s", cfg.TraceExporter, svc)
 
 		tp := sdktrace.NewTracerProvider(
-			sdktrace.WithBatcher(exp),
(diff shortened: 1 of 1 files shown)


yurishkuro (approved):

yurishkuro:
@haanhvu in the future, use a branch in a fork to create PRs, don't do it from main. It both breaks the CI and makes it harder for you to keep main in sync with upstream.

haanhvu:
> @haanhvu in the future, use a branch in a fork to create PRs, don't do it from main. It both breaks the CI and makes it harder for you to keep main in sync with upstream.

Yeah, I just did a super quick change on main since it's a tiny PR. Forgot that CI didn't accept it...

Issue #4576, opened 2023-07-11
[test]: panic: Log in goroutine after ... has completed: Starting main loop


kennyaz:
### What happened?

panic: Log in goroutine after ... has completed:  Starting main loop

### Steps to reproduce

Not Applicable

### Expected behavior

There is a possibility that the [goroutine](https://github.com/jaegertracing/jaeger/blob/main/cmd/ingester/app/consumer/consumer.go#L79) is not finished when [Close ](https://github.com/jaegertracing/jaeger/blob/main/cmd/ingester/app/consumer/consumer.go#L94)is called since there is no waitGroup that is waiting for the `c.logger.Info("Starting main loop")` to be done executing before the application is shutdown

### Relevant log output

```shell
panic: Log in goroutine after ... has completed:  Starting main loop

goroutine 39 [running]:
testing.(*common).logDepth(0xc00052bd40, {0xc0001b95f0, 0x30}, 0x3)
        GOROOT/src/testing/testing.go:1003 +0x4e7
testing.(*common).log(...)
        GOROOT/src/testing/testing.go:985
testing.(*common).Logf(0xc00052bd40, {0x5e4ddb?, 0x14aceb7?}, {0xc0000401c0?, 0xffffffffffffffff?, 0x31?})
        GOROOT/src/testing/testing.go:1036 +0x5a
go.uber.org/zap/zaptest.testingWriter.Write({{0x8eecf0?, 0xc00052bd40?}, 0x0?}, {0xc000241400?, 0x31, 0xc0000401b0?})
        external/org_uber_go_zap/zaptest/logger.go:130 +0xe6
go.uber.org/zap/zapcore.(*ioCore).Write(0xc0001cd860, {0x0, {0xc1233bb7a60da4ca, 0x1863da2, 0x2bf1e00}, {0x0, 0x0}, {0x5fa77a, 0x12}, {0x0, ...}, ...}, ...)
        external/org_uber_go_zap/zapcore/core.go:99 +0xb5
go.uber.org/zap/zapcore.(*CheckedEntry).Write(0xc0001b71e0, {0x0, 0x0, 0x0})
        external/org_uber_go_zap/zapcore/entry.go:255 +0x1d9
go.uber.org/zap.(*Logger).Info(0x0?, {0x5fa77a?, 0x0?}, {0x0, 0x0, 0x0})
        external/org_uber_go_zap/logger.go:220 +0x59
[...]

PR #4582, opened 2023-07-14
[kafka-consumer] Use wait group to ensure goroutine is finished before returning from Close

kennyaz:
Resolves # [4576](https://github.com/jaegertracing/jaeger/issues/4576)

Once we call [start](https://github.com/jaegertracing/jaeger/blob/main/cmd/ingester/app/consumer/consumer.go#L76) and call [close](https://github.com/jaegertracing/jaeger/blob/main/cmd/ingester/app/consumer/consumer.go#L76) again with fxtest module, there is a chance that the application will close the goroutine before the l[ogline](https://github.com/jaegertracing/jaeger/blob/main/cmd/ingester/app/consumer/consumer.go#L79) is executed causing **panic Log in routine after .. has completed**. 

Fix:
Add a wait group that will wait for the goroutine to finish before closing the channel.

### DIFF
--- a/cmd/ingester/app/consumer/consumer.go
+++ b/cmd/ingester/app/consumer/consumer.go
@@ -15,6 +15,7 @@
 package consumer
 
 import (
+	"fmt"
 	"sync"
 	"time"
 
@@ -77,8 +78,8 @@ func (c *Consumer) Start() {
 	c.deadlockDetector.start()
 	c.doneWg.Add(1)
 	go func() {
+		defer c.doneWg.Done()
 		c.logger.Info("Starting main loop")
-		c.doneWg.Done()
 		for pc := range c.internalConsumer.Partitions() {
 			c.partitionMapLock.Lock()
 			c.partitionIDToState[pc.Partition()] = &consumerState{partitionConsumer: pc}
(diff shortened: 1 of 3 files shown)


cmd/ingester/app/consumer/consumer.go
```diff
@@ -75,8 +75,10 @@ func New(params Params) (*Consumer, error) {
 // Start begins consuming messages in a go routine
 func (c *Consumer) Start() {
 	c.deadlockDetector.start()
+	c.doneWg.Add(1)
 	go func() {
 		c.logger.Info("Starting main loop")
+		c.doneWg.Done()
```
yurishkuro, line 81:
```suggestion
		defer c.doneWg.Done()
		c.logger.Info("Starting main loop")
```

yurishkuro (commented):
I think it's indeed a good idea to keep track of all goroutines that begin in Start, and reusing the existing `doneWg` for this makes sense. However, you're still not tracking the deadlock detector goroutine started in L77, which also performs logging, so I don't think this change will fully address your issue.

cmd/ingester/app/consumer/consumer.go
kennyaz: Doing, that cause test to timeout because this [channel](https://github.com/jaegertracing/jaeger/blob/main/cmd/ingester/app/consumer/consumer.go#L80) is not Closed at all for some reason. I investigate it further and it seems like this [method ](https://github.com/jaegertracing/jaeger/blob/main/cmd/ingester/app/consumer/consumer_test.go#L81)is not closing the partition channel properly at least. The test that causes the code to timeout is https://github.com/jaegertracing/jaeger/blob/main/cmd/ingester/app/consumer/consumer_test.go#L122. It seems like the bug is happening on the "github.com/Shopify/sarama/mocks" side.

kennyaz: I take it back. It is happening on the Jaeger Side. I will submit a correct push shortly

yurishkuro (approved):

PR #4586, opened 2023-07-16
Enable CORS settings on OTLP HTTP endpoint

### LINKED ISSUE #4459: [Feature]: Support CORS headers on OTEL HTTP endpoint

pmuls99:
- Resolves #4459 

-  Added a new package corscfg to handle the CORS requests on OTLP HTTP Endpoint and Zipkin


pmuls99:
@yurishkuro  , had to delete the previous PR, had signed off commits that were not authored by me. I will make the updates on this PR. Sorry for the inconvinience.

yurishkuro:
That's fine, but please refer to comments on the old PR

yurishkuro:
please run `make fmt`

cmd/collector/app/server/zipkin.go
```diff
@@ -85,14 +85,13 @@ func serveZipkin(server *http.Server, listener net.Listener, params *ZipkinServe
 	r := mux.NewRouter()
 	zHandler := zipkin.NewAPIHandler(params.Handler)
 	zHandler.RegisterRoutes(r)
-
-	origins := strings.Split(strings.ReplaceAll(params.AllowedOrigins, " ", ""), ",")
-	headers := strings.Split(strings.ReplaceAll(params.AllowedHeaders, " ", ""), ",")
+	allowedOrigins := strings.Split(strings.ReplaceAll(params.CORSConfig.AllowedOrigins, " ", ""), ",")
+	allowedHeaders := strings.Split(strings.ReplaceAll(params.CORSConfig.AllowedHeaders, " ", ""), ",")
```
yurishkuro, line 89: I think this can be done once in corscfg/flags when parsing from Viper

pmuls99: Done!

pkg/config/corscfg/flags.go
```diff
@@ -0,0 +1,49 @@
...
+	corsAllowedHeaders = corsPrefix + ".allowed-headers"
+	corsAllowedOrigins = corsPrefix + ".allowed-origins"
+)
+
+type Flags struct {
+	Prefix string
+}
+
+func (c Flags) AddFlags(flags *flag.FlagSet) {
+	flags.String(c.Prefix+corsAllowedHeaders, "content-type", "Allowed headers for the HTTP port , default content-type")
```
yurishkuro, line 35:
```suggestion
	flags.String(c.Prefix+corsAllowedHeaders, "content-type", "Comma-separated CORS allowed headers. Cf. https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Headers")
```

pkg/config/corscfg/flags.go
```diff
@@ -0,0 +1,49 @@
...
+	flags.String(c.Prefix+corsAllowedOrigins, "*", "Allowed origins for the HTTP port , default accepts all")
```
yurishkuro, line 36:
```suggestion
	flags.String(c.Prefix+corsAllowedOrigins, "", "Comma-separated COSR allowed origins. Cf. https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin")
```
Should the default really be `*`? That does not seem like a safe default.

pkg/config/corscfg/flags_test.go
```diff
@@ -0,0 +1,55 @@
...
+	t.Run("CORS Flags", func(t *testing.T) {
+		v := viper.New()
+		command := cobra.Command{}
+		flagSet := &flag.FlagSet{}
+		flagCfg := Flags{
+			Prefix: "prefix",
+		}
+		flagCfg.AddFlags(flagSet)
+		command.PersistentFlags().AddGoFlagSet(flagSet)
+		v.BindPFlags(command.PersistentFlags())
```
yurishkuro, line 42: Use `Viperize`, e.g. https://github.com/jaegertracing/jaeger/blob/587fb749ad3c680ce0b6b23fb334f594bc09ae58/cmd/query/app/flags_test.go#L35

pkg/config/corscfg/options.go
```diff
@@ -0,0 +1,20 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package corscfg
+
+type Options struct {
+	AllowedOrigins []string `mapstructure:"allowed_origins"`
+	AllowedHeaders []string `mapstructure:"allowed_headers"`
```
yurishkuro, line 19:
```suggestion
	AllowedOrigins []string
	AllowedHeaders []string
```

yurishkuro:
almost done. Please make sure all commits are signed, see CONTRIBUTING

pkg/config/corscfg/flags.go
pmuls99: I actually had the thought. But then saw the default values for Zipkin endpoint be set as * . Keeping the default value empty instead feels right?

pkg/config/corscfg/flags.go
```diff
@@ -0,0 +1,49 @@
...
+	corsAllowedHeaders = corsPrefix + ".allowed-headers"
+	corsAllowedOrigins = corsPrefix + ".allowed-origins"
+)
+
+type Flags struct {
+	Prefix string
+}
+
+func (c Flags) AddFlags(flags *flag.FlagSet) {
+	flags.String(c.Prefix+corsAllowedHeaders, "content-type", "Comma-separated CORS allowed headers, default accepts content-type. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Headers")
```
yurishkuro, line 35:
```suggestion
	flags.String(c.Prefix+corsAllowedHeaders, "content-type", "Comma-separated CORS allowed headers. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Headers")
```
context-type is not the only allowed by default header, this is why I added the link.

pkg/config/corscfg/flags.go
```diff
@@ -0,0 +1,49 @@
...
+	flags.String(c.Prefix+corsAllowedOrigins, "*", "Comma-separated CORS allowed headers, default accepts all. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin")
```
yurishkuro, line 36:
```suggestion
	flags.String(c.Prefix+corsAllowedOrigins, "*", "Comma-separated CORS allowed headers. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin")
```

pkg/config/corscfg/flags.go
pmuls99: @yurishkuro should I change this as well?

pmuls99:
Even after signing off previous commits, the DCO check doesnt seem to pass. i.e if I signoff the previous commit, it creates a new commit hash and the previous commit stays on the branch. Is there an alternate strategy you have for this @yurishkuro ?

yurishkuro:
See CONTRIBUTING_GUIDELINES.md for how to deal with unsigned commits

pmuls99:
Done, @yurishkuro  should I change the default value of Allowed origins to nil ? Previously with Zipkin endpoint , * was the default value , so I had set that. But as you said, that doesnt seem right.

yurishkuro:
Yes, I would change the default to `""`. However, I question if we should check for both fields being empty before adding the CORS handler in the first place - if user didn't configure it, maybe they don't need CORS.

pmuls99:
Yes, we could do that. I checked the implementation of CORS in the confighttp.ServerSettings, they also seem to use cors.New to configure cors. This sets the default value for Origins being All, and Headers being the sensible ones. Should we prompt the user in the usage string regarding the same ?

yurishkuro:
can you elaborate or point to code?

pmuls99:
If we absoutely dont want any CORS, we can do that. But if we set the value to "" in the cors flags default string, then here are the default values for them https://github.com/rs/cors/blob/e90f167479505c4dbe1161306c3c977f162c1442/cors.go#L136

yurishkuro:
Let's just go with blank defaults and not over-complicate the logic. Sounds like doing so will be backwards compatible with both Zipkin and OTEL endpoints.

pkg/config/corscfg/flags.go
```diff
@@ -0,0 +1,49 @@
...
+	corsAllowedOrigins = corsPrefix + ".allowed-origins"
+)
+
+type Flags struct {
+	Prefix string
+}
+
+func (c Flags) AddFlags(flags *flag.FlagSet) {
+	flags.String(c.Prefix+corsAllowedHeaders, "", "Comma-separated CORS allowed headers. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Headers")
+	flags.String(c.Prefix+corsAllowedOrigins, "", "Comma-separated CORS allowed headers. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin")
```
yurishkuro, line 36:
```suggestion
	flags.String(c.Prefix+corsAllowedOrigins, "", "Comma-separated CORS allowed origins. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin")
```

yurishkuro (approved):

yurishkuro:
Thanks!

pmuls99:
Sorry for the trouble I gave with some mistakes!

vikashpisces:
@pmuls99 While scrolling issues tab for CORS solution, I came across this PR.

I have been strruggling in configuring/ disabling cors for OTEL's http receiver endpoint on port 4318. I'm getting below error on latest all-in-one build:

Access to XMLHttpRequest at 'http://localhost:4318/v1/traces' from origin 'https://localhost:5173' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource.

I did try configuring over yaml, env and cli option for the flag: --collector.otlp.http.cors.allowed-origins

While the same port 4318 is working fine for sending otel traces from nodejs backend service, however same fails from browser application on the same endpoint http://localhost:4318/v1/traces.

Any help or direction what is going wrong or anything that I'm missing in configuring it.

Thanks in Advance.

pmuls99:
@vikashpisces , is your nodejs backend hosted somewhere and not running on the local host?

vikashpisces:
@pmuls99 Never mind, I was only passing allowed_origins and it did'nt work. I had to add allowed_headers as well and it worked, not sure why. Ideally Access-Control-Allow-Origin header should have been added without allowed_headers being not there.

yurishkuro:
@vikashpisces please post the full command that worked for you so that it's easier for people to reuse when someone comes across this

blm768:
This worked for me:

```bash
jaeger-all-in-one \
  --collector.otlp.http.cors.allowed-origins https://my.origin \
  --collector.otlp.http.cors.allowed-headers '*' \
  --collector.otlp.http.tls.enabled \
  --collector.otlp.http.tls.cert jaeger.crt \
  --collector.otlp.http.tls.key jaeger.key
```
This didn't.

```bash
jaeger-all-in-one \
  --collector.otlp.http.cors.allowed-origins https://my.origin \
  --collector.otlp.http.tls.enabled \
  --collector.otlp.http.tls.cert jaeger.crt \
  --collector.otlp.http.tls.key jaeger.key
```

The TLS stuff is probably only necessary for setups like mine that have the root page served over HTTPS.

TechWatching:
> This worked for me:
> 
> ```shell
> jaeger-all-in-one \
>   --collector.otlp.http.cors.allowed-origins https://my.origin \
>   --collector.otlp.http.cors.allowed-headers '*' \
>   --collector.otlp.http.tls.enabled \
>   --collector.otlp.http.tls.cert jaeger.crt \
>   --collector.otlp.http.tls.key jaeger.key
> ```
> 
> This didn't.
> 
> ```shell
> jaeger-all-in-one \
>   --collector.otlp.http.cors.allowed-origins https://my.origin \
>   --collector.otlp.http.tls.enabled \
>   --collector.otlp.http.tls.cert jaeger.crt \
>   --collector.otlp.http.tls.key jaeger.key
> ```
> 
> The TLS stuff is probably only necessary for setups like mine that have the root page served over HTTPS.

I can't make it work with the jaeger v2. Is there something wrong with this command: `podman  run --rm --name jaeger -p 16686:16686 -p 4318:4318 jaegertracing/jaeger  --collector.otlp.http.cors.allowed-origins=https://localhost:3000` . It tells me the flag is unknown

yurishkuro:
Jaeger v2 takes a config file, not cli flags. https://www.jaegertracing.io/docs/2.4/configuration/

TechWatching:
> Jaeger v2 takes a config file, not cli flags. https://www.jaegertracing.io/docs/2.4/configuration/

Thanks, that explains why. Do you know what the config file would look like to do that ? I have trouble seing all the available properties, I guess it's because the documentation is still in construction.

TechWatching:
I started with this config file but it does not seem to be enough to make it work (I have EMPTY RESPONSE errors now):

```yaml
receivers:
  otlp:
    protocols:
      http:
        include_metadata: true
        cors:
          allowed_origins:
            - https://localhost:3000
exporters:
  debug:
    verbosity: detailed
service:
  extensions: [jaeger_query, jaeger_storage]
  pipelines:
    traces:
      receivers: [otlp]
      exporters: [debug]

extensions:
  jaeger_query:
    storage:
      traces: query_storage
  jaeger_storage:
    backends:
      query_storage:
        memory:
          max_traces: 100000
```

alvaradojl:
@TechWatching I also get the same behaviour, empty response. 
this is my full config.

```
receivers:
  otlp:
    protocols:
      http:
        endpoint: 0.0.0.0:4318
        cors:
          allowed_origins:
            - http://localhost:3000  # Your SvelteKit dev server
          allowed_headers:
            - "*"

processors:
  batch:

exporters:
  jaeger_storage_exporter:
    trace_storage: memory_store
  debug:
    verbosity: detailed

service:
  extensions: [jaeger_storage, jaeger_query, healthcheckv2]
  pipelines:
    traces:
      receivers: [otlp]
      processors: [batch]
      exporters: [jaeger_storage_exporter, debug]

extensions:
  healthcheckv2:
    use_v2: true
    http:
  
  jaeger_query:
    storage:
      traces: memory_store
    ui:
      config_file: ./cmd/jaeger/config-ui.json

  jaeger_storage:
    backends:
      memory_store:
        memory:
          max_traces: 50000

```

Same receiver config in opentelemetry-collector-contrib works as expected.

yurishkuro:
> Same receiver config in opentelemetry-collector-contrib works as expected.

@alvaradojl why would that be? Jaeger is using the same receiver from OTEL.

PR #4595, opened 2023-07-23
feat: Add `OTEL tracer` to metricsstore reader component

afzal442:
* Part of #3381 
* This PR adds OTEL Provider to metrics reader component

- Replaces OT tracer with OTEL tracer to support jtracer pkg

### DIFF
--- a/pkg/jtracer/jtracer.go
+++ b/pkg/jtracer/jtracer.go
@@ -61,7 +61,7 @@ func New(serviceName string) (*JTracer, error) {
 }
 
 func NoOp() *JTracer {
-	return &JTracer{OT: opentracing.NoopTracer{}, OTEL: trace.NewNoopTracerProvider(), closer: func(ctx context.Context) error { return nil }}
+	return &JTracer{OT: opentracing.NoopTracer{}, OTEL: trace.NewNoopTracerProvider(), closer: nil}
 }
 
 // initOTEL initializes OTEL Tracer
@@ -107,5 +107,8 @@ func otelExporter(ctx context.Context) (sdktrace.SpanExporter, error) {
 
 // Shutdown the tracerProvider to clean up resources
 func (jt *JTracer) Close(ctx context.Context) error {
-	return jt.closer(ctx)
+	if jt.closer != nil {
+		return jt.closer(ctx)
+	}
+	return nil
 }
--- a/pkg/jtracer/jtracer.go
+++ b/pkg/jtracer/jtracer.go
@@ -61,7 +61,7 @@ func New(serviceName string) (*JTracer, error) {
 }
 
 func NoOp() *JTracer {
-	return &JTracer{OT: opentracing.NoopTracer{}, OTEL: trace.NewNoopTracerProvider()}
+	return &JTracer{OT: opentracing.NoopTracer{}, OTEL: trace.NewNoopTracerProvider(), closer: func(ctx context.Context) error { return nil }}
 }
 
 // initOTEL initializes OTEL Tracer
--- a/plugin/metrics/prometheus/factory.go
+++ b/plugin/metrics/prometheus/factory.go
@@ -18,6 +18,8 @@ import (
 	"flag"
 
 	"github.com/spf13/viper"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/trace"
 	"go.uber.org/zap"
 
 	"github.com/jaegertracing/jaeger/plugin"
@@ -31,11 +33,13 @@ var _ plugin.Configurable = (*Factory)(nil)
 type Factory struct {
 	options *Options
 	logger  *zap.Logger
+	tracer  trace.TracerProvider
 }
 
 // NewFactory creates a new Factory.
 func NewFactory() *Factory {
 	return &Factory{
+		tracer:  otel.GetTracerProvider(),
 		options: NewOptions("prometheus"),
 	}
 }
@@ -60,5 +64,5 @@ func (f *Factory) Initialize(logger *zap.Logger) error {
 
 // CreateMetricsReader implements storage.MetricsFactory.
 func (f *Factory) CreateMetricsReader() (metricsstore.Reader, error) {
-	return prometheusstore.NewMetricsReader(f.logger, f.options.Primary.Configuration)
+	return prometheusstore.NewMetricsReader(f.options.Primary.Configuration, f.logger, f.tracer)
 }
--- a/plugin/metrics/prometheus/factory.go
+++ b/plugin/metrics/prometheus/factory.go
@@ -16,10 +16,12 @@ package prometheus
 
 import (
 	"flag"
+	"log"
 
 	"github.com/spf13/viper"
 	"go.uber.org/zap"
 
+	"github.com/jaegertracing/jaeger/pkg/jtracer"
 	"github.com/jaegertracing/jaeger/plugin"
 	prometheusstore "github.com/jaegertracing/jaeger/plugin/metrics/prometheus/metricsstore"
 	"github.com/jaegertracing/jaeger/storage/metricsstore"
@@ -31,12 +33,18 @@ var _ plugin.Configurable = (*Factory)(nil)
 type Factory struct {
 	options *Options
 	logger  *zap.Logger
+	tracer  *jtracer.JTracer
 }
 
 // NewFactory creates a new Factory.
 func NewFactory() *Factory {
+	jt, err := jtracer.New("metricsreader")
+	if err != nil {
+		log.Fatal("Failed to initialize tracer: %w", err)
+	}
 	return &Factory{
 		options: NewOptions("prometheus"),
+		tracer:  jt,
 	}
 }
 
@@ -60,5 +68,5 @@ func (f *Factory) Initialize(logger *zap.Logger) error {
 
 // CreateMetricsReader implements storage.MetricsFactory.
 func (f *Factory) CreateMetricsReader() (metricsstore.Reader, error) {
-	return prometheusstore.NewMetricsReader(f.logger, f.options.Primary.Configuration)
+	return prometheusstore.NewMetricsReader(f.logger, f.options.Primary.Configuration, f.tracer)
 }
--- a/plugin/metrics/prometheus/factory.go
+++ b/plugin/metrics/prometheus/factory.go
@@ -16,12 +16,10 @@ package prometheus
 
 import (
 	"flag"
-	"log"
 
 	"github.com/spf13/viper"
 	"go.uber.org/zap"
 
-	"github.com/jaegertracing/jaeger/pkg/jtracer"
 	"github.com/jaegertracing/jaeger/plugin"
 	prometheusstore "github.com/jaegertracing/jaeger/plugin/metrics/prometheus/metricsstore"
 	"github.com/jaegertracing/jaeger/storage/metricsstore"
@@ -33,18 +31,12 @@ var _ plugin.Configurable = (*Factory)(nil)
 type Factory struct {
 	options *Options
 	logger  *zap.Logger
-	tracer  *jtracer.JTracer
 }
 
 // NewFactory creates a new Factory.
 func NewFactory() *Factory {
-	jt, err := jtracer.New("metricsreader")
-	if err != nil {
-		log.Fatal("Failed to initialize tracer: %w", err)
-	}
 	return &Factory{
 		options: NewOptions("prometheus"),
-		tracer:  jt,
 	}
 }
 
@@ -68,5 +60,5 @@ func (f *Factory) Initialize(logger *zap.Logger) error {
 
 // CreateMetricsReader implements storage.MetricsFactory.
 func (f *Factory) CreateMetricsReader() (metricsstore.Reader, error) {
-	return prometheusstore.NewMetricsReader(f.logger, f.options.Primary.Configuration, f.tracer)
+	return prometheusstore.NewMetricsReader(f.logger, f.options.Primary.Configuration)
 }
--- a/plugin/metrics/prometheus/metricsstore/reader.go
+++ b/plugin/metrics/prometheus/metricsstore/reader.go
@@ -26,14 +26,15 @@ import (
 	"time"
 	"unicode"
 
-	"github.com/opentracing/opentracing-go"
-	ottag "github.com/opentracing/opentracing-go/ext"
-	otlog "github.com/opentracing/opentracing-go/log"
 	"github.com/prometheus/client_golang/api"
 	promapi "github.com/prometheus/client_golang/api/prometheus/v1"
+	"go.opentelemetry.io/otel/attribute"
+	semconv "go.opentelemetry.io/otel/semconv/v1.20.0"
+	"go.opentelemetry.io/otel/trace"
 	"go.uber.org/zap"
 
 	"github.com/jaegertracing/jaeger/pkg/bearertoken"
+	"github.com/jaegertracing/jaeger/pkg/jtracer"
 	"github.com/jaegertracing/jaeger/pkg/prometheus/config"
 	"github.com/jaegertracing/jaeger/plugin/metrics/prometheus/metricsstore/dbmodel"
 	"github.com/jaegertracing/jaeger/proto-gen/api_v2/metrics"
@@ -54,6 +55,7 @@ type (
 		latencyMetricName string
 		callsMetricName   string
 		operationLabel    string
+		tracer            *jtracer.JTracer
 	}
 
 	promQueryParams struct {
@@ -73,7 +75,7 @@ type (
 )
 
 // NewMetricsReader returns a new MetricsReader.
-func NewMetricsReader(logger *zap.Logger, cfg config.Configuration) (*MetricsReader, error) {
+func NewMetricsReader(logger *zap.Logger, cfg config.Configuration, jt *jtracer.JTracer) (*MetricsReader, error) {
 	logger.Info("Creating metrics reader", zap.Any("configuration", cfg))
 
 	roundTripper, err := getHTTPRoundTripper(&cfg, logger)
@@ -101,6 +103,7 @@ func NewMetricsReader(logger *zap.Logger, cfg config.Configuration) (*MetricsRea
 		callsMetricName:   buildFullCallsMetricName(cfg),
 		latencyMetricName: buildFullLatencyMetricName(cfg),
 		operationLabel:    operationLabel,
+		tracer:            jt,
 	}
 
 	logger.Info("Prometheus reader initialized", zap.String("addr", cfg.ServerURL))
@@ -224,8 +227,11 @@ func (m MetricsReader) executeQuery(ctx context.Context, p metricsQueryParams) (
 	}
 	promQuery := m.buildPromQuery(p)
 
-	span, ctx := startSpanForQuery(ctx, p.metricName, promQuery)
-	defer span.Finish()
+	ctx, span := startSpanForQuery(ctx, p.metricName, promQuery, m.tracer.OTEL)
+	defer span.End()
+	if err := m.tracer.Close(ctx); err != nil {
+		m.logger.Error("Error shutting down tracer provider", zap.Error(err))
+	}
 
 	queryRange := promapi.Range{
 		Start: p.EndTime.Add(-1 * *p.Lookback),
@@ -287,17 +293,18 @@ func promqlDurationString(d *time.Duration) string {
 	return string(b)
 }
 
-func startSpanForQuery(ctx context.Context, metricName, query string) (opentracing.Span, context.Context) {
-	span, ctx := opentracing.StartSpanFromContext(ctx, metricName)
-	ottag.DBStatement.Set(span, query)
-	ottag.DBType.Set(span, "prometheus")
-	ottag.Component.Set(span, "promql")
-	return span, ctx
+func startSpanForQuery(ctx context.Context, metricName, query string, tp trace.TracerProvider) (context.Context, trace.Span) {
+	ctx, span := tp.Tracer("prom-metrics-reader").Start(ctx, metricName)
+	span.SetAttributes(
+		attribute.Key(semconv.DBStatementKey).String(query),
+		attribute.Key("db.type").String("prometheus"),
+		attribute.Key("component").String("promql"),
+	)
+	return ctx, span
 }
 
-func logErrorToSpan(span opentracing.Span, err error) {
-	ottag.Error.Set(span, true)
-	span.LogFields(otlog.Error(err))
+func logErrorToSpan(span trace.Span, err error) {
+	span.AddEvent(err.Error(), trace.WithAttributes(semconv.OTelStatusCodeError))
 }
 
 func getHTTPRoundTripper(c *config.Configuration, logger *zap.Logger) (rt http.RoundTripper, err error) {
--- a/plugin/metrics/prometheus/metricsstore/reader.go
+++ b/plugin/metrics/prometheus/metricsstore/reader.go
@@ -28,7 +28,6 @@ import (
 
 	"github.com/prometheus/client_golang/api"
 	promapi "github.com/prometheus/client_golang/api/prometheus/v1"
-	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
 	semconv "go.opentelemetry.io/otel/semconv/v1.20.0"
 	"go.opentelemetry.io/otel/trace"
@@ -50,6 +49,7 @@ type (
 	MetricsReader struct {
 		client promapi.API
 		logger *zap.Logger
+		tracer trace.Tracer
 
 		metricsTranslator dbmodel.Translator
 		latencyMetricName string
@@ -74,7 +74,7 @@ type (
 )
 
 // NewMetricsReader returns a new MetricsReader.
-func NewMetricsReader(logger *zap.Logger, cfg config.Configuration) (*MetricsReader, error) {
+func NewMetricsReader(cfg config.Configuration, logger *zap.Logger, tracer trace.TracerProvider) (*MetricsReader, error) {
 	logger.Info("Creating metrics reader", zap.Any("configuration", cfg))
 
 	roundTripper, err := getHTTPRoundTripper(&cfg, logger)
@@ -97,6 +97,7 @@ func NewMetricsReader(logger *zap.Logger, cfg config.Configuration) (*MetricsRea
 	mr := &MetricsReader{
 		client: promapi.NewAPI(client),
 		logger: logger,
+		tracer: tracer.Tracer("prom-metrics-reader"),
 
 		metricsTranslator: dbmodel.New(operationLabel),
 		callsMetricName:   buildFullCallsMetricName(cfg),
@@ -225,7 +226,7 @@ func (m MetricsReader) executeQuery(ctx context.Context, p metricsQueryParams) (
 	}
 	promQuery := m.buildPromQuery(p)
 
-	ctx, span := startSpanForQuery(ctx, p.metricName, promQuery)
+	ctx, span := startSpanForQuery(ctx, p.metricName, promQuery, m.tracer)
 	defer span.End()
 
 	queryRange := promapi.Range{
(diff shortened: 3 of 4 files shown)


plugin/metrics/prometheus/factory.go
```diff
@@ -31,12 +33,18 @@ var _ plugin.Configurable = (*Factory)(nil)
 type Factory struct {
 	options *Options
 	logger  *zap.Logger
+	tracer  *jtracer.JTracer
 }
 
 // NewFactory creates a new Factory.
 func NewFactory() *Factory {
+	jt, err := jtracer.New("metricsreader")
```
yurishkuro, line 41: You have to utilize tracer provider already created in main.

plugin/metrics/prometheus/metricsstore/reader.go
```diff
@@ -224,8 +227,11 @@ func (m MetricsReader) executeQuery(ctx context.Context, p metricsQueryParams) (
 	}
 	promQuery := m.buildPromQuery(p)
 
-	span, ctx := startSpanForQuery(ctx, p.metricName, promQuery)
-	defer span.Finish()
+	ctx, span := startSpanForQuery(ctx, p.metricName, promQuery, m.tracer.OTEL)
+	defer span.End()
+	if err := m.tracer.Close(ctx); err != nil {
+		m.logger.Error("Error shutting down tracer provider", zap.Error(err))
+	}
```
yurishkuro, line 234:
```suggestion
```

plugin/metrics/prometheus/metricsstore/reader.go
```diff
@@ -287,17 +293,18 @@ func promqlDurationString(d *time.Duration) string {
...
-	span, ctx := opentracing.StartSpanFromContext(ctx, metricName)
-	ottag.DBStatement.Set(span, query)
-	ottag.DBType.Set(span, "prometheus")
-	ottag.Component.Set(span, "promql")
-	return span, ctx
+func startSpanForQuery(ctx context.Context, metricName, query string, tp trace.TracerProvider) (context.Context, trace.Span) {
+	ctx, span := tp.Tracer("prom-metrics-reader").Start(ctx, metricName)
+	span.SetAttributes(
+		attribute.Key(semconv.DBStatementKey).String(query),
+		attribute.Key("db.type").String("prometheus"),
```
yurishkuro, line 300: Pretty sure there is otel key for this

plugin/metrics/prometheus/metricsstore/reader.go
```diff
@@ -287,17 +293,18 @@ func promqlDurationString(d *time.Duration) string {
...
+		attribute.Key("component").String("promql"),
+	)
+	return ctx, span
 }
 
-func logErrorToSpan(span opentracing.Span, err error) {
-	ottag.Error.Set(span, true)
-	span.LogFields(otlog.Error(err))
+func logErrorToSpan(span trace.Span, err error) {
+	span.AddEvent(err.Error(), trace.WithAttributes(semconv.OTelStatusCodeError))
```
yurishkuro, line 307: What does span.RecordError do?

plugin/metrics/prometheus/metricsstore/reader.go
```diff
@@ -73,7 +75,7 @@ type (
 )
 
 // NewMetricsReader returns a new MetricsReader.
-func NewMetricsReader(logger *zap.Logger, cfg config.Configuration) (*MetricsReader, error) {
+func NewMetricsReader(logger *zap.Logger, cfg config.Configuration, jt *jtracer.JTracer) (*MetricsReader, error) {
```
yurishkuro, line 78: By convention, observability components like logger and tracer are passed as last arguments. Since you're changing signature anyway, let's fix it.

plugin/metrics/prometheus/metricsstore/reader_test.go
```diff
@@ -639,12 +645,13 @@ func prepareMetricsReaderAndServer(t *testing.T, config config.Configuration, wa
 	mockPrometheus := startMockPrometheusServer(t, wantPromQlQuery, wantWarnings)
 
 	logger := zap.NewNop()
+	tracer := jtracer.NoOp()
```
yurishkuro, line 648: Looks like the tests are not testing tracing code. I would create a recording tracer here and add some validation in certain tests that the correct span is created.

plugin/metrics/prometheus/factory.go
afzal442: Thanks. That means I can use it from `cmd/all-in-one` and `cmd/app/query` by passing the jtracer to `storageFactory.InitFromViper(v, logger, jtracer)`.

yurishkuro: yes. And you don't need to worry about closing it.

pkg/jtracer/jtracer.go
```diff
@@ -61,7 +61,7 @@ func New(serviceName string) (*JTracer, error) {
 }
 
 func NoOp() *JTracer {
-	return &JTracer{OT: opentracing.NoopTracer{}, OTEL: trace.NewNoopTracerProvider()}
+	return &JTracer{OT: opentracing.NoopTracer{}, OTEL: trace.NewNoopTracerProvider(), closer: func(ctx context.Context) error { return nil }}
```
yurishkuro, line 64: instead, you could do a check in the Close function if `jt.closer != nil`

plugin/metrics/prometheus/metricsstore/reader.go
```diff
@@ -287,17 +288,19 @@ func promqlDurationString(d *time.Duration) string {
...
 
-func startSpanForQuery(ctx context.Context, metricName, query string) (opentracing.Span, context.Context) {
-	span, ctx := opentracing.StartSpanFromContext(ctx, metricName)
-	ottag.DBStatement.Set(span, query)
-	ottag.DBType.Set(span, "prometheus")
-	ottag.Component.Set(span, "promql")
-	return span, ctx
+func startSpanForQuery(ctx context.Context, metricName, query string) (context.Context, trace.Span) {
+	tp := otel.GetTracerProvider()
+	ctx, span := tp.Tracer("prom-metrics-reader").Start(ctx, metricName)
```
yurishkuro, line 293: This is almost correct, but you don't want to be asking for the tracer on critical path of every request, since the tracer never changes. Instead, you can create the tracer in `NewMetricsReader` and store it as member variable.

yurishkuro: even cleaner is to make the Factory call `GetTracerProvider()` and pass the provider to `NewMetricsReader`

plugin/metrics/prometheus/metricsstore/reader_test.go
```diff
@@ -61,22 +64,36 @@ var defaultConfig = config.Configuration{
 	LatencyUnit:                 "ms",
 }
 
+func createTracingProvider() trace.TracerProvider {
+	exporter, err := stdout.New(stdout.WithPrettyPrint())
```
yurishkuro, line 68: We don't need tests printing traces. 
```suggestion
	exporter := tracetest.NewInMemoryExporter()
```
Add assertions to some tests to check that the spans were created correctly.

plugin/metrics/prometheus/metricsstore/reader_test.go
```diff
@@ -61,22 +64,36 @@ var defaultConfig = config.Configuration{
...
+	if err != nil {
+		panic(err)
+	}
+	tp := sdktrace.NewTracerProvider(
+		sdktrace.WithSampler(sdktrace.AlwaysSample()),
+		sdktrace.WithBatcher(exporter),
+	)
+	return tp
```
yurishkuro, line 76:
```suggestion
	return tp, exporter
```

plugin/metrics/prometheus/metricsstore/reader_test.go
```diff
@@ -61,22 +64,36 @@ var defaultConfig = config.Configuration{
 	LatencyUnit:                 "ms",
 }
 
+func createTracingProvider() trace.TracerProvider {
+	exporter, err := stdout.New(stdout.WithPrettyPrint())
+	if err != nil {
+		panic(err)
```
yurishkuro, line 70: you won't need it because of above change, but if you did, the correct way to do this is to pass `t` into the function and do `require.NoError(t, err)`

afzal442: Thanks much. Do I also need to change the name of the function? like `TestInitTracingProvider`.

plugin/metrics/prometheus/metricsstore/reader_test.go
afzal442: Sorry! I couldn't get that. Why `exporter`?

plugin/metrics/prometheus/metricsstore/reader_test.go
yurishkuro: no, what you have is fine. You could make it shorter `tracerProvider(t)`. You certainly don't want to name it TestXxx, as that would make it an actual unit test, while it's just a helper function.

plugin/metrics/prometheus/metricsstore/reader_test.go
yurishkuro: because that allows you to do `exporter.GetSpans()` and validate that the correct spans were created

afzal442: @yurishkuro while testing out I am not getting spans to validate that the correct ones were created
`assert.NotEmpty(t, exp.GetSpans(), "No spans recorded during the test.")`
```
Error:          Should NOT be empty, but was []
                Test:           TestInvalidCertFile
                Messages:       No spans recorded during the test.
```
LMK if I have missed somewhere

plugin/metrics/prometheus/metricsstore/reader_test.go
```diff
@@ -64,36 +64,36 @@ var defaultConfig = config.Configuration{
...
-		panic(err)
-	}
+func tracerProvider(t *testing.T) (trace.TracerProvider, *tracetest.InMemoryExporter) {
+	exporter := tracetest.NewInMemoryExporter()
 	tp := sdktrace.NewTracerProvider(
 		sdktrace.WithSampler(sdktrace.AlwaysSample()),
 		sdktrace.WithBatcher(exporter),
 	)
-	return tp
+	defer tp.Shutdown(context.Background())
```
yurishkuro, line 73: defer fires when you exit this function. You're disabling provider even before you returned it.

afzal442: I tried testing this out without shutdown, but still failed to get span recorded.

yurishkuro: shutdown needs to be called from test functions
```
tp, exporter = tracerProvider(t)
defer tp.Shutdown()
```

yurishkuro: and you don't want `sdktrace.WithBatcher(exporter)`, you want to export immediately. Maybe `withSyncer`

afzal442: Ah, I see. I tried that option `WithSyncer`, but no chance. 
And If I change its argument to use `sdktrace`, `NewMetricsReader()` will have to be changed to support it. Right? But then, `otel.GetTr..()` doesn't support it.

plugin/metrics/prometheus/factory.go
```diff
@@ -31,11 +33,13 @@ var _ plugin.Configurable = (*Factory)(nil)
 type Factory struct {
 	options *Options
 	logger  *zap.Logger
+	tracer  trace.TracerProvider
 }
 
 // NewFactory creates a new Factory.
 func NewFactory() *Factory {
 	return &Factory{
+		tracer:  otel.GetTracerProvider(),
```
afzal442, line 42: I mean this one.

plugin/metrics/prometheus/metricsstore/reader_test.go
```diff
@@ -85,14 +102,16 @@ func TestNewMetricsReaderInvalidAddress(t *testing.T) {
...
 	require.NoError(t, err)
 	assert.NotNil(t, listener)
 
-	reader, err := NewMetricsReader(logger, config.Configuration{
+	reader, err := NewMetricsReader(config.Configuration{
 		ServerURL:      "http://" + listener.Addr().String(),
 		ConnectTimeout: defaultTimeout,
-	})
+	}, logger, tracer)
+	assert.NotEmpty(t, exp.GetSpans(), "No spans recorded during the test.")
```
yurishkuro, line 114: why do you expect spans created at this point? There is no instrumentation in `NewMetricsReader` function. You know which functions do have it, it's in this PR, so only assert in tests for those functions.

afzal442: Now I got that. I found three but they don't invoke `NewMetricsReader` function. e.g. `TestGetLatencies` only tests against query svc.

yurishkuro: `TestGetLatencies` calls `prepareMetricsReaderAndServer` which calls `NewMetricsReader`

afzal442: On it.

plugin/metrics/prometheus/metricsstore/reader_test.go
```diff
@@ -64,42 +64,49 @@ var defaultConfig = config.Configuration{
...
 
-func tracerProvider(t *testing.T) (trace.TracerProvider, *tracetest.InMemoryExporter) {
+func tracerProvider(t *testing.T) (trace.TracerProvider, *tracetest.InMemoryExporter, func(ctx context.Context) error) {
 	exporter := tracetest.NewInMemoryExporter()
 	tp := sdktrace.NewTracerProvider(
 		sdktrace.WithSampler(sdktrace.AlwaysSample()),
 		sdktrace.WithSyncer(exporter),
 	)
-	return tp, exporter
+	closer := func(ctx context.Context) error {
```
yurishkuro, line 73: you don't need to return a closer, the callers can call tp.Shutdown themselves, you do not provide any extra capability by wrapping it.

plugin/metrics/prometheus/metricsstore/reader_test.go
```diff
@@ -64,42 +64,49 @@ var defaultConfig = config.Configuration{
...
 	logger := zap.NewNop()
-	tracer, _ := tracerProvider(t)
+	tracer, _, closer := tracerProvider(t)
 	reader, err := NewMetricsReader(config.Configuration{
 		ServerURL:      "http://localhost:1234",
 		ConnectTimeout: defaultTimeout,
 	}, logger, tracer)
 	require.NoError(t, err)
 	assert.NotNil(t, reader)
+	err = closer(context.Background())
```
yurishkuro, line 88: closer should be invoked via defer right after the object is created. Otherwise require in L86 may panic and you would never reach this line.

plugin/metrics/prometheus/metricsstore/reader_test.go
yurishkuro: however, what you COULD do is this:
```
closer := func() {
		assert.NoError(t, tp.Shutdown(context.Background())
	}
```
This way you make it very simple for the caller to do `defer closer()`

plugin/metrics/prometheus/metricsstore/reader_test.go
afzal442: I see, I will add defer to it.

plugin/metrics/prometheus/metricsstore/reader_test.go
afzal442: As I told, the return type is trace.TracerProvider which is not sdktrace. If I change it, `otel.Tp` won't support it.

plugin/metrics/prometheus/metricsstore/reader_test.go
```diff
@@ -121,18 +145,21 @@ func TestMetricsServerError(t *testing.T) {
...
 		ServerURL:      "http://" + address,
 		ConnectTimeout: defaultTimeout,
-	})
+	}, logger, tracer)
 	require.NoError(t, err)
 
 	m, err := reader.GetCallRates(context.Background(), &params)
 	assert.NotNil(t, m)
-
+	assert.NotEmpty(t, exp.GetSpans(), "No spans recorded during the test.")
```
yurishkuro, line 158: 1. this code would read better if the message was "spans should be recorded". Otherwise when reading this code all I see is "no spans recorded" - is that good/expected?.
2. why not validate the exact number of spans you expect?

plugin/metrics/prometheus/metricsstore/reader_test.go
```diff
@@ -99,8 +98,7 @@ func TestNewMetricsReaderInvalidAddress(t *testing.T) {
 	require.Error(t, err)
 	assert.Contains(t, err.Error(), "failed to initialize prometheus client")
 	assert.Nil(t, reader)
-	err = closer(context.Background())
-	require.NoError(t, err)
+	defer closer()
```
yurishkuro, line 101: reposting from https://github.com/jaegertracing/jaeger/pull/4595#pullrequestreview-1549000731

> closer should be invoked via defer **right after the object is created**. Otherwise `require` before it may panic and you would never reach this line.

afzal442: my bad. if panic, closer() won't be executed.

plugin/metrics/prometheus/metricsstore/reader_test.go
```diff
@@ -121,16 +142,19 @@ func TestMetricsServerError(t *testing.T) {
...
 		ServerURL:      "http://" + address,
 		ConnectTimeout: defaultTimeout,
-	})
+	}, logger, tracer)
 	require.NoError(t, err)
 
 	m, err := reader.GetCallRates(context.Background(), &params)
+	assert.NotEmpty(t, exp.GetSpans(), "Expected spans are recorded")
+	assert.Len(t, exp.GetSpans(), 1, "HTTP request was traced and span reported")
+	defer closer()
```
yurishkuro, line 156: Too late

plugin/metrics/prometheus/metricsstore/reader_test.go
```diff
@@ -221,14 +245,18 @@ func TestGetLatencies(t *testing.T) {
...
 			}
-			reader, mockPrometheus := prepareMetricsReaderAndServer(t, cfg, tc.wantPromQlQuery, nil)
+			reader, mockPrometheus := prepareMetricsReaderAndServer(t, cfg, tc.wantPromQlQuery, nil, tracer)
 			defer mockPrometheus.Close()
 
 			m, err := reader.GetLatencies(context.Background(), &params)
+			assert.NotEmpty(t, exp.GetSpans(), "Spans recorded during the test.")
+			assert.Len(t, exp.GetSpans(), 1, "HTTP request was traced and span reported")
+			defer closer()
 			require.NoError(t, err)
```
yurishkuro, line 260: Too late

plugin/metrics/prometheus/metricsstore/reader_test.go
```diff
@@ -318,14 +346,18 @@ func TestGetCallRates(t *testing.T) {
...
 				cfg = tc.updateConfig(cfg)
 			}
-			reader, mockPrometheus := prepareMetricsReaderAndServer(t, cfg, tc.wantPromQlQuery, nil)
+			reader, mockPrometheus := prepareMetricsReaderAndServer(t, cfg, tc.wantPromQlQuery, nil, tracer)
 			defer mockPrometheus.Close()
 
 			m, err := reader.GetCallRates(context.Background(), &params)
+			assert.NotEmpty(t, exp.GetSpans(), "Spans recorded during the test.")
+			assert.Len(t, exp.GetSpans(), 1, "HTTP request was traced and span reported")
+			defer closer()
```
yurishkuro, line 360: Too late

plugin/metrics/prometheus/metricsstore/reader_test.go
```diff
@@ -318,14 +346,18 @@ func TestGetCallRates(t *testing.T) {
...
 			cfg := defaultConfig
 			if tc.updateConfig != nil {
 				cfg = tc.updateConfig(cfg)
 			}
-			reader, mockPrometheus := prepareMetricsReaderAndServer(t, cfg, tc.wantPromQlQuery, nil)
+			reader, mockPrometheus := prepareMetricsReaderAndServer(t, cfg, tc.wantPromQlQuery, nil, tracer)
 			defer mockPrometheus.Close()
 
 			m, err := reader.GetCallRates(context.Background(), &params)
+			assert.NotEmpty(t, exp.GetSpans(), "Spans recorded during the test.")
```
yurishkuro, line 358: NotEmpty check is redundant when followed by Len=1 check.

plugin/metrics/prometheus/metricsstore/reader_test.go
```diff
@@ -121,16 +142,19 @@ func TestMetricsServerError(t *testing.T) {
...
+	reader, err := NewMetricsReader(config.Configuration{
 		ServerURL:      "http://" + address,
 		ConnectTimeout: defaultTimeout,
-	})
+	}, logger, tracer)
 	require.NoError(t, err)
 
 	m, err := reader.GetCallRates(context.Background(), &params)
+	assert.NotEmpty(t, exp.GetSpans(), "Expected spans are recorded")
+	assert.Len(t, exp.GetSpans(), 1, "HTTP request was traced and span reported")
```
yurishkuro, line 155: Since server is returning an error, I assume the span status should also be error, which we can assert.

plugin/metrics/prometheus/metricsstore/reader_test.go
```diff
@@ -85,14 +104,16 @@ func TestNewMetricsReaderInvalidAddress(t *testing.T) {
...
 	listener, err := net.Listen("tcp", "localhost:")
 	require.NoError(t, err)
 	assert.NotNil(t, listener)
 
-	reader, err := NewMetricsReader(logger, config.Configuration{
+	reader, err := NewMetricsReader(config.Configuration{
 		ServerURL:      "http://" + listener.Addr().String(),
 		ConnectTimeout: defaultTimeout,
-	})
+	}, logger, tracer)
```
yurishkuro, line 115: GetMinDtepDuration sounds like something that should produce traces.

plugin/metrics/prometheus/metricsstore/reader_test.go
afzal442: sorry! why we got it too late?

plugin/metrics/prometheus/metricsstore/reader_test.go
afzal442: That means I have to rmv every assert NotEmpty check.

plugin/metrics/prometheus/metricsstore/reader_test.go
yurishkuro: Defer close is always called as soon as object is created.

plugin/metrics/prometheus/metricsstore/reader_test.go
afzal442: Nope, it doesn't.

plugin/metrics/prometheus/metricsstore/reader_test.go
afzal442: Do you mean exp.GetSpans()[0].Status ?

yurishkuro: yes, but not checking it for null, checking it for expected value.
Two other problems:
  1. the asserts are in the wrong place. This test is called TestMetricsServerError. It is testing specific business functionality. Testing spans is a side effect, it should be done later, not as the very first (i.e. most important) thing after calling the business function.
  2. The order of the two asserts doesn't make sense. If the list of spans is empty, then `Spans()[0]` panics, so you never even get to assert.Len(). If you switch the order, it's still incorrect because `assert` does not abort the test, so if the list of spans is empty, then `assert.Len` will log an error and continue, but `Spans()[0]` will panic again. The right way to approach this is to make first check a `require` instead of `assert`. `require` aborts the test if not satisfied.

afzal442: Insightful noted. 🚀

plugin/metrics/prometheus/metricsstore/reader_test.go
```diff
@@ -121,18 +142,20 @@ func TestMetricsServerError(t *testing.T) {
...
 	require.NoError(t, err)
-
 	m, err := reader.GetCallRates(context.Background(), &params)
 	assert.NotNil(t, m)
-
 	require.Error(t, err)
 	assert.Contains(t, err.Error(), "failed executing metrics query")
+	require.Len(t, exp.GetSpans(), 1, "HTTP request was traced and span reported")
+	assert.Equal(t, exp.GetSpans()[0].Name, "service_call_rate")
```
afzal442, line 158: `exp.GetSpans()[0].Status` is having null values in trace.Status fmt.

yurishkuro, line 158: please read the signature of Equal() more carefully and find a mistake

afzal442: so silly. on it.

plugin/metrics/prometheus/metricsstore/reader.go
```diff
@@ -224,8 +226,8 @@ func (m MetricsReader) executeQuery(ctx context.Context, p metricsQueryParams) (
 	}
 	promQuery := m.buildPromQuery(p)
 
-	span, ctx := startSpanForQuery(ctx, p.metricName, promQuery)
-	defer span.Finish()
+	ctx, span := startSpanForQuery(ctx, p.metricName, promQuery, m.tracer)
+	defer span.End()
 
 	queryRange := promapi.Range{
 		Start: p.EndTime.Add(-1 * *p.Lookback),
```
yurishkuro, line 233: GH doesn't allow commenting below. On L239 I suggest to change it to
```
	if err != nil {
		err = fmt.Errorf("failed executing metrics query: %w", err)
		logErrorToSpan(span, err)
		return &metrics.MetricFamily{}, err
	}
```

plugin/metrics/prometheus/metricsstore/reader_test.go
afzal442: the value comes up is `trace.Status(trace.Status{Code:0x0, Description:""})` which is null contained.

yurishkuro: that means you're not setting it correctly

afzal442: Again stuck. Not able to figure out what you said.

yurishkuro: when an error happens, your code needs to set span.Status. From what I see you're only doing span.RecordError, that does not set the status.

plugin/metrics/prometheus/metricsstore/reader.go
```diff
@@ -297,6 +298,7 @@ func startSpanForQuery(ctx context.Context, metricName, query string, tp trace.T
 		attribute.Key(semconv.DBSystemKey).String("prometheus"),
 		attribute.Key("component").String("promql"),
 	)
+	span.SetStatus(codes.Ok, "Span started")
```
yurishkuro, line 301: there is no need to set OK status explicitly, only error

plugin/metrics/prometheus/metricsstore/reader.go
```diff
@@ -287,17 +291,19 @@ func promqlDurationString(d *time.Duration) string {
...
+		attribute.Key("component").String("promql"),
+	)
+	return ctx, span
 }
 
-func logErrorToSpan(span opentracing.Span, err error) {
-	ottag.Error.Set(span, true)
-	span.LogFields(otlog.Error(err))
+func logErrorToSpan(span trace.Span, err error) {
+	span.RecordError(err, trace.WithAttributes(semconv.OTelStatusCodeError))
```
yurishkuro, line 305: what do we get from `trace.WithAttributes(semconv.OTelStatusCodeError)`?

plugin/metrics/prometheus/metricsstore/reader_test.go
```diff
@@ -121,18 +142,20 @@ func TestMetricsServerError(t *testing.T) {
...
+	}, logger, tracer)
 	require.NoError(t, err)
-
 	m, err := reader.GetCallRates(context.Background(), &params)
 	assert.NotNil(t, m)
-
 	require.Error(t, err)
 	assert.Contains(t, err.Error(), "failed executing metrics query")
+	require.Len(t, exp.GetSpans(), 1, "HTTP request was traced and span reported")
+	assert.Equal(t, "Error", exp.GetSpans()[0].Status.Code.String())
```
yurishkuro, line 158:
```suggestion
	assert.Equal(t, codes.Error, exp.GetSpans()[0].Status.Code)
```

yurishkuro (approved):

PR #4596, opened 2023-07-23
Replace es-spanstore tracing instrumentation with OpenTelemetry

afzal442:
* Part of #3381 
* This PR adds OTEL Provider to es-spanstore component

- Replaces OT tracer with OTEL tracer to support jtracer pkg

### DIFF
--- a/plugin/storage/es/factory.go
+++ b/plugin/storage/es/factory.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/spf13/viper"
 	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/trace"
 	"go.uber.org/zap"
 
 	"github.com/jaegertracing/jaeger/pkg/es"
@@ -51,6 +52,7 @@ type Factory struct {
 
 	metricsFactory metrics.Factory
 	logger         *zap.Logger
+	tracer         trace.TracerProvider
 
 	newClientFn func(c *config.Configuration, logger *zap.Logger, metricsFactory metrics.Factory) (es.Client, error)
 
@@ -65,6 +67,7 @@ func NewFactory() *Factory {
 	return &Factory{
 		Options:     NewOptions(primaryNamespace, archiveNamespace),
 		newClientFn: config.NewClient,
+		tracer:      otel.GetTracerProvider(),
 	}
 }
 
@@ -109,17 +112,17 @@ func (f *Factory) Initialize(metricsFactory metrics.Factory, logger *zap.Logger)
 
 // CreateSpanReader implements storage.Factory
 func (f *Factory) CreateSpanReader() (spanstore.Reader, error) {
-	return createSpanReader(f.metricsFactory, f.logger, f.primaryClient, f.primaryConfig, false)
+	return createSpanReader(f.primaryClient, f.primaryConfig, false, f.metricsFactory, f.logger, f.tracer)
 }
 
 // CreateSpanWriter implements storage.Factory
 func (f *Factory) CreateSpanWriter() (spanstore.Writer, error) {
-	return createSpanWriter(f.metricsFactory, f.logger, f.primaryClient, f.primaryConfig, false)
+	return createSpanWriter(f.primaryClient, f.primaryConfig, false, f.metricsFactory, f.logger)
 }
 
 // CreateDependencyReader implements storage.Factory
 func (f *Factory) CreateDependencyReader() (dependencystore.Reader, error) {
-	return createDependencyReader(f.logger, f.primaryClient, f.primaryConfig)
+	return createDependencyReader(f.primaryClient, f.primaryConfig, f.logger)
 }
 
 // CreateArchiveSpanReader implements storage.ArchiveFactory
@@ -127,7 +130,7 @@ func (f *Factory) CreateArchiveSpanReader() (spanstore.Reader, error) {
 	if !f.archiveConfig.Enabled {
 		return nil, nil
 	}
-	return createSpanReader(f.metricsFactory, f.logger, f.archiveClient, f.archiveConfig, true)
+	return createSpanReader(f.archiveClient, f.archiveConfig, true, f.metricsFactory, f.logger, f.tracer)
 }
 
 // CreateArchiveSpanWriter implements storage.ArchiveFactory
@@ -135,23 +138,22 @@ func (f *Factory) CreateArchiveSpanWriter() (spanstore.Writer, error) {
 	if !f.archiveConfig.Enabled {
 		return nil, nil
 	}
-	return createSpanWriter(f.metricsFactory, f.logger, f.archiveClient, f.archiveConfig, true)
+	return createSpanWriter(f.archiveClient, f.archiveConfig, true, f.metricsFactory, f.logger)
 }
 
 func createSpanReader(
-	mFactory metrics.Factory,
-	logger *zap.Logger,
 	client es.Client,
 	cfg *config.Configuration,
 	archive bool,
+	mFactory metrics.Factory,
+	logger *zap.Logger,
+	tp trace.TracerProvider,
 ) (spanstore.Reader, error) {
 	if cfg.UseILM && !cfg.UseReadWriteAliases {
 		return nil, fmt.Errorf("--es.use-ilm must always be used in conjunction with --es.use-aliases to ensure ES writers and readers refer to the single index mapping")
 	}
 	return esSpanStore.NewSpanReader(esSpanStore.SpanReaderParams{
 		Client:                        client,
-		Logger:                        logger,
-		MetricsFactory:                mFactory,
 		MaxDocCount:                   cfg.MaxDocCount,
 		MaxSpanAge:                    cfg.MaxSpanAge,
 		IndexPrefix:                   cfg.IndexPrefix,
@@ -163,16 +165,18 @@ func createSpanReader(
 		UseReadWriteAliases:           cfg.UseReadWriteAliases,
 		Archive:                       archive,
 		RemoteReadClusters:            cfg.RemoteReadClusters,
-		Tracer:                        otel.GetTracerProvider().Tracer("span-reader"),
+		Logger:                        logger,
+		MetricsFactory:                mFactory,
+		Tracer:                        tp.Tracer("esSpanStore.SpanReader"),
 	}), nil
 }
 
 func createSpanWriter(
-	mFactory metrics.Factory,
-	logger *zap.Logger,
 	client es.Client,
 	cfg *config.Configuration,
 	archive bool,
+	mFactory metrics.Factory,
+	logger *zap.Logger,
 ) (spanstore.Writer, error) {
 	var tags []string
 	var err error
@@ -199,8 +203,6 @@ func createSpanWriter(
 	}
 	writer := esSpanStore.NewSpanWriter(esSpanStore.SpanWriterParams{
 		Client:                 client,
-		Logger:                 logger,
-		MetricsFactory:         mFactory,
 		IndexPrefix:            cfg.IndexPrefix,
 		SpanIndexDateLayout:    cfg.IndexDateLayoutSpans,
 		ServiceIndexDateLayout: cfg.IndexDateLayoutServices,
@@ -209,6 +211,8 @@ func createSpanWriter(
 		TagDotReplacement:      cfg.Tags.DotReplacement,
 		Archive:                archive,
 		UseReadWriteAliases:    cfg.UseReadWriteAliases,
+		Logger:                 logger,
+		MetricsFactory:         mFactory,
 	})
 
 	// Creating a template here would conflict with the one created for ILM resulting to no index rollover
@@ -222,9 +226,9 @@ func createSpanWriter(
 }
 
 func createDependencyReader(
-	logger *zap.Logger,
 	client es.Client,
 	cfg *config.Configuration,
+	logger *zap.Logger,
 ) (dependencystore.Reader, error) {
 	reader := esDepStore.NewDependencyStore(esDepStore.DependencyStoreParams{
 		Client:              client,
--- a/plugin/storage/es/factory.go
+++ b/plugin/storage/es/factory.go
@@ -21,6 +21,7 @@ import (
 	"io"
 
 	"github.com/spf13/viper"
+	"go.opentelemetry.io/otel"
 	"go.uber.org/zap"
 
 	"github.com/jaegertracing/jaeger/pkg/es"
@@ -162,6 +163,7 @@ func createSpanReader(
 		UseReadWriteAliases:           cfg.UseReadWriteAliases,
 		Archive:                       archive,
 		RemoteReadClusters:            cfg.RemoteReadClusters,
+		Tracer:                        otel.GetTracerProvider().Tracer("span-reader"),
 	}), nil
 }
 
--- a/plugin/storage/es/spanstore/reader.go
+++ b/plugin/storage/es/spanstore/reader.go
@@ -144,7 +144,9 @@ func NewSpanReader(p SpanReaderParams) *SpanReader {
 	if err != nil {
 		log.Fatal("Failed to initialise tracer", zap.Error(err))
 	}
-	defer jt.Close(context.Background())
+	if err := jt.Close(context.Background()); err != nil {
+		log.Fatal("Error shutting down tracer provider", zap.Error(err))
+	}
 	return &SpanReader{
 		client:                        p.Client,
 		logger:                        p.Logger,
--- a/plugin/storage/es/spanstore/reader.go
+++ b/plugin/storage/es/spanstore/reader.go
@@ -21,7 +21,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"log"
 	"time"
 
 	"github.com/olivere/elastic"
@@ -31,7 +30,6 @@ import (
 
 	"github.com/jaegertracing/jaeger/model"
 	"github.com/jaegertracing/jaeger/pkg/es"
-	"github.com/jaegertracing/jaeger/pkg/jtracer"
 	"github.com/jaegertracing/jaeger/pkg/metrics"
 	"github.com/jaegertracing/jaeger/plugin/storage/es/spanstore/dbmodel"
 	"github.com/jaegertracing/jaeger/storage/spanstore"
@@ -140,13 +138,6 @@ func NewSpanReader(p SpanReaderParams) *SpanReader {
 	if p.UseReadWriteAliases {
 		maxSpanAge = rolloverMaxSpanAge
 	}
-	jt, err := jtracer.New("spanstore")
-	if err != nil {
-		log.Fatal("Failed to initialise tracer", zap.Error(err))
-	}
-	if err := jt.Close(context.Background()); err != nil {
-		log.Fatal("Error shutting down tracer provider", zap.Error(err))
-	}
 	return &SpanReader{
 		client:                        p.Client,
 		logger:                        p.Logger,
@@ -163,7 +154,7 @@ func NewSpanReader(p SpanReaderParams) *SpanReader {
 		sourceFn:                      getSourceFn(p.Archive, p.MaxDocCount),
 		maxDocCount:                   p.MaxDocCount,
 		useReadWriteAliases:           p.UseReadWriteAliases,
-		tracer:                        jt.OTEL.Tracer("spanreader"),
+		tracer:                        p.Tracer,
 	}
 }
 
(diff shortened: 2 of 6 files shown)


plugin/storage/es/spanstore/reader.go
```diff
@@ -360,9 +371,9 @@ func (s *SpanReader) FindTraceIDs(ctx context.Context, traceQuery *spanstore.Tra
 }
 
 func (s *SpanReader) multiRead(ctx context.Context, traceIDs []model.TraceID, startTime, endTime time.Time) ([]*model.Trace, error) {
-	childSpan, _ := opentracing.StartSpanFromContext(ctx, "multiRead")
-	childSpan.LogFields(otlog.Object("trace_ids", traceIDs))
-	defer childSpan.Finish()
+	ctx, childSpan := s.tracer.Start(ctx, "multiRead")
+	childSpan.SetAttributes(attribute.Key("trace_ids").String(""))
```
afzal442, line 375: Need some suggestion on how to adjust []model type here.

plugin/storage/es/spanstore/reader.go
```diff
@@ -137,6 +140,13 @@ func NewSpanReader(p SpanReaderParams) *SpanReader {
 	if p.UseReadWriteAliases {
 		maxSpanAge = rolloverMaxSpanAge
 	}
+	jt, err := jtracer.New("spanstore")
+	if err != nil {
+		log.Fatal("Failed to initialise tracer", zap.Error(err))
+	}
+	if err := jt.Close(context.Background()); err != nil {
```
afzal442, line 147: Not able to figure out where to close.

plugin/storage/es/spanstore/reader.go
```diff
@@ -137,6 +140,13 @@ func NewSpanReader(p SpanReaderParams) *SpanReader {
 	if p.UseReadWriteAliases {
 		maxSpanAge = rolloverMaxSpanAge
 	}
+	jt, err := jtracer.New("spanstore")
```
yurishkuro, line 143: Once again, we do not create tracers in random places, only in mains. You already used jtracer.New twice, in two mains - that's it, you cannot call it again. You have to pass the tracers you creates in mains down to these components.

plugin/storage/es/spanstore/reader_test.go
```diff
@@ -170,63 +171,64 @@ func TestSpanReaderIndices(t *testing.T) {
...
+	tracer := jtracer.NoOp().OTEL.Tracer("")
 
 	testCases := []struct {
 		indices []string
 		params  SpanReaderParams
 	}{
 		{
 			params: SpanReaderParams{
-				Client: client, Logger: logger, MetricsFactory: metricsFactory,
+				Client: client, Logger: logger, MetricsFactory: metricsFactory, Tracer: tracer,
```
yurishkuro, line 182: These tests are using a poor pattern, let's fix it. When parameterizing tests, params should only contain things that are different between test cases. These three values are always the same, they can be assigned inside the loop below.

plugin/storage/es/spanstore/reader.go
yurishkuro: Loop through trace IDs and call String() on each, and pass in a slice of strings. And before doing that you may want to check if the span is recording, since it's expensive operation.

plugin/storage/es/spanstore/reader.go
```diff
@@ -360,9 +362,16 @@ func (s *SpanReader) FindTraceIDs(ctx context.Context, traceQuery *spanstore.Tra
...
+	ctx, childSpan := s.tracer.Start(ctx, "multiRead")
+	defer childSpan.End()
+
+	tracesIDs := make([]string, len(traceIDs))
+	if childSpan.IsRecording() {
+		for i, traceID := range traceIDs {
+			tracesIDs[i] = traceID.String()
+		}
+	}
+	childSpan.SetAttributes(attribute.Key("trace_ids").StringSlice(tracesIDs))
```
yurishkuro, line 374:
```suggestion
	if childSpan.IsRecording() {
		tracesIDs := make([]string, len(traceIDs))
		for i, traceID := range traceIDs {
			tracesIDs[i] = traceID.String()
		}
		childSpan.SetAttributes(attribute.Key("trace_ids").StringSlice(tracesIDs))
	}
```

plugin/storage/es/factory.go
```diff
@@ -162,6 +163,7 @@ func createSpanReader(
 		UseReadWriteAliases:           cfg.UseReadWriteAliases,
 		Archive:                       archive,
 		RemoteReadClusters:            cfg.RemoteReadClusters,
+		Tracer:                        otel.GetTracerProvider().Tracer("span-reader"),
```
yurishkuro, line 166:
```suggestion
		Tracer:                        otel.GetTracerProvider().Tracer("esSpanStore.SpanReader"),
```

plugin/storage/es/spanstore/reader_test.go
```diff
@@ -282,6 +273,13 @@ func TestSpanReaderIndices(t *testing.T) {
 		},
 	}
 	for _, testCase := range testCases {
+		testCase.params = SpanReaderParams{
```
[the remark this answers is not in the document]
yurishkuro: yes - don't override the whole Params struct, just few fields

plugin/storage/integration/elasticsearch_test.go
```diff
@@ -151,6 +152,7 @@ func (s *ESStorageIntegration) initSpanstore(allTagsAsFields, archive bool) erro
 		TagDotReplacement: tagKeyDeDotChar,
 		Archive:           archive,
 		MaxDocCount:       defaultMaxDocCount,
+		Tracer:            jtracer.NoOp().OTEL.Tracer(""),
```
yurishkuro, line 155: keep observability objects together (logger, metrics, traces)

plugin/storage/es/factory.go
yurishkuro: also, this should be split: here you only pass `Tracer:  otel.GetTracerProvider()`, and `esSpanStore.NewSpanReader()` calls `.Tracer()` on it. Because in the future it may need Provider for other purposes, you don't know. Provider is the shared object, Tracers are created by individual instrumentations as needed.

plugin/storage/es/spanstore/reader.go
```diff
@@ -329,8 +331,8 @@ func bucketToStringArray(buckets []*elastic.AggregationBucketKeyItem) ([]string,
 
 // FindTraces retrieves traces that match the traceQuery
 func (s *SpanReader) FindTraces(ctx context.Context, traceQuery *spanstore.TraceQueryParameters) ([]*model.Trace, error) {
-	span, ctx := opentracing.StartSpanFromContext(ctx, "FindTraces")
-	defer span.Finish()
+	ctx, span := s.tracer.Start(ctx, "GetOperations")
```
yurishkuro, line 334:
```suggestion
	ctx, span := s.tracer.Start(ctx, "FindTraces")
```

plugin/storage/es/factory.go
```diff
@@ -50,6 +52,7 @@ type Factory struct {
 
 	metricsFactory metrics.Factory
 	logger         *zap.Logger
+	tracer         trace.TracerProvider
```
afzal442, line 55: @yurishkuro I am still facing an error here quite scary. Maybe I am missed out here.
```
```

yurishkuro: look at full stack trace in the panic log and find where nil is being passed.

afzal442: I drilled down to this `trace, err := r.reader.GetTrace(context.Background(), model.NewTraceID(0, 1))`

afzal442: sorry you mean e.g `jaeger-all-in-one` main? Then yes.

yurishkuro: I don't see it in the code

afzal442: I tried this out in my local test. I still failed to get the traces. I don't understand this `github.com/jaegertracing/jaeger/plugin/storage/es/spanstore.(*SpanReader).GetTrace(0xc000000cc0, {0xd64a90?, 0xc0000440d8?}, {0xc00007f480?, 0x20?})` I guess this has pointer to address which is not nil.

yurishkuro: Please push changes to pr so that I can reproduce. Is this when running tests or main binaries? (Also, you can avoid this back and forth by posting full stack trace)

yurishkuro: in the spanreader_test, I see many tests are using `withSpanReader()`, which creates a reader like this:
```
		reader: NewSpanReader(SpanReaderParams{
			Client:            client,
			Logger:            zap.NewNop(),
			MaxSpanAge:        0,
			IndexPrefix:       "",
			TagDotReplacement: "@",
			MaxDocCount:       defaultMaxDocCount,
		}),
```
The tracer is not initialized there at all, so not surprising you get a null pointer panic

plugin/storage/es/spanstore/reader_test.go
```diff
@@ -118,6 +120,7 @@ func withArchiveSpanReader(readAlias bool, fn func(r *spanReaderTest)) {
 		reader: NewSpanReader(SpanReaderParams{
 			Client:              client,
 			Logger:              zap.NewNop(),
+			Tracer:              jtracer.NoOp().OTEL.Tracer("test"),
```
yurishkuro, line 123: how will we know if the traces are created if it's noop? You already went through this with metricsreader PR

plugin/storage/es/spanstore/reader.go
```diff
@@ -238,8 +240,8 @@ func indexNames(prefix, index string) string {
 
 // GetTrace takes a traceID and returns a Trace associated with that traceID
 func (s *SpanReader) GetTrace(ctx context.Context, traceID model.TraceID) (*model.Trace, error) {
-	span, ctx := opentracing.StartSpanFromContext(ctx, "GetTrace")
-	defer span.Finish()
+	ctx, span := s.tracer.Start(ctx, "GetTrace")
```
afzal442, line 243: I am getting nooptracer here.

plugin/storage/es/spanstore/reader_test.go
```diff
@@ -86,19 +89,37 @@ type spanReaderTest struct {
...
 func withSpanReader(fn func(r *spanReaderTest)) {
 	client := &mocks.Client{}
+	tracer, exp, closer := tracerProvider()
 	logger, logBuffer := testutils.NewLogger()
 	r := &spanReaderTest{
 		client:    client,
 		logger:    logger,
 		logBuffer: logBuffer,
+		exporter:  exp,
+		closer:    closer(),
```
yurishkuro, line 118: Closer is too generic name in the context of this struct

plugin/storage/es/spanstore/reader_test.go
```diff
@@ -282,7 +297,13 @@ func TestSpanReaderIndices(t *testing.T) {
 		},
 	}
 	for _, testCase := range testCases {
+		testCase.params.Client = client
+		testCase.params.Logger = logger
+		testCase.params.MetricsFactory = metricsFactory
+		testCase.params.Tracer = tracer.Tracer("test")
 		r := NewSpanReader(testCase.params)
+		assert.NotEmpty(t, exp.GetSpans(), "Spans recorded")
+		assert.NoError(t, closer())
```
yurishkuro, line 306: How is the next test in the loop going to work if you close the tracer here?

plugin/storage/es/spanstore/reader_test.go
```diff
@@ -438,6 +468,8 @@ func TestSpanReader_GetTraceQueryError(t *testing.T) {
 			Return(&elastic.MultiSearchResult{
 				Responses: []*elastic.SearchResult{},
 			}, nil)
+		assert.NotEmpty(t, r.exporter.GetSpans(), "Spans recorded")
+		assert.NoError(t, r.closer)
```
yurishkuro, line 472: NoError expects error type, closer is a function. How does this even compile?

plugin/storage/es/spanstore/reader.go
```diff
@@ -153,6 +153,8 @@ func NewSpanReader(p SpanReaderParams) *SpanReader {
 		sourceFn:                      getSourceFn(p.Archive, p.MaxDocCount),
 		maxDocCount:                   p.MaxDocCount,
 		useReadWriteAliases:           p.UseReadWriteAliases,
+		logger:                        p.Logger,
+		tracer:                        p.Tracer,
```
afzal442, line 157: tracer is not getting initialised, I think. 😕

yurishkuro: easy debugging trick:
```
if p.Tracer == nil {
     panic("nil tracer")
}
```
This gives you a stack trace showing the order of execution. If that does not explain where nil is coming from, then bubble this code up the stack.

plugin/storage/es/spanstore/reader_test.go
```diff
@@ -307,6 +328,8 @@ func TestSpanReader_GetTrace(t *testing.T) {
 			}, nil)
 
 		trace, err := r.reader.GetTrace(context.Background(), model.NewTraceID(0, 1))
+		require.NotEmpty(t, r.exporter.GetSpans(), "Spans recorded")
```
afzal442, line 331: @yurishkuro after workaround, I am still getting Empty spans. Any help?

yurishkuro: L118:
```
	tracer, exp, closer := tracerProvider()
	logger, logBuffer := testutils.NewLogger()
	r := &spanReaderTest{
		client:    client,
		logger:    logger,
		logBuffer: logBuffer,
		exporter:  exp,
		err:       closer(), // what do you suppose happens here?
```

afzal442: thanks much. that got called and shutdown the tracer.

plugin/storage/es/spanstore/reader_test.go
```diff
@@ -115,7 +115,7 @@ func withSpanReader(fn func(r *spanReaderTest)) {
 		logger:    logger,
 		logBuffer: logBuffer,
 		exporter:  exp,
-		err:       closer(),
+		err:       closer,
```
yurishkuro, line 118: why do you think `err` is an appropriate name?

afzal442: Would it be a `closer` obj?

yurishkuro: Here's how your struct would look:
```
type spanReaderTest struct {
	client    *mocks.Client
	logger    *zap.Logger
	logBuffer *testutils.Buffer
	exporter  *tracetest.InMemoryExporter
	closer    func() error
	reader    *SpanReader
}
```
From this definition alone, can you explain what `closer` does? Don't say "it shuts down the tracer", because that information is not in this struct definition (whereas all other fields are self-explanatory).

afzal442: That's true. how I can use my golang dumb skill. thanks for your advice as always. it may feel you annoying, I guess.

plugin/storage/es/spanstore/reader_test.go
```diff
@@ -82,23 +85,43 @@ var exampleESSpan = []byte(
...
 	}`)
 
+type closed func() error
+
 type spanReaderTest struct {
 	client    *mocks.Client
 	logger    *zap.Logger
 	logBuffer *testutils.Buffer
+	exporter  *tracetest.InMemoryExporter
+	closer    closed
```
yurishkuro, line 95: this isn't any better, I can already tell from the field name that it "closes something", just not what it closes. Adding a type like that does not help. But naming the field `tracerCloser` make is obvious, without any additional type.

plugin/storage/es/spanstore/reader.go
```diff
@@ -137,9 +139,11 @@ func NewSpanReader(p SpanReaderParams) *SpanReader {
 	if p.UseReadWriteAliases {
 		maxSpanAge = rolloverMaxSpanAge
 	}
+	if p.Tracer == nil {
+		p.Tracer = otel.Tracer("eSpanstore.SpanReader")
+	}
```
yurishkuro, line 144:
```suggestion
```

plugin/storage/es/spanstore/reader.go
```diff
@@ -686,7 +699,6 @@ func (s *SpanReader) buildObjectQuery(field string, k string, v string) elastic.
 	return elastic.NewBoolQuery().Must(keyQuery)
 }
 
-func logErrorToSpan(span opentracing.Span, err error) {
-	ottag.Error.Set(span, true)
-	span.LogFields(otlog.Error(err))
+func logErrorToSpan(span trace.Span, err error) {
+	span.RecordError(err)
```
yurishkuro, line 703: missing status again?

plugin/storage/es/spanstore/reader_test.go
```diff
@@ -83,22 +86,40 @@ var exampleESSpan = []byte(
...
+	tracer, exp, closer := tracerProvider()
 	logger, logBuffer := testutils.NewLogger()
 	r := &spanReaderTest{
-		client:    client,
-		logger:    logger,
-		logBuffer: logBuffer,
+		client:       client,
+		logger:       logger,
+		logBuffer:    logBuffer,
+		exporter:     exp,
```
yurishkuro, line 117:
```suggestion
		traceBuffer:     exp,
```

plugin/storage/es/spanstore/reader_test.go
```diff
@@ -307,6 +327,8 @@ func TestSpanReader_GetTrace(t *testing.T) {
 			}, nil)
 
 		trace, err := r.reader.GetTrace(context.Background(), model.NewTraceID(0, 1))
+		require.NotEmpty(t, r.exporter.GetSpans(), "Spans recorded")
+		require.NoError(t, r.tracerCloser())
```
yurishkuro, line 331: one of the nice thing about doing tests in a functional style via `withSpanReader` is that `withSpanReader` can do both initialization and tear-down, since it's frame is lower in the stack trace than the actual test function and `defer`s inside `withSpanReader` always get executed even if the test function panics. 
```
   TestSomething()
       withSpanReader(testFn)
            d = initialize()
            defer d.teardown()
            testFn(d)
```

So I don't think we even need to expose traceCloser in the struct, we can just defer-call it inside `withSpanReader`.

This is very different from helper functions that simply _return_ a tracer. They have no room to defer-schedule shutdown, and you have to do it manually in the caller.

afzalbin64: That's nit. Cherry picked. 😄

plugin/storage/es/spanstore/reader_test.go
```diff
@@ -126,6 +124,7 @@ func withSpanReader(fn func(r *spanReaderTest)) {
 			MaxDocCount:       defaultMaxDocCount,
 		}),
 	}
+	defer closer()
```
yurishkuro, line 127: we talked about where defers should be placed

plugin/storage/es/spanstore/reader_test.go
```diff
@@ -282,12 +296,17 @@ func TestSpanReaderIndices(t *testing.T) {
...
+		testCase.params.Logger = logger
+		testCase.params.MetricsFactory = metricsFactory
+		testCase.params.Tracer = tracer.Tracer("test")
 		r := NewSpanReader(testCase.params)
 
 		actualSpan := r.timeRangeIndices(r.spanIndexPrefix, r.spanIndexDateLayout, date, date, -1*time.Hour)
 		actualService := r.timeRangeIndices(r.serviceIndexPrefix, r.serviceIndexDateLayout, date, date, -24*time.Hour)
 		assert.Equal(t, testCase.indices, append(actualSpan, actualService...))
 	}
+	require.NoError(t, closer())
```
yurishkuro, line 309: defer after creation

plugin/storage/integration/elasticsearch_test.go
```diff
@@ -160,6 +177,10 @@ func (s *ESStorageIntegration) initSpanstore(allTagsAsFields, archive bool) erro
 		MaxDocCount:     defaultMaxDocCount,
 	})
 
+	if closer != nil {
+		return err
```
yurishkuro, line 181: ?

plugin/storage/es/spanstore/reader_test.go
afzalbin64: sorry Yuri! I did it inside withSpanReader function call. but How can I defer it? I think I have to modify the closer to use `func()` and `fmt.Errorf`.

yurishkuro: you create the object in L110, so `defer closer()` should be L111. Always keep defer next to creation.

plugin/storage/es/spanstore/reader_test.go
```diff
@@ -83,48 +86,69 @@ var exampleESSpan = []byte(
...
 			Logger:              zap.NewNop(),
+			Tracer:              tracer.Tracer("test"),
 			MaxSpanAge:          0,
 			IndexPrefix:         "",
 			TagDotReplacement:   "@",
 			Archive:             true,
 			UseReadWriteAliases: readAlias,
 		}),
 	}
+	defer closer()
```
yurishkuro, line 151: to L134

plugin/storage/es/spanstore/reader_test.go
yurishkuro: L198

plugin/storage/es/spanstore/reader_test.go
afzalbin64: I did that. Just to re-commit.

yurishkuro (approved):
lgtm

yurishkuro:
minor clean-up #4605

PR #4614, opened 2023-08-01
Replace usage of OpenTracing constants with OpenTelemetry in the domain model

afzal442:
* Part of #3381 
* This PR adds `otel trace` spans type to span model

- Replace OT trace with `otel trace` spans type i.e. `ext.SpanKind..` with `trace.SpanKind..`

### DIFF
--- a/cmd/query/app/apiv3/otlp_translator.go
+++ b/cmd/query/app/apiv3/otlp_translator.go
@@ -25,6 +25,7 @@ import (
 
 	"go.opentelemetry.io/collector/pdata/ptrace"
 	semconv "go.opentelemetry.io/collector/semconv/v1.5.0"
+	"go.opentelemetry.io/otel/trace"
 
 	"github.com/jaegertracing/jaeger/model"
 	commonv1 "github.com/jaegertracing/jaeger/proto-gen/otel/common/v1"
@@ -112,7 +113,7 @@ func jSpanToOTLP(jSpan *model.Span) (*tracev1.Span, resource, instrumentationLib
 		Kind:              tracev1.Span_SPAN_KIND_UNSPECIFIED,
 	}
 	if kind, found := jSpan.GetSpanKind(); found {
-		s.Kind = jSpanKindToInternal(kind.String())
+		s.Kind = jSpanKindToInternal(kind)
 		ignoreKeys[tagSpanKind] = true
 	}
 
@@ -333,17 +334,17 @@ func getStatusCodeFromHTTPStatusTag(tag model.KeyValue) (int, error) {
 	return int(statusCodeFromHTTP(statusCode)), nil
 }
 
-func jSpanKindToInternal(spanKind string) tracev1.Span_SpanKind {
+func jSpanKindToInternal(spanKind trace.SpanKind) tracev1.Span_SpanKind {
 	switch spanKind {
-	case "client":
+	case trace.SpanKindClient:
 		return tracev1.Span_SPAN_KIND_CLIENT
-	case "server":
+	case trace.SpanKindServer:
 		return tracev1.Span_SPAN_KIND_SERVER
-	case "producer":
+	case trace.SpanKindProducer:
 		return tracev1.Span_SPAN_KIND_PRODUCER
-	case "consumer":
+	case trace.SpanKindConsumer:
 		return tracev1.Span_SPAN_KIND_CONSUMER
-	case "internal":
+	case trace.SpanKindInternal:
 		return tracev1.Span_SPAN_KIND_INTERNAL
 	}
 	return tracev1.Span_SPAN_KIND_UNSPECIFIED
--- a/cmd/query/app/apiv3/otlp_translator_test.go
+++ b/cmd/query/app/apiv3/otlp_translator_test.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	semconv "go.opentelemetry.io/collector/semconv/v1.5.0"
+	"go.opentelemetry.io/otel/trace"
 
 	"github.com/jaegertracing/jaeger/model"
 	commonv1 "github.com/jaegertracing/jaeger/proto-gen/otel/common/v1"
@@ -151,27 +152,27 @@ func TestTranslateSpan(t *testing.T) {
 
 func TestTranslateSpanKind(t *testing.T) {
 	tests := []struct {
-		kind         string
+		kind         trace.SpanKind
 		otelSpanKind v1.Span_SpanKind
 	}{
 		{
-			kind:         "client",
+			kind:         trace.SpanKindClient,
 			otelSpanKind: v1.Span_SPAN_KIND_CLIENT,
 		},
 		{
-			kind:         "server",
+			kind:         trace.SpanKindServer,
 			otelSpanKind: v1.Span_SPAN_KIND_SERVER,
 		},
 		{
-			kind:         "producer",
+			kind:         trace.SpanKindProducer,
 			otelSpanKind: v1.Span_SPAN_KIND_PRODUCER,
 		},
 		{
-			kind:         "consumer",
+			kind:         trace.SpanKindConsumer,
 			otelSpanKind: v1.Span_SPAN_KIND_CONSUMER,
 		},
 		{
-			kind:         "internal",
+			kind:         trace.SpanKindInternal,
 			otelSpanKind: v1.Span_SPAN_KIND_INTERNAL,
 		},
 		{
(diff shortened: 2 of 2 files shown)


model/span.go
```diff
@@ -49,19 +49,19 @@ func (s *Span) Hash(w io.Writer) (err error) {
...
 }
 
 // GetSpanKind returns value of `span.kind` tag and whether the tag can be found
 func (s *Span) GetSpanKind() (spanKind string, found bool) {
-	if tag, ok := KeyValues(s.Tags).FindByKey(string(ext.SpanKind)); ok {
+	if tag, ok := KeyValues(s.Tags).FindByKey("span.kind"); ok {
 		return tag.AsString(), true
 	}
-	return "", false
+	return trace.SpanKindUnspecified.String(), false
```
yurishkuro, line 64: could we return the enum `trace.SpanKind` here instead of string? I don't suppose too many places in the code check for this.

model/span.go
```diff
@@ -49,19 +49,19 @@ func (s *Span) Hash(w io.Writer) (err error) {
 }
 
 // HasSpanKind returns true if the span has a `span.kind` tag set to `kind`.
-func (s *Span) HasSpanKind(kind ext.SpanKindEnum) bool {
-	if tag, ok := KeyValues(s.Tags).FindByKey(string(ext.SpanKind)); ok {
-		return tag.AsString() == string(kind)
+func (s *Span) HasSpanKind(kind trace.SpanKind) bool {
+	if tag, ok := KeyValues(s.Tags).FindByKey("span.kind"); ok {
```
yurishkuro, line 53: > "span.kind"

instead of using a string, it's better to define a (private) constant, e.g. `const keySpanKind = "span.kind"`

model/span.go
afzalbin64: I didn't find that. I think enum is not supported for otel tracer span.

albertteoh: @afzal442 [trace.SpanKind is an enum](https://github.com/open-telemetry/opentelemetry-go/blob/trace/v1.16.0/trace/trace.go#L428).

I like the idea of returning `trace.SpanKind` so it's more strongly and explicitly typed, but there are two drawbacks I see:
1. We'd need to map from `string` to `trace.SpanKind`. This is trivial.
2. If we want to use this type deep in the call stack, the changes may not be trivial and would need to be well tested, but doable as a single separate PR.

We could go half-way and return `trace.SpanKind` then call `spanKind.String()` wherever it's used but I don't think there's much value there.

Overall, my vote is for 1. as it's the most pragmatic.

model/span_test.go
```diff
@@ -227,12 +227,12 @@ func TestIsFirehoseEnabled(t *testing.T) {
 func TestGetSpanKind(t *testing.T) {
 	span := makeSpan(model.String("sampler.type", "lowerbound"))
 	spanKind, found := span.GetSpanKind()
-	assert.Equal(t, "", spanKind)
+	assert.Equal(t, trace.SpanKind.String(0), spanKind)
```
albertteoh, line 230: Using the predefined enum consts would be more readable:
- `trace.SpanKindUnspecified.String()`
- `trace.SpanKindClient.String()`

I think it's also okay to just leave it as `""` and `"client"` since it's better to decouple expected values from the thing that's being tested, and is the final displayed value in the span tag anyway.

albertteoh (commented):
Looks pretty good to me, just a couple of nits.

model/span_test.go
afzalbin64: Thanks for this direction, I will revert it as it was.

albertteoh:
> but the concern is why the unit test is still failing. sorry!

Ah, based on what I see [here](https://github.com/jaegertracing/jaeger/actions/runs/5735977952/job/15544657259?pr=4614#step:7:5219):

```
integration.go:184: 
        	Error Trace:	/home/runner/work/jaeger/jaeger/plugin/storage/integration/integration.go:184
        	Error:      	Should be true
        	Test:       	TestMemoryStorage/GetOperations
    integration.go:185: 	 Expected: [{example-operation-1 } {example-operation-3 server} {example-operation-4 client}]
    integration.go:186: 	 Actual  : [{example-operation-1 unspecified} {example-operation-3 server} {example-operation-4 client}]
```

So perhaps, the expectation should be?

```go
assert.Equal(t, "unspecified", spanKind)
```

model/span.go
```diff
@@ -34,12 +34,22 @@ const (
...
 	samplerType        = "sampler.type"
+	keySpanKind        = "span.kind"
 	samplerTypeUnknown = "unknown"
 )
 
 // Flags is a bit map of flags for a span
 type Flags uint32
 
+// Map from string to trace.SpanKind
+var spanKindMapping = map[string]trace.SpanKind{
```
albertteoh, line 45: nit: I tend to prefer avoid package-level vars where practical, especially in this case where the lookup should never change, because as a package-level var, it can be mutated throughout the lifetime of the program.

What do you think about making this readonly?

```go
// Map from string to trace.SpanKind.
func toSpanKind(s string) (trace.SpanKind, bool) {
	sk, found := map[string]trace.SpanKind{
		"client":   trace.SpanKindClient,
		"server":   trace.SpanKindServer,
		"producer": trace.SpanKindProducer,
		"consumer": trace.SpanKindConsumer,
		"internal": trace.SpanKindInternal,
	}[s]
	return sk, found
}
```

albertteoh (dismissed):
LGTM with a tiny nit.

model/span_test.go
```diff
@@ -188,7 +188,7 @@ func TestSpanIDUnmarshalJSONErrors(t *testing.T) {
 func TestIsRPCClientServer(t *testing.T) {
 	span1 := &model.Span{
 		Tags: model.KeyValues{
-			model.String(string(ext.SpanKind), string(ext.SpanKindRPCClientEnum)),
+			model.String(string("span.kind"), trace.SpanKindClient.String()),
```
yurishkuro, line 191:
```suggestion
			model.String(keySpanKind, trace.SpanKindClient.String()),
```

cmd/query/app/apiv3/otlp_translator.go
```diff
@@ -112,7 +112,7 @@ func jSpanToOTLP(jSpan *model.Span) (*tracev1.Span, resource, instrumentationLib
 		Kind:              tracev1.Span_SPAN_KIND_INTERNAL,
 	}
 	if kind, found := jSpan.GetSpanKind(); found {
-		s.Kind = jSpanKindToInternal(kind)
+		s.Kind = jSpanKindToInternal(kind.String())
```
yurishkuro, line 115: isn't this now simply `s.Kind = kind`? And `jSpanKindToInternal()` not neeed?

cmd/query/app/apiv3/otlp_translator.go
```diff
@@ -112,7 +112,7 @@ func jSpanToOTLP(jSpan *model.Span) (*tracev1.Span, resource, instrumentationLib
 		Kind:              tracev1.Span_SPAN_KIND_INTERNAL,
```
yurishkuro, line 112: should this be UNSPECIFIED?

model/span.go
yurishkuro: would this allocate a map on every invocation?

albertteoh: Potentially, but worst case, I expect it to be very cheap compared to I/O, for instance; but hard to know the actual impact without profiling. If we think this will be on a hot path and materially impact performance, I'm happy to leave it as is.

cmd/query/app/apiv3/otlp_translator.go
afzalbin64: Yes! that should be initialised to "0".

cmd/query/app/apiv3/otlp_translator.go
afzalbin64: I think it's needed. s.Kind is defined as `tracev1.Span_SpanKind` whereas kind is `trace.SpanKind`. both are different.

model/span.go
albertteoh: By the way, since this is a variable declared inside the stack frame of this function call, I expect it to be allocated on the stack, which should be substantially more efficient than heap allocation.

afzalbin64: thanks for that pro tip. I discovered new way of doing things sorted.

yurishkuro: it's not just the matter of allocating, actually initializing a map and laying out its structure is unnecessary work. It's normal to keep private maps as package var.

cmd/query/app/apiv3/otlp_translator.go
yurishkuro: then at least let's not do it based on the string.

yurishkuro (approved):

PR #4624, opened 2023-08-04
[test] Replace OT const with OTEL

afzal442:
* Part of #3381 
* This PR adds `otel trace` spans type to span model

- Replace OT const with `otel trace` spans const i.e. `ext.SpanKind..` with `trace.SpanKind..`

### DIFF
--- a/cmd/query/app/handler_options.go
+++ b/cmd/query/app/handler_options.go
@@ -61,7 +61,7 @@ func (handlerOptions) QueryLookbackDuration(queryLookbackDuration time.Duration)
 	}
 }
 
-// Tracer creates a HandlerOption that initializes OpenTracing tracer
+// Tracer creates a HandlerOption that passes the tracer to the handler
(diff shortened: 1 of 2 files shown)


cmd/query/app/handler_options.go
```diff
@@ -61,10 +61,10 @@ func (handlerOptions) QueryLookbackDuration(queryLookbackDuration time.Duration)
 	}
 }
 
-// Tracer creates a HandlerOption that initializes OpenTracing tracer
+// Tracer creates a HandlerOption that initializes Opentelemetry tracer
 func (handlerOptions) Tracer(tracer *jtracer.JTracer) HandlerOption {
 	return func(apiHandler *APIHandler) {
-		apiHandler.tracer = tracer
+		apiHandler.tracer.OTEL = tracer.OTEL
```
yurishkuro, line 67: why this change?

afzalbin64: I did this to workaround if it was desirable.

PR #4635, opened 2023-08-07
[hotrod/observer_test] Switch to OpenTelemetry

afzal442:
* Part of #3381 
* This PR adds `otel` traces and const to hotROD test file

- Replace OT const with `otel trace` spans const i.e. `ext.SpanKind..` with `trace.SpanKind..` and its trace API


examples/hotrod/pkg/tracing/rpcmetrics/observer_test.go
```diff
@@ -47,46 +48,39 @@ func withTestTracer(runTest func(tt *testTracer)) {
...
+		finishOptions := trace.WithTimestamp(ts.Add((50 * time.Millisecond)))
 
 		testCases := []struct {
 			name           string
-			tag            opentracing.Tag
+			tag            trace.SpanKind
 			opNameOverride string
 			err            bool
 		}{
-			{name: "local-span", tag: opentracing.Tag{Key: "x", Value: "y"}},
```
yurishkuro, line 70: in OT span kind is an example of tag, but this test was also using another tag, unrelated to span kind. Your change removes that - did not not have any impact on the tests?

afzal442: Sorry Yuri. How would I replace it with otel traces spankind? Do I need to use attribute keys?

yurishkuro: span kind can be a separate field in the test case, and tag can remain as arbitrary tag (decoupled from span kind)

afzal442: arbitrary tag means?

examples/hotrod/pkg/tracing/rpcmetrics/observer_test.go
```diff
@@ -152,9 +142,20 @@ func TestTags(t *testing.T) {
 		}
 		t.Run(fmt.Sprintf("%s-%v-%s", testCase.key, testCase.value, testCase.variant), func(t *testing.T) {
 			withTestTracer(func(testTracer *testTracer) {
-				span := testTracer.tracer.StartSpan("span", ext.SpanKindRPCServer)
-				span.SetTag(testCase.key, testCase.value)
-				span.Finish()
+				_, span := testTracer.tracer.Start(context.Background(), "span", trace.WithSpanKind(trace.SpanKindServer))
+				switch v := testCase.value.(type) {
```
yurishkuro, line 146: the only values in the matrix that I see are `42` and `true`. Why do you need this large switch statement?

Also, you can initialize the attribute directly in the test cases initialization, and only do `span.SetAttributes(testCase.attr)` here

examples/hotrod/pkg/tracing/rpcmetrics/observer_test.go
```diff
@@ -103,14 +103,15 @@ func TestTags(t *testing.T) {
 		key     string
 		variant string
 		value   interface{}
+		attr    attribute.KeyValue
```
yurishkuro, line 106: why do you keep key/value fields?

examples/hotrod/pkg/tracing/rpcmetrics/observer_test.go
```diff
@@ -103,14 +103,15 @@ func TestTags(t *testing.T) {
...
 		metrics []u.ExpectedMetric
 	}
 
 	testCases := []tagTestCase{
-		{key: "something", value: 42, metrics: []u.ExpectedMetric{
+		{key: "something", value: 42, attr: attribute.Key("something").String(fmt.Sprint(42)), metrics: []u.ExpectedMetric{
```
yurishkuro, line 111: the original type in the test is int, not string

examples/hotrod/pkg/tracing/rpcmetrics/observer_test.go
```diff
@@ -47,46 +48,39 @@ func withTestTracer(runTest func(tt *testTracer)) {
...
+			{name: "get-user-client", spanKind: trace.SpanKindClient},
 		}
 
 		for _, testCase := range testCases {
-			span := testTracer.tracer.StartSpan(
-				testCase.name,
-				testCase.tag,
-				opentracing.StartTime(ts),
-			)
+			_, span := testTracer.tracer.Start(context.Background(), testCase.name, trace.WithSpanKind(testCase.spanKind), trace.WithTimestamp(ts))
```
yurishkuro, line 76: use multi-line formatting for long statements

examples/hotrod/pkg/tracing/rpcmetrics/observer_test.go
```diff
@@ -152,9 +144,9 @@ func TestTags(t *testing.T) {
 		}
 		t.Run(fmt.Sprintf("%s-%v-%s", testCase.key, testCase.value, testCase.variant), func(t *testing.T) {
 			withTestTracer(func(testTracer *testTracer) {
-				span := testTracer.tracer.StartSpan("span", ext.SpanKindRPCServer)
-				span.SetTag(testCase.key, testCase.value)
-				span.Finish()
+				_, span := testTracer.tracer.Start(context.Background(), "span", trace.WithSpanKind(trace.SpanKindServer))
```
yurishkuro, line 147: line too long

examples/hotrod/pkg/tracing/rpcmetrics/observer_test.go
```diff
@@ -109,20 +103,17 @@ func TestTags(t *testing.T) {
 		key     string
 		variant string
```
yurishkuro, line 104: where is it used?

yurishkuro (dismissed):

examples/hotrod/pkg/tracing/rpcmetrics/observer_test.go
afzalbin64: I found here https://github.com/jaegertracing/jaeger/pull/4635/files#diff-763ae75271b1c221da3048d9f551c0fc66cdb9b1b48fbed0d98753bb204c1a6cL141

yurishkuro: that means it's not used, that link is still initialization path

afzalbin64:
Thanks again. But I am wondering why our test is still failing.

examples/hotrod/pkg/tracing/rpcmetrics/observer_test.go
```diff
@@ -106,23 +104,18 @@ func TestObserver(t *testing.T) {
...
 
 	testCases := []tagTestCase{
-		{key: "something", value: 42, metrics: []u.ExpectedMetric{
+		{attr: attribute.Key("something").Int(42), metrics: []u.ExpectedMetric{
 			{Name: "requests", Value: 1, Tags: tags("error", "false")},
 		}},
-		{key: "error", value: true, metrics: []u.ExpectedMetric{
+		{attr: attribute.Key("error").Bool(true), metrics: []u.ExpectedMetric{
 			{Name: "requests", Value: 1, Tags: tags("error", "true")},
 		}},
```
yurishkuro, line 118: this is not going to work anymore. The new code in the observer generates error metrics based on span status
```
	if status := sp.Status(); status.Code == codes.Error {
		mets.RequestCountFailures.Inc(1)
		mets.RequestLatencyFailures.Record(latency)
```
not based on error=true tag that is no longer used/recognized in OTEL instrumentation

yurishkuro: so you need to generate spans with/without error status in order to test different metrics. I am not sure if something=42 has any bearing anymore, it's probably just a placehoder. But you do need the `attr` in order to test `"http.status_code"` tag.

examples/hotrod/pkg/tracing/rpcmetrics/observer_test.go
```diff
@@ -136,8 +129,7 @@ func TestTags(t *testing.T) {
 		}
 		for _, v := range status_codes {
 			testCases = append(testCases, tagTestCase{
-				key:     "http.status_code",
-				value:   v.value,
+				attr:    attribute.Key("http.status_code").String(fmt.Sprint(v.value)),
```
yurishkuro, line 132: There should be a constant in semconv for this.

yurishkuro: Does the OTEL instrumentation actually use string or int for status_code tag?

examples/hotrod/pkg/tracing/rpcmetrics/observer_test.go
```diff
@@ -106,39 +104,20 @@ func TestObserver(t *testing.T) {
...
-		}},
-		{key: "error", value: true, metrics: []u.ExpectedMetric{
-			{Name: "requests", Value: 1, Tags: tags("error", "true")},
-		}},
-		// OTEL bridge does not interpret string "true" as error status
-		// {key: "error", value: "true", variant: "string", metrics: []u.ExpectedMetric{
-		// 	{Name: "requests", Value: 1, Tags: tags("error", "true")},
-		// }},
-	}
+	testCases := []tagTestCase{}
```
yurishkuro, line 110: I don't understand your change - why are you deleting test cases? They had distinct outcomes in the expected metrics.

afzalbin64: sorry! I thought they are getting ignored against the metrics created during spans starting since error=true has no significance for OTEL.

yurishkuro: error cases have the same significance in OTEL, they are just expressed differently, via status instead of a tag

afzalbin64: so how I can re declare it as status? any OTEL fmt?

yurishkuro: didn't my comment answer that yesterday? https://github.com/jaegertracing/jaeger/pull/4635/files#r1290509959

examples/hotrod/pkg/tracing/rpcmetrics/observer_test.go
```diff
@@ -106,39 +104,27 @@ func TestObserver(t *testing.T) {
...
-			value   interface{}
-			variant string
+			value int64
 		}{
-			{value: i},
-			{value: uint16(i), variant: "uint16"},
-			{value: fmt.Sprintf("%d", i), variant: "string"},
+			{value: int64(i)},
 		}
 		for _, v := range status_codes {
```
yurishkuro, line 125: there is only one variant, no need to loop through it. Just use `i` as the value of the attribute.

examples/hotrod/pkg/tracing/rpcmetrics/observer_test.go
```diff
@@ -106,39 +104,27 @@ func TestObserver(t *testing.T) {
...
 			testCases = append(testCases, tagTestCase{
-				key:     "http.status_code",
-				value:   v.value,
-				variant: v.variant,
+				attr: attribute.Key(semconv.HTTPStatusCodeKey).Int64(v.value),
```
yurishkuro, line 127:
```suggestion
				attr: semconv.HTTPStatusCode(i),
```

examples/hotrod/pkg/tracing/rpcmetrics/observer_test.go
```diff
@@ -150,11 +136,17 @@ func TestTags(t *testing.T) {
...
 			withTestTracer(func(testTracer *testTracer) {
-				span := testTracer.tracer.StartSpan("span", ext.SpanKindRPCServer)
-				span.SetTag(testCase.key, testCase.value)
-				span.Finish()
+				_, span := testTracer.tracer.Start(
+					context.Background(),
+					"span", trace.WithSpanKind(trace.SpanKindServer),
+				)
+				span.SetAttributes(testCase.attr)
+				if testCase.attr.Key == attribute.Key(codes.Error.String()) {
```
yurishkuro, line 146:
```suggestion
				if testCase.err {
```

yurishkuro (approved):

PR #4652, opened 2023-08-12
Add shell script to ensure consistent version of OTEL semconv

### LINKED ISSUE #4646: [ci] Add CI check to ensure consistent version of OTEL semconv
# Problem

As seen in this PR #4645, we used a random collection of semconv versions.

# Proposal

* Add a script to search the code for all usages of `"go.opentelemetry.io/otel/semconv/v1.xx.x"` and fail if they are not identical, with an appropriate error message and the exact files that are behind.
* Run that script as part of `make lint`

# Stretch goal

1. A script that automatically fixes the versions (suggest running the script in the output of the lint script)
2. Another check that determines the latest version of semconv available (1.20.0 is the latest right now), and ensure that all files reference that version.

perhapsmaple:
Resolves #4646 

## Description of the changes
- Added scripts/check-semconv-version.sh to search the code for all usages of "go.opentelemetry.io/otel/semconv/v1.xx.x" and fail if they are not identical, with an error message and the source files along with their respective versions.
- Added scripts/update-semconv-version.sh to fetch the latest version of semconv with curl and replace all references of 
"go.opentelemetry.io/otel/semconv/v1.xx.x" with the latest version.

I have tested the changes locally by modifying import statements
Mismatch_Version_Test

## Checklist

### DIFF
--- a/scripts/check-semconv-version.sh
+++ b/scripts/check-semconv-version.sh
@@ -2,20 +2,22 @@
 
 declare -A semconv_map
 
+package_name="go.opentelemetry.io/otel/semconv"
+version_regex="v[0-9]\.[0-9]\+\.[0-9]\+"
+
 while IFS=: read -r file_name package_string; do
-    version_number=$(echo "$package_string" | grep -o -E "v1\.[0-9]+\.[0-9]+")
-    semconv_map["$file_name"]=$version_number
-done < <(find . -type f -name "*.go" -exec grep -o -H "go.opentelemetry.io/otel/semconv/v1\.[0-9]\+\.[0-9]\+" {} +)
+    semconv_map["$file_name"]="${package_string##*/}"
+done < <(find . -type f -name "*.go" -exec grep -o -H "$package_name/$version_regex" {} +)
 
 semconv_versions=($(printf "%s\n" "${semconv_map[@]}" | sort -u))
 
 if [ ${#semconv_versions[@]} -gt 1 ]; then
-    echo "Error: semconv version mismatch detected. Run ./scripts/update-semconv-version.sh to update semconv to latest version."
+    echo "Error: semconv version mismatch detected."
     {
         for key in "${!semconv_map[@]}"; do
-            printf "Source File: %-30s | Semconv Version: %s\n" "$key" "${semconv_map[$key]}"
+            printf "Source File: %-50s | Semconv Version: %s\n" "$key" "${semconv_map[$key]}"
         done
     } | column -t -s '|'
-
+    echo "Run ./scripts/update-semconv-version.sh to update semconv to latest version."
     exit 1
 fi
--- a/scripts/check-semconv-version.sh
+++ b/scripts/check-semconv-version.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+declare -A semconv_map
+
+while IFS=: read -r file_name package_string; do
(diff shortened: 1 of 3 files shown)


Mayhul-Jindal:
@perhapsmaple did you get any error while running `make install-tools` ?

perhapsmaple:
@Mayhul-Jindal I had no errors while running make. Although, install-tools did not download the tparallel module on my machine

perhapsmaple:
@yurishkuro I wanted to know if there was any way other than using curl to determines the latest version of semconv available. The go list command only seems to work for the go.opentelemetry.io/otel package. I would be willing to work on another script that updates all references to the latest version of semconv.

scripts/check-semconv-version.sh
```diff
@@ -10,12 +10,12 @@ done < <(find . -type f -name "*.go" -exec grep -o -H "go.opentelemetry.io/otel/
 semconv_versions=($(printf "%s\n" "${semconv_map[@]}" | sort -u))
 
 if [ ${#semconv_versions[@]} -gt 1 ]; then
-    echo "Error: semconv version mismatch detected"
+    echo "Error: semconv version mismatch detected. Run ./scripts/update-semconv-version.sh to update semconv to latest version."
```
yurishkuro, line 13: I would recommend printing "Run <script>" line at the end of the output, after the listing of files.

scripts/check-semconv-version.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+declare -A semconv_map
+
+while IFS=: read -r file_name package_string; do
+    version_number=$(echo "$package_string" | grep -o -E "v1\.[0-9]+\.[0-9]+")
```
yurishkuro, line 6:
```suggestion
    version_number=$(echo "$package_string" | grep -o -E "v[0-9]\.[0-9]+\.[0-9]+")
```

yurishkuro: also, this regex is used more then once, let's move it to a variable

yurishkuro: also, why do you need this grep here in the first place, isn't it guaranteed to match because of L8?

scripts/check-semconv-version.sh
```diff
@@ -0,0 +1,21 @@
...
+    semconv_map["$file_name"]=$version_number
+done < <(find . -type f -name "*.go" -exec grep -o -H "go.opentelemetry.io/otel/semconv/v1\.[0-9]\+\.[0-9]\+" {} +)
+
+semconv_versions=($(printf "%s\n" "${semconv_map[@]}" | sort -u))
+
+if [ ${#semconv_versions[@]} -gt 1 ]; then
+    echo "Error: semconv version mismatch detected. Run ./scripts/update-semconv-version.sh to update semconv to latest version."
+    {
+        for key in "${!semconv_map[@]}"; do
+            printf "Source File: %-30s | Semconv Version: %s\n" "$key" "${semconv_map[$key]}"
```
yurishkuro, line 16:
```suggestion
            printf "Source File: %-50s | Semconv Version: %s\n" "$key" "${semconv_map[$key]}"
```
30 may not be long enough

perhapsmaple:
Hi @yurishkuro ,I have addressed all of your initial comments. Let me know if there are any further changes required

yurishkuro (approved):
Thanks

PR #4655, opened 2023-08-13
Implement N-1 support policy for Go versions

### LINKED ISSUE #4653: Implement N-1 support policy for Go versions
Follow up to https://github.com/jaegertracing/jaeger/issues/4329.

* [ ] implement a linter check `scripts/check-go-version.sh` that verifies that all our build scripts are using the same Go version N, but go.mod is N-1. This includes GH actions, Alpine images, etc.
* [x] implement a convenience script `scripts/update-go-version.sh` that updates those scripts to another version, to make these upgrades easier in the future (see #4652 for reference). It should also bump go.mod to N-1.
* [x] add a section to README (similar to `Version Compatibility Guarantees`) explaining the policy. For reference: https://github.com/open-telemetry/opentelemetry-collector#compatibility

Since Go 1.21 is out, this is an opportunity to test this new process.

perhapsmaple:
Resolves  #4653

## Description of the changes
- Current files tracked are go.mod, all yml files in ./github/workflows/ and ./docker/Makefile. If there are any more build scripts to track, I would be happy to update the PR.
- Added scripts/check-go-version.sh to verify that all our build scripts are using the same Go version N, but go.mod is N-1 and fail if not, with an error message and the source files along with their respective versions.
- Added scripts/update-go-version.sh to update those scripts to latest version, to make these upgrades easier in the future. Also bumps go.mod to version N-1.
- Updated README explaining the support policy.

- I have tested the changes locally by modifying go versions
Screenshot_20230813_120937
Screenshot_20230813_121021

## Checklist


README.md
```diff
@@ -141,6 +141,16 @@ For example, consider a scenario where v1.28.0 is released on 01-Jun-2021 contai
...
 It may remain deprecated for longer than the aforementioned grace period.
 
+## Go Version Compatibility Guarantees
+
+The Jaeger project attempts to track the currently supported versions of Go, as [defined by the Go team](https://go.dev/doc/devel/release#policy).
+Removing support for an unsupported Go version is not considered a breaking change.
+
+Starting with the release of Go 1.21, support for Go versions will be updated as follows:
+
+1. The first release after the release of a new Go minor version `N` will add build and tests steps for the new Go minor version.
```
yurishkuro, line 151: I am not sure we want to promise that such changes will be precisely in the "**next release** after Go release". Let's change to a bit more flexible phrasing like "soon after Go release"

scripts/check-go-version.sh
```diff
@@ -0,0 +1,45 @@
...
+
+# Fetch latest go release version
+go_latest_version=$(curl -s https://go.dev/dl/?mode=json | jq -r '.[0].version' | awk -F'.' '{gsub("go", ""); print $1"."$2}')
+
+# Extract go version from go.mod
+go_mod_path="./go.mod"
+go_mod_version=$(grep -oP "go\s+\K$version_regex+" $go_mod_path)
+
+# Ensure go.mod uses go_latest_version - 1
+if [ "${go_mod_version%.*}" != "${go_latest_version%.*}" ] || [ $((10#${go_latest_version#*.} - 10#${go_mod_version#*.})) != 1 ]; then
```
yurishkuro, line 14: I don't think this is what we want. With this check the CI will start failing immediately after Go release, which is unfair to PR authors. My intention was that the script checks that all other build scripts are on some version N and the go.mod is on N-1, but N does not have to be latest. Latest can be printed, but not enforced. It should only be used in the update script.

scripts/update-go-version.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
```
yurishkuro, line 1: This file has a large overlap with the check script, they need to be in sync regarding all the places they check. It would make sense to combine them into one and have an optional "update" behavior. Eg check script can take -u flag, and update script simply calls 'check -u'

perhapsmaple:
Hi @yurishkuro , I have updated the scripts according. I'm wondering if the update script is required, since it just calls the check-go-version script with a flag.

README.md
```diff
@@ -141,6 +141,16 @@ For example, consider a scenario where v1.28.0 is released on 01-Jun-2021 contai
...
 
+## Go Version Compatibility Guarantees
+
+The Jaeger project attempts to track the currently supported versions of Go, as [defined by the Go team](https://go.dev/doc/devel/release#policy).
+Removing support for an unsupported Go version is not considered a breaking change.
+
+Starting with the release of Go 1.21, support for Go versions will be updated as follows:
+
+1. Soon after the release of a new Go minor version `N`, updates will be made to the build and tests steps to accommodate the latest Go minor version. 
+2. Soon after the release of a new Go minor version `N`, support for Go version `N-2` will be removed.
```
yurishkuro, line 152:
```suggestion
2. Soon after the release of a new Go minor version `N`, support for Go version `N-2` will be removed and version `N-1` will become the minimum required version.
```

scripts/check-go-version.sh
```diff
@@ -0,0 +1,96 @@
...
+    case $opt in
+        u) update=true ;;
+        *) echo "Usage: $0 [-u]" >&2
+           exit 1
+           ;;
+    esac
+done
+
+# Fetch latest go release version
+go_latest_version=$(curl -s https://go.dev/dl/?mode=json | jq -r '.[0].version' | awk -F'.' '{gsub("go", ""); print $1"."$2}')
```
yurishkuro, line 16: I suggest parsing it here into `major` and `minor` to simplify the rest of the code

scripts/check-go-version.sh
```diff
@@ -0,0 +1,96 @@
...
+
+# Extract go version from go.mod
+go_mod_path="./go.mod"
+go_mod_version=$(grep -oP "go\s+\K$version_regex+" $go_mod_path)
+
+# If update, set go.mod version as latest version - 1 (N-1)
+if [ "$update" = true ]; then
+    go_mod_updated="${go_latest_version%.*}.$((10#${go_latest_version#*.} - 1))"
```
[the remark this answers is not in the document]
yurishkuro: `go_mod_updated` is odd name, perhaps `go_mod_desired`

scripts/check-go-version.sh
```diff
@@ -0,0 +1,96 @@
...
+fi
+
+# Extract go version from GitHub actions scripts
+while IFS=: read -r build_script version_string; do
+    
+    # If update, set latest go version for all GitHub actions scripts
+    if [ "$update" = true ]; then
+        if [ "${version_string##*: }" != "$go_latest_version" ]; then
+            updated_version="${version_string%%:*}: ${go_latest_version}"
+            sed -i "s/$version_string/$updated_version/g" "$build_script"
```
yurishkuro, line 40: how do you know this will not update some other part of the file? E.g. some GH actions are included via hash, which could in theory match the string you're substituting

scripts/check-go-version.sh
```diff
@@ -0,0 +1,96 @@
...
+        {
+            printf "Build script: %-60s | Previous Go Version: %s | Updated Go Version: %s\n" "$alpine_makefile_path" "${go_alpine_version##*:}" "${go_latest_version}"
+        } | column -t -s '|'
+    fi
+else
+    go_version_map["$alpine_makefile_path"]="${go_alpine_version##*:}"
+fi
+
+# Ensure N-1 support policy
+if [ "$update" != true ]; then
```
yurishkuro, line 68: this is not really making the code shorter, you still have completely independent code paths

scripts/update-go-version.sh
```diff
@@ -0,0 +1,3 @@
...
+
+./scripts/check-go-version.sh -u
```
yurishkuro, line 3: I'm fine to drop this script

yurishkuro (approved):

yurishkuro:
This did not fully resolve the ticket because there is no linter check, but let's first start with manual updates.

PR #4662, opened 2023-08-14
[fix] Disable tracing of OTLP Receiver

### LINKED ISSUE #4657: [Bug]: recursive tracing in all-in-one
### What happened?

image

### Steps to reproduce

1. run all-in-one
2. reload screen a few times
3. observe above trace

### Expected behavior

In the past only read queries to the query-service were logged. The span above is the write part, where Jaeger submits tracing spans back to itself. We never instrument the write path as that creates recursive trace collection. We need to exclude the write endpoints from being traced by OTEL instrumentation.

cc @afzal442 

### Relevant log output

_No response_

### Screenshot

_No response_

### Additional context

_No response_

### Jaeger backend version

_No response_

### SDK

_No response_

### Pipeline

_No response_

### Stogage backend

_No response_

### Operating system

_No response_

### Deployment model

_No response_

### Deployment configs

_No response_

yurishkuro:
- Resolves #4657

## Description of the changes
- use NoOp tracer provider in the OTLP receiver configuration

- ran the UI, reloaded page many times, this endpoint no longer shows up

### DIFF
--- a/cmd/collector/app/handler/otlp_receiver.go
+++ b/cmd/collector/app/handler/otlp_receiver.go
@@ -28,8 +28,8 @@ import (
 	"go.opentelemetry.io/collector/pdata/ptrace"
 	"go.opentelemetry.io/collector/receiver"
 	"go.opentelemetry.io/collector/receiver/otlpreceiver"
-	"go.opentelemetry.io/otel"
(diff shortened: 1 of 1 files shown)


albertteoh (approved):

albertteoh:
Thanks!

PR #4674, opened 2023-08-15
[zipkin] Replace zipkin exporter from jaeger sdk with otel zipkin exp

afzal442:
* Part of #3381 
* This PR removes `zipkin exp` from jaeger sdk and replaces it with otel zipkin exp


go.mod
```diff
@@ -4,7 +4,7 @@ go 1.20
 
 require (
 	github.com/HdrHistogram/hdrhistogram-go v1.1.2
-	github.com/Shopify/sarama v1.33.0
+	github.com/Shopify/sarama v1.37.2
```
mmorel-35, line 7: How is it related to the zipkin exporter ?

afzal442: that's right. I spotted that. When I added otel zipkin exporter pkg, I ran `go mod tidy` which caused abv pkg upgraded.

cmd/collector/app/zipkin/http_handler_test.go
```diff
@@ -72,39 +71,13 @@ func initializeTestServer(err error) (*httptest.Server, *APIHandler) {
...
-			return len(handler.getSpans()) == expecting
-		},
-		2*time.Second,
-		time.Millisecond,
-		"expecting to receive %d span(s), have %d span(s)",
-		expecting,
-		len(handler.getSpans()),
-	)
+	assert.NoError(t, zexp.Shutdown(context.Background()))
+	assert.NoError(t, zexp.ExportSpans(context.Background(), nil))
```
yurishkuro, line 80: why do you think we had `waitForSpan` function?

yurishkuro: So you don't see a difference between _exporting_ spans and _receiving_ them? What is this whole file is supposed to test?

afzalbin64: If I misunderstood here, would I just need to rmv it from that TestViaClient()? Since L139 requires this.

yurishkuro: This is what the test was doing before your change:
  * create HTTP server/handler receiving Zipkin span
  * use Jaeger SDK with Zipkin exporter to generate spans and export them to that server/handler
  * validate that the handler received the expected spans

afzalbin64: Thanks Yuri for this info. So I removed waitForSpans(...) from TestViaClient() to get it not checked for spans received. I wondered if we need to init otel tracer for it with Zipkin exporter to generate spans. If yes, how?

yurishkuro: > so I removed waitForSpans(...) from TestViaClient() to get it not checked for spans received. 

so you still this it was the right thing to do to remove it?

> we need to init otel tracer for it with Zipkin exporter to generate spans. If yes, how?

why is this different from all the other times you already initialized OTEL tracer?

afzalbin64: So now I got you. That had not to be removed against the spans generated with some purpose.
> why is this different from all the other times you already initialized OTEL tracer?

I think becaz otel doesn't support for this `NewRemoteReporter(zipkinSender)`.

yurishkuro: > I think becaz otel doesn't support for this NewRemoteReporter(zipkinSender).

you previously instantiated tracer with OTLP exporter. Why is Zipkin exporter any different?

afzalbin64: @yurishkuro I am still getting no spans after init tracer. LMK if I am missing something.

yurishkuro: You need to call waitForSpans, export happens async.

afzalbin64: When I call it, zipkin exporter doesn't reflect exported spans. I mean `len(handler.zipkinSpansHandler.(*mockZipkinHandler).spans)` remains 0.

yurishkuro: Did you debug it? Is the exporter successful in exporting data? (I' pretty sure about what the issue is, but you should be able to troubleshoot issues like this yourself, especially in this very narrow setup).

afzalbin64: Yes Yuri, that's what I have been doing debugging the test so far. Let me take a look again after many attempts. Thanks.

yurishkuro: When the client can't connect to the server, one of the first things to check if the client has the right address. But also, client should be able to provide / log an error message indicating why it cannot send data.

afzal442: Seems to be listening on different port.

yurishkuro: Port is provided in server.URL, it should be correct. Anything else?

image

afzalbin64: Sorry Yuri, I have changed that code a bit locally. After debugging, I am stuck.
image
 LMK!

yurishkuro: see my previous comment. You changed the URL for v2.

afzalbin64: Yes I know I did it intentionally to figure out the zipkin exporter. But still I don't get despite `/api/v1/spans`

cmd/collector/app/zipkin/http_handler_test.go
```diff
@@ -72,25 +84,23 @@ func initializeTestServer(err error) (*httptest.Server, *APIHandler) {
...
 func TestViaClient(t *testing.T) {
-	server, handler := initializeTestServer(nil)
+	server, _ := initializeTestServer(nil)
 	defer server.Close()
 
-	zipkinSender, err := zipkinTransport.NewHTTPTransport(
-		server.URL+`/api/v1/spans`,
-		zipkinTransport.HTTPBatchSize(1),
-	)
+	zexp, err := zipkin.New(server.URL + `/api/v2/spans`)
```
yurishkuro, line 90:
```suggestion
	zexp, err := zipkin.New(server.URL + `/api/v1/spans`)
```

cmd/collector/app/zipkin/http_handler_test.go
```diff
@@ -72,25 +84,23 @@ func initializeTestServer(err error) (*httptest.Server, *APIHandler) {
...
 	require.NoError(t, err)
+	trace, closer := initTracer(t, zexp)
+	defer closer()
+	assert.NoError(t, zexp.Shutdown(context.Background()))
+	assert.NoError(t, zexp.ExportSpans(context.Background(), nil))
```
yurishkuro, line 95:
```suggestion
```

cmd/collector/app/zipkin/http_handler_test.go
```diff
@@ -72,25 +84,23 @@ func initializeTestServer(err error) (*httptest.Server, *APIHandler) {
...
 
-	tracer, closer := jaegerClient.NewTracer(
-		"test",
-		jaegerClient.NewConstSampler(true),
-		jaegerClient.NewRemoteReporter(zipkinSender),
-	)
-	defer closer.Close()
+	_, span := trace.Tracer("test").Start(context.Background(), "root")
+	defer span.End()
```
yurishkuro, line 98:
```suggestion
	span.End()
```

cmd/collector/app/zipkin/http_handler_test.go
```diff
@@ -72,25 +84,23 @@ func initializeTestServer(err error) (*httptest.Server, *APIHandler) {
...
 
-	tracer.StartSpan("root").Finish()
+	// ft := handler.zipkinSpansHandler.(*mockZipkinHandler).spans
+	// require.Len(t, len(ft), 1)
 
-	waitForSpans(t, handler.zipkinSpansHandler.(*mockZipkinHandler), 1)
+	// waitForSpans(t, handler.zipkinSpansHandler.(*mockZipkinHandler), 1)
```
yurishkuro, line 103:
```suggestion
	waitForSpans(t, handler.zipkinSpansHandler.(*mockZipkinHandler), 1)
```

cmd/collector/app/zipkin/http_handler_test.go
```diff
@@ -72,25 +84,23 @@ func initializeTestServer(err error) (*httptest.Server, *APIHandler) {
 }
 
 func TestViaClient(t *testing.T) {
-	server, handler := initializeTestServer(nil)
+	server, _ := initializeTestServer(nil)
```
yurishkuro, line 87:
```suggestion
	server, handler := initializeTestServer(nil)
```

yurishkuro (commented):
works for me with these changes

cmd/collector/app/zipkin/http_handler_test.go
```diff
@@ -75,20 +87,13 @@ func TestViaClient(t *testing.T) {
 	server, handler := initializeTestServer(nil)
 	defer server.Close()
 
-	zipkinSender, err := zipkinTransport.NewHTTPTransport(
-		server.URL+`/api/v1/spans`,
-		zipkinTransport.HTTPBatchSize(1),
-	)
+	zexp, err := zipkin.New(server.URL + `/api/v1/spans`)
```
afzal442, line 90: Actually Yuri, I shut downed the exporter after sending the spans frequently. Thanks for quick kind help

yurishkuro (approved):

PR #4681, opened 2023-08-17
[chore] Stop using constants from jaeger-client-go

afzal442:
* Part of #3381 
* This PR removes `jaeger-client` pkg constant and replaces it with const variable

### DIFF
--- a/cmd/collector/app/root_span_handler.go
+++ b/cmd/collector/app/root_span_handler.go
@@ -34,7 +34,7 @@ func handleRootSpan(aggregator strategystore.Aggregator, logger *zap.Logger) Pro
 			return
 		}
 		samplerType, samplerParam := span.GetSamplerParams(logger)
-		if samplerType == "" {
+		if samplerType == 0 {
 			return
 		}
 		aggregator.RecordThroughput(service, span.OperationName, samplerType, samplerParam)
--- a/cmd/collector/app/root_span_handler.go
+++ b/cmd/collector/app/root_span_handler.go
@@ -34,7 +34,7 @@ func handleRootSpan(aggregator strategystore.Aggregator, logger *zap.Logger) Pro
 			return
 		}
 		samplerType, samplerParam := span.GetSamplerParams(logger)
-		if samplerType == 0 {
+		if samplerType == model.SamplerTypeUnrecognized {
 			return
 		}
 		aggregator.RecordThroughput(service, span.OperationName, samplerType, samplerParam)
--- a/model/span.go
+++ b/model/span.go
@@ -20,7 +20,6 @@ import (
 	"io"
 	"strconv"
 
-	"github.com/uber/jaeger-client-go"
 	"go.opentelemetry.io/otel/trace"
 	"go.uber.org/zap"
 )
@@ -57,12 +56,16 @@ var toSpanKind = map[string]trace.SpanKind{
 	"internal": trace.SpanKindInternal,
 }
 
-var toSamplerType = map[string]SamplerType{
-	"unknown":       SamplerTypeUnrecognized,
-	"probabilistic": SamplerTypeProbabilistic,
-	"lowerbound":    SamplerTypeLowerBound,
-	"ratelimiting":  SamplerTypeRateLimiting,
-	"const":         SamplerTypeConst,
+var toSamplerType = map[SamplerType]string{
+	SamplerTypeUnrecognized:  "unrecognized",
+	SamplerTypeProbabilistic: "probabilistic",
+	SamplerTypeLowerBound:    "lowerbound",
+	SamplerTypeRateLimiting:  "ratelimiting",
+	SamplerTypeConst:         "const",
+}
+
+func (s SamplerType) String() string {
+	return toSamplerType[s]
 }
 
 // Hash implements Hash from Hashable.
@@ -97,8 +100,10 @@ func (s *Span) GetSamplerType() SamplerType {
 	if tag, ok := KeyValues(s.Tags).FindByKey(keySamplerType); ok {
 		if tag.VStr == "" {
 			return SamplerTypeUnrecognized
-		} else if tag.VStr == "probabilistic" {
+		} else if tag.VStr == toSamplerType[SamplerTypeProbabilistic] {
 			return SamplerTypeProbabilistic
+		} else if tag.VStr == toSamplerType[SamplerTypeRateLimiting] {
+			return SamplerTypeRateLimiting
 		} else {
 			return SamplerTypeLowerBound
 		}
@@ -163,23 +168,12 @@ func (s *Span) ReplaceParentID(newParentID SpanID) {
 
 // GetSamplerParams returns the sampler.type and sampler.param value if they are valid.
 func (s *Span) GetSamplerParams(logger *zap.Logger) (SamplerType, float64) {
-	tag, ok := KeyValues(s.Tags).FindByKey(jaeger.SamplerTypeTagKey)
-	if !ok {
-		return SamplerTypeUnrecognized, 0
-	}
-	if tag.VType != StringType {
-		logger.
-			With(zap.String("traceID", s.TraceID.String())).
-			With(zap.String("spanID", s.SpanID.String())).
-			Warn("sampler.type tag is not a string", zap.Any("tag", tag))
-		return SamplerTypeUnrecognized, 0
-	}
-	samplerType := toSamplerType[tag.AsString()]
+	samplerType := s.GetSamplerType()
 	if samplerType != SamplerTypeProbabilistic && samplerType != SamplerTypeLowerBound &&
 		samplerType != SamplerTypeRateLimiting {
 		return SamplerTypeUnrecognized, 0
 	}
-	tag, ok = KeyValues(s.Tags).FindByKey(keySamplerParam)
+	tag, ok := KeyValues(s.Tags).FindByKey(keySamplerParam)
 	if !ok {
 		return SamplerTypeUnrecognized, 0
 	}
--- a/model/span.go
+++ b/model/span.go
@@ -62,6 +62,7 @@ var toSamplerType = map[string]SamplerType{
 	"probabilistic": SamplerTypeProbabilistic,
 	"lowerbound":    SamplerTypeLowerBound,
 	"ratelimiting":  SamplerTypeRateLimiting,
+	"const":         SamplerTypeConst,
 }
 
 func (s SamplerType) String() string {
@@ -74,6 +75,8 @@ func (s SamplerType) String() string {
 		return "lowerbound"
 	case SamplerTypeRateLimiting:
 		return "ratelimiting"
+	case SamplerTypeConst:
+		return "const"
 	default:
 		return ""
 	}
@@ -109,10 +112,9 @@ func (s *Span) GetSpanKind() (spanKind trace.SpanKind, found bool) {
 func (s *Span) GetSamplerType() SamplerType {
 	// There's no corresponding opentelemetry tag label corresponding to sampler.type
 	if tag, ok := KeyValues(s.Tags).FindByKey(keySamplerType); ok {
-		if tag.VStr == "" {
-			return toSamplerType[SamplerTypeUnrecognized.String()]
+		if s, ok := toSamplerType[tag.VStr]; ok {
+			return s
 		}
-		return toSamplerType[tag.VStr]
 	}
 	return SamplerTypeUnrecognized
 }
--- a/model/span.go
+++ b/model/span.go
@@ -57,6 +57,13 @@ var toSpanKind = map[string]trace.SpanKind{
 	"internal": trace.SpanKindInternal,
 }
 
+var toSamplerType = map[string]SamplerType{
+	"unrecognized":  SamplerTypeUnrecognized,
+	"probabilistic": SamplerTypeProbabilistic,
+	"lowerbound":    SamplerTypeLowerBound,
+	"ratelimiting":  SamplerTypeRateLimiting,
+}
+
 func (s SamplerType) String() string {
 	switch s {
 	case SamplerTypeUnrecognized:
@@ -67,8 +74,6 @@ func (s SamplerType) String() string {
 		return "lowerbound"
 	case SamplerTypeRateLimiting:
 		return "ratelimiting"
-	case SamplerTypeConst:
-		return "const"
 	default:
 		return ""
 	}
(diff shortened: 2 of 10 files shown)


plugin/sampling/strategystore/adaptive/aggregator.go
```diff
@@ -18,16 +18,15 @@ import (
...
 	"github.com/jaegertracing/jaeger/cmd/collector/app/sampling/model"
 	"github.com/jaegertracing/jaeger/cmd/collector/app/sampling/strategystore"
 	"github.com/jaegertracing/jaeger/pkg/metrics"
 	"github.com/jaegertracing/jaeger/storage/samplingstore"
 )
 
 const (
-	maxProbabilities = 10
+	maxProbabilities         = 10
+	SamplerTypeProbabilistic = "probabilistic"
```
albertteoh, line 29: Does this need to be exported?

mmorel-35: As they can be found in several places, maybe they could be moved in an internal package dedicated to those constants and they won't be public but only dedicated to jaeger.

afzal442: Thanks. into it.

model/span.go
```diff
@@ -82,6 +83,8 @@ func (s *Span) GetSamplerType() string {
 	if tag, ok := KeyValues(s.Tags).FindByKey(samplerType); ok {
 		if tag.VStr == "" {
 			return samplerTypeUnknown
+		} else if tag.VStr == "probabilistic" {
+			return samplerTypeProbabilistic
```
yurishkuro, line 87: this seems redundant

model/span.go
```diff
@@ -33,9 +33,10 @@ const (
 	// FirehoseFlag is the bit in Flags in order to define a span as a firehose span
 	FirehoseFlag = Flags(8)
 
-	samplerType        = "sampler.type"
-	keySpanKind        = "span.kind"
-	samplerTypeUnknown = "unknown"
+	samplerType              = "sampler.type"
```
yurishkuro, line 36: to be consistent
```suggestion
	keySamplerType              = "sampler.type"
```

model/span.go
```diff
@@ -33,9 +33,10 @@ const (
...
+	keySpanKind              = "span.kind"
+	samplerTypeUnknown       = "unknown"
+	samplerTypeProbabilistic = "probabilistic"
```
yurishkuro, line 39: These could be made public, since they are returned from the public method GetSamplerType, and you need at least one of them in the aggregator code.

An even better solution is to make this strongly typed:
```
type SamplerType int

const (
    SamplerTypeUnknown SamplerType = iota
    SamplerTypeProbabilistic
}

func (s *Span) GetSamplerType() SamplerType {}
```

model/span.go
```diff
@@ -144,34 +160,34 @@ func (s *Span) ReplaceParentID(newParentID SpanID) {
 }
 
 // GetSamplerParams returns the sampler.type and sampler.param value if they are valid.
-func (s *Span) GetSamplerParams(logger *zap.Logger) (string, float64) {
+func (s *Span) GetSamplerParams(logger *zap.Logger) (SamplerType, float64) {
 	tag, ok := KeyValues(s.Tags).FindByKey(jaeger.SamplerTypeTagKey)
 	if !ok {
-		return "", 0
+		return 0, 0
```
yurishkuro, line 166:
```suggestion
		return SamplerTypeUnrecognized, 0
```
this is why we're defining named constants

afzal442:
But Yuri, I got to find chaining effect due to that.
https://github.com/jaegertracing/jaeger/blob/10d18e680e5b3c9043c7d344156848905787dd78/cmd/collector/app/metrics.go#L234

I am stuck here since its util is having unchanged method. LMK!
https://github.com/jaegertracing/jaeger/blob/10d18e680e5b3c9043c7d344156848905787dd78/cmd/collector/app/metrics.go#L281

model/span.go
```diff
@@ -144,34 +160,34 @@ func (s *Span) ReplaceParentID(newParentID SpanID) {
...
 			With(zap.String("spanID", s.SpanID.String())).
 			Warn("sampler.type tag is not a string", zap.Any("tag", tag))
-		return "", 0
+		return 0, 0
 	}
-	samplerType := tag.AsString()
-	if samplerType != jaeger.SamplerTypeProbabilistic && samplerType != jaeger.SamplerTypeLowerBound &&
-		samplerType != jaeger.SamplerTypeRateLimiting {
-		return "", 0
+	samplerType := toSamplerType[tag.AsString()]
```
yurishkuro, line 177:
```suggestion
	samplerType, ok := toSamplerType[tag.AsString()]
```
and check for ok

model/span.go
```diff
@@ -77,15 +90,18 @@ func (s *Span) GetSpanKind() (spanKind trace.SpanKind, found bool) {
...
 // GetSamplerType returns the sampler type for span
-func (s *Span) GetSamplerType() string {
+func (s *Span) GetSamplerType() SamplerType {
 	// There's no corresponding opentelemetry tag label corresponding to sampler.type
-	if tag, ok := KeyValues(s.Tags).FindByKey(samplerType); ok {
+	if tag, ok := KeyValues(s.Tags).FindByKey(keySamplerType); ok {
 		if tag.VStr == "" {
-			return samplerTypeUnknown
+			return SamplerTypeUnrecognized
+		} else if tag.VStr == "probabilistic" {
```
yurishkuro, line 100: why are you not using `toSamplerType` here?

model/span.go
```diff
@@ -25,17 +25,23 @@ import (
 	"go.uber.org/zap"
 )
 
+type SamplerType int
+
 const (
+	SamplerTypeUnrecognized SamplerType = iota
+	SamplerTypeProbabilistic
+	SamplerTypeLowerBound
+	SamplerTypeRateLimiting
```
yurishkuro, line 34:
```suggestion
	SamplerTypeRateLimiting

```

model/span.go
```diff
@@ -50,6 +56,13 @@ var toSpanKind = map[string]trace.SpanKind{
 	"internal": trace.SpanKindInternal,
 }
 
+var toSamplerType = map[string]SamplerType{
+	"unknown":       SamplerTypeUnrecognized,
```
yurishkuro, line 61: this is not needed, you won't get a string like that in the span tags

model/span.go
```diff
@@ -144,34 +160,34 @@ func (s *Span) ReplaceParentID(newParentID SpanID) {
 }
 
 // GetSamplerParams returns the sampler.type and sampler.param value if they are valid.
-func (s *Span) GetSamplerParams(logger *zap.Logger) (string, float64) {
+func (s *Span) GetSamplerParams(logger *zap.Logger) (SamplerType, float64) {
 	tag, ok := KeyValues(s.Tags).FindByKey(jaeger.SamplerTypeTagKey)
```
yurishkuro, line 166: reuse GetSamplerType() function here

model/span_test.go
```diff
@@ -398,106 +400,106 @@ func TestGetSamplerParams(t *testing.T) {
...
+		expectedType  model.SamplerType
 		expectedParam float64
 	}{
 		{
 			tags: model.KeyValues{
 				model.String("sampler.type", "probabilistic"),
 				model.String("sampler.param", "1e-05"),
 			},
-			expectedType:  "probabilistic",
+			expectedType:  1,
```
yurishkuro: why are you using numbers?

model/span.go
```diff
@@ -50,6 +57,14 @@ var toSpanKind = map[string]trace.SpanKind{
 	"internal": trace.SpanKindInternal,
 }
 
+var toSamplerType = map[string]SamplerType{
+	"unknown":       SamplerTypeUnrecognized,
+	"probabilistic": SamplerTypeProbabilistic,
+	"lowerbound":    SamplerTypeLowerBound,
+	"ratelimiting":  SamplerTypeRateLimiting,
+	"const":         SamplerTypeConst,
+}
```
yurishkuro, line 66: you may also want to define `func (t SamplerType) String() string` function, since it's needed in cmd/collector/app/metrics.go

model/span.go
```diff
@@ -77,15 +95,20 @@ func (s *Span) GetSpanKind() (spanKind trace.SpanKind, found bool) {
...
-	if tag, ok := KeyValues(s.Tags).FindByKey(samplerType); ok {
+	if tag, ok := KeyValues(s.Tags).FindByKey(keySamplerType); ok {
 		if tag.VStr == "" {
-			return samplerTypeUnknown
+			return SamplerTypeUnrecognized
+		} else if tag.VStr == toSamplerType[SamplerTypeProbabilistic] {
+			return SamplerTypeProbabilistic
+		} else if tag.VStr == toSamplerType[SamplerTypeRateLimiting] {
+			return SamplerTypeRateLimiting
+		} else {
```
yurishkuro, line 107: Why is "else" a lower bound? What if the value is "xyz"?

model/span.go
```diff
@@ -77,15 +95,20 @@ func (s *Span) GetSpanKind() (spanKind trace.SpanKind, found bool) {
 }
 
 // GetSamplerType returns the sampler type for span
-func (s *Span) GetSamplerType() string {
+func (s *Span) GetSamplerType() SamplerType {
 	// There's no corresponding opentelemetry tag label corresponding to sampler.type
-	if tag, ok := KeyValues(s.Tags).FindByKey(samplerType); ok {
+	if tag, ok := KeyValues(s.Tags).FindByKey(keySamplerType); ok {
 		if tag.VStr == "" {
```
yurishkuro, line 101: Redundant check

model/span.go
```diff
@@ -77,15 +95,20 @@ func (s *Span) GetSpanKind() (spanKind trace.SpanKind, found bool) {
...
-			return samplerTypeUnknown
+			return SamplerTypeUnrecognized
+		} else if tag.VStr == toSamplerType[SamplerTypeProbabilistic] {
```
yurishkuro, line 103: Why use map for static lookups? It's a waste of cpu cycles

cmd/collector/app/root_span_handler.go
```diff
@@ -34,7 +34,7 @@ func handleRootSpan(aggregator strategystore.Aggregator, logger *zap.Logger) Pro
 			return
 		}
 		samplerType, samplerParam := span.GetSamplerParams(logger)
-		if samplerType == "" {
+		if samplerType == model.SamplerTypeConst {
```
yurishkuro, line 37: Why is "" equivalent to const?

model/span.go
yurishkuro: keep blank line between enums and other constants, this is not a logical continuation

model/span.go
```diff
@@ -20,22 +20,28 @@ import (
 	"io"
 	"strconv"
 
-	"github.com/uber/jaeger-client-go"
 	"go.opentelemetry.io/otel/trace"
 	"go.uber.org/zap"
 )
 
+type SamplerType int
```
mmorel-35, line 27: @yurishkuro ,
Why making it an int type with a string value while you can making it a string directly ?
```suggestion
type SamplerType string
```

model/span.go
```diff
@@ -20,22 +20,28 @@ import (
...
+
 const (
+	SamplerTypeUnrecognized SamplerType = iota
```
afzal442, line 30:
```suggestion
	SamplerTypeUnrecognized SamplerType = "unrecognized"
	SamplerTypeProbabilistic = "probabilistic"
```

model/span.go
yurishkuro: int type is faster for comparisons

mmorel-35: Ok 👍🏻

model/span.go
```diff
@@ -77,15 +101,23 @@ func (s *Span) GetSpanKind() (spanKind trace.SpanKind, found bool) {
...
 // GetSamplerType returns the sampler type for span
-func (s *Span) GetSamplerType() string {
+func (s *Span) GetSamplerType() SamplerType {
 	// There's no corresponding opentelemetry tag label corresponding to sampler.type
-	if tag, ok := KeyValues(s.Tags).FindByKey(samplerType); ok {
-		if tag.VStr == "" {
-			return samplerTypeUnknown
+	if tag, ok := KeyValues(s.Tags).FindByKey(keySamplerType); ok {
+		switch tag.VStr {
+		case SamplerTypeProbabilistic.String():
```
yurishkuro, line 108: this doesn't make sense to me. String() is already implemented as a switch statement, which is sequential comparison, so you're using O(n^2) resolution here, instead of O(1) when you had a map.

model/span.go
```diff
@@ -144,34 +176,23 @@ func (s *Span) ReplaceParentID(newParentID SpanID) {
...
-		return "", 0
-	}
-	samplerType := tag.AsString()
-	if samplerType != jaeger.SamplerTypeProbabilistic && samplerType != jaeger.SamplerTypeLowerBound &&
-		samplerType != jaeger.SamplerTypeRateLimiting {
-		return "", 0
+func (s *Span) GetSamplerParams(logger *zap.Logger) (SamplerType, float64) {
+	samplerType := s.GetSamplerType()
+	if samplerType != SamplerTypeProbabilistic && samplerType != SamplerTypeLowerBound &&
+		samplerType != SamplerTypeRateLimiting {
```
yurishkuro, line 182:
```suggestion
	if samplerType == SamplerTypeUnrecognized {
```

model/span.go
afzal442: > Why use map for static lookups? It's a waste of cpu cycles

Sorry! but you told map is not needed. Or maybe I misunderstood it

yurishkuro: you're taking it out of context. 
  * `someMap[someLiteralConstant]` is a waste of CPU. 
  * `someMap[someVariable]` is not a waste, but likely more efficient than a switch statement

model/span.go
```diff
@@ -50,6 +57,28 @@ var toSpanKind = map[string]trace.SpanKind{
...
+func (s SamplerType) String() string {
+	switch s {
+	case SamplerTypeUnrecognized:
+		return "unrecognized"
+	case SamplerTypeProbabilistic:
+		return "probabilistic"
+	case SamplerTypeLowerBound:
+		return "lowerbound"
+	case SamplerTypeRateLimiting:
+		return "ratelimiting"
```
yurishkuro, line 76: missing const?

model/span.go
```diff
@@ -77,15 +106,15 @@ func (s *Span) GetSpanKind() (spanKind trace.SpanKind, found bool) {
...
 
 // GetSamplerType returns the sampler type for span
-func (s *Span) GetSamplerType() string {
+func (s *Span) GetSamplerType() SamplerType {
 	// There's no corresponding opentelemetry tag label corresponding to sampler.type
-	if tag, ok := KeyValues(s.Tags).FindByKey(samplerType); ok {
+	if tag, ok := KeyValues(s.Tags).FindByKey(keySamplerType); ok {
 		if tag.VStr == "" {
-			return samplerTypeUnknown
+			return toSamplerType[SamplerTypeUnrecognized.String()]
```
yurishkuro, line 113: can you explain what you're trying to do here?

model/span.go
```diff
@@ -77,15 +106,15 @@ func (s *Span) GetSpanKind() (spanKind trace.SpanKind, found bool) {
...
 		}
-		return tag.VStr
+		return toSamplerType[tag.VStr]
```
yurishkuro, line 115:
```suggestion
		if s, ok := toSamplerType[tag.VStr]; ok {
				return s
		}
```

yurishkuro: and `if tag.VStr == "" {` above is not necessary, because having `sampler.type=""` in the span is very unlikely, but the map lookup for `""` will take care of it anyway

afzal442: Thanks again Yuri for this hint. 😄

yurishkuro (approved):

PR #4697, opened 2023-08-22
3085 - Change type of logs from nested to object to reduce fixedbit memory consumption

### LINKED ISSUE #3085: [Jaeger Collector - ES Backend ]Use object type instead of nested type for logs field in index mapping for jaeger-span

bhiravabhatla:
Resolves #3085 
## Description of the changes
Adds a flag to disable search on logs field and changes the type of logs from nested to object to reduce  fixedbit memory consumption.

Added required unit tests. Tested in one of our project's lower environment end to end.

## Checklist
- [yes] I have read https://github.com/jaegertracing/jaeger/blob/master/CONTRIBUTING_GUIDELINES.md
- [yes ] I have signed all commits
- [ yes] I have added unit tests for the new functionality
- [ yes] I have run lint and test steps successfully


wasup-yash (approved):

plugin/storage/es/mappings/fieldType.go
```diff
@@ -0,0 +1,59 @@
...
+		case "object", "true":
+			return ObjectFieldType
+		default:
+			return NestedFieldType
+		}
+	}
+	return NestedFieldType
+}
+
+func (field FieldType) Format(f fmt.State, verb rune) {
```
yurishkuro, line 41: which interface does this implement?  Please add a check, e.g.
```
var _ ExpectedInterface = (*MyType)(nil)
```

plugin/storage/es/mappings/fieldType.go
```diff
@@ -0,0 +1,59 @@
...
+package mappings
+
+import (
+	"fmt"
+	"strconv"
+)
+
+type FieldType int
+
+func ParseFieldType(v any) FieldType {
```
yurishkuro, line 24: I don't understand the purpose of this function, please document it. It seems like you're using it by passing a Boolean value of the CLI option, which doesn't make sense to me - it's fine to declare enums in this module, but the mapping from CLI flag to the mapping type should be done close to where the CLI flag is defined. I also think that instead of pushing down DisableLogsFieldSearch flag all the way down to storage, you could've instead declared `LogsFieldsMappingType` parameter in the storage, and map boolean CLI flag to one of the values.

plugin/storage/es/mappings/fieldType.go
```diff
@@ -0,0 +1,59 @@
...
+	default:
+		_, _ = f.Write([]byte(strconv.Itoa(int(field))))
+
+	}
+}
+
+const (
+	NestedFieldType FieldType = iota
+	ObjectFieldType
+)
```
yurishkuro, line 59: these should go to the top, next to type declaration

plugin/storage/es/mappings/fieldType.go
```diff
@@ -0,0 +1,59 @@
...
+// limitations under the License.
+
+package mappings
+
+import (
+	"fmt"
+	"strconv"
+)
+
+type FieldType int
```
yurishkuro, line 22: why do this as `int`, instead of `string` with the expected values of `"object"` and `"nested"`?

yurishkuro:
```suggestion
type FieldMappingType string
const (
	FieldMappingObject = "object"
	FieldMappingNested = "nested"
)
```

plugin/storage/es/mappings/fieldType.go
```diff
@@ -0,0 +1,59 @@
+// Copyright (c) 2023 The Jaeger Authors.
```
yurishkuro, line 1: rename file to `mapping_type.go`, "field type" does not match ES nomenclature

plugin/storage/es/mappings/fixtures/jaeger-dependencies-7.json
```diff
@@ -1,17 +1,10 @@
 {
   "index_patterns": "*jaeger-dependencies-*",
-  "aliases": {
-    "test-jaeger-dependencies-read" : {}
-  },
```
yurishkuro, line 5: why is this changing in this PR?

plugin/storage/es/mappings/fixtures/jaeger-dependencies-with-ilm-7.json
```diff
@@ -0,0 +1,17 @@
+{
```
yurishkuro, line 1: please explain motivation for adding this. How is your PR related to ILM ?

plugin/storage/es/mappings/fixtures/jaeger-service-with-ilm-7.json
```diff
@@ -0,0 +1,48 @@
+{
```
yurishkuro, line 1: how different is it from `-7`? I would rather use some template / substitution, not duplicate a pretty large file

bhiravabhatla:
@yurishkuro  thanks for the review. I will spend sometime over weekend and revert on the comments.

pc-coder:
@yurishkuro @wasup-yash We have refactored the code and made the requested changes. Can you please take a look?

PR #4726, opened 2023-09-05
Allow disabling brearer token override from request in metrics store

pavolloffay:
Relates to https://github.com/grafana/tempo-operator/issues/526


plugin/metrics/prometheus/options.go
```diff
@@ -88,6 +89,8 @@ func (opt *Options) AddFlags(flagSet *flag.FlagSet) {
 		"The period to wait for a connection to Prometheus when executing queries.")
 	flagSet.String(nsConfig.namespace+suffixTokenFilePath, defaultTokenFilePath,
 		"The path to a file containing the bearer token which will be included when executing queries against the Prometheus API.")
+	flagSet.Bool(nsConfig.namespace+suffixOverrideFromContext, true,
+		"Whether the bearer token should be overridden from context (incoming request)")
```
yurishkuro, line 93: the phrasing is a bit confusing to me. In terms of functionality, it seems a priority list would be a more elegant solution that covers more use cases, e.g., e.g. `--token-priority=context,file` or `--token-priority=file` (the latter will achieve what this PR needs I think)

pavolloffay: The priority is not supported by our `bearertoken` package https://github.com/jaegertracing/jaeger/blob/4aa436bc53fa8868a3c278ec6cbe244ac1e9f7f7/pkg/bearertoken/transport.go#L42 

I don't need the priority. I want to for the transport to always use the token regardless of what is in the context.

Issue #4736, opened 2023-09-09
[deprecate] Remove support for OTEL spanmetrics processor


yurishkuro:
Support for spanmetrics connector was added in v1.46.0 (#4452, June) and made to be the default in v1.49.0 (#4704, Sep).

* [x] Add deprecation warning to the current flag that allows choosing between connector/processor -- #4741
* [x] Per our 3-moonths/2-releases deprecation policy, remove support for that flag and the corresponding code. -- #5539
* [x] Update README and website to avoid referencing the old flag or the processor -- https://github.com/jaegertracing/documentation/pull/709

cc @albertteoh

varshith257:
@yurishkuro Can I ahead to make updates in docs repo?

yurishkuro:
Yes

PR #4741, opened 2023-09-10
[SPM] Deprecate support for Spanmetrics Processor naming convention

yurishkuro:
- Part 1 of #4736

## Description of the changes
- Add deprecation notice to CLI flag
- Add log message when using span metrics processor notation

```
$ go run ./cmd/all-in-one help
      --prometheus.query.support-spanmetrics-connector            (deprecated, will be removed after 2024-01-01 or in release v1.53.0, whichever is later)  Controls whether the metrics queries should match the OpenTelemetry Collector's spanmetrics connector naming (when true) or spanmetrics processor naming (when false). (default true)

$ PROMETHEUS_QUERY_SUPPORT_SPANMETRICS_CONNECTOR=false METRICS_STORAGE_TYPE=prometheus go run ./cmd/all-in-one
2023/09/09 20:29:19 using Spanmetrics Processor's metrics naming conventions (deprecated, will be removed after 2024-01-01 or in release v1.53.0, whichever is later)
```

cc @albertteoh

### DIFF
--- a/plugin/metrics/prometheus/options.go
+++ b/plugin/metrics/prometheus/options.go
@@ -17,6 +17,7 @@ package prometheus
 import (
 	"flag"
 	"fmt"
+	"log"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


albertteoh (approved):
Thanks!

Issue #4743, opened 2023-09-10
[test] Flaky test TestPasswordFromFile in plugin/storage/es/factory_test.go


yurishkuro:
https://github.com/jaegertracing/jaeger/actions/runs/6134973238/job/16648278200?pr=4742#step:7:4523

One strange thing I noticed in the output that the very last logged request to the server was with Basic Auth string looking like the first password, not the second password, as if the request with the second password was made but then someone again made a request with the old one. I wonder if the es.Client object can continue sending pings to the server if it's not Close-ed (which we don't do).

MeenuyD:
Hello @yurishkuro I would like to work on this issue can you please guide me on how to get start

yurishkuro:
The test is here: https://github.com/jaegertracing/jaeger/blob/c8bcf2814aa0a91516dc9871d23b078428460513/plugin/storage/es/factory_test.go#L264

yurishkuro:
not solved, the #4751 seems like the right fix, but not finished.

yurishkuro:
Still failing https://github.com/jaegertracing/jaeger/actions/runs/6606707454/job/17943197226

Still seeing `POST /_bulk` with 1st password coming in after the client has already been swapped and the HEAD pings were going in with 2nd password. The 2nd `POST /_bulk` comes with the correct password.

Several options to fix this:
  * in the fake server ignore HEAD requests, only capture password from the POST requests. This should ensure that  the code blocks until the 2nd POST request comes. However, if there is an inherent race condition in the es client where the bulk processor's Close function is not flushing batch synchronously, then we may still observe issues, i.e. the first POST request could actually come last.
  * another option is to make the fake server remember all passwords it sees, not just the last one. This will make the test more resilient to race conditions.

yurishkuro:
this test still keeps failing, maybe not functionally but it leaks goroutine https://github.com/jaegertracing/jaeger/actions/runs/8216894810/job/22471977947#step:7:210

yurishkuro:
Most likely the same issue as #5083

madugushivakumar:
Hi @yurishkuro 👋  
I’d like to work on this issue.

I’ll investigate the goroutine leak and the relationship with #5083,
and try to stabilize TestPasswordFromFile by ensuring proper ES client
and bulk processor shutdown.

Please let me know if this approach sounds good.

Br1an67:
[...]

PR #4751, opened 2023-09-12
Fix: Ensure Proper Handling of Password Changes

MeenuyD:
It was observed that the last logged request to the server used a Basic Auth string that resembled the old password, even after a new password was set. This could potentially happen if the es.Client object continued sending pings to the server without being properly closed.
This commit addresses the issue by ensuring that the es.Client object is correctly closed when needed, preventing any further requests with outdated passwords.
#4743

### DIFF
--- a/plugin/storage/es/factory.go
+++ b/plugin/storage/es/factory.go
@@ -312,8 +312,12 @@ func (f *Factory) onClientPasswordChange(cfg *config.Configuration, client *atom
 	if err != nil {
 		f.logger.Error("failed to recreate Elasticsearch client with new password", zap.Error(err))
 	} else {
-		client.Store(&primaryClient)
+		oldClient := client.Swap(&newClient)
+        if oldClient != nil {
+            oldClient.Close()
+        }
 	}
+    
 }
 
 func loadTokenFromFile(path string) (string, error) {


plugin/storage/es/factory.go
```diff
@@ -312,8 +312,12 @@ func (f *Factory) onClientPasswordChange(cfg *config.Configuration, client *atom
 	if err != nil {
 		f.logger.Error("failed to recreate Elasticsearch client with new password", zap.Error(err))
 	} else {
-		client.Store(&primaryClient)
+		oldClient := client.Swap(&newClient)
```
yurishkuro, line 315: yep, exactly what I had in mind. It would be good to also add this to factory.Close(), i.e. swap both primary and archive clients for nil and close them

MeenuyD: should I change this to primary client?
oldClient := client.Swap(primaryClient)
then close the old client

yurishkuro:
please ensure commits are signed off

plugin/storage/es/factory.go
yurishkuro: no, I have no problem with this code. One improvement you could make is rename primaryClient->newClient.

I was referring to adding Close of clients to Factory.Close() method. Right now we leave them hanging.

yurishkuro:
@MeenuyD can you make sure that the commits are signed? Easiest is to squash - see https://github.com/jaegertracing/jaeger/blob/main/CONTRIBUTING_GUIDELINES.md#missing-sign-offs

Issue #3500, opened 2022-01-28
Jaeger v2 based on OpenTelemetry collector


pavolloffay:
## Update (Sept 2023)

Superseded by #4843

---

## Proposal (2022)

Creating top-level issue for Jaeger v2 based on OpenTelemetry collector. 

This has already been discussed a couple of times:
* roadmap https://www.jaegertracing.io/roadmap/#Integration%20with%20OpenTelemetry%20collector 
* two prototypes https://github.com/jaegertracing/jaeger/tree/v1.22.0/cmd/opentelemetry and https://github.com/jaegertracing/jaeger-otelcol. Docs https://www.jaegertracing.io/docs/1.20/opentelemetry/
* Many issue from: https://github.com/jaegertracing/jaeger/issues?q=is%3Aissue+opentelemetry+is%3Aclosed

I would like to bring the topic back as I think it is vital for the future success of Jaeger and staying relevant. Since we worked on v2 a couple of things changed, and therefore, I have created a new proposal on how v2 could be designed https://docs.google.com/document/d/1d7j956tDVYacKHF-l0JL9sVhpbYkt0JECDyGeup-olc/edit?usp=sharing and what is the impact on the ecosystem. 

I would like to open this for discussion.  If you are Jaeger user please upvote the issue if you would like the proposal including the breaking changes and impact on deployment helm/operator.

eugeniyk:
Direction wise makes total sense as jaeger pipeline looks rudimentary, however I believe there are some features that are not yet avail in otel collectors

What's left to Jaeger then? custom protocol as receiver (which should go away at some point), UI as extension and custom storage integration? Means just coupling querier with particular exporters?

pavolloffay:
Jaeger will provide the storage layer and UI. 

Once we are based on the OTEL collector it will be easier to write extensions and provide additional functionality.

yurishkuro:
> What's left to Jaeger then?

@eugeniyk Think of this not from the architecture point of view but from the user impact - *everything* about Jaeger is still there, plus some additional capabilities we could inherit from OTEL (the most critical one - OTLP receiver). Jaeger is an end-to-end tracing platform, OTEL collector is not, just a piece of it.

jkowall:
So, what about the data model which is used on storage, and how about the terminology in Jaeger would evolve to align with Otel versus OT? I don't think we need to make those changes, but they would help for the future. Unfortunatley they don't really help users so much, but are more for maintainers and the project.

[...]

PR #4766, opened 2023-09-16
Add jaeger-v2 single binary based on OTEL collector

yurishkuro:
- Third prototype of "Jaeger-v2"
- Another alternative approach to #3500

## Description of the changes
- Adds a new binary `jaeger-v2` using OTEL Collector framework
- Minimal amount of extensions is included, to mimic what `jaeger-collector` normally has
- It will combine all previous functions of agent/collector/query in one binary, but controllable via config file

```
$ go run -tags=ui ./cmd/jaeger-v2 --config ./cmd/jaeger-v2/config.yaml
```

## Roadmap

https://docs.google.com/document/d/1s4_6VgAS7qAVp6iEm5KYvpiGw3h2Ja5T5HpKo29iv00/edit

## Design

* the ingestion and storing of traces will be done via standard receivers/processors/exporters OTEL Collector components
* the jaeger-query and UI are implemented as `jaeger_query` extension (already working in this PR)

### Storage
In order to keep the flexibility of mixing & matching storage implementations, all backends can be configured via `jaeger_storage` extension (we may need to add `jaeger_metrics_storage` extension in the future). It might look like this:
```yaml
jaeger_storage:
  memory:  # defines Factory
    memstore:
      max_traces: 100000
  cassandra:
    cassandra_primary:
      servers: [...]
      namespace: jaeger
    cassandra_archive:
      servers: [...]
      namespace: jaeger_archive
```

The `jaeger_query` extension then references specific storage factories by name:
```yaml
  jaeger_query:
    trace_storage: memstore
    dependencies: something_else
    metrics_store: prometheus_store
```

It's not clear yet if `jaeger_query` extension should simply subsume `jaeger_storage` extension, because Query is the only one that needs this _generic_ access to storage, while things like exporters or Kafka ingester (receiver) always deal with a single implementation (because OTEL Coll pipeline allows to connect them with each other, which is not possible with extensions).

## Trade-offs
- This not using OTEL Collector builder `ocb`. That means people won't be able to assemble a different version of the collector with other extensions.
- We may want to support `ocb` in the future, as it makes it easier to write custom in-process exporters for custom storage. It will require converting all the components into their own modules.

## Next steps

* [x] Get feedback from the community on the approach
* [x] Fully implement all-in-one by wiring receivers / exporters correctly

## Open Questions
* How can we implement all-in-one equivalent that can be run without any config file?
* Do we want [healthcheckextension](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/extension/healthcheckextension/README.md) to be included by default?
* Investigate startup error `2023-09-23T19:55:46.661-0400	warn	zapgrpc/zapgrpc.go:195	[core] [Channel #2 SubChannel #3] grpc: addrConn.createTransport failed to connect to {Addr: ":16685", ServerName: "localhost:16685", }. Err: connection error: desc = "transport: Error while dialing: dial tcp :16685: connect: connection refused"	{"grpc_log": true}`


pavolloffay:
I am not sure what benefits this approach will give us from the functionality standpoint. Also the `otel` mode is significantly different and our existing ecosystem is not compatible with it (operator, helm).

Personally, I would prefer maintaining ecosystem that is closer to OTEL (collector builder) and have an opinionated OTELcol build with Jaeger components.

yurishkuro:
@pavolloffay yeah, I tend to agree. Especially the binary size increase seems unreasonable to me.

The alternative approach, still similar to this one, is to define another binary, but still with a hardcoded list of extensions, like in this PR, without using `ocb` builder.

Do you have some thoughts on why we would want to use `ocb` for assembling the binary? If we organize all the extensions similar to OTEL contrib, then someone in theory could still use `ocb` for a custom build of Jaeger, but it's not clear to me what benefits this approach provides for our official binary.

cmd/collector/internal/otelcmd/components.go
```diff
@@ -0,0 +1,96 @@
...
+
+	factories.Processors, err = processor.MakeFactoryMap(
+		// standard
+		batchprocessor.NewFactory(),
+		memorylimiterprocessor.NewFactory(),
+		// add-ons
+		attributesprocessor.NewFactory(),
+		resourcedetectionprocessor.NewFactory(),
+		resourceprocessor.NewFactory(),
+		spanmetricsprocessor.NewFactory(),
```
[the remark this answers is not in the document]
yurishkuro: done

cmd/jaeger-v2/config.yaml
```diff
@@ -0,0 +1,45 @@
+service:
+  extensions: [jaeger_storage, jaeger_query]
+  pipelines:
+    traces:
+      receivers: [otlp, jaeger, zipkin]
```
afzal442, line 5: Just one question, why we need `zipkin` receiver since it's been deprecated.

yurishkuro: Who says it was deprecated? There's plenty of zipkin instrumentation out there.

afzal442: Sorry I thought it was deprecated against OTEL instrumentation.

cmd/jaeger-v2/internal/components.go
```diff
@@ -0,0 +1,98 @@
...
+
+	factories.Exporters, err = exporter.MakeFactoryMap(
+		// standard
+		loggingexporter.NewFactory(),
+		otlpexporter.NewFactory(),
+		otlphttpexporter.NewFactory(),
+		// add-ons
+		storageexporter.NewFactory(), // generic exporter to Jaeger v1 spanstore.SpanWriter
+		kafkaexporter.NewFactory(),
+		// elasticsearch.NewFactory(),
```
albertteoh, line 70: Why exclude the elasticsearch exporter? Is it because storageexporter should already cover the ability to write to elasticsearch using jaeger's native elasticsearch span writer?

plugin/storage/memory/factory.go
```diff
@@ -44,6 +45,18 @@ func NewFactory() *Factory {
 	return &Factory{}
 }
 
+// NewFactoryWithConfig is used from jaeger-v2
```
albertteoh, line 48:
```suggestion
// NewFactoryWithConfig is used from jaeger-v2.
```

albertteoh (approved):
👍🏼 Amazing! Keen to see how this all fits together with the SPM feature, would be awesome to not require running a separate OTEL collector for the setup.

cmd/jaeger-v2/internal/components.go
yurishkuro: oh, because there isn't one, it's just the code I copied from Pavol's repo where es exporter module was defined, but was not implemented.

Initially yes, the generic support for existing storage backends is done via `storageexporter`, which converts OTLP into Jaeger model and sends to SpanWriter one span at a time. As we build OTLP-native exporters people can migrate to those.

PR #4772, opened 2023-09-22
Add e2e test for sampling storage

slayer321:
- Related #3305

## Description of the changes
-  Adding e2e test for Implemented Storage backends for adaptive sampling i.e (cassendra, in-memory)

- 

## Checklist

### DIFF
--- a/plugin/storage/integration/cassandra_test.go
+++ b/plugin/storage/integration/cassandra_test.go
@@ -86,6 +86,10 @@ func (s *CassandraStorageIntegration) initializeCassandra() error {
 	if s.SpanReader, err = f.CreateSpanReader(); err != nil {
 		return err
 	}
+	if s.SamplingStore, err = f.CreateSamplingStore(0); err != nil {
+		return err
+	}
+
 	if err = s.initializeDependencyReaderAndWriter(f); err != nil {
 		return err
 	}
--- a/plugin/storage/integration/integration.go
+++ b/plugin/storage/integration/integration.go
@@ -32,8 +32,10 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
+	samplemodel "github.com/jaegertracing/jaeger/cmd/collector/app/sampling/model"
 	"github.com/jaegertracing/jaeger/model"
 	"github.com/jaegertracing/jaeger/storage/dependencystore"
+	"github.com/jaegertracing/jaeger/storage/samplingstore"
 	"github.com/jaegertracing/jaeger/storage/spanstore"
 )
 
@@ -50,6 +52,7 @@ type StorageIntegration struct {
 	SpanReader       spanstore.Reader
 	DependencyWriter dependencystore.Writer
 	DependencyReader dependencystore.Reader
+	SamplingStore    samplingstore.Store
 	Fixtures         []*QueryFixtures
 
 	// TODO: remove this after all storage backends return spanKind from GetOperations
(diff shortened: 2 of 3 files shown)


plugin/storage/integration/integration.go
```diff
@@ -393,6 +396,59 @@ func (s *StorageIntegration) testGetDependencies(t *testing.T) {
...
+		actual, err := s.SamplingStore.GetThroughput(start, start.Add(time.Second*time.Duration(10)))
+		require.NoError(t, err)
+		fmt.Printf("actual: %v\n", actual)
+		fmt.Printf("expected: %v\n", expected)
+		return assert.ObjectsAreEqualValues(expected, len(actual))
+	})
+	if !assert.True(t, found) {
+		t.Log("\t Expected:", expected)
+		t.Log("\t Actual  :", actual)
+	}
```
yurishkuro, line 420:
```suggestion
	assert.Len(t, actual, expected)
```

plugin/storage/integration/cassandra_test.go
```diff
@@ -134,4 +138,5 @@ func TestCassandraStorage(t *testing.T) {
 	require.NoError(t, s1.initializeCassandra())
 	// TODO: Support all other tests.
 	t.Run("GetDependencies", s1.testCassandraGetDependencies)
+	s1.IntegrationTestAll(t)
```
yurishkuro, line 141: oh, wow, we never tested Cassandra?

slayer321: yes,`IntegrationTestAll` was never called for cassendra

plugin/storage/integration/integration.go
```diff
@@ -422,4 +477,10 @@ func (s *StorageIntegration) IntegrationTestAll(t *testing.T) {
 	t.Run("GetLargeSpans", s.testGetLargeSpan)
 	t.Run("FindTraces", s.testFindTraces)
 	t.Run("GetDependencies", s.testGetDependencies)
+	// Remove the below line check once we have sampling store implemented for all the backend
+	if os.Getenv("STORAGE") != "cassendra" || os.Getenv("STORAGE") != "in-memory" {
```
yurishkuro, line 481: this is not how we exclude tests. When a backend does not support sampling, it cannot populate `s.SamplingStore`, so the test can simply check for this being nil and call Skip with appropriate message.

plugin/storage/integration/integration.go
```diff
@@ -126,7 +130,8 @@ func (s *StorageIntegration) testGetServices(t *testing.T) {
 
 	var actual []string
 	found := s.waitForCondition(t, func(t *testing.T) bool {
-		actual, err := s.SpanReader.GetServices(context.Background())
+		var err error
+		actual, err = s.SpanReader.GetServices(context.Background())
```
yurishkuro, line 134: unnecessary change

plugin/storage/integration/memstore_test.go
```diff
@@ -36,8 +37,10 @@ func (s *MemStorageIntegrationTestSuite) initialize() error {
 	s.logger, _ = testutils.NewLogger()
 
 	store := memory.NewStore()
+	sampleStore := memory.NewSamplingStore(2)
 	s.SpanReader = store
 	s.SpanWriter = store
+	s.SamplingStore = sampleStore
```
yurishkuro, line 43: can combine L40 and 43 in one statement

plugin/storage/integration/memstore_test.go
```diff
@@ -57,5 +60,10 @@ func (s *MemStorageIntegrationTestSuite) cleanUp() error {
 func TestMemoryStorage(t *testing.T) {
 	s := &MemStorageIntegrationTestSuite{}
 	require.NoError(t, s.initialize())
+	// Remove the below line check once we have sampling store implemented for all the storage backend
+	err := os.Setenv("STORAGE", "in-memory")
+	if err != nil {
+		t.Fatal(err)
+	}
```
yurishkuro, line 67: not needed

plugin/storage/integration/integration.go
slayer321: adding the new variable named `err` is required because  if we don't do that then a new `actual` variable get's Initialized at L134 and if the `GetServices` method fails we get the empty slice for the actual value as you can see below.

```
    integration.go:139: 
        	Error Trace:	/home/infracloud/go/src/github.com/jaeger/plugin/storage/integration/integration.go:139
        	Error:      	Should be true
        	Test:       	TestMemoryStorage/GetServices
    integration.go:140: 	 Expected: [example-service-1 example-service-2 example-service-3123]
    integration.go:141: 	 Actual  : []

```

plugin/storage/integration/integration.go
```diff
@@ -422,4 +476,9 @@ func (s *StorageIntegration) IntegrationTestAll(t *testing.T) {
 	t.Run("GetLargeSpans", s.testGetLargeSpan)
 	t.Run("FindTraces", s.testFindTraces)
 	t.Run("GetDependencies", s.testGetDependencies)
+	if s.SamplingStore == nil {
```
yurishkuro, line 479: this should be done inside the individual tests that depend on sampling store, not in the top runner

slayer321: I also thought of that but then I realized that in all the individual tests we are calling `s.IntegrationTestAll(t)` which will run the integration test for them and we can skip sampling store test at the top runner level. 
If I try to skip it at the individual level I'm not sure how can we skip this two specific sampling store test because there we can check for the samplingStore is nil or not .. but we cannot specify that we want to skip this two test.

plugin/storage/integration/integration.go
```diff
@@ -414,6 +420,56 @@ func (s *StorageIntegration) testGetDependencies(t *testing.T) {
 	assert.EqualValues(t, expected, actual)
 }
 
+// === Sampling Store Integration Tests ===
+
+func (s *StorageIntegration) testGetThroughput(t *testing.T) {
+	s.skipIfNeeded(t)
```
yurishkuro, line 426:
```suggestion
	s.skipIfNeeded(t)
	if s.SamplingStore == nil {
	   // skip
	}
```

PR #4776, opened 2023-09-25
[tests] Fix failing e2e test  for Cassandra storage

### LINKED ISSUE #4782: [Bug]: Storage integration tests can fail queitly
https://github.com/jaegertracing/jaeger/actions/runs/6308158203/job/17125895911

The run is successful, but the logs show
image

* [ ] fix the CI scripts to fail the run if the unit tests fail (the error is probably ignored somewhere)
* [ ] disable Cassandra tests if necessary

slayer321:

Resolves #4782
## Description of the changes
-  Recently a [PR](https://github.com/jaegertracing/jaeger/pull/4773) was raised to run all integration tests against Cassandra while checking it's [CI for cassandra](https://github.com/jaegertracing/jaeger/actions/runs/6287799686/job/17072619668) I found that the some test where failing but CI passed. 
-  Fixed the script so that if the test fail we get exit code as non-zero.
-  Fixed  `TestCassandraStorage/GetLargeSpans` test which was failing due to older traces not getting cleanup from db.

- run the cassandra script to test it.

## Checklist

### DIFF
--- a/plugin/storage/cassandra/factory.go
+++ b/plugin/storage/cassandra/factory.go
@@ -218,6 +218,7 @@ func (f *Factory) Close() error {
 	return f.Options.GetPrimary().TLS.Close()
 }
 
-func (f *Factory) CassendraSession() cassandra.Session {
+// PrimarySession is used from integration tests to clean database between tests
+func (f *Factory) PrimarySession() cassandra.Session {
 	return f.primarySession
 }
--- a/plugin/storage/cassandra/factory.go
+++ b/plugin/storage/cassandra/factory.go
@@ -62,7 +62,7 @@ type Factory struct {
 	tracer                trace.TracerProvider
 
 	primaryConfig  config.SessionBuilder
-	PrimarySession cassandra.Session
+	primarySession cassandra.Session
 	archiveConfig  config.SessionBuilder
 	archiveSession cassandra.Session
 }
@@ -104,11 +104,11 @@ func (f *Factory) Initialize(metricsFactory metrics.Factory, logger *zap.Logger)
 	f.archiveMetricsFactory = metricsFactory.Namespace(metrics.NSOptions{Name: "cassandra-archive", Tags: nil})
 	f.logger = logger
 
-	PrimarySession, err := f.primaryConfig.NewSession(logger)
+	primarySession, err := f.primaryConfig.NewSession(logger)
 	if err != nil {
 		return err
 	}
-	f.PrimarySession = PrimarySession
+	f.primarySession = primarySession
 
 	if f.archiveConfig != nil {
 		if archiveSession, err := f.archiveConfig.NewSession(logger); err == nil {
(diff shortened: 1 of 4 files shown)


plugin/storage/cassandra/factory.go
```diff
@@ -62,7 +62,7 @@ type Factory struct {
 	tracer                trace.TracerProvider
 
 	primaryConfig  config.SessionBuilder
-	primarySession cassandra.Session
+	PrimarySession cassandra.Session
```
yurishkuro, line 65: add a method that returns it, instead of changing field visibility

plugin/storage/integration/cassandra_test.go
```diff
@@ -77,6 +87,9 @@ func (s *CassandraStorageIntegration) initializeCassandra() error {
 	if s.SpanReader, err = f.CreateSpanReader(); err != nil {
 		return err
 	}
+	if s.CleanUp, err = s.cleanUp(f.PrimarySession); err != nil {
```
yurishkuro, line 90: you can assign s.CleanUp function at suite construction time and it will be called automatically before each test.

scripts/cassandra-integration-test.sh
```diff
@@ -51,7 +51,7 @@ apply_schema() {
 run_integration_test() {
   local version=$1
   local schema_version=$2
-  local cid=$(setup_cassandra ${version})
+  cid=$(setup_cassandra ${version})
```
yurishkuro, line 54: why this change?

slayer321: as in this script at the end we want to remove the cassandra container and we are using trap for that with `EXIT` signal, and inside the trap we are passing `cid` as the argument to the `teardown_cassandra` function. Earlier when we are running this script it was giving below error 
```
scripts/cassandra-integration-test.sh: line 1: cid: unbound variable
```
as the trap is listing for the `EXIT` signal `teardown_cassandra` function only executes once the script completed and by that time we don't have the `cid` variable in the scope as it was Initialized with local keyword which are bound to theat functions only. 
Below is the exam script I wrote to show the trap working with `EXIT` signal
```
sachin@SACHIN-MAURYA:~/tp/script-test $ cat trap-test.sh 
#!/bin/bash

first(){
echo "this is first function"
}

last(){
echo "Last function call"
}

second(){
echo "this is second function"
first
trap 'last' EXIT
}

main(){
echo "starting with main"
second
echo "ending with main"
}

main
```

Output
```
sachin@SACHIN-MAURYA:~/tp/script-test $ bash trap-test.sh 
starting with main
this is second function
this is first function
ending with main
Last function call
```

yurishkuro: disagree with your analysis, but you're right that there is a bug. The bug is that (a) the trap instruction is at the end of the function and (b) it's using non-interpolated string ''. I think the trap instruction should follow immediately after `local cid=` is assigned, and it should use interpolated string `"teardown_cassandra ${cid}"` which will be resolved at the time when `cid` is in scope.

slayer321: yes, after testing it I found just having `interpolated string` works there is no need to changes the position of the trap call.

plugin/storage/integration/cassandra_test.go
slayer321: I didn't get it properly can you please explain it in more details
Do you mean to assign CleanUp at `newCassandraStorageIntegration` function level ? I think as we required to pass the `f.CassendraSession()`  function to cleanup function we cannot setup at suite level i.e `newCassandraStorageIntegration` function.

plugin/storage/cassandra/factory.go
```diff
@@ -217,3 +217,7 @@ func (f *Factory) Close() error {
 	}
 	return f.Options.GetPrimary().TLS.Close()
 }
+
+func (f *Factory) CassendraSession() cassandra.Session {
```
yurishkuro, line 221:
```suggestion
func (f *Factory) PrimarySession() cassandra.Session {
```

yurishkuro (approved):

PR #4783, opened 2023-09-26
Use non-root user in built containers

### LINKED ISSUE #4781: [Bug]: Image should be created with a non-root user
### What happened?

We ran a twistlock scan of:
```
jaegertracing/jaeger-ingester  1.48.0
jaegertracing/jaeger-collector 1.48.0
jaegertracing/jaeger-query 1.48.0
```
And it returned that the image was built with "Image should be created with a non-root user".

### Steps to reproduce

Twistlock scan of the images above will return the results.

### Expected behavior

Image is created with non-root user instead of root user.

### Relevant log output

_No response_

### Screenshot

_No response_

### Additional context

_No response_

### Jaeger backend version

_No response_

### SDK

_No response_

### Pipeline

_No response_

### Stogage backend

_No response_

### Operating system

_No response_

### Deployment model

_No response_

### Deployment configs

_No response_

nikzayn:
Resolves #4781

## Description of the changes
- I have added the non-root user in all images so that a twist lock scan could happen

- By running the jaeger make test commands and also check using docker.

## Checklist
- [-] I have read https://github.com/jaegertracing/jaeger/blob/master/CONTRIBUTING_GUIDELINES.md
- [-] I have signed all commits
- [-] I have run lint and test steps successfully


go.mod
```diff
@@ -43,15 +43,15 @@ require (
 	github.com/stretchr/testify v1.8.4
 	github.com/uber/jaeger-client-go v2.30.0+incompatible
 	github.com/xdg-go/scram v1.1.2
-	go.opentelemetry.io/collector/component v0.85.0
```
yurishkuro, line 46: this is not part of the PR

yurishkuro (approved):
thanks

PR #4796, opened 2023-10-02
Fix batchProcessor to set correct span format flags

k0zl:

Recently I've noticed that collector sends wrong tags "format" and "transport" in metric `jaeger_collector_spans_received_total`. 

Root cause: when batchProcessor was extended to be called from OTLP Receiver, it was given SpanOptions parameter that could specify a different span format OTLP, vs. always setting Proto. But the processor was still using hardcoded Proto format when invoking the next processor.

Added unit test

## Checklist
- [ x ] I have read https://github.com/jaegertracing/jaeger/blob/master/CONTRIBUTING_GUIDELINES.md
- [ x ] I have signed all commits
- [ - ] I have added unit tests for the new functionality (not needed)
- [ x ] I have run lint and test steps successfully


albertteoh (approved):
Thanks! Please [sign your commit](https://github.com/jaegertracing/jaeger/pull/4796/checks?check_run_id=17310162478).

k0zl:
> Thanks! Please [sign your commit](https://github.com/jaegertracing/jaeger/pull/4796/checks?check_run_id=17310162478).

Yep, already did

k0zl:
@albertteoh could you approve the required workflows please?

yurishkuro (changes requested):
please add unit test that illustrates the issue

k0zl:
Added unit test

yurishkuro:
Sorry, I am still missing the use case. Can you elaborate on the problem?
  * What is the setup? I.e. how are you receiving spans?
  * How are the metrics looking today and what you expect them to look?

yurishkuro (dismissed):

yurishkuro (approved):

yurishkuro:
Thanks!

PR #4811, opened 2023-10-06
Add product/file version in Windows executables

### LINKED ISSUE #3414: Product/File version in Windows executables

ResamVi:
This resolves #3414

## Description of the changes
Add to the windows build step in `Makefile` to create a `resource.syso`  using [goversioninfo](https://github.com/josephspurrier/goversioninfo).  The version is extracted from the latest git tag.

`go build` will embed this information in the final executable

image

(Forgive the German in this screenshot)

1. Run ` make build-binaries-windows` 
2. Navigate to the folder containing binaries (e.g. ` cmd/agent/agent-windows-amd64`)
3. Add `.exe` to the file name for the changes to be seen: `agent-windows-amd64.exe`

Ran other targets (`make build-binaries-linux`) to see if it broke other targets. Looked fine so far.

## Notes
There is the hassle of creating a `.syso` file that must be in the [same directory as the main.go file](https://github.com/golang/go/issues/16090#issuecomment-258956641) so there's a lot of copying to each `cmd/*`. 

I tried to keep it centralized to a `prepare-winres` Makefile target and the files are cleaned up afterwards.

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,6 +48,9 @@ GIT_CLOSEST_TAG=$(shell git describe --abbrev=0 --tags)
 ifneq ($(GIT_CLOSEST_TAG),$(shell echo ${GIT_CLOSEST_TAG} | grep -E "$(semver_regex)"))
 	$(warning GIT_CLOSEST_TAG=$(GIT_CLOSEST_TAG) is not in the semver format $(semver_regex))
 endif
+GIT_CLOSEST_TAG_MAJOR := $(shell echo $(GIT_CLOSEST_TAG) | sed -n 's/v\([0-9]*\)\.[0-9]*\.[0-9]/\1/p')
+GIT_CLOSEST_TAG_MINOR := $(shell echo $(GIT_CLOSEST_TAG) | sed -n 's/v[0-9]*\.\([0-9]*\)\.[0-9]/\1/p')
+GIT_CLOSEST_TAG_PATCH := $(shell echo $(GIT_CLOSEST_TAG) | sed -n 's/v[0-9]*\.[0-9]*\.\([0-9]\)/\1/p')
 DATE=$(shell TZ=UTC0 git show --quiet --date='format-local:%Y-%m-%dT%H:%M:%SZ' --format="%cd")
 BUILD_INFO_IMPORT_PATH=$(JAEGER_IMPORT_PATH)/pkg/version
 BUILD_INFO=-ldflags "-X $(BUILD_INFO_IMPORT_PATH).commitSHA=$(GIT_SHA) -X $(BUILD_INFO_IMPORT_PATH).latestVersion=$(GIT_CLOSEST_TAG) -X $(BUILD_INFO_IMPORT_PATH).date=$(DATE)"
@@ -73,6 +76,8 @@ DOCKER_NAMESPACE?=jaegertracing
 DOCKER_TAG?=latest
 
 MOCKERY=mockery
+GOVERSIONINFO=goversioninfo
+SYSOFILE=resource.syso
 
 .DEFAULT_GOAL := test-and-lint
 
@@ -178,27 +183,27 @@ build-examples:
 
 .PHONY: build-tracegen
 build-tracegen:
-	$(GOBUILD) $(BUILD_INFO) -o ./cmd/tracegen/tracegen-$(GOOS)-$(GOARCH) ./cmd/tracegen/main.go
+	$(GOBUILD) $(BUILD_INFO) -o ./cmd/tracegen/tracegen-$(GOOS)-$(GOARCH) ./cmd/tracegen/
 
 .PHONY: build-anonymizer
 build-anonymizer:
-	$(GOBUILD) $(BUILD_INFO) -o ./cmd/anonymizer/anonymizer-$(GOOS)-$(GOARCH) $(BUILD_INFO) ./cmd/anonymizer/main.go
+	$(GOBUILD) $(BUILD_INFO) -o ./cmd/anonymizer/anonymizer-$(GOOS)-$(GOARCH) $(BUILD_INFO) ./cmd/anonymizer/
 
 .PHONY: build-esmapping-generator
 build-esmapping-generator:
-	$(GOBUILD) -o ./plugin/storage/es/esmapping-generator-$(GOOS)-$(GOARCH) $(BUILD_INFO) ./cmd/esmapping-generator/main.go
+	$(GOBUILD) -o ./plugin/storage/es/esmapping-generator-$(GOOS)-$(GOARCH) $(BUILD_INFO) ./cmd/esmapping-generator/
 
 .PHONY: build-esmapping-generator-linux
 build-esmapping-generator-linux:
-	 GOOS=linux GOARCH=amd64 $(GOBUILD) -o ./plugin/storage/es/esmapping-generator $(BUILD_INFO) ./cmd/esmapping-generator/main.go
+	 GOOS=linux GOARCH=amd64 $(GOBUILD) -o ./plugin/storage/es/esmapping-generator $(BUILD_INFO) ./cmd/esmapping-generator/
 
 .PHONY: build-es-index-cleaner
 build-es-index-cleaner:
-	$(GOBUILD) -o ./cmd/es-index-cleaner/es-index-cleaner-$(GOOS)-$(GOARCH) ./cmd/es-index-cleaner/main.go
+	$(GOBUILD) -o ./cmd/es-index-cleaner/es-index-cleaner-$(GOOS)-$(GOARCH) ./cmd/es-index-cleaner/
 
 .PHONY: build-es-rollover
 build-es-rollover:
-	$(GOBUILD) -o ./cmd/es-rollover/es-rollover-$(GOOS)-$(GOARCH) ./cmd/es-rollover/main.go
+	$(GOBUILD) -o ./cmd/es-rollover/es-rollover-$(GOOS)-$(GOARCH) ./cmd/es-rollover/
 
 .PHONY: docker-hotrod
 docker-hotrod:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -44,6 +44,12 @@ IMPORT_LOG=.import.log
 
 GIT_SHA=$(shell git rev-parse HEAD)
 GIT_CLOSEST_TAG=$(shell git describe --abbrev=0 --tags)
+ifeq ($(GIT_CLOSEST_TAG), main)
+	GIT_CLOSEST_TAG=v0.0.0
```
yurishkuro, line 48: I think we should fail instead. I don't even know how this 'main' tag ended up being created, but it certainly messed up the release process. Having a default like v0.0.0 means that that process might have succeeded and created v0.0.0 release - not at all what we want.

Makefile
```diff
@@ -174,27 +182,27 @@ build-examples:
 
 .PHONY: build-tracegen
 build-tracegen:
-	$(GOBUILD) $(BUILD_INFO) -o ./cmd/tracegen/tracegen-$(GOOS)-$(GOARCH) ./cmd/tracegen/main.go
+	$(GOBUILD) $(BUILD_INFO) -o ./cmd/tracegen/tracegen-$(GOOS)-$(GOARCH) ./cmd/tracegen/
```
yurishkuro, line 185: was there a particular reason to change this to package?

versioninfo.json
```diff
@@ -0,0 +1,33 @@
...
+        "FileFlagsMask": "3f",
+        "FileFlags ": "00",
+        "FileOS": "040004",
+        "FileType": "01",
+        "FileSubType": "00"
+    },
+    "StringFileInfo": {
+        "FileDescription": "Jaeger, Distributed Tracing Platform.",
+        "FileVersion": "MAJOR.MINOR.PATCH.0",
+        "LegalCopyright": "2015-2019 The Jaeger Project Authors",
```
yurishkuro, line 24: this should use current year, not 2019

versioninfo.json
```diff
@@ -0,0 +1,33 @@
...
+        "ProductVersion": "MAJOR.MINOR.PATCH"
+    },
+    "VarFileInfo": {
+        "Translation": {
+            "LangID": "0409",
+            "CharsetID": "04B0"
```
yurishkuro, line 30: what are these magic numbers?

Makefile
```diff
@@ -403,6 +429,7 @@ install-tools:
 	$(GO) install github.com/vektra/mockery/v2@v2.14.0
 	$(GO) install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.52.1
 	$(GO) install mvdan.cc/gofumpt@latest
+	$(GO) install github.com/josephspurrier/goversioninfo/cmd/goversioninfo@latest
```
yurishkuro, line 432: this needs to be pinned to specific version, otherwise it will result in non-reproducible builds

versioninfo.json
```diff
@@ -0,0 +1,33 @@
...
+            "Build": 0
+        },
+        "FileFlagsMask": "3f",
+        "FileFlags ": "00",
+        "FileOS": "040004",
+        "FileType": "01",
+        "FileSubType": "00"
+    },
+    "StringFileInfo": {
+        "FileDescription": "Jaeger, Distributed Tracing Platform.",
```
yurishkuro, line 22: shouldn't this be different for each binary?

Makefile
```diff
@@ -229,39 +237,57 @@ build-all-in-one-debug build-agent-debug build-query-debug build-collector-debug
...
-	$(GOBUILD) $(DISABLE_OPTIMIZATIONS) -o ./cmd/remote-storage/remote-storage$(SUFFIX)-$(GOOS)-$(GOARCH) $(BUILD_INFO) ./cmd/remote-storage/main.go
+	$(GOBUILD) $(DISABLE_OPTIMIZATIONS) -o ./cmd/remote-storage/remote-storage$(SUFFIX)-$(GOOS)-$(GOARCH) $(BUILD_INFO) ./cmd/remote-storage/
 
 .PHONY: build-binaries-linux
 build-binaries-linux:
 	GOOS=linux GOARCH=amd64 $(MAKE) build-platform-binaries
 
+.PHONY: prepare-winres
+prepare-winres:
+	sed 's/MAJOR/$(GIT_CLOSEST_TAG_MAJOR)/;s/MINOR/$(GIT_CLOSEST_TAG_MINOR)/;s/PATCH/$(GIT_CLOSEST_TAG_PATCH)/' versioninfo.json \
```
yurishkuro, line 272: you have `s/a/b/;...` - do you need to end with `/g` so that all instances are replaced?

Makefile
yurishkuro: I'm actually having second thoughts about it. I was about to commit this:
```diff
+semver_regex := ^v\d+\.\d+\.\d+$$
 GIT_SHA=$(shell git rev-parse HEAD)
 GIT_CLOSEST_TAG=$(shell git describe --abbrev=0 --tags)
+ifneq ($(GIT_CLOSEST_TAG),$(shell echo ${GIT_CLOSEST_TAG} | grep -E "$(semver_regex)"))
+       $(error GIT_CLOSEST_TAG=$(GIT_CLOSEST_TAG) is not in the semver format $(semver_regex))
+endif
```
but this would break the whole Makefile in case there are some custom tags. I don't think the Makefile entry is the place for such check, it should be done only in places where having a valid semver is actually necessary.

Makefile
```diff
@@ -229,39 +234,98 @@ build-all-in-one-debug build-agent-debug build-query-debug build-collector-debug
...
+        }
+    }
+}
+endef
+
+export VERSIONINFO
+
+.PHONY: prepare-winres
+prepare-winres:
+	echo $$VERSIONINFO | sed "s/NAME/Jaeger Agent/g" | $(GOVERSIONINFO) -o="cmd/agent/$(SYSOFILE)" -
```
yurishkuro, line 314: I think you can use `define VERSIONINFO` as a function:
```
echo $(call VERSIONINFO,Jaeger Agent) | ...
```
and access the parameter as `$1`

You can also define another helper target to reduce boiler plate, e.g. reduce each command to 
```
  $(MAKE) prepare-winres-once NAME="Jaeger Agent" PATH="cmd/agent"
```
(or use another template / function for that)

Makefile
```diff
@@ -233,39 +238,98 @@ build-all-in-one-debug build-agent-debug build-query-debug build-collector-debug
...
+	echo $$VERSIONINFO | sed "s/NAME/Jaeger Tracegen/g" | $(GOVERSIONINFO) -o="cmd/tracegen/$(SYSOFILE)" -
+	echo $$VERSIONINFO | sed "s/NAME/Jaeger Anonymizer/g" | $(GOVERSIONINFO) -o="cmd/anonymizer/$(SYSOFILE)" -
+	echo $$VERSIONINFO | sed "s/NAME/Jaeger ES-Index-Cleaner/g" | $(GOVERSIONINFO) -o="cmd/es-index-cleaner/$(SYSOFILE)" -
+	echo $$VERSIONINFO | sed "s/NAME/Jaeger ES-Rollover/g" | $(GOVERSIONINFO) -o="cmd/es-rollover/$(SYSOFILE)" -
+
 .PHONY: build-binaries-windows
-build-binaries-windows:
+build-binaries-windows: prepare-winres
 	GOOS=windows GOARCH=amd64 $(MAKE) build-platform-binaries
+	find . -name $(SYSOFILE) -type f -delete
```
yurishkuro, line 332:
```suggestion
	find cmd/ -name $(SYSOFILE) -type f -delete
```

yurishkuro: or even better `rm ./cmd/*/$(SYSOFILE)`

Makefile
```diff
@@ -229,39 +234,101 @@ build-all-in-one-debug build-agent-debug build-query-debug build-collector-debug
...
+	NAME="Jaeger Agent"				PKGPATH="cmd/agent"				$(MAKE) prepare-winres-helper
+	NAME="Jaeger Collector" 		PKGPATH="cmd/collector" 		$(MAKE) prepare-winres-helper
+	NAME="Jaeger Query"				PKGPATH="cmd/query"				$(MAKE) prepare-winres-helper
+	NAME="Jaeger Ingester"			PKGPATH="cmd/ingester"			$(MAKE) prepare-winres-helper
+	NAME="Jaeger Remote Storage"	PKGPATH="cmd/remote-storage"	$(MAKE) prepare-winres-helper
+	NAME="Jaeger All-In-One"		PKGPATH="cmd/all-in-one"		$(MAKE) prepare-winres-helper
+	NAME="Jaeger Tracegen"			PKGPATH="cmd/tracegen"			$(MAKE) prepare-winres-helper
+	NAME="Jaeger Anonymizer"		PKGPATH="cmd/anonymizer"		$(MAKE) prepare-winres-helper
+	NAME="Jaeger ES-Index-Cleaner"	PKGPATH="cmd/es-index-cleaner"	$(MAKE) prepare-winres-helper
+	NAME="Jaeger ES-Rollover"		PKGPATH="cmd/es-rollover"		$(MAKE) prepare-winres-helper
```
yurishkuro, line 322: The formatting looks messed up, can you use spaces instead of tabs (except the first one which has to be a tab)
image

Makefile
```diff
@@ -229,39 +234,101 @@ build-all-in-one-debug build-agent-debug build-query-debug build-collector-debug
...
+            "Build": 0
+        },
+        "FileFlagsMask": "3f",
+        "FileFlags ": "00",
+        "FileOS": "040004",
+        "FileType": "01",
+        "FileSubType": "00"
+    },
+    "StringFileInfo": {
+        "FileDescription": "NAME",
```
yurishkuro, line 294: why not `$(NAME)` as with all other vars?

Makefile
```diff
@@ -229,39 +234,101 @@ build-all-in-one-debug build-agent-debug build-query-debug build-collector-debug
...
+        }
+    }
+}
+endef
+
+export VERSIONINFO
+
+.PHONY: prepare-winres
+prepare-winres:
+	NAME="Jaeger Agent"				PKGPATH="cmd/agent"				$(MAKE) prepare-winres-helper
```
yurishkuro, line 313: you don't need to pass these values as env vars, `make {target} x=1 y=2` works fine.

Makefile
```diff
@@ -229,39 +234,101 @@ build-all-in-one-debug build-agent-debug build-query-debug build-collector-debug
...
+	NAME="Jaeger Remote Storage"	PKGPATH="cmd/remote-storage"	$(MAKE) prepare-winres-helper
+	NAME="Jaeger All-In-One"		PKGPATH="cmd/all-in-one"		$(MAKE) prepare-winres-helper
+	NAME="Jaeger Tracegen"			PKGPATH="cmd/tracegen"			$(MAKE) prepare-winres-helper
+	NAME="Jaeger Anonymizer"		PKGPATH="cmd/anonymizer"		$(MAKE) prepare-winres-helper
+	NAME="Jaeger ES-Index-Cleaner"	PKGPATH="cmd/es-index-cleaner"	$(MAKE) prepare-winres-helper
+	NAME="Jaeger ES-Rollover"		PKGPATH="cmd/es-rollover"		$(MAKE) prepare-winres-helper
+
+.PHONY: prepare-winres-helper
+prepare-winres-helper:
+	echo $$VERSIONINFO | sed "s/NAME/$$NAME/g" | $(GOVERSIONINFO) -o="$$PKGPATH/$(SYSOFILE)" -
```
yurishkuro, line 326: Why not just use `$(NAME)` in the JSON string as you do with other vars? I prefer this to be consistent, one way.

yurishkuro (approved):

yurishkuro:
Thanks! 🎉

PR #4819, opened 2023-10-07
Fix build-all-in-one-image script

albertteoh:
## Description of the changes
- Fixes [this error in CI builds](https://github.com/jaegertracing/jaeger/actions/runs/6442084773/job/17492537692#step:19:744) and may explain why the jaeger all-in-one v1.50 image is [missing in docker hub](https://hub.docker.com/r/jaegertracing/all-in-one/tags?page=1&name=1.50):

```
+ '[main' '!=' 'pr-only]'
scripts/build-all-in-one-image.sh: line 57: [main: command not found
```

- Tested using a simple bash script to reproduce the problem.

## Checklist

### DIFF
--- a/scripts/build-all-in-one-image.sh
+++ b/scripts/build-all-in-one-image.sh
@@ -54,7 +54,7 @@ bash scripts/build-upload-a-docker-image.sh -b -c all-in-one -d cmd/all-in-one -
 
 
 #do not run debug image build when it is pr-only
-if ["$mode" != "pr-only"]; then
+if [ "$mode" != "pr-only" ]; then
   make build-all-in-one-debug GOOS=linux GOARCH=$GOARCH
   repo=${repo}-debug
   #build all-in-one-debug image locally for integration test


yurishkuro (approved):
usually double `[[]]` are used for expressions

yurishkuro:
if it's missing from docker hub, the workflow should've failed, let's make sure this happens

scripts/build-all-in-one-image.sh
```diff
@@ -54,7 +54,7 @@ bash scripts/build-upload-a-docker-image.sh -b -c all-in-one -d cmd/all-in-one -
 
 
 #do not run debug image build when it is pr-only
-if ["$mode" != "pr-only"]; then
+if [ "$mode" != "pr-only" ]; then
```
yurishkuro, line 57:
```suggestion
if [[ "$mode" != "pr-only" ]]; then
```

yurishkuro:
I think the reason there is no 1.50 image is because the build did not recognize semver tag and published this as `latest` only:
```
+ docker buildx build --output 'type=image, push=true' --progress=plain --target release --build-arg base_image=localhost:5000/baseimg_alpine:latest --build-arg debug_image=localhost:5000/debugimg_alpine:latest --platform=linux/amd64,linux/s390x,linux/ppc64le,linux/arm64 --file cmd/all-in-one/Dockerfile --tag docker.io/jaegertracing/all-in-one --tag docker.io/jaegertracing/all-in-one:latest --tag quay.io/jaegertracing/all-in-one --tag quay.io/jaegertracing/all-in-one:latest --tag docker.io/jaegertracing/all-in-one-snapshot:2bb7ee46f8f8ed9310515f4e68eca32afeae4589 --tag quay.io/jaegertracing/all-in-one-snapshot:2bb7ee46f8f8ed9310515f4e68eca32afeae4589 cmd/all-in-one
```

yurishkuro:
So the workflow was indeed successful. My other fix should solve that, hopefully.

albertteoh:
> if it's missing from docker hub, the workflow should've failed, let's make sure this happens

I was wondering why we didn't get a non-zero exit status on this failure even though we have `set -exu`.

From this [SO answer](https://unix.stackexchange.com/questions/650942/shell-script-returns-0-exit-status-despite-syntax-error):

> The exit status of the if command shall be the exit status of the then or else compound-list that was executed, or zero, if none was executed.

So since the `then` block wasn't executed, it just defaults to zero.

I'm not sure if there's anything reasonable we can do except making sure we use the correct bash syntax.

albertteoh:
> I'm not sure if there's anything reasonable we can do except making sure we use the correct bash syntax.

Perhaps we _could_ do something to prevent this sort of error like using https://github.com/koalaman/shellcheck (github action available: https://github.com/marketplace/actions/shellcheck). 

What do you think?

yurishkuro:
great idea -- https://github.com/jaegertracing/jaeger/issues/4822

PR #4826, opened 2023-10-08
Add CI validation of shell scripts using shellcheck

### LINKED ISSUE #4822: [Feature]: Implement CI validation of shell scripts using shellcheck
According to its docs, the [shellcheck tool](https://github.com/koalaman/shellcheck) already comes pre-installed in Github runners, so we could add a workflow that will validate all build scripts we use.

Currently running it against our scripts produces 83 warnings, so first we'd need to review them and fix the issues (or disable some lint checks), before we can enable the linter in CI.

```shell
$ shellcheck scripts/*sh | grep '^In ' | wc -l
      83
```

akagami-harsh:
Resolves #4822 

## Description of the changes
- fixed all 83 warnings produced by shellcheck
- added a workflow for shellcheck

- CI

## Checklist


yurishkuro (changes requested):
I reviewed the first few files, the same mistake is repeated over and over. When a variable is meant to represent multiple arguments like `x="a b c"`, using it in quotes `"${x}"` means all those values will be given to the command as a single argument, the same as passing `a\ b\ c`. This breaks the semantic of most of the scripts.

scripts/build-upload-a-docker-image.sh
```diff
@@ -62,11 +62,11 @@ else
...
 docker buildx build --output "${PUSHTAG}" \
-	--progress=plain ${target_arg} ${base_debug_img_arg}\
-	--platform=${platforms} \
-	--file ${docker_file_arg} \
-	${IMAGE_TAGS} \
-	${dir_arg}
+	--progress=plain "${target_arg}" "${base_debug_img_arg}"\
+	--platform="${platforms}" \
+	--file "${docker_file_arg}" \
+	"${IMAGE_TAGS}" \
```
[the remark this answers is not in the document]
akagami-harsh:
```
# shellcheck disable=SC2086
docker buildx build --output ${PUSHTAG} \
	--progress=plain ${target_arg} ${base_debug_img_arg}\
	--platform=${platforms} \
	--file ${docker_file_arg} \
	${IMAGE_TAGS} \
	${dir_arg}
```
i kept it as it is, and disabled warnings

akagami-harsh:
> I reviewed the first few files, the same mistake is repeated over and over. When a variable is meant to represent multiple arguments like `x="a b c"`, using it in quotes `"${x}"` means all those values will be given to the command as a single argument, the same as passing `a\ b\ c`. This breaks the semantic of most of the scripts.

@yurishkuro Thank you for taking the time to review my code and providing valuable feedback. I was so fixated on resolving those shellcheck warnings, that I forgot it was breaking the sole purpose of the code. I revisited all of it again and made some changes , can you review it again

scripts/cassandra-integration-test.sh
```diff
@@ -23,14 +23,16 @@ setup_cassandra() {
     --publish 9042:9042
     --publish 9160:9160
   )
-  local cid=$(docker run ${params[@]} ${image}:${tag})
-  echo ${cid}
+  local cid
+  # shellcheck disable=SC2068
+  cid=$(docker run ${params[@]} "${image}:${tag}")
```
yurishkuro, line 28: quoting `${params[@]}` should work. Try this script:
```shell
function pp {
  for i; do
    echo x=$i
  done
}

x=(1 2 3)
pp ${x[@]}
pp "${x[@]}"
```

scripts/cassandra-integration-test.sh
```diff
@@ -45,17 +47,19 @@ apply_schema() {
     --network host
   )
   docker build -t ${image} ${schema_dir}
+  # shellcheck disable=SC2068
   docker run ${params[@]} ${image}
```
yurishkuro, line 51: array is ok to quote

scripts/cassandra-integration-test.sh
```diff
@@ -45,17 +47,19 @@ apply_schema() {
...
   local version=$1
   local schema_version=$2
-  local cid=$(setup_cassandra ${version})
+  local cid
+  cid=$(setup_cassandra "${version}")
   apply_schema "$2"
   STORAGE=cassandra make storage-integration-test
   exit_status=$?
-  trap "teardown_cassandra ${cid}" EXIT
+  trap "teardown_cassandra \${cid}" EXIT
```
yurishkuro, line 62: our intention here was to evaluate `$cid` right away, so that the trap command becomes `teardown_cassandra some_id`. Using `\$` prevents expansion altogether:
```shell
$ x=123
$ echo "teardown ${x}"
teardown 123
$ echo "teardown \${x}"
teardown ${x}
```

scripts/check-go-version.sh
```diff
@@ -8,13 +8,17 @@ while getopts "uvd" opt; do
...
         u) update=true ;;
         v) verbose=true ;;
-        x) set -x ;;
         *) echo "Usage: $0 [-u] [-v] [-d]" >&2
            exit 1
            ;;
     esac
 done
 
+# Debugging option (-x) used separately
```
yurishkuro, line 17: I don't follow. The linter warning was `This case is not specified by getopts.`, so we just need to add it there
```
while getopts "uvdx" opt; do
```

scripts/update-semconv-version.sh
```diff
@@ -6,7 +6,7 @@ version_regex="v[0-9]\.[0-9]\+\.[0-9]\+"
 latest_semconv_version=$(
     curl -s https://pkg.go.dev/$package_name \
     | grep -oP 'data-id="v\d+\.\d+\.\d+"' \
-    | sed -E 's/\"($version_regex)\"/v\1/' \
+    | sed -E "s/\"($version_regex)\"/v\1/" \
```
yurishkuro, line 9: needs to be tested

scripts/package-deploy.sh
```diff
@@ -9,12 +9,12 @@ function stage-platform-files {
     local -r PACKAGE_STAGING_DIR=$2
     local -r FILE_EXTENSION=${3:-}
 
-    cp ./cmd/all-in-one/all-in-one-$PLATFORM  $PACKAGE_STAGING_DIR/jaeger-all-in-one$FILE_EXTENSION
-    cp ./cmd/agent/agent-$PLATFORM            $PACKAGE_STAGING_DIR/jaeger-agent$FILE_EXTENSION
-    cp ./cmd/query/query-$PLATFORM            $PACKAGE_STAGING_DIR/jaeger-query$FILE_EXTENSION
-    cp ./cmd/collector/collector-$PLATFORM    $PACKAGE_STAGING_DIR/jaeger-collector$FILE_EXTENSION
-    cp ./cmd/ingester/ingester-$PLATFORM      $PACKAGE_STAGING_DIR/jaeger-ingester$FILE_EXTENSION
-    cp ./examples/hotrod/hotrod-$PLATFORM     $PACKAGE_STAGING_DIR/example-hotrod$FILE_EXTENSION
+    cp ./cmd/all-in-one/all-in-one-"$PLATFORM"  "$PACKAGE_STAGING_DIR"/jaeger-all-in-one"$FILE_EXTENSION"
```
yurishkuro, line 12:
```suggestion
    cp "./cmd/all-in-one/all-in-one-${PLATFORM}"  "${PACKAGE_STAGING_DIR}/jaeger-all-in-one${FILE_EXTENSION}"
```

scripts/hotrod-integration-test.sh
```diff
@@ -13,9 +13,10 @@ make prepare-docker-buildx
 #build image locally for integration test
 bash scripts/build-upload-a-docker-image.sh -l -c example-hotrod -d examples/hotrod -p "${platforms}"
 
-export CID=$(docker run -d -p 8080:8080 localhost:5000/$REPO:${GITHUB_SHA})
+export CID
+CID=$(docker run -d -p 8080:8080 localhost:5000/$REPO:"${GITHUB_SHA}")
```
yurishkuro, line 17:
```suggestion
CID=$(docker run -d -p 8080:8080 "localhost:5000/${REPO}:${GITHUB_SHA}")
```

scripts/hotrod-integration-test.sh
```diff
@@ -13,9 +13,10 @@ make prepare-docker-buildx
...
 i=0
-while [[ "$(curl -s -o /dev/null -w ''%{http_code}'' localhost:8080)" != "200" && ${i} -lt 30 ]]; do
+while [[ "$(curl -s -o /dev/null -w "%{http_code}" localhost:8080)" != "200" && ${i} -lt 30 ]]; do
```
yurishkuro, line 19:
```
"$(curl -s -o /dev/null -w "
```
the last `"` breaks the string

scripts/generate-help-output.sh
```diff
@@ -14,7 +14,7 @@ function gen {
   shift
   for s in "$@"
   do
-    SPAN_STORAGE_TYPE=$s go run ./cmd/$bin help > $dir/$bin-$s.txt
+    SPAN_STORAGE_TYPE=$s go run ./cmd/"$bin" help > "$dir"/"$bin"-"$s".txt
```
yurishkuro, line 17: best to quote string only once

scripts/es-integration-test.sh
```diff
@@ -43,8 +45,10 @@ setup_opensearch() {
     --env "transport.host=127.0.0.1"
     --env "plugins.security.disabled=true"
   )
-  local cid=$(docker run ${params[@]} ${image}:${tag})
-  echo ${cid}
+  local cid
+  # shellcheck disable=SC2068
+  cid=$(docker run ${params[@]} "${image}:${tag}")
```
yurishkuro, line 50: can be quoted without disabling, i think

scripts/check-test-files.sh
```diff
@@ -7,14 +7,14 @@ COLOR_FIXME=$(printf "\033[31mFIXME\033[0m")
...
+  mainFile=$(find "${dir}" -maxdepth 1 -name 'main.go')
+  testFiles=$(find "${dir}" -maxdepth 1 -name '*_test.go')
   if [ -z "${testFiles}" ]; then
     if [ -n "${mainFile}" ]; then
       continue # single main does not require tests
     fi
-    if [ -e ${dir}/.nocover ]; then
-      reason=$(cat ${dir}/.nocover)
+    if [ -e "${dir}"/.nocover ]; then
+      reason=$(cat "${dir}"/.nocover)
```
yurishkuro, line 17:
```suggestion
    if [ -e "${dir}/.nocover" ]; then
      reason=$(cat "${dir}/.nocover")
```

scripts/check-go-version.sh
```diff
@@ -68,16 +72,17 @@ function check() {
...
 
-check go.mod "^go\s\+$version_regex" $go_previous_version
+check go.mod "^go\s\+$version_regex" "$go_previous_version"
 
-check docker/Makefile "^.*golang:$version_regex" $go_latest_version
+check docker/Makefile "^.*golang:$version_regex" "$go_latest_version"
 
 gha_workflows=$(grep -rl go-version .github)
+# shellcheck disable=SC2068
 for gha_workflow in ${gha_workflows[@]}; do
```
yurishkuro, line 81: can be quoted

scripts/update-semconv-version.sh
akagami-harsh:
```bash
package_name="go.opentelemetry.io/otel/semconv"
version_regex="v[0-9]\.[0-9]\+\.[0-9]\+"

latest_semconv_version=$(
    curl -s https://pkg.go.dev/$package_name \
    | grep -oP 'data-id="v\d+\.\d+\.\d+"' \
    | sed -E "s/\"($version_regex)\"/v\1/" \
    | sort -Vr \
    | head -n 1 \
    | awk -F'"' '{print $2}'
)

echo "$latest_semconv_version"
```
`output> v1.21.0
`
- I tested it in my local environment, and it produces the desired output that is the latest version of semconv

scripts/hotrod-integration-test.sh
akagami-harsh: That was my mistake, you are right 
`
while [[ "$(curl -s -o /dev/null -w '%{http_code}' localhost:8080)" != "200" && ${i} -lt 30 ]]; do 
`
adding single quotes should work fine

scripts/cassandra-integration-test.sh
akagami-harsh: could not find a workaround for this, so I disabled it

yurishkuro: Does this work?
```
teardown_cmd="teardown_cassandra ${cid}"
trap $teardown_cmd EXIT
```

akagami-harsh:
can you please add the Hacktoberfest label

scripts/cassandra-integration-test.sh
akagami-harsh: - No, it will produce this warning 
```bash
In scripts/cassandra-integration-test.sh line 61:
  trap "$teardown_cmd" EXIT
        ^-----------^ SC2064 (warning): Use single quotes, otherwise this expands now rather than when signalled.
```

- I think, disabling it, is the approach for it, if we want `$cid` to expand right away

akagami-harsh:
@yurishkuro , i have made some changes, can you please review it.

scripts/build-crossdock.sh
```diff
@@ -17,7 +17,7 @@ if [[ "$BRANCH" == "main" ]]; then
   docker buildx build --push \
     --progress=plain \
     --platform=linux/amd64 \
-    ${IMAGE_TAGS} \
+    "${IMAGE_TAGS[@]}" \
```
yurishkuro, line 20: IMAGE_TAGS needs to be assembled as array in order to use `[@]` syntax. Currently it is assembled as a string. This change is the same as `"${IMAGE_TAGS}"`:
```shell
$ cat 1.sh
function pr {
  echo "arg1=$1"
  echo "arg2=$2"
  echo "arg3=$3"
  echo "arg4=$4"
}

x="x y z"
x="${x} w"
pr "${x[@]}"
```
```shell
$ bash 1.sh
arg1=x y z w
arg2=
arg3=
arg4=

File: scripts/compute-tags.sh [hunk 15]
```diff
@@ -9,6 +9,7 @@ BRANCH=${BRANCH:?'expecting BRANCH env var'}
 
 ## if we are on a release tag, let's extract the version number
 ## the other possible value, currently, is 'main' (or another branch name)
+# shellcheck disable=SC2086
```
yurishkuro, line 12: this looks pretty straightforward, why are you disabling?

scripts/compute-tags.sh
```diff
@@ -36,4 +37,5 @@ fi
 
 IMAGE_TAGS="${IMAGE_TAGS} --tag docker.io/${SNAPSHOT_TAG} --tag quay.io/${SNAPSHOT_TAG}"
 
+#shellcheck disable=SC2086
```
yurishkuro, line 40: `echo "${IMAGE_TAGS}"` - why wouldn't this work? We are not passing the string as argument to another script, we just print it, and it should work fine with spaces.

scripts/es-integration-test.sh
```diff
@@ -87,21 +89,21 @@ bring_up_storage() {
...
     fi
-    wait_for_storage ${distro} "http://localhost:9200" ${cid}
+    wait_for_storage "${distro}" "http://localhost:9200" "${cid}"
     if [ ${db_is_up} = "1" ]; then
       break
     fi
   done
   if [ ${db_is_up} = "1" ]; then
-    trap "teardown_storage ${cid}" EXIT
+    trap 'teardown_storage ${cid}' EXIT
```
yurishkuro, line 106: `cid` is local here, we cannot extrapolate it later.

scripts/generate-help-output.sh
```diff
@@ -14,7 +14,7 @@ function gen {
   shift
   for s in "$@"
   do
-    SPAN_STORAGE_TYPE=$s go run ./cmd/$bin help > $dir/$bin-$s.txt
+    SPAN_STORAGE_TYPE=$s go run ./cmd/"$bin" help > "$dir/$bin-$s".txt
```
yurishkuro, line 17:
```suggestion
    SPAN_STORAGE_TYPE=$s go run "./cmd/$bin" help > "$dir/$bin-$s.txt"
```

scripts/package-deploy.sh
```diff
@@ -73,7 +74,7 @@ package tar linux-arm64
 package tar linux-ppc64le
 
 # Create a checksum file for all non-checksum files in the deploy directory. Strips the leading 'deploy/' directory from filepaths. Sort by filename.
-find deploy \( ! -name '*sha256sum.txt' \) -type f -exec shasum -b -a 256 {} \; | sed -r 's#(\w+\s+\*?)deploy/(.*)#\1\2#' | sort -k2 | tee ./deploy/jaeger-$VERSION.sha256sum.txt
+find deploy \( ! -name '*sha256sum.txt' \) -type f -exec shasum -b -a 256 {} \; | sed -r 's#(\w+\s+\*?)deploy/(.*)#\1\2#' | sort -k2 | tee ./deploy/jaeger-"$VERSION".sha256sum.txt
```
yurishkuro, line 77:
```suggestion
find deploy \( ! -name '*sha256sum.txt' \) -type f -exec shasum -b -a 256 {} \; | sed -r 's#(\w+\s+\*?)deploy/(.*)#\1\2#' | sort -k2 | tee "./deploy/jaeger-${VERSION}.sha256sum.txt"
```

scripts/build-crossdock.sh
akagami-harsh: does this work?
```bash
x=("x" "y" "z")
x+=("w")
for z in "${x[@]}"; do
  echo "x=$z"
done 
```
```
output
x=x
x=y
x=z
x=w
```

scripts/es-integration-test.sh
akagami-harsh: should i disable it?

scripts/build-crossdock.sh
yurishkuro: yes

scripts/es-integration-test.sh
yurishkuro: yes

scripts/compute-tags.sh
```diff
@@ -36,4 +36,4 @@ fi
 
 IMAGE_TAGS="${IMAGE_TAGS} --tag docker.io/${SNAPSHOT_TAG} --tag quay.io/${SNAPSHOT_TAG}"
 
-echo ${IMAGE_TAGS}
+echo "${IMAGE_TAGS}"
```
yurishkuro, line 39:
```suggestion
echo "${IMAGE_TAGS[@]}"
```
otherwise it only prints the first item

scripts/build-upload-a-docker-image.sh
```diff
@@ -61,7 +61,8 @@ else
     fi
 fi
 
-docker buildx build --output "${PUSHTAG}" \
+# shellcheck disable=SC2086
+docker buildx build --output ${PUSHTAG} \
```
yurishkuro, line 65: this is breaking the CI. Wy did you remove the quotes?

yurishkuro (approved):
looking good! 🎉

yurishkuro:
Thanks!

PR #4829, opened 2023-10-10
Support Elasticsearch 8.x

### LINKED ISSUE #3571: Elasticsearch 8.x support

Elastic recently released Elasticsearch 8.0, by now they've released 8.1.

The ES index can't be created anymore, I think the legacy template api part is now completely deprecated:
```go

jaeger-collector    | {"level":"info","ts":1646824639.4679213,"caller":"config/config.go:205","msg":"Elasticsearch detected","version":8}
jaeger-collector    | {"level":"info","ts":1646824639.4697242,"caller":"zapgrpc/zapgrpc.go:129","msg":"Deprecation warning: 299 Elasticsearch-8.1.0-3700f7679f7d95e36da0b43762189bab189bc53a \"Legacy index templates are deprecated in favor of composable templates.\""}
jaeger-collector    | {"level":"fatal","ts":1646824639.4699457,"caller":"./main.go:84","msg":"Failed to create span writer","error":"elastic: Error 400 (Bad Request): unknown key [template] in the template  [type=parse_exception]","stacktrace":"main.main.func1\n\t./main.go:84\ngithub.com/spf13/cobra.(*Command).execute\n\tgithub.com/spf13/cobra@v1.3.0/command.go:856\ngithub.com/spf13/cobra.(*Command).ExecuteC\n\tgithub.com/spf13/cobra@v1.3.0/command.go:974\ngithub.com/spf13/cobra.(*Command).Execute\n\tgithub.com/spf13/cobra@v1.3.0/command.go:902\nmain.main\n\t./main.go:147\nruntime.main\n\truntime/proc.go:255"}
```
The https://www.elastic.co/guide/en/elasticsearch/reference/8.1/migrating-8.0.html lists now (somewhere in the middle)
```ini
The [create or update index template API](https://www.elastic.co/guide/en/elasticsearch/reference/8.1/indices-templates-v1.html)’s template parameter has been removed.

Details
In 6.0, we deprecated the template parameter in create or update index template requests in favor of using index_patterns. Support for the template parameter is now removed in 8.0.

Impact
Use the create or update index template API's index_patterns parameter. Requests that include the template parameter will return an error.
```
I think the code falls back to the unversioned mappings if it does not find a specific mapping 

So, in my opinion the code should not fall back to the [jaeger-span.json](https://github.com/jaegertracing/jaeger/blob/7872d1b07439c3f2d316065b1fd53e885b26a66f/plugin/storage/es/mappings/jaeger-span.json) mapping but instead use the 
[jaeger-span-7.json](https://github.com/jaegertracing/jaeger/blob/7872d1b07439c3f2d316065b1fd53e885b26a66f/plugin/storage/es/mappings/jaeger-span-7.json) for newer versions, e.g. use the -7 jsons for 8 etc.

e.g. change line 42 from == 7 to >=7
https://github.com/jaegertracing/jaeger/blob/7872d1b07439c3f2d316065b1fd53e885b26a66f/plugin/storage/es/mappings/mapping.go#L40-L46

## Workaround
Force the ES Version to 7 via --es.version=7

pmuls99:
- Resolves #3571 

## Description of the changes
- Added index templates for esv8
- Added a esv8 client to handle the put index template request that is not currently possible by olivere/elastic
- After these changes are merged , users no more need to use create.index-templates= false unless they want to add their custom index templates and force the es.version = 7


plugin/storage/es/factory.go
```diff
@@ -241,8 +241,15 @@ func createSpanWriter(
 	if err != nil {
 		return nil, err
 	}
+
+	elasticsearch8Client, err := config.NewElasticSearch8Client(cfg, logger)
```
yurishkuro, line 245: Should this be done conditionally on the desired version? Would this not fail if the actual backend is <= v7?

plugin/storage/es/spanstore/writer.go
```diff
@@ -107,15 +112,36 @@ func (s *SpanWriter) CreateTemplates(spanTemplate, serviceTemplate, indexPrefix
...
-	_, err := s.client().CreateTemplate(indexPrefix + "jaeger-span").Body(spanTemplate).Do(context.Background())
-	if err != nil {
-		return err
-	}
-	_, err = s.client().CreateTemplate(indexPrefix + "jaeger-service").Body(serviceTemplate).Do(context.Background())
-	if err != nil {
-		return err
+	esVersion := s.client().GetVersion()
+	if esVersion > 7 {
+		spanTemplateResponse, err := s.v8client.Indices.PutIndexTemplate(indexPrefix+"jaeger-span", strings.NewReader(spanTemplate))
```
yurishkuro, line 117: nit: extract these two branches into helper functions createTemplatesV7 and createTemplatesV8

pkg/es/config/config.go
```diff
@@ -29,6 +29,7 @@ import (
 	"sync"
 	"time"
 
+	elasticsearch8 "github.com/elastic/go-elasticsearch/v8"
```
jkowall, line 32: URL is invalid

pmuls99: Yes, it is probably because of the specific version that we have used here i.e. v8. github.com/elastic/go-elasticsearch is a valid url.

plugin/storage/es/factory.go
pmuls99: I am not sure why it would fail, because in the local testing , the templates do get added. If the actual backend is <=v7 there would be a client declaration and we would just not use it anywhere. We can do it conditionally but I dont see where it would fail if the backend in <=v7.

plugin/storage/es/spanstore/writer.go
pmuls99: Yes, missed that.

plugin/storage/es/factory.go
yurishkuro: I suppose it will come up in integration tests if it fails

yurishkuro:
in the unit tests the mocks are unhappy about unexpected GetVersion call

in the CIT, ES8 complains about `    --env "xpack.monitoring.enabled=false"` param passed from here: [scripts/es-integration-test.sh](https://github.com/jaegertracing/jaeger/blob/2769d97fa79fa06131be429ebf8ce6a338115514/scripts/es-integration-test.sh#L30)
```
Unknown setting [xpack.monitoring.enabled] did you mean any of [xpack.profiling.enabled, xpack.monitoring.templates.enabled]?
```
You will need to change this flag for v8

yurishkuro:
this PR https://github.com/jaegertracing/jaeger/pull/4512/files had an example how to deal with the deprecated CLI flag. But I think we should not just remove the flag, we should change it for v8 to whatever the replacement is.

pmuls99:
> this PR https://github.com/jaegertracing/jaeger/pull/4512/files had an example how to deal with the deprecated CLI flag. But I think we should not just remove the flag, we should change it for v8 to whatever the replacement is.

So basically this whole set of flags have been deprecated since the version 7.16(https://www.elastic.co/guide/en/elasticsearch/reference/current/monitoring-settings.html).

yurishkuro:
```
if (( major_version < 8 )); then
    params+=(--env "xpack.monitoring.enabled=false")
else
    params+=(--env "xpack.monitoring.collection.enabled=false")
fi
```

yurishkuro (approved):

yurishkuro (approved):

cmd/es-rollover/app/init/action.go
```diff
@@ -39,13 +39,16 @@ type Action struct {
...
-		Shards:          int64(c.Config.Shards),
-		Replicas:        int64(c.Config.Replicas),
-		IndexPrefix:     c.Config.IndexPrefix,
-		UseILM:          c.Config.UseILM,
-		ILMPolicyName:   c.Config.ILMPolicyName,
-		EsVersion:       version,
+		TemplateBuilder:              es.TextTemplateBuilder{},
+		PrioritySpanTemplate:         int64(c.Config.PrioritySpanTemplate),
+		PriorityServiceTemplate:      int64(c.Config.PriorityServiceTemplate),
+		PriorityDependenciesTemplate: int64(c.Config.PriorityDependenciesTemplate),
```
yurishkuro, line 45: please make sure PR description reflects all the changes, and explain why they are needed (especially the priority)

plugin/storage/integration/es_index_cleaner_test.go
```diff
@@ -116,24 +119,25 @@ func TestIndexCleaner(t *testing.T) {
...
 
-func runIndexCleanerTest(t *testing.T, client *elastic.Client, prefix string, expectedIndices, envVars []string) {
+func runIndexCleanerTest(t *testing.T, client *elastic.Client, v8Client *elasticsearch8.Client, prefix string, expectedIndices, envVars []string) {
 	// make sure ES is clean
 	_, err := client.DeleteIndex("*").Do(context.Background())
 	require.NoError(t, err)
-
+	err = cleanESIndexTemplates(t, client, v8Client, prefix)
+	require.NoError(t, err)
+	defer cleanESIndexTemplates(t, client, v8Client, prefix)
```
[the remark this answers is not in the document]
pmuls99: So , in L137 we call the function createAllIndices which adds the templates.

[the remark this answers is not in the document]
yurishkuro: you call "clean" in L134 and deferred in L136. Why is L136 needed if you're deleting templates at the start of the test anyway? Only one of those should be needed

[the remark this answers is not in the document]
pmuls99: So the first one is to keep a clean slate before starting the test, and the next one is for post test cleanup.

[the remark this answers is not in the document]
yurishkuro: don't you see that this is redundant? If each test cleans after itself (via defer call), the next test always starts with a clean slate, it doesn't need to pre-clean.

plugin/storage/integration/elasticsearch_test.go
```diff
@@ -284,3 +316,43 @@ func (s *StorageIntegration) testArchiveTrace(t *testing.T) {
...
+		_, err = v8Client.Indices.DeleteIndexTemplate(prefixWithSeparator + dependenciesTemplateName)
+		require.NoError(t, err)
+	} else {
+		_, err := client.IndexDeleteTemplate("*").Do(context.Background())
+		require.NoError(t, err)
+	}
+	return nil
+}
+
+func getESVersion(client *elastic.Client) (uint, error) {
```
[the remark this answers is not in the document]
pmuls99: The problem for using that is , cleanESIndexTemplates is also used in index-cleaner and index-rollover integration tests.

plugin/storage/integration/es_index_cleaner_test.go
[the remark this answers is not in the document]
pmuls99: Yes that is right. But I had seen the initial implementation and followed the same. Will take care.

yurishkuro:
run `make fmt`

yurishkuro:
Please squash commits and sign

pkg/es/client/index_client.go
```diff
@@ -249,6 +249,24 @@ func (i IndicesClient) CreateTemplate(template, name string) error {
 	return nil
 }
 
+// Create Template for ESV8
```
yurishkuro, line 252: as far as I can see, the only difference in this method is the endpoint `_index_template` instead of `_template` in the function above. I think this is not a good reason to have a full new function declared at the interface level. Can we not just pass the version to the IndicesClient and make the endpoint name conditional on that version?

yurishkuro (approved):

yurishkuro:
great job @pmuls99 !  🎉

yurishkuro:
within a week - cc @albertteoh

Issue #4534, opened 2023-06-17
[OTEL] Process info is missing from the spans


yurishkuro:
When query-service span is generated by Jaeger client, we get some Process information filled out:
<img width="536" alt="image" src="https://github.com/jaegertracing/jaeger/assets/3523016/342f63ca-9085-4780-a71f-c0e712b7b059">

However, when the spans are produced by OTEL SDK, Process is empty.

<img width="687" alt="image" src="https://github.com/jaegertracing/jaeger/assets/3523016/54d4db7a-7f74-443b-967f-b2408c7e4328">

We need to find out
  * why is the server information not being filled in - doesn't OTEL SDK capture it by default? Maybe we need to enable some  resource discovery module?
  * Some attributes produced by OTEL SDK, like `otel.library.name`, should not be in the span Tags, but in the process. Ideally this should be fixed upstream in OTEL, but we could also implement it internally as a `model/adjuster`.

Michealajit:
Hi Sir,
I'm new here can i work on this ?

Michealajit:
@yurishkuro  is it a good first issue ?

afzal442:
Hey Yuri, I just came over this. how do you think we can adjust Process to have `otel.library.name` and enable some resource discovery module? Is there any other tags adjusted the same way? How do you see the changes when take this into account? Thanks

yurishkuro:
> adjust Process to have otel.library.name

The simplest way would be to add a new adjuster (under `model/adjuster`) to Jaeger that would move certain attributes from span.tags to span.process.tags.

> enable some resource discovery module

That I don't know, it needs looking into OTEL SDK code.

NavinShrinivas:
Hey @yurishkuro, I'm looking to making my first contribution to jaeger...This issue seems like a good one, Can I be assigned this? I see it's been a while since this issue was last picked up.

yurishkuro:
Feel free to pick it up. We do not explicitly assign issues.

james-ryans:
Sorry... I've encountered the same issue with the previous PR contributor. I've tackled the resource discovery and tag adjuster. Should I remove my tag adjuster and only apply the resource discovery to respect the previous PR contributor?

NavinShrinivas:
Hey, I was having trouble even getting this off the ground...I think you should go ahead and merge your prs. I'll look into some easier issues to get started with

PR #4841, opened 2023-10-14
[WIP] feat: implement adjuster moving tags from span to process

dtsulik:
One possible solution for  #4534 

## Description of the changes
- added adjuster that moves set of tags from span to process

- added unit test for new adjuster
- tested by launching all-in-one and producing traces from test app with otel-go

## Checklist

### DIFF
--- a/model/adjuster/span_tag_to_process.go
+++ b/model/adjuster/span_tag_to_process.go
@@ -1,0 +1,52 @@
+// Copyright (c) 2019 The Jaeger Authors.
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package adjuster
+
+import (
+	"github.com/jaegertracing/jaeger/model"
+)
+
+var spanTagsToMove = map[string]struct{}{
+	"otel.library.name": {},
+}
+
+// SpanTagsToProcessAdjuster moves certain tags from span.tags to
+// span.process.tags that should be there. This should ideally be
+// fixed in upstream OTEL.
+func SpanTagsToProcessAdjuster() Adjuster {
+	return Func(func(trace *model.Trace) (*model.Trace, error) {
+		for _, span := range trace.Spans {
+			processTagsMap := make(map[string]struct{})
+			for _, check := range span.Process.Tags {
+				processTagsMap[check.Key] = struct{}{}
+			}
+
+			index := 0
+			for _, tag := range span.Tags {
+				if _, ok := spanTagsToMove[tag.Key]; ok {
+					if _, exists := processTagsMap[tag.Key]; !exists {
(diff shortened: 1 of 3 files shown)


dtsulik:
Hi @yurishkuro added simple adjuster for tags related to #4534 

It could also be extended to be more generic
```go
type TagOperation int

const (
	SPANTOPROC TagOperation = iota
	PROCTOSPAN
)

var tagOperationMap = map[string]TagOperation{
    "otel.library.name": SPANTOPROC,
}
```

If you think this is better idea I will update the logic. In meantime I will try to check the OTEL side.

model/adjuster/span_tag_to_process.go
```diff
@@ -0,0 +1,52 @@
...
+func SpanTagsToProcessAdjuster() Adjuster {
+	return Func(func(trace *model.Trace) (*model.Trace, error) {
+		for _, span := range trace.Spans {
+			processTagsMap := make(map[string]struct{})
+			for _, check := range span.Process.Tags {
+				processTagsMap[check.Key] = struct{}{}
+			}
+
+			index := 0
+			for _, tag := range span.Tags {
```
yurishkuro, line 38:
```suggestion
			for i, tag := range span.Tags {
```

model/adjuster/span_tag_to_process.go
```diff
@@ -0,0 +1,52 @@
...
+				if _, ok := spanTagsToMove[tag.Key]; ok {
+					if _, exists := processTagsMap[tag.Key]; !exists {
+						span.Process.Tags = append(span.Process.Tags, tag)
+						continue
+					}
+				}
+				span.Tags[index] = tag
```
yurishkuro, line 45:
```suggestion
       if i != index {
				span.Tags[index] = tag
       }
```
avoid unnecessary copying

model/adjuster/span_tag_to_process.go
```diff
@@ -0,0 +1,52 @@
+// Copyright (c) 2019 The Jaeger Authors.
+// Copyright (c) 2017 Uber Technologies, Inc.
```
yurishkuro, line 2:
```suggestion
// Copyright (c) 2023 The Jaeger Authors.
```

model/adjuster/span_tag_to_process.go
```diff
@@ -0,0 +1,52 @@
...
+// limitations under the License.
+
+package adjuster
+
+import (
+	"github.com/jaegertracing/jaeger/model"
+)
+
+var spanTagsToMove = map[string]struct{}{
+	"otel.library.name": {},
```
yurishkuro, line 23: Resource attributes are the primary targets here

dtsulik: I see them in the [spec](https://github.com/open-telemetry/semantic-conventions/blob/main/model/resource/process.yaml), should this be parsed at build time to populate above map or should I add some of the main culprits manually, taken from [here](https://github.com/open-telemetry/opentelemetry-go/blob/main/semconv/v1.21.0/resource.go) for example, since the map itself is a temporary solution to an external problem.

yurishkuro: Yes, the `process` semconv certainly belongs to Span.Process.

Taking a step back though, I think it's weird that we have to do this translation *after* OTLP data was already translated into Jaeger - I think it's that translation that should be fixed.

dtsulik: Will take a look [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/translator/jaeger/traces_to_jaegerproto.go).

This PR however is more of dirty/quick fix utility that could be adapted for more general use as suggested [above](https://github.com/jaegertracing/jaeger/pull/4841#issuecomment-1762819773).

dtsulik:
Closing this since  #4844 implemented this approach.

Issue #4799, opened 2023-10-04
[Feature]: (Re)Introduce changelog-related labels and require them for PRs


yurishkuro:
We already tried this in the past, but there was no enforcement and it didn't take.

* [x] Add a GitHub Action that will verify that a PR has a specific categorization label before it can be merged. We already have a bunch of those labels `changelog:***` labels
  * example of bot: https://github.com/asyncapi/website/pull/2170#issuecomment-1735255198
  * example: https://twitter.com/benasher44/status/1503460237331734528
* [x] Alter the release notes script to validate that all PRs have changelog labels, fail if some don't and print their URLs so that the maintainer can fix them manually. This should ignore dependabot PRs automatically.
* [x] Alter the release notes script to automatically group changes by the labels
* [ ] Apply similar changes to jaeger-ui repo

sanyamjain036:
I'm a beginner to github actions but I want to work on this issue. I am learning about github actions online. May you please guide how to learn and solve this issue?

Suraj-kumar00:
Hi @yurishkuro, How are you doing?
I saw no one is working on this issue so can you please assign it to me so that i can work on this issue?

yurishkuro:
https://github.com/jaegertracing/jaeger/blob/029e759d3087de9ef55f388c2f0a425b1274f549/CONTRIBUTING_GUIDELINES.md?plain=1#L35-L41

Suraj-kumar00:
> https://github.com/jaegertracing/jaeger/blob/029e759d3087de9ef55f388c2f0a425b1274f549/CONTRIBUTING_GUIDELINES.md?plain=1#L35-L41

Yeah thanks, Now I'm doing this!

kanha-gupta:
Hey @Suraj-kumar00 are you currently working on this ?

devidasjadhav:
@yurishkuro Can you please explain " to validate that all PRs have changelog labels" means ?
https://github.com/jaegertracing/jaeger/pull/4735/commits/a44a2452219f9ed246892a633bee3f0c6fa686dc

As this was the only PR with changelog. 
So I am definitely missing something here.
Please explain.

jkowall:
@devidasjadhav We are trying to ensure that each PR has a changelog label, which would allow us to generate release notes more easily and in an automated way. Currently, the release process generates the list, but requires some manual processing of the ones which are not tagged.

jkowall:
[...]

PR #4849, opened 2023-10-16
Feature: updating release-note script to validate that all PRs have changelog label 

akagami-harsh:
- #4799 
 - [x] Alter the release notes script to validate that all PRs have changelog labels, fail if some don't and print their URLs so that the maintainer can fix them manually. This should ignore dependabot PRs automatically.

### DIFF
--- a/scripts/release-notes.py
+++ b/scripts/release-notes.py
@@ -16,6 +16,7 @@ import os.path
 import urllib.parse
 from os.path import expanduser
 import sys
+from github import Github
 from urllib.request import (
     urlopen,
     Request
@@ -42,6 +43,13 @@ def num_commits_since_prev_tag(token, base_url):
 
     print(f"There are {num_commits} new commits since {prev_release_tag}")
     return num_commits
+
+
+def validate_changelog_labels(pr):
+    for label in pr.get_labels():
+        if label.name.lower().startswith('changelog:'):
+            return True
+    return False
 
 
 def main(token, repo, num_commits, exclude_dependabot):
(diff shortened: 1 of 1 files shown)


scripts/release-notes.py
```diff
@@ -99,6 +107,28 @@ def main(token, repo, num_commits, exclude_dependabot):
...
     if skipped_dependabot:
         print(f"\n(Skipped {skipped_dependabot} dependabot commit{'' if skipped_dependabot == 1 else 's'})")
+     
+    # Print PRs that don't have changelog label 
+    github = Github(token)
+    repository = github.get_repo(f'jaegertracing/{repo}')
+    changelog_prs = []
+    other_prs = []
+    
+    for pr in repository.get_pulls(state='closed'):
```
yurishkuro, line 117: this will load all historical PRs, but we only need those which are in the commit history since the last releast

scripts/release-notes.py
```diff
@@ -99,6 +107,28 @@ def main(token, repo, num_commits, exclude_dependabot):
...
+
+        if validate_changelog_labels(pr):
+            changelog_prs.append(pr)
+        else:
+            other_prs.append(pr)
+
+    if other_prs:
+        print('PRs without "changelog:" labels:')
+        for pr in other_prs:
+            print(f'{pr.title} - {pr.html_url}')
```
yurishkuro, line 129: tip: 
  * trim title to fixed length like 50chars
  * use whitespace padding so that URL column always starts from the same place

scripts/release-notes.py
```diff
@@ -16,6 +16,7 @@
 import urllib.parse
 from os.path import expanduser
 import sys
+from github import Github
```
yurishkuro, line 19: I think the nice feature of the existing script is that it runs without extra dependencies that need to be pip-installed

scripts/release-notes.py
yurishkuro: basically in L91 we already determine PRs

yurishkuro:
High-level comment: the idea in the ticket is not that the release script must validate PR labels, but that it uses those labels to automatically organize changelog entries into different sections

akagami-harsh:
#4850, did better than mine , closing this one

PR #4853, opened 2023-10-17
Visualize uninstrumented services in the dependency diagrams

### LINKED ISSUE #3804: [Feature]: Visualize uninstrumented services in the dependency diagrams

nidhey27:
- Closes #3804 

## Description of the changes
- Updated `GetDependencies` to identify and include client spans without corresponding server spans, ensuring a complete dependency graph.
- Implemented `updateServiceDependencyLinks` to streamline the update of service dependencies, reducing code duplication.
- Enabled the representation of uninstrumented services in the dependency graph by inferring names from client spans lacking server spans.

- Introduced TestInferredServiceDependency to specifically evaluate the functionality of inferred service dependencies in the scenario of uninstrumented services.
- Created a client span without a corresponding server span and asserted the correct formation of dependencies, ensuring the inclusion of inferred services in the dependency graph.
- Employed assertions to confirm the presence and accuracy of inferred service dependencies, ensuring they are adequately represented with the correct attributes and relationships.

## Checklist

### DIFF
--- a/examples/hotrod/services/customer/database.go
+++ b/examples/hotrod/services/customer/database.go
@@ -21,7 +21,6 @@ import (
 	"fmt"
 
 	"go.opentelemetry.io/otel/attribute"
-	semconv "go.opentelemetry.io/otel/semconv/v1.21.0"
 	"go.opentelemetry.io/otel/trace"
 	"go.uber.org/zap"
 
@@ -75,9 +74,8 @@ func newDatabase(tracer trace.Tracer, logger log.Factory) *database {
 func (d *database) Get(ctx context.Context, customerID int) (*Customer, error) {
 	d.logger.For(ctx).Info("Loading customer", zap.Int("customer_id", customerID))
 
-	ctx, span := d.tracer.Start(ctx, "SQL SELECT", trace.WithSpanKind(trace.SpanKindClient))
+	_, span := d.tracer.Start(ctx, "SQL SELECT")
 	span.SetAttributes(
-		semconv.PeerServiceKey.String("mysql"),
 		attribute.
 			Key("sql.query").
 			String(fmt.Sprintf("SELECT * FROM customer WHERE customer_id=%d", customerID)),
--- a/plugin/storage/memory/memory.go
+++ b/plugin/storage/memory/memory.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"sort"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/gogo/protobuf/proto"
@@ -30,6 +31,9 @@ import (
 	"github.com/jaegertracing/jaeger/pkg/tenancy"
 	"github.com/jaegertracing/jaeger/storage/spanstore"
 )
+
+// Global counter
+var counter uint64
 
 // Store is an in-memory store of traces
 type Store struct {
@@ -116,6 +120,43 @@ func (st *Store) GetDependencies(ctx context.Context, endTs time.Time, lookback 
 						deps[depKey] = &model.DependencyLink{
 							Parent:    parentSpan.Process.ServiceName,
 							Child:     s.Process.ServiceName,
+							CallCount: 1,
+						}
+					} else {
+						deps[depKey].CallCount++
+					}
+				} else if isClientSpan(s) && !hasCorrespondingServerSpan(s, trace.Spans) {
+
+					// Generate a unique ID based on current time and a counter
+					currentTime := uint64(time.Now().UnixNano())
+					uniqueID := (currentTime << 32) | (atomic.AddUint64(&counter, 1) & 0xFFFFFFFF)
+
+					// Creating an artificial server span
+					artificialSpan := &model.Span{
+						TraceID:       s.TraceID,                                         // Keeping the same trace ID
+						SpanID:        model.NewSpanID(uniqueID),                         // Generating a new span ID
+						OperationName: "artificial-" + s.OperationName,                   // TODO
+						StartTime:     s.StartTime,                                       // Starting time can be the same as the client span
+						Duration:      s.Duration,                                        // Duration can also be copied or set as needed
+						Flags:         s.Flags,                                           // Flags can be copied
+						Process:       &model.Process{ServiceName: "artificial-service"}, // TODO: Set service name as needed
+						Tags: model.KeyValues{
+							model.String("span.kind", "server"), // This is an artificial server span
+							model.String("artificial", "true"),  // Tag to identify it as artificial
+						},
+					}
+
+					// Adding the artificial span to the trace
+					trace.Spans = append(trace.Spans, artificialSpan)
+
+					// Handle dependencies if needed, for example, increment the call count between services
+					parentService := s.Process.ServiceName
+					childService := "artificial-" + s.OperationName // TODO: another method to name the uninstrumented service
+					depKey := parentService + "&&&" + childService
+					if _, ok := deps[depKey]; !ok {
+						deps[depKey] = &model.DependencyLink{
+							Parent:    parentService,
+							Child:     childService,
 							CallCount: 1,
 						}
 					} else {
@@ -338,4 +379,22 @@ func flattenTags(span *model.Span) model.KeyValues {
 		retMe = append(retMe, l.Fields...)
 	}
 	return retMe
+}
+
+func hasCorrespondingServerSpan(clientSpan *model.Span, spans []*model.Span) bool {
+	for _, span := range spans {
+		if span.ParentSpanID() == clientSpan.SpanID {
+			return true
+		}
+	}
+	return false
+}
+
+func isClientSpan(span *model.Span) bool {
+	for _, tag := range span.Tags {
+		if tag.Key == "span.kind" && tag.VStr == "client" {
+			return true
+		}
+	}
+	return false
 }
--- a/plugin/storage/memory/memory.go
+++ b/plugin/storage/memory/memory.go
@@ -114,7 +114,8 @@ func (st *Store) GetDependencies(ctx context.Context, endTs time.Time, lookback 
 					updateServiceDependencyLinks(parentSpan.Process.ServiceName, s.Process.ServiceName, deps)
 				} else if isClientSpan(s) && !hasCorrespondingServerSpan(s, trace.Spans) {
 					parentService := s.Process.ServiceName
-					childService := "inferred-" + s.OperationName // TBD: call a function here to infer the name
+					childService := inferServiceName(s)
+					// childService := "inferred-" + s.OperationName // TBD: call a function here to infer the name
 					updateServiceDependencyLinks(parentService, childService, deps)
 				}
 			}
@@ -364,4 +365,48 @@ func updateServiceDependencyLinks(parentService, childService string, deps map[s
 	} else {
 		deps[depKey].CallCount++
 	}
+}
+
+func inferServiceName(span *model.Span) string {
+    serviceName := "inferred"
+
+    // Check for peer.service tag
+    if peerService, found := getTagValue(span, "peer.service"); found {
+        return serviceName + "-" + peerService
+    }
+
+    // Check for RPC service name
+    if rpcService, found := getTagValue(span, "rpc.service"); found {
+        return serviceName + "-rpc-" + rpcService
+    }
+
+    // Check for HTTP service name via route or URL
+    if httpRoute, found := getTagValue(span, "http.route"); found {
+        return serviceName + "-http-" + httpRoute
+    } else if httpURL, found := getTagValue(span, "http.url"); found {
+        return serviceName + "-http-" + httpURL 
+    }
+
+    // Check for Database service name via DB name or connection string
+    if dbSystem, found := getTagValue(span, "db.system"); found {
+        dbName := dbSystem
+        if dbName == "" {
+            dbName = "unknown"
+        }
+        if dbStatement, found := getTagValue(span, "db.statement"); found {
+            dbName += "-" + dbStatement 
+        }
+        return serviceName + "-db-" + dbName
+    }
+
+    return serviceName + "-" + span.OperationName
+}
+
+func getTagValue(span *model.Span, key string) (string, bool) {
+    for _, tag := range span.Tags {
+        if tag.Key == key {
+            return tag.VStr, true
+        }
+    }
+    return "", false
 }
--- a/plugin/storage/memory/memory.go
+++ b/plugin/storage/memory/memory.go
@@ -107,23 +107,16 @@ func (st *Store) GetDependencies(ctx context.Context, endTs time.Time, lookback 
 		if traceIsBetweenStartAndEnd(startTs, endTs, trace) {
 			for _, s := range trace.Spans {
 				parentSpan := findSpan(trace, s.ParentSpanID())
-				if parentSpan != nil {
-					if parentSpan.Process.ServiceName == s.Process.ServiceName {
-						continue
-					}
+				if parentSpan != nil && parentSpan.Process.ServiceName != s.Process.ServiceName {
 					updateServiceDependencyLinks(parentSpan.Process.ServiceName, s.Process.ServiceName, deps)
-				} else if isLeaf(s, trace.Spans) {
-					serviceName := s.Process.ServiceName
-					var childService string
-
-					if isClientSpan(s) {
-						childService = inferServiceName(s, !isLeaf(s, trace.Spans))
-					} else {
-						// childService = "inferred-leaf-" + s.OperationName
-						childService = inferServiceName(s, isLeaf(s, trace.Spans))
-					}
-
-					updateServiceDependencyLinks(serviceName, childService, deps)
+				}
+				// Handle client spans separately
+				if isClientSpan(s) {
+					childService := inferServiceName(s, false)
+					updateServiceDependencyLinks(s.Process.ServiceName, childService, deps)
+				} else if isLeaf(s, trace.Spans) { // Handle non-client leaf spans
+					childService := inferServiceName(s, true)
+					updateServiceDependencyLinks(s.Process.ServiceName, childService, deps)
 				}
 			}
 		}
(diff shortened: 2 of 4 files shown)


plugin/storage/memory/memory.go
```diff
@@ -121,6 +125,43 @@ func (st *Store) GetDependencies(ctx context.Context, endTs time.Time, lookback
...
 						deps[depKey].CallCount++
 					}
+				} else if isClientSpan(s) && !hasCorrespondingServerSpan(s, trace.Spans) {
+
+					// Generate a unique ID based on current time and a counter
+					currentTime := uint64(time.Now().UnixNano())
+					uniqueID := (currentTime << 32) | (atomic.AddUint64(&counter, 1) & 0xFFFFFFFF)
+
+					// Creating an artificial server span
+					artificialSpan := &model.Span{
```
yurishkuro, line 135: why do you need to create artificial span? All this method needs to do is to return dependency links, which you already construct in L157

plugin/storage/memory/memory.go
```diff
@@ -121,6 +125,43 @@ func (st *Store) GetDependencies(ctx context.Context, endTs time.Time, lookback
...
+					depKey := parentService + "&&&" + childService
+					if _, ok := deps[depKey]; !ok {
+						deps[depKey] = &model.DependencyLink{
+							Parent:    parentService,
+							Child:     childService,
+							CallCount: 1,
+						}
+					} else {
+						deps[depKey].CallCount++
+					}
```
yurishkuro, line 164: this is duplication, you can move that code into anonymous function and reuse it

plugin/storage/memory/memory.go
```diff
@@ -121,6 +125,43 @@ func (st *Store) GetDependencies(ctx context.Context, endTs time.Time, lookback
...
+							model.String("artificial", "true"),  // Tag to identify it as artificial
+						},
+					}
+
+					// Adding the artificial span to the trace
+					trace.Spans = append(trace.Spans, artificialSpan)
+
+					// Handle dependencies if needed, for example, increment the call count between services
+					parentService := s.Process.ServiceName
+					childService := "artificial-" + s.OperationName // TODO: another method to name the uninstrumented service
```
yurishkuro, line 154:
```suggestion
					childService := "inferred-" + s.OperationName // TODO: another method to name the uninstrumented service
```

plugin/storage/memory/memory.go
nidhey27: I've updated the code to focus on updating dependency links directly, eliminating the need for artificial spans.

plugin/storage/memory/memory.go
nidhey27: Refactored to use `updateServiceDependencyLinks`, removing code duplication.

yurishkuro (commented):
👍

plugin/storage/memory/memory.go
```diff
@@ -339,3 +335,78 @@ func flattenTags(span *model.Span) model.KeyValues {
...
+        return serviceName + "-http-" + httpURL 
+    }
+
+    // Check for Database service name via DB name or connection string
+    if dbSystem, found := getTagValue(span, "db.system"); found {
+        dbName := dbSystem
+        if dbName == "" {
+            dbName = "unknown"
+        }
+        if dbStatement, found := getTagValue(span, "db.statement"); found {
```
yurishkuro, line 396: db.statement has no relation to service name

plugin/storage/memory/memory.go
```diff
@@ -339,3 +335,78 @@ func flattenTags(span *model.Span) model.KeyValues {
...
+
+    // Check for RPC service name
+    if rpcService, found := getTagValue(span, "rpc.service"); found {
+        return serviceName + "-rpc-" + rpcService
+    }
+
+    // Check for HTTP service name via route or URL
+    if httpRoute, found := getTagValue(span, "http.route"); found {
+        return serviceName + "-http-" + httpRoute
+    } else if httpURL, found := getTagValue(span, "http.url"); found {
```
yurishkuro, line 386: URL should not be used as service name

plugin/storage/memory/memory.go
nidhey27: Addressed!

plugin/storage/memory/memory.go
nidhey27: Addressed!

plugin/storage/memory/memory.go
```diff
@@ -339,3 +335,73 @@ func flattenTags(span *model.Span) model.KeyValues {
 	}
 	return retMe
 }
+
+func hasCorrespondingServerSpan(clientSpan *model.Span, spans []*model.Span) bool {
```
yurishkuro, line 339:
```suggestion
func isLeaf(clientSpan *model.Span, spans []*model.Span) bool {
```
isn't "leaf" what you're looking for?

plugin/storage/memory/memory.go
```diff
@@ -111,16 +111,12 @@ func (st *Store) GetDependencies(ctx context.Context, endTs time.Time, lookback
...
-						deps[depKey] = &model.DependencyLink{
-							Parent:    parentSpan.Process.ServiceName,
-							Child:     s.Process.ServiceName,
-							CallCount: 1,
-						}
-					} else {
-						deps[depKey].CallCount++
-					}
+					updateServiceDependencyLinks(parentSpan.Process.ServiceName, s.Process.ServiceName, deps)
+				} else if isClientSpan(s) && !hasCorrespondingServerSpan(s, trace.Spans) {
```
yurishkuro, line 115: why is `else if` appropriate here? A leaf span will have a valid parent so the above branch will be executed, and not this one.

plugin/storage/memory/memory.go
```diff
@@ -339,3 +335,73 @@ func flattenTags(span *model.Span) model.KeyValues {
...
+	for _, span := range spans {
+		if span.ParentSpanID() == clientSpan.SpanID {
+			return true
+		}
+	}
+	return false
+}
+
+func isClientSpan(span *model.Span) bool {
+	for _, tag := range span.Tags {
```
yurishkuro, line 349: iirc `span.Tags.FindByKey()` function exists

plugin/storage/memory/memory.go
```diff
@@ -339,3 +335,73 @@ func flattenTags(span *model.Span) model.KeyValues {
...
+			dbName = "unknown"
+		}
+		return serviceName + "-db-" + dbName
+	}
+
+	return serviceName + "-" + span.OperationName
+}
+
+func getTagValue(span *model.Span, key string) (string, bool) {
+	for _, tag := range span.Tags {
```
yurishkuro, line 401: `span.Tags.FindByKey(key)`

plugin/storage/memory/memory_test.go
```diff
@@ -523,3 +523,92 @@ func makeTestingSpan(traceID model.TraceID, suffix string) *model.Span {
...
 }
+
+// TestInferredServiceDependency tests if the service dependencies include an inferred service
+// when there is a client span without a corresponding server span.
+func TestInferredServiceDependency(t *testing.T) {
+	// Create a memory store
+	store := NewStore()
+
+	// Create a client span without a corresponding server span
+	clientSpan := &model.Span{
```
yurishkuro, line 534: this is rather simplistic fixture. I'd say you also want to test:
  * span.kind=client which is NOT a leaf (has a child)
  * leaf span that is NOT span.kind=client

plugin/storage/memory/memory.go
nidhey27: Yes, I have updated the function name.

plugin/storage/memory/memory.go
nidhey27: Updated the code block to use `span.Tags.FindByKey()`

plugin/storage/memory/memory.go
nidhey27: Addressed!

plugin/storage/memory/memory_test.go
nidhey27: Modified the test to check above two scenarios as well.

plugin/storage/memory/memory.go
```diff
@@ -111,16 +111,19 @@ func (st *Store) GetDependencies(ctx context.Context, endTs time.Time, lookback
...
 					}
-					depKey := parentSpan.Process.ServiceName + "&&&" + s.Process.ServiceName
-					if _, ok := deps[depKey]; !ok {
-						deps[depKey] = &model.DependencyLink{
-							Parent:    parentSpan.Process.ServiceName,
-							Child:     s.Process.ServiceName,
-							CallCount: 1,
-						}
+					updateServiceDependencyLinks(parentSpan.Process.ServiceName, s.Process.ServiceName, deps)
+				} else if isLeaf(s, trace.Spans) {
```
yurishkuro, line 115: reposting previous question - why is `else if` appropriate here? A leaf span will have a valid parent so the above branch will be executed, and not this one.

nidhey27: After reviewing the code and your comments, I see the concern regarding leaf spans potentially having valid parents and therefore not being processed correctly in the `else if` structure.
I've made adjustments so that all relevant spans, including client and leaf spans, are handled correctly:
- revised logic begins by verifying the existence of a **parent span (parentSpan != nil)**. When a parent is present and belongs to a different service, a dependency link is established between the parent's service and the current span's service. This ensures that inter-service dependencies are captured regardless of the span type.
- I introduced a conditional check to specifically address client spans and leaf spans. This approach enables us to infer service dependencies correctly for leaf spans, even when they have a valid parent span. It's crucial for cases where a client span is not a leaf span, ensuring that these spans are still evaluated for potential service dependencies.
- for both **client** spans and **leaf** spans, the child service name is now inferred using a function, `inferServiceName`. This function takes into account whether the span is a leaf, allowing for accurate determination of the child service name in the dependency link.

The revised logic should fix the issue you pointed out. I appreciate your feedback and guidance in refining this.

plugin/storage/memory/memory.go
```diff
@@ -107,20 +107,16 @@ func (st *Store) GetDependencies(ctx context.Context, endTs time.Time, lookback
...
-						}
-					} else {
-						deps[depKey].CallCount++
-					}
+				if parentSpan != nil && parentSpan.Process.ServiceName != s.Process.ServiceName {
+					updateServiceDependencyLinks(parentSpan.Process.ServiceName, s.Process.ServiceName, deps)
+				}
+				// Handle client spans separately
+				if isClientSpan(s) {
+					childService := inferServiceName(s, false)
```
yurishkuro, line 115: I don't think this is needed. All you're adding is a use case where both `isClientSpan && isLeaf` are true.

nidhey27: I've removed the redundant check for cases where a span is both a client and a leaf, simplifying the logic.

plugin/storage/memory/memory.go
```diff
@@ -339,3 +331,68 @@ func flattenTags(span *model.Span) model.KeyValues {
...
+		}
+	} else {
+		deps[depKey].CallCount++
+	}
+}
+
+func inferServiceName(span *model.Span, isLeaf bool) string {
+	serviceName := "inferred"
+
+	if isLeaf {
```
yurishkuro, line 362: when would this be false?

plugin/storage/memory/memory.go
```diff
@@ -339,3 +331,68 @@ func flattenTags(span *model.Span) model.KeyValues {
...
+			Child:     childService,
+			CallCount: 1,
+		}
+	} else {
+		deps[depKey].CallCount++
+	}
+}
+
+func inferServiceName(span *model.Span, isLeaf bool) string {
+	serviceName := "inferred"
```
yurishkuro, line 360: this doesn't make a lot of sense in many cases. E.g. if you have `peer.service`, then why should you use `inferred-` prefix? If that `peer.service` value is actually consistent across the architecture (i.e. there may be proper spans from this service in other traces), then using the same name will reuse the same node, which is better representation of the service map than splitting into `interred-` + another

plugin/storage/memory/memory.go
```diff
@@ -339,3 +331,68 @@ func flattenTags(span *model.Span) model.KeyValues {
...
+	// Check for Database service name via DB name or connection string
+	if dbSystem, found := getTagValue(span, "db.system"); found {
+		dbName := dbSystem
+		if dbName == "" {
+			dbName = "unknown"
+		}
+		return serviceName + "-db-" + dbName
+	}
+
+	return serviceName + "-" + span.OperationName
```
yurishkuro, line 390:
```suggestion
	return "inferred::" + span.OperationName
```

plugin/storage/memory/memory.go
nidhey27: I see your point about the "inferred-" prefix not always making sense, especially when we already have a clear identifier like `peer.service`. It's true that using the `peer.service` tag directly would be more consistent and give us a better picture of the service interactions. I'll adjust the code to use `peer.service` as the service name without the "inferred-" prefix when it's available. This should help us avoid creating unnecessary distinctions in the service map.

I propose the following adjustment to the `inferServiceName` function:

```
func inferServiceName(span *model.Span, isLeaf bool) string {
   // If peer.service tag is present, use it directly as it is a reliable indicator of the service name.
   if peerService, found := getTagValue(span, "peer.service"); found {
      serviceName := peerService
      if isLeaf {
         serviceName += "-leaf"
      }
      return serviceName
   }
   ...
}
```

plugin/storage/memory/memory.go
nidhey27: I've reconsidered the isLeaf parameter and agree that it's not needed right now since we only use inferServiceName for leaf spans. Keeping the code simple and avoiding unnecessary complexity is a priority, so I removed the isLeaf check. This will make the function more straightforward for our current needs.

plugin/storage/memory/memory.go
```diff
@@ -107,20 +107,12 @@ func (st *Store) GetDependencies(ctx context.Context, endTs time.Time, lookback
...
-							Child:     s.Process.ServiceName,
-							CallCount: 1,
-						}
-					} else {
-						deps[depKey].CallCount++
-					}
+				if parentSpan != nil && parentSpan.Process.ServiceName != s.Process.ServiceName {
+					updateServiceDependencyLinks(parentSpan.Process.ServiceName, s.Process.ServiceName, deps)
+				}
+				if isLeaf(s, trace.Spans) {
```
yurishkuro, line 113: A span being a leaf in the call graph does not mean there's some uninstrumented service below it, it could be a genuine leaf. In addition to leaf you also need to check that the span kind is client or producer, as that would definitively indicate that there is supposed to be "the other side" of that transmission, which is not found in the trace

nidhey27: Agreed! Being a leaf doesn't necessarily mean an uninstrumented service. I have updated the function.

yurishkuro (approved):

plugin/storage/memory/memory_test.go
```diff
@@ -523,3 +523,260 @@ func makeTestingSpan(traceID model.TraceID, suffix string) *model.Span {
...
+		},
+		{
+			tags: model.KeyValues{
+				model.String("http.route", "users"),
+			},
+			expectedName: "http-users",
+		},
+		{
+			tags: model.KeyValues{
+				model.String("db.system", "mysql"),
```
yurishkuro, line 763: add a test for blank to increase code coverage

nidhey27: Done!

yurishkuro:
I suggest you take a look at hotrod's graph with this logic, it looks odd. Originally HotROD instrumentation was intentionally tweaked to produce a nice graph since we didn't have this derived client nodes displaying. Perhaps you can undo some of those changes and/or change instrumentation to follow semantic conventions better (e.g. so that Redis is recognized). There's also an odd loop around mysql, it shouldn't be there.

image

examples/hotrod/services/customer/database.go
```diff
@@ -75,9 +74,8 @@ func newDatabase(tracer trace.Tracer, logger log.Factory) *database {
 func (d *database) Get(ctx context.Context, customerID int) (*Customer, error) {
 	d.logger.For(ctx).Info("Loading customer", zap.Int("customer_id", customerID))
 
-	ctx, span := d.tracer.Start(ctx, "SQL SELECT", trace.WithSpanKind(trace.SpanKindClient))
+	_, span := d.tracer.Start(ctx, "SQL SELECT")
```
yurishkuro, line 77: don't follow the reasoning. Both ctx and span kind were appropriate

yurishkuro (approved):

examples/hotrod/services/customer/database.go
yurishkuro: What I mean is I don't agree with the change, it doesn't make sense to me. This particular code was already correct.

nidhey27: Alright, I have tested the HotROD code by reverting the changes, and it looks as we expected.

yurishkuro:
Replaced by #5062

PR #4862, opened 2023-10-19
[fix]: Close previous es.Client when creating new one on password fil…

MeenuyD:
It was observed that the last logged request to the server used a Basic Auth string that resembled the old password, even after a new password was set. This could potentially happen if the es.Client object continued sending pings to the server without being properly closed.
This commit addresses the issue by ensuring that the es.Client object is correctly closed when needed, preventing any further requests with outdated passwords.

Resolves https://github.com/jaegertracing/jaeger/issues/4743


yurishkuro:
please rebase on recent main branch

yurishkuro:
addressed in #4863

PR #4864, opened 2023-10-19
Add OTel resource detector to Jaeger components

### LINKED ISSUE #4534: [OTEL] Process info is missing from the spans

james-ryans:
- Resolves #4534
- Continuation from #4844

## Description of the changes
- Add OTel resources to Jaeger components and tracegen tool.

- Tested manually in local
image
image

## Checklist

### DIFF
--- a/cmd/tracegen/main.go
+++ b/cmd/tracegen/main.go
@@ -82,12 +82,21 @@ func createTracers(cfg *tracegen.Config, logger *zap.Logger) ([]trace.Tracer, fu
 		}
 		logger.Sugar().Infof("using %s trace exporter for service %s", cfg.TraceExporter, svc)
 
+		res, err := resource.New(
+			context.Background(),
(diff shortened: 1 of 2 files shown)


james-ryans:
Hi @yurishkuro, I'm not sure if I have covered all the SDK traces with the OTel detector resource. Can you point out any that I may have missed? Thanks! Also I'll update jaeger-ui's SDK trace later.

Issue #4843, opened 2023-10-15
feat: Jaeger v2 based on OpenTelemetry collector


yurishkuro:
Proposal: https://docs.google.com/document/d/1s4_6VgAS7qAVp6iEm5KYvpiGw3h2Ja5T5HpKo29iv00/view

This replaces the previous attempt: #3500.

### Proof of concept

* [x] A single binary with memory storage and all-in-one functionality #4766
* [ ] Validate that the current storage configuration works for primary / archive storage.
* [ ] Default configuration solution (do not require config file for all-in-one) #4842
* [ ] Demonstrate Kafka exporter / receiver (ideally from OTEL), compatible with existing Jaeger Kafka writer
* [ ] Define v2 storage API that operates on OTLP batches, see if internal & remote APIs can be identical

### Feature parity

* [ ] Extend configuration of `jaeger_storage_exporter` to support all existing backends
* [ ] Metrics
* [ ] Traces
* [ ] What to do with admin port / services
* [ ] Adaptive sampling processor

### Prepare for Beta

* [ ] Add to integration tests
* [ ] Add to release pipeline
* [ ] Documentation

### Prepare for GA

* [ ] Helm chart
* [ ] Operator
* [ ] Clarify https://github.com/jaegertracing/jaeger#version-compatibility-guarantees 

### Enhancements

* [ ] SPM built-in
* [ ] Integrate with OTEL collector builder `ocb`
* [ ] Print active configuration (unresolved: [Print final configuration OTEL #1791](https://github.com/open-telemetry/opentelemetry-collector/issues/1791)), https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/27003

pavolloffay:
The V2 is as well affected by https://github.com/open-telemetry/opentelemetry-collector/pull/8768, It can cause startup failure.

yurishkuro:
The new `jaeger` binary is available as Docker image https://hub.docker.com/repository/docker/jaegertracing/jaeger/tags

yurishkuro:
I wrote down some thoughts on what [Jaeger Storage API v2 might look like](https://docs.google.com/document/d/1s4_6VgAS7qAVp6iEm5KYvpiGw3h2Ja5T5HpKo29iv00/edit#heading=h.hlhu8pegwdcj). Interested in feedback.

hdost:
[...]

PR #4890, opened 2023-10-24
ci: Test jaeger-v2 as all-in-one in CI

yurishkuro:
- Enable testing of jaeger-v2 in all-in-one mode in CI
- Part of #4843

## Description of the changes
- Add `cmd/jaeger-v2/Dockerfile`
- Parameterize `scripts/build-all-in-one-image.sh` to allow running `jaeger-v2` instead of `all-in-one`
- Call this script with `jaeger-v2` from .github/workflows/ci-all-in-one-build.yml

```shell
$ BINARY=jaeger-v2 BRANCH=v2-docker GITHUB_SHA=abcd scripts/build-all-in-one-image.sh pr-only
```


yurishkuro:
Marked this WIP because it fails to start jaeger-v2 due to what looks like the random ordering of extensions startup sequence, whereas jaegerquery extension can only run after jaegerstorage

.github/workflows/ci-all-in-one-build.yml
```diff
@@ -51,6 +51,10 @@ jobs:
       if: github.ref_name != 'main'
       run: bash scripts/build-all-in-one-image.sh pr-only
 
+    - name: Build and test jaeger-v2 as all-in-one
+      if: github.ref_name != 'main'
+      run: BINARY=jaeger-v2 bash scripts/build-all-in-one-image.sh pr-only
```
yurishkuro, line 56: this only added 39s to the total run, I think it's worth keeping in this workflow for now

scripts/build-all-in-one-image.sh
```diff
@@ -35,35 +43,40 @@ run_integration_test() {
   docker kill "$CID"
 }
 
-if [ "$mode" = "pr-only" ]; then
+if [ -n "$pull_request" ]; then
```
albertteoh, line 46: I suggest something like the following to help with readability:

```bash
is_pull_request=true

if [ "$is_pull_request" = true ]; then
if [ "$is_pull_request" = false ]; then
```

cmd/jaeger-v2/internal/extension/jaegerquery/server.go
```diff
@@ -35,6 +38,11 @@ func newServer(config *Config, otel component.TelemetrySettings) *server {
 	}
 }
 
+// Dependencies implements extension.Dependent to ensure this always starts after jaegerstorage extension.
```
albertteoh, line 41: 👍🏼

albertteoh (dismissed):

go.mod
```diff
@@ -61,7 +61,7 @@ require (
 	go.opentelemetry.io/collector/exporter/loggingexporter v0.88.0
 	go.opentelemetry.io/collector/exporter/otlpexporter v0.88.0
 	go.opentelemetry.io/collector/exporter/otlphttpexporter v0.88.0
-	go.opentelemetry.io/collector/extension v0.88.0
+	go.opentelemetry.io/collector/extension v0.88.1-0.20231102040825-2e44da36e2c6
```
yurishkuro, line 64: pinned to main branch for now, it will get updates once OTEL does a new release, which should happen before our next release in Dec

PR #4896, opened 2023-10-26
feat: Add blackhole storage, for benchmarking

yurishkuro:
- When benchmarking jaeger-v2 vs. jaeger-collector, it's useful to have a no-op storage.

## Description of the changes
- Add `blackhole` storage type.

- unit tests


albertteoh (approved):
blackhole storage is a cool name, although "no-op" storage is immediately intuitive to me, probably because it's a common parlance.

yurishkuro:
I was going for cool.

PR #4907, opened 2023-10-30
fix dependency policy and add to SECURITY-INSIGHTS.yml

jkowall:
## Description of the changes
Adding security policy to SECURITY.md 
Adding a couple of sections to the SECURITY-INSIGHTS.yml to fix missing items.
Next step is to check the CLOMonitor for other gaps - Need to build a threat model too.

Only text changes in this PR.

## Checklist
- [Not needed] I have added unit tests for the new functionality
- [Not needed] I have run lint and test steps successfully

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -16,7 +16,7 @@ Dependencies are evaluated before being introduced to ensure they:
 
 1) are actively maintained
 2) are maintained by trustworthy maintainers
-3) are licensed in a way not to impact the Jaeger license.
+3) are licensed in a way not to impact the Jaeger license based on [the CNCF license allowlist](https://github.com/cncf/foundation/blob/main/allowed-third-party-license-policy.md).
 
 These evaluations vary from dependency to dependencies.
 
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -6,6 +6,25 @@ The Jaeger project provides community support only for last minor version: bug f
 
 Security fixes are given priority and might be enough to cause a new version to be released.
 
+### Security Patch Policy
+
+CVEs in Jaeger code will be patched in the newest Jaeger releases. 
+
+### Dependencies Lifecycle Policy
+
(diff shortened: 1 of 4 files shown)


SECURITY.md
```diff
@@ -6,6 +6,25 @@ The Jaeger project provides community support only for last minor version: bug f
...
+
+CVEs in Jaeger code will be patched in the newest Jaeger releases. 
+
+### Dependencies Lifecycle Policy
+
+Dependencies are evaluated before being introduced to ensure they:
+
+1) are actively maintained
+2) are maintained by trustworthy maintainers
+3) are licensed in a way not to impact the Jaeger license.
```
yurishkuro, line 19: can be more precise here - they must be using a license on CNCF allowlist https://github.com/cncf/foundation/blob/main/allowed-third-party-license-policy.md

jkowall: Should be good with latest push

SECURITY.md
```diff
@@ -6,6 +6,25 @@ The Jaeger project provides community support only for last minor version: bug f
...
+1) are actively maintained
+2) are maintained by trustworthy maintainers
+3) are licensed in a way not to impact the Jaeger license based on [the CNCF license allowlist](https://github.com/cncf/foundation/blob/main/allowed-third-party-license-policy.md).
+
+These evaluations vary from dependency to dependencies.
+
+Dependencies are also scheduled for removal if the project has been deprecated or if the project is no longer maintained. Additionally based on license changes we replace dependencies as necessary.
+
+CVEs in dependencies will be patched for all supported versions if the CVE is applicable and is assessed by Snyk to be
+of high or critical severity. Automation generates a new dependabot scan daily and alerts are addressed.
```
yurishkuro, line 26: we also have FOSSA scanning enabled in case some dependency changes to a non-allowed license. Unfortunately, I don't think FOSSA reports are publicly available (very annoying)

yurishkuro (approved):

Issue #4906, opened 2023-10-30
[Bug]: Default badger path is no longer writtable


freef4ll:
### What happened?

With the release of 1.50, [documented badger default path](https://www.jaegertracing.io/docs/1.50/deployment/#badger---local-storage) is no longer writable as it is owned by root after the changes in https://github.com/jaegertracing/jaeger/pull/4783

### Steps to reproduce

```
docker run \
  -e SPAN_STORAGE_TYPE=badger \
  -e BADGER_EPHEMERAL=false \
  -e BADGER_DIRECTORY_VALUE=/badger/data \
  -e BADGER_DIRECTORY_KEY=/badger/key \
  -v /tmp/badger:/badger   -p 16686:16686   jaegertracing/all-in-one:1.50
```

> {"level":"fatal","ts":1698680105.917989,"caller":"./main.go:110","msg":"Failed to init storage factory","error":"Error Creating Dir: \"/badger/key\" error: mkdir /badger/key: permission denied","stacktrace":"main.main.func1\n\t./main.go:110\ngithub.com/spf13/cobra.(*Command).execute\n\tgithub.com/spf13/cobra@v1.7.0/command.go:940\ngithub.com/spf13/cobra.(*Command).ExecuteC\n\tgithub.com/spf13/cobra@v1.7.0/command.go:1068\ngithub.com/spf13/cobra.(*Command).Execute\n\tgithub.com/spf13/cobra@v1.7.0/command.go:992\nmain.main\n\t./main.go:243\nruntime.main\n\truntime/proc.go:267"}

### Expected behavior

Jaeger to start up

### Jaeger backend version

1.50

yurishkuro:
Makes sense, but I don't think we're going to fix that. You have a manual workaround of changing permissions on the existing files to non-root user.

To plat devil's advocate, how would you expect Jaeger to handle that? Changing to non-root user was the right choice from security perspective.

The other thing is we should call this out in the release notes as a breaking change and suggest remediation steps, like running chown - would you like to submit a PR for that?

miparnisari:
> The other thing is we should call this out in the release notes as a breaking change and suggest remediation steps, like running chown

Yes please

yurishkuro:
updated changelog and release notes

yurishkuro:
thanks for reporting!

miparnisari:
I'm not a Docker expert, how would I prevent this problem from occurring with my existing Docker-compose file?

[...]

PR #4910, opened 2023-10-31
Pre-create badger directory in Docker image

freef4ll:

#4906

## Description of the changes
- Create directory for badger in a non-root environment so that existing setups do not break.

- Tested permissions in a minimal container.

## Checklist


yurishkuro:
I don't think this change is going to help. First, there is no requirement that the actual directory is `/badger`, it's up to user to specify what it is. Second, all this change is doing is creating a dir inside the container file system during image creation, whereas in real deployment such directory would need to be mapped to host file system.

yurishkuro:
> Correct, but if it is mapped by a named volume then it works for the default configuration.

why would it work? if you have files on host file system owned by a different user-id, then the owner of in-container directory still not going to match the permissions on those files. Can you show commands that demonstrate a successful test?

PR #4913, opened 2023-10-31
Make ArchiveTrace button auto-configurable

### LINKED ISSUE #4874: [Feature] Make ArchiveTrace button in the UI auto-configurable
When archive storage is configured in the query-service it can receive POST requests from the UI with a trace ID and clone that trace from primary storage into archive, useful for longer retention. However, the Archive Trace button has to be separately enabled in the UI for that, even though the query-service can check at runtime if archive storage is available.

## Proposal

When query-service embeds the UI config into UI loading page, it can also include a structure describing the capabilities of the backend. One of the flags in this structure would be whether the archive storage is available, so that users wouldn't need to configure it separately in the UI. The same struct can be used to inform the UI of other things, for example whether more flexible queries are supported by the storage, like regex on tags (supported in ES but not in Cassandra).

We should still keep the enableArchive in the main UI config, but use it as a kill switch / override for archive function. When the UI decides whether to show the button, it would be `backendSupportsArchive && config.archiveEnabled`.

Implementing this ticket requires some knowledge of both Go and Node.js

## Details, Pointers

The `query-service` injects a config into UI loading page in the static handler: https://github.com/jaegertracing/jaeger/blob/089fc3c878d7def9ce92ab7194012061a8f185de/cmd/query/app/static_handler.go#L120

thecoons:
- backend part of #4874

## Description of the changes
The button to archive a trace is now configured based on the state of
the QueryService in addition to the UI configuration. It is now
possible to request features from the QueryService to inject them
into the UI.

Related UI change https://github.com/jaegertracing/jaeger-ui/pull/1944

All corresponding tests have been updated.

## Checklist
- [X] I have read
https://github.com/jaegertracing/jaeger/blob/master/CONTRIBUTING_GUIDELINES.md
- [X] I have added unit tests for the new functionality

---------

Signed-off-by: Antonin Barthelemy <antobarth@gmail.com>


thecoons:
I am still working on the feature but would appreciate any feedback on the design choices. Thank you in advance.

cmd/query/app/static_handler.go
```diff
@@ -71,12 +73,14 @@ type StaticAssetsHandlerOptions struct {
...
 	LogAccess    bool
+	Features     querysvc.QueryServiceFeature
 	Logger       *zap.Logger
 }
 
 type loadedConfig struct {
-	regexp *regexp.Regexp
-	config []byte
+	regexp                 *regexp.Regexp
+	backendSupportsArchive bool
```
yurishkuro, line 82: this doesn't really belong in this struct. I would instead store the `StorageCapabilities` inside `staticHandler` struct and change the functions below to be on `staticHandler` receiver

yurishkuro (commented):
as I can see you're not actually using the capability flag in generating the UI config.

I suggest following the pattern already used in index.html and defining another function `getStorageCapabilities` to be substituted by the query-service's static handler. I just added a bit of documentation how this is currently done: https://github.com/jaegertracing/jaeger-ui/pull/1926

cmd/query/app/querysvc/query_service.go
```diff
@@ -41,6 +41,13 @@ type QueryServiceOptions struct {
 	Adjuster          adjuster.Adjuster
 }
 
+// QueryServiceFeature is a feature flag for query service
+type QueryServiceFeature struct {
```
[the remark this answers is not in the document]
thecoons: Done

cmd/query/app/querysvc/query_service.go
```diff
@@ -41,6 +41,13 @@ type QueryServiceOptions struct {
...
+	SupportArchive bool
```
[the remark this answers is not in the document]
thecoons: Done

cmd/query/app/querysvc/query_service.go
```diff
@@ -151,3 +165,8 @@ func (opts *QueryServiceOptions) InitArchiveStorage(storageFactory storage.Facto
 	opts.ArchiveSpanWriter = writer
 	return true
 }
+
+// isSupportsArchive returns true if archive storage reader/writer are initialized.
+func (opts *QueryServiceOptions) isSupportsArchive() bool {
```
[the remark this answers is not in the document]
thecoons: Done

cmd/query/app/static_handler.go
thecoons: Indeed, the functions loadAndEnrichIndexHTML and loadUIConfig are dependent on the instance of StaticAssetsHandlerOptions, but I left loadIndexHtml without a receiver as it seems to me completely static.

Done

thecoons: Would you like the refactoring of the receivers to be in a separate commit to facilitate reading, or keep a single commit that contains all the content of the feature?

thecoons:
> as I can see you're not actually using the capability flag in generating the UI config.
> 
> I suggest following the pattern already used in index.html and defining another function `getStorageCapabilities` to be substituted by the query-service's static handler. I just added a bit of documentation how this is currently done: [jaegertracing/jaeger-ui#1926](https://github.com/jaegertracing/jaeger-ui/pull/1926)

I've taken a look. Do you think we should include the `QueryServiceCapabilities` within the `redux.config` state, or should we create a dedicated state?

yurishkuro:
Yes, we can add it to `redux.config`

cmd/query/app/querysvc/query_service_test.go
```diff
@@ -285,6 +285,24 @@ func TestGetDependencies(t *testing.T) {
 	assert.Equal(t, expectedDependencies, actualDependencies)
 }
 
+// Test QueryService.GetFeatures()
+func TestGetFeatures(t *testing.T) {
```
yurishkuro, line 289:
```suggestion
func TestGetCapabilities(t *testing.T) {
```

cmd/query/app/static_handler.go
```diff
@@ -112,75 +116,92 @@ func NewStaticAssetsHandler(staticAssetsRoot string, options StaticAssetsHandler
...
+	sH.options.Logger.Info("reloading UI config", zap.String("filename", sH.options.UIConfigPath))
+	content, err := sH.options.loadAndEnrichIndexHTML(sH.assetsFS.Open)
+	if err != nil {
+		sH.options.Logger.Error("error while reloading the UI config", zap.Error(err))
+	}
+	sH.indexHTML.Store(content)
+	sH.options.Logger.Info("reloaded UI config", zap.String("filename", sH.options.UIConfigPath))
+}
+
+func (sH *StaticAssetsHandler) loggingHandler(handler http.Handler) http.Handler {
```
yurishkuro, line 129: please keep the methods where they were. This reordering is introducing changes to the PR that are unnecessary and make it much more difficult to review.

cmd/query/app/static_handler.go
```diff
@@ -30,26 +30,29 @@ import (
...
+	configPattern      = regexp.MustCompile("JAEGER_CONFIG *= *DEFAULT_CONFIG;")
+	configJsPattern    = regexp.MustCompile(`(?im)^\s*\/\/\s*JAEGER_CONFIG_JS.*\n.*`)
+	versionPattern     = regexp.MustCompile("JAEGER_VERSION *= *DEFAULT_VERSION;")
+	compabilityPattern = regexp.MustCompile("JAEGER_STORAGE_CAPABILITIES *= *DEFAULT_STORAGE_CAPABILITIES;")
+	basePathPattern    = regexp.MustCompile(`<base href="/"`) // Note: tag is not closed
 )
 
 // RegisterStaticHandler adds handler for static assets to the router.
-func RegisterStaticHandler(r *mux.Router, logger *zap.Logger, qOpts *QueryOptions) {
+func RegisterStaticHandler(r *mux.Router, logger *zap.Logger, qOpts *QueryOptions, qFeatures querysvc.StorageCapabilities) {
```
yurishkuro, line 49:
```suggestion
func RegisterStaticHandler(r *mux.Router, logger *zap.Logger, qOpts *QueryOptions, qCapabilities querysvc.StorageCapabilities) {
```

cmd/query/app/querysvc/query_service_test.go
thecoons: Done

cmd/query/app/static_handler.go
thecoons: Done

thecoons:
I always have problems integrating the feature, can you confirm that in its current state the feature is not functional?

yurishkuro (approved):
lgtm

thecoons:
Thank you for the help; I'll try to be more independent with the upcoming features on Jaeger.

Issue #4196, opened 2023-01-31
ClickHouse as a core storage backend


yurishkuro:
## Summary

Build first-class support for [ClickHouse ](https://github.com/ClickHouse/ClickHouse) as an official Jaeger backend. ClickHouse is an open-source column-oriented database for OLAP use cases. It is highly efficient and performant for high volumes of ingestion and search making it a good database for tracing and logging data specifically. It can also do aggregates very quickly which will come in handy for several features in Jaeger. 

Benefits to the users:

* Efficient backend
* Powerful search
* Analytics capability, e.g. the possibility to support the APM function (Monitoring tab in Jaeger) directly from ClickHouse

## Background
* This ticket is a part of CNCG GSoC program (see [Jaeger project entry](https://github.com/cncf/mentoring/blob/main/summerofcode/2023.md#jaeger)). 
* Applications will be open from March-20 to April-4: https://summerofcode.withgoogle.com/

This is a continuation of #1438. Currently, there is one (or more) community-supported plugins for ClickHouse as a Jaeger storage backend. This ticket is about making it one of the core backends supported by Jaeger, on the level of Cassandra/Elasticsearch.

## Scope
* Design table schema, document trade-offs
* Implement `plugin/storage/clickhouse` module and integrate it with `storage.Factory`
  * ClickHouse is not optimized for individual inserts, need to address that (cf. OTel Collector ClickHouse exporter)
  * For deployments with Kafka, investigate Kafka Connector for ClickHouse for direct ingestion
* Address schema creation problem (similar to `plugin/storage/cassandra/schema`)
* Address management needs if any (e.g., ES has index rollover, not sure if ClickHouse needs something similar)
* Add relevant documentation to the Jaeger website

### Expected outcomes
* Design table schema, document trade-offs
* Implement ClickHouse support as core storage backend
* Address schema creation problem / tooling
* Add relevant documentation to the Jaeger website

### Stretch goals
* Use Kafka Connector for ClickHouse for bulk inserts
* Support Monitoring tab (https://www.jaegertracing.io/docs/1.43/spm/) in Jaeger directly from ClickHouse

yurishkuro:
For reference: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/clickhouseexporter

abhi1287:
Suggestion: It would be nice to somehow retain the type information of custom tags. 
The otel implementation you linked above simply stores a vector of string key value pairs.
[...]

PR #4941, opened 2023-11-12
[WIP] Implement spans exporting for ClickHouse storage for Jaeger V2

haanhvu:
First implementation step of #4196 for Jaeger V2

Still in WIP. Add tests later.


haanhvu:
@yurishkuro This still needs some supplements and a lot of cleanups and tests. But could you take an initial look to see if I'm going the right way? Thanks!

cmd/jaeger/internal/extension/jaegerstorage/config.go
```diff
@@ -13,9 +14,16 @@ type Config struct {
...
+
+	ClickHouse map[string]ch.Config `mapstructure:"clickhouse"`
 }
 
 type MemoryStorage struct {
 	Name string `mapstructure:"name"`
 	memoryCfg.Configuration
 }
+
+type ClickHouseStorage struct {
```
yurishkuro, line 26: is this type used?

plugin/storage/clickhouse/factory.go
```diff
@@ -0,0 +1,71 @@
...
+	} else {
+		f.client = client
+	}
+
+	return f
+
+	//TODO: Move some steps to Initialize()
+}
+
+func (f *Factory) CreateSpansTable(ctx context.Context) {
```
yurishkuro, line 43: why is this function public, who would call it?

plugin/storage/clickhouse/factory.go
```diff
@@ -0,0 +1,71 @@
...
+	if err := chSpanStore.CreateSpansTable(ctx, f.client, f.spansTableName); err != nil {
+		f.logger.Error("failed to create spans table", zap.Error(err))
+	}
+}
+
+func (f *Factory) ExportSpans(ctx context.Context, td ptrace.Traces) error {
```
yurishkuro, line 49: why is ExportSpans defined on the factory instead of span store?

haanhvu: The logic is defined in `plugin/storage/spanstore/clickhouse/exporter.go` ([link](https://github.com/jaegertracing/jaeger/pull/4941/files#diff-e85bace6bfd6e03dd602c1d1bd2ec27367e539b554e541d6f30786a63d652f2eR43)). But the exporter is depending on the real type of Factory to decide whether to call `SpanWriter.WriteSpan()` or `ExportSpans()` ([link](https://github.com/jaegertracing/jaeger/pull/4941/files#diff-d11f17877ed77e35143c698010924340c6cf95489656e3ccb0b3c2654c9b9853R42-R52). So I think making `ExportSpans()` a func of clickhouse Factory may be a good idea.

Of course, a better idea is to define a new API for `ExportSpans()`, but maybe we can do that later? Or you have a better idea?

cmd/jaeger/internal/extension/jaegerstorage/config.go
haanhvu: Ah, I followed your MemoryStorage struct above. Now realize I'm currently not needing this for anything. Do you use MemoryStorage struct for anything?

plugin/storage/clickhouse/factory.go
haanhvu: Currently, Jaeger V2's exporter checks the type of Factory. If it's a clickhouse Factory, then f.CreateSpansTable() is called, as you can see in this [piece](https://github.com/jaegertracing/jaeger/pull/4941/files#diff-d11f17877ed77e35143c698010924340c6cf95489656e3ccb0b3c2654c9b9853R42-R52).

plugin/storage/clickhouse/factory.go
yurishkuro: I have a better idea, but it is not well formed yet, I need to write it down and maybe put together the new interfaces. Will try to do this later this weekend. Are you blocked on it?

haanhvu: No I don't think this blocks the progress. Can continue.

k0zl:
👍 I've been waiting clickhouse support for a long time. Is there anything I can help you with?

haanhvu:
> 👍 I've been waiting clickhouse support for a long time. Is there anything I can help you with?

@k0zl Sure! For this PR (spans exporting), the remaining parts are:

- Finish schema implementation in `plugin/storage/clickhouse/spanstore/schema.go`. We described our schema choices in our [benchmark report](https://medium.com/jaegertracing/making-design-decisions-for-clickhouse-as-a-core-storage-backend-in-jaeger-62bf90a979d)). As you can see in the current `schema.go`, I just implemented the spans table, not the materialized views yet.
- Add tests: I'm pushing an unit `exporter_test.go` this week. There are other unit tests, e2e test, or even local test with Jaeger v2 binary to do. (Jaeger v2 issue/decription is [here](https://github.com/jaegertracing/jaeger/issues/4843))

After finishing with spans exporting, we'll move to spans reading,...

Also, since clickhouse storage is supported in Jaeger v2, to make this realeased you could also help with Jaeger v2 roadmap, in the issue I linked above.

Please choose whatever interests you. Thanks!

yurishkuro:
@haanhvu it would be useful to write a design doc / roadmap documenting what we're doing and what tasks need to be done, are you up for that?  You can model it after the Jaeger V2 document.

haanhvu:
> @haanhvu it would be useful to write a design doc / roadmap documenting what we're doing and what tasks need to be done, are you up for that? You can model it after the Jaeger V2 document.

Yeah I'll create an issue for supporting clickhouse in jaeger v2 tomorrow

haanhvu:
@yurishkuro Could you take a look at the failed [all-in-one build](https://github.com/jaegertracing/jaeger/actions/runs/7358386059/job/20031566276?pr=4941)? `make all-in-one-integration-test` passed locally for me. Not sure why it fails in CI.

yurishkuro:
I am seeing error
```
Error: failed to build pipelines: failed to create "jaeger_storage_exporter" exporter for data type "traces": nil PushTraces
2023/12/29 16:57:59 failed to build pipelines: failed to create "jaeger_storage_exporter" exporter for data type "traces": nil PushTraces
+ exit 1
```

cmd/jaeger/internal/extension/jaegerstorage/config.go
yurishkuro: not needed, you can remove both

cmd/jaeger/internal/exporters/storageexporter/exporter.go
```diff
@@ -14,13 +14,16 @@ import (
...
 )
 
 type storageExporter struct {
-	config     *Config
-	logger     *zap.Logger
-	spanWriter spanstore.Writer
+	config             *Config
+	logger             *zap.Logger
+	spanWriter         spanstore.Writer
+	exportTraces       func(ctx context.Context, td ptrace.Traces) error
```
yurishkuro, line 25: nit: I would call this `consumeTraces`, because the argument that we pass to the exporter helper is `consumer.ConsumeTracesFunc`

cmd/jaeger/internal/exporters/storageexporter/factory.go
```diff
@@ -34,8 +34,21 @@ func createDefaultConfig() component.Config {
 func createTracesExporter(ctx context.Context, set exporter.CreateSettings, config component.Config) (exporter.Traces, error) {
 	cfg := config.(*Config)
 	ex := newExporter(cfg, set.TelemetrySettings)
+
+	if ex.requireBatchInsert {
```
yurishkuro, line 38: you don't need an if statement:
```go
exporterhelper.WithQueue(exporterhelper.QueueSettings{Enabled: ex.requireBatchInsert}),
```

cmd/jaeger/internal/exporters/storageexporter/exporter.go
```diff
@@ -30,14 +33,23 @@ func newExporter(config *Config, otel component.TelemetrySettings) *storageExpor
...
+	case *ch.Factory:
+		t.CreateSpansTable(ctx)
+		exp.requireBatchInsert = true
+		exp.exportTraces = t.ExportSpans
+	default:
+		if exp.spanWriter, err = f.CreateSpanWriter(); err != nil {
+			return fmt.Errorf("cannot create span writer: %w", err)
+		}
+		exp.requireBatchInsert = false
+		exp.exportTraces = exp.pushTraces
```
yurishkuro, line 52: nit: I'd rename `s/pushTraces/writeSpans/` to indicate jaeger-v1 style storage

cmd/jaeger/internal/exporters/storageexporter/exporter.go
```diff
@@ -30,14 +33,23 @@ func newExporter(config *Config, otel component.TelemetrySettings) *storageExpor
...
+func (exp *storageExporter) start(ctx context.Context, host component.Host) error {
 	f, err := jaegerstorage.GetStorageFactory(exp.config.TraceStorage, host)
 	if err != nil {
 		return fmt.Errorf("cannot find storage factory: %w", err)
 	}
 
-	if exp.spanWriter, err = f.CreateSpanWriter(); err != nil {
-		return fmt.Errorf("cannot create span writer: %w", err)
+	switch t := f.(type) {
+	case *ch.Factory:
```
yurishkuro, line 43: it seems you're doing this too late, I think `createTracesExporter` would be called before `start`

haanhvu:
> I am seeing error
> 
> ```
> Error: failed to build pipelines: failed to create "jaeger_storage_exporter" exporter for data type "traces": nil PushTraces
> 2023/12/29 16:57:59 failed to build pipelines: failed to create "jaeger_storage_exporter" exporter for data type "traces": nil PushTraces
> + exit 1
> ```

@yurishkuro This error didn't come up in my `make all-in-one-integration-test`. Probably because my `make all-in-one-integration-test` checked v1, not v2? If that's the case how to specify v2 in this test locally?

yurishkuro:
> This error didn't come up in my make `all-in-one-integration-test`

This test doesn't just run on its own, it expects the binary running, so how are you starting the binary?

For example, this works:
```shell
# terminal 1
$ go run -tags=ui ./cmd/jaeger

# terminal 2
$ SKIP_SAMPLING=true make all-in-one-integration-test
```

But this is running with default aio config for OTEL, whereas you need to provide your own config in order to test the same against CH.

haanhvu:
> ```shell
> go run -tags=ui ./cmd/jaeger
> ```

Ah, I ran the binary with `make run-all-in-one`... That's why v2 wasn't checked... Let me try again

haanhvu:
> But this is running with default aio config for OTEL, whereas you need to provide your own config in order to test the same against CH.

So I'll have to run clickhouse server, make my own jaeger v2 config (that exports to clickhouse) and `go run -tags=ui -config=config.yaml ./cmd/jaeger` right? I'll try it tomorrow.

haanhvu:
@yurishkuro I ran the binary with my [config.yaml](https://github.com/jaegertracing/jaeger/pull/4941/files#diff-d8d0422389f03d783e32e627250fe29834bd09c6361640d1ff00661dd6820034), spans were exported successfully to clickhouse:
Screenshot from 2024-01-03 03-07-22

Should we add an e2e test now, or wait until spans reading are implemented?

Also, can you check the unit tests? I just have one `factory_test.go`, no test files for `config.go` or `spanstore` because factory already calls config and spanstore.

Is there anything else you think we should test in this PR?

yurishkuro:
> Should we add an e2e test now, or wait until spans reading are implemented?

You wouldn't be able to do e2e without reading, that's how the tests verify that traces are saved.

cmd/jaeger/internal/exporters/storageexporter/exporter.go
```diff
@@ -30,14 +35,22 @@ func newExporter(config *Config, otel component.TelemetrySettings) *storageExpor
...
+		exp.clickhouse = true
+		t.CreateSpansTable(ctx)
```
yurishkuro, line 47: why can't we do this inside `clickhouse.NewFactory`? `storageexporter` doesn't need to know about internal details of each storage implementation

haanhvu:
@yurishkuro I rebased and made a little change following what we're having with other V2 storage. I checked the change locally and it worked:
Screenshot from 2024-06-13 15-15-16

There's one goleak failure in unit tests. However, the goleak failure is in `database/sql` package, not our code. Should we ignore or do what with this?

Also, the code coverage is a bit skewed. Seems like codecov checks per file or per package and crossed-files and cross-packages function calls are not counted. Do you want to rearrange the files/packages so that code coverage is more accurate?

Please also checks if we need more tests in this PR.

yurishkuro:
> There's one goleak failure in unit tests. However, the goleak failure is in database/sql package, not our code. Should we ignore or do what with this?

We shouldn't ignore leaks. Most likely it's from not closing resources in tests properly, you can see which goroutine is remaining running and understand where it starts. There are rare cases where libraries introduce unsolvable leaks (like glog), but I am skeptical database/sql would be such lib.

The codecov for storage is usually combined from unit tests and e2e tests.

Note that V2 Storage API [already exists](https://github.com/jaegertracing/jaeger/pull/5497), and we have an adapter to v1 that will be used by the `jaegerexporter`. The Clickhouse implementation can be the first one to be natively supporting V2.

haanhvu:
> Note that V2 Storage API https://github.com/jaegertracing/jaeger/pull/5497, and we have an adapter to v1 that will be used by the jaegerexporter. The Clickhouse implementation can be the first one to be natively supporting V2.

Yeah I saw the conflicts in `exporter.go`. I'll rebase and make the current implementation work with the rebase first, then make clickhouse v2-native next.

haanhvu:
The latest rebased implementation works (not v2-native yet)
Screenshot from 2024-06-23 13-19-14

JellyTony:
Is there any progress on this PR currently, and when will it be merged

PR #4964, opened 2023-11-24
Collect code coverage from integration tests and upload to codecov

### LINKED ISSUE #1516: Use integration tests to generate coverage

### LINKED ISSUE #4965: Remove unused token-propagation-integration-test
This makefile target is not called from anywhere, a leftover from #4162
```
$ grx token-propagation-integration-test .
./Makefile:157:.PHONY: token-propagation-integration-test
./Makefile:158:token-propagation-integration-test:

$ grx TestBearTokenPropagation .
./Makefile:160:	bash -c "set -e; set -o pipefail; $(GOTEST) -tags token_propagation -run TestBearTokenPropagation $(STORAGE_PKGS) | $(COLORIZE)"
```

yurishkuro:
- Some of our code is not testable unless we run integration tests, e.g. places that require db connection
- But we don't collect coverage from integration tests
- Resolves #1516
- Resolves #4965

## Description of the changes
- Add coverage generation to storage tests and upload to codecov
- Remove several `.nocover` since their packages are now included in coverage via integration tests anyway. Add corresponding `empty_test.go` to avoid linter error.
- Remove `println` from badger & sampling storage integration
- Change COLORIZE macro to include `| ...` so that it can be overwritten to nothing for faster std output. When output is piped, even to `| cat -`, the buffering prevents incremental test output, so one cannot easily observe test progress

- CI is green
- [Comment below by codecov](https://github.com/jaegertracing/jaeger/pull/4964#issuecomment-1825819949) shows 11 reports (inside expander, Flags table)
- The global code coverage dropped -0.82%, which is not surprising since some previously excluded packages are now counted, but not all the code in them is exercised even in integration tests (e.g. kafka/auth/tls.go). However, we can ramp it up in future PRs. One area that seems particularly neglected is calling Close on the storage from integration tests, which we're not doing.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -6,9 +6,11 @@ codecov:
 
 ignore:
   - "**/*.pb.go"
-  - "**/*-gen/*"
   - "**/mocks/*"
+  - "proto-gen/*/*"
+  - "thrift-gen/*/*"
   - "**/thrift-0.9.2/*"
+  - "swagger-gen/*/*"
 
 coverage:
   precision: 2
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,7 +1,7 @@
 codecov:
   notify:
     require_ci_to_pass: yes
-    after_n_builds: 3
+    after_n_builds: 11
   strict_yaml_branch: main  # only use the latest copy on the main branch
 
 ignore:
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -4,6 +4,12 @@ codecov:
     after_n_builds: 3
   strict_yaml_branch: main  # only use the latest copy on the main branch
 
+ignore:
+  - "**/*.pb.go"
+  - "**/*-gen/*"
+  - "**/mocks/*"
+  - "**/thrift-0.9.2/*"
+
 coverage:
   precision: 2
   round: down
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,6 +1,7 @@
 codecov:
   notify:
     require_ci_to_pass: yes
+    after_n_builds: 3
   strict_yaml_branch: main  # only use the latest copy on the main branch
 
 coverage:
--- a/Makefile
+++ b/Makefile
@@ -129,7 +129,7 @@ storage-integration-test: go-gen
 	# Expire tests results for storage integration tests since the environment might change
 	# even though the code remains the same.
 	go clean -testcache
-	bash -c "set -e; set -o pipefail; $(GOTEST) $(STORAGE_PKGS) | $(COLORIZE)"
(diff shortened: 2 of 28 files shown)


Makefile
```diff
@@ -129,47 +133,34 @@ storage-integration-test: go-gen
...
 index-rollover-integration-test: docker-images-elastic
 	# Expire test results for storage integration tests since the environment might change
 	# even though the code remains the same.
 	go clean -testcache
-	bash -c "set -e; set -o pipefail; $(GOTEST) -tags index_rollover $(STORAGE_PKGS) | $(COLORIZE)"
-
-.PHONY: token-propagation-integration-test
-token-propagation-integration-test:
-	go clean -testcache
-	bash -c "set -e; set -o pipefail; $(GOTEST) -tags token_propagation -run TestBearTokenPropagation $(STORAGE_PKGS) | $(COLORIZE)"
```
yurishkuro, line 160: was not used anywhere

Makefile
```diff
@@ -61,6 +61,7 @@ GOFMT=gofmt
 GOFUMPT=gofumpt
 FMT_LOG=.fmt.log
 IMPORT_LOG=.import.log
+COLORIZE ?= | $(SED) 's/PASS/✅ PASS/g' | $(SED) 's/FAIL/❌ FAIL/g' | $(SED) 's/SKIP/☠️ SKIP/g'
```
yurishkuro, line 64: Changed COLORIZE macro to include `| ...` so that it can be overwritten to nothing for faster std output. When output is piped, even to `| cat -`, the buffering prevents incremental test output, so one cannot easily observe test progress

.codecov.yml
```diff
@@ -1,8 +1,17 @@
 codecov:
   notify:
     require_ci_to_pass: yes
+    after_n_builds: 11
```
albertteoh, line 4: Why specifically 11?

albertteoh (approved):
Nice 👍🏼

.codecov.yml
yurishkuro: that's how many reports we upload

PR #4966, opened 2023-11-26
fix(badger): Add missing SamplingStoreFactory.CreateLock method

### LINKED ISSUE #4963: [Bug]: badger as sampling store not work
### What happened?

Error starting "storage factory of type badger does not support sampling store"

badger storage not implement SamplingStoreFactory interface -> no method CreateLock

### Steps to reproduce

docker run --rm --env SAMPLING_STORAGE_TYPE=badger --env SAMPLING_CONFIG_TYPE=adaptive jaegertracing/all-in-one:1.51.0

### Expected behavior

work with badger as sampling store

### Relevant log output

_No response_

### Screenshot

_No response_

### Additional context

_No response_

### Jaeger backend version

v1.51.0

### SDK

_No response_

### Pipeline

_No response_

### Stogage backend

badger

### Operating system

Linux

### Deployment model

Docker

### Deployment configs

_No response_

slayer321:
- Resolves #4963

## Description of the changes
-   Implement missing `CreateLock` method on `SamplingStoreFactory` interface for badger

-  run the Integration and make test command

## Checklist

### DIFF
--- a/plugin/storage/badger/factory.go
+++ b/plugin/storage/badger/factory.go
@@ -53,9 +53,7 @@ var ( // interface comformance checks
 	// TODO badger could implement archive storage
 	// _ storage.ArchiveFactory       = (*Factory)(nil)
 
-	// TODO CreateLock function is missing
-	// Being fixed in https://github.com/jaegertracing/jaeger/pull/4966
-	// _ storage.SamplingStoreFactory = (*Factory)(nil)
+	_ storage.SamplingStoreFactory = (*Factory)(nil)
 )
 
 // Factory implements storage.Factory for Badger backend.
--- a/plugin/storage/badger/factory.go
+++ b/plugin/storage/badger/factory.go
@@ -26,6 +26,7 @@ import (
 	"github.com/spf13/viper"
 	"go.uber.org/zap"
 
+	"github.com/jaegertracing/jaeger/pkg/distributedlock"
 	"github.com/jaegertracing/jaeger/pkg/metrics"
 	"github.com/jaegertracing/jaeger/plugin"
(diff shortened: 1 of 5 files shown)


yurishkuro:
@slayer321 this issue should've been caught at compile time, and via integration test, but it looks like this PR is not changing any of those. The storage factory must implement this interface https://github.com/jaegertracing/jaeger/blob/47f39a8f4f4279c7532aef75a85af4d31d253933/storage/factory.go#L54

I think the issue with integration tests is that the StorageIntegration struct is defined with direct access to SpanWriter etc. that are pre-initialized by the concrete implementation. Instead it would be better if implementations would only initialize the respective factories and let the test retrieve writer, etc. (we could do it in another PR).

plugin/storage/badger/factory.go
```diff
@@ -177,6 +178,11 @@ func (f *Factory) CreateSamplingStore(maxBuckets int) (samplingstore.Store, erro
 	return badgerSampling.NewSamplingStore(f.store), nil
 }
 
+// CreateLock implements storage.SamplingStoreFactory
+func (f *Factory) CreateLock() (distributedlock.Lock, error) {
```
yurishkuro, line 182: please add a test

yurishkuro:
Cf: https://github.com/jaegertracing/jaeger/pull/4967

plugin/storage/badger/factory.go
```diff
@@ -186,6 +187,11 @@ func (f *Factory) CreateSamplingStore(maxBuckets int) (samplingstore.Store, erro
...
```
yurishkuro, line 191: please fix L58

yurishkuro (approved):

PR #4968, opened 2023-11-26
Add codecov to all-in-one integration test

yurishkuro:
- All-in-one integration test was not included in the code coverage

## Description of the changes
- Generate code coverage from all-in-one integration test and upload to codecov
- Split build-all-in-one workflow into two jobs to reduce code repetition

- CI

### DIFF
--- a/.github/workflows/ci-all-in-one-build.yml
+++ b/.github/workflows/ci-all-in-one-build.yml
@@ -18,6 +18,14 @@ permissions:
 jobs:
   all-in-one:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        mode:
+        - name: v1
+          binary: all-in-one
+        - name: v2
+          binary: jaeger
+
     steps:
     - name: Harden Runner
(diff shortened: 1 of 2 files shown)


cmd/all-in-one/all_in_one_test.go
```diff
@@ -37,58 +35,100 @@ import (
...
 
-	t.Run("Check if the favicon icon is available", jaegerLogoCheck)
+	t.Run("checkWebUI", checkWebUI)
 	t.Run("createTrace", createTrace)
 	t.Run("getAPITrace", getAPITrace)
 	t.Run("getSamplingStrategy", getSamplingStrategy)
 	t.Run("getServicesAPIV3", getServicesAPIV3)
 }
 
+func healthCheck(t *testing.T) {
```
yurishkuro, line 75: moved two methods from below to match the order of calls from the test above

jkowall (approved):
lgtm! Nice to have these split out for the future. Should speed up the CI as well.

Issue #4868, opened 2023-10-20
[Jaeger v2] Add Kafka exporter and receiver


yurishkuro:
Part of #4843

## Objective
  * Demonstrate using Kafka exporter & receiver from OTEL Collector Contrib with `jaeger-v2` as a replacement for `jaeger-collector` / `jaeger-ingester`
  * Document gaps in backwards compatibility

## Design

We want to rely on OTel's usual pipeline setup, which should allow us to use OTel's Kafka receiver and exporter without changes. For example, this is how it might look

### Replacement for jaeger-collector
```shell
jaeger-v2 --config collector-with-kafka.yaml
```

```yaml
# collector-with-kafka.yaml
service:
  # extensions: [jaeger_storage, jaeger_query] # don't really need these for collector with Kafka
  pipelines:
    traces:
      receivers: [otlp, jaeger, zipkin]
      processors: [batch]
      exporters: [kafka]  # using standard Kafka exporter as the only one
```

### Replacement for jaeger-ingester
```shell
jaeger-v2 --config ingester.yaml
```

```yaml
# ingester.yaml
service:
  extensions: [jaeger_storage, jaeger_query]
  pipelines:
    traces:
      receivers: [kafka] # using standard Kafka receiver as the only one
      processors: [batch]
      exporters: [jaeger_storage_exporter]  # same as in the default cmd/jaeger-v2/config.yaml
```

## Steps
  * [ ] Add wiring and sample configuration to prototype #4766
  * [ ] Test backwards compatibility:
    * [ ] Sending spans in Jaeger format to Kafka
    * [ ] Receiving spans in all formats that `jaeger-ingester` supports today
  * [ ] Document gaps
  * [ ] Provide reproducible integration test scripts (manually run)

This issue is not a one-evening-worth kind, but the steps are simple enough that I think it can be `good-first-issue`.

NavinShrinivas:
Hey, I'd like to try this 

Just to warrant my understanding : 
- First include `Kafka` configurations in the config files ( receiver : https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/kafkareceiver/README.md and exporter : https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/kafkaexporter/README.md)
- I'm assuming here that the single v2 binary is responsible for ingesting and querying this means : 
   - kafka recvier will get traces (in jeager format from various data sources)
[...]

PR #4971, opened 2023-11-27
[jaeger-v2] Add kafka exporter and receiver configuration

james-ryans:
- Add wiring and sample kafka receiver and exporter configuration to jaeger-v2 prototype #4868

## Description of the changes
- Create collector-with-kafka.yaml and ingester.yaml, to mimic what `jaeger-collector` with kafka span storage type and `jaeger-ingester` already have.
- ingester.yaml configuration exposes prometheus metrics at 8889 port to avoid conflict with collector-with-kafka.yaml.
- Execute the below command to run the collector and ingester components.
```
$ go run -tags=ui ./cmd/jaeger --config ./cmd/jaeger/collector-with-kafka.yaml
$ go run -tags=ui ./cmd/jaeger --config ./cmd/jaeger/ingester.yaml
```

- Execute `./scripts/otel-kafka-integration-test.sh` to test the overall pipelines.
- The testbed module is used to implement the integration test without data race detection because it has goroutines that read/write the same variable.
- The testbed validator checks if the kafka message and memory remote storage have the same spans from the load generator.

## Checklist


cmd/jaeger/ingester.yaml
```diff
@@ -0,0 +1,45 @@
+# ingester.yaml
+service:
+  extensions: [jaeger_storage, jaeger_query]
```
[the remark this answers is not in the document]
james-ryans: Isn't the `jaeger_query` required for the UI on port 16686?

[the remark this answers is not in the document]
yurishkuro: Yes but ingester does not offer ui

james-ryans:
HI! I am currently halfway through implementing the e2e test. I've discoveered testbed pkg from [opentelemetry-collector-contrib](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/testbed) repo, designed for Otel collector integration tests. I've successfully test e2e from load generator (otlp) -> collector-with-kafka -> kafka -> mock backend (kafka receiver), and now I'm stuck to test load generator -> collector-with-kafka -> kafka -> ingester -> mock backend because ingester doesn't actually export the traces instead it stores them to memory. Do you have any suggestions on how should the test be done?

Here's the diagram of testbed and my e2e test structure.
testbed
Testbed structure

kafka-receiver
The successful tested collector with kafka

ingester-receiver
The actual e2e collector needs to be tested

cmd/jaeger/integration/goldendataset/pict_input_traces.txt
```diff
@@ -0,0 +1,3 @@
+Resource: Empty, VMOnPrem, VMCloud, K8sOnPrem, K8sCloud, Faas, Exec
```
yurishkuro, line 1: what is the meaning of the "pict_" prefix?

cmd/jaeger/integration/datareceivers/kafka.go
```diff
@@ -0,0 +1,53 @@
+package datareceivers
```
yurishkuro, line 1: > cmd/jaeger/integration/datareceivers/kafka.go

can we keep this all under $root/internal/cit/?

yurishkuro:
> because ingester doesn't actually export the traces instead it stores them to memory. Do you have any suggestions on how should the test be done?

This is the first time I see the testbed framework, I will need to understand it better. What is the expectation of the mock backend - is it expected to "push" the data somewhere?

If it can be pulled for data by the framework, then you could utilize a remote memory storage (i.e. running as a separate process), similar to how our existing integration test works: https://github.com/jaegertracing/jaeger/blob/9f63383ecf65c1022ef855d5526b1c42c0e235cd/crossdock/jaeger-docker-compose.yml#L4

cmd/jaeger/integration/goldendataset/pict_input_traces.txt
james-ryans: PICT stands for Pairwise Independent Combinatorial Tool (https://github.com/microsoft/pict/), apparently correctnesstest in testbed uses this tool as its source of combinatoric data. Here's the example e2e test with correctnesstest: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/testbed/correctnesstests/traces/correctness_test.go.

james-ryans: Testbed has 3 data provider types to generate the testdata:
1. perfTestDataProvider: randomly generate traces that might not cover all trace combinations.
2. goldenDataProvider: generate traces from PICT generated file with the purpose of covering all combinatorial trace data.
3. fileDataProvider: parse provided traces from a file in yaml or json format.

I decided to use `goldenDataProvider` because `perfTest` and `file` data providers are created for use in performance tests.

james-ryans: > also, why "golden"? Usually we place test input files under ./fixtures/

Because the testbed example stores the test input files under the 'goldendataset/' directory, I'll change that to align with our convention.

cmd/jaeger/integration/datareceivers/kafka.go
james-ryans: Sure!

james-ryans:
> What is the expectation of the mock backend - is it expected to "push" the data somewhere?

Nope, it can be a "pull" strategy too! The mock backend uses the receiver component to retrieve the traces, and there are some receivers in a "pull" strategy, such as kafka_receiver.

> If it can be pulled for data by the framework, then you could utilize a remote memory storage ... similar to how our existing integration test works.

Ah, okay, I think this can do. I'll try to read it.

james-ryans:
Hi, I've almost finished the integration test. Currently, I'm blocked on a mismatch model between OTLP Span Link ([here](https://github.com/open-telemetry/opentelemetry-collector/blob/main/pdata/internal/data/protogen/trace/v1/trace.pb.go#L605)) and Jaeger Span Reference, OTLP Span Link has attributes but Jaeger Span Reference doesn't thus the data sent with span link attributes will be discarded at the Jaeger storage.
```
2024/01/01 21:55:23 Sent and received data counters match.
    validator.go:92: 
        	Error Trace:	/Users/james_ryans/go/src/github.com/james-ryans/jaeger/vendor/github.com/open-telemetry/opentelemetry-collector-contrib/testbed/testbed/validator.go:92
        	            				/Users/james_ryans/go/src/github.com/james-ryans/jaeger/vendor/github.com/open-telemetry/opentelemetry-collector-contrib/testbed/testbed/test_case.go:242
        	            				/Users/james_ryans/go/src/github.com/james-ryans/jaeger/cmd/jaeger/integration/kafka_test.go:96
        	Error:      	Not equal: 
        	            	expected: 0
        	            	actual  : 11824
        	Test:       	TestKafkaStorage
        	Messages:   	There are span data mismatches.
2024-01-01T21:55:23.870+0700	info	otelcol/collector.go:266	Received shutdown request
2024-01-01T21:55:23.870+0700	info	service/service.go:185	Starting shutdown...
2024-01-01T21:55:24.034+0700	info	kafkareceiver/kafka_receiver.go:179	Consumer stopped	{"kind": "receiver", "name": "kafka", "data_type": "traces", "error": "context canceled"}
2024-01-01T21:55:24.039+0700	info	extensions/extensions.go:61	Stopping extensions...
2024-01-01T21:55:24.039+0700	info	service/service.go:199	Shutdown complete.
2024/01/01 21:55:24 assertion failures "Link.Attributes[messaging.system]": expected="kafka" actual=null
2024/01/01 21:55:24 assertion failures "Link.Attributes[messaging.destination]": expected="infrastructure-events-zone1" actual=null
2024/01/01 21:55:24 assertion failures "Link.Attributes[messaging.operation]": expected="receive" actual=null
2024/01/01 21:55:24 assertion failures "Link.Attributes[net.peer.ip]": expected="2600:1700:1f00:11c0:4de0:c223:a800:4e87" actual=null
2024/01/01 21:55:24 assertion failures "Link.Attributes[enduser.id]": expected="unittest" actual=null
2024/01/01 21:55:24 assertion failures "Link.Attributes[app.inretry]": expected=true actual=null
2024/01/01 21:55:24 assertion failures "Link.Attributes[app.progress]": expected=0.6 actual=null
2024/01/01 21:55:24 assertion failures "Link.Attributes[app.statemap]": expected="14|5|202" actual=null
2024/01/01 21:55:24 assertion failure count: {"Link.Attributes[app.inretry]":658,"Link.Attributes[app.progress]":658,"Link.Attributes[app.statemap]":658,"Link.Attributes[enduser.id]":1970,"Link.Attributes[messaging.destination]":1970,"Link.Attributes[messaging.operation]":1970,"Link.Attributes[messaging.system]":1970,"Link.Attributes[net.peer.ip]":1970}
```
How should we solve this? Does it make sense if we add attributes to our Span Reference model?

yurishkuro:
It's going to be pretty significant change to add attributes to SpanRefs. Is it possible to configure the test controller not to generate attributes for the links?

james-ryans:
> It's going to be pretty significant change to add attributes to SpanRefs. Is it possible to configure the test controller not to generate attributes for the links?

The only way to not generate attributes is not to generate a span link entirely. Are you okay with that?

yurishkuro:
Yes, that's fine - we're not testing model conversion here, but the ingestion pipeline integration, it doesn't matter much _what_ is in the spans.

Makefile
```diff
@@ -4,6 +4,8 @@
 SHELL := /bin/bash
 JAEGER_IMPORT_PATH = github.com/jaegertracing/jaeger
 STORAGE_PKGS = ./plugin/storage/integration/...
+OTEL_INTEGRATION_PATH = ./cmd/jaeger/integration/...
+GO = go
```
yurishkuro, line 8: already defined in L55

Makefile
```diff
@@ -141,6 +143,15 @@ index-rollover-integration-test: docker-images-elastic
 	go clean -testcache
 	bash -c "set -e; set -o pipefail; $(GOTEST) -tags index_rollover -coverpkg=./... -coverprofile cover-index-rollover.out $(STORAGE_PKGS) $(COLORIZE)"
 
+# Don't detect data race because testbed has race condition issue
```
yurishkuro, line 146: you mean the OTEL testbed? Can we fix it?

cmd/jaeger/internal/receivers/storagereceiver/factory.go
```diff
@@ -0,0 +1,33 @@
...
+
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/consumer"
+	"go.opentelemetry.io/collector/receiver"
+)
+
+// componentType is the name of this extension in configuration.
+const componentType = component.Type("jaeger_storage_receiver")
+
+func NewFactory() receiver.Factory {
```
yurishkuro, line 17: Sorry, this is confusing to me. It looks like you're trying to make it similar to `exporters/storageexporter`, but I don't understand why. The latter is needed to support a transition from OTEL pipeline into Jaeger storage world, where there could be any number of storage backends. Similar functionality is not needed in OTEL itself because their exporters are one-way (write), while Jaeger Storage is read/write.

However, on the receiver side there is no such situation. If we need to receive over Kafka, we can use official OTEL Kafka receiver directly in the pipeline, without any additional machinery from Jaeger.

Please elaborate.

cmd/jaeger/internal/extension/jaegerstorage/config.go
```diff
@@ -5,11 +5,13 @@ package jaegerstorage
 
 import (
 	memoryCfg "github.com/jaegertracing/jaeger/pkg/memory/config"
+	grpcCfg "github.com/jaegertracing/jaeger/plugin/storage/grpc/config"
 )
 
 // Config has the configuration for jaeger-query,
 type Config struct {
 	Memory map[string]memoryCfg.Configuration `mapstructure:"memory"`
+	GRPC   map[string]grpcCfg.Configuration   `mapstructure:"grpc-plugin"`
```
yurishkuro, line 14:
```suggestion
	GRPC   map[string]grpcCfg.Configuration   `mapstructure:"grpc"`
```

cmd/jaeger/internal/receivers/storagereceiver/factory.go
james-ryans: Actually, the `storagereceiver` is only used for integration test. The pipeline I used for the test is `collector -> kafka -> ingester -> remote storage` and the validator checks if the spans stored in remote storage match exactly with the provided data. Therefore, I need `storagereceiver` to retrieve spans from the remote storage for validation.

Here's the diagram of the integration test I created. The first test case verifies if the spans sent to Kafka are correct, and the second one checks the spans stored in the remote storage.
Untitled-2023-12-16-1017

Makefile
james-ryans: yes, it's OTEL testbed. I've tried to fix it but got stuck. Apparently, in the `opentelemetry-collector-contrib` repository, the OTEL testbed run the tests without detecting data race which I overlooked before trying to use it. I also opened a discussion to about this (https://github.com/open-telemetry/opentelemetry-collector-contrib/discussions/30287) but still got no answer.

The main problem is that when we create a new testbed testcase with NewTestCase(), it starts a goroutine that log stats of the data sent (LoadGenerator) dan received (MockBackend) (https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/testbed/testbed/test_case.go#L126C18-L126C18) but later we must start a MockBackend which it write a startTime variable that is used by the log stats goroutine hence the data race occur.

cmd/jaeger/internal/receivers/storagereceiver/receiver.go
```diff
@@ -0,0 +1,139 @@
...
+			host.ReportFatalError(err)
+		}
+	}()
+
+	return nil
+}
+
+func (r *storageReceiver) consumeLoop(ctx context.Context) error {
+	// golden data provider can produce one of these service names
+	services := []string{"", "customers", "OTLPResourceNoServiceName"}
```
yurishkuro, line 74: I think it would be more robust if you just loaded services from SpanReader, in case the testbed changes its behavior

cmd/jaeger/internal/receivers/storagereceiver/factory.go
yurishkuro: In this case could we move it under integration tests? I see that it is not included in the Components() output, so why does it need to be in jaeger/internal? The reason I don't like it is it looks like another production component. but it is not. 

Also, please add READMEs to new components, and save these awesome diagrams in them!

yurishkuro (changes requested):
tagging as needing followup

cmd/jaeger/internal/receivers/storagereceiver/factory.go
james-ryans: Absolutely! I've moved it under the integration test and added the README.

james-ryans:
> tagging as needing followup

Hi, I've raised a PR to fix the OTEL testbed data race and currently waiting for their review here https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/30549.

james-ryans:
Hi @yurishkuro, sorry, this PR became messy as I tried to solve the conflict merge with rebases, some failing CIs, an unsigned commit, and now I'm stuck with Build binaries CI. Can you help me find out what's wrong with my commits? I've successfully run `make build-binaries-linux` locally but somehow it is failing in CI.

Locally
```
+ git log --oneline --decorate=full -n 10
+ cat
53976a5e (HEAD, refs/remotes/origin/main, refs/remotes/origin/HEAD) Bump follow-redirects from 1.15.1 to 1.15.4 (#2108)
9be986e7 Bump actions/upload-artifact from 3.1.0 to 4.0.0 (#2106)
9bb97358 (tag: refs/tags/v1.37.0) Prepare release v1.37.0 (#2097)
a6fb5824 Setup OSSF Scorecard workflow  (#2096)
039d8d6d Bump classnames from 2.3.3 to 2.5.1 (#2085)
b574d7a9 Bump the eslint group with 2 updates (#2095)
9fadfff1 Fix broken Hot Reload of Plexus Package (#2089)
868bb058 Use fake/fixed date in unit tests (#2091)
911ccf54 feat: remove `is-promise` library (#2080)
cdd251ba feat: remove `ReactGA` and migrate to `GA4` for tracking (#2071)
++ git describe --tags --dirty
+ last_tag=v1.37.0-2-g53976a5e
+ [[ v1.37.0-2-g53976a5e =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]
+ yarn install --frozen-lockfile
yarn install v1.22.18
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
[3/4] 🔗  Linking dependencies...
...
```

CI
```
++ git describe --tags --dirty
+ last_tag=
make[2]: *** [Makefile:242: rebuild-ui] Error 128
make[2]: Leaving directory '/home/runner/work/jaeger/jaeger'
make[1]: *** [Makefile:238: jaeger-ui/packages/jaeger-ui/build/index.html] Error 2
make[1]: Leaving directory '/home/runner/work/jaeger/jaeger'
make: *** [Makefile:352: build-binaries-linux] Error 2
```

yurishkuro:
The next collector release will be around Feb 11.

> Should I create a CI for this kafka integration test?

Yes. Thanks.

james-ryans:
> The next collector release will be around Feb 11.

So, do we wait until the next official opentelemetry-collector-contrib is released?

james-ryans:
Sorry, the codecov CI is failing due to lots of uncovered lines in the `cmd/jaeger/integration` directory (mostly error returns that never occur and some unused functions due to the interface contract). However, I don’t see any benefit for testing on these files since these are packages to help the e2e test. What is the best way to solve this? Should I ignore these errors by replacing them with `_` and write `panic("not implemented");` line to the unused functions? Or Is it better not to upload the coverage file? Or maybe there is another workaround for this?

plugin/storage/grpc/factory_test.go
```diff
@@ -143,6 +143,13 @@ func TestGRPCStorageFactory(t *testing.T) {
 	assert.Equal(t, f.store.DependencyReader(), depReader)
 }
 
+func TestGRPCStorageFactoryWithConfig(t *testing.T) {
+	cfg := grpcConfig.Configuration{}
+	_, err := NewFactoryWithConfig(cfg, metrics.NullFactory, zap.NewNop())
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "grpc-plugin builder failed to create a store: error connecting to remote storage")
```
yurishkuro, line 150:
```suggestion
	assert.ErrorContains(t, err, "grpc-plugin builder failed to create a store: error connecting to remote storage")
```

yurishkuro:
> codecov CI is failing due to lots of uncovered lines in the cmd/jaeger/integration directory

we should be able to control it with .nocover files, but perhaps those only work for the main test script. Another alternative is to add the dir to .codecov.yaml as exclusion

james-ryans:
> we should be able to control it with .nocover files, but perhaps those only work for the main test script. Another alternative is to add the dir to .codecov.yaml as exclusion

Yes, I've added the .nocover files but it doesn't take any effect. I'll add an ignore to that dir in .codecov.yaml.

plugin/storage/grpc/factory_test.go
yurishkuro: the require.ErrorContains

yurishkuro:
does this release have what you need? #5143

plugin/storage/grpc/factory_test.go
james-ryans: Ohh, silly me.. Okay, fixed it.

james-ryans:
> does this release have what you need? #5143

Wow, yes, it has! I've updated the go.mod but encountered an issue that the OTEL testcase is not closing.. I'll get back to you after I've found and fixed the issue. 😄

jkowall (approved):
Looks good, thanks for taking this on Yuri!

yurishkuro:
@jkowall I'm just reviewing, this is a community contribution from @james-ryans

.github/workflows/ci-kafka.yml
```diff
@@ -18,6 +18,10 @@ permissions:  # added using https://github.com/step-security/secure-workflows
 jobs:
   kafka:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        name: [v1, v2]
```
yurishkuro, line 23: will `name` be also the name of the job? Or do you need to add a `name: ` directive using this var?

yurishkuro: nit: I would use `version` instead of `name` if it's just a var

james-ryans: > nit: I would use `version` instead of `name` if it's just a var

Yes, it is just a var. I'll change that.

james-ryans:
I think the issue in our CIT kafka v2 is because of OTEL kafka receiver v0.93.0 bug, the receiver is shutting down indefinitely. I've raised an issue on OTEL contrib repository here https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/30789. If it is actually the OTEL contrib bug, we'll need to wait until the next release again.

yurishkuro:
@james-ryans the ticket you referenced sounds like a deadlock issue during abnormal conditions, not when everything is running well, is that not true? Because if it's actually blocking normal operations it seems like a much bigger issue in OTEL.

james-ryans:
> @james-ryans the ticket you referenced sounds like a deadlock issue during abnormal conditions, not when everything is running well, is that not true? Because if it's actually blocking normal operations it seems like a much bigger issue in OTEL.

I'm not sure, but I think this actually is a big issue? 🤔 They have labeled the ticket to `priority:p1` since the deadlock always happens 100% when it's shutting down, there is no workaround to avoid this.

Or what you meant by "running well" is that when the collector never shut down? If yes, we can avoid this bug by not shutting them down after the test cases have been tested.

yurishkuro:
I didn't get into the meat of the issue, got the impression that deadlock happens when something first happens during startup. Otherwise how do any of their unit tests work, do they not test shutdown?

cmd/jaeger/ingester-with-remote.yaml
```diff
@@ -0,0 +1,31 @@
...
+      memstore:
+        server: localhost:17271
+        connection-timeout: 5s
+
+receivers:
+  kafka:
+    brokers:
+      - localhost:9092
+    encoding: otlp_proto # available encodings are otlp_proto, jaeger_proto, jaeger_json, zipkin_proto, zipkin_json, zipkin_thrift
+    initial_offset: earliest # consume messages from the beginning
```
yurishkuro, line 24: perhaps we should parameterize this and default to latest checkpoint rather than earliest, since it would be very bad to run with earliest in production. The integration tests can override the value via env var.

cmd/jaeger/ingester-with-remote.yaml
```diff
@@ -0,0 +1,31 @@
...
+      processors: [batch]
+      exporters: [jaeger_storage_exporter]  # same as in the default cmd/jaeger-v2/config.yaml
+  telemetry:
+    metrics:
+      address: 0.0.0.0:8889 # to avoid port conflict with collector-with-kafka.yaml
+
+extensions:
+  jaeger_storage:
+    grpc:
+      memstore:
```
yurishkuro, line 15: nit: we don't know it's a `memstore` in this config, let's call it 'external-storage'

cmd/jaeger/integration/receivers/storagereceiver/factory.go
```diff
@@ -0,0 +1,33 @@
...
+)
+
+// componentType is the name of this extension in configuration.
+const componentType = component.Type("jaeger_storage_receiver")
+
+func NewFactory() receiver.Factory {
+	return receiver.NewFactory(
+		componentType,
+		createDefaultConfig,
+		receiver.WithTraces(createTraces, component.StabilityLevelDevelopment),
```
yurishkuro, line 21:
```suggestion
		receiver.WithTraces(tracesReceiver, component.StabilityLevelDevelopment),
```

cmd/jaeger/integration/receivers/storagereceiver/factory.go
```diff
@@ -0,0 +1,33 @@
...
+	)
+}
+
+func createDefaultConfig() component.Config {
+	return &Config{}
+}
+
+func createTraces(ctx context.Context, set receiver.CreateSettings, config component.Config, nextConsumer consumer.Traces) (receiver.Traces, error) {
```
yurishkuro, line 29:
```suggestion
func tracesReceiver(ctx context.Context, set receiver.CreateSettings, config component.Config, nextConsumer consumer.Traces) (receiver.Traces, error) {
```

cmd/jaeger/integration/receivers/storagereceiver/receiver.go
```diff
@@ -0,0 +1,142 @@
...
+func newReceiver(config *Config, otel component.TelemetrySettings, nextConsumer consumer.Traces) (*storageReceiver, error) {
+	f, err := grpc.NewFactoryWithConfig(
+		config.GRPC,
+		metrics.NullFactory,
+		otel.Logger,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to init storage factory: %w", err)
+	}
+	// TODO add support for other backends
```
yurishkuro, line 43: Can't we use jaegerextension here, same as `storageexporter`?

cmd/jaeger/integration/receivers/storagereceiver/receiver.go
```diff
@@ -0,0 +1,142 @@
...
+
+func (r *storageReceiver) consumeTraces(ctx context.Context, serviceName string) error {
+	traces, err := r.spanReader.FindTraces(ctx, &spanstore.TraceQueryParameters{
+		ServiceName: serviceName,
+	})
+	if err != nil {
+		return err
+	}
+
+	cnt := 0
```
yurishkuro, line 100: ?

cmd/jaeger/integration/receivers/storagereceiver/receiver.go
```diff
@@ -0,0 +1,142 @@
...
+		if err := r.consumeLoop(ctx); err != nil {
+			host.ReportFatalError(err)
+		}
+	}()
+
+	return nil
+}
+
+func (r *storageReceiver) consumeLoop(ctx context.Context) error {
+	for {
```
yurishkuro, line 73: how do we exit this loop? Should it be checking `ctx.Done` for that?

cmd/jaeger/integration/receivers/storagereceiver/receiver.go
```diff
@@ -0,0 +1,142 @@
...
+			r.logger.Error("Failed to get services from consumer", zap.Error(err))
+			return err
+		}
+
+		for _, svc := range services {
+			if err := r.consumeTraces(ctx, svc); err != nil {
+				r.logger.Error("Failed to consume traces from consumer", zap.Error(err))
+			}
+			if ctx.Err() != nil {
+				r.logger.Error("Consumer stopped", zap.Error(ctx.Err()))
```
yurishkuro, line 85: Why is this an error for the consumer?

cmd/jaeger/integration/receivers/storagereceiver/receiver.go
```diff
@@ -0,0 +1,142 @@
...
+	cnt := 0
+	for _, trace := range traces {
+		cnt += len(trace.Spans)
+		traceID := trace.Spans[0].TraceID
+		if _, ok := r.consumedTraces[traceID]; !ok {
+			r.consumedTraces[traceID] = &consumedTrace{
+				spanIDs: make(map[model.SpanID]struct{}),
+			}
+		}
+		if len(trace.Spans) > len(r.consumedTraces[traceID].spanIDs) {
```
yurishkuro, line 109: is this check important? the next function will skip spans that are already received

cmd/jaeger/integration/receivers/storagereceiver/receiver.go
```diff
@@ -0,0 +1,142 @@
...
+		}
+
+		for _, svc := range services {
+			if err := r.consumeTraces(ctx, svc); err != nil {
+				r.logger.Error("Failed to consume traces from consumer", zap.Error(err))
+			}
+			if ctx.Err() != nil {
+				r.logger.Error("Consumer stopped", zap.Error(ctx.Err()))
+				return ctx.Err()
+			}
```
yurishkuro, line 87: would it make sense to sleep here for a bit (configurable), rather than continuously crunching through the same spans?

cmd/jaeger/integration/receivers/storagereceiver/config.go
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) 2024 The Jaeger Authors.
```
yurishkuro, line 1: please add a README to this package, e.g.

`storagereceiver` is a fake receiver that creates an artificial stream of traces by:
  * repeatedly querying one of Jaeger storage backends for all traces (by service)
  * tracking new traces / spans and passing them to the next component in the pipeline.

cmd/jaeger/integration/receivers/storagereceiver/config.go
```diff
@@ -0,0 +1,19 @@
...
+package storagereceiver
+
+import (
+	"github.com/asaskevich/govalidator"
+
+	grpcCfg "github.com/jaegertracing/jaeger/plugin/storage/grpc/config"
+)
+
+type Config struct {
+	GRPC grpcCfg.Configuration `mapstructure:"grpc"`
```
yurishkuro, line 13: similar to storage exporter, this should be just a name of the storage that can be retrieved from `jaeger_storage_extension`

cmd/jaeger/internal/extension/jaegerstorage/extension.go
```diff
@@ -88,6 +89,20 @@ func (s *storageExt) Start(ctx context.Context, host component.Host) error {
 		s.factories[name] = factory
 	}
 
+	for name, g := range s.config.GRPC {
```
yurishkuro, line 92: Suggestion: could you extract changes to `jaegerstorage` into a new PR that we can merge separately? It's largely independent of what's blocking the integration tests.

cmd/jaeger/integration/kafka_test.go
```diff
@@ -0,0 +1,100 @@
...
+		factories, err := internal.Components()
+		require.NoError(t, err, "default components resulted in: %v", err)
+
+		runner := testbed.NewInProcessCollector(factories)
+		validator := testbed.NewCorrectTestValidator(sender.ProtocolName(), test.receiver.ProtocolName(), dataProvider)
+
+		config, err := os.ReadFile(test.configFile)
+		if err != nil {
+			t.Fatal(err)
+		}
```
yurishkuro, line 62: require.NoError(t, err)

cmd/jaeger/integration/kafka_test.go
```diff
@@ -0,0 +1,100 @@
...
+		configCleanup, err := runner.PrepareConfig(string(config))
+		require.NoError(t, err, "collector configuration resulted in: %v", err)
```
yurishkuro, line 64:
```suggestion
		require.NoError(t, err)
```

cmd/jaeger/integration/kafka_test.go
```diff
@@ -0,0 +1,100 @@
...
+		defer tc.Stop()
+
+		tc.EnableRecording()
+		tc.StartBackend()
+		tc.StartAgent()
+
+		load := i == 0
+		if load {
+			tc.StartLoad(testbed.LoadOptions{
+				DataItemsPerSecond: 16,
```
yurishkuro, line 84: Does this mean there's no limit on the number of traces the generator creates? Seems like it would result in a random amount in each test. Not necessarily a problem as long as it doesn't cause flakiness.

cmd/jaeger/integration/kafka_test.go
```diff
@@ -0,0 +1,100 @@
...
+				ItemsPerBatch:      1,
+			})
+		}
+		tc.Sleep(5 * time.Second)
+		if load {
+			tc.StopLoad()
+		}
+
+		tc.WaitForN(func() bool { return tc.LoadGenerator.DataItemsSent() == tc.MockBackend.DataItemsReceived() },
+			10*time.Second, "all data items received")
```
yurishkuro, line 94:
```suggestion
		tc.WaitForN(func() bool { 
				return tc.LoadGenerator.DataItemsSent() == tc.MockBackend.DataItemsReceived()
			},
			10*time.Second,
			"all data items received")
```

cmd/jaeger/integration/receivers/storagereceiver/.nocover
```diff
@@ -0,0 +1 @@
+FIXME
```
yurishkuro, line 1: may want to pull storagereceiver into a separate PR as well, and add the tests to it (most other v2 packages already have tests)

yurishkuro (commented):
please rebase on top of #5171

james-ryans:
Hi @yurishkuro, thank you for all of this invaluable feedback! I just recovered from dental implant post-OP fever, I'll get back to this issue as soon as I can. Sorry if the revision takes long enough. 😄

yurishkuro:
@james-ryans was the upstream issue in OTEL resolved in their latest release?

james-ryans:
> @james-ryans was the upstream issue in OTEL resolved in their latest release?

Not yet. I've tried @v0.95.0, but I still encountered issues with terminating the Kafka receiver. I haven't fully understood the root cause yet. The issue they referenced occurs at startup, but our issue happens on shutdown, which is confusing. I plan to help pinpoint and possibly fix the problem or find out if there is a workaround after I finish refactoring this PR based on your feedbacks.

cmd/jaeger/integration/kafka_test.go
james-ryans: It have a limit, it will stop once it has used all the combinations of traces and spans from the fixtures. It consistently stops at 3,524 traces every time I run the test.

cmd/jaeger/integration/receivers/storagereceiver/receiver.go
james-ryans: Oops, sorry, seems I have changed the implementation but forgot to remove that variable.

cmd/jaeger/integration/receivers/storagereceiver/receiver.go
james-ryans: Yes, I think we can. I'll change the implementation in a new PR of this receiver.

cmd/jaeger/integration/receivers/storagereceiver/receiver.go
james-ryans: I actually follow kafkareceiver from otelcol-contrib since they're both a "pull" strategy to retrieve traces. https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/3652b444802fad4819b50fbda41e0957000d8279/receiver/kafkareceiver/kafka_receiver.go#L171-L185

cmd/jaeger/integration/receivers/storagereceiver/receiver.go
james-ryans: Hmm... my thought on this logic is that it will reduce the time complexity because I won't need to loop through every span in the trace if it's fully stored. For example, if the average trace contains 10 spans, the time complexity would be reduced by 10x for the next iteration. Fun fact: I've just compared it, and with this check, it improves the ./scripts/otel-kafka-integration-test.sh runtime by 1-2 seconds. Do you think it's somewhat overkill for an artificial receiver?

cmd/jaeger/integration/receivers/storagereceiver/receiver.go
james-ryans: Ahh.. yes. I'll add that at the new PR.

james-ryans:
> how do you feel about this approach?

I think this approach is now outdated, same as the last discussion we did at https://github.com/jaegertracing/jaeger/issues/5254#issuecomment-2016858385, it only solves the (1) requirement. This PR only checks if the span is correctly sent to Kafka and stored in remote storage.

> Do you think we still need it, or could we also cover Kafka as part of the same v2 e2e tests?

I don't think this PR is relevant to the current solution anymore. We should change this to the proposed solution as the other v2 e2e tests.

PR #4980, opened 2023-12-01
Fix example/grafana-integration

angristan:

Currently, the `grafana-integration` example doesn't work properly: if you run `docker-compose up` in that folder, services will start but only logging will work, the metrics and tracing won't.

## Description of the changes

* Fix traces endpoint config for hotrod (`traces export: Post "http://localhost:4318/v1/traces": dial tcp 127.0.0.1:4318: connect: connection refused`)
* Fix hotrod metrics scraping (endpoint has moved to the frontend service)
* Pin Jaeger/hotrod tags to prevent future issues
* Fix Grafana dashboard (metric names, labels, migrate to new time series panel)
* Add default Grafana credentials to README
* Fix the loki container being stuck on shutdown by setting shorter timeouts (bug with the driver: https://github.com/grafana/loki/issues/2361#issuecomment-718024318)
* Update Grafana/Loki/Prom pinned versions (especially to get Grafana UI improvements)

`docker-compose up` 🙂 

SCR-20231201-cohy
SCR-20231201-cxos
SCR-20231201-coke

## Checklist

### DIFF
--- a/examples/grafana-integration/README.md
+++ b/examples/grafana-integration/README.md
@@ -20,11 +20,11 @@ grafana/loki-docker-driver:latest \
 
 ### Run the services
 
-`docker-compose up` 
+`docker-compose up`
 
 ### Access the services
 * HotROD application at http://localhost:8080
-* Grafana UI at http://localhost:3000
+* Grafana UI at http://localhost:3000 (default credentials: admin/admin)
 
 ### Explore with Loki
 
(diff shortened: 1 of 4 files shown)


yurishkuro:
Thanks for the fixes. 

Could you elaborate on the value you're getting from this example? It seems 99% about Grafana, Jaeger almost not needed or used. I am tempted to delete this example altogether.

angristan:
Sure :) Well, I would agree Prometheus + Loki might not be relevant in this context indeed. We could probably delete them from the example to keep it leaner.
However, the value I got from it was to see how Grafana could use Jaeger as a data source, I'm pretty pleased with the result.

I'm happy to rework the MR and remove Loki and Prometheus, but feel free to close it if you want to delete it altogether 👍

PR #4985, opened 2023-12-03
[SPM] Differentiate null from no error data

### LINKED ISSUE #4229: [Feature]: SPM Provide option to fill null gaps with zero in monitor tab
### Requirement

Currently, if there is no error in a traces(ie. no error=true tag in spans), otel-collector doesn't send any error metrics to prometheus (i.e. it doesn't create any metric with prometheus label as status_code = "STATUS_CODE_ERROR"). This generates empty dashboard on Jaeger monitor tab.

### Problem

It creates empty dashboards on monitor tab which can create confusion whether:
- data is not being ingested through the pipeline
- data has no error

### Proposal

In order to prevent confusion, it will be better if instead of showing empty dashboard, we show dashboard containing data value as zero.

cc @albertteoh

albertteoh:
- Resolves #4229

## Description of the changes
- Ensures SPM displays a 0% error rate if there are no error metrics _and_ call rates exist.
- If call rates don't exist, the error rate will also be null.
- This ensures SPM is able to differentiate "no data" from "no errors".

- Add unit tests to cover happy and error cases.
- Tested locally to confirm "No data" is shown in the Error graph when there is no data, then when call rates are available, a 0% rate is displayed.

Screenshot 2023-12-03 at 8 01 36 pm

Screenshot 2023-12-03 at 8 00 45 pm

## Checklist


docker-compose/monitor/Makefile
```diff
@@ -2,7 +2,7 @@
 build: export DOCKER_TAG = dev
 build: clean-jaeger
 	cd ../../ && \
-	make build-all-in-one && \
+	make build-all-in-one-linux && \
```
albertteoh, line 5: The docker images rely on linux binaries.

plugin/metrics/prometheus/metricsstore/reader.go
```diff
@@ -211,7 +211,38 @@ func (m MetricsReader) GetErrorRates(ctx context.Context, requestParams *metrics
 			)
 		},
 	}
-	return m.executeQuery(ctx, metricsParams)
+	errorMetrics, err := m.executeQuery(ctx, metricsParams)
```
albertteoh, line 214: This is the main change, the rest supports this change, mostly for testing purposes.

yurishkuro (approved):

docker-compose/monitor/README.md
```diff
@@ -62,13 +62,13 @@ make build
 ## Bring up the dev environment
 
 ```bash
-make run-dev
+make dev
```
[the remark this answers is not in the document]
albertteoh: Interesting idea, thanks. I'll have a think about this and put together a separate PR for it.

PR #4989, opened 2023-12-04
[ci]:Improve Kafka integration test self-sufficiency

### LINKED ISSUE #4982: [ci] Make Kafka integration test more self-siufficient
The test script `scripts/kafka-integration-test.sh` is currently very simple, it could use some improvements:
  * Instead of starting Kafka in the CI it could be started from the script when an extra `-k` parameter is passed. This makes it easier to run the test locally. Example of starting Kafka manually in #4981
  * The CI workflow `.github/workflows/ci-kafka.yml` starts both Kafka and ZK. Is ZK actually needed with the latest versions of Kafka?
  * The script has an infinite `while true` loop waiting for Kafka to start - if should instead use a loop counter and be time bound, say no more than 3min, with checks every 5sec or so

RipulHandoo:
Resolves #4982 

## Description of the changes
- Introduced a -k parameter to the kafka-integration-test.sh script.
- Modified the script to wait for Kafka availability before proceeding.

## Checklist

### DIFF
--- a/.github/workflows/ci-kafka.yml
+++ b/.github/workflows/ci-kafka.yml
@@ -31,7 +31,7 @@ jobs:
         go-version: 1.21.x
 
     - name: Run kafka integration tests
-      run: bash scripts/kafka-integration-test.sh
+      run: bash scripts/kafka-integration-test.sh -k
 
     - name: Output Kafka logs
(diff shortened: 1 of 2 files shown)


yurishkuro:
Thanks!

PR #4995, opened 2023-12-10
Remove custom GOCACHE location

yurishkuro:
- Even though we're using setup-go v4.1 where caching is supposed to be enabled by default, it does not seem to work as the dependencies are still being downloaded when we run the tests

## Description of the changes
- remove custom GOCACHE
- point to go.sum file

- CI
- the change itself did not seem to help, still seeing lots of "downloading" when running the tests
- needs more investigation, but I suggest we still merge this as an improvement

### DIFF
--- a/.github/workflows/ci-unit-tests.yml
+++ b/.github/workflows/ci-unit-tests.yml
@@ -26,9 +26,10 @@ jobs:
 
     - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11
 
-    - uses: actions/setup-go@93397bea11091df50f3d7e59dc26a7711a8bcfbe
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-unit-tests.yml
```diff
@@ -26,9 +26,10 @@ jobs:
 
     - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11
 
-    - uses: actions/setup-go@93397bea11091df50f3d7e59dc26a7711a8bcfbe
+    - uses: actions/setup-go@93397bea11091df50f3d7e59dc26a7711a8bcfbe # v4.1.0
```
yurishkuro, line 29: sometimes dependabot adds version at the end, but not everywhere. I wonder if adding it manually once will cause it to keep adding it.

jkowall (approved):
Looks good, if not we can always revert :)

PR #5000, opened 2023-12-11
Add dependency version numbers as comments

yurishkuro:
- We import most GH actions via commit hash (for repeatable builds), but it's hard to tell which version
- Dependabot can add this version as a comment, but only when it's already there

## Description of the changes
- Add current versions as comments (functionally no-op)
- Upgrade checkout action in two workflows that were out of sync with the rest of the workflows

- CI


jkowall (approved):
Good improvement for version tracking! lgtm.

PR #5008, opened 2023-12-15
Make all-in-one metric names match metrics from standalone components

### LINKED ISSUE #5007: [Bug]: Jaeger prometheus metrics are missing "collector" prefix?
### What happened?

The jaeger docs [here](https://www.jaegertracing.io/docs/1.52/performance-tuning/) as well as sample Grafana dashboard [here](https://github.com/jaegertracing/jaeger/blob/b5f73aede2d76249cbd6d9855d46ed158c0419f9/monitoring/jaeger-mixin/dashboard-for-grafana.json#L573) reference a metric `jaeger_collector_queue_length`. However, when running the latest jaeger image `jaegertracing/all-in-one:1.52`, the "collector" prefix seems to be missing - it only exports a `jaeger_queue_length` metric. So it seems like either the docs/grafana dashboard are wrong, or there's a bug where the "collector" prefix is missing on the exported metric? Or maybe these are actually different metrics and the "jaeger_collector_*" metrics are just missing?

Here are all the `jaeger_collector_*` metrics that are referenced from the Grafana dashboard:

```
jaeger_collector_in_queue_latency_bucket
jaeger_collector_queue_length
jaeger_collector_spans_dropped_total
jaeger_collector_spans_received_total
```

Unless these are actually different metrics, it looks like the actual metric names that are exported are:

```
jaeger_in_queue_latency_bucket
jaeger_queue_length
jaeger_spans_dropped_total
jaeger_spans_received_total
```

### Steps to reproduce

Run the latest all-in-one image:

```
$ docker run --rm -it --net=host jaegertracing/all-in-one:1.52
```

Check which metrics are exported:

```
$ curl -fsSL http://localhost:14269/metrics | grep queue_length
jaeger_queue_length{host="..."}
```

Observe that the metric is called "jaeger_queue_length", not "jaeger_collector_queue_length" as referenced in both docs and sample Grafana dashboards.

### Expected behavior

The metric should either be called `jaeger_collector_queue_length` or the docs + sample Grafana dashboard should reference the actual metric value `jaeger_queue_length`, depending on which is the "correct" / intended value.

### Relevant log output

_No response_

### Screenshot

_No response_

### Additional context

_No response_

### Jaeger backend version

1.52

### SDK

_No response_

### Pipeline

_No response_

### Stogage backend

_No response_

### Operating system

_No response_

### Deployment model

_No response_

### Deployment configs

_No response_

yurishkuro:
- Resolves #5007

## Description of the changes
- provide proper namespaces for different sub-components (agent, collector, query) running in all-in-one, to match the metrics names produces by these components when running standalone

* `go run ./cmd/all-in-one`
* `curl http://localhost:14269/metrics | grep -v '#'`
* compare before/after [textcompare-diff.pdf](https://github.com/jaegertracing/jaeger/files/13689954/textcompare-diff.pdf)

### DIFF
--- a/cmd/all-in-one/main.go
+++ b/cmd/all-in-one/main.go
@@ -95,10 +95,13 @@ by default uses only in-memory database.`,
 				return err
 			}
 			logger := svc.Logger // shortcut
-			metricsFactory := fork.New("internal",
+			baseFactory := fork.New("internal",
 				expvar.NewFactory(10), // backend for internal opts
 				svc.MetricsFactory.Namespace(metrics.NSOptions{Name: "jaeger"}))
-			version.NewInfoMetrics(metricsFactory)
+			version.NewInfoMetrics(baseFactory)
+			agentMetricsFactory := baseFactory.Namespace(metrics.NSOptions{Name: "agent"})
+			collectorMetricsFactory := baseFactory.Namespace(metrics.NSOptions{Name: "collector"})
+			queryMetricsFactory := baseFactory.Namespace(metrics.NSOptions{Name: "query"})
 
 			tracer, err := jtracer.New("jaeger-all-in-one")
 			if err != nil {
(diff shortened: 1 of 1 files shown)


albertteoh (approved):
LGTM. I wonder if it's easy to write a unit test to validate the namespaces attached to a metrics factory to prevent future regressions?

yurishkuro:
main isn't easily testable without serious refactoring. This will be better in jaeger-v2

PR #5010, opened 2023-12-17
Feature: add goleak to check goroutine leak in tests

### LINKED ISSUE #5006: Enable and enforce goroutine leak checks in tests
Enforce the use of https://github.com/uber-go/goleak across the repo:
 * [x] a linter similar to scripts/check-test-files.sh that verifies that all packages with tests have goleak in TestMain
   * first deploy the script in audit mode where it only prints offending packages
   * then make the script fail the CI once the next item is addressed
   * Done in #5010
 * [x] add goleak to all packages via TestMain. Some packages may pass the validation, some may fail and would need to be fixed. It's best to break this into several PRs, first to add goleak to packages that do not fail, and then potentially separate PRs for each package where goleak does fail with the corresponding fixes to the tests.
   * Run `make goleak` to see the list of packages missing the linter
   * In packages with multiple `*_test.go` files add the check to a new file `package_test.go`
     * Remember to add copyright header - run `make fmt` after committing the new files
   * Examples: 
     * #5015 
     * #5017
 * [x] Fix leaks in the integration tests (see #6339)
 * [x] Change linter script to fail CI on errors going forward

State on March 8 2024:
```
$ make goleak
🔴 Error(check-goleak): no goleak check in package ./cmd/collector/app/
🔴 Error(check-goleak): no goleak check in package ./cmd/internal/printconfig/
🔴 Error(check-goleak): no goleak check in package ./cmd/query/app/
🔴 Error(check-goleak): no goleak check in package ./pkg/cassandra/gocql/testutils/
🔴 Error(check-goleak): no goleak check in package ./pkg/testutils/
🔴 Error(check-goleak): no goleak check in package ./plugin/metrics/prometheus/metricsstore/
🔴 Error(check-goleak): no goleak check in package ./plugin/sampling/strategystore/adaptive/
Error(check-goleak): no goleak check in 8 package(s).
```

akagami-harsh:
- Partially Fixes #5006 

## Description of the changes
- added a linter to check implementation of goleak in tests
- added go leak in some tests 

- 

## Checklist


akagami-harsh:
@yurishkuro, i have goleak to some packages (that do not fail) and added a linter to verify goleak is present in tests. just want some pointers if i am on the right track.

scripts/check-goleak-files.sh
```diff
@@ -0,0 +1,29 @@
...
+
+NO_GOLEAK_FILE_DIRS=""
+
+# shellcheck disable=SC2048
+for dir in $*; do
+  testFiles=$(find "${dir}" -maxdepth 1 -name '*_test.go')
+  for test in ${testFiles}; do
+    if grep -q "TestMain" "${test}" | grep -q "goleak" "${test}"; then
+      break
+    elif [[ ${test} == "/" ]]; then
```
yurishkuro, line 13: When would this match if you're finding go files only?

scripts/check-goleak-files.sh
```diff
@@ -0,0 +1,29 @@
...
+set -euo pipefail
+
+NO_GOLEAK_FILE_DIRS=""
+
+# shellcheck disable=SC2048
+for dir in $*; do
+  testFiles=$(find "${dir}" -maxdepth 1 -name '*_test.go')
+  for test in ${testFiles}; do
+    if grep -q "TestMain" "${test}" | grep -q "goleak" "${test}"; then
+      break
```
yurishkuro, line 12: Since break does early exit there is no need for elseif

scripts/check-goleak-files.sh
```diff
@@ -0,0 +1,29 @@
...
+    elif [[ ${test} == "/" ]]; then
+      continue
+    else
+      if [ -z "${NO_GOLEAK_FILE_DIRS}" ]; then
```
yurishkuro, line 16: Don't think you need this if you treat the var as array. See other scripts where we do that.

scripts/check-goleak-files.sh
```diff
@@ -0,0 +1,29 @@
...
+        NO_GOLEAK_FILE_DIRS="${NO_GOLEAK_FILE_DIRS} ${dir}"
+      fi
+    fi
+  done
+done
+
+if [ -n "${NO_GOLEAK_FILE_DIRS}" ]; then
+  echo "*** directories without Goleak implemented:" >&2
+  echo "${NO_GOLEAK_FILE_DIRS}" | tr ' ' '\n' | sort -u >&2
+  exit 1
```
yurishkuro, line 28: I don't think you want error code at this point while there are packages without goleak. Once we fix them all then we can add error code, for now just print.

scripts/check-goleak-files.sh
```diff
@@ -0,0 +1,29 @@
...
+      if [ -z "${NO_GOLEAK_FILE_DIRS}" ]; then
+        NO_GOLEAK_FILE_DIRS="${dir}"
+      else
+        NO_GOLEAK_FILE_DIRS="${NO_GOLEAK_FILE_DIRS} ${dir}"
+      fi
+    fi
+  done
+done
+
+if [ -n "${NO_GOLEAK_FILE_DIRS}" ]; then
```
yurishkuro, line 25: I don't think you need to accumulate packages in a variety of, just print errors as you're finding them and only keep a counter that you can check at the end.

Makefile
```diff
@@ -163,6 +163,11 @@ nocover:
 	@echo Verifying that all packages have test files to count in coverage
 	@scripts/check-test-files.sh $(ALL_PKGS)
 
+.PHONY: goleak
+goleak:
```
yurishkuro, line 167: Add this target as a dependency of lint (see how nocover target is called)

go.mod
```diff
@@ -221,6 +221,7 @@ require (
 	go.opentelemetry.io/otel/exporters/stdout/stdoutmetric v0.44.0 // indirect
 	go.opentelemetry.io/otel/sdk/metric v1.21.0 // indirect
 	go.opentelemetry.io/proto/otlp v1.0.0 // indirect
+	go.uber.org/goleak v1.3.0 // indirect
```
yurishkuro, line 224: Why is it indirect? Try go mod tidy.

scripts/check-goleak-files.sh
```diff
@@ -0,0 +1,29 @@
...
+
+set -euo pipefail
+
+NO_GOLEAK_FILE_DIRS=""
+
+# shellcheck disable=SC2048
+for dir in $*; do
+  testFiles=$(find "${dir}" -maxdepth 1 -name '*_test.go')
+  for test in ${testFiles}; do
+    if grep -q "TestMain" "${test}" | grep -q "goleak" "${test}"; then
```
yurishkuro, line 11: I think you can just check for "goleak.VerifyTestMain"

yurishkuro (commented):
Thanks, looks good as a direction.

yurishkuro:
Let's fix the comments but not add any more packages, I prefer to merge smaller PRs.

scripts/check-goleak-files.sh
akagami-harsh: if some package have multiple tests in it then the package will be printed multiple times, in that way we can't use `sort -u` to filter out repeated packages

yurishkuro: I think that just points to an error in the logic: if the first test file in package contains goleak, you break out, otherwise you add the package to the list, even though a later file in that package might contain goleak

scripts/check-goleak-files.sh
akagami-harsh: yes, removed error code for now

scripts/check-goleak-files.sh
akagami-harsh: done

scripts/check-goleak-files.sh
akagami-harsh: fixed it like this `elif [[ ${dir} == "./"]]; then` 
i add this because there is dummy test present in the root directory

scripts/check-goleak-files.sh
akagami-harsh: removed it!

scripts/check-goleak-files.sh
akagami-harsh: yes!, there was a error in logic but i think i fixed it now

Makefile
akagami-harsh: do you mean like this
 ```make
 .PHONY: cover
cover: nocover goleak
	$(GOTEST) -tags=memory_storage_integration -timeout 5m -coverprofile cover.out ./...
	go tool cover -html=cover.out -o cover.html`

yurishkuro:
I simplified the script and moved `goleak` to `lint` target.

In the previous run some of the packages failed on goleak, so you need to remove them from this PR.

yurishkuro:
we could also add goleak into all empty_test.go files, right now they do come up in the output

akagami-harsh:
there are still some packages that do not fail, should i also add them

yurishkuro:
let's do that in future PRs

PR #5011, opened 2023-12-18
Allow specifying version for HotROD docker-compose

yurishkuro:
- `examples/hotrod/docker-compose.yml` always used `latest` tag, which is not being refreshed by Docker unless you proactively do that. There was no way to override that.

## Description of the changes
- Add `JAEGER_VERSION` environment variable

- Previously my `latest` tags were pointing to stale versions, and the example was not actually working (likely due to earlier changes to introducing OTEL SDK)
- After this change: `$ JAEGER_VERSION=1.52 docker compose up` downloaded new versions and the tracing was working

### DIFF
--- a/examples/hotrod/docker-compose.yml
+++ b/examples/hotrod/docker-compose.yml
@@ -1,7 +1,7 @@
 version: '3.7'
 services:
   jaeger:
-    image: jaegertracing/all-in-one:latest
+    image: jaegertracing/all-in-one:${JAEGER_VERSION:-latest}
     ports:
       - "16686:16686"
       - "4318:4318"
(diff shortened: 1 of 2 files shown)


examples/hotrod/docker-compose.yml
```diff
@@ -1,7 +1,11 @@
 version: '3.7'
+
+# To run a specific version of Jaeger, use environment variable, e.g.:
+#     JAEGER_VERSION=1.52 docker compose up
```
albertteoh, line 4: Perhaps we should move this into https://github.com/jaegertracing/jaeger/tree/main/examples/hotrod#run-everything-via-docker-compose (and also update the command)?

albertteoh (approved):

examples/hotrod/docker-compose.yml
yurishkuro: added to README

PR #5017, opened 2023-12-19
Add goleak check to packages with empty tests

yurishkuro:
- Part of #5006

## Description of the changes
- Add goleak checks to all packages with empty_test.go as well as ./storage/... (which mostly contains APIs)

- CI


yurishkuro:
cc @akagami-harsh

jkowall (approved):
lgtm! Thanks.

PR #5020, opened 2023-12-19
Move Go Tip installation into sub-action

yurishkuro:
- Unit tests on Go Tip sometimes fail because Go Tip fails to install, but the output is non-descriptive, e.g. https://github.com/jaegertracing/jaeger/actions/runs/7263410887/job/19788655227

image

## Description of the changes
- Move Go Tip installation into a separate sub-action
- Add some debug logging to see the progress to make it easier to debug in case of failures (right now it's hard to tell which step is actually failing and resulting in the errors above)

- Tested in personal repo: https://github.com/yurishkuro/test-project-3/actions/runs/7264047725/job/19790690846

image


jkowall (approved):
lgtm!

PR #5022, opened 2023-12-19
ci: add retry logic in the Install Go Tip GitHub Action

### LINKED ISSUE #5021: Implement retry logic in the Install Go Tip GitHub Action
See https://github.com/jaegertracing/jaeger/pull/5020#issuecomment-1863045673

Proposal:
* Check if curl returns 404
* Add a loop that retries the download every 30sec for about 5min total, with adequate logging

akagami-harsh:
- Fixes #5021 

## Description of the changes
- added retry logic in Go Tip Github Action 

- in below case it ran successfully
- but the logic was not used because the `curl` command did not any return code 
- https://github.com/akagami-harsh/jaeger/actions/runs/7266233916/job/19797565337

## Checklist

### DIFF
--- a/.github/actions/setup-go-tip/action.yml
+++ b/.github/actions/setup-go-tip/action.yml
@@ -15,31 +15,25 @@ runs:
           tip=$(git ls-remote https://github.com/golang/go.git HEAD | awk '{print $1;}')
           echo "Go Tip version: ${tip}"
           if curl_output=$(curl -fsSL -w "%{http_code}" -o gotip.tar.gz https://storage.googleapis.com/go-build-snap/go/linux-amd64/${tip}.tar.gz); then
-            http_code=$(tail -n1 <<< "$curl_output")
-            
-            if [[ "$http_code" == "404" ]]; then
-              echo "HTTP 404: File not found"
-            else
-              echo "Downloaded bundle:"
-              ls -lah gotip.tar.gz
-              success=true
-              break
-            fi
+            echo "Downloaded bundle:"
+            ls -lah gotip.tar.gz
+            success=true
+            break
           else
             echo "Failed to download. Retrying in $wait_time seconds..."
             sleep $wait_time
           fi
         done
 
-        if [[ "$success" == true ]]; then
-          export GOROOT="$HOME/sdk/gotip"
-          mkdir -p $GOROOT
-          tar -C $GOROOT -xzf gotip.tar.gz
-          export PATH="$GOROOT/bin/:$PATH"
-          echo "GOROOT=$GOROOT" >> $GITHUB_ENV
-          echo "PATH=$PATH" >> $GITHUB_ENV
-          echo "Active Go version:"
-          go version
-        else
+        if [[ "$success" == false ]]; then
           echo "Failed to download Go Tip after $retries attempts"
+          exit 1
         fi
+        export GOROOT="$HOME/sdk/gotip"
+        mkdir -p $GOROOT
+        tar -C $GOROOT -xzf gotip.tar.gz
+        export PATH="$GOROOT/bin/:$PATH"
+        echo "GOROOT=$GOROOT" >> $GITHUB_ENV
+        echo "PATH=$PATH" >> $GITHUB_ENV
+        echo "Active Go version:"
+        go version
--- a/.github/actions/setup-go-tip/action.yml
+++ b/.github/actions/setup-go-tip/action.yml
@@ -19,16 +19,17 @@ runs:
             ls -lah gotip.tar.gz
             success=true
             break
-          else
-            echo "Failed to download. Retrying in $wait_time seconds..."
-            sleep $wait_time
           fi
+          echo "Failed to download. Retrying in $wait_time seconds..."
+          sleep $wait_time
         done
 
         if [[ "$success" == false ]]; then
           echo "Failed to download Go Tip after $retries attempts"
           exit 1
         fi
+        echo "Downloaded bundle:"
+        ls -lah gotip.tar.gz
         export GOROOT="$HOME/sdk/gotip"
         mkdir -p $GOROOT
         tar -C $GOROOT -xzf gotip.tar.gz
--- a/.github/actions/setup-go-tip/action.yml
+++ b/.github/actions/setup-go-tip/action.yml
@@ -8,16 +8,38 @@ runs:
       shell: bash
       run: |
         set -euo pipefail
-        tip=$(git ls-remote https://github.com/golang/go.git HEAD | awk '{print $1;}')
-        echo "Go Tip version: ${tip}"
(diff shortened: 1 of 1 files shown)


.github/actions/setup-go-tip/action.yml
```diff
@@ -8,16 +8,38 @@ runs:
...
+              success=true
+              break
+            fi
+          else
+            echo "Failed to download. Retrying in $wait_time seconds..."
+            sleep $wait_time
+          fi
+        done
+
+        if [[ "$success" == true ]]; then
```
yurishkuro, line 34: you can simply execute `[[ "$success" == true ]]` as a statement and it will abort the script if not satisfied

yurishkuro: if you want to print the message, then 
```
if [[ "$success" == false ]]; then
  echo "Failed to download Go Tip after $retries attempts"
  exit 1
fi
```

.github/actions/setup-go-tip/action.yml
```diff
@@ -8,16 +8,38 @@ runs:
...
+        wait_time=30
+        success=false
+        for ((i=1; i<=retries; i++)); do
+          tip=$(git ls-remote https://github.com/golang/go.git HEAD | awk '{print $1;}')
+          echo "Go Tip version: ${tip}"
+          if curl_output=$(curl -fsSL -w "%{http_code}" -o gotip.tar.gz https://storage.googleapis.com/go-build-snap/go/linux-amd64/${tip}.tar.gz); then
+            http_code=$(tail -n1 <<< "$curl_output")
+            
+            if [[ "$http_code" == "404" ]]; then
+              echo "HTTP 404: File not found"
```
yurishkuro, line 21: so you're tracking two conditions, (1) error code from curl or (2) the HTTP status code. Why is sleep/retry only on the former?

.github/actions/setup-go-tip/action.yml
```diff
@@ -8,16 +8,38 @@ runs:
...
+            else
+              echo "Downloaded bundle:"
```
yurishkuro, line 23: this could be some other failing HTTP code, `not 404` does not mean success

akagami-harsh: yes, another thing i noticed now if `curl` returns error code then the nested `if` will never execute
```bash
if curl_output=$(curl -fsSL -w "%{http_code}" -o gotip.tar.gz https://storage.googleapis.com/go-build-snap/go/linux-amd64/${tip}.tar.gz); then
  http_code=$(tail -n1 <<< "$curl_output")
  if [[ "$http_code" == "404" ]]; then
    echo "HTTP 404: File not found. Retrying in $wait_time seconds..."
    sleep $wait_time
```
I don't think we need to check for error code `404` specifically

.github/actions/setup-go-tip/action.yml
akagami-harsh: addressed!

.github/actions/setup-go-tip/action.yml
```diff
@@ -8,16 +8,32 @@ runs:
...
+        success=false
+        for ((i=1; i<=retries; i++)); do
+          tip=$(git ls-remote https://github.com/golang/go.git HEAD | awk '{print $1;}')
+          echo "Go Tip version: ${tip}"
+          if curl_output=$(curl -fsSL -w "%{http_code}" -o gotip.tar.gz https://storage.googleapis.com/go-build-snap/go/linux-amd64/${tip}.tar.gz); then
+            echo "Downloaded bundle:"
+            ls -lah gotip.tar.gz
+            success=true
+            break
+          else
```
yurishkuro, line 22: don't need else nesting, since you break out in if()

.github/actions/setup-go-tip/action.yml
```diff
@@ -8,16 +8,32 @@ runs:
...
-        ls -lah gotip.tar.gz
+        retries=10
+        wait_time=30
+        success=false
+        for ((i=1; i<=retries; i++)); do
+          tip=$(git ls-remote https://github.com/golang/go.git HEAD | awk '{print $1;}')
+          echo "Go Tip version: ${tip}"
+          if curl_output=$(curl -fsSL -w "%{http_code}" -o gotip.tar.gz https://storage.googleapis.com/go-build-snap/go/linux-amd64/${tip}.tar.gz); then
+            echo "Downloaded bundle:"
+            ls -lah gotip.tar.gz
```
yurishkuro, line 19: I would move L18-19 to after L31, keep only iteration logic in the loop

.github/actions/setup-go-tip/action.yml
```diff
@@ -8,16 +8,32 @@ runs:
...
-        curl -fsSL https://storage.googleapis.com/go-build-snap/go/linux-amd64/${tip}.tar.gz -o gotip.tar.gz
-        echo "Downloaded bundle:"
-        ls -lah gotip.tar.gz
+        retries=10
+        wait_time=30
+        success=false
+        for ((i=1; i<=retries; i++)); do
+          tip=$(git ls-remote https://github.com/golang/go.git HEAD | awk '{print $1;}')
+          echo "Go Tip version: ${tip}"
+          if curl_output=$(curl -fsSL -w "%{http_code}" -o gotip.tar.gz https://storage.googleapis.com/go-build-snap/go/linux-amd64/${tip}.tar.gz); then
```
yurishkuro, line 17:
```suggestion
          url="https://storage.googleapis.com/go-build-snap/go/linux-amd64/${tip}.tar.gz"
          if curl -fsSL -o gotip.tar.gz "$url"; then
```

.github/actions/setup-go-tip/action.yml
akagami-harsh: done!

.github/actions/setup-go-tip/action.yml
akagami-harsh: done!

.github/actions/setup-go-tip/action.yml
```diff
@@ -8,9 +8,25 @@ runs:
...
         set -euo pipefail
-        tip=$(git ls-remote https://github.com/golang/go.git HEAD | awk '{print $1;}')
-        echo "Go Tip version: ${tip}"
-        curl -fsSL https://storage.googleapis.com/go-build-snap/go/linux-amd64/${tip}.tar.gz -o gotip.tar.gz
+        retries=10
+        wait_time=30
+        success=false
+        for ((i=1; i<=retries; i++)); do
+          tip=$(git ls-remote https://github.com/golang/go.git HEAD | awk '{print $1;}')
+          echo "Go Tip version: ${tip}"
```
yurishkuro, line 16: L15-16 should be outside of the loop, I'd put them right after L10

akagami-harsh: addressed!

yurishkuro (approved):
thanks!

PR #5025, opened 2023-12-21
add goleak check in tests that do not fail

akagami-harsh:
- part of #5006 

## Description of the changes
- added goleak checks to all tests that do not fail
- minor bug fix in `check-goleak-file.sh`
  - `make goleak` was listing directories that do not contain any test, added a `if` condition to prevent that 

- `make test`


yurishkuro (approved):
Awesome!

yurishkuro:
@akagami-harsh some packages do fail, e.g.
```
goleak: Errors on successful test run: found unexpected goroutines:
[Goroutine 6 in state select, with github.com/golang/glog.(*fileSink).flushDaemon on top of the stack:
github.com/golang/glog.(*fileSink).flushDaemon(0x322b1d8)
	/home/runner/go/pkg/mod/github.com/golang/glog@v1.1.2/glog_file.go:351 +0x17c
created by github.com/golang/glog.init.1 in goroutine 1
	/home/runner/go/pkg/mod/github.com/golang/glog@v1.1.2/glog_file.go:166 +0x225
]
❌ FAIL	github.com/jaegertracing/jaeger/plugin/storage/integration	3.360s
```

What method did you use to validate "tests that do not fail"?

yurishkuro:
The integration package contains numerous conditions (build tags and env var checks) that disable/enable various storage tests, so it's difficult to test locally to ensure full coverage. I removed it for now.

akagami-harsh:
> What method did you use to validate "tests that do not fail"?

i tested i locally by using `make test` and by running package tests

PR #5026, opened 2023-12-22
Fix goroutine leaks in several packages

yurishkuro:
- Part of #5006

## Description of the changes
- Fix goroutine leaks in several packages

- go test

### DIFF
--- a/pkg/jtracer/jtracer.go
+++ b/pkg/jtracer/jtracer.go
@@ -75,6 +75,11 @@ func initHelper(
 	otelExporter func(ctx context.Context) (sdktrace.SpanExporter, error),
 	otelResource func(ctx context.Context, svc string) (*resource.Resource, error),
 ) (*sdktrace.TracerProvider, error) {
+	res, err := otelResource(ctx, svc)
+	if err != nil {
+		return nil, err
+	}
+
 	traceExporter, err := otelExporter(ctx)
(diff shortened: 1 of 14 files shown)


pkg/jtracer/jtracer.go
```diff
@@ -84,11 +89,6 @@ func initHelper(
 	// span processor to aggregate spans before export.
 	bsp := sdktrace.NewBatchSpanProcessor(traceExporter)
 
-	res, err := otelResource(ctx, svc)
-	if err != nil {
-		return nil, err
```
yurishkuro, line 89: this was leaking exporter created above. Better to create resource first since it doesn't start any background work.

plugin/storage/kafka/factory.go
```diff
@@ -90,6 +85,11 @@ func (f *Factory) Initialize(metricsFactory metrics.Factory, logger *zap.Logger)
 	default:
 		return errors.New("kafka encoding is not one of '" + EncodingJSON + "' or '" + EncodingProto + "'")
 	}
+	p, err := f.NewProducer(logger)
```
yurishkuro, line 88: produced starts bg activity so create it last

yurishkuro:
cc @akagami-harsh

jkowall (approved):
lgtm

PR #5029, opened 2023-12-23
Upgrade golangci-lint to v1.55.2 and fix errors

yurishkuro:
## Description of the changes
- Upgrade linter version
- Update configuration of depguard
  - Disable uber-go/atomic and crossdock as deps
- Replace invalid crossdock imports with stretchr/testify
- Replace uber-go/atomic usage with sync/atomic
  - in most cases pointers to atomics are not necessary, and likewise initialization to zero values
- Fix other test and linter errors revealed by new linter

### DIFF
--- a/cmd/collector/app/span_processor_test.go
+++ b/cmd/collector/app/span_processor_test.go
@@ -21,12 +21,12 @@ import (
 	"io"
 	"reflect"
 	"sync"
+	"sync/atomic"
 	"testing"
 	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	"go.uber.org/atomic"
 	"go.uber.org/zap"
 
 	"github.com/jaegertracing/jaeger/cmd/collector/app/handler"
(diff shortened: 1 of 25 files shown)


cmd/collector/app/span_processor_test.go
```diff
@@ -456,7 +459,14 @@ func TestSpanProcessorCountSpan(t *testing.T) {
 			p.background(10*time.Millisecond, p.updateGauges)
 
 			p.processSpan(&model.Span{}, "")
-			assert.NotEqual(t, uint64(0), p.bytesProcessed)
```
yurishkuro, line 459: this was always a bogus check since it compared pointer with uint64, and it was flagged by the new linter. But then the tests required the conditional check

pkg/queue/bounded_queue.go
```diff
@@ -39,9 +37,9 @@ type Consumer interface {
 type BoundedQueue struct {
 	workers       int
 	stopWG        sync.WaitGroup
-	size          *uatomic.Uint32
-	capacity      *uatomic.Uint32
-	stopped       *uatomic.Uint32
+	size          atomic.Int32
```
yurishkuro, line 40: atomic.Uint32 does not support subtracting, so switch to signed. OTEL collector has a cleaner implementation of bounded queue that we will be able to use in jaeger-v2

albertteoh (approved):

PR #5030, opened 2023-12-24
Publish test report

albertteoh:
## Which problem is this PR solving?
- Resolves #2859

## Description of the changes
- Publish unit test results using https://github.com/EnricoMi/publish-unit-test-result-action

## How was this change tested?
- Tested in my fork: https://github.com/albertteoh/jaeger/actions/runs/7312986726

## Checklist
- [x] I have read https://github.com/jaegertracing/jaeger/blob/master/CONTRIBUTING_GUIDELINES.md
- [x] I have signed all commits

PR #5035, opened 2023-12-24
Upload test report

albertteoh:
- Relates to https://github.com/jaegertracing/jaeger/issues/2859
- Continues from https://github.com/jaegertracing/jaeger/pull/5030

## Description of the changes
- Enables a test report summary in the PR comment and check summary.

- Confirm upload job is running successfully in my fork: https://github.com/albertteoh/jaeger/actions/runs/7316648725.
- Can only know if it's working once the PR is created.

## Checklist


albertteoh:
Set to draft to test a few things.

.github/workflows/ci-unit-tests-report.yml
```diff
@@ -0,0 +1,41 @@
...
+      actions: read
+
+    steps:
+      - name: Harden Runner
+        uses: step-security/harden-runner@eb238b55efaa70779f274895e782ed17c84f2895 # v2.6.1
+        with:
+          egress-policy: audit # TODO: change to 'egress-policy: block' after couple of runs
+
+      - name: Download and Extract Artifacts
+        uses: dawidd6/action-download-artifact@v3
```
yurishkuro, line 30: do we know what's different about it?  It's preferable to use official actions like https://github.com/actions/download-artifact

yurishkuro:
@albertteoh report upload just failed on main - https://github.com/jaegertracing/jaeger/actions/runs/7324976182/job/19949034415

Can we make it so that it doesn't fail the overall workflow?

.github/workflows/ci-unit-tests-report.yml
albertteoh: Not specifically, but I did spend a good number of hours trying out actions/download-artifact without any luck. Oddly, it wasn't able to discover the uploaded files, but there may have been a silly mistake on my part 🤷🏼 Happy to take suggestions!

.github/workflows/ci-unit-tests-report.yml
albertteoh, line 1: This file needs to be merged into `main` first before we can see the results of the test report summary in PRs, which is why we can't see it in this PR.

.github/workflows/ci-unit-tests-report.yml
```diff
@@ -0,0 +1,42 @@
+name: Unit Test Results
+
+on:
+  workflow_run:
+    workflows: ["Unit Tests"]
+    types:
+      - completed
+
+jobs:
+  unit-test-results:
```
yurishkuro, line 10: Why is it necessary to separate this into a different workflow? Doing so requires all the juggling with artifact files. Can we not do all it at once directly in the unit test workflow, where we already call publish reports action?

albertteoh: > Why is it necessary to separate this into a different workflow? Doing so requires all the juggling with artifact files. Can we not do all it at once directly in the unit test workflow, where we already call publish reports action?

It's a valid question, because one would think the action to execute the unit test and generate the results artifact is done as part of this repo's CI pipeline, so the relevant files should be easily accessible. 

In fact, we're trying to do that now (publish test results as a comment to this PR) because, I agree, it's a lot simpler than juggling artifact uploads and downloads; but it's failing with this error message (from [example run](https://github.com/jaegertracing/jaeger/actions/runs/7377433266/job/20071402739#step:10:32)):

> 2024-01-01 12:55:11 +0000 - publish -  INFO - This action is running on a pull_request event for a fork repository. Pull request comments and check runs cannot be created, so disabling these features. To fully run the action on fork repository pull requests, see https://github.com/EnricoMi/publish-unit-test-result-action/blob/v2.12.0/README.md#support-fork-repositories-and-dependabot-branches

I don't fully understand why, but from the log message, it almost sounds like a security measure to prevent PRs from another fork from writing to a PR owned by the "parent" repo, which seems to make sense.

Can you see any other alternative?

yurishkuro (approved):

yurishkuro:
The output of this reporting action keeps rising my blood pressure... (not your fault, Albert)

https://github.com/jaegertracing/jaeger/actions/runs/7387766706

What is "1 🔥"? When I look at JSON output there is clearly a test that failed:
```
{"Time":"2024-01-02T16:00:59.335298108Z","Action":"fail","Package":"github.com/jaegertracing/jaeger/plugin/storage/es","Elapsed":0.635}
```

albertteoh:
> What is "1 🔥"? When I look at JSON output there is clearly a test that failed

It's documented [here](https://github.com/EnricoMi/publish-unit-test-result-action?tab=readme-ov-file#the-symbols), albeit a little vaguely that it's an "erroneous" test, as opposed to a "failed" test.

Based on your [example "erroneous" test](https://github.com/jaegertracing/jaeger/actions/runs/7387766706/job/20097178910#step:8:8652) and an [example "failed" test](https://github.com/jaegertracing/jaeger/actions/runs/7394608292/job/20116252656#step:8:9531) my interpretation is:
- "erroneous": a test failure caused by a panic or exit. In your example, it's triggered by a goleak check which invokes `syscall.Exit`.
- "failed": a test failure caused by a failed assertion.

It is a little confusing though that, in both cases, the "Action" is "failed".

yurishkuro:
@albertteoh one of the tests failed in Go Tip run: https://github.com/jaegertracing/jaeger/actions/runs/7428429111

Is there a way to see detailed test report? The one in the summary doesn't seem to have any links.

albertteoh:
> Is there a way to see detailed test report? The one in the summary doesn't seem to have any links.

Yes (I see you've already approved, thanks!): https://github.com/jaegertracing/jaeger/pull/5082

PR #5037, opened 2023-12-25
Refactor Protobuf types generation

yurishkuro:
- It was no longer possible to run `make proto` on Macs with Apple silicon since the Docker image with protobuf was only built for amd64.
- We are generating unoptimized protobuf types for OTLP without Gogo, which will be a problem for jaeger-v2

## Description of the changes
- Upgrade jaegertracing/protobuf to https://github.com/jaegertracing/docker-protobuf/releases/tag/v0.5.0
- Move proto/thrift/crossdock-related targets into separate files `Makefile.*.mk`
- Proto targets improvements
  - DRY targets by using a macro for invoking protoc
  - Separate different modules into different targets
  - Add some pretty logging

- `make proto` regenerates all the files without any additional changes

## Next steps
- [ ] `proto-hotrod` produces a different file now, need to regenerate and test with HotROD, and then include it into the main `make proto` target
- [ ] Thrift gen produces a different output (can't run it on Mac)
- [ ] Utilize sed script from OTEL to add Gogo annotations to OTLP protos, to generate more efficient representation


.github/workflows/ci-protogen-tests.yml
```diff
@@ -26,11 +26,14 @@ jobs:
...
       with:
         go-version: 1.21.x
 
-    - name: Run protogen validation
-      run: make init-submodules && make proto && git diff --name-status --exit-code
+    - name: Verify Protobuf types are up to date
+      run: make proto && git diff --name-status --exit-code
+
+    # - name: Verify Thrift types are up to date
+    #   run: make thrift && git diff --name-status --exit-code
```
yurishkuro, line 39: will add in the future, right now it produces different output

Makefile.Protobuf.mk
```diff
@@ -0,0 +1,160 @@
...
+proto-hotrod:
+	$(call proto_compile, , examples/hotrod/services/driver/driver.proto)
+
+.PHONY: proto-zipkin
+proto-zipkin:
+	$(call proto_compile, proto-gen/zipkin, idl/proto/zipkin.proto, -Iidl/proto)
+
+# Target 'proto-prepare-otel' modifies OTEL proto to use proto-import path jaeger.proto.*
+# The modification is needed because OTEL collector already uses opentelemetry.proto.*
+# and two complied protobuf types cannot have the same import path. The root cause is that the compiled OTLP
```
albertteoh, line 120:
```suggestion
# and two compiled protobuf types cannot have the same import path. The root cause is that the compiled OTLP
```

albertteoh (approved):

Makefile.Protobuf.mk
yurishkuro: will fix in the next PR

yurishkuro: https://github.com/jaegertracing/jaeger/pull/5039

.github/workflows/ci-protogen-tests.yml
yurishkuro: https://github.com/jaegertracing/jaeger/pull/5039

PR #5038, opened 2023-12-25
Fix goroutine leaks in some packages

akagami-harsh:
- part of #5006 

## Description of the changes
- fix goroutine leaks in some package and add goleak check

- go test


yurishkuro (approved):
🎉 
Thanks!

PR #5040, opened 2023-12-26
Regenerate HotROD proto

yurishkuro:
- HotROD Proto was not checked for being up to date in the standard CI check
- And it did not match the output of the generator

## Description of the changes
- Include `proto-hotrod` in the main `proto` target so that it always runs and is checked for being up to date
- Change settings to exclude all *.pb.go files from linting and formatting
- Remove unnecessary wrapper script for checking license headers
- Remove unnecessary wrapper script for checking imports order
- Do not exclude `doc.go` from linting
  - Add empty tests to 3 packages that started to get tested now because of sole doc.go file

- CI


Makefile.Protobuf.mk
```diff
@@ -79,7 +79,14 @@ endef
 
 # TODO add proto-hotrod to the list after regenerating its file (may need linter tweaking)
 .PHONY: proto
-proto: proto-model proto-api-v2 proto-storage-v1 proto-zipkin proto-openmetrics proto-otel proto-api-v3
+proto: proto-model \
+	proto-api-v2 \
+	proto-storage-v1 \
+	proto-hotrod \
```
yurishkuro, line 85: added

albertteoh (approved):

PR #5044, opened 2023-12-26
Fix go tip unit test runs

### LINKED ISSUE #5043: Unit tests on tip fail when uploading test report
https://github.com/jaegertracing/jaeger/actions/runs/7330978812/job/19962951731

The tool is being installed, although strangely it lists both v1 and v2
```
go install github.com/jstemmer/go-junit-report/v2@latest
go: downloading github.com/jstemmer/go-junit-report/v2 v2.1.0
go: downloading github.com/jstemmer/go-junit-report v1.0.0
```
but then it fails saying `command not found`
```
go tool cover -html=cover.out -o cover.html
cat test-results.json | go-junit-report -parser gojson > junit-report.xml
/bin/bash: line 1: go-junit-report: command not found
cat: write error: Broken pipe
make: *** [Makefile:465: test-report] Error 127
```

cc @albertteoh

albertteoh:
- Resolves #5043

## Description of the changes
- `go-junit-report` was getting installed successfully into the default `$GOPATH` (`$HOME/go/bin`) but the GOPATH wasn't in the `$PATH` environment variable, which is what caused the `go-junit-report: command not found` error.
- It worked for the normal (non-go-tip) unit tests because the `$GOPATH` was already set and inserted into `$PATH`.
- This explicitly sets the `$GOPATH` and inserts it into `$PATH` prior to installing test tools.

- Tested in [my own fork](https://github.com/albertteoh/jaeger/actions/runs/7332463204/job/19966641893) with some debug logging in the Makefile.

## Checklist


.github/actions/setup-go-tip/action.yml
```diff
@@ -30,10 +30,12 @@ runs:
         echo "Downloaded bundle:"
         ls -lah gotip.tar.gz
         export GOROOT="$HOME/sdk/gotip"
+        export GOPATH="$HOME/go"
```
yurishkuro, line 33: Are you sure this is a good idea. By doing that we are mixing the default Go installation (e.g. some 1.21.x version) with `gotip`, which makes it difficult to understand which version exactly is being used for tests. 

Why do we need GOPATH? It was supposed to be deprecated some time ago. If we must have it, could it be set to GOROOT too?

albertteoh: > By doing that we are mixing the default Go installation (e.g. some 1.21.x version) with `gotip`, which makes it difficult to understand which version exactly is being used for tests.

I've checked that `go install ...` uses the `gotip` version instead of the default 1.21.x because `$GOROOT` is exported here, so IIUC, it _should_ be using consistent go versions, but please correct me if I've misunderstood something.

> Why do we need GOPATH?

Ah, that's where I thought `go install` places the binaries:

```
$ go help install
usage: go install [build flags] [packages]

Install compiles and installs the packages named by the import paths.

Executables are installed in the directory named by the GOBIN environment
variable, which defaults to $GOPATH/bin or $HOME/go/bin if the GOPATH
environment variable is not set. Executables in $GOROOT
are installed in $GOROOT/bin or $GOTOOLDIR instead of $GOBIN.
...
```

>  It was supposed to be deprecated some time ago.

It's the GOPATH _development mode_ that's getting deprecated

From https://zchee.github.io/golang-wiki/GOPATH/:

"Note that removing GOPATH development mode does not mean removing the GOPATH variable. It will still be used for the purposes listed at the top of this page.

FAQ
Is the GOPATH variable being removed?
No. The GOPATH variable (set in the environment or by go env -w) is not being removed. It will still be used to determine the default binary install location, module cache location, and checksum database cache location, as mentioned at the top of this page."

> If we must have it, could it be set to GOROOT too?

I'm happy to set GOPATH to the same path as GOROOT.

yurishkuro (approved):

.github/actions/setup-go-tip/action.yml
yurishkuro: ok fair enough

PR #5045, opened 2023-12-26
💤 Swap Zipkin server for Zipkin Receiver from OTel Collector Contrib

yurishkuro:
- Remove duplicated code for handling Zipkin formats that requires a dependency on open-api packages
- This will also help with jaeger-v2 since we'll just reuse the same OTEL Zipkin receiver, no more breaking changes

## Description of the changes
- Swap custom Zipkin server for Zipkin Receiver from OTel Collector Contrib
  - (**breaking** 🛑)  Zipkin Proto receiver requires 128bit trace ID, otherwise `TraceID: has length 8 yet wanted length 16`. Jaeger receiver did not have such limitation.
  - (**breaking** 🛑) the `--collector.zipkin.keep-alive` setting is not supported in the OTEL receiver
- Delete a bunch of packages and files that are no longer needed (5k lines, including code-gen)
- `go mod tidy` removed all `go-openapi` dependencies and a few others

- Added unit tests `cmd/collector/app/handler/zipkin_receiver_test.go` that submit various Zipkin formats to the receiver
- Crossdock tests also cover similar exports in CI:
```
Executing Matrix...
✓ [endtoend] test_driver→ (services=go) ⇒ Pass
✓ [endtoend] test_driver→ (services=java) ⇒ Pass
✓ [endtoend] test_driver→ (services=node) ⇒ Pass
✓ [endtoend] test_driver→ (services=python) ⇒ Pass
✓ [endtoend] test_driver→ (services=zipkin-brave-json) ⇒ Pass
✓ [endtoend] test_driver→ (services=zipkin-brave-json-v2) ⇒ Pass
✓ [endtoend] test_driver→ (services=zipkin-brave-proto) ⇒ Pass
✓ [endtoend] test_driver→ (services=zipkin-brave-thrift) ⇒ Pass
```

## Checklist


cmd/collector/app/handler/testdata/zipkin_thrift_v1_merged_spans.json
```diff
@@ -0,0 +1,79 @@
+[
```
yurishkuro, line 1: I used the existing `zipkin_v1_merged_spans.json` file (Zipkin JSON v1), converged into object model, and re-serialized as JSON. It uses different representations for many fields, like raw numbers instead of hex strings for IDs.

cmd/collector/app/handler/testdata/zipkin_proto_02.json
```diff
@@ -1,7 +1,7 @@
 {
   "spans": [
     {
-      "trace_id": "AAAAAAAAAAI=",
+      "trace_id":"QoVouqoJRn5CjmiKqglGfw==",
```
yurishkuro, line 4: (breaking 🛑) Zipkin Proto receiver requires 128bit trace ID, otherwise TraceID: has length 8 yet wanted length 16. Jaeger receiver did not have such limitation.

cmd/collector/app/handler/zipkin_receiver_tls_test.go
```diff
@@ -12,59 +12,29 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package server
+package handler
```
yurishkuro, line 15: This was testing exhaustively TLS settings on Zipkin endpoint, so I decided to keep it, just move from a different location. It works against the Zipkin receiver now.

jkowall (approved):
lgtm. One potential issue is keeping this sync with the upstream Otel changes. I do not think this will happen much with Zipkin being stagnant but it's a potential point of additional work.

yurishkuro:
What sync are you referring to? Basically my plan is to only use OTEL's implementation going forward and if any issues direct then upstream.

We still support Zipkin Thrift in the agent, which is a separate implementation, but we should be able to remove it once we sunset the agent.

PR #5046, opened 2023-12-27
Clean-up APIv3 tests

yurishkuro:
- I would like to remove dependency on grpc-gateway because it causes issues with using Gogo in OTLP protobufs
- The tests were not fully exercising all of the HTTP service functions

## Description of the changes
- DRY the tests
- Ensure no goroutines are leaked - pass tlscfg.Options by reference

- unit tests


albertteoh (approved):
I don't quite understand how this removes the dependency on GRPC Gateway, but that's probably because I'm not so familiar with this part of the code.

yurishkuro:
It does not remove the dependency yet. I first want to have good tests that exercise v3 api via http, and then replace the gateway implementation.

PR #5047, opened 2023-12-27
Fix goroutine leaks in plugin/sampling/strategystore/static/

akagami-harsh:
- part of #5006 

## Description of the changes
- fix goroutine leaks and add goleak check

- go test

## Checklist


plugin/sampling/strategystore/static/strategy_store_test.go
```diff
@@ -381,6 +382,7 @@ func TestAutoUpdateStrategyWithFile(t *testing.T) {
 
 func TestAutoUpdateStrategyWithURL(t *testing.T) {
 	mockServer, mockStrategy := mockStrategyServer()
+	defer mockServer.Close()
```
yurishkuro, line 385: a better fix would be to pass `t` above: `mockStrategyServer(t)` and inside that function add
```
t.Cleanup(func() { mockServer.Close() })
```
This way you don't have to keep cleaning it up everywhere

yurishkuro (approved):

plugin/sampling/strategystore/static/strategy_store_test.go
akagami-harsh: I've implemented cleanup approach into tests and removed redundant cleanups

yurishkuro (approved):
Thanks!

PR #5048, opened 2023-12-27
Fix goroutine leaks in some packages

akagami-harsh:
- part of #5006

## Description of the changes
- fix goroutine leaks and add goleak check 

- go test

## Checklist


yurishkuro (approved):
Thanks! 🙏

PR #5050, opened 2023-12-27
Change test stdout redirect to tee

albertteoh:
- Addresses https://github.com/jaegertracing/jaeger/pull/5030#discussion_r1436685933

## Description of the changes
- Forks the stdout to both stdout and json file to allow for debugging test runs.

- Ran locally.

## Checklist


yurishkuro:
already added in #5045

PR #5051, opened 2023-12-27
Add grpc-gateway tests for all APIv3 methods

yurishkuro:
- Part of #5052
- Continuation of #5046

## Description of the changes
- Add tests for all HTTP APIs, not just GetTrace
- Use snapshots to make validation of HTTP/JSON response from the server easier
- Replace grpc-gateway/runtime JSONPb marshaler (in tests) with gogo/jsonpb marshaler

## Gaps
- The error conditions are not being tested currently, such as not specifying the timestamps in the query

- Unit tests


cmd/query/app/apiv3/grpc_gateway_test.go
```diff
@@ -144,36 +222,100 @@ func testGRPCGatewayWithTenancy(
...
+	t.Run("GetOperations", func(t *testing.T) {
+		runGatewayGetOperations(t, gw, setupRequest)
+	})
+	t.Run("GetTrace", func(t *testing.T) {
+		runGatewayGetTrace(t, gw, setupRequest)
+	})
+	t.Run("FindTraces", func(t *testing.T) {
+		runGatewayFindTraces(t, gw, setupRequest)
+	})
+}
```
yurishkuro, line 237: @albertteoh this was the point of cleaning up tests - how they are testing all gateway APIs and we can try things like upgrading the gateway dep to v2.x or completely re-implementing it. My next step would be to try an upgrade, which requires building a new protoc image that I just kicked off (takes 4hrs to build), and maybe with v2 gateway we could find a way to integrate gogo marshaling.

jkowall (approved):
looks good to me.

Issue #5052, opened 2023-12-27
Define internal version of OTLP proto types using Gogo optimizations


yurishkuro:
Per [Jaeger v2 design doc](https://docs.google.com/document/d/1s4_6VgAS7qAVp6iEm5KYvpiGw3h2Ja5T5HpKo29iv00/edit#heading=h.hlhu8pegwdcj), we need to have our internal copy of code-generated Proto types for OTLP data model (because OTel Collector keeps it private). My current experiments in https://github.com/yurishkuro/jaeger/tree/otel-proto-gogo ran into an issue with grpc-gateway that already uses internally built OTLP protos, but without Gogo optimizations.

* Part of #4843

## Plan of action

* [ ] Upgrade grpc-gateway to v2
  * [x] clean up tests https://github.com/jaegertracing/jaeger/pull/5046
  * [x] add tests for all API v3 methods to ensure stable JSON output https://github.com/jaegertracing/jaeger/pull/5051
* [ ] Try to use v2 gateway with gogoproto marshaling
  * [ ] if it does not work, consider reimplementing the gateway manually, it's only 4 functions in the API. It will also be more efficient since we won't need to do a roundtrip to the gRPC service.

PR #5054, opened 2023-12-28
Add manual implementation of APIv3 HTTP endpoints

yurishkuro:
- Part of #5052

## Description of the changes
- Pull in IDL change https://github.com/jaegertracing/jaeger-idl/pull/102
- Re-implement APIv3 HTTP endpoints without the use of grpc-gateway
- Share tests from grpc-gateway for manual implementation
  - Refactor tests to avoid duplication of snapshots
- Fix inconsistency between http and grpc tenancy interceptors where HTTP was returning Unauthenticated in certain cases, but GRPC was always returning Forbidden. Make them consistent: missing tenant header results in Unauthenticated.

## Follow-ups
- [x] http implementation needs more unit tests (mostly error handling and parameter variations)
- [ ] the new implementation is not hooked up into production code yet, I first want to confirm it works with model-v2, and just in general minimize the scope of a single PR

- Using unit tests added in #5051, with additional enhancements


cmd/query/app/apiv3/grpc_gateway_test.go
```diff
@@ -60,6 +61,9 @@ const (
 //	REGENERATE_SNAPSHOTS=true go test -v ./cmd/query/app/apiv3/...
 var regenerateSnapshots = os.Getenv("REGENERATE_SNAPSHOTS") == "true"
 
+// The tests in http_gateway_test.go set this to true to use manual gateway implementation.
+var useHTTPGateway = false
```
yurishkuro, line 65: not the cleanest way, but hopefully we'll just delete grpc-gateway based implementation and simplify/consolidate the tests

cmd/query/app/apiv3/grpc_gateway_test.go
```diff
@@ -177,17 +185,6 @@ func parseResponse(t *testing.T, body []byte, obj gogoproto.Message) {
 	require.NoError(t, gogojsonpb.Unmarshal(bytes.NewBuffer(body), obj))
 }
 
-func parseChunkResponse(t *testing.T, body []byte, obj gogoproto.Message) {
-	// Unwrap the 'result' container generated by the gateway.
-	// See https://github.com/grpc-ecosystem/grpc-gateway/issues/2189
-	type resultWrapper struct {
-		Result json.RawMessage `json:"result"`
```
yurishkuro, line 184: to void this ugliness use the new helper types from IDL and marshal in one go.

PR #5055, opened 2023-12-28
test: Added goleak to TestMain in cmd/collector/app/server package folder

### LINKED ISSUE #5006: Enable and enforce goroutine leak checks in tests

limistah:
- Part of #5006 

## Description of the changes
- Added TestMain to package folders
- Added make goleak into unit test ci

- please run make goleak 

## Checklist


.github/workflows/ci-unit-tests.yml
```diff
@@ -35,6 +35,9 @@ jobs:
     - name: Run unit tests
       run: make test-ci
 
+    - name: Run go leak
+      run: make goleak
```
yurishkuro, line 39: this is already called as part of the lint workflow, please remove

cmd/agent/app/reporter/grpc/grpc_test.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"testing"
+
+	"go.uber.org/goleak"
+)
+
+func TestMain(m *testing.M) {
+	goleak.VerifyTestMain(m)
+}
```
yurishkuro, line 25:
```
$ go test ./cmd/agent/app/reporter/grpc/
PASS
goleak: Errors on successful test run: found unexpected goroutines:
```

cmd/collector/app/app_test.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"testing"
+
+	"go.uber.org/goleak"
+)
+
+func TestMain(m *testing.M) {
+	goleak.VerifyTestMain(m)
+}
```
yurishkuro, line 25:
```
$ go test ./cmd/collector/app/
```
finds the leaks

yurishkuro (changes requested):
I spot-checked a couple packages with added goleak checks and the tests now fail with actual goleaks. The objective of the ticket was not just add the checks and break the CI, but to fix the leaks and only then add the checks,

yurishkuro:
btw the test workflow was successful here because of a recently introduced bug -- #5057

limistah:
> I spot-checked a couple packages with added goleak checks and the tests now fail with actual goleaks. The objective of the ticket was not just add the checks and break the CI, but to fix the leaks and only then add the checks.

Cool, I'd check in the leaks.

Did a quick run on them and noticed a few of them, some were floated from vendor packages.

I might have to reduce the scope of the fix as the errors returned by the go test path/to/package returned quite a lot of them.

yurishkuro:
> I might have to reduce the scope of the fix as the errors returned by the go test path/to/package returned quite a lot of them.

this was exactly the recommendation in the ticket.

limistah:
> > I might have to reduce the scope of the fix as the errors returned by the go test path/to/package returned quite a lot of them.
> 
> this was exactly the recommendation in the ticket.

Great, I added the goleak fixes for `cmd/collector/app/server` package.

cmd/collector/app/server/grpc_test.go
```diff
@@ -35,6 +36,13 @@ import (
 	"github.com/jaegertracing/jaeger/proto-gen/api_v2"
 )
 
+func stopOpenCensus(t *testing.T) {
```
yurishkuro, line 39: this is wrong. We don't call opencensus so should not be stopping it manually either. It's most likely some other component starts it and we're not shutting down that component.

cmd/collector/app/server/http_test.go
```diff
@@ -240,18 +248,21 @@ func TestSpanCollectorHTTPS(t *testing.T) {
 			} else {
 				require.NoError(t, requestError)
 				require.NotNil(t, response)
+				response.Body.Close()
```
yurishkuro, line 251: this should be on L245 as `defer ...`

cmd/collector/app/server/http_test.go
```diff
@@ -64,26 +64,30 @@ func TestCreateTLSHTTPServerError(t *testing.T) {
...
 	server := httptest.NewServer(nil)
-	defer server.Close()
 
 	serveHTTP(server.Config, server.Listener, params)
 
 	response, err := http.Post(server.URL, "", nil)
 	require.NoError(t, err)
 	assert.NotNil(t, response)
+	response.Body.Close()
+	server.Close()
```
yurishkuro, line 90: why did you move it? `defer close` after creating is the correct pattern

cmd/collector/app/server/http_test.go
limistah: In a case when there is an error in the request, it throws a nil pointer since the request.Body might not have been initialized 

github.com/jaegertracing/jaeger/cmd/collector/app/server.serveHTTP.func1
        /Users/limistah/Work/oss/jaeger/cmd/collector/app/server/http.go:111
--- FAIL: TestSpanCollectorHTTPS (0.02s)
    --- FAIL: TestSpanCollectorHTTPS/should_fail_with_TLS_client_to_untrusted_TLS_server (0.02s)
[signal SIGSEGV: segmentation violation code=0x2 addr=0x40 pc=0x104e6664c]

cmd/collector/app/server/grpc_test.go
limistah: So, I removed the code.

The reason why I added it is that [opencensus has an init](https://github.com/census-instrumentation/opencensus-go/blob/01e6da5fc01c42aca1e0ce315f41744876e9fcbb/stats/view/worker.go#L33) that starts the worker 


I have searched for [NewMeter](https://github.com/census-instrumentation/opencensus-go/blob/01e6da5fc01c42aca1e0ce315f41744876e9fcbb/stats/view/worker.go#L261) within this codebase, the best that I could find is inside of the cmd/collector/app/handler using [opentelemtry's NewMeterProvider](https://pkg.go.dev/go.opentelemetry.io/otel/metric@v1.21.0/noop#NewMeterProvider) but none of these exposes the worker.

Here is a screenshot of the error:

image

Any suggestion here, I sincerely don't want to silent the leak.

cmd/collector/app/server/server_test.go
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) 2023 The Jaeger Authors.
```
yurishkuro, line 1: we call these `package_test.go`

cmd/collector/app/server/grpc_test.go
yurishkuro: I prefer to silence it. Starting goroutines in init() is a stupid pattern and it does not indicate a leak in our code, so why bother handling it manually. See #5066 for similar issue.

I believe newer versions of OTEL Collector will soon remove opencensus dependency altogether.

limistah: Fair enough

yurishkuro (approved):

pkg/testutils/leakcheck_test.go
```diff
@@ -24,3 +24,10 @@ func TestIgnoreGlogFlushDaemonLeak(t *testing.T) {
 		t.Errorf("IgnoreGlogFlushDaemonLeak() returned nil, want non-nil goleak.Option")
 	}
 }
+
+func TestIgnoreOpenCensusWorkerLeak(t *testing.T) {
+	opt := IgnoreOpenCensusWorkerLeak()
+	if opt == nil {
+		t.Errorf("IgnoreOpenCensusWorkerLeak() returned nil, want non-nil goleak.Option")
+	}
```
yurishkuro, line 32:
```suggestion
	assert.NotNil(t, IgnoreOpenCensusWorkerLeak())
```

pkg/testutils/leakcheck_test.go
```diff
@@ -24,3 +24,7 @@ func TestIgnoreGlogFlushDaemonLeak(t *testing.T) {
 		t.Errorf("IgnoreGlogFlushDaemonLeak() returned nil, want non-nil goleak.Option")
 	}
 }
+
+func TestIgnoreOpenCensusWorkerLeak(t *testing.T) {
+	assert.NotNil(t, IgnoreOpenCensusWorkerLeak())
```
yurishkuro, line 29: this will not compile, you need an import from stretchr/testify

yurishkuro (approved):

limistah:
I ran make cover on [this commit](https://github.com/jaegertracing/jaeger/pull/5055/commits/2151038a6d9c95cb8ca4580f7918a1c529dd5aaf)

Which gave the below result for `pkg/testutils/leakcheck_test.go`
image

yurishkuro (approved):

PR #5061, opened 2023-12-29
Separate test report collection from the main target

yurishkuro:
- When the tests fail, the target `test-report` is never run and thus no summary can be created

## Description of the changes
- Invoke the target from the GH workflow as an explicit `if: always()` step


albertteoh (approved):
Thanks!

PR #5062, opened 2023-12-30
Visualize uninstrumented services in the dependency diagrams

### LINKED ISSUE #3804: [Feature]: Visualize uninstrumented services in the dependency diagrams

nidhey27:
- Closes #3804 
- Replaces #4853

## Description of the changes
- Updated `GetDependencies` to identify and include client spans without corresponding server spans, ensuring a complete dependency graph.
- Implemented `updateServiceDependencyLinks` to streamline the update of service dependencies, reducing code duplication.
- Enabled the representation of uninstrumented services in the dependency graph by inferring names from client spans lacking server spans.

- Introduced TestInferredServiceDependency to specifically evaluate the functionality of inferred service dependencies in the scenario of uninstrumented services.
- Created a client span without a corresponding server span and asserted the correct formation of dependencies, ensuring the inclusion of inferred services in the dependency graph.
- Employed assertions to confirm the presence and accuracy of inferred service dependencies, ensuring they are adequately represented with the correct attributes and relationships.

## Checklist


nidhey27:
Ref PR: https://github.com/jaegertracing/jaeger/issues/3804

examples/hotrod/services/customer/database.go
```diff
@@ -77,7 +76,6 @@ func (d *database) Get(ctx context.Context, customerID int) (*Customer, error) {
 
 	ctx, span := d.tracer.Start(ctx, "SQL SELECT", trace.WithSpanKind(trace.SpanKindClient))
 	span.SetAttributes(
-		semconv.PeerServiceKey.String("mysql"),
```
yurishkuro, line 80: I don't understand your thought process behind these changes. Why is it appropriate to remove this attribute? Just because it happens to help your specific graph implementation doesn't make it the right thing to do.

nidhey27: Hi @yurishkuro, I apologize for making the above changes. A problem was identified in the code where the dependency links were created, which did not contain a check to prevent the addition of parent services with the same name as child services. The new changes have been added. Could you please review them?

kannan-ak:
+1, we are awaiting this feature. 
Currently the deep dependency graph is not showing the relationship with the databases and other non-instrumented services, for the given service.

jkowall:
This is a dupe of #4853 which, as you can see, was a very long conversation. This PR is a 10 line change. It's very frustrating for the maintainers who need to spend a ton of time fixing such a small issue. I spoke to the author as well and explained this decision we are making. The decision is that this will not be merged as it would need to be rebased and once again create a lot of work for the team. Sorry.

PR #5066, opened 2023-12-30
Fix Goroutine leaks in several packages

akagami-harsh:
- fix gorourine leak in some packages 

## Description of the changes
- part of #5006 

- go test

## Checklist


akagami-harsh:
I've ignored `goleak` check for `github.com/golang/glog.(*fileSink).flushDaemon` in certain packages, as is couldn't find a fix for it

cmd/internal/env/command_test.go
```diff
@@ -30,3 +31,7 @@ func TestCommand(t *testing.T) {
 	assert.True(t, strings.Contains(buf.String(), "METRICS_BACKEND"))
 	assert.True(t, strings.Contains(buf.String(), "SPAN_STORAGE"))
 }
+
+func TestMain(m *testing.M) {
+	goleak.VerifyTestMain(m, goleak.IgnoreTopFunction("github.com/golang/glog.(*fileSink).flushDaemon"))
```
yurishkuro, line 36: > goleak.IgnoreTopFunction("github.com/golang/glog.(*fileSink).flushDaemon")

Let's move this to pkg/testutils (with comment explaining why it's needed) and reuse instead of repeating everywhere.

cmd/agent/app/reporter/grpc/builder.go
```diff
@@ -102,6 +102,9 @@ func (b *ConnBuilder) CreateConnection(logger *zap.Logger, mFactory metrics.Fact
 	dialOptions = append(dialOptions, grpc.WithUnaryInterceptor(grpc_retry.UnaryClientInterceptor(grpc_retry.WithMax(b.MaxRetry))))
 	dialOptions = append(dialOptions, b.AdditionalDialOptions...)
 
+	connCtx, cancel := context.WithCancel(ctx)
+	defer cancel()
```
yurishkuro, line 106: `CreateConnection` is not a blocking function, it will exit after setting things up and this defer will be executed. Is that what you intended?

Please move this specific change to a different PR, as other packages don't require substantial changes and the rest of this PR can be merged.

cmd/internal/env/command_test.go
akagami-harsh: yes, that would be better

cmd/agent/app/reporter/grpc/builder.go
akagami-harsh: >CreateConnection is not a blocking function, it will exit after setting things up and this defer will be executed. Is that what you intended?

ohh my mistake, i should have just returned the `cancel()` function

akagami-harsh: >Please move this specific change to a different PR, as other packages don't require substantial changes and the rest of this PR can be merged.

Okay!

cmd/internal/env/command_test.go
```diff
@@ -19,7 +19,9 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/jaegertracing/jaeger/pkg/testutils"
 	"github.com/stretchr/testify/assert"
+	"go.uber.org/goleak"
```
yurishkuro, line 24: run `make fmt`

pkg/testutils/leakcheck.go
```diff
@@ -0,0 +1,13 @@
...
+	"go.uber.org/goleak"
+)
+
+// IgnoreGlogFlushDaemonLeak returns a goleak.Option that ignores the flushDaemon function
+// from the glog package that can cause false positives in leak detection.
+// This is necessary because glog starts a goroutine in the background that may not
+// be stopped when the test finishes, leading to a detected but expected leak.
+func IgnoreGlogFlushDaemonLeak() goleak.Option {
+	return goleak.IgnoreTopFunction("github.com/golang/glog.(*fileSink).flushDaemon")
+}
```
yurishkuro, line 13: add package_test invoking this function to appease code coverage

pkg/testutils/leakcheck.go
```diff
@@ -0,0 +1,13 @@
+package testutils
+
+import (
+	"go.uber.org/goleak"
+)
+
+// IgnoreGlogFlushDaemonLeak returns a goleak.Option that ignores the flushDaemon function
+// from the glog package that can cause false positives in leak detection.
+// This is necessary because glog starts a goroutine in the background that may not
+// be stopped when the test finishes, leading to a detected but expected leak.
```
yurishkuro, line 24: great explanation!

pkg/testutils/leakcheck.go
yurishkuro: Also, I just realized, we could've done this differently by adding `testutils.CheckGoroutineLeaks` function that would be called from the rest of the codebase, instead of having two dependencies everywhere (goleak and testutils).

yurishkuro (approved):

pkg/testutils/leakcheck.go
akagami-harsh: yeah, we can also do that.

yurishkuro:
Thanks!

PR #5067, opened 2023-12-31
Add gogo/protobuf to OpenTelemetry OTLP data model

### LINKED ISSUE #5052: Define internal version of OTLP proto types using Gogo optimizations

yurishkuro:
- Resolves #5052

## Description of the changes
- Apply gogo annotations to OTLP proto for generating more efficient data model

- Unit tests


cmd/query/app/apiv3/http_gateway.go
```diff
@@ -124,17 +124,6 @@ func (h *HTTPGateway) returnSpansTestable(
 	if h.tryHandleError(w, err, http.StatusInternalServerError) {
 		return
 	}
-	for _, rs := range resourceSpans {
-		for _, ss := range rs.ScopeSpans {
-			for _, s := range ss.Spans {
-				if len(s.ParentSpanId) == 0 {
-					// If ParentSpanId is empty array then gogo/jsonpb renders it as empty string.
-					// To match the output with grpc-gateway we set it to nil and it won't be included.
-					s.ParentSpanId = nil
```
yurishkuro, line 133: field is not nullable so this fix is no longer possible

cmd/query/app/apiv3/snapshots/FindTraces.json
```diff
@@ -9,10 +9,14 @@
...
                 "endTimeUnixNano": "11651379494838206464",
+                "kind": "SPAN_KIND_SERVER",
                 "name": "foobar",
+                "parentSpanId": "",
                 "spanId": "AAAAAAAAALQ=",
                 "startTimeUnixNano": "11651379494838206464",
-                "status": {},
+                "status": {
+                  "code": "STATUS_CODE_ERROR"
+                },
```
yurishkuro, line 19: Added this and span kind as examples of enums. This matches the previous output when using grpc-gateway marshaling (and is different from OTLP-JSON encoding).

cmd/query/app/apiv3/snapshots/FindTraces.json
```diff
@@ -9,10 +9,14 @@
             "spans": [
               {
                 "endTimeUnixNano": "11651379494838206464",
+                "kind": "SPAN_KIND_SERVER",
                 "name": "foobar",
+                "parentSpanId": "",
```
yurishkuro, line 14: Could not find a way to suppress this empty field. In practice not a big deal all but one span in a trace will have non-empty parent. Also, verified that OTLP-JSON also includes empty parent ID.

model/v2/jsonid.go
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) 2023 The Jaeger Authors.
+// Copyright The OpenTelemetry Authors
```
yurishkuro, line 2: Copied from OTEL, but had to change to marshal IDs as base64, to keep with the original grpc-gateway version.

PR #5094, opened 2024-01-11
Fix flaky test: plugin/storage/badger/samplingstore

### LINKED ISSUE #5090: [Bug] Flaky test: plugin/storage/badger/samplingstore
Sporadically fails, e.g. I was able to reproduce with this:
```
$ go test -count 10 ./plugin/storage/badger/samplingstore
```

```
--- FAIL: TestGetThroughput (0.08s)
    storage_test.go:58:
        	Error Trace:	/Users/ysh/dev/jaegertracing/jaeger/plugin/storage/badger/samplingstore/storage_test.go:58
        	            				/Users/ysh/dev/jaegertracing/jaeger/plugin/storage/badger/samplingstore/storage_test.go:142
        	            				/Users/ysh/dev/jaegertracing/jaeger/plugin/storage/badger/samplingstore/storage_test.go:47
        	Error:      	Not equal:
        	            	expected: []*model.Throughput{(*model.Throughput)(0x14000092270), (*model.Throughput)(0x140000922a0)}
        	            	actual  : []*model.Throughput(nil)

        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,15 +1,2 @@
        	            	-([]*model.Throughput) (len=2) {
        	            	- (*model.Throughput)({
        	            	-  Service: (string) (len=6) "my-svc",
        	            	-  Operation: (string) (len=2) "op",
        	            	-  Count: (int64) 0,
        	            	-  Probabilities: (map[string]struct {}) <nil>
        	            	- }),
        	            	- (*model.Throughput)({
        	            	-  Service: (string) (len=7) "our-svc",
        	            	-  Operation: (string) (len=3) "op2",
        	            	-  Count: (int64) 0,
        	            	-  Probabilities: (map[string]struct {}) <nil>
        	            	- })
        	            	-}
        	            	+([]*model.Throughput) <nil>

        	Test:       	TestGetThroughput
```

akagami-harsh:
- Resolves #5090

## Description of the changes
- fixed flaky test
- fixed a small typo

### DIFF
--- a/plugin/storage/badger/samplingstore/storage.go
+++ b/plugin/storage/badger/samplingstore/storage.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"encoding/binary"
 	"encoding/json"
+	"fmt"
 	"time"
 
 	"github.com/dgraph-io/badger/v3"
@@ -95,7 +96,7 @@ func (s *SamplingStore) GetThroughput(start, end time.Time) ([]*model.Throughput
 			if err != nil {
 				return err
 			}
-			if t.Before(end) || t.Equal(end) {
+			if t.After(start) && (t.Before(end) || t.Equal(end)) {
 				retSlice = append(retSlice, throughputs...)
 			}
 		}
@@ -251,6 +252,7 @@ func initalStartTime(timeBytes []byte) (time.Time, error) {
 	buf := bytes.NewReader(timeBytes)
 
 	if err := binary.Read(buf, binary.BigEndian, &usec); err != nil {
+		fmt.Println("error-------------------------------------")
 		return time.Time{}, err
 	}
 
--- a/plugin/storage/badger/samplingstore/storage.go
+++ b/plugin/storage/badger/samplingstore/storage.go
@@ -18,7 +18,6 @@ import (
 	"bytes"
 	"encoding/binary"
 	"encoding/json"
-	"fmt"
 	"time"
 
 	"github.com/dgraph-io/badger/v3"
@@ -252,7 +251,6 @@ func initalStartTime(timeBytes []byte) (time.Time, error) {
 	buf := bytes.NewReader(timeBytes)
 
 	if err := binary.Read(buf, binary.BigEndian, &usec); err != nil {
-		fmt.Println("error-------------------------------------")
 		return time.Time{}, err
 	}
 
--- a/plugin/storage/badger/samplingstore/storage.go
+++ b/plugin/storage/badger/samplingstore/storage.go
@@ -96,7 +96,7 @@ func (s *SamplingStore) GetThroughput(start, end time.Time) ([]*model.Throughput
 				return err
 			}
 
-			if (t.After(start) || t.Equal(start)) && (t.Before(end) || t.Equal(end)) {
+			if t.After(start) && (t.Before(end) || t.Equal(end)) {
 				retSlice = append(retSlice, throughputs...)
 			}
 		}
--- a/plugin/storage/badger/samplingstore/storage.go
+++ b/plugin/storage/badger/samplingstore/storage.go
@@ -95,8 +95,7 @@ func (s *SamplingStore) GetThroughput(start, end time.Time) ([]*model.Throughput
 			if err != nil {
 				return err
 			}
-
-			if t.After(start) && (t.Before(end) || t.Equal(end)) {
+			if t.Before(end) || t.Equal(end) {
 				retSlice = append(retSlice, throughputs...)
 			}
 		}
--- a/plugin/storage/badger/samplingstore/storage.go
+++ b/plugin/storage/badger/samplingstore/storage.go
@@ -96,7 +96,7 @@ func (s *SamplingStore) GetThroughput(start, end time.Time) ([]*model.Throughput
 				return err
 			}
 
-			if t.After(start) && (t.Before(end) || t.Equal(end)) {
+			if (t.After(start) || t.Equal(start)) && (t.Before(end) || t.Equal(end)) {
 				retSlice = append(retSlice, throughputs...)
 			}
 		}
--- a/plugin/storage/badger/samplingstore/storage.go
+++ b/plugin/storage/badger/samplingstore/storage.go
@@ -95,6 +95,7 @@ func (s *SamplingStore) GetThroughput(start, end time.Time) ([]*model.Throughput
 			if err != nil {
 				return err
 			}
+
 			if t.After(start) && (t.Before(end) || t.Equal(end)) {
 				retSlice = append(retSlice, throughputs...)
 			}
--- a/plugin/storage/badger/samplingstore/storage.go
+++ b/plugin/storage/badger/samplingstore/storage.go
@@ -91,7 +91,7 @@ func (s *SamplingStore) GetThroughput(start, end time.Time) ([]*model.Throughput
 			if err != nil {
 				return err
 			}
-			throughputs, err := decodeThroughtputValue(val)
+			throughputs, err := decodeThroughputValue(val)
 			if err != nil {
 				return err
 			}
@@ -225,7 +225,7 @@ func (s *SamplingStore) createThroughputKV(throughput []*model.Throughput, start
 	return key, bb, err
 }
 
-func decodeThroughtputValue(val []byte) ([]*model.Throughput, error) {
+func decodeThroughputValue(val []byte) ([]*model.Throughput, error) {
 	var throughput []*model.Throughput
 
 	err := json.Unmarshal(val, &throughput)
--- a/plugin/storage/badger/samplingstore/storage_test.go
+++ b/plugin/storage/badger/samplingstore/storage_test.go
@@ -45,11 +45,13 @@ func TestInsertThroughput(t *testing.T) {
 
 func TestGetThroughput(t *testing.T) {
 	runWithBadger(t, func(t *testing.T, store *SamplingStore) {
-		start := time.Now().Truncate(time.Millisecond)
+		start := time.Now()
 		expected := []*samplemodel.Throughput{
 			{Service: "my-svc", Operation: "op"},
 			{Service: "our-svc", Operation: "op2"},
 		}
+
+		assert.Eventually(t, func() bool { return true }, time.Second, time.Microsecond*10)
 		err := store.InsertThroughput(expected)
 		require.NoError(t, err)
 
--- a/plugin/storage/badger/samplingstore/storage_test.go
+++ b/plugin/storage/badger/samplingstore/storage_test.go
@@ -53,7 +53,6 @@ func TestGetThroughput(t *testing.T) {
 		err := store.InsertThroughput(expected)
 		require.NoError(t, err)
 
-		time.Sleep(10 * time.Millisecond)
 		actual, err := store.GetThroughput(start, start.Add(time.Second*time.Duration(10)))
 		require.NoError(t, err)
 		assert.Equal(t, expected, actual)
@@ -121,7 +120,7 @@ func TestDecodeThroughtputValue(t *testing.T) {
 
 	marshalBytes, err := json.Marshal(expected)
 	require.NoError(t, err)
-	acrual, err := decodeThroughtputValue(marshalBytes)
+	acrual, err := decodeThroughputValue(marshalBytes)
 	require.NoError(t, err)
 	assert.Equal(t, expected, acrual)
 }
--- a/plugin/storage/badger/samplingstore/storage_test.go
+++ b/plugin/storage/badger/samplingstore/storage_test.go
@@ -45,7 +45,7 @@ func TestInsertThroughput(t *testing.T) {
 
 func TestGetThroughput(t *testing.T) {
 	runWithBadger(t, func(t *testing.T, store *SamplingStore) {
-		start := time.Now().Truncate(time.Microsecond)
+		start := time.Now()
 		expected := []*samplemodel.Throughput{
 			{Service: "my-svc", Operation: "op"},
 			{Service: "our-svc", Operation: "op2"},
@@ -54,7 +54,7 @@ func TestGetThroughput(t *testing.T) {
 		err := store.InsertThroughput(expected)
 		require.NoError(t, err)
 
-		actual, err := store.GetThroughput(start, start.Add(time.Second*time.Duration(10)))
+		actual, err := store.GetThroughput(start.Add(-time.Millisecond), start.Add(time.Second*time.Duration(10)))
 		require.NoError(t, err)
 		assert.Equal(t, expected, actual)
 	})
--- a/plugin/storage/badger/samplingstore/storage_test.go
+++ b/plugin/storage/badger/samplingstore/storage_test.go
@@ -50,7 +50,6 @@ func TestGetThroughput(t *testing.T) {
 			{Service: "my-svc", Operation: "op"},
 			{Service: "our-svc", Operation: "op2"},
 		}
-
 		err := store.InsertThroughput(expected)
 		require.NoError(t, err)
 
--- a/plugin/storage/badger/samplingstore/storage_test.go
+++ b/plugin/storage/badger/samplingstore/storage_test.go
@@ -45,7 +45,7 @@ func TestInsertThroughput(t *testing.T) {
 
 func TestGetThroughput(t *testing.T) {
 	runWithBadger(t, func(t *testing.T, store *SamplingStore) {
-		start := time.Now()
+		start := time.Now().Truncate(time.Millisecond)
 		expected := []*samplemodel.Throughput{
 			{Service: "my-svc", Operation: "op"},
 			{Service: "our-svc", Operation: "op2"},
@@ -53,6 +53,7 @@ func TestGetThroughput(t *testing.T) {
 		err := store.InsertThroughput(expected)
 		require.NoError(t, err)
 
+		time.Sleep(10 * time.Millisecond)
 		actual, err := store.GetThroughput(start, start.Add(time.Second*time.Duration(10)))
 		require.NoError(t, err)
 		assert.Equal(t, expected, actual)
--- a/plugin/storage/badger/samplingstore/storage_test.go
+++ b/plugin/storage/badger/samplingstore/storage_test.go
@@ -45,13 +45,12 @@ func TestInsertThroughput(t *testing.T) {
 
 func TestGetThroughput(t *testing.T) {
 	runWithBadger(t, func(t *testing.T, store *SamplingStore) {
-		start := time.Now()
+		start := time.Now().Truncate(time.Microsecond)
 		expected := []*samplemodel.Throughput{
 			{Service: "my-svc", Operation: "op"},
 			{Service: "our-svc", Operation: "op2"},
 		}
 
-		assert.Eventually(t, func() bool { return true }, time.Second, time.Microsecond*10)
 		err := store.InsertThroughput(expected)
 		require.NoError(t, err)
 


plugin/storage/badger/samplingstore/storage_test.go
```diff
@@ -45,7 +45,7 @@ func TestInsertThroughput(t *testing.T) {
 
 func TestGetThroughput(t *testing.T) {
 	runWithBadger(t, func(t *testing.T, store *SamplingStore) {
-		start := time.Now()
+		start := time.Now().Truncate(time.Millisecond)
```
yurishkuro, line 48: why does it solve the issue?

akagami-harsh: `Truncate(time.Millisecond)` adjusts the time by removing any fractional parts smaller than a millisecond, effectively rounding it down to the nearest millisecond. If the test uses `time.Now()` without truncating, it includes nanoseconds, which might not be stored or retrieved with the same precision in BadgerDB. By truncating to milliseconds, it ensure that the precision used in the test matches the precision stored in the database. i initially tried using `Truncate(time.Millisecond)` but is didn't work.

yurishkuro: > which might not be stored or retrieved with the same precision in BadgerDB

So is the example here would be `start = xxxx.yyy123`, but Badger stores it truncated as `xxxx.yyy`, and then the check is performed `start <= dbobj.start` which fails?

yurishkuro: If yes, can you find out exactly what precision Badger users to store the timestamps? Jaeger's data model captures them in microseconds, so truncating to `us` seems like it should do the trick as well, and be justified, vs. `ms` which feels rather arbitrary.

On a separate note, you may also want to check what precision the query service allows for timestamps - that _might_ actually be in `ms` (easier for the UI).

akagami-harsh: Yes, that's a plausible scenario.

akagami-harsh: > If yes, can you find out exactly what precision Badger users to store the timestamps? 

tried but couldn't find.

>Jaeger's data model captures them in microseconds, so truncating to us seems like it should do the trick as well, and be justified, vs. ms which feels rather arbitrary.

you got a point. yeah! it should have worked with `us` . i am also confused. i think this needs some more research

yurishkuro: This is why it's useful to reproduce the issue with sufficient observability, before trying to fix it. For example, if you suspect that the issue was due to time precision mismatch, you could add a log statement right where the comparison happen. May not always work if you call DB with "where ts1 < ts2" kind of query, but it seems your original fix attempt already identified a comparison in the code you thought was responsible, so that would be a candidate for logging. And then run the test until it fails and look at the timestamps logged to see if your hypothesis is confirmed by data.

btw this is where Badger uses query timestamp, it represents it in microseconds

https://github.com/jaegertracing/jaeger/blob/09f7a30ad705c0d9838032e19a7da1417781ad77/plugin/storage/badger/spanstore/reader.go#L481

akagami-harsh: i think, i have figured it out. 
so what happens is when we try to use microseconds precision sometimes the timestamp becomes equal like this
image

and the `t.After(start)` tries to compare equal timestamps and returns `false`  and the test fails


We can truncate it to microseconds. 
```go 
start := time.Now().Truncate(time.Microsecond)
```
and use a condition to handle the case when the timestamp becomes equal 
something like this. and it works fine
```go
	if (t.After(start) || t.Equal(start)) && (t.Before(end) || t.Equal(end)) {
				retSlice = append(retSlice, throughputs...)
	}
```

akagami-harsh: Or we can just add a `time.Sleep` or `assert.Eventually`(as you suggested)  in the test. In this case we don't really need to truncate start timestamp to `us` because it is already in `us` precision.

plugin/storage/badger/samplingstore/storage_test.go
```diff
@@ -50,6 +50,8 @@ func TestGetThroughput(t *testing.T) {
 			{Service: "my-svc", Operation: "op"},
 			{Service: "our-svc", Operation: "op2"},
 		}
+
+		assert.Eventually(t, func() bool { return true }, time.Second, time.Microsecond*10)
```
yurishkuro, line 54: This doesn't do anything. The point was to wait on some business specific condition.

akagami-harsh: i didn't find any solid condition there to wait on. i think modifying the condition in `storage.go` would be a better approach because it doesn't rely on arbitrary delays, which can be unreliable and may not scale well with different testing environments. It also makes the intent clearer that operations occurring exactly at the start time are included in the results.

plugin/storage/badger/samplingstore/storage.go
```diff
@@ -91,12 +91,12 @@ func (s *SamplingStore) GetThroughput(start, end time.Time) ([]*model.Throughput
...
 				return err
 			}
-			throughputs, err := decodeThroughtputValue(val)
+			throughputs, err := decodeThroughputValue(val)
 			if err != nil {
 				return err
 			}
 
-			if t.After(start) && (t.Before(end) || t.Equal(end)) {
+			if (t.After(start) || t.Equal(start)) && (t.Before(end) || t.Equal(end)) {
```
yurishkuro, line 99: Why is this the right change?

akagami-harsh: i made this changes to handle the condition when timestamps sometimes becomes equal (you can see at the top left corner) .
image

>Why is this the right change?

Because it tackles the issue at the root cause, and we don't have to rely on some arbitrary delay. It will also ensure that operations occurring exactly at the start time are included in the results.

yurishkuro: > you can see at the top left corner

I am not sure what you're pointing at - the two INFO logs? The timestamps there are truncated to full seconds, so impossible to say anything (since we know already the difference in milliseconds affects the behavior).

> Because it tackles the issue at the root cause

What is your evidence of that? This change modifies the semantics of the query, changing timestamp interval treatment from open-closed to closed-closed.

akagami-harsh: > The timestamps there are truncated to full seconds

The timestamps are not truncated to the full second. It has a precision of microseconds, which means it includes fractions of a second up to six decimal places (624305 microseconds) 
The presence of digits after the decimal point in the seconds field indicates microsecond precision.
Truncation to the full second would remove those digits, leaving only 21:35:23

akagami-harsh: Here is another example of a passing test 
image
there is as slight difference of 1 microsecond

yurishkuro: Sorry, I don't understand what you are showing with this screenshot

akagami-harsh: i am trying to show that the test passes when there is some difference in the timestamps, and it fails when the timestamps are same.

yurishkuro: how does the screenshot illustrate that?

akagami-harsh: sorry, the above screenshot doesn't shows full screen shot
here are full logs of the **passed test**
```
harsh@MSI:~/jaeger/plugin/storage/badger/samplingstore$ go test
badger 2024/01/15 01:54:04 INFO: All 0 tables opened in 0s
badger 2024/01/15 01:54:04 INFO: Discard stats nextEmptySlot: 0
badger 2024/01/15 01:54:04 INFO: Set nextTxnTs to 0
badger 2024/01/15 01:54:04 INFO: Lifetime L0 stalled for: 0s
badger 2024/01/15 01:54:04 INFO:
Level 0 [ ]: NumTables: 01. Size: 282 B of 0 B. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 64 MiB
Level 1 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 2 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 3 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 4 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 5 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 6 [B]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level Done
badger 2024/01/15 01:54:04 INFO: All 0 tables opened in 0s
badger 2024/01/15 01:54:04 INFO: Discard stats nextEmptySlot: 0
badger 2024/01/15 01:54:04 INFO: Set nextTxnTs to 0
usec:  2024-01-15 01:54:04.181913 +0530 IST
start:  2024-01-15 01:54:04.181912 +0530 IST
badger 2024/01/15 01:54:04 INFO: Lifetime L0 stalled for: 0s
badger 2024/01/15 01:54:04 INFO:
Level 0 [ ]: NumTables: 01. Size: 282 B of 0 B. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 64 MiB
Level 1 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 2 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 3 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 4 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 5 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 6 [B]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level Done
badger 2024/01/15 01:54:04 INFO: All 0 tables opened in 0s
badger 2024/01/15 01:54:04 INFO: Discard stats nextEmptySlot: 0
badger 2024/01/15 01:54:04 INFO: Set nextTxnTs to 0
badger 2024/01/15 01:54:04 INFO: Lifetime L0 stalled for: 0s
badger 2024/01/15 01:54:04 INFO:
Level 0 [ ]: NumTables: 01. Size: 278 B of 0 B. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 64 MiB
Level 1 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 2 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 3 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 4 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 5 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 6 [B]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level Done
badger 2024/01/15 01:54:04 INFO: All 0 tables opened in 0s
badger 2024/01/15 01:54:04 INFO: Discard stats nextEmptySlot: 0
badger 2024/01/15 01:54:04 INFO: Set nextTxnTs to 0
badger 2024/01/15 01:54:04 INFO: Lifetime L0 stalled for: 0s
badger 2024/01/15 01:54:04 INFO:
Level 0 [ ]: NumTables: 01. Size: 332 B of 0 B. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 64 MiB
Level 1 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 2 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 3 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 4 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 5 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 6 [B]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level Done
PASS
ok      github.com/jaegertracing/jaeger/plugin/storage/badger/samplingstore     0.496s
```

you can see in the logs that there is a difference of 1 microsecond and because of that difference test passed 
```
usec:  2024-01-15 01:54:04.181913 +0530 IST
start:  2024-01-15 01:54:04.181912 +0530 IST
```

akagami-harsh: here are logs of failing test 
```
badger 2024/01/15 02:03:49 INFO: All 0 tables opened in 0s
badger 2024/01/15 02:03:49 INFO: Discard stats nextEmptySlot: 0
badger 2024/01/15 02:03:49 INFO: Set nextTxnTs to 0
usec:  2024-01-15 02:03:49.674922 +0530 IST
start:  2024-01-15 02:03:49.674922 +0530 IST
badger 2024/01/15 02:03:49 INFO: Lifetime L0 stalled for: 0s
badger 2024/01/15 02:03:49 INFO:
Level 0 [ ]: NumTables: 01. Size: 282 B of 0 B. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 64 MiB
Level 1 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 2 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 3 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 4 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 5 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 6 [B]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level Done
--- FAIL: TestGetThroughput (0.06s)
    storage_test.go:59:
                Error Trace:    /home/harsh/jaeger/plugin/storage/badger/samplingstore/storage_test.go:59
                                                        /home/harsh/jaeger/plugin/storage/badger/samplingstore/storage_test.go:143
                                                        /home/harsh/jaeger/plugin/storage/badger/samplingstore/storage_test.go:47
                Error:          Not equal:
                                expected: []*model.Throughput{(*model.Throughput)(0xc000400150), (*model.Throughput)(0xc000400180)}
                                actual  : []*model.Throughput(nil)

                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1,15 +1,2 @@
                                -([]*model.Throughput) (len=2) {
                                - (*model.Throughput)({
                                -  Service: (string) (len=6) "my-svc",
                                -  Operation: (string) (len=2) "op",
                                -  Count: (int64) 0,
                                -  Probabilities: (map[string]struct {}) <nil>
                                - }),
                                - (*model.Throughput)({
                                -  Service: (string) (len=7) "our-svc",
                                -  Operation: (string) (len=3) "op2",
                                -  Count: (int64) 0,
                                -  Probabilities: (map[string]struct {}) <nil>
                                - })
                                -}
                                +([]*model.Throughput) <nil>
```

you can see the timestamp is same 
```
usec:  2024-01-15 02:03:49.674922 +0530 IST
start:  2024-01-15 02:03:49.674922 +0530 IST
```

yurishkuro: what are `usec` and `start`? who logs them? where are they measured?

akagami-harsh: those are just print statements that i used to print timestamps.

akagami-harsh:
```go
func (s *SamplingStore) GetThroughput(start, end time.Time) ([]*model.Throughput, error) {
	var retSlice []*model.Throughput
	prefix := []byte{throughputKeyPrefix}

	err := s.store.View(func(txn *badger.Txn) error {
		opts := badger.DefaultIteratorOptions
		it := txn.NewIterator(opts)
		defer it.Close()

		val := []byte{}
		for it.Seek(prefix); it.ValidForPrefix(prefix); it.Next() {
			item := it.Item()
			k := item.Key()
			startTime := k[1:9]
			val, err := item.ValueCopy(val)
			if err != nil {
				return err
			}
			t, err := initalStartTime(startTime)
			if err != nil {
				return err
			}
			throughputs, err := decodeThroughputValue(val)
			if err != nil {
				return err
			}
			fmt.Println("start: ", start.Truncate(time.Microsecond))
			if t.After(start) && (t.Before(end) || t.Equal(end)) {
				retSlice = append(retSlice, throughputs...)
			}
		}
		return nil
	})
	if err != nil {
		return nil, err
	}

	return retSlice, nil
}
```

```go
func initalStartTime(timeBytes []byte) (time.Time, error) {
	var usec int64

	buf := bytes.NewReader(timeBytes)

	if err := binary.Read(buf, binary.BigEndian, &usec); err != nil {
		return time.Time{}, err
	}
	fmt.Println("usec: ", time.UnixMicro(usec).Truncate(time.Microsecond))
	t := time.UnixMicro(usec)
	return t, nil
}
```

yurishkuro: These print statements are in unrelated placed, they do not tell me anything. What I would rather see is a print before `if t.After(start) && (t.Before(end) || t.Equal(end))` that outputs `t`, `start`, and `end`, with max precision. Then we can compare their values in successful and failed tests.

akagami-harsh: logs of a successful test.
```
harsh@MSI:~/jaeger/plugin/storage/badger/samplingstore$ go test
badger 2024/01/15 10:55:41 INFO: All 0 tables opened in 0s
badger 2024/01/15 10:55:41 INFO: Discard stats nextEmptySlot: 0
badger 2024/01/15 10:55:41 INFO: Set nextTxnTs to 0
badger 2024/01/15 10:55:41 INFO: Lifetime L0 stalled for: 0s
badger 2024/01/15 10:55:41 INFO:
Level 0 [ ]: NumTables: 01. Size: 282 B of 0 B. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 64 MiB
Level 1 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 2 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 3 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 4 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 5 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 6 [B]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level Done
badger 2024/01/15 10:55:41 INFO: All 0 tables opened in 0s
badger 2024/01/15 10:55:41 INFO: Discard stats nextEmptySlot: 0
badger 2024/01/15 10:55:41 INFO: Set nextTxnTs to 0
---------timestamps-----------
t:  2024-01-15 10:55:41.423455 +0530 IST
start:  2024-01-15 10:55:41.423454 +0530 IST
end:  2024-01-15 10:55:51.423454 +0530 IST
badger 2024/01/15 10:55:41 INFO: Lifetime L0 stalled for: 0s
badger 2024/01/15 10:55:41 INFO:
Level 0 [ ]: NumTables: 01. Size: 282 B of 0 B. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 64 MiB
Level 1 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 2 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 3 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 4 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 5 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 6 [B]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level Done
badger 2024/01/15 10:55:41 INFO: All 0 tables opened in 0s
badger 2024/01/15 10:55:41 INFO: Discard stats nextEmptySlot: 0
badger 2024/01/15 10:55:41 INFO: Set nextTxnTs to 0
badger 2024/01/15 10:55:41 INFO: Lifetime L0 stalled for: 0s
badger 2024/01/15 10:55:41 INFO:
Level 0 [ ]: NumTables: 01. Size: 278 B of 0 B. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 64 MiB
Level 1 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 2 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 3 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 4 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 5 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 6 [B]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level Done
badger 2024/01/15 10:55:41 INFO: All 0 tables opened in 0s
badger 2024/01/15 10:55:41 INFO: Discard stats nextEmptySlot: 0
badger 2024/01/15 10:55:41 INFO: Set nextTxnTs to 0
badger 2024/01/15 10:55:41 INFO: Lifetime L0 stalled for: 0s
badger 2024/01/15 10:55:41 INFO:
Level 0 [ ]: NumTables: 01. Size: 332 B of 0 B. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 64 MiB
Level 1 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 2 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 3 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 4 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 5 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 6 [B]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level Done
PASS
ok      github.com/jaegertracing/jaeger/plugin/storage/badger/samplingstore     0.268s
```


logs of a failed test
```
badger 2024/01/15 10:56:38 INFO: All 0 tables opened in 0s
badger 2024/01/15 10:56:38 INFO: Discard stats nextEmptySlot: 0
badger 2024/01/15 10:56:38 INFO: Set nextTxnTs to 0
badger 2024/01/15 10:56:38 INFO: Lifetime L0 stalled for: 0s
badger 2024/01/15 10:56:38 INFO:
Level 0 [ ]: NumTables: 01. Size: 282 B of 0 B. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 64 MiB
Level 1 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 2 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 3 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 4 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 5 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 6 [B]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level Done
badger 2024/01/15 10:56:38 INFO: All 0 tables opened in 0s
badger 2024/01/15 10:56:38 INFO: Discard stats nextEmptySlot: 0
badger 2024/01/15 10:56:38 INFO: Set nextTxnTs to 0
---------timestamps-----------
t:  2024-01-15 10:56:38.23525 +0530 IST
start:  2024-01-15 10:56:38.23525 +0530 IST
end:  2024-01-15 10:56:48.23525 +0530 IST
badger 2024/01/15 10:56:38 INFO: Lifetime L0 stalled for: 0s
badger 2024/01/15 10:56:38 INFO:
Level 0 [ ]: NumTables: 01. Size: 282 B of 0 B. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 64 MiB
Level 1 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 2 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 3 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 4 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 5 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 6 [B]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level Done
--- FAIL: TestGetThroughput (0.06s)
    storage_test.go:59:
                Error Trace:    /home/harsh/jaeger/plugin/storage/badger/samplingstore/storage_test.go:59
                                                        /home/harsh/jaeger/plugin/storage/badger/samplingstore/storage_test.go:143
                                                        /home/harsh/jaeger/plugin/storage/badger/samplingstore/storage_test.go:47
                Error:          Not equal:
                                expected: []*model.Throughput{(*model.Throughput)(0xc0002c34a0), (*model.Throughput)(0xc0002c34d0)}
                                actual  : []*model.Throughput(nil)

                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1,15 +1,2 @@
                                -([]*model.Throughput) (len=2) {
                                - (*model.Throughput)({
                                -  Service: (string) (len=6) "my-svc",
                                -  Operation: (string) (len=2) "op",
                                -  Count: (int64) 0,
                                -  Probabilities: (map[string]struct {}) <nil>
                                - }),
                                - (*model.Throughput)({
                                -  Service: (string) (len=7) "our-svc",
                                -  Operation: (string) (len=3) "op2",
                                -  Count: (int64) 0,
                                -  Probabilities: (map[string]struct {}) <nil>
                                - })
                                -}
                                +([]*model.Throughput) <nil>

                Test:           TestGetThroughput
```

plugin/storage/badger/samplingstore/storage_test.go
```diff
@@ -45,11 +45,12 @@ func TestInsertThroughput(t *testing.T) {
 
 func TestGetThroughput(t *testing.T) {
 	runWithBadger(t, func(t *testing.T, store *SamplingStore) {
-		start := time.Now()
+		start := time.Now().Truncate(time.Microsecond)
```
yurishkuro, line 48: So this alone does not fix the issue?

akagami-harsh: no

plugin/storage/badger/samplingstore/storage_test.go
```diff
@@ -45,11 +45,12 @@ func TestInsertThroughput(t *testing.T) {
...
 		expected := []*samplemodel.Throughput{
 			{Service: "my-svc", Operation: "op"},
 			{Service: "our-svc", Operation: "op2"},
 		}
+
 		err := store.InsertThroughput(expected)
 		require.NoError(t, err)
```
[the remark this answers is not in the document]
akagami-harsh: I've re-run the tests with this modification, and they now pass consistently also without truncating.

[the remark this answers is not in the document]
akagami-harsh: applied suggested changes.

yurishkuro (approved):

Issue #5068, opened 2024-01-02
Implement missing unit tests


yurishkuro:
Our CI enforces that all packages with .go files must have at least one _test.go file, to ensure the package is included in the code coverage, OR that it has a `.nocover` file when testing it is very difficult (e.g. requires external DB connection).

However, we somewhat abused the `.nocover` capability and have a bunch of packages that *could* have tests with high coverage. This ticket is about incrementally fixing those, which is a good first issue since the changes can be reasonably small if scoped to one package.

To see which packages need these tests, run `make nocover`.

Regarding `./examples/`, where lots of `.nocover` files are found today, I think we can replace them with `empty_test.go` as we do in some other dirs, because examples will still be excluded from the overall coverage percentage, so there's no need to be excluding them via `.nocover` mechanism (it just creates noise).

## Update Jun 17 2024

We are actually down to a single package remaining
```
$ make nocover
Verifying that all packages have test files to count in coverage
Package excluded from coverage: ./pkg/es/config/
  reason: requires connection to Elasticsearch
```
The reason stated in that package is not very valid - it only checks for the version of the server which can be easily mocked with a simple HTTP server as we already do in other ES-related tests. The majority of methods in the package do not require connection at all and can be 100% covered by tests.

chirag-ghosh:
I would like to work on this

yurishkuro:
@chirag-ghosh thanks. I suggest you pick one of the packages and maybe post here that you're working on it, in case someone else also wants to pick another package. Don't combine multiple packages in one PR.

chirag-ghosh:
Sure. Since I'm new to the project, I'll first setup it in my local and the comment the package I'm working on ASAP.

chirag-ghosh:
I am working on the storageexporter package (https://github.com/jaegertracing/jaeger/tree/main/cmd/jaeger/internal/exporters/storageexporter)

akagami-harsh:
working on `cmd/jeager/internal`

Pushkarm029:
working on `./pkg/version/`

chirag-ghosh:
Working on jaegerstorage

tarhphilomina:
working on [Jaeger/plugin/](https://github.com/jaegertracing/jaeger/tree/main/plugin)

VaibhavMalik4187:
I'd like to work on `internal/tracegen`

VaibhavMalik4187:
[...]

PR #5096, opened 2024-01-12
Add tests for cmd/jaeger/internal/extension/jaegerstorage

chirag-ghosh:
- #5068 

## Description of the changes
-  Add unit tests for jaegerstorage

- 

## Checklist


cmd/jaeger/internal/extension/jaegerstorage/extension_test.go
```diff
@@ -0,0 +1,115 @@
...
+	require.EqualError(t, err, fmt.Sprintf("cannot find extension '%s' (make sure it's defined earlier in the config)", ID))
+}
+
+func TestStorageExtension(t *testing.T) {
+	ctx := context.Background()
+	const memstoreName = "memstore"
+
+	storageExtension := makeStorageExtension(t, memstoreName)
+
+	host := storageHost{t: t, storageExtension: storageExtension}
```
yurishkuro, line 83: Why do you need custom host that simply returns back the extension itself? Would it not work with NopHost?

yurishkuro (commented):
Another suggestion: the config does not have a Validate function like the other components have, can you add it? And then you can add a test calling getDefaultConfig and checking that Validate returns an error (because I don't think empty config is valid for this extension).

cmd/jaeger/internal/extension/jaegerstorage/extension_test.go
```diff
@@ -0,0 +1,115 @@
...
+func TestGetStorageFactoryError(t *testing.T) {
+	ctx := context.Background()
+	const memstoreName = "memstore"
+
+	storageExtension := makeStorageExtension(t, memstoreName)
+
+	host := componenttest.NewNopHost()
+	err := storageExtension.Start(ctx, host)
+	require.NoError(t, err)
+	t.Cleanup(func() { require.NoError(t, storageExtension.Shutdown(ctx)) })
```
yurishkuro, line 70: could this be moved to makeStorageExtension? I am not sure if it's safe to call shutdown before Start

yurishkuro:
btw very nice PR?

yurishkuro:
Nit: be more specific in the PR title - use full package path

cmd/jaeger/internal/extension/jaegerstorage/extension_test.go
yurishkuro: nvm

cmd/jaeger/internal/extension/jaegerstorage/extension_test.go
chirag-ghosh: Done. Although t.Cleanup should ensure that the function is called after all tests are done. Atleast that's what the function description is.

chirag-ghosh:
The config did not have any required fields. `Memory` seems to be a required field atleeast till other storages are supported. I have therefore made it required. Please check.

cmd/jaeger/internal/extension/jaegerstorage/config.go
```diff
@@ -19,3 +21,8 @@ type MemoryStorage struct {
 	Name string `mapstructure:"name"`
 	memoryCfg.Configuration
 }
+
+func (cfg *Config) Validate() error {
+	_, err := govalidator.ValidateStruct(cfg)
```
yurishkuro, line 26: Memory isn't required, what's required is that at least one storage is present. You don't need to use external validator for that.

chirag-ghosh: Made changes. Please review

cmd/jaeger/internal/extension/jaegerstorage/config.go
```diff
@@ -19,3 +22,12 @@ type MemoryStorage struct {
 	Name string `mapstructure:"name"`
 	memoryCfg.Configuration
 }
+
+func (cfg *Config) Validate() error {
+	emptyCfg := createDefaultConfig().(*Config)
+	if reflect.DeepEqual(*cfg, *emptyCfg) {
+		return errors.New("no storage type present in config")
```
yurishkuro, line 29:
```suggestion
		return fmt.Errorf("%s: no storage type present in config", ID)
```

cmd/jaeger/internal/extension/jaegerstorage/extension_test.go
```diff
@@ -0,0 +1,115 @@
...
+
+type storageHost struct {
+	t                *testing.T
+	storageExtension component.Component
+}
+
+func (host storageHost) GetExtensions() map[component.ID]component.Component {
+	myMap := make(map[component.ID]component.Component)
+	myMap[ID] = host.storageExtension
+	return myMap
```
yurishkuro, line 30:
```suggestion
	return map[component.ID]component.Component {
	   ID: host.storageExtension,
	}
```

cmd/jaeger/internal/extension/jaegerstorage/extension_test.go
```diff
@@ -0,0 +1,115 @@
...
+
+func TestExtensionConfigError(t *testing.T) {
+	config := createDefaultConfig().(*Config)
+	err := config.Validate()
+	require.EqualError(t, err, "no storage type present in config")
+}
+
+func TestStartStorageExtensionError(t *testing.T) {
+	ctx := context.Background()
+	const memstoreName = "memstore"
```
yurishkuro, line 53: Pull to file level, too many uses

yurishkuro (commented):
Lgtm

cmd/jaeger/internal/extension/jaegerstorage/extension_test.go
```diff
@@ -0,0 +1,112 @@
...
+
+func TestStartStorageExtensionError(t *testing.T) {
+	ctx := context.Background()
+
+	storageExtension := makeStorageExtension(t, memstoreName)
+
+	host := componenttest.NewNopHost()
+	err := storageExtension.Start(ctx, host)
+	require.Error(t, err)
+	require.EqualError(t, err, fmt.Sprintf("duplicate memory storage name %s", memstoreName))
```
yurishkuro, line 61: Where is the error in this case?

chirag-ghosh: We are starting the storageExtension again. This will lead to an error. [Ref](https://github.com/jaegertracing/jaeger/blob/main/cmd/jaeger/internal/extension/jaegerstorage/extension.go#L65)

cmd/jaeger/internal/extension/jaegerstorage/extension_test.go
```diff
@@ -0,0 +1,112 @@
...
+	return nil
+}
+
+func TestExtensionConfigError(t *testing.T) {
+	config := createDefaultConfig().(*Config)
+	err := config.Validate()
+	require.EqualError(t, err, fmt.Sprintf("%s: no storage type present in config", ID))
+}
+
+func TestStartStorageExtensionError(t *testing.T) {
```
yurishkuro, line 53:
```suggestion
func TestStartStorageExtensionTwiceError(t *testing.T) {
```

cmd/jaeger/internal/extension/jaegerstorage/extension_test.go
yurishkuro: but the errer message doesn't seem to be about extension starting.

yurishkuro: If you're testing specifically the dual entry condition, it would be better to test it explicitly, not via a side-effect of starting extension twice. I assume it's possible to create dual entry in the config if using a different type of storage and then giving it the same name.

chirag-ghosh: Currently only Memory type storage is supported. So I was unable to recreate this. Have changed the function name for now.

yurishkuro (changes requested):
1. you can have a test for bad name passed to GetStorageFactory
2. you can have a test for error path in Shutdown by manually inserting a mock Factory into `extension.factories` which throws an error on Close

chirag-ghosh:
1. Done
2. extension.factories isn't supposed to be accessible by public methods. The only place it is set is [https://github.com/jaegertracing/jaeger/blob/3c9da0df1a28a7191a74a5c48a7f7537c0098ae4/cmd/jaeger/internal/extension/jaegerstorage/extension.go#L67](here)

yurishkuro:
> extension.factories isn't supposed to be accessible by public methods.

tests can be defined in the same package and have access to private members

chirag-ghosh:
Done. Please check

cmd/jaeger/internal/extension/jaegerstorage/extension_test.go
```diff
@@ -0,0 +1,165 @@
...
+func (storageHost) GetExporters() map[component.DataType]map[component.ID]component.Component {
+	return nil
+}
+
+type errorFactory struct {
+	closeErr error
+}
+
+func (e errorFactory) Initialize(metricsFactory metrics.Factory, logger *zap.Logger) error {
+	panic("implement me")
```
yurishkuro, line 56: nit: change to "not implemented", otherwise it reads as if these *need* to be implemented

cmd/jaeger/internal/extension/jaegerstorage/extension_test.go
```diff
@@ -0,0 +1,165 @@
...
+func TestBadHostGetStorageFactoryError(t *testing.T) {
+	makeStorageExtension(t, memstoreName)
+
+	host := componenttest.NewNopHost()
+	_, err := GetStorageFactory(memstoreName, host)
+	require.Error(t, err)
+	require.EqualError(t, err, fmt.Sprintf("cannot find extension '%s' (make sure it's defined earlier in the config)", ID))
+}
+
+func TestBadNameGetStorageFactoryError(t *testing.T) {
```
yurishkuro, line 101: nit: the test names are meant to be hierarchical {componentName}{testScenario}. I.e. everything here should have the prefix TestStorageFactory***

cmd/jaeger/internal/extension/jaegerstorage/extension_test.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+func TestBadFactoryShutdownError(t *testing.T) {
+	storageExtension := storageExt{
+		factories: make(map[string]storage.Factory),
+	}
+	badFactoryError := fmt.Errorf("error factory")
+	storageExtension.factories[memstoreName] = errorFactory{closeErr: badFactoryError}
+
+	t.Cleanup(func() {
```
yurishkuro, line 119: nit: no need to do this as cleanup, move to the main code of the test

cmd/jaeger/internal/extension/jaegerstorage/extension_test.go
```diff
@@ -0,0 +1,165 @@
...
+		err := storageExtension.Shutdown(context.Background())
+		require.Error(t, err)
+		require.EqualError(t, err, badFactoryError.Error())
```
yurishkuro, line 122: would require.ErrorIs() work here? And then you can skip L121

cmd/jaeger/internal/extension/jaegerstorage/extension_test.go
chirag-ghosh: Done. Maybe we should also change [this](https://github.com/jaegertracing/jaeger/blob/main/plugin/storage/factory_test.go#L443) then.

yurishkuro (approved):
Thanks!

Issue #5079, opened 2024-01-04
Implement Storage API v2 that operates on OTLP batches


yurishkuro:
## Summary

In OTEL-based Jaeger-v2 we don't want to force OTLP trace data to go through unnecessary transformation into Jaeger's current data model. We want to define a V2 version of the storage API so that the OTLP data can be passed through the receiver-exporter pipeline without additional conversions, for better efficiency.

## Problem

More discussion in the doc https://docs.google.com/document/d/1s4_6VgAS7qAVp6iEm5KYvpiGw3h2Ja5T5HpKo29iv00/edit#heading=h.hlhu8pegwdcj

## Progress

* [x] Define Storage V2 API #5399
* Write path
  * [x] Change `jaegerstorage` extension to expose Storage V2 API #5497
  * [x] Change `storageexporter` to use V2 API exclusively #5497
    * This is achieved by using an adapter that provides v2 API over v1 storage implementation, for writes only
* Read path (more complicated)
  * discussed in the [Migration](https://docs.google.com/document/d/1s4_6VgAS7qAVp6iEm5KYvpiGw3h2Ja5T5HpKo29iv00/edit?tab=t.0#heading=h.qi7357ivoogu) section of the doc
* [x] Remove `jaegerstorage.GetStorageFactory` once not used

tarhphilomina:
Can I take  this issue ?

yurishkuro:
you can try - this is a design issue

mahadzaryab1:
@yurishkuro For the read path implementation, do we just want to expand the v2 adapter to implement `spanstore.Reader` from v2 that just wraps the v1 implementation?

yurishkuro:
Not quite, please see the Google doc.

mahadzaryab1:
@yurishkuro Here are the remaining usages of `GetStorageFactory` that cannot be removed right now because they do runtime casts to other factories which the factoryadapter does not implement currently:
- Remote Sampling Extension - requires a cast to `storage.SamplingStoreFactory` (https://github.com/jaegertracing/jaeger/blob/main/cmd/jaeger/internal/extension/remotesampling/extension.go#L184-L189)
- Storage Cleaner - requires a cast to `storage.Purger` (https://github.com/jaegertracing/jaeger/blob/main/cmd/jaeger/internal/integration/storagecleaner/extension.go#L48-L54)
[...]

PR #5098, opened 2024-01-14
Use official OTLP types in api_v3 and avoid triple-marshaling

yurishkuro:
- Part of #5079
- Avoid triple marshaling due to our use of internally generated OTLP proto types, which prevents us from directly using the output of jaeger->otlp conversion from collector contrib.

## Description of the changes
- 🛑 breaking: the JSON format is changed to match [OTLP-JSON][otlp-json], specifically
  - the trace & span IDs are returned as hex-encoded strings, not base64 as required by Proto-JSON
  - enums are returned as integers, not strings
- Use Proposal 1 from https://github.com/open-telemetry/opentelemetry-collector/issues/9233#issuecomment-1883868433
- API-V3 proto is already declared to use official OTLP types; remove the overrides to our internally generated OTLP proto types
- Replace `SpansResponseChunk` with official `otlp.TracesData`, but override it internally to use a custom type
- Depends on https://github.com/jaegertracing/jaeger-idl/pull/103

- Unit tests

[otlp-json]: https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#json-protobuf-encoding


Makefile.Protobuf.mk
```diff
@@ -146,14 +146,13 @@ proto-otel: proto-prepare-otel
...
+# which at runtime is mapped to a custom type in cmd/query/app/internal/api_v3/traces.go
+API_V3_PATH=cmd/query/app/internal/api_v3
 .PHONY: proto-api-v3
 proto-api-v3:
-	$(call print_caption, Enriching OpenTelemetry Protos into $(PATCHED_OTEL_PROTO_DIR))
-	rm -rf $(PATCHED_OTEL_PROTO_DIR)/*
-	cp -R idl/opentelemetry-proto/* $(PATCHED_OTEL_PROTO_DIR)
-	find $(PATCHED_OTEL_PROTO_DIR) -name "*.proto" | xargs -L 1 $(SED) -i 's+go.opentelemetry.io/proto/otlp+github.com/jaegertracing/jaeger/proto-gen/otel+g'
-
-	$(call proto_compile, proto-gen/api_v3, idl/proto/api_v3/query_service.proto, -I$(PATCHED_OTEL_PROTO_DIR))
```
yurishkuro, line 159: * enriching is no longer needed
* types are generated into an internal module now

cmd/all-in-one/all_in_one_test.go
```diff
@@ -180,9 +179,11 @@ func getServicesAPIV3(t *testing.T) {
 	require.NoError(t, err)
 	body, _ := io.ReadAll(resp.Body)
 
-	var servicesResponse api_v3.GetServicesResponse
```
yurishkuro, line 183: This was unnecessary coupling. `api_v3` is not internal and not accessible from here.

cmd/query/app/apiv3/otlp_translator.go
```diff
@@ -15,33 +15,13 @@
...
-	b, err := req.MarshalProto()
-	if err != nil {
-		return nil, fmt.Errorf("cannot marshal OTLP: %w", err)
-	}
-	// use api_v3.SpansResponseChunk which has the same shape as otlp.ExportTraceServiceRequest
-	var chunk api_v3.SpansResponseChunk
-	if err := proto.Unmarshal(b, &chunk); err != nil {
-		return nil, fmt.Errorf("cannot marshal OTLP: %w", err)
-	}
-	return chunk.ResourceSpans, nil
```
yurishkuro, line 46: no triple marshaling!

jkowall (approved):
LGTM! pretty big PR and chunk of work. Nice job.

PR #5109, opened 2024-01-16
Update Makefile

RupeshHacker:
Changed Go Image version from 1.21-alpine
 to 1.22rc1-alpine3.19


yurishkuro:
we don't use RC for official builds

PR #5123, opened 2024-01-19
Add tests for internal/extension/jaegerquery

### LINKED ISSUE #5068: Implement missing unit tests

VaibhavMalik4187:
- Part of #5068

## Description of the changes
- This commit adds tests for the `cmd/jaeger/internal/extension/jaegerquery` package.

- make test

## Checklist


cmd/jaeger/internal/extension/jaegerquery/config_test.go
```diff
@@ -0,0 +1,52 @@
...
+			name:        "Empty config",
+			config:      &Config{},
+			expectedErr: errors.New("TraceStoragePrimary: non zero value required"),
+		},
+		{
+			name: "Non empty-config",
+			config: &Config{
+				TraceStoragePrimary: "some-storage",
+			},
+			expectedErr: errors.New("TraceStoragePrimary: non zero value required"),
```
yurishkuro, line 40: is it expected?

cmd/jaeger/internal/extension/jaegerquery/config_test.go
```diff
@@ -0,0 +1,52 @@
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := tt.config.Validate()
+			if err != nil {
```
yurishkuro, line 47: this is a leaky test - you check for error only if there an error. The usual way we do this is
```
if tt.expectedErr == "" { // note: expectedErr does not need to be of type error 
  require.NoError(t, err)
} else {
  require.ErrorContains(...)
}
```

cmd/jaeger/internal/extension/jaegerquery/factory_test.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+func Test_NewFactory(t *testing.T) {
+	factory := NewFactory()
+	assert.Equal(t, factory.Type(), component.Type("jaeger_query"))
+	assert.Equal(t, factory.CreateDefaultConfig(), createDefaultConfig())
+}
+
+func Test_CreateDefaultConfig(t *testing.T) {
+	expectedDefaultConfig := &Config{
```
yurishkuro, line 35: I don't like this - if default config keeps growing in the future, we'd need to keep duplicating it here. What benefit would this test provide?

cmd/jaeger/internal/extension/jaegerquery/server_test.go
```diff
@@ -0,0 +1,87 @@
...
+// mockHost implements the Host interface for testing purposes
+type mockHost struct {
+	logger zap.Logger
+	// Captures arguments for expected methods
+	reportFatalErrorCalled     bool
+	reportFatalErrorArg        error
+	getFactoryCalled           bool
+	getFactoryKindArg          component.Kind
+	getFactoryComponentTypeArg component.Type
+	getExtensionsCalled        bool
```
yurishkuro, line 35: I don't see you using any of these. If you don't need special behavior of the host you can use Nop available in OTEL (you can check other tests in parallel packages using it)

yurishkuro (commented):
please provide output of
```
go test -cover ./cmd/jaeger/internal/extension/jaegerquery/
```

cmd/jaeger/internal/extension/jaegerquery/config_test.go
VaibhavMalik4187: It was not, thanks for pointing this out.

cmd/jaeger/internal/extension/jaegerquery/config_test.go
VaibhavMalik4187: Understood, thanks!

cmd/jaeger/internal/extension/jaegerquery/factory_test.go
VaibhavMalik4187: Replaced the config with a non-empty check to make sure that the default config is not an empty object.

cmd/jaeger/internal/extension/jaegerquery/server_test.go
VaibhavMalik4187: Removed the unnecessary variables in the latest version

VaibhavMalik4187:
> please provide output of
> 
> ```
> go test -cover ./cmd/jaeger/internal/extension/jaegerquery/
> ```

`ok  	github.com/jaegertracing/jaeger/cmd/jaeger/internal/extension/jaegerquery	0.013s	coverage: 84.1% of statements`

cmd/jaeger/internal/extension/jaegerquery/factory_test.go
```diff
@@ -0,0 +1,35 @@
...
+func Test_CreateDefaultConfig(t *testing.T) {
+	defaultConfig := createDefaultConfig()
+	assert.NotEmpty(t, defaultConfig)
+}
+
+func Test_CreateExtension(t *testing.T) {
+	set := extension.CreateSettings{
+		ID: component.NewID(component.Type("jaeger_query")),
+	}
+	defaultConfig := createDefaultConfig()
```
yurishkuro, line 30:
```suggestion
	cfg := createDefaultConfig()
```
it's not relevant to var name that it's "default"

cmd/jaeger/internal/extension/jaegerquery/server_test.go
```diff
@@ -0,0 +1,223 @@
...
+		Logger: zap.NewNop(),
+	}
+	server := newServer(createDefaultConfig().(*Config), telemetrySettings)
+	assert.NotNil(t, server)
+	assert.NotNil(t, server.config)
+	assert.NotNil(t, server.logger)
+}
+
+func Test_Dependencies(t *testing.T) {
+	expectedDependencies := []component.ID{component.NewID(component.Type("jaeger_storage"))}
```
yurishkuro, line 59: move this to just before the assert

yurishkuro: also, is `component.NewID(component.Type("jaeger_storage"))` necessary? There is a declared constant for that

cmd/jaeger/internal/extension/jaegerquery/server_test.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+
+	server := newServer(createDefaultConfig().(*Config), telemetrySettings)
+	dependencies := server.Dependencies()
+
+	assert.Equal(t, expectedDependencies, dependencies)
+}
+
+func Test_Start(t *testing.T) {
+	expectedErrMessage := "cannot find primary storage : cannot find extension 'jaeger_storage' (make sure it's defined earlier in the config)"
```
yurishkuro, line 71: move next to assert, or even inline

cmd/jaeger/internal/extension/jaegerquery/server_test.go
```diff
@@ -0,0 +1,223 @@
...
+	telemetrySettings := component.TelemetrySettings{
+		Logger: zap.NewNop(),
+	}
+
+	server := newServer(createDefaultConfig().(*Config), telemetrySettings)
+	host := componenttest.NewNopHost()
+	err := server.Start(context.Background(), host)
+	assert.Equal(t, expectedErrMessage, err.Error())
```
yurishkuro, line 79: there's no value in matching on the full wrapped error string, you only need to be checking for something specific, e.g.
```
require.ErrorContains(t, err, "cannot find primary storage : cannot find extension")
```

cmd/jaeger/internal/extension/jaegerquery/server_test.go
```diff
@@ -0,0 +1,223 @@
...
+}
+
+func Test_StartStorageHost(t *testing.T) {
```
yurishkuro, line 82: from this name I am not sure what is meant to be tested here

cmd/jaeger/internal/extension/jaegerquery/server_test.go
```diff
@@ -0,0 +1,223 @@
...
+			logger, buf := testutils.NewLogger()
+			telemetrySettings := component.TelemetrySettings{
+				Logger: logger,
+			}
+			server := newServer(tt.config, telemetrySettings)
+			err := server.addArchiveStorage(tt.qSvcOpts, host)
+			if tt.expectedErr == "" {
+				assert.NoError(t, err)
+			} else {
+				assert.Equal(t, tt.expectedErr, err.Error())
```
yurishkuro, line 136:
```suggestion
				assert.ErrorContains(t, err, tt.expectedErr)
```

cmd/jaeger/internal/extension/jaegerquery/server_test.go
```diff
@@ -0,0 +1,223 @@
...
+			}
+
+			assert.Equal(t, tt.expectedOutput, buf.String())
+		})
+	}
+}
+
+func Test_MakeQueryOptions(t *testing.T) {
```
yurishkuro, line 144: Seems like you're going overboard with the tests - isn't this method going to be called anyway from some other tests? The two primary approaches for tests are:
  1. test all public methods
  2. add more tests as needed to improve code coverage

yurishkuro (changes requested):
Need to fix linter errors (run `make lint`)

```
go test -cover ./cmd/jaeger/internal/extension/jaegerquery/
ok github.com/jaegertracing/jaeger/cmd/jaeger/internal/extension/jaegerquery 0.013s coverage: 84.1% of statements
```
needs to be above 95%

VaibhavMalik4187:
> Need to fix linter errors (run `make lint`)
> 
> ```
> go test -cover ./cmd/jaeger/internal/extension/jaegerquery/
> ok github.com/jaegertracing/jaeger/cmd/jaeger/internal/extension/jaegerquery 0.013s coverage: 84.1% of statements
> ```
> 
> needs to be above 95%

I'm trying but the uncovered statements are kind of difficult to reach. For example:
image

cmd/jaeger/internal/extension/jaegerquery/server_test.go
VaibhavMalik4187: Understood, thanks for the pointer.

cmd/jaeger/internal/extension/jaegerquery/server_test.go
VaibhavMalik4187: My bad, fixed this issue in the latest commit.

cmd/jaeger/internal/extension/jaegerquery/server_test.go
VaibhavMalik4187: Testing only the public methods makes sense. Thanks for sharing. I'll keep that in mind.

cmd/jaeger/internal/extension/jaegerquery/server_test.go
```diff
@@ -0,0 +1,231 @@
...
+
+func (storageHost) GetFactory(_ component.Kind, _ component.Type) component.Factory {
+	return nil
+}
+
+func (storageHost) GetExporters() map[component.DataType]map[component.ID]component.Component {
+	return nil
+}
+
+func Test_Dependencies(t *testing.T) {
```
yurishkuro, line 49:
```suggestion
func TestServerDependencies(t *testing.T) {
```

cmd/jaeger/internal/extension/jaegerquery/server_test.go
```diff
@@ -0,0 +1,231 @@
...
+		Logger: zap.NewNop(),
+	}
+
+	server := newServer(createDefaultConfig().(*Config), telemetrySettings)
+	dependencies := server.Dependencies()
+
+	assert.Equal(t, expectedDependencies, dependencies)
+}
+
+func Test_Start(t *testing.T) {
```
yurishkuro, line 61:
```suggestion
func TestServerStart(t *testing.T) {
```

cmd/jaeger/internal/extension/jaegerquery/server_test.go
```diff
@@ -0,0 +1,231 @@
...
+			expectedErr:    "",
+		},
+		{
+			name: "Archive storage set",
+			config: &Config{
+				TraceStorageArchive: "random-value",
+			},
+			qSvcOpts:       &querysvc.QueryServiceOptions{},
+			expectedOutput: "",
+			expectedErr:    "cannot find archive storage factory: cannot find extension 'jaeger_storage' (make sure it's defined earlier in the config)",
```
yurishkuro, line 129:
```suggestion
			expectedErr:    "cannot find archive storage factory: cannot find extension",
```

yurishkuro:
I added example of another test: [7af4dc1](https://github.com/jaegertracing/jaeger/pull/5123/commits/7af4dc1189b2d5ad738c6bc9070c82905a10b180)

`coverage: 86.4% of statements`

yurishkuro:
This is what I would suggest:
  * in order to simulate the errors you really need to be able to provide a fake storage factory
  * the factories are retrieved via `jaegerstorage.GetStorageFactory` util function, which internally looks up the extension in the `host` and tries to cast it to private type to access the factories map `f, ok := comp.(*storageExt).factories[name]`
  * to allow this to be mocked, we can change this to `f, ok := comp.(StorageExt).Factory(name)` with a new interface:
```
type StorageExt interface {
  Factory(string) (storage.Factory, error)
}
```
and have `storageExt` implement this interface. Then in the test you can create a fake implementation of `StorageExt` that returns a fake factory, and store that fake impl in your fake `hostExtension`

yurishkuro:
@VaibhavMalik4187 do you plan to finish this?

VaibhavMalik4187:
> @VaibhavMalik4187 do you plan to finish this?

I'm sorry for being inactive. I'll finish this soon. Thanks for bearing with me.

yurishkuro:
No worries, I just don't want this to rot & get lost, as these are valuable changes.

VaibhavMalik4187:
> No worries, I just don't want this to rot & get lost, as these are valuable changes.

Quick update, I've tried to incorporate your suggestions and now the code coverage of this package has crossed the 95% mark. Could you please go through this PR once again? Many thanks!

cmd/jaeger/internal/extension/jaegerstorage/extension.go
```diff
@@ -44,8 +48,9 @@ func GetStorageFactory(name string, host component.Host) (storage.Factory, error
 			componentType,
 		)
 	}
-	f, ok := comp.(*storageExt).factories[name]
-	if !ok {
+	// f, ok := comp.(*storageExt).factories[name]
+	f, err := comp.(StorageExt).Factory(name)
```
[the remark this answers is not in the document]
VaibhavMalik4187: It's already there :)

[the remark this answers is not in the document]
yurishkuro: where? The previous code was doing a direct map lookup `comp.(*storageExt).factories[name]`

cmd/jaeger/internal/extension/jaegerstorage/extension.go
```diff
@@ -29,6 +29,19 @@ type storageExt struct {
 	factories map[string]storage.Factory
 }
 
+func (se storageExt) Factory(name string) (storage.Factory, error) {
```
VaibhavMalik4187, line 32: @yurishkuro I've implemented the Factory method here.

cmd/jaeger/internal/extension/jaegerquery/server_test.go
```diff
@@ -0,0 +1,308 @@
...
+						},
+					},
+				}
+				tenancyMgr := tenancy.NewManager(&serverOptions.Tenancy)
+				spanReader := &spanstoremocks.Reader{}
+				dependencyReader := &depsmocks.Reader{}
+				querySvc := querysvc.NewQueryService(spanReader, dependencyReader, querysvc.QueryServiceOptions{})
+				queryAppServer, err := queryApp.NewServer(zap.NewNop(), querySvc, nil,
+					serverOptions, tenancyMgr,
+					jtracer.NoOp())
```
yurishkuro, line 295: simplify (remove deps that don't make a difference)
```suggestion
				serverOptions := &queryApp.QueryOptions{
					HTTPHostPort: ":0",
					GRPCHostPort: ":0",
				}
				var tenancyMgr tenancy.Manager
				querySvc := querysvc.NewQueryService(nil, nil, querysvc.QueryServiceOptions{})
				queryAppServer, err := queryApp.NewServer(zap.NewNop(), querySvc, nil,
					serverOptions, &tenancyMgr, jtracer.NoOp())
```

cmd/jaeger/internal/extension/jaegerquery/server_test.go
```diff
@@ -0,0 +1,308 @@
...
+			assert.Contains(t, buf.String(), tt.expectedOutput)
+		})
+	}
+}
+
+func TestServerShutdown(t *testing.T) {
+	tests := []struct {
+		name           string
+		hasQueryServer bool
+	}{
```
yurishkuro, line 260: there is no benefit in making this a data driven test. This would be more readable:
```
t.Run("without query server", (t *testing.T) { /* first test /* })
t.Run("without query server", (t *testing.T) { /* 2nd test /* })
```
You can reuse telemetry settings across these, but aside from that there's little overlap

cmd/jaeger/internal/extension/jaegerquery/server_test.go
```diff
@@ -0,0 +1,308 @@
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			telemetrySettings := component.TelemetrySettings{
+				Logger: zap.NewNop(),
+			}
+			server := newServer(createDefaultConfig().(*Config), telemetrySettings)
+			assert.NotNil(t, server)
```
yurishkuro, line 277: not needed. The test will blow up anyway if it's nil

cmd/jaeger/internal/extension/jaegerquery/server_test.go
```diff
@@ -0,0 +1,308 @@
...
+				spanReader := &spanstoremocks.Reader{}
+				dependencyReader := &depsmocks.Reader{}
+				querySvc := querysvc.NewQueryService(spanReader, dependencyReader, querysvc.QueryServiceOptions{})
+				queryAppServer, err := queryApp.NewServer(zap.NewNop(), querySvc, nil,
+					serverOptions, tenancyMgr,
+					jtracer.NoOp())
+				require.NoError(t, err)
+
+				err = queryAppServer.Start()
+				require.NoError(t, err)
```
yurishkuro, line 299: because you have a bunch of potential exit points here, you should schedule the actual validation of Shutdown as soon as the jaegerquery.server was created:
```
server := newServer(createDefaultConfig().(*Config), telemetrySettings)
defer func() {
  require.NoError(t, server.Shutdown(context.Background()))
}()
```

yurishkuro (commented):
please address all open comments

VaibhavMalik4187:
> please address all open comments

On it.

yurishkuro (approved):

yurishkuro:
Thanks, @VaibhavMalik4187

PR #5126, opened 2024-01-21
Remove test summary reports

### LINKED ISSUE #5114: [Bug] Test reports don't seem to work
@albertteoh maybe it's just me, but I am having no luck with this new test reports thing. On the workflow Summary page it only shows statistics with no links (nothing there is clickable):

https://github.com/jaegertracing/jaeger/actions/runs/7552405639?pr=5101

image

If instead I find a comment in the PR, it does have a link:

> For more details on these failures, see [this check](https://github.com/jaegertracing/jaeger/runs/20561351443).

However, on that page no error messages are displayed, just the name of the test

image

Overall, I am having much harder time analyzing the failures than before, since the logs of the run are now unreadable JSON, and these summaries are not helpful (and why so hard to find...).

Suggestion:
  * either find another test results uploader that has better integration with GitHub, or
  * go back to plain text logs

BTW, with plain logs my go-to trick was to search for `--- FAIL` string.

albertteoh:
- Resolves #5114

## Description of the changes
- We had a go at using a test summary reporter github action, but it appeared to lack sufficient support for "unexpected" test failures such as panics, while also making it harder to manually inspect test log output.
- The original log output still satisfies our requirements, as it is still easy to search through to identify the failing test.

## Checklist


Makefile
```diff
@@ -455,10 +454,6 @@ install-ci: install-test-tools install-build-tools
 test-ci: GOTEST := $(GOTEST_QUIET) -json
```
yurishkuro, line 454: do we still need this?

albertteoh: Nope, good catch; thanks! Removed `-json` flag in: [5675e49](https://github.com/jaegertracing/jaeger/pull/5126/commits/5675e49cf86e78f9b9ce3d5b652f4fa1f4835d02)

albertteoh:
Note the test is failing on "Download and extract artifacts" but this should no longer exist once this PR is merged into main.

yurishkuro (approved):

yurishkuro:
Thanks!

PR #5127, opened 2024-01-21
Update ids-to-base64.py

RupeshHacker:
Refactored traceId and spanId processing for improved performance with compiled regex and bulk string replacement.Optimized base64 encoding of IDs using direct byte conversion and efficient string handling.

### DIFF
--- a/scripts/utils/ids-to-base64.py
+++ b/scripts/utils/ids-to-base64.py
@@ -5,30 +5,39 @@ import os
 import re
 import sys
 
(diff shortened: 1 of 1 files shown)


RupeshHacker:
@yurishkuro sir can I get any feedback on this PR

yurishkuro:
this script does not require efficiency improvements

RupeshHacker:
> this script does not require efficiency improvements

okay sir thanks for reply

Issue #5083, opened 2024-01-06
[Bug]: Goroutine leak in plugin/storage/es/


yurishkuro:
### What happened?

Unrelated PR failed with goleak:
```
Runtime error
goleak: Errors on successful test run: found unexpected goroutines:
[Goroutine 106 in state select, with github.com/olivere/elastic.(*Client).healthchecker on top of the stack:
github.com/olivere/elastic.(*Client).healthchecker(0xc0002561a0)
	/home/runner/go/pkg/mod/github.com/olivere/elastic@v6.2.37+incompatible/client.go:1060 +0x229
created by github.com/olivere/elastic.DialContext in goroutine 102
	/home/runner/go/pkg/mod/github.com/olivere/elastic@v6.2.37+incompatible/client.go:404 +0xcfa
 Goroutine 107 in state select, with github.com/olivere/elastic.(*bulkWorker).work on top of the stack:
github.com/olivere/elastic.(*bulkWorker).work(0xc000476180, {0x1de9fb8, 0x2906f40})
	/home/runner/go/pkg/mod/github.com/olivere/elastic@v6.2.37+incompatible/bulk_processor.go:480 +0x1c5
created by github.com/olivere/elastic.(*BulkProcessor).Start in goroutine 102
	/home/runner/go/pkg/mod/github.com/olivere/elastic@v6.2.37+incompatible/bulk_processor.go:336 +0x5dc
]
```

### Steps to reproduce

Unknown, this appears to be a spurious condition when ES clients are not being stopped properly (they kick off a bg healthcheck go routine). It's possible that there is a race condition between starting and stopping the client, since we already spent time fixing these go leaks.

### Expected behavior

No go leaks in the package.

akagami-harsh:
I'll look into it

akagami-harsh:
i think, this error could be because of Elasticsearch client is not properly closed at some parts of code in [plugin/storage/es/factory_test.go](https://github.com/jaegertracing/jaeger/blob/main/plugin/storage/es/factory_test.go) 

The `Close()` method of the Factory struct wasn't invoked at the end of some tests. While calling `f.Close()` might resolve the issue, I'm unable to confirm this without a reliable method of reproducing the error.

yurishkuro:
@akagami-harsh do you see specific places where Close is not called?

yurishkuro:
One possible approach is to switch the tests to use this constructor from `olivere` lib:
```
[...]

PR #5128, opened 2024-01-21
[test] Add per-test goleak check for storage/es

yurishkuro:
- Part of #5083

## Description of the changes
- Add goleak checks to individual tests suspected of causing the leaks, this might help confirming which ones are at fault, since the main VerifyGoLeak check cannot distinguish between tests in the package.

```
$ go test -count 100 ./plugin/storage/es
ok  	github.com/jaegertracing/jaeger/plugin/storage/es	8.246s
```

### DIFF
--- a/pkg/testutils/leakcheck_test.go
+++ b/pkg/testutils/leakcheck_test.go
@@ -18,6 +18,10 @@ import (
 	"testing"
 )
 
+func TestVerifyGoLeaksOnes(t *testing.T) {
+	defer VerifyGoLeaksOnce(t)
+}
+
 func TestMain(m *testing.M) {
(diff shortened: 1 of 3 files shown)


pkg/testutils/leakcheck_test.go
```diff
@@ -18,6 +18,10 @@ import (
 	"testing"
 )
 
+func TestVerifyGoLeaksOnes(t *testing.T) {
```
albertteoh, line 21:
```suggestion
func TestVerifyGoLeaksOnce(t *testing.T) {
```

albertteoh (approved):

PR #5129, opened 2024-01-21
Added unit test in internal/metricstest/keys_test.go

### LINKED ISSUE #5068: Implement missing unit tests

sivasathyaseeelan:
- Partially fixes #5068 

## Description of the changes
- This PR adds test for keys.go in internal/metricstest

- run `make test`

## Checklist


yurishkuro:
I don't see much value in this test, we already have 100% coverage in the package.

PR #5134, opened 2024-01-22
testifylint: enable compares

mmorel-35:
- Apply compares rule of testifylint

## Description of the changes
-  More appropriate testify API with clearer failure message.

- CI 

## Checklist

### DIFF
--- a/plugin/sampling/strategystore/static/strategy_store_test.go
+++ b/plugin/sampling/strategystore/static/strategy_store_test.go
@@ -333,7 +333,6 @@ func TestDeepCopy(t *testing.T) {
 		},
 	}
 	cp := deepCopy(s)
-	assert.False(t, cp == s)
 	assert.EqualValues(t, cp, s)
 }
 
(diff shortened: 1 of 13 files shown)


plugin/sampling/strategystore/static/strategy_store_test.go
```diff
@@ -333,7 +333,6 @@ func TestDeepCopy(t *testing.T) {
 		},
 	}
 	cp := deepCopy(s)
-	assert.False(t, cp == s)
```
yurishkuro, line 336: this seems important - it was ensuring the two structs are not the same, even if equal in content

yurishkuro (commented):
+1

plugin/sampling/strategystore/static/strategy_store_test.go
yurishkuro: can we fix this?

mmorel-35: If I try to comply with testifylint it will be 
assert.NotEqual(t, cp, s) but this fails .
Maybe I am applying the wrong correction .
Any better idea?

yurishkuro: Is there something like Same() check?

mmorel-35: Yes Same and NotSame

yurishkuro (approved):

PR #5135, opened 2024-01-22
Define GOBIN targets for tools

mmorel-35:
-  avoid tools reinstall 

## Description of the changes
-  define targets for tools so they are not reinstalled once they are installed in GOBIN dir

-  CI

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,6 +53,7 @@ SED=sed
 GO=go
 GOOS ?= $(shell $(GO) env GOOS)
 GOARCH ?= $(shell $(GO) env GOARCH)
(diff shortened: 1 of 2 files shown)


yurishkuro (commented):
problem with that is there is no way to enforce the versions we have pinned, other than by manually deleting the binaries

mmorel-35:
Would having a clean target that would rm those binaries help ?
Or is this just a false good idea?

yurishkuro:
I don't think we need it. You don't install tools all the time to warrant a speed-up, and we don't call the target from local testing, only in CI.

PR #5138, opened 2024-01-23
Add endpoints to allow uploading OTLP traces

### LINKED ISSUE #4949: Ability to upload OTLP data into UI
### What happened?

I'm running OTel Collector with one of the exporters exporting traces to a file via https://opentelemetry.io/docs/specs/otel/protocol/file-exporter/

Jaeger is running with `COLLECTOR_OTLP_ENABLED:  true`

When trying to import this data to a Jaeger UI via 
Screenshot 2023-11-14 at 12 27 43

I'm getting

```
There was an error querying for traces:
Error parsing JSON: Unexpected non-whitespace character after JSON at position 9465 (line 2 column 1)
```

### Steps to reproduce

1. Export traces via file exporter(otel collector) to a file
2. Try to import them manually via Jaeger UI

### Expected behavior

To have my parses interpreted correctly as it happens with any other OTLP ingestion

### Relevant log output

```shell
There was an error querying for traces:
Error parsing JSON: Unexpected non-whitespace character after JSON at position 9465 (line 2 column 1)
```

### Screenshot

_No response_

### Additional context

_No response_

### Jaeger backend version

1.50

### SDK

- jaegertracing/all-in-one:1.50
- otel col 0.86

### Pipeline

OTel SDK -> OTel Collector -> file -> Jaeger manual import via UI

### Stogage backend

a file

### Operating system

Linux

### Deployment model

k8s

### Deployment configs

```
COLLECTOR_OTLP_ENABLED:  true
```

```shell
receivers:
  otlp:
    protocols:
      grpc:

exporters:
  file:
    path: "/tmp/otel-output.json"

processors:

service:
  pipelines:
    traces:
      receivers: [otlp]
      processors: []
      exporters: [file]

```
```

### PR DESCRIPTION
**Pull request by NavinShrinivas**
Signed-off-by: Navin Shrinivas <karupal2002@gmail.com>

partial work for issue #4949

-  Resolved #4949 

## Description of the changes
- Adds an endpoint that converts OTLP traces to jaeger traces

- Tested using example traces placed in `idl/otel-proto/examples/traces.json`

The commit in the PR is rather rough, looking for feedback as it's only a draft PR for now. I haven't formatted and there are some extra prints here and there that I used.

## Checklist

### DIFF
--- a/cmd/query/app/apiv3/otlp_translator.go
+++ b/cmd/query/app/apiv3/otlp_translator.go
@@ -17,8 +17,10 @@ package apiv3
 import (
 	"fmt"
 
+	"github.com/gogo/protobuf/jsonpb"
 	"github.com/gogo/protobuf/proto"
 	model2otel "github.com/open-telemetry/opentelemetry-collector-contrib/pkg/translator/jaeger"
+	"go.opentelemetry.io/collector/pdata/ptrace"
 	"go.opentelemetry.io/collector/pdata/ptrace/ptraceotlp"
 
 	"github.com/jaegertracing/jaeger/model"
@@ -44,4 +46,27 @@ func modelToOTLP(spans []*model.Span) ([]*tracev1.ResourceSpans, error) {
 		return nil, fmt.Errorf("cannot marshal OTLP: %w", err)
 	}
 	return chunk.ResourceSpans, nil
+}
+
+func OTLP2model(OTLPSpans []byte) ([]model.Trace, error) {
+	ptrace_unmarshaler := ptrace.JSONUnmarshaler{}
+	otlp_traces, err := ptrace_unmarshaler.UnmarshalTraces(OTLPSpans)
+	if err != nil {
+		fmt.Println(err)
+		return nil, fmt.Errorf("cannot marshal OTLP : %w", err)
+	}
+	batches, err := model2otel.ProtoFromTraces(otlp_traces)
+   fmt.Println(otlp_traces.ResourceSpans())
+	if err != nil {
+		fmt.Println(err)
+		return nil, fmt.Errorf("cannot marshal OTLP : %w", err)
+	}
+   jaeger_traces := make([]model.Trace, len(batches) )
+	for _, v := range batches {
+		mar := jsonpb.Marshaler{}
+		fmt.Println(mar.MarshalToString(v))
+      jaeger_trace := v.ConvertToTraces()
+      jaeger_traces = append(jaeger_traces, *jaeger_trace)
+	}
+	return jaeger_traces, nil
 }
--- a/cmd/query/app/http_handler.go
+++ b/cmd/query/app/http_handler.go
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	"net/url"
 	"strconv"
@@ -31,6 +32,7 @@ import (
 	"go.opentelemetry.io/otel/propagation"
 	"go.uber.org/zap"
 
+	"github.com/jaegertracing/jaeger/cmd/query/app/apiv3"
 	"github.com/jaegertracing/jaeger/cmd/query/app/querysvc"
 	"github.com/jaegertracing/jaeger/model"
 	uiconv "github.com/jaegertracing/jaeger/model/converter/json"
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: cmd/query/app/apiv3/otlp_translator.go [hunk 1]
```diff
@@ -45,3 +47,26 @@ func modelToOTLP(spans []*model.Span) ([]*tracev1.ResourceSpans, error) {
 	}
 	return chunk.ResourceSpans, nil
 }
+
+func OTLP2model(OTLPSpans []byte) ([]model.Trace, error) {
+	ptrace_unmarshaler := ptrace.JSONUnmarshaler{}
+	otlp_traces, err := ptrace_unmarshaler.UnmarshalTraces(OTLPSpans)
+	if err != nil {
+		fmt.Println(err)
+		return nil, fmt.Errorf("cannot marshal OTLP : %w", err)
```
yurishkuro, line 56:
```suggestion
		return nil, fmt.Errorf("cannot unmarshal OTLP: %w", err)
```

model/model.pb.go
```diff
@@ -597,6 +597,22 @@ func (*Batch) ProtoMessage()    {}
 func (*Batch) Descriptor() ([]byte, []int) {
 	return fileDescriptor_4c16552f9fdb66d8, []int{6}
 }
+
+func (m *Batch)ConvertToTraces()(*Trace){
```
yurishkuro, line 601: this is auto-generated file, cannot change it manually.  Also, I am not sure why you need it - batch is just a collection of spans with a shared Process, and the HTTP handler already has code to translate a list of spans into  UI-compatible output.

cmd/query/app/http_handler.go
```diff
@@ -158,6 +161,39 @@ func (aH *APIHandler) formatRoute(route string, args ...interface{}) string {
 	return fmt.Sprintf("/%s"+route, args...)
 }
 
+func (aH *APIHandler) transformOTLP(w http.ResponseWriter, r *http.Request) {
```
yurishkuro, line 164: please move the function to a place in the file so that it reflects its relative position in RegisterRoutes

yurishkuro (commented):
approach looks good

model/model.pb.go
NavinShrinivas: the `fromDomain` function that is used by `convertModelToUI` tries to deref Span.Process (I get a nil pointer derefencing error) : https://github.com/jaegertracing/jaeger/blob/d18b5f91e288d8af37a517ca274c7651f47c83cc/model/converter/json/from_domain.go#L60 and doesn't consider a shared common Process, would it be better to modify the `fromDomain` function instead?

yurishkuro: you need to prepare data for that function by denormalizing Batch.Process -> Span.Process

NavinShrinivas: That is what im trying to do in this function by copying m.Process into v.process

yurishkuro: I know, but (a) you cannot do this in this file, and (b) that's not all you're doing, L607-610 don't make sense, it's already being done on the GET /v1/trace code path.

NavinShrinivas: Ahhh, that makes sense, thanks! I'll revert back with the changes :)

NavinShrinivas: Yikes! I forgot models.pb is a file generated by protoc from grpc definitions...

cmd/query/app/apiv3/otlp_translator.go
```diff
@@ -45,3 +47,26 @@ func modelToOTLP(spans []*model.Span) ([]*tracev1.ResourceSpans, error) {
 	}
 	return chunk.ResourceSpans, nil
 }
+
+func OTLP2model(OTLPSpans []byte) ([]model.Trace, error) {
+	ptrace_unmarshaler := ptrace.JSONUnmarshaler{}
+	otlp_traces, err := ptrace_unmarshaler.UnmarshalTraces(OTLPSpans)
```
charconstpointer, line 53:
```suggestion
	ptraceUnmarshaler := ptrace.JSONUnmarshaler{}
	otlpTraces, err := ptrace_unmarshaler.UnmarshalTraces(OTLPSpans)
```
same applies to `jaeger_traces` and `jaeger_traces`

context: https://go.dev/doc/effective_go#mixed-caps

PR #5144, opened 2024-01-25
Separate /model into its own module

### LINKED ISSUE #4899: feat: Separate /model into its own module
Jaeger repo has always been a monorepo with a single module / go.mod, because it was not meant to be used as a library. But it is often used as a library, in particular in the OpenTelemetry Collector Contrib, where many modules depend on our data model. This creates an undesired coupling, e.g. all our dependencies get pull as OTEL dependencies, sometimes causing conflicts.

In contrast, OTEL's repos are organized as multiple modules, which allows more selective dependencies.

The ask from OTEL project is to make `/model` as separate module (https://github.com/open-telemetry/opentelemetry-collector/issues/8743#issuecomment-1781923246).

I don't know all that it takes to achieve that. Some trial and error will be needed. OTEL repos may serve as a blueprint.

* [ ] Do we need special handling of the `model` module in the main  `go.mod` for local development?
* [ ] How does this change the build?
* [ ] How does this affect code coverage computations?
* [ ] How does this change the release / versioning process?
* [ ] How will dependency updates be handled? OTEL apparently switched from dependabot to renovatebot because of multi-module setup.
* [ ] What additional maintenance changes are needed? E.g. we have `go 1.20` in `go.mod`, which is checked by `scripts/check-go-version.sh` - are there more places like that?

chirag-ghosh:
- Resolves #4899 

## Description of the changes
- Created go module at `/model`. Used replace in go.mod at `/` and `/model`
- Updated `/scripts/check-go-version.sh` to check `/model/go.mod` as well.
- Added support in dependabot (opentelemetry moved from dependabot to renovate-bot [here](https://github.com/open-telemetry/opentelemetry-collector/issues/4885). We can use dependabot for multiple modules though.)
- For releases, we can use [multimod](https://pkg.go.dev/go.opentelemetry.io/build-tools/multimod#section-readme) by opentelemetry. A [versions.yaml](https://github.com/open-telemetry/opentelemetry-collector/blob/main/versions.yaml) will define the different versions of different modules. (I have not implemented this yet.)
- Code coverage should not be affected I feel. Although let's see the codecov report.

- make lint test

## Checklist


model/go.mod
```diff
@@ -0,0 +1,34 @@
+module github.com/jaegertracing/jaeger/model
+
+go 1.20
+
+replace github.com/jaegertracing/jaeger => ../
+
+require (
+	github.com/apache/thrift v0.19.0
+	github.com/gogo/protobuf v1.3.2
+	github.com/jaegertracing/jaeger v0.0.0-00010101000000-000000000000
```
yurishkuro, line 10: Does this mean we have circular dependency? Because in that case the whole exercise is pointless, the whole reason for separating model was so that OTEL contrib could import it with the rest of jaeger deps. I think model should have very minimal deps, we can probably separate them.

go.mod
```diff
@@ -22,6 +22,7 @@ require (
 	github.com/grpc-ecosystem/go-grpc-middleware v1.4.0
 	github.com/hashicorp/go-hclog v1.6.2
 	github.com/hashicorp/go-plugin v1.6.0
+	github.com/jaegertracing/jaeger/model v0.0.0-00010101000000-000000000000
```
yurishkuro, line 25: What is the significance of this weird version? Is dependabot going to try to change it? Did you try on a separate repo?

model/go.mod
chirag-ghosh: i think we can remove this. this was automatically added when i ran `go mod tidy`.

go.mod
chirag-ghosh: I will try in my repository and see what happens

model/go.mod
chirag-ghosh: The only dependency of jaeger in `/model` is in tests like [here](https://github.com/jaegertracing/jaeger/blob/main/model/adjuster/package_test.go#L20). I think we need to have the testutils inside /model to remove the dependency.

chirag-ghosh:
@yurishkuro Can you please help me figure out why the code coverage went down?

yurishkuro:
> Can you please help me figure out why the code coverage went down?

In the report I see that all /model/ files are down to 0% coverage. I suspect `go test ./...` does not include sub-modules - this is one of the things that needs to be figured out.

yurishkuro:
another issue - the vulnerability checker flagged /model/ as vulnerable dependency, I suspect because of 0.x version number. What is the story with this version number? In OTEL collector I see that their versions are in sync across modules, so they must have some mechanism for managing/updating those after release.
```
Vulnerabilities
  model/go.mod » github.com/jaegertracing/jaeger@0.0.0-00010101000000-000000000000 – A stored XSS in jaeger UI might allow an attacker who controls a trace to perform arbitrary jaeger queries (moderate severity)
    ↪ https://github.com/advisories/GHSA-2w[8](https://github.com/jaegertracing/jaeger/actions/runs/7656173413/job/20863882760?pr=5144#step:5:9)w-qhg4-f78j
  model/go.mod » github.com/jaegertracing/jaeger@0.0.0-00010101000000-000000000000 – Information Exposure in jaeger (moderate severity)
    ↪ https://github.com/advisories/GHSA-gh32-pc56-4c[9](https://github.com/jaegertracing/jaeger/actions/runs/7656173413/job/20863882760?pr=5144#step:5:10)6
  Error: Dependency review detected vulnerable packages.
```

model/go.mod
mmorel-35: I see model/adjuster/package_test.go
Pointing to pkg/utils, which can explain the circular dependency

chirag-ghosh: to remove the circular dependency we will need to either copy the same things inside /model or make them separate packages as well

chirag-ghosh:
The wierd versioning happended after I aded the replace lines and ran `go mod tidy`. [Here](https://stackoverflow.com/questions/58012771/go-mod-fails-to-find-version-v0-0-0-00010101000000-000000000000-of-a-dependency) is a StackOverflow question regarding it.

yurishkuro:
closing https://github.com/jaegertracing/jaeger/issues/4899#issuecomment-2134139934

PR #5152, opened 2024-01-29
[jaeger-v2] Add support for Elasticsearch

akagami-harsh:
- part of #4843

## Description of the changes
- add support for elasticsearch in jaeger-v2

- tested locally

## Checklist


cmd/jaeger/es_config.yaml
```diff
@@ -0,0 +1,34 @@
...
+    trace_storage: es_main
+    trace_storage_archive: es_archive
+    ui_config: ./cmd/jaeger/config-ui.json
+
+  jaeger_storage:
+    elasticsearch:
+      es_main:
+        server_urls: http://127.0.0.1:9200
+        log_level: "error"
+      es_archive:
```
yurishkuro, line 20: is this a good example? archive storage implementation is identical to primary, so if you given them identical parameters they will be trying to write to the same data

akagami-harsh: added `index_prefix` field in archive storage

cmd/jaeger/internal/extension/jaegerstorage/extension.go
```diff
@@ -88,6 +89,22 @@ func (s *storageExt) Start(ctx context.Context, host component.Host) error {
 		s.factories[name] = factory
 	}
 
+	for name, e := range s.config.Elasticsearch {
+		if _, ok := s.factories[name]; ok {
+			return fmt.Errorf("duplicate elasticsearch storage name %s", name)
+		}
+		var err error
```
yurishkuro, line 96:
```suggestion
```

plugin/storage/es/factory.go
```diff
@@ -83,6 +83,23 @@ func NewFactory() *Factory {
...
 
+func NewFactoryWithConfig(
+	cfg config.Configuration,
+	metricsFactory metrics.Factory,
+	logger *zap.Logger,
+) (*Factory, error) {
+	f := NewFactory()
+	f.InitFromOptions(Options{
+		Primary: namespaceConfig{Configuration: cfg},
+		others:  make(map[string]*namespaceConfig),
```
yurishkuro, line 94: I wonder, if only Primary and not Archive is ever initialized, would it actually provide an archive storage to the UI? Did you try it out?

cmd/jaeger/internal/extension/jaegerstorage/extension.go
```diff
@@ -88,6 +89,22 @@ func (s *storageExt) Start(ctx context.Context, host component.Host) error {
 		s.factories[name] = factory
 	}
 
+	for name, e := range s.config.Elasticsearch {
```
yurishkuro, line 92:
```suggestion
	for name, cfg := range s.config.Elasticsearch {
```

plugin/storage/es/factory.go
akagami-harsh: yes, you are right it dose not work with ui. it throw's this error 
this docker container used to test it  
```
docker run --name es01 -p 9200:9200 -e "discovery.type=single-node" -e "xpack.security.enabled=false" elasticsearch:8.12.0
```
image
do you know what causes it?🤔 , i tried fixing it with different configuration of `Options` but couldn't fix.

yurishkuro:
please rebase on top of #5171

cmd/jaeger/internal/extension/jaegerstorage/extension_test.go
```diff
@@ -85,6 +88,9 @@ func TestStorageExtensionNameConflict(t *testing.T) {
 		Badger: map[string]badgerCfg.NamespaceConfig{
 			"foo": {},
 		},
+		Elasticsearch: map[string]esCfg.Configuration{
```
yurishkuro, line 91: not necessary, does not add anything to the test

cmd/jaeger/internal/extension/jaegerstorage/extension_test.go
```diff
@@ -151,6 +157,43 @@ func TestBadgerStorageExtensionError(t *testing.T) {
...
+
+func TestESStorageExtensionError(t *testing.T) {
+	ext := makeStorageExtenion(t, &Config{
+		Elasticsearch: map[string]esCfg.Configuration{
+			"foo": {},
+		},
+	})
+	err := ext.Start(context.Background(), componenttest.NewNopHost())
+	require.ErrorContains(t, err, "failed to initialize elasticsearch storage")
+	require.ErrorContains(t, err, "no servers specified")
```
yurishkuro, line 194: I think this should be caught during config validation. We need to add annotations to esCfg.Config struct.

plugin/storage/es/factory.go
yurishkuro: Better to look in the server logs.

akagami-harsh:
I am currently working on my LFX proposals and will address the reviewed comment soon.

cmd/jaeger/internal/extension/jaegerstorage/extension_test.go
akagami-harsh: removed it!

cmd/jaeger/internal/extension/jaegerstorage/extension_test.go
akagami-harsh: added annotations

akagami-harsh:
>Better to look in the server logs.

looked at the server logs. getting this err when trying to access ui, not sure why.
```
2024-02-17T01:40:04.265+0530    info    service@v0.93.0/service.go:165  Everything is ready. Begin running and processing data.
2024-02-17T01:40:08.406+0530    error   app/http_handler.go:492 HTTP handler, Internal Server Error     {"kind": "extension", "name": "jaeger_query", "error": "search services failed: elastic: Error 400 (Bad Request): all shards failed [type=search_phase_execution_exception]"}
github.com/jaegertracing/jaeger/cmd/query/app.(*APIHandler).handleError
        /home/harsh/OpenSource/JAEGER/es/jaeger/cmd/query/app/http_handler.go:492
github.com/jaegertracing/jaeger/cmd/query/app.(*APIHandler).getServices
        /home/harsh/OpenSource/JAEGER/es/jaeger/cmd/query/app/http_handler.go:163
net/http.HandlerFunc.ServeHTTP
        /usr/local/go/src/net/http/server.go:2136
```

plugin/storage/es/factory.go
yurishkuro: Can you put together a docker-compose that illustrates this issue?

This "all shards failed" error has always annoyed me, I wonder if there are more details in the error returned from ES that we're not logging that could be more descriptive. Like, it's ridiculous for a database to respond with "your query failed" instead of telling why (the fact that it failed on "all shards" is not an explanation of why).

pkg/es/config/config.go
```diff
@@ -45,7 +45,7 @@ import (
 
 // Configuration describes the configuration properties needed to connect to an ElasticSearch cluster
 type Configuration struct {
-	Servers                        []string       `mapstructure:"server_urls"`
+	Servers                        []string       `mapstructure:"server_urls" validate:"required,min=1,dive,url"`
```
yurishkuro, line 48: I wonder if there's a test we could write for ensuring that these annotations are correct (btw, what is "dive"?)

We could do it brute-force and just write individual tests for individual fields, but it seems redundant as that would be already done in the validator package (hopefully). But at the same time, having no tests to even ensure that the syntax of the annotations is correct is troublesome.

yurishkuro: Comment for L58 and below - can you try removing `yaml:...` annotations? I think they are left-overs from earlier attempts to build on OTEL Coll, I don't think we're using them anywhere.

plugin/storage/es/factory.go
akagami-harsh: >Can you put together a docker-compose that illustrates this issue?

sure, i'll make one.

pkg/es/config/config.go
akagami-harsh: >Comment for L58 and below - can you try removing yaml:... annotations? I think they are left-overs from earlier attempts to build on OTEL Coll, I don't think we're using them anywhere.

removed it

akagami-harsh: > (btw, what is "dive"?)

dive instructs the validation library to examine each element within the slice individually

akagami-harsh: > I wonder if there's a test we could write for ensuring that these annotations are correct 
> 
> We could do it brute-force and just write individual tests for individual fields, but it seems redundant as that would be already done in the validator package (hopefully). But at the same time, having no tests to even ensure that the syntax of the annotations is correct is troublesome.

we can write unit tests that check whether the validation behaves as expected. Instead of testing individual fields, we can create a set of test cases that cover various scenarios, including both valid and invalid configurations.
should i create one?

yurishkuro: but that wouldn't be what I am after, such test would simply verify that the Validator is working as expected, on select fields. But if I declare a field with annotations that don't make sense, I want to catch that in unit tests. E.g. misspell `dive` as `diev` (btw weird option, "recursive" would've been much more self explanatory).

cmd/jaeger/es_config.yaml
```diff
@@ -0,0 +1,37 @@
+service:
```
yurishkuro, line 1: rename this file to config-elasticsearch.yaml

pkg/es/config/config.go
```diff
@@ -54,13 +54,13 @@ type Configuration struct {
 	AllowTokenFromContext          bool           `mapstructure:"-"`
 	Sniffer                        bool           `mapstructure:"sniffer"` // https://github.com/olivere/elastic/wiki/Sniffing
 	SnifferTLSEnabled              bool           `mapstructure:"sniffer_tls_enabled"`
-	MaxDocCount                    int            `mapstructure:"-"`                     // Defines maximum number of results to fetch from storage per query
-	MaxSpanAge                     time.Duration  `yaml:"max_span_age" mapstructure:"-"` // configures the maximum lookback on span reads
```
yurishkuro, line 58: please do not remove comments

plugin/storage/es/factory.go
yurishkuro: so when I run v2 with your new config file (and without `-tags ui`), then placeholder UI shows this:
```
JAEGER_CONFIG = {"archiveEnabled":true};
JAEGER_STORAGE_CAPABILITIES = {"archiveStorage":false};
JAEGER_VERSION = {"gitCommit":"","gitVersion":"","buildDate":""};
```
We expect `{"archiveStorage":true}` in order for archiving to work

yurishkuro:
I ran with your docker-compose and config and (with the help of #5209) it returns this error:

`RootCause[Fielddata is disabled on [serviceName] in [jaeger-service-]. Text fields are not optimised for operations that require per-document field data like aggregations and sorting, so these operations are disabled by default. Please use a keyword field instead. Alternatively, set fielddata=true on [serviceName] in order to load field data by uninverting the inverted index. Note that this can use significant memory. [type=illegal_argument_exception]]`

pkg/es/config/config.go
akagami-harsh: sorry!, added the commnets back

akagami-harsh:
> I ran with your docker-compose and config and (with the help of #5209) it returns this error:
> 
> `RootCause[Fielddata is disabled on [serviceName] in [jaeger-service-]. Text fields are not optimised for operations that require per-document field data like aggregations and sorting, so these operations are disabled by default. Please use a keyword field instead. Alternatively, set fielddata=true on [serviceName] in order to load field data by uninverting the inverted index. Note that this can use significant memory. [type=illegal_argument_exception]]`

thanks a lot for the help. now i think it is working fine, setting `UseReadWriteAliases` to `true` fixes it.

cmd/jaeger/Dockerfile.debug
```diff
@@ -0,0 +1,11 @@
+FROM alpine:latest
```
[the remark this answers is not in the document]
akagami-harsh: I was having issues with the main Dockerfile and created a simpler one as a solution. We don't need the Dockerfile and Docker-compose file anymore, so I'm removing them.

pkg/es/config/config.go
akagami-harsh: We can use the "[github.com/go-playground/validator](https://github.com/go-playground/validator)" library in our unit tests to detect misspelled annotations and fields with invalid annotation combinations.
```go 
package main

import (
	"github.com/go-playground/validator"
)

type Webhook struct {
	CallbackURL string `validate:"url,badAnnotation"` // badAnnotation is not a valid tag
}

func main() {
	webhook := Webhook{
		CallbackURL: "https://www.example.com/",
	}
	validate := validator.New()
	validate.Struct(webhook)

}
```
```
>output 
panic: Undefined validation function 'badAnnotation' on field 'CallbackURL'

goroutine 1 [running]:
github.com/go-playground/validator.(*Validate).parseFieldTagsRecursive(0xc00002e240, {0x526987?, 0x55812d?}, {0x526971, 0xb}, {0x0, 0x0}, 0x0)
        /home/harsh/go/pkg/mod/github.com/go-playground/validator@v9.31.0+incompatible/cache.go:289 +0x9d9
github.com/go-playground/validator.(*Validate).extractStructCache(0xc00002e240, {0x538520?, 0xc00020a950?, 0x7fda1aac3198?}, {0x52160c, 0x7})
        /home/harsh/go/pkg/mod/github.com/go-playground/validator@v9.31.0+incompatible/cache.go:150 +0x4ec
github.com/go-playground/validator.(*validate).validateStruct(0xc000118480, {0x58b0c0, 0x68faa0}, {0x538520?, 0xc00020a950?, 0xc000242000?}, {0x538520?, 0xc00020a950?, 0x40e3a5?}, {0x58cd60, ...}, ...)
        /home/harsh/go/pkg/mod/github.com/go-playground/validator@v9.31.0+incompatible/validator.go:37 +0x196
github.com/go-playground/validator.(*Validate).StructCtx(0xc00002e240, {0x58b0c0, 0x68faa0}, {0x538520?, 0xc00020a950?})
        /home/harsh/go/pkg/mod/github.com/go-playground/validator@v9.31.0+incompatible/validator_instance.go:304 +0x491
github.com/go-playground/validator.(*Validate).Struct(...)
        /home/harsh/go/pkg/mod/github.com/go-playground/validator@v9.31.0+incompatible/validator_instance.go:277
main.main()
        /home/harsh/Testing/main.go:16 +0x65
exit status 2

Comment (yurishkuro) [reply pkg/es/config/config.go]: what about "github.com/asaskevich/govalidator", will it do the same? We do not use "github.com/go-playground/validator"

Comment (akagami-harsh) [reply pkg/es/config/config.go]: >will it do the same?

Nope, it doesn't do the same. I actually tried using the existing validator at first.

File: plugin/storage/es/factory.go (hunk 3 above)
Comment (yurishkuro) [reply plugin/storage/es/factory.go:99]: where are we with this one?

Comment (akagami-harsh) [reply plugin/storage/es/factory.go:99]: I didn't check this one previously because there were no errors in the UI. However, I've just reviewed it and found an issue.
```
  function getJaegerStorageCapabilities() {
        const DEFAULT_STORAGE_CAPABILITIES = { "archiveStorage": false };
        const JAEGER_STORAGE_CAPABILITIES = {"archiveStorage":false};
        return JAEGER_STORAGE_CAPABILITIES;
      }
```

Comment (akagami-harsh) [reply plugin/storage/es/factory.go:99]: when i checked with `badger` and `memstore` strangely `{"archiveStorage":false}` in both of them

Comment (yurishkuro) [reply plugin/storage/es/factory.go:99]: At least for memstore you need to add archive to the config explicitly:
```diff
diff --git a/cmd/jaeger/internal/all-in-one.yaml b/cmd/jaeger/internal/all-in-one.yaml
index d17c52fc..a6189c8e 100644
--- a/cmd/jaeger/internal/all-in-one.yaml
+++ b/cmd/jaeger/internal/all-in-one.yaml
@@ -9,11 +9,14 @@ service:
 extensions:
   jaeger_query:
     trace_storage: memstore
+    trace_storage_archive: memstore2

   jaeger_storage:
     memory:
       memstore:
         max_traces: 100000
+      memstore2:
+        max_traces: 100000
```
But maybe this is ok - the idea in v2 was to decouple primary and archive storage altogether, so that one could use different types of backends for those.

yurishkuro: So in theory the config you have should've worked, because you did define archive storage, but something isn't working.

akagami-harsh: added a small fix for above problem, which sets `{"archiveEnabled":true}`

plugin/storage/es/factory.go
```diff
@@ -83,6 +83,39 @@ func NewFactory() *Factory {
...
+	}
+
+	cfg.MaxDocCount = defaultMaxDocCount
+	cfg.Enabled = true
+
+	archive := make(map[string]*namespaceConfig)
+	archive[archiveNamespace] = &namespaceConfig{
+		Configuration: cfg,
+		namespace:     archiveNamespace,
+	}
```
yurishkuro, line 102: ok, this is logically wrong, but will work for now. The underlying problem is the existence of `storage.ArchiveFactory` interface, we need to get rid of it long term, because as far as interface goes the archive storage is just like a regular storage, only with different configuration.

yurishkuro: #5230

plugin/storage/es/factory.go
```diff
@@ -83,6 +83,39 @@ func NewFactory() *Factory {
...
 }
 
+func NewFactoryWithConfig(
+	cfg config.Configuration,
+	metricsFactory metrics.Factory,
+	logger *zap.Logger,
+) (*Factory, error) {
+	if err := cfg.Validate(); err != nil {
+		return nil, err
+	}
```
yurishkuro, line 93: please add a test for this

yurishkuro (approved):
Thanks!

PR #5154, opened 2024-01-29
Add reliable check for Kafka broker being ready

### LINKED ISSUE #5148: [fix] Ensure Kafka broker is started before running integration tests
Kafka integration tests started failing frequently. I suspect our `scripts/kafka-integration-test.sh` script is not waiting until Kafka is properly started. We need to find a more reliable check.

There are a bunch of scripts available in the container:
```
docker run -it --rm bitnami/kafka:3.6.0 /bin/sh
$ ls -lF /opt/bitnami/kafka/bin/
```
We can probably use one of them to more reliably check for running server, e.g. `kafka-topics.sh` or `kafka-broker-api-versions.sh`, or even read & write a test message via `kafka-console-producer.sh` and `kafka-console-consumer.sh`.

Some documentation on how to run the container is available from bitnami: https://github.com/bitnami/containers/tree/main/bitnami/kafka#how-to-use-this-image

shashank-iitbhu:
Resolves #5148

## Description of the changes
- The previous approach used ```nc``` to check the availability of the ```Kafka port```, which was not reliable, especially when the Kafka container was started asynchronously. The updated script now ensures a more reliable check for Kafka readiness by attempting to list topics using ```kafka-topics.sh``` within a loop. This modification ensures that the script waits until Kafka is fully initialized before proceeding with integration tests.

- This modification ensures that the script waits until Kafka is fully initialized before proceeding with integration tests.
- Ran ```bash scripts/kafka-integration-test.sh -k``` from the project root.

## Checklist

### DIFF
--- a/scripts/kafka-integration-test.sh
+++ b/scripts/kafka-integration-test.sh
@@ -35,7 +35,8 @@ interval=5
 end_time=$((SECONDS + timeout))
 
 while [ $SECONDS -lt $end_time ]; do
-    if nc -z localhost 9092; then
+    # Check if Kafka is ready by attempting to describe a topic
+    if docker exec kafka /opt/bitnami/kafka/bin/kafka-topics.sh --list --bootstrap-server localhost:9092 >/dev/null 2>&1; then
         break
     fi
     echo "Kafka broker not ready, waiting ${interval} seconds"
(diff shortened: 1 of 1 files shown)


yurishkuro (approved):
Thanks!

PR #5155, opened 2024-01-30
Support uploading traces to UI in OpenTelemetry format (OTLP JSON)

NavinShrinivas:
* Solves #4949 
* In combination with https://github.com/jaegertracing/jaeger-ui/pull/2145

## Description of the changes
- Incorporates changes from previous draft PR review 
- Adds middleware in frontend as well

- Unit tests and supplying valid and invalid OTLP traces using insomnia to test the API

## Checklist


cmd/query/app/apiv3/otlp_translator.go
```diff
@@ -45,3 +46,26 @@ func modelToOTLP(spans []*model.Span) ([]*tracev1.ResourceSpans, error) {
...
+		fmt.Println(err)
+		return nil, fmt.Errorf("cannot unmarshal OTLP : %w", err)
+	}
+	jaegerBatches, err := model2otel.ProtoFromTraces(otlpTraces)
+	if err != nil {
+		fmt.Println(err)
+		return nil, fmt.Errorf("cannot transform OTLP to Jaeger: %w", err)
+	}
+	var jaegerTraces []model.Trace
+	for _, v := range jaegerBatches {
```
yurishkuro, line 63: This may be incorrect. OTEL uses the notion of "resource" that is equivalent to Jaeger's Process. `otlpTraces` is a collection of `resourceSpans` where each entry encapsulates a single resource and a bunch of spans - this is equivalent to Jaeger's Batch. We need to check how `model2otel.ProtoFromTraces` works, but it's quite possible that is just maps a single `resourceSpans` entry into a single Batch, since that's the equivalent granularity. Which means at this point even you have multiple batches you may still have one trace, and vice versa, a single batch may contain spans from different traces. This is all because these structures are optimized for ingestion, not retrieval. So the correct way would be to tease apart all batches into individual spans and group-by those spans by trace-id, only then will you be sure.

yurishkuro: btw, please make sure you have unit tests for all new functionality.

yurishkuro (commented):
there should be no UI submodule changes in this PR.

NavinShrinivas:
> there should be no UI submodule changes in this PR.

How do I make pr's for jaeger-ui changes?

yurishkuro:
in jaeger-ui repository

cmd/query/app/apiv3/otlp_translator.go
NavinShrinivas: Would it be fine to do this process manually? by maybe pushing these spans in a map of (trace_id, span) kv pair and later combining these spans into a single trace?

NavinShrinivas: > btw, please make sure you have unit tests for all new functionality.

Im not sure how to test the http endpoint thats supposed to have OTLP type traces

NavinShrinivas:
hey @yurishkuro, I have implemented the group by traceID and I have excluded jaeger-ui changes and added some tests.

cmd/query/app/apiv3/package_test.go
```diff
@@ -6,9 +6,67 @@ package apiv3
 import (
```
yurishkuro, line 6: tests are always named after the file they are testing: `otlp_translator_test.go`

model/trace.go
```diff
@@ -32,3 +32,21 @@ func (t *Trace) NormalizeTimestamps() {
...
 	}
 }
+
+func (m *Trace) DenormalizeProcess(p *Process) {
+	for _, v := range m.Spans {
+		v.Process = p
+	}
+}
+
+func (m *Trace) FlattenToSpansMaps(spanMap map[TraceID][]*Span) {
```
yurishkuro, line 42: this is used in a single place, just define the function there, not at the model level

yurishkuro: same for DenormalizeProcess - it's an absolutely trivial loop, no reason to have a model-level function for it

cmd/query/app/apiv3/otlp_translator.go
```diff
@@ -45,3 +46,38 @@ func modelToOTLP(spans []*model.Span) ([]*tracev1.ResourceSpans, error) {
 	}
 	return chunk.ResourceSpans, nil
 }
+
+func OTLP2model(OTLPSpans []byte) ([]*model.Batch, error) {
+	ptraceUnmarshaler := ptrace.JSONUnmarshaler{}
+	otlpTraces, err := ptraceUnmarshaler.UnmarshalTraces(OTLPSpans)
+	if err != nil {
+		fmt.Println(err)
```
yurishkuro, line 54: remove these

cmd/query/app/apiv3/otlp_translator.go
```diff
@@ -45,3 +46,38 @@ func modelToOTLP(spans []*model.Span) ([]*tracev1.ResourceSpans, error) {
...
+	return jaegerBatches, nil
+}
+
+func BatchesToTraces(jaegerBatches []*model.Batch) ([]model.Trace, error) {
+	var jaegerTraces []model.Trace
+	spanMap := make(map[model.TraceID][]*model.Span)
+	for _, v := range jaegerBatches {
+		jaegerTrace := model.Trace{
+			Spans: v.Spans,
+		}
```
yurishkuro, line 72: what's the point of creating a Trace object here?

cmd/query/app/http_handler.go
```diff
@@ -193,6 +196,43 @@ func (aH *APIHandler) getOperationsLegacy(w http.ResponseWriter, r *http.Request
...
 }
 
+func (aH *APIHandler) transformOTLP(w http.ResponseWriter, r *http.Request) {
+	body, err := ioutil.ReadAll(r.Body)
+	if aH.handleError(w, err, http.StatusBadRequest) {
+		return
+	}
+
+	var uiErrors []structuredError
+	batches, err := apiv3.OTLP2model(body)
```
yurishkuro, line 206: there is no need for those functions to be defined in apiv3 package, you can create a similar `transform_otlp.go` in this package.

yurishkuro (commented):
> and I have excluded jaeger-ui changes

can you submit another PR for UI changes? The functionality is not useful without both of them.

NavinShrinivas:
> > and I have excluded jaeger-ui changes
> 
> can you submit another PR for UI changes? The functionality is not useful without both of them.

I wont be able to link the issue in the jaeger-ui repo, is that alright? Or do you want me to push a seperate pr changing the commit version tracked in this repo for jaeger-ui?

cmd/query/app/http_handler.go
```diff
@@ -193,6 +195,42 @@ func (aH *APIHandler) getOperationsLegacy(w http.ResponseWriter, r *http.Request
 	aH.writeJSON(w, r, &structuredRes)
 }
 
+func (aH *APIHandler) transformOTLP(w http.ResponseWriter, r *http.Request) {
```
yurishkuro, line 198: * we have unit tests for all API endpoints, please add one for /transform
* similar to my comment in the UI PR (https://github.com/jaegertracing/jaeger-ui/pull/2145#discussion_r1472233464), I would suggest saving an OTLP test fixture file, and the corresponding UI JSON, and implementing submit/validate in the unit test

cmd/query/app/otlp_translator.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	"fmt"
+
+	model2otel "github.com/open-telemetry/opentelemetry-collector-contrib/pkg/translator/jaeger"
+	"go.opentelemetry.io/collector/pdata/ptrace"
+
+	"github.com/jaegertracing/jaeger/model"
+)
+
+func OTLP2model(OTLPSpans []byte) ([]*model.Batch, error) {
```
yurishkuro, line 26: none of these functions need to be public

cmd/query/app/http_handler.go
```diff
@@ -193,6 +195,42 @@ func (aH *APIHandler) getOperationsLegacy(w http.ResponseWriter, r *http.Request
...
+
+	if aH.handleError(w, err, http.StatusInternalServerError) {
+		return
+	}
+
+	traces, err := BatchesToTraces(batches)
+
+	if aH.handleError(w, err, http.StatusInternalServerError) {
+		return
+	}
```
yurishkuro, line 215: why does the handler need to call two functions? the whole conversion can be encapsulated in one helper, having fewer error branches here

NavinShrinivas: I wanted to make it easier to test, would you recommend me putting it together as a single function?

cmd/query/app/otlp_translator.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+	jaegerBatches, err := model2otel.ProtoFromTraces(otlpTraces)
+	if err != nil {
+		return nil, fmt.Errorf("cannot transform OTLP to Jaeger: %w", err)
+	}
+
+	return jaegerBatches, nil
+}
+
+func BatchesToTraces(jaegerBatches []*model.Batch) ([]model.Trace, error) {
```
yurishkuro, line 40: this seems over-complicated. How about this:
```
var traces []*model.Trace
traceMap := make(map[model.TraceID]*Trace)
for _, batch := range jaegerBatches {
  for _, span := range batch.Spans {
    if span.Process == nil {
      span.Process = batch.Process
    }
    trace, ok := traceMap[span.TraceID]
    if !ok {
      // create new trace, add to traces slice
    }
    trace.Spans = append(trace.Spans, span)
  }
}
```

yurishkuro:
> I wont be able to link the issue in the jaeger-ui repo, is that alright?

Yes you can, see how I just edited UI PR description - you just include URL to the ticket and GitHub links them

cmd/query/app/http_handler.go
yurishkuro: handler doesn't need to know about multiple phases of transformation. And handler is harder to test than the same two-step logic in a helper function

cmd/query/app/http_handler.go
NavinShrinivas: Is it ok to have a json file in cmd/query/app/fixture and read that file in the .._test.go file and pass it to the test http server? None of the API endpoints seem to be reading a file and hence this question

yurishkuro: Yes. Probably because other tests are using dynamically generated data. Here it's important to use the input exactly as produced by otel collector.

NavinShrinivas: I have written a basic unit test for the API in a new file handler_otlp_translator_test.go

cmd/query/app/otlp_translator.go
NavinShrinivas: Yes indeed, this is much better! I have implemented a version of this now.

cmd/query/app/fixture/OTLP-transform.json
```diff
@@ -0,0 +1,234 @@
+{
```
yurishkuro, line 1: let's reduce to a single span

cmd/query/app/handler_otlp_translator_test.go
```diff
@@ -0,0 +1,77 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/jaegertracing/jaeger/cmd/query/app/querysvc"
+)
+
+func readOTLPTraces() (interface{}, error) {
+	dat, err := os.ReadFile("./fixture/OTLP-transform.json")
```
yurishkuro, line 29: naming suggestion: otlp2jaeger-in.json, otlp2jaeger-out.json

cmd/query/app/handler_otlp_translator_test.go
```diff
@@ -0,0 +1,77 @@
...
+	if err != nil {
+		return nil, err
+	} else {
```
yurishkuro, line 32:
```suggestion
	require.NoError(t, err)
```
and pass `t` as arg

cmd/query/app/handler_otlp_translator_test.go
```diff
@@ -0,0 +1,77 @@
...
+		var out interface{}
+		json.Unmarshal(dat, out)
```
yurishkuro, line 34: check error via require.NoError

cmd/query/app/handler_otlp_translator_test.go
```diff
@@ -0,0 +1,77 @@
...
+		require.NoError(t, err)
+		corectResponse, err := readJaegerTraces()
+		require.NoError(t, err)
+		assert.Equal(t, response, corectResponse)
+	}, querysvc.QueryServiceOptions{})
+}
+
+
+func TestTransformOTLPEmptySucess(t *testing.T) {
+	// mockReader := &spanstoremocks.Reader{}
```
yurishkuro, line 68: weird whitespace

yurishkuro: run `make fmt`

cmd/query/app/handler_otlp_translator_test.go
```diff
@@ -0,0 +1,77 @@
...
+	withTestServer(func(ts *testServer) {
+		// make main reader return NotFound
+		var response interface{}
+      var request interface{} // Keeping request empty for checking behaviour
+      err := postJSON(ts.server.URL+"/api/transform", request, response)
```
yurishkuro, line 73: not sure what the objective here. If no OTLP is send in the request, shouldn't it be a BadRequest error?

cmd/query/app/handler_otlp_translator_test.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+func TestTransformOTLPSuccess(t *testing.T) {
+	// mockReader := &spanstoremocks.Reader{}
+	withTestServer(func(ts *testServer) {
+		// make main reader return NotFound
+		var response interface{}
+		request, err := readOTLPTraces()
+		require.NoError(t, err)
+		err = postJSON(ts.server.URL+"/api/transform", request, response)
```
yurishkuro, line 58: I think there is unnecessary JSON marshalling happening in this test. There's definitely no need in trying to parse the input file, just submit the bytes from the file as is. For the output file you may want to parse and re-serialize with  a predefined indentation, so that you can compare it with the fixture.

cmd/query/app/handler_otlp_translator_test.go
```diff
@@ -0,0 +1,77 @@
+// Copyright (c) 2024 The Jaeger Authors.
```
yurishkuro, line 1: why not add the tests to http_handler_test.go?

cmd/query/app/handler_otlp_translator_test.go
NavinShrinivas: Initially the test I wrote was doing this, but was failing. Later I found out that `postJSON` (that's being used in the test) converts the given "JSON" into a stream of bytes itself. If it converts a stream of bytes into a stream of bytes (thinking its s JSON type) the functions give an error 

https://github.com/jaegertracing/jaeger/blob/e08f576fd64a992ef0396112bc8401472cc9dd92/cmd/query/app/http_handler_test.go#L837-L842

yurishkuro: yes, but you don't need to use postJSON

NavinShrinivas: using `execJSON` and making the http.request manually with the buffer?

NavinShrinivas:
Im unable to add the required lavel for ci/cd
\

NavinShrinivas:
I just saw the missing tests, from what I can see, golang testify lets you mock functions that are methods of a struct, here the model2otlp functions are one-off separate functions. Would it be okay to include them as part of aH (APIHandler), allowing me to mock them like `tracesByID` function? Or is there some alternative method I'm missing out on?

cmd/query/app/http_handler.go
```diff
@@ -193,6 +195,42 @@ func (aH *APIHandler) getOperationsLegacy(w http.ResponseWriter, r *http.Request
...
+
+	var uiErrors []structuredError
+	batches, err := otlp2model(body)
+
+	if aH.handleError(w, err, http.StatusInternalServerError) {
+		return
+	}
+
+	traces, err := batchesToTraces(batches)
```
yurishkuro, line 211: you need to combine otlp2model and batchesToTraces into one function called by this handler

cmd/query/app/http_handler_test.go
```diff
@@ -623,6 +623,38 @@ func TestGetOperationsLegacyStorageFailure(t *testing.T) {
...
+		err := json.Unmarshal(requestBytes, request)
+		require.NoError(t, err)
+
+		err = postJSON(ts.server.URL+"/api/transform", request, response)
+		require.NoError(t, err)
+
+		corectResponseBytes := readJaegerTraces(t)
+		corectResponse := new(interface{})
+
+		err = json.Unmarshal(corectResponseBytes, corectResponse)
```
yurishkuro, line 642: none of this marshaling is required for the test. You need to read a fixture as bytes, submit them to the server, receive back bytes, and compare them with the expected fixture.

cmd/query/app/http_handler_test.go
```diff
@@ -623,6 +632,39 @@ func TestGetOperationsLegacyStorageFailure(t *testing.T) {
...
+		requestBytes := readOTLPTraces(t)
+
+		resp, err := ts.server.Client().Post(ts.server.URL+"/api/transform", "application/json", bytes.NewReader(requestBytes))
+		require.NoError(t, err)
+
+		responseBytes, err := io.ReadAll(resp.Body)
+		require.NoError(t, err)
+
+		corectResponseBytes := readJaegerTraces(t)
+		assert.Equal(t, corectResponseBytes, responseBytes)
```
yurishkuro, line 646: nit: by convention tests use `expected` not `correct`

yurishkuro (commented):
Found one issue on the UI side: https://github.com/jaegertracing/jaeger-ui/pull/2145#issuecomment-1937870768

cmd/query/app/http_handler_test.go
```diff
@@ -623,6 +632,39 @@ func TestGetOperationsLegacyStorageFailure(t *testing.T) {
...
+	}, querysvc.QueryServiceOptions{})
+}
+
+func TestTransformOTLPReadError(t *testing.T) {
+	withTestServer(func(ts *testServer) {
+		bytesReader := &IoReaderMock{}
+		bytesReader.On("Read", mock.AnythingOfType("[]uint8")).Return(0, errors.New("Mocked error"))
+		_, err := ts.server.Client().Post(ts.server.URL+"/api/transform", "application/json", bytesReader)
```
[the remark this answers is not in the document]
NavinShrinivas: yes this will never read the logic we written but by using this mocked IO reader I'm  trying to test the error block in httphandler API to increase test percentages (which are still low as I haven't been able to cover convertToUI)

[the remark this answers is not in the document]
yurishkuro: ok. But I see coverage already very high, you don't need to add any more
```
$ go test -cover ./cmd/query/app
ok  	github.com/jaegertracing/jaeger/cmd/query/app	3.683s	coverage: 99.0% of statements
```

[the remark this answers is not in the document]
NavinShrinivas: Do you suggest removing this test as well?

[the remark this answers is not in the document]
yurishkuro: let's keep it

yurishkuro (approved):

cmd/query/app/http_handler.go
```diff
@@ -193,6 +195,22 @@ func (aH *APIHandler) getOperationsLegacy(w http.ResponseWriter, r *http.Request
...
+		return
+	}
+
+	traces, err := otlp2traces(body)
+	if aH.handleError(w, err, http.StatusInternalServerError) {
+		return
+	}
+
+	var uiErrors []structuredError
+	structuredRes := aH.tracesToResponse(traces, false, uiErrors)
```
yurishkuro, line 210: moved to helper function. Also improved code coverage since the ui-conversion error is tested in one of the scenarios already

cmd/query/app/http_handler_test.go
```diff
@@ -623,6 +633,52 @@ func TestGetOperationsLegacyStorageFailure(t *testing.T) {
...
 
+func TestTransformOTLPSuccess(t *testing.T) {
+	reformat := func(in []byte) []byte {
+		obj := new(interface{})
+		require.NoError(t, json.Unmarshal(in, obj))
+		// format json similar to `jq .`
+		out, err := json.MarshalIndent(obj, "", "  ")
+		require.NoError(t, err)
+		return out
+	}
```
yurishkuro, line 644: you were write the first time about employing marshaling to ensure consistent formatting. I changed the expected fixture to be readable - also added a 2nd trace to test a branch in the translator

yurishkuro (approved):

yurishkuro (approved):

PR #5158, opened 2024-02-02
Add Elasticsearch storage support for adaptive sampling

Pushkarm029:
- #3305

## Description of the changes
- Implemented Elasticsearch storage for adaptive sampling

- not tested yet

## Checklist


Pushkarm029:
Raised PR to get to know from reviewers whether I am on the right track or not.

yurishkuro (commented):
* run `make fmt`
* please add tests to all new functionality
* enable integration tests (see plugin/storage/integration/integration.go:423)

Pushkarm029:
Enabling integration tests

Pushkarm029:
The Tests are failing because the `GetLatestProbability` is failing
```
    integration.go:466: 
        	Error Trace:	/home/runner/work/jaeger/jaeger/plugin/storage/integration/integration.go:466
        	Error:      	Should be true
        	Test:       	TestElasticsearchStorage/GetLatestProbability
    integration.go:467: 	 Expected: map[new-srv:map[op:0.1]]
    integration.go:468: 	 Actual  : map[new-srv3:map[op:0.123]]
```

Pushkarm029:
Tests are not failing anymore. But I am still suspicious. That's why adding more cases in `testGetLatestProbability` in the integration test.

Pushkarm029:
any idea @yurishkuro, why it is failing on es v5 but not on v7 and v8.

plugin/storage/integration/integration.go
```diff
@@ -453,9 +453,12 @@ func (s *StorageIntegration) testGetLatestProbability(t *testing.T) {
 	defer s.cleanUp(t)
 
 	s.SamplingStore.InsertProbabilitiesAndQPS("newhostname1", samplemodel.ServiceOperationProbabilities{"new-srv3": {"op": 0.123}}, samplemodel.ServiceOperationQPS{"new-srv2": {"op": 11}})
-	s.SamplingStore.InsertProbabilitiesAndQPS("dell11eg843d", samplemodel.ServiceOperationProbabilities{"new-srv": {"op": 0.1}}, samplemodel.ServiceOperationQPS{"new-srv": {"op": 4}})
+	s.SamplingStore.InsertProbabilitiesAndQPS("dell11eg843d", samplemodel.ServiceOperationProbabilities{"new-srv": {"op": 0.23}}, samplemodel.ServiceOperationQPS{"new-srv": {"op": 4}})
+	s.SamplingStore.InsertProbabilitiesAndQPS("sghdvavdasvd", samplemodel.ServiceOperationProbabilities{"new-srv": {"op": 0.2}}, samplemodel.ServiceOperationQPS{"new-srv": {"op": 8}})
+	s.SamplingStore.InsertProbabilitiesAndQPS("sjsvdfgvdsavg", samplemodel.ServiceOperationProbabilities{"new-srv": {"op": 0.1}}, samplemodel.ServiceOperationQPS{"new-srv": {"op": 2}})
+	s.SamplingStore.InsertProbabilitiesAndQPS("newjaegerv2", samplemodel.ServiceOperationProbabilities{"new-srv": {"op": 0.106}}, samplemodel.ServiceOperationQPS{"new-srv": {"op": 9}})
```
yurishkuro, line 459: what are you trying to achieve with these additional inserts?

plugin/storage/integration/elasticsearch_test.go
```diff
@@ -139,6 +143,27 @@ func (s *ESStorageIntegration) esCleanUp(allTagsAsFields, archive bool) error {
...
 
+func (s *ESStorageIntegration) initSamplingStore() error {
+	bp, _ := s.client.BulkProcessor().BulkActions(1).FlushInterval(time.Nanosecond).Do(context.Background())
+	s.bulkProcessor = bp
+	esVersion, err := s.getVersion()
+	if err != nil {
+		return err
+	}
+	client := eswrapper.WrapESClient(s.client, bp, esVersion, s.v8Client)
+	clientFn := func() estemplate.Client { return client }
```
yurishkuro, line 154: please don't duplicate code, extract into a helper function and reuse

plugin/storage/es/samplingstore/storage.go
```diff
@@ -0,0 +1,197 @@
...
+type TimeProbabilitiesAndQPS struct {
+	Timestamp           time.Time           `json:"timestamp"`
+	ProbabilitiesAndQPS ProbabilitiesAndQPS `json:"probabilitiesandqps"`
+}
+
+type ProbabilitiesAndQPS struct {
+	Hostname      string
+	Probabilities model.ServiceOperationProbabilities
+	QPS           model.ServiceOperationQPS
+}
```
yurishkuro, line 68: I would move these to a separate file like `dbmodel.go`

plugin/storage/es/factory.go
```diff
@@ -272,6 +274,25 @@ func createSpanWriter(
...
+func createSamplingStore(
+	clientFn func() es.Client,
+	cfg *config.Configuration,
+	logger *zap.Logger,
+) (samplingstore.Store, error) {
+	store := esSampling.NewSamplingStore(esSampling.SamplingStoreParams{
+		Client:          clientFn,
+		Logger:          logger,
+		IndexPrefix:     cfg.IndexPrefix,
+		IndexDateLayout: cfg.IndexDateLayoutDependencies,
```
yurishkuro, line 290: why "dependencies"?

plugin/storage/integration/integration.go
Pushkarm029: Previously, there were only two tests, and the code is this:
``` go
lengthOfSearchResult := len(searchResult.Hits.Hits)
if lengthOfSearchResult == 0 {
	return nil, nil
}
hit := searchResult.Hits.Hits[lengthOfSearchResult-1]
```
With the two test cases, it was sure that the test would not fail. That's why I added more inserts. To increase my surety.

If you want, I can remove it.

plugin/storage/es/factory.go
Pushkarm029: sorry, i should have checked it before pushing.

plugin/storage/integration/integration.go
yurishkuro: unless there is an extra edge condition triggered, how is 6 inserts different from 2?

plugin/storage/es/samplingstore/storage.go
```diff
@@ -0,0 +1,197 @@
...
+
+type SamplingStore struct {
+	client              func() es.Client
+	logger              *zap.Logger
+	samplingIndexPrefix string
+	indexDateLayout     string
+	maxDocCount         int
+}
+
+type SamplingStoreParams struct {
```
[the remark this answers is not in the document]
Pushkarm029: I tried creating a common store-params type, but it looks less clean and concise than the current one.

plugin/storage/integration/integration.go
Pushkarm029: You are right. 2 inserts function the same as 6 inserts. Removed in the current push

yurishkuro:
es-5 tests seems to consistently fail here, and precisely on the new tests, so must be related.

Pushkarm029:
Tests are not failing anymore..

yurishkuro:
> codecov/patch — 84.29% of diff hit (target 95.00%)

can you improve test coverage?

Pushkarm029:
done

yurishkuro (commented):
please undo jaeger-ui submodule change

Pushkarm029:
done

plugin/storage/es/samplingstore/dbmodel/model.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"time"
+
+	"github.com/jaegertracing/jaeger/cmd/collector/app/sampling/model"
+)
+
+type TimeThroughput struct {
+	Timestamp  time.Time           `json:"timestamp"`
+	Throughput []*model.Throughput `json:"throughputs"`
```
yurishkuro, line 25: It is not safe to depend on the `model.*` type for DB Objects - `model.*` is internal, it can change, which should not affect how JSON is stored inside of ES. I suggest re-defining fully internal data types here and converting from `model` to `dbmodel` fully.

plugin/storage/integration/elasticsearch_test.go
```diff
@@ -139,14 +143,41 @@ func (s *ESStorageIntegration) esCleanUp(allTagsAsFields, archive bool) error {
 	return s.initSpanstore(allTagsAsFields, archive)
 }
 
-func (s *ESStorageIntegration) initSpanstore(allTagsAsFields, archive bool) error {
+func (s *ESStorageIntegration) initSamplingStore() error {
+	client, err := s.getEsClient()
+	if err != nil {
```
yurishkuro, line 148: One useful improvement we could make to these helper functions is to pass `t` all the way from the test entry point and just use `require.NoError(t, err)` instead of checks & return values.

Pushkarm029: working on this one

Pushkarm029: I added some changes in the last commit. please check

plugin/storage/es/samplingstore/dbmodel/converter.go
```diff
@@ -0,0 +1,51 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package dbmodel
+
+import (
+	"github.com/jaegertracing/jaeger/cmd/collector/app/sampling/model"
+)
+
+func FromThroughputs(throughputs []*model.Throughput) []model.Throughput {
```
yurishkuro, line 21: I don't follow - you are converting between identical objects, just from pointers to values. The point was not to use jaeger/model types in the structs defined in dbmodel/model.go, but re-declare similar structure and then transform between that and jaeger/model

Pushkarm029: got it

plugin/storage/es/samplingstore/storage.go
```diff
@@ -0,0 +1,189 @@
...
+		Probabilities: probabilities,
+		QPS:           qps,
+	}
+	s.writeProbabilitiesAndQPS(writeIndexName, ts, val)
+	return nil
+}
+
+func (s *SamplingStore) writeProbabilitiesAndQPS(indexName string, ts time.Time, pandqps dbmodel.ProbabilitiesAndQPS) {
+	s.client().Index().Index(indexName).Type(probabilitiesType).
+		BodyJson(&dbmodel.TimeProbabilitiesAndQPS{
```
yurishkuro, line 89: data conversion it not necessary again

plugin/storage/es/samplingstore/storage.go
```diff
@@ -0,0 +1,189 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to search for throughputs: %w", err)
+	}
+	var retSamples []dbmodel.Throughput
+	hits := searchResult.Hits.Hits
+	for _, hit := range hits {
+		source := hit.Source
+		var tToD dbmodel.TimeThroughput
+		if err := json.Unmarshal(*source, &tToD); err != nil {
+			return nil, errors.New("unmarshalling ElasticSearch documents failed")
```
yurishkuro, line 124: don't swallow errors
```suggestion
			return nil, errors.New("unmarshalling ElasticSearch documents failed: %w", err)
```

yurishkuro: don't need ElasticSearch in the message

plugin/storage/es/samplingstore/storage.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+	lengthOfSearchResult := len(searchResult.Hits.Hits)
+	if lengthOfSearchResult == 0 {
+		return nil, nil
+	}
+
+	var latestProbabilities dbmodel.TimeProbabilitiesAndQPS
+	latestTime := time.Time{}
+	for _, hit := range searchResult.Hits.Hits {
+		var unMarshalProbabilities dbmodel.TimeProbabilitiesAndQPS
```
yurishkuro, line 149:
```suggestion
		var data dbmodel.TimeProbabilitiesAndQPS
```

plugin/storage/es/samplingstore/storage.go
```diff
@@ -0,0 +1,189 @@
...
+		err = json.Unmarshal(*hit.Source, &unMarshalProbabilities)
+		if err != nil {
```
yurishkuro, line 151:
```suggestion
		if err = json.Unmarshal(*hit.Source, &unMarshalProbabilities); err != nil {
```

plugin/storage/es/samplingstore/storage.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+	return latestProbabilities.ProbabilitiesAndQPS.Probabilities, nil
+}
+
+func buildTSQuery(start, end time.Time) elastic.Query {
+	return elastic.NewRangeQuery("timestamp").Gte(start).Lte(end)
+}
+
+func (s *SamplingStore) getLatestIndices() []string {
+	currTime := time.Now().UTC()
```
yurishkuro, line 167: this doesn't look good. If a user rotates indices every hour, and for some reason the adaptive sampling hasn't run for the last hour, this will never find the records that may actually be in the database. Maybe it needs some kind of lookback configuration parameter.

plugin/storage/es/samplingstore/storage.go
```diff
@@ -0,0 +1,189 @@
...
+
+func (s *SamplingStore) writeProbabilitiesAndQPS(indexName string, ts time.Time, pandqps dbmodel.ProbabilitiesAndQPS) {
+	s.client().Index().Index(indexName).Type(probabilitiesType).
+		BodyJson(&dbmodel.TimeProbabilitiesAndQPS{
+			Timestamp:           ts,
+			ProbabilitiesAndQPS: pandqps,
+		}).Add()
+}
+
+func indexWithDate(indexNamePrefix, indexDateLayout string, date time.Time) string {
```
yurishkuro, line 95: could we move it to the bottom of the file where the other helper functions are located? next to prefixIndexName

plugin/storage/integration/elasticsearch_test.go
```diff
@@ -100,27 +101,26 @@ func (s *ESStorageIntegration) getVersion() (uint, error) {
 	return uint(esVersion), nil
 }
 
-func (s *ESStorageIntegration) initializeES(allTagsAsFields, archive bool) error {
+func (s *ESStorageIntegration) initializeES(allTagsAsFields, archive bool, t *testing.T) error {
```
yurishkuro, line 104: nit: testing.T is usually passed as the 1st param

plugin/storage/integration/elasticsearch_test.go
```diff
@@ -100,27 +101,26 @@ func (s *ESStorageIntegration) getVersion() (uint, error) {
...
 		Addresses:            []string{queryURL},
 		DiscoverNodesOnStart: false,
 	})
-	if err != nil {
-		return err
-	}
+	require.NoError(t, err)
+
+	err = s.initSpanstore(allTagsAsFields, archive)
+	require.NoError(t, err)
```
yurishkuro, line 119:
```suggestion
	s.initSpanstore(t, allTagsAsFields, archive)
```
and let it check the error internally

plugin/storage/integration/elasticsearch_test.go
```diff
@@ -139,14 +139,39 @@ func (s *ESStorageIntegration) esCleanUp(allTagsAsFields, archive bool) error {
 	return s.initSpanstore(allTagsAsFields, archive)
 }
 
-func (s *ESStorageIntegration) initSpanstore(allTagsAsFields, archive bool) error {
+func (s *ESStorageIntegration) initSamplingStore(t *testing.T) error {
+	client, err := s.getEsClient()
+	require.NoError(t, err)
```
yurishkuro, line 144:
```suggestion
	client, err := s.getEsClient(t)
```

plugin/storage/integration/elasticsearch_test.go
```diff
@@ -139,14 +139,39 @@ func (s *ESStorageIntegration) esCleanUp(allTagsAsFields, archive bool) error {
...
+			IndexPrefix:     indexPrefix,
+			IndexDateLayout: indexDateLayout,
+			MaxDocCount:     defaultMaxDocCount,
+		})
+	s.SamplingStore = w
+	return nil
+}
+
+func (s *ESStorageIntegration) getEsClient() (eswrapper.ClientWrapper, error) {
 	bp, _ := s.client.BulkProcessor().BulkActions(1).FlushInterval(time.Nanosecond).Do(context.Background())
```
yurishkuro, line 159:
```suggestion
	bp, err := s.client.BulkProcessor().BulkActions(1).FlushInterval(time.Nanosecond).Do(context.Background())
	require.NoError(t, err)
```

plugin/storage/integration/elasticsearch_test.go
```diff
@@ -269,7 +294,7 @@ func TestElasticsearchStorage_IndexTemplates(t *testing.T) {
 		t.Fatal(err)
 	}
 	s := &ESStorageIntegration{}
-	require.NoError(t, s.initializeES(true, false))
+	require.NoError(t, s.initializeES(true, false, t))
```
yurishkuro, line 297: it should not return an error, it can check it internally

plugin/storage/es/samplingstore/storage.go
Pushkarm029: It is necessary because in this type conversion(`dbmodel.ProbabilitiesAndQPS` -> `dbmodel.TimeProbabilitiesAndQPS`), we are adding a timestamp also, which is important for it to work properly.

plugin/storage/es/samplingstore/storage.go
Pushkarm029: added an algorithm in recent commit to get latest existing index.

plugin/storage/es/samplingstore/storage.go
```diff
@@ -0,0 +1,199 @@
...
+	for {
+		indexName := indexWithDate(s.samplingIndexPrefix, s.indexDateLayout, now)
+		exists, err := s.client().IndexExists(indexName).Do(ctx)
+		if err != nil {
+			return nil, err
+		}
+		if exists {
+			return []string{indexName}, nil
+		}
+		now = now.Add(-time.Hour * 24)
```
yurishkuro, line 170: this is not exactly what I had in mind. First, this loop can be infinite on an empty database. Second, 24hrs is pretty arbitrary; we don't know how often users rotate indices, so T-24h could skip the latest index and go to earlier one.

My suggestion was to have a user-provided configuration parameter how far back the search should go.

plugin/storage/es/samplingstore/storage.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+}
+
+func (s *SamplingStore) getReadIndices(start, end time.Time) []string {
+	var indices []string
+	firstIndex := indexWithDate(s.samplingIndexPrefix, s.indexDateLayout, start)
+	currentIndex := indexWithDate(s.samplingIndexPrefix, s.indexDateLayout, end)
+	for currentIndex != firstIndex {
+		indices = append(indices, currentIndex)
+		end = end.Add(-24 * time.Hour)
```
yurishkuro, line 180: similar to above, why 24hrs? How does this logic look in our main ES Span Store?

Pushkarm029: in es span store, it is decided by this fn.
```go
func (c *Configuration) GetIndexRolloverFrequencySpansDuration() time.Duration {
	if c.IndexRolloverFrequencySpans == "hour" {
		return -1 * time.Hour
	}
	return -24 * time.Hour
}
```

Pushkarm029: implemented the same here

Pushkarm029:
I added a user-provided configuration parameter(MaxSampleTime) to limit scanning for the latest index. 
- Currently, I set the default for the config `MaxSampleTime` as `72 * time.Hour`.
- I will add tests if you think the current approach is good.

plugin/storage/es/samplingstore/storage.go
```diff
@@ -0,0 +1,214 @@
...
+	firstIndex := indexWithDate(indexName, indexDateLayout, startTime)
+	currentIndex := indexWithDate(indexName, indexDateLayout, endTime)
+	for currentIndex != firstIndex {
+		if len(indices) == 0 || indices[len(indices)-1] != currentIndex {
+			indices = append(indices, currentIndex)
+		}
+		endTime = endTime.Add(reduceDuration)
+		currentIndex = indexWithDate(indexName, indexDateLayout, endTime)
+	}
+	indices = append(indices, firstIndex)
```
yurishkuro, line 197: the logic above is rather convoluted, could we not make it simpler?
```suggestion
	lastIndex := indexWithDate(indexName, indexDateLayout, endTime)
	indices := []string{lastIndex} // always include the latest
	currentIndex := indexWithDate(indexName, indexDateLayout, startTime)
	for currentIndex != lastIndex {
		indices = append(indices, currentIndex)
		startTime = startTime.Add(lookback)
		currentIndex = indexWithDate(indexName, indexDateLayout, startTime)
	}
```

plugin/storage/integration/elasticsearch_test.go
```diff
@@ -100,53 +101,67 @@ func (s *ESStorageIntegration) getVersion() (uint, error) {
...
 
-func (s *ESStorageIntegration) esCleanUp(allTagsAsFields, archive bool) error {
+func (s *ESStorageIntegration) esCleanUp(t *testing.T, allTagsAsFields, archive bool) error {
 	_, err := s.client.DeleteIndex("*").Do(context.Background())
-	if err != nil {
-		return err
-	}
-	return s.initSpanstore(allTagsAsFields, archive)
+	require.NoError(t, err)
+	return s.initSpanstore(t, allTagsAsFields, archive)
```
yurishkuro, line 134: don't return, check (e.g. error is not checked in L125)

plugin/storage/integration/elasticsearch_test.go
```diff
@@ -100,53 +101,67 @@ func (s *ESStorageIntegration) getVersion() (uint, error) {
...
 	s.Refresh = s.esRefresh
-	s.esCleanUp(allTagsAsFields, archive)
+	s.esCleanUp(t, allTagsAsFields, archive)
 	// TODO: remove this flag after ES support returning spanKind when get operations
 	s.GetOperationsMissingSpanKind = true
 	return nil
 }
 
-func (s *ESStorageIntegration) esCleanUp(allTagsAsFields, archive bool) error {
+func (s *ESStorageIntegration) esCleanUp(t *testing.T, allTagsAsFields, archive bool) error {
```
yurishkuro, line 131:
```suggestion
func (s *ESStorageIntegration) esCleanUp(t *testing.T, allTagsAsFields, archive bool) {
```

plugin/storage/integration/elasticsearch_test.go
```diff
@@ -100,53 +101,67 @@ func (s *ESStorageIntegration) getVersion() (uint, error) {
...
-		return err
-	}
-	return s.initSpanstore(allTagsAsFields, archive)
+	require.NoError(t, err)
+	return s.initSpanstore(t, allTagsAsFields, archive)
 }
 
-func (s *ESStorageIntegration) initSpanstore(allTagsAsFields, archive bool) error {
-	bp, _ := s.client.BulkProcessor().BulkActions(1).FlushInterval(time.Nanosecond).Do(context.Background())
+func (s *ESStorageIntegration) initSamplingStore(t *testing.T) error {
```
yurishkuro, line 137:
```suggestion
func (s *ESStorageIntegration) initSamplingStore(t *testing.T) {
```

plugin/storage/integration/elasticsearch_test.go
```diff
@@ -100,53 +101,67 @@ func (s *ESStorageIntegration) getVersion() (uint, error) {
...
+		})
+	s.SamplingStore = w
+	return nil
+}
+
+func (s *ESStorageIntegration) getEsClient(t *testing.T) eswrapper.ClientWrapper {
+	bp, err := s.client.BulkProcessor().BulkActions(1).FlushInterval(time.Nanosecond).Do(context.Background())
+	require.NoError(t, err)
 	s.bulkProcessor = bp
+	var client eswrapper.ClientWrapper
```
yurishkuro, line 156: not needed

plugin/storage/integration/elasticsearch_test.go
```diff
@@ -100,53 +101,67 @@ func (s *ESStorageIntegration) getVersion() (uint, error) {
...
 	esVersion, err := s.getVersion()
-	if err != nil {
-		return err
-	}
-	client := eswrapper.WrapESClient(s.client, bp, esVersion, s.v8Client)
+	require.NoError(t, err)
+	client = eswrapper.WrapESClient(s.client, bp, esVersion, s.v8Client)
+	return client
```
yurishkuro, line 160:
```suggestion
	return eswrapper.WrapESClient(s.client, bp, esVersion, s.v8Client)
```

plugin/storage/es/samplingstore/storage.go
Pushkarm029: I am wondering if the order or position of indices matters because the tests are currently failing due to it.
In suggested logic : if start=10 feb, end = 12 feb, we get : 
```go
"prefix-jaeger-sampling-2024-02-12",
"prefix-jaeger-sampling-2024-02-10", 
"prefix-jaeger-sampling-2024-02-11"
```
if order matters, then this code lgtm : 
```go
var indices []string
firstIndex := indexWithDate(indexName, indexDateLayout, startTime)
currentIndex := indexWithDate(indexName, indexDateLayout, endTime)
	endTime = endTime.Add(lookback)
	currentIndex = indexWithDate(indexName, indexDateLayout, endTime)
}
indices = append(indices, firstIndex)
```

yurishkuro: whether the order matters or not depends on how you use the returned data. Since you're picking the latest entry based on the timestamp, the order should not matter. However, in theory it's more efficient to search indices in reverse chronological order because the most recent one is most likely to have the needed data. Instead, if you search through all indices in the lookback interval then you will be crunching a lot of data on every iteration, even though you're only interested in the latest entry.

Pushkarm029: ok thanks, got it.

Pushkarm029:
I am working to increase code coverage.

Pushkarm029:
please review

pkg/es/config/config.go
```diff
@@ -69,9 +69,12 @@ type Configuration struct {
 	IndexPrefix                    string         `mapstructure:"index_prefix"`
 	IndexDateLayoutSpans           string         `mapstructure:"-"`
 	IndexDateLayoutServices        string         `mapstructure:"-"`
+	IndexDateLayoutSampling        string         `mapstructure:"-"`
```
yurishkuro, line 72: do we not need to define templates for the new indices you're introducing? `plugin/storage/es/mappings/`

plugin/storage/es/samplingstore/storage.go
```diff
@@ -0,0 +1,213 @@
...
+}
+
+type SamplingStoreParams struct {
+	Client                         func() es.Client
+	Logger                         *zap.Logger
+	IndexPrefix                    string
+	IndexDateLayout                string
+	MaxDocCount                    int
+	SamplingIndexRolloverFrequency time.Duration
+	AdaptiveSamplingLookback       time.Duration
```
yurishkuro, line 55: there's no need to keep referring to 'sampling' here, it's already clear contextually (unlike es/config/Config where all kinds of attributes are mixed in one place)

```suggestion
	IndexRolloverFrequency time.Duration
	Lookback       time.Duration
	MaxDocCount                    int
```

pkg/es/config/config.go
Pushkarm029: Yes, we do need it. I am working on it.

Pushkarm029:
I added mappings for the new indices. also, I tried to add support for rollover for es sampling store. but, it fails tests and makes this PR very complex to work on.

We can create another PR to add support. Do you think we can leave this for now?

plugin/storage/integration/es_index_cleaner_test.go
```diff
@@ -158,7 +159,7 @@ func createAllIndices(client *elastic.Client, prefix string) error {
 	// create daily indices and archive index
 	err := createEsIndices(client, []string{
 		prefixWithSeparator + spanIndexName, prefixWithSeparator + serviceIndexName,
-		prefixWithSeparator + dependenciesIndexName, prefixWithSeparator + archiveIndexName,
+		prefixWithSeparator + dependenciesIndexName, prefixWithSeparator + samplingIndexName, prefixWithSeparator + archiveIndexName,
```
yurishkuro, line 162: one per line is better for readability

yurishkuro:
> To simplify the process, We can create another PR to add support. Do you think we can leave rollover support for now?

since you already have the code in here, let's just finish it in one place.

Pushkarm029:
added rollover support

Pushkarm029:
Please review.

cmd/es-rollover/app/index_options.go
```diff
@@ -65,6 +65,12 @@ func RolloverIndices(archive bool, skipDependencies bool, prefix string) []Index
 		})
 	}
```
yurishkuro, line 67: I see that above is conditional on `!skipDependencies`
```
      --skip-dependencies           Disable rollover for dependencies index
```
I think we need a similar one for adaptive sampling, since not everyone will be using it. I would also make it skipped by default, otherwise the rollover might start failing for people after upgrade - they should explicitly opt-in for this. Suggestion:
```
      --adaptive-sampling           Enable rollover for adaptive sampling index
```

cmd/es-rollover/app/index_options_test.go
```diff
@@ -152,6 +180,7 @@ func TestRolloverIndices(t *testing.T) {
 				test.prefix += "-"
 			}
 			result := RolloverIndices(test.archive, test.skipDependencies, test.prefix)
+			// assert.Equal(t, len(result), len(test.expected))
```
yurishkuro, line 183: feel free to leave it in

cmd/es-rollover/app/index_options.go
yurishkuro: and we should have unit test that checks the opt-in nature

pkg/es/config/config.go
```diff
@@ -299,6 +305,14 @@ func (c *Configuration) GetIndexRolloverFrequencyServicesDuration() time.Duratio
 	return -24 * time.Hour
 }
 
+// GetIndexRolloverFrequencySamplingDuration returns jaeger-sampling index rollover frequency duration
+func (c *Configuration) GetIndexRolloverFrequencySamplingDuration() time.Duration {
+	if c.IndexRolloverFrequencySampling == "hour" {
+		return -1 * time.Hour
+	}
+	return -24 * time.Hour
+}
```
yurishkuro, line 314: you can refactor all 3 functions to invoke a shared helper that would convert a frequency string to Duration.

plugin/storage/es/factory.go
```diff
@@ -272,6 +274,27 @@ func createSpanWriter(
 	return writer, nil
 }
 
+func (f *Factory) CreateSamplingStore(maxBuckets int) (samplingstore.Store, error) {
+	return createSamplingStore(f.getPrimaryClient, f.primaryConfig, f.logger)
+}
+
+func createSamplingStore(
```
yurishkuro, line 281: unnecessary function, never called by anything else, can inline into CreateSamplingStore

plugin/storage/es/mappings/jaeger-sampling-7.json
```diff
@@ -0,0 +1,21 @@
...
+  {{- if .UseILM }}
+  "aliases": {
+    "{{ .IndexPrefix }}jaeger-sampling-read" : {}
+  },
+  {{- end }}
+  "settings":{
+    "index.number_of_shards": {{ .Shards }},
+    "index.number_of_replicas": {{ .Replicas }},
+    "index.mapping.nested_fields.limit":50,
+    "index.requests.cache.enable":true
```
yurishkuro, line 12: Are you sure about this? The nature of throughput is such that you always want to read & return the latest, I don't think caching is productive for it.

plugin/storage/es/samplingstore/storage.go
```diff
@@ -0,0 +1,221 @@
...
+)
+
+type SamplingStore struct {
+	client                         func() es.Client
+	logger                         *zap.Logger
+	samplingIndexPrefix            string
+	indexDateLayout                string
+	maxDocCount                    int
+	samplingIndexRolloverFrequency time.Duration
+	adaptiveSamplingLookback       time.Duration
```
yurishkuro, line 45:
```suggestion
	indexRolloverFrequency time.Duration
	lookback       time.Duration
```

plugin/storage/es/samplingstore/storage.go
```diff
@@ -0,0 +1,221 @@
...
+		Size(s.maxDocCount).
+		Query(buildTSQuery(start, end)).
+		IgnoreUnavailable(true).
+		Do(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to search for throughputs: %w", err)
+	}
+	var retSamples []dbmodel.Throughput
+	hits := searchResult.Hits.Hits
+	for _, hit := range hits {
```
yurishkuro, line 90: unnecessary deref
```suggestion
	for _, hit := range searchResult.Hits.Hits {
```

plugin/storage/es/samplingstore/storage.go
```diff
@@ -0,0 +1,221 @@
...
+	indices := getReadIndices(s.samplingIndexPrefix, s.indexDateLayout, start, end, s.samplingIndexRolloverFrequency)
+	searchResult, err := s.client().Search(indices...).
+		Size(s.maxDocCount).
+		Query(buildTSQuery(start, end)).
+		IgnoreUnavailable(true).
+		Do(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to search for throughputs: %w", err)
+	}
+	var retSamples []dbmodel.Throughput
```
yurishkuro, line 88: you can pre-allocate based on `len(searchResult.Hits.Hits)` and assign by position instead of using append

plugin/storage/es/samplingstore/storage.go
```diff
@@ -0,0 +1,221 @@
...
+			ProbabilitiesAndQPS: pandqps,
+		}).Add()
+}
+
+func (s *SamplingStore) CreateTemplates(samplingTemplate string) error {
+	_, err := s.client().CreateTemplate("jaeger-sampling").Body(samplingTemplate).Do(context.Background())
+	if err != nil {
+		return err
+	}
+	return nil
```
yurishkuro, line 171:
```suggestion
	return err
```

plugin/storage/es/samplingstore/storage.go
```diff
@@ -0,0 +1,221 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to check index existence: %w", err)
+		}
+		if exists {
+			return []string{currentIndex}, nil
+		}
+		if currentIndex == endIndex {
+			return nil, fmt.Errorf("falied to find latest index")
+		}
+		now = now.Add(rollover)
```
yurishkuro, line 191: I don't know why rollover is negative (bad coding pattern)
```suggestion
		now = now.Add(rollover) // rollover is negative
```

plugin/storage/es/samplingstore/storage.go
```diff
@@ -0,0 +1,221 @@
...
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func getLatestIndices(indexPrefix, indexDateLayout string, clientFn es.Client, rollover time.Duration, maxDuration time.Duration) ([]string, error) {
+	ctx := context.Background()
+	now := time.Now().UTC()
+	end := now.Add(-maxDuration)
```
yurishkuro, line 177: `end` is quite confusing in this context
```suggestion
	earliest := now.Add(-maxDuration)
```

plugin/storage/es/samplingstore/storage.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+}
+
+func getReadIndices(indexName, indexDateLayout string, startTime time.Time, endTime time.Time, rollover time.Duration) []string {
+	var indices []string
+	firstIndex := indexWithDate(indexName, indexDateLayout, startTime)
+	currentIndex := indexWithDate(indexName, indexDateLayout, endTime)
+	for currentIndex != firstIndex {
+		indices = append(indices, currentIndex)
+		endTime = endTime.Add(rollover)
```
yurishkuro, line 201:
```suggestion
		endTime = endTime.Add(rollover) // rollover is negative
```

plugin/storage/es/samplingstore/storage_test.go
```diff
@@ -0,0 +1,417 @@
...
+		searchResult   *elastic.SearchResult
+		searchError    error
+		expectedError  string
+		expectedOutput []*samplemodel.Throughput
+		indexPrefix    string
+		maxDocCount    int
+		indices        []interface{}
+	}{
+		{
+			searchResult: createSearchResult(goodThroughputs),
```
yurishkuro, line 276: it's customary to have a `name` property for the tests that makes it easier to locate in case of failure. Right now your subtests will be named /1, /2, ...

yurishkuro (commented):
almost there!

plugin/storage/es/samplingstore/storage.go
Pushkarm029: I think append will be better because both are arrays; if we assign by position, we have to create a 2d array, and then convert this 2d array of dbmodel throughput into a 1d array of model throughput.

Pushkarm029:
tests are also failing. looks like i have to add adaptive sampling in this `cmd/es-index-cleaner/app/flags.go Config` also????

cmd/es-index-cleaner/app/index_filter.go
```diff
@@ -50,9 +50,9 @@ func (i *IndexFilter) filter(indices []client.Index) []client.Index {
 		// archive works only for rollover
 		reg, _ = regexp.Compile(fmt.Sprintf("^%sjaeger-span-archive-\\d{6}", i.IndexPrefix))
 	case i.Rollover:
-		reg, _ = regexp.Compile(fmt.Sprintf("^%sjaeger-(span|service|dependencies)-\\d{6}", i.IndexPrefix))
+		reg, _ = regexp.Compile(fmt.Sprintf("^%sjaeger-(span|service|dependencies|sampling)-\\d{6}", i.IndexPrefix))
 	default:
-		reg, _ = regexp.Compile(fmt.Sprintf("^%sjaeger-(span|service|dependencies)-\\d{4}%s\\d{2}%s\\d{2}", i.IndexPrefix, i.IndexDateSeparator, i.IndexDateSeparator))
+		reg, _ = regexp.Compile(fmt.Sprintf("^%sjaeger-(span|service|dependencies|sampling)-\\d{4}%s\\d{2}%s\\d{2}", i.IndexPrefix, i.IndexDateSeparator, i.IndexDateSeparator))
```
yurishkuro, line 55: hm, arguably, this is a pretty weird design (not your fault, predates your PR). If es-rollover has flags to include deps/sampling, one would expect es-cleaner to also include such flags, not to blast all possible indices. But not worth tackling now.

plugin/storage/es/samplingstore/storage.go
yurishkuro: not sure what you mean. I am talking about this pattern:
```
input := ...
output = make([]Output, len(input))
for i, v := range input {
  json.Unmarshal(v, &output[i])
}
```
In this style you pre-allocate the output array of exactly the size you need, which can be more efficient than doing append and potentially having to extend the slice. You can also unmarshal directly into array element (if the slice is defined over values, not pointers) and avoid extra copying.

Btw, in `retSamples = append(retSamples, tToD.Throughput...)` the `...` doesn't make sense since you're appending a single element, not another slice.

plugin/storage/es/samplingstore/storage_test.go
```diff
@@ -0,0 +1,486 @@
...
+		start time.Time
+		end   time.Time
+	}{
+		{
+			name:  "",
+			start: time.Date(2024, time.February, 10, 0, 0, 0, 0, time.UTC),
+			end:   time.Date(2024, time.February, 12, 0, 0, 0, 0, time.UTC),
+		},
+	}
+	for _, test := range tests {
```
yurishkuro, line 108: there's no value in this array and for loop if you're only doing a single test

plugin/storage/es/samplingstore/storage_test.go
```diff
@@ -0,0 +1,486 @@
...
+func TestInsertThroughput(t *testing.T) {
+	tests := []struct {
+		name          string
+		expectedError string
+		esVersion     uint
+	}{
+		{
+			name:          "es v6",
+			expectedError: "",
+			esVersion:     6,
```
yurishkuro, line 186: did you mean to add v7 / v8? Is the logic actually different by version and against a mock storage? If not, there's no need to have a loop

plugin/storage/es/samplingstore/storage_test.go
```diff
@@ -0,0 +1,486 @@
...
+			"Hostname": "dell11eg843d",
+			"Probabilities": {
+				"new-srv": {"op": 0.1}
+			},
+			"QPS": {
+				"new-srv": {"op": 4}
+			}
+		}
+	}`
+	badProbabilities := `badJson{hello}world`
```
yurishkuro, line 378: nit: if you're not reusing this constant, then inline it for better readability

yurishkuro:
> tests are also failing. looks like i have to add adaptive sampling in this cmd/es-index-cleaner/app/flags.go Config also????

heh, I left a comment about that, but why would have any impact on the tests? The cleaner is just over-eager in which indices it removes, but unless the integration tests are actually testing for this behavior I would expect to be a no-op (because either rollover will create extra indices and things would work, or it wouldn't create them and cleaner will be no-op).

are you able to run these integration tests locally?

yurishkuro:
I see this error, maybe you're not passing extra flags inside the integration test to actually generate sampling-related index?
```
        	            	listA:
        	            	([]string) (len=3) {
        	            	 (string) (len=34) "integration-test-jaeger-span-write",
        	            	 (string) (len=37) "integration-test-jaeger-service-write",
        	            	 (string) (len=42) "integration-test-jaeger-dependencies-write"
        	            	}
        	            	
        	            	
        	            	listB:
        	            	([]string) (len=4) {
        	            	 (string) (len=37) "integration-test-jaeger-service-write",
        	            	 (string) (len=34) "integration-test-jaeger-span-write",
        	            	 (string) (len=42) "integration-test-jaeger-dependencies-write",
        	            	 (string) (len=38) "integration-test-jaeger-sampling-write"
        	            	}
```

plugin/storage/es/samplingstore/storage.go
Pushkarm029: > Btw, in retSamples = append(retSamples, tToD.Throughput...), the ... doesn't make sense since you're appending a single element, not another slice.

tTod.Throughput is also a slice
```go
type TimeThroughput struct 
    Timestamp  time.Time    `json:"timestamp"`
    Throughput []Throughput `json:"throughputs"`
}
```
I agree that it may create extra copies, which may cause vulnerabilities. working on it

cmd/es-index-cleaner/app/index_filter.go
Pushkarm029: Yeah, we can tackle it in another PR.

Pushkarm029:
> are you able to run these integration tests locally?

YES, but sometimes it fails with a timeout, in that case, I rely on gh workflow

> I see this error, maybe you're not passing extra flags inside the integration test to actually generate sampling-related index?

POSSIBLE SOLUTION: Create a flag similar to archive, rollover, or useilm. and then use it in the integration test.

plugin/storage/es/samplingstore/storage.go
Pushkarm029: I think my current approach is wrong. Currently I am inserting throughputs like one timestamp-> array of throughputs,
```go
type TimeThroughput struct {
	Timestamp  time.Time    `json:"timestamp"`
	Throughput []Throughput `json:"throughputs"`
}
```
maybe we want something like this :
```go
type TimeThroughput struct {
	Timestamp  time.Time    `json:"timestamp"`
	Throughput Throughput   `json:"throughput"`
}
```

Pushkarm029:
🤞everything seems good now.

yurishkuro (approved):
🎉

PR #5162, opened 2024-02-04
Add unit tests for anonymizer/app/writer

h4shk4t:
- Part of [#5068](https://github.com/jaegertracing/jaeger/issues/5068)

## Description of the changes
- Added unit tests for `cmd/anonymizer/app/writer/`

- `go test -cover -v ./cmd/anonymizer/app/writer`
- `make lint test`

## Checklist


cmd/anonymizer/app/writer/writer_test.go
```diff
@@ -0,0 +1,148 @@
...
+	require.NoError(t, err)
+
+	for i := 0; i <= 99; i++ {
+		err = writer.WriteSpan(span)
+		require.NoError(t, err)
+	}
+}
+
+func TestWriter_WriteSpan_Exits(t *testing.T) {
+	if os.Getenv("BE_SUBPROCESS") == "1" {
```
yurishkuro, line 117: where would this be set?

cmd/anonymizer/app/writer/writer_test.go
```diff
@@ -0,0 +1,148 @@
...
+		writer, err := New(config, zap.NewNop())
+		require.NoError(t, err)
+
+		err = writer.WriteSpan(span)
+		require.NoError(t, err)
+
+		err = writer.WriteSpan(span)
+		if err == nil {
+			t.Fatal("expected an error but got none")
+		}
```
yurishkuro, line 135:
```suggestion
		require.Error(t, writer.WriteSpan(span))
```

cmd/anonymizer/app/writer/writer_test.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	cmd := exec.Command(os.Args[0], "-test.run=TestWriter_WriteSpan_Exits")
+	cmd.Env = append(os.Environ(), "BE_SUBPROCESS=1")
+	err := cmd.Run()
+	// The process should have exited with status 1, but exec.Command returns
+	// an *ExitError when the process exits with a non-zero status.
+	if err != nil {
+		t.Fatalf("process ran with err %v, want exit status 1", err)
+	}
```
yurishkuro, line 147:
```suggestion
	require.NoError(t, err)
```

cmd/anonymizer/app/writer/writer_test.go
```diff
@@ -0,0 +1,148 @@
...
+				MaxSpansCount:  10,
+				CapturedFile:   tempDir + "/captured.json",
+				AnonymizedFile: tempDir + "/anonymized.json",
+				MappingFile:    tempDir + "/mapping.json",
+			},
+		},
+		{
+			name: "captured.json doesn't exist",
+			config: Config{
+				CapturedFile:   tempDir + "/nonexistent_directory/captured.json",
```
yurishkuro, line 67: I don't understand - how is this different from the above case? There are no files in `tempDir` in each test.

cmd/anonymizer/app/writer/writer_test.go
```diff
@@ -0,0 +1,148 @@
...
+				AnonymizedFile: tempDir + "/nonexistent_directory/anonymized.json",
+				MappingFile:    tempDir + "/mapping.json",
+			},
+		},
+	}
+
+	t.Run(tests[0].name, func(t *testing.T) {
+		_, err := New(tests[0].config, nopLogger)
+		require.NoError(t, err)
+	})
```
yurishkuro, line 85: The point of doing data driven tests is so that you don't have to repeat code like this. You can either run them in a loop, or don't bother defining array in the first place and just initialize arguments right here, in the sub-tests (I personally prefer the latter as it makes it easier to understand the stacktrace, but when there's certain repeated logic for each test then data driven makes more sense).

cmd/anonymizer/app/writer/writer_test.go
h4shk4t: This was to increase the code coverage in the tests. The nonexistent directory doesn't exist inside the `tempDir` so it would force an error and increase the code coverage.

cmd/anonymizer/app/writer/writer_test.go
h4shk4t: Alright I will make the change.

cmd/anonymizer/app/writer/writer_test.go
h4shk4t: This is set on line 141. Actually the code I was testing here was exiting with `os.Exit(0)` so the tests were not running properly. My first thought was to look for tests dealing with `os.Exit()` in the codebase itself, however there were none. So I referred to an [article](https://sr-taj.medium.com/how-to-test-methods-that-kill-your-program-in-golang-e3b879185b8a) and found this implementation. If this is incorrect let me know I will look for other methods and fix it.

cmd/anonymizer/app/writer/writer_test.go
h4shk4t: Made the changes.

cmd/anonymizer/app/writer/writer_test.go
h4shk4t: Made the changes.

cmd/anonymizer/app/writer/writer_test.go
h4shk4t: Made the fixes

cmd/anonymizer/app/writer/writer_test.go
```diff
@@ -0,0 +1,124 @@
...
+	_, err := New(config, nopLogger)
+	require.NoError(t, err)
+
+	config = Config{
+		CapturedFile:   tempDir + "/nonexistent_directory/captured.json",
+		AnonymizedFile: tempDir + "/anonymized.json",
+		MappingFile:    tempDir + "/mapping.json",
+	}
+	_, err = New(config, nopLogger)
+	require.Error(t, err)
```
yurishkuro, line 66: Is there a specific error expected? I prefer `require.ErrorContains(t, err, "some identifying substring")`, provided that the substring is produced in our code, not in std library (which usually means we need to wrap the error via `fmt.Errorf("some text: %w", err)`)

cmd/anonymizer/app/writer/writer_test.go
```diff
@@ -0,0 +1,124 @@
...
+	tempDir := t.TempDir()
+
+	config := Config{
+		MaxSpansCount:  10,
+		CapturedFile:   tempDir + "/captured.json",
+		AnonymizedFile: tempDir + "/anonymized.json",
+		MappingFile:    tempDir + "/mapping.json",
+	}
+	_, err := New(config, nopLogger)
+	require.NoError(t, err)
```
yurishkuro, line 58: you could've done this:
```suggestion
t.Run("descriptive name", func(t *testing.T) {
	config := Config{
		MaxSpansCount:  10,
		CapturedFile:   tempDir + "/captured.json",
		AnonymizedFile: tempDir + "/anonymized.json",
		MappingFile:    tempDir + "/mapping.json",
	}
	_, err := New(config, nopLogger)
	require.NoError(t, err)
})
```

cmd/anonymizer/app/writer/writer_test.go
```diff
@@ -0,0 +1,124 @@
...
+		MaxSpansCount:  101,
+		CapturedFile:   tempDir + "/captured.json",
+		AnonymizedFile: tempDir + "/anonymized.json",
+		MappingFile:    tempDir + "/mapping.json",
+	}
+
+	writer, err := New(config, nopLogger)
+	require.NoError(t, err)
+
+	for i := 0; i <= 99; i++ {
```
yurishkuro, line 91: if you write just a few spans, are you going to get less signal than if writing a 100? There's no reason to make the test run longer than necessary

cmd/anonymizer/app/writer/writer_test.go
```diff
@@ -0,0 +1,124 @@
...
+
+	config := Config{
+		MaxSpansCount:  101,
+		CapturedFile:   tempDir + "/captured.json",
+		AnonymizedFile: tempDir + "/anonymized.json",
+		MappingFile:    tempDir + "/mapping.json",
+	}
+
+	writer, err := New(config, nopLogger)
+	require.NoError(t, err)
```
yurishkuro, line 89:
```suggestion
	require.NoError(t, err)
	defer writer.Close()
```

cmd/anonymizer/app/writer/writer_test.go
```diff
@@ -0,0 +1,124 @@
...
+			MappingFile:    tempDir + "/mapping.json",
+		}
+
+		writer, err := New(config, zap.NewNop())
+		require.NoError(t, err)
+
+		err = writer.WriteSpan(span)
+		require.NoError(t, err)
+
+		require.Error(t, writer.WriteSpan(span))
```
yurishkuro, line 113: why is the second write an error? Here it's good to add documentation in the form of a message passed to require.Error as 3rd arg

cmd/anonymizer/app/writer/writer_test.go
```diff
@@ -0,0 +1,124 @@
...
+
+		return
+	}
+
+	cmd := exec.Command(os.Args[0], "-test.run=TestWriter_WriteSpan_Exits")
+	cmd.Env = append(os.Environ(), "BE_SUBPROCESS=1")
+	err := cmd.Run()
+	// The process should have exited with status 1, but exec.Command returns
```
yurishkuro, line 121: > The process should have exited with status 1

why 1? The code does `os.Exit(0)`

cmd/anonymizer/app/writer/writer_test.go
```diff
@@ -0,0 +1,124 @@
...
+
+		err = writer.WriteSpan(span)
+		require.NoError(t, err)
+
+		require.Error(t, writer.WriteSpan(span))
+
+		return
+	}
+
+	cmd := exec.Command(os.Args[0], "-test.run=TestWriter_WriteSpan_Exits")
```
yurishkuro, line 118: I see what you meant. I would rather change the main code to make it more testable. Instead of calling exit, the function should return a predeclated public error, e.g.
```
var ErrMaxSpansCount = errors.New("Saved MaxSpansCount")
```
and the caller of Write() should check it and call os.Exit itself (or whatever else it needs to do).

cmd/anonymizer/app/writer/writer_test.go
h4shk4t: Made the fixes

cmd/anonymizer/app/writer/writer_test.go
h4shk4t: Made the fixes

cmd/anonymizer/app/writer/writer_test.go
h4shk4t: Made the proposed changes to `writer.go` and `writer_test.go`

cmd/anonymizer/app/writer/writer_test.go
h4shk4t: Got rid of the original implementation altogether. Now `WriteSpan` function returns an appropriate error in the given condition.

cmd/anonymizer/app/writer/writer_test.go
h4shk4t: Added a descriptive error message in the error and verifying in the test if the `ErrorContains` the same substring.

cmd/anonymizer/app/writer/writer_test.go
h4shk4t: Made the suggested changes

cmd/anonymizer/app/writer/writer.go
```diff
@@ -130,7 +131,7 @@ func (w *Writer) WriteSpan(msg *model.Span) error {
 	if w.config.MaxSpansCount > 0 && w.spanCount >= w.config.MaxSpansCount {
 		w.logger.Info("Saved enough spans, exiting...")
 		w.Close()
-		os.Exit(0)
+		return errors.New("saved MaxSpansCount")
```
yurishkuro, line 134: * this needs to be a constant that can be checked externally
* you need to make changes to main.go that calls Write so that it checks for the error code and does the exit

cmd/anonymizer/app/writer/writer_test.go
h4shk4t: Forgot to update here but made the changes.

cmd/anonymizer/main.go
```diff
@@ -53,7 +54,7 @@ func main() {
 				},
 			}
 
-			writer, err := writer.New(conf, logger)
+			writerObj, err := writer.New(conf, logger)
```
yurishkuro, line 57: if you don't like the name shadowing, then use `w`

cmd/anonymizer/main.go
```diff
@@ -69,9 +70,15 @@ func main() {
 			}
 
 			for _, span := range spans {
-				writer.WriteSpan(&span)
+				err := writerObj.WriteSpan(&span)
+				if err != nil {
```
yurishkuro, line 74:
```suggestion				
				if err := writerObj.WriteSpan(&span); err != nil {
```

cmd/anonymizer/main.go
```diff
@@ -69,9 +70,15 @@ func main() {
...
+					logger.Fatal("error while writing span", zap.Error(err))
```
yurishkuro, line 75: it's not fatal if it's `ErrMaxSpansCountReached`. Also, Fatal will exit so you won't ever reach the next line.
* move logging after `if error.Is` check
* use `logger.Error`, not `Fatal`

cmd/anonymizer/app/writer/writer_test.go
```diff
@@ -0,0 +1,117 @@
...
+		tempDir := t.TempDir()
+		config := Config{
+			MaxSpansCount:  1,
+			CapturedFile:   tempDir + "/captured.json",
+			AnonymizedFile: tempDir + "/anonymized.json",
+			MappingFile:    tempDir + "/mapping.json",
+		}
+
+		writer, err := New(config, zap.NewNop())
+		require.NoError(t, err)
```
yurishkuro, line 111: wouldn't this already be ErrMaxSpanCountReached?

h4shk4t: This is just creating a new Writer Object. It will give ErrMaxSpanCountReached when at least one span is written.

cmd/anonymizer/main.go
h4shk4t: Made the changes.

cmd/anonymizer/main.go
h4shk4t: Made the change.

cmd/anonymizer/main.go
h4shk4t: Made the changes.

cmd/anonymizer/app/writer/writer_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+		err = writer.WriteSpan(span)
+		require.ErrorContains(t, err, "max spans count reached")
```
yurishkuro, line 114: this should be ErrorIs check, the contract is that a specific error object is returned, not that it has a specific message in it (the code using it should not be checking error messages for decisions)

cmd/anonymizer/main.go
```diff
@@ -69,9 +70,14 @@ func main() {
 			}
 
 			for _, span := range spans {
-				writer.WriteSpan(&span)
+				if err := w.WriteSpan(&span); err != nil {
+					if errors.Is(err, writer.ErrMaxSpansCountReached) {
+						logger.Error("error while writing span", zap.Error(err))
+						os.Exit(0)
+					}
```
yurishkuro, line 77:
```suggestion
					if errors.Is(err, writer.ErrMaxSpansCountReached) {
						os.Exit(0)
					}
					logger.Error("error while writing span", zap.Error(err))
```

h4shk4t: In this case if ErrMaxSpansCountReached error is encountered, how would the error be logged? As `os.Exit(0)` would cause the program to stop.

yurishkuro: it's not an error, it does not need to be logged as such. If you want you can add a separate log statement inside if() with Info level saying "max span count processed"

h4shk4t: Okay, making the changes

cmd/anonymizer/app/writer/writer_test.go
h4shk4t: Made the changes

cmd/anonymizer/main.go
h4shk4t: Made the changes

cmd/anonymizer/app/writer/writer_test.go
```diff
@@ -0,0 +1,117 @@
...
+		}
+
+		writer, err := New(config, nopLogger)
+		require.NoError(t, err)
+
+		for i := 0; i < 9; i++ {
+			err = writer.WriteSpan(span)
+			require.NoError(t, err)
+		}
+		defer writer.Close()
```
yurishkuro, line 99: doing `defer` as the last line of a function is anti-pattern, how is that different from direct call? The correct pattern is to do defer right after the object was created, to guarantee it's closed regardless of how the rest of the function happens to run

h4shk4t: Oh yes... Fixed it and made the changes.

yurishkuro (approved):

yurishkuro:
Coverage 94.09% ==> 94.38%, nice! 🎉

PR #5171, opened 2024-02-06
[jaeger-v2] Streamline storage initialization

yurishkuro:
- There is duplicated code in the extension Start, which is being copy/pasted in other PRs

## Description of the changes
- Refactor Start to avoid duplication and to make testing simpler (no need to test all edge cases for every storage type)
- Clean-up tests

- 

## Checklist

### DIFF
--- a/cmd/jaeger/internal/extension/jaegerstorage/extension.go
+++ b/cmd/jaeger/internal/extension/jaegerstorage/extension.go
@@ -13,8 +13,10 @@ import (
 	"go.opentelemetry.io/collector/extension"
 	"go.uber.org/zap"
 
+	memoryCfg "github.com/jaegertracing/jaeger/pkg/memory/config"
 	"github.com/jaegertracing/jaeger/pkg/metrics"
 	"github.com/jaegertracing/jaeger/plugin/storage/badger"
+	badgerCfg "github.com/jaegertracing/jaeger/plugin/storage/badger"
 	"github.com/jaegertracing/jaeger/plugin/storage/memory"
 	"github.com/jaegertracing/jaeger/storage"
 )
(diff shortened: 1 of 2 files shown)


cmd/jaeger/internal/extension/jaegerstorage/extension.go
```diff
@@ -60,35 +62,62 @@ func newStorageExt(config *Config, otel component.TelemetrySettings) *storageExt
...
+	return nil
+}
 
-	// TODO add support for other backends
+func (s *storageExt) Start(ctx context.Context, host component.Host) error {
+	memStarter := &starter[memoryCfg.Configuration, *memory.Factory]{
+		ext:         s,
+		storageKind: "memory",
+		cfg:         s.config.Memory,
+		// memory factory does not return an error, so need to wrap it
```
albertteoh, line 95: Why not change `memory.NewFactoryWithConfig` to return the error? It doesn't look like it's used anywhere else, so it shouldn't be a large breaking change.

cmd/jaeger/internal/extension/jaegerstorage/extension.go
```diff
@@ -60,35 +62,62 @@ func newStorageExt(config *Config, otel component.TelemetrySettings) *storageExt
...
+	ext         *storageExt
+	storageKind string
+	cfg         map[string]Config
+	builder     func(Config, metrics.Factory, *zap.Logger) (Factory, error)
+}
 
-	for name, b := range s.config.Badger {
-		if _, ok := s.factories[name]; ok {
-			return fmt.Errorf("duplicate badger storage name %s", name)
+func (s *starter[Config, Factory]) build(ctx context.Context, host component.Host) error {
```
albertteoh, line 72: Internal state isn't persisted between method calls so all methods are effectively read-only; we could assert this by using value receivers instead of pointer receivers.

albertteoh (approved):

cmd/jaeger/internal/extension/jaegerstorage/extension.go
yurishkuro: Because I hate functions that are declared with errors but always return nil. The caller either has to ignore the error, or it has no easy way of unit testing the error handing path (which will never be used). Lose-lose. I think if the function can never return an error it should never declare it as a possibility (unless you're dealing with an interface where there's no choice).

cmd/jaeger/internal/extension/jaegerstorage/extension.go
yurishkuro: true, but makes no difference in this context since the code runs exactly once. Don't want to reset the approval just for this fix :-)

PR #5172, opened 2024-02-06
Prepare Release 1.54.0

joe-elliott:
Prepares Jaeger for release 1.54.0.

Includes jaeger-ui submodule updates.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,32 @@ run `make changelog` to generate content
 
 </details>
 
+1.54.0 (2024-02-06)
+-------------------
+
+#### ⛔ Breaking Changes:
+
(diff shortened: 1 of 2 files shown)


albertteoh (approved):

yurishkuro:
I think we should undo idl submodule change, this is not a part of normal release process.

I through it was already at latest:
```
bdd5ab5 - (HEAD, origin/main, main) Upgrade to opentelemetry-proto v1.1.0 (#104) (3 weeks ago) <Yuri Shkuro>
```

Since it's not, I would like to look into it separately and have a dedicated PR, not sneak it into a release update.

joe-elliott:
thx yuri! 

> Since it's not, I would like to look into it separately and have a dedicated PR, not sneak it into a release update.

I was following this step:

https://github.com/jaegertracing/jaeger/blob/main/RELEASE.md?plain=1#L8

Will submit a PR to clarify.

PR #5188, opened 2024-02-07
Release notes update

joe-elliott:
Suggested update to the release instructions to clarify that the jaeger-ui submodule only should be checked for updates.


yurishkuro (approved):
thx

Issue #5203, opened 2024-02-13
Extract archive storage integration test from ES and generalize to all storage backends


yurishkuro:
I just noticed, to my surprise, that archiving test is only implemented in the ES integration tests

https://github.com/jaegertracing/jaeger/blob/8ae2e05878a1ab38509e6678f6336b6947c9faba/plugin/storage/integration/elasticsearch_test.go#L295

## Objective
* Extract this test into the shared test suite
* Make sure it works with the backends that support archiving (which may be all of them at this point)

Wise-Wizard:
Should I create a new test file under integration itself, to test archiving for all Backends? Or can I just add the tests to the pre-existing test files for each storage?

yurishkuro:
we already have integration.go as the shared framework for all backends, it should go there

Wise-Wizard:
Hi, I believe the Archive Trace test functionality is only remaining for Cassandra and Badger after my PR. Should I continue working on them and close this issue?

Wise-Wizard:
@yurishkuro, may I pick up on this?

PR #5207, opened 2024-02-16
Move Archive test into shared integration test suite

Wise-Wizard:

This PR addresses the issue [#5203 ](https://github.com/jaegertracing/jaeger/issues/5203)

**Description of the changes**
This PR addresses the issue [#5203](https://github.com/jaegertracing/jaeger/issues/5203) by integrating the testArchiveTrace function into the common StorageIntegration type. This modification ensures compatibility with multiple storage backends, allowing the archiving test to be executed across different environments.

```bash
make test
```

**Checklist**


plugin/storage/integration/integration.go
```diff
@@ -121,6 +121,33 @@ func (s *StorageIntegration) waitForCondition(t *testing.T, predicate func(t *te
...
+	expected := &model.Span{
+		OperationName: "archive_span",
+		StartTime:     time.Now().Add(-time.Hour * 72 * 5),
+		TraceID:       tID,
+		SpanID:        model.NewSpanID(55),
+		References:    []model.SpanRef{},
+		Process:       model.NewProcess("archived_service", model.KeyValues{}),
+	}
+
+	require.NoError(t, s.SpanWriter.WriteSpan(context.Background(), expected))
```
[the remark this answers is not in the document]
Wise-Wizard: Understood, will make the changes

plugin/storage/integration/integration.go
```diff
@@ -121,6 +121,33 @@ func (s *StorageIntegration) waitForCondition(t *testing.T, predicate func(t *te
...
+	s.Refresh()
+
+	var actual *model.Trace
+	found := s.waitForCondition(t, func(t *testing.T) bool {
+		var err error
+		actual, err = s.SpanReader.GetTrace(context.Background(), tID)
```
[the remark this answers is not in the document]
Wise-Wizard: Noted

plugin/storage/integration/integration.go
```diff
@@ -121,6 +121,33 @@ func (s *StorageIntegration) waitForCondition(t *testing.T, predicate func(t *te
 	return predicate(t)
 }
 
+func (s *StorageIntegration) testArchiveTrace(t *testing.T) {
```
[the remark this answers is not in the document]
Wise-Wizard: Pardon me, I didn't quite understand you here

plugin/storage/integration/integration.go
```diff
@@ -480,6 +507,7 @@ func (s *StorageIntegration) insertThroughput(t *testing.T) {
 
 // IntegrationTestAll runs all integration tests
 func (s *StorageIntegration) IntegrationTestAll(t *testing.T) {
+	t.Run("ArchiveTrace", s.testArchiveTrace)
```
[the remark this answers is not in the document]
Wise-Wizard: Oh okay, will look into adding the necessary conditions..
The integration test for Archive did not fail though in the GitHub actions?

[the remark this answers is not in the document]
yurishkuro: because it didn't use archive reader/writer

plugin/storage/integration/integration.go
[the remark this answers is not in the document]
Wise-Wizard: Hi @yurishkuro , I am having some difficulties and confusion in integrating the archive writer. After looking into archive.go and grpc_client.go. I assume I have to instantiate the archive Reader/Writer 
```bash
archiveReaderClient storage_v1.ArchiveSpanReaderPluginClient
archiveWriterClient storage_v1.ArchiveSpanWriterPluginClient
```
which is in turn asking for a gRPC connection type while creating an instance in the Archive Test. Is there something I am doing wrong or missing?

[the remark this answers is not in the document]
yurishkuro: I think you're looking at the wrong API, you need this: https://github.com/jaegertracing/jaeger/blob/5375e1e854336e66fb0862f4f7bcc75437579c3c/storage/factory.go#L70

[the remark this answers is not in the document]
Wise-Wizard: Thanks a lot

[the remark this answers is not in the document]
Wise-Wizard: Hi, can you please confirm if I have implemented the archive reader and writer correctly?

plugin/storage/integration/integration.go
```diff
@@ -121,6 +125,54 @@ func (s *StorageIntegration) waitForCondition(t *testing.T, predicate func(t *te
 	return predicate(t)
 }
 
+func (s *StorageIntegration) InitArchiveStorage(storageFactory storage.Factory, logger *zap.Logger) bool {
```
yurishkuro, line 128: who calls it?

plugin/storage/integration/integration.go
```diff
@@ -121,6 +125,54 @@ func (s *StorageIntegration) waitForCondition(t *testing.T, predicate func(t *te
...
+	archiveFactory, ok := storageFactory.(storage.ArchiveFactory)
+	if !ok {
+		logger.Info("Archive storage not supported by the factory")
+		return false
+	}
+	reader, err := archiveFactory.CreateArchiveSpanReader()
+	if err != nil {
+		logger.Error("Cannot init archive storage reader", zap.Error(err))
+		return false
+	}
```
yurishkuro, line 138: pass `t` to this function and just do `require.NoError(t, err)`. It's ok to fail (require will panic) because if the storage implements archive factory per above, there's no reason for it to not return a reader, it's no longer a soft error.

yurishkuro (commented):
there should not be submodule changes in this PR

Wise-Wizard:
> there should not be submodule changes in this PR

Yeah, actually before making any changes at all. In the main branch after executing git pull, the submodule showed M (Modified) icon automatically. So, I thought those changes were coming from git pull itself?

plugin/storage/integration/integration.go
Wise-Wizard: Done 👍

plugin/storage/integration/integration.go
Wise-Wizard: Done! PTAL.

yurishkuro:
> there should not be submodule changes in this PR

You can do this:
```shell
git checkout main
git pull 
git submodule update --recursive
git checkout pr-branch
git commit -sm 'fix submodules'
```
Your `main` branch must be up to date with upstream.

plugin/storage/integration/integration.go
```diff
@@ -121,6 +125,52 @@ func (s *StorageIntegration) waitForCondition(t *testing.T, predicate func(t *te
...
+func (s *StorageIntegration) InitArchiveStorage(t *testing.T, storageFactory storage.Factory, logger *zap.Logger) bool {
+	archiveFactory, ok := storageFactory.(storage.ArchiveFactory)
+	if !ok {
+		logger.Info("Archive storage not supported by the factory")
+		return false
+	}
+	reader, errReader := archiveFactory.CreateArchiveSpanReader()
+	require.NoError(t, errReader)
+	writer, errWriter := archiveFactory.CreateArchiveSpanWriter()
+	require.NoError(t, errWriter)
```
yurishkuro, line 137: no need to name errors
```suggestion
	reader, err := archiveFactory.CreateArchiveSpanReader()
	require.NoError(t, err)
	writer, err := archiveFactory.CreateArchiveSpanWriter()
	require.NoError(t, err)
```

plugin/storage/integration/integration.go
```diff
@@ -121,6 +125,52 @@ func (s *StorageIntegration) waitForCondition(t *testing.T, predicate func(t *te
...
+	writer, err := archiveFactory.CreateArchiveSpanWriter()
+	require.NoError(t, err)
+	s.ArchiveSpanReader = reader
+	s.ArchiveSpanWriter = writer
+	return true
+}
+
+func (s *StorageIntegration) testArchiveTrace(t *testing.T) {
+	defer s.CleanUp()
+	success := s.InitArchiveStorage(t, s.storageFactory, zap.NewNop())
```
yurishkuro, line 145: None of the other testSomething tests are calling init storage. Please use the same initialization pattern.

yurishkuro (commented):
where is the "extract" part of this PR? The ES storage integration test still contains archive logic.

Wise-Wizard:
> where is the "extract" part of this PR? The ES storage integration test still contains archive logic.

Yes, I am yet to remove the archive logic from ES Storage Integration. Thought will do that after the logic is approved.

plugin/storage/integration/integration.go
Wise-Wizard: Done! PTAL.

yurishkuro: Don't see much change. You're introducing storageFactory member field that is never set, that makes all archive tests to be skipped.

Wise-Wizard: Yes while R&D, to create an ArchiveSpan Reader/Writer it needed an archiveFactory initialization that is why introduced the storageFactory member field. Had a hunch that all archive tests will be skipped due to this, wanted to run it through by you once.

Wise-Wizard: I have my college Mid-Semester exams starting from today. Can I resume working on this around next week please?

Wise-Wizard:
Hi, @yurishkuro. I had a query regarding this issue. You mentioned that the archive span reader/writer is not being ued. But, in the existing code of elasticsearch_test.go, the archive trace test is using the same Span Reader/Writer received from integration.go field. So, is that wrong itself?
Screenshot 2024-03-21 213816

yurishkuro:
> Why can't I just extract this to integration.go, and add a condition to check if the specific storage supports archiving and then only execute the test?

That was exactly the objective of the ticket, but it's not what your PR is doing. "extract" means "move", not duplicate.

Wise-Wizard:
> > Why can't I just extract this to integration.go, and add a condition to check if the specific storage supports archiving and then only execute the test?
> 
> That was exactly the objective of the ticket, but it's not what your PR is doing. "extract" means "move", not duplicate.

Yes, I do know that clearly. Like I said, I was going to remove it from the elasticsearch_test after I have cloned it succesfully in Integration.go. And, in my previous commit I had done the exact same thing but I believe there was some kind of misunderstanding and you mentioned that I needed to create an Archive Span Reader/Writer seperately.

plugin/storage/integration/integration.go
```diff
@@ -481,6 +507,7 @@ func (s *StorageIntegration) insertThroughput(t *testing.T) {
 // IntegrationTestAll runs all integration tests
 func (s *StorageIntegration) IntegrationTestAll(t *testing.T) {
 	t.Run("GetServices", s.testGetServices)
+	t.Run("ArchiveTrace", s.testArchiveTrace)
```
yurishkuro, line 510: so there's nothing in this function that makes it "archive" - it's just using the main writer/reader. In some sense this is not bad conceptually (I even have another ticket to deprecate "archive" APIs), but when the storage has specific behavior only for archive mode your test is not going to exercise that behavior. Specifically, although the `testArchiveTrace` in ES looked like it also uses primary reader/writer, those were actually initialized with an archive=true flag (see L211 in plugin/storage/es/spanstore/reader.go), unlike the rest of the tests.

Wise-Wizard: Ok, thanks a lot for the clarification. I went through the ES Test and did notice the archive bool value initializing ES. I just had one more query, what about the other backends like Cassandra, Badger etc. Because, they aren't initialized in the same manner. They aren't initialized with the archive bool value, so how do I get them to exercise the archive mode behaviour?

yurishkuro: Right now the shared tests run by assuming that SpanReader etc. fields in the struct are set, and it's up to individual storage implementations to initialize those. What we can do is add ArchiveSpanReader/etc to the main struct and have the implementations initialize them as they wish. This way ES can create separate reader/write with archive=true, while other implementations can just use the same reader/writer since they don't have functional differences.

Wise-Wizard: Ok got it, I did initialize Archive Span Reader/Writer i the main struct earlier, but writing the test required initialization of a storage factory, will revert back to the change and work on it. Thanks!

plugin/storage/integration/integration.go
```diff
@@ -144,6 +144,32 @@ func (s *StorageIntegration) testGetServices(t *testing.T) {
...
+	expected := &model.Span{
+		OperationName: "archive_span",
+		StartTime:     time.Now().Add(-time.Hour * 72 * 5),
+		TraceID:       tID,
+		SpanID:        model.NewSpanID(55),
+		References:    []model.SpanRef{},
+		Process:       model.NewProcess("archived_service", model.KeyValues{}),
+	}
+
+	require.NoError(t, s.SpanWriter.WriteSpan(context.Background(), expected))
```
yurishkuro, line 159: I suggest changing this to use `s.ArchiveSpanWriter` (to be initialized by the actual implementation), and to skip the test if that writer is nil

Wise-Wizard: Yes, Understood that. Already made that change.

Pushkarm029:
Cassandra e2e is failing due to config error
```bash
=== RUN   TestCassandraStorage
    cassandra_test.go:134: 
        	Error Trace:	/home/runner/work/jaeger/jaeger/plugin/storage/integration/cassandra_test.go:134
        	Error:      	Received unexpected error:
make: *** [Makefile:120: storage-integration-test] Error 1
        	            	archive storage not configured
        	Test:       	TestCassandraStorage
--- ❌ FAIL: TestCassandraStorage (0.19s)
```

Wise-Wizard:
> Cassandra e2e is failing due to config error
> 
> ```shell
> === RUN   TestCassandraStorage
>     cassandra_test.go:134: 
>         	Error Trace:	/home/runner/work/jaeger/jaeger/plugin/storage/integration/cassandra_test.go:134
>         	Error:      	Received unexpected error:
> make: *** [Makefile:120: storage-integration-test] Error 1
>         	            	archive storage not configured
>         	Test:       	TestCassandraStorage
> --- ❌ FAIL: TestCassandraStorage (0.19s)
> ```

Yes, I know that so now for each backend supporting archiving, I should configure the archive storage separately in each test right?

plugin/storage/integration/badgerstore_test.go
```diff
@@ -49,6 +48,7 @@ func (s *BadgerIntegrationStorage) initialize() error {
 	if err != nil {
 		return err
 	}
+
```
Pushkarm029, line 51: remove this unnecessary change

plugin/storage/integration/memstore_test.go
```diff
@@ -39,6 +39,8 @@ func (s *MemStorageIntegrationTestSuite) initialize() error {
 	s.SamplingStore = memory.NewSamplingStore(2)
 	s.SpanReader = store
 	s.SpanWriter = store
+	s.ArchiveSpanReader = store
+	s.ArchiveSpanWriter = store
```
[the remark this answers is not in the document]
Wise-Wizard: Noted

plugin/storage/integration/integration.go
```diff
@@ -144,6 +146,34 @@ func (s *StorageIntegration) testGetServices(t *testing.T) {
...
+	tID := model.NewTraceID(uint64(11), uint64(22))
+	expected := &model.Span{
+		OperationName: "archive_span",
+		StartTime:     time.Now().Add(-time.Hour * 72 * 5),
+		TraceID:       tID,
+		SpanID:        model.NewSpanID(55),
+		References:    []model.SpanRef{},
+		Process:       model.NewProcess("archived_service", model.KeyValues{}),
+	}
+	if s.ArchiveSpanReader == nil || s.ArchiveSpanWriter == nil {
```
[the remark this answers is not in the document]
Wise-Wizard: Done, PTAL!

plugin/storage/integration/integration.go
```diff
@@ -144,6 +146,34 @@ func (s *StorageIntegration) testGetServices(t *testing.T) {
...
+
+	var actual *model.Trace
+	s.waitForCondition(t, func(t *testing.T) bool {
+		var err error
+		actual, err = s.ArchiveSpanReader.GetTrace(context.Background(), tID)
+		return err == nil && len(actual.Spans) == 1
+	})
+	// if !assert.True(t, found) {
+	// 	CompareTraces(t, &model.Trace{Spans: []*model.Span{expected}}, actual)
+	// }
```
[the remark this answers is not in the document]
Wise-Wizard: CodeCov is not covering this. Can I move this test directly inside the s.waitForCondition function instead?

plugin/storage/integration/elasticsearch_test.go
```diff
@@ -126,6 +125,7 @@ func (s *ESStorageIntegration) initializeES(t *testing.T, allTagsAsFields, archi
 	s.esCleanUp(t, allTagsAsFields, archive)
 	// TODO: remove this flag after ES support returning spanKind when get operations
 	s.GetOperationsMissingSpanKind = true
+	s.SkipArchiveTest = !archive
```
yurishkuro, line 128: I don't think this `archive` flag makes sense anymore - the archive test is always included in the harness, so it should not be conditional that ES test initializes archive reader/writer, it always needs to create both primary and archive namespaces

plugin/storage/integration/integration.go
```diff
@@ -144,6 +150,34 @@ func (s *StorageIntegration) testGetServices(t *testing.T) {
...
+
+	var actual *model.Trace
+	found := s.waitForCondition(t, func(t *testing.T) bool {
+		var err error
+		actual, err = s.ArchiveSpanReader.GetTrace(context.Background(), tID)
+		return err == nil && len(actual.Spans) == 1
+	})
+	if !assert.True(t, found) {
+		CompareTraces(t, &model.Trace{Spans: []*model.Span{expected}}, actual)
+	}
```
yurishkuro, line 178: I suggest changing this to
```
require.True(t, found)
CompareTraces(...)
```
there's no need for the latter to be conditional, since require would abort the test anyway

Wise-Wizard: Noted

plugin/storage/integration/elasticsearch_test.go
Wise-Wizard: Without this flag, the ES test was failing.
In the logs it was written as a nil pointer deference error.
I assumed it was coming from the SkipArchiveTrace itself, and after initialising it to !archive, the tests were successfully passing

plugin/storage/integration/integration.go
Wise-Wizard: The ES tests are failing after this change. Even if found is false, CompareTraces should only run if the found value is false right?

yurishkuro: if the value is not found the test failed

Wise-Wizard: Yes, the test should be skipped instead and compare traces shouldn't run, yes?

yurishkuro: no the test should. In the previous test `assert.True` was always called, so if that assert fails the whole test fails. It was just used to guard comparison which would create extra error noise if nothing was found. But if it's changed to require() then it simply doesn't reach that line. The test fails in either case.

plugin/storage/integration/integration.go
```diff
@@ -144,6 +150,33 @@ func (s *StorageIntegration) testGetServices(t *testing.T) {
...
+	tID := model.NewTraceID(uint64(11), uint64(22))
+	expected := &model.Span{
+		OperationName: "archive_span",
+		StartTime:     time.Now().Add(-time.Hour * 72 * 5),
+		TraceID:       tID,
+		SpanID:        model.NewSpanID(55),
+		References:    []model.SpanRef{},
+		Process:       model.NewProcess("archived_service", model.KeyValues{}),
+	}
+	if s.SkipArchiveTest {
```
yurishkuro, line 164: should be second line in the function

plugin/storage/integration/integration.go
```diff
@@ -144,6 +150,33 @@ func (s *StorageIntegration) testGetServices(t *testing.T) {
 	}
 }
 
+func (s *StorageIntegration) testArchiveTrace(t *testing.T) {
+	defer s.CleanUp()
```
yurishkuro, line 154:
```suggestion
	s.skipIfNeeded(t)
	defer s.cleanUp(t)
```

plugin/storage/integration/elasticsearch_test.go
```diff
@@ -259,8 +258,12 @@ func testElasticsearchStorage(t *testing.T, allTagsAsFields, archive bool) {
 	s.Fixtures = LoadAndParseQueryTestCases(t, "fixtures/queries_es.json")
 
 	if archive {
```
yurishkuro, line 260: * this flag should be deleted from tests
* TestElasticsearchStorage_Archive() should be deleted
* we only need to call IntegrationTestAll(), not manually call testArchiveTrace

Wise-Wizard: Done! PTAL

plugin/storage/integration/elasticsearch_test.go
```diff
@@ -206,7 +205,6 @@ func (s *ESStorageIntegration) initSpanstore(t *testing.T, allTagsAsFields, arch
 		IndexPrefix:       indexPrefix,
 		MaxSpanAge:        maxSpanAge,
 		TagDotReplacement: tagKeyDeDotChar,
-		Archive:           archive,
```
yurishkuro, line 209: so where do you initialize ArchiveSpanReader now?

Wise-Wizard: Yes, I apologise for my absent-mindedness 😅. Made the changes now.

Wise-Wizard: Hi, I tried looking into it, but why is the ES expected trace and actual trace differing? Is it because of the way I have initialized the Archive Reader/Writer and there is a specific way to do so?

plugin/storage/integration/elasticsearch_test.go
```diff
@@ -184,6 +184,7 @@ func (s *ESStorageIntegration) initSpanstore(t *testing.T, allTagsAsFields, arch
 	require.NoError(t, err)
 	clientFn := func() estemplate.Client { return client }
 
+	// Initializing Span Reader and Writer
```
yurishkuro, line 187: thinking further through this, it's weird why we init the reader/writer manually instead of using es.Factory. But it's for another round of PRs.

plugin/storage/integration/integration.go
```diff
@@ -40,27 +40,34 @@ import (
 )
 
 const (
-	iterations = 100
+	iterations  = 100
+	maxSpanTime = 72 * time.Hour
```
yurishkuro, line 44:
```suggestion
	maxSpanAge = 72 * time.Hour
```

plugin/storage/integration/integration.go
```diff
@@ -144,6 +151,35 @@ func (s *StorageIntegration) testGetServices(t *testing.T) {
...
+		OperationName: "archive_span",
+		StartTime:     time.Now().Add(-maxSpanTime * 5),
+		TraceID:       tID,
+		SpanID:        model.NewSpanID(55),
+		References:    []model.SpanRef{},
+		Process:       model.NewProcess("archived_service", model.KeyValues{}),
+	}
+
+	require.NoError(t, s.ArchiveSpanWriter.WriteSpan(context.Background(), expected))
+	s.Refresh()
```
yurishkuro, line 171:
```suggestion
	s.refresh(t)
```
match all all other tests are doing

plugin/storage/integration/elasticsearch_test.go
Wise-Wizard: Yes exactly, I had the same doubt. Do you want me to initialize it using es.Factory instead?

plugin/storage/integration/integration.go
Wise-Wizard: Done

Wise-Wizard: Initially it was directly defined itself, but in es_test it was defined as constant, that's why changed it to constant here as well

plugin/storage/integration/elasticsearch_test.go
yurishkuro: I'd say more important is to actually enable archive tests in other storages, that was the main point of this ticket, but right now it's all disabled except for ES.

yurishkuro (approved):

plugin/storage/integration/elasticsearch_test.go
yurishkuro: ah, good. There's no reason why Badger cannot support it, just point badger to another directory for data. Cassandra certainly supports it. Kafka isn't really a "storage" so yeah, it's ok to skip there.

Wise-Wizard: Yes, I didn't knew there was a workaround for badger just noticed it did not have CreateArchiveReader method and Stack Overflow said it didn't support archiving, so had skipped Badger. As for Cassandra, I was working on it but threw some error, so had removed the archive method, initially and would work on it after ES was fixed.

Wise-Wizard: What can I work on next, or should continue on this itself?

yurishkuro: up to you. May be worth finishing the ES by cleaning it up to make it use Factory rather than instantiate things manually.

Wise-Wizard: On it!

PR #5209, opened 2024-02-18
Return more detailed errors from ES storage

yurishkuro:
- In many cases when ES operations fail, the only error returned to the UI or written to Jaeger log is `all shards fail`, which is pretty useless for the actual troubleshooting.
- Meanwhile, the driver actually returns an error struct that contains a `RootCause` field.

## Description of the changes
- Inspect the error and include root cause if present

Run all-in-one and `curl 'http://localhost:16686/api/services'`

Before:
`$ curl 'http://localhost:16686/api/services'`
`{"data":null,"total":0,"limit":0,"offset":0,"errors":[{"code":500,"msg":"search services failed: elastic: Error 400 (Bad Request): all shards failed [type=search_phase_execution_exception]"}]}`

After:
`$ curl 'http://localhost:16686/api/services'`
`{"data":null,"total":0,"limit":0,"offset":0,"errors":[{"code":500,"msg":"search services failed: elastic: Error 400 (Bad Request): all shards failed [type=search_phase_execution_exception]: RootCause[Fielddata is disabled on [serviceName] in [jaeger-service-]. Text fields are not optimised for operations that require per-document field data like aggregations and sorting, so these operations are disabled by default. Please use a keyword field instead. Alternatively, set fielddata=true on [serviceName] in order to load field data by uninverting the inverted index. Note that this can use significant memory. [type=illegal_argument_exception]]"}]}`

### DIFF
--- a/pkg/es/errors.go
+++ b/pkg/es/errors.go
@@ -1,0 +1,33 @@
+// Copyright (c) 2024 The Jaeger Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package es
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/olivere/elastic"
+)
+
+// DetailedError creates a more detailed error if the error stack contains elastic.Error.
+// This is useful because by default olivere/elastic returns errors that print like this:
+//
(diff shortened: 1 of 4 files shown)


jkowall (approved):
LGTM!

PR #5216, opened 2024-02-20
Fix groutine leaks in cmd/anonymizer/app/writer

akagami-harsh:
- part of #5006

## Description of the changes
- Fix groutine leaks in cmd/anonymizer/app/writer

- `go test`

## Checklist

### DIFF
--- a/cmd/anonymizer/app/anonymizer/anonymizer.go
+++ b/cmd/anonymizer/app/anonymizer/anonymizer.go
@@ -15,6 +15,7 @@
 package anonymizer
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"hash/fnv"
@@ -57,6 +58,8 @@ type Anonymizer struct {
 	lock        sync.Mutex
 	mapping     mapping
 	options     Options
+	ctx         context.Context
+	cancel      context.CancelFunc
 }
 
 // Options represents the various options with which the anonymizer can be configured.
@@ -70,6 +73,7 @@ type Options struct {
 // New creates new Anonymizer. The mappingFile stores the mapping from original to
 // obfuscated strings, in case later investigations require looking at the original traces.
 func New(mappingFile string, options Options, logger *zap.Logger) *Anonymizer {
+	ctx, cancel := context.WithCancel(context.Background())
 	a := &Anonymizer{
 		mappingFile: mappingFile,
 		logger:      logger,
@@ -78,6 +82,8 @@ func New(mappingFile string, options Options, logger *zap.Logger) *Anonymizer {
 			Operations: make(map[string]string),
 		},
 		options: options,
+		ctx:     ctx,
+		cancel:  cancel,
 	}
 	if _, err := os.Stat(filepath.Clean(mappingFile)); err == nil {
 		dat, err := os.ReadFile(filepath.Clean(mappingFile))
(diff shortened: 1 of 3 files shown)


cmd/anonymizer/app/writer/writer_test.go
```diff
@@ -53,8 +53,9 @@ func TestNew(t *testing.T) {
 			AnonymizedFile: tempDir + "/anonymized.json",
 			MappingFile:    tempDir + "/mapping.json",
 		}
-		_, err := New(config, nopLogger)
+		writer, err := New(config, nopLogger)
 		require.NoError(t, err)
+		defer writer.anonymizer.Stop()
```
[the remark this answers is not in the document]
akagami-harsh: added it in existing `close()` method

cmd/anonymizer/app/anonymizer/anonymizer.go
```diff
@@ -89,13 +95,24 @@ func New(mappingFile string, options Options, logger *zap.Logger) *Anonymizer {
...
+				a.SaveMapping()
+			case <-a.ctx.Done():
+				return
+			}
 		}
 	}()
 	return a
 }
 
+func (a *Anonymizer) Stop() {
```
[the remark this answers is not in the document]
akagami-harsh: done!

cmd/anonymizer/app/writer/writer.go
```diff
@@ -146,4 +146,5 @@ func (w *Writer) Close() {
 	w.anonymizedFile.WriteString("\n]\n")
 	w.anonymizedFile.Close()
 	w.anonymizer.SaveMapping()
+	w.anonymizer.Stop()
```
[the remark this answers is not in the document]
akagami-harsh: moved `Stop()` method above `SaveMapping()` and also added a wait group

cmd/anonymizer/app/anonymizer/anonymizer.go
```diff
@@ -88,14 +95,31 @@ func New(mappingFile string, options Options, logger *zap.Logger) *Anonymizer {
...
+				a.SaveMapping()
+			case <-a.ctx.Done():
+				return
+			}
 		}
 	}()
 	return a
 }
 
+func (a *Anonymizer) WaitForStop() {
```
yurishkuro, line 115: I don't think this needs to be a public function, you just need to call Wait after cancel() in Stop()

yurishkuro: this way you only return from Stop() once we confirmed that all background work has stopped (even though goroutine itself might linger before exiting)

akagami-harsh: Done!, `Stop()` now calls `wg.Wait()` after cancelling the context to block until the background goroutine exits

cmd/anonymizer/app/anonymizer/anonymizer.go
```diff
@@ -78,6 +83,8 @@ func New(mappingFile string, options Options, logger *zap.Logger) *Anonymizer {
 			Operations: make(map[string]string),
 		},
 		options: options,
+		ctx:     ctx,
```
yurishkuro, line 86: nit: didn't need to capture ctx in the struct since you're only using it in the same function

akagami-harsh: removed it from the struct

yurishkuro (approved):
Thanks!

PR #5189, opened 2024-02-07
Added Feature Support  for External URL prefix

varshith257:
## Which problem is this PR solving?
- 
Resolves #5157 
## Description of the changes
 The changes include:

- Added a new field` UIBasePath` to the `StaticAssetsHandlerOptions` struct.
- I created a new `assetsFS` instance using the `UIBasePath` field if it is non-empty.
- Replaced the `BasePath ` with the `UIBasePath` field.

## How was this change tested?
-  Done

## Checklist
- [ ] I have read https://github.com/jaegertracing/jaeger/blob/master/CONTRIBUTING_GUIDELINES.md
- [ ] I have signed all commits
- [ ] I have added unit tests for the new functionality
- [ ] I have run lint and test steps successfully
  - for `jaeger`: `make lint test`
  - for `jaeger-ui`: `yarn lint` and `yarn test`

PR #5219, opened 2024-02-21
Feature Implementation for External URL prefix

### LINKED ISSUE #5157: [Feature]: Support external prefix
### Requirement

There are 2 prefixes, local (current implementation) and external (to be used from the UI). Jaeger supports the local prefix and does not support external prefix currently.
Please check the below discussion for more details.
https://github.com/orgs/jaegertracing/discussions/5042#discussioncomment-8236479

### Problem

The traffic flow is gateway>virtual service>gateway>jaeger query.

We are setting the query_base_path as "/jaeger/viewer" and then doing the rewrite using the virtual service configuration mentioned below.

    match:
        uri:
        prefix: /jaeger/viewer
        rewrite:
        uri: /hub/jaeger-query/v1/jaeger/viewer
        retries:
        attempts: 3
        perTryTimeout: 15s
        retryOn: connect-failure,refused-stream,reset,unavailable,cancelled,resource-exhausted
        route:
        destination:
        host: api-gateway
        port:
        number: 8080
        timeout: 15s

We are getting issue in the Jaeger UI.The index.html is getting loaded but the static files like index.js, index.css are showing loading error(404 not found) in the console tab of the browser.

### Proposal

_No response_

### Open questions

_No response_

Sidhartha-Biswal:
Resolves #5157 

## Description of the changes
-     Implemented code for query.ui-base-path CLI flag.
-     Added a new field UIBasePath to the StaticAssetsHandlerOptions struct.
-     Created a new assetsFS instance using the UIBasePath field if it is non-empty.

-     Tested manually and it did not work. So need some suggestions to proceed ahead.


Sidhartha-Biswal:
Hi @yurishkuro,

Please find the below configuration.
QUERY_BASE_PATH: /jaeger
QUERY_UI_BASE_PATH: /hub/jaeger

The static files loading error is still there.

Could you please check the code changes and give some suggestions to proceed ahead?

yurishkuro:
How is this different from the other PR?

Sidhartha-Biswal:
Hi @yurishkuro,

We have tried setting the QUERY_UI_BASE_PATH="/hub/jaeger" and tried to access the Jaeger UI. But still we found that static assets are not getting loaded.  
With the changes we have tried, the QUERY_UI_BASE_PATH value is reflecting in the <base href="/hub/jaeger"> in the index.html. This way the changes are different from the other PR.

manisha5464:
Hi @yurishkuro 
Can you please describe what is the relation between QUERY_BASE_PATH and QUERY_UI_BASE_PATH?
Can we set both the parameters at the same time?
Like for example QUERY_BASE_PATH: /jaeger and QUERY_UI_BASE_PATH: /hub/jaeger, then which path should take the priority?
Which path should reflect in index.html?

I was also facing a similar issue, unable to load the static assets in Jaeger UI when both the parameters are set. I was trying with the code changes in this PR and also with those mentioned in this [https://github.com/jaegertracing/jaeger/pull/5189]. Unfortunately still the issue exists and not able to load the static assets.

Can you please provide some suggestions that would help to proceed further.

Sidhartha-Biswal:
Hi @yurishkuro,

We have tried adding  the QUERY_UI_BASE_PATH parameter.
We tested the latest code changes by setting the QUERY_UI_BASE_PATH ="/hub/jaeger" and QUERY_BASE_PATH="/jaeger"
With the latest changes pushed we see the below log in Query Service. The QUERY_UI_BASE_PATH is not getting appended to static folder. Therefore static files are not loaded into Jaeger UI.

staticlog

Expected log entry should have url entry like below
"url": "/hub/jaeger/static/index.css" 
 
Below is the code change I have tried for appending the UIBasePath to static folder, but still we see the issue.
 
staticasset

Can you please review the latest changes and let us know where we are missing.

cmd/query/app/static_handler.go
```diff
@@ -139,11 +140,21 @@ func (sH *StaticAssetsHandler) loadAndEnrichIndexHTML(open func(string) (http.Fi
 	if sH.options.BasePath == "" {
 		sH.options.BasePath = "/"
 	}
+	if sH.options.UIBasePath == "" {
+		sH.options.UIBasePath = "/ui" // default value
```
yurishkuro, line 144: It should default to BasePath

cmd/query/app/static_handler.go
```diff
@@ -139,11 +140,21 @@ func (sH *StaticAssetsHandler) loadAndEnrichIndexHTML(open func(string) (http.Fi
...
+	}
 	if sH.options.BasePath != "/" {
```
yurishkuro, line 146: BasePath no longer relevant to indexBytes, only ui path

cmd/query/app/static_handler.go
```diff
@@ -139,11 +140,21 @@ func (sH *StaticAssetsHandler) loadAndEnrichIndexHTML(open func(string) (http.Fi
...
+		if sH.options.UIBasePath != "/ui" {
+			if !strings.HasPrefix(sH.options.UIBasePath, "/") || strings.HasSuffix(sH.options.UIBasePath, "/") {
+				return nil, fmt.Errorf("invalid base path '%s'. Must start but not end with a slash '/', e.g. '/jaeger/ui'", sH.options.BasePath)
+			}
+			indexBytes = basePathPattern.ReplaceAll(indexBytes, []byte(fmt.Sprintf(`<base href="%s/"`, sH.options.UIBasePath)))
+		} else {
+			if !strings.HasPrefix(sH.options.BasePath, "/") || strings.HasSuffix(sH.options.BasePath, "/") {
+				return nil, fmt.Errorf("invalid base path '%s'. Must start but not end with a slash '/', e.g. '/jaeger/ui'", sH.options.BasePath)
+			}
+			indexBytes = basePathPattern.ReplaceAll(indexBytes, []byte(fmt.Sprintf(`<base href="%s/"`, sH.options.BasePath)))
```
yurishkuro, line 156:
```suggestion
```

cmd/query/app/static_handler.go
```diff
@@ -139,11 +140,21 @@ func (sH *StaticAssetsHandler) loadAndEnrichIndexHTML(open func(string) (http.Fi
 	if sH.options.BasePath == "" {
 		sH.options.BasePath = "/"
 	}
+	if sH.options.UIBasePath == "" {
+		sH.options.UIBasePath = "/ui" // default value
+	}
 	if sH.options.BasePath != "/" {
-		if !strings.HasPrefix(sH.options.BasePath, "/") || strings.HasSuffix(sH.options.BasePath, "/") {
-			return nil, fmt.Errorf("invalid base path '%s'. Must start but not end with a slash '/', e.g. '/jaeger/ui'", sH.options.BasePath)
+		if sH.options.UIBasePath != "/ui" {
```
yurishkuro, line 147: Check not needed

cmd/query/app/static_handler.go
```diff
@@ -225,8 +236,8 @@ func (sH *StaticAssetsHandler) loggingHandler(handler http.Handler) http.Handler
 // RegisterRoutes registers routes for this handler on the given router
 func (sH *StaticAssetsHandler) RegisterRoutes(router *mux.Router) {
 	fileServer := http.FileServer(sH.assetsFS)
-	if sH.options.BasePath != "/" {
-		fileServer = http.StripPrefix(sH.options.BasePath+"/", fileServer)
+	if sH.options.UIBasePath != "/" {
```
yurishkuro, line 239: File serving is unaffected by ui path

yurishkuro (commented):
Getting closer

Sidhartha-Biswal:
Hi @yurishkuro,

Tested with the latest changes, we still see issues with loading the static assets as shown below.
uierror

yurishkuro (commented):
Your PR still has no changes to the configuration of the reverse proxy. How are you expecting it to work if you're giving different prefixes and nothing in between that is rewriting the URLs?

Sidhartha-Biswal:
Hi @yurishkuro,

Is it mandatory to use a reverse proxy when both the QUERY_BASE_PATH & QUERY_UI_BASE_PATH parameters are set at the same time?

The changes in reverseproxy configuration which you mentioned above are they related to source code changes or test code changes?

yurishkuro:
It's a test file

Sidhartha-Biswal:
Hi @yurishkuro,

Thanks for the support. 
Currently when we set both paths and only using QUERY_UI_BASE_PATH we are able to access query UI but not able to access using QUERY_BASE_PATH, we have a requirement to set both the paths( QUERY_BASE_PATH and QUERY_UI_BASE_PATH) at the same time, and the expectation is that query UI should be able to access using both paths. Is this valid use case and is it possible?

yurishkuro:
This is a valid use case:
```mermaid
flowchart LR
    Browser -->|/foo/bar/...\nUI_BASE_PATH| Proxy
    Proxy -->|/baz/...\nBASE_PATH| JaegerQuery
    InternalClient -->|/baz/api/traces/...\nBASE_PATH| JaegerQuery
```

Sidhartha-Biswal:
Hi @yurishkuro,

Just to clarify my understanding based on the code changes and above diagram, I am writing again here the below scenarios. Please correct it if I am wrong.
 
1) When proxy is involved and both the parameters are set, say UI_BASE_PATH = /foo/bar and BASE_PATH=/baz, Query-UI can only be accessed via UI_BASE_PATH. There is no possibility to access the Query-UI via BASE_PATH. Please correct it if I am wrong.
 
2) When proxy is not involved, there is no need to set the UI_BASE_PATH and Query-UI can be accessed via BASE_PATH

3) When proxy is not involved, Is it still a valid case to set both the parameters UI_BASE_PATH and BASE_PATH?

yurishkuro:
> There is no possibility to access the Query-UI via BASE_PATH. Please correct it if I am wrong.

Correct. Jaeger can serve only **one version** of the UI assets which are dynamically modified to match UI_BASE_PATH.

> When proxy is not involved, there is no need to set the UI_BASE_PATH and Query-UI can be accessed via BASE_PATH

Correct, because Jaeger will simply default UI_BASE_PATH to BASE_PATH if the former is not set.

> When proxy is not involved, Is it still a valid case to set both the parameters UI_BASE_PATH and BASE_PATH?

No. Traffic coming into Jaeger is only recognized if it's relative to BASE_PATH. If you try to access it via different URL you will get 404.

Sidhartha-Biswal:
Hi @yurishkuro,

Thanks for all your support and inputs.
We have a new use-case for this requirement where we have two products "X" & "Y", both the products are currently using gateway1(has no-proxy i.e no-rewrite mechanism). Now product "Y" is migrating from gateway1(has no-proxy i.e no-rewrite mechanism) to gateway2( has proxy i.e rewrite mechanism). Both the products "X" & "Y" use a single configuration file where QUERY_BASE_PATH = /jaeger and QUERY_UI_BASE_PATH = /hub/jaeger. Since product "Y" is being migrated from gateway1 to gateway2, expectation is Query-UI has to be accessed using both the gateways parallelly.
 
We have tested accessing Query-UI for both the products using this configuration with the latest changes.
 
But currently when both values are set, Product "Y" using gateway2 is able to access Query-UI.
Product "X" using gateway1 and Product "Y" using gateway1 have the issue accessing Query-UI
 
Unfortunately we don't have a possibility to have a separate configuration for both the gateways. So in this case, do you think can we achieve this kind of use-case requirement from Jaeger side in any way?

yurishkuro:
The only way to run UI under two external URLs is if the rewriting is taken care of by the proxy, not only on the way in, but in the responses as well (i.e. altering the returned HTML and static assets by replacing the prefix). I don't know how difficult the output rewriting would be and if proxies support that out of the box.

If you're in the middle of migration, I would just run a second instance of Jaeger UI, since it's stateless (unless you're using in-process storage like Badger, then you would need to reconfigure it as a remote storage so that it can be shared across multiple Jaeger processes).

Sidhartha-Biswal:
Hi @yurishkuro,

Thanks for all the support. I am confused with the below two responses given by you. Can you please clarify them
For the below case:

**1.   When proxy is involved and both the parameters are set, say UI_BASE_PATH = /foo/bar and BASE_PATH=/baz, Query-UI can only be accessed via UI_BASE_PATH. There is no possibility to access the Query-UI via BASE_PATH. Please correct it if I am wrong.**

But in the latest comment, you have mentioned that it is still possible to run UI under two external URL's if rewriting can be take care by proxy.....

**The only way to run UI under two external URLs is if the rewriting is taken care of by the proxy, not only on the way in, but in the responses as well (i.e. altering the returned HTML and static assets by replacing the prefix). I don't know how difficult the output rewriting would be and if proxies support that out of the box.**
 
The above two comments seem to be confusing & contradicting for me. Which one is correct? 
Can you please clarify on this.

Sidhartha-Biswal:
Hi @yurishkuro,

Thanks for the support, sorry for delayed reply we were busy with some other works, we wanted resume to this activity.
Currently we wanted to merge these changes which are working fine in proxy case in our environment.
 
But in Jaeger not sure how to test it, can you please suggest where tests needs to be added and how it should be tested.

yurishkuro:
the end to end test is here: https://github.com/jaegertracing/jaeger/tree/main/examples/reverse-proxy

Sidhartha-Biswal:
Hi @yurishkuro,

Thanks for the inputs. 
In our environment with a proxy setup when we try to test with the latest changes, we don't see any issue accessing the Jaeger-UI. But in order to manually test using the reverseproxy example provided by you above, we need to build the jaeger-all-in-one image with our changes I guess. But when we try to run the ./build-all-in-one-image in our local, we are facing some issues with node.js version while building it. We are trying to fix it. But in meantime, we observed that a job was triggered for the commit where different stages like build, test, publish were executed. So if possible can you please provide us the jaeger-all-in-one-image tag with our changes. So that we can use that image and try to execute the docker compose up and check if we are able to access the Jaeger-UI using the UI-Base-Path.

Sidhartha-Biswal:
Hi @yurishkuro,

We were going through the reverse-proxy file shared by you above. We have made the the changes in the PR as per our understanding. We are facing some issues while building the jaeger-all-in-one-image locally which is preventing us from manually checking the Jaeger-UI. Can you please review the changes and let us know if we have to make some modifications in the test files.

Sidhartha-Biswal:
Thanks for all the inputs.
In our environment with a proxy setup when we try to test with the latest changes, we don't see any issue accessing the Jaeger-UI. The changes seems to be working fine as we are able to access the Jaeger-UI via UI-Base-Path.
 
But in order to manually test using the reverseproxy example provided by you above, we need to build the jaeger-all-in-one image with our changes I guess. But when we try to run the ./build-all-in-one-image in our local, we are facing one or other issue. Currently we are seeing the below issue. 

error

We are trying to fix the above error. But in meantime, we observed that a job was triggered for the commit where different stages like build, test, publish were executed. So if possible can you please provide us the jaeger-all-in-one-image tag with our changes. So that we can use that image and try to execute the docker compose up and check if we are able to access the Jaeger-UI using the UI-Base-Path.
 
Can you please help us with this & also can you please confirm if the changes in this PR are fine?

yurishkuro:
no follow-up, closing

PR #5228, opened 2024-02-24
[jaeger-v2] Add support for GRPC storarge

james-ryans:
- Part of #4843
- Separate GRPC storage PR from and will be used by jaeger-v2 Kafka PR #4971 

## Description of the changes
- Implement GRPC storage backend for Jaeger-V2 storage

- Run two `jaegertracing/jaeger-remote-storage` at `17271` and `17281` ports
- Execute `go run -tags=ui ./cmd/jaeger --config ./cmd/jaeger/grpc_config.yaml`

## Checklist

### DIFF
--- a/cmd/jaeger/internal/extension/jaegerstorage/config.go
+++ b/cmd/jaeger/internal/extension/jaegerstorage/config.go
@@ -9,12 +9,14 @@ import (
 
 	memoryCfg "github.com/jaegertracing/jaeger/pkg/memory/config"
 	badgerCfg "github.com/jaegertracing/jaeger/plugin/storage/badger"
+	grpcCfg "github.com/jaegertracing/jaeger/plugin/storage/grpc/config"
 )
 
 // Config has the configuration for jaeger-query,
 type Config struct {
 	Memory map[string]memoryCfg.Configuration   `mapstructure:"memory"`
 	Badger map[string]badgerCfg.NamespaceConfig `mapstructure:"badger"`
+	GRPC   map[string]grpcCfg.Configuration     `mapstructure:"grpc"`
 	// TODO add other storage types here
 	// TODO how will this work with 3rd party storage implementations?
 	//      Option: instead of looking for specific name, check interface.
--- a/plugin/storage/grpc/factory_test.go
+++ b/plugin/storage/grpc/factory_test.go
@@ -143,6 +143,12 @@ func TestGRPCStorageFactory(t *testing.T) {
 	assert.Equal(t, f.store.DependencyReader(), depReader)
 }
 
+func TestGRPCStorageFactoryWithConfig(t *testing.T) {
+	cfg := grpcConfig.Configuration{}
+	_, err := NewFactoryWithConfig(cfg, metrics.NullFactory, zap.NewNop())
+	require.ErrorContains(t, err, "grpc-plugin builder failed to create a store: error connecting to remote storage")
+}
+
 func TestGRPCStorageFactory_Capabilities(t *testing.T) {
 	f := NewFactory()
 	v := viper.New()
--- a/cmd/jaeger/grpc_config.yaml
+++ b/cmd/jaeger/grpc_config.yaml
@@ -1,0 +1,35 @@
+service:
+  extensions: [jaeger_storage, jaeger_query]
+  pipelines:
+    traces:
+      receivers: [otlp]
+      processors: [batch]
+      exporters: [jaeger_storage_exporter]
+
+extensions:
+  jaeger_query:
+    trace_storage: external-storage
+    trace_storage_archive: external-storage-archive
+    ui_config: ./cmd/jaeger/config-ui.json
+
+  jaeger_storage:
+    grpc:
+      external-storage:
+        server: localhost:17271
+        connection-timeout: 5s
+      external-storage-archive:
+        server: localhost:17281
+        connection-timeout: 5s
+
+receivers:
+  otlp:
+    protocols:
+      grpc:
+      http:
+
+processors:
+  batch:
+
+exporters:
+  jaeger_storage_exporter:
+    trace_storage: external-storage
--- a/cmd/jaeger/internal/extension/jaegerstorage/extension.go
+++ b/cmd/jaeger/internal/extension/jaegerstorage/extension.go
@@ -17,6 +17,8 @@ import (
 	"github.com/jaegertracing/jaeger/pkg/metrics"
 	"github.com/jaegertracing/jaeger/plugin/storage/badger"
 	badgerCfg "github.com/jaegertracing/jaeger/plugin/storage/badger"
+	"github.com/jaegertracing/jaeger/plugin/storage/grpc"
+	grpcCfg "github.com/jaegertracing/jaeger/plugin/storage/grpc/config"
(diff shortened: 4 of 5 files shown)


plugin/storage/grpc/config/config.go
```diff
@@ -78,6 +79,9 @@ func (c *Configuration) Close() error {
 		c.pluginHealthCheck.Stop()
 		c.pluginHealthCheckDone <- true
 	}
+	if c.remoteRPCClient != nil {
```
james-ryans, line 82: To pass the code coverage, I set up a gRPC server within the unit test for the factory to use, since we can't just mock the gRPC configuration. Because the current gRPC storage doesn't close the remote gRPC client, I added a method to close the client right inside the factory's `Close()` function. However, it looks like the code coverage tool isn't recognizing this particular line as covered. Even though I've called the Close() function in my unit test, which should hit this line, the method is abstracted behind an interface. Could this be why it's not showing up as covered?

plugin/storage/grpc/config/config.go
```diff
@@ -106,12 +110,13 @@ func (c *Configuration) buildRemote(logger *zap.Logger, tracerProvider trace.Tra
 		opts = append(opts, grpc.WithUnaryInterceptor(tenancy.NewClientUnaryInterceptor(tenancyMgr)))
 		opts = append(opts, grpc.WithStreamInterceptor(tenancy.NewClientStreamInterceptor(tenancyMgr)))
 	}
-	conn, err := grpc.DialContext(ctx, c.RemoteServerAddr, opts...)
+	var err error
+	c.remoteRPCClient, err = grpc.DialContext(ctx, c.RemoteServerAddr, opts...)
```
yurishkuro, line 114: this is not a client
```suggestion
	c.remoteConn, err = grpc.DialContext(ctx, c.RemoteServerAddr, opts...)
```

plugin/storage/grpc/factory_test.go
```diff
@@ -143,6 +147,36 @@ func TestGRPCStorageFactory(t *testing.T) {
...
 
+func TestGRPCStorageFactoryWithConfig(t *testing.T) {
+	cfg := grpcConfig.Configuration{}
+	_, err := NewFactoryWithConfig(cfg, metrics.NullFactory, zap.NewNop())
+	require.ErrorContains(t, err, "grpc-plugin builder failed to create a store: error connecting to remote storage")
+
+	lis, err := net.Listen("tcp", ":0")
+	if err != nil {
+		t.Fatalf("failed to listen: %v", err)
+	}
```
yurishkuro, line 158:
```suggestion
	require.NoError(t, err, "failed to listen")
```

plugin/storage/grpc/config/config.go
yurishkuro, line 114: this is not a client
```suggestion
	c.remoteConn, err = grpc.DialContext(ctx, c.RemoteServerAddr, opts...)
```

plugin/storage/grpc/config/config.go
yurishkuro: in Go coverage is computed in a weird way, it's package local, i.e. coverage is only collected for a package from tests running in the same package. This behavior can be altered by passing additional packages explicitly, as we do for storage integration tests (we tell Go to count coverage in all other packages):
```
$(GOTEST) -coverpkg=./... -coverprofile cover.out $(STORAGE_PKGS)
```
So when the integration test runs for grpc-plugin it should include your Close changes, but not if you just run `go test` manually without `-coverpkg`

cmd/jaeger/internal/extension/jaegerstorage/config.go
```diff
@@ -9,12 +9,14 @@ import (
...
 	memoryCfg "github.com/jaegertracing/jaeger/pkg/memory/config"
 	badgerCfg "github.com/jaegertracing/jaeger/plugin/storage/badger"
+	grpcCfg "github.com/jaegertracing/jaeger/plugin/storage/grpc/config"
 )
 
 // Config has the configuration for jaeger-query,
 type Config struct {
 	Memory map[string]memoryCfg.Configuration   `mapstructure:"memory"`
 	Badger map[string]badgerCfg.NamespaceConfig `mapstructure:"badger"`
+	GRPC   map[string]grpcCfg.Configuration     `mapstructure:"grpc"`
```
yurishkuro, line 19: I don't think this is how we want to do the configuration in V2, but we can address it more holistically later -- #5229

plugin/storage/grpc/factory_test.go
```diff
@@ -143,6 +147,34 @@ func TestGRPCStorageFactory(t *testing.T) {
...
+	defer s.Stop()
+
+	cfg.RemoteServerAddr = lis.Addr().String()
+	cfg.RemoteConnectTimeout = 1 * time.Second
+	f, err := NewFactoryWithConfig(cfg, metrics.NullFactory, zap.NewNop())
+	defer func() {
+		err := f.Close()
+		if err != nil {
+			log.Fatalf("Client exited with error: %v", err)
+		}
```
yurishkuro, line 173:
```suggestion
		require.NoError(t, f.Close())
```

plugin/storage/grpc/factory_test.go
```diff
@@ -143,6 +147,34 @@ func TestGRPCStorageFactory(t *testing.T) {
...
+	}()
+	require.NoError(t, err)
```
yurishkuro, line 175: incorrect order - when creating a closable object, you first need to check that no error was returned from constructor, and only the do defer/close (in this case you don't need defer since it's the last line of the function).

plugin/storage/integration/grpc_test.go
```diff
@@ -96,14 +96,20 @@ func (s *gRPCServer) Restart() error {
...
+	logger  *zap.Logger
+	flags   []string
+	factory *grpc.Factory
+	server  *gRPCServer
 }
 
 func (s *GRPCStorageIntegrationTestSuite) initialize() error {
 	s.logger, _ = testutils.NewLogger()
 
+	if s.factory != nil {
```
yurishkuro, line 108: how would it be not nil at this point if the factory is only created in L119? Is it from another run of the test? If so we should close it as part of the clean-up, not as part of starting a new test

plugin/storage/grpc/factory_test.go
james-ryans: > in this case you don't need defer since it's the last line of the function

Ah yes, that is a good point.

plugin/storage/integration/grpc_test.go
james-ryans: Okay, my initial thought was because it restarts the server when initializing, so I think it makes sense if I put it there too. But, moving it to the clean-up process does make things clearer. I've changed it.

james-ryans:
Do you know why the previous Protogen Validation CI failed? I didn't change anything in the idl module, and now it succeeded without me fixing anything. https://github.com/jaegertracing/jaeger/actions/runs/8066523840

yurishkuro:
Unclear why it failed:
```
--gogo_out: protoc-gen-gogo: Plugin failed with status code 127.
make: *** [Makefile.Protobuf.mk:97: proto-api-v2] Error 1
```
It's strange that the container doesn't print any more logs about it.

yurishkuro (approved):

PR #5242, opened 2024-03-02
[jaeger-v2] Add support for artificial jaeger storage receiver

james-ryans:
- Part of #4843
- Separate Jaeger storage receiver PR from and will be used by jaeger-v2 Kafka PR #4971

## Description of the changes
- Implement Jaeger storage receiver to be used by Jaeger-v2 Kafka integration test.

- Added some unit tests.

## Checklist

### DIFF
--- a/cmd/jaeger/integration/receivers/storagereceiver/README.md
+++ b/cmd/jaeger/integration/receivers/storagereceiver/README.md
@@ -9,7 +9,7 @@
 
 The following settings are required:
 
-- `trace_storage` (no default): Jaeger's storage extension name
+- `trace_storage` (no default): name of a storage backend defined in `jaegerstorage` extension
 
 The following settings can be optionally configured:
 
--- a/cmd/jaeger/integration/receivers/storagereceiver/README.md
+++ b/cmd/jaeger/integration/receivers/storagereceiver/README.md
@@ -1,4 +1,23 @@
+# Storage Receiver
+
 `storagereceiver` is a fake receiver that creates an artificial stream of traces by:
 
 - repeatedly querying one of Jaeger storage backends for all traces (by service).
 - tracking new traces / spans and passing them to the next component in the pipeline.
+
+# Getting Started
+
+The following settings are required:
+
+- `trace_storage` (no default): Jaeger's storage extension name
+
+The following settings can be optionally configured:
+
+- `pull_interval` (default = 0s): The delay between each iteration of pulling traces.
+
+```yaml
+receivers:
+  jaeger_storage_receiver:
+    trace_storage: external-storage
+    pull_interval: 0s
+```
--- a/cmd/jaeger/integration/receivers/storagereceiver/README.md
+++ b/cmd/jaeger/integration/receivers/storagereceiver/README.md
@@ -1,0 +1,4 @@
+`storagereceiver` is a fake receiver that creates an artificial stream of traces by:
+
+- repeatedly querying one of Jaeger storage backends for all traces (by service).
+- tracking new traces / spans and passing them to the next component in the pipeline.
--- a/cmd/jaeger/integration/receivers/storagereceiver/receiver.go
+++ b/cmd/jaeger/integration/receivers/storagereceiver/receiver.go
@@ -1,0 +1,132 @@
+// Copyright (c) 2024 The Jaeger Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package storagereceiver
+
+import (
+	"context"
+	"fmt"
+
+	jaeger2otlp "github.com/open-telemetry/opentelemetry-collector-contrib/pkg/translator/jaeger"
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/consumer"
+	"go.uber.org/zap"
+
+	"github.com/jaegertracing/jaeger/cmd/jaeger/internal/extension/jaegerstorage"
+	"github.com/jaegertracing/jaeger/model"
+	"github.com/jaegertracing/jaeger/storage/spanstore"
+)
+
+type storageReceiver struct {
+	cancelConsumeLoop context.CancelFunc
+	config            *Config
+	logger            *zap.Logger
+	consumedTraces    map[model.TraceID]*consumedTrace
+	nextConsumer      consumer.Traces
+	spanReader        spanstore.Reader
+}
+
+type consumedTrace struct {
+	spanIDs map[model.SpanID]struct{}
+}
+
+func newReceiver(config *Config, otel component.TelemetrySettings, nextConsumer consumer.Traces) (*storageReceiver, error) {
+	return &storageReceiver{
+		config:         config,
+		logger:         otel.Logger,
+		consumedTraces: make(map[model.TraceID]*consumedTrace),
+		nextConsumer:   nextConsumer,
+	}, nil
+}
+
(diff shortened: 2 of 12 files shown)


james-ryans:
Hi @yurishkuro, I've implemented the Jaeger storage receiver as you stated it's better to separate this PR. There are still things to add here:
1. Improve coverage but I will need to abstract the jaeger storage extension with interface similar to #5123
2. Add a configurable pull interval delay

cmd/jaeger/integration/receivers/storagereceiver/receiver.go
```diff
@@ -0,0 +1,134 @@
...
+		for _, svc := range services {
+			if err := r.consumeTraces(ctx, svc); err != nil {
+				r.settings.Logger.Error("Failed to consume traces from consumer", zap.Error(err))
+			}
+		}
+
+		if ctx.Err() != nil {
+			r.settings.Logger.Error("Consumer stopped", zap.Error(ctx.Err()))
+			return ctx.Err()
+		}
```
james-ryans, line 82: > Why is this an error for the consumer?

I actually follow kafkareceiver from otelcol-contrib since they're both a "pull" strategy to retrieve traces. https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/3652b444802fad4819b50fbda41e0957000d8279/receiver/kafkareceiver/kafka_receiver.go#L171-L185

cmd/jaeger/integration/receivers/storagereceiver/receiver.go
```diff
@@ -0,0 +1,134 @@
...
+		if err := r.consumeLoop(ctx); err != nil {
+			r.settings.ReportStatus(component.NewFatalErrorEvent(err))
+		}
+	}()
+
+	return nil
+}
+
+func (r *storageReceiver) consumeLoop(ctx context.Context) error {
+	for {
```
james-ryans, line 66: > how do we exit this loop? Should it be checking ctx.Done for that?

When the `r.cancelConsumeLoop()` is called from `Shutdown`, we can capture that with `ctx.Err()`. Is it better to check with `ctx.Done`? I followed kafkareceiver implementation from otelcol-contrib actually https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/3652b444802fad4819b50fbda41e0957000d8279/receiver/kafkareceiver/kafka_receiver.go#L171-L185

cmd/jaeger/integration/receivers/storagereceiver/receiver.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	for _, trace := range traces {
+		traceID := trace.Spans[0].TraceID
+		if _, ok := r.consumedTraces[traceID]; !ok {
+			r.consumedTraces[traceID] = &consumedTrace{
+				spanIDs: make(map[model.SpanID]struct{}),
+			}
+		}
+		if len(trace.Spans) > len(r.consumedTraces[traceID].spanIDs) {
```
james-ryans, line 101: > is this check important? the next function will skip spans that are already received

My thought on this logic is that it will reduce the time complexity because I won't need to loop through every span in the trace once it's fully stored. For example, if the average trace contains 10 spans, the time complexity would be reduced by 10x for the next iteration. Fun fact: I've just compared it, and with this check, it improves the ./scripts/otel-kafka-integration-test.sh runtime by 1-2 seconds. Do you think it's somewhat overkill for an artificial receiver?

yurishkuro: I do think it's overkill (rather skeptical of 1-2sec to be coming from just the check against a map). It is also more robust to actually go through all spans and check against IDs map, rather than rely on the count

cmd/jaeger/integration/receivers/storagereceiver/receiver.go
yurishkuro: I think it's wrong to treat orderly shutdown as error and log all kinds of error messages. You can check Done instead
```go
        select {
        case <-ctx.Done():
            return
        default:
            time.Sleep(time.Second)
        }
```

cmd/jaeger/integration/receivers/storagereceiver/receiver_test.go
```diff
@@ -0,0 +1,264 @@
...
+
+type storageHost struct {
+	t                *testing.T
+	storageExtension component.Component
+}
+
+func (host storageHost) GetExtensions() map[component.ID]component.Component {
+	myMap := make(map[component.ID]component.Component)
+	myMap[jaegerstorage.ID] = host.storageExtension
+	return myMap
```
yurishkuro, line 41:
```suggestion
	return map[component.ID]component.Component{
	  jaegerstorage.ID: host.storageExtension,
	}
```

cmd/jaeger/integration/receivers/storagereceiver/receiver_test.go
```diff
@@ -0,0 +1,264 @@
...
+				reader.On(
+					"FindTraces",
+					mock.AnythingOfType("*context.cancelCtx"),
+					&spanstore.TraceQueryParameters{ServiceName: service},
+				).Return(test.traces, test.tracesErr)
+			}
+			r.spanReader = reader
+
+			require.NoError(t, r.Shutdown(ctx))
+			err := r.consumeLoop(ctx)
```
yurishkuro, line 202: are you sure you're actually testing different branches of accumulation of spans? Since you call consumeLoop after canceling the context, it will run exactly once. In other words, I don't really see how different test cases would actually result in different behavior, since some of the behavior is only triggered if the loop runs more than once.

cmd/jaeger/integration/receivers/storagereceiver/receiver_test.go
```diff
@@ -0,0 +1,264 @@
...
+			actualTraces := sink.AllTraces()
+			assert.Equal(t, expectedTraces, actualTraces)
+		})
+	}
+}
+
+func newStorageHost(t *testing.T, traceStorage string) *storageHost {
+	ctx := context.Background()
+
+	lis, err := net.Listen("tcp", ":0")
```
yurishkuro, line 225: if you're mocking the SpanReader, when do you need a remote service?

cmd/jaeger/integration/receivers/storagereceiver/receiver_test.go
james-ryans: Yes, because I use the same receiver for every test case, the whole test sequence is a continuation from the previous case. Also, the behavior of calling consumeLoop after cancelling the context is expected. This approach allows me to control the simulation of consumeLoop, assuming that each test case represents an iteration of the for loop in the consumeLoop function and read the updated traces.

cmd/jaeger/integration/receivers/storagereceiver/README.md
```diff
@@ -0,0 +1,23 @@
...
+`storagereceiver` is a fake receiver that creates an artificial stream of traces by:
+
+- repeatedly querying one of Jaeger storage backends for all traces (by service).
+- tracking new traces / spans and passing them to the next component in the pipeline.
+
+# Getting Started
+
+The following settings are required:
+
+- `trace_storage` (no default): Jaeger's storage extension name
```
yurishkuro, line 12:
```suggestion
- `trace_storage` (no default): name of a storage backend defined in `jaegerstorage` extension
```

yurishkuro (approved):

PR #5253, opened 2024-03-06
[jaeger-v2] Add support for Cassandra

Pushkarm029:
- part of #4843 

## Description of the changes
- add support for cassandra

- currently, tests are failing.

## Checklist
- [X] I have added unit tests for the new functionality


Pushkarm029:
@james-ryans @akagami-harsh, please review.

Currently, TestCassandraStorageFactoryWithConfig is failing with a timeout error. Maybe some config issue???

james-ryans:
> Currently, TestCassandraStorageFactoryWithConfig is failing with a timeout error. Maybe some config issue???

It is because you are trying to connect to HTTP server with Cassandra client, if you set the config with `ConnectTimeout` to `1 * time.Second` you will get the connection timeout error
```
2024/03/06 23:25:39 gocql: unable to dial control conn 127.0.0.1:50134: gocql: no response to connection startup within timeout
```

That is because Cassandra connection is over TCP instead of HTTP. Perhaps, you should try to spawn a TCP server instead of HTTP server, here I found a guide to create one but I haven't successfully implement it for Cassandra connection. https://okanexe.medium.com/the-complete-guide-to-tcp-ip-connections-in-golang-1216dae27b5a

Pushkarm029:
@yurishkuro, I cannot create a connection between the test and Cassandra. Everytime, i try something, I get this error : 
```bash
Received unexpected error: gocql: unable to create session: unable to discover protocol version
```
Do you have any idea how I can do it?

In other tests, the first factory is created with f.NewFactory(), the mock session is passed in, and then it checks noError in f.initialize. But, here, we can't do the same. since both f.NewFactory() and f.Initialize() is part of the function NewFactoryWithConfig().
```go
f := NewFactory()
f.primaryConfig = newMockSessionBuilder(session, nil)
require.NoError(t, f.Initialize(metrics.NullFactory, logger))
```

yurishkuro:
now do we do that in Elastic? It also uses NewFactoryWithConfig pattern.

yurishkuro:
Creating a custom server is a useful pattern. Can we do the same for Cassandra? It may be more difficult as I don't think it's using HTTP protocol.

Pushkarm029:
Yeah, we can create a custom server for Cassandra. Can you provide some hints/examples?

yurishkuro:
Well again, you can see the example in ES tests. But ES uses HTTP so it's much easier to mock. I don't know if Cassandra can use HTTP protocol. If it's only some bespoke binary protocol, than it's harder to mock (perhaps there are libraries out there for that, or perhaps gocql has a place were we can substitute the actual server communication).

There are some tricks we can do internally by not combining creation and connection in a single step.

Pushkarm029:
removing unit tests

james-ryans:
Pushkar, I believe what Yuri meant previously was that we should only leave the unit test that needs to connect to Cassandra (which we can't mock) and keep all other unit tests you successfully created. What do you think?

plugin/storage/cassandra/factory.go
```diff
@@ -78,6 +80,56 @@ func NewFactory() *Factory {
...
+	f := NewFactory()
+	f.InitFromOptions(&Options{
+		Primary: namespaceConfig{
+			Configuration: cfg,
+			servers:       serverURL,
+			namespace:     primaryStorageConfig,
+			Enabled:       true,
+		},
+		others:                 archive,
+		SpanStoreWriteCacheTTL: time.Hour * 12,
```
yurishkuro, line 110: Why magic number here?

plugin/storage/cassandra/factory.go
```diff
@@ -78,6 +80,56 @@ func NewFactory() *Factory {
...
+	})
+	err = f.Initialize(metricsFactory, logger)
+	if err != nil {
+		return nil, err
+	}
+	return f, nil
+}
+
+func getServers(servers []string) (string, error) {
```
yurishkuro, line 119: Not following why we need to go from structured array back to a string

plugin/storage/cassandra/factory.go
Pushkarm029: left it by mistake

plugin/storage/cassandra/factory.go
Pushkarm029: We are converting an array to a string because namespaceConfig.servers want a string. Which on further processing in f.Options.GetPrimary() or f.Options.Get() gets converted back to []string.

if we don't want this conversion. we have to change it in these functions.

pkg/cassandra/config/config.go
```diff
@@ -29,21 +30,21 @@ import (
 
 // Configuration describes the configuration properties needed to connect to a Cassandra cluster
 type Configuration struct {
-	Servers              []string       `validate:"nonzero" mapstructure:"servers"`
+	Servers              []string       `validate:"nonzero" mapstructure:"servers" valid:"required,url"`
```
yurishkuro, line 33: do we know what would be using the `validate:` tags? Maybe we should convert them to `valid:`?

plugin/storage/cassandra/factory.go
yurishkuro: ok, but we don't need to be calling InitFromOptions function, which is designed to transfer information from CLI flags into Config. You already have config here, just assign it to the factory, and we wouldn't need these shenanigans.

Pushkarm029:
working on it

pkg/cassandra/config/config.go
Pushkarm029: go-playground/validator is using the `validate:` tags. Since we are using asaskevich/govalidator, we should change this to `valid:`.

yurishkuro (approved):

Issue #5254, opened 2024-03-06
[jaeger-v2] Storage backend integration tests


james-ryans:
### Requirement

With the Jaeger storage extension for Jaeger-v2 is going to have full support of Jaeger-v1's storage backends, some unit tests on every storage backends are not enough. We need to conduct end-to-end tests of OpenTelemetry Collector pipeline to the targeted database.

### Problem

There are still no integration tests to test the actual stored traces to the database from V2 Jaeger storage extension.

### Proposal

Fortunately, OpenTelemetry Collector already has a [testbed](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/testbed) framework to help us conducting the end-to-end tests.

Testbed is a controlled environment and tools for conducting end-to-end tests for the OpenTelemetry Collector, including reproducible short-term benchmarks, correctness tests, long-running stability tests and maximum load stress tests. However, we will only utilize the correctness tests from testbed, it generates and sends every combinatorial trace attributes and matches every single of them with the received traces from another end.

## Architecture of the integration test
Here's the architecture we will use to test the OpenTelemetry Collector pipeline from end-to-end with the designated storage backends.
![jaeger-v2-testbed](https://github.com/jaegertracing/jaeger/assets/46216691/00f5a9ee-1b14-4b2d-8de7-c51767e50574)
Testbed components:
- `LoadGenerator` - encapsulates DataProvider and DataSender in order to generate and send data.
  - `Golden DataProvider` - generates traces from the "Golden" dataset generated using pairwise combinatorial testing techniques. Testbed example uses [PICT](https://github.com/microsoft/pict/) to generate the test data, e.g. [testdata](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/internal/coreinternal/goldendataset/testdata).
  - `OTLP Trace DataSender` - with the generated traces from DataProvider, the DataSender sends traces to OTLP receiver in the collector instance.
- `Mockbackend` - encapsulates DataReceiver and provides consume functionality.
  - `DataReceiver` - we will create a custom DataReceiver that will host a Jaeger storage extension to retrieve traces from the database by pulling them using our artificial Jaeger storage receiver https://github.com/jaegertracing/jaeger/pull/5242.
  - `Consumer` - consumer does not actually a thing in MockBackend but only to make the diagram intuitive, the traces received from our artificial receiver will be stored inside MockBackend.
[...]

PR #5255, opened 2024-03-07
Prepare v1.55.0 release

jkowall:
Prepares Jaeger for release 1.55.0.

Includes jaeger-ui submodule updates.

I can add more updates for the RELEASE.md if needed based on the Slack discussion. Thank you.

PR #5259, opened 2024-03-08
[jaeger-v2] add GRPC storage backend integration test

james-ryans:
- Resolves GRPC integration test sub-task at #5254

## Description of the changes
- Created a `grpc-integration-test.sh` script to run a jaeger remote storage and execute the end-to-end test through the OpenTelemetry Collector pipeline with the jaeger storage extension inside connected to the remote storage. To have a visualization of this architecture, see the proposal at #5254
- Separate the GRPC and Badger integration test CI because GRPC need to be run twice for the v1 and v2 versions.

- Run `./scripts/grpc-integration-test.sh latest` and the whole remote storage and pipeline will be built and executed for you.
- I also ran a `jaeger-query`component to query traces from the remote storage for manual checks. 
Screenshot 2024-03-08 at 09 58 13

## Checklist


james-ryans:
Am I allowed to split the `grpc-and-badger` CI? I need to do this because the CIT gRPC requires a matrix for Jaeger v1 and v2. But the `grpc-and-badger` is stuck on `Expected — Waiting for status to be reported`. Is this something that needs to be tinkered with in the repository settings?

Makefile
```diff
@@ -118,9 +119,16 @@ storage-integration-test:
 	go clean -testcache
 	bash -c "set -e; set -o pipefail; $(GOTEST) -coverpkg=./... -coverprofile cover.out $(STORAGE_PKGS) $(COLORIZE)"
 
+.PHONY: jaeger-storage-integration-test
+jaeger-storage-integration-test:
```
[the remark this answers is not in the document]
yurishkuro: At least add a comment explaining what it does

cmd/jaeger/integration/fixtures/grpc_config.yaml
```diff
@@ -0,0 +1,23 @@
...
+  extensions: [jaeger_storage]
+  pipelines:
+    traces:
+      receivers: [otlp]
+      exporters: [jaeger_storage_exporter]
+
+extensions:
+  jaeger_storage:
+    grpc:
+      external-storage:
```
yurishkuro, line 11:
```suggestion
      some-external-storage:
```
With the current name the config reads as if this is a part of the actual config structure; adding "some-" emphasizes that it's just a user-given name

scripts/grpc-integration-test.sh
```diff
@@ -0,0 +1,105 @@
+#!/bin/bash
+
+PS4='T$(date "+%H:%M:%S") '
```
yurishkuro, line 3: TIL!

scripts/grpc-integration-test.sh
```diff
@@ -0,0 +1,105 @@
...
+db_is_up=
+
+usage() {
+  echo $"Usage: $0 <image_version>"
+  exit 1
+}
+
+check_arg() {
+  if [ ! $# -eq 1 ]; then
+    echo "ERROR: need exactly one argument, <image_version>"
```
yurishkuro, line 16: It seems like you're trying to run already published image of jaeger-remote-storage? We don't have to fix it in this pr but the preferred way is to run with the code built from THIS version of the repo, not some other version.

yurishkuro: Please add a ticket

cmd/jaeger/integration/grpc_test.go
```diff
@@ -0,0 +1,84 @@
...
+import (
+	"log"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/open-telemetry/opentelemetry-collector-contrib/testbed/testbed"
+	"github.com/stretchr/testify/require"
+
+	"github.com/jaegertracing/jaeger/cmd/jaeger/integration/datareceivers"
```
yurishkuro, line 15: Do you foresee other data receivers? If not we could define this directly in integration pkg since they are closely coupled. And we could also move integration pkg under internal so that you can call components() function

Makefile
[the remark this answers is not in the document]
james-ryans: Okay, I've added an explanation, please take a look!

scripts/grpc-integration-test.sh
james-ryans: Mm.. sorry, I don't quite follow. What does TIL mean?

yurishkuro: TIL- today I learned

cmd/jaeger/integration/grpc_test.go
yurishkuro: Thoughts?

james-ryans: > Do you foresee other data receivers?

We will need another data receivers for Kafka, I've created one at #4971. Actually, I've just read your comment at Kafka data receiver that you prefer this package to be moved under "$root/internal/cit/". Is that still relevant?

> we could also move integration pkg under internal so that you can call components() function

I've just tried that to move `cmd/jaeger/integration` to `cmd/jaeger/internal/integration` but I can't call the components() function when it is in private. Am I miss understood something here?

yurishkuro: Yeah we'd still need Components as public, I suppose. Although another alternative is to define your own components for integration tests, as you probably won't need everything from production build.

james-ryans: Okay, I just felt like making the Components function public, since it will still be kept private within the internal directory. Defining a different components function means maintaining two similar functions with different environments, which could lead to oversight in the future (probably? this is only my thought process).

james-ryans:
The unit tests CI is failing due to jaeger storage extension goroutine leak but I still cannot find the cause. It's so hard to replicate, I can't even reproduce the error once.
```
[Goroutine 96 in state chan receive, with net.(*Resolver).lookupIPAddr.func2 on top of the stack:
net.(*Resolver).lookupIPAddr.func2(0xc0001107e0, 0x225ebb0)
	/opt/hostedtoolcache/go/1.22.0/x64/src/net/lookup.go:338 +0x31
created by net.(*Resolver).lookupIPAddr in goroutine 57
	/opt/hostedtoolcache/go/1.22.0/x64/src/net/lookup.go:353 +0x7dc
 Goroutine 58 in state chan receive, with net.(*Resolver).goLookupIPCNAMEOrder.func4 on top of the stack:
 ...
 Goroutine 60 in state IO wait, with internal/poll.runtime_pollWait on top of the stack:
 ...
]
FAIL	github.com/jaegertracing/jaeger/cmd/jaeger/internal/extension/jaegerstorage	6.670s
```

james-ryans:
Not sure if this actually solves the issue but I've changed the `TestESStorageExtensionError` at extension_test.go to connect to `http://127.0.0.1:65535` instead of `http://badurl` to avoid connecting to an external network, which caused the goroutine leaks of looking for badurl IP addr.

yurishkuro:
> Not sure if this actually solves the issue but I've changed ...

This may cause the execution to go through a different code path and reduce code coverage (just guessing).

james-ryans:
> > Not sure if this actually solves the issue but I've changed ...
> 
> This may cause the execution to go through a different code path and reduce code coverage (just guessing).

Hmm.. from my understanding, I believe this unit test is to cover this line of code, which both `http://badurl` and `http://localhost:65535` failed to initialize the storage due to no elasticsearch server hosted at that URL. https://github.com/jaegertracing/jaeger/blob/b990fe91c3d863cb9b9cc3e7bc6bf467b91eeb87/cmd/jaeger/internal/extension/jaegerstorage/extension.go#L91-L93

And also, to back up my argument, I've done test coverage locally specific only to the `TestESStorageExtensionError` unit test, and both are exactly the same.

Or do you have any suggestions on how to fix this? Perhaps I'm doing it wrong. Because I've looked into another similar issue https://gitlab.com/gitlab-org/gitaly/-/issues/4916 and they fixed it by changing the URL to connect to localhost in the unit test too.

cmd/jaeger/internal/integration/grpc_test.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/open-telemetry/opentelemetry-collector-contrib/testbed/testbed"
+	"github.com/stretchr/testify/require"
+
+	"github.com/jaegertracing/jaeger/cmd/jaeger/internal"
+	"github.com/jaegertracing/jaeger/cmd/jaeger/internal/extension/jaegerstorage"
+	"github.com/jaegertracing/jaeger/cmd/jaeger/internal/integration/datareceivers"
+	grpcCfg "github.com/jaegertracing/jaeger/plugin/storage/grpc/config"
+)
+
+func TestGRPCStorage(t *testing.T) {
```
yurishkuro, line 21: This looks like way too much boiler plate that I assume will be common across all storage tests. Can we refactor to make it reusable?

yurishkuro (commented):
I'm good with badurl change

james-ryans:
I've made some adjustments:
- Refactor to remove the boilerplate of testbed codes and now we just need to pass the config file
- I also implemented to automatically create the data receiver from that config file since the data receiver needs to host the storage extension exactly as the collector.
- Fixed unit test goroutine leaking at `/plugin/storage/es` similar to the previous issue, instead this time fails at `FAIL: TestPasswordFromFile` because it contains individual goleak check which detected goroutine leaking exactly from the previous test at `TestESStorageFactoryWithConfigError`.

cmd/jaeger/internal/integration/integration.go
```diff
@@ -0,0 +1,122 @@
...
+	cfg, err := configProvider.Get(context.Background(), factories)
+	require.NoError(t, err)
+
+	config, ok := cfg.Extensions[jaegerstorage.ID].(*jaegerstorage.Config)
+	require.True(t, ok, "no jaeger storage extension found in the config")
+
+	// Hacky way to get the storage extension name.
+	// This way we don't need to modify this,
+	// when a new storage backend is added.
+	name := ""
```
yurishkuro, line 50: Please explain what you are trying to do. Can we not create the component always with the same name for integration tests?

yurishkuro (commented):
jaeger-ui sub module should not be changing in this PR, right?

cmd/jaeger/internal/integration/integration.go
james-ryans: From the architecture at issue #5254, I've designed that the data receiver will utilize the artificial jaeger receiver to pull the traces from the database which is through jaeger storage extension. Because of that, we need to host another jaeger storage extension that lives in the data receiver, and the jaeger receiver will point to that extension.

So, since the jaeger storage extension in the data receiver is a duplication from the collector's extension, I unmarshal the extension config to be used for the data receiver to host it. But now the jaeger receiver needs to know the name of that storage to be set at `TraceStorage`, so I need to scrape it through the config whether it is in Memory/Badger/GRPC/Elasticsearch.

But while explaining, I've thought of another easier way to set the jaeger receiver `TraceStorage`. Which is I just need to get it from the exporter. I'm changing the implementation right now and add the explanation to `newDataReceiver` function.

cmd/jaeger/internal/integration/grpc_test.go
```diff
@@ -0,0 +1,20 @@
...
+
+import (
+	"os"
+	"testing"
+)
+
+func TestGRPCStorage(t *testing.T) {
+	if os.Getenv("STORAGE") != "jaeger_grpc" {
+		t.Skip("Integration test against Jaeger-V2 GRPC skipped; set STORAGE env var to jaeger_grpc to run this")
+	}
```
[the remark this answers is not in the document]
james-ryans: Okay, I think it's always going to be used. Will move it.

cmd/jaeger/internal/integration/grpc_test.go
```diff
@@ -0,0 +1,20 @@
...
+
+package integration
+
+import (
+	"os"
+	"testing"
+)
+
+func TestGRPCStorage(t *testing.T) {
+	if os.Getenv("STORAGE") != "jaeger_grpc" {
```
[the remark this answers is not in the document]
james-ryans: Actually, now that you mention it, we don't need the `jaeger_` prefix. I thought we might need it to differentiate the v1 and v2 envvar so they wouldn't collide, but they are already separated by the Makefile target. V1 calls `storage-integration-test`, and v2 calls `jaeger-storage-integration-test`. I'll remove the prefix.

james-ryans:
> jaeger-ui sub module should not be changing in this PR, right?

This is because I merged this change #5255 from `main` into this branch

yurishkuro (commented):
to fix jaeger-ui:
  * git checkout main
  * git submodule update
  * git checkout your-branch
  * git status (should show jaeger-ui as changed)
  * git commit -s

cmd/jaeger/internal/integration/integration.go
```diff
@@ -0,0 +1,121 @@
...
+	validator := testbed.NewCorrectTestValidator(
+		sender.ProtocolName(),
+		receiver.ProtocolName(),
+		dataProvider,
+	)
+	correctnessResults := &testbed.CorrectnessResults{}
+
+	config, err := os.ReadFile(s.ConfigFile)
+	require.NoError(t, err)
+	log.Println(string(config))
```
yurishkuro, line 86: t.Log

yurishkuro (approved):
You used to have all kinds of diagrams explaining the set up of these integration tests, where are those? I only see one README, which is quite short.

james-ryans:
> You used to have all kinds of diagrams explaining the set up of these integration tests, where are those? I only see one README, which is quite short.

I forgot to add the README but I have now added it. Currently, there's only one diagram for the general architecture of the integration test. For additional diagrams, I've referred to the Testbed README since they have their own diagram and another diagram will be included with the Jaeger-v2 Kafka PR.

yurishkuro (approved):

yurishkuro:
@james-ryans can we add goleaks check?
```
🔴 Error(check-goleak): no goleak check in package ./cmd/jaeger/internal/integration/
```

james-ryans:
> @james-ryans can we add goleaks check?
> 
> ```
> 🔴 Error(check-goleak): no goleak check in package ./cmd/jaeger/internal/integration/
> ```

Yes, but apparently it has a goroutine leaking from the testbed. There's no functionality to close the data sender's gRPC connection, I'll try to solve the upstream issue first then get back here.

PR #5263, opened 2024-03-09
[kafka-producer] Support setting max message size

sappusaketh:
- Exposes MaxMessageBytes config mentioned in https://github.com/jaegertracing/jaeger/issues/1335

## Description of the changes
- Adds support to set MaxMessageBytes config on producer
- Also setting a default which matches with [sarama](https://github.com/IBM/sarama/blob/main/config.go#L177) default unclear if we need to set that default explicitly from jaeger producer initialization 

- Tested on our development jaeger cluster. After seeing below error we made following changes and built a new Image and verified that my changes fixed the issue
```
{"level":"error","ts":1550003610.8029132,"caller":"kafka/writer.go:59","msg":"kafka server: Message was too large, server rejected it to avoid allocation error.","stacktrace":"[github.com/jaegertracing/jaeger/plugin/storage/kafka.NewSpanWriter.func2](http://github.com/jaegertracing/jaeger/plugin/storage/kafka.NewSpanWriter.func2)\n\t/home/travis/gopath/src/github.com/jaegertracing/jaeger/plugin/storage/kafka/writer.go:59"}
```

## Checklist

### DIFF
--- a/plugin/storage/kafka/options.go
+++ b/plugin/storage/kafka/options.go
@@ -47,6 +47,7 @@ const (
 	suffixBatchSize        = ".batch-size"
 	suffixBatchMinMessages = ".batch-min-messages"
 	suffixBatchMaxMessages = ".batch-max-messages"
+	suffixMaxMessageBytes  = ".max-message-bytes"
 
 	defaultBroker           = "127.0.0.1:9092"
 	defaultTopic            = "jaeger-spans"
@@ -58,6 +59,7 @@ const (
 	defaultBatchSize        = 0
 	defaultBatchMinMessages = 0
 	defaultBatchMaxMessages = 0
+	defaultMaxMessageBytes  = 1000000 // https://github.com/IBM/sarama/blob/main/config.go#L177
 )
 
 var (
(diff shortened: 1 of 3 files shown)


plugin/storage/kafka/options.go
```diff
@@ -152,6 +154,11 @@ func (opt *Options) AddFlags(flagSet *flag.FlagSet) {
 		defaultBatchMaxMessages,
 		"(experimental) Maximum number of message to batch before sending records to Kafka",
 	)
+	flagSet.Int(
+		configPrefix+suffixMaxMessageBytes,
+		defaultMaxMessageBytes,
+		"(experimental) The maximum permitted size of a message (defaults to 1000000). Should be set equal to or smaller than the broker's `message.max.bytes`.",
```
yurishkuro, line 160:
```suggestion
		"(experimental) The maximum permitted size of a message. Should be set equal to or smaller than the broker's `message.max.bytes`.",
```

yurishkuro (approved):
thanks

Issue #5270, opened 2024-03-12
[Bug]: Operation based sampling strategies are not being returned for ratelimiting service strategies


kuujis:
Edit by @yurishkuro 

* [x] introduce new flag with correct behavior #5277
* [x] change that flag to be on by default (v2.2.0/v1.65.0) #6441
* [ ] remove the flag and the old code (circa March 2025) #6872
 
1. (bug) When service has rate limiting strategy at the service level, and defines no per-operation strategies, the global per-operation strategies were also ignored
2. (change) When service has probabilistic service-level strategy, and defines no per-operation strategies, the global per-operation strategies were used, but the default sampling rate was taken from global settings, whereas arguably service-level setting should take priority.

Example: given this config file where services A and B do not define their own per-operation strategies:

```json
{ 
      "service_strategies": [ 
        { 
          "service": "ServiceA", 
          "type": "probabilistic", 
          "param": 1.0 
        }, 
        { 
         "service": "ServiceB", 
         "type": "ratelimiting", 
         "param": 3 
       } 
      ], 
      "default_strategy": { 
        "type": "probabilistic", 
        "param": 0.2, 
        "operation_strategies": [ 
          { 
            "operation": "/health", 
            "type": "probabilistic", 
            "param": 0.1
          } 
        ] 
      } 
    } 
```

We expect ServiceA to have the following strategy:
  * it inherits per-operation strategies from global config
  * except for `defaultSamplingProbability` which is kept the same `1.0` as at the service-level
```json
{
  "probabilisticSampling": {
    "samplingRate": 1
  },
  "operationSampling": {
    "defaultSamplingProbability": 1,
    "perOperationStrategies": [
      {
        "operation": "/health",
        "probabilisticSampling": {
          "samplingRate": 0.1
        }
      }
    ]
  }
}
```
We expect ServiceB to have the following strategy:
  * it inherits per-operation strategies from global config as is
```json
{
  "strategyType": 1,
  "rateLimitingSampling": {
    "maxTracesPerSecond": 3
  },
  "operationSampling": {
    "defaultSamplingProbability": 0.2,
    "perOperationStrategies": [
      {
        "operation": "/health",
        "probabilisticSampling": {
          "samplingRate": 0.1
        }
      }
    ]
  }
}
```

Original ticket below:
---

### What happened?

[...]

PR #5277, opened 2024-03-15
[sampling] Fix merging of per-operation strategies into service strategies without them

kuujis:
- Part 1 of [#5270]

## Description of the changes
- See issue for description of the bug and of the two changed behaviors
  - Default `probabilistic` operation level strategies will now be returned if service strategy is of `ratelimiting` type
  - If the service itself has `probabilistic` strategy, its sampling rate takes priority for operation-level definition

- Via updated unit test
- By building local all-in-one executable and using it as a otel instrumentation backend for simple java service https://github.com/kuujis/opentelemetry-java-5504 and using the below `sampling_strategies.json` from https://github.com/open-telemetry/opentelemetry-java/issues/5504
```
{
  "service_strategies": [
    {
      "service": "foobar",
      "type": "ratelimiting",
      "param": 2
    }
  ],
  "default_strategy": {
    "type": "probabilistic",
    "param": 0.2,
    "operation_strategies": [
      {
        "operation": "metrics",
        "type": "probabilistic",
        "param": 0.0
      }
    ]
  }
}
```

## Checklist


yurishkuro (commented):
Sorry, it's been quite a while since this code was written, and I am not able to follow either your change or your unit test changes. I would recommend unit tests to be structured as follows:
  1. define config file
  2. execute request to retrieve sampling strategy for a service
  3. compare it with a pre-constructed response that is easy to read

Steps 1 and 3 are best described via JSON docs under /fixtures, or embedded JSON strings

plugin/sampling/strategystore/static/strategy_store_test.go
```diff
@@ -465,18 +467,44 @@ func TestAutoUpdateStrategyErrors(t *testing.T) {
...
 	require.NoError(t, err)
-	assert.Equal(t, 1.0, s.OperationSampling.DefaultSamplingProbability)
+	require.NotNil(t, strategy.OperationSampling)
+
+	assert.Equal(t, expectedServiceAResponse, strategy)
 
-	s, err = store.GetSamplingStrategy(context.Background(), "ServiceB")
+	// expected response for ServiceB (which has ratelimiting service level sampling strategy)
+	// shall contain /health operation with sampling probability taken from defaul operation level strategy
+	expectedServiceBResponse, errServiceB := makeServiceStrategyResponseFromFile("fixtures/ratelimiting_with_default_operation_strategy.json")
```
yurishkuro, line 487: did you forget to commit the new files?

kuujis: I indeed did forget two new files :(

kuujis:
> Sorry, it's been quite a while since this code was written, and I am not able to follow either your change or your unit test changes. I would recommend unit tests to be structured as follows:
> 
> 1. define config file
> 2. execute request to retrieve sampling strategy for a service
> 3. compare it with a pre-constructed response that is easy to read
> 
> Steps 1 and 3 are best described via JSON docs under /fixtures, or embedded JSON strings

After looking over the test - extracting expected response to separate file makes things a lot more readable :)

plugin/sampling/strategystore/static/fixtures/probablistic_with_default_operation_strategy.json
```diff
@@ -0,0 +1,14 @@
+{
+    "probabilisticSampling": {
+        "samplingRate": 1
+    },
+    "operationSampling": {
+        "defaultSamplingProbability": 1,
+        "perOperationStrategies": [
+            {
+                "operation": "/health",
+                "probabilisticSampling": {}
```
yurishkuro, line 10: This does not seem like a valid output, where is the rate?

This could be a side effect of JSON marshaling which has an option of omitting fields that have empty value (in this case prob=0). I suggest passing an argument to the marshaler to include empty fields.

plugin/sampling/strategystore/static/fixtures/probablistic_with_default_operation_strategy.json
```diff
@@ -0,0 +1,14 @@
+{
```
yurishkuro, line 1: I would recommend naming the file after the test name and some other dimension like serviceA.

plugin/sampling/strategystore/static/fixtures/probablistic_with_default_operation_strategy.json
kuujis: This is indeed the case. However - from what I see in the suggested fixes on the internet - the options are to either have ProbabilisticSamplingStrategy.SamplingRate declared as `*float64` or remove `omitempty` tag from the field, but as those are generated - does not look like it should be done. 

I was not able to find any property/parameter/option, which I could pass to the marshaler/unmarshaler to control this behavior (I might be missing something too, as this is literally the first time I write something in GO :) )

So took the easy way out and added non-zero values to the test, so now the operation level strategy sampling probability values are there and are verified. 

I also checked how the configuration is received on Java app side - it does have 0.0 probability if that is the value set in strategy configuration and accordingly sampling does not happen for the identified operation. So maybe its ok to leave the tests with 0.1?

plugin/sampling/strategystore/static/strategy_store_test.go
```diff
@@ -465,18 +467,44 @@ func TestAutoUpdateStrategyErrors(t *testing.T) {
...
 
 func TestServiceNoPerOperationStrategies(t *testing.T) {
+	// given setup of strategy store with no specific per operation sampling strategies
 	store, err := NewStrategyStore(Options{StrategiesFile: "fixtures/service_no_per_operation.json"}, zap.NewNop())
 	require.NoError(t, err)
 
-	s, err := store.GetSamplingStrategy(context.Background(), "ServiceA")
+	// expected response for ServiceA (which has probablistic service level sampling strategy)
+	// shall contain /health operation with default service level sampling strategy as well
+	expectedServiceAResponse, errServiceA := makeServiceStrategyResponseFromFile("fixtures/TestServiceNoPerOperationStrategies_response_for_ServiceA.json")
```
yurishkuro, line 476: nit: one of the useful techniques I used when dealing with fixtures for __expected__ outcomes (aka "snapshots") is to have an env var that allows the test to re-generate the snapshots (e.g. cmd/query/app/apiv3/gateway_test.go). Since you have relatively small outputs I don't think that specific mechanism is necessary, but the way snapshots are generated is by using the test name as the file name: https://github.com/jaegertracing/jaeger/blob/55ceba9467114d3bbbfad9c4bdfa27b96c28b1f5/cmd/query/app/apiv3/gateway_test.go#L80

You did it partially, but you still have a manually added suffix `response_for_ServiceA`. You can use just the test name if you create sub-tests within this function, e.g. 
```
t.Run("serviceA", func(t) {
  expectedFile := basePath + t.Name()
});
```

plugin/sampling/strategystore/static/strategy_store_test.go
```diff
@@ -465,18 +467,44 @@ func TestAutoUpdateStrategyErrors(t *testing.T) {
...
-	expected := makeResponse(api_v2.SamplingStrategyType_RATE_LIMITING, 3)
-	assert.Equal(t, *expected.RateLimitingSampling, *s.RateLimitingSampling)
+	assert.Equal(t, expectedServiceBResponse, strategy)
+}
+
+func makeServiceStrategyResponseFromFile(serviceStrategyFile string) (resp *api_v2.SamplingStrategyResponse, err error) {
+	strategyBytes, err := os.ReadFile(filepath.Clean(serviceStrategyFile))
+	if err != nil {
+		return nil, fmt.Errorf("failed to read SamplingStrategyResponse file %s: %w", serviceStrategyFile, err)
+	}
```
yurishkuro, line 501: a much cleaner and simpler approach is to pass `t` to this function and simply do `require.NoError(t, err)`

plugin/sampling/strategystore/static/strategy_store_test.go
```diff
@@ -465,18 +467,44 @@ func TestAutoUpdateStrategyErrors(t *testing.T) {
...
 
-	s, err = store.GetSamplingStrategy(context.Background(), "ServiceB")
+	// expected response for ServiceB (which has ratelimiting service level sampling strategy)
+	// shall contain /health operation with sampling probability taken from default operation level strategy
+	expectedServiceBResponse, errServiceB := makeServiceStrategyResponseFromFile("fixtures/TestServiceNoPerOperationStrategies_response_for_ServiceB.json")
+	require.NoError(t, errServiceB)
+
+	strategy, err = store.GetSamplingStrategy(context.Background(), "ServiceB")
 	require.NoError(t, err)
+	require.NotNil(t, strategy.OperationSampling)
```
yurishkuro, line 492: since you're comparing with a snapshot, do you need this assert?

plugin/sampling/strategystore/static/strategy_store_test.go
```diff
@@ -465,18 +467,44 @@ func TestAutoUpdateStrategyErrors(t *testing.T) {
...
 
-	expected := makeResponse(api_v2.SamplingStrategyType_RATE_LIMITING, 3)
-	assert.Equal(t, *expected.RateLimitingSampling, *s.RateLimitingSampling)
+	assert.Equal(t, expectedServiceBResponse, strategy)
+}
+
+func makeServiceStrategyResponseFromFile(serviceStrategyFile string) (resp *api_v2.SamplingStrategyResponse, err error) {
```
yurishkuro, line 497: I'd argue it's better to do comparison on JSON strings rather than objects. JSON comparison is more reliable, the diff output is more intuitive.

kuujis:
Thank you for the comments, I will address them, but I'm a bit swamped with life things, so this gets a bit of a back seat for now :(

plugin/sampling/strategystore/static/strategy_store_test.go
kuujis: Sounds like a way to be lazy about updating JSON's... I love it :D

plugin/sampling/strategystore/static/strategy_store_test.go
kuujis: That whole function was a mess and not needed if one focuses on string comparison instead, but I'll keep this in mind for future.

plugin/sampling/strategystore/static/strategy_store_test.go
kuujis: Nop :)

plugin/sampling/strategystore/static/strategy_store_test.go
kuujis: After playing around a bit - you are definitely right. Originally I was thinking of using baseline object loaded from file and modified by applying some parameters, but in retrospect - not worth it :)

plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -216,35 +219,42 @@ func (h *strategyStore) parseStrategies(strategies *strategies) {
...
+			continue
+		}
+
 		opS := newStore.serviceStrategies[s.Service].OperationSampling
 		if opS == nil {
-			if newStore.defaultStrategy.OperationSampling == nil ||
+
+			if !h.options.IncludeDefaultOpStrategies &&
 				newStore.serviceStrategies[s.Service].ProbabilisticSampling == nil {
+				h.logger.Warn("Default operations level strategies will not be included for this Ratelimiting service strategy." +
```
yurishkuro, line 238: it doesn't make sense to log this on every retrieval, I would move this check and the logging statement to constructor

plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -216,35 +219,42 @@ func (h *strategyStore) parseStrategies(strategies *strategies) {
...
+		if newStore.defaultStrategy.OperationSampling == nil {
+			// Default strategy doens't have them either, nothing to do.
+			continue
+		}
+
 		opS := newStore.serviceStrategies[s.Service].OperationSampling
 		if opS == nil {
-			if newStore.defaultStrategy.OperationSampling == nil ||
+
+			if !h.options.IncludeDefaultOpStrategies &&
```
yurishkuro, line 236: let's not mix up the logic in one place. I suggest copying the original function from main and renaming it xxx_deprecatedI(), so that when the time comes we can just delete it.

plugin/sampling/strategystore/static/strategy_store_test.go
```diff
@@ -464,19 +473,64 @@ func TestAutoUpdateStrategyErrors(t *testing.T) {
 	assert.Len(t, logs.FilterMessage("failed to update sampling strategies").All(), 2)
 }
 
-func TestServiceNoPerOperationStrategies(t *testing.T) {
-	store, err := NewStrategyStore(Options{StrategiesFile: "fixtures/service_no_per_operation.json"}, zap.NewNop())
+func TestServiceNoPerOperationStrategies5720ToggleOn(t *testing.T) {
```
yurishkuro, line 476: similar to the main function, I would prefer the test name to be "clean" from bug-fix ticket, and only the to-be-remove test to have a more complex name

plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -58,11 +60,20 @@ func NewStrategyStore(options Options, logger *zap.Logger) (ss.StrategyStore, er
...
+		options:    options,
 	}
 	h.storedStrategies.Store(defaultStrategies())
 
 	if options.StrategiesFile == "" {
-		h.parseStrategies(nil)
+
+		if !h.options.IncludeDefaultOpStrategies {
+			h.logger.Warn("Default operations level strategies will not be included for Ratelimiting service strategies." +
+				"This behavior will be deprecated in release 1.57.0 . See https://github.com/jaegertracing/jaeger/issues/5270 for more information.")
```
yurishkuro, line 71:
```suggestion
				"This behavior will be deprecated in future releases. See https://github.com/jaegertracing/jaeger/issues/5270 for more information.")
```

plugin/sampling/strategystore/static/options.go
```diff
@@ -33,17 +34,23 @@ type Options struct {
...
+	// strategies when calculating Ratelimiting type service level strategy
+	// more information https://github.com/jaegertracing/jaeger/issues/5270
+	IncludeDefaultOpStrategies bool
 }
 
 // AddFlags adds flags for Options
 func AddFlags(flagSet *flag.FlagSet) {
 	flagSet.Duration(samplingStrategiesReloadInterval, 0, "Reload interval to check and reload sampling strategies file. Zero value means no reloading")
 	flagSet.String(samplingStrategiesFile, "", "The path for the sampling strategies file in JSON format. See sampling documentation to see format of the file")
+	flagSet.Bool(samplingStrategiesBugfix5270, false, "Disabled by default. If enabled - will include default operation level strategies for Ratesampling type service level strategy. See https://github.com/jaegertracing/jaeger/issues/5270 for more information.")
```
yurishkuro, line 47:
```suggestion
	flagSet.Bool(samplingStrategiesBugfix5270, false, "Include default operation level strategies for Ratesampling type service level strategy. Cf. https://github.com/jaegertracing/jaeger/issues/5270")
```

plugin/sampling/strategystore/static/strategy_store.go
yurishkuro, line 71:
```suggestion
```
is this relevant if the user did not provide strategies file?

yurishkuro: I don't even know why we call parse function - all it does is log a warning and exit. Why don't we log that warning here and remove `if-nil` check from the parse functions?

plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -71,7 +82,13 @@ func NewStrategyStore(options Options, logger *zap.Logger) (ss.StrategyStore, er
 	if err != nil {
 		return nil, err
 	}
-	h.parseStrategies(strategies)
+	if !h.options.IncludeDefaultOpStrategies {
+		h.logger.Warn("Default operations level strategies will not be included for Ratelimiting service strategies." +
+			"This behavior will be deprecated in release 1.57.0 . See https://github.com/jaegertracing/jaeger/issues/5270 for more information.")
```
yurishkuro, line 87:
```suggestion
			"This behavior will be changed in future releases. " +
			"Cf. https://github.com/jaegertracing/jaeger/issues/5270")
```

plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -216,6 +233,49 @@ func (h *strategyStore) parseStrategies(strategies *strategies) {
...
+
+		// If the service did have its own per-operation strategies, then merge them with the default ones.
+		opS.PerOperationStrategies = mergePerOperationSamplingStrategies(
+			opS.PerOperationStrategies,
+			newStore.defaultStrategy.OperationSampling.PerOperationStrategies)
+	}
+	h.storedStrategies.Store(newStore)
+}
+
+func (h *strategyStore) parseStrategies_deprecated(strategies *strategies) {
```
yurishkuro, line 269: to minimize code changes diff, I would prefer if you renamed the original function to parseStrategies_deprecated (leaving it in the same place where it's defined now) and added the new function below. This would make the diff more readable and easier to see that the original function did not change. Not it all shows as changed / added, nothing to compare against.

plugin/sampling/strategystore/static/strategy_store.go
kuujis: Well, I think the check inside the parsing function was for the case, where URL is provided as a source, but its unavailable or returns no strategies. However - in this case I think there are two cases to handle: no source provided (i.e. `options.StrategiesFile` being empty) and no strategies being loaded from a provided source. I added different messages and clarified handling for both, hope this looks better.

yurishkuro (approved):
Thanks!

plugin/sampling/strategystore/static/strategy_store.go
```diff
@@ -249,6 +259,45 @@ func (h *strategyStore) parseStrategies(strategies *strategies) {
...
+			}
+			newStore.serviceStrategies[s.Service].OperationSampling = &newOpS
+			continue
+		}
+
+		// If the service did have its own per-operation strategies, then merge them with the default ones.
+		opS.PerOperationStrategies = mergePerOperationSamplingStrategies(
+			opS.PerOperationStrategies,
+			newStore.defaultStrategy.OperationSampling.PerOperationStrategies)
+	}
```
yurishkuro, line 297: the corresponding section of the code is being covered in the "deprecated" function, but not here. Maybe you can parameterize that test to toggle the new bugfix option so that both functions are exercised?

yurishkuro:
please revert ui submodule change, it's not related
```
git checkout main
git submodule update
git checkout your-branch
git commit -asm 'fix ui'
```

kuujis:
> please revert ui submodule change, it's not related
> 
> ```
> git checkout main
> git submodule update
> git checkout your-branch
> git commit -asm 'fix ui'
> ```

Done, sorry, should not have happened.

yurishkuro:
I edited description to NOT resolve the issue, to make sure we do the follow-ups.

yurishkuro (approved):

PR #5279, opened 2024-03-16
Remove archive storage factory 

### LINKED ISSUE #5230: [jaeger-v2] Do away with storage.ArchiveFactory
Per comment https://github.com/jaegertracing/jaeger/pull/5152/files#r1501916983

We need to get rid of `storage.ArchiveFactory` interface because it is conceptually meaningless - archive storage is just a storage that implements SpanStorage interface. It should be the same thing with Factory, just have a `storage.Factory` that in context can be interpreted as primary or archive. We unnecessarily bundle these two.

akagami-harsh:
- Simplify Storage: Remove ArchiveFactory, Use Storage.Factory

## Description of the changes
- Fixes #5230 

- some test are still failing working on fixing them

## Checklist


cmd/query/app/querysvc/query_service.go
```diff
@@ -138,12 +138,7 @@ func (qs QueryService) GetCapabilities() StorageCapabilities {
 
 // InitArchiveStorage tries to initialize archive storage reader/writer if storage factory supports them.
 func (opts *QueryServiceOptions) InitArchiveStorage(storageFactory storage.Factory, logger *zap.Logger) bool {
-	archiveFactory, ok := storageFactory.(storage.ArchiveFactory)
-	if !ok {
-		logger.Info("Archive storage not supported by the factory")
-		return false
-	}
-	reader, err := archiveFactory.CreateArchiveSpanReader()
+	reader, err := storageFactory.CreateSpanReader()
```
yurishkuro, line 141: This is not equivalent to previous behavior. Whoever calls this function is passing the primary storage factory, and you're just using it straight up.

yurishkuro:
@akagami-harsh I see you fixing tests, but the direction itself doesn't seem right to me (https://github.com/jaegertracing/jaeger/pull/5279#discussion_r1528655325). Can you elaborate?

akagami-harsh:
> @akagami-harsh I see you fixing tests, but the direction itself doesn't seem right to me ([#5279 (comment)](https://github.com/jaegertracing/jaeger/pull/5279#discussion_r1528655325)). Can you elaborate?

@yurishkuro , i am trying do to what you explained in the comment. first i removed the storage.archiveFactory interface and then trying to find a way to pass in two separate factories in the query-service for primary and archive.

yurishkuro:
I also said we should only do it in the scope of jaeger-v2, not do a major change in v1 storage

> If at all possible we should try to avoid changing anything in v1 storage interface, because it's not necessary for v2.

yurishkuro:
I think we can close this

PR #5283, opened 2024-03-18
[kafka-consumer] Add support for setting fetch message max bytes

sappusaketh:
- PR #5263 added support for setting max message size a producer can send that solved the issue on collector end now we are seeing our consumer getting stuck consuming messages larger that 1MB(default fetch max message bytes) so adding an option to set this value via flags

## Description of the changes
- adding an option to set fetch max message bytes value via flags

- Tested against our dev jaeger where we have spans larger than 1 MB

## Checklist


yurishkuro (approved):

yurishkuro:
thanks

PR #5288, opened 2024-03-21
[jaeger-v2] fix storage receiver consume traces query params

james-ryans:
- Fixes #5281 issue for not being able to consume traces

## Description of the changes
- Added `StartTimeMin` and `StartTimeMax` query params with 1h lookback to `consumeTraces` function.
- Added logging to the data receivers with the OTelCol codes below as the references.
  - https://github.com/open-telemetry/opentelemetry-collector/blob/main/otelcol/unmarshaler.go#L45
  - https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/telemetry/telemetry.go#L118

- Tested manually.

## Checklist


cmd/jaeger/internal/integration/datareceivers/jaegerstorage.go
```diff
@@ -37,19 +40,41 @@ func NewJaegerStorageDataReceiver(traceStorage string, storageConfig *jaegerstor
...
+		Encoding:          "console",
+		EncoderConfig:     zap.NewProductionEncoderConfig(),
+		OutputPaths:       []string{"stderr"},
+		ErrorOutputPaths:  []string{"stderr"},
+		DisableCaller:     false,
+		DisableStacktrace: false,
+		InitialFields:     map[string]any(nil),
+	}
+	zapCfg.EncoderConfig.EncodeTime = zapcore.ISO8601TimeEncoder
+	logger, err := zapCfg.Build()
```
yurishkuro, line 55: I think there's a much simpler way, something like `zap.NewDevelopment()`

yurishkuro: also, it's usually a bad pattern for a component to initialize telemetry, telemetry objects should be injected. The production OTEL components receive TelemetrySettings, which I think you also initialize somewhere at the start of the integration test, so the logger should be set up there.

james-ryans: I see, I've refactored it. Actually, the TelemetrySettings is set from the config file at `service.telemetry.logs`, so I've refactored the data receiver to get that setting from the config file and injected the logger.

cmd/jaeger/internal/integration/integration.go
```diff
@@ -46,13 +52,20 @@ func (s *StorageIntegration) newDataReceiver(t *testing.T, factories otelcol.Fac
 	cfg, err := configProvider.Get(context.Background(), factories)
 	require.NoError(t, err)
 
+	tel, err := telemetry.New(context.Background(), telemetry.Settings{}, cfg.Service.Telemetry)
```
yurishkuro, line 55: Maybe I'm missing something. You're passing logger to the function, but not using it here. Then in l65 you're reading logger from tel - how is it initialized there?

And what does cfg.Service.Telemetry do? There's no config file changes in this PR.

james-ryans: Oops.. sorry, I've caused some confusion here. Previously, I tried to initialize the `zap.NewDevelopment()` and pass it to the data receiver but then I realized that it would be better if I used the similar logic with the OTEL collector by retrieving the log settings from the config file ([open-telemetry/opentelemetry-collector/otelcol/collector.go#L187](https://github.com/open-telemetry/opentelemetry-collector/blob/15201f1e5967fd1bc1e51c118178ab109170fe07/otelcol/collector.go#L187) -> [open-telemetry/opentelemetry-collector/service/service.go#L90](https://github.com/open-telemetry/opentelemetry-collector/blob/15201f1e5967fd1bc1e51c118178ab109170fe07/service/service.go#L90)). I'll delete the logger passed to the function.

`cfg.Service.Telemetry` is the struct that contains telemetry settings, I don't change the config file because OTEL already has the default values for this ([open-telemetry/opentelemetry-collector/otelcol/unmarshaler.go#L45](https://github.com/open-telemetry/opentelemetry-collector/blob/main/otelcol/unmarshaler.go#L45)).

My intention by implementing this `telemetry.New()` is to have the exact logger settings with the OTEL collector instead of initializing a new `zap.NewDevelopment()`

cmd/jaeger/internal/integration/datareceivers/jaegerstorage.go
```diff
@@ -37,19 +45,24 @@ func NewJaegerStorageDataReceiver(traceStorage string, storageConfig *jaegerstor
 func (dr *jaegerStorageDataReceiver) Start(tc consumer.Traces, _ consumer.Metrics, _ consumer.Logs) error {
 	ctx := context.Background()
 
-	extFactory := jaegerstorage.NewFactory()
-	ext, err := extFactory.CreateExtension(ctx, extension.CreateSettings{
+	extSet := extension.CreateSettings{
+		ID:                jaegerstorage.ID,
 		TelemetrySettings: componenttest.NewNopTelemetrySettings(),
```
yurishkuro, line 50: I see you still create telemetry settings in this component. My suggestion is that you get them as constructor argument, instead of only getting a logger, and you create the whole telemetry settings in integration.go

yurishkuro (commented):
Did you also mean to add any additional logging in the datareceiver? Or does it do the logging already and you just weren't giving it a good logger?

james-ryans:
> Did you also mean to add any additional logging in the datareceiver? Or does it do the logging already and you just weren't giving it a good logger?

Yes, it already has the logging there. This is where the logs have captured the error from Badger's integration test. https://github.com/jaegertracing/jaeger/blob/d887b74a504a8e25e1c4b36661ca42f4050edf5b/cmd/jaeger/internal/integration/receivers/storagereceiver/receiver.go#L74-L78

cmd/jaeger/internal/integration/integration.go
```diff
@@ -46,13 +48,23 @@ func (s *StorageIntegration) newDataReceiver(t *testing.T, factories otelcol.Fac
...
 
 	exporterCfg, ok := cfg.Exporters[storageexporter.ID].(*storageexporter.Config)
 	require.True(t, ok, "no jaeger storage exporter found in the config")
 
-	receiver := datareceivers.NewJaegerStorageDataReceiver(exporterCfg.TraceStorage, storageCfg)
+	telemetrySettings := componenttest.NewNopTelemetrySettings()
+	telemetrySettings.Logger = tel.Logger()
+
+	receiver := datareceivers.NewJaegerStorageDataReceiver(
+		telemetrySettings,
```
yurishkuro, line 64: Any reason why can't pass tel directly?

james-ryans: Because tel (or `telemetry.Telemetry`) and `component.TelemetrySettings` is two different component.
[telemetry.Telemetry](https://github.com/open-telemetry/opentelemetry-collector/blob/15201f1e5967fd1bc1e51c118178ab109170fe07/service/telemetry/telemetry.go#L61) initialize the `zap.Logger` from the config.
[component.TelemetrySettings](https://github.com/open-telemetry/opentelemetry-collector/blob/15201f1e5967fd1bc1e51c118178ab109170fe07/component/telemetry.go#L20) is the struct that contains `zap.Logger` that is used by components in OTEL to report telemetry matters.

This means that the extension and receiver in the data receiver are using `component.TelemetrySettings` to write logs. Based on your suggestion, I thought it is better if we pass the initialized struct and use it directly in the data receiver.

yurishkuro: Got it. Looking through OTEL code the situation seems to be quite messy, I don't know if they have a ticket to clean it up. E.g.
  * why are `Telemetry` and `TelemetrySettings` two different structs? They seem to server the same purpose (the latter being more comprehensive)
  * `TelemetrySettings` comment also mentions yet another similar struct `servicetelemetry.TelemetrySettings`

yurishkuro (approved):

yurishkuro:
> Yes, it already has the logging there. This is where the logs have captured the error from Badger's integration test.

I was rather thinking of INFO logs as the data receiver polls the storage - it's a test scenario, we don't have to be stingy about logs. E.g. something like
  * querying for traces, params: ...
  * retrieved N spans
  * sleeping till the next retry

PR #5306, opened 2024-03-28
[jaeger-v2] Add GRPC Unit Test

Pushkarm029:
- Part of #5254

## Description of the changes
- Resolves point 1 of the [requirements](https://github.com/jaegertracing/jaeger/issues/5254#issuecomment-2016858385).
- Utilizes existing configuration files located in `./cmd/jaeger` to initialize the storage extension.
- Utilizes the storage extension to obtain a new storage factory.
- Initializes necessary components such as spanReader, spanWriter, etc.
- Executes test cases (located in plugin/storage/integration/integration.go) on remote storage utilizing these components.
- done some refactoring to reuse multiple code segments from the existing testbed based integration test.

### existing integration tests
- directly, initializes `grpc.Factory`.

### unit tests
- starts storage extension. based on the config provided, currently in this test, we are reusing configs present in `./cmd/jaeger`.
- initializes `storage.Factory` using storage extension. which internally initializes `grpc.Factory` to create necessary components like spanReader, spanWriter.

- `./scripts/grpc-unit-test.sh latest`

## Checklist


Pushkarm029:
If this fulfils our 1st requirement (to keep integration tests operating as unit tests). Then, I will start working on the e2e test.

yurishkuro:
not sure I follow. How is this different from `plugin/storage/integration/grpc_test.go`?

Pushkarm029:
> not sure I follow. How is this different from `plugin/storage/integration/grpc_test.go`?

I copied it from there to make sure that when we add more tests, such as e2e, on top of it, they won't affect the v1 integration tests. Basically, to separate v1 and v2 tests. However, if we want to keep them in the same place, we can do so.

yurishkuro:
There is no reason to duplicate the code. The existing integration tests are specific to Jaeger overall, in how it uses storage backends, they are not specific to v1 or v2 architecture. The real task is this one:

> abstract how integration tests write and read span data

I added diagrams https://github.com/jaegertracing/jaeger/issues/5254#issuecomment-2025604602

Makefile
```diff
@@ -5,6 +5,7 @@ SHELL := /bin/bash
 JAEGER_IMPORT_PATH = github.com/jaegertracing/jaeger
 STORAGE_PKGS = ./plugin/storage/integration/...
 JAEGER_STORAGE_PKGS = ./cmd/jaeger/internal/integration
+JAEGER_UNIT_STORAGE_PKGS = ./cmd/jaeger/internal/unittest
```
yurishkuro, line 8: Please rebase on main, this is already in the makefile

yurishkuro: why not put this into ./cmd/jaeger/internal/integration?

cmd/jaeger/internal/unittest/unittest.go
```diff
@@ -0,0 +1,115 @@
...
+	Name       string
+	ConfigFile string
+}
+
+func (s *StorageTest) Test(t *testing.T) {
+	if os.Getenv("STORAGE") != s.Name {
+		t.Skipf("Integration test against Jaeger-V2 %[1]s skipped; set STORAGE env var to %[1]s to run this", s.Name)
+	}
+	var v integration.StorageIntegration
+	factories, err := internal.Components()
```
yurishkuro, line 41: @james-ryans if feels like there is a lot of repetition here compared to `cmd/jaeger/internal/integration/integration.go`.

plugin/storage/integration/integration.go
```diff
@@ -532,7 +532,7 @@ func (s *StorageIntegration) RunAll(t *testing.T) {
 	t.Run("ArchiveTrace", s.testArchiveTrace)
 	t.Run("GetOperations", s.testGetOperations)
 	t.Run("GetTrace", s.testGetTrace)
-	t.Run("GetLargeSpans", s.testGetLargeSpan)
+	t.Run("GetLargeSpans", s.TestGetLargeSpan)
```
yurishkuro, line 535: the intention is that the other tests would call RunAll, not cherry-pick specific tests

scripts/grpc-unit-test.sh
```diff
@@ -0,0 +1,105 @@
+#!/bin/bash
```
yurishkuro, line 1: How is this different from `scripts/grpc-integration-test.sh`? Our goal is to **simplify** and **consolidate** code, not duplicate it.

cmd/jaeger/internal/integration/integration.go
```diff
@@ -130,3 +129,49 @@ func (s *StorageIntegration) Test(t *testing.T) {
...
+	require.NoError(t, err)
+	v := &integration.StorageIntegration{
+		SpanReader:        spanReader,
+		SpanWriter:        spanWriter,
+		ArchiveSpanReader: archiveSpanReader,
+		ArchiveSpanWriter: archiveSpanWriter,
+		Refresh:           func(_ *testing.T) {},
+		CleanUp:           func(_ *testing.T) {},
+	}
+	v.RunAll(t)
```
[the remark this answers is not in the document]
yurishkuro: add debug logging to see what's going on. When using memory storage it's not difficult to troubleshoot.

cmd/jaeger/internal/integration/integration.go
```diff
@@ -130,3 +129,63 @@ func (s *StorageIntegration) Test(t *testing.T) {
...
+	s.tests.Refresh = func(_ *testing.T) {}
+	s.tests.CleanUp = s.cleanUp
+}
+
+func (s *StorageIntegration) cleanUp(t *testing.T) {
+	require.NoError(t, s.storageExtension.Shutdown(context.Background()))
+
+	time.Sleep(20 * time.Second)
+	s.initComponents(t)
+}
```
Pushkarm029, line 191: i think the issue is with the cleanup process. In the existing integration test, we simply close the factory and initialize it again. However, here, we don't have anything similar to close factory, so to tackle this, I tried the following approach -> in cleanup function
- i shutted down the storage extension and then restarted it again. got this error,
```bash
Error:          Received unexpected error:
                duplicate grpc storage name external-storage

plugin error: rpc error: code = Canceled desc = grpc: the client connection is closing

```

cmd/jaeger/internal/extension/jaegerstorage/extension_test.go
```diff
@@ -220,7 +220,7 @@ func makeStorageExtenion(t *testing.T, config *Config) component.Component {
 	return storageExtension
 }
 
-func startStorageExtension(t *testing.T, memstoreName string) component.Component {
+func StartStorageExtension(t *testing.T, memstoreName string) component.Component {
```
yurishkuro, line 223: is this change necessary? I don't see this function being called from outside of this package (and if it were it would be a code smell, we don't share test code like that)

cmd/jaeger/internal/integration/integration.go
```diff
@@ -11,30 +11,23 @@ import (
...
-}
-
-// The data receiver will utilize the artificial jaeger receiver to pull
-// the traces from the database which is through jaeger storage extension.
-// Because of that, we need to host another jaeger storage extension
-// that is a duplication from the collector's extension. And get
-// the exporter TraceStorage name to set it to receiver TraceStorage.
-func (s *StorageIntegration) newDataReceiver(t *testing.T, factories otelcol.Factories) testbed.DataReceiver {
-	fmp := fileprovider.NewWithSettings(confmap.ProviderSettings{})
+func (s *StorageIntegration) initTelemetry(t *testing.T, factories otelcol.Factories) (*jaegerstorage.Config, *storageexporter.Config, component.TelemetrySettings) {
```
yurishkuro, line 29: why does `initTelemetry` method return storage / exporter configs?

yurishkuro:
I am really confused by this PR. Please explain in the description what it's trying to achieve. E.g. it says "A test has been added to perform test cases (present in plugin/storage/integration/integration.go) on remote storage by calling the storage API." - we already do that in plugin/storage/integration/grpc_test, what is different here? Explain the differences.

Pushkarm029:
> I am really confused by this PR. Please explain in the description what it's trying to achieve. E.g. it says "A test has been added to perform test cases (present in plugin/storage/integration/integration.go) on remote storage by calling the storage API." - we already do that in plugin/storage/integration/grpc_test, what is different here? Explain the differences.

I apologize, I have just added it now. please check it.

cmd/jaeger/internal/integration/integration.go
```diff
@@ -130,3 +129,63 @@ func (s *StorageIntegration) Test(t *testing.T) {
...
 }
+
+func (s *StorageIntegration) UnitTest(t *testing.T) {
+	if os.Getenv("STORAGE") != s.Name {
+		t.Skipf("Integration test against Jaeger-V2 %[1]s skipped; set STORAGE env var to %[1]s to run this", s.Name)
+	}
+	factories, err := internal.Components()
+	require.NoError(t, err)
+
+	storageCfg, exporterCfg, telemetrySettings := s.initTelemetry(t, factories)
```
james-ryans, line 140: Looks like you're trying to get the configuration from the `./cmd/jaeger` config file, it isn't necessary because this PR is a unit test mode and the scopes are only to solve (2), (3), and (5). The (4) requirement (use config from `./cmd/jaeger`) can be covered by e2e test mode.

You're allowed to initialize the jaeger storage manually with your own defined config, which might be more readable. Also, this initialization step should be clearer if it is moved to `initComponents` too, then you'll not need to pass the `storageExtension` and `exporterCfg` through the struct.

yurishkuro: I agree, and if we ignore the config file part because it will be exercised by e2e test, then I still don't see what else this PR is testing that is not already tested by the existing integration tests. I showed two diagrams in the ticket - this PR doesn't fit either of them.

james-ryans: Oh, I thought the second point in the ticket stated that "unit mode" and "e2e mode" refer to v2's storage integration tests, so we need to create both "unit" and "e2e" tests for the v2 storage backend. This PR represents the unit mode for that, by reading/writing span data through the Jaeger storage extension.
The second point of the ticket states:
> we can abstract how integration tests write and read span data, such that in the unit test mode they would call storage API as a library, but in e2e mode they will do the same via RPC requests.

After giving it some thought, this PR only gives an additional benefit to solve point (5) that "generate code coverage for some parts of the code that do not get exercised in unit tests", and perhaps also indirectly solves (2) and (3) since this extension is somewhat a wrapper to storage API. But I realized maybe what you meant by unit mode is exactly the existing integration tests (?), since points (2) and (3) are already solved by them and point (5) will be covered by e2e mode.

So, do you think we still need to add the existing integration test to test through the Jaeger storage extension? Not through the whole collector pipeline nor directly to the storage API.

yurishkuro: > do you think we still need to add the existing integration test to test through the Jaeger storage extension?

The logic in the storage extension can be unit-tested directly, because all the extension does is delegate to factories. It already has 100% coverage:
```
$ go test -cover ./cmd/jaeger/internal/extension/jaegerstorage/
ok  	github.com/jaegertracing/jaeger/cmd/jaeger/internal/extension/jaegerstorage	6.055s	coverage: 100.0% of statements
```

So yes, I did mean we should reuse the existing integration tests. My comment said:
> Our storage integration tests solve (2), (3) and (5), and I think (6)

My proposal was to extend (not duplicate!) those integration tests to also enable e2e testing, by replacing direct function calls to storage API with RPC calls to the separately running collector/query pipeline. And I think it's preferable if that separately running pipeline is actually running the real jaeger-v2 binary, instead of being orchestrated in-process by the unit tests themselves, because then it wouldn't be an e2e test.

Pushkarm029:
so, there were some misunderstandings. this pr does not solve anything.

PR #5310, opened 2024-03-29
Fix goroutine leaks in plugin/sampling/strategystore/adaptive

WillSewell:

## Description of the changes
- This mainly involved ensuring that all goroutines started by the Processor are shut down in a Close method (which also blocks on them returning via a WaitGroup).
- Adding this flagged an issue where the `runUpdateProbabilitiesLoop` had a long delay, so tests need to be able to override the default Processor.followerRefreshInterval, or `Close` would take up to ~20s to return. More context on this here https://github.com/jaegertracing/jaeger/issues/5006#issuecomment-2027292668 - specifically regarding how to override this in the `Factory`.
- I also needed to fix a deadlock where a test was not unlocking a lock which would allow the Close method to return: https://github.com/jaegertracing/jaeger/pull/5310/files#diff-c9d6c33e36122502911585c65618d1af863079a70a68543adcc0ae2798faa348R468

- `make test lint`

## Checklist

### DIFF
--- a/cmd/collector/app/sampling/strategystore/interface.go
+++ b/cmd/collector/app/sampling/strategystore/interface.go
@@ -24,6 +24,9 @@ import (
 
 // StrategyStore keeps track of service specific sampling strategies.
 type StrategyStore interface {
+	// Close() from io.Closer  stops the processor from calculating probabilities.
+	io.Closer
+
 	// GetSamplingStrategy retrieves the sampling strategy for the specified service.
 	GetSamplingStrategy(ctx context.Context, serviceName string) (*api_v2.SamplingStrategyResponse, error)
 }
--- a/plugin/sampling/strategystore/adaptive/processor_test.go
+++ b/plugin/sampling/strategystore/adaptive/processor_test.go
@@ -359,6 +359,7 @@ func TestRunCalculationLoop(t *testing.T) {
 	}
 	p, err := newProcessor(cfg, "host", mockStorage, mockEP, metrics.NullFactory, logger)
 	require.NoError(t, err)
+	p.followerRefreshInterval = time.Millisecond
 	p.Start()
 
 	for i := 0; i < 1000; i++ {
@@ -392,9 +393,17 @@ func TestRunCalculationLoop_GetThroughputError(t *testing.T) {
 	}
 	p, err := newProcessor(cfg, "host", mockStorage, mockEP, metrics.NullFactory, logger)
 	require.NoError(t, err)
+	p.followerRefreshInterval = time.Millisecond
 	p.shutdown = make(chan struct{})
-	defer close(p.shutdown)
-	go p.runCalculationLoop()
+	done := make(chan struct{})
+	go func() {
+		p.runCalculationLoop()
+		close(done)
+	}()
+	defer func() {
+		close(p.shutdown)
+		<-done
+	}()
 
 	for i := 0; i < 1000; i++ {
 		// match logs specific to getThroughputErrMsg. We expect to see more than 2, once during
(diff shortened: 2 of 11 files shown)


yurishkuro:
Haven't looked at the code yet, but from the description it sounds like the exit is simply not properly implemented if you have to wait for 20sec. If there is a loop with a timer, probably blocking in select, the good approach is to add another "stop" channel that Close function can close and it would cause an exit from select.

WillSewell:
Ah yes, that would be better. I'll have a go at reworking it.

plugin/sampling/strategystore/adaptive/processor_test.go
```diff
@@ -393,8 +393,15 @@ func TestRunCalculationLoop_GetThroughputError(t *testing.T) {
...
-	defer close(p.shutdown)
-	go p.runCalculationLoop()
+	done := make(chan struct{})
+	go func() {
+		p.runCalculationLoop()
+		close(done)
+	}()
+	defer func() {
+		close(p.shutdown)
+		<-done
```
yurishkuro, line 403: do we really need `done` sync here? as long as goroutines are not blocked they will exit and goleak will tolerate that, it does not expect immediate clean state when it's called.

cmd/collector/app/sampling/strategystore/interface.go
```diff
@@ -24,6 +24,9 @@ import (
 
 // StrategyStore keeps track of service specific sampling strategies.
 type StrategyStore interface {
+	// Close() from io.Closer  stops the processor from calculating probabilities.
+	io.Closer
```
yurishkuro, line 28: You added this to the interface, but I am not seeing any non-test code that's actually calling it.

Adding Closer to interface it often contentious, some people argue that if you create an object via `NewX() *X`, you already have the ability to call Close on it without adding Close function to the interface that X implements. This doesn't work well when factories are involved since the factory does return an interface, not an actual struct. One other workaround to that is doing a runtime check for `io.Closer` interface and only then calling close - this is why I am asking about prod code calling it.

I'm ok to keep io.Closer in the interface because both real implementations are now closable (static store used to not have close before we added file watcher to it)

plugin/sampling/strategystore/adaptive/processor_test.go
WillSewell: Ah thanks for flagging this. They are not required (test pass without it). I was also misunderstanding the goleak semantics.

I think the issue I was having is before https://github.com/jaegertracing/jaeger/pull/5310/commits/dfe7adc7fd34389e2e507d9512eb64b5e19048f7, these tests were failing because a goroutine could be blocked in time.Sleep.

I'll remove this.

cmd/collector/app/sampling/strategystore/interface.go
WillSewell: Yes I think that makes sense.

> This doesn't work well when factories are involved since the factory does return an interface, not an actual struct.

Is there a fundamental reason why factories shouldn't return a struct instead of an interface? (Other than it being a breaking change to make in this instance).

> One other workaround to that is doing a runtime check for io.Closer interface and only then calling close - this is why I am asking about prod code calling it.

Prod code is not calling `Close` - do you have a preference between the current implementation vs the runtime check in tests? I don't feel strongly.

yurishkuro: > Is there a fundamental reason why factories shouldn't return a struct instead of an interface? 

Yes - polymorphism. The whole point of a factory is to abstract what underlying implementation it creates, which means it always returns an interface.

> Prod code is not calling Close

I actually think our pattern is that the main code only calls factory.Close() and the factory is generally responsible for releasing any resources. E.g. we don't call Close on SpanReader that we obtain from the factory.

yurishkuro (approved):

PR #5311, opened 2024-03-29
Fix goroutine leaks raised by plugin/storage/integration/cassandra_test.go

WillSewell:

## Description of the changes
- Fixes goroutine leaks in the shutdown of cassandra Factory
- It does not add the goleak detection because there are still failures in this package related to other components (e.g. grpc, elasticsearch)

- Add a goleak detection (i.e. this file https://github.com/jaegertracing/jaeger/pull/5311/commits/105a4d76e584c50e4260527519eb57fcc61ab521)
- `STORAGE=cassandra go test -v  ./plugin/storage/integration`

## Checklist


yurishkuro (approved):

yurishkuro:
I think `make test lint` doesn't actually run integration tests, so in real CI they are failing on leaks.

plugin/storage/integration/cassandra_test.go
```diff
@@ -81,28 +82,28 @@ func (s *CassandraStorageIntegration) initializeCassandraFactory(flags []string)
 	return f, nil
```
yurishkuro, line 82:
```suggestion
	t.Cleanup(func() {
		require.NoError(t, f.Close())
	})
	return f, nil
```

plugin/storage/integration/cassandra_test.go
```diff
@@ -81,28 +82,28 @@ func (s *CassandraStorageIntegration) initializeCassandraFactory(flags []string)
...
 }
 
-func (s *CassandraStorageIntegration) initializeCassandra() error {
+func (s *CassandraStorageIntegration) initializeCassandra() (io.Closer, error) {
```
yurishkuro, line 85: instead, add a cleanup lambda in the previous function.

yurishkuro (commented):
please rebase, lots of function signatures have changes

plugin/storage/cassandra/factory.go
```diff
@@ -232,6 +232,13 @@ var _ io.Closer = (*Factory)(nil)
...
 // Close closes the resources held by the factory
 func (f *Factory) Close() error {
+	if f.primarySession != nil {
+		f.primarySession.Close()
+	}
+	if f.archiveSession != nil {
+		f.archiveSession.Close()
+	}
+
 	f.Options.Get(archiveStorageConfig)
```
yurishkuro, line 242: looks like this line is not needed

plugin/storage/cassandra/factory.go
```diff
@@ -232,6 +232,13 @@ var _ io.Closer = (*Factory)(nil)
 
 // Close closes the resources held by the factory
 func (f *Factory) Close() error {
+	if f.primarySession != nil {
+		f.primarySession.Close()
+	}
```
yurishkuro, line 237:
```suggestion
	var errs []error
	if f.primarySession != nil {
		errs = append(errs, f.primarySession.Close())
	}
```
and at the end return `errors.join(errs...)`

plugin/storage/integration/cassandra_test.go
```diff
@@ -125,6 +126,8 @@ func TestCassandraStorage(t *testing.T) {
 		t.Skip("Integration test against Cassandra skipped; set STORAGE env var to cassandra to run this")
 	}
 	s := newCassandraStorageIntegration()
-	require.NoError(t, s.initializeCassandra())
+	closer, err := s.initializeCassandra()
+	require.NoError(t, err)
 	s.IntegrationTestAll(t)
```
yurishkuro, line 131: this has been renamed to RunAll, you must have stale branch

plugin/storage/integration/cassandra_test.go
yurishkuro: > this gets called on each subtest

do you mean Cleanup of the parent `t` gets call for each of the child `t`s? I would've thought they would be independent, and all actual tests are run as `t.Run()` on the parent `t`

WillSewell: Yes - I think because `Cleanup` is called _within_ each subtest ([example](https://github.com/jaegertracing/jaeger/blob/ea53d8e6ce6e0c40197dd2cdc6ffadff152f78ac/plugin/storage/integration/integration.go#L150)).

Here is a simplified version of what is happening:

```golang
func TestCleanup(t *testing.T) {
	t.Log("Start of test")
	for _, test := range []string{"one", "two"} {
		t.Run(test, func(t *testing.T) {
			t.Cleanup(func() {
				t.Log("Running cleanup")
			})
			t.Log("Running test", test)
		})
	}
	t.Log("End of test")
}
```

```
--- PASS: TestDummy (0.00s)
=== RUN   TestCleanup
    example_test.go:6: Start of test
=== RUN   TestCleanup/one
    example_test.go:12: Running test one
    example_test.go:10: Running cleanup
=== RUN   TestCleanup/two
    example_test.go:12: Running test two
    example_test.go:10: Running cleanup
=== NAME  TestCleanup
    example_test.go:15: End of test
--- PASS: TestCleanup (0.00s)
    --- PASS: TestCleanup/one (0.00s)
    --- PASS: TestCleanup/two (0.00s)
```

plugin/storage/cassandra/factory.go
WillSewell: The `Close` function in this case does not return an error - I think the concrete implementation in this case comes from here: https://github.com/gocql/gocql/blob/1c9ec8f651a2f47e2ac7d7b7e5030b1a6d7a2575/session.go#L463.

Having said that there is existing code that further down which swallows an error, so I can address that.

plugin/storage/integration/cassandra_test.go
yurishkuro: `t.Cleanup` is called in the context of `t` where it's added. If you add a cleanup at the top of your test, it will run only once, whereas the one that's called inside sub-test function is indeed called for each subtest

WillSewell: 🤦‍♂️ sorry I overlooked your original suggestion

WillSewell: Address in https://github.com/jaegertracing/jaeger/pull/5311/commits/0970720bdd8b21c2c8f51876776aca8ca2e477a7

plugin/storage/integration/cassandra_test.go
```diff
@@ -109,7 +112,7 @@ func (s *CassandraStorageIntegration) initializeDependencyReaderAndWriter(t *tes
 }
 
 func TestCassandraStorage(t *testing.T) {
-	skipUnlessEnv(t, "cassandra")
+	// skipUnlessEnv(t, "cassandra")
```
yurishkuro, line 115: need to revert this

yurishkuro (approved):

WillSewell:
@yurishkuro this has stalled because I got bogged down fixing various other goroutine leaks that goleak detects in this package.

I fixed the ones in the grpc tests, but lost momentum on the ones in the elasticsearch tests because at least one leak is actually originating in the `github.com/olivere/elastic` library.

So my thinking is rather than a single big PR that fixes all leaks (and risks not getting done at all), I'll instead aim to do some smaller PRs to cover the things I have managed to fixed. No guarantees whether I'll get to the elasticsearch fixes...

PR #5313, opened 2024-03-30
[es cit] Use storage factory instead of manual creation

Wise-Wizard:

This PR addresses a new found issue which is a part of the issue [#5203 ](https://github.com/jaegertracing/jaeger/issues/5203).
It initializes ES Factory in the Integration Test Suite.

Description:
This PR introduces changes to the integration tests by initializing an ElasticSearch factory (es.Factory) and utilizing it for the creation of span readers and writers. Previously, the code instantiated a new span reader/writer from the SpanStore directly, which has been replaced with the more appropriate approach of leveraging the ElasticSearch factory.

```bash
make test
```

**Checklist**


Wise-Wizard:
Hi @yurishkuro, Can you please guide me on how to test these Github Workflows locally? I have to push everytime I make changes to see if it is right or not.

yurishkuro:
You can check the code for the workflows in .github dir. usually we keep them very simple, all of the test business logic is in the script that the workflow runs, precisely to allow local debugging. You can run the script manually.

Wise-Wizard:
> You can check the code for the workflows in .github dir. usually we keep them very simple, all of the test business logic is in the script that the workflow runs, precisely to allow local debugging. You can run the script manually.

Understood Thanks!
Since, I am using WSL I will have to install "act" and run a specific job using act -j right? Or, is there a simpler way?

yurishkuro:
don't know what `act` is - you should only need Docker

Wise-Wizard:
> don't know what `act` is - you should only need Docker

While googling how to run github actions locally, almost 10-12 sites used a github repo nektos/act to run the actions locally. I already am using docker, will look into it further. Thanks for your guidance.

yurishkuro:
you don not need to run actions locally. Actions just call a shell script, you can run the shell script locally. This is the only line relevant to running ES test locally: https://github.com/jaegertracing/jaeger/blob/bae96f7c0f1177978a9eb47fd43d1ba403d945e6/.github/workflows/ci-elasticsearch.yml#L61

yurishkuro:
please rebase, there were changes in the signatures

Wise-Wizard:
> you don not need to run actions locally. Actions just call a shell script, you can run the shell script locally. This is the only line relevant to running ES test locally:
> 
> https://github.com/jaegertracing/jaeger/blob/bae96f7c0f1177978a9eb47fd43d1ba403d945e6/.github/workflows/ci-elasticsearch.yml#L61
Got it, Thanks a lot!

Wise-Wizard:
`func (f *Factory) CreateArchiveSpanReader() (spanstore.Reader, error) {
	if !f.archiveConfig.Enabled {
		return nil, nil
	}
	return createSpanReader(f.getArchiveClient, f.archiveConfig, true, f.metricsFactory, f.logger, f.tracer)
}

// CreateArchiveSpanWriter implements storage.ArchiveFactory
func (f *Factory) CreateArchiveSpanWriter() (spanstore.Writer, error) {
	if !f.archiveConfig.Enabled {
		return nil, nil
	}
	return createSpanWriter(f.getArchiveClient, f.archiveConfig, true, f.metricsFactory, f.logger)
}`

Wise-Wizard:
`f.archiveCOnfig`

Wise-Wizard:
```if !f.archiveConfig.Enabled {
		return nil, nil
	}```

File: plugin/storage/integration/elasticsearch_test.go [hunk 1]
```diff
@@ -105,14 +104,31 @@ func (s *ESStorageIntegration) initializeES(t *testing.T, allTagsAsFields bool)
...
 		Addresses:            []string{queryURL},
 		DiscoverNodesOnStart: false,
 	})
 	require.NoError(t, err)
 
+	// Initialize ES Factory
+	f := es.NewFactory()
+	v, _ := config.Viperize(f.AddFlags)
+	f.InitFromViper(v, zap.NewNop())
+	err = f.Initialize(metrics.NullFactory, s.logger)
```
yurishkuro, line 118: maybe you can use InitFromConfig instead of CLI-based functions. You would need to set some defaults, however.

Wise-Wizard: Yes, I was thinking about doing the same thing..
Use NewFactoryWithConfig to initialise instead, will do that!
Thanks a lot

Wise-Wizard: Hi, Can you please review and lmk if I am on the right path?

plugin/storage/integration/elasticsearch_test.go
```diff
@@ -105,14 +102,40 @@ func (s *ESStorageIntegration) initializeES(t *testing.T, allTagsAsFields bool)
...
-
 	s.client = rawClient
 	s.v8Client, err = elasticsearch8.NewClient(elasticsearch8.Config{
 		Addresses:            []string{queryURL},
 		DiscoverNodesOnStart: false,
 	})
 	require.NoError(t, err)
 
+	// Initialize ES Factory
+	cfg := es.NewFactory().Options.Primary.Configuration
```
yurishkuro, line 113: why do you need to go through NewFactory? The factory is leaked this way. What if you just instantiate the config struct directly? Are you relying on some initialization happening in NewFactory?

yurishkuro (commented):
you need to rebase from main, your branch is stale

plugin/storage/integration/elasticsearch_test.go
Wise-Wizard: I did initially initialise config struct directly in the below manner. Many tests were failing, so I presumed the default initialization would not have been occurring, that's why thought of initializing using NewFactory, and change the Fields manually.

Wise-Wizard: Done! Didn't need the NewFactory, initialized it using NewOptions struct.

Wise-Wizard: Hi, I am not able to debug why the GetOperations test is failing.
I have went through the initialisation of GetOperations and used Debugger, but "Actual" is always being initialised as nil.
Is it due to a difference in the Client Fn?
I would really appreciate your insight..

yurishkuro: My recommendation is to add logging of the query that we send to ES (a JSON). You can then run this query directly against ES and troubleshoot why it does not return the data.

Wise-Wizard: Yes, I did that. The Example_Trafe query is being sent to the loadParseandWrite method successfully, I e writing the trace to ES but it's failing (Returning nil) while fetching it from the SpanReader

yurishkuro: I was referring to logging reads, not writes. I see that ES client has this logger that is used to log requests:
```
	tracelog                  Logger       // trace log for debugging
```
it can be set via SetTraceLog option when creating the client

Wise-Wizard: Got it, will try this.
Thank you!

plugin/storage/integration/elasticsearch_test.go
```diff
@@ -105,14 +104,38 @@ func (s *ESStorageIntegration) initializeES(t *testing.T, allTagsAsFields bool)
 		elastic.SetSniff(false))
```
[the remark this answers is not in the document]
Wise-Wizard: Done!

Wise-Wizard:
Hi, Took a long time to debug the issue as TraceLog also didn't help, but according to me the client being intialized is the reason. The client used in Factory initialization differs from the actual initialization of client, specific to the tests which is causing the Errors.
PTAL, would really appreciate your insights.

yurishkuro:
could you please document what debugging you did and what you found? Were you able to log the queries?

Wise-Wizard:
> could you please document what debugging you did and what you found? Were you able to log the queries?

Yes, I was able to log the queries using SetTraceLog, but the logs didn't show anything evident enough for debugging.
So, instead I ran a debugger in the Main code and my Branch code.
The debugger showed the only difference in the initialisation of SpanReader...
The difference was the client passed to SpanReader which was causing the Error

yurishkuro:
so what's the state now, were you able to resolve the issue? I see the CI is still failing.

Wise-Wizard:
> so what's the state now, were you able to resolve the issue? I see the CI is still failing.

Yes, the GetOperation and the first 2 tests were successful, but I had to do something unorthodox i.e to change the Factory file of ES and do things differently. Can you please look at it and see if it's acceptable? Meanwhile I will look at the next failing test.

Wise-Wizard:
> > so what's the state now, were you able to resolve the issue? I see the CI is still failing.
> 
> Yes, the GetOperation and the first 2 tests were successful, but I had to do something unorthodox i.e to change the Factory file of ES and do things differently. Can you please look at it and see if it's acceptable? Meanwhile I will look at the next failing test.

I mean these 2 below tests were successful.
Screenshot 2024-04-05 105946

Wise-Wizard:
Update: Figured out the reason for the IndexTempelate test failing. That is also initialized with PrimaryClient instead of the client specifically written in the code. The client initialization is itself at fault.

Wise-Wizard:
> Please remove commented out code

Done, PTAL!

yurishkuro:
I removed custom initialization in favor of CLI-based factory creation. There is room for more similar clean-up for sampling store, but it also has a bug that needs to be fixed https://github.com/jaegertracing/jaeger/issues/5333

yurishkuro (approved):

Wise-Wizard:
> I removed custom initialization in favor of CLI-based factory creation. There is room for more similar clean-up for sampling store, but it also has a bug that needs to be fixed #5333

Sure!
Should I look into that next?

PR #5316, opened 2024-04-01
Allow go-leak linter to fail CI

yurishkuro:
- Part of #5006
- We have only two packages remaining without goroutine leak checks, both integration tests, which are being worked on
- Meanwhile new code / packages do not benefit from enforcement of goleak checks because the linter does not fail CI

## Description of the changes
- Add the two remaining packages into exclusion list and force a failure of the CI if anything else is missing goleak checks

Successful exit:
```shell
$ make goleak
🔴 Error(check-goleak): no goleak check in package ./cmd/jaeger/internal/integration/
	this package is temporarily allowed and will not cause linter failure
	this package is temporarily allowed and will not cause linter failure
🐞 Warning(check-goleak): no goleak check in 2 package(s).
	See https://github.com/jaegertracing/jaeger/pull/5010/files
	for examples of adding the checks.
```

Unsuccessful exit (forced by removing one of the existing checks):
```shell
$ make goleak
🔴 Error(check-goleak): no goleak check in package ./cmd/jaeger/internal/integration/
	this package is temporarily allowed and will not cause linter failure
🔴 Error(check-goleak): no goleak check in package ./internal/tracegen/
	this package is temporarily allowed and will not cause linter failure
⛔ Fatal(check-goleak): no goleak check in 3 package(s), 1 of which not allowed.
	See https://github.com/jaegertracing/jaeger/pull/5010/files
	for examples of adding the checks.
make: *** [goleak] Error 1
```

### DIFF
--- a/scripts/check-goleak-files.sh
+++ b/scripts/check-goleak-files.sh
@@ -3,6 +3,7 @@
 set -euo pipefail
 
 bad_pkgs=0
+failed_pkgs=0
 
 # shellcheck disable=SC2048
 for dir in $*; do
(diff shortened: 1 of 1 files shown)


jkowall (approved):
lgtm, thanks!

PR #5322, opened 2024-04-03
[jaeger-v2] add gRPC integration e2e test mode

james-ryans:
- Part of https://github.com/jaegertracing/jaeger/issues/5254#issuecomment-2016858385 solves (1) and (4) for gRPC storage integration test in e2e mode.

## Description of the changes
- Utilizing existing `StorageIntegration` to test the jaeger-v2 OTel Collector and gRPC storage backend with the provided config file at `cmd/jaeger/grpc_config.yaml`.
- Creates an abstraction for e2e test mode that initializes the collector, span writer to the receiver, and span reader from jaeger_query.

- Run `STORAGE=grpc SPAN_STORAGE_TYPE=memory make jaeger-v2-storage-integration-test`

## Checklist


cmd/jaeger/internal/integration/e2e/integration.go
```diff
@@ -0,0 +1,242 @@
...
+	integration.StorageIntegration
+	ConfigFile string
+
+	runner        testbed.OtelcolRunner
+	configCleanup func()
+}
+
+// e2eInitialize starts the Jaeger-v2 collector with the provided config file,
+// it also initialize the SpanWriter and SpanReader below.
+func (s *E2EStorageIntegration) e2eInitialize() error {
```
yurishkuro, line 50: please use the pattern of passing `*testing.T` to all functions and doing `require.NoError(t, err)` instead of returning errors. This makes the code cleaner and easier to debug.

cmd/jaeger/internal/integration/e2e/integration.go
```diff
@@ -0,0 +1,242 @@
...
+		return err
+	}
+
+	config, err := os.ReadFile(s.ConfigFile)
+	if err != nil {
+		return err
+	}
+	config = []byte(strings.Replace(string(config), "./cmd/jaeger/", "../../../", 1))
+
+	s.runner = testbed.NewInProcessCollector(factories)
```
yurishkuro, line 62: Personally I have a preference towards starting the actual jaeger-v2 binary because that is a much more real e2e test. From performance considerations I think it doesn't matter much, because starting the storage backend is the more expensive part anyway. We could even go further and run the binary as a Docker image, for an even deeper e2e testing which is still needed (otherwise how do we know the Docker image we built is actually working?).

It's ok to leave this for now, but I would suggest in the follow-ups to refactor this piece into a swappable component, so that we could have different modes of e2e testing.

cmd/jaeger/internal/integration/e2e/integration.go
```diff
@@ -0,0 +1,242 @@
...
+func (s *E2EStorageIntegration) e2eCleanUp() error {
+	s.configCleanup()
+	_, err := s.runner.Stop()
+	return err
+}
+
+// SpanWriter utilizes the OTLP exporter to send span data to the Jaeger-v2 receiver
+type spanWriter struct {
+	testbed.TraceDataSender
+}
```
yurishkuro, line 89: I would recommend moving these into separate go files, span_writer.go, span_reader.go, etc.

cmd/jaeger/internal/integration/e2e/integration.go
yurishkuro: except for internals of span reader / writer, those are better as proper (non-test) components.

plugin/storage/integration/integration.go
```diff
@@ -538,3 +538,12 @@ func (s *StorageIntegration) RunAll(t *testing.T) {
 	t.Run("GetThroughput", s.testGetThroughput)
 	t.Run("GetLatestProbability", s.testGetLatestProbability)
 }
+
+// RunTestSpanstore runs only span related integration tests
+func (s *StorageIntegration) RunTestSpanstore(t *testing.T) {
```
yurishkuro, line 543: ok, but invoke it from RunAll instead of repeating
```suggestion
func (s *StorageIntegration) RunSpanStoreTests(t *testing.T) {
```

cmd/jaeger/internal/integration/e2e/grpc_test.go
```diff
@@ -0,0 +1,95 @@
...
+	if err != nil {
+		return err
+	}
+
+	s.Refresh = func(_ *testing.T) {}
+	s.CleanUp = s.cleanUp
+	return nil
+}
+
+func (s *GRPCStorageIntegration) startServer() error {
```
yurishkuro, line 42: This again feels like duplication of that the existing grpc_test is already doing. From my perspective, the only differences in the e2e test case are these:
  * it needs to have jaeger-v2 process running (in-process or as binary or as container - see the other comment)
  * it needs to use RPC proxies for SpanReader/Writer instead of in-process function calls

yurishkuro (commented):
This is great!

cmd/jaeger/internal/integration/e2e/integration.go
james-ryans: > Personally I have a preference towards starting the actual jaeger-v2 binary... We could even go further and run the binary as a Docker image,

Working on it!

plugin/storage/integration/grpc_server.go
```diff
@@ -0,0 +1,67 @@
...
+	googleGRPC "google.golang.org/grpc"
+
+	grpcMemory "github.com/jaegertracing/jaeger/plugin/storage/grpc/memory"
+	"github.com/jaegertracing/jaeger/plugin/storage/grpc/shared"
+	"github.com/jaegertracing/jaeger/plugin/storage/memory"
+	"github.com/jaegertracing/jaeger/ports"
+)
+
+type GRPCServer struct {
+	errChan chan error
```
yurishkuro, line 19: Is this actually needed? If it is, I would prefer to use a callback function, not a channel, because channels like this tend to cause deadlocks when nothing is reading them, while the callback achieves the same result of communicating the status without the risk of blocking (and the implementation of the callback can still use channel if it wants to, but that channel will be more local and less prone to deadlocks)

plugin/storage/integration/grpc_test.go
```diff
@@ -16,94 +16,37 @@
...
 }
 
 func (s *GRPCStorageIntegrationTestSuite) initialize(t *testing.T) {
 	s.logger, _ = testutils.NewLogger()
 
 	if s.server != nil {
-		err := s.server.Restart()
+		err := s.server.Start()
 		require.NoError(t, err)
 	}
```
yurishkuro, line 51: This doesn't look equivalent. The Restart function was stopping and replacing the server, you're starting the server that was already stopped (not even sure if it's safe to do)

plugin/storage/integration/grpc_test.go
```diff
@@ -132,8 +75,10 @@ func (s *GRPCStorageIntegrationTestSuite) initialize(t *testing.T) {
 }
 
 func (s *GRPCStorageIntegrationTestSuite) cleanUp(t *testing.T) {
-	err := s.factory.Close()
-	require.NoError(t, err)
+	require.NoError(t, s.factory.Close())
+	if s.server != nil {
+		require.NoError(t, s.server.Close())
```
yurishkuro, line 80: Would it make sense to release the server here (s.server = nil)?

cmd/jaeger/internal/integration/span_reader.go
```diff
@@ -0,0 +1,155 @@
...
+	"github.com/jaegertracing/jaeger/model"
+	"github.com/jaegertracing/jaeger/ports"
+	"github.com/jaegertracing/jaeger/proto-gen/api_v2"
+	"github.com/jaegertracing/jaeger/storage/spanstore"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/grpc/status"
+)
+
+var _ spanstore.Reader = (*spanReader)(nil)
```
yurishkuro, line 24: since it encapsulates an open connection, do we also need io.Closer?

cmd/jaeger/internal/integration/span_writer.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"context"
+
+	"github.com/jaegertracing/jaeger/model"
+	"github.com/jaegertracing/jaeger/storage/spanstore"
+	jaeger2otlp "github.com/open-telemetry/opentelemetry-collector-contrib/pkg/translator/jaeger"
+	"github.com/open-telemetry/opentelemetry-collector-contrib/testbed/testbed"
+)
+
+var _ spanstore.Writer = (*spanWriter)(nil)
```
yurishkuro, line 15: io.Closer to clean up the sender?

cmd/jaeger/internal/integration/span_writer.go
```diff
@@ -0,0 +1,46 @@
...
+
+// SpanWriter utilizes the OTLP exporter to send span data to the Jaeger-v2 receiver
+type spanWriter struct {
+	testbed.TraceDataSender
```
yurishkuro, line 19: do you want spanWriter __to be__ a TraceDataSender (i.e. expose all of the sender's API), or is it just an implementation detail of the spanWriter? If it's the latter, you don't want to embed the struct, but have a named field.

plugin/storage/integration/grpc_test.go
james-ryans: I believe the Restart function is an idempotent version of Start & Close functions.

When the Restart function is first called before running all the tests, it just starts the server. And then the Restart function is called in cleanUp function, which I've looked that it is called right before the end of every subtest, it stops the server and then replaces it with a new one.

With the new Start and Close functions, I "Start" the server before running the tests. Then in the cleanUp function, I "Close" the server and then "Start" a new one. After that, at the end of all tests, I "Close" the server (which the Restart function cannot do and causes goroutine leak).

By the way, the `if s.server != nil` is a check if the current test suite uses remote storage or a plugin (if it uses plugin, we don't start (or restart) the server. But, I'll add additional checking to the Start function to Close it first if it is already running.

plugin/storage/integration/grpc_test.go
james-ryans: Hmm.. actually the server (GRPCServer) here is a declaration if the current test suite uses remote storage or a plugin. If it uses remote storage then it closes and starts the server again.

But the server inside GPRCServer struct is the gRPC server that handles requests, and I'll add to release the server there.

cmd/jaeger/internal/integration/span_writer.go
james-ryans: Yes, added it

james-ryans: Actually, the sender doesn't have the functionality to close the client, so I use OTLP exporter instead.

cmd/jaeger/internal/integration/span_writer.go
james-ryans: I see. It is just an implementation detail actually, named it in the recent changes.

cmd/jaeger/grpc_config.yaml
```diff
@@ -9,17 +9,13 @@ service:
...
     ui_config: ./cmd/jaeger/config-ui.json
 
   jaeger_storage:
     grpc:
       external-storage:
         server: localhost:17271
         connection-timeout: 5s
-      external-storage-archive:
-        server: localhost:17281
-        connection-timeout: 5s
```
yurishkuro, line 22: hm, ok, to be revisited - we should be testing archive too

cmd/jaeger/internal/integration/grpc_test.go
```diff
@@ -5,12 +5,47 @@ package integration
...
+	s.initialize(t)
+}
+
 func TestGRPCStorage(t *testing.T) {
-	s := &StorageIntegration{
-		Name:       "grpc",
-		ConfigFile: "fixtures/grpc_config.yaml",
+	integration.SkipUnlessEnv(t, "grpc")
+
+	server, err := integration.NewGRPCServer()
```
yurishkuro, line 38: you are already calling initialize() in L45

cmd/jaeger/internal/integration/grpc_test.go
```diff
@@ -5,12 +5,47 @@ package integration
...
+	s.server, err = integration.NewGRPCServer()
+	require.NoError(t, err)
+	require.NoError(t, s.server.Start())
+
+	s.Refresh = func(_ *testing.T) {}
+	s.CleanUp = s.cleanUp
+}
+
+func (s *GRPCStorageIntegration) cleanUp(t *testing.T) {
+	require.NoError(t, s.server.Close())
```
yurishkuro, line 31: I am curious, is it really enough to close the server, which only closes the connection to remote process, but doesn't actually perform storage clean-up? The intention of CleanUp was to delete data from storage to avoid data interference across tests.

plugin/storage/integration/grpc_test.go
james-ryans: Could you please point out where the replacing behavior is? I'm afraid I missed your point.

cmd/jaeger/internal/integration/integration.go
```diff
@@ -4,129 +4,61 @@
...
+	s.SpanReader = spanReader
+}
 
-	tc.StopBackend()
+func (s *E2EStorageIntegration) e2eCleanUp(t *testing.T) {
+	require.NoError(t, s.SpanReader.(io.Closer).Close())
+	require.NoError(t, s.SpanWriter.(io.Closer).Close())
 
-	tc.ValidateData()
+	s.binaryProcess.Kill()
```
yurishkuro, line 63: nit: this should probably be in `defer` inside e2eCleanUp, to ensure that it's always run (the two require above could throw and exit). I think even better it to add it to t.Cleanup in e2eInitialize

cmd/jaeger/internal/integration/span_reader.go
```diff
@@ -0,0 +1,172 @@
...
+func unwrapNotFoundErr(err error) error {
+	if s, _ := status.FromError(err); s != nil {
+		if strings.Contains(s.Message(), spanstore.ErrTraceNotFound.Error()) {
+			return spanstore.ErrTraceNotFound
+		}
+	}
+	return err
+}
+
+func (r *spanReader) GetTrace(ctx context.Context, traceID model.TraceID) (*model.Trace, error) {
```
yurishkuro, line 76: a bit sad that you had to implement all of that. This is a takeway for V2 storage - perhaps it's possible to have remote storage API to be the same as query service API. Then we could've used plugin/storage/grpc/shared/client

cmd/jaeger/internal/integration/grpc_test.go
james-ryans: You're right, overlooked it

plugin/storage/integration/grpc_test.go
yurishkuro: Old line 67
```
s.server = googleGRPC.NewServer()
```

cmd/jaeger/internal/integration/span_reader.go
```diff
@@ -0,0 +1,172 @@
...
+	Port int
+
+	clientConn *grpc.ClientConn
+	client     api_v2.QueryServiceClient
+}
+
+func createSpanReader(port int) *spanReader {
+	return &spanReader{
+		Port: port,
+	}
```
yurishkuro, line 41: you may be missing coverage instruction and upload

also, wonder if https://github.com/jaegertracing/jaeger/blob/main/plugin/storage/grpc/shared/grpc_client_test.go could be repurposed for testing these span reader/writer, since integration testing of them will not provide full coverage

plugin/storage/integration/grpc_test.go
james-ryans: Oh, the Start function actually also replaces the server too. But I've refactored it so that the Start function is identical to Restart, with the extracted closing server logic into the Close function (so I can close the server after the end of all tests).

cmd/jaeger/internal/integration/span_reader.go
james-ryans: Okay, I'll create a new issue for this task.

cmd/jaeger/internal/integration/span_reader.go
james-ryans: https://github.com/jaegertracing/jaeger/pull/5322#issuecomment-2033748607 
Have you read this comment? I think the main reason is that the gRPC (v2) integration test is failing due to trace data translation between Jaeger proto and OTLP tag attributes. Jaeger proto's BINARY tag type is translated into STRING type in the OTLP format and remains as STRING type when translated back into Jaeger proto.

yurishkuro: I think it's a bug in the translator. OTLP supports bytes in the AnyValue, the translator did not have to convert it to a string.
```
message AnyValue {
  // The value is one of the listed fields. It is valid for all values to be unspecified
  // in which case this AnyValue is considered to be "empty".
  oneof value {
    string string_value = 1;
    bool bool_value = 2;
    int64 int_value = 3;
    double double_value = 4;
    ArrayValue array_value = 5;
    KeyValueList kvlist_value = 6;
    bytes bytes_value = 7;
  }
}
```

yurishkuro: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/32204

plugin/storage/integration/grpc_server.go
james-ryans: I see, I have updated this to pass the error through the struct field

cmd/jaeger/internal/integration/integration.go
```diff
@@ -4,129 +4,59 @@
...
+	s.SpanWriter = spanWriter
 
-	tc.StopBackend()
+	spanReader := createSpanReader(ports.QueryGRPC)
+	require.NoError(t, spanReader.Start())
+	s.SpanReader = spanReader
+}
 
-	tc.ValidateData()
+func (s *E2EStorageIntegration) e2eCleanUp(t *testing.T) {
```
yurishkuro, line 59: It's becoming very difficult to keep track of who is supposed to call which methods. Please add comments.

cmd/jaeger/internal/integration/span_reader.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+func (r *spanReader) Start() error {
```
yurishkuro, line 44: I suggest merging this into create function. What's the benefit of separating Start?

plugin/storage/integration/grpc_server.go
```diff
@@ -0,0 +1,69 @@
...
+
+	googleGRPC "google.golang.org/grpc"
+
+	grpcMemory "github.com/jaegertracing/jaeger/plugin/storage/grpc/memory"
+	"github.com/jaegertracing/jaeger/plugin/storage/grpc/shared"
+	"github.com/jaegertracing/jaeger/plugin/storage/memory"
+	"github.com/jaegertracing/jaeger/ports"
+)
+
+type GRPCServer struct {
```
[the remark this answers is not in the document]
yurishkuro: Also, what is the difference between this server and cmd/remote-storage/app/server.go? Could we reuse that one?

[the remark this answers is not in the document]
james-ryans: Actually yes, we could reuse `cmd/remote-storage/app/server.go`. This server is just an extraction of the gRPCServer from existing grpc_test.go so I can reuse this code for the new e2e grpc_test, but I'll change this to reuse the remote-storage server.

cmd/jaeger/internal/integration/span_reader.go
james-ryans: Okay. I've seen a lot of structs that separate the creation and start functions so I thought it was cleaner that way. I'll refactor this.

yurishkuro:
@james-ryans another possible way to unblock from waiting on OTEL contrib is to tweak the tests to not use binary tags (controlled by a flag, with a new issue to fix it once OTEL fix is released)

plugin/storage/integration/integration.go
```diff
@@ -75,6 +75,10 @@ type StorageIntegration struct {
 	// Skip Archive Test if not supported by the storage backend
 	SkipArchiveTest bool
 
+	// TODO: remove this after upstream issue in OTEL jaeger translator is fixed
+	// Skip testing trace binary tags, logs, and process
+	SkipBinaryAttrs bool
```
yurishkuro, line 80: please book a ticket referring to this commit

plugin/storage/integration/remote_memory_storage.go
```diff
@@ -0,0 +1,63 @@
...
+
+	return &RemoteMemoryStorage{
+		server:         server,
+		storageFactory: storageFactory,
+	}, nil
+}
+
+func (s *RemoteMemoryStorage) Close() error {
+	if err := s.server.Close(); err != nil {
+		return err
```
yurishkuro, line 60: nit: this potentially leaves storageFactory not closed. The pattern we use for multi-close is
```go
	return errors.Join(
		b.reporter.Close(),
		b.tlsCloser.Close(),
		b.GetConn().Close(),
	)
```

plugin/storage/integration/remote_memory_storage.go
```diff
@@ -0,0 +1,63 @@
...
+	opts := &app.Options{
+		GRPCHostPort: ports.PortToHostPort(ports.RemoteStorageGRPC),
+		Tenancy: tenancy.Options{
+			Enabled: false,
+		},
+	}
+	tm := tenancy.NewManager(&opts.Tenancy)
+	storageFactory, err := storage.NewFactory(storage.FactoryConfigFromEnvAndCLI(os.Args, os.Stderr))
+	if err != nil {
+		return nil, err
```
yurishkuro, line 35: wdyt about taking t.Testing arg and just calling require.NoError? This will eliminate the uncovered lines.

plugin/storage/integration/es_index_cleaner_test.go
```diff
@@ -60,7 +60,7 @@ func TestIndexCleaner_doNotFailOnEmptyStorage(t *testing.T) {
 }
 
 func TestIndexCleaner_doNotFailOnFullStorage(t *testing.T) {
-	skipUnlessEnv(t, "elasticsearch", "opensearch")
+	SkipUnlessEnv(t, "elasticsearch", "opensearch")
```
yurishkuro, line 63: this little change is a good candidate for a separate PR. It would significantly reduce the number of files in this PR, making it more focused.

plugin/storage/integration/integration.go
james-ryans: Booked it at https://github.com/jaegertracing/jaeger/issues/5341

plugin/storage/integration/remote_memory_storage.go
james-ryans: Yes, this is the easiest approach. Will do

plugin/storage/integration/es_index_cleaner_test.go
james-ryans: I see, I'll do that

Issue #5229, opened 2024-02-24
[jaeger-v2] Rethink components configuration to align with OTEL


yurishkuro:
Right now several PRs that are trying to add support for standard storage backends to v2 are reusing the same configuration for storage as we had in v1. Now is a good time to rethink that configuration, in particular:
  * better align with standard components of OTEL Collector. For example, the gRPC storage (https://github.com/jaegertracing/jaeger/pull/5228) is conceptually very similar to OTLP Exporter - they both send data over gRPC to a remote service. Our configuration is pretty slim, while OTLP Exporter's configuration has [a lot more features](https://github.com/open-telemetry/opentelemetry-collector/blob/6be5fe9fecd12a39adf093a2e7c5b55700bc0518/exporter/otlpexporter/config.go#L16), there is no reason why we cannot support them (would be even better if we could simply reuse OTEL's configuration utilities for gRPC)
  * our storage configs have been accumulating over time, we can see if there are better logical groupings. For example, ES config is a [flat list of everything](https://github.com/jaegertracing/jaeger/blob/ef4791ec7b0761f7a6f9ac3cace54a2039483d8e/pkg/es/config/config.go#L47), we can group many flags into sub-types so that the YAML config looks a lot more structured than using long field names like `priority_dependencies_template`
  * ensure that we have sensible defaults. For example, in #5398 basic options had to be added to Cassandra configuration, even though in v1 many of them are already defaulted to these values, so we're clearly missing defaulting mechanism. It's somewhat complicated by the fact that the storage extension cannot actually define the defaults in `CreateDefaultConfig` because at that point we don't known which storage will be used.

## Scope
  * [x] Cassandra storage config -- #5928
  * [ ] Elasticsearch / OpenSearch storage config -- #6059
  * [x] Badger storage config -- #5926
  * [x] Memory storage config -- #5923
  * [x] Kafka config -- #6027
  * [x] grpc storage (e.g. issue #6030) -- #6041
  * [x] Adaptive sampling processor / extension config -- #6021 
  * [x] Query service config -- #5996
  * [x] Collector-specific config -- #6040
  * [x] Utilize env var defaults available since OTEL v0.110 e.g. https://github.com/jaegertracing/jaeger/blob/c655e56694a5340520b330968452c8db99db27eb/cmd/jaeger/internal/integration/kafka_test.go#L24

## Tasks / outcomes
For each of the above components, we need to ensure
[...]

PR #5331, opened 2024-04-05
Create new grpc storage configuration to align with OTEL

akagami-harsh:
- part of #5229 

## Description of the changes
- added more grpc storage client configuration to align with otel
 ```
	configgrpc.ClientConfig        `mapstructure:",squash"`
	exporterhelper.TimeoutSettings `mapstructure:",squash"`
 ```
These are not all the configs, but i'll add more based on feedback on this initial approach.

- not tested yet 

## Checklist


go.mod
```diff
@@ -151,7 +151,7 @@ require (
 	github.com/mitchellh/reflectwalk v1.0.2 // indirect
 	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
 	github.com/modern-go/reflect2 v1.0.2 // indirect
-	github.com/mostynb/go-grpc-compression v1.2.2 // indirect
+	github.com/mostynb/go-grpc-compression v1.2.2
```
yurishkuro, line 154: this seems unrelated to this PR

plugin/storage/grpc/config/config.go
```diff
@@ -90,7 +100,28 @@ func (c *Configuration) buildRemote(logger *zap.Logger, tracerProvider trace.Tra
 	opts := []grpc.DialOption{
 		grpc.WithStatsHandler(otelgrpc.NewClientHandler(otelgrpc.WithTracerProvider(tracerProvider))),
 		grpc.WithBlock(),
+		grpc.WithDefaultCallOptions(grpc.WaitForReady(c.WaitForReady)),
+	}
+
+	if c.Compression.IsCompressed() {
```
yurishkuro, line 106: Doesn't OTEL provide a utility function in their config packages that helps with translating gRPC server config into settings? We should avoid reinventing the wheel and just reuse what they have.

akagami-harsh: To reuse that, we need to fully utilize all the features it has. We can do this for every feature except `auth` (authentication) , since it needs to be supported by both the client and server. and also the `auth` is tightly coupled with auth extension https://github.com/open-telemetry/opentelemetry-collector/tree/main/config/configauth

akagami-harsh: > Do we need to do something special for auth?

yes, i think from the client side we can reuse the otel's authentication extensions but, from the server side we need to find a find a way to support authentication.

plugin/storage/grpc/config/config.go
```diff
@@ -47,10 +50,11 @@ type Configuration struct {
...
 
-	pluginHealthCheck     *time.Ticker
-	pluginHealthCheckDone chan bool
-	pluginRPCClient       plugin.ClientProtocol
-	remoteConn            *grpc.ClientConn
+	pluginHealthCheck       *time.Ticker
+	pluginHealthCheckDone   chan bool
+	pluginRPCClient         plugin.ClientProtocol
+	remoteConn              *grpc.ClientConn
+	configgrpc.ClientConfig `mapstructure:",squash"`
```
yurishkuro, line 57: Let's create a new config struct like ConfigV2 and only keep the fields that are really needed there. For instance, we do not plan to support sidecar plugin model, so the first three fields at the top are not needed.

Also, I assume otel ClientConfig would already include things like TLS and timeout and address, so fields 4-6 also should not be needed in the new config.

Basically, I would suggest this refactoring:
  * add ConfigV2 with only what we need
  * refactor remote storage initialization code to use OTEL helpers for interpreting ClientConfig into a gRPC connection
  * write a translator from v1 Configuration into ConfigV2 (V2 should be a superset, except for the sidecar mode).

akagami-harsh: now the sidecar plugin config fields are removed, do we still need to create a new configV2 and a translator?

cmd/remote-storage/app/server.go
```diff
@@ -22,6 +22,7 @@ import (
 	"go.uber.org/zap"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
+	_ "google.golang.org/grpc/encoding"
```
yurishkuro, line 25: Is this the encoding we need, or the gogo one from pkg/ ?

plugin/storage/grpc/config/config.go
```diff
@@ -41,6 +46,30 @@ type Configuration struct {
 	remoteConn *grpc.ClientConn
 }
 
+type ConfigV2 struct {
+	Configuration `mapstructure:",squash"`
```
yurishkuro, line 50: The idea of cfg v2 is to have a clean break, not to keep using v1 config (ok to use it internally when starting the implementation, but not as external config struct)

plugin/storage/grpc/factory.go
```diff
@@ -62,7 +62,7 @@ func NewFactory() *Factory {
 
 // NewFactoryWithConfig is used from jaeger(v2).
 func NewFactoryWithConfig(
-	cfg config.Configuration,
+	cfg config.ConfigV2,
```
yurishkuro, line 65: +1

cmd/remote-storage/app/server.go
akagami-harsh: earlier i added it because i was getting a warning when used the `Compression` config from `configgrpc.ClientConfig` . but now i am not getting any such warning. so removed it for now

plugin/storage/grpc/config/config.go
```diff
@@ -37,10 +42,34 @@ type Configuration struct {
 	RemoteTLS            tlscfg.Options
 	RemoteConnectTimeout time.Duration `yaml:"connection-timeout" mapstructure:"connection-timeout"`
 	TenancyOpts          tenancy.Options
+}
+
+type ConfigV2 struct {
+	Configuration
```
yurishkuro, line 48: You are still redefining v2 in terms of v1 - you need to break v1 dependency completely (super easy in case of this specific storage since v1 config is quire small). Otherwise you're just duplicating the settings, e.g. tenancy and TLS are also defined in v1 Configuration

plugin/storage/grpc/config/config.go
```diff
@@ -37,10 +42,34 @@ type Configuration struct {
...
+	TenancyOpts                    tenancy.Options
+	configgrpc.ClientConfig        `mapstructure:",squash"`
+	exporterhelper.TimeoutSettings `mapstructure:",squash"`
 
 	remoteConn *grpc.ClientConn
```
yurishkuro, line 53: can we avoid having runtime objects in the configuration structs? It was a bad pattern in v1

plugin/storage/grpc/config/config.go
```diff
@@ -55,17 +84,21 @@ type PluginBuilder interface {
 }
 
 // Build instantiates a PluginServices
-func (c *Configuration) Build(logger *zap.Logger, tracerProvider trace.TracerProvider) (*ClientPluginServices, error) {
-	return c.buildRemote(logger, tracerProvider, grpc.NewClient)
+func (c *ConfigV2) Build(logger *zap.Logger, tracerProvider trace.TracerProvider) (*ClientPluginServices, error) {
```
yurishkuro, line 87: I suggest 
```
func NewRemoteStorage(*ConfigV2)
```

leave Build where it is, with Configuration, but re-implement as
```
func (c *Configuration) Build(...) {
    v2cfg = c.AsV2()
    s := NewRemoteStorage(v2cfg)
}
```

akagami-harsh: PTAL, is this what you intended?

plugin/storage/grpc/config/config.go
akagami-harsh: moved it in `factory` struct

plugin/storage/grpc/config/config.go
```diff
@@ -37,8 +39,32 @@ type Configuration struct {
...
+	exporterhelper.TimeoutSettings `mapstructure:",squash"`
+}
+
+func (c *Configuration) translateToConfigV2() *ConfigV2 {
+	v2 := &ConfigV2{}
+	v2.Endpoint = c.RemoteServerAddr
+	v2.Timeout = c.RemoteConnectTimeout
 
-	remoteConn *grpc.ClientConn
+	v2.ClientConfig.TLSSetting.Insecure = !c.RemoteTLS.Enabled
```
yurishkuro, line 55: let's move all these assignments into a method on the `tlscfg` object
```
func (o *Options) ToOtelClientConfig()
```

plugin/storage/grpc/config/config.go
```diff
@@ -50,60 +76,53 @@ type ClientPluginServices struct {
...
 // Build instantiates a PluginServices
-func (c *Configuration) Build(logger *zap.Logger, tracerProvider trace.TracerProvider) (*ClientPluginServices, error) {
-	return c.buildRemote(logger, tracerProvider, grpc.NewClient)
+func (c *Configuration) Build(logger *zap.Logger, tracerProvider trace.TracerProvider) (*ClientPluginServices, *grpc.ClientConn, error) {
+	v2Cfg := c.translateToConfigV2()
+	s, remoteConn, err := newRemoteStorage(v2Cfg, tracerProvider)
+	if err != nil {
+		return nil, nil, err
+	}
+	return s, remoteConn, nil
```
yurishkuro, line 89:
```suggestion
	return v2Cfg.Build(...)
```

plugin/storage/grpc/config/config.go
```diff
@@ -50,60 +76,53 @@ type ClientPluginServices struct {
...
 
 	tenancyMgr := tenancy.NewManager(&c.TenancyOpts)
 	if tenancyMgr.Enabled {
 		opts = append(opts, grpc.WithUnaryInterceptor(tenancy.NewClientUnaryInterceptor(tenancyMgr)))
 		opts = append(opts, grpc.WithStreamInterceptor(tenancy.NewClientStreamInterceptor(tenancyMgr)))
 	}
-	var err error
-	c.remoteConn, err = newClient(c.RemoteServerAddr, opts...)
+
+	remoteConn, err := c.ToClientConn(context.Background(), componenttest.NewNopHost(), component.TelemetrySettings{}, opts...)
```
yurishkuro, line 115: > `componenttest.NewNopHost()`, `component.TelemetrySettings{}`

that doesn't make sense in production code, if these args are required we need to somehow pass them in as arguments.

plugin/storage/grpc/config/config.go
```diff
@@ -50,60 +76,53 @@ type ClientPluginServices struct {
...
+}
 
-func (c *Configuration) buildRemote(logger *zap.Logger, tracerProvider trace.TracerProvider, newClient newClientFn) (*ClientPluginServices, error) {
+func newRemoteStorage(c *ConfigV2, tracerProvider trace.TracerProvider) (*ClientPluginServices, *grpc.ClientConn, error) {
 	opts := []grpc.DialOption{
 		grpc.WithStatsHandler(otelgrpc.NewClientHandler(otelgrpc.WithTracerProvider(tracerProvider))),
 		grpc.WithBlock(),
 	}
-	if c.RemoteTLS.Enabled {
-		tlsCfg, err := c.RemoteTLS.Config(logger)
```
yurishkuro, line 70: This used to create a background thread that was watching for certificate file changes. How is this done in OTEL? It won't be an equivalent change if we don't support cert reloads.

plugin/storage/grpc/factory.go
```diff
@@ -92,14 +95,17 @@ func (f *Factory) InitFromViper(v *viper.Viper, logger *zap.Logger) {
 func (f *Factory) configureFromOptions(opts Options) {
 	f.options = opts
 	f.builder = &f.options.Configuration
+	f.builder = &f.options.ConfigV2
```
yurishkuro, line 98: why have `options.Configuration` if we won't use it?

plugin/storage/grpc/config/config.go
```diff
@@ -50,60 +76,53 @@ type ClientPluginServices struct {
...
+func (c *ConfigV2) Build(logger *zap.Logger, tracerProvider trace.TracerProvider) (*ClientPluginServices, *grpc.ClientConn, error) {
+	s, remoteConn, err := newRemoteStorage(c, tracerProvider)
+	if err != nil {
+		return nil, nil, err
+	}
+	return s, remoteConn, nil
+}
 
-func (c *Configuration) buildRemote(logger *zap.Logger, tracerProvider trace.TracerProvider, newClient newClientFn) (*ClientPluginServices, error) {
+func newRemoteStorage(c *ConfigV2, tracerProvider trace.TracerProvider) (*ClientPluginServices, *grpc.ClientConn, error) {
```
yurishkuro, line 100: Do yo think it's actually important to the caller of this function to get back `*grpc.ClientConn`? I understand that you are returning it in order to ensure it's closed in the end, but from the caller perspective what it's trying to do is instantiate `*ClientPluginServices`. The connection is just an implementation detail of `*ClientPluginServices`, the caller does not need to know about it. Instead we can capture the connection inside `*ClientPluginServices` and have `*ClientPluginServices` implement a Close method.

plugin/storage/grpc/options.go
```diff
@@ -37,6 +37,7 @@ const (
 // to bind them to command line flags
 type Options struct {
 	Configuration config.Configuration `mapstructure:",squash"`
+	ConfigV2      config.ConfigV2      `mapstructure:",squash"`
```
yurishkuro, line 40: I recommend leaving Options only as a container for Configuration. It's still needed to support v1 CLI parsing, but in the factory when time comes to start creating services we can convert it to ConfigV2 once.

yurishkuro (commented):
it looks like you're picking some go.mod changes because your branch is not up to date. Please rebase on main

yurishkuro (commented):
I refactored & cleaned-up a bit more to make the code simpler.

pkg/config/tlscfg/options.go
```diff
@@ -135,6 +136,23 @@ func (p Options) loadCertPool() (*x509.CertPool, error) {
 	return certPool, nil
 }
 
+func (o *Options) ToOtelClientConfig() configtls.ClientConfig {
```
yurishkuro, line 139: we should add a test for this to keep the coverage up

plugin/storage/grpc/config/config.go
```diff
@@ -37,72 +39,80 @@ type Configuration struct {
 	RemoteTLS            tlscfg.Options
 	RemoteConnectTimeout time.Duration `yaml:"connection-timeout" mapstructure:"connection-timeout"`
 	TenancyOpts          tenancy.Options
+}
+
+type ConfigV2 struct {
+	TenancyOpts                    tenancy.Options
```
yurishkuro, line 45:
```suggestion
	Tenancy                    tenancy.Options
```
(why would we want "Opts" suffix in the config files?)

yurishkuro (approved):

PR #5340, opened 2024-04-08
Fix goroutine leaks in grpc integration tests

WillSewell:

## Description of the changes
- The key issue was that the grpc client was only being killed in a `runtime.SetFinalizer` - i.e. when it is GCed. I think in the tests this was not shutting down before goleak had decided that the goroutine had leaked.
- This change switches to a more conventional approach of killing the client as part of the Close method and then ensuring this is called in each of the tests.

- While this change does not include a call to activate goleak, it _was_ tested with this. The PR does not include goleak detection because there are still other violations related to elasticsearch in this package.

## Checklist


yurishkuro:
I merged a large PR that affected some of the test files. There are now merge conflicts (which would've been more difficult to resolve in that large PR than here, so I picked this one to resolve conflicts).

WillSewell:
No worries. The changes actually make this PR simpler.

One area that is pretty unpleasant about this change is the fact that that each subtest is calling `CleanUp` which in this case is actually _initializing_ for the following subtest. This means the final subtest pointlessly initializes. For now I've taken the approach of just explicitly closing this after each test.

I think a nicer approach would be to introduce separate a `Initialize` callback (separate from `CleanUp`) but this is a larger change because I think it should be done across all integration tests.

yurishkuro:
I also don't like how CleanUp functions often call initialize(), this was not at all how it was meant to work, the cleanup functions were intended to remove data from the DB, not to reset the connection completely. 

I am not opposed to larger refactoring (may not be as large as it seems) if we can make the workflow more robust and more intuitive.

I don't even remember why Refresh callback was needed - nothing implements it except for ES, and in case of ES the reason seems to be to ensure that the data is queryable after write once Refresh returns, which actually doesn't seem to work reliably anyway and we already have most of the assertions on reading data executing within waitForCondition, which makes synchronous flushing irrelevant.

plugin/storage/integration/grpc_test.go
```diff
@@ -109,6 +113,7 @@ func TestGRPCStorage(t *testing.T) {
 	}
 	s.initialize(t)
 	s.RunAll(t)
+	s.close(t)
```
yurishkuro, line 116: nit: should really be called before RunAll via defer()

yurishkuro (approved):

PR #5349, opened 2024-04-11
Create sampling templates when creating sampling store

### LINKED ISSUE #5333: es.SamplingStore.CreateTemplates is never called
In case of Span Storage the templates are created when SpanWriter is instantiated. Sampling templates could be created when SamplingStore is being created. Right now the method exists but is never called, except from the integration test.

JaeguKim:
- Resolves #5333

## Description of the changes
- Create sampling templates when creating sampling store since doing so is intended.

- CI

## Checklist

### DIFF
--- a/plugin/storage/es/factory.go
+++ b/plugin/storage/es/factory.go
@@ -74,6 +74,8 @@ type Factory struct {
 	archiveClient atomic.Pointer[es.Client]
 
 	watchers []*fswatcher.FSWatcher
+
+	templateBuilder es.TemplateBuilder
 }
 
 // NewFactory creates a new Factory.
@@ -171,6 +173,8 @@ func (f *Factory) Initialize(metricsFactory metrics.Factory, logger *zap.Logger)
 		}
 	}
 
+	f.templateBuilder = es.TextTemplateBuilder{}
+
 	return nil
 }
 
@@ -195,7 +199,7 @@ func (f *Factory) CreateSpanReader() (spanstore.Reader, error) {
 
 // CreateSpanWriter implements storage.Factory
 func (f *Factory) CreateSpanWriter() (spanstore.Writer, error) {
-	return createSpanWriter(f.getPrimaryClient, f.primaryConfig, false, f.metricsFactory, f.logger)
+	return createSpanWriter(f.getPrimaryClient, f.primaryConfig, false, f.metricsFactory, f.logger, f.templateBuilder)
 }
 
 // CreateDependencyReader implements storage.Factory
@@ -216,7 +220,7 @@ func (f *Factory) CreateArchiveSpanWriter() (spanstore.Writer, error) {
 	if !f.archiveConfig.Enabled {
 		return nil, nil
 	}
-	return createSpanWriter(f.getArchiveClient, f.archiveConfig, true, f.metricsFactory, f.logger)
+	return createSpanWriter(f.getArchiveClient, f.archiveConfig, true, f.metricsFactory, f.logger, f.templateBuilder)
 }
 
 func createSpanReader(
@@ -255,6 +259,7 @@ func createSpanWriter(
 	archive bool,
 	mFactory metrics.Factory,
 	logger *zap.Logger,
+	templateBuilder es.TemplateBuilder,
 ) (spanstore.Writer, error) {
 	var tags []string
 	var err error
@@ -282,7 +287,7 @@ func createSpanWriter(
 
 	// Creating a template here would conflict with the one created for ILM resulting to no index rollover
 	if cfg.CreateIndexTemplates && !cfg.UseILM {
-		mappingBuilder := mappingBuilderFromConfig(cfg)
+		mappingBuilder := getMappingBuilder(cfg, templateBuilder)
 		spanMapping, serviceMapping, err := mappingBuilder.GetSpanServiceMappings()
 		if err != nil {
 			return nil, err
@@ -307,7 +312,7 @@ func (f *Factory) CreateSamplingStore(maxBuckets int) (samplingstore.Store, erro
 	})
 
 	if f.primaryConfig.CreateIndexTemplates && !f.primaryConfig.UseILM {
-		mappingBuilder := mappingBuilderFromConfig(f.primaryConfig)
+		mappingBuilder := getMappingBuilder(f.primaryConfig, f.templateBuilder)
 		sampleMapping, err := mappingBuilder.GetSamplingMappings()
 		if err != nil {
 			return nil, err
@@ -320,9 +325,9 @@ func (f *Factory) CreateSamplingStore(maxBuckets int) (samplingstore.Store, erro
 	return store, nil
 }
 
-func mappingBuilderFromConfig(cfg *config.Configuration) mappings.MappingBuilder {
+func getMappingBuilder(cfg *config.Configuration, templateBuilder es.TemplateBuilder) mappings.MappingBuilder {
 	return mappings.MappingBuilder{
-		TemplateBuilder:              es.TextTemplateBuilder{},
+		TemplateBuilder:              templateBuilder,
 		Shards:                       cfg.NumShards,
 		Replicas:                     cfg.NumReplicas,
 		EsVersion:                    cfg.Version,
--- a/plugin/storage/es/factory.go
+++ b/plugin/storage/es/factory.go
@@ -266,23 +266,6 @@ func createSpanWriter(
 		return nil, err
 	}
 
-	mappingBuilder := mappings.MappingBuilder{
-		TemplateBuilder:              es.TextTemplateBuilder{},
-		Shards:                       cfg.NumShards,
-		Replicas:                     cfg.NumReplicas,
-		EsVersion:                    cfg.Version,
-		IndexPrefix:                  cfg.IndexPrefix,
-		UseILM:                       cfg.UseILM,
-		PrioritySpanTemplate:         cfg.PrioritySpanTemplate,
-		PriorityServiceTemplate:      cfg.PriorityServiceTemplate,
-		PriorityDependenciesTemplate: cfg.PriorityDependenciesTemplate,
-	}
-
-	spanMapping, serviceMapping, err := mappingBuilder.GetSpanServiceMappings()
-	if err != nil {
-		return nil, err
-	}
-
 	writer := esSpanStore.NewSpanWriter(esSpanStore.SpanWriterParams{
 		Client:                 clientFn,
 		IndexPrefix:            cfg.IndexPrefix,
@@ -299,7 +282,12 @@ func createSpanWriter(
 
 	// Creating a template here would conflict with the one created for ILM resulting to no index rollover
 	if cfg.CreateIndexTemplates && !cfg.UseILM {
-		err := writer.CreateTemplates(spanMapping, serviceMapping, cfg.IndexPrefix)
+		mappingBuilder := mappingBuilderFromConfig(cfg)
+		spanMapping, serviceMapping, err := mappingBuilder.GetSpanServiceMappings()
+		if err != nil {
+			return nil, err
+		}
+		err = writer.CreateTemplates(spanMapping, serviceMapping, cfg.IndexPrefix)
 		if err != nil {
 			return nil, err
 		}
@@ -319,29 +307,31 @@ func (f *Factory) CreateSamplingStore(maxBuckets int) (samplingstore.Store, erro
 	})
 
 	if f.primaryConfig.CreateIndexTemplates && !f.primaryConfig.UseILM {
-		mappingBuilder := mappings.MappingBuilder{
-			TemplateBuilder:              es.TextTemplateBuilder{},
-			Shards:                       f.primaryConfig.NumShards,
-			Replicas:                     f.primaryConfig.NumReplicas,
-			EsVersion:                    f.primaryConfig.Version,
-			IndexPrefix:                  f.primaryConfig.IndexPrefix,
-			UseILM:                       f.primaryConfig.UseILM,
-			PrioritySpanTemplate:         f.primaryConfig.PrioritySpanTemplate,
-			PriorityServiceTemplate:      f.primaryConfig.PriorityServiceTemplate,
-			PriorityDependenciesTemplate: f.primaryConfig.PriorityDependenciesTemplate,
-		}
-
+		mappingBuilder := mappingBuilderFromConfig(f.primaryConfig)
 		sampleMapping, err := mappingBuilder.GetSamplingMappings()
 		if err != nil {
 			return nil, err
 		}
-
 		if err := store.CreateTemplates(sampleMapping); err != nil {
 			return nil, err
 		}
 	}
 
 	return store, nil
+}
+
+func mappingBuilderFromConfig(cfg *config.Configuration) mappings.MappingBuilder {
+	return mappings.MappingBuilder{
+		TemplateBuilder:              es.TextTemplateBuilder{},
+		Shards:                       cfg.NumShards,
+		Replicas:                     cfg.NumReplicas,
+		EsVersion:                    cfg.Version,
+		IndexPrefix:                  cfg.IndexPrefix,
+		UseILM:                       cfg.UseILM,
+		PrioritySpanTemplate:         cfg.PrioritySpanTemplate,
+		PriorityServiceTemplate:      cfg.PriorityServiceTemplate,
+		PriorityDependenciesTemplate: cfg.PriorityDependenciesTemplate,
+	}
 }
 
 func createDependencyReader(
--- a/plugin/storage/es/factory.go
+++ b/plugin/storage/es/factory.go
@@ -48,9 +48,9 @@ import (
 )
 
 const (
-	primaryNamespace    = "es"
-	archiveNamespace    = "es-archive"
-	indexTemplatePrefix = "jaeger-sampling-"
+	primaryNamespace       = "es"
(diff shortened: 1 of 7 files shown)


plugin/storage/es/factory.go
```diff
@@ -317,6 +317,30 @@ func (f *Factory) CreateSamplingStore(maxBuckets int) (samplingstore.Store, erro
 		Lookback:               f.primaryConfig.AdaptiveSamplingLookback,
 		MaxDocCount:            f.primaryConfig.MaxDocCount,
 	})
+
+	if f.primaryConfig.CreateIndexTemplates && !f.primaryConfig.UseILM {
+		mappingBuilder := mappings.MappingBuilder{
```
yurishkuro, line 322: this is largely identical to L269, please refactor into a helper function that accepts a `cfg`

yurishkuro (commented):
Thanks!

The full scope of this ticket includes:
  * adding tests to validate this change
  * changing existing integration test that manually creates the mapping to not do that and instead rely on this new built-in behavior

Question: what happens if the mapping already exists?

plugin/storage/es/factory.go
```diff
@@ -171,6 +173,8 @@ func (f *Factory) Initialize(metricsFactory metrics.Factory, logger *zap.Logger)
 		}
 	}
 
+	f.templateBuilder = es.TextTemplateBuilder{}
```
yurishkuro, line 176: If this is a stateless class that requires no parameters to initialize, why pass it around? It can be instantiated where needed.

JaeguKim:
@yurishkuro Apologies for the delayed response.
I've implemented tests to ensure coverage of code changes:
- 5c5c5b5b643f4604da8c1a005b22a0fb58edbb2d
- aae3616be29bcb6d2e1f7521df5e545402078825

I've refactored the integration test to utilize the CreateSamplingStore method instead of manual creation.
- 035603954a91113f258a0b7478a83c12fa20c1aa

When mapping was already created and SamplingStore is created again via CreateSamplingStore, this could result the existing indexes or entire indexes becoming unsearchable depending on the compatibility between existing and new index template.

I think there are several strategies.
- Merge existing index template with the new template
  - This approach involves handling cases, such as conflicting field types, to ensure that existing indexes remain searchable. 
- Create a new index template with a different name
  - By supporting search for indexes with different index template versions, we can maintain compatibility while introducing the new template. But storage cost should be considered.

plugin/storage/es/factory.go
JaeguKim: I added following helper function. https://github.com/JaeguKim/jaeger/blob/aae3616be29bcb6d2e1f7521df5e545402078825/plugin/storage/es/factory.go#L328-L340

yurishkuro:
for some reason this test consistently fails in your branch, while always succeeds in main:
```
$ go test ./plugin/storage/es
ok  	github.com/jaegertracing/jaeger/plugin/storage/es	(cached)
```

plugin/storage/es/factory.go
```diff
@@ -317,9 +310,35 @@ func (f *Factory) CreateSamplingStore(maxBuckets int) (samplingstore.Store, erro
...
+			return nil, err
+		}
+	}
+
 	return store, nil
 }
 
+func getMappingBuilder(cfg *config.Configuration, templateBuilder es.TemplateBuilder) mappings.MappingBuilder {
+	return mappings.MappingBuilder{
+		TemplateBuilder:              templateBuilder,
```
yurishkuro, line 330: why do you need this to be passed as parameter? Previously it was initialized in-place:
```
	mappingBuilder := mappings.MappingBuilder{
		TemplateBuilder:              es.TextTemplateBuilder{},
```

yurishkuro:
> When mapping was already created and SamplingStore is created again via CreateSamplingStore, this could result the existing indexes or entire indexes becoming unsearchable depending on the compatibility between existing and new index template.

why is that? doesn't CreateSamplingStore create **different** index templates? It should not be affecting span templates.

plugin/storage/es/factory.go
```diff
@@ -317,9 +310,35 @@ func (f *Factory) CreateSamplingStore(maxBuckets int) (samplingstore.Store, erro
 		Lookback:               f.primaryConfig.AdaptiveSamplingLookback,
 		MaxDocCount:            f.primaryConfig.MaxDocCount,
 	})
+
+	if f.primaryConfig.CreateIndexTemplates && !f.primaryConfig.UseILM {
+		mappingBuilder := getMappingBuilder(f.primaryConfig, f.templateBuilder)
+		sampleMapping, err := mappingBuilder.GetSamplingMappings()
```
yurishkuro, line 316:
```suggestion
		samplingMapping, err := mappingBuilder.GetSamplingMappings()
```

plugin/storage/es/factory.go
```diff
@@ -317,9 +310,35 @@ func (f *Factory) CreateSamplingStore(maxBuckets int) (samplingstore.Store, erro
...
+		if err != nil {
+			return nil, err
+		}
+		if err := store.CreateTemplates(sampleMapping); err != nil {
```
yurishkuro, line 320: can we delete CreateTemplates methods and inline it here? It's a one-liner that should not be called from anywhere else.

JaeguKim:
> why is that? doesn't CreateSamplingStore create **different** index templates? It should not be affecting span templates.

https://github.com/jaegertracing/jaeger/blob/aae3616be29bcb6d2e1f7521df5e545402078825/plugin/storage/es/samplingstore/storage.go#L165

It appears that the template ID is hardcoded as jaeger-sampling. Am I interpreting this correctly?

plugin/storage/es/factory.go
JaeguKim: 76397788b289326949557c0c7b7a988903e23fb0

plugin/storage/es/factory.go
JaeguKim: I reverted the code back to its original state. eaa6bdce979d6f4779ffe52df358a43fcf5ac38a

plugin/storage/es/factory.go
JaeguKim: I reverted the code back to its original state. https://github.com/jaegertracing/jaeger/commit/eaa6bdce979d6f4779ffe52df358a43fcf5ac38a

plugin/storage/es/factory.go
```diff
@@ -317,9 +307,39 @@ func (f *Factory) CreateSamplingStore(maxBuckets int) (samplingstore.Store, erro
...
+
+	if f.primaryConfig.CreateIndexTemplates && !f.primaryConfig.UseILM {
+		mappingBuilder := mappingBuilderFromConfig(f.primaryConfig)
+		samplingMapping, err := mappingBuilder.GetSamplingMappings()
+		if err != nil {
+			return nil, err
+		}
+		primaryClient := f.getPrimaryClient()
+		if primaryClient == nil {
+			return nil, errors.New("primary client is nil")
```
yurishkuro, line 319: does this even happen? You already called getPrimaryClient in L302

plugin/storage/es/factory.go
```diff
@@ -317,9 +307,39 @@ func (f *Factory) CreateSamplingStore(maxBuckets int) (samplingstore.Store, erro
...
+		}
+		if _, err := primaryClient.CreateTemplate(indexTemplateId).Body(samplingMapping).Do(context.Background()); err != nil {
+			return nil, err
```
yurishkuro, line 322:
```suggestion
			return nil, errors.Wrap("failed to create template: %w", err)
```

plugin/storage/es/factory.go
yurishkuro: CreateSamplingStore(), as all other CreateXYZ methods, are supposed to be called after the call to `factory.Initialize()`, which sets the primary client. So this null check is not needed.

plugin/storage/es/factory.go
```diff
@@ -317,9 +308,40 @@ func (f *Factory) CreateSamplingStore(maxBuckets int) (samplingstore.Store, erro
...
+		mappingBuilder := mappingBuilderFromConfig(f.primaryConfig)
+		samplingMapping, err := mappingBuilder.GetSamplingMappings()
+		if err != nil {
+			return nil, err
+		}
+		primaryClient := f.getPrimaryClient()
+		if primaryClient == nil {
+			return nil, errors.New("primary client is nil")
+		}
+		templateId := samplingTemplatePrefix + time.Now().UTC().Format(f.primaryConfig.IndexDateLayoutSampling)
```
yurishkuro, line 322: this is not an acceptable hack. None of the other code does this, e.g. in the Writer:
```
	_, err := s.client().CreateTemplate(indexPrefix + "jaeger-span").Body(spanTemplate).Do(context.Background())
```
The template names are supposed to be stable, fixed names, not randomized strings.

plugin/storage/es/factory.go
JaeguKim: 0e90660db25c69d63468aafab357b974cbf92d9e

plugin/storage/es/factory.go
JaeguKim: I updated to use a fixed name and added a prefix to the sampling template name.
8118847da8567f9b95257740bb91213299022667
04ca6ff4033a8f1b2c54909c6259ff1383aeb64d

plugin/storage/integration/elasticsearch_test.go
```diff
@@ -100,9 +96,10 @@ func (s *ESStorageIntegration) initializeES(t *testing.T, allTagsAsFields bool)
 		DiscoverNodesOnStart: false,
 	})
 	require.NoError(t, err)
+	s.bulkProcessor, err = s.client.BulkProcessor().BulkActions(1).FlushInterval(time.Nanosecond).Do(context.Background())
```
yurishkuro, line 99: I don't believe we need bulk processor. First, it's not actually the one that is being used to write the data (which is encapsulated in the Client). Second, I just tried an experiment (#5357) of making esRefresh function no-op, and all the tests passed just fine. I suggest removing esRefresh and removing all mentioning of bulkProcessor in this test.

plugin/storage/integration/elasticsearch_test.go
```diff
@@ -120,48 +117,21 @@ func (s *ESStorageIntegration) esCleanUp(t *testing.T, allTagsAsFields bool) {
...
-	esVersion, err := s.getVersion()
-	require.NoError(t, err)
-	return eswrapper.WrapESClient(s.client, bp, esVersion, s.v8Client)
-}
-
 func (s *ESStorageIntegration) initializeESFactory(t *testing.T, allTagsAsFields bool) *es.Factory {
 	s.logger = zaptest.NewLogger(t)
 	f := es.NewFactory()
 	v, command := config.Viperize(f.AddFlags)
+	esVersion, err := s.getVersion()
```
yurishkuro, line 124: why is this needed? I see you pass it as an argument, but isn't it circular? When the storage is initialized it would execute the same logic to get the version: https://github.com/jaegertracing/jaeger/blob/a9a7bc22df9fd63d7eb591af2ad2c4be730b4831/pkg/es/config/config.go#L170

In fact, I would remove getVersion function from this test altogether (can be a separate PR) - the es.Client already identifies the version, we can simply expose it in its interface: https://github.com/jaegertracing/jaeger/blob/a9a7bc22df9fd63d7eb591af2ad2c4be730b4831/pkg/es/wrapper/wrapper.go#L50

plugin/storage/es/factory.go
```diff
@@ -46,8 +47,9 @@ import (
 )
 
 const (
-	primaryNamespace = "es"
-	archiveNamespace = "es-archive"
+	primaryNamespace   = "es"
+	archiveNamespace   = "es-archive"
+	samplingTemplateId = "jaeger-sampling"
```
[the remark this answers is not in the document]
JaeguKim: f482946a1aba538e2c3711fa38f1bde7d956b16e

plugin/storage/es/factory.go
```diff
@@ -317,9 +306,39 @@ func (f *Factory) CreateSamplingStore(maxBuckets int) (samplingstore.Store, erro
...
+	if f.primaryConfig.CreateIndexTemplates && !f.primaryConfig.UseILM {
+		mappingBuilder := mappingBuilderFromConfig(f.primaryConfig)
+		samplingMapping, err := mappingBuilder.GetSamplingMappings()
+		if err != nil {
+			return nil, err
+		}
+		normalizedPrefix := f.primaryConfig.IndexPrefix
+		if normalizedPrefix != "" && !strings.HasSuffix(normalizedPrefix, "-") {
+			normalizedPrefix += "-"
+		}
```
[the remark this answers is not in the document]
JaeguKim: 61ffa9b0b782281c1c2159914dc34db507780979

plugin/storage/integration/elasticsearch_test.go
JaeguKim: 9f801dd0d23acbc4ced4111cf9b697769838df73

plugin/storage/integration/elasticsearch_test.go
JaeguKim: that logic was redundant. I fixed it in commit b558e9290dd16fd79d50a1a324facab632ccf1f4.
Regarding the removal of the getVersion function, are you planning to handle it as indicated in the TODO comment in the following link?
https://github.com/jaegertracing/jaeger/blob/a9a7bc22df9fd63d7eb591af2ad2c4be730b4831/plugin/storage/integration/elasticsearch_test.go#L245

yurishkuro:
seeing this failure
```
    --- ❌ FAIL: TestIndexCleaner/RemoveDailyIndices_with_adaptiveSampling_prefix,_[] (2.78s)
=== RUN   TestIndexRollover_FailIfILMNotPresent
Error: ILM policy jaeger-ilm-policy doesn't exist in Elasticsearch. Please create it and re-run init
```

plugin/storage/integration/elasticsearch_test.go
```diff
@@ -120,48 +117,18 @@ func (s *ESStorageIntegration) esCleanUp(t *testing.T, allTagsAsFields bool) {
...
 func (s *ESStorageIntegration) initializeESFactory(t *testing.T, allTagsAsFields bool) *es.Factory {
 	s.logger = zaptest.NewLogger(t)
 	f := es.NewFactory()
 	v, command := config.Viperize(f.AddFlags)
 	args := []string{
-		fmt.Sprintf("--es.tags-as-fields.all=%v", allTagsAsFields),
+		fmt.Sprintf("--es.num-shards=%v", 5),
+		fmt.Sprintf("--es.num-replicas=%v", 1),
 		fmt.Sprintf("--es.index-prefix=%v", indexPrefix),
+		fmt.Sprintf("--es.use-ilm=%v", false),
```
yurishkuro, line 128: looks like some of the tests are dependent on ILM. Maybe you need to parameterize this.
```
    --- ❌ FAIL: TestIndexCleaner/RemoveDailyIndices_with_adaptiveSampling_prefix,_[] (2.78s)
=== RUN   TestIndexRollover_FailIfILMNotPresent
Error: ILM policy jaeger-ilm-policy doesn't exist in Elasticsearch. Please create it and re-run init
```

yurishkuro:
@Pushkarm029 this is the PR I was referring to.

yurishkuro:
@JaeguKim could you please rebase?

plugin/storage/integration/elasticsearch_test.go
JaeguKim: It's taking some time to pinpoint the source of the integration test failure among the recent changes I made. It appears that the ILM policy was not created in elasticsearch_test.go. I'm not sure if parameterizing it resolves the issue. I'll take a closer look at it.

JaeguKim:
> @JaeguKim could you please rebase?

@yurishkuro I rebased!

plugin/storage/integration/integration.go
```diff
@@ -299,6 +299,7 @@ func (s *StorageIntegration) testFindTraces(t *testing.T) {
 		}
 		expectedTracesPerTestCase = append(expectedTracesPerTestCase, expected)
 	}
+	time.Sleep(time.Second)
```
JaeguKim, line 302: @yurishkuro I attempted to address the [test failure](https://github.com/jaegertracing/jaeger/actions/runs/8739681859/job/23982877531?pr=5349) occurring with ES7 by incorporating t.Parallel() to ensure the test keeps retrying until 100 seconds have elapsed. Despite this effort, the test continued to fail. While I understand that relying on time.Sleep is not ideal, I'm currently exploring alternative solutions.

yurishkuro: Sleep alone is not a reliable mechanism to deal with concurrency, retries are. We already have sleep inside `waitForCondition`, which is used by `findTracesByQuery` called in this test. Why doesn't that work?

JaeguKim: It turns out that the retry condition in waitForCondition was incorrect. The retry condition should match not only the trace count but also the span count. I addressed this issue in commit 8c8f2e3495b396d41a5191aae1ee3af9ad4671d6. I'm curious why the test only fails intermittently when using other Elasticsearch versions.

yurishkuro (approved):

yurishkuro:
Thanks!

Pushkarm029:
I think we still get this [error](https://github.com/jaegertracing/jaeger/issues/3305#issuecomment-2049269010). Because `CreateLock()` for es is yet to be implemented.

yurishkuro:
> Because CreateLock() for es is yet to be implemented.

yeah, we should add this to integration test

PR #5355, opened 2024-04-12
[jaeger-v2] add badger e2e integration test

akagami-harsh:
- part of #5254
## Description of the changes
- added badger e2e integration test

- tested locally 

## Checklist


akagami-harsh:
integration tests are passing but failing of goroutine leaks

plugin/storage/badger/spanstore/reader.go
```diff
@@ -565,6 +565,10 @@ func (r *TraceReader) scanIndexKeys(indexKeyValue []byte, plan *executionPlan) (
 	return indexResults, err
 }
 
+func (s *TraceReader) Close() error {
```
yurishkuro, line 568: Why is it necessary? The reader/writer interfaces do not require Close.

akagami-harsh: i was getting this error 
```bash
panic: interface conversion: *spanstore.TraceReader is not io.Closer: missing method Close [recovered]
        panic: interface conversion: *spanstore.TraceReader is not io.Closer: missing method Close
```
when on calling `e2eCleanUp` in the test, that is why i add added `Close` method
 ```
func (s *E2EStorageIntegration) e2eCleanUp(t *testing.T) {
	require.NoError(t, s.SpanReader.(io.Closer).Close())
	require.NoError(t, s.SpanWriter.(io.Closer).Close())
}
```

File: cmd/jaeger/internal/integration/badger_test.go [hunk 2]
```diff
@@ -0,0 +1,73 @@
...
+	require.NoError(t, err)
+
+	s.SpanWriter, err = s.factory.CreateSpanWriter()
+	require.NoError(t, err)
+
+	s.SpanReader, err = s.factory.CreateSpanReader()
+	require.NoError(t, err)
+
+	s.SamplingStore, err = s.factory.CreateSamplingStore(0)
+	require.NoError(t, err)
```
james-ryans, line 37: I don't think this is correct. These lines of code replace the initialized SpanWriter and SpanReader from e2eInitialize() function, which writes span data to the OTLP receiver and reads from jaeger_query. But these read/write from badger storage directly and are not the goal of integration tests in `cmd/jaeger/internal/integration`.

plugin/storage/badger/spanstore/reader.go
yurishkuro: I suggest we change that function to be conditional, e.g.
```
if closer, ok := s.SpanReader.(io.Closer); ok {
  require.NoError(t, closer.Close())
}
```

cmd/jaeger/internal/integration/badger_test.go
yurishkuro: Overall my preference would be to have the individual test be only responsible for constructing a factory and passing it over to the overall framework. Right now we're still repeating a lot of boilerplate code in each storage test.

yurishkuro: And in case of e2e tests even the factory is not necessary, only implementing the cleanup() function is needed to delete data between tests (incidentally, if we were to better support tenancy, even deletion of the data would not be necessary).

akagami-harsh: >I don't think this is correct. These lines of code replace the initialized SpanWriter and SpanReader from e2eInitialize() function, which writes span data to the OTLP receiver and reads from jaeger_query. But these read/write from badger storage directly and are not the goal of integration tests in cmd/jaeger/internal/integration.

oops sorry, i'll fix it asap

plugin/storage/badger/spanstore/reader.go
akagami-harsh: i think we don't need to use it because now we are using SpanWriter and SpanReader from e2eInitialize(), which implements `Close` method

akagami-harsh:
Some tests are failing, but pass when run individually.

.github/workflows/ci-badger.yaml
```diff
@@ -31,7 +34,17 @@ jobs:
         go-version: 1.22.x
 
     - name: Run Badger storage integration tests
-      run: make badger-storage-integration-test
+      run: |
+        case ${{ matrix.version }} in
+          v1)
+            SPAN_STORAGE_TYPE=badger \
```
yurishkuro, line 40: This wasn't passed before, is it needed?

cmd/jaeger/internal/integration/badger_test.go
```diff
@@ -0,0 +1,64 @@
...
+	s.initialize(t)
+}
+
+func TestBadgerStorage(t *testing.T) {
+	integration.SkipUnlessEnv(t, "badger")
+
+	s := &BadgerStorageIntegration{}
+	s.ConfigFile = "cmd/jaeger/badger_config.yaml"
+	s.SkipBinaryAttrs = true
```
yurishkuro, line 55:
```suggestion
	s := &BadgerStorageIntegration{
	    ConfigFile: "cmd/jaeger/badger_config.yaml"
	    SkipBinaryAttrs: true,
   }
```

cmd/jaeger/internal/integration/badger_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+	err := s.factory.Initialize(metrics.NullFactory, zap.NewNop())
+	require.NoError(t, err)
+
+	s.Refresh = func(_ *testing.T) {}
+	s.CleanUp = s.cleanUp
+
+	s.logger, _ = testutils.NewLogger()
+
+	// TODO: remove this badger supports returning spanKind from GetOperations
```
yurishkuro, line 35:
```suggestion
	// TODO: remove this once badger supports returning spanKind from GetOperations
```

cmd/jaeger/internal/integration/badger_test.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+type BadgerStorageIntegration struct {
+	E2EStorageIntegration
+	logger  *zap.Logger
+	factory *badger.Factory
+}
+
+func (s *BadgerStorageIntegration) initialize(t *testing.T) {
+	s.factory = badger.NewFactory()
```
yurishkuro, line 25: What is the purpose of the factory if we're using remote storage?

cmd/jaeger/internal/integration/badger_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+	err := s.factory.Initialize(metrics.NullFactory, zap.NewNop())
+	require.NoError(t, err)
+
+	s.Refresh = func(_ *testing.T) {}
+	s.CleanUp = s.cleanUp
+
+	s.logger, _ = testutils.NewLogger()
```
yurishkuro, line 33: I think there is zap.NewNop() for this.

yurishkuro (commented):
Tests may be failing because we do not provide a mechanism to clean up the data between runs. Simply reinitializing the remote storage (by restarting the binary) worked for the memstore, but won't work for persistent backends.

cc @james-ryans

james-ryans:
> Tests may be failing because we do not provide a mechanism to clean up the data between runs. Simply reinitializing the remote storage (by restarting the binary) worked for the memstore, but won't work for persistent backends.
> 
> cc @james-ryans

Yes, it is because the span data are persisted between test cases. In this case for badger, restarting the binary is the only way for now.

yurishkuro:
> Yes, it is because the span data are persisted between test cases. In this case for badger, restarting the binary is the only way for now.

I don't think restarting the binary would help, because the data will still be in the file system (we're using persistent file system as I can see from the config, not the ephemeral mode). We will have similar problems with Cassandra and ES/OS. However, in the case of read DBs we could actually have a separate connection opened from the test that performs a cleanup, but Badger is not a sharable DB, only one process can access the data. So here are some options:

1. we could use ephemeral mode for Badger tests. It's probably the easiest solution, but not ideal since I would prefer e2e tests to be more realistic and people use Badger for its file storage, not for ephemeral mode.
2. we could implement a backdoor specifically in the Badger factory and expose it as an optional endpoint in the remote storage binary (e.g. POST /purge or even DELETE /purge )

I'd say let's reduce the scope and go with (1) for now. This is what the existing integration tests were doing, I believe, since unlike ES/Cassandra tests that were actually deleting data, the Badger test was only closing the factory (so I assume it was using ephemeral mode).

But @james-ryans , didn't your implementation of cleanup already kill & restart the remote storage binary? Or was it only done at the whole test suite scope, not for each test? Is it going to be too expensive to restart for each test?

.github/workflows/ci-badger.yaml
akagami-harsh: No, removed it.

cmd/jaeger/internal/integration/badger_test.go
```diff
@@ -0,0 +1,49 @@
...
+	s.e2eInitialize(t)
+
+	s.Refresh = func(_ *testing.T) {}
+	s.CleanUp = s.cleanUp
+
+	s.logger = zap.NewNop()
+
+	// TODO: remove this once badger supports returning spanKind from GetOperations
+	s.GetOperationsMissingSpanKind = true
+	s.SkipArchiveTest = true
```
yurishkuro, line 29: since these are static, they should be assigned when creating BadgerStorageIntegration struct, not in the initialize function

cmd/jaeger/internal/integration/badger_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+func (s *BadgerStorageIntegration) initialize(t *testing.T) {
+	s.e2eInitialize(t)
+
+	s.Refresh = func(_ *testing.T) {}
+	s.CleanUp = s.cleanUp
+
+	s.logger = zap.NewNop()
+
+	// TODO: remove this once badger supports returning spanKind from GetOperations
```
yurishkuro, line 27:
```suggestion
	// TODO: remove this once badger supports returning spanKind from GetOperations
	// Cf https://github.com/jaegertracing/jaeger/issues/1922
```

cmd/jaeger/internal/integration/badger_test.go
```diff
@@ -37,9 +33,19 @@ func (s *BadgerStorageIntegration) cleanUp(t *testing.T) {
...
-	s.ConfigFile = "cmd/jaeger/badger_config.yaml"
-	s.SkipBinaryAttrs = true
+	s := &BadgerStorageIntegration{
+		E2EStorageIntegration: E2EStorageIntegration{
+			ConfigFile: "cmd/jaeger/badger_config.yaml",
+			StorageIntegration: integration.StorageIntegration{
+				SkipBinaryAttrs: true,
+
+				// TODO: remove this once badger supports returning spanKind from GetOperations
+				// Cf https://github.com/jaegertracing/jaeger/issues/1922
```
yurishkuro, line 43: this comment only applies to span kind flag, not to archive test

akagami-harsh:
tests are passing in my local machine but failing here.

cmd/jaeger/internal/integration/integration.go
```diff
@@ -66,4 +65,5 @@ func (s *E2EStorageIntegration) e2eInitialize(t *testing.T) {
 func (s *E2EStorageIntegration) e2eCleanUp(t *testing.T) {
 	require.NoError(t, s.SpanReader.(io.Closer).Close())
 	require.NoError(t, s.SpanWriter.(io.Closer).Close())
+	require.NoError(t, s.jaegerBinary.Process.Kill())
```
yurishkuro, line 68: I feel like this should be conditional. Badger is the only implementation that requires killing the collector binary between tests, all other storage backends will be fine without it. cc @james-ryans

james-ryans:
> I don't think restarting the binary would help, because the data will still be in the file system (we're using persistent file system as I can see from the config, not the ephemeral mode).

Oh, I see.

> But @james-ryans , didn't your implementation of cleanup already kill & restart the remote storage binary? Or was it only done at the whole test suite scope, not for each test? Is it going to be too expensive to restart for each test?

It is only killed after the entire test suite is finished. As for whether it is expensive or not, I've never thought about it because grpc_test didn't need it and did not foresee encountering this with Badger.

cmd/jaeger/internal/integration/integration.go
james-ryans: Yes, this should be conditional. But I'm thinking is there a way to kill and start the binary synchronously? Because restarting it between tests brings 2 issues:
1. The new binary doesn't wait until the previous binary has successfully shutdown thus binding port will collide (as we can see in the new CIT Badger (v2) we encountered `listen tcp :8888: bind: address already in use`)
2. SpanReader and SpanWriter will receive `connection refused` error for a while until the new binary is up. While SpanReader doesn't have any problem with that because It will keep retrying for 100 secs, SpanWriter will fail instantly because the current implementation doesn't expect the `WriteSpan` to fail.
    - Perhaps we can solve this by also retrying the write until it doesn't have any error but not sure if this will cause duplicate data in some cases.
    - I've just found out that the OTLP exporter has `WaitForReady` setting that instead of throwing errors, we can hold the RPC calls until the connection has successfully established. Perhaps this is worth trying.

yurishkuro: Very good points. My preferences would be not to use killing a binary as a way to clean up the data, it seems like a massive overkill, and causes all these other issues. It will also not help if we want to use badger with real file system. 

So what can we do that is not very intrusive?

yurishkuro:
```mermaid
flowchart LR
    Receiver --> Processor
    Processor --> Exporter
    JaegerStorageExension -->|"(1) get storage"| Exporter
    Exporter -->|"(2) write trace"| Badger

    Badger_e2e_test -->|"(1) POST /cleanup"| HTTP_endpoint
    JaegerStorageExension -->|"(2) getStorage()"| HTTP_endpoint
    HTTP_endpoint -.->|"(3) storage.(*Badger).cleanup()"| Badger

    subgraph Jaeger Collector
        Receiver
        Processor
        Exporter
        
        Badger
        BadgerCleanerExtension
        HTTP_endpoint
        subgraph JaegerStorageExension
            Badger
        end
        subgraph BadgerCleanerExtension
            HTTP_endpoint
        end
    end
```

This is the solution I was thinking with the custom extension. It's pretty clean from implementation point of view, the only issue I see is that the plan for all e2e test was to use the official configs, e.g. `badger_config.yaml` in this case. But those are "production" configs, so this test-only extension does not belong there. I think the workaround for this is to have a pre-precessing script (before starting the collector binary) that will read the official `badger_config.yaml`, inject a section with the test extension, and use that output as the config for the binary.

yurishkuro: BTW, before we jump on this approach, we need to de-risk it by implementing the clean-up functionality and test it in the regular integration test (i.e. use persistent badger instead of reinitializing ephemeral one between tests). We can do this as a separate PR.

cmd/jaeger/badger_config.yaml
```diff
@@ -15,17 +15,19 @@ extensions:
...
+        span_store_ttl: 72h
       badger_archive:
-        directory_key: "/tmp/jaeger_archive/"
-        directory_value: "/tmp/jaeger_archive/"
+        directory_key: /tmp/jaeger_archive/
+        directory_value: /tmp/jaeger_archive/
         ephemeral: false
         maintenance_interval: 5
         metrics_update_interval: 10
+        span_store_ttl: 72h
```
yurishkuro, line 30: as an illustration this should be longer, e.g. 30d

cmd/jaeger/internal/integration/badgercleaner/factory.go
```diff
@@ -0,0 +1,34 @@
...
+		createExtension,
+		component.StabilityLevelBeta,
+	)
+}
+
+func createDefaultConfig() component.Config {
+	return &Config{}
+}
+
+func createExtension(_ context.Context, set extension.CreateSettings, cfg component.Config) (extension.Extension, error) {
```
yurishkuro, line 32:
```suggestion
func createExtension(
	_ context.Context,
	set extension.CreateSettings,
	cfg component.Config,
) (extension.Extension, error) {
```

cmd/jaeger/internal/integration/badgercleaner/extension.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/jaegertracing/jaeger/plugin/storage/badger"
+)
+
+var (
+	_ extension.Extension = (*cleaner)(nil)
+	_ extension.Dependent = (*cleaner)(nil)
+)
+
+const (
+	cleanerPort = "9231"
```
yurishkuro, line 26: this should be part of the config

scripts/prepare-badger-integration-tests.py
```diff
@@ -0,0 +1,18 @@
...
+    config = yaml.safe_load(f)
+if 'badger_cleaner' not in config['service']['extensions'] :
+    config['service']['extensions'].insert(1, 'badger_cleaner') 
+
+config['extensions']['badger_cleaner'] = {
+    'trace_storage': 'badger_main'
+}
+
+with open(config_file, 'w') as f:
+    yaml.dump(config, f, sort_keys=False)
```
yurishkuro, line 18: You don't want to change the file in place, because then it can be accidentally committed. Let's create a new temporary file.

scripts/prepare-badger-integration-tests.py
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env python3
```
yurishkuro, line 1: doesn't look like this script is called from anywhere

cmd/jaeger/internal/integration/badger_test.go
```diff
@@ -0,0 +1,75 @@
...
+func (s *BadgerStorageIntegration) cleanUp(t *testing.T) {
+	r, err := http.NewRequest(http.MethodPost, cleanerAddr+cleanerURL, nil)
+	require.NoError(t, err)
+
+	client := &http.Client{}
+
+	resp, err := client.Do(r)
+	require.NoError(t, err)
+	defer resp.Body.Close()
+}
```
yurishkuro, line 45: you should test the status code

cmd/jaeger/internal/integration/badgercleaner/README.md
```diff
@@ -0,0 +1,47 @@
+# badger_cleaner
```
yurishkuro, line 1: We already discovered a similar need to have purger for ES tests. So I would like to refactor this extension so that it never mentions Badger, it should be completely generic and work with any factory. We need to add a new interface `storage.Purger` to the main storage package and do a check in this extension if the factory implements it (error out if it does not).

Does not need to be in this PR, let's first make sure Badger tests work.

cmd/jaeger/internal/integration/badgercleaner/extension.go
```diff
@@ -0,0 +1,92 @@
...
+		}
+		return nil
+	}
+
+	cleanerHandler := func(w http.ResponseWriter, r *http.Request) {
+		if err := purgeBadger(); err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+		w.WriteHeader(http.StatusCreated)
```
yurishkuro, line 60: why Created? Usually OK or Accepted are used.

cmd/jaeger/internal/integration/badgercleaner/.nocover
```diff
@@ -0,0 +1 @@
+FIXME
```
yurishkuro, line 1: are you planning to add tests in this PR?

akagami-harsh: yes, i am working on it.

scripts/prepare-badger-integration-tests.py
akagami-harsh: it is called form the `badger_test.go`

scripts/prepare-badger-integration-tests.py
akagami-harsh: modified the script to create a temp file for the new config. and also added a mechanism in the test to remove the temp config file.

scripts/prepare-badger-integration-tests.py
yurishkuro: Oh, in this case could we not do this in Go?

cmd/jaeger/badger_config.yaml
```diff
@@ -20,12 +20,14 @@ extensions:
...
         maintenance_interval: 5
         metrics_update_interval: 10
+        span_store_ttl: 720h
       badger_archive:
         directory_key: "/tmp/jaeger_archive/"
         directory_value: "/tmp/jaeger_archive/"
         ephemeral: false
         maintenance_interval: 5
         metrics_update_interval: 10
+        span_store_ttl: 720h
```
yurishkuro, line 30: 30d, why do this in hours?

cmd/jaeger/badger_config.yaml
```diff
@@ -20,12 +20,14 @@ extensions:
...
```
yurishkuro, line 23:
```suggestion
        span_store_ttl: 72h
```

cmd/jaeger/internal/integration/badger_test.go
```diff
@@ -0,0 +1,87 @@
...
+	s.RunAll(t)
+}
+
+func createBadgerCleanerConfig(t *testing.T) string {
+	cmd := exec.Command("../../../../scripts/prepare-badger-integration-tests.py")
+	data, err := cmd.Output()
+	require.NoError(t, err)
+	tempFile := string(data)
+	tempFile = strings.ReplaceAll(tempFile, "\n", "")
+	return tempFile
```
yurishkuro, line 81: Can rm file via t.cleanup here

cmd/jaeger/internal/integration/badgercleaner/config.go
```diff
@@ -0,0 +1,18 @@
...
+
+package badgercleaner
+
+import (
+	"github.com/asaskevich/govalidator"
+)
+
+type Config struct {
+	TraceStorage string `valid:"required" mapstructure:"trace_storage"`
+	cleanerPort  string `mapstructure:"cleaner_port"`
```
yurishkuro, line 12: "cleaner" is redundant, it's clear from context

cmd/jaeger/internal/integration/badgercleaner/extension.go
```diff
@@ -0,0 +1,95 @@
...
+		}
+	}()
+
+	return nil
+}
+
+func (c *cleaner) Shutdown(ctx context.Context) error {
+	if c.server != nil {
+		err := c.server.Shutdown(ctx)
+		if err != nil {
```
yurishkuro, line 86: Combine into one line

cmd/jaeger/internal/integration/badgercleaner/factory.go
```diff
@@ -0,0 +1,38 @@
...
+	return extension.NewFactory(
+		componentType,
+		createDefaultConfig,
+		createExtension,
+		component.StabilityLevelBeta,
+	)
+}
+
+func createDefaultConfig() component.Config {
+	return &Config{}
```
yurishkuro, line 29: Port should be defaulted here

cmd/jaeger/badger_config.yaml
akagami-harsh: because i was getting this error :
```
 * error decoding 'badger[badger_main].span_store_ttl': time: unknown unit "d" in duration "30d"
2024/04/21 20:19:40 failed to get config: cannot unmarshal the configuration: 1 error(s) decoding:
```

cmd/jaeger/internal/integration/badger_test.go
akagami-harsh: yes, but we will need to do some refactoring in the test. because the function that removes the file is a method of `BadgerStorageIntegration` and the above function is not.

cmd/jaeger/internal/integration/badger_test.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/stretchr/testify/require"
+	"go.uber.org/zap"
+
+	"github.com/jaegertracing/jaeger/plugin/storage/integration"
+)
+
+const (
+	host = "0.0.0.0"
+	Port = "9231"
+	URL  = "/purge"
```
yurishkuro, line 22: I would recommend moving these to storagecleaner package as public constants

cmd/jaeger/internal/integration/badger_test.go
yurishkuro: it does not need to be a method. `createBadgerCleanerConfig(t)` creates a temp file, it can also register a cleanup hook to remove that file once the test is done. 

An even better approach is to use tempFile functionality of `testing.T`, especially if you refactor to use Go code for enriching the config instead of depending on external Python script.

yurishkuro:
so Badger tests are still failing, what's the cause?

akagami-harsh:
> so Badger tests are still failing, what's the cause?

i wasn't able to figure it out yet because the tests sometimes passes (like this [run](https://github.com/jaegertracing/jaeger/actions/runs/8779342177/job/24087228968?pr=5355)) but it fails most of the time. And also on my local machine it passes every time i run it.

cmd/jaeger/internal/integration/storagecleaner/extension.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+func (c *storageCleaner) Start(ctx context.Context, host component.Host) error {
+	storageFactory, err := jaegerstorage.GetStorageFactory(c.config.TraceStorage, host)
+	if err != nil {
+		return fmt.Errorf("cannot find storage factory for Badger: %w", err)
+	}
+
+	purgeBadger := func() error {
+		badgerFactory := storageFactory.(*badger.Factory)
```
yurishkuro, line 48: Don't need to depend on badger explicitly, create storage.Purger interface with a single method and make Badger implement it.

cmd/jaeger/internal/integration/badger_test.go
```diff
@@ -0,0 +1,86 @@
...
+	"go.uber.org/zap"
+
+	"github.com/jaegertracing/jaeger/cmd/jaeger/internal/integration/storagecleaner"
+	"github.com/jaegertracing/jaeger/plugin/storage/integration"
+)
+
+const (
+	host = "0.0.0.0"
+	Addr = "http://" + host + ":" + storagecleaner.Port + storagecleaner.URL
+)
```
yurishkuro, line 23: Don't really need these as constants. Can just do fmt.Sprintf inside cleanup function.

cmd/jaeger/internal/integration/badger_test.go
akagami-harsh: > An even better approach is to use tempFile functionality of testing.T, especially if you refactor to use Go code for enriching the config instead of depending on external Python script.

i tried doing it in Go. it's a lot complicated and in golang.

cmd/jaeger/internal/integration/storagecleaner/extension.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+func (c *storageCleaner) Start(ctx context.Context, host component.Host) error {
+	storageFactory, err := jaegerstorage.GetStorageFactory(c.config.TraceStorage, host)
+	if err != nil {
+		return fmt.Errorf("cannot find storage factory for Badger: %w", err)
+	}
+
+	purgeStorage := func() error {
+		purger := storageFactory.(storage.Purger)
```
yurishkuro, line 48: we should not assume every factory implements this.
```suggestion
		purger, ok := storageFactory.(storage.Purger)
		if !ok {
		  return fmt.Errorf("storage %s does not implement Purger interface", c.config.TraceStorage)
		}
```

storage/factory.go
```diff
@@ -49,6 +49,12 @@ type Factory interface {
 	CreateDependencyReader() (dependencystore.Reader, error)
 }
 
+// Purger defines an interface that is capable of purging the storage.
```
yurishkuro, line 52:
```suggestion
// Purger defines an interface that is capable of purging the storage.
// Only meant to be used from integration tests.
```

cmd/jaeger/internal/integration/badger_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+func (s *BadgerStorageIntegration) initialize(t *testing.T) {
+	s.e2eInitialize(t)
+
+	s.CleanUp = s.cleanUp
+	s.logger = zap.NewNop()
+}
+
+func (s *BadgerStorageIntegration) cleanUp(t *testing.T) {
+	Addr := fmt.Sprintf("http://%s:%s%s", "0.0.0.0", storagecleaner.Port, storagecleaner.URL)
```
yurishkuro, line 34:
```suggestion
	addr := fmt.Sprintf("http://0.0.0.0:%s%s", storagecleaner.Port, storagecleaner.URL)
```

cmd/jaeger/internal/integration/badger_test.go
yurishkuro: You're sure? You can parse into a dynamic structure
```go
  var config map[interface{}]interface{}
  err := yaml.Unmarshal(data, &config)
  require.NoError(t, err)

  service, ok := config["service"].(map[interface{}]interface{})
  require.True(t, ok)

  // etc, and at some point add your own map

  newData, err := yaml.Marshal(config)
  // write to temp file
```

cmd/jaeger/internal/integration/storagecleaner/README.md
```diff
@@ -0,0 +1,47 @@
+# storage_cleaner
+
+This module implements an extension that allows purging the backend storage by exposing a HTTP endpoint for making POST requests.
```
yurishkuro, line 3:
```suggestion
This module implements an extension that allows purging the backend storage by making an HTTP POST request to it. 
```

cmd/jaeger/internal/integration/storagecleaner/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+The storage_cleaner extension is intended to be used only in tests, providing a way to clear the storage between test runs. Making a POST request to the exposed endpoint will delete all data in storage.
+
+
+```mermaid
+flowchart LR
+    Receiver --> Processor
+    Processor --> Exporter
+    JaegerStorageExension -->|"(1) get storage"| Exporter
+    Exporter -->|"(2) write trace"| Badger
```
yurishkuro, line 13: suggest replacing `Badger` with `Storage`, and `Badger_e2e_test` with `E2E_test`

cmd/jaeger/internal/integration/storagecleaner/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+    Badger_e2e_test -->|"(1) POST /purge"| HTTP_endpoint
+    JaegerStorageExension -->|"(2) getStorage()"| HTTP_endpoint
+    HTTP_endpoint -.->|"(3) storage.(*Badger).Purge()"| Badger
```
yurishkuro, line 17:
```suggestion
    HTTP_endpoint -.->|"(3) storage.(*storage.Purger).Purge()"| Badger
```

cmd/jaeger/internal/integration/storagecleaner/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+    subgraph Jaeger Collector
+        Receiver
+        Processor
+        Exporter
+        
+        Badger
+        BadgerCleanerExtension
```
yurishkuro, line 25:
```suggestion
        StorageCleanerExtension
```

cmd/jaeger/internal/integration/storagecleaner/README.md
```diff
@@ -0,0 +1,47 @@
...
+# Getting Started
+
+The following settings are required:
+
+- `trace_storage` : name of a storage backend defined in `jaegerstorage` extension
+
+```yaml
+extensions:
+  storage_cleaner:
+    trace_storage: external-storage
```
yurishkuro, line 45:
```suggestion
    trace_storage: storage_name
```

cmd/jaeger/internal/integration/badger_test.go
```diff
@@ -0,0 +1,80 @@
...
+		},
+	}
+	s.initialize(t)
+	t.Cleanup(func() {
+		s.e2eCleanUp(t)
+	})
+	s.RunAll(t)
+}
+
+func createBadgerCleanerConfig(t *testing.T) string {
```
yurishkuro, line 70: suggest moving to integration_test.go as `createStorageCleanerConfig`, since we may call it from ES tests as well.

In fact, I don't think we should be calling this function explicitly from Badger test at all, just pass the badger config as property to `E2EStorageIntegration` and let it create a temp file with extension (regardless of whether the specific storage test needs it or not). Less code / initialization this way.

cmd/jaeger/internal/integration/badger_test.go
```diff
@@ -0,0 +1,80 @@
...
+type BadgerStorageIntegration struct {
+	E2EStorageIntegration
+	logger *zap.Logger
+}
+
+func (s *BadgerStorageIntegration) initialize(t *testing.T) {
+	s.e2eInitialize(t)
+
+	s.CleanUp = s.cleanUp
+	s.logger = zap.NewNop()
```
yurishkuro, line 30: this can be assigned when the struct is initialized

cmd/jaeger/internal/integration/badger_test.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/jaegertracing/jaeger/cmd/jaeger/internal/integration/storagecleaner"
+	"github.com/jaegertracing/jaeger/plugin/storage/integration"
+)
+
+type BadgerStorageIntegration struct {
+	E2EStorageIntegration
+	logger *zap.Logger
+}
+
+func (s *BadgerStorageIntegration) initialize(t *testing.T) {
```
yurishkuro, line 26: Consider if we actually need this function.

cmd/jaeger/internal/integration/badger_test.go
akagami-harsh: it worked. now we can easily generalize the config for other storages too

cmd/jaeger/internal/integration/badger_test.go
```diff
@@ -0,0 +1,51 @@
...
+			// Cf https://github.com/jaegertracing/jaeger/issues/1922
+			GetOperationsMissingSpanKind: true,
+		},
+	}
+	s.e2eInitialize(t)
+	t.Cleanup(func() {
+		s.e2eCleanUp(t)
+	})
+	s.RunAll(t)
+}
```
yurishkuro, line 51: So much simpler! 10x easier to read. Very nice!

cmd/jaeger/internal/integration/integration.go
```diff
@@ -4,12 +4,14 @@
...
 
 import (
+	"fmt"
 	"io"
 	"os"
 	"os/exec"
 	"testing"
 
 	"github.com/stretchr/testify/require"
+	"gopkg.in/yaml.v2"
```
yurishkuro, line 14: not `gopkg.in/yaml.v3`?

cmd/jaeger/internal/integration/integration.go
```diff
@@ -67,3 +71,41 @@ func (s *E2EStorageIntegration) e2eCleanUp(t *testing.T) {
...
+	extensions, ok := config["extensions"].(map[interface{}]interface{})
+	require.True(t, ok)
+	trace_storage := getTraceStorage(extensions)
+	extensions["storage_cleaner"] = map[string]string{"trace_storage": trace_storage}
+
+	newData, err := yaml.Marshal(config)
+	require.NoError(t, err)
+	tempFile, err := os.Create("storageCleaner_config.yaml")
+	require.NoError(t, err)
+	err = os.WriteFile(tempFile.Name(), newData, 0o600)
```
yurishkuro, line 95:
```suggestion
	tempFile := filepath.Join(t.TempDir(), "mapping.json")
	err = os.WriteFile(tempFile, newData, 0o600)
```
the dir will be automatically deleted, you don't need to rm it manually

cmd/jaeger/internal/integration/integration.go
```diff
@@ -38,7 +40,9 @@ type E2EStorageIntegration struct {
 // This function should be called before any of the tests start.
 func (s *E2EStorageIntegration) e2eInitialize(t *testing.T) {
 	logger, _ := testutils.NewLogger()
+	s.ConfigFile = "./cmd/jaeger/internal/integration/" + createStorageCleanerConfig(t, s.ConfigFile)
```
yurishkuro, line 43:
```suggestion
	s.ConfigFile = createStorageCleanerConfig(t, s.ConfigFile)
```

cmd/jaeger/internal/integration/integration.go
```diff
@@ -67,3 +71,41 @@ func (s *E2EStorageIntegration) e2eCleanUp(t *testing.T) {
...
+	require.NoError(t, err)
+	tempFile, err := os.Create("storageCleaner_config.yaml")
+	require.NoError(t, err)
+	err = os.WriteFile(tempFile.Name(), newData, 0o600)
+	require.NoError(t, err)
+
+	t.Cleanup(func() {
+		os.Remove(tempFile.Name())
+	})
+	return tempFile.Name()
```
yurishkuro, line 101:
```suggestion
	return tempFile
```

cmd/jaeger/internal/integration/integration.go
```diff
@@ -67,3 +71,41 @@ func (s *E2EStorageIntegration) e2eCleanUp(t *testing.T) {
...
+func createStorageCleanerConfig(t *testing.T, configFile string) string {
+	data, err := os.ReadFile(configFile)
+	require.NoError(t, err)
+	var config map[interface{}]interface{}
+	err = yaml.Unmarshal(data, &config)
+	require.NoError(t, err)
+
+	service, ok := config["service"].(map[interface{}]interface{})
+	require.True(t, ok)
+	service["extensions"] = []string{"jaeger_storage", "jaeger_query", "storage_cleaner"}
```
yurishkuro, line 84: this is not safe assumption - you should read the array and append to it

cmd/jaeger/internal/integration/integration.go
```diff
@@ -67,3 +71,41 @@ func (s *E2EStorageIntegration) e2eCleanUp(t *testing.T) {
...
+	require.NoError(t, err)
+
+	t.Cleanup(func() {
+		os.Remove(tempFile.Name())
+	})
+	return tempFile.Name()
+}
+
+func getTraceStorage(extensions map[interface{}]interface{}) string {
+	for k := range extensions["jaeger_query"].(map[interface{}]interface{}) {
```
yurishkuro, line 105: why do you need to loop instead of just retrieving by key?

yurishkuro (approved):

cmd/jaeger/internal/integration/integration.go
akagami-harsh: this returns false when using v3 but works fine with v2
```go
service, ok := config["service"].(map[interface{}]interface{})
```

cmd/jaeger/internal/integration/integration.go
```diff
@@ -38,7 +41,9 @@ type E2EStorageIntegration struct {
 // This function should be called before any of the tests start.
 func (s *E2EStorageIntegration) e2eInitialize(t *testing.T) {
 	logger, _ := testutils.NewLogger()
+	s.ConfigFile = createStorageCleanerConfig(t, s.ConfigFile)
 
+	fmt.Println(s.ConfigFile)
```
yurishkuro, line 46: can remove this

cmd/jaeger/internal/integration/integration.go
```diff
@@ -38,7 +41,9 @@ type E2EStorageIntegration struct {
 // This function should be called before any of the tests start.
 func (s *E2EStorageIntegration) e2eInitialize(t *testing.T) {
 	logger, _ := testutils.NewLogger()
+	s.ConfigFile = createStorageCleanerConfig(t, s.ConfigFile)
```
yurishkuro, line 44:
```suggestion
	configFile := createStorageCleanerConfig(t, s.ConfigFile)
```
it's better not to override the field in the struct, you only need this value in L49

cmd/jaeger/internal/integration/integration.go
```diff
@@ -38,7 +41,9 @@ type E2EStorageIntegration struct {
...
 
+	fmt.Println(s.ConfigFile)
 	cmd := exec.Cmd{
 		Path: "./cmd/jaeger/jaeger",
 		Args: []string{"jaeger", "--config", s.ConfigFile},
```
yurishkuro, line 49: curious, what does the first argument "jaeger" refer to?

cmd/jaeger/internal/integration/integration.go
yurishkuro: odd. Perhaps instead of doing conditional cast and checking OK we could do `require.IsType()` first, and if the types do not match it should print the actual type

akagami-harsh: it worked. the type was `map[string]interface{}`

cmd/jaeger/internal/integration/integration.go
akagami-harsh: Args slice includes the actual command itself as the first element `(Args[0])`.

cmd/jaeger/internal/integration/integration.go
```diff
@@ -38,10 +40,11 @@ type E2EStorageIntegration struct {
 // This function should be called before any of the tests start.
 func (s *E2EStorageIntegration) e2eInitialize(t *testing.T) {
 	logger, _ := testutils.NewLogger()
+	ConfigFile := createStorageCleanerConfig(t, s.ConfigFile)
```
yurishkuro, line 43:
```suggestion
	configFile := createStorageCleanerConfig(t, s.ConfigFile)
```

yurishkuro (commented):
extension.go is having many coverage misses, do you want to add tests?

https://github.com/jaegertracing/jaeger/pull/5355/checks?check_run_id=24174925354

cmd/jaeger/internal/integration/storagecleaner/extension_test.go
```diff
@@ -0,0 +1,92 @@
...
+		if err == nil {
+			ok = true
+			break
+		}
+		time.Sleep(1 * time.Second)
+	}
+	require.True(t, ok, "could not reach storage cleaner server")
+	defer resp.Body.Close()
+	require.Equal(t, http.StatusOK, resp.StatusCode)
+	s.Dependencies()
```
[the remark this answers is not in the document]
akagami-harsh: No!, removed it because the mock `purge()` will never return err

cmd/jaeger/internal/integration/storagecleaner/extension.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	r := mux.NewRouter()
+	r.HandleFunc(URL, purgeHandler).Methods(http.MethodPost)
+	c.server = &http.Server{
+		Addr:              ":" + c.config.Port,
+		Handler:           r,
+		ReadHeaderTimeout: 3 * time.Second,
+	}
+	go func() error {
```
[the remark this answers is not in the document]
akagami-harsh: fixed it!

cmd/jaeger/internal/integration/storagecleaner/extension_test.go
```diff
@@ -0,0 +1,84 @@
...
+	Addr := fmt.Sprintf("http://0.0.0.0:%s%s", Port, URL)
+	client := &http.Client{}
+	var resp *http.Response
+	require.Eventually(t, func() bool {
+		r, err := http.NewRequest(http.MethodPost, Addr, nil)
+		require.NoError(t, err)
+		resp, err = client.Do(r)
+		return err == nil
+	}, 5*time.Second, 100*time.Millisecond)
+	defer resp.Body.Close()
```
yurishkuro, line 68: lambda runs in a loop. You need to close the body inside the lambda, not outside, otherwise it may leak goroutines.

Something like this:
```
resp, err = client.Do(r)
require.NoError(t, err)
defer resp.Body.Close()
return resp.Status == OK
```

cmd/jaeger/internal/integration/storagecleaner/extension_test.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+	s := newStorageCleaner(config, component.TelemetrySettings{})
+	host := storagetest.NewStorageHost()
+	host.WithExtension(jaegerstorage.ID, &mockStorageExt{
+		name:    "storage",
+		factory: &factoryMocks.Factory{},
+	})
+	ctx := context.Background()
+	err := s.Start(ctx, host)
+	require.NoError(t, err)
```
yurishkuro, line 57:
```suggestion
	require.NoError(t, err)
	defer s.Shutdown(context.Background())
```

cmd/jaeger/internal/integration/storagecleaner/extension_test.go
```diff
@@ -0,0 +1,84 @@
...
+	var resp *http.Response
+	require.Eventually(t, func() bool {
+		r, err := http.NewRequest(http.MethodPost, Addr, nil)
+		require.NoError(t, err)
+		resp, err = client.Do(r)
+		return err == nil
+	}, 5*time.Second, 100*time.Millisecond)
+	defer resp.Body.Close()
+	s.Dependencies()
+	require.NoError(t, s.Shutdown(ctx))
```
yurishkuro, line 70: do this after Start via defer

storage/mocks/Factory.go
```diff
@@ -111,5 +111,9 @@ func (_m *Factory) Initialize(metricsFactory metrics.Factory, logger *zap.Logger
 
 	return r0
 }
+func (_m *Factory) Purge() error {
+	return nil
+}
 
 var _ storage.Factory = (*Factory)(nil)
+var _ storage.Purger = (*Factory)(nil)
```
yurishkuro, line 119: we don't want to do that. Purger is an optional interface, by force-adding it to the mock you cannot test cases when factory does not implement it. Instead, in your test you can do this:
```
type PurgerFactory struct {
	factoryMocks.Factory
	err error
}

var _ storage.Purger = (*PurgerFactory)(nil)

func (f *PurgerFactory) Purge() error {
	return f.err
}
```

cmd/jaeger/internal/integration/storagecleaner/extension_test.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/jaegertracing/jaeger/cmd/jaeger/internal/extension/jaegerstorage"
+	"github.com/jaegertracing/jaeger/storage"
+	factoryMocks "github.com/jaegertracing/jaeger/storage/mocks"
+)
+
+var _ jaegerstorage.Extension = (*mockStorageExt)(nil)
+
+type mockStorageExt struct {
+	name    string
+	factory *factoryMocks.Factory
```
yurishkuro, line 26:
```suggestion
	factory storage.Factory
```

cmd/jaeger/internal/integration/storagecleaner/extension_test.go
```diff
@@ -0,0 +1,84 @@
...
+func TestStorageCleanerExtension(t *testing.T) {
+	config := &Config{
+		TraceStorage: "storage",
+		Port:         Port,
+	}
+	s := newStorageCleaner(config, component.TelemetrySettings{})
+	host := storagetest.NewStorageHost()
+	host.WithExtension(jaegerstorage.ID, &mockStorageExt{
+		name:    "storage",
+		factory: &factoryMocks.Factory{},
```
yurishkuro, line 53: see comment in the mock file

cmd/jaeger/internal/integration/storagecleaner/extension_test.go
```diff
@@ -0,0 +1,84 @@
...
+	})
+	ctx := context.Background()
+	err := s.Start(ctx, host)
+	require.NoError(t, err)
+
+	Addr := fmt.Sprintf("http://0.0.0.0:%s%s", Port, URL)
```
yurishkuro, line 59:
```suggestion
	addr := fmt.Sprintf("http://0.0.0.0:%s%s", Port, URL)
```
local variables are always lower-case

cmd/jaeger/internal/integration/storagecleaner/extension_test.go
```diff
@@ -0,0 +1,84 @@
...
+
+func (m *mockStorageExt) Factory(name string) (storage.Factory, bool) {
+	if m.name == name {
+		return m.factory, true
+	}
+	return nil, false
+}
+
+func TestStorageCleanerExtension(t *testing.T) {
+	config := &Config{
```
yurishkuro, line 45: you can parameterize this test to test more conditions:
```
	tests := []struct {
		name    string
		factory storage.Factory
		status  int
	}{
		{"good storage", &PurgerFactory{}, http.StatusOK},
		{"good storage with error", &PurgerFactory{err: fmt.Errorf("error")}, http.StatusInternalServerError},
		{"bad storage", &factoryMocks.Factory{}, http.StatusInternalServerError},
	}

	for _, test := range tests {
		t.Run(test.name, func(t *testing.T) {
			// your test here, using values from `test`
```

yurishkuro:
we're on a final stretch, let's finish it @akagami-harsh

akagami-harsh:
> nish it @akagami-harsh

yes i had an exam yesterday. i am working on it now

Issue #5350, opened 2024-04-11
[test] Crossdock tests do not need to build Docker images for all architectures


yurishkuro:
I just noticed in the [crossdock CI test logs](https://github.com/jaegertracing/jaeger/actions/runs/8649480331/job/23715804829?pr=5347):
```
#17 [linux/arm64 cert 2/2] RUN apk add --update --no-cache ca-certificates mailcap
#17 2.550 Executing ca-certificates-20240226-r0.trigger
#17 ...
#18 [linux/ppc64le stage-1 2/3] COPY --from=cert /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
#18 DONE 0.0s
#19 [linux/ppc64le stage-1 3/3] COPY --from=cert /etc/mime.types /etc/mime.types
#19 DONE 0.0s
#20 [linux/s390x cert 2/2] RUN apk add --update --no-cache ca-certificates mailcap
```
It means we're building images for arm64, ppc64le, s390x, etc. while we only need arm64 linux images for this test. We already have a shortcut for not building other architectures in the `docker-images` CI workflow for PRs, we should do the same for crossdock in all situations. https://github.com/jaegertracing/jaeger/blob/14aa43d4929ef344df722081aafd1a76c93c4f22/.github/workflows/ci-docker-build.yml#L53

PR #5356, opened 2024-04-13
CI-Crossdock removed unnecessary arch builds in docker image

Wise-Wizard:

This PR addresses the issue [#5350 ](https://github.com/jaegertracing/jaeger/issues/5350).

Description:
Initialized QEMU only for linux/arm64 architecture.

```bash
make test
```

**Checklist**


.github/workflows/ci-crossdock.yml
```diff
@@ -43,9 +43,13 @@ jobs:
     - name: Install tools
       run: make install-ci
 
-    - uses: docker/setup-qemu-action@68827325e0b33c7199eb31dd4e31fbe9023e06e3 # v3.0.0
+    # Conditionally setup QEMU only for arm64 Linux architecture
+    - name: Setup QEMU for arm64 Linux
+      if: runner.os == 'Linux' && runner.arch == 'arm64'
```
yurishkuro, line 48: This condition is always true, and the build is still going to go through all arch

yurishkuro (changes requested):

.github/workflows/ci-crossdock.yml
Wise-Wizard: Noted.

yurishkuro:
Please use descriptive title for pr

scripts/build-crossdock.sh
```diff
@@ -11,14 +11,19 @@ make build-and-run-crossdock
 if [[ "$BRANCH" == "main" ]]; then
   echo 'upload images to dockerhub/quay.io'
   REPO=jaegertracing/test-driver
-  IFS=" " read -r -a IMAGE_TAGS <<< "$(bash scripts/compute-tags.sh ${REPO})"
-  IMAGE_TAGS+=("--tag" "docker.io/${REPO}:${COMMIT}" "--tag" "quay.io/${REPO}:${COMMIT}")
+
+  TARGET_ARCH="linux/arm64"
+
+  IMAGE_TAGS=("--tag" "docker.io/${REPO}:${COMMIT}-${TARGET_ARCH}" "--tag" "quay.io/${REPO}:${COMMIT}-${TARGET_ARCH}")
```
yurishkuro, line 17: we do not include arch in the image tags. You can see in the production images that the same tag could represent multiple architectures, e.g.

image

yurishkuro:
https://github.com/jaegertracing/jaeger/actions/runs/8921061123/job/24500365309?pr=5356#step:10:197
```
docker buildx build -t localhost:5000/baseimg_alpine:latest --push \
	--build-arg root_image=alpine:3.16 \
	--build-arg cert_image=alpine:3.16 \
	--platform=linux/amd64,linux/s390x,linux/ppc64le,linux/arm64 \
	docker/base
```
still builds many platforms

scripts/build-crossdock.sh
Wise-Wizard: Yes, thanks for clarifying. Tried a different approach now, went through the QEMU documentation, and noticed we can specify platforms as an input to QEMU, if we want only specific platforms to be built? Is this approach right?

yurishkuro:
the approach was outlined in the ticket - to use the same mechanism that is used for not building all platforms for all other images when not on main branch https://github.com/jaegertracing/jaeger/blob/14aa43d4929ef344df722081aafd1a76c93c4f22/.github/workflows/ci-docker-build.yml#L53

Wise-Wizard:
HI, found the issue. The actual docker image being built is from another Makefile inside the docker folder.
Screenshot 2024-05-03 021523
So, we have to pass the platform as an argument from the build-crossodck.sh file itself, to only build the amd-64 architecture instead?

Wise-Wizard:
> Found the issue. The actual docker image being built is from another Makefile inside the docker folder. Screenshot 2024-05-03 021523 So, we have to pass the platform as an argument from the build-crossodck.sh file itself, to only build the amd-64 architecture instead?

On hardcoding, the platform variable as linux/amd-64, it is not building any other architectures!

.github/workflows/ci-crossdock.yml
```diff
@@ -45,12 +45,18 @@ jobs:
 
     - uses: docker/setup-qemu-action@68827325e0b33c7199eb31dd4e31fbe9023e06e3 # v3.0.0
 
-    - name: Build, test, and publish crossdock image
+    - name: Build crossdock image (main branch only)
+      if: github.ref_name == 'main'
```
yurishkuro, line 49: we don't need these ifs, unlike main binaries, crossdock always can be built on one architecture only

docker/Makefile
```diff
@@ -14,13 +14,13 @@ create-baseimg: prepare-docker-buildx
 	docker buildx build -t $(BASE_IMAGE) --push \
 		--build-arg root_image=$(ROOT_IMAGE) \
 		--build-arg cert_image=$(CERT_IMAGE) \
-		--platform=$(PLATFORMS) \
+		--platform=linux/amd64 \
```
yurishkuro, line 17: again, if you need to change the behavior, just override the PLATFORMS var, don't change the actual target code

.github/workflows/ci-crossdock.yml
Wise-Wizard: Yes, I will be removing that.
The main change of this commit was only the Makefile

docker/Makefile
Wise-Wizard: Yes, doing that right now.

yurishkuro:
The issue has been solve in #5410. Closing this PR, thanks for trying.

Wise-Wizard:
> The issue has been solve in #5410. Closing this PR, thanks for trying.

I had made the same change, and was going to push it just now. He built upon my work anyways.
No Worries :)

varshith257:
> > The issue has been solve in #5410. Closing this PR, thanks for trying.
> 
> I had made the same change, and was going to push it just now. He built upon my work anyways.
> No Worries :)

@Wise-Wizard I haven't taken your work. You can see my closed PR of my attempts to rectify the root cause and @yurishkuro also knows it. And this suggestion made by yurishkuro before here suggested for you to override variable. You can follow up in my closed PR :)

varshith257:
@yurishkuro PTAL . This is something going wrong in our organization. If someone working on issue please don't make others to work on them until closes/merge PR and simply waste of the efforts of others.  

I think it can better handled by assignment of issues to contributer to avoid this conflicts

Wise-Wizard:
> > > The issue has been solve in #5410. Closing this PR, thanks for trying.
> > 
> > 
> > I had made the same change, and was going to push it just now. He built upon my work anyways.
> > No Worries :)
> 
> @Wise-Wizard I haven't taken your work. You can see my closed PR of my attempts to rectify the root cause and @yurishkuro also knows it. And this suggestion made by yurishkuro to me before here suggested for you to override variable. You can follow up in my closed PR :)

No worries Varshith, but as you mentioned you weren't able to find the root cause. I found the root cause and was on the verge of completing it, so you should have just let me go through with it. Again. it's all in a good cause and the issue was resolved, it's all that matters :)

PR #5357, opened 2024-04-13
WIP Remove dependency on ES Refresh

yurishkuro:


yurishkuro:
Conclusion - all ES tests passed with this change

PR #5358, opened 2024-04-14
[storage integration tests] Remove Refresh

yurishkuro:
- Refresh was only used by Elasticsearch tests, and required too many low-level clients to do correctly.
- Meanwhile, since all the tests are using waitForCondition checks, forcing a refresh isn't necessary, the test will simply wait for ES background indexing to finish

## Description of the changes
- Remove Refresh as a concept from integration tests

- CI


jkowall (approved):
LGTM, thanks!

PR #5364, opened 2024-04-16
fix go.mod to fix CodeQL workflows

jkowall:
Correct the issue in codeQL

image

## Checklist
- [NA] I have added unit tests for the new functionality

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/jaegertracing/jaeger
 
-go 1.21
+go 1.21.0
 
 require (
 	github.com/HdrHistogram/hdrhistogram-go v1.1.2


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/jaegertracing/jaeger
 
-go 1.21
+go 1.21.0
```
yurishkuro, line 3: according to LLM specifying patch version in this directive is not recommended. The directive controls source code compatibility, which is not affected by the patch version.

On that topic, most of our CI workflows are using `go-version: 1.22.x`. It's fine for test-only workflows, but others are used to build final artifacts, and you could argue not fixing a precise patch version means the builds are not repeatable. That sounds like the issue the security scanner should be flagging, not the go.mod directive

yurishkuro: For reference, all official examples of `go` directive use major.minor version -- https://go.dev/ref/mod#go-mod-file-go

PR #5366, opened 2024-04-17
Add `CleanUp` method for badger storage

akagami-harsh:
- https://github.com/jaegertracing/jaeger/pull/5355#discussion_r1566052787

## Description of the changes
- added clean up method for badger 
- This cleanup method is used in the badger integration test for cleanup between tests 
- Use persistent badger instead of reinitializing ephemeral one between tests

- tested locally

## Checklist


plugin/storage/badger/factory.go
```diff
@@ -303,3 +303,9 @@ func (f *Factory) registerBadgerExpvarMetrics(metricsFactory metrics.Factory) {
 		}
 	})
 }
+
+func (f *Factory) CleanUp() error {
```
yurishkuro, line 307: I recommend the name Purge, as it needs to be explicit that all data will be lost ("clean-up" is much weaker).

Also, add a comment explaining that this function is only meant to be called from tests, it's not for production use.

plugin/storage/integration/badgerstore_test.go
```diff
@@ -60,9 +61,7 @@ func (s *BadgerIntegrationStorage) clear() error {
 }
 
 func (s *BadgerIntegrationStorage) cleanUp(t *testing.T) {
-	err := s.clear()
```
yurishkuro, line 63: clear() function should be removed?

yurishkuro: I'm asking if the function itself needs to be removed.

akagami-harsh: yes, we can also remove the `clear()` function and directly call `s.factory.Close()` in the `TestBadgerStorage` function

plugin/storage/integration/badgerstore_test.go
```diff
@@ -55,20 +56,14 @@ func (s *BadgerIntegrationStorage) initialize(t *testing.T) {
...
+	s.factory.Purge()
 }
 
 func TestBadgerStorage(t *testing.T) {
 	SkipUnlessEnv(t, "badger")
 	s := &BadgerIntegrationStorage{}
 	s.initialize(t)
 	s.RunAll(t)
-	defer s.clear()
+	defer s.factory.Close()
```
yurishkuro, line 68: this is the wrong way to use defer - if RunAll panics you will never reach this point. It can be called right after `initialize()`, but even better is to add it as a `t.CleanUp()` callback inside `initialize()`.

plugin/storage/integration/badgerstore_test.go
```diff
@@ -55,20 +56,14 @@ func (s *BadgerIntegrationStorage) initialize(t *testing.T) {
...
 func (s *BadgerIntegrationStorage) cleanUp(t *testing.T) {
-	err := s.clear()
-	require.NoError(t, err)
-	s.initialize(t)
+	s.factory.Purge()
 }
 
 func TestBadgerStorage(t *testing.T) {
 	SkipUnlessEnv(t, "badger")
 	s := &BadgerIntegrationStorage{}
```
yurishkuro, line 65: setting of properties in L50-L56 should be here, not inside `initialize()`

plugin/storage/integration/badgerstore_test.go
```diff
@@ -49,26 +53,22 @@ func (s *BadgerIntegrationStorage) initialize(t *testing.T) {
 	s.CleanUp = s.cleanUp
 
 	s.logger, _ = testutils.NewLogger()
```
yurishkuro, line 55: these two should also move to L64 - these fields are not changing once initialized. Also, this purpose of this function is to initialize the _factory_, not to initialize the _test suite_. So let's keep the code more tidy.

plugin/storage/integration/badgerstore_test.go
```diff
@@ -33,6 +33,10 @@ type BadgerIntegrationStorage struct {
 
 func (s *BadgerIntegrationStorage) initialize(t *testing.T) {
 	s.factory = badger.NewFactory()
+	t.Cleanup(func() {
+		s.factory.Close()
+	})
```
yurishkuro, line 38: Are you sure this needs to be right after NewFactory()? Usually the constructors NewFactory don't do any processing, and the actual work happens in `factory.Initialize`, so if `factory.Initialize` returns no error only then do you need to schedule a call to Close.

akagami-harsh: yes, you are right. moved it after `factory.Initialize`.

yurishkuro (approved):

PR #5377, opened 2024-04-21
chore: use errors.New to replace fmt.Errorf with no parameters will much better

ChengenH:
use errors.New to replace fmt.Errorf with no parameters will much better


ChengenH:
> Isn’t there any linter like errorlint that could ensure such best practice ?

It's not serious, so...

yurishkuro (approved):

yurishkuro (commented):
If you're interested in this clean-up, there are many other places with similar usage:
```
./model/ids.go:90:		return TraceID{}, fmt.Errorf("invalid length for TraceID")
./model/ids.go:97:	return nil, fmt.Errorf("unsupported method TraceID.MarshalText; please use github.com/gogo/protobuf/jsonpb for marshalling")
./model/ids.go:102:	return fmt.Errorf("unsupported method TraceID.UnmarshalText; please use github.com/gogo/protobuf/jsonpb for marshalling")
./model/ids.go:127:		return 0, fmt.Errorf("buffer is too short")
./model/ids.go:184:		return SpanID(0), fmt.Errorf("invalid length for SpanID")
./model/ids.go:191:	return nil, fmt.Errorf("unsupported method SpanID.MarshalText; please use github.com/gogo/protobuf/jsonpb for marshalling")
./model/ids.go:196:	return fmt.Errorf("unsupported method SpanID.UnmarshalText; please use github.com/gogo/protobuf/jsonpb for marshalling")
./pkg/config/tlscfg/options.go:78:			return nil, fmt.Errorf("minimum tls version can't be greater than maximum tls version")
./pkg/config/tlscfg/options.go:108:		return nil, fmt.Errorf("for client auth via TLS, either both client certificate and key must be supplied, or neither")
./pkg/config/tlscfg/options_test.go:162:					return nil, fmt.Errorf("fake system pool")
./pkg/clientcfg/clientcfghttp/thrift-0.9.2/ttypes.go:59:	return SamplingStrategyType(0), fmt.Errorf("not a valid SamplingStrategyType string")
./pkg/es/errors_test.go:15:	require.ErrorContains(t, fmt.Errorf("some err"), "some err", "no panic")
./pkg/es/config/config.go:366:		return nil, fmt.Errorf("both Password and PasswordFilePath are set")
```

yurishkuro:
alternatively, I agree with @mmorel-35 , we should find a linter that catches this, rather than doing manually. I thought revive linter would do that (#5505), but it doesn't.

yurishkuro:
no follow-up, closing

